PR #28935, opened 2023-02-14
Stats: remove unnecessary params passed to wpcom

kangzj:
## Proposed changes:
The PR add a general function to filter and build the query pass to WPCOM. The reason we are doing this is that, `rest_route` sometime breaks the WPCOM API. An example is `/jetpack/v4/stats-app/sites/{site-id}/site-has-never-published-post`.

The PR also takes the chance to add a method `get_wp_error` to unify remote error code and message.

### Other information:

- [x] Have you written new tests for your changes, if applicable?
- [x] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion
n/a

## Does this pull request change what data or activity we track or use?
no

## Testing instructions:
* Use branch of https://github.com/Automattic/wp-calypso/pull/73301 to build Odyssey Stats
* Choose `Plain` for Permanent Links `/wp-admin/options-permalink.php`
* Open `/wp-admin/admin.php?page=stats`
* Inspect network request
* Ensure no 404s for API requests. Requests should be like `https://jetpack.xxxx.com/index.php?rest_route=/jetpack/v4/stats-app/sites/{site-id}/site-has-never-published-post&include-pages=true`
* Ensure all tests pass

PR #30103, opened 2023-04-14
Blaze: add new dashboard menu

jeherve:
## Proposed changes:

This registers a new "Tools > Advertising" menu, appearing when your site is eligible for Blaze.

The contents of the menu will be loaded remotely from Calypso.

This borrows from the Stats Admin package, where we've implemented something similar (the Stats page is loaded from Calypso).

**To do**

- [x] Add tests
- [x] Figure out if the menu location should change in different environments (e.g. sites that do not use the Jetpack plugin, but may require Blaze on its own). peeHDf-Zg-p2
- [x] Maybe add gating so the menu is not publicly accessible until this is ready for prime-time.
- [x] Figure out if this menu should be hidden or handled differently on WordPress.com sites or WoA sites, where we already have the Tools > Advertising menu. peeHDf-Zg-p2

### Other information:

## Jetpack product discussion

* https://github.com/Automattic/wp-calypso/pull/76431
* pedfti-1M1-p2
* Related: pbIy4N-3vY-p2
* p1681412672232459-slack-C053FAERCDP
* peeHDf-Zg-p2
* D112493-code

* No

## Testing instructions:

1. Ensure that Blaze links still appear under the published posts in Posts > All Posts, and that the links work.
2. Ensure that you're still offered to Blaze a post in the post-publish panel in the editor (after publishing a post), and that the Blaze link properly redirects you to the Advertising dashboard in Calypso.
3. Add `add_filter( 'jetpack_blaze_dashboard_enable', '__return_true' );` to a functionality plugin on your site.
4. You should see a new Tools > Advertising menu when your site is connected to WordPress.com.

*******************
To populate the contents of the dashboard, you have 2 options:

* You can patch D112493-code in your WordPress.com sandbox.
* You can build your own copy of the dashboard:
    * Build Blaze and Jetpack: `jetpack build packages/blaze && jetpack build plugins/jetpack`
    * Clone https://github.com/Automattic/wp-calypso
    * Install deps with `yarn`
    * Go to `apps/blaze-dashboard`
    * Launch the app with `BLAZE_DASHBOARD_PACKAGE_PATH=/.../jetpack/projects/packages/blaze yarn dev`, while entering the correct path to your package on your machine.
*******************

5. Visit the Tools > Advertising menu on your local installation.
6. Now re-test items 1 and 2 of this list; they should still work, but the links should now point to the wp-admin dashboard.


projects/packages/blaze/src/class-dashboard-config-data.php
```diff
@@ -0,0 +1,147 @@
...
+						"$blog_id" => array(
+							'ID'           => $blog_id,
+							'URL'          => site_url(),
+							'jetpack'      => true,
+							'visible'      => true,
+							'capabilities' => $empty_object,
+							'products'     => array(),
+							'plan'         => $empty_object, // we need this empty object, otherwise the front end would crash on insight page.
+							'options'      => array(
+								'wordads'    => ( new Modules() )->is_active( 'wordads' ),
```
jeherve: I'm not sure I understand why we need the connection to the WordAds module here and downer. Shouldn't Blaze work even when the module is disabled?

projects/packages/blaze/src/class-dashboard-config-data.php
```diff
@@ -0,0 +1,147 @@
...
+	/**
+	 * Get the features of the current plan.
+	 */
+	protected function get_plan_features() {
+		$plan = Current_Plan::get();
+		if ( empty( $plan['features'] ) ) {
+			return array();
+		}
+		return $plan['features'];
+	}
```
jeherve, line 131: Down the road (not in this PR), I think it may be nice to extract those instead of copying them from the Stats Admin class.

projects/packages/blaze/src/class-dashboard-rest-controller.php
```diff
@@ -0,0 +1,270 @@
...
+		);
+
+		register_rest_route(
+			static::$namespace,
+			sprintf( '/sites/%d/wordads/dsp/api/(?P<sub_path>.+)', $site_id ),
+			array(
+				'methods'             => WP_REST_Server::EDITABLE,
+				'callback'            => array( $this, 'edit_wordads_dsp_generic' ),
+				'permission_callback' => array( $this, 'can_user_view_wordads_dsp_callback' ),
+			)
```
jeherve: Maybe we should add some checks for the `sub_path`, so folks do not input malicious things or paths that will not lead to a proper response when making the WordPress.com request?

projects/packages/blaze/src/class-dashboard-rest-controller.php
```diff
@@ -0,0 +1,270 @@
...
+			array( 'method' => 'GET' )
+		);
+	}
+
+	/**
+	 * Only administrators or users with capability `activate_wordads` can access the API.
+	 */
+	public function can_user_view_wordads_dsp_callback() {
+		// phpcs:ignore WordPress.WP.Capabilities.Unknown
+		if ( current_user_can( 'manage_options' ) || current_user_can( 'activate_wordads' ) ) {
```
jeherve: Same here, I'm not sure I understand the WordAds link.

projects/packages/blaze/src/class-dashboard.php
```diff
@@ -0,0 +1,179 @@
...
+					style="position: absolute; left: 50%; top: 50%;"
+					alt=<?php echo esc_attr( __( 'Loading', 'jetpack-blaze' ) ); ?>
+					src="//en.wordpress.com/i/loading/loading-64.gif"
+				/>
+			</div>
+		</div>
+		<script>
+			jQuery(document).ready(function($) {
+				// Load SVG sprite.
+				$.get("https://widgets.wp.com/blaze-dashboard/common/gridicons-506499ddac13811fee8e.svg", function(data) {
```
jeherve, line 84: This one doesn't load for me yet, I get a 404.

projects/packages/blaze/src/class-blaze.php
```diff
@@ -57,6 +61,27 @@ public static function add_post_links_actions() {
 		}
 	}
 
+	/**
+	 * Enable the Blaze menu.
+	 *
+	 * @return void
+	 */
+	public static function enable_blaze_menu() {
+		if ( self::should_initialize() ) {
```
jeherve: Maybe we should add a filter to gate the feature until it's ready?

jeherve (commented):
This is looking good! 🌟 I pushed a few commits and left a few notes, let me know what you think!

projects/packages/blaze/src/class-blaze.php
sbarbosa: This is a good idea! I don't know if we can add FF or something here to close the functionality until is ready

projects/packages/blaze/src/class-dashboard.php
sbarbosa: Yes, this file will be included on the widgets CDN. I am already working on pushing the first version of the app there. 
I will let you know when I have the phabricator patch so you can test the integration like in production.

projects/packages/blaze/src/class-dashboard-config-data.php
sbarbosa: Let me check this point; I don't think we need wordads enabled for Blaze. Some of the end points in the DSP have the wordads prefix, but those are under the DSP, they probably don't have anything to do with that module.

projects/packages/blaze/src/class-dashboard-rest-controller.php
sbarbosa: I wanted to have this open because we are making some changes on the DSP API and I didn't want to have Jetpack too wired to the endpoints names. 

My line of thought was that the API is exposed to the public, so users can make the same request they do here from a browser. 

But, if you want, I can add a whitelist of paths that we are allowed to call (maybe high-order paths like /campaigns, /posts, etc.) and leave the sub-path free to use (like /campaings/stats, etc.).

What do you think? If not I can go for an approach similar to the stats-admin and have each endpoint described and not use wildcards here.

jeherve: Thanks for the extra details! Since you're still iterating I can see how limiting ourselves would be problematic. I like the idea of a general whitelist though, that seems like a good middle-ground. 👍

projects/packages/blaze/src/class-blaze.php
jeherve: 👍 I've just pushed 4f2d68ed7d10391f1529dfff4917428cb93ae7d3 to gate access to the feature for now.

sbarbosa: We may need to roll back the entry points I updated or add the same gate so you don't get redirected to the module if it is not opened.

I will try to make that change, it should be easy in PHP (I can use the same code as yours), but I don't know if we can use the same code in JS. I will try and see if I find any examples.

jeherve: I've just pushed a few commits to make that happen. Let me know what you think about it.

Ideally I would have preferred to build the full management link in one place instead of 2 (we now build it in php and in js), but the dynamic post ID would make this less clean in my opinion.

projects/packages/blaze/src/class-blaze.php
```diff
@@ -161,6 +206,47 @@ public static function should_initialize() {
...
+	/**
+	 * Get URL to create a Blaze campaign for a specific post.
+	 *
+	 * This can return 2 different types of URL:
+	 * - Calypso Links
+	 * - wp-admin Links if access to the wp-admin Blaze Dashboard is enabled.
+	 *
+	 * @param int $post_id Post ID.
+	 *
+	 * @return array An array with the link, and whether this is a Calypso or a wp-admin link.
```
jeherve, line 218: I'm not a huge fan of returning an array here just to get the info on whether we should link outside of wp-admin or not, I would rather return just the URL. That said, this seemed cleaner than having to check the string returned by this method later on just to find out if it was an external link or not.

projects/packages/blaze/src/js/editor.js
```diff
@@ -99,12 +105,15 @@ const BlazePostPublishPanel = () => {
 				onClick={ trackClick }
 				onKeyDown={ trackClick }
 			>
-				<Button variant="secondary" href={ blazeUrl } target="_top">
+				<Button variant="secondary" href={ blazeUrl( postId ) } target="_top">
```
sbarbosa: You don't need to pass the postId because the `blazeUrl ` function doesn't expect it.

jeherve: Good point. This was a leftover :) Removed in 3fa209950f4c428007c1e7b363d3936e8e499a2e

projects/packages/blaze/src/class-dashboard-config-data.php
jeherve: Fixed in a20893ce4aba6dcee51413949e06a43c86d6d14e

jjolmo:
Code LGTM. Awesome work there

jjolmo (dismissed):

projects/packages/blaze/src/class-dashboard-rest-controller.php
sergeymitr: I can see how creating a list of allowed sub-paths can be limiting, but I think we should try our best to follow that route. If that would limit us too much, then the partial-listing sounds good. Not allowing user-provided `/` in the URL would be a big win.

We could also add some basic validation there via some generic regex to make sure it matches the format and only contains symbols we would use in the endpoint path.
The parameter description in `args` would come handy too.

sergeymitr (dismissed):
Left a comment about the endpoint parameters: https://github.com/Automattic/jetpack/pull/30103#discussion_r1232335197

I'd also appreciate a bit more details on how to enable Blaze, as I wasn't able to activate it on my local and WoA sites.

sbarbosa:
Sure @sergeymitr! 
To test this locally, you first need to enable this filter: `add_filter( 'jetpack_blaze_dashboard_enable', '__return_true' );`
I do that by installing the plugin `Code Snippets` and then creating a snippet with that filter. I think there are easier ways to do this, but this is the only one I know.

After that, you need to sandbox `widgets.wp.com` and apply this patch D112493-code. That patch is the one that has the Blaze Dashboard SPA.

With that, you should be able to see the Tools->Advertising option in the menu, and the Dashboard if you enter there. 
We are still working in the last bugs in the dashboard, that's what is still a patch, so you may find some small bugs while testing the page. 

Let me know if I can help you with something more. I will try and restrict the rest endpoints a little more.

sbarbosa:
Ok @jeherve, @sergeymitr. I changed the rest controller to be more restrictive in the paths.
The changes were:
* Added a validation in the path to only allow letters, numbers, and symbols (`-_/`), that are what the DSP is expecting.
* Changed to register only the base roots that we are using.

Let me know what you think of this change. I can do it differently if you want.

projects/packages/blaze/src/class-dashboard-config-data.php
```diff
@@ -0,0 +1,145 @@
...
+	/**
+	 * Return the config for the app.
+	 */
+	public function get_data() {
+		$blog_id      = Jetpack_Options::get_option( 'id' );
+		$empty_object = json_decode( '{}' );
+		return array(
+			'admin_page_base'          => $this->get_admin_path(),
+			'api_root'                 => esc_url_raw( rest_url() ),
+			'blog_id'                  => Jetpack_Options::get_option( 'id' ),
```
timur987: It seems we can use `$blog_id` there.

timur987 (commented):
Thanks, @jeherve!

I tested according to testing instructions (the approach of local build and using Jurassic Tube).

It works fine for me (including opportunity to start blazing from All Posts page (JITM and action under post title within post lists table), also Tools > Advertising section works as expected.

What I've found is that a user can scroll down and partially see the side menu:

https://github.com/Automattic/jetpack/assets/115007291/839be0a3-96ed-4c1c-a8a7-90e14534fd9f

Other than that it looks awesome!

sbarbosa:
Thanks @timur987! I will create a ticket for the scroll bug in the dashboard. It's probably a bug in the dashboard when we load the Create campaign widget.

sergeymitr:
Hi @sbarbosa, thanks for clarifying the instructions.
I'm able to see the "Tools -> Advertising" menus both locally and on WPCOM.
However, I get a bunch of 404 API responses in my local Jetpack:
Screen Shot 2023-06-20 at 11 07 58
Is there anything else I need to do to activate Blaze on my site?

sbarbosa:
Wow, that is strange. Did you have the same result in other browsers?
From the screenshots, the endpoint is not correct. The rest route is, but I don't know what that `index.php?rest_route=` is. It should be made to `domain/wp-json/xxx`

@jeherve, do you have any idea what could be causing that problem?

jeherve:
> From the screenshots, the endpoint is not correct. The rest route is, but I don't know what that `index.php?rest_route=` is. It should be made to `domain/wp-json/xxx`

This is expected when [pretty permalinks](https://wordpress.org/documentation/article/customize-permalinks/) aren't enabled on the site.

The endpoints should still work then though. I'm not sure yet why the endpoints aren't working in this scenario.

sergeymitr:
The endpoint makes its way to `public-api.wordpress.com` [here](https://github.com/Automattic/jetpack/blob/c8f8b49223febed6392f3c2996e2adc9c5260cab/projects/packages/connection/src/class-client.php#L264). The 404 response (route not found) comes straight from WPCOM.

The WPCOM request URL looks like this:
```
https://public-api.wordpress.com/wpcom/v2/sites/[redacted_blog_id]/wordads/dsp/api/v1/credits/balance?rest_route=%2Fjetpack%2Fv4%2Fblaze-app%2Fsites%2F[redacted_blog_id]%2Fwordads%2Fdsp%2Fapi%2Fv1%2Fcredits%2Fbalance&home=https%3A%2F%2F[redacted_url]&siteurl=https%3A%2F%2F[redacted_url]&idc=1
```

I'm sandboxed.
Is there anything other than D112493 I need to checkout in my sandbox?

jeherve:
> Is there anything other than D112493 I need to checkout in my sandbox?

No, this isn't you, it's happening to anyone with plain permalinks.

I wonder if we may need to do something like what was done in #28935 for stats.

sbarbosa:
Thanks @jeherve. That probably is the cause for this. I will apply a similar filter here and see if that fixes the problem.

sbarbosa:
Ok @jeherve, @sergeymitr. I filtered the wrong parameters that we were sending to WordPress.com API. 

I used the same approach as the commit mentioned by Jeremy; the only problem was that it only worked when the route didn't expect any query parameters. 
For the rest of the routes, I had to change the initial regex to include the possibility of those query parameters (without this, I was getting 404 from Jetpack for requests like `index.php?rest_route=/jetpack/v4/blaze-app/sites/[site_id]/blaze/posts?page=1&order_by=modified&order=desc` ).

Let me know what you think and if there is a better way to consider the query params in the `register_rest_route` when the site is using plain permalinks.

projects/packages/blaze/src/class-blaze.php
```diff
@@ -34,7 +34,8 @@ class Blaze {
 	public static $script_path = '../build/editor.js';
 
 	/**
-	 * The configuration method that is called from the jetpack-config package.
+	 * Initializer.
+	 * Used to configure the stats package, eg when called via the Config package.
```
samiff:
```suggestion
	 * Used to configure the blaze package, eg when called via the Config package.
```

samiff (commented):
This tested well for me on my local dev site, even trying with plain permalinks. Two quick questions.

1. Are Woo products intended to show the Blaze option?
2. Where are we tracking issues with the Advertising page contents? For example, one of the buttons seems off on my end:
Screenshot 2023-06-21 at 2 59 38 PM

jeherve:
> 1. Are Woo products intended to show the Blaze option?

Yes. Blaze supports posts, pages, and products.

> Where are we tracking issues with the Advertising page contents? For example, one of the buttons seems off on my end:

I think posting in #blaze-dashboard-redesign about it is fine, but since you mentioned it here @sbarbosa will know what to do with it :)

sbarbosa:
Thanks for the test @samiff. You can have an updated version of the dashboard in this Calypso PR:
https://github.com/Automattic/wp-calypso/pull/78209

You can follow the testing steps in that PR to see that page version. I adjusted some styles and changed the header.

I will update the Phabricator diff when I merge that PR.

sergeymitr:
I noticed a couple of potential issues with the `/wpcom/v2/sites/%d/blaze/posts` endpoint.

I have permalinks set to "plain", meaning the page URL's looks like `/?p=123`.
Calypso page list takes that into account, using the `/?page_id=123` links.
However, when I go to `https://wordpress.com/advertising/[site-url]`, it tries to use post-name permalinks (e.g. `/cart/`), which lead to 404.
So for some reason the endpoint doesn't take permalink settings into account.

Another problem appeared on a new JN site where the same page started displaying "About" page which never existed on that website (clean JN site with only Jetpack installed).
Then I installed WooCommerce, which created a bunch of pages, but they never appeared in Calypso's "Tools -> Advertising", only displaying that non-existing "About" page.

The `blaze/posts` endpoint does not see them, but sees some other page that never existed (ID: 1):
Screen Shot 2023-06-22 at 10 10 32

Pages in the database.
Post ID: 1 is the default "Hello World" that gets created automatically. Never been edited.
Screen Shot 2023-06-22 at 10 05 28

They show up in "Calypso -> Posts", so they did get synced and WPCOM knows about them:
Screen Shot 2023-06-22 at 10 12 09

jeherve:
> I noticed a couple of potential issues with the `/wpcom/v2/sites/%d/blaze/posts` endpoint.

These may be sync issues since the endpoint does a direct `WP_Query` on WordPress.com (see the `WPCOM_REST_API_V2_Endpoint_Blaze` class on WordPress.com). Do you think you could take a look at that?

Since those issues are not directly related to this PR, but to the endpoint it is using, and since this PR doesn't surface the UI just yet (it's still hidden behind a filter), do you think we could move forward with merging the PR?

samiff (approved):
> since this PR doesn't surface the UI just yet (it's still hidden behind a filter), do you think we could move forward with merging the PR?

I think that's a good plan of action since we've also got some other dependent PRs.

PR #30105, opened 2023-04-14
Social: Pull dynamic values from DB in pricing table

gmjuhasz:
This change replaces the hardcoded values on the Social admin page, and pull the current prices based on the users locale.

## Proposed changes:

* Added new endpoint to fetch the product info
* Created a hook that calls the endpoint
* Used the hook to get the data. Spinners are present while loading

### Other information:

## Jetpack product discussion

1204272127368434-as-1204008786549223

No.

## Testing instructions:

* Go to the Social admin page
* In `projects/plugins/social/src/js/components/admin-page/index.jsx` change the logic so the `PricingPage` is shown, or create a new site.
* Test that the prices are pulled in correctly. Should be the same as on https://jetpack.com/social/


projects/packages/publicize/src/class-rest-controller.php
```diff
@@ -70,6 +78,17 @@ public function register_rest_routes() {
 			)
 		);
 
+		// Get currently promoted product from the product's endpoint.
+		register_rest_route(
+			'jetpack/v4',
+			'/social-promoted-product-info',
```
danielpost: Doesn't this get all the prices, not just promoted prices?

projects/packages/publicize/src/class-rest-controller.php
```diff
@@ -156,6 +175,34 @@ public function get_publicize_shares_count() {
...
+		$request_url   = 'https://public-api.wordpress.com/rest/v1.1/products?locale=' . get_user_locale() . '&type=jetpack';
+		$wpcom_request = wp_remote_get( esc_url_raw( $request_url ) );
+		$response_code = wp_remote_retrieve_response_code( $wpcom_request );
+		if ( 200 === $response_code ) {
+			$products = json_decode( wp_remote_retrieve_body( $wpcom_request ) );
+			return array(
+				'advanced' => $products->{self::JETPACK_SOCIAL_ADVANCED_YEARLY},
+				'basic'    => $products->{self::JETPACK_SOCIAL_BASIC_YEARLY},
+			);
+		} else {
```
danielpost: This `else` is unnecessary because there's a `return` in the `if` statement already. Personally, I also prefer flipping the logic and returning the `WP_Error` early if the response code isn't `200` but that's more of a personal preference.

What happens if one of the products is missing (e.g. `$products->{self::JETPACK_SOCIAL_ADVANCED_YEARLY}`)? Does the page get a fatal error or does it just give an infinite spinner?

It might be nice to add some kind of fallback for that, but if it just keeps spinning (no error) I think that's okay.

projects/plugins/social/src/js/hooks/use-product-info/test/index.test.js
```diff
@@ -0,0 +1,44 @@
...
+					introOffer: null,
+				},
+				advanced: {
+					price: 20,
+					introOffer: 60,
+				},
+			} );
+		} );
+	} );
+} );
```
danielpost, line 44: Nice, I appreciate these tests!

projects/plugins/social/src/js/hooks/use-product-info/index.js
```diff
@@ -0,0 +1,46 @@
...
+			} );
+			setProductInfo( parsePromotedProductInfo( socialPromotedProductInfo ) );
+		} catch {
+			setProductInfo( null );
+		}
+	}, [] );
+
+	useEffect( () => {
+		getAsyncInfo();
+	}, [ getAsyncInfo ] );
```
danielpost, line 43: Did you create `getAsyncInfo` to be able to use async/await inside the `useEffect`? If so, you could also define the whole function inside the `useEffect` callback and call it right after defining it, or wrap it in an IIFE.

Something like: 

```js
useEffect(() => {
	(async () => {
		try {
			const socialPromotedProductInfo = await apiFetch( {
				path: '/jetpack/v4/social-promoted-product-info',
			} );
			setProductInfo( parsePromotedProductInfo( socialPromotedProductInfo ) );
		} catch {
			setProductInfo( null );
		}
	})();
}, []);
```

But again, it's mostly personal preference 😅

danielpost (dismissed):
Looks good and tests well! I've added a few thoughts but nothing blocking, feel free to implement or ignore as you see fit.

projects/plugins/social/src/js/hooks/use-product-info/index.js
gmjuhasz: Yeah, I just like the way it looks, to me it looks cleaner, I was never a fan of IIFEs, but yeah as you said it's a matter of preference 😄 I can change it if you want.

projects/packages/publicize/src/class-rest-controller.php
gmjuhasz: > This else is unnecessary

Yeah nice catch, to be honest I grabbed this from Backup. I should have noticed this!

> What happens if one of the products is missing

Right now it's infinite spinner, I am not sure what to do then, we could show a notice or something, but that could be another change with design involved, what do you think? I [changed](https://github.com/Automattic/jetpack/pull/30105/commits/e7c5a723a1925d12235b3948ec9ce93c9f52a781) now, that if only one is missing, we show the spinner infinitely only for that one.

projects/plugins/social/src/js/hooks/use-product-info/index.js
danielpost: No all good, feel free to leave it! Ideally we should have an ESLint rule that enforces one way or another to improve consistency.

danielpost (approved):
Looks good to me!

PR #30107, opened 2023-04-14
CRM: Resolves 2840 - tag existing contacts with WooSync

tbradsha:
Resolves Automattic/zero-bs-crm#2840 - tag existing contacts with WooSync

## Proposed changes:

TL;DR: we captured order tags, but tags weren't updated on existing contacts, because we used `addUpdateContact()` with `do_not_update_blanks`, which converts things to `limitedFields` without blanks or arrays. The `tags` param is an array, so is removed. Confused? It's messy.

This PR calls `$zbs->DAL->contacts->addUpdateContactTags()` directly to add any new tags.

Notes:
* This does NOT affect invoices or transactions, which have different logic in the DAL.
* If one deletes a tag, then saves/resyncs a WooCommerce order, the tag will be added back. I don't see this as a problem.

### Other information:

## Testing instructions:

1. Create an order with a product.
2. Save.
3. Verify the contact's profile was tagged with the product.
4. Add another product to the order (or create a new order for the same user).
5. Save.

In `trunk`, the contact is tagged with the original product but not subsequent ones.

In `fix/crm/2840-woosync_tag_existing_contacts`, the contact is tagged with the additional products as well.

### DIFF
--- a/projects/plugins/crm/changelog/fix-crm-2840-woosync_tag_existing_contacts
+++ b/projects/plugins/crm/changelog/fix-crm-2840-woosync_tag_existing_contacts
@@ -1,0 +1,4 @@
+Significance: minor
+Type: fixed
+
+WooSync: tag existing contacts with new orders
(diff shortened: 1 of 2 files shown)


cleacos (dismissed):
I tested it before. It works as expected.

LGTM!

cleacos (approved):
Re-approved!! LGTM!

PR #30110, opened 2023-04-14
CRM: Resolves 3059 - PHP notice when sending invoice to contact

tbradsha:
Resolves Automattic/zero-bs-crm#3059 - PHP notice when sending invoice to contact

## Proposed changes:

We didn't account for the "notification" template when generating invoice HTML, so it threw a PHP notice.

This PR adds a catch for that notification, thus enabling the `##INVOICE-HTML-STATUS##` placeholder in email context. As a bonus, I also added the status placeholder to the default email template.

### Other information:

## Testing instructions:

1. Enable full debugging.
2. Create an invoice and assign it to a contact.
3. Send the invoice to the contact via email (with or without PDF).

In `trunk`, the email will send, but will throw a PHP notice and will not show the invoice status.

In `fix/crm/3059_php-notice_when_sending_invoice`, the email will not generate a PHP notice and it will show the invoice status.

Note: simply switching between branches locally in this instance wouldn't be enough, as the invoice email template HTML is stored in the database on site init, loaded from `html/templates/invoicesent.html`. If you want to use the new HTML on an existing site, you can instead copy the HTML from that page and paste it here: `/wp-admin/admin.php?page=zbs-email-templates&zbs_template_id=3`

### DIFF
--- a/projects/plugins/crm/changelog/fix-crm-3059_php-notice_when_sending_invoice
+++ b/projects/plugins/crm/changelog/fix-crm-3059_php-notice_when_sending_invoice
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Invoices: PHP notice when sending contact an invoice via email
--- a/projects/plugins/crm/html/templates/invoicesent.html
+++ b/projects/plugins/crm/html/templates/invoicesent.html
@@ -1,57 +1,61 @@
 <div class="container" style="max-width:1200px;margin-left:auto;margin-right:auto">
-  <div id="zbs_top_wrapper">
-      <div id="zbs_invoice_logos" style="width:50%;float:left">
-          <div class="wh-logo-set ##LOGO-CLASS##">
-              <img id="wh-logo-set-img" src="##LOGO-URL##" style="max-width:50%" />
-          </div>
-      </div>
-      <div class="zbs-invoice-topper" style="width: 50%;float: right;text-align: right">
-          <table class="form-table" style="text-align: right;float: right">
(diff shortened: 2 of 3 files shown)


coder-karen:
Can confirm the PHP error is gone when testing with this PR vs trunk.

> Send the invoice to the contact via email (with or without PDF).

I might not be testing this right though, as both emails don't have a status at the top. Is this referring to the 'Email Invoice' button at the top of the 'Edit Invoice' page?

tbradsha:
> I might not be testing this right though, as both emails don't have a status at the top. Is this referring to the 'Email Invoice' button at the top of the 'Edit Invoice' page?

To confirm, did you test on a JN/new site (I updated the instructions with that now)? The email should look like this on the fix branch (note the status field in the top right):

image

I should have specified that simply switching between branches locally in this instance wouldn't be enough, as the invoice email template HTML is stored in the database on site init, loaded from `html/templates/invoicesent.html`. If you want to use the new HTML on an existing site, you can instead copy the HTML from that page and paste it here: `/wp-admin/admin.php?page=zbs-email-templates&zbs_template_id=3`

coder-karen (dismissed):
Gotcha, thanks for the explanation! I was testing on a Pressable test site using the Beta plugin so same issue as on a JN site - needed to copy that info over. Works now! :shipit:

tbradsha:
@coder-karen I did a rebase per the checklist (it showed a fail on that one point), and then doing a force-push dismissed your review. Can you re-approve? :^)

coder-karen (approved):

PR #29482, opened 2023-03-15
CRM: Checking segment index exists to prevent export issues

coder-karen:
## Proposed changes:

* In some instances, exports were not working. PHP warnings and errors were displaying after clicking the export button, and the export did not take place. This was the case exporting quotes, transactions, contacts and invoices.
* It looks like the issues occur when wp_debug is set to true on a site (`define( 'WP_DEBUG', 'true' ); ` in wp-config) as well as the debug display not explicitly disabled (so `define( 'WP_DEBUG_DISPLAY', false );` not added to the wp-config file) , however as well as printing out the errors is still prevents the export from taking place.
* This PR adds a check to see if the segment index exists before checking to see if it is an array, [on this line](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Export.php#L254). This prevents the error messages and allows the exports to go ahead on the sites where the issue existed.

The warnings / errors:

```
Notice: Undefined index: segment in /srv/users/usered83e0e7/apps/usered83e0e7/public/wp-content/plugins/zero-bs-crm/includes/ZeroBSCRM.DAL3.Export.php on line 230

Warning: Cannot modify header information - headers already sent by (output started at /srv/users/usered83e0e7/apps/usered83e0e7/public/wp-content/plugins/zero-bs-crm/includes/ZeroBSCRM.DAL3.Export.php:230) in /srv/users/usered83e0e7/apps/usered83e0e7/public/wp-content/plugins/zero-bs-crm/includes/ZeroBSCRM.DAL3.Export.php on line 244

Warning: Cannot modify header information - headers already sent by (output started at /srv/users/usered83e0e7/apps/usered83e0e7/public/wp-content/plugins/zero-bs-crm/includes/ZeroBSCRM.DAL3.Export.php:230) in /srv/users/usered83e0e7/apps/usered83e0e7/public/wp-content/plugins/zero-bs-crm/includes/ZeroBSCRM.DAL3.Export.php on line 245
```

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

https://github.com/Automattic/zero-bs-crm/issues/2919

## Does this pull request change what data or activity we track or use?

No.

## Testing instructions:

[...]

PR #30111, opened 2023-04-14
CRM: Resolves 3031 - PHP notice when exporting subset of objects

tbradsha:
Resolves Automattic/zero-bs-crm#3031 - PHP notice when exporting subset of objects

## Proposed changes:

Segment export was introduced in Automattic/zero-bs-crm#2735. Similar to #29482, there were some conditions that weren't accounted for. This PR catches a PHP notice I ran across (line 300), and also mitigates a potential one (line 226). 

### Other information:

## Testing instructions:

1. Have at least one transaction and enable full debugging.
2. Go to the transaction list: `/wp-admin/admin.php?page=manage-transactions`
3. Select one or more transactions and choose "Export" from the bulk actions dropdown.
4. On the export page, proceed with the export.

On `trunk`, one will get a PHP notice.

On `fix/crm/3031-catch_PHP_error_on_export`, the PHP notice is no more.

### DIFF
--- a/projects/plugins/crm/changelog/fix-crm-3031-catch_PHP_error_on_export
+++ b/projects/plugins/crm/changelog/fix-crm-3031-catch_PHP_error_on_export
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Export: catch PHP notice when exporting a subset of objects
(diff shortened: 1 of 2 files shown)


coder-karen (approved):
Works well, no PHP errors when exporting single or multiple transactions. :shipit:

PR #30114, opened 2023-04-15
CRM: Cleanup after deleting custom fields

gogdzl:
## Proposed changes:
* This PR eliminates orphaned references to custom fields within field sorting settings when removing custom fields.

### Other information:

## Jetpack product discussion
https://github.com/Automattic/zero-bs-crm/issues/3057

No.

## Testing instructions:

1. In settings, create a new address custom field. (`wp-admin/admin.php?page=zerobscrm-plugin-settings&tab=customfields`)
2. In settings, go to field sorts and click save. (`/wp-admin/admin.php?page=zerobscrm-plugin-settings&tab=fieldsorts`).
3. Go back to the previous settings page and delete the newly created custom field.
4. Go back to the field sorts page, you should already see a php warning.
image
image
5. Go to the edit page for any contact. (e.g. `/wp-admin/admin.php?page=zbs-add-edit&action=edit&zbstype=contact&zbsid=1`)
6. Click Update Contact.
7. You should see more warnings.
image
8. Do the same for fields where `Hidden` was checked in `/wp-admin/admin.php?page=zerobscrm-plugin-settings&tab=fieldsorts`

In `fix/crm-3057-cleanup-after-deleting-custom-fields` none of this happen.


projects/plugins/crm/admin/settings/custom-fields.page.php
```diff
@@ -293,6 +293,89 @@
...
+				unset( $fields[ $sort_type ] );
+				continue;
+			}
+
+			$fields[ $sort_type ] = array_values(
+				array_filter(
+					$sort_names,
+					function ( $field_name ) use ( $custom_field_names, $custom_type ) {
+						return isset( $custom_field_names[ $custom_type ][ $field_name ] );
+					}
```
coder-karen: I've been unable to reproduce, but I was initially seeing this error in my error log after saving a custom address field:
``` PHP Fatal error:  Uncaught TypeError: array_filter(): Argument #1 ($array) must be of type array, string given in /usr/local/src/jetpack-monorepo/projects/plugins/crm/admin/settings/custom-fields.page.php:368```.

That said, on doing a trunk merge I see PHP tests are now failing with the same error: https://github.com/Automattic/jetpack/actions/runs/4784812679/jobs/8506768712?pr=30114

**Review by coder-karen — COMMENTED** [review]
Those original errors seem to be gone now!

I left a comment with one issue I found.

Another issue I noticed is that if I delete a custom address field while I have the field sorts page open in a separate browser window, then go back to the field sorts page without clicking refresh and then click save, I do see the blank lines showing (no label) instead of the fields being deleted, and get this error in my log: `PHP Warning:  Undefined array key 1 in /usr/local/src/jetpack-monorepo/projects/plugins/crm/admin/settings/field-sorts.page.php on line 229`.

However I assume it is unexpected behaviour to already have the Field Sorts page open when making changes to custom fields. I notice this was also an issue prior to this PR (it exists on trunk), so not relevant to this PR (just spotted and sharing now, can create an issue if we think it's behaviour that ought to be different?)

File: projects/plugins/crm/admin/settings/custom-fields.page.php (hunk 1 above)
Comment (tbradsha) [reply projects/plugins/crm/admin/settings/custom-fields.page.php]: Looks like it failed prior to the `trunk` merge as well:
https://github.com/Automattic/jetpack/actions/runs/4705060841

Comment (tbradsha) [reply projects/plugins/crm/admin/settings/custom-fields.page.php]: ~Resolved this in a050b84bf4.~ Prematurely declared fixed. 😅

File: projects/plugins/crm/admin/settings/custom-fields.page.php [hunk 2]
```diff
@@ -293,6 +293,89 @@
...
+			'obj_key'    => 'zbsCompanyFields',
+		),
+	);
+
+	/*
+	 * Creating a lookup table so we don't need to have multiple nested loops in
+	 * the next step. We do this by first adding all currently valid custom
+	 * fields, then adding all others (that should be the default ones).
+	 */
+	$custom_field_names = array();
```
tbradsha: This variable name was throwing me off until I tested a bit deeper...really it returns an array of all fields for the given object, not just custom fields. Maybe `$sort_field_names` would be more readable?

projects/plugins/crm/admin/settings/custom-fields.page.php
tbradsha: There, fixed in a050b84 + 0fa1ddf.

tbradsha:
I've resolved the fatal error, which only occurred if one hadn't previously saved field sorts (either through the Field Sorts settings page, or, as of this PR, with the Custom Fields settings page). As it turns out, the sort and hide settings created on site init didn't have valid keys and even if they did, they had a bad value (`false` instead of `array()`). New sites now init with the proper setting keys and with an empty array.

I also solved the data race issue @coder-karen reported: aad9062

coder-karen:
Thanks @tbradsha ! Both of those two issues look resolved now 🥳 

I noticed something else (sorry, only spotted now while testing with a fresh installation on a JN site) - when adding the first custom field and heading to Field Sorts, I see that new custom field in each of the three sort areas (Address, Contact and Company Fields).

With a subsequent custom field, this only shows under 'Address Fields' (as do all custom fields), until clicking save. Then they or it shows up under the other two field sections.

Seems that was also behaviour from before though so nothing new, for that I could create a new issue too.

projects/plugins/crm/admin/settings/custom-fields.page.php
coder-karen: I agree. The comment above the variable name helps to some degree, though ideally I imagine we wouldn't be relying on the comment for an explanation of what the variable holds.

tbradsha: Updated var name: 0f600c79c7

tbradsha:
> I noticed something else

Let's file this as a new issue to keep this PR from ballooning. :^)

tbradsha (approved):

coder-karen (approved):

PR #30119, opened 2023-04-17
CRM: Allowing minification of JS in development builds

coder-karen:
## Proposed changes:

* This PR makes sure JavaScript files are properly minified in development builds. Prior to this the `.min.js` files were created in development builds, but were not actually minified.
* This PR also then makes sure that `.min.js.LICENSE.txt` files aren't created in development builds, while still being created for production builds.

### Other information:

## Jetpack product discussion

https://github.com/Automattic/zero-bs-crm/issues/3035

No.

## Testing instructions:

* To test this PR out, with this PR applied run the following command: `jetpack build plugins/crm`. You should notice JS files being properly minified, and no `.min.js.LICENSE.txt` files created.
* Add a small change to a JS file (I prefer to make the change at the end of the file so it's easy to track), then run the command again. Check the minified file to make sure the change is applied.
* Make sure the CRM dashboard and pages are loading and displaying as expected with this development build.


cleacos (approved):
I checked it and it creates the expected minified files.

LGTM!

PR #30130, opened 2023-04-17
[RNMobile] Cover case of GUID being empty after upload finish

### LINKED ISSUE #5650: Cover case of GUID being empty after upload finish
_**Internal reference:**_ p1680888431063159/1680539601.569399-slack-C04LWMHSGLC

All video uploads are automatically added to VideoPress so the GUID should be always included. However, seems that in some cases this happens a bit later than the upload, hence this value is not returned in the media metadata.

A potential workaround would be to try to [fetch the media metadata](https://developer.wordpress.com/docs/api/1.1/get/sites/%24site/media/%24media_ID/) until we get the GUID after the upload finishes.

fluiddot:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/5650.

## Proposed changes:

* When the upload finish, we'll check if the returned VideoPress GUID is empty. If so, we'll fetch it manually via the `media` endpoint and notify the block. During this state, the block will be displayed as being uploaded but without progress.
* This is an edge case but since we managed to reproduce it occasionally, this PR aims to cover that possibility.

### Other information:

## Jetpack product discussion

N/A

N/A

## Testing instructions:

Since it's hard to reproduce the case of getting an empty GUID after upload, we needed to apply the following patch on the native side to force it.

<details><summary>WordPress-iOS</summary>

```patch
diff --git forkSrcPrefix/WordPress/Classes/ViewRelated/Gutenberg/GutenbergMediaInserterHelper.swift forkDstPrefix/WordPress/Classes/ViewRelated/Gutenberg/GutenbergMediaInserterHelper.swift
index 24823ca107e701b5fbb7e295b78504a4fe2e398c..1b06b0890e50b5dc8e6fe333a5744834eae2a26b 100644
--- forkSrcPrefix/WordPress/Classes/ViewRelated/Gutenberg/GutenbergMediaInserterHelper.swift
+++ forkDstPrefix/WordPress/Classes/ViewRelated/Gutenberg/GutenbergMediaInserterHelper.swift
@@ -263,6 +263,8 @@ class GutenbergMediaInserterHelper: NSObject {
             }
             switch media.mediaType {
             case .video:
+                // TEST: Force VideoPress GUID to be empty.
+                media.videopressGUID = nil
                 // Fetch metadata when is a VideoPress video
                 if media.videopressGUID != nil {
                     EditorMediaUtility.fetchVideoPressMetadata(for: media, in: post) { [weak self] (result) in

```

</details> 

<details><summary>WP-Android</summary>

```patch
diff --git forkSrcPrefix/libs/editor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java forkDstPrefix/libs/editor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
index 9494980db126c5cbbafd4236340c87ef0205bb73..1ff521dcc3533feea35b3aaab99123d1bd23cdd0 100644
--- forkSrcPrefix/libs/editor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
+++ forkDstPrefix/libs/editor/src/main/java/org/wordpress/android/editor/gutenberg/GutenbergEditorFragment.java
@@ -1377,9 +1377,10 @@ public class GutenbergEditorFragment extends EditorFragmentAbstract implements
         mUploadingMediaProgressMax.remove(localMediaId);
 
         WritableNativeMap metadata = new WritableNativeMap();
-        if (mediaFile.getVideoPressGuid() != null) {
-            metadata.putString("videopressGUID", mediaFile.getVideoPressGuid());
-        }
+        // TEST: Force VideoPress GUID to be empty.
+        // if (mediaFile.getVideoPressGuid() != null) {
+        //     metadata.putString("videopressGUID", mediaFile.getVideoPressGuid());
+        // }
 
         getGutenbergContainerFragment()
                 .mediaFileUploadSucceeded(Integer.parseInt(localMediaId), mediaFile.getOptimalFileURL(),

```

</details> 

**Steps:**

* Apply the above patches.
* Build and open the app.
* Add a VideoPress block.
* Upload a video from the local device.
* Wait until the upload finish.
* Observe that the video is displayed and can be played.


SiobhyB (approved):
Followed the testing steps for both Android and iOS, confirmed this works perfectly for me and the code looks good too. Thanks @fluiddot 🔥

PR #30136, opened 2023-04-17
Menu icon: avoid error when used with older version of Logo

jeherve:
## Proposed changes:

Avoid fatal errors when combining multiple plugins with different package versions.

### Other information:

## Jetpack product discussion

See p1681747330884549-slack-CBG1CP4EN

* No

## Testing instructions:

This is a bit involved (see Slack discussion). It relies on using an old version of the Logo package, alongside this branch. You can achieve that by using the Jetpack plugin with this branch alongside another plugin that relies on an old version of the Logo package, like the VaultPress plugin I believe.

### DIFF
--- a/projects/packages/admin-ui/changelog/fix-method-admin-ui
+++ b/projects/packages/admin-ui/changelog/fix-method-admin-ui
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Avoid errors when used in combination with an older version of the Assets package.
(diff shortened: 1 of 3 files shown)


projects/packages/admin-ui/changelog/fix-method-admin-ui
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Avoid errors when used in combination with an older version of the Assets package.
```
anomiex:
```suggestion
Avoid errors when used in combination with an older version of the Logo package.
```

anomiex (commented):
Logo package, not Assets. Otherwise LGTM.

anomiex (approved):

PR #30142, opened 2023-04-17
VideoPress: add is_videopress_url() helper method

retrofox:
This PR adds the `is_videopress_url()` Util method that we'll use in follow-up janitorial tasks.

Fixes #

## Proposed changes:

* VideoPress: add is_videopress_url() helper method

### Other information:

## Testing instructions:

Since the helper is not used yet:
* Take a look at the function
* Run tests

```cli
jetpack test packages/videopress php -
```

image

### DIFF
--- a/projects/packages/videopress/changelog/update-videopress-add-util-videopress-url
+++ b/projects/packages/videopress/changelog/update-videopress-add-util-videopress-url
@@ -1,0 +1,4 @@
+Significance: patch
+Type: added
+
+VideoPress: add is_videopress_url() helper function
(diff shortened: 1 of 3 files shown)


nunyvega (changes requested):
Tests work well. While looking for all VideoPress URLs, I came across `https://videos.files.wordpress.com/*`. Should we include it or are we intentionally not adding this one?

retrofox:
> videos.files.wordpress.com

This is a valid URL for the video file: `https://videos.files.wordpress.com/<guid>/<filename>,` I think it shouldn't be included as a valid URL.

Specifically for the follow-up tasks that are coming on, we need to check that the URL is a VideoPress one when filtering it to get the oEmbed HTML. AFAIK, we don't filter the video file but the video URL.

nunyvega (approved):
Code look good and tests as expected

PR #29910, opened 2023-04-04
Show social review prompt with both Social and Jetpack installed

jboland88:
## Proposed changes:

* This diff allows the Jetpack Social review prompt to show if the main Jetpack plugin is active. Previously, if you were using the Jetpack Social plugin and also had the main Jetpack plugin active, the review request prompt would not show
* This is a follow-up to #27917 

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

pbNhbs-5DY-p2#comment-13811

## Does this pull request change what data or activity we track or use?

* No, it just shows the review prompt in more cases. There are some analytics events associated with the review prompt which will now be recorded in more cases.

## Testing instructions:

* Follow the testing plan on #27917
* Note that if you are testing with the Jetpack Beta plugin, you will need to set the Jetpack plugin to use the Release Candidate instead of the latest stable.
* Test with both Jetpack Social and the main Jetpack plugin active and confirm that you still receive the review prompt when the conditions in the test plan are met.

PR #27917, opened 2022-12-13
Add review prompt for social plugin

jboland88:
#### Changes proposed in this Pull Request:

* This PR adds a review request prompt for the Jetpack Social plugin after a post is shared to connected social media accounts

#### Other information:
- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

#### Jetpack product discussion

pbNhbs-3wm-p2

#### Does this pull request change what data or activity we track or use?

Yes, this PR adds 3 new analytics events around the new review prompt:
- `jetpack_social_plugin_review_prompt_view` (recorded when review prompt is rendered)
- `jetpack_social_plugin_review_prompt_new_review_click` (recorded when "Leave a review" is clicked)
- `jetpack_social_plugin_review_prompt_dismiss_click` (recorded when "Dismissed" is clicked)

#### Testing instructions:

- You will need a Jetpack site with the Jetpack Social plugin installed
- Note that if you also have the Jetpack plugin installed, you will need to deactivate it for this test
- From your test site, you can apply this branch to Jetpack Social using the Jetpack Beta testing plugin
- Connect at least one social account for your site using Jetpack Social
- After your social account is connected, create a new post and publish it. Social sharing should be on by default and your post should be shared to your connected social media account
- In the post-publish sidebar of the editor, you should see the new review prompt

Screen Shot 2022-12-13 at 11 41 10 AM

- When the prompt renders, the `jetpack_social_plugin_review_prompt_view` analytics event should fire. Search for this in your browser's network panel to confirm.
- Confirm that clicking on "Leave a Review" takes you to the reviews page for the Jetpack Social plugin
- Clicking on "Leave a Review" should also fire the `jetpack_social_plugin_review_prompt_new_review_click` analytics event, search for this in your browser's network panel.
- Do not dismiss the prompt yet, we'll test that later :)
- Now, go create another new post. This time, use the Jetpack social sidebar (click on the megaphone icon in the upper right hand corner of the editor) to disable social sharing for this post
- Publish this new post with social sharing disabled. The review prompt should not show in the post-publish sidebar since sharing was disabled.
- Run this test again with a 3rd post, but disable the individual sharing account(s) before publishing instead of using the toggle to disable social sharing all together. You should not see the review prompt when publishing this 3rd post.
[...]

PR #30148, opened 2023-04-18
Revert "Show social review prompt with both Social and Jetpack installed"

spsiddarthan:
Reverts Automattic/jetpack#29910

We are going to revert this PR to fix the initial state of the plugin p1681469685034749-slack-C02JJ910CNL and pick up 1198218726984184-as-1204393860935466 in the following sprints. The social plugin will have the review prompts working when it's active. The review prompts won't show up when both Social and Jetpack are active. 

This is the https://github.com/Automattic/jetpack/pull/27917 which originally added the review prompts. But this [line](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/social/src/class-jetpack-social.php#L295) sets the state for the Social plugin only if the Jetpack plugin is not active which leads to https://github.com/Automattic/jetpack/pull/29910 being created, I presume. That accidentally stomped over the plugin's initial state.

* Yes, this PR reinstates 3 new analytics events around the new review prompt:

- jetpack_social_plugin_review_prompt_view (recorded when review prompt is rendered)
- jetpack_social_plugin_review_prompt_new_review_click (recorded when "Leave a review" is clicked)
- jetpack_social_plugin_review_prompt_dismiss_click (recorded when "Dismissed" is clicked)

## Testing instructions:

- When the prompt renders, the jetpack_social_plugin_review_prompt_view analytics event should fire. Search for this in your browser's network panel to confirm.
- Confirm that clicking on "Leave a Review" takes you to the reviews page for the Jetpack Social plugin
- Create another new post, but disable sharing. No prompt will show up since sharing was disabled. 
- Publish a post again, but disable the individual connections. Make sure the prompts don't show up. 
- Publish a post again with sharing enabled and notice that the prompt shows up. Dismiss the prompt
- Publish a post again with sharing enabled and notice that no prompt shows up.


danielpost:
@spsiddarthan I created a new site to test this. Connected Social, added an account and published a new post. Instead of the prompt I got an empty panel (see screenshot). Any idea what might cause this?

Screenshot 2023-04-18 at 10 05 57@2x

spsiddarthan:
>Any idea what might cause this?

I'll have to check that, Daniel! Will get back.

pablinos:
> Instead of the prompt I got an empty panel (see screenshot). Any idea what might cause this?

If that was with Jetpack Social rather than Jetpack then I think that's the problem the original PR was trying to fix. Maybe we should fix it rather than reverting it.

pablinos:
The empty [panel comes from here](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js#L209-L244). `showNotices` is false causing an empty panel to be rendered. I'm not sure why we don't get one in the pre-publish sidebar, but the problem doesn't seem related to this PR.

pablinos (approved):
This is working well for me. It reverts the changes, and the review prompt is still displayed when Social is active, but not Jetpack.

spsiddarthan:
>If that was with Jetpack Social rather than Jetpack then I think that's the problem the original PR was trying to fix. 

Right, it is happening only in the Jetpack plugin.

>Maybe we should fix it rather than revert it.
Definitely, I tried my hand at it but it seems a like a bit of work. So, I decided to do that in the next sprint. 

I am merging this fix for now, so the social plugin's reviews work correctly.

PR #30152, opened 2023-04-18
Animate the response view in JP Forms dashboard

ice9js:
## Proposed changes:

This patch adds an animation for the responses view on the JP Forms dashboard, like requested by @digitalwaveride.

https://user-images.githubusercontent.com/8056203/232763777-085ead36-5477-44ef-b917-de407d2d4b73.mov

In order to accomplish this, I opted to create a new `SwitchTransition` component which should nicely encapsulate the logic when switching between two views where they both need to be visible during the transition. This is different from `react-transition-group`'s own `SwitchTransition` where it will only start animating the new element once the previous one has fully exited.  
The component is only responsible for managing the class names on the wrapper, the actual animations would still need to be defined individually for each use case through regular CSS.

### Other information:

No.

## Testing instructions:

- Go to `admin.php?page=jetpack-forms` and wait for the responses to load.
- When switching between the responses on the left, you should see the right pane animate.


CGastrell (dismissed):
Looks fine, works great! :shipit:

digitalwaveride (dismissed):
Thanks so much Kuba!!! 🚀 
Awesome first version implementation! 👍

digitalwaveride (approved):
👍

PR #30156, opened 2023-04-18
VideoPress: enqueue token bridge file in the front-end only when required

### LINKED ISSUE #24765: VideoPress: avoid enqueuing js on pages that do not need it.
This is a follow-up to #23949. We now enqueue `modules/videopress/js/videopress-token-bridge.js` on all pages of websites where the module is enabled:
https://github.com/Automattic/jetpack/blob/304e134beec8ec1d3dd79e04288bc14a5a849c6c/projects/plugins/jetpack/modules/videopress/class.jetpack-videopress.php#L72-L95

That seems problematic since we do not need it on all those pages.

@Automattic/greenhouse Do you think we could be a bit more restrictive there?

retrofox:
This PR iterates over how the app enqueues the token-bridge file. Currently, it's always enqueued in the front end.

* First, it removes the code line when enqueued in the front end.
* And it keeps relying only on hooking to the embed_oembed_html filter.

Fixes https://github.com/Automattic/jetpack/issues/24765

## Proposed changes:

* VideoPress: enqueue token bridge file in the front-end only when required

### Other information:

## Testing instructions:

* Go to the block editor
* Create a post without a VideoPress video block
* Save the post
* Go to the front end
* **Before**, confirm the client loads the token-bridge file

Screen Shot 2023-04-18 at 10 56 35

*  **After**,  confirm the client does not load the token-bridge file.

* Go back to the editor
* Create a VideoPress video block
* Set video privacy private 
* Go to the front end again
* Confirm the token-bridge file is now loaded
* You can confirm the bridge provides the token to the client by using debug() lines

```
localStorage.setItem( 'debug', 'videopress:*' )
```


projects/packages/videopress/src/class-jwt-token-bridge.php
```diff
@@ -30,8 +30,7 @@ public static function init() {
 			return;
 		}
 
-		// Expose the VideoPress token to the Block Editor context, including the front-end editor.
-		add_action( 'enqueue_block_assets', array( __CLASS__, 'enqueue_jwt_token_bridge' ), 1 );
```
retrofox, line 34: with this line, the file is always enqueued in the front end.

dhasilva (approved):
Tests fine on JN and WoA :+1:

projects/packages/videopress/src/class-jwt-token-bridge.php
jgcaruso: I think this was necessary since we don't know ahead of time if there will be a video block added to a new post.

But if you have some way of detecting that and enqueuing it afterwards then that might be OK?

retrofox: First and just in case, we need to keep in mind that we are enqueuing the token-bridge file in the block editor content line below via the `enqueue_block_editor_assets` action. It guarantees to deal with private videos in this context (block editor)

The `enqueue_block_assets` action [fires after enqueuing block assets for both editor and front-end](https://developer.wordpress.org/reference/hooks/enqueue_block_assets/).

Enqueuing the token bridge by hooking to `enqueue_block_assets` is not optimal (IMHO) since it could be possible to create a VideoPress video instance using other methods, like a shortcode or a simple URL?
Ultimately, the app will filter the VideoPress URL request to the oEmbed markup to serve it to the client.

I tested with a shortcode:
image

With a raw URL:
image

VideoPress embed block
image

Extended Core Video block (v5)

image

...and the token bridge is always enqueued via the `embed_oembed_html` hook.

If we don't remove that line, the script is always enqueued. It doesn't matter if the site has or hasn't a VideoPess video block, a shortcode, or whatever.

fgiannar (approved):
Tested using the provided instructions and works as expected 👍

retrofox (commented):
Tested with a wpvideo shortcode in the footer:

image

The JS is enqueued

PR #30160, opened 2023-04-18
CRM: Resolves 3063 - broken link in settings on white label installs

tbradsha:
Resolves Automattic/zero-bs-crm#3063 - broken link in settings on white label installs

## Proposed changes:

We linked to the extensions page on the settings page, but on white label said page is removed. This PR updates the link to go to the modules page, which was probably the original intent (extensions and modules were originally on the same page), and it also updates the button from "Manage Features" to "Manage modules".

### Other information:

## Testing instructions:

There's not an easy way to test white label, but in this case we're just changing the link to a valid one, so just make sure the new link works on a normal site.

1. Go to the settings page: `/wp-admin/admin.php?page=zerobscrm-plugin-settings`
2. Click on "Manage modules".

It should take you to the modules page.

### DIFF
--- a/projects/plugins/crm/changelog/fix-crm-3063-fix_manage_features_button
+++ b/projects/plugins/crm/changelog/fix-crm-3063-fix_manage_features_button
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Settings: fixed broken link on white label installs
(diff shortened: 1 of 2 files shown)


cleacos (approved):
LGTM!

PR #30161, opened 2023-04-18
Forms: Create dashboard landing page

jcheringer:
## Proposed changes:
* Create Forms dashboard landing page

### Other information:

## Jetpack product discussion
N/A

No 

## Testing instructions:

Make sure you have the Forms package and the new Dashboard enabled

* You should be able to access the landing page by hitting this URL: `wp-admin/admin.php?page=jetpack-forms#/landing`
* Also, if you have no responses when accessing the new Dashboard, you should see the Landing page instead of the responses Inbox


projects/packages/forms/src/blocks/contact-form/edit.js
```diff
@@ -140,6 +140,18 @@ export const JetpackContactFormEdit = forwardRef(
...
+				! hasInnerBlocks &&
+				registerBlockVariation &&
+				! isPatternsModalOpen &&
+				window.location.search.indexOf( 'showPatterns' ) !== -1
+			) {
+				setIsPatternsModalOpen( true );
+			}
+			// eslint-disable-next-line react-hooks/exhaustive-deps
+		}, [] );
+
```
jcheringer: Would we have a better way to do this?
The idea here is to be able to open the Patterns modal when the user enters the Editor.
That's something triggered by a button on the landing page.

ice9js: Hmm, it doesn't look too bad, I think it's fine. Just maybe let's call the var `showJetpackFormsPatterns` or something like that to minimize chances of any conflicts down the line.

projects/packages/forms/src/class-jetpack-forms.php
```diff
@@ -46,6 +46,13 @@ public static function plugin_url() {
 		return plugin_dir_url( __FILE__ );
 	}
 
+	/**
+	 * Get the plugin URL.
```
ice9js: Nit: should it be 'assets URL' ?

projects/packages/forms/src/dashboard/components/details/index.js
```diff
@@ -0,0 +1,16 @@
+import './style.scss';
+import { Icon, chevronRight } from '@wordpress/icons';
+
+const Details = ( { children, summary } ) => {
+	return (
+		<details className="jp-forms__details">
```
ice9js, line 6: Fancy!

projects/packages/forms/src/dashboard/inbox/index.js
```diff
@@ -103,6 +108,12 @@ const Inbox = () => {
 		} );
 	}, [ currentPage, fetchResponses, query ] );
 
+	useEffect( () => {
```
ice9js: So, here's an idea, how about instead of dealing with state and everything we check on the backend if you have any feedback posts and return a separate config variable or simply redirect to the landing page if empty?

As is, there are some glitches. When you select a tab or a filter that happens to have no responses but you have plenty of responses in general, it'd still redirect you to the landing page. We can't deal with all of it on the frontend alone and doing it on the backend would also make it more responsive by immediately showing you what you need to see :)

projects/packages/forms/src/dashboard/landing/style.scss
```diff
@@ -0,0 +1,344 @@
+.jp-forms__landing {
```
ice9js, line 1: Will we also need WP.com styles for this?

ice9js (changes requested):
Looks good but I think now that you've dealt with the router, it'd be a better idea to deal with the redirect and related logic on the backend making the whole thing more bulletproof and responsive. Let me know what you think.

projects/packages/forms/src/dashboard/landing/index.js
```diff
@@ -0,0 +1,204 @@
+import { createInterpolateElement } from '@wordpress/element';
```
ice9js, line 5: Should we just turn that into a 'view'? The way I imagined it initially `components/` were meant to be reusable and then we'd create 'top-level directories' for everything else.

So, shall we move it to `/dashboard/landing-page` ?

ice9js (commented):
Looks good but I think now that you've dealt with the router, it'd be a better idea to deal with the redirect and related logic on the backend making the whole thing more bulletproof and responsive. Let me know what you think.

CGastrell:
Not sure if this is supposed to be functional, I'm finding some weirdness. Like, whenever I try to switch tabs (Inbox/Spam/Trash) it will immediately jump to the landing page. Then, trying to hit "back" button will just not take me anywhere, it'll glitch and be back at the landing (worth mentioning: I'm not switching to an empty tab, I have responses on all tabs).

Other than that, change looks great! I'd defer to @digitalwaveride for this img that seems a bit "blurry"?
image

Maybe the screenshot doesn't do justice, worth taking a look at the landing on browser.

projects/packages/forms/src/dashboard/landing/style.scss
jcheringer: @digitalwaveride, will we need it?

projects/packages/forms/src/dashboard/landing/index.js
jcheringer: Not sure if understood what you mean.
The landing page is already under `/dashboard`, but it's called `landing`.
Are you suggesting renaming it to `landing-page`?

CGastrell:
Nice! Image looks sharper, more crispy.

Switching tabs still gets me to the landing page even with content on all tabs.

jcheringer:
> Nice! Image looks sharper, more crispy.
> 
> Switching tabs still gets me to the landing page even with content on all tabs.

Yeah... Handling the redirect logic on the frontend seems to not be the best idea.
Kuba suggested doing it on the backend, but I need to figure out how.
He gave me some insights. Need to see if I can get it through

projects/packages/forms/src/dashboard/landing/style.scss
digitalwaveride: We will work with a different screenshot for wp.com ... but keep the JP branding.

jcheringer:
I don't know if it's the best solution, but on c073889, I added some logic to handle the redirection.
Due to our current limitation of using the HashNavigation, I added a new property to the API response to check if there are posts, regardless of the filters.
Then, I'm using this on the Frontend to handle the redirection.

cc @ice9js @CGastrell

ice9js:
Hmm, yeah, I see how we can't really do the redirection on the backend, but I guess it's also not really necessary. I think what you have now is a good idea, just if we simplify it a little more.

Instead of adding the prop to the responses endpoint, add it to `$config` in `Dashboard::render_dashboard()` - that way it's value is available as soon as the page is loaded and we can immediately display the correct view instead of having to make another request and wait for the response.

ice9js:
I've made the change. `hasFeedback` is now available on the global config immediately so we don't need to wait on the responses request to come back.  

I've also taken the opportunity to move the inbox view under `/responses` in case we ever need or want to add more views. For example, originally we planned on there being a forms overview which would be the default, so this way we're avoiding any future conflicts. So the entire flow is now like this:

- `/`: Redirects to `/responses`.
- `/responses`: Redirects to `/landing` when `hasFeedback` is `false`. Shows the responses inbox otherwise.
- `/landing`: Landing page. Always accessible.

By the way, the landing page looks great @jcheringer 👍!

ice9js (approved):
The landing page looks great! I think we're good to go if you're happy with the redirect change.

jcheringer:
Thanks for handling the redirection, @ice9js!
That solution is much faster than waiting for the API response.
Looks good to me 🚀

CGastrell (approved):
Looks awesome, and works nicely! Nice job! :shipit:

PR #30162, opened 2023-04-18
Jetpack Social: Use image from Media Picker for OpenGraph image

danielpost:
## Proposed changes:

This PR changes the behavior of the Media Picker (part of Jetpack Social's Advanced Plan). With this PR, the Media Picker not just sets the image that gets uploaded during Publicize, but also sets the OpenGraph image that's used whenever a post gets shared.

It also changed the Media Picker to always be visible in the sidebar, since it now has a use even when no Publicize connections are toggled on.

As a bonus, it also fixes 1202583289493814-as-1204345803096704/f.

### Other information:

## Jetpack product discussion

1202583289493814-as-1204411041263280/f
1202583289493814-as-1204345803096704/f

No.

## Testing instructions:

* Add the Jetpack Social Advanced plan to your site.
* Create a new post and pick an image from the media picker.
* Publish the post.
* View the post's source code and verify that `og:image` and `twitter:image` are set to the URL of the image you picked.
* Verify that `twitter:card` is set to `summary_large_image`.
* Add a featured image and verify that `og:image` and `twitter:image` are still set to the image from the media picker.
* Remove the image from the media picker and check that and `og:image` and `twitter:image` are set to the featured image.

**With Social Image Generator:**
* Add Social Image Generator to your site.
* Verify that `og:image` and `twitter:image` use the image from the media picker if set, and use the SIG image if not.
* Verify that `twitter:card` is set to `summary_large_image` in both cases.

### DIFF
--- a/projects/js-packages/publicize-components/changelog/change-jetpack-social-use-media-picker-for-og-image
+++ b/projects/js-packages/publicize-components/changelog/change-jetpack-social-use-media-picker-for-og-image
@@ -1,0 +1,4 @@
+Significance: minor
+Type: changed
+
+Use attached media for the OpenGraph image
--- a/projects/js-packages/publicize-components/package.json
+++ b/projects/js-packages/publicize-components/package.json
@@ -1,7 +1,7 @@
 {
 	"private": true,
 	"name": "@automattic/jetpack-publicize-components",
-	"version": "0.20.2",
+	"version": "0.21.0-alpha",
 	"description": "A library of JS components required by the Publicize editor plugin",
 	"homepage": "https://github.com/Automattic/jetpack/tree/HEAD/projects/js-packages/publicize-components/#readme",
 	"bugs": {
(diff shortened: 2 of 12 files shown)


gmjuhasz (dismissed):
This looks and tests well!

Just a question:
> Verify that og:image and twitter:image use the image from the media picker if set, and use the SIG image if not.

Will we use the media picker image image even if SIG is enabled? I thought SIG is "stronger" than the picked image?

danielpost:
> Will we use the media picker image image even if SIG is enabled? I thought SIG is "stronger" than the picked image?

@gmjuhasz Mhmm, maybe you're right, because right now it's impossible to use SIG and upload an image at the same time. I'll change it.

That being said, in the next version of the UI it won't be possible to have both, since you have to either pick an image/video or create an image.

gmjuhasz (approved):
Thanks for the answers, looks good!

Issue #29901, opened 2023-04-04
My Jetpack: Introduce standalone options in product card menu


renatoagds:
Depends on #29904 #29903 #29902 

You can add in the product menu the default option to install or activate the standalone plugin (if available).

![Image](https://user-images.githubusercontent.com/1663717/229812665-92da35e0-0405-4fb1-a1bd-c6e8ab650680.png)

It will require the adjustment to force API to install and activate the standalone plugin and not Jetpack.

Some steps to acomplish this:

- [x] add side-effect action to trigger the installation, the same way we have one for activating the product - PR: #30143 
- [x] add button to "Install X plugin" if the product has a standalone plugin and it's not installed or active - PR: #30153
- [x] handle the click of the button to dispatch the side-effect action - PR: #30168

lhkowalski:
I will take a look into the first step, the side-effect action.

lhkowalski:
Going to look into adding the menu option for installing the plugin as well. We have some code history to help this taks on [this old closed PR](https://github.com/Automattic/jetpack/pull/29840/files#diff-0f409ceb5046d993dec9c29a4ed60662af5899bc168d13994b84b97a8244f803R77).

lhkowalski:
Now I am going to connect the action menu with the action, and also fix the small menu issues that @renatoagds commented about on the last PR.

lhkowalski:
We merged the changes that we needed to make this happens. There are some UI/UX polishing being made as follow-ups to this issue, but they all have their own issue here on GH.

PR #30168, opened 2023-04-18
My Jetpack: Connect standalone plugin menu actions

lhkowalski:
Relates to #29901.

## Proposed changes:

* Change the `use-product` hook to export the standalone plugin installation action
* Create function to handle the call for the installation and pass it to the `ProductCard`
* Fix menu visibility control to consider the standalone plugin status
* Enable the action menu for VideoPress
* Track events related to install and activate standalone plugin

### Other information:

### Testing the installation + activation of a standalone product

* Spin a new JN site with only Jetpack
* Connect the site
* Go to `Jetpack > My Jetpack` and check the VideoPress card; it should show as inactive since we don't activate it yet; it will also show the three dots to open the actions menu:

Screen Shot 2023-04-18 at 18 44 38

* Click on the action menu icon, on the top right corner of the VideoPress card
* You should see the "Install Plugin" action available:

Screen Shot 2023-04-18 at 18 45 44

* Click on "Install Plugin"
* The card will switch to a loading state and then the plugin will be installed and activated
* After the loading state finishes, refresh the page to get the card with the new status; there is a known issue that the install standalone endpoint is returning a stale product status after the installation; I will fix it on a follow up;
* After the refresh, the card is now shown as active and the actions menu only shows the "Manage" option:

Screen Shot 2023-04-18 at 18 48 29

* Go to `Plugins > Installed Plugins` and confirm that the VideoPress plugin is now installed:

Screen Shot 2023-04-17 at 18 38 56

### Testing the activation of a standalone product

* We are going to disable the plugin to check the action menu again
* On the same test site, go to `Plugins > Installed Plugins` and disable the VideoPress plugin
* Go to `Jetpack > My Jetpack` and confirm that the VideoPress card still show as active (since the module on Jetpack got enable on the prior step):

Screen Shot 2023-04-18 at 18 50 02

* Click on the action menu icon, on the top right corner of the VideoPress card
* You should see the "Manage" (since the product is active) and "Activate Plugin" (since the plugin is installed but not active) actions available:

Screen Shot 2023-04-18 at 18 51 04

* Click on "Activate Plugin"
* The card will switch to a loading state and then the plugin will be activated on the backend
* After the loading state finishes, the card still shows as active and the actions menu only shows the "Manage" option:

Screen Shot 2023-04-18 at 18 51 37

* Go to `Plugins > Installed Plugins` and confirm that the VideoPress plugin is now active again:

Screen Shot 2023-04-17 at 18 38 56

### DIFF
--- a/projects/packages/my-jetpack/_inc/components/connected-product-card/index.jsx
+++ b/projects/packages/my-jetpack/_inc/components/connected-product-card/index.jsx
@@ -63,6 +63,7 @@ const ConnectedProductCard = ( { admin, slug, children, showMenu = false } ) => 
 			hasStandalonePlugin={ standalonePluginInfo?.hasStandalonePlugin }
 			isStandaloneInstalled={ standalonePluginInfo?.isStandaloneInstalled }
 			isStandaloneActive={ standalonePluginInfo?.isStandaloneActive }
+			isConnected={ isRegistered && isUserConnected }
 		>
 			{ children }
 		</ProductCard>
--- a/projects/packages/my-jetpack/_inc/components/connected-product-card/index.jsx
+++ b/projects/packages/my-jetpack/_inc/components/connected-product-card/index.jsx
@@ -41,30 +41,6 @@ const ConnectedProductCard = ( { admin, slug, children, showMenu = false } ) => 
 		navigateToConnectionPage,
 	] );
 
-	/*
-	 * Redirect to connection page if the user is not connected
-	 */
-	const handleInstallAndActivateStandalonePlugin = useCallback( () => {
-		if ( ( ! isRegistered || ! isUserConnected ) && requiresUserConnection ) {
-			navigateToConnectionPage();
-			return;
-		}
-
-		/**
-		 * For both installing and activating the plugin, the action is the same
-		 * because the backend endpoint performs both actions
-		 * - installing when is not installed
-		 * - activating when is not active
-		 */
-		installStandalonePlugin();
-	}, [
-		installStandalonePlugin,
-		isRegistered,
-		isUserConnected,
-		requiresUserConnection,
-		navigateToConnectionPage,
-	] );
-
 	const Icon = getIconBySlug( slug );
 
 	return (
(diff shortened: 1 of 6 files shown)


projects/packages/my-jetpack/_inc/components/product-card/index.jsx
```diff
@@ -34,6 +35,30 @@ const Menu = ( {
...
+		} );
+	}
+
+	if ( showStandaloneOption && isStandaloneInstalled && ! isStandaloneActive ) {
+		items.push( {
+			label: __( 'Activate Plugin', 'jetpack-my-jetpack' ),
+			onClick: onActivate,
+			icon: check,
+		} );
+	}
```
renatoagds: I think it is better not to manipulate the prop directly. We're changing the original array, and if we need to rely on it at some moment, it will be another one. And this way, we change the layout, not including the divider between "actions" and "install/activate." If you prefer to keep using array, better to split between custom items and default items, but I vote to keep the default items declarative.

projects/packages/my-jetpack/_inc/components/connected-product-card/index.jsx
```diff
@@ -40,6 +41,30 @@ const ConnectedProductCard = ( { admin, slug, children } ) => {
 		navigateToConnectionPage,
 	] );
 
+	/*
+	 * Redirect to connection page if the user is not connected
+	 */
```
renatoagds: I vote to remove the installation option if the user isn't connected. Strange experience to be moved for connection after clicking on install, and in the end, nothing will be installed. We have a big notice at the top warning the user that isn't connected.

projects/packages/my-jetpack/_inc/components/product-card/index.jsx
lhkowalski: Good point!

About the divider, I ended up having the menu showing it above the install option but with nothing above the divider, I think when the module is not active and the plugin is not installed.

I will move back to the declarative format and work on a small condition to show/not show the divider.

projects/packages/my-jetpack/_inc/components/connected-product-card/index.jsx
lhkowalski: Technically it's not impossible to trigger the installation for a disconnected site, but I agree that it may be a sub-optimal experience. I will try removing the option to see how things show up.

projects/packages/my-jetpack/_inc/components/product-card/index.jsx
lhkowalski: Changed the code back to the declarative format and also added a check to show the divider.

projects/packages/my-jetpack/_inc/components/connected-product-card/index.jsx
lhkowalski: Removed the connection redirect from the code and added a check for the `error` status, but I think it's not enough yet. I will figure out a way to have the connection state available on the menu to include it on the button condition.

lhkowalski: Done. I injected the connection state into the product card and use it to decide between showing or not the menu. For me it does make sense to hide the menu everytime there is a connection problem, since it will not show anything that could work disconnected and will not help fix the connection issue, at least for now. Do you agree?

renatoagds (approved):

PR #30175, opened 2023-04-19
AI Chat Block

glendaviesnz:
## Proposed changes:

* Wires in the AI Chat beta block into the search API so it can be tested against live content.

### Other information:

 - Still a beta block, so no tests added yet - will add some when more of structure finalised.

## Jetpack product discussion
Some background discussion here peGwbA-2b-p2

No

## Todo

- [ ] Work out how to pass in the blog id to search - probably just the current blog id for starters, and needs to be limitied to public blogs only so private blog data is not passed out to OpenAI API 
- [ ] Add the option to specify default questions to choose from
- [ ] Add a little more design finesse, some fade-in, fade-out of text changes, etc.
- [ ] There is a CORs error when using the block on the frontend of sandbox server
- [ ] Work out how the block can be restricted to only sites with Jetpack search subs and how summarization of results will be metered

## Testing instructions:

* Sync PR to sandbox and add the AI Chat block to a post and try asking a WPcom related question

https://user-images.githubusercontent.com/3629020/236056356-f266b146-4f74-4e12-98c3-0562a7ab36d0.mp4

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-chat/loading.js
+++ b/projects/plugins/jetpack/extensions/blocks/ai-chat/loading.js
@@ -1,0 +1,39 @@
+/**
+ * WordPress dependencies
+ */
+import { SVG, Circle } from '@wordpress/components';
+export default function Loading() {
+	return (
+		<SVG width="24" height="24" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">
+			<Circle cx="4" cy="12" r="3" opacity="1">
+				<animate
+					id="spinner_qYjJ"
+					begin="0;spinner_t4KZ.end-0.25s"
+					attributeName="opacity"
+					dur="0.75s"
+					values="1;.2"
+					fill="freeze"
+				/>
+			</Circle>
+			<Circle cx="12" cy="12" r="3" opacity=".4">
+				<animate
+					begin="spinner_qYjJ.begin+0.15s"
+					attributeName="opacity"
+					dur="0.75s"
+					values="1;.2"
+					fill="freeze"
+				/>
+			</Circle>
+			<Circle cx="20" cy="12" r="3" opacity=".3">
+				<animate
+					id="spinner_t4KZ"
+					begin="spinner_qYjJ.begin+0.3s"
+					attributeName="opacity"
+					dur="0.75s"
+					values="1;.2"
+					fill="freeze"
+				/>
+			</Circle>
+		</SVG>
+	);
+}
--- a/projects/plugins/jetpack/extensions/blocks/ai-chat/question-answer.js
+++ b/projects/plugins/jetpack/extensions/blocks/ai-chat/question-answer.js
@@ -1,18 +1,13 @@
 /**
  * WordPress dependencies
  */
-import {
-	Button,
-	TextControl,
-	Spinner,
-	KeyboardShortcuts,
-	ExternalLink,
-} from '@wordpress/components';
+import { Button, TextControl, KeyboardShortcuts, ExternalLink } from '@wordpress/components';
 import { RawHTML, useEffect, useState } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 /**
  * Internal dependencies
  */
+import Loading from './loading';
 import useSubmitQuestion from './use-submit-question';
 
 // This component displays the text word by word if show animation is true
@@ -89,7 +84,7 @@ export default function QuestionAnswer() {
 				<div className="jetpack-ai-chat-answer-container">
 					{ isLoading ? (
 						<>
-							<Spinner />
+							<Loading />
 							{ waitString }
 						</>
 					) : (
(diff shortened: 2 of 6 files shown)


projects/plugins/jetpack/extensions/blocks/ai-chat/use-submit-question.js
```diff
@@ -1,53 +1,67 @@
...
-		}, 3000 );
+		setWaitString( __( 'Let me think about that for a moment.', 'jetpack' ) );
+		const encoded = encodeURIComponent( question );
+
+		const { terms } = await apiFetch( {
+			path: addQueryArgs( `wpcom/v2/sites/${ blogId }/jetpack-search/ai/search`, {
+				stop_at: 'terms',
+				query: encoded,
+			} ),
+		} );
```
glendaviesnz: Need to add some error handling in here

glendaviesnz: Done.

projects/plugins/jetpack/extensions/blocks/ai-chat/use-submit-question.js
```diff
@@ -1,53 +1,69 @@
...
-			url: 'http://en.support.wordpress.com/video-tutorials/using-the-block-editor/',
-			title: 'Video Tutorials: Using the Block Editor',
-		},
-		{
-			url: 'http://en.support.wordpress.com/wordpress-editor/add-content-blocks/',
-			title: 'Add Content Using Blocks',
-		},
-		{ url: 'http://en.support.wordpress.com/wordpress-editor/blocks/', title: 'Blocks' },
-	],
-};
```
glendaviesnz, line 28: This should be replaced with the current blog id somehow as the idea of this block is to provide AI chat across the user's own blog content. We also need to check that it is a public blog and has jetpack search available and not show the block if not.

projects/plugins/jetpack/extensions/blocks/ai-chat/use-submit-question.js
```diff
@@ -12,22 +12,55 @@ import { referenceTexts, initialWaitTexts } from './response-strings';
...
+				`https://public-api.wordpress.com/wpcom/v2/sites/${ blogId }/jetpack-search/ai/search`,
+				{
+					stop_at: 'response',
+					query: encoded,
+					stream: true,
+				}
+			);
+			source = new EventSource( path );
+
+			source.addEventListener( 'open', () => {
```
glendaviesnz, line 44: This streaming of answers from OpenAI is dependent on D109907 being merged and an additional `stream` param being piped through API calls

projects/plugins/jetpack/extensions/blocks/ai-chat/loading.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * WordPress dependencies
+ */
+import { SVG, Circle } from '@wordpress/components';
+export default function Loading() {
```
jeherve, line 5: Should we use the core spinner instead of bringing our own loading indicator?
https://github.com/WordPress/gutenberg/tree/trunk/packages/components/src/spinner#readme

glendaviesnz: I will leave this to Agora to make a call on. My thinking at the time was it needed something different to the standard loading spinner in order to give the impression that rather than there being a delay with loading, the AI just needed time to think 😄, which is sort of the case. 

This may not be such an issue if we do switch to streaming from the API as the response will start arriving much quicker in chunks.

projects/plugins/jetpack/extensions/blocks/ai-chat/question-answer.js
```diff
@@ -1,77 +1,40 @@
...
-];
+export default function QuestionAnswer() {
+	const [ answer, setAnswer ] = useState( '' );
+	const [ aiResponse, setAiResponse ] = useState( '' );
 
-// This component displays the text word by word if show animation is true
-function ShowLittleByLittle( { html, showAnimation, onAnimationDone } ) {
-	// This is the HTML to be displayed.
-	const [ displayedRawHTML, setDisplayedRawHTML ] = useState( '' );
+	const markdownConverter = new MarkdownIt( { html: true } );
```
glendaviesnz, line 18: In theory there should only be markdown in the response from OpenAI, so `html: true` is only needed as we keep parsing the same string, and it turns existing html to html entities if this is not set - but we probably need to have a good think about any security implications, re XSS, etc. around this approach.

projects/plugins/jetpack/extensions/blocks/ai-chat/question-answer.js
```diff
@@ -1,77 +1,40 @@
...
-		<div className="content">
-			<RawHTML>{ displayedRawHTML }</RawHTML>
-		</div>
-	);
-}
-export default function QuestionAnswer() {
-	const { question, setQuestion, answer, isLoading, submitQuestion, references } =
-		useSubmitQuestion();
+		}
+	}, [ aiResponse ] );
```
glendaviesnz, line 28: we definitely don't want `answer` as a dependency as it will trigger an infinite rerender, so if this approach is kept we should do a functional update to  setAnswer instead to avoid this dependency warning `setAnswer(a => ...)`

glendaviesnz:
Closing as looks like the chatbot is going to be in its own package.

PR #30178, opened 2023-04-19
CRM: Profile summary - removing deleted invoice counts and totals

coder-karen:
## Proposed changes:

* This PR ensures that total amounts (current total value and invoice total, plus the invoice count in brackets) are factoring in deleted invoices (by that I mean not including the deleted invoices), on contact and company profile summaries. It also corrects the invoice total counts and amount on both the Contact and Company 'View All' pages as well.

### Other information:

## Jetpack product discussion

https://github.com/Automattic/zero-bs-crm/issues/2673

No.

## Testing instructions:

* Testing locally on trunk, or on a test site with the latest stable version of Jetpack CRM, create both a company and contact with multiple invoices attached to them. On the profile summary, view the current total value and invoice total (plus count in brackets next to the total).
  * Contacts: `/wp-admin/admin.php?page=zbs-add-edit&action=view&zbstype=contact&zbsid={contactid}
  * Companies: `http://localhost/wp-admin/admin.php?page=zbs-add-edit&action=view&zbstype=company&zbsid={companyid}`
* Take note of the totals and invoice count - these should include the totals from deleted invoices.
* You can also test the totals box on the contact 'View All' page - the invoice total (and total) should exclude deleted invoices as well: `/wp-admin/admin.php?page=manage-customers`.
* In the developer console on that same contacts 'View All' page, type in `jpcrm_totals_table` and hit enter. You should see the updated totals, including some new variables such as `invoices_total_inc_deleted` and `total_sum_inc_deleted` that can be used at any stage.
* Another thing to test is the modified list view for both contacts and companies on their respective 'View All' pages, adding in invoice counts and invoice totals: https://kb.jetpackcrm.com/knowledge-base/how-do-i-change-columns-on-contact-list-view/
* These should also be correct (not including deleted invoices)
* Using the [Jetpack Beta plugin](https://github.com/Automattic/jetpack-beta) on a test site with this branch applied or testing locally with the branch checked out, open up the same pages again. The counts should be reduced to factor in deleted invoices.

## Screenshots:

A contact profile summary before a $3 deleted invoice was removed:
Screenshot 2023-04-19 at 09 34 29

And after:

Screenshot 2023-04-19 at 09 33 16

Info box changes:

Screenshot 2023-04-19 at 09 40 45
Screenshot 2023-04-19 at 09 40 11

Contacts 'View All' page including invoice totals and counts in the list, plus the totals box (bottom right):

Screenshot 2023-05-05 at 10 50 10


tbradsha (changes requested):
This works as written on the contact/company profile, but it'd be better to implement in DAL so that (for example) `$contact['total_value']` would always reflect the total value. That way it'd properly show the correct values everywhere else it's used (e.g. list view, API calls).

I think the proper place to fix values would be in DAL's `tidy_contact()` and `tidy_company()`.

Likewise, for counts you'd want to update `contactHasCountObjType()` (company doesn't appear to have this, so maybe a new DAL function like contacts have)?

coder-karen:
Thanks for the review @tbradsha !

> but it'd be better to implement in DAL so that (for example) $contact['total_value'] would always reflect the total value. That way it'd properly show the correct values everywhere else it's used (e.g. list view, API calls).

I thought about doing something like that initially, but stopped. Do we really want a contact's / company's `total_value` to always factor in deleted invoice quantities? If we're 100% positive of that, then yes it makes sense. I doubted it, so figured a helper function to delete the quantities and counts make sense.

Like you said, that would affect list view total and API calls, and also the totals box on the list view pages - though the invoice (and total) sums come from `invoices_total` rather than `total_value` in that case ( in that respect it seems changing `invoices_total` makes more sense as `total_value` is calculated from that plus transaction value anyway).

I guess thinking ahead - if we were to build out anything that makes use of invoice totals, would we default to thinking invoice totals = those not in deleted status, or they include them all? Or would we be better off creating a helper function that calculates totals including deleted invoices, if needed. I imagine that all depends on the use-case.

I realize I'm actually on the line on this and because of that I'm deferring to wanting a helper function, but you've got much more experience with CRM and also with what the customers would expect.

So that in mind, I'm happy to make the changes! But want to run my thinking by you first ^ just in case that changes your mind 😄

tbradsha:
> Do we really want a contact's / company's total_value to always factor in deleted invoice quantities? If we're 100% positive of that, then yes it makes sense. I doubted it, so figured a helper function to delete the quantities and counts make sense.

I'd say yes. :^) At the end of the day, I can't think of any contexts where one would want the total value to include deleted invoices, so changing one value further up the chain is a better move.

That said, you can also add an additional key (e.g. `invoices_value_with_deleted`) in case for some reason there's a need for it.

On top of that, you might add two invoice count keys (e.g. `invoice_count` and `invoice_count_with_deleted`) in the same place (`tidy_contact()` and `tidy_company()`)to prevent an additional call to the DAL.

Just suggestions; feel free to adjust/ignore as desired!

> in that respect it seems changing `invoices_total` makes more sense as `total_value`

Agreed; we'd want to filter early for best readability.

coder-karen:
> I'd say yes. :^) 

Perfect - I'll get on to this tomorrow.

> That said, you can also add an additional key (e.g. invoices_value_with_deleted) in case for some reason there's a need for it.
> On top of that, you might add two invoice count keys (e.g. invoice_count and invoice_count_with_deleted) in the same place (tidy_contact() and tidy_company())to prevent an additional call to the DAL.

Great idea!

coder-karen:
Since the same type of function to count obj type is now used for both companies and contacts, should that become one single helper function - and if so where should it best be placed? in Dal3.Helpers.php, or DAL3.php? 

Related - by copying over the logic PHPCS was picking up several issues with the DB call, so this looks like something for the PHPCS cleanup (didn't feel I should spend too much time digging in to the best way to do that right now, knowing it will get picked up in that cleanup as well).

The existing helper functions, they are still in DAL3.Helpers.php. I found I was still duplicating code so it was useful to keep those, but just call them closer to the source. Perhaps there is a better location than DAL3.Helpers.php for them?

Lastly, it looks like more work is needed to updated the totals box on the list view page. Perhaps one for another issue? it currently seems a little confusing with total quantities - possibly needs an explanation added with an info icon 🤔

tbradsha:
> Since the same type of function to count obj type is now used for both companies and contacts, should that become one single helper function - and if so where should it best be placed? in Dal3.Helpers.php, or DAL3.php?

It's a mess, so your discretion. I'd either:

1. Copy it to Companies DAL.
2. Make it more generic, and drop it in `DAL3.php`. If you went with this route, I'd remove the Contacts one and update any references to it.

That said, an alternative (or additional option) might be to simply add the key/value to the main Contact object within `tidy_contact()`, thereby potentially rendering the aforementioned function obsolete.

That would depend if there's ever a context where one would want the count of objects attached to a contact without retrieving contact info. Again, your choice! :^)

Hopefully that made sense.

> by copying over the logic PHPCS was picking up several issues with the DB call, so this looks like something for the PHPCS cleanup

🙈 

> Lastly, it looks like more work is needed to updated the totals box on the list view page. Perhaps one for another issue? it currently seems a little confusing with total quantities - possibly needs an explanation added with an info icon

Ah, good catch. Probably one for another issue (and another helper function of sorts 😬 ).

coder-karen:
I found the newly renamed and moved function `customer_has_count_obj_type` is still useful as it's used for quotes and events as well, but I removed the extra logic from it and added that to the tidy functions - and also added the logic for transactions as well to stay still somewhat within scope (re counts including deleted items). So this should give more count options for invoices and transactions - both for contacts and companies.

projects/plugins/crm/includes/ZeroBSCRM.DAL3.php
```diff
@@ -571,6 +571,57 @@ public function fields_to_hide_on_frontend( $obj_type_int = false ){
...
+	 * Returns a count of objects of a type which are associated with a company or contact
+	 * Supports Quotes, Invoices, Transactions, Events currently
+	 *
+	 * @param int $cust_id The company / contact ID.
+	 * @param int $obj_type_id The type constant being checked (eg ZBS_TYPE_QUOTE).
+	 * @param int $zbs_type The contact or company type, for example ZBS_TYPE_COMPANY, ZBS_TYPE_CONTACT (default contact).
+	 *
+	 * @return int The count of relevant objects of the given type.
+	 */
+	public function customer_has_count_obj_type( $cust_id, $obj_type_id, $zbs_type = ZBS_TYPE_CONTACT ) {
```
tbradsha: A nitpick and not a blocker, but this should probably be `contact_has_count_obj_type()` with a `$contact_id` param. If you do change it, make sure to update its references elsewhere.

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Helpers.php
```diff
@@ -5696,6 +5696,35 @@ function zeroBSCRM_check_amount_due_mark_paid($invoice_id=-1){
...
+		return array(
+			'total' => 0,
+			'count' => 0,
+		);
+	}
+	$count_deleted = 0;
+	$total_deleted = 0;
+
+	foreach ( $all_invoices as $invoice ) {
+		if ( $invoice['status'] === __( 'Deleted', 'zero-bs-crm' ) ) {
```
tbradsha, line 5500: Good job making sure this is comparing with the translated value!

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Helpers.php
```diff
@@ -5696,6 +5696,35 @@ function zeroBSCRM_check_amount_due_mark_paid($invoice_id=-1){
 		return false;
 	}
 
+/**
+ * Helper function calculate the number of deleted invoices or transactions and the
```
tbradsha:
```suggestion
 * Helper function to calculate the number of deleted invoices or transactions and the
```

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Helpers.php
```diff
@@ -5696,6 +5696,35 @@ function zeroBSCRM_check_amount_due_mark_paid($invoice_id=-1){
...
+ * total value of those invoices, for any particular contact / company.
+ *
+ * @param array $all_invoices An array of all invoice or transaction data for a contact / company.
+ *
+ * @returns array An array with the deleted invoices total, and the deleted invoices count.
+ */
+function jetpackCRM_deleted_totals( $all_invoices = null ) {
```
tbradsha: Typically we've prefixed new functions with `jpcrm_` (eventually namespace will remove the need for a prefix), but not a blocker. However:

* the name of the function here isn't very descriptive as to what it does/what it returns
* the `$all_invoices` param and the logic throughout indicates it only handles invoices

It seems we've expanded some of the logic to transactions as well as invoices. For better or worse, transaction counts/statuses are manually controlled in the settings and [filtered a level up in the query](https://github.com/Automattic/jetpack/blob/0835a5018bcc12737ed94ba66fdae6960496417a/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Contacts.php#L1861) (if I'm reading the code correctly, and [here's companies](https://github.com/Automattic/jetpack/blob/0835a5018bcc12737ed94ba66fdae6960496417a/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Companies.php#L468)). As such, won't this only affect invoices?

If correct, I'd propose reducing the scope of the function and calling it something like:

`jpcrm_deleted_invoice_total_and_counts()`

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Companies.php
```diff
@@ -2759,8 +2761,22 @@ private function tidy_company($obj=false,$withCustomFields=false){
...
+			if ( isset( $obj->invoices_total ) ) {
+				$company                            = $zbs->DAL->companies->getCompany( $res['id'], array( 'withInvoices' => true ) ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+				$deleted_invoice_details            = jetpackCRM_deleted_totals( $company['invoices'] );
+				$res['invoices_total']              = $obj->invoices_total - $deleted_invoice_details['total'];
+				$res['invoices_total_with_deleted'] = $deleted_invoice_details['total'];
+				$res['invoices_count_with_deleted'] = $zbs->DAL->customer_has_count_obj_type( $res['id'], ZBS_TYPE_INVOICE, ZBS_TYPE_COMPANY ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+				$res['invoices_count']              = $res['invoices_count_with_deleted'] - $deleted_invoice_details['count'];
+			}
+			if ( isset( $obj->transactions_total ) ) {
+				$company                                = $zbs->DAL->companies->getCompany( $res['id'], array( 'withTransactions' => true ) ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
```
tbradsha: I think most of this block isn't needed due to earlier filtering of transactions?

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Contacts.php
```diff
@@ -3622,8 +3622,22 @@ private function tidy_contact($obj=false,$withCustomFields=false){
...
+				if ( isset( $obj->invoices_total ) ) {
+					$contact                            = $zbs->DAL->contacts->getContact( $res['id'], array( 'withInvoices' => true ) ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+					$deleted_invoice_details            = jetpackCRM_deleted_totals( $contact['invoices'] );
+					$res['invoices_total']              = $obj->invoices_total - $deleted_invoice_details['total'];
+					$res['invoices_total_with_deleted'] = $deleted_invoice_details['total'];
+					$res['invoices_count_with_deleted'] = $zbs->DAL->customer_has_count_obj_type( $res['id'], ZBS_TYPE_INVOICE, ZBS_TYPE_CONTACT ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+					$res['invoices_count']              = $res['invoices_count_with_deleted'] - $deleted_invoice_details['count'];
+				}
+				if ( isset( $obj->transactions_total ) ) {
+					$contact                                = $zbs->DAL->contacts->getContact( $res['id'], array( 'withTransactions' => true ) ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
```
tbradsha: I think this isn't needed due to earlier filtering of transactions?

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Companies.php
```diff
@@ -2759,8 +2761,22 @@ private function tidy_company($obj=false,$withCustomFields=false){
 
             // if have totals, pass them :)
             if (isset($obj->quotes_total)) $res['quotes_total'] = $obj->quotes_total;
-            if (isset($obj->invoices_total)) $res['invoices_total'] = $obj->invoices_total;
-            if (isset($obj->transactions_total)) $res['transactions_total'] = $obj->transactions_total;
+			if ( isset( $obj->invoices_total ) ) {
+				$company                            = $zbs->DAL->companies->getCompany( $res['id'], array( 'withInvoices' => true ) ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+				$deleted_invoice_details            = jetpackCRM_deleted_totals( $company['invoices'] );
+				$res['invoices_total']              = $obj->invoices_total - $deleted_invoice_details['total'];
+				$res['invoices_total_with_deleted'] = $deleted_invoice_details['total'];
```
tbradsha: Doesn't this just return `the total value of deleted invoices`, as opposed to `the total value of all invoices including deleted`?

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Companies.php
```diff
@@ -2759,8 +2761,22 @@ private function tidy_company($obj=false,$withCustomFields=false){
...
+				$res['invoices_count_with_deleted'] = $zbs->DAL->customer_has_count_obj_type( $res['id'], ZBS_TYPE_INVOICE, ZBS_TYPE_COMPANY ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
```
tbradsha: Perhaps simpler (if I'm understanding correctly):

```suggestion
				$res['invoices_count_with_deleted'] = count( $company['invoices'] );
```

Or even:

```suggestion
				$res['invoices_count_with_deleted'] = count( $obj->invoices );
```

projects/plugins/crm/includes/ZeroBSCRM.DAL3.php
tbradsha: Now reading this again after looking through the code, I see that you've made this more generic. "Customer" is pretty consistently related to "Contact" in the CRM (and in large part we've tried to remove references to customer. Likewise, `has` typically means it'll return a `bool`. Perhaps instead:

`assigned_obj_type_count()`

I'm not in love with that function name either, so feel free to think on it!

projects/plugins/crm/includes/ZeroBSCRM.DAL3.php
```diff
@@ -571,6 +571,57 @@ public function fields_to_hide_on_frontend( $obj_type_int = false ){
...
+			case ZBS_TYPE_EVENT:
+				$table_name = $ZBSCRM_t['events'];
+				break;
+
+			default:
+				// For any unsupported objtype.
+				return -1;
+		}
+
+		$obj_query = 'SELECT COUNT(obj_table.id) FROM ' . $table_name . ' obj_table'
```
tbradsha, line 611: This is legacy code, but at first I was wondering why there's a join on a given table. In the end it's probably so orphan references are ignored.

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Companies.php
```diff
@@ -2759,8 +2761,22 @@ private function tidy_company($obj=false,$withCustomFields=false){
 
             // if have totals, pass them :)
             if (isset($obj->quotes_total)) $res['quotes_total'] = $obj->quotes_total;
-            if (isset($obj->invoices_total)) $res['invoices_total'] = $obj->invoices_total;
-            if (isset($obj->transactions_total)) $res['transactions_total'] = $obj->transactions_total;
+			if ( isset( $obj->invoices_total ) ) {
+				$company                            = $zbs->DAL->companies->getCompany( $res['id'], array( 'withInvoices' => true ) ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
```
tbradsha: This seems unnecessarily recursive? I suspect what you're wanting is already available within `$obj`, so no need to grab `$company` anew.

tbradsha (changes requested):
Several items inline; hopefully my questions make sense.

projects/plugins/crm/includes/ZeroBSCRM.DAL3.php
coder-karen: Yeah, I know we have history around the usage of customer - it did seem to be a compromise between contact and company so as not to be one or the other. How about `client_obj_type_count` using that same logic? 🤔

tbradsha: "Client" is also used specifically in reference to "Contact" too (e.g. Client Portal). The function is generic, and would work with transactions assigned to invoices, for example, so I don't see a need to define it narrowly?

coder-karen: > would work with transactions assigned to invoices,

I see! I honestly think your suggestion may be best - I've gone with `specific_obj_type_count_for_assignee` for now and modified the assignments, and related variables and comments.

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Companies.php
coder-karen: It does, I spotted that too when going through everything again. Changed that now.

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Companies.php
coder-karen: > $res['invoices_count_with_deleted'] = count( $company['invoices'] );

Yes, thank you, that would work. It is simpler, with no DB query (added it to a new commit).

>  count( $obj->invoices );

The $obj object doesn't return the list of invoices so that doesn't work.

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Companies.php
coder-karen: Unfortunately $obj itself doesn't include the list of invoices or transactions in order to perform calculations on them. It does include totals, which are not correct.

coder-karen:
Thank you once again, @tbradsha !

II've added a commit with some further changes based on some of the reviews.

What I've also done is simplify things back to how they were pre-transaction edits (so, not including changes to transaction totals / counts). Plus some cleanup on the contact / company view pages - where the actual (deleted included) - invoice (and transaction) count was needed to determine what to show under the transactions tab (same with invoices).

With transactions, I'll create a new issue ([created here](https://github.com/Automattic/zero-bs-crm/issues/3078)). Based on what you mentioned with the calculations happening further up, I then realized we have an option on the transaction settings page to check which statuses to include:
`wp-admin/admin.php?page=zerobscrm-plugin-settings&tab=transactions`.
Therefore the logic will be different - and it isn't quite right as it doesn't modify counts (likely why I started looking into adding it here).

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Helpers.php
coder-karen: > For better or worse, transaction counts/statuses are manually controlled in the settings and [filtered a level up in the query](https://github.com/Automattic/jetpack/blob/0835a5018bcc12737ed94ba66fdae6960496417a/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Contacts.php#L1861) (if I'm reading the code correctly, and [here's companies](https://github.com/Automattic/jetpack/blob/0835a5018bcc12737ed94ba66fdae6960496417a/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Companies.php#L468)). As such, won't this only affect invoices?

Thank you for pointing this out! I've added more in [a comment here](https://github.com/Automattic/jetpack/pull/30178#issuecomment-1531296759) - basically I think better to leave fixing any details around transactions to another issue now.

> If correct, I'd propose reducing the scope of the function and calling it something like:

I've now changed the scope back to transactions so the params and logic will be ok, but I've changed the function name, thanks!

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Companies.php
coder-karen: I've removed transaction logic (gone back to how it was) for now - counts aren't correct even though the totals seem to be, but this can be something for another issue.

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Contacts.php
coder-karen: Taken out now, left a comment [with more info here](https://github.com/Automattic/jetpack/pull/30178#issuecomment-1531296759) :)

projects/plugins/crm/admin/company/view.page.php
```diff
@@ -594,7 +590,7 @@ function jpcrm_render_company_view_page( $id = -1 ) {
 								// prep link to create a new invoice
 								$new_invoice_url = jpcrm_esc_link( 'create', -1, ZBS_TYPE_INVOICE ) . '&prefillco=' . $company['id'];
 
-								if ( $company_invoices_count > 0 ) {
+								if ( count( $company['invoices'] ) > 0 ) {
```
tbradsha: FYI, `$company_invoices_count` still exists. :^)

projects/plugins/crm/admin/contact/view.page.php
```diff
@@ -872,7 +869,7 @@ function jpcrm_render_contact_view_page( $id = -1 ) {
 
 							// prep link to create a new invoice
 							$new_invoice_url = jpcrm_esc_link( 'create', -1, ZBS_TYPE_INVOICE ) . '&zbsprefillcust=' . $contact['id'];
-							if ( $contact_invoice_count > 0 ) {
+							if ( count( $contact['invoices'] ) > 0 ) {
```
tbradsha: `$contact_invoice_count` still exists if you want to use it.

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Companies.php
tbradsha: Hm... I still don't like that within `getCompany()` we're calling `getCompany()`.

Currently we've got a query (actually, two depending on the params) that gets `invoices_total` (sum of all invoice values), and then we make another call to grab the entire contact including invoices, and then we overwrite `invoices_total` (hopefully in all cases).

If `invoices_total` already exists at this point (and would be "incorrect", as it includes deleted invoices), perhaps we're trying to fix it too late and should handle it earlier (e.g. in the SQL queries).

As written it works, but it's not terribly efficient. At minimum I'd swap it out for a call to `$zbs->DAL->invoices->getInvoices()`.

projects/plugins/crm/admin/company/view.page.php
coder-karen: Right, but as I understand it there we want the real count (without deleted invoices). If it was 0 factoring in a deleted invoice, it won't show that deleted invoice under the 'invoices' tab on the company view page.

projects/plugins/crm/admin/contact/view.page.php
coder-karen: I believe that's the same situation as above: https://github.com/Automattic/jetpack/pull/30178#discussion_r1182581342 (we want the real, full count).

tbradsha: In that case you can use `$contact['invoices_count_with_deleted']`. But either way works. :^)

projects/plugins/crm/admin/company/view.page.php
tbradsha: In that case you can use `$company['invoices_count_with_deleted']`. But either way works. :^)

coder-karen: Ha... if I have it now, I ought to use it 🤦

projects/plugins/crm/admin/contact/view.page.php
coder-karen: Got it, makes sense to use the new key!

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Companies.php
coder-karen: > perhaps we're trying to fix it too late and should handle it earlier (e.g. in the SQL queries

Given this issue - https://github.com/Automattic/zero-bs-crm/issues/3077 - it does seem that handling it earlier, in the SQL queries, may be better. A bit like `getTransactionStatusesToIncludeQuery` but for invoices.

coder-karen:
I've now redone much of this PR so the logic exists at the stage where we we are making SQL queries to gather information from the database (for contacts and customers). Because of this, the invoice totals are now correct on the Customer 'View All' page as well, and I've updated the testing instructions to include these two steps:

* You can also test the totals box on the contact 'View All' page - the invoice total (and total) should exclude deleted invoices as well: `/wp-admin/admin.php?page=manage-customers`.
* In the developer console on that same contacts 'View All' page, type in `jpcrm_totals_table` and hit enter. You should see the updated totals, including some new variables such as `invoices_total_inc_deleted` and `total_sum_inc_deleted` that can be used at any stage.

projects/plugins/crm/admin/contact/view.page.php
```diff
@@ -74,10 +74,16 @@ function jpcrm_render_contact_view_page( $id = -1 ) {
 		}
 
 		// contact obj counts
-		$contact_quote_count       = $zbs->DAL->contacts->contactHasCountObjType( $id, ZBS_TYPE_QUOTE );
-		$contact_invoice_count     = $zbs->DAL->contacts->contactHasCountObjType( $id, ZBS_TYPE_INVOICE );
-		$contact_transaction_count = $zbs->DAL->contacts->contactHasCountObjType( $id, ZBS_TYPE_TRANSACTION );
-		$contact_task_count        = $zbs->DAL->contacts->contactHasCountObjType( $id, ZBS_TYPE_EVENT );
+		$contact_quote_count   = $zbs->DAL->specific_obj_type_count_for_assignee( $id, ZBS_TYPE_QUOTE, ZBS_TYPE_CONTACT ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
```
tbradsha, line 77: Interesting we don't don't grab this for transactions and quotes within the contact object, but that's how it was before and is out of scope.

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Companies.php
```diff
@@ -466,12 +466,16 @@ public function getCompany($id=-1,$args=array()){
...
-                    // trans (with status):
+					$extraSelect .= ',(SELECT SUM(invstotal.zbsi_total) FROM ' . $ZBSCRM_t['invoices'] . ' as invstotal WHERE invstotal.ID IN (SELECT DISTINCT zbsol_objid_from FROM ' . $ZBSCRM_t['objlinks'] . ' WHERE zbsol_objtype_from = ' . ZBS_TYPE_INVOICE . ' AND zbsol_objtype_to = ' . ZBS_TYPE_COMPANY . ' AND zbsol_objid_to = company.ID)' . $inv_status_query_add . ') as invoices_total'; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
+					// invs including deleted:
+					$extraSelect .= ',(SELECT SUM(invstotalincdeleted.zbsi_total) FROM ' . $ZBSCRM_t['invoices'] . ' as invstotalincdeleted WHERE invstotalincdeleted.ID IN (SELECT DISTINCT zbsol_objid_from FROM ' . $ZBSCRM_t['objlinks'] . ' WHERE zbsol_objtype_from = ' . ZBS_TYPE_INVOICE . ' AND zbsol_objtype_to = ' . ZBS_TYPE_COMPANY . ' AND zbsol_objid_to = company.ID)) as invoices_total_inc_deleted'; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
```
tbradsha: In an ideal world this should probably only occur if the "invoices" module is enabled, but at this stage we'd need to ensure no implications everywhere else, and thus it's out of scope.

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Invoices.php
```diff
@@ -2681,6 +2681,23 @@ public function getInvoiceStatus($id=-1){
...
+		 * that should be included in "total value" fields, excluding 'deleted' status invoices.
+		 *
+		 * @param str $table_alias_sql - if using a table alias pass that here, e.g. `invoices.`.
+		 * @return array
+		 */
+		public function get_invoice_status_except_deleted_for_query( $table_alias_sql = '' ) {
+
+			$invoice_statuses_except_deleted = array( 'Draft', 'Unpaid', 'Paid', 'Overdue' );
+
+			$inv_statuses_str = $this->build_csv( $invoice_statuses_except_deleted );
```
tbradsha, line 2695: Huh, didn't know about `$this→build_csv()`.

coder-karen: Much of the logic here is thanks to `getTransactionStatusesToIncludeQuery` - including that call :)

coder-karen:
Fresh refactor done - @tbradsha pointed out that it's possible to add items to the 'list view' page (invoice totals and counts) with the drag and drop interface, and as I hadn't realized that those totals also weren't working properly - now fixed. Separately I also discovered that the invoice totals column in companies 'list view' wasn't working at all (not working in production, that is), leading to some changes to the Company list view call to not only fix that but make sure the relevant counts and totals were properly available on the list view page.

That plus changes to the counts - the counts are also calculated as early as possible (via SQL queries), thanks to a suggestion from @tbradsha - and some SQL help as well.

projects/plugins/crm/admin/contact/view.page.php
```diff
@@ -890,7 +890,7 @@ function jpcrm_render_contact_view_page( $id = -1 ) {
 
 								// if we have more than we're showing, communicate that
 								// Note this is defunct until we add contact filters to our object list views.
-								if ( $contact_invoice_count_inc_deleted > $contact_invoice_count ) {
+								if ( count( $contact['invoices'] ) > $contact_invoice_count_inc_deleted ) {
```
tbradsha, line 893: I'm pretty sure by definition these are already equal as written. :^)

tbradsha (dismissed):
The code looks good. I've also checked the following places for inconsistencies with deleted invoice totals and counts, but didn't notice any:

* contact profile
* company profile
* contact listview columns
* contact listview totals table
* company listview columns

I did find one bug we introduced: company was not sorting because we didn't include all the params needed. I fixed that in 17179af122.

I also chopped some unneeded lines in 0a138df382 and 461003acd6.

projects/plugins/crm/admin/contact/view.page.php
coder-karen: Yes, I believe that's currently the intention based just on the comment above the if statement, also based on how it was written before 😄 
I did think about chopping the whole if statement (or rather just checking the variable exists) but it seems like there may be future plans for adding a filter, in which case checking the manual count with the variable that currently holds the count would apparently be the way that's used to check if something else is being filtered... either way I thought it best to leave the same kind of logic still in place for now.

coder-karen:
Thank you for going through and fixing those inconsistencies @tbradsha !

I've taken another look through and spotted a preexisting typo as well affecting `$withQuotes`. Was about to fix it up, then wondered why that even exists for companies as there is no option to add quotes let alone filter by them, unless I'm missing something? I've removed the `$withQuotes` variables for now. Was tempted to do the same with `$withInvoices` but I can see how it is helpful to see that it is deliberately set to false. Not so sure on that one. Either way with the removals I've just added this still needs another review ;)

projects/plugins/crm/admin/contact/view.page.php
tbradsha: Yep, that's fine! Just figured I'd note it.

tbradsha:
> I've taken another look through and spotted a preexisting typo as well affecting `$withQuotes`. Was about to fix it up, then wondered why that even exists for companies as there is no option to add quotes let alone filter by them, unless I'm missing something? I've removed the $withQuotes variables for now. 

I considered chopping those when working on 0a138df. I'm okay with removing it like you did; I'd consider it basically boilerplate code that never was implemented. I didn't notice the typo, though, so good eye!

> Was tempted to do the same with `$withInvoices` but I can see how it is helpful to see that it is deliberately set to false.

I would lean toward removing the var throughout the file, as it doesn't appear to ever be used (anymore); I'm indifferent regarding the use of the param in the DAL call.

coder-karen:
Thanks @tbradsha - I think we should be all good now (I removed `$withInvoices` usage as well)!

tbradsha (approved):
Looks good!

PR #30179, opened 2023-04-19
Social: Add option to flag post as social post

gmjuhasz:
This change adds a new option to our `jetpack_social_options` array, which is used to flag a post whether it is a social post

## Proposed changes:

* Added new option to meta
* Extended `useAttachedMedia` hook to be able to get/set this option.

### Other information:

## Jetpack product discussion

1204272127368434-as-1204411047112874

No.
## Testing instructions:

* in `projects/js-packages/publicize-components/src/components/media-section/index.js` import `isSocialPost` and `updateIsSocialPost` from the `useAttachedMedia` hook.
* Console log out `isSocialPost`.
* Open a new post, and check the logs, it should log out `false`.
* Refresh the page with `updateIsSocialPost( true );` called, and save as draft.
* Log out `isSocialPost` again, now it should be true for this post even after refresh 
* Saving attachedMedia should work as before - select a media with the media picker, save post, refresh, and media should be selected

### DIFF
--- a/projects/js-packages/publicize-components/src/hooks/use-attached-media/index.js
+++ b/projects/js-packages/publicize-components/src/hooks/use-attached-media/index.js
@@ -2,6 +2,8 @@ import { useSelect, useDispatch } from '@wordpress/data';
 import { store as editorStore } from '@wordpress/editor';
 import { useCallback } from '@wordpress/element';
 
+const PUBLICIZE_STORE = 'jetpack/publicize';
+
 /**
  * @typedef {object} AttachedMediaHook
  * @property {Array} attachedMedia - List of media with ID, URL, and metadata.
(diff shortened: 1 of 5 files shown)


projects/js-packages/publicize-components/src/hooks/use-attached-media/index.js
```diff
@@ -32,8 +37,21 @@ export default function useAttachedMedia() {
 		[ currentOptions, editPost ]
 	);
 
+	const updateIsSocialPost = useCallback(
+		option => {
+			editPost( {
+				meta: {
+					jetpack_social_options: { ...currentOptions, is_social_post: option },
```
danielpost: I wonder if it's worth it to combine this method and the `updateAttachedMedia` into a shared method that takes a key/value pair to update the relevant option.

danielpost (commented):
This looks good so far. Are you planning on adding the checkbox in the sidebar in a follow-up PR?

I'm a little concerned about the wording of `isSocialPost`. I know that's what it will be called in the sidebar, but I think from a code perspective it's not very clear what it implies. What do you think about using something like `shouldUploadAttachedMedia` and `setShouldUploadAttachedMedia`? I think that makes it a lot more clear from just reading the code.

Having the "Share as a social post" checkbox toggle the `shouldUploadAttachedMedia` setting should also immediately tell anyone reading the code what the toggle does behind the scenes.

gmjuhasz:
Thanks for the review @danielpost!
> Are you planning on adding the checkbox in the sidebar in a follow-up PR?

 Yeah, I thought it makes sense to add the option logic first, but I can also add the other changes to this one. - Also it's good for Sid with the other task on this

> I'm a little concerned about the wording of isSocialPost.

I see your point, I was thinking of the name of the option and I thought `is_social_post` makes more sense there when we want to look up stats on how many posts have this feature on, rather than searching for `should_upload_attached_media`. Then I just matched that in the hook too. Would you change both places?

danielpost:
Follow-up PR is a good idea I think, keeps it small and contained.

> I see your point, I was thinking of the name of the option and I thought is_social_post makes more sense there when we want to look up stats on how many posts have this feature on, rather than searching for should_upload_attached_media. Then I just matched that in the hook too. Would you change both places?

We could add some tracking or something similar if we want to track it. I think it's alright either way, but without prior context  somebody coming in and looking at the `useAttachedMedia` hook would have no idea what a social post is or why it is in this particular hook. Personally I'd change it everywhere and only mention "social post" in the actual checkbox on the front-end.

gmjuhasz:
@danielpost I changed the naming, could you give another look?

danielpost (approved):
Looks good to me!

pablinos:
Looking at the testing instructions. I wonder if we should add a test for that. We could maybe implement it with the toggle.

PR #30189, opened 2023-04-19
Forms: Fix response details date value

jcheringer:
## Proposed changes:

* Fix the Form response date value in the response details section
* Add a link to the source value in the response details section

### Other information:

## Jetpack product discussion
N/A

No

## Testing instructions:

* Make sure you have the Forms package enabled
* Go to the response inbox in the ` wp-admin`
* Select a response and check if the issues mentioned above were fixed in the response details section


CGastrell (dismissed):
Works fine, thanks Joao! :shipit:

projects/packages/forms/src/dashboard/state/selectors.js
```diff
@@ -11,6 +11,7 @@ export const getResponses = state =>
 	map( state.responses, response => {
 		return {
 			...response,
+			rawDate: response.date,
 			date: dateI18n( 'M j, Y', response.date ),
```
ice9js: Actually... I'd argue `date` was always intended to be what `rawDate` is and any formatting to be done in the view layer. If we're not using `date` anywhere else now, can we just replace it and keep things clean?

ice9js (changes requested):
Hmm... do we really need two properties though? I feel like the selector should always be returning the 'raw' version anyway and it's up to the view layer to format it.

digitalwaveride (dismissed):
Date issue is fixed for me 👍 !

Thanks João!!! 🚀

jcheringer:
Thanks for pointing that out, @ice9js.
My last commit should address the issue and keep things cleaner.

CGastrell (dismissed):
LGTM!

projects/packages/forms/src/dashboard/inbox/list.js
```diff
@@ -13,6 +14,7 @@ const COLUMNS = [
 	{
 		key: 'date',
 		label: __( 'Date', 'jetpack-forms' ),
+		getValue: item => dateI18n( 'M j, Y', item.date ),
```
ice9js, line 27: Instead of adding complexity to the `<Table>` component, if it's just a simple property, not a component or anything, couldn't we just add the new property/value to the item [here](https://github.com/Automattic/jetpack/blob/trunk/projects/packages/forms/src/dashboard/inbox/list.js#L45)?

ice9js (changes requested):
Thanks Joao! I've had another look over it and I think we could maybe get away with it without altering the `<Table>` component.

Also, let's maybe use this opportunity to update the format [on the response tab](https://github.com/Automattic/jetpack/blob/trunk/projects/packages/forms/src/dashboard/inbox/response.js#L63) as @jeherve pointed out in pcsBup-1aJ-p2#comment-419.

projects/packages/forms/src/dashboard/inbox/list.js
jeherve, line 27: Should the date format be localized, like we do here?
https://github.com/Automattic/jetpack/blob/a05d65b2ed858afe5710e43d17ee07f6b6e79947/projects/packages/forms/src/dashboard/inbox/response.js#L66

Or maybe that wouldn't look that good in the UI, I'm not sure?

jcheringer: Yeah. The idea here was to make the date as short as possible, as the space is quite limited on smaller breakpoints.

cc @ice9js

jcheringer: Yeah, adding the property value to the `tableItems` would do the job.
But I'd argue if we're indeed adding complexity to the `<Table>` component.

IMHO, I think the `getValue` property gives us the flexibility to compose the cell values the way we desire.
We could even get rid of the `component` and `getProps` properties, since we could use the `getValue` function to return any markup, like:

```Javascript
		getValue: ( item ) => (
			<Button href={ item.entry_permalink } variant="link">{ item.source }</Button>
		),
```

Also, it'd make it clear to understand what a column value represents, since it'd be tied to the `COLUMNS` instead of anywhere else in the code.

ice9js: Hmm... fair enough, maybe that's indeed simpler. I'll allow it 😉.
But then maybe let's remove `component` and `getProps` later like you suggested - doesn't need to be in this PR though :)

jcheringer: Cool! Thanks for your blessing! 😆 
I've already removed the `component` and `getProps` and made the necessary adjustments to the code to work with the `getValue` function

ice9js: Thanks for cleaning it up! 🙇‍♂️

ice9js (dismissed):
Looks good to me 👍

ice9js (approved):

PR #26903, opened 2022-10-18
Jetpack: Add Salesforce Lead form template

CGastrell:
#### Changes proposed in this Pull Request:
This PR adds a new form template: Salesforce Lead Form. The form is a variation of the contact-form with a fixed set of inputs:
- Email
- First name
- Last name
- Job/Title
- Company
- Phone number

Sidebar settings include the Organization ID and (optional) a Campaign ID

Closes #26868 

#### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?

#### Jetpack product discussion
The change is being discussed/tracked at pcsBup-dV-p2

#### Does this pull request change what data or activity we track or use?

No

#### Testing instructions:
You'll need a Salesforce developer account, you can get one for free at salesforce.com.
Once you do, use the Jurassic Ninja link to create a new instance with this PR.
Change the Jetpack Constant JETPACK_BETA_BLOCKS to `true` by visiting the Settings->Jetpack Constants entry on WP-ADMIN

**NOTE: do NOT check JETPACK_EXPERIMENTAL_BLOCKS as it will override the BETA list. If it's checked, uncheck it.**

Edit a post and add a `/form`.

- [ ] you should see a new entry: Salesforce Lead Form

Add a Salesforce Lead Form.

- [ ] mentioned list of inputs should be present
- [ ] settings on the sidebar should allow you to enter an Organization ID

Follow the instructions on the settings to get your Organization ID, enter it on the settings
Visit the post and fill the form. Then, on your Salesforce instance, go to Apps -> Marketing. You should see a "Leads" tab.

- [ ] verify a lead now exists with the data you filled the form with

PR #30191, opened 2023-04-19
Forms: add post to url hook

CGastrell:
This PR implements the post-to-url feature on the Forms package as a replacement to the (left behind) [Salesforce integration implementation on the plugin](https://github.com/Automattic/jetpack/pull/26903).

## Proposed changes:
Add a new class to hook on feedback post creation. If the form attributes specify a post-to-url, it will trigger an HTTP post using the submit values. Salesforce configurations are accepted and turned into a generic configuration as a backwards compatibility feature.

The change also adds new expected attributes on the form to pave the road for new features:
- hiddenFields: TBD, simple name/value pairs added to forms
- postToUrl: TBD, post to url configuration
- salesforceData (backwards compatibility with plugin)

### Other information:

No

## Testing instructions:

Make sure you enable the package:
```php
add_filter( 'jetpack_contact_form_use_package', '__return_true' );
```

Run install on both the plugin and package (jetpack install packages/forms && jetpack install plugins/jetpack) to ensure autoloader update.

Publish a post with a Salesforce Lead form (or use an existing one). Make sure you use a valid `organizationID` on the form setup. Fill the form and submit. You should see a new lead created on your Salesforce instance.

Salesforce instance can be acquired free for testing, I can't share mine. Use the form's instructions to get your organization ID.


projects/packages/forms/src/class-jetpack-forms.php
```diff
@@ -62,4 +62,14 @@ public static function is_feedback_dashboard_enabled() {
...
 	}
+
+	/**
+	 * Initialization, calls for register pattern and loads any extra file/libs.
+	 */
+	public static function init() {
+		Util::register_pattern();
+
+		// Load any extra file/libs so they are available on the namespace for use (Forms\Post_To_URL_Hook::initialize)
+		require_once __DIR__ . '/class-post-to-url-hook.php';
```
ice9js: I don't understand why we need this at all? The file should already be loaded through the autoloader and you can just proceed to use `Automattic\Jetpack\Forms\Post_To_Url_Hook` where necessary.

projects/packages/forms/src/class-post-to-url-hook.php
```diff
@@ -0,0 +1,182 @@
...
+ */
+
+namespace Automattic\Jetpack\Forms;
+
+/**
+ * Class Post_To_Url_Hook
+ *
+ * Hooks on Jetpack's Contact form to post form data to some URL.
+ */
+class Post_To_Url_Hook {
```
ice9js: Not sure about the name of this class quite yet - it seems the Salesforce connection is currently hardcoded to let's name it accordingly.

Either way I think it'd make more sense inside the `service/` subdirectory - where the google drive class is located as well. Make sure to update the namespace as well when you move it.

ice9js (commented):
I left some initial comments on code organization, still need to test this.

projects/packages/forms/src/class-post-to-url-hook.php
CGastrell: > Not sure about the name of this class quite yet - it seems the Salesforce connection is currently hardcoded to let's name it accordingly.

Well, actually that's the big change here: it should be able to handle any form with the proper configuration to send submissions to a remote URL. Salesforce is just one configuration and here is being adapted to this new, more generic syntax/structure. Agree on the naming though.

> Either way I think it'd make more sense inside the service/ subdirectory - where the google drive class is located as well. Make sure to update the namespace as well when you move it.

Makes sense.

projects/packages/forms/src/class-jetpack-forms.php
CGastrell: Right! I was installing the package with no success, then I realized I needed to install the plugin for the autoloader to update. Thanks!

ice9js (dismissed):
Looks good, thanks Christian!

ice9js (approved):

Issue #30108, opened 2023-04-14
My Jetpack: change video-plays stats endpoint on WPCOM to skip Stats module requirement


lhkowalski:
We are using a WPCOM `v1.1` `/sites/$site/stats/video-plays` endpoint to fetch the stats related to VideoPress, but it has a check for the presence of the `Stats` module on the site being requested that is causing some troubles.

We tried to embed and enable `Stats` module inside the VideoPress plugin to pass the check (#29668), but it caused problematic side-effects (#29960) and we reverted it (#29966).

To solve it in another way, we are proposing to bypass the check entirely. VideoPress does not depends on the Stats module to collect the stats related to video consumption, so not having it is not a blocker for the stats to be collected (more info on this here: p1HpG7-lHv-p2#comment-62164).

The class responsible for handling the endpoint is `WPCOM_JSON_API_V1_1_Stats_Video_Plays_Endpoint`. The `callback` method on this class handles the request processing and uses the `$this->validate_stats_user( $site_id );` method to check for permissions.

The `validate_stats_user` method does the check for the `Stats` module and already supports a way to short-circuit the check, using the second parameter `$check_stats_module` (that is `true` by default).

So what we need to do is change the `$this->validate_stats_user( $site_id );` call to `$this->validate_stats_user( $site_id, false );` and we can by-pass the check.

My proposal is to make this bypass an optional one. We add a `check_stats_module` `GET` parameter that is set to `no` when the requester wants to bypass the check and defaults to `yes`. This way, every existing application keeps working the same way.

renatoagds:
Implemented at: D108216-code

lhkowalski:
Added my approval there. Nice work!

renatoagds:
Done: r265288-wpcom

PR #30194, opened 2023-04-19
VideoPress: Move out from WPCOM_Stats class

### LINKED ISSUE #30109: VideoPress: remove Stats module dependency from featured stats endpoint
We need to revisit the code for the `videopress/v1/stats/featured ` endpoint, created by #29785, to remove the dependency of it on the `Stats` module.

It needs to request the data directly to the WPCOM `v1.1` `/sites/$site/stats/video-plays` endpoint without relying on the WPCOM_Stats class. It needs to include on the request the `GET` parameter that will bypass the `Stats` module checking. The parameter will be added by #30108.

Depends on #30108.

renatoagds:
Closes #30109 

## Proposed changes:

* Remove usage of `WPCOM_Stats` class.
* Request directly to `WPCOM` using `check_stats_module` parameter.

### Other information:

## Jetpack product discussion

p1HpG7-lHv-p2#comment-62164

N/A

## Testing instructions:

* Spawn a JN site or run it locally
* Navigate to the `My Jetpack` page (or another page).
* Disable `Jetpack Stats` module (`wp-admin/admin.php?page=jetpack_modules`)
* ~Sandbox `public-api` using this patch: D108216-code~ - Patch Deployed #30108 
* Run: 

```
wp.apiFetch({ path: "/videopress/v1/stats/featured" })
  .then((data) => console.log(data));
```

* It should return stats

_If you stop to sandbox, it should works too, but returning an error message._

### DIFF
--- a/projects/packages/videopress/changelog/update-videopress-stats-endpoint
+++ b/projects/packages/videopress/changelog/update-videopress-stats-endpoint
@@ -1,0 +1,4 @@
+Significance: patch
+Type: changed
+
+VideoPress: Move out from Stats class
--- a/projects/packages/videopress/src/class-stats.php
+++ b/projects/packages/videopress/src/class-stats.php
@@ -8,13 +8,52 @@
 namespace Automattic\Jetpack\VideoPress;
 
 use Automattic\Jetpack\Connection\Client;
-use Automattic\Jetpack\Stats\WPCOM_Stats;
 use WP_Error;
 
 /**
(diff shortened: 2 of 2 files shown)


retrofox (approved):
LGTM. Confirmed the `https://fox.jurassic.tube/wp-json/videopress/v1/stats/featured` responses when the Jetpack Stats module is inactive. Even when the Jetpack plugin is not activated, too.

Nice one! 🚀

PR #30060, opened 2023-04-12
3rd Party: improve loading

kraftbj:
## Proposed changes:

* Updating the loading system to only include things that need to be loaded instead of loading everything with an over-reliance on hooks.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

p1HpG7-lOi-p2

## Does this pull request change what data or activity we track or use?

No.

## Testing instructions:

* Needs performance profiling.
* bbPress: sharing links should appear under bbPress items (discussion threads, etc)
* Bitly: ensure that Bitly OG tags are removed, but I'm pretty sure that plugin is no longer available. 
* Creative Mail, Jetpack Backup, Jetpack Boost: ensure that JITMs to install these plugins work.
* On WoA, need to check it out on a Atomic dev site where we can edit the `jetpack` plugin. (`jetpack-dev` still should return as as a development site).
* ...

PR #30195, opened 2023-04-19
jetpack: Load `Jetpack_AMP_Support` from 'amp_init'

anomiex:
## Proposed changes:

When #30060 moved `Jetpack_AMP_Support` to the autoloader, that slowed things down on wpcom. There they were still loading the AMP plugin unconditionally but were using 'amp_is_enabled' to prevent it from initializing, and instead of loading `Jetpack_AMP_Support` unconditionally they were using 'amp_init' to only load it if AMP actually initialized.

We may as well take the same approach. But that requires we add `class_exists` checks to a few code paths that were previously lacking it (apparently these aren't currently used on wpcom or someone would have noticed there).

### Other information:

## Jetpack product discussion

p1681891005968299-slack-CBG1CP4EN

No

## Testing instructions:

* On a site without the AMP plugin, make sure everything still works (i.e. no fatals from trying to call `Jetpack_AMP_Support` when it doesn't exist).
* On a site with the AMP plugin active, verify that Jetpack's AMP-related features still work.
* On a site with the AMP plugin loaded but deactivated via the 'amp_is_enabled' filter, verify that the `3rd-party/amp.php` file is not loaded and everything still works.
* Load this onto a wpcom sandbox and check that time used by `cd ~/public_html/bin/tests/isolated/; phpunit --testsuite=post-1` does not significantly increase versus 12.1-a.7 (which is what is currently in wpcom trunk).


samiff (approved):
Tested well for me 👍

PR #30197, opened 2023-04-19
My Jetpack: Add stats to VideoPress card

### LINKED ISSUE #29702: My Jetpack: Implement rich VideoPress card
We need to implement the VideoPress rich card using the enhanced component from #29787.

The rich card will show the number of plays in the last 7 days, and the variation when compared to 7 days before it:

Image

It will be shown only when the VideoPress product is active and the metrics to render it are available. A selector for the product stats was added by #29764. The format of the data available for the stats is described on that PR.

dhasilva:
Fixes #29702 

## Proposed changes:

* Adds the necessary presentational components for many rich cards
* Applies it to the VideoPress card, adding stats

### Other information:

No.

## Testing instructions:

* Install VideoPress
* Go to My Jetpack
* Check that the stats are loaded in the card

Loading | Positive | No change | Negative | No available stats
-|-|-|-|-
 |  |  |  |


renatoagds:
@dhasilva Maybe is good to wait for #30194 land? This way, we can test it working even without the Stats module.

renatoagds:
Merged #30194 ... Can you rebase here @dhasilva ?

dhasilva:
> Merged #30194 ... Can you rebase here @dhasilva ?

Sure. Just waiting for the first build to finish so I can test it as is on JN first.

lhkowalski:
Tested the branch, can see the card with the stats.

Zero:

Screen Shot 2023-04-20 at 18 16 23

Some views:

Screen Shot 2023-04-20 at 18 18 39

In this case the percentage is not correct, but it would actually be `Infinity` so probably we should show a big number instead (like 999+ %, for example).

The loading state is working nicely:

https://user-images.githubusercontent.com/6760046/233492979-a5419d7b-7412-46de-86dd-255162e0bc16.mov

I also faked some data on the backend to get other results.

Positive one:

Screen Shot 2023-04-20 at 18 37 57

Positive with big difference:

Screen Shot 2023-04-20 at 18 38 53

Negative one:

Screen Shot 2023-04-20 at 18 39 55

Negative with big difference:

Screen Shot 2023-04-20 at 18 41 01

From 1 to 100K:

Screen Shot 2023-04-20 at 18 42 59

From 100K to 1:

Screen Shot 2023-04-20 at 18 43 49

---

The card is behaving nicely on the big differences and big numbers.

---

What we may want to do is make all the cards with the same height. Makes sense?

---

Going to take a look in the code now.

lhkowalski:
Code change is neat. Found the line where you decide to return `0` as the percentage when the previous value is `0`. As a suggestion, we could return `Infinity` and then handle it when formating the number to display.

renatoagds (approved):

PR #30198, opened 2023-04-19
[CRM] Add API v4 base logic

kallehauge:
## Summary

The aim of this PR is to create a new place for our API v4 to exist, but also implement base functionality we need to start developing the same way like documentation, base controllers and test resources.

We try to achieve this by implementing a simple version of `[GET] /contacts/:id` so we have something concrete to compare to.

Fixes https://github.com/Automattic/zero-bs-crm/issues/3066

## Proposed changes:

**Feature(s)**

High level then we're adding a `[GET] /contacts/:id` endpoint since it's one of the simplest endpoints we can implement and it will serve as an example and proof of concept that our base logic is functioning.

* Add a new feature flag for experimental API v4 endpoints that we want to merge into trunk but don't want to officially ship yet (`jetpack_crm_feature_flag_api_v4`).
* Add API v4 documentation to explain the principles we follow.
* Add `REST_Base_Controller` class: Currently it "just" defines the namespace for our endpoints, but will probably grow in the future when we start implementing more endpoints and we identify generic logic.
* Add `REST_Base_Objects_Controller`: This class will serve as the base class for any DAL object endpoints since we'll have a lot of generic logic that can be applied across objects.
  * E.g.: Fetch the DAL service, but also convert object structure/models to endpoint args when we start to create create/update endpoints.
* Add `REST_Contacts_Controller`: An example of how we're going to work with API v4 moving forward and serves as the example that our API namespace + test logic works as expected.

**Tests**

High level then we're adding a REST API test suite that has the base functionality/bootrapping sorted out to be able to test future CRM API v4 endpoints:

* Add `REST_Base_Test_Case` that our endpoints can extend in order to test CRM endpoints.
* Add `REST_Authentication_Test` to provide easy ways to ensure that all of our custom endpoints fail authentication if we do not have a valid current_user().
  * To make it easier, then it should be enough to just provide new URLs to the data provider(s).
* Add `REST_Contacts_Controller_Test` to verify that our `[GET] /contacts/:id` endpoint works as expected.

### Other information:

## Jetpack product discussion

We don't have any direct discussion for implementing an endpoint that returns a contact, but it's prep work for the upcoming automations refactor: pbhBOz-3ss-p2

No.

## Automated testing

Verify that the new `rest-api` test suite doesn't throw any errors.

This can either be done by going through the CI steps on this PR or by following the below instructions on your local environment.

* Checkout branch locally
* Run: `jetpack install plugins/crm`
* Go to the CRM directory: `projects/plugins/crm`
* Run `composer run-script phpunit`

## Manual testing

* Add snippet below to your local environment (e.g. in `0-sandbox.php`).
* Update the `$contact_id` variable to a contact ID that exists in your local environment.
* Login to the website (the endpoint requires a logged in administrator user).
* Go to any front-end page and either:
  * Inspect the Network for the XHR request.
  * Inspect console since we `console.log()` the response in the snippet.
* Verify that we return the contact that corresponds to the ID.

```php
// Enable the feature flag that initialise the in progress code.

// Output a JS script to `wp_footer` that includes a nonce value so the WP REST registers you as a logged in user.
add_action( 'wp_footer', function () {
	$nonce = wp_create_nonce( 'wp_rest' );
	$contact_id = 1;

	$code = "
fetch( '/wp-json/jetpack-crm/v4/contacts/$contact_id', { headers: { 'X-WP-Nonce': '$nonce' } } ).then( function ( response ) {
	return response.json();
} ).then( function( data ) {
	console.log( data );
} ).catch( function ( err ) {
	console.log( err );
} );";

	printf( '<script>%s</script>', $code );
} );

```


projects/plugins/crm/src/rest-api/v4/class-base.php
```diff
@@ -0,0 +1,197 @@
...
+use Automattic\Jetpack_CRM\REST_API\Util\Authentication;
+use Automattic\Jetpack_CRM\REST_API\Util\Route_Scope;
+use WP_Error;
+
+/**
+ * Jetpack Partner API REST base class.
+ *
+ * @package Automattic\Jetpack_CRM\REST_API\V4
+ */
+abstract class Base {
```
cleacos: The `Base` name sounds quite generic. Maybe `API_Base` or `API_Controller` are more specific names.

projects/plugins/crm/src/rest-api/v4/class-base.php
```diff
@@ -0,0 +1,197 @@
...
+ */
+
+namespace Automattic\Jetpack_CRM\REST_API\V4;
+
+use Automattic\Jetpack_CRM\REST_API\Util\Authentication;
+use Automattic\Jetpack_CRM\REST_API\Util\Route_Scope;
+use WP_Error;
+
+/**
+ * Jetpack Partner API REST base class.
```
cleacos: Don't forget to change this

projects/plugins/crm/src/rest-api/exceptions/class-base-exception.php
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * @package Automattic\Jetpack_CRM\REST_API
+ */
+
+namespace Automattic\Jetpack_CRM\REST_API\Exception;
+
+defined( 'ABSPATH' ) || exit;
+
+use Exception;
+use Throwable;
```
cleacos: Cool, I usually use the backslash `\Exception` to refer to global classes.

cleacos:
Other question, do we need all that authentication stuff? It seems that we can rely on the WP API auth system. For our case (React app using the nonce mechanism, it'd be enough), we don't need to generate/manage API keys for third parties, at least for now, like for the Jetpack partners' connections.

projects/plugins/crm/src/rest-api/exceptions/class-base-exception.php
kallehauge: I'm actually kinda mixed 😅 
I've started to generally add all "dependencies" at the top of the file so it fits the React import pattern, but I'd be up for whatever the team wants to do as our standard (WordPress core doesn't use namespaces, so they don't have anything we can lean up against)

projects/plugins/crm/src/rest-api/v4/class-base.php
kallehauge: Agreed - although we use namespaces, then it still didn't provide a lot of context in the scope itself 😄 

The updated code uses:

* `REST_Base_Controller`: A base controller that - for now - just contains the namespace for our APIs
  * I'm thinking e.g. automations will extend this one which is why it's been separated out from the one below
* `REST_Base_Objects_Controller`: A base objects controller for DAL specific resource endpoints that extends the base class

cleacos: > although we use namespaces, then it still didn't provide a lot of context in the scope itself

Yes, sure, but my """complain""" is about the naming, `Base` is quite a neutral word (adjective). The question I had was, but a `base` of what? We could have many types of bases in this context: An API controller base, an endpoint base, even an exception base...

projects/plugins/crm/src/rest-api/exceptions/class-base-exception.php
cleacos: I don't mind. I'm easy here, without any strong opinion here. The `use` clause doesn't add relevant information to me as a developer, only increases the number of lines 😆. I usually use `use` for those long namespaces and use only the class name, and global classes don't have a namespace so I don't feel the need to add them to the `use` clauses.

kallehauge: In theory it should make the code more DRY since you're able to replace it if it's using the same interface. E.g.:

If we start out using the native `\Exception` class but chose to extend it with our own `class JPCRM_Exception extends \Exception {}` then we'd only have to replace the use statement at the top of the file instead of the places all uses in the code itself, so it becomes more of an actual "import" (like in JS) versus just being a quality of life improvement not having to write the full namespace every time we want to use classes 😛

kallehauge:
> Other question, do we need all that authentication stuff?

Nope! Most of the PR has been rewritten since what I initially based it on was too "Automattic specific".

projects/plugins/crm/tests/php/bootstrap.php
```diff
@@ -14,3 +14,17 @@
...
+add_filter( 'jetpack_crm_feature_flag_api_v4', '__return_true' );
+
+/**
+ * Load Jetpack CRM.
+ *
+ * Not all code is automatically loaded any we depend on a lot of global
+ * variables, so the easiest path forward (for now at least) is to just
+ * load the core plugin file so everything is initiated.
+ */
+require_once __DIR__ . '/../../ZeroBSCRM.php';
```
tbradsha, line 30: Note that this won't always work depending on the setup; we should avoid hard-coding path traversal.

tbradsha (commented):
I haven't looked at the actual code yet other than skimming it. I tried to test it, but got a 500 error:

`PHP Fatal error:  Uncaught Error: Class 'Automattic\Jetpack_CRM\REST_API\V4\REST_Contacts_Controller' not found in /includes/ZeroBSCRM.REST.php:109`

kallehauge:
> I haven't looked at the actual code yet other than skimming it. I tried to test it, but got a 500 error

It could sound like you have to run `jetpack install plugins/crm` after checking out the branch so the new classes are added to [the Jetpack autoloader](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/crm/ZeroBSCRM.php#L38-L84) classmap 😄 

**Context:**

Copy+🍝  from the Jetpack Autoloader:

> This is a custom autoloader generator that uses a classmap to always load the latest version of a class.
>
>The problem this autoloader is trying to solve is conflicts that arise when two or more plugins use the same package, but one of the plugins uses an older version of said package.
>
>This is solved by keeping an in memory map of all the different classes that can be loaded, and updating the map with the path to the latest version of the package for the autoloader to find when we instantiate the class.
> 
> [Source](https://github.com/Automattic/jetpack/tree/trunk/projects/packages/autoloader)

TL;DR: We create a map of classes to try and prevent internal package conflicts. The map is generated when running the `jetpack install` command (and maybe the `jetpack build` command as well?)

tbradsha:
Great! I was able to confirm the function of the snippet. As @cleacos is in context, I'll let him review this further.

cleacos:
I'll take another look at the code soon, and I'll test it.

projects/plugins/crm/includes/ZeroBSCRM.Permissions.php
```diff
@@ -692,13 +692,49 @@ function zeroBSCRM_permsObjType($objTypeID=-1){
...
+ */
+function zeroBSCRM_permsCustomers() {
+	$current_user = wp_get_current_user();
+
+	if ( ! $current_user instanceof WP_User ) {
+		return false;
 	}
 
+	return jpcrm_can_user_manage_contacts( $current_user ) === true;
+}
```
coder-karen:
```suggestion
 * @deprecated $$next-version$$ Use jpcrm_can_user_manage_contacts()
 *
 * @return bool
 *
 * @phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
 */
function zeroBSCRM_permsCustomers() {
	_deprecated_function( __FUNCTION__, 'crm-$$next-version$$' );
	$current_user = wp_get_current_user();

	if ( ! $current_user instanceof WP_User ) {
		return false;
	}

	return jpcrm_can_user_manage_contacts( $current_user ) === true;
}
```

coder-karen: Just a suggestion, no harm in adding this in just now given the deprecation suggestions :)

coder-karen (commented):
Also tested the snippet and works well (left a very minor suggestion).

For me it's good to just read through this and see how the API is being set up.

projects/plugins/crm/includes/ZeroBSCRM.Permissions.php
tbradsha: If we do deprecate this function, at minimum let's update all calls to said function in core and make an issue in the extensions repo to update the various extensions that use said call.

That said, do we really need to introduce another helper function + filter? Why not just use the native `has_cap()` WP function directly?

cleacos: > That said, do we really need to introduce another helper function + filter? Why not just use the native has_cap() WP function directly?

I was wondering the same... I was trying to see a use case for allowing third parties to change the user permissions.

kallehauge: > do we really need to introduce another helper function + filter? Why not just use the native has_cap() WP function directly?

**Filter**

This is just my own opinion about plugin development, but In general I tend to lean towards making our systems "Integrable" even in situations where we don't currently have a need for the hook(s) ourselves.

It's kind of a judgement call when to add hooks if that's the case, but to give this one some specific context, then we'd be able to create an extension (e.g.: "Granular Team and User Permissions") by extending the filter - or someone else can come up with other types of ideas. If we never open up to these extensions/modifications, we never really allow the community to get involved in "the extension market".

_Note to self: if we disagree about this PoV, [then the API docs need to be updated as well](https://github.com/Automattic/jetpack/pull/30198/files#diff-cf70cf276da969c5b82c375e193eb07baf8b07f7cb125c35a0824ef905b2c66bR23)_ 📔 

**Function**

If I remember correctly, then we have ~50 instances of `zeroBSCRM_permsCustomers` in core right now so it's definitely something we're repeating a lot which makes me immediately think it's supposed to follow the DRY principle and therefor make it easier to maintain (e.g.: especially if we keep the filter around - otherwise it would be a potential nightmare to update it in the future 😄)

kallehauge: >If we do deprecate this function, at minimum let's update all calls to said function in core and make an issue in the extensions repo to update the various extensions that use said call.

I'm leaning towards this as well, but I'd prefer to keep it as a separate issue for two reasons:

* A separate PR makes sense since the old function is used in ~50 places in the code and it will take away focus from the main objective here.
* If we add `_deprecated_function( __FUNCTION__, 'crm-$$next-version$$' );` in this PR, then we'd be flooding our users error logs with warning since it's used in so many spaces, so the condition would (IMO) belong in the separate PR in this specific instance 🗒️
  * If it was a smaller change to replace all the old uses, then I wouldn't mind adding `_deprecated_function` in this PR just to reduce overhead.

 (I'll create an issue depending on the feedback about keeping the new permission check function)

projects/plugins/crm/tests/php/bootstrap.php
kallehauge: >we should avoid hard-coding path traversal

That's my default approach when working inside the plugin as well, but we can't use e.g. `ZBS_PLUGIN_DIR` since the file that defines it hasn't been required at this point.
AFAIK then we can't rely on other types of e.g. WP constants like `WP_PLUGIN_DIR` or `ABSPATH` as well because that would assume we know where the plugin is located - although they should exist because we run `\WorDBless\Load::load()`.

Because the tests lives "inside" the CRM directory, then AFAIK this is probably the most robust way to bootstrap it; like when we require the autoloader:

```php
require_once __DIR__ . '/../../vendor/autoload.php';
``` 

>Note that this won't always work depending on the setup

What specific setup(s) do you have in mind where this would fail?

projects/plugins/crm/includes/ZeroBSCRM.Permissions.php
gogdzl: > A separate PR makes sense since the old function is used in ~50 places in the code and it will take away focus from the main objective here.

I Agree.

projects/plugins/crm/tests/php/rest-api/v4/class-rest-authentication-test.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+namespace Automattic\Jetpack_CRM\Tests;
```
gogdzl:
```suggestion
namespace Automattic\Jetpack\CRM\Tests;
```

projects/plugins/crm/tests/php/rest-api/v4/class-rest-contacts-controller-test.php
```diff
@@ -0,0 +1,111 @@
+<?php
+
+namespace Automattic\Jetpack_CRM\Tests;
```
gogdzl:
```suggestion
namespace Automattic\Jetpack\CRM\Tests;
```

gogdzl (changes requested):
I like it!
It is simple, it leaves room for plugins to extend it.

I left just two inline suggestions, but there are some other ` Automattic\Jetpack_CRM` leftovers that could be addressed in this PR - some in the package names and some in docblocks.

Other than that, great job!

projects/plugins/crm/tests/php/rest-api/v4/class-rest-authentication-test.php
kallehauge: Awesome, thank you for catching these @gogdzl 🙇 

I found a few DocBlock instances as well, but all should be fixed since https://github.com/Automattic/jetpack/pull/30198/commits/466f2b6285f71f30b2a592badad23aecbc5e4099

projects/plugins/crm/tests/php/rest-api/v4/class-rest-contacts-controller-test.php
kallehauge: Awesome, thank you for catching these @gogdzl 🙇 

I found a few DocBlock instances as well, but all should be fixed since https://github.com/Automattic/jetpack/pull/30198/commits/466f2b6285f71f30b2a592badad23aecbc5e4099

projects/plugins/crm/tests/php/rest-api/v4/class-dummy-test.php
kallehauge, line 1: We no longer need a dummy test since this PR introduces actual integration test(s).

(Read: this was just a placeholder for the PR that implemented PHPUnit and a few other dependencies for this PR).

cleacos:
This PR could be merged after the JPCRMv6 release.

kallehauge:
> This PR could be merged after the JPCRMv6 release.

The functionality is technically hidden behind a feature flag, so it shouldn't do much difference, but I don't mind waiting 😄

cleacos:
> but I don't mind waiting 😄

Okay, then wait! 😄

> The functionality is technically hidden behind a feature flag

I know it, but including it in the 6.0 release will not add any value for our users... for now. Correct me if I'm wrong. It'd be nice to ship it in 6.1 with the first API endpoints implementation.

gogdzl (approved):
Jetpack CRM v6.0.0 has already been shipped and the code looks good.
Merge it!

PR #30213, opened 2023-04-20
Dynamically remove and append rows in the forms dashboard

ice9js:
## Proposed changes:

This PR updates the forms dashboard to dynamically remove and append rows when using bulk actions, instead of fully refreshing the list every time.

### Other information:

No.

## Testing instructions:

- Go to `admin.php?page=jetpack-forms`.
- You should notice the new loading placeholders on the table.
- Select some responses and perform a bulk action.
- The selected rows should disappear immediately while the rest remains unaffected.
- You should see a loading placeholder appear at the bottom of the list and it should eventually append any new rows that now fit on the page if applicable.


digitalwaveride:
😻 Nice Kuba!

Works for me as advertised, when using the bulk options, noticed though when using the "dropdown" via "...", then it seems to do a full page reload ... not sure if that is expected.

https://user-images.githubusercontent.com/5120357/233386913-c2814312-5c7e-4902-aba0-a9ea8a716c6f.mov

jcheringer:
I like the new animation for the full responses loading.
But I found it a bit confusing adding the one-line loading at the bottom while processing the bulk request.
Wouldn't it be possible to put only the lines being processed in a loading (or disabled) state while we execute the bulk action and load the updated results, and, once the requests are complete, we replace the response list?

ice9js:
>Wouldn't it be possible to put only the lines being processed in a loading (or disabled) state while we execute the bulk action and load the updated results, and, once the requests are complete, we replace the response list?

The placeholders are at the bottom because when the list is reloaded, the moved rows aren't replaced, they're gone. Instead there's now a few more rows that fit onto this page that otherwise would be on the next one. I think this part makes total sense.

But you're right that it can feel a little confusing, but I'd attribute that to the rows being moved or deleted being gone instantly. Ideally we'd animate that too so it's more clear for the user what is going on. But that's a little more complicated so while we may deem it worthy to put time into at some point, it's for another PR.

jcheringer:
> The placeholders are at the bottom because when the list is reloaded, the moved rows aren't replaced, they're gone. Instead there's now a few more rows that fit onto this page that otherwise would be on the next one. I think this part makes total sense.

I see your point. Maybe I got that feeling because my response list is too short, and I don't even get a second page.
Because of that, I just see a placeholder at the bottom which is replaced by nothing once the action is done.

I've played a bit with the code, and the following video is more or less what I had in mind:

[Screen Recording on 2023-04-20 at 15_24_34.webm](https://user-images.githubusercontent.com/7811225/233454766-9c5ac32a-2dac-4e9a-8a2a-a3fd04fb11ae.webm)

ice9js:
>Maybe I got that feeling because my response list is too short, and I don't even get a second page.

Yeah... technically, we could detect that because we have the totals. And in that case we wouldn't show any loading placeholders at all.

I still don't think we should be showing placeholders for the things you're moving. One of the points of this change is that as far as the user is concerned, their action should be instant and they're just waiting for the list to update following their action, not on the action itself.

CGastrell (dismissed):
LGTM! I think there are things that still need fine tuning (removed rows animation, kebab menu causing reload). We can address those in upcoming PRs or keep pushing to this one.

ice9js:
I've added some changes:

- I updated the `<Table>` component so rows can be animated in and out when necessary, but in a way where it's still up to the implementation to add the actual animation rather than forcing it everywhere where it appears.
- I added the animation for removing rows from the table.
- I've updated tab totals to also dynamically recalculate whenever things are moved from one tab to the other.
- We show 10 placeholders initially. Once the tab totals are known and kept track of, the number of loading placeholders will always exactly match the results. Magic.
- Tabs are now disabled while results are loading as it led to a bunch of race conditions. The only way to prevent that is through moving the 'query state' into the URL, which we definitely should do, just not in this PR.

There's just one annoying thing I haven't yet found the cure for. Because we have alternating row backgrounds, when you delete an uneven amount of rows the backgrounds will swap slightly deteriorating the experience.

jcheringer (approved):
This is looking great! Nice job! 🚀

CGastrell:
Animation looks good, but kebab dropdown button stopped working for me. Does it work for you? Unrelated?

ice9js:
> Animation looks good, but kebab dropdown button stopped working for me. Does it work for you? Unrelated?

Yeah, that seems unrelated. I can see the same issues happening on `trunk`.

PR #30215, opened 2023-04-20
My Jetpack: Activate product module after installing the standalone plugin

### LINKED ISSUE #30188: My Jetpack: product install-standalone endpoint is returning stale status for the product
After we finished some changes on the frontend, we realised that the status returned for the product after the installation is stale, and the updated state can be fetched on a new request to the product endpoint.

**How to reproduce**

- Spin up a fresh JN site
- Connect the site
- Go to `Jetpack > My Jetpack` and open the browser console
- Run the following command:

```
wp.apiFetch({
  path: "/my-jetpack/v1/site/products/videopress/install-standalone",
  method: 'POST',
})
.then((data) => console.log(data));
```

- Notice that the product representation returned still shows the status as inactive, and it should be active by this time, if the request succeeds.

lhkowalski:
Fixes #30188.

## Proposed changes:

* Changed the installation method to call `Modules` `activate` function so the module for the product gets enabled as well
* Avoid activating the module for products that does not have the required plan to be activated
   * Search is the example, it requires a plan even for free usage (that also explain why the product stays inactive even when the plugin is installed and activated, and the action button stays as "Start for free")

### Other information:

## Testing instructions:

* Spin up a new JN site with only Jetpack
* Go to `Jetpack > My Jetpack` and open up the browser console
* Run the following command:

```
wp.data.dispatch( 'my-jetpack' ).installStandalonePluginForProduct( 'videopress' ).then(result => console.log(result))
```

* This will trigger the standalone plugin installation for VideoPress
* Notice that the VideoPress product card will enter a loading state
* After the loading finishes, the card will then change it's status to active, reflecting the correct status returned by the endpoint:

Screen Shot 2023-04-20 at 15 26 42

Screen Shot 2023-04-20 at 15 29 58

* You can test this for all the hybrid products: `videopress`, `social`, `backup` and `search`
* For some products, the next status after installing the plugin may not be `active` because they may require a plan to work
* For `Search`, for example, the status stays as `needs_purchase_or_free` resulting on a "Start for free" call-to-action button, but the plugin gets installed and activated correctly:

Screen Shot 2023-04-20 at 15 33 26

Screen Shot 2023-04-20 at 15 34 06

* The menu options under `Jetpack` also get enabled, as expected (you may need to refresh the page to see it):

Screen Shot 2023-04-20 at 15 36 34


lhkowalski:
I thought the change I did would fix the issue for all hybrid products, but seems like it's not working for Search, so I will dig to fix it and update the PR.

lhkowalski:
> I thought the change I did would fix the issue for all hybrid products, but seems like it's not working for Search, so I will dig to fix it and update the PR.

Ok, now I have a fix that works for every hybrid product.

dhasilva (approved):
LGTM :rocket:

PR #30221, opened 2023-04-20
My Jetpack: Reload page after standalone install/activation

### LINKED ISSUE #30220: My Jetpack: Reload page after standalone activation or install
After installing or activating the standalone plugin through the menu option, we should reload the page to make the new jetpack menu item visible.

This way, we provide a better experience for the user since the menu is the essential link to the standalone plugin's main page.

renatoagds:
Closes #30220 

## Proposed changes:

* Add `handleInstallStandalone` who calls `window.location.reload` after success.

### Other information:

## Jetpack product discussion

N/A

N/A

## Testing instructions:

* Navigate to My Jetpack page.
* Make sure that you have Jetpack core plugin installed and VideoPress standalone not installed or active.
* Open VideoPress card menu and install/activate plugin.
* The page should be reload after success.

### DIFF
--- a/projects/packages/my-jetpack/_inc/components/connected-product-card/index.jsx
+++ b/projects/packages/my-jetpack/_inc/components/connected-product-card/index.jsx
@@ -41,6 +41,12 @@ const ConnectedProductCard = ( { admin, slug, children, showMenu = false } ) => 
 		navigateToConnectionPage,
 	] );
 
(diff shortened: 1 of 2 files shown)


dhasilva (approved):
Working as intended.
There is a small gap between the time the `manage` button is re-enabled and the page refresh where the user could press it and be redirected to the wrong page. Let's leave it for a follow-up later.

PR #30225, opened 2023-04-20
Update README.md to point to jetpack-autoloader v2.x

nextgenthemes:
Version 1 has a unsolvable deps issue in my case.

Is this thing ready for public use in plugins in general outside of Jetpack? The idea seems pretty cool. Seems version 2 is the one that should be used.

### DIFF
--- a/projects/packages/autoloader/README.md
+++ b/projects/packages/autoloader/README.md
@@ -21,7 +21,7 @@ In your project's `composer.json`, add the following lines:
 ```json
 {
     "require-dev": {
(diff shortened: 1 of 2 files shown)


anomiex (dismissed):
Good catch, thanks!

anomiex (approved):

nextgenthemes:
@anomiex Yeah thanks for ignoring my question.

anomiex:
Sorry, I didn't see the question. Yes, it should be usable in other plugins.

PR #30232, opened 2023-04-20
actions: Improve build and post-build workflows on forks

anomiex:
## Proposed changes:

The "Trigger TeamCity build" cannot run on a fork as it needs secrets, so we'll have to skip it.

The post-build workflow is fine to run, but it will think there are no plugins because the "Create artifact for Jetpack Beta plugin" job doesn't run and it looks for that job's artifact to decide if there were plugins. It doesn't actually _use_ it, it just looks for its presence. While we could download and read the zip file instead, it seems easier to just create a dummy artifact to serve as the signal instead.

### Other information:

## Jetpack product discussion

None, spotted the problem while trying to merge #30225.

No

## Testing instructions:

* Does the post-build workflow run correctly, i.e. it doesn't think "no plugins"?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -112,6 +112,15 @@ jobs:
           # Only need to retain for a day since the beta builder slurps it up to distribute.
           retention-days: 1
 
+      - name: Store plugins.tsv as artifact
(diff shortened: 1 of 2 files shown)


kraftbj (approved):
let's try it.

PR #30239, opened 2023-04-21
Anchor: Control CTA with a filter

andres-blanco:
## Proposed changes:

The idea in this PR is to control wether to show/hide the Convert to audio prompt with a filter that defaults to the scheduled sunsetting date.

| Before  | After |
| image  | image |

### Other information:

## Jetpack product discussion
paYKcK-2Rq-p2
p1682014164018689-slack-C0Q664T29

## Testing instructions:

* Follow instructions to apply this PR on [this comment](https://github.com/Automattic/jetpack/pull/30239#issuecomment-1517792531).
* Set up a new site.
* Go to https://anchor.fm/wordpressdotcom (verify you are logged out from Anchor).
* Create a new account and link it to the site that you just created.
* **NOTE: in a previous PR it was noted that sometimes it's not possible to create an Anchor account, sadly there's no way around it. If you are experiencing trouble with this ping me on slack and I will add you to an already connected site.**
* Sandbox the site, create a new post and publish it.
* Verify you see the `Convert to audio` prompt in the post publish menu.
* Change the filter value to false.
* Verify you don't see the `Convert to audio` prompt.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/anchor-fm/anchor-fm.php
+++ b/projects/plugins/jetpack/extensions/blocks/anchor-fm/anchor-fm.php
@@ -200,11 +200,13 @@ function process_anchor_params() {
 	wp_localize_script( 'jetpack-blocks-editor', 'Jetpack_AnchorFm', $data );
 }
 
+function is_anchor_enabled() {
+	return time() < strtotime( '2023-05-11 00:00:00 UTC' );
+}
+
 add_filter(
 	'jetpack_is_anchor_enabled',
-	function ( $is_enabled ) {
-		return time() < strtotime( '2023-05-11 00:00:00 UTC' );
-	}
+	__NAMESPACE__ . '\is_anchor_enabled'
 );
 
 add_action( 'init', __NAMESPACE__ . '\register_extension' );
--- a/projects/plugins/jetpack/extensions/blocks/anchor-fm/anchor-fm.php
+++ b/projects/plugins/jetpack/extensions/blocks/anchor-fm/anchor-fm.php
@@ -183,16 +183,16 @@ function process_anchor_params() {
 	 *
 	 * @since $$next-version$$
 	 *
-	 * @param bool $is_enabled Whether the prompt is enabled or not.
+	 * @param bool $is_anchor_enabled Whether the prompt is enabled or not.
 	 */
-	$is_enabled = apply_filters( 'jetpack_is_anchor_enabled', true );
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/anchor-fm/anchor-fm.php
```diff
@@ -177,17 +177,43 @@ function process_anchor_params() {
...
 	$blog_connected_to_anchor = function_exists( 'get_blog_option' ) && false !== get_blog_option( get_current_blog_id(), 'anchor_podcast' );
+
+	/**
+	 * Allows disabling the Anchor convert to audio prompt.
+	 *
+	 * @since $$next-version$$
+	 *
+	 * @param bool $is_enabled Whether the prompt is enabled or not.
+	 */
+	$is_enabled = apply_filters( 'jetpack_is_anchor_enabled', true );
```
daledupreez: Nit: It may be worth calling the local variable `$is_anchor_enabled` to be more self-descriptive, but definitely not blocking.

projects/plugins/jetpack/extensions/blocks/anchor-fm/anchor-fm.php
```diff
@@ -177,17 +177,43 @@ function process_anchor_params() {
...
 
+/**
+ * Function to check if the Anchor feature is enabled. This is used in the
+ * filter below.
+ *
+ * @return bool whether the feature is enabled or not.
+ */
+function is_anchor_enabled() {
+	return time() < strtotime( '2023-05-11 00:00:00 UTC' );
+}
```
daledupreez: The filter should check whether the value has been overridden already, and not just return its own computation, otherwise overrides need to specify later priority to override this filter.
```suggestion
function is_anchor_enabled( $is_anchor_enabled ) {
	if ( ! $is_anchor_enabled ) {
		return false;
	}

	return time() < strtotime( '2023-05-11 00:00:00 UTC' );
}
```

daledupreez (commented):
I have one request about making the filter be sensitive to existing overrides via other filters that I think we need to fix before this is ready to ship.

I also noticed during my testing that this flag is set _at the time that the post/editor is loaded_. That isn't a blocker, but I think we should be aware of that as we're preparing to communicate that to users.

daledupreez (dismissed):
The changes LGTM!

kraftbj (approved):
🚢

PR #30240, opened 2023-04-21
admin-menu: Use .getAttribute('aria-expanded') instead of .ariaExpanded

mreishus:
## Proposed changes:

* Just a minor compatibility fix for firefox:

It seems that `.ariaExpanded` doesn't work in firefox:

https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaExpanded

### Other information:

## Testing instructions:

* Visit wp-admin 
* Check `console.log( document.getElementById("collapse-button").ariaExpanded ); ` in both Chrome and FF
* Check `console.log( document.getElementById("collapse-button").getAttribute('aria-expanded' ) ); ` in both Chrome and FF
* If you have a full setup, you can expand/collapse the bar and watch the ajax requests going out

* "Expanded" should no longer be undefined in FF

### DIFF
--- a/projects/plugins/jetpack/changelog/update-admin-menu-collapse-button
+++ b/projects/plugins/jetpack/changelog/update-admin-menu-collapse-button
@@ -1,0 +1,4 @@
+Significance: patch
+Type: other
+
+Fix firefox not detecting the collapsed state of the sidebar properly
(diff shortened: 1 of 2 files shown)


cpapazoglou (approved):
Thanks for tracking and fixing this Matthew!

PR #30251, opened 2023-04-21
[RNMobile] Update embed loading screen

### LINKED ISSUE #5654: Loading state UI
This loading state appears when the Player block is mounted and before the embed html is available

When uploading a new video it appears after the media upload period and before the VideoPress embed had time to load.

Image

jhnstn:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/5654

## Proposed changes:

* Updates the styling of the embed loading view 
* Renders the loading screen while the preview is being prepared

### Screens

**Still**

**Recording**

https://user-images.githubusercontent.com/744755/234397494-8547d636-60b1-4368-a25a-1c0681374d1e.mov

### Other information:

## Testing instructions:

* Add a VideoPress block and upload a video with a tall aspect ratio
* Verify that the loading screen appears the same as the embed player "processing" screen
* Wait for the video to be processed
* Notice the loading screen is rendered again at the default height ( ~200px)
* Notice the player height expand when the video loads

Note: VideoPress sometimes responds with the correct height for the video preview. In these cases the loading screen is not displayed after the video is processed.


projects/packages/videopress/src/client/block-editor/blocks/video/components/player/style.native.scss
```diff
@@ -10,3 +10,21 @@
 	width: 100%;
 	z-index: 1;
 }
+
+.videopress-player__loading {
```
jhnstn, line 14: There is no difference between dark and light modes for the embed processing screen. I opted to keep the loading view as close to the embed processing view so I didn't add light/dark versions.

projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
```diff
@@ -104,26 +109,37 @@ export default function Player( { isSelected, attributes } ) {
 		}
 	}, [] );
 
-	const loadingStyle = {};
-	if ( ! isPreviewReady( preview ) ) {
-		loadingStyle.height = 250;
+	const containerLoadingStyle = {};
+	if ( ! isPreviewAvailable( preview ) ) {
+		containerLoadingStyle.height = loadingStyle.height;
+		containerLoadingStyle.backgroundColor = loadingStyle.backgroundColor;
```
jhnstn: There is a flicker between the initial loading screen and when the embed html is rendered. I'm not sure how prevalent this is on a production build but using the loading background reduces the flicker.

projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
```diff
@@ -79,13 +77,20 @@ export default function Player( { isSelected, attributes } ) {
 	const { invalidateResolution } = useDispatch( coreStore );
 	const invalidatePreview = () => invalidateResolution( 'getEmbedPreview', [ videoPressUrl ] );
 
+	const iconStyle = style[ 'videopress-player__loading-icon' ];
+	const loadingStyle = style[ 'videopress-player__loading' ];
+
+	// Check if the preview is ready or we ran out of attempts.
+	const isPreviewAvailable = preview =>
```
fluiddot: We'd need to rename the function's argument as it conflicts with the `preview` attribute. However, I think we could turn this into a constant:
```
const isPreviewAvailable =
		!! preview?.height || previewCheckAttempts > VIDEO_PREVIEW_ATTEMPTS_LIMIT;
```

fluiddot (commented):
@jhnstn I tested uploading a portrait video and noticed about an extra space between the content and the caption. It's likely that we had it before but it wasn't as visible as it's now with the black background.

| Conversion | Player |
|--------|--------|
|  |  | 

Additionally, I also noticed that the video is cropped. Maybe it's related to [this comment](https://github.com/Automattic/jetpack/pull/30251#discussion_r1175170295) I shared.

_Tested on iOS simulator iPhone 14._

projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
jhnstn: ah yeah, I didn't notice any warnings or errors but you're right about the naming conflict. I remember having an issue using a constant but I've made that switch and it's working.

jhnstn:
@fluiddot 

I had to drop trying to set the background color on the `Player` component. The default embed aspect ratio is fixed and scales proportionally on different devices. What I had before looked fine on an iPhone 14 max pro simulator and on physical iPhone SE. I did however see the background bleed on an iPhone 14 simulator and on a physical pixel 6. 

I tried using a dynamic height to adjust for the device width but it was still difficult to match the loading screen with the embed iframe. I did end up keeping the dynamic height calculation to reduce the height jumping between loading and processing.

Without the background there is a white flicker between mounting the `Player` component and when the embed markup is ready. The flicker seems to be from the `WebView` loading it's DOM. I tried setting the background color on the embed iframe in the `Sandbox` but there wasn't any noticeable improvement. I'm wondering if the flicker will be as noticeable when not using a development build.  

I've updated the screen recording in the description to show the flicker on an iPhone 14 simulator.

projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
```diff
@@ -17,12 +17,13 @@ import getMediaToken from '../../../../../lib/get-media-token/index.native';
...
+import { VideoPressIcon } from '../icons';
 import style from './style.scss';
 
 const VIDEO_PREVIEW_ATTEMPTS_LIMIT = 10;
-
-// The preview is ready when it has a height property
-const isPreviewReady = preview => !! preview?.height;
+const DEFAULT_PLAYER_ASPECT = 380 / 600; // This is the observed default aspect ratio from VideoPress embeds.
+const LOADING_OFFSET_HEIGHT = 37;
+const MAX_WIDTH = 548;
```
jhnstn: I noticed this is the max with of the `Player` component on wide screens.

jhnstn:
> Additionally, I also noticed that the video is cropped. Maybe it's related to https://github.com/Automattic/jetpack/pull/30251#discussion_r1175170295 I shared.

I didn't notice the `preview` scope issue causing the video to be cropped. I _think_ that was still valid JS but just won't pass ESLint. Regardless I've fixed that specific issue.

I have noticed that every once in a while the embed response comes back with `width` and `height` properties but they are wrong and the video looks cropped. It's not very often and hard to reproduce since it's tied to the embed response. We might need to open an issue on VideoPress core if we continue to see that behavior.

fluiddot (commented):
@jhnstn as you mentioned in https://github.com/Automattic/jetpack/pull/30251#issuecomment-1522397351, I noticed that the preview flickers. This is also noticeable when opening a post with VideoPress blocks.

https://user-images.githubusercontent.com/14905380/234575690-0467c9fe-2853-4267-adb6-a85e0cdfb739.mp4

I'm wondering if the loading screen we are introducing could avoid it by being displayed as an overlay. I tested updating some parts of the component and seems feasible, sharing a patch file with the changes in case you'd like to take a look.

The changes in the below patch are:
- Use `aspectRatio` style attribute to control the dimensions of the container.
- Listen to `videopress_ready` event instead of `videopress_loading_state`. I noticed that the latter is only sent after the video is converted on iOS.
- Update styles and rendering logic to display the loading UI over the preview.

```patch
diff --git forkSrcPrefix/projects/packages/videopress/src/client/block-editor/blocks/video/components/player/style.native.scss forkDstPrefix/projects/packages/videopress/src/client/block-editor/blocks/video/components/player/style.native.scss
index b128bb6aadb460c6ba8526fd512a0606ad26e7ad..7e165e34dd226668c3dd60306fd05c8edbc6d8e6 100644
--- forkSrcPrefix/projects/packages/videopress/src/client/block-editor/blocks/video/components/player/style.native.scss
+++ forkDstPrefix/projects/packages/videopress/src/client/block-editor/blocks/video/components/player/style.native.scss
@@ -1,23 +1,20 @@
-.videopress-player {
+.videopress-player__container {
 	width: 100%;
 }
 
-.videopress-player__overlay {
-	position: absolute;
-	top: 0;
-	left: 0;
-	height: 100%;
-	width: 100%;
-	z-index: 1;
-}
-
 .videopress-player__loading {
-	display: flex;
 	justify-content: space-evenly;
 	align-items: center;
-	width: 100%;
 	background-color: #000000;
+
+	position: absolute;
+	top: 0;
+	left: 0;
+	right: 0;
+	bottom: 0;
+	z-index: 10;
 }
+
 .videopress-player__loading-icon,
 .videopress-player__loading-text {
 	color: $white;
diff --git forkSrcPrefix/projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js forkDstPrefix/projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
index 55e9c00392662c5065610a7db0a2e9403f4de1a1..55c2b9f4cb0e43a57ca25f49beed81e38eb8f682 100644
--- forkSrcPrefix/projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
+++ forkDstPrefix/projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
@@ -22,8 +22,6 @@ import style from './style.scss';
 
 const VIDEO_PREVIEW_ATTEMPTS_LIMIT = 10;
 const DEFAULT_PLAYER_ASPECT_RATIO = 380 / 600; // This is the observed default aspect ratio from VideoPress embeds.
-const LOADING_OFFSET_HEIGHT = 37;
-const MAX_WIDTH = 548;
 
 /**
  * VideoPlayer react component
@@ -33,7 +31,7 @@ const MAX_WIDTH = 548;
  * @param {boolean} props.isSelected - Whether the block is selected.
  * @returns {import('react').ReactElement} - React component.
  */
-export default function Player( { isSelected, attributes } ) {
+export default function Player( { attributes } ) {
 	const {
 		controls,
 		guid,
@@ -47,13 +45,13 @@ export default function Player( { isSelected, attributes } ) {
 		seekbarPlayedColor,
 	} = attributes;
 
+	const containerStyle = style[ 'videopress-player__container' ];
 	const iconStyle = style[ 'videopress-player__loading-icon' ];
 	const loadingViewStyle = style[ 'videopress-player__loading' ];
 
-	const [ isPlayerLoaded, setIsPlayerLoaded ] = useState( false );
+	const [ isPlayerReady, setIsPlayerReady ] = useState( false );
 	const [ token, setToken ] = useState();
 	const [ previewCheckAttempts, setPreviewCheckAttempts ] = useState( 0 );
-	const [ loadingHeight, setLoadingHeight ] = useState( loadingViewStyle.height );
 	const previewCheckTimer = useRef();
 
 	// Fetch token for a VideoPress GUID
@@ -88,9 +86,11 @@ export default function Player( { isSelected, attributes } ) {
 	const isPreviewAvailable =
 		!! preview?.height || previewCheckAttempts > VIDEO_PREVIEW_ATTEMPTS_LIMIT;
 
+	const aspectRatio = preview?.width / preview?.height || DEFAULT_PLAYER_ASPECT_RATIO;
+
 	// Fetch the preview until it's ready
 	useEffect( () => {
-		if ( ! isPlayerLoaded || isRequestingEmbedPreview ) {
+		if ( ! isPlayerReady || isRequestingEmbedPreview ) {
 			return;
 		}
 
@@ -105,54 +105,43 @@ export default function Player( { isSelected, attributes } ) {
 		}, 1000 );
 
 		return () => clearTimeout( previewCheckTimer.current );
-	}, [ preview, isPlayerLoaded, isRequestingEmbedPreview, previewCheckAttempts ] );
+	}, [ preview, isPlayerReady, isRequestingEmbedPreview, previewCheckAttempts ] );
+
+	// Consider the player not ready when the preview changes.
+	// The player will be ready when receiving the first `videopress_ready` event.
+	useEffect( () => {
+		setIsPlayerReady( false );
+	}, [ preview ] );
 
 	const onSandboxMessage = useCallback( message => {
-		if ( message.event === 'videopress_loading_state' && message.state === 'loaded' ) {
-			setIsPlayerLoaded( true );
+		if ( message.event === 'videopress_ready' ) {
+			setIsPlayerReady( true );
 		}
 	}, [] );
 
-	// Set up container loading styles
-	const loadingStyle = {};
-	if ( ! isPreviewAvailable ) {
-		loadingStyle.height = loadingHeight;
-	}
-
-	const onLayout = useCallback( event => {
-		const { width } = event.nativeEvent.layout;
-		const scaledHeight = width * DEFAULT_PLAYER_ASPECT_RATIO;
-		const scaledOffset = ( width / MAX_WIDTH ) * LOADING_OFFSET_HEIGHT;
-		setLoadingHeight( scaledHeight - scaledOffset );
-	}, [] );
-
-	const renderEmbed = () => {
-		// Show the loading view if the embed html is not available or
-		// if still preparing the preview
-		if ( ! html || ( isPlayerLoaded && ! isPreviewAvailable ) ) {
-			return (
-				<View style={ [ loadingViewStyle, loadingStyle ] } onLayout={ onLayout }>
-					<Icon icon={ VideoPressIcon } size={ iconStyle?.size } style={ iconStyle } />
-					<Text style={ style[ 'videopress-player__loading-text' ] }>
-						{ __( 'Loading', 'jetpack-videopress-pkg' ) }
-					</Text>
-				</View>
-			);
-		}
+	const loadingUI = (
+		<View style={ loadingViewStyle }>
+			<Icon icon={ VideoPressIcon } size={ iconStyle?.size } style={ iconStyle } />
+			<Text style={ style[ 'videopress-player__loading-text' ] }>
+				{ __( 'Loading', 'jetpack-videopress-pkg' ) }
+			</Text>
+		</View>
+	);
 
-		return (
-			<SandBox
-				html={ html }
-				onWindowEvents={ { message: onSandboxMessage } }
-				viewportProps="user-scalable=0"
-			/>
-		);
-	};
+	const sandBox = (
+		<SandBox
+			html={ html }
+			onWindowEvents={ {
+				message: onSandboxMessage,
+			} }
+			viewportProps="user-scalable=0"
+		/>
+	);
 
 	return (
-		<View style={ [ style[ 'videopress-player' ], loadingStyle ] }>
-			{ ! isSelected && <View style={ style[ 'videopress-player__overlay' ] } /> }
-			{ renderEmbed() }
+		<View style={ [ containerStyle, { aspectRatio } ] }>
+			{ ! isPlayerReady || ! isPreviewAvailable ? loadingUI : null }
+			{ html ? sandBox : null }
 		</View>
 	);
 }
```

Here you can see how it looks:

https://user-images.githubusercontent.com/14905380/234577380-e5b9793c-7ff6-42b3-9c92-7872df06fc2e.mp4

projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
```diff
@@ -99,33 +109,41 @@ export default function Player( { isSelected, attributes } ) {
...
 
+	// Show the loading overlay when: 
+	// 1. Player is not ready
+	// 2. Player is loaded but preview is not ready
+	const showLoadingOverlay = ! isPlayerReady || ( isPlayerLoaded && ! isPreviewReady );
+	
 	return (
-		<View style={ [ style[ 'videopress-player' ], loadingStyle ] }>
+		<View style={ [ style[ 'videopress-player' ], { aspectRatio } ] }>
 			{ ! isSelected && <View style={ style[ 'videopress-player__overlay' ] } /> }
```
jhnstn, line 147: This deselected overlay is still needed to prevent the video from playing when selecting the block.

projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
```diff
@@ -99,33 +109,41 @@ export default function Player( { isSelected, attributes } ) {
...
+					{ __( 'Loading', 'jetpack-videopress-pkg' ) }
+				</Text>
+			</View>
+		</View>
+	);
 
+	// Show the loading overlay when: 
+	// 1. Player is not ready
+	// 2. Player is loaded but preview is not ready
+	const showLoadingOverlay = ! isPlayerReady || ( isPlayerLoaded && ! isPreviewReady );
```
jhnstn, line 143: We still need the `isPlayerLoaded` state to show the VP processing window if the preview is not ready.

@fluiddot I noticed in the screen recording with your patch applied the video adjusts to the correct aspect ratio while the video is processing. While that can happen from time to time the more common behavior I see is that the preview is not ready till a few moments after the video is processed. So in these cases we need to hide the loading screen while the preview is not ready but the player has loaded.

projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
```diff
@@ -99,33 +109,41 @@ export default function Player( { isSelected, attributes } ) {
...
+	
 	return (
-		<View style={ [ style[ 'videopress-player' ], loadingStyle ] }>
+		<View style={ [ style[ 'videopress-player' ], { aspectRatio } ] }>
```
jhnstn, line 146: @fluiddot thanks for suggesting the `aspectRatio` style!

projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
```diff
@@ -17,12 +17,11 @@ import getMediaToken from '../../../../../lib/get-media-token/index.native';
...
 import { usePreview } from '../../../../hooks/use-preview';
 import addTokenIntoIframeSource from '../../../../utils/add-token-iframe-source';
+import { VideoPressIcon } from '../icons';
 import style from './style.scss';
 
 const VIDEO_PREVIEW_ATTEMPTS_LIMIT = 10;
-
-// The preview is ready when it has a height property
-const isPreviewReady = preview => !! preview?.height;
+const DEFAULT_PLAYER_ASPECT_RATIO =  16 / 9; // This is the observed default aspect ratio from VideoPress embeds.
```
jhnstn: I found the `16:9` aspect ratio best matches the default embed aspect ratio. I did find [this reference](https://github.com/Automattic/jetpack/blame/rnmobile/update/vp-loading-screen/projects/packages/videopress/src/class-block-editor-content.php#L108) to the `16:9` aspect ratio in the VP package but I don't think that's what's served to the editor.  Either way I found this is working well

projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
```diff
@@ -99,33 +109,41 @@ export default function Player( { isSelected, attributes } ) {
 	}, [ preview, isPlayerLoaded, isRequestingEmbedPreview, previewCheckAttempts ] );
 
 	const onSandboxMessage = useCallback( message => {
-		if ( message.event === 'videopress_loading_state' && message.state === 'loaded' ) {
-			setIsPlayerLoaded( true );
+		switch ( message.event ) {
+			case 'videopress_ready':
+				setIsPlayerReady( true );
+				break;
+			case 'videopress_loading_state':
```
jhnstn, line 123: @fluiddot you mentioned you were not seeing the `videopress_loading_state` fire on iOS when loading an existing VP block. Can you confirm this?

I've consistently seen that event fire during both uploading or existing cases. However I tried disabling the `setIsPlayerLoaded` call to see how the block behaves. It looks fine for an existing block but it will hide the processing screen if the preview is not ready (see this [comment](https://github.com/Automattic/jetpack/pull/30251/files#r1178316365)). That's probably not an issue since you mentioned not seeing the event on existing VP blocks.

jhnstn:
@fluiddot Thanks for the suggestions!
I adopted most of them and made a few other modifications. I left comments that I hope explain the reasons.

Now the fun part. This is only working on iOS 😭 

It looks like the `videopress_ready` event does not fire on Android. Or at least I have not been able to see it fire on my Pixel 6 running Android 13. I've inspected the `WebView` and I don't see any issues. Also all the other player events are firing as expected so I think we just have to deal with the ready event not firing

I started on a few workarounds but opted to leave them out of this PR since we are targeting to release on iOS first. I will open a follow up PR for Android.

projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
fluiddot: > @fluiddot you mentioned you were not seeing the `videopress_loading_state` fire on iOS when loading an existing VP block. Can you confirm this?

Yep, although I only managed to reproduce this issue a couple of times when replacing a video with a new upload 🤔. The main problem I saw related to that event is that `videopress_loading_state` is sent after the video is processed, so we can't use it to control the visibility of the loading overlay.

fluiddot:
> It looks like the `videopress_ready` event does not fire on Android. Or at least I have not been able to see it fire on my Pixel 6 running Android 13. I've inspected the `WebView` and I don't see any issues. Also all the other player events are firing as expected so I think we just have to deal with the ready event not firing

I think that event is triggered by the `load` event of the iframe, which doesn't work either on Android for some reason. As an alternative, we could use the [`onLoadEnd` prop from the WebView's library](https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#onloadend), which should be triggered when the content is loaded. In any case, let's stick to this approach for now in order to release the block on iOS 👍.

fluiddot (commented):
I tested the following cases and found an issue related to replacing a video with an item from the WP library.

- Loading post with a VideoPress video
    - Landscape video: ✅
    - Portrait video: ✅
- Upload a video
    - Landscape video: ✅
    - Portrait video: ⚠️ - Can't be reproduced consistently but sometimes the loading screen is displayed after processing the video. 

https://user-images.githubusercontent.com/14905380/234836616-002e47ee-ba22-4048-b487-2cfa235f5b40.mp4

- Replace with Media item
    - Landscape video: ❌ - The loading screen is not displayed.
    - Portrait video: ❌ - The loading screen is not displayed.

https://user-images.githubusercontent.com/14905380/234836866-c236332e-4a5c-4cfa-9864-73206996baf1.mp4

- Replace with upload
    - Landscape video: ✅
    - Portrait video: ⚠️ - Can't be reproduced consistently but sometimes the loading screen is displayed after processing the video. 

<hr>
> - Replace with Media item - Landscape/Portrait video: ❌ - The loading screen is not displayed.

I think we can fix this by simply resetting (i.e. setting to `false`) `isPlayerLoaded` and `isPlayerReady` state values when the video changes. I'll push a quick fix.

fluiddot (approved):
I re-tested the flows I shared in [this comment](https://github.com/Automattic/jetpack/pull/30251#pullrequestreview-1403722486) and all worked as expected 🎊 .

@jhnstn I'd like to defer the merge to you so you can check the last change I made (https://github.com/Automattic/jetpack/pull/30251/commits/2058ce30aa6af5c6ef097e01ff76381f5e2cfbe1). Thanks!

- Loading post with a VideoPress video
    - Landscape video: ✅
    - Portrait video: ✅
- Upload a video
    - Landscape video: ✅
    - Portrait video: ⚠️ - Can't be reproduced consistently but sometimes the loading screen is displayed after processing the video. 
- Replace with Media item
    - Landscape video: ✅ 
    - Portrait video: ✅
- Replace with URL
    - Landscape video: ✅ 
    - Portrait video: ✅
- Replace with upload
    - Landscape video: ✅
    - Portrait video: ⚠️ - Can't be reproduced consistently but sometimes the loading screen is displayed after processing the video.

PR #30256, opened 2023-04-24
Add a 'copy' button for emails on the jetpack forms dashboard

ice9js:
## Proposed changes:

This PR adds a 'Copy' button for the email on the response tab in Jetpack Forms dashboard. The button will appear when hovering over the line and will copy the email address to the clipboard.

Screenshot 2023-04-24 at 09 47 25

### Other information:

No.

## Testing instructions:

- Go to `admin.php?page=jetpack-forms`. Make sure you already have some responses in there.
- Hover over the email address in the header of the response tab.
- You should see a 'Copy' button next to the email.
- Clicking the button should copy the email address to the clipboard.


digitalwaveride (changes requested):
😻 Nice Kuba! Thanks a ton for bringing this in!
It worked for me as advertised ... 
image

Just for the responsive views the button behaviour was not optimal... 

image

For these extreme cases (on mobile there would be no hover) my proposal would be to make sure that button would not leave the screen, but it's ok if the button is on hover overlapping, being above, the actual email.
 
image

In a best case scenario giving it some "left padding" via the overlay... if doable:

image

digitalwaveride (approved):
Works! 🚀

PR #30257, opened 2023-04-24
Fix html entities not displaying correctly in the forms dashboard

ice9js:
## Proposed changes:

This patch fixes HTML entities not being rendered correctly on the forms dashboard. Instead of applying the fix in every place we render a piece of information, I just added an additional step to when a response comes in to make sure the data is already in the correct format before it goes into the store. I think this will also help prevent accidental regressions in the future.

### Other information:

No.

## Testing instructions:

- Create a post with a contact form.
- Respond to the created contact form and make sure to include some special characters inside the name or message field. `&` or `<` are good first candidates.
- Go to `admin.php?page=jetpack-forms` and view the new response.
- All response values should be rendered correctly in both list and response views.

### DIFF
--- a/projects/packages/forms/changelog/fix-forms-dashboard-html-entities
+++ b/projects/packages/forms/changelog/fix-forms-dashboard-html-entities
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fixed html entities not displaying correctly in the forms dashboard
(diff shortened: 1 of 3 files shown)


digitalwaveride (approved):
Thanks Kuba!!! 🚀 

Fixed for me:

image

CGastrell (approved):
Works as advertised :shipit:

PR #30258, opened 2023-04-24
Fix forms dashboard view setting not saving on WP.com

ice9js:
## Proposed changes:

This PR fixes an issue with Jetpack Forms' view setting not saving on WP.com.

This was caused by a race condition with [the code handling a similar setting](https://github.com/Automattic/jetpack/blob/5387e38c2960c52d347f4ade3abc118361fc987a/projects/plugins/jetpack/modules/masterbar/admin-menu/class-base-admin-menu.php#L679) for 'Calypsoified views'. Specifically, that function would trigger a redirect before our code was executed.

Renaming our query param from `preferred-view` to `dashboard-preferred-view` fixes the issue.

### Other information:

No.

## Testing instructions:

Note that this shouldn't change anything on local or stand-alone environments and will only affect WP.com.
Make sure the forms package and the new dashboard is enabled for testing if necessary:

```
add_filter( 'jetpack_contact_form_use_package', '__return_true' );
add_filter( 'jetpack_forms_dashboard_enable', '__return_true' );
```

- Go to WP Admin and click feedback.
- You should be redirected to your currently preferred forms dashboard view.
- Use the `view` dropdown in the upper right corner to switch to the different view version.
- You should be redirected to the other view, which should now become your preferred option.
- Navigate to outside of the forms WP Admin area.
- Click 'feedback' again.
- Your setting should be remembered and you should be redirected to the version you last selected.


CGastrell (approved):
Works! :shipit:

PR #30260, opened 2023-04-24
Forms: add field id setting

CGastrell:
This PR adds a new text control to the field block's inspector, allowing to change the name/ID for the input.

## Proposed changes:
Add the name/ID setting on the inspector's Advance section for form input blocks
image

This is replacing our per-input implementation where it would be referred as "Unique CSS ID".

### Other information:

No

## Testing instructions:
Insert a form on a post. Select any input and check the inspector's Advanced section, you should be able to change the input's name/ID.
It should be restricted to alphanumeric, dash and underscore (this restriction can be bypassed using the code editor).
Look at the published post, see that the input has the name/ID set for it.


digitalwaveride (changes requested):
Can we move the Name ID field into the "Advanced" section, as the functionality is not really "self-explanatory", very technically, and for most users not really relevant?

The feature feels on a similar level like "customizing css classes", and would be a good fit under "advanced".

image

CGastrell:
> Can we move the Name ID field into the "Advanced" section, as the functionality is not really "self-explanatory", very technically, and for most users not really relevant?
> 
> The feature feels on a similar level like "customizing css classes", and would be a good fit under "advanced".
> 
It not only feels similar, it's the exact same thing. The reason to put it up more prominently is that it's relevant mostly for inputs, where you not only control the CSS/styling, but actual form functionality.

If we aim to put it in the "Advanced" section, then it's already there and we only need to instruct users on where to find it.

digitalwaveride:
Perfect, let's put it into the advanced section! There is little chance to understand what this field ID is about with out any documentation. Let's keep it in the advanced section, and let's include in the description below the ID field a "learn more" link, that points to the documentation.

digitalwaveride:
Great! I like it Christian!

Just wondering if we should simplify the instructions, and link out to the documentation ... which we obviously still would need to create ;) ... something like this?

image

CGastrell:
> Great! I like it Christian!
> 
> Just wondering if we should simplify the instructions, and link out to the documentation ... which we obviously still would need to create ;) ... something like this?
> 

Let's make it an issue (lacks documentation, link on settings) and let it for next iterations (either us or whoever takes over)

digitalwaveride (approved):
👍  Let's do this.

> Let's make it an issue (lacks documentation, link on settings) and let it for next iterations (either us or whoever takes over)

Create [issue](https://github.com/orgs/Automattic/projects/491/views/1?pane=issue&itemId=27447778) draft here.

tbradsha:
I'll note that "name" and "ID" have separate meanings/purposes for inputs, so it's a bit confusing to combine them. It looks like it's technically the ID attribute on the backend.

@mikestottuk @StefMattana might be looking into verifying this didn't break anything on the CRM side. I think not, but worth a check and update to the docs:
https://kb.jetpackcrm.com/knowledge-base/jetpack-contact-forms/

StefMattana:
@tbradsha I added an ID value in a contact form Name field here: https://steftest.mystagingwebsite.com/2025/01/23/test-form-id-with-crm/
I sent a message with a fake email address, I received the form in the Feedback section in WP Admin, and it created a CRM contact with no problems. 

I guess it doesn't conflict as you anticipated, but let me know if I need to test it further.

tbradsha:
@StefMattana Thanks for testing! Note that the JPCRM doc linked above still mentions "Unique CSS ID" instead of "Name/ID".

StefMattana:
Updated!

PR #30135, opened 2023-04-17
CRM: Learn More buttons and links to be consistent with Jetpack style guide

coder-karen:
## Proposed changes:

* This PR changes the style of the Learn More button to be an info icon in line with Jetpack styles and following the Jetpack CRM style guide.
* It also changes the Learn More link in the Learn More popup to include an external link icon.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

https://github.com/Automattic/zero-bs-crm/issues/2998

## Does this pull request change what data or activity we track or use?

No.

## Testing instructions:

* Using the [Jetpack Beta plugin](https://github.com/Automattic/jetpack-beta) on a test site with this branch applied or testing locally with the branch checked out, open up any page that previously had a Learn More button in the navigation (for example the Dashboard, any Contact / Transaction / Company / Invoice / Quote dashboard pages).
* You should now see an info icon in place of the Learn More button.
* On clicking the icon, the popup should include a Learn More link followed by an external link icon.

Info icon style: 

Screenshot 2023-04-17 at 16 50 26

Screenshot 2023-04-17 at 16 50 35

PR #30268, opened 2023-04-24
CRM: Add PHPUnit dependencies 

kallehauge:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3069

## Proposed changes:

* Add WorDBless as a dependency so we can start doing integration tests.
* Adding WorDBless will also add PHPUnit as a dependency which means we add PHP unit tests.
* Add a dummy test just to prove that it successfully runs (it will be removed in #30198)
* Specifically require PHPUnit 8 (instead of 9 that would be installed otherwise) since we require PHP `7.2`.
* Move Codeception acceptance tests into its own directory to prevent confusion with the `tests/php` directory.

### Other information:

## Jetpack product discussion

It's technically related to pbhBOz-3ss-p2 since it's prep work, but there's nothing that specifically addresses adding PHPUnit to the CRM project.

Nope.

## Testing instructions:

**Automated testing**

All of the tests in GitHub should appear as ✅ 

You can also go check e.g. `PHP tests: PHP 7.3 WP latest > Run project tests` to verify that both PHPUnit executed with 1 successful test and that all of our previous acceptance tests also appears in the terminal as ✅ 

That should verify that changing the Composer scripts and moving the conception files haven't broken anything and everything is running.

**Manual testing**

* Checkout branch locally
* Run: `jetpack install plugins/crm`
* Go to the CRM directory: `projects/plugins/crm`
* Run `composer run-script phpunit`

_(Codeception prevents us from running `jetpack test plugins/crm` locally, so we have to call the PHPUnit directly)._


projects/plugins/crm/tests/php/rest-api/v4/class-dummy-test.php
```diff
@@ -0,0 +1,29 @@
...
+class Dummy_Test extends BaseTestCase {
+
+	/**
+	 * Dummy test to ensure PHPUnit works.
+	 */
+	public function test_that_our_tests_are_testing_the_test_system() {
+		add_option( 'jpcrm_test_option', 'hello' );
+
+		$this->assertSame( 'hello', get_option( 'jpcrm_test_option' ) );
+	}
```
kallehauge, line 27: This test class is simply to verify that everything works, so you don't need to focus on the test itself 😄

anomiex:
> It is not super ideal but PHPUnit 9+ requires PHP 7.3 or newer.
> We could, technically, try and downgrade some of our packages but we have a lot of conflicts with:
>
> 1. We need to use PHPUnit 8 or older for PHP 7.2 to be supported.

The yoast/phpunit-polyfills package should already be handling getting you the right version of PHPUnit for the version of PHP in use. The reason it's not is because you [override the version detection](https://github.com/Automattic/jetpack/blob/a1c838e349d01127a5a2fd5a97381fb98efb7790/projects/plugins/crm/composer.json#L47-L49).

When writing tests, that package also provides traits to make newer assertions available under older versions of PHPUnit.

> 2. Many of our dependencies only support PHPUnit 7 or 9 (so not 8) which means we're limited to 7.

yoast/phpunit-polyfills goes all the way back to PHPUnit 4.8.36 if necessary (although it shouldn't be, and they're upping that to 5.7 in v2). But which dependencies support 7 or 9 but not 8? Things seem to work fine when I set up a 7.2 environment and remove the override mentioned above so PHPUnit 8 gets installed.

kallehauge:
> The reason it's not is because you [override the version detection](https://github.com/Automattic/jetpack/blob/a1c838e349d01127a5a2fd5a97381fb98efb7790/projects/plugins/crm/composer.json#L47-L49).

Oh, I love code reviews! ❤️ I hadn't even considered the platform config to be the potential culprit since our 7.2 compatibility test have pass up until now (e.g.: https://github.com/Automattic/jetpack/pull/30135) and I never noticed any mentions of the platform version in the logs.
 
> But which dependencies support 7 or 9 but not 8? Things seem to work fine when I set up a 7.2 environment and remove the override mentioned above so PHPUnit 8 gets installed.

`automattic/wordbless` doesn't mention PHPUnit 8 but it seems to work fine now that my other conflicts have been fixed 🤷 

```
"require-dev": {
	"phpunit/phpunit": "^5.7 || ^6.5 || ^7.5 || ^9.5"
},
```
https://github.com/Automattic/wordbless/blob/master/composer.json#L17

anomiex:
Ah, since that's in `require-dev` it doesn't matter for anyone using the package, it's what WorDBless uses for its own unit tests.

kallehauge:
> Ah, since that's in require-dev it doesn't matter for anyone using the package, it's what WorDBless uses for its own unit tests.

I'm not coding enough anymore; that's so obvious 😂 Thank you for all your feedback @anomiex 🙇

projects/plugins/crm/tests/php/rest-api/v4/class-dummy-test.php
cleacos: 👍

projects/plugins/crm/.gitignore
```diff
@@ -19,4 +20,5 @@
 /templates/quotes/.htaccess
 /templates/quotes/index.html
 /tests/acceptance.suite.yml
```
cleacos:
```suggestion
/tests/acceptance.suite.yml
/tests/codeception/acceptance.suite.yml
```

cleacos:
Great work! It works as expected!

Things to take into account when we are adding tests:

- New test files must be ended by the `-test.php` suffix
- To execute the old acceptance tests locally, we have to copy/move our local config file `acceptance.suite.yml` to the new `codeception` folder: `/tests/codeception/acceptance.suite.yml`

**Note** this PR should add to the ignore config files the `acceptance.suite.yml` file: `/tests/codeception/acceptance.suite.yml` ( I left the code  suggestion )

projects/plugins/crm/.gitignore
cleacos: File for local configuration should be added to the `.gitignore` file

kallehauge: Nice catch @cleacos 👏
I have added the new path to .gitignore and a comment for the old path so we can remove it in the future if we want to keep the file clean

cleacos (dismissed):
👍  LGTM!

cleacos (approved):
re-approved

PR #30275, opened 2023-04-24
[Earn] Make the Donation block available on Free plans

romarioraffington:
---

* Relates to https://github.com/Automattic/gold/issues/11
* To be tested with D108873-code
* Relates to https://github.com/Automattic/wpcomsh/pull/1358

## Proposed changes:

* Make Donation block free for all plans and charge a 10% commission for free blogs on free plans. `should_upgrade_to_access_memberships` is no longer needed as recurring payments will be free for all users

### Other information:

No
## Testing instructions:

1. Sync code to WPCOM
2. Sandbox your site, `subscribe.wordpress.com` and `public-api.wordpress.com`
3.  On a site with a free plan, add a Donation block to post and save it
4.  Open the post in an incognito window and verify the user can checkout successfully


projects/plugins/jetpack/extensions/blocks/donations/edit.js
```diff
@@ -61,9 +60,8 @@ const Edit = props => {
 			return;
 		}
 
-		// Set fake products when plan should be upgraded or there is no connection to Stripe so users can still try the
-		// block in the editor.
-		if ( result.should_upgrade_to_access_memberships || result.connect_url ) {
+		// Set fake products when there is no connection to Stripe so users can still try the block in the editor.
+		if ( result.connect_url ) {
```
romarioraffington, line 64: @millerf  it's unclear if we want to make the Donation block free for all users as well

projects/plugins/jetpack/extensions/blocks/donations/edit.js
```diff
@@ -61,9 +60,8 @@ const Edit = props => {
 			return;
 		}
 
-		// Set fake products when plan should be upgraded or there is no connection to Stripe so users can still try the
-		// block in the editor.
-		if ( result.should_upgrade_to_access_memberships || result.connect_url ) {
```
romarioraffington, line 66: `should_upgrade_to_access_memberships` is currently expected to passed via the API to the Donations block. As such I chose not to remove the param from the API in D108873-code until this code is deployed on WPCOM and released in a new JP version.

However, in JavaScript `result.should_upgrade_to_access_memberships` would return `undefined` if `should_upgrade_to_access_memberships` is not passed which is okay in this case. Feedback appreciated

projects/plugins/jetpack/extensions/blocks/donations/edit.js
millerf: Yes, all Earn products.

millerf:
While you are working on Donation blocks, can you look into this issue: https://github.com/Automattic/jetpack/issues/30299 ?

If it is an easy one-liner, do fix it here, otherwise we can use another PR...

projects/plugins/jetpack/extensions/blocks/donations/edit.js
millerf: I believe we should remove all `should_upgrade_to_access_memberships` in this PR.

> However, in JavaScript result.should_upgrade_to_access_memberships would return undefined if should_upgrade_to_access_memberships is not passed which is okay in this case. 

That is actually great news as it will make the feature reto-compatible once D108873-code is out!

millerf:
I answered on all diffs...

projects/plugins/jetpack/extensions/blocks/recurring-payments/test/edit.js
```diff
@@ -44,7 +44,6 @@ describe( 'MembershipsButtonEdit', () => {
 		connect_url: '',
 		connected_account_id: 1,
 		products: [],
-		should_upgrade_to_access_memberships: false,
 		site_slug: 'test',
 		upgrade_url: 'https://wordpress.com/checkout/test/jetpack_security_daily_monthly',
```
millerf: Should be removed?

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/memberships.php
```diff
@@ -262,7 +262,7 @@ public function create_products( $request ) {
 	 *
 	 * @param \WP_REST_Request $request - request passed from WP.
 	 *
-	 * @return WP_Error|array ['products','connected_account_id','connect_url','should_upgrade_to_access_memberships','upgrade_url']
+	 * @return WP_Error|array ['products','connected_account_id','connect_url','upgrade_url']
```
millerf: I guess `upgrade_url` should maybe be removed as well?

romarioraffington: Good point. I think we should re-add `should_upgrade_to_access_memberships` since we choose not to remove it from the WPCOM endpoint as yet

millerf: > I think we should re-add should_upgrade_to_access_memberships since we choose not to remove it from the WPCOM endpoint as yet

We chose not to remove it to make sure we are backward-compatible for users that don't update. But we need to remove it here...

romarioraffington: Ok, I'll remove it. IMHO, removing it pre-maturely wouldn't accurately represent what the endpoint returns.

millerf: You need to remove it from WPCOM end anyway. So there will always be a lapse where it won't be accurate, one way or the other.

romarioraffington: Fixed in [1ceba3f](https://github.com/Automattic/jetpack/pull/30275/commits/1ceba3f369e043a98a79e35d1c5107d4483e04be)

projects/plugins/jetpack/extensions/blocks/recurring-payments/test/edit.js
romarioraffington: Fixed in [1ceba3f](https://github.com/Automattic/jetpack/pull/30275/commits/1ceba3f369e043a98a79e35d1c5107d4483e04be)

millerf (dismissed):
LGTM

projects/plugins/jetpack/extensions/blocks/recurring-payments/test/edit.js
anomiex, line 1: Should the skipped test at https://github.com/Automattic/jetpack/blob/1ceba3f369e043a98a79e35d1c5107d4483e04be/projects/plugins/jetpack/extensions/blocks/recurring-payments/test/edit.js#L118-L130 be removed now?

projects/plugins/jetpack/extensions/store/membership-products/resolvers.js
```diff
@@ -54,7 +53,6 @@ const mapAPIResponseToMembershipProductsStoreData = ( response, registry, dispat
 	const postId = registry.select( editorStore ).getCurrentPostId();
 
 	dispatch( setConnectUrl( getConnectUrl( postId, response.connect_url ) ) );
-	dispatch( setShouldUpgrade( response.should_upgrade_to_access_memberships ) );
 	dispatch( setSiteSlug( response.site_slug ) );
 	dispatch( setUpgradeUrl( response.upgrade_url ) );
```
anomiex: Should this line go away too, since it seems the plan is to remove `upgrade_url` from the response as well?

projects/plugins/jetpack/extensions/blocks/recurring-payments/test/edit.js
romarioraffington: Fixed in b4d9f8157e851d4817a67c38e277eb6b9c45dcaa

projects/plugins/jetpack/extensions/store/membership-products/resolvers.js
romarioraffington: Thanks! Fixed in [2d92290](https://github.com/Automattic/jetpack/pull/30275/commits/2d922908ef672a095c1e18979bfc2877c343a9b9)

anomiex (approved):

PR #30280, opened 2023-04-25
Social: Add support to flag unsupported connections

gmjuhasz:
This change adds support for flagging connections as not supported.

## Proposed changes:

* Added error code in a WPCOM change, which is received here and processed to display the notice if needed.
* Previously broken connections got the 'broken' status code.

### Other information:

No.
## Testing instructions:

* Test that the broken connection logic works as before. You can make a connection broken by revoking the access from the social media site.
* For the classic editor, you have to run 'jetpack build packages/publicize' to see the changes
* You can hardcode flag a connection as unsupported, by returning 1 as 'error_code' in 'projects/packages/publicize/src/class-publicize-base.php' for a specific connection. In that case the unsupported notice should appear


projects/packages/publicize/src/class-publicize.php
```diff
@@ -613,6 +614,7 @@ public function test_connection( $service_name, $connection ) {
 			'user_can_refresh' => $user_can_refresh,
 			'refresh_text'     => $refresh_text,
 			'refresh_url'      => $refresh_url,
+			'code'             => $connection_error_code,
```
pablinos: I'm not sure but I'm not seeing the notices as the error code isn't coming through for some reason - this is with Jetpack Social.

I wondered if this should be 
```suggestion
			'error_code'             => $connection_error_code,
```

But it doesn't seem to make a difference!

pablinos:
This is working well for me. I had to explicitly build the publicize package to get the changes to show in the classic editor, but that's not a problem with this PR.

projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -47,7 +47,12 @@ export default function PublicizeForm( {
 
 	const Wrapper = isPublicizeDisabledBySitePlan ? Disabled : Fragment;
 
-	const brokenConnections = connections.filter( connection => false === connection.is_healthy );
+	const brokenConnections = connections.filter(
+		connection => false === connection.is_healthy && 'broken' === connection.error_code
```
pablinos: Is is_healthy ever not a boolean? We could change these checks to `! connection.is_healthy`

pablinos (commented):
I just realised that we needed to test this on simple sites too, and it's not working for me, the any services marked as unsupported are coming back as working from the `connection-test-results` call.

projects/js-packages/publicize-components/src/components/form/index.js
gmjuhasz: Yeah I think it was changed, I remember I added this for a reason back then, but probably it's not needed anymore

danielpost:
I can't approve because it's a draft, but this tested as expected on a simple site.

projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -70,6 +75,43 @@ export default function PublicizeForm( {
...
+							fixLink: <ExternalLink href={ connectionsAdminUrl } />,
+						}
+					) }
+				</Notice>
+			) }
+			{ unsupportedConnections.length > 0 && (
+				<Notice type={ 'error' }>
+					{ createInterpolateElement(
+						_n(
+							'One of your social connection is not supported anymore. <moreInfo>Learn more here</moreInfo>.',
```
pablinos: I think 'connection' should still be plural here

```suggestion
							'One of your social connections is not supported anymore. <moreInfo>Learn more here</moreInfo>.',
```

I was wondering if we should change this to 'Twitter is not supported anymore', and then we don't have to worry about how many unsupported connections there are.

projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -70,6 +75,43 @@ export default function PublicizeForm( {
...
+							'Some of your social connections are not supported anymore. <moreInfo>Learn more here</moreInfo>.',
+							unsupportedConnections.length,
+							'jetpack'
+						),
+						{
+							moreInfo: (
+								<ExternalLink href={ getRedirectUrl( 'publicize-connection-not-supported' ) } />
```
pablinos: It's seems that we can only have different redirect URLs for Jetpack and WPCOM if we include the site ID/slug in the target URL!

We could fake it and and the site identifier in a query parameter, even though it won't be used, make a change to Jetpack redirects, or pass in a different URL in the initial state, like we do with the `connection-test-results` URL. The last might be the quickest and simplest for now.

projects/packages/publicize/src/js/classic-editor-connections.js
```diff
@@ -56,6 +65,29 @@ jQuery( function ( $ ) {
 				.addClass( 'publicize-token-refresh-message' )
 				.append( msg.replace( '%s', connectionsUrl ) );
 		}
+
+		if ( unsupportedConnections ) {
+			/* translators: %s is the link to the connections page in Calypso */
+			const msg = _n(
+				'One of your social connection is not supported anymore. <a href="%s" rel="noopener noreferrer" target="_blank">Learn more here</a>.',
```
pablinos: Depending on what message we're using, we'll need to fix this one too.

pablinos (commented):
Just some slight niggles around the message and where to link to, other than that, this is working well for me.

projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -53,7 +53,12 @@ export default function PublicizeForm( {
 		isSocialImageGeneratorEnabled && isSocialImageGeneratorEnabledForPost;
 	const Wrapper = isPublicizeDisabledBySitePlan ? Disabled : Fragment;
 
-	const brokenConnections = connections.filter( connection => false === connection.is_healthy );
+	const brokenConnections = connections.filter(
+		connection => false === connection.is_healthy && 'broken' === connection.error_code
```
pablinos: I wonder if we should check for `connection.error_code` being `undefined` here. Until the corresponding change is deployed on WPCOM then you don't get any notices for broken connections because of this change, so something like that could mean we don't need to worry about the order things are deployed.

projects/packages/publicize/src/js/classic-editor-connections.js
```diff
@@ -20,6 +20,23 @@ jQuery( function ( $ ) {
 		timer = setTimeout( publicizeConnTestStart, 2000 );
 	} );
 
+	const countConnectionsBy = ( status, response ) => {
+		return ! response.data
+			? 0
+			: response.data.reduce( ( count, testResult ) => {
+					if (
+						! testResult.connectionTestPassed &&
+						status === testResult.connectionTestErrorCode
```
pablinos: ...we'd need to make a similar change here, maybe something like 
```suggestion
						status === testResult.connectionTestErrorCode ?? 'broken'
```

pablinos (dismissed):
This is working well for me. I just wondered about whether we should default the error code, so that it can be merged and will still work while we wait for the change to hit WPCOM

projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -22,6 +22,9 @@ import Notice from '../notice';
 import PublicizeSettingsButton from '../settings-button';
 import styles from './styles.module.scss';
 
+const checkConnectionCode = ( connection, code ) =>
+	false === connection.is_healthy && code === ( connection.error_code ?? 'broken' );
```
gmjuhasz, line 26: @pablinos we have to keep false here, because is_healthy is only defined after we come back from the connection test call. Otherwise everything becomes broken, until that call is finisehd

pablinos (approved):
This is working well for me. As discussed, Social users will still see the connection as broken rather than 'unsupported' until the API change is made on WPCOM, but that's fine.

We might want to try and trunk merge to see if we can get the e2e tests passing.

projects/packages/publicize/src/class-publicize.php
```diff
@@ -600,6 +600,7 @@ public function test_connection( $service_name, $connection ) {
 
 		$xml_response            = $xml->getResponse();
 		$connection_test_message = $xml_response['faultString'];
+		$connection_error_code   = $xml_response['faultCode'];
```
anomiex, line 603: We're seeing some errors that make me think this is coming in as `0` (or `''`, but 0 seems more likely) rather than a valid code.

When a falsey value such as 0 is passed to `WP_Error::construct()` (line 619), you wind up with an "empty" WP_Error object instead of one with the passed message and data.

That results in `$error_data` being null in class-publicize-base.php at line 684, resulting in errors about trying to index null a few lines later.

pablinos: Oh no, that's not great! You're right, it comes back as 0 I think, as the IXRError class tries to parse the error code as an int, but it's a string, so it's converted to 0.

Sounds like we should make it `'pub_conn_test_failed'` again if the `faultCode` is `empty`.

How serious is this? Should we be reverting it?

@gmjuhasz Could you have a look at fixing it?

anomiex: The errors weren't enough to have to revert the 12.2-a.5 rollout to Atomic, so we should be ok on that front. I have no idea what effect if any this might have for users.

It would be good to have a fix in before Monday, to make it into 12.2-a.7. Or maybe it can be entirely fixed on the wpcom side where the error response is generated?

pablinos: Great, we can definitely get something in by Monday, and we can possibly fix it on the WPCOM side. It's a type casting issue, but we can maybe use another non-falsey error code.

gmjuhasz: Sorry, my Github notifications got messed up! I can take a look at this tomorrow morning. It's definitely the issue I described in the description too, I'll try to fix it on WPCOM

gmjuhasz: Hi! We hopefully fixed this in D110540-code, but also created a Jetpack side PR #30748 so it cannot happen in the future.

PR #30282, opened 2023-04-25
Visual Refresh: Update modules list view

### LINKED ISSUE #29836: Jetpack Plugin Visual Refresh: Modules list
Good 'ol `/wp-admin/admin.php?page=jetpack_modules` modules list. Not react. No functionality changes. 

p1HpG7-lx8-p2#module-list-debugger

Image

dereksmart:
Fixes #29836 

## Proposed changes:

Updates the `/wp-admin/admin.php?page=jetpack_modules` page with current styles. 

### Other information:

I'm not super great at css. But these styles shouldn't conflict with others at least. 

## Jetpack product discussion

p1HpG7-lx8-p2#module-list-debugger

No

## Testing instructions:

Visit `/wp-admin/admin.php?page=jetpack_modules` 
It should look great!
- In mobile too. 
It should not have lost any functionality. 
It should not have impacted any other pages.


projects/plugins/jetpack/_inc/lib/admin-pages/class.jetpack-admin-page.php
```diff
@@ -283,7 +283,7 @@ public static function load_wrapper_styles() {
 	 */
 	public static function wrap_ui( $callback, $args = array() ) {
 		$defaults          = array(
-			'is-wide'  => false,
+			'is-wide'  => true,
```
dereksmart: Sergey's approach in https://github.com/Automattic/jetpack/pull/30277/files#diff-77cc6b621a06e9e27a68759973cdb0af136471286592eb40392f46f568f70b04R156 is probably safer...

darssen (dismissed):
Thanks for working on this @sergeymitr !

I found a couple of things that might need some work.

When clicking on the checkbox of a module for the Bulk Actions, color is Blue and the check itself is not centered.
image

In Mobile the check not being centered is even more clear.
image

Also, for the disabled modules, before they were clearly disabled, and there was no checkbox available.
image
Now, they don't seem disabled. The grayed-out is the same as for Active Modules, and they do have the checkbox present.
image

jeffgolenski:
Great work @sergeymitr & @dereksmart. I only found a couple of small items to address (aside from the feedback from @darssen). Great work on the mobiles, since we didn't explore those during the design phase!

- [x] The "Apply" button has some weird text alignment happening in the latest version of chrome on MacOS.
- [x] Segmented control seems to now have a gray border / outline around it. Can we remove that?
- [x] For any module that is "not supported by current plan" let's disable the checkbox
- [x] The bottom border of the last `tr` should be hidden. the `last-of-type` selector should handle that nicely.

Mobile only

- [ ] When possible, the "Filter" button should be on the same line as the "Apply" button. On smaller viewports, it disappears completely. Let's make sure it's always visible, even if it drops to the second line on very small screens.

module-list

sergeymitr:
Thanks @darssen, @jeffgolenski.
Fixed all but one issue, asked Jeff about it in Slack: p1682692733033599-slack-C051H2TQR54

jeffgolenski:
Had a chat with @sergeymitr in slack. Feel free to ship this as-is for this release and we'll follow up with the mobile adjustment in the subsequent release. 

> When possible, the "Filter" button should be on the same line as the "Apply" button. On smaller viewports, it disappears completely. Let's make sure it's always visible, even if it drops to the second line on very small screens.

nateweller (dismissed):

nateweller:
Code looks good, page tests well, LGTM :+1:

Some minor style adjustments for follow-up:

1. [x] Misaligned checkmark icon for Bulk Actions checkbox

Screenshot 2023-05-01 at 11 21 34 AM

2. [ ] Highlighted/active filter background is filling the width of the container, instead of matching the text width.

Screenshot 2023-05-01 at 11 22 27 AM

3. Missing a `:focus` state on the view and sort filters, there should be an outline or indicator of some kind.

sergeymitr:
Hi @nateweller, thanks for taking a look.

> Highlighted/active filter background is filling the width of the container, instead of matching the text width.
It looks a bit unexpected, I agree, but actually follows the design:
Screen Shot 2023-05-01 at 13 53 18

sergeymitr:
The other two issues are fixed.

nateweller (approved):

PR #30287, opened 2023-04-25
Add map_block_settings global to jetpack-mu-wpcom

TimBroddin:
## Proposed changes:
This PR adds the  map_block_settings global to jetpack-mu-wpcom

### Other information:

## Testing instructions:
- Checkout this PR
- Follow instructions here: PCYsg-Osp-p2
- Add a new post/page
- Open devtools & make sure the iframe (post-new.php) is selected
- Output `Jetpack_Maps`:

CleanShot 2023-04-26 at 08 40 18@2x

(This will say mapbox in your testing)


projects/packages/jetpack-mu-wpcom/src/class-jetpack-mu-wpcom.php
```diff
@@ -97,6 +100,19 @@ public static function load_wpcom_rest_api_endpoints() {
...
+	 * Adds a global variable containing the map provider in a map_block_settings object to the window object
+	 *
+	 * @return void
+	 */
+	public static function load_map_block_settings() {
+		$map_provider = apply_filters( 'wpcom_map_block_map_provider', 'mapbox' );
+
+		wp_register_script( 'map-block-settings', '', array(), '1.0', true );
+		wp_localize_script( 'map-block-settings', 'map_block_settings', array( 'provider' => $map_provider ) );
+		wp_enqueue_script( 'map-block-settings', '', array(), '1.0', true );
```
jeherve: Instead of registering a new script, what do you think about hooking into the existing Jetpack block bundle, where the map block is being enqueued? You can use this as an example:

https://github.com/Automattic/jetpack/blob/fc41f14311eff83abe2188c4aa6e6307287f3113/projects/plugins/jetpack/extensions/blocks/anchor-fm/anchor-fm.php#L200

jeherve (changes requested):

projects/packages/jetpack-mu-wpcom/src/class-jetpack-mu-wpcom.php
TimBroddin: Thanks for pointing this out! I made the change.

jeherve (approved):
This should work. As we discussed before, I think this could have been added directly to the Jetpack plugin though. I'm not sure I fully understand why we need this in a separate plugin, but I may have missed something with this.

TimBroddin:
> This should work. As we discussed before, I think this could have been added directly to the Jetpack plugin though. I'm not sure I fully understand why we need this in a separate plugin, but I may have missed something with this.

You're probably right. I think our reasoning was not to pollute self-hosted sites with WPCOM only globals. But then again, we also ship them a bunch of Mapkit code.

PR #30290, opened 2023-04-25
CRM: Resolves 2873 - update to moment.js library

tbradsha:
Resolves Automattic/zero-bs-crm#2873 - update to moment.js library

## Proposed changes:

JPCRM was using `moment.js` 2.24.0, which was flagged in some systems as having a vulnerability. The security risk really only lies in `npm` server instances:
https://github.com/moment/moment/security/advisories/GHSA-8hfj-j24r-96c4

That said, it's best to avoid even false positives, so I've updated to the latest `moment.js` (with locales), which is `v2.29.4`.

Someday it'd be nice to excise this legacy library, but this is currently used by FullCalendar, and also for legacy datepickers.

### Other information:

## Testing instructions:

In the `fix/crm/2873-update_moment.js` branch:

1. Go to any JPCRM page, and run `moment.version` in the JS console. It should show `2.29.4`.
2. Go to the task manager calendar. It should load without errors.
3. Make sure the datepicker works when creating a new task.

Everything should work as before.


coder-karen (commented):
Works well, and the correct version is showing!

I do notice a new one: `devTools failed to load source map: Could not load content for .../wp-content/plugins/crm/js/lib/moment-with-locales.min.js.map: HTTP error: status code 404, net::ERR_HTTP_RESPONSE_CODE_FAILURE`

... and I notice this extra line at the bottom of `moment-with-locales.min.js.map` which seems to be causing that: `//# sourceMappingURL=moment-with-locales.min.js.map`

I know it only a source map issue that will show in the console, but we could remove it to keep things clean in the console since there are no map files?

tbradsha:
I had source map loading disabled on my browser, so didn't see this error originally, thanks! Resolved in 5fdb805.

coder-karen (commented):
Just to check - is it expected to see a console deprecation warning when loading the task calendar, related to `warning: value provided is not in a recognized RFC2822 or ISO format. moment construction falls back to js Date(), which is not reliable across all browsers and versions. Non RFC2822/ISO date formats are discouraged. Please refer to http://momentjs.com/guides/#/warnings/js-date/ for more info.`?

The calendar itself works well, `moment.version` is showing as 2.29.4.

tbradsha:
> is it expected to see a console deprecation warning when loading the task calendar

It's not new, anyway. 😅 We can certainly do better about dates/times: pbhBOz-1s2-p2

~Go ahead and log it as a new issue; while I'm in similar context in Automattic/zero-bs-crm#300 I'll see if it's an easy fix.~

Confirmed that I'll be fixing this anyway when working on #300. :^)

coder-karen (approved):
In that case, all looks good!

PR #30295, opened 2023-04-25
CRM: Resolves 1610 -  translate object names in PDF filename

tbradsha:
Resolves Automattic/zero-bs-crm#1610 -  translate object names in PDF filename

## Proposed changes:

Quote and invoice PDF filenames were inconsistent depending if they were downloaded directly or emailed as an attachment, and always included the object name in English.

This PR fixed the above, and did a fair amount of cleanup around this area, as there was a bunch of redundant code:

* There is now a generic `jpcrm_generate_pdf()` function that handles the generation of code.
* We now properly use the central `wp-content/jpcrm-storage/tmp` folder as a working dir instead of `wp-content/uploads/quotes` and wp-content/uploads/invoices`.
* The same filename is used whether emailed or downloaded directly.
* The object name is now translated in the filename.
* Invoice PDF filenames now consistently use the reference ID instead of the invoice ID if available.

Note that I also cleaned things up per coding standards, which involved updating some var and function names. Filenames are sanitised with `sanitize_title()` to ensure safe alphanumeric names.

### Other information:

## Testing instructions:

1. Switch your WP install to Spanish: `/wp-admin/options-general.php`
2. Create an invoice.
3. Download the invoice PDF.
4. Email the invoice to a contact with the PDF attached.
5. Add a reference number to the invoice.
6. Download the invoice PDF.
7. Email the invoice to a contact with the PDF attached.
8. Create a quote.
9. Download the quote PDF.
10. Email the quote to a contact with the PDF attached.

In `trunk`, you'll notice the object name in the filename is in English, and there's no consistency between the filenames in invoices. Quotes were consistent but not translated.

In `fix/crm/1610-adjust_PDF_names`, the object name is translated and the filenames are consistent.

### DIFF
--- a/projects/plugins/crm/includes/ZeroBSCRM.QuoteBuilder.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.QuoteBuilder.php
@@ -1,5 +1,6 @@
-<?php 
-/*!
+<?php // phpcs:ignore WordPress.Files.FileName.NotHyphenatedLowercase
+
+/*
  * Jetpack CRM
  * https://jetpackcrm.com
  * V1.20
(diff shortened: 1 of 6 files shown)


projects/plugins/crm/includes/ZeroBSCRM.QuoteBuilder.php
```diff
@@ -9,125 +10,111 @@
...
+			header( 'Content-Length: ' . filesize( $pdf_path ) );
+			header( 'Accept-Ranges: bytes' );
+			readfile( $pdf_path ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_operations_readfile
 
-        }
-
-        exit();
-    }
+			// delete the PDF file once it's been read (i.e. downloaded)
+			wp_delete_file( $pdf_path );
```
gogdzl, line 66: =)

gogdzl (approved):
Works as advertised.
Nice cleanup!

PR #30301, opened 2023-04-26
Social Image Generator: Reduce template image file size

pablinos:
While testing, the images in the template picker were loading slowly. They were all at least 100kb in size. Compressing them a bit and reducing the quality to 80% has them all <50kb, reducing the bundle size and making the loading much faster.

### Other information:

No

## Testing instructions:
* Enable SIG on a site
* Click the change template button
* With this change the previews should load a bit quicker, but look the same as without it.


danielpost (approved):
LGTM, thanks!

PR #30306, opened 2023-04-26
CRM: Cleaning up WP Editor helper functions and wp_editor usage

coder-karen:
## Proposed changes:

* All `zeroBSCRM_io_WPEditor` helper functions have been removed in this PR: `zeroBSCRM_io_WPEditor_WPEditorToDB`, `zeroBSCRM_io_WPEditor_DBToWPEditor`, `zeroBSCRM_io_WPEditor_DBToHTML`, `zeroBSCRM_io_WPEditor_DBToHTMLExcerpt`. As such, any place where they were called has been cleaned up to work without them, with replacement functions added where needed.
  * The `DBToWPEditor`, `DBToHTML`, and `DBToHTMLExcerpt` helper functions were all decoding HTML, which is not the expected behaviour in TinyMCE editors. The areas where these functions were being used have now been changed to reflect what is expected in those cases.
* Similarly, `wp_editor` function usage has also been cleaned up. In two cases we previously overcomplicated the `wp_editor` usage - in [this example](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/crm/admin/email/main.page.php?rgh-link-date=2023-04-25T15%3A27%3A42Z#L303) plus [this example](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/crm/admin/email/main.page.php?rgh-link-date=2023-04-25T15%3A27%3A42Z#L558), `$content` is always an empty string (it's the default info in the editor.) That has been changed as well. In the remaining case where a change was made, decoding wasn't needed.

### Other information:

## Jetpack product discussion

https://github.com/Automattic/zero-bs-crm/issues/2662

No.

## Testing instructions:

To test the `wp_editor` changes:

* The two `wp_editor` changes affecting the email page affect content added to the editor before you start typing - and in both cases this is set to nothing (so it doesn't need decoding). 
* To confirm the editor isn't affected by the changes, on the email page - `/wp-admin/admin.php?page=zerobscrm-emails` - click on compose and manually add some HTML to the editor. Alongside that, use the editor mark up some text (eg bold, italic).
* Click send and check the content of the email in the sent folder. The HTML that was manually added should show as HTML (unrendered).
* Do the same but from within the thread of a sent email, the results should be the same.
* This behaviour should be the same both before and after this PR is applied. 
* The third `wp_editor` change affects [email templates](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/crm/includes/ZeroBSCRM.AdminPages.php#L1192)
* To test this, visit the email templates page: `wp-admin/admin.php?page=zbs-email-templates&zbs_template_id=1`. 
* Add some acceptable HTML (another bold title for example), and preview the email. The HTML should render as before. 
* Add some unacceptable HTML (a script tag), that will be stripped out. Both before and after previewing this PR, the results should be the same.
* Open a quote proposal in the Client Portal - and the result should be consistent with previous changes (manual HTML not rendered).

To test the removal of the ZBS helper functions:
* On the email page, open a sent email where some manual HTML has been added to an email body. The HTML should now not be rendered (it was before). Using the editor toolbar itself to add the HTML should still render the HTML.
* Check email excerpts - no HTML should be rendered.
* With Quote templates - create a new template and manually add HTML in the visual editor. When adding that quote template to a quote, the HTML should not render. If you add the HTML using the editor, it should render as expected. Placeholders shouldn't be affected.
* Similarly, add manual HTML to the quote editor (visual editor) and save it - it shouldn't render unless using the toolbar to add HTML.


tbradsha:
> All zeroBSCRM_io_WPEditor helper functions have been removed in this PR 

Have you verified none of these functions are called in extensions (including Mail Campaigns), perchance? Hopefully not, but if they are we should probably deprecate them gracefully.

Planning to review this tomorrow; excited to have this cleaned up!

coder-karen:
> Have you verified none of these functions are called in extensions (including Mail Campaigns), perchance? Hopefully not, but if they are we should probably deprecate them gracefully.

Thank you for this reminder! I had not looked - looking now I see there is no mention anywhere in the extensions, but those functions should still be deprecated so I've added them back and deprecated them following the Jetpack monorepo standards. I understand CRM has it's own helper function - ZeroBSCRM_DEPRECATEDMSG - however I think we should move to conforming with the monorepo (added an issue here - https://github.com/Automattic/zero-bs-crm/issues/3075 (and now P2 post here - pbhBOz-3vv-p2 ) ). As that's now up for discussion that might slow down the review of this PR but seems a good conversation to have!

projects/plugins/crm/admin/email/ajax.php
```diff
@@ -141,17 +141,16 @@ function zeroBSCRM_emails_customer_panel() {
 
 	zeroBSCRM_mark_as_read( $threadID );
 
-	global $zbs;
-
-	$e = 0;
+	$e         = 0;
+	$email_ret = array();
```
tbradsha, line 147: 😅

projects/plugins/crm/admin/email/ajax.php
```diff
@@ -141,17 +141,16 @@ function zeroBSCRM_emails_customer_panel() {
...
 	foreach ( $email as $em ) {
 		$email_ret[ $e ]['the_id']          = $em->ID;
 		$email_ret[ $e ]['date']            = zeroBSCRM_locale_utsToDate( $em->zbsmail_created );
 		$email_ret[ $e ]['zbsmail_subject'] = $em->zbsmail_subject;
 		if ( $em->zbsmail_content == null ) {
 			$email_ret[ $e ]['zbsmail_content'] = __( 'No content was stored for this message', 'zero-bs-crm' );
 		} else {
-			$email_ret[ $e ]['zbsmail_content'] = wp_kses( zeroBSCRM_io_WPEditor_DBToHTML( $em->zbsmail_content ), $zbs->acceptable_html );
+			$email_ret[ $e ]['zbsmail_content'] = $em->zbsmail_content;
```
tbradsha: This still needs output escaping (e.g. try adding `<script>alert();</script>` directly in the `wp_zbs_sys_email_hist` table, `zbsmail_content` column).

```suggestion
			$email_ret[ $e ]['zbsmail_content'] = wp_kses( $em->zbsmail_content, $zbs->acceptable_html );
```

projects/plugins/crm/admin/email/main.page.php
```diff
@@ -159,7 +159,7 @@ function jpcrm_render_emailbox() {
 				// echo '<img class="ui avatar image" src="/images/avatar/small/helen.jpg">';
 					echo '<div class="content">';
 					echo '<div class="header">' . esc_html( $email->zbsmail_subject ) . '</div>';
-						echo '<div class="the_content">' . esc_html( zeroBSCRM_io_WPEditor_DBToHTMLExcerpt( $email->zbsmail_content, 200 ) ) . '</div>';
+						echo '<div class="the_content">' . esc_html( wp_html_excerpt( $email->zbsmail_content, 200 ) ) . '</div>';
```
tbradsha, line 162: Didn't know about `wp_html_excerpt()`! Do we still need `esc_html()` in this context?

tbradsha: Shouldn't hurt either way. Looks like WP core has one instance where they use it:

https://github.com/WordPress/WordPress/blob/1d561e1fa6b8f965e039dd18f7029d41296af240/wp-admin/async-upload.php#L66

And several instances where they don't.

projects/plugins/crm/admin/email/main.page.php
```diff
@@ -291,16 +291,15 @@ function jpcrm_render_emailbox() {
...
+				$editor_settings = array(
 					'media_buttons' => false,
 					'quicktags'     => false,
 					'tinymce'       => array(
 						'toolbar1' => 'bold,italic,underline,bullist,numlist,link,unlink,forecolor,undo,redo',
 					),
 					'editor_class'  => 'ui textarea zbs-email-thread',
 				);
-				wp_editor( htmlspecialchars_decode( $content, ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401 ), 'zbs_send_email_thread', $editorSettings );
+				wp_editor( '', 'zbs_send_email_thread', $editor_settings );
```
tbradsha, line 302: 👏

projects/plugins/crm/includes/ZeroBSCRM.DataIOValidation.php
```diff
@@ -90,45 +90,88 @@ function zeroBSCRM_io_sanitizeInt($input=''){
...
+ *
+ * @return string
+ */
+function zeroBSCRM_io_WPEditor_WPEditorToDB( $string = '' ) {
 
-        global $zbs;
+	global $zbs;
 
-        return wp_kses( $string, $zbs->acceptable_html );
+	if ( function_exists( '_deprecated_function' ) ) {
```
tbradsha: Let's get rid of the `function_exists() line`, as it's been around [since WP 2.5](https://developer.wordpress.org/reference/functions/_deprecated_function/#changelog).

tbradsha: But also, I think in this case let's chop 'em. They're doing no one any good. :^)

projects/plugins/crm/includes/ZeroBSCRM.Mail.php
```diff
@@ -375,7 +375,7 @@ function zeroBSCRM_mailDelivery_sendMessage($mailDeliveryAccKey='',$mail=-1){
 					// NOTE: we can apply wpeditor here because the only part of zbs using this 'content' addition is mikes emails
 					//... but this does FORCE us to be using same wp_editor format unless we do a migration
 					//... for now living with that [WH 18/10/18]
-					$emailContent = zeroBSCRM_io_WPEditor_WPEditorToDB($mail['content']);
+					$emailContent = wp_kses( $mail['content'], $zbs->acceptable_html ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
```
tbradsha, line 378: Much better, as it's now clear we're escaping things.

projects/plugins/crm/includes/ZeroBSCRM.MetaBoxes3.QuoteTemplates.php
```diff
@@ -198,7 +197,7 @@ public function save_data( $quoteTemplateID, $quoteTemplate ) {
                 if (isset($_POST['zbs_quotetemplate_content'])) {
 
                     #} Save content
-                    $quoteTemplate['content'] = zeroBSCRM_io_WPEditor_WPEditorToDB($_POST['zbs_quotetemplate_content']);                        
+					$quoteTemplate['content'] = wp_kses( wp_unslash( $_POST['zbs_quotetemplate_content'] ), $zbs->acceptable_html ); //phpcs:ignore WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase,WordPress.Security.NonceVerification.Missing -- to follow up with.
```
tbradsha: Why `wp_unslash()` here? This results in us losing backslashes on save.

projects/plugins/crm/includes/ZeroBSCRM.MetaBoxes3.QuoteTemplates.php
```diff
@@ -155,8 +155,7 @@ public function html( $quoteTemplate, $metabox ) {
                         echo '<p style="text-align:right"><span class="ui basic blue label">'.esc_html__('Did you know: You can now use Quote Placeholders?','zero-bs-crm').' <a href="' . esc_url( $zbs->urls['kbquoteplaceholders'] ) . '" target="_blank">' . esc_html__('Read More','zero-bs-crm') . '</a></span></p>';
                         ##/WLREMOVE
 
-                        #http://stackoverflow.com/questions/3493313/how-to-add-wysiwyg-editor-in-wordpress-meta-box
-                        $content = zeroBSCRM_io_WPEditor_DBToWPEditor($quoteTemplateContent);
+						$content = $quoteTemplateContent; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
```
tbradsha: This is a preexisting issue, but let's use `wp_kses()` to filter output from the database:

```suggestion
						$content = wp_kses( $quoteTemplateContent, $zbs->acceptable_html ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
```

projects/plugins/crm/includes/ZeroBSCRM.MetaBoxes3.Quotes.php
```diff
@@ -450,7 +450,7 @@ public function save_data( $quoteID, $quote ) {
 
                         #} Save content
                         //$data=htmlspecialchars($_POST['zbs_quote_content'], ENT_COMPAT);
-                        $quote['content'] = zeroBSCRM_io_WPEditor_WPEditorToDB($_POST['zbs_quote_content']);                        
+						$quote['content'] = wp_kses( wp_unslash( $_POST['zbs_quote_content'] ), $zbs->acceptable_html ); // phpcs:ignore WordPress.Security.NonceVerification.Missing -- to follow up with.
```
tbradsha: I don't think we should use `wp_unslash()` here either.

projects/plugins/crm/includes/ZeroBSCRM.MetaBoxes3.Quotes.php
```diff
@@ -675,10 +675,12 @@ public function html( $quote, $metabox ) {
...
             $quoteID = -1; if (is_array($quote) && isset($quote['id'])) $quoteID = (int)$quote['id'];
-            $quoteContent = ''; if (is_array($quote) && isset($quote['content'])) $quoteContent = $quote['content'];
+			$quote_content = '';
+		if ( is_array( $quote ) && isset( $quote['content'] ) ) {
+			$quote_content = $quote['content'];
+		}
             
-            #http://stackoverflow.com/questions/3493313/how-to-add-wysiwyg-editor-in-wordpress-meta-box
-            $content = zeroBSCRM_io_WPEditor_DBToWPEditor($quoteContent);
+			$content = $quote_content;
```
tbradsha: As above, let's use `wp_kses()` here. We also should remove `$content` and just pass `$quote_content` to `wp_editor()`.

tbradsha (changes requested):
I took quite some time reviewing the code and its security implications, and left my comments inline. I haven't yet gone through the test steps.

Other than the security concerns, the only bug I saw directly was that we're using `wp_unslash()`, so when saving a quote or quote template, backslashes are unslashed (try saving `\\\\\\\\` in the editor and each save you'll get half the backslashes).

Also, I'd prefer just chopping out those old functions. :^)

projects/plugins/crm/admin/email/ajax.php
coder-karen: Ah yes! Thanks for the reminder, never trust the DB, even if saved content is already escaped.

projects/plugins/crm/admin/email/main.page.php
coder-karen: Ah this is partly to appease PHPCS, which suggests 'All output should be run through an escaping function ' for this line. And yes, like you spotted it also appears to be in use in WordPress core 😄  The main difference appears to be that `esc_html` also converts some additional special characters into their HTML entities, particularly quotes.
Could add an ignore for PHPCS but also like you say it doesn't hurt in this case with the extra escape function.

projects/plugins/crm/includes/ZeroBSCRM.DataIOValidation.php
coder-karen: I was inspired by the [extra check that WP Super Cache has been implementing ](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/super-cache/wp-cache.php#L139) when using that function. Happy to remove them!

> But also, I think in this case let's chop 'em. 

😱 

All four functions? I'd love to follow the monorepo guidelines, but will also follow your lead on this if that's what you prefer! Maybe we need some more engagement on the post for a team decision?

projects/plugins/crm/includes/ZeroBSCRM.MetaBoxes3.QuoteTemplates.php
coder-karen: Also a PHPCS suggestion in this case, but I didn't spot the implications. There are additional slashes being added here, but they are removed before output in the editor (so wp_unslash is duplicating the removal). I think the best thing is to ignore that sniff here.

projects/plugins/crm/includes/ZeroBSCRM.MetaBoxes3.Quotes.php
coder-karen: Ah yes, same issue.

projects/plugins/crm/includes/ZeroBSCRM.MetaBoxes3.Quotes.php
coder-karen: Good points, thank you.

coder-karen:
Thank you @tbradsha , appreciate it!

I've followed up / commented on all your reviews - in the end I did also delete those helper functions (partly as I also know they can be added back in before the next release if needed, depending on any further discussion on the deprecation post!) 😄

projects/plugins/crm/includes/ZeroBSCRM.MetaBoxes3.QuoteTemplates.php
```diff
@@ -1,4 +1,5 @@
 <?php 
+//phpcs:disable
```
tbradsha: Was this added intentionally?

tbradsha (dismissed):
Looks good, though there appears perhaps to be a stray `phpcs:ignore` at the top of `includes/ZeroBSCRM.MetaBoxes3.QuoteTemplates.php`?

projects/plugins/crm/includes/ZeroBSCRM.MetaBoxes3.QuoteTemplates.php
coder-karen: It is stray, thanks for spotting that.
However having committed it, I now can't remove it without the whole file being flagged for linting issues.
I've added the commit anyway, but will now just need to bypass the check in order to merge the PR.

coder-karen:
> Looks good, though there appears perhaps to be a stray phpcs:ignore at the top of includes/ZeroBSCRM.MetaBoxes3.QuoteTemplates.php?

Yes it was a stray unfortunately ([left a reply here](https://github.com/Automattic/jetpack/pull/30306#discussion_r1181703371)) - which means we'll need to bypass the linter checks that will fail now (to merge the PR).

tbradsha:
> we'll need to bypass the linter checks that will fail now

Looks like they still passed?

coder-karen:
> Looks like they still passed?

🥳 

That's good! I know it happened before some months ago but must have been with a merged PR, so it was spotted in time here. Locally PHPCS spat out a page of issues though hence the no-verify.

tbradsha (approved):

PR #30311, opened 2023-04-26
Jetpack Social: Disable Media Picker in sidebar when Social Image Generator is active

danielpost:
With this change we disable the media picker if SIG is enabled, to handle some confusion while we are not implementing the new UI.

More here: pdrWKz-TB-p2

## Proposed changes:

* Updated media section so it accepts a notice and a disabled prop
* Set the disabled prop based on SIG availability.

### Other information:

## Jetpack product discussion

pdrWKz-TB-p2

No.
## Testing instructions:

* Open the editor with the Advanced plan.
* You should no be able to pick/remove/interact with the media picker if SIG is enabled. Turning SIG off should re-enable the picker.

https://user-images.githubusercontent.com/36671565/234886786-35759da8-678e-4c01-a967-b1514609a0b3.mp4


projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -188,7 +190,19 @@ export default function PublicizeForm( {
 							/>
 						</>
 					) }
-					{ isEnhancedPublishingEnabled && <MediaSection /> }
+					{ isEnhancedPublishingEnabled && (
+						<MediaSection
+							disabled={ isImageGeneratorEnabled }
```
pablinos: This is definitely an odd state, but I published a post with SIG enabled and then turned the feature off. That meant that `isImageGeneratorEnabled` here is true, but the panel is hidden as isSocialImageGeneratorEnabled from `usePublicizeConfig` is false.

Do you think we should try and use both of these values? It seems like useImageGeneratorConfig shouldn't come back as enabled if the feature isn't available, although that could mean people lose the OG images if they edit a post after letting the Advanced plan lapse.

image

pablinos (commented):
I think this is probably fine as is, but I just wondered about the 'enabled' flags.

projects/js-packages/publicize-components/src/components/form/index.js
pablinos: I've added this change in [1996bbd](https://github.com/Automattic/jetpack/pull/30311/commits/1996bbd0840814208c86e4640d14fc9f7ada949b), but it's possibly not a situation that will ever occur!

projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -188,7 +192,19 @@ export default function PublicizeForm( {
 							/>
 						</>
 					) }
-					{ isEnhancedPublishingEnabled && <MediaSection /> }
+					{ isEnhancedPublishingEnabled && (
+						<MediaSection
+							disabled={ isSocialImageGeneratorEnabled && isImageGeneratorEnabled }
```
spsiddarthan: Super minor nitpick and we can address it later, `isImageGeneratorEnabled` should perhaps be renamed to `isImageGeneratorSelected`, the word 'enabled' makes me think if the feature is enabled or not.

spsiddarthan (dismissed):
The feature works as intended. Just added a minor comment. I had to enable SIG from the admin page to test this. If the default state of SIG is disabled, without the admin screen in Jetpack, we can't really offer it to users on Jetpack, but we know that already, don't we?

projects/js-packages/publicize-components/src/components/form/index.js
pablinos: Yeah I was a bit concerned about these names too. Perhaps `isImageGeneratorEnabledForPost`?

pablinos (approved):
LGTM!

zinigor:
Cherry picked in 339299f14113b7c929f93030672328251acbe9c9.

PR #30313, opened 2023-04-26
Remove builds on `pnpm install`

anomiex:
## Proposed changes:

It's annoying to have to build these on `pnpm install`, and slows down the install unnecessarily if the built packages aren't going to be used. Let's fix that.

To keep the actual npmjs packages usable, we still need external users to get the built versions while getting the sources within the monorepo. package.json exports custom conditions are a way to do that: we specify a custom condition "jetpack:src" first pointing at the source index file, then "types" for the index.d.ts where appropriate, then "default" for the built index.js.

Then we have to configure tooling inside the monorepo to look for that "jetpack:src" condition:

* Webpack uses `.resolve.conditionNames`. But in case we ever publish our webpack-config package, let's avoid the problem we had with calypso-build and their "calypso:src" condition by setting an environment variable via `.npmrc` and using that instead of hard-coding 'jetpack:src'.
  * Unlike Calypso, we're not going to also set `.resolve.mainFields`. Exports are the modern way to do things like this.
* Typescript's tsc uses `.compilerOptions.customConditions`. Plus we also have to set `.compilerOptions.moduleResolution` to "bundler" for it to pay attention to exports at all (or "nodenext", but that has a lot of other side effects).
* Jest (at least for the jsdom and node environments) uses `.testEnvironmentOptions.customExportConditions`.
* Eslint's `eslint-plugin-import` needs a custom resolver (sigh).
* Rollup uses `exportConditions` in the options for `@rollup/plugin-node-resolve`.

Then, for good measure, let's also have the cli strip the "jetpack:src" from the mirror repos to make it even less likely that some external user gets confused by it.

### Other information:

## Jetpack product discussion

None, but p1671775119411749-slack-C01U2KGS2PQ is relevant.

No

## Testing instructions:

* In a fresh checkout, run `pnpm install`. There should be no building of js-packages/image-guide or js-packages/svelte-data-sync-client anymore.
* In that same checkout, run `pnpm lint-required`. It should pass.
* In that same checkout, run `jetpack build -v plugins/boost`. It should succeed.
* Check that the CI linting job is happy.
* Check the build artifacts, the ones for image-guide and svelte-data-sync-client should not contain "jetpack:src" entries in package.json exports.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -8,3 +8,6 @@ engine-strict = true
 # No hoisting by default.
 hoist-pattern=[]
 public-hoist-pattern=[]
+
+# @automattic/jetpack-webpack-config looks for this.
+jetpack-webpack-config-resolve-conditions=jetpack:src
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -3722,6 +3722,9 @@ importers:
       eslint-config-prettier:
         specifier: 8.8.0
         version: 8.8.0(eslint@8.39.0)
+      eslint-import-resolver-exports:
+        specifier: 1.0.0-beta.5
+        version: 1.0.0-beta.5(eslint-plugin-import@2.27.5)(eslint@8.39.0)
       eslint-plugin-es:
         specifier: 4.1.0
         version: 4.1.0(eslint@8.39.0)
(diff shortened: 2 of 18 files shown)


zinigor:
Does what it says on the tin, and the tests seems to pass, except for the Boost e2e test. I've re-run it to make sure.

zinigor (approved):
Boost seems to be OK based on the tests!

PR #30317, opened 2023-04-26
Fix array diff key warnings

### LINKED ISSUE #30304: Jetpack Forms - Potential issue with responses endpoint (Inbox not loading) - PHP warnings
[Slack conversation](p1682348735794039-slack-C01CSBEN0QZ)

I noticed an issue with our responses endpoint for the new form responses inbox. My main inbox was not loading responses, even though the Spam and Trash worked as expected, as well as the "classic view".

It seems that this line (https://github.com/Automattic/jetpack/blame/trunk/projects/packages/forms/src/class-wpcom-rest-api-v2-endpoint-forms.php#L168) is throwing some PHP warnings, I wonder if it could be related to the data structure changes we made for new responses?

```
<b>Warning</b>: array_diff_key(): Expected parameter 1 to be an array, null given in <b>/home/wpcom/public_html/wp-content/mu-plugins/jetpack-plugin/production/jetpack_vendor/automattic/jetpack-forms/src/class-wpcom-rest-api-v2-endpoint-forms.php</b> on line <b>171</b><br />
[public-api.wordpress.com/wpcom/v2/sites/204035811/forms/responses?_envelope=1&amp;limit=50&amp;offset=0&amp;status=inbox&amp;_gutenberg_nonce=9f929356d8&amp;_locale=user]<br />
```

CGastrell:
Ensure array is provided for array_diff_key, fix status handling when untrashing.

Fixes #30304 

## Proposed changes:
This PR ensures an array is provided as argument for array_diff_key as we've been seeing warnings about it. ~The underlying cause is still TBD, couldn't reproduce.~

Marking as `spam` was unescaping content, making JSON strings not decodable. This PR changes the use of `wp_insert_post` for `wp_update_post` and canceling any after insert actions. **Should this be a concern?**

It also adds a `wp_untrash_post_status` filter handler, so untrashing a feedback sends it back to where it was before trashing. Defaults to `publish`

### Other information:

No

## Testing instructions:
Submit responses to some form, make sure to use double quotes `"` and strange characters (tilde is fine, á, ê, ï, etc).
Back at inbox, move those feedbacks to/from inbox/spam/trash.

See that the feedbacks don't break (feedbacks not showing basic properties, like URL or detailed fields).
See that moving out from trash (on legacy view) doesn't turn feedbacks into `drafts` (you can trash your drafts and then untrash to see them go into the inbox)

### DIFF
--- a/projects/packages/forms/changelog/fix-array-diff-key-warnings
+++ b/projects/packages/forms/changelog/fix-array-diff-key-warnings
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Ensure array is provided to array_diff_key to avoid warnings
--- a/projects/packages/forms/src/class-wpcom-rest-api-v2-endpoint-forms.php
+++ b/projects/packages/forms/src/class-wpcom-rest-api-v2-endpoint-forms.php
@@ -371,7 +371,7 @@ class WPCOM_REST_API_V2_Endpoint_Forms extends WP_REST_Controller {
 			$status = wp_update_post(
 				array(
 					'ID'          => $post_id,
-					'post_status' => 'published',
+					'post_status' => 'publish',
 				),
 				false,
 				false
--- a/projects/packages/forms/src/class-wpcom-rest-api-v2-endpoint-forms.php
+++ b/projects/packages/forms/src/class-wpcom-rest-api-v2-endpoint-forms.php
@@ -175,7 +175,7 @@ class WPCOM_REST_API_V2_Endpoint_Forms extends WP_REST_Controller {
 					'entry_permalink'         => $data['all_fields']['entry_permalink'],
 					'subject'                 => $data['_feedback_subject'],
 					'fields'                  => array_diff_key(
-						$data['all_fields'],
+						empty( $data['all_fields'] ) ? array() : $data['all_fields'],
 						array(
 							'email_marketing_consent' => '',
 							'entry_title'             => '',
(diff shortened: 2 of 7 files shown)


digitalwaveride:
My issue still exists... when switching to a the new inbox view, very briefly it flickers, and seems to load... but then getting the white screen.

- not sure if it was expected to go away with this fix:

image


image


image


On Classic View things work as expected.

digitalwaveride (changes requested):
Christian, thank you so much for the digging and finding the issue! 🙇 

I can load now all tabs, even for the "incomplete" remaining responses with the lost detail data 👍 

One thing I noticed though:

Inside of the spam tab, the functionality of the "remove spam" action changed.
It seems to delete this responses now permanently.
image

What it actually should do (like on Trunk currently) it should move the response back to the inbox.

You mentioned already before that this label naming might be not intuitive, so I would suggest renaming the label - like in the classic view - from "Remove from spam" to "Not spam".

image

image

https://user-images.githubusercontent.com/5120357/235120649-3676584e-3e12-4fbf-9e5f-814b9051de58.mov

digitalwaveride (changes requested):
GitHub 🤷 
https://github.com/Automattic/jetpack/pull/30317#pullrequestreview-1405605457

digitalwaveride (approved):
Thanks a ton Christian!!!

- Trash with "special responses" works! 👍 
- "Remove from Spam" behavior is as expected 👍 
- "Copy edit" is there 👍

PR #30318, opened 2023-04-26
[RNMobile] Update Android VP player loading screen

jhnstn:
Follow up to https://github.com/Automattic/jetpack/pull/30251 to fix the loading screen on Android.

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*


projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
```diff
@@ -88,6 +90,20 @@ export default function Player( { isSelected, attributes } ) {
...
 
+	// Android doesn't receive the videopress_ready event,
+	// so we need to set the player as ready after a delay.
+	useEffect( () => {
+		if ( IS_ANDROID && !! html ) {
+			androidPlayerReadyTimer.current = setTimeout( () => {
+				if ( ! isPlayerLoaded ) {
+					setIsPlayerReady( true );
+				}
+			}, 500 );
```
jhnstn, line 101: This gives the embed a half a second to load. It's not ideal since it will not be the same on slower connections but we can increase the timeout. 500ms seems like a good starting point if we choose this route

PR #30320, opened 2023-04-26
Make the Donation Block Stipe Connect Link the Primary Color

### LINKED ISSUE #30300: Payments block: Stripe connection button visually broken
### Impacted plugin

Jetpack

### Quick summary

At least in Lettre theme, the Stripe banner has visually broken "connect" button:

image

### Steps to reproduce

1. on site with stripe disconnected, and Lettre theme
1. add payments block in the page editor
1. Note how stripe banner is broken

_No response_

### What actually happened

_No response_

### Impact

Some (< 50%)

### Available workarounds?

Simple

### Logs or notes

- Chrome

I've pinged Team Gold in p81Rsd-1hl-p2#comment-1436

romarioraffington:
Fixes https://github.com/Automattic/jetpack/issues/30300

## Proposed changes:

* Make the Donation Block Stipe Connect Link the Primary Color

| Before | After |
| Screenshot 2023-04-26 at 5 40 25 PM | Screenshot 2023-04-26 at 5 33 15 PM |

### Other information:

## Testing instructions:

* Sync changes to WPCOM
* Sandbox `romarioraffearn.wordpress.com`
* Go to https://wordpress.com/page/romarioraffearn.wordpress.com/242, edit the post and verify the Stripe "Connect" link is visible for both instances of the Donation block on the page

### DIFF
--- a/projects/plugins/jetpack/extensions/shared/components/stripe-nudge/style.scss
+++ b/projects/plugins/jetpack/extensions/shared/components/stripe-nudge/style.scss
@@ -16,6 +16,6 @@
 	padding: 6px;
 }
 
-..jetpack-stripe-nudge__banner span a.jetpack-stripe-nudge__link {
+.jetpack-stripe-nudge__banner span a.jetpack-stripe-nudge__link {
 	color: var( --color-primary ) !important;
 }
--- a/projects/plugins/jetpack/extensions/shared/components/stripe-nudge/style.scss
+++ b/projects/plugins/jetpack/extensions/shared/components/stripe-nudge/style.scss
@@ -16,6 +16,6 @@
 	padding: 6px;
 }
 
-.jetpack-stripe-nudge__banner span.block-editor-warning__action a {
-	color: var( --color-primary );
+.jetpack-stripe-nudge__banner .jetpack-stripe-nudge__link {
+	color: #0675c4 !important;
 }
--- a/projects/plugins/jetpack/extensions/shared/components/stripe-nudge/style.scss
+++ b/projects/plugins/jetpack/extensions/shared/components/stripe-nudge/style.scss
@@ -17,5 +17,5 @@
 }
 
 .jetpack-stripe-nudge__banner span a.jetpack-stripe-nudge__link {
-	color: var( --color-primary ) !important;
+	color: var( --color-primary );
 }
--- a/projects/plugins/jetpack/extensions/shared/components/stripe-nudge/style.scss
+++ b/projects/plugins/jetpack/extensions/shared/components/stripe-nudge/style.scss
@@ -16,6 +16,6 @@
 	padding: 6px;
 }
 
-.jetpack-stripe-nudge__banner .jetpack-stripe-nudge__link {
+..jetpack-stripe-nudge__banner span a.jetpack-stripe-nudge__link {
 	color: var( --color-primary ) !important;
 }
--- a/projects/plugins/jetpack/extensions/shared/components/stripe-nudge/style.scss
+++ b/projects/plugins/jetpack/extensions/shared/components/stripe-nudge/style.scss
@@ -17,5 +17,5 @@
 }
 
 .jetpack-stripe-nudge__banner .jetpack-stripe-nudge__link {
-	color: #0675c4 !important;
+	color: var( --color-primary ) !important;
 }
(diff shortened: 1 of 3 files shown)


projects/plugins/jetpack/extensions/shared/components/stripe-nudge/style.scss
```diff
@@ -15,3 +15,7 @@
 	margin-right: 16px;
 	padding: 6px;
 }
+
+.jetpack-stripe-nudge__banner span.block-editor-warning__action a {
```
simison: Is it possible to add custom class to the button, because `span.block-editor-warning__action a` could get changed over time in Gutenberg?

romarioraffington: Thanks for the suggestion! I couldn't find a way to target the `span`, however, I added a CSS class to the link and made the CSS more robust. Let me know if that works for you.

See [d674f3b](https://github.com/Automattic/jetpack/pull/30320/commits/d674f3bfed112de2876360855e4f1c83bf51b70d)

projects/plugins/jetpack/extensions/shared/components/stripe-nudge/style.scss
```diff
@@ -15,3 +15,7 @@
 	margin-right: 16px;
 	padding: 6px;
 }
+
+.jetpack-stripe-nudge__banner .jetpack-stripe-nudge__link {
+	color: #0675c4 !important;
```
millerf: I think the !important should not be needed, right?

millerf (dismissed):
LGTM

projects/plugins/jetpack/extensions/shared/components/stripe-nudge/style.scss
romarioraffington: It is needed as the editor has CSS that is more specific that'll override this

projects/plugins/jetpack/extensions/shared/components/stripe-nudge/style.scss
```diff
@@ -16,6 +16,6 @@
 	padding: 6px;
 }
 
-.jetpack-stripe-nudge__banner span.block-editor-warning__action a {
-	color: var( --color-primary );
+.jetpack-stripe-nudge__banner .jetpack-stripe-nudge__link {
+	color: #0675c4 !important;
```
simison: Is there no sass variable instead of HEX?

projects/plugins/jetpack/extensions/shared/components/stripe-nudge/style.scss
simison: You could add [specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) to overcome it and not rely on `!important` which is a bit blunt. A bit of a hack is to do `.jetpack-stripe-nudge__link.jetpack-stripe-nudge__link` or include `+ span` or `span.class`

millerf: I was asking because I thought you [created a custom class](https://github.com/Automattic/jetpack/pull/30320#discussion_r1179310488), and I thought that would be specific enough.

projects/plugins/jetpack/extensions/shared/components/stripe-nudge/style.scss
romarioraffington: Ah yes, I copied this directly from the browser while testing. Thanks for catching this. Fixed in [be1b85a](https://github.com/Automattic/jetpack/commit/be1b85a3a22b0eacf363db4b27251904b3879a2d)

projects/plugins/jetpack/extensions/shared/components/stripe-nudge/style.scss
romarioraffington: Thanks for the feedback!

> You could add [specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) to overcome it and not rely on `!important which is a bit blunt. A bit of a hack is to do .jetpack-stripe-nudge__link.jetpack-stripe-nudge__link

`jetpack-stripe-nudge__link.jetpack-stripe-nudge__link` would not work in this case

> + span or span.class

Umm, is `span.class` similar to the initial approach above or are you referring to a different `span.css` class? ` + span` would not be specific enough in this case as well.

Made the CSS more specific without `!important`.

jeherve (approved):

PR #30328, opened 2023-04-27
Extensions: add blogroll block

heavyweight:
## Proposed changes:

* Introduce a new Blogroll block

### Other information:

## Jetpack product discussion

* 1721-gh-Automattic/dotcom-forge
 
* No

## Testing instructions:

* TBD


jeherve (changes requested):
The classic blogroll is back! ❤️ 

While you're still working on the block itself, what would you think about adding only the structure of the new block in a first PR, with nothing in it just yet, using the scaffold command?
`jetpack docker wp jetpack scaffold block blogroll`

Splitting the work into multiple PRs (one that adds the structure, and then multiple iterations adding different parts of the interface) would make PR reviews easier for both your team and anyone following your work in the monorepo. It would also ensure we do not forget anything when adding the block structure.

heavyweight:
> The classic blogroll is back! ❤️
> 
> While you're still working on the block itself, what would you think about adding only the structure of the new block in a first PR, with nothing in it just yet, using the scaffold command? `jetpack docker wp jetpack scaffold block blogroll`
> 
> Splitting the work into multiple PRs (one that adds the structure, and then multiple iterations adding different parts of the interface) would make PR reviews easier for both your team and anyone following your work in the monorepo. It would also ensure we do not forget anything when adding the block structure.

Thanks for the feedback. Much appreciated.

I will have to close this PR.
We will move to use new endpoints instead of legacy.
The block is out for now but will come back stronger.

PR #30338, opened 2023-04-27
Photon: remove unused method

jeherve:
## Proposed changes:

This hasn't been used since 2013 in 9b86a288d4e07c385109d931fcb31128d7e7622d.

We do not use that method on WordPress.com either.

We also do not need that method, since we already get Photonized images when we get images from the `Jetpack_PostImages` class (core functions like `wp_get_attachment_image_src` return a photonized image).

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

* Go to Jetpack > Settings > Performance and enable the site accelerator.
* Go to Jetpack > Settings > Social and enable sharing buttons.
* Create a new post, and add a featured image to it.
* View the post on the frontend, and view source.
* Find the `og:image` meta. 
    * It should use Photon.
* Check your error logs; you should not see any notices.


projects/packages/image-cdn/changelog/rm-unused-photon-method
```diff
@@ -0,0 +1,4 @@
+Significance: patch
```
anomiex: This should be a "major" change, since if any third-party user of the package was calling the public method then they'll break.

Note that since the package is still at a 0.x version (0.1.0), a major change will still only do a minor bump (to 0.2.0). That's expected.

anomiex (commented):
Confirmed that this function doesn't seem to be used anywhere.

projects/packages/image-cdn/changelog/rm-unused-photon-method
jeherve: 👍 done in 56537d16fb1dae37c5b97b77d2a5af688a57dcc5

anomiex (approved):

PR #30339, opened 2023-04-27
beta plugin: Fix markdown rendering

anomiex:
## Proposed changes:

The markdown library we're using seems to get confused by HTML-style tags and might let some through that shouldn't be let through. Since it has a flag to disable HTML-style tags entirely, and we don't use those much in PR summaries or the relevant bundled .md files, let's set that flag to avoid the problem.

### Other information:

## Jetpack product discussion

p1682615499416869-slack-CBG1CP4EN

No

## Testing instructions:

* Stuff still renders at the bottom of pages like `/wp-admin/admin.php?page=jetpack-beta&plugin=jetpack`?

### DIFF
--- a/projects/plugins/beta/changelog/fix-beta-plugin-markdown-render
+++ b/projects/plugins/beta/changelog/fix-beta-plugin-markdown-render
@@ -1,0 +1,4 @@
+Significance: patch
+Type: security
+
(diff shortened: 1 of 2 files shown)


samiff (approved):
Still able to see Markdown to-test info rendered as expected.

PR #30340, opened 2023-04-27
Repo Gardening: automatically add legacy Photon label

jeherve:
## Proposed changes:

Photon now lives on in the Image CDN package. We can use the new "[Package] Image CDN" label to look for changes to the feature, but the existing Photon label is still widely used. Let's add that legacy label to PRs that touch the package, so we can continue to use the Photon label for sorting through issues and PRs in the repo.

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

This is better tested in a fork of this repo. I'll add a link below:

* https://github.com/jeherve/jetpack/pull/94

### DIFF
--- a/projects/github-actions/repo-gardening/changelog/add-photon-legacy-label
+++ b/projects/github-actions/repo-gardening/changelog/add-photon-legacy-label
@@ -1,0 +1,4 @@
+Significance: patch
+Type: added
+
+Add the legacy Photon label for PRs that make changes to the Image CDN package.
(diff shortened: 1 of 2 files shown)


anomiex (approved):
At some point we might want to look at moving both this and the isDraft check outside of the `files.map()` callback, as neither depends on the individual files, but for now this works.

jeherve:
> At some point we might want to look at moving both this and the isDraft check outside of the `files.map()` callback, as neither depends on the individual files, but for now this works.

That's a good point. Done in #30350

PR #30353, opened 2023-04-28
Fix issues with loading placeholders on the forms dashboard

ice9js:
## Proposed changes:

This patch fixes some minor issues introduced with #30213 that I caught while working on other bugs. Namely:

- Fixed the equation that'd still cause an incorrect number of placeholders to be shown in some cases. Particularly when there's multiple pages of responses.
- Updated the handlers for single- and bulk-action menus to ensure they wait for the `fetchResponses` action to finish before setting `loading` to `false`.
- Old responses will now be cleared immediately when changing pages which will prevent unwanted animations in this case.

### Other information:

No.

## Testing instructions:

While testing you can set `RESPONSES_FETCH_LIMIT` constant to something low which will help notice the issues mentioned above.

- Go to `admin.php?page=jetpack-forms` and make sure there are multiple pages worth of responses.
- Try moving items between different tabs using bulk actions.
- You should always see the exact number of placeholders replace the items you just moved.
- When the placeholders disappear, they should immediately be replaced by responses. Previously there was a noticeable delay.
- There shouldn't be any buggy animations on the response list when switching pages.

### DIFF
--- a/projects/packages/forms/changelog/fix-forms-dashboard-loading-placeholders
+++ b/projects/packages/forms/changelog/fix-forms-dashboard-loading-placeholders
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fixed buggy behavior of loading placeholders in the forms dashboard.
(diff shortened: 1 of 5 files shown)


jcheringer (approved):
Nice! 👍

PR #30354, opened 2023-04-28
Fix responses endpoint returning the wrong totals when 'offset' is set

### LINKED ISSUE #30332: JP Forms - WPAdmin - Issue with the "tab counter", when you use pagination
Originally [reported here](pcsBup-1aJ-p2#comment-421):

> When I go to the Spam view, and then use the navigation option at the bottom of the screen to load more spam entries, the counter for the Inbox resets:

Video: https://video.wordpress.com/v/6Y4zg3dc

-> Inbox and Trash counter go to zero:

Image

ice9js:
Fixes #30332.

## Proposed changes:

This PR fixes an issue where incorrect totals would be reported on the tabs in the forms dashboard when the results contain more than one page and any but the first page is selected.

This was due to `offset` and `limit` properties on the API request being appended to every query, causing it to also offset the numbers for the other tabs.

By setting `posts_per_page` to `-1` for queries where we're only fetching totals, `WP_Query` will ignore the `offset` param as well and return the correct counts.

### Other information:

No.

## Testing instructions:

While testing you can set `RESPONSES_FETCH_LIMIT` constant to something low which will help notice the issues mentioned above.

- Go to `/wp-admin/admin.php?page=jetpack-forms` and make sure there are multiple pages worth of responses.
- Try switching between different pages using page navigation.
- The totals on the tabs in the header should remain constant regardless of what page you're on.

### DIFF
--- a/projects/packages/forms/changelog/fix-forms-dashboard-tab-totals
+++ b/projects/packages/forms/changelog/fix-forms-dashboard-tab-totals
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fixed invalid totals being reported for different tabs in the forms dashboard.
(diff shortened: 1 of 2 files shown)


jcheringer (approved):
Works for me 👍

PR #30355, opened 2023-04-28
[Newsletter] Return early to prevent spamming DB for post access

millerf:
## Proposed changes:
* Return early to prevent spamming DB for post access

### Other information:
* See p1682634536650749-slack-C0299DMPG

 No

## Testing instructions:
- Create a post with a newsletter access level
- Verify the post is correctly gated
- Repeat steps 1 -2 with the other newsletter access level


projects/plugins/jetpack/modules/memberships/class-jetpack-memberships.php
```diff
@@ -480,7 +481,13 @@ public static function user_can_view_post_nocache() {
 		require_once JETPACK__PLUGIN_DIR . 'extensions/blocks/premium-content/_inc/subscription-service/include.php';
 
 		$newsletter_access_level = self::get_newsletter_access_level();
-		$paywall                 = \Automattic\Jetpack\Extensions\Premium_Content\subscription_service();
+
+		if ( null === $newsletter_access_level || Token_Subscription_Service::POST_ACCESS_LEVEL_EVERYBODY === $newsletter_access_level ) {
```
artpi: I am not confident in `Token_Subscription_Service ` being available here

artpi (dismissed):
Yup, code looks good - it could use some thorough testing though :)

projects/plugins/jetpack/modules/memberships/class-jetpack-memberships.php
millerf: It is included in line 481

```
```

millerf: See this: https://github.com/Automattic/jetpack/blob/94ddae8284758db97529fad30752b9e11a556bb1/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/include.php#L17

artpi: Seems ok then :)

projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-wpcom-offline-subscription-service.php
```diff
@@ -46,7 +46,13 @@ public function subscriber_can_receive_post_by_mail( $user_id, $post_id ) {
 		$previous_user = wp_get_current_user();
 		wp_set_current_user( $user_id );
 
-		$access_level       = get_post_meta( $post_id, '_jetpack_newsletter_access', true );
+		$access_level = get_post_meta( $post_id, '_jetpack_newsletter_access', true );
+
+		if ( null === $access_level || self::POST_ACCESS_LEVEL_EVERYBODY === $access_level ) {
```
romarioraffington: Would `empty` be better to use here? Could `$access_level` be an empty string as well?

anomiex: Just do `! $access_level` instead of `empty( $access_level )` if that's the direction you want to go with it.

`empty( $access_level )` does the same thing as `( ! isset( $access_level ) || ! $access_level )`, but we know `$access_level` is already set because we just set it above.

zinigor:
Hey @millerf ! Are you planning to include @anomiex 's suggestion? I have removed the Needs Review label - if not, please add it back, so we can continue with review and testing.

millerf:
@zinigor Can I get a check on this?

millerf:
See https://github.com/Automattic/jetpack/pull/30807 for more details

anomiex (approved):
Code looks ok to me. I have not tested though, someone should do that before this is actually merged.

n3f:
Related: D111411-code

PR #30356, opened 2023-04-28
Don't re-send emails when marking responses as not-spam

### LINKED ISSUE #30305: Jetpack Forms - Email notifications - when responses get restored from Spam/Trash a new email notification is triggered
Email notifications - when responses get restored from Spam/Trash a new email notification is triggered

This seems to be a longstanding issue, nevertheless is not really expected and can be confusing for users. Especially when a lot of responses are being moved or in the context of a multi-editor site, when other users triggering these changes, and the creator of the form is getting emails about old responses (-> risk of double orders / double bookings).

ice9js:
Fixes #30305.

## Proposed changes:

This patch will prevent emails from being send when items are being restored from spam on the new forms dashboard.

### Other information:

No.

## Testing instructions:

- Go to `/wp-admin/admin.php?page=jetpack-forms`.
- Make sure there are some responses with valid emails (that you have access to) and contact details and mark some of them as spam.
- Go to the spam tab.
- Select those responses and remove them from spam.
- You should not be getting any emails.


CGastrell (approved):
LGTM!

PR #30358, opened 2023-04-28
Jetpack Social: Always render Social Image Generator panel when it is available

danielpost:
## Proposed changes:

This PR fixes the Social Image Generator panel to always be rendered when the feature is available. The toggle on the Jetpack Social admin page now simply decides whether or not SIG is enabled by default for new posts.

### Other information:

No.

## Testing instructions:

1. Create a new test site with Jetpack and Jetpack Social.
2. Disable Jetpack Social, make sure only Jetpack is enabled.
3. Add Social Image Generator to your site.
4. Go to the "New Post" screen and verify that the SIG panel is visible but toggled off.
5. Do NOT toggle it on and save the post.
6. Verify that SIG is disabled and no token was generated: `wp post meta list POST_ID`.
7. Create a new post, toggle SIG on and verify that a token is generated..
8. Enable Jetpack Social and disable Jetpack.
9. Go to the Social admin page and check that the SIG toggle is off.
10. Repeat steps 4-7.
11. Turn on the SIG toggle on the admin page, create a new post and verify that SIG is toggled on for the post.
12. Save the post and check that SIG meta was added.

### DIFF
--- a/projects/packages/publicize/src/social-image-generator/class-setup.php
+++ b/projects/packages/publicize/src/social-image-generator/class-setup.php
@@ -73,7 +73,9 @@ class Setup {
 			return;
 		}
 
-		if ( ! $publicize->has_social_image_generator_feature() ) {
+		$settings = new Settings();
+
+		if ( ! $settings->is_available() ) {
 			return;
 		}
 
@@ -81,14 +83,12 @@ class Setup {
 			return;
 		}
 
-		if ( ! ( new Settings() )->is_enabled() ) {
-			return;
-		}
-
+		// Set SIG to be enabled by default for new posts if the toggle is on.
 		$post_settings = new Post_Settings( $post_id );
 		if (
 			! $update &&
 			'auto-draft' === $post->post_status &&
+			$settings->is_enabled() &&
 			empty( $post_settings->get_settings( true ) )
 		) {
 			$post_settings->update_setting( 'enabled', true );
@@ -96,6 +96,10 @@ class Setup {
 		}
 
 		if ( $post->post_status === 'auto-draft' ) {
+			return;
+		}
+
+		if ( ! $post_settings->is_enabled() ) {
 			return;
 		}
 
--- a/projects/js-packages/publicize-components/changelog/fix-jetpack-social-render-sig-panel-even-when-disabled-by-default
+++ b/projects/js-packages/publicize-components/changelog/fix-jetpack-social-render-sig-panel-even-when-disabled-by-default
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Jetpack Social: Render Social Image Generator panel even when SIG's default is disabled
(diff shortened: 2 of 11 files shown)


projects/packages/publicize/src/social-image-generator/class-setup.php
```diff
@@ -73,22 +73,22 @@ public function generate_token_on_save( $post_id, $post, $update ) {
 			return;
 		}
 
-		if ( ! $publicize->has_social_image_generator_feature() ) {
-			return;
-		}
+		$settings = new Settings();
 
-		if ( wp_is_post_autosave( $post ) || wp_is_post_revision( $post_id ) ) {
+		if ( ! $settings->is_available() ) {
```
danielpost, line 78: This is the same as `$publicize->has_social_image_generator_feature()`.

projects/packages/publicize/src/social-image-generator/class-setup.php
```diff
@@ -73,22 +73,22 @@ public function generate_token_on_save( $post_id, $post, $update ) {
...
 			return;
 		}
 
-		if ( ! ( new Settings() )->is_enabled() ) {
```
danielpost, line 84: This prevents us from generating a token if the toggle on the admin page is off, so it needs to be removed.

projects/packages/publicize/src/social-image-generator/class-setup.php
```diff
@@ -73,22 +73,22 @@ public function generate_token_on_save( $post_id, $post, $update ) {
...
+		if ( wp_is_post_autosave( $post ) || wp_is_post_revision( $post_id ) ) {
 			return;
 		}
 
+		// Set SIG to be enabled by default for new posts if the toggle is on.
 		$post_settings = new Post_Settings( $post_id );
 		if (
 			! $update &&
 			'auto-draft' === $post->post_status &&
+			$settings->is_enabled() &&
```
danielpost, line 91: This line ensures that if the SIG toggle is enabled on the admin page, SIG gets enabled by default for new posts.

gmjuhasz:
This looks good and tests well, I just have a question, not sure if it should work like this:

When saving a post as draft with SIG off like in step 4, after that turning on SIG and saving the post again won't add the token to the meta. It is there if I save the post the first time with SIG on, but not if it's saved as off before

pablinos:
> When saving a post as draft with SIG off like in step 4, after that turning on SIG and saving the post again won't add the token to the meta. It is there if I save the post the first time with SIG on, but not if it's saved as off before

This is working ok for me.

The possible issue with this is that with it toggled off we're saving the defaults to the post meta, which seems a bit unnecessary. I suppose as the setting changes that default we can't rely on it being the same value as when the post was saved, so we can't get around that.

spsiddarthan (approved):
This tests well!

zinigor:
Cherry picked in https://github.com/Automattic/jetpack/commit/56c0cc78aba6c11a328ad162f001929e12d91ada.

PR #30361, opened 2023-04-28
Launchpad API: Address excess queries

edanzer:
## Proposed changes:

It appears some code in our new backend Launchpad logic is resulting in excess queries. As a temporary fix to unblock deployment of other code, we're adding a 'false' bypass to avoid invoking the problematic method call. Once deployed, we'll investigate and push more complete fix. 

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

See p1682634536650749-slack-C0299DMPG

## Does this pull request change what data or activity we track or use?

No.

## Testing instructions:

The behavior of our backend logic and endpoints should be the same except for the return value of the `domain_upsell` task. For this task, 'completed' will now always be false and 'badge_text' will always be Upgrade plan. Other than that, we just want to ensure no regressive breakage.

1. Run `bin/jetpack-downloader test jetpack-mu-wpcom-plugin update/launchpad-text-string-case` to build and sync this branch to your sandbox.
2. Create or use a launchpad enabled site and sandbox the site.
3. Go to https://developer.wordpress.com/docs/api/console/, select Rest API and wpcom/v2, and input the following: `/sites/YOURSITESLUG/launchpad/checklist?checklist_slug=free`. Confirm correct checklist is returned, with correct values above for domain_upsell.

PR #30362, opened 2023-04-28
Bypass launchpad domain upsell queries

Addison-Stavlo:
Fixes p1682634633096559/1682634536.650749-slack-C0299DMPG to unblock jetpack release. We will have to follow up to update how launchpad.php handles these queries.

## Proposed changes:

* Adds a `false` to the conditional triggering the query.
* This is not yet consumed elsewhere and should be fine to bypass in this manner until we can follow up with a better solution.

### Other information:

## Testing instructions:

* Go to '..'
*


Addison-Stavlo:
closing in favor of duplicate - https://github.com/Automattic/jetpack/pull/30361 (Erick was throwing up these changes there while i was fixing environment issues required to push)

PR #30366, opened 2023-04-28
CRM: New Automation System

cleacos:
## Proposed changes:

This is the main branch for the new Automation system we implement in JPCRM. This PR contents mainly the Automation Engine code with all the required unit tests. Once the main system is done, we will merge it intro trunk.

**Build JPCRM before:** `jetpack build plugins/crm` or `composer dumpautoload` in CRM folder
**For running the tests:** `composer phpunit -- --testdox --testsuite automation`

### Other information:

## Testing instructions:

This PR was reviewed along the multiple comments and PRs merged.
- Ensure that all tests pass


projects/plugins/crm/src/automation/class-base-trigger.php
```diff
@@ -0,0 +1,80 @@
...
+	 * Base_Trigger constructor.
+	 *
+	 * @param array $trigger_data The trigger data.
+	 */
+	public function __construct( array $trigger_data ) {
+		$this->name        = $trigger_data['name'];
+		$this->title       = $trigger_data['title'] ?? '';
+		$this->description = $trigger_data['description'] ?? '';
+		$this->category    = $trigger_data['category'] ?? '';
+	}
```
kallehauge: Should we make the dependency for `name`, `title`, `description` and `category`  more explicit by adding them as parameters to the construct method instead of an array that might / might now listen for the data that is being passed? 😄  AFAIK then we don't have scenarios where we don't need any of the data:

```suggestion
	public function __construct( $name, $title, $description, $category ) {
		$this->name        = $name;
		$this->title       = $title;
		$this->description = $description;
		$this->category    = $category;
	}
```

projects/plugins/crm/src/automation/commons/triggers/class-contact-updated.php
```diff
@@ -0,0 +1,35 @@
...
+
+class Contact_Updated extends Base_Trigger {
+	
+	private $contact_before_update = array();
+	
+	public function __construct() {
+		$trigger_data = array(
+			'name'        => 'contact_updated',
+			'title'       => 'Contact Updated',
+			'description' => 'Triggered when a CRM contact is updated',
```
kallehauge: These looks like they could benefit from translation 😄

```suggestion
			'title'       => __( 'Contact Updated', 'zero-bs-crm' ),
			'description' => __( 'Triggered when a CRM contact is updated', 'zero-bs-crm' ),
```

projects/plugins/crm/src/automation/class-automation-workflow.php
```diff
@@ -0,0 +1,234 @@
...
+	 * 
+	 * @return int
+	 */
+	public function get_id(): int {
+		return $this->id;
+	}
+
+	public function set_triggers( array $triggers ) {
+		$this->triggers = $triggers;
+	}
```
kallehauge, line 77: This method seems to be unused 😄

projects/plugins/crm/src/automation/class-automation-workflow.php
```diff
@@ -0,0 +1,234 @@
...
+		$this->logger = $logger;
+	}
+	
+	/**
+	 * Set Automation Engine
+	 * @param Automation_Engine $automation_engine
+	 */
+	public function set_automation_engine( Automation_Engine $automation_engine ) {
+		$this->automation_engine = $automation_engine;
+	}
```
kallehauge, line 264: This method doesn't seem to be used anywhere 😄

projects/plugins/crm/src/automation/class-automation-workflow.php
```diff
@@ -0,0 +1,234 @@
...
+	 * Start the workflow execution once a trigger is activated
+	 * 
+	 * @param Trigger $trigger
+	 * @param array $data
+	 * @return bool
+	 * @throws Automation_Exception
+	 */
+	public function execute( Trigger $trigger, array $data ): bool {
+		$this->logger->log( 'Trigger activated: ' . $trigger->get_name() );
+		$this->logger->log( 'Executing workflow: ' . $this->name );
```
kallehauge: Should we combine these into a single log entry? Technically, when we separate the to two different log entries, there could be a race condition with two automations happening at the same time and the order of them could be mixed up (although it's very theoretical)

```suggestion
		$this->logger->log( sprintf( '[%s] Executing trigger: %s', $this->name, $trigger->get_name() ) );
```

projects/plugins/crm/src/automation/class-base-trigger.php
```diff
@@ -0,0 +1,80 @@
...
+namespace Automattic\Jetpack\CRM\Automation;
+
+/**
+ * Base Trigger implementation
+ *
+ * @inheritDoc
+ */
+abstract class Base_Trigger implements Trigger {
+
+	/** @var mixed|string The trigger name */
```
kallehauge: AFAIK then this will always be a string since we register triggers in code, right?

```suggestion
	/** @var string The trigger name */
	protected $name = '';
```

projects/plugins/crm/src/automation/class-base-trigger.php
```diff
@@ -0,0 +1,80 @@
...
+	protected $name = '';
```
kallehauge: Should we perhaps rename it from `name` to avoid mixing it up with `title`?
What about `id` or `machine_name` instead?

projects/plugins/crm/src/automation/commons/triggers/class-contact-updated.php
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Automattic\Jetpack\CRM\Automation\Triggers;
+
+use Automattic\Jetpack\CRM\Automation\Automation_Exception;
+use Automattic\Jetpack\CRM\Automation\Automation_Workflow;
+use Automattic\Jetpack\CRM\Automation\Base_Trigger;
+
+class Contact_Updated extends Base_Trigger {
+	
+	private $contact_before_update = array();
```
kallehauge: We don't seem to use this property. Should we remove it to improve readability since it kinda confused me when I saw it? 😝

projects/plugins/crm/src/automation/class-automation-workflow.php
```diff
@@ -0,0 +1,234 @@
+<?php
+
+namespace Automattic\Jetpack\CRM\Automation;
+
+class Automation_Workflow {
```
kallehauge, line 13: Should we remove the `Automation_` prefix from our class names?

AFAIK then they won't really be used outside of the Automation context anyway and, if specific code have name conflicts, then they can always use an alias:

```php
use Automattic\Jetpack\CRM\Automation\Workflow as Automation_Workflow;
```

projects/plugins/crm/src/automation/class-automation-engine.php
```diff
@@ -0,0 +1,191 @@
...
+	 * @throws Automation_Exception
+	 */
+	public function get_trigger_class( string $trigger_name ): string {
+		
+		if ( ! isset( $this->triggers_map[ $trigger_name ] ) ) {
+			throw new Automation_Exception( sprintf( __( 'Trigger %s does not exist', 'zero-bs-crm' ), $trigger_name ), Automation_Exception::TRIGGER_CLASS_NOT_FOUND );
+		}
+		
+		return $this->triggers_map[ $trigger_name ];
+	}
```
kallehauge: This method doesn't seem to actually do what the DocBlock says but IMO it would make sense to refactor it to do so since it would simplify how it's being used in the code:

```suggestion
	public function get_trigger( string $trigger_name ): Trigger {

		if ( ! isset( $this->triggers_map[ $trigger_name ] ) ) {
			throw new Automation_Exception( sprintf( __( 'Trigger %s does not exist', 'zero-bs-crm' ), $trigger_name ), Automation_Exception::TRIGGER_CLASS_NOT_FOUND );
		}

		return new $this->triggers_map[ $trigger_name ]();
	}
```

That way this implementation:

```php
$class_name = $this->automation_engine->get_trigger_class( $trigger_name );

/** @var Trigger $trigger */
$trigger = new $class_name();
$trigger->init( $this );
```

would look like this instead:

```php
$trigger = $this->automation_engine->get_trigger( $trigger_name );
$trigger->init( $this );
```

projects/plugins/crm/src/automation/class-automation-engine.php
```diff
@@ -0,0 +1,191 @@
...
+
+	/**
+	 * Register a trigger
+	 *
+	 * @param string Trigger $trigger_name
+	 * @param string         $class_name
+	 * @throws Automation_Exception
+	 */
+	public function register_trigger( string $trigger_name, string $class_name ) {
```
kallehauge: Should we try to simplify this slightly and find the `$trigger_name` value on the trigger itself?

projects/plugins/crm/src/automation/class-automation-engine.php
```diff
@@ -0,0 +1,191 @@
...
+
+	/**
+	 * Build and add a workflow
+	 *
+	 * @param array $workflow_data
+	 * @param bool $init_workflow
+	 * @return Automation_Workflow
+	 * @throws Workflow_Exception
+	 */
+	public function build_add_workflow( array $workflow_data, bool $init_workflow = false ): Automation_Workflow {
```
kallehauge, line 161: This method doesn't seem to be used anywhere 😄

projects/plugins/crm/tests/php/automation/tools/class-event-emitter.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+namespace Automattic\Jetpack\CRM\Automation\Tests;
+
+class Event_Emitter {
```
kallehauge, line 5: Do we need our own emitter system? Could we make the overall experience more WP native hooking in to WordPress' existing hook system?

A few benefits of using the WordPress hooks system:

1. We make it easier for other WordPress developers to use our system (be it 3. party developers or other a12s)
2. We will make it consistent with other event we'll have to listen for (e.g.: WooCommerce hooks for order status changes, etc)
3. If we start adding WordPress hooks to our own code (e.g. when a contact is updated) then other plugins and code can hook in to the same events, so the events we create aren't limited to the automations system

projects/plugins/crm/src/automation/class-automation-workflow.php
```diff
@@ -0,0 +1,234 @@
...
+
+		return $workflow;
+	}
+
+	/**
+	 * Get the initial step of this workflow
+	 * 
+	 * @return array
+	 */
+	public function get_initial_step(): array {
```
kallehauge, line 147: This method doesn't seem to be used anywhere 😄

projects/plugins/crm/src/automation/class-recipe-exception.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Automattic\Jetpack\CRM\Automation;
+
+class Workflow_Exception extends \Exception {
+    const INVALID_RECIPE                = 10;
+	const RECIPE_REQUIRE_A_TRIGGER      = 11;
+	const RECIPE_REQUIRE_A_INITIAL_STEP = 12;
```
kallehauge: These constants wasn't refactored to `workflow`

projects/plugins/crm/tests/php/automation/tools/class-event-emitter.php
kallehauge: (Other discussion about it: https://github.com/Automattic/jetpack/pull/30815/files#r1200482305)

kallehauge (commented):
I'm not done going through the code, but I have to step away and wanted to share what I've found so far (there's no reason waiting 😛)

(I _**LOVE**_ that we finally have PHPUnit tests 😍)

projects/plugins/crm/src/automation/class-base-trigger.php
cleacos: > we don't have scenarios where we don't need any of the data

We could have scenarios where we don't have a category or an empty description, or even a title. For example, internal triggers that we don't want to expose for any reason. The unique required field would be the $name.

Really, since properties are `protected`, the trigger to be implemented, could add their data directly in their constructor without calling the parent constructor:

```
	public function __construct() {
		$this->name 		= 'contact_updated';
		$this->title 		= 'Contact Updated';
		$this->description 	= 'Triggered when a CRM contact is updated';
		$this->category 	= 'contact';
	}
```

In fact, I'd remove the constructor from the `Base_Trigger` since we have to set it from the implemented trigger. The Automation System must be responsible for rejecting/ignoring invalid triggers, for example, without a name.

projects/plugins/crm/src/automation/class-automation-workflow.php
cleacos: For now... The Automation Module, in JPCRM, could use it when it's CRUDing it.

projects/plugins/crm/src/automation/class-automation-workflow.php
cleacos: For now...?

projects/plugins/crm/src/automation/class-automation-workflow.php
cleacos: > Should we combine these into a single log entry? 

It could be I've not gone deeper at this level of detail. 

> Technically, when we separate the to two different log entries, there could be a race condition with two automations happening at the same time and the order of them could be mixed up (although it's very theoretical)

That is not possible here. We don't have that kind of concurrency (multithreading). Each execution (PHP request) goes isolated from others, and the logger collects only the logs of the current request, and at the end of the automation execution, the log is available for the Automation System that will store it on the database or whatever, for example only store it if something went wrong and discard if not.

projects/plugins/crm/src/automation/class-automation-engine.php
cleacos: We are mapping/registering classes with the name that we will identify the trigger.

`"contact_updated" => "Automattic\Jetpack\CRM\Automation\Triggers\Contact_Updated"`

That "simplification" will imply instancing the trigger only to get the name. Or, setting the `name` as a static field.

The part of how we are going to discover triggers, actions, and conditions, is not totally defined yet. But my first idea is that probably all those fields (name, title, category,...) will become static fields or constants. We don't need to instance them until we really need them.

projects/plugins/crm/src/automation/class-automation-engine.php
cleacos: For now... still in progress

projects/plugins/crm/tests/php/automation/tools/class-event-emitter.php
cleacos: The Event Emitter is only a tool for testing purposes: `/tests/php/automation/tools/class-event-emitter.php` It's really simple, only 40 lines of code.

The idea is to emulate an event system to trigger automation workflow to test without depending on WP or third parties libraries (target: light unit testing)

projects/plugins/crm/src/automation/class-automation-workflow.php
cleacos: For now... still in progress.

projects/plugins/crm/src/automation/class-recipe-exception.php
cleacos: Thanks

cleacos: and the file name either... :)

cleacos:
PR fixed! Linter tests passed with PHP 7.2 syntax.

cleacos:
Updated with trunk (method: merge)

cleacos:
This PR is ready to merge. The code cleanup in the Internal Automator file is fine

gogdzl (approved):
Time to merge!

PR #30367, opened 2023-04-28
Add URL navigation for the Jetpack Forms dashboard

### LINKED ISSUE #30314: Jetpack Forms - WP-Admin Form Responses -> The back button doesnt work as expected
Originally [reported here](pcsBup-1aJ-p2#comment-409).

Feedback as part of testing the WP-Admin Form Responses Area:
> The back button doesn’t quite work as expected when navigating around. I would expect the URL to change and the back button to work when navigating between inbox, spam, trash tabs as well as when you open a response. This in particularly impacts the mobile web side of things where the user are more likely to use the swipe back feature.

ice9js:
Fixes #30314.

## Proposed changes:

This patch adds full support for URL based navigation for the new forms dashboard. All view settings are now persisted in the url, including: current tab & filters, current page and active response ID. This makes it much easier to access specific views/responses directly or share them with others.

This change also adds support for the browser's back button as each selection/filter/tab change now counts as a history item.

### Other information:

## Jetpack product discussion

pcsBup-1aJ-p2#comment-421

No.

## Testing instructions:

This is quite a large change so it's best to be thorough with testing:

- Go to `/wp-admin/admin.php?page=jetpack-forms`.
- You should notice the URL changing when you apply any filters, actively select a response, or change tabs.
- Refreshing the page should now bring you back to the exact same view you had before the refresh as all view settings are stored in the URL.
- Make sure date and source filters still work as expected.
- Make sure tabs still work as expected.
- Make sure we're always showing the correct active response.
- Make sure that the above settings are also correctly reflected when you export.


jcheringer:
It works very well for me on a desktop resolution.
But I noticed something when testing it simulating a mobile device using the browser DevTools.
From the list of responses, if I select a response and then hit the browser back button, I'm not being redirected to the response list.
Not sure if this will happen on an actual device or if it's just something with the DevTools.

ice9js:
>From the list of responses, if I select a response and then hit the browser back button, I'm not being redirected to the response list.

You're right, I think I missed the point there - the view was still independent of the URL.

I updated it so now it should show you the response view if you selected a valid response, or redirect you back to the inbox view if the response ID isn't part of the returned responses. Similarly the tab should switch when using the back button.

digitalwaveride (commented):
This is super cool Kuba!!!! 😻 

> You should notice the URL changing when you apply any filters, actively select a response, or change tabs.
☑️

> Refreshing the page should now bring you back to the exact same view you had before the refresh as all view settings are stored in the URL.
☑️

> Make sure date and source filters still work as expected.
🔴 Not sure what happened, and if related to this PR, but the filter work not as expected for me, when switching back to "All dates" or "All sources":

image


> Make sure tabs still work as expected.
☑️

> Make sure we're always showing the correct active response.
☑️

> Make sure that the above settings are also correctly reflected when you export.
☑️

The only other thing I noticed, it did break with my "special responses", but this is because https://github.com/Automattic/jetpack/pull/30317 was not merged yet.

ice9js:
>🔴 Not sure what happened, and if related to this PR, but the filter work not as expected for me, when switching back to "All dates" or "All sources":

I think this should be fixed now.

digitalwaveride (dismissed):
> 

>    🔴 Not sure what happened, and if related to this PR, but the filter work not as expected for me, when switching back to "All dates" or "All sources":

>> I think this should be fixed now.

👍  Fixed for me! Things work as advertised! NICE!

jcheringer (approved):
Navigation was fixed for me!
There's a known issue regarding the selected tab when using the browser navigation buttons, but as discussed with @ice9js, we'll handle that in another ticket.

PR #30393, opened 2023-05-01
CRM: Resolves 98 - export company data with segments

tbradsha:
Resolves Automattic/zero-bs-crm#98 - export company data with segments

## Proposed changes:

The original issue was to export company data with contacts. When testing, I confirmed that company data was properly exported, but not when exporting a segment, which gave this error:

`PHP Notice:  Undefined index: company in /includes/ZeroBSCRM.DAL3.Export.php on line 462`

Finding the root cause was a bit of a rabbit hole, but ultimately the fix seems to be that `getSegmentAudience()` didn't use the `withAssigned` param when running `getContacts()`, so it never got company data.

This PR adds that param, thereby fixing the issue. I also code-standardised the function and removed this silly line:
https://github.com/Automattic/jetpack/blob/db8ae2cb6530941e400139f28ff2e6d946c37551/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Segments.php#L698

### Other information:

## Testing instructions:

1. Ensure some contacts have company data in them.
2. Go to segments: `/wp-admin/admin.php?page=manage-segments`
3. Click "Export .CSV" next to one of the segments that has contacts with company data.
4. Ensure company fields are ticked.
5. Export the segment.

In `trunk`, one gets a PHP notice and the company data is blank.

In `fix/crm/98-export_company_data_with_segments`, the PHP notice is no more, and the company data is present.

### DIFF
--- a/projects/plugins/crm/changelog/fix-crm-98-export_company_data_with_segments
+++ b/projects/plugins/crm/changelog/fix-crm-98-export_company_data_with_segments
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Export: contact segments now export company info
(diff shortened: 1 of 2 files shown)


cleacos (approved):
Tested and it works as expected.

LGTM!

PR #30397, opened 2023-05-01
Launchpad Checklist API Refactor/Fixes: Task completion status and performance optimizations

### LINKED ISSUE #76058: Launchpad: Move Launchpad.php plugin to Jetpack
### Details

Right now, we keep a lot of backend launchpad logic in the launchpad.php plugin on WordPress.com. We then duplicate that logic to wpcomsh. In order to consolidate and reduce duplication, we're going to move all the backend launchpad logic to the new launchpad.php file in Jetpack. From there it will be deployed automatically to both WordPress.com and to atomic sites via wpcomsh.

mattwiebe:
Fixes https://github.com/Automattic/wp-calypso/issues/76058

## Proposed changes:

* This PR adds changes to sync WPCOM Launchpad code with Jetpack
* Use with https://github.com/Automattic/wp-calypso/pull/76467 to use launchpad checklist API on Calypso
* Task attributes ('completion', 'disabled', etc.) are now determined when the user makes a request to the Launchpad checklist API.

### Other information:

## Testing instructions:

* Checkout Calypso PR with updated launchpad checklist API changes (https://github.com/Automattic/wp-calypso/pull/76467)
* Sandbox public api and your test site(s).
* Run `bin/jetpack-downloader test jetpack-mu-wpcom-plugin add/launchpad-completion-logic` on your sandbox.
* Test the different launchpad flows (newsletter, link-in-bio, write, build, free, videopress) and ensure that the launchpad checklist tasks are marked as completed correctly when you go through them. Keep the production version of the launchpad open in a different tab and compare it with calypso localhost launchpad as you walk through each checklist task.


projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -185,13 +185,76 @@ public function build( $id ) {
...
+	/**
+	 * Loads a subtitle for a task, calling the callback if it exists.
+	 *
+	 * @param Task $task A task definition.
+	 * @return string The subtitle for the task.
+	 */
+	private function load_subtitle( $task ) {
+		if ( isset( $task['subtitle'] ) ) {
+			if ( is_callable( $task['subtitle'] ) ) {
+				return call_user_func( $task['subtitle'] );
```
jeyip, line 223: `NB:`

Does PHP not allow for invocation of the anonymous function like so?

`$task['is_disabled']( )`

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -185,13 +185,76 @@ public function build( $id ) {
...
+
+	/**
+	 * Checks if a task is complete, relying on task-defined callbacks if available
+	 *
+	 * @param Task $task Task definition.
+	 * @return boolean
+	 */
+	public function is_task_complete( $task ) {
+		if ( isset( $task['is_complete_callback'] ) && is_callable( $task['is_complete_callback'] ) ) {
+			return call_user_func( $task['is_complete_callback'] );
```
jeyip, line 251: `NB:`

Does PHP not allow for invocation of the anonymous function like so?

`$task['is_complete_callback']()`

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -185,13 +185,76 @@ public function build( $id ) {
...
+	 * Checks if a task is disabled
+	 *
+	 * @param array $task Task definition.
+	 * @return boolean
+	 */
+	public function is_task_disabled( $task ) {
+		if ( isset( $task['is_disabled'] ) && is_callable( $task['is_disabled'] ) ) {
+			return call_user_func( $task['is_disabled'] );
+		}
+		return false;
```
jeyip, line 240: `NB:`

Should we return what is defined on the task, if it's a boolean, instead of defaulting to false?

`return is_bool( $task['is_disabled'] ) ? $task['is_disabled'] : false`

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
mattwiebe: It would fail in the cases where we supply string or array-type callback identifiers like
`'some_function_callback'` or `array( $object, 'method_callback' )`. `call_user_func` allows us leverage any of the three formats (1. anonymous functions, 2. string identifier, 3. array identifier for object->method).

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
mattwiebe: The idea here was that `is_disabled` would _only_ be a callback to prevent trying to evaluate things at definition time, which is almost always too early.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
jeyip: > string identifier,  array identifier for object->method)

Ah these are what I was missing. Makes sense 👍

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
jeyip: In that case, should we rename `is_disabled` to `is_disabled_callback`?

mattwiebe: That is a solid idea and this is just the right time to make a change like that

jeyip:
## Testing
### Requirements
Test the different launchpad flows (newsletter, link-in-bio, write, build, free, videopress) and ensure that the launchpad checklist tasks are marked as completed correctly when you go through them.

- [x] Free flow tasks completed as expected
- [x] Build flow tasks completed as expected
- [ ] Write flow tasks completed as expected ( Set up your site task is unexpectedly enabled, nonblocking in my opinion )
- [x] Link in bio flow tasks completed as expected

### Browsers
- [x] Chrome

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
mattwiebe: The last two were formerly the only way to do callbacks in PHP, before anonymous functions landed in PHP 5.3

projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php
```diff
@@ -118,7 +118,7 @@ public function can_access() {
 	 *               and `checklist`.
 	 */
 	public function get_data( $request ) {
-		$checklist_slug = $request['checklist_slug'];
+		$checklist_slug = isset( $request['checklist_slug'] ) ? $request['checklist_slug'] : get_option( 'site_intent' );
```
mattwiebe, line 122: This will allow calling the endpoint without  a `checklist_slug` at all, since most of the time all we want is the checklist for a site's current `site_intent`.

jeyip (dismissed):

jeyip:
My only concern is edge cases for the task building logic which we might not be considering. I have to hop off, but I think when we're confident with how we're handling `is_completed_callback`, `is_disabled_callback`, `subtitle`, etc., let's merge this.

jeyip (dismissed):

edanzer:
Thanks for all the work on this @mattwiebe and @agrullon95. There is a lot being tackled here.

**Review.** With respect to reviewing, there is a LOT being changed here, combined with a lot of new code being added and/or transferred from the wpcom launchpad.php plugin. And all that change is on top of the existing, also-newish abstraction code that I didn't have a good handle on in the first place. 

Short of sitting down with someone to walk through the purpose of various changes in details, it's honestly hard for me to review. That said, I know that our current code is a) returning wrong values for checklist statuses and b) creating significant performance concerns, and I know this PR fixes both of those things. 

So on the basis of that, I'd be in favor of getting this merged. 

**Testing.** As far as testing goes, this seems to test as expected, which is great. I did the following.

1. Sync this branch to my sandbox and sandbox public api plus each newly created test site. 
2. Checkout the suggest frontend calypso branch at https://github.com/Automattic/wp-calypso/pull/76467
3. Tested Free flow.
   - Confirm initial launchpad statuses are correct. 
   - Confirm statuses update correctly as I completed tasks. 
   - Checked return values on frontend via frontend PR, but also directly from endpoint via dev console. 
4. Tested Newsletter flow.
   - Confirm initial launchpad statuses are correct. 
   - Confirm statuses update correctly as I completed tasks. 
   - Checked return values on frontend via frontend PR, but also directly from endpoint via dev console. 
5. Tested Link in Bio flow.
   - Confirm initial launchpad statuses are correct. 
   - Confirm statuses update correctly as I completed tasks. 
   - Checked return values on frontend via frontend PR, but also directly from endpoint via dev console. 
6. Tested Build flow.
   - Confirm initial launchpad statuses are correct. 
   - Confirm statuses update correctly as I completed tasks. 
   - Checked return values on frontend via frontend PR, but also directly from endpoint via dev console. 
7. Confirmed that correct values are returned from the endpoint whether I send request to /launchpad or /launchpad?checklist_slug=CHECKLIST_SLUG (ie with or without providing a checklist slug). I think returning a default checklist based on site_intent is a good change here. 

**Notes:** 
- The verify_email task always returns false right now from endpoint for newsletters. We update that on the frontend. I think that's fine for now. But I wonder if we should not populate that with 'false' if we know the value may actually be wrong? This would apply for any task where we're depending on frontend logic rather than backend logic to determine completion status.
- This was generating some e2e errors related to the domain upsell page. I re-ran those and they passed. But there are still some failing Jetpack checks we'll need to resolve before merging.

projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php
edanzer: This was a good 'extra' addition in this PR.

edanzer (dismissed):

mattwiebe:
@edanzer 

Thanks for the thorough review!

> The verify_email task always returns false right now from endpoint for newsletters. We update that on the frontend. I think that's fine for now. But I wonder if we should not populate that with 'false' if we know the value may actually be wrong? This would apply for any task where we're depending on frontend logic rather than backend logic to determine completion status.

We're only prepopulating so it didn't accidentally show up as enabled anywhere, and I knew we'd be dynamically managing it on the frontend. We can/should revisit if it causes problems.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
```diff
@@ -17,12 +17,170 @@
 require_once __DIR__ . '/class-launchpad-task-lists.php';
 
 /**
- * Returns the list of tasks by flow or checklist id.
- *
- * @return array Associative array with checklist task data
+ * Registers all default launchpad checklists
  */
```
lezama, line 21: shouldn't this file have a NAMESPACE?

lezama: or find better unique names for functions 😄

Addison-Stavlo: Totally agree. A namespace or `wpcom_*` prefix could probably mitigate a lot of potential for conflicts here.

jeyip (approved):
Found a small namespacing bug with Anthony that he fixed.

## Testing
- [x] Retested all tasks in the free flow and they continue to behave as expected. 
- [x] Retested all tasks in the write flow and they continue to behave as expected. 

## Browsers
- [x] Chrome

edanzer (approved):
I retested this again following recent changes, and it is testing as expected. I had an issue with the lauchpad/checklist endpoint but that's been resolved in the latest commit. 

As noted in my other comment, the code looks good to me, but the scope of code changes and the fact that I wasn't too familiar with the registry code means I may miss things on the review side.

But our current code is returning wrong values and has performance issues, and this PR is blocking up other things. So I'd say let merge and we can deal with follow ups if they arise.

PR #30400, opened 2023-05-02
Settings: Use Constants to check value of REST_API_REQUEST inside the is_syncing function

### LINKED ISSUE #76343: Site Assembler - Patterns - Fix patterns with Jetpack blocks like forms and maps
## What

Patterns with Jetpack blocks like forms and maps are not renderer.

|Assembler|Editor|
|---|---|
|Image|Image|

## Why

It seems like Jetpack blocks are not ready for our server-side rendering system. 

## How

> The "Submit a form" link makes it seem like your server-side rendering is failing the jetpack_is_frontend() check [here](https://github.com/Automattic/jetpack/blob/a27cfebbccfec63dda77b550af5466eb22f581b7/projects/plugins/jetpack/modules/contact-form/grunion-contact-form.php#L524) or [here](https://github.com/Automattic/jetpack/blob/a27cfebbccfec63dda77b550af5466eb22f581b7/projects/packages/forms/src/contact-form/class-contact-form-plugin.php#L419), so you're getting the fallback rendering that's used for emails and notifications and such.

Find a way to workaround it and ask Jetpack for help. See p1681882551868759-slack-CBG1CP4EN and p1683021691694579-slack-CDLH4C1UZ

Test them in Horizon and remove the tag `assembler-waitlist` when they are fixed.

arthur791004:
Fixes https://github.com/Automattic/wp-calypso/issues/76343

## Proposed changes:

We want to render the `jetpack/contact-form` block via the rest API endpoint but it's disabled since it will check the value of [Settings::is_syncing](https://github.com/Automattic/jetpack/blob/fb116496543a656cde086d4bdbd37c3817e783e6/projects/packages/forms/src/contact-form/class-contact-form.php#L243). However, the value is always true if it's in the context of the [REST API request](https://github.com/Automattic/jetpack/blob/fb116496543a656cde086d4bdbd37c3817e783e6/projects/packages/sync/src/class-settings.php#L511).

So, this PR is proposing to use `Constants` class from Jetpack to check the defined value and then we're able to overwrite the value to bypass the check of `Settings::is_syncing` to have a chance to render `jetpack/contact-form` block via the REST API request.

We can also add a new filter, `jetpack_sync_settings_is_syncing`, to filter the value of `Settings::is_syncing`  as below. I'm also fine with this way. WDYT?

```php
public static function is_syncing() {
  $is_syncing = self::$is_syncing || ( defined( 'REST_API_REQUEST' ) && REST_API_REQUEST );
  return (bool) apply_filters( 'jetpack_sync_settings_is_syncing', $is_syncing );
}
```

### Screenshots

| Before | After |
| - | - |
| image | image |
| image | image |
| image | image |
| image | image |

### Other information:

## Testing instructions:

* Apply D109531-code to your sandbox
* Apply this PR to your sandbox as below
  * `bin/jetpack-downloader test jetpack feat/rest-api-jetpack-block`
* Go to /setup/site-setup?siteSlug=<your_free_site>
* Continue until you land on the Design Picker
* On the Design Picker screen, scroll down to the bottom and select BCPA CTA
* On the Pattern Assembler screen
  * Select Sections > Add patterns > Contact
  * Ensure the patterns with the contact form are rendered correctly


bisko (dismissed):
LGTM! Thanks for improving that!

bisko (approved):

PR #30415, opened 2023-05-03
Bit.ly: avoid errors when using non-official Bit.ly plugins alongside Jetpack.

### LINKED ISSUE #30412: Third-Party Bit.ly Code does not check for method
### Impacted plugin

Social

### Quick summary

A third-party plugin is included to change the default behavior of the [Official Bit.ly Plugin](https://wordpress.org/plugins/bitly/) (which hasn't been updated in over 10 years and is now closed). This plugin assumes that the method `get_bitly_link_for_post_id` will be present if the `Bitly` `$_GLOBAL` is set:
https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/jetpack/3rd-party/bitly.php#L26

However, I recently ran into a situation where another plugin was using the `Bitly` namespace causing a fatal error with this bit of Jetpack code.

This code should be wrapped in `method_exists` or removed from Jetpack due to it being quite dated :) 

### Steps to reproduce

1. add a plugin (or code) that sets the `Bitly` class name, but is not the official WordPress plugin
3. Enable Jetpack Social 
4. Get a fatal error that `get_bitly_link_for_post_id` method can't be found.

I would expect that PHP's [method_exists](https://www.php.net/manual/en/function.method-exists.php) function is used to ensure the method exists before trying to call it. 

### What actually happened

A fatal error due to the lack of method.

### Impact

One

### Available workarounds?

Yes, easy to implement

Self-hosted

### Logs or notes

WordPress VIP

jeherve:
Fixes #30412

## Proposed changes:

Follow-up from #30060.

Multiple plugins use the Bitly class, and one of them does not have a `get_bitly_link_for_post_id` method. Let's not assume it exists.

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

I do not know the exact plugin one can use to test this. @BrookeDot Any chance you could give it a try on the site where you ran into the issue?

### DIFF
--- a/projects/plugins/jetpack/3rd-party/bitly.php
+++ b/projects/plugins/jetpack/3rd-party/bitly.php
@@ -22,7 +22,14 @@ function jetpack_bitly_og_tag() {
 		if ( method_exists( $GLOBALS['bitly'], 'og_tags' ) ) {
 			$GLOBALS['bitly']->og_tags();
 		}
-	} elseif ( isset( $GLOBALS['posts'] ) && $GLOBALS['posts'][0]->ID > 0 ) {
-		printf( "<meta property=\"bitly:url\" content=\"%s\" /> \n", esc_attr( $GLOBALS['bitly']->get_bitly_link_for_post_id( $GLOBALS['posts'][0]->ID ) ) );
+	} elseif (
+		isset( $GLOBALS['posts'] )
+		&& $GLOBALS['posts'][0]->ID > 0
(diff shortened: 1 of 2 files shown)


anomiex (approved):
Looks ok, just adds one `method_exists` check and reformats some code. Haven't tested it though.

BrookeDot:
This got merged before I could test, but this was on a VIP site, the customer also resolved it before I was able to sandbox. Thanks for the quick fix!

PR #28906, opened 2023-02-10
Forms: Update patterns modal default view mode

jcheringer:
## Proposed changes:

* Updates the Form patterns modal default view mode to Grid

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion
N/A

## Does this pull request change what data or activity we track or use?
No

## Testing instructions:

* Create a post and add a Form block
* Click the "Explore Forms Patterns" button
* Check if the modal opens with the grid view mode selected by default

PR #30419, opened 2023-05-03
e2e tests: add new step to form block test

jeherve:
## Proposed changes:

Following #28906, you now need one additional step to pick a form in the block editor: the default view mode is now the grid mode, so we need to switch to the carousel mode so the choose button can appear, and so we can select the default form pattern.

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

* Is CI happy? Block tests are failing, but not the form ones:
https://github.com/Automattic/jetpack/actions/runs/4870781951/jobs/8687816981

I think the failures with the tiled gallery block aren't related and can be ignored here.


jcheringer (approved):
Thanks for taking care of this, @jeherve!
And sorry for missing the failing tests on the other PR.
Tests are now working fine 👍

PR #30420, opened 2023-05-03
theme-tools: remove jQuery from responsive-videos

sgomes:
This PR reimplements the `responsive-videos.js` script in vanilla JS, removing the dependency on `jQuery`.

## Proposed changes:
* Reimplement `responsive-videos.js` without `jQuery`
* Remove the dependency on `jQuery` for this script
* Add `responsive-videos` to list of paths to be compiled
* Modify `moduleSources` handling to support sub-paths

### Other information:

## Jetpack product discussion
None; this is a simple under-the-hood change.

No.

## Testing instructions:
- For your test blog, pick a theme that doesn't require `jQuery` by itself but makes use of `responsive-videos.js`, e.g. Independent Publisher 2.
- Create a test page with a video shortcode.
   <details>
   <summary>Example content</summary>

   ```html
   <!-- wp:paragraph {"align":"center"} -->
   <p class="has-text-align-center">A centered paragraph.</p>
   <!-- /wp:paragraph -->
   
   <!-- wp:shortcode -->
   [youtube https://www.youtube.com/watch?v=dQw4w9WgXcQ]
   <!-- /wp:shortcode -->
   ```

   </details>
- Load the test page.
- Ensure that `jQuery` is not loaded.
- Ensure that the video resizes as you shrink and grow the window, and that it does so in a manner identical to the current script in trunk.


sgomes:
The test failures seem to be unrelated, possibly due to flakiness. I've retried many times but was never able to get a successful run 🤷

anomiex:
PR looks fine to me at a quick read-through. Haven't tested yet.

>  things could probably be improved by having the build process generate responsive-videos.min.js, instead of expecting the developer to do so manually. I have no idea what the idiomatic way of accomplishing that in this repo would be, however confused

First, probably add `...glob.sync( 'modules/theme-tools/responsive-videos/*.js' ),` into https://github.com/Automattic/jetpack/blob/26f0fba6ca6ebe80766e3e785b49a39ba70138a5/projects/plugins/jetpack/tools/webpack.config.js#L94-L97 Unfortunately we can't just add it to `supportedModules` because `glob()` can't handle a slash inside `@(...)`. That will generate files in `_inc/build/theme-tools/responsive-videos/`.

Then the enqueuing at https://github.com/Automattic/jetpack/blob/26f0fba6ca6ebe80766e3e785b49a39ba70138a5/projects/plugins/jetpack/modules/theme-tools/responsive-videos.php#L67-L71 should be changed to be more like this
```js
Assets::register_script(
    'jetpack-responsive-videos',
    '_inc/build/theme-tools/responsive-videos/responsive-videos.js',
    JETPACK__PLUGIN_FILE,
    array(
        'in_footer' => true,
        'enqueue' => true,
    )
);  
```
(with `use Automattic\Jetpack\Assets` added at the top of the file, of course).

You could also throw `'nonmin_path' => 'modules/theme-tools/responsive-videos/responsive-videos.js'` in the array if you really want, but probably we should move away from that method of debugging scripts.

You might even take it a step further and `import './responsive-videos.css';` in the JS file so you can get rid of the [separate style enqueue](https://github.com/Automattic/jetpack/blob/26f0fba6ca6ebe80766e3e785b49a39ba70138a5/projects/plugins/jetpack/modules/theme-tools/responsive-videos.php#L73-L79) (`Assets::register_script` will automatically register a css file if one exists alongside the js file, and that import will cause a minified css file to be generated), although that would also need a change [here](https://github.com/Automattic/jetpack/blob/26f0fba6ca6ebe80766e3e785b49a39ba70138a5/projects/plugins/jetpack/class.jetpack.php#L97) for the different handle.

sgomes:
Thank you, @anomiex, that was a great step-by-step! 🙌

I've applied the JS-related changes you mentioned, and I even found a way of adding the directory path to `supportedModules` 🙂 Please let me know what you think!

I didn't import the CSS, however, as that seemed to bloat the JS somewhat with webpack scaffolding, and it would require linting changes somewhere to start treating `responsive-videos.js` as an ESM file.

sgomes:
I'm having some trouble loading the minified script; it seems to want to load `.js` instead of the generated `.min.js`. Investigating.

sgomes:
I had to go back to using `wp_enqueue_script`, as it seems that `Assets::register_script` does not correctly load the `.min.js` file, and instead tries to load `.js`. The other modules in the jetpack plugin seem to use the `wp_enqueue_script` approach, so it seems safe to do the same here.

sgomes:
I restored the code changes to `supportedModules` / `moduleSources`, as it seems they weren't behind the issues I was seeing.

sgomes:
@anomiex : I think this is ready for another look! 👍

anomiex:
> I'm having some trouble loading the minified script; it seems to want to load `.js` instead of the generated `.min.js`. Investigating.

Is it because I had the wrong path in the example code earlier? I should have said `'_inc/build/theme-tools/responsive-videos/responsive-videos.min.js'`, since that's the actual filename.

anomiex:
> I've applied the JS-related changes you mentioned, and I even found a way of adding the directory path to `supportedModules` slightly_smiling_face Please let me know what you think!

Nice! In some quick testing, it seems to be approximately the same speed.

> I didn't import the CSS, however, as that seemed to bloat the JS somewhat with webpack scaffolding,

When I try it locally, it looks like it's only 15 bytes (1160 vs 1175) for the production build. 🤷 But it's fine to leave it out if you want.

> and it would require linting changes somewhere to start treating `responsive-videos.js` as an ESM file.

Speaking of which, we should probably do that anyway. Currently `modules/.eslintrc.js` sets up very different rules because the module source files are loaded directly (and in particular wpcom's minifier doesn't like es6). When we switch a module to using a built version and Assets instead, we can use the same modern linting rules as we do for the rest of the JS.

For now, the simple change to the linting rules would be to add `modules/theme-tools/responsive-videos/.eslintrc.js` with contents like this:
```js
/**
 * @type {import("eslint").Linter.Config}
 */
module.exports = {
	// eslint config in modules/.eslintrc.js is screwy for historical reasons that don't apply to built modules like this one.
	// Reset to the rules from Jetpack's root dir.
	extends: [ '../../../.eslintrc.js' ],
};
```
Eventually we'd delete `modules/.eslintrc.js` along with `modules/**/.eslintrc.js`, once all modules are being built and can be written in modern syntax.

sgomes:
> Is it because I had the wrong path in the example code earlier? I should have said '_inc/build/theme-tools/responsive-videos/responsive-videos.min.js', since that's the actual filename.

That is exactly why, oops! 😅 My bad, didn't put two and two together.

There seems to be some weirdness with `.min.*`, however, in that the built JS file ends with `.min.js`, but the built CSS file ends with just `.css`, at least for local builds. Because of that, I had to use `css_path` to get CSS loading to work, since `Assets::register_script` wouldn't automatically match the two in that situation.

> > I didn't import the CSS, however, as that seemed to bloat the JS somewhat with webpack scaffolding,
> 
> When I try it locally, it looks like it's only 15 bytes (1160 vs 1175) for the production build. 🤷 But it's fine to leave it out if you want.

Nope, you're absolutely right! The cruft goes away entirely in the production build 👍

> > and it would require linting changes somewhere to start treating `responsive-videos.js` as an ESM file.
> 
> Speaking of which, we should probably do that anyway.

Done, thanks for the step-by-step! 🙂

---

Please let me know what you think about the latest changes, @anomiex! 🙏

projects/plugins/jetpack/tools/webpack.config.css.js
```diff
@@ -132,7 +132,6 @@ const weirdRtlEntries = {
 		'modules/shortcodes/css/style.css', // TODO: Should be renamed to shortcode-presentations
 		'modules/shortcodes/css/quiz.css',
 		'modules/subscriptions/subscriptions.css',
-		'modules/theme-tools/responsive-videos/responsive-videos.css',
```
anomiex: I'm not sure that we should actually remove this here. It seems the purpose of this is to speed up frontend page loads by having Jetpack load one big frontend css file instead of all these small ones (because WP Core doesn't/didn't concatenate them, apparently).

So probably we should keep it for now, pending some investigation as to the performance impact of removing it for all these files. Maybe after minified versions of them all can be loaded instead, as that may have been part of the issue.

You'll still need to change class.jetpack.php, but to rename the handle rather than removing it.

anomiex (commented):
> There seems to be some weirdness with `.min.*`, however, in that the built JS file ends with `.min.js`, but the built CSS file ends with just `.css`, at least for local builds. Because of that, I had to use `css_path` to get CSS loading to work, since `Assets::register_script` wouldn't automatically match the two in that situation.

Yeah, now that I think about it, we don't normally name files ".min.js" because that stops translate.wordpress.org from finding the i18n strings. That doesn't matter for this file since it doesn't currently use `@wordpress/i18n` and so has no i18n strings.

[There's a "fix this" comment in the webpack config](https://github.com/Automattic/jetpack/blob/73c759e6e6e76a6de6842ebc8226855889ef0d81/projects/plugins/jetpack/tools/webpack.config.js#L133) about the ".min.js" naming, but actually fixing it is probably something for a separate PR since it will affect all the modules. `css_path` is good for now.

projects/plugins/jetpack/tools/webpack.config.css.js
sgomes: I see, that makes sense! I modified this file to add this back, and used the correct handle in `class.jetpack.php` 👍

anomiex (approved):

anomiex:
Thanks for all the work here!

sgomes:
Thank you for the review and all the great guidance, @anomiex ! 👍

Issue #219, opened 2014-02-14
Sharing: Lacking a loading@2x.gif


kraftbj:
All of the images in the Sharing module include a 2x version, except for loading.gif. This is resulting in some 404s, as reported by an user in 1761270-t.

PR #30422, opened 2023-05-03
CRM: Fixing segments by email and doesnotcontain, by reordering checks in segmentConditionArgs()

coder-karen:
## Proposed changes:

* When adding new segments using the 'Does Not Contain' condition in a few situations (using Email, or Address fields), an error popup will show and the segment won't save. This PR addresses the issue with the Email field by changing the order in which condition types are checked in `segmentConditionArgs`.
* This PR also changes the `segment_condition_produces_no_args` error code to produce a 400 response instead of 500, when previewing segments.

### Other information:

## Jetpack product discussion

https://github.com/Automattic/jetpack-crm-extensions/issues/672
Related PR: https://github.com/Automattic/jetpack-crm-extensions/pull/675

No.

## Testing instructions:

1. To test this out, install and activate the Advanced Segments extension on a test site, as well as having the CRM plugin installed and active.
2. Open up the Segments dashboard, and visit the extensions page: [`/wp-admin/admin.php?page=manage-segments`](http://norubbish.local/wp-admin/admin.php?page=manage-segments), then click 'Add New'.
3. Attempt to create the following segment:
   1. Email / Does not contain / {example-email-string}
4. Each time you click 'Preview Segment', you will see a popup with 'General Error `#219` There was a general error' on it
5. Each time you click 'Save', the page will display the following message:

235457834-71b07c6f-0807-4b90-92ce-a67191c09d3f

6. Apply this PR, and follow the same steps - the segments should save properly and a preview be visible as well.
7. To test the error code change, the best bet is to apply this PR but not the Advanced Segments PR (ideally done before that PR is merged - otherwise comment out the changed lines in `contact/class-segment-condition-contact-address-field.php`).
8. Then, try adding a 'Does Not Contain' condition to an address field when building a segment. Click 'Preview' and check the developer console - you should see `POST .../wp-admin/admin-ajax.php 400 (Bad Request)` instead of `POST .../wp-admin/admin-ajax.php 500 (Internal Server Error)`


projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
```diff
@@ -5349,13 +5349,18 @@ function zeroBSCRM_AJAX_previewSegment() {
 
 			// return error str
 			$error_string = $exception->get_error_code();
+			$status       = 500;
+			if ( $error_string === 'segment_condition_produces_no_args' ) {
+				$status = 400;
+			}
```
tbradsha, line 5350: ❤️ Probably could switch it to 500 all around, but I'm not familiar with fail cases and this works for now.

tbradsha (changes requested):
This changes the error code as expected, and lays the groundwork for email to support "Does not contain".

However:

1. We don't support "Does not contain" in any of the other fields in core. Perhaps we should keep this consistent.
2. If we do decide to implement this in core, we'd need to enable the condition for email here (took me a while to figure this out):
https://github.com/Automattic/jetpack/blob/4a6949976b856f375f6e8fd930d0929a94267d35/projects/plugins/crm/includes/jpcrm-segment-conditions.php#L41

~This is because Advanced Segments overrides contact field segment conditions with its own.~

Edit: I believe Advanced Segments is supposed to override core contact field segments, but it does not. As such, there are some other issues at play. 🤔

coder-karen:
> Edit: I believe Advanced Segments is supposed to override core contact field segments, but it does not. As such, there are some other issues at play. 🤔

Thanks @tbradsha , I've now added an if / else statement within `segmentConditionArgs`, focussing on checking the 'default' case first and then having the core case as a fallback, based on what you mentioned in Slack (as well as cleaning up that initial default case as you mentioned as well).

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Segments.php
```diff
@@ -1435,7 +1435,24 @@ public function segmentConditionArgs($condition=array(),$conditionKeySuffix=''){
 
                     global $zbs,$wpdb,$ZBSCRM_t;
 
-                    switch ( $condition['type'] ){
+					$advanced_segments_active = defined( 'JPCRM_ADVANCED_SEGMENTS_ROOTFILE' );
```
gogdzl: Just dropping my two cents here: personally I dislike the idea of the core plugin referencing extensions. I think that the core plugin should not be aware of the existence of specific extensions, and should be prepared to receive input from any code that calls a filter/action/hook.

Not a blocker though.

gogdzl:
Ignore this comment, made it in the wrong PR.

gogdzl (changes requested):
Ignore this comment, made it in the wrong PR.

gogdzl (dismissed):
It works!
I think it could be improved by taking away the reference to advanced segments extension, but it is not a blocker.

coder-karen:
Thanks @gogdzl ,
I'm not too sure what a good alternative would be - well, other than adding the `doesnotcontain` option to emails in core as I'd done initially (it doesn't show on the frontend if the Advanced Segments plugin is not active). But as @tbradsha  mentioned we don't support `doesnotcontain` in core for anything. Maybe laying the groundwork in core would be ok though, as long as we don't enable it in core as that will take away some of the benefits that the Advanced Segments plugin provides.
I imagine there are other things we could do such as adding additional keys to conditions added in the AS plugin, then checking for the existence of those keys. But that's still checking something that's within the AS plugin (unless the main issue here was that the check for the AS plugin is very obvious).

tbradsha:
I don't think we even need a reference to Advanced Segments. It used to do this:

1. Check for core conditions
2. Check for non-core conditions

All we've done is switch the order:

1. Check for non-core conditions
2. Check for core conditions

If there aren't any non-core conditions defined (be it through Advanced Segments or any other third-party setup), it'll still fall back to any core conditions, right?

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Segments.php
tbradsha: I don't think we need a check at all; we're just changing the order of precedence (non-core, then core).

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Segments.php
```diff
@@ -1435,7 +1435,24 @@ public function segmentConditionArgs($condition=array(),$conditionKeySuffix=''){
...
+
+			// If the Advanced Segments plugin is active, we want to use the logic defined there instead.
+			if ( ! empty( $condition['type'] ) && $advanced_segments_active ) {
+				$filter_tag = $this->makeSlug( $condition['type'] ) . '_zbsSegmentArgumentBuild';
+				if ( substr( $condition['type'], 0, 5 ) === 'zbsc_' ) {
+					$filter_tag = $this->makeSlug( substr( $condition['type'], 5 ) ) . '_zbsSegmentArgumentBuild';
+				}
```
tbradsha: I'd simplify this:

```suggestion
				$filter_tag = $this->makeSlug( preg_replace( '/^zbsc_/', '', $condition['type'] ) ) . '_zbsSegmentArgumentBuild';
```

Or perhaps more readable:

```suggestion
				// normalise type string
				$condition_type  = preg_replace( '/^zbsc_/', '', $condition['type'] );
				$filter_tag = $this->makeSlug( $condition_type ) . '_zbsSegmentArgumentBuild';
```

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Segments.php
tbradsha: In fact, adding the check actually breaks conditions added by the core Woo module.

tbradsha (changes requested):
We'll need to remove the Advanced Segments check; otherwise segments added by the Woo module will not work.

coder-karen:
> If there aren't any non-core conditions defined (be it through Advanced Segments or any other third-party setup), it'll still fall back to any core conditions, right?

That seems to be the case after trying that out, yes.

I've followed up with another commit. Thanks for checking this, both of you.

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Segments.php
coder-karen: Ah interesting, thanks for checking this.

tbradsha: Fixed in 07a1fd2.

tbradsha (approved):
This PR is good to go!

PR #30430, opened 2023-05-03
My Jetpack: Add check for product status before requesting stats

### LINKED ISSUE #30387: My Jetpack: VideoPress featured stats being requested even when the product is not active
Steps to replicate:

- launch a new JN site with only Jetpack installed
- finish connecting Jetpack on the site
- open the browser console
- go to `Jetpack > My Jetpack`
- notice that a request to `videopress/v1/stats/featured` is made even when the VideoPress module is not enabled, and that the request fails because the endpoint is not available before the module is active:

Image

Since the module is not enabled, we should prevent trying to request the data since the endpoint will not be available.

dhasilva:
Fixes #30387

## Proposed changes:

* Checks for the product status and, if not active, does not send a request for stats

### Other information:

## Testing instructions:

* Test on a site with Jetpack, without the VideoPress plugin and VideoPress stats enabled
* Make sure VideoPress is disabled (Jetpack > Settings > Media > `Enable Videopress` is off)
* With the `network` dev tools tab opnen, go to My Jetpack
* Check that there is no request to `videopress/v1/stats/featured`
* Enable VideoPress via the Media settings page (if the plugin is installed and disabled, enabling via the card will enable the plugin as well, which we don't want now)
* Go back to My Jetpack
* Check that there is a request to `videopress/v1/stats/featured` and the card shows the stats as intended
* Now install/activate the VideoPress plugin (if on JN, use Jetpack Beta to install this PR's branch version of the plugin, as installing via the car will install the production version)
* Check that there is a request to `videopress/v1/stats/featured` and the card shows the stats as intended

https://user-images.githubusercontent.com/8486249/235983710-7a497c09-10b6-4265-a16d-e71074bbf69f.mp4

### DIFF
--- a/projects/packages/my-jetpack/_inc/components/connected-product-card/index.jsx
+++ b/projects/packages/my-jetpack/_inc/components/connected-product-card/index.jsx
@@ -15,9 +15,9 @@ import ProductCard, { PRODUCT_STATUSES } from '../product-card';
 const ConnectedProductCard = ( { admin, slug, children, showMenu = false, menuItems = [] } ) => {
 	const { isRegistered, isUserConnected } = useConnection();
 
-	const { detail, status, activate, deactivate, isFetching, installStandalonePlugin } =
-		useProduct( slug );
-	const { name, description, manageUrl, requiresUserConnection, standalonePluginInfo } = detail;
+	const { detail, activate, deactivate, isFetching, installStandalonePlugin } = useProduct( slug );
+	const { name, description, manageUrl, requiresUserConnection, standalonePluginInfo, status } =
+		detail;
 	const [ installingStandalone, setInstallingStandalone ] = useState( false );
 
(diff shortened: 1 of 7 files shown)


lhkowalski:
With VideoPress disabled, there is no request being made to `videopress/v1/stats/featured`:

Screen Shot 2023-05-03 at 14 08 57

After enabling the VideoPress module on Jetpack (keeping the VideoPress plugin disabled), the request is made and is returning the correct data:

Screen Shot 2023-05-03 at 14 12 41

After enabling the plugin, the request is made and is also returning the correct data:

Screen Shot 2023-05-03 at 14 18 01

So it's working as expected. Going to take a look in the code.

lhkowalski (approved):
Code change is great. Nice improvement on moving the `stats` selector call to the `useProduct` hook.

PR #30431, opened 2023-05-03
CRM: Resolves 300 and 2884 - fix task date inconsistencies

tbradsha:
Resolves Automattic/zero-bs-crm#300 and Automattic/zero-bs-crm#2884 - fix task date inconsistencies

## Proposed changes:

Tasks have plenty of issues related to timezones and errant conversions/storage. This PR seeks to solve those, as they're all interconnected.

A high-level summary of the issue is that when we saved a task via the task editor, it would add the timezone offset to the UTS (which is never a good thing). When viewing the task in the editor or the calendar, the datetime was shown without adjustment so looked correct, but in various other places (e.g. task list view, contact profile, notification emails, etc.) the time was displayed with the local WP timezone adjustment (correctly processed, but looking incorrect because of the source data).

This PR does the following:
* Swaps out daterangepicker inputs to use native browser inputs (yay, we're slowly removing our dependency on that library)
* Removes our use of `strftime()`, which is deprecated in PHP 8.1
* Swaps out legacy (and often errant) functions like `zeroBSCRM_date_forceEN()` and `zeroBSCRM_locale_dateToUTS()` for safer ones like `jpcrm_uts_to_date_str()` and `jpcrm_datetime_post_keys_to_uts()`
* Removed legacy functions that are now no longer used (e.g. `zeroBSCRM_locale_setServerLocale()`, `zeroBSCRM_locale_resetServerLocale()`, and `zeroBSCRM_date_forceEN()`)
* Fixed `jpcrm_datetime_post_keys_to_uts()` to properly support its `$format` param
* Removed some pre-DAL3 code
* Fixed the `##TASK-BODY##` placeholder (it was using the wrong key)
* Added a migration to subtract the timezone offset from stored values in the database
* Caught [a moment.js notice](https://github.com/Automattic/jetpack/pull/30290#pullrequestreview-1408534150) due to use of fallback formats (h/t @coder-karen)

A note on scope: when swapping out the inputs on the task editor page, I went for "good enough". That page could really use some love, but that's for another day.

### Other information:

## Testing instructions:

Most of the testing isn't too complicated right off, and the code shouldn't be too complex (most of the changes are chops). Do the following on the `fix/crm/300-task_date_inconsistencies` branch:

1. Switch to a non-UTC timezone in WP.
2. Create a new task and assign it to a contact. Verify the following:
    * the UTS value in the database matches what you submitted: https://www.epochconverter.com/
    * the date and time are consistent correct in all the places:
        * the task editor
        * the task calendar
        * the task listview
        * the contact profile under the tasks tab
3. Set the end date to before the start date, and save. The end date should show as start date + 1 hour.
4. Go to the Calendar view. You should no longer see a `moment.js` notice in the JS console.
5. Switch to PHP 8.1 and visit the Calendar view again. You'll no longer see PHP notices regarding the deprecated `strftime()` function.

There's not an easy way to test migrations (they only run once), but this one is pretty important to test. I recommend doing the following:

6. Start with a fresh `trunk` install with a non-UTC WP timezone.
7. Create a task. Note the date and time you selected. Observe that the date and time is incorrect in the task list view and the contact profile, for example.
8. Check the value in the database, which will be incorrect.
9. Install a copy of the `fix/crm/300-task_date_inconsistencies` branch. When you load a WP page, the migration should run.
10. Repeat the checks in step 2. The UTS value should be corrected in the database and the date outputs should match what you selected when creating the task in step 7.


cleacos:
I'm curious about the name `task_offset_fix` for the migration instead of the version.

tbradsha:
> I'm curious about the name `task_offset_fix` for the migration instead of the version.

Consider it a test. I can change it, but my thoughts:

* Version numbers are now a bit arbitrary as of the monorepo, which means it's hard to know what one to put.
* Some versions don't have migrations. It looks odd when there's a gap.
* Other versions have multiple migrations, which has led to merge conflicts in the past (or weird version numbers like 5.5.a).

In the end, I think it's more useful to have each migration be tied to a specific fix. This might be easier to clean up as well.

cleacos:
> Consider it a test. I can change it, but my thoughts:

I'm fine with it, they seem good reasons for it. I was curious about the motivation.

gogdzl (dismissed):
Everything works.
Migration worked flawlessly.
As usual: Great job!

tbradsha:
Rebased.

gogdzl (dismissed):

gogdzl (dismissed):

gogdzl (approved):

PR #30398, opened 2023-05-01
Remove the upgrade nudge for the Premium Block

romarioraffington:
Relates to https://github.com/Automattic/gold/issues/11
Fixes https://github.com/Automattic/jetpack/issues/30349

## Proposed changes:

Removes the upgrade nudge for the Premium block

| Before  | After |
| ------------- | ------------- |
| Screenshot 2023-05-01 at 4 44 22 PM  | Screenshot 2023-05-01 at 4 43 54 PM  | 

### Other information:
- [ ] Have you written new tests for your changes, if applicable?
- [x] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [x] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Does this pull request change what data or activity we track or use?

No

## Testing instructions:

* Add the Premium block to a site with a free plan
* Verify that the upgrade nudge is no longer present
* Open the post in an incognito window
* Checkout to ensure the block works as expected

PR #30432, opened 2023-05-03
[Earn] Make Earn products free for all Jetpack plans

romarioraffington:
Relates to https://github.com/Automattic/gold/issues/11

## Proposed changes:

Make Earn products free for all Jetpack plans.  Follow up from https://github.com/Automattic/jetpack/pull/30398

### Other information:

## Jetpack product discussion

p1HpG7-lSW-p2

No

## Testing instructions:
- Run `bin/jetpack-downloader test jetpack add/make-earn-products-free-for-all` on your sanbox
- Sandbox public-api.wordpress.com and subscribe.wordpress.com
- Either create a site via WPCOM (and sanbox it), or use jetpack-beta.
- Connect Jetpack free version
- Create a donation block in a post
- Test it
Result:
https://total-antelope.jurassic.ninja/2023/05/08/donate/

See my videos:

https://user-images.githubusercontent.com/790558/236858249-fff8db35-9be5-4ea7-bd5d-faf4949ad4cf.mov

https://user-images.githubusercontent.com/790558/236857412-77ef764a-d761-4fe1-aaf3-5cfa44bb6373.mov


millerf:
@jeherve or @romarioraffington 
Do you have any documentation on those files? How come this seems to duplicate whatever is in `class-wpcom-features.php`. I understand the idea behind it, but it feels like there should be only one source of truth, especially for this kind of features...

jeherve (changes requested):
This doesn't seem to work for me:

**WordPress.com Simple**

image

**Self-hosted**

When trying to purchase something:

image

I assume this will require changes on the WordPress.com side as well.

> what would the best way to test this?

I would say loading each block (and widget in the case of Simple Payments) and attempting to use them on each platform.

********

We may be able to remove some other things to simplify things further. No need for any plan check at all when we know they're free.
https://github.com/Automattic/jetpack/blob/41f3afbb906bcda8f933194e95414afa6ea69539/projects/plugins/jetpack/modules/simple-payments/simple-payments.php#L223
https://github.com/Automattic/jetpack/blob/3a3c4adc1c0913bac224ddb832713bd6101fe1ab/projects/plugins/jetpack/extensions/blocks/simple-payments/simple-payments.php#L28
https://github.com/Automattic/jetpack/blob/daab12c35c81cf407f626981f495220021ad90fe/projects/plugins/jetpack/extensions/blocks/donations/donations.php#L28
https://github.com/Automattic/jetpack/blob/ba6a8391edb84d72637cb6a9fd9033b0093a2f4f/projects/plugins/jetpack/modules/memberships/class-jetpack-memberships.php#L569-L589
https://github.com/Automattic/jetpack/blob/ba6a8391edb84d72637cb6a9fd9033b0093a2f4f/projects/plugins/jetpack/modules/memberships/class-jetpack-memberships.php#L500
https://github.com/Automattic/jetpack/blob/ba6a8391edb84d72637cb6a9fd9033b0093a2f4f/projects/plugins/jetpack/modules/memberships/class-jetpack-memberships.php#L510
https://github.com/Automattic/jetpack/blob/893632f924f9b06071978cc253af8ab5719d0997/projects/plugins/jetpack/extensions/blocks/recurring-payments/edit.jsx#L54-L58
https://github.com/Automattic/jetpack/blob/893632f924f9b06071978cc253af8ab5719d0997/projects/plugins/jetpack/extensions/blocks/recurring-payments/edit.jsx#L119-L137

millerf:
@jeherve 

> I assume this will require changes on the WordPress.com side as well.

Everything should be set on WPCOM side already.

jeherve:
> Do you have any documentation on those files? How come this seems to duplicate whatever is in `class-wpcom-features.php`. I understand the idea behind it, but it feels like there should be only one source of truth, especially for this kind of features...

Yeah, it's a bit messy. `Current_Plan` and the Plans package was supposed to replace `Jetpack_Plan`, but this wasn't completed yet. It is supposed to match the class on WordPress.com indeed.

millerf:
@jeherve I did work on it today. It seems to work. I'll update the testing instructions so it makes it easier to test.

projects/plugins/jetpack/extensions/shared/components/stripe-nudge/index.jsx
```diff
@@ -17,7 +17,7 @@ export const StripeNudge = ( { blockName } ) => {
 			block: blockName,
 		} );
 
-	if ( store.getShouldUpgrade() || ! stripeConnectUrl ) {
```
jeherve, line 20: I think we'll need to update the tests accordingly before we can merge:
https://github.com/Automattic/jetpack/actions/runs/4928397209/jobs/8806728553?pr=30432

projects/plugins/jetpack/modules/memberships/class-jetpack-memberships.php
```diff
@@ -492,23 +491,8 @@ public static function user_can_view_post_nocache() {
...
-
-	/**
-	 * Whether the site's plan supports the Recurring Payments block.
-	 */
-	public static function is_supported_jetpack_recurring_payments() {
-		$api_available     = ( ( defined( 'IS_WPCOM' ) && IS_WPCOM ) || Jetpack::is_connection_ready() );
-		$supported_in_plan = Jetpack_Plan::supports( 'recurring-payments' );
-		return apply_filters( 'test_jetpack_is_supported_jetpack_recurring_payments', $api_available && $supported_in_plan );
+		$api_available = ( ( defined( 'IS_WPCOM' ) && IS_WPCOM ) || Jetpack::is_connection_ready() );
+		return apply_filters( 'test_jetpack_is_supported_jetpack_recurring_payments', $api_available );
```
jeherve: Maybe we can remove the filter entirely now?

jeherve (changes requested):
This tests well for me. Once the tests are fixed I think we should be good to go!

projects/plugins/jetpack/modules/memberships/class-jetpack-memberships.php
millerf: Nice catch!

millerf: Fixed in 50ff034069f2b4f23b7b382277960bd9f3e7cf0a

jeherve (dismissed):

jeherve (approved):

PR #30437, opened 2023-05-03
Forms: Improve Multiple/Single choice fields sidebar style settings

jcheringer:
## Proposed changes:
Update Multiple Choice and Single Choice fields Sidebar style settings: 
- Rename some labels to match the elements on the MC/SC fields
- Hide some style options when the Form style is Default or Animated

Form style: Default or Animated
image

Form style: Outlined
image

### Other information:

## Jetpack product discussion
N/A

No

## Testing instructions:

Make sure you have the Forms package enabled

* Create a post, add a Form block and a MC/SC field
* Check if the Sidebar works as describe above

### DIFF
--- a/projects/packages/forms/changelog/update-forms-mc-sc-sidebar
+++ b/projects/packages/forms/changelog/update-forms-mc-sc-sidebar
@@ -1,0 +1,4 @@
+Significance: minor
+Type: changed
+
+Forms: Update Multiple Choice and Single Choice fields Sidebar style settings
--- a/projects/packages/forms/src/blocks/contact-form/components/jetpack-field-controls.js
+++ b/projects/packages/forms/src/blocks/contact-form/components/jetpack-field-controls.js
@@ -50,15 +50,15 @@ const JetpackFieldControls = ( {
 
 	const inputColorLabel = isChoicesBlock
 		? __( 'Option Text', 'jetpack-forms' )
-		: __( 'Field Text', 'jetpack-forms' );
+		: __( 'Field Text', 'jetpack-forms', 0 );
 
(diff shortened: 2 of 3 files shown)


digitalwaveride (commented):
Thanks so much João!!! 😻 

It works as advertised! Nice!

Unfortunately though we might have missed something:
The styling for the new button style?


For the Color category we would need a "Button Background".
For the Option Styles a "Border Radius" and "Border Width"
image


Do you want to do this in a separate PR?

jcheringer:
> Thanks so much João!!! 😻
> 
> It works as advertised! Nice!
> 
> Unfortunately though we might have missed something: The styling for the new button style?
> 
> For the Color category we would need a "Button Background". For the Option Styles a "Border Radius" and "Border Width" 
> Do you want to do this in a separate PR?

Yeah, I'm not sure if that should be part of this PR.
The idea here was to adjust the Sidebar, so it'd make sense.

If we want to add a background option for the buttons, we also need to think about how we will handle the 'selected' state.
Right now, the selected state mimics the Primary Button style. If the user changes the background color for the default state, they might want to change it for the selected state too, as the primary style might not fit their intended design.

Regarding the "Border Radius" and the "Border width", these settings already exist.
They are visible for the MC/SC field when the Form is in the Outlined state, but they affect the block 'container'.
So, if we want to add these settings for the Buttons style, we need to think about how we'll present them to the user, especially for the Outlined style.

digitalwaveride (approved):
As discussed in our call yesterday, let's address the "button styling" in a separate PR.

Thanks again João!
🚀

PR #30440, opened 2023-05-03
My Jetpack: Remove manage option from menu

### LINKED ISSUE #30396: My Jetpack: duplicate Manage option being showed as primay action + action menu option
When a hybrid product is enabled and ready for use, the "Manage" option is being showed twice, as the primary action and also as an option on the actions menu:

Screen Shot 2023-05-01 at 18 01 43

This will only happen for hybrid products (VideoPress, Backup, Search and Social) and represents a different behavior from the other products.

How to replicate:

- on a site with Jetpack, enable VideoPress, install and enable the Jetpack VideoPress plugin and visit `Jetpack > My Jetpack`
- notice the option being showed twice

This also happens when the module is already active but the plugin is not installed/active (notice the two "Manage" options):

Screen Shot 2023-05-01 at 18 00 23

We may not need both actions being available.

renatoagds:
Closes #30396 

## Proposed changes:

* Remove the default `Manage` option from the menu.
* Update logic to make the menu visible.

### Other information:

## Jetpack product discussion

N/A

N/A

## Testing instructions:

* Start `My Jetpack`
* Make sure that you have only `Jetpack` plugin installed.
* Check if the `Install or Activate plugin` is the only action in the cards (`VideoPress`, `Backup`, `Search` and `Social`).

## Demo

CleanShot 2023-05-03 at 16 40 34@2x

### DIFF
--- a/projects/packages/my-jetpack/_inc/components/connected-product-card/index.jsx
+++ b/projects/packages/my-jetpack/_inc/components/connected-product-card/index.jsx
@@ -25,22 +25,22 @@ const ConnectedProductCard = ( { admin, slug, children, showMenu = false, menuIt
 
 	/* Menu Handling */
 	const hasStandalonePlugin = standalonePluginInfo?.hasStandalonePlugin;
-	const isConnected = isRegistered && isUserConnected;
 	const isStandaloneInstalled = standalonePluginInfo?.isStandaloneInstalled;
 	const isStandaloneActive = standalonePluginInfo?.isStandaloneActive;
+	const showActivateOption = hasStandalonePlugin && isStandaloneInstalled && ! isStandaloneActive;
+	const showInstallOption = hasStandalonePlugin && ! isStandaloneInstalled;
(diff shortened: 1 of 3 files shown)


lhkowalski:
Tested it using the VideoPress card.

When the product is inactive and the plugin is not installed, I see only the install option:

Screen Shot 2023-05-03 at 18 32 51

When the module is active on Jetpack but the plugin is not installed, I see only the install option again:

Screen Shot 2023-05-03 at 18 26 34

When the module is active on Jetpack, the plugin is installed but not active, I see only the activate option:

Screen Shot 2023-05-03 at 18 29 45

After activating the plugin, since there is no other menu option to show, the menu is gone:

Screen Shot 2023-05-03 at 18 29 57

It's working the right way for the other products too.

lhkowalski (approved):
Code looks good. Good work on moving forward the refactor of the flags that control the menu.

PR #22912, opened 2022-02-15
Subscribe Block: Add pre/post-publish panels

jsnmoon:
Built off of #22694, will change base branch to `master` once it has been merged. Feature requested by @mtias.

#### Changes proposed in this Pull Request:

* Initializes a `subscriptions` Gutenberg plugin to enable pre/post publish panels.
* Adds pre/post publish sidebar panels indicating the number of subscribers.
<table>
<tr>
<td>Screen Shot 2022-02-15 at 4 07 22 PM
<td>Screen Shot 2022-02-15 at 4 07 09 PM
</table>

#### Does this pull request change what data or activity we track or use?

No.

#### Testing instructions:

* Apply this change to your connected test site.
* Enable subscriptions in `Jetpack` -> `Settings` -> `Discussions`.
* Write a new post and press "Publish."
* Ensure that the new pre-publish panel is present.
* Press "Publish" again to confirm.
* Ensure that the new post-publish panel is present.

PR #30446, opened 2023-05-04
Refactor and Improve Paid Newsletter User Experience

romarioraffington:
### Notes
This PR refactors the [panel.js](https://github.com/Automattic/jetpack/compare/add/update-newsletter-panel-design?expand=1#diff-582aef850107fc9949275668f0ca61696c0e19729528837a5becf25c9229cdbb) file and the [settings.js](https://github.com/Automattic/jetpack/compare/add/update-newsletter-panel-design?expand=1#diff-eba2301ede604e2450a0b3ea7d13cdc88d427f67d0f1777028aa9e2e154623c3) files. Both files are used by the **Subscription block** and the **Paid Newsletter**.  While making the changes, it was difficult to determine the Subscription block logic from the Paid Newsletter. Nevertheless, I went ahead with the refactor as it would be best to refactor this code before launch since it posed less of a risk.  I'm sure this PR has broken some of the existing Subscription block logic.

@jeherve, @kozer, @jsnmoon, based on git blame, you all have added Subscription block logic to one of the two files mentioned above. To prevent this from re-occurring, I would appreciate any suggestions on decoupling the Paid Subscription logic from the Newsletter logic. This would allow us to quickly iterate the Paid Newsletter features without breaking the Subscription block in the future

@millerf I removed misconfiguration logic for now since there are no designs for these errors at the moment. Ultimately, I think it would be best to re-add the logic when the direction of these notices are clear.

Fixes https://github.com/Automattic/gold/issues/23
Fixes https://github.com/Automattic/gold/issues/28
Related PR: D109789-code

## Proposed changes:

* Refactor Paid Newsletter to make it easier to make changes
* Update Paid Newsletter Flow and User Experience based on designs

See p81Rsd-1iu-p2 for more context.

### Other information:

## Jetpack product discussion

No

No

## Testing instructions:

**Paid Newsletter**

1. Verify the UI and UX improvements matches the designs
2. Perform various sanity checks to ensure nothing major has broken

**Subscription Block**

_Pending Feedback from the JP team_


projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
```diff
@@ -67,18 +70,88 @@ function AccessLevelSelectorPanel( {
...
+} ) {
+	const { tracks } = useAnalytics();
+	const { changeStatus, isLoadingModules, isChangingStatus } = useModuleStatus( name );
+
+	const enableSubscriptionsModule = () => {
+		tracks.recordEvent( 'jetpack_editor_subscriptions_enable' );
+		return changeStatus( true );
+	};
+
+	// Subscriptions will not be triggered for a post that was already published in the past.
```
simison, line 116: Does this mean that the panel becomes hidden after publishing? 

If so, I think it should stay visible and allow changing the access setting, or at least allow seeing what was the setting.

projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
```diff
@@ -67,18 +70,88 @@ function AccessLevelSelectorPanel( {
...
+						isBusy={ isChangingStatus }
+						disabled={ isModuleActive || isChangingStatus }
+						onClick={ enableSubscriptionsModule }
+					>
+						{ isChangingStatus
+							? __( 'Activating Subscriptions', 'jetpack' )
+							: __(
+									'Activate Subscriptions',
+									'jetpack',
+									/* dummy arg to avoid bad minification */ 0
```
simison, line 160: Is there any more context on this comment?

projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
```diff
@@ -91,196 +164,35 @@ export default function SubscribePanels() {
...
-
-	// Only show this for posts for now (subscriptions are only available on posts).
-	const postWasEverPublished = useSelect(
-		select =>
-			select( editorStore ).getEditedPostAttribute( 'meta' )?.jetpack_post_was_ever_published,
-		[]
-	);
-
-	// Subscriptions will not be triggered on private sites (on WordPress.com simple and WoA),
+	// Subscriptions will not be triggered on private sites ( on WordPress.com simple and WoA ),
```
simison: Also P2s should be excluded, even if they're not set to private?

projects/plugins/jetpack/extensions/blocks/subscriptions/settings.scss
```diff
@@ -14,3 +14,21 @@
 	padding: 1em;
 	text-align: justify;
 }
+
+.edit-post-post-visibility {
```
simison, line 21: Do we need to modify core blocks, using core classes? These tend to be fragile long term.

If we must, I'd at least try adding custom classname prefixed with `jetpack-`.

Also a nit but `0px` can just be written `0`.

projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
```diff
@@ -3,96 +3,232 @@ import { getRedirectUrl } from '@automattic/jetpack-components';
...
 				__(
-					'Private or password-protected posts cannot be assigned a newsletter setting of "%1$s". Please update the setting to "%2$s", or update the post visibility setting.',
+					'Restrict your post to subscribers. <learnMoreLink>Learn more.</learnMoreLink>',
 					'jetpack'
 				),
-				accessOptions[ accessLevel ].label,
-				accessOptions.everybody.label
+				{
+					learnMoreLink: (
+						<Link
```
simison, line 64: Why not just `ExternalLink` from `@wordpress/components`?

projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
```diff
@@ -3,96 +3,232 @@ import { getRedirectUrl } from '@automattic/jetpack-components';
...
+					{ createInterpolateElement(
+						__(
+							'You don’t have any subscribers yet. How about <importingLink>importing</importingLink> some? Or check out <thisGuideLink>this guide</thisGuideLink> on how to grow your audience.',
+							'jetpack'
+						),
+						{
+							importingLink: (
+								<Link
+									slug={
+										'https://wordpress.com/support/launch-a-newsletter/import-subscribers-to-a-newsletter/'
```
simison: We should localize the link. There's a function for it but can't remember now how to do that in Jetpack.

projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
```diff
@@ -3,96 +3,232 @@ import { getRedirectUrl } from '@automattic/jetpack-components';
...
+									}
+								/>
+							),
+							thisGuideLink: (
+								<Link slug={ 'https://wordpress.com/support/category/grow-your-audience' } />
```
simison: Same as above support links should be localized.

projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
```diff
@@ -3,96 +3,232 @@ import { getRedirectUrl } from '@automattic/jetpack-components';
...
+	}
+
+	if ( accessOptions.everybody.key === accessLevel ) {
+		reachCount = reachCount + '+'; // Concat "+"
+	}
+
+	return (
+		<FlexBlock>
+			<Notice status="info" isDismissible={ false } className="edit-post-post-visibility__notice">
+				{ __( 'This will be sent to ', 'jetpack' ) }
```
simison: No partial strings; they don't work for translating.

See PCYsg-3hI-p2#partial-strings-and-concatenation

projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
romarioraffington: This code is already on `trunk`. I suspect this is for minification purposes. I can clarify with the JP team

projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
romarioraffington: I agree! I think this boolean only gates the Subscription Block code. It shouldn't impact the Paid Newsletter

projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
romarioraffington: I went with this approach because the `ExternalLink` includes an icon that the designs doesn't have.  I felt this was an easier approach. Open to changing it!

simison: @crisbusquets what do you think about having the "external link" icon at least in some of the links that open in new tabs? It's not a strict convention in Gutenberg but used a lot. Makes it less scary to click while they're editing the content, since they know it opens a new tab. :-)

Example:

Screenshot 2023-05-04 at 11 37 57

crisbusquets: @simison From an accessibility point of view, we should always identify links that open on new tabs/windows. Having said that, I'm concerned that we would have too many icons. Could you try this to see how it would look?

n3f:
Fix for notices background: #30473

projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
jeherve: Using Jetpack Redirects can help here:
PCYsg-pY7-p2

millerf: This is done

projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
millerf: Fixed in e56f586038a32690555beee5e0c5243156ea1315

projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
millerf: Done

projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
millerf: I tried. It looks pretty bad.

projects/plugins/jetpack/extensions/blocks/subscriptions/settings.scss
millerf: Done in d5ade0bd7fe4739f48d11cb685f6168f98dc5638

Noting here that it does not really matter as the CSS compiler will translate to 0 anyway.

projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
millerf: > If so, I think it should stay visible and allow changing the access setting, or at least allow seeing what was the setting.

This issue happened because the newsletter has been set already and therefore it would make little sense to change its reach.

But I definitely agree with changing the visibility of the post after publish.

Maybe we could simply change the wording and keep the panel displayed. @crisbusquets ?

@simison I'll add a separate task once this is finalized.

millerf (dismissed):
WFM 🎉

projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
```diff
@@ -43,42 +38,124 @@ const SubscriptionsPanelPlaceholder = ( { children } ) => {
...
+	};
+
+	// Subscriptions will not be triggered for a post that was already published in the past.
+	const shouldLoadSubscriptionPlaceholder = useSelect( select => {
+		const meta = select( editorStore ).getEditedPostAttribute( 'meta' );
+		return ! isModuleActive && ! isLoadingModules && ! meta?.jetpack_post_was_ever_published;
+	} );
+
+	return (
+		<PluginPrePublishPanel
```
jeherve, line 123: This panel ends up showing, but without anything in it on self-hosted sites where the module is enabled:

image

Maybe we should hide it altogether unless either the module is off, or the newsletter feature is enabled?

jeherve (dismissed):
This looks kinda good now. It does remove the logic that was first added in #22912, but I assume that's because it wasn't on the planned design?

I noticed an issue on self-hosted sites, noted it below.

millerf:
> This looks kinda good now. It does remove the logic that was first added in https://github.com/Automattic/jetpack/pull/22912, but I assume that's because it wasn't on the planned design?

That is a good remark. AFAIK there was no talk to have a post-publish panel. @crisbusquets what do you think? I will treat it in a separate PR in case we need one.

projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
millerf: Good point. Let me look into it...

millerf: I prevent displaying any panel if the feature is not enabled.

crisbusquets:
> That is a good remark. AFAIK there was no talk to have a post-publish panel. @crisbusquets what do you think? I will treat it in a separate PR in case we need one.

Hi @millerf,
I wasn't aware of this, sorry. I'll work on a proposal today/tomorrow.

Thanks for the heads up!

simison:
@crisbusquets @millerf I think the post-publish panel can be there as it is today? Improvements to it can be follow ups, but let's not hide it indeed.

crisbusquets:
@millerf @simison 

Here's the post-publish panel design (on the right side). [Figma link](h8tMpJlCqNFemEerU9owHI-fi-333_5053).

I would add the "Newsletter: All Subscribers" header to make it more clear who you have sent this to.

postpublish-checklist

crisbusquets:
Feedback (cc @millerf, @romarioraffington ).

First of all, good work everyone 🙌🏼

I tested this PR and all the links and only saw two things:

**Alert message color**
I noticed that the notice alert (the blue one) has a different left border color depending on my user theme. This doesn't happen with other notices, like warning (screenshot attached). The blue one needs to keep the left border color blue — it's a semantic notice, and the blue color shouldn't change.

notice


**Disabled "Paid subscribers" radio button**
When I don't have any paid subscribers, and I hover over that radio button, the cursor is a `pointer`. Would it be possible to change it to `no-drop` (the one that's a pointer but also has a forbidden symbol) if that option is disabled?

Thank you!

PD: Let me know if you'd prefer this feedback somewhere else or more detail.

millerf:
> This doesn't happen with other notices, like warning (screenshot attached). The blue one needs to keep the left border color blue — it's a semantic notice, and the blue color shouldn't change.

This is the normal behaviour of the Notice component. The color changes depending on the theme.
The background color should adapt as well. But, because the normal background is white and we had to override it with `blue`, it doesn't change anymore.
I'll try and get some kind of fix. 

> When I don't have any paid subscribers, and I hover over that radio button, the cursor is a pointer. Would it be possible to change it to no-drop (the one that's a pointer but also has a forbidden symbol) if that option is disabled?

Sure thing.

millerf:
@simison 

It is not hidden. It has been removed during the refactor.

millerf:
@crisbusquets 
- Added the `no-drop` cursor
- Slightly greyed-out the disabled option so it looks more obvious
- Forced the blue colors on the notice, independently of the theme

Let me know what you think...

Screenshot 2023-05-11 at 13 07 06

millerf (dismissed):
LGTM :)

pablohoneyhoney:
@crisbusquets some notes based on what I'm seeing in the Figma file, and above. 

Semantics and wording, and refine a few details in the UI. nothing blocking for a v1, but worth taking a look at my notes here https://www.figma.com/file/h8tMpJlCqNFemEerU9owHI/Paid-Newsletters-flows?type=design&node-id=333%3A5053&t=Tp4OzyZniy4g9iR1-1, and discuss them if I'm missing context or previous discussions.

projects/plugins/jetpack/extensions/blocks/subscriptions/paid-plan.js
```diff
@@ -1,36 +0,0 @@
-import { PanelBody, PanelRow } from '@wordpress/components';
-import { compose } from '@wordpress/compose';
-import { withSelect } from '@wordpress/data';
-import { __, _x } from '@wordpress/i18n';
-import GetAddPaidPlanButton from './utils';
-
-function PaidPlanPanel( { hasNewsletterPlans } ) {
```
romarioraffington, line 7: Should we re-add this?  I suspect we still need this for the Subscribe block?

projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
```diff
@@ -157,133 +300,146 @@ export function NewsletterAccess( {
...
-							{ ! showVisibilityRestrictedMessage && ! withModal && canEdit && (
+	const _accessLevel = accessLevel ?? accessOptions.everybody.key;
+	const accessLabel = accessOptions[ _accessLevel ]?.label;
+
+	return (
+		<PostVisibilityCheck
+			render={ ( { canEdit } ) => (
+				<PanelRow className="edit-post-post-visibility">
+					<Flex direction="column">
+						{ showMisconfigurationWarning && MisconfigurationWarning() }
```
romarioraffington, line 422: Should we render `MisconfigurationWarning` as a React component verses calling it as function here?

millerf: Good point. I missed that.

projects/plugins/jetpack/extensions/blocks/subscriptions/paid-plan.js
millerf: Do we?

projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
millerf: Yes, and it [this is how it is right now](https://github.com/Automattic/jetpack/pull/30446/files#diff-eba2301ede604e2450a0b3ea7d13cdc88d427f67d0f1777028aa9e2e154623c3R318-R319). You are looking at an outdated version (not sure how you got there).

millerf (approved):

PR #30448, opened 2023-05-04
Boost: Update Minify to include latest page optimize updates

dilirity:
## Proposed changes:

* During the migration of Page Optimize into Boost, there were some updates made to [Page Optimize's codebase](https://github.com/Automattic/page-optimize/commits/master). This PR adds those changes to Minify.

### Other information:

## Jetpack product discussion

p1683185893703879-slack-C016BBAFHHS

n/a

## Testing instructions:

* Install the plugin;
* Turn on debugging;
* Turn on CSS/JS concatenate;
* View a few pages to generate cache;
* Check logs to make sure there are no Minify related warnings/notices.

### DIFF
--- a/projects/plugins/boost/app/lib/minify/functions-service.php
+++ b/projects/plugins/boost/app/lib/minify/functions-service.php
@@ -235,7 +235,7 @@ function jetpack_boost_page_optimize_build_output() {
 			if ( 0 === strpos( $buf, '@charset' ) ) {
 				preg_replace_callback(
 					'/(?P<charset_rule>@charset\s+[\'"][^\'"]+[\'"];)/i',
-					function ( $match ) use ( $pre_output ) {
+					function ( $match ) use ( &$pre_output ) {
 						if ( 0 === strpos( $pre_output, '@charset' ) ) {
 							return '';
 						}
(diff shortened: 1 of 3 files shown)


thingalon (approved):

thingalon (dismissed):

projects/plugins/boost/app/lib/minify/functions-service.php
```diff
@@ -238,7 +238,7 @@ function jetpack_boost_page_optimize_build_output() {
 					function ( $match ) {
 						global $pre_output;
 
-						if ( 0 === strpos( $pre_output, '@charset' ) ) {
+						if ( 0 === strpos( (string) $pre_output, '@charset' ) ) {
```
pyronaur: I think a guard clause for non-strings would be better here

pyronaur (changes requested):

projects/plugins/boost/app/lib/minify/functions-service.php
dilirity: I've reworked how the variable is used in aff554fb6b5d1a2c21116df82d3c8456db55e98a, so casting is no longer necessary.

pyronaur (dismissed):

pyronaur (approved):

PR #30450, opened 2023-05-04
WP Super Cache: make preload faster and more robust

donnchawp:
The preload function in WP Super Cache allows a website owner to prime the cache on their site by creating cache files for all the taxonomy archives (first page), pages, and posts. It does this by fetching each page separately.

* Instead of getting 100 pages at a time, it will request 10 instead. This is because some sites are very slow or have pages that are slow to generate. If you're using FastCGI then set_time_limit() doesn't reset the read timeout.
* It will preload all taxonomies first now. This avoids a problem where you have lots of taxonomies and not enough posts to run the preload loop enough times to preload everything. The preload loop used to run on the number of posts on the site.
* I removed many pauses as it's only 10 pages per preload loop now, fetched sequentially, with a 3 second pause in between.
* The preload notification JS only requests every 3 seconds now as there's a 3 second wait between jobs, it avoids requesting a notification with the same filename over and over again.

https://wordpress.org/support/topic/preload-mode-skips-caching-pages-or-deletes-already-cached-pages/#post-16706554

## Testing instructions:
You can copy wp-cache.php from this branch into a live site. Also copy inc/preload-notifcation.js but that was a very minor change.
Run a preload and note that taxonomies are cached first.

### Other information:

No


projects/plugins/super-cache/wp-cache.php
```diff
@@ -3214,10 +3212,16 @@ function wp_cron_preload_cache() {
...
+			$preload_more_taxonomies = false;
+
 			foreach( $taxonomies as $taxonomy => $path ) {
 				$taxonomy_filename = $cache_path . "taxonomy_" . $taxonomy . ".txt";
-				if ( $c == 0 )
+
+				if ( ! get_transient( 'taxonomy_preload' ) ) {
 					@unlink( $taxonomy_filename );
+				}
+				set_transient( 'taxonomy_preload', 1, 600 );
```
thingalon, line 3242: I think it would be worth adding a comment explaining the purpose of `'taxonomy_preloead'` to make this code easier to understand on future reads.

projects/plugins/super-cache/wp-cache.php
```diff
@@ -3271,14 +3275,39 @@ function wp_cron_preload_cache() {
...
+				) {
+					$preload_more_taxonomies = true;
+				}
+			}
+
+			if ( $preload_more_taxonomies === true ) {
+				if ( defined( 'DOING_CRON' ) ) {
+					wp_cache_debug( 'wp_cron_preload_cache: scheduling the next taxonomy preload in 3 seconds.' );
+					wp_schedule_single_event( time() + 3, 'wp_cache_preload_hook' );
+					wp_delete_file( $mutex );
```
thingalon: Is it worth taking this opportunity to move the scheduling and mutex stuff into its own function, to separate it from the work of preloading - to make the code easier to maintain in future?

projects/plugins/super-cache/wp-cache.php
```diff
@@ -3271,14 +3275,39 @@ function wp_cron_preload_cache() {
...
 		}
 	}
 
+	update_option(
+		'preload_cache_counter',
+		array(
+			'c' => ( $c + 10 ),
+			't' => time(),
+		)
+	);
```
thingalon: It would also be neat to make less code need to understand the difference between `'c'` and `'t'` in the preload option.

projects/plugins/super-cache/wp-cache.php
```diff
@@ -3294,7 +3323,8 @@ function wp_cron_preload_cache() {
 			$page_on_front = $page_for_posts = 0;
 		}
 		if ( $wp_cache_preload_email_me && $wp_cache_preload_email_volume == 'many' )
-			wp_mail( get_option( 'admin_email' ), sprintf( __( '[%1$s] Refreshing posts from %2$d to %3$d', 'wp-super-cache' ), home_url(), $c, ($c+100) ), ' ' );
+		/* translators: 1: home url, 2: start post id, 3: end post id */
+			wp_mail( get_option( 'admin_email' ), sprintf( __( '[%1$s] Refreshing posts from %2$d to %3$d', 'wp-super-cache' ), home_url(), $c, ( $c + 10 ) ), ' ' );
```
thingalon: Might be worth making a constant / define / variable which sets the size of preloads, so that:
a) Each instance of `10` or `100` in code is easier to understand because it will be a label like `PRELOAD_POST_COUNT`, and 
b) Any changes in future to preload size become easier.

thingalon (commented):
I hope you don't mind, but I have some suggestions for changes we can make now while in this bit of code which will push us towards a more maintainable codebase :)

projects/plugins/super-cache/wp-cache.php
donnchawp: Good point. I added comments above this line explaining what it does, which I hope helps, but the code could be tightened up a bit - when the preload is cancelled in either loop, the code is duplicated. I'll put that in it's own function.

projects/plugins/super-cache/wp-cache.php
donnchawp: Good idea. Just added that!

projects/plugins/super-cache/wp-cache.php
donnchawp: Done!

projects/plugins/super-cache/wp-cache.php
thingalon: Thanks. I think that helps. I would like to move it all into a couple of functions that know how to deal with that data, but we can make that a separate PR. :)

thingalon (approved):
Looks good to me :) let's merge it

PR #30458, opened 2023-05-04
My Jetpack: Remove icon from plugin activation action in product card

### LINKED ISSUE #30395: My Jetpack: improve icon choice for the action menu options
We need to review the icon choice on the action menu to make sure they are helping the user:

- showing the Manage action menu with the external icon seems confusing
- the checkmark next to “Activate plugin” can be confused with a toggle which is currently on; the user may think it is telling that the plugin is active, and that clicking it would deactivate it

Context:

- pedxs5-ck-p2

dhasilva:
Fixes #30395

## Proposed changes:

* Just removes the icon for now. There is no state where there are more than one action, including the `Activate` action, so the center text alignment does not contrast with other options.
* The `Manage` action was removed entirely in a previous PR.

### Other information:

## Testing instructions:

* Go to My Jetpack with Search, VideoPress or Social plugin installed, but not activated
* Check that the Activation action in the actions menu does not have an icon

### DIFF
--- a/projects/packages/my-jetpack/_inc/components/product-card/index.jsx
+++ b/projects/packages/my-jetpack/_inc/components/product-card/index.jsx
@@ -1,7 +1,7 @@
 import { Button, Text } from '@automattic/jetpack-components';
 import { Dropdown } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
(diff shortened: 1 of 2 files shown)


retrofox (approved):
LGTM

PR #30469, opened 2023-05-04
Fix link area glith on forms dashboard

### LINKED ISSUE #30330: JP Forms - WP-Admin Form Responses - optimizing link space for source column
Originally [reported here](pcsBup-1aJ-p2#comment-417):

> It would seem that one gets redirected to the form’s page even when clicking next to the link, elsewhere on the row. See the screencast below in Google Chrome:

The link on the source column takes the full width of the column. That can be confusing as the full row in most places is for selecting the row, to see the details in the right panel. 

The link should be limited to just the text.

Image

ice9js:
Fixes #30330.

## Proposed changes:

This fixes an issue where the hitbox for source links on the forms dashboard response list takes up the entire column width, even if the text is shorter. After the change, it should only cover the actual text.

### Other information:

## Jetpack product discussion

pcsBup-1aJ-p2#comment-417

No.

## Testing instructions:

- Go to `/wp-admin/admin.php?page=jetpack-forms`
- The links in the source column should only trigger when clicking on the text.
- When clicking outside of the text but swill within the source column, it should only select that response.
- Overflow ellipsis should still work as expected on narrower screens.

### DIFF
--- a/projects/packages/forms/changelog/fix-forms-dashboard-link-area
+++ b/projects/packages/forms/changelog/fix-forms-dashboard-link-area
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fixed the hitbox for the source link on the forms dashboard response list.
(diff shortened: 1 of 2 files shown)


CGastrell (approved):
Works nicely! Thanks Kuba :shipit:

PR #30470, opened 2023-05-04
CRM: Resolves 3081 - expand log types that update Last Contacted timestamp

tbradsha:
Resolves Automattic/zero-bs-crm#3081 - expand log types that update Last Contacted timestamp

## Proposed changes:

Being a CRM, it's important that the "Last Contacted" date is accurate. Our system updates that when a specific log type is added against a contact. This PR updates those types to include Facebook, Tweet, Feedback, and a new "Other contact" log type, so that now the following log types will update the last contacted timestamp:

`public $contact_log_types = array( 'call', 'email', 'mail', 'meeting', 'feedback', 'invoice__sent', 'quote__sent', 'feedback', 'tweet', 'facebook_post', 'other_contact' );`

I also:

* updated the above array to use keys instead of friendly names, as previously the code was trying to lowercase+compare to keys, but this didn't match when there were spaces or special chars
* chopped some red herring code that was outdated
* fixed a bug that occurred when trying to submit two logs in a row

This became a coding standards rabbit hole; sorry in advance to anyone reviewing. Logic should have remained largely untouched (other than the first point above and an obsolete DAL check I removed).

### Other information:

## Testing instructions:

1. Create a contact.
2. Pull up the contact's `zbsc_lastcontacted` value in the database.
3. Add a "Note" log to the contact. Observe the database value.
4. Add an "Email" log to the contact. Observe the database value.
5. Add a "Facebook Post" log to the contact. Observe the database value.

In `trunk`, you'll notice only step 4 updates the database value. Also, when adding a second log (without a page refresh), the log type dropdown will be blank.

In the `fix/crm/3081-update_last_contacted_date_when_appropriate` branch, step 5 also updates the database value. Try adding an "Other contact" log to the contact, which should also update the database value. After the initial log add, the dropdown will reset to the default "Note" log type when starting to add the next log.

6. Add a new contact with a "Meeting" log type.
7. Add a second contact with a "Tweet" log type.
8. Go to the list view: `/wp-admin/admin.php?page=manage-customers`
9. Add the "Last Contacted" column to the list view settings.
10. Sort by said column.

In `trunk`, the column will show "Not contacted" and the new contact will show in the second position.

In the `fix/crm/3081-update_last_contacted_date_when_appropriate` branch, the column will show "Today", and will properly show in the first position.


coder-karen (approved):
Looks good to me! Tests work well.

Heads up we've duplicated some linting efforts in `DAL3.Obj.Contacts.php` from what I can see (see `30178` ) so depending on which PR makes it over the line first the other may end up with possible merge conflicts.

PR #30472, opened 2023-05-04
Launchpad: migrate business logic to Task definitions

mattwiebe:
See Automattic/wp-calypso#75504

## Proposed changes:
This finishes work begun in #30397 to consolidate all task-completing logic from the old mu-plugin to the Task registry, using `add_listener_callback`.

### Other information:

- [x] Have you written new tests for your changes, if applicable? **Fixed tests **

## Jetpack product discussion
n/a

No.

## Testing instructions:

1. apply this to your wpcom sandbox
2. sandbox `public-api.wordpress.com`
3. sandbox the URL of the site **you create before you create it** (eg `wiebelaunchpadwrite.wordpress.com`)

With each type of Launchpad-supporting site, go through the full flow and ensure that:

1. All tasks are marked off when they should be
2. Launching works

## Known issues
The VideoPress upload detector does not work properly. I've added a workaround in 284c921e4726976fec70ddd12281e69d5376652e that mimics the old mu-plugin logic that worked. Unfortunately this can't be tested until this lands in production because it needs to be on job servers.

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/changelog/update-migrate-launchpad-business-logic
+++ b/projects/packages/jetpack-mu-wpcom/changelog/update-migrate-launchpad-business-logic
@@ -1,0 +1,4 @@
+Significance: minor
+Type: changed
+
+Migrates Launchpad business logic from the mu-plugin to being defined by tasks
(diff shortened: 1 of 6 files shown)


Addison-Stavlo (commented):
tested with most recent changes for:

✅ Free, Write, Build, Newsletter, Link-in-bio flows
❌  videopress upload video task doesnt seem to be working.

syncing this jetpack build to sandbox
sandboxing public-api and test sites
preventing previous wpcom_launchpad class from initializing (shouldnt be necessary but extra safety)

more testing discussion context in this slack thread p1683237029135389-slack-CHN6J22MP

Addison-Stavlo:
The videopress changes seem like they should work.  Matt mentions a related idea here p1683312690247109/1683237029.135389-slack-CHN6J22MP they they may require being in production.

Addison-Stavlo (approved):
All tests well except that one item that might require production. Approval with the obvious caveat of test the in question videopress task in production immediately and be ready to revert if there are issues.

mattwiebe:
> test the in question videopress task in production immediately and be ready to revert if there are issues.

We can also just push a quick update to the existing mu-plugin to continue running but only with the videopress logic hooked up. Probably easier than a full revert for an edge case.

PR #30473, opened 2023-05-05
Fix notices background

n3f:
Fixes the notices background:

image

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/settings.scss
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/settings.scss
@@ -18,6 +18,10 @@
 .edit-post-post-visibility {
 	&__notice {
 		margin: 0;
+		// General notice
+		&.is-info {
+			background: #e9eff5; // Found using color picker on figma.
+		}
 	}
 
 	.components-notice__content {
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/settings.scss
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/settings.scss
@@ -20,7 +20,7 @@
 		margin: 0;
 		// General notice
 		&.is-info {
-			background: #e9eff5; // Found using color picker on figma.
+			background: var(--color-primary-0, #e9f0f5);
 		}
 	}
 


projects/plugins/jetpack/extensions/blocks/subscriptions/settings.scss
```diff
@@ -18,6 +18,10 @@
 .edit-post-post-visibility {
 	&__notice {
 		margin: 0px;
+		// General notice
+		&.is-info {
+			background: #e9eff5; // Found using color picker on figma.
```
n3f: I looked through the all the jetpack code to see if this background color existed somewhere else, or if the background was getting applied to notices and I didn't find any.  That's why I just put in the color straight from the figma file.

n3f: @crisbusquets, I'm not sure if it matters or not, but the border color looks a tad different from Figma.

crisbusquets: Hi @n3f !

Sorry I missed this comment.

I've used Blue / Blue 0 (`studio-blue-0`), but maybe de HEX code is slightly different. Do you have that studio color?

If you don't have it, can you please share with me the blue hex you have, so I can find what can we use for this particular case?

As a side note: the `notice` component on Gutenberg has some accessibility issues. I plan to open an issue in the Gutenberg repo so we can address it (and this would include add the blue background as well).

millerf: I am not sure hard-coding the background is a good idea: people that are using different theme could have some issues: https://a8c.slack.com/archives/C052XEUUBL4/p1683645627179279?thread_ts=1683639652.285329&cid=C052XEUUBL4

millerf: I discussed with Brent to use the SCSS lighten with:

```
lighten(var(--wp-components-color-accent,var(--wp-admin-theme-color,#007CBA)), 0.5) //adapt to fit Figma
```

n3f: After talking with @millerf I found the right var (`--color-primary-0`), this is actually `studio-blue-0`, but it can be/is changed by the theme.

millerf (approved):
WFM. Please merge..


Screenshot 2023-05-09 at 18 52 24

millerf:
@n3f It looks like this was the wrong var to use: https://github.com/Automattic/jetpack/pull/30446#issuecomment-1543510120.
I'll try and do something in https://github.com/Automattic/jetpack/pull/30446

PR #30475, opened 2023-05-05
Forms: denormalize comment column

CGastrell:
This PR is the second attempt to make export process to be a more faithful representation of the form's response.

## Proposed changes:
The export process now checks if the feedback post has new response format and uses it instead of relying on the post's meta for fields label/value pairs. 

This way, new responses won't have a "normalized" main _comment_ or _feedback_, but instead will export any `textarea` field as a column named after the input's label.

### Other information:

## Jetpack product discussion
p1683220500744339-slack-C01CSBEN0QZ

No

## Testing instructions:
Apply and try exporting single a multi (add as much as you can) responses, both to CSV and GSheets. Compare to same exports checking out `trunk`.

### DIFF
--- a/projects/packages/forms/changelog/forms-denormalize-comment-column
+++ b/projects/packages/forms/changelog/forms-denormalize-comment-column
@@ -1,0 +1,4 @@
+Significance: minor
+Type: changed
+
+Do not normalize feedback posts main comment when possible, allowing fexports to not guess which is the Comment and simply adding a column with the input's label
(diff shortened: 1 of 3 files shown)


digitalwaveride (dismissed):
Thank you so much Christian for fixing this!!! 🙏🙏🙏

The issue with the hardcoded "comment"/"feedback" field is gone.
Order of columns matches the order of the form fields.
Data export works as expected!
More context here: p1683220500744339-slack-C01CSBEN0QZ

🚀

digitalwaveride (approved):
Thank you, Christian! The "decoding" issue on export is gone! 👍

PR #30479, opened 2023-05-05
Social: Skip posts based on the new post meta key _wpas_skip_publicize_

spsiddarthan:
Fixes 1204432374898749-as-1204468895818601

## Proposed changes:

* Introduced new post meta key _wpas_skip_publicize_ which will be appended with the connection_id to make sure a post gets skipped. 
* Tweaked the code in `projects/packages/publicize/src/class-connections-post-field.php` to use the above key. 
*  The `projects/plugins/jetpack/_inc/lib/core-api/wpcom-fields/post-fields-publicize-connections.php` changes are needed to get the skip meta changes to work correctly on WPCOM. 

### Other information:

## Jetpack product discussion

1204432374898749-as-1204468895818601

No
## Testing instructions:

For Jetpack, Jetpack Social and WordPress.com:
* Set up 3 publicize connections and try publicizing to only two of them while disabling the other and making sure that only the selected networks get the post. 
* Repeat the above for the classic editor.


projects/packages/publicize/src/class-publicize-base.php
```diff
@@ -1264,33 +1278,28 @@ public function save_meta( $post_id, $post ) {
...
-				}
-
 				// This was a wp-admin request, so we need to check the state of checkboxes.
 				if ( $from_web ) {
+					$connection_id = $this->get_connection_id( $connection );
 					// Delete stray service-based post meta.
 					delete_post_meta( $post_id, $this->POST_SKIP . $service_name );
 
 					// We *unchecked* this stream from the admin page, or it's set to readonly, or it's a new addition.
-					if ( empty( $admin_page['submit'][ $unique_id ] ) ) {
```
spsiddarthan, line 1298: A test triggered this, but I need to see how to test this method manually

spsiddarthan: We need to test this using the classic editor from the looks of it

projects/packages/publicize/src/class-publicize-base.php
```diff
@@ -1264,33 +1278,28 @@ public function save_meta( $post_id, $post ) {
...
+					if ( empty( $admin_page['submit'][ $connection_id ] ) ) {
 						// Also make sure that the service-specific input isn't there.
 						// If the user connected to a new service 'in-page' then a hidden field with the service
 						// name is added, so we just assume they wanted to Publicize to that service.
 						if ( empty( $admin_page['submit'][ $service_name ] ) ) {
 							// Nothing seems to be checked, so we're going to mark this one to be skipped.
-							update_post_meta( $post_id, $this->POST_SKIP . $unique_id, 1 );
+							update_post_meta( $post_id, $this->POST_DONE . $connection_id, 1 );
```
danielpost: Should this be `$this->POST_SKIP_PUBLICIZE` instead of `$this->POST_DONE`?

projects/plugins/jetpack/_inc/lib/core-api/wpcom-fields/post-fields-publicize-connections.php
```diff
@@ -306,14 +306,14 @@ public function rest_insert( $post, $request, $is_new ) {
...
-	 * @param array $requested_connections Publicize conenctions to update.
+	 * @param array $requested_connections Publicize connections to update.
 	 *              Items are either `{ id: (string) }` or `{ service_name: (string) }`.
 	 * @param int   $post_id    Post ID.
 	 */
 	protected function get_meta_to_update( $requested_connections, $post_id = 0 ) {
 		global $publicize;
 
-		if ( ! $publicize ) {
+		if ( ! $publicize || ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE ) ) {
```
danielpost: Why does this need to be added now and not before?

danielpost:
I tested this on Jetpack and it works well. I also scheduled a post on `trunk` with some connections disabled, and then switched to this branch—the connections were still correct in the editor and the post only got Publicized to the right connections.

I'll go through the code a bit more and do some more testing but I think this is very close, if not already ready to merge.

projects/plugins/jetpack/tests/php/core-api/wpcom-fields/test-post-fields-publicize-connections.php
```diff
@@ -153,6 +153,7 @@ public static function setup_connections_jetpack() {
 						'id_number' => array(
 							'connection_data' => array(
 								'user_id'  => self::$user_id,
+								'id'       => '456',
 								'token_id' => 'test-unique-id456',
```
danielpost, line 157: Do we still need the `token_id` in all these tests?

danielpost (approved):
Nice work, I think this is ready to go!

projects/plugins/jetpack/tests/php/core-api/wpcom-fields/test-post-fields-publicize-connections.php
spsiddarthan: If I removed this method `get_connection_unique_id` starts breaking: https://github.com/search?q=repo%3AAutomattic%2Fjetpack+get_connection_unique_id&type=code

PR #30483, opened 2023-05-05
CRM: Removing HTML from all invoice date HTML functions, renaming functions

coder-karen:
## Proposed changes:

* This PR removes HTML from the `zeroBSCRM_html_invoiceDate`, `zeroBSCRM_html_transactionDate` and `zeroBSCRM_html_QuoteDate` functions (and renames those functions accordingly).
* The reason for the change was due to an observable issue with incorrect output for invoice dates showing on Company pages due to over-escaping (HTML was showing).

### Other information:

## Jetpack product discussion

https://github.com/Automattic/zero-bs-crm/issues/3076

No.

## Testing instructions:

* Using the [Jetpack Beta plugin](https://github.com/Automattic/jetpack-beta) on a test site with this branch applied or testing locally with the branch checked out, open up the 'View Contact' page for a contact (`/wp-admin/admin.php?page=zbs-add-edit&action=view&zbstype=contact&zbsid={contactid}`), and 'View Company page for a company (`/wp-admin/admin.php?page=zbs-add-edit&action=view&zbstype=company&zbsid={companyid}`).
* Click on the 'Quotes', 'Invoices' and 'Transactions' tabs for Contacts and the 'Invoices' and 'Transactions' tabs for Companies.
* View the Date column for each - the date of the relevant Quote/Invoice/Transaction should show, with no HTML visible. The date formats are all different but that is a [separate issue](https://github.com/Automattic/zero-bs-crm/issues/3084).
* To test the original issue with Invoice dates showing HTML for companies, check out trunk or open a test site with CRM 5.7.0 running, and on the View Company page open up the Invoices tab. The date field should show HTML.

The View Company page Invoice date styling before this PR is applied:

235151987-b5242bd8-4210-449d-b77a-63d8cbab9d4d

After:

Screenshot 2023-05-05 at 12 54 29

### DIFF
--- a/projects/plugins/crm/admin/company/view.page.php
+++ b/projects/plugins/crm/admin/company/view.page.php
@@ -591,57 +591,33 @@ item"><?php esc_html_e( 'Tasks', 'zero-bs-crm' ); ?></div><?php } ?>
 									foreach ( $company['invoices'] as $invoice ) {
 										// debugecho '<pre>'; print_r($invoice); echo '</pre><hr>';
 
-										$idRefStr = '';
-										// DAL3 change of field name
-										if ( $zbs->isDAL3() ) {
+										$id_ref_str = '';
 
-											// 3.0
-											if ( isset( $invoice['id'] ) ) {
-												$idRefStr = '#' . $invoice['id'];
+										if ( isset( $invoice['id'] ) ) {
+											$id_ref_str = '#' . $invoice['id'];
+										}
+										if ( isset( $invoice['id_override'] ) && ! empty( $invoice['id_override'] ) ) {
+											if ( ! empty( $id_ref_str ) ) {
+												$id_ref_str .= ' -';
 											}
-											if ( isset( $invoice['id_override'] ) && ! empty( $invoice['id_override'] ) ) {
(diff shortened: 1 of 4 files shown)


tbradsha (dismissed):
This works, and I'm approving for you to merge it if you'd like.

That said, here's another case where I think it'd be better to simply chop the functions.
We're passing the object to the function, then returning the object's date; it'd be just as easy to directly reference the object's date without a call.

From what I can see, the functions had a fallback for < DAL3, which is no longer relevant.

As an aside, there's a visual quirk that you might consider fixing at the same time. If an invoice doesn't have a reference number, it looks like this:
image

Contacts handle such cases like this:
image

coder-karen:
> That said, here's another case where I think it'd be better to simply chop the functions.

Yes, makes sense here. Chopped.

> As an aside, there's a visual quirk that you might consider fixing at the same time. If an invoice doesn't have a reference number, it looks like this:

Good catch, also chopped!

tbradsha (dismissed):
Looks great!

tbradsha (approved):
✨ Beautiful!

PR #29542, opened 2023-03-17
API endpoints: handle updated Requests library

jeherve:
## Proposed changes:

Let's start using the updated library now available in WordPres 6.2.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

Follow-up of #27981
See #27795

Dev-note: https://make.wordpress.org/core/2023/03/08/requests-library-upgraded-to-2-0-5-in-wordpress-6-2/

## Does this pull request change what data or activity we track or use?

* No

## Testing instructions:

Start with 2 sites, connected to WordPress.com, with the Publicize module active, and connected to a Twitter account:

- One running WordPress 6.1
- One running WordPress 6.2 Beta

We'll test Twitter features to see that things still work., while tailing logs for the 2 sites to watch for any notices.

* Go to Posts > Add New, and add some content
* Add a Twitter block and paste a Twitter URL for a thread, e.g. `https://twitter.com/doctorow/status/1636413428943278081`
* Click on Unroll
* Watch the tweet unroll
* Open the Jetpack sidebar
* Change Publicize options to publish a thread instead of a single tweet.
* Watch the separators added to the post content.
* Click on the "Social Previews" section at the bottom of the sidebar.
* You should see previews under some of the tweets, and you should not see any notices in your logs.
* Now add an image block, and select "Select Image > Google Photos".
* If you have not connected your Google account yet, you should be able to do so now.
* Once you've connected, hit the disconnect button in the top right corner of the modal.
* It should disconnect without generating any notices in your logs.

PR #30485, opened 2023-05-05
External Media: Define request method directly

obenland:
Fixes a bug on WP installations that are not fully updated to 6.2.0 yet and are missing `\WpOrg\Requests\Requests`.

Introduced in #29542.

## Proposed changes:

* Removes the use of the Requests library entirely, in favor of defining the request method directly.

### Other information:

- [X] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

p1683302169075199-slack-CDLH4C1UZ

No

## Testing instructions:

* Go to Posts > Add New, and add some content
* Add a Twitter block and paste a Twitter URL for a thread, e.g. https://twitter.com/kelseyhightower/status/1654098279116992513
* Click on Unroll
* Watch the tweet unroll
* Open the Jetpack sidebar
* Change Publicize options to publish a thread instead of a single tweet.
* Watch the separators added to the post content.
* Click on the "Social Previews" section at the bottom of the sidebar.
* You should see previews under some of the tweets, and you should not see any notices in your logs.
* Now add an image block, and select "Select Image > Google Photos".
* If you have not connected your Google account yet, you should be able to do so now.
* Once you've connected, hit the disconnect button in the top right corner of the modal.
* It should disconnect and not show a never ending spinner.

### DIFF
--- a/projects/plugins/jetpack/_inc/lib/class-jetpack-tweetstorm-helper.php
+++ b/projects/plugins/jetpack/_inc/lib/class-jetpack-tweetstorm-helper.php
@@ -1626,8 +1626,6 @@ class Jetpack_Tweetstorm_Helper {
 	 * @return array The Twitter card data.
 	 */
 	public static function generate_cards( $urls ) {
-		global $wp_version;
-
 		$validator = new Twitter_Validator();
 
 		$requests = array_map(
@@ -1649,7 +1647,7 @@ class Jetpack_Tweetstorm_Helper {
 		$requests = array_filter( $requests );
 
 		// Remove this check once WordPress 6.2 is the minimum supported version.
-		if ( version_compare( $wp_version, '6.2-alpha', '<' ) ) {
+		if ( ! class_exists( '\WpOrg\Requests\Hooks' ) ) {
 			$hooks = new Requests_Hooks();
 		} else {
 			$hooks = new \WpOrg\Requests\Hooks();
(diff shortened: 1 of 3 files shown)


jeherve (dismissed):
Should we update the other places we rely on `\WpOrg\Requests\Requests` as well?
https://github.com/Automattic/jetpack/blob/c2f963d23f9634ee687942f5c07df978da02bc88/projects/plugins/jetpack/_inc/lib/class-jetpack-tweetstorm-helper.php#L1666
https://github.com/Automattic/jetpack/blob/c2f963d23f9634ee687942f5c07df978da02bc88/projects/plugins/jetpack/_inc/lib/class-jetpack-tweetstorm-helper.php#L1750

projects/plugins/jetpack/changelog/fix-google-photos-disconnect
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: other
+
+Fixes a bug on WP installations that are not fully updated to 6.2.0 yet.
```
anomiex: That's either sites that installed a five-month-old 6.2 alpha version and then haven't updated since for some reason, or WordPress.com Simple that's apparently slowly updating commit-by-commit and hasn't gotten to [54997](https://core.trac.wordpress.org/changeset/54997) yet.

Might be clearer to just say "WordPress.com Simple" since I doubt there are any sites people care about in the first category.

projects/plugins/jetpack/_inc/lib/class-jetpack-tweetstorm-helper.php
```diff
@@ -1649,7 +1647,7 @@ function ( $url ) use ( $validator ) {
 		$requests = array_filter( $requests );
 
 		// Remove this check once WordPress 6.2 is the minimum supported version.
-		if ( version_compare( $wp_version, '6.2-alpha', '<' ) ) {
+		if ( ! class_exists( '\WpOrg\Requests\Hooks' ) ) {
```
anomiex, line 1650: Nitpick, feel free to ignore
```suggestion
		if ( ! class_exists( \WpOrg\Requests\Hooks::class ) ) {
```
Same for the other instances below.

anomiex (dismissed):
Looks sensible. Haven't tested.

Left some inline comments, but nothing I would block a merge on.

anomiex (approved):

PR #30492, opened 2023-05-07
Social: Use publicize connection_id to identify the connection in the editor sidebar

spsiddarthan:
For multiple connections belonging to the same token to work correctly, the connections shown in the editor sidebar need to be identified by the connection id and not the token_id.  That being said, connection_id is not available in the editor sidebar and it needs to be added in the result of the connection test results. 

## Proposed changes:

* In the case of Jetpack, I tweaked  the `class-connections-post-field` and the publicize connection test results 
* For simple sites, you need to apply the WPCOM generated by this PR to your sandbox

### Other information:

## Jetpack product discussion

1204432374898749-as-1204478638443353

## Testing instructions:

For Jetpack Block Editor
* Go to the editor sidebar after creating publicize connections, make sure you have three of them. They need not be from the same service. 
* Publish to two of the connections and make sure only the selected connections got publicized. 
* Inspect the publicized connection element on the UI (The connection-facebook-<id> bit on the imagebelow) and make sure the id used is that of publicize connection and that of the token_id. 
Screenshot 2023-05-08 at 13 37 36
* Boot up a site on trunk on JN and add connections to it. Switch to this branch. The site should continue to work without any errors or warnings. 

For Jetpack Classic Editor
* Same as above, but use the classic editor. 

For Jetpack Social, follow the same steps as above, but sandbox this patch D110694-code

To test WPCOM we need to perform an additional step:

 **Sandbox the changes generated by bin/jetpack-downloader test jetpack add/connectio-id-in-editor-sidebar** to test WordPress.com
For WordPress.com:
* Prepare a simple site and repeat the same steps from your testing above For Jetpack Block Editor
* Repeat the same steps from your testing above For Jetpack Classic Editor


gmjuhasz (dismissed):
This works well for me, for all of the mentioned scenarios 👍

projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -194,8 +195,8 @@ export default function PublicizeForm( {
 											false === is_healthy
 										}
 										enabled={ enabled && ! isPublicizeDisabledBySitePlan && false !== is_healthy }
-										key={ id }
-										id={ id }
+										key={ connection_id ? connection_id : id }
+										id={ connection_id ? connection_id : id }
```
pablinos, line 199: I've just realised that this could be:
```suggestion
										key={ connection_id ?? id }
										id={ connection_id ?? id }
```

projects/packages/publicize/tests/php/test-connections-post-field.php
```diff
@@ -254,10 +261,11 @@ public function get_connections() {
 				'facebook' => array(
 					'id_number' => array(
 						'connection_data' => array(
-							'user_id'  => self::$user_id,
-							'id'       => 'test-unique-id456',
-							'token_id' => 'test-unique-id456',
-							'meta'     => array(
+							'user_id'       => self::$user_id,
+							'id'            => '456',
```
pablinos, line 265: Do we change this as it's being used now? If so, should these be numbers rather than strings?

pablinos (approved):
Looks good! Just a couple of minor, non-blocking comments.

PR #30495, opened 2023-05-08
WIP script and style concatination

DavidCramer:
## Proposed changes:

* Adds script and style concatenation for enabled modules

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/projects/plugins/jetpack/functions.global.php
+++ b/projects/plugins/jetpack/functions.global.php
@@ -515,7 +515,7 @@ function jetpack_mastodon_get_instance_list() {
 function jetpack_enqueue_script( $handle, $src = '', $deps = array(), $ver = false, $in_footer = false ) {
 	static $concat;
 	if ( ! $concat ) {
-		$concat = new Jetpack_Concat();
+		$concat = Jetpack_Concat::get_instance();
 	}
 	wp_register_script( $handle, $src, $deps, $ver, $in_footer );
 	if ( is_admin() ) {
@@ -543,7 +543,7 @@ function jetpack_enqueue_script( $handle, $src = '', $deps = array(), $ver = fal
 function jetpack_enqueue_style( $handle, $src = '', $deps = array(), $ver = false, $media = 'all' ) {
 	static $concat;
 	if ( ! $concat ) {
-		$concat = new Jetpack_Concat();
+		$concat = Jetpack_Concat::get_instance();
 	}
 	wp_register_style( $handle, $src, $deps, $ver, $media );
 	if ( is_admin() ) {
--- a/projects/plugins/jetpack/functions.global.php
+++ b/projects/plugins/jetpack/functions.global.php
@@ -52,7 +52,7 @@ function jetpack_deprecated_function( $function, $replacement, $version ) { // p
 	) {
 		error_log( // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
 			sprintf(
-				/* Translators: 1. Function name. 2. Jetpack version number. */
+			/* Translators: 1. Function name. 2. Jetpack version number. */
 				__( 'The %1$s function will be removed from the Jetpack plugin in version %2$s.', 'jetpack' ),
 				$function,
 				$removed_version
@@ -61,6 +61,7 @@ function jetpack_deprecated_function( $function, $replacement, $version ) { // p
 
 	}
 }
+
 add_action( 'deprecated_function_run', 'jetpack_deprecated_function', 10, 3 );
 
 /**
@@ -92,7 +93,7 @@ function jetpack_deprecated_file( $file, $replacement, $version, $message ) { //
 	) {
 		error_log( // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
 			sprintf(
-				/* Translators: 1. File name. 2. Jetpack version number. */
+			/* Translators: 1. File name. 2. Jetpack version number. */
 				__( 'The %1$s file will be removed from the Jetpack plugin in version %2$s.', 'jetpack' ),
 				$file,
 				$removed_version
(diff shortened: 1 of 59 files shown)


projects/plugins/jetpack/functions.global.php
```diff
@@ -485,11 +489,66 @@ function jetpack_mastodon_get_instance_list() {
...
+ *                                    Default empty.
+ * @param string[]         $deps      Optional. An array of registered script handles this script depends on. Default empty array.
+ * @param string|bool|null $ver       Optional. String specifying script version number, if it has one, which is added to the URL
+ *                                    as a query string for cache busting purposes. If version is set to false, a version
+ *                                    number is automatically added equal to current installed WordPress version.
+ *                                    If set to null, no version is added.
+ * @param bool             $in_footer Optional. Whether to enqueue the script before `</body>` instead of in the `<head>`.
+ *                                    Default 'false'.
+ */
+function jetpack_enqueue_script( $handle, $src = '', $deps = array(), $ver = false, $in_footer = false ) {
```
jeherve, line 515: Since we'll be using that function in different Composer packages that are available and used on sites where the Jetpack plugin itself is not active, I would recommend moving this new function to the Assets package, alongside the new `Jetpack_Concat` class.

jeherve (commented):
Thanks for pushing that first branch! I left a note below on the general implementation of `jetpack_enqueue_script` and `Jetpack_Concat`. 

In general, we intend for our different Composer packages to be usable without the Jetpack plugin. In fact, some of those packages are already used by other WordPress plugins, developed by Automattic or by others in the WordPress community. 

As a result, I would suggest extracting the new functionality to a package, that will in turn be required by the Jetpack plugin. The existing Assets package seems like a good fit for what you're doing here.

projects/plugins/jetpack/functions.global.php
jeherve: I would still move this over to the Assets package, along with some tests for it. We do https://github.com/Automattic/jetpack/blob/a2e9a178e4c685d0b030f6dc2d24a60cd2097bf5/projects/packages/assets/src/class-assets.php#L328 , and I think this logic would be a good addition.

projects/plugins/jetpack/modules/concat.php
```diff
@@ -0,0 +1,25 @@
+<?php
+/**
+ * Module Name: Concatenation
```
jeherve, line 3: I don't think it's ideal to have this as a module, that needs to be activated for folks to benefit from it. I would rather have it automatically activated as soon as folks have connected their site to their WordPress.com account. It would also match the existing behavior of our concatenating style logic:
https://github.com/Automattic/jetpack/blob/5673953cf79c16533c6ab161dfff915c33a0b0bf/projects/plugins/jetpack/class.jetpack.php#L6217

Ideally, the code would be part of the Assets package, and then would be enabled in the `Jetpack` class. Extracting the logic into a package would allow other plugins to benefit from it; we could start using it in other Jetpack plugins right away.

projects/plugins/jetpack/modules/concat/class.jetpack-concat.php
```diff
@@ -0,0 +1,282 @@
+<?php //phpcs:ignore WordPress.Files.FileName.InvalidClassFileName
```
jeherve, line 1: Since those are new files, we can respect coding standards for them.

projects/plugins/jetpack/modules/concat/class.jetpack-scripts.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+class JetPack_Scripts extends WP_Scripts {
```
jeherve, line 3: We typically do not capitalize the P in the Jetpack brand.

projects/plugins/jetpack/modules/concat/class.jetpack-concat.php
```diff
@@ -0,0 +1,282 @@
...
+	 * @var string
+	 */
+	protected $cache_folder;
+
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		add_action( 'jetpack_pre_activate_module', array( $this, 'flush_cache' ) );
+		add_action( 'jetpack_pre_deactivate_module', array( $this, 'flush_cache' ) );
```
jeherve, line 45: While this will work well for resources that get enqueued by modules, it won't work for other CSS and JS items that may get enqueued, such as block resources for example. I wonder if there would be a different way to detect the change in the list of Jetpack resources that need to be loaded.

projects/plugins/jetpack/modules/concat/class.jetpack-concat.php
```diff
@@ -0,0 +1,282 @@
...
+		$wp_styles->done[]                = $handle;
+	}
+
+	/**
+	 * Handle a scripts dependancies.
+	 *
+	 * @param _WP_Dependency $object The script object.
+	 */
+	public function add_script_deps( $object ) {
+		$wp_scripts = wp_scripts();
```
jeherve, line 159: Is this so we do not enqueue core dependencies twice, once via Core if a plugin or the theme declares that dependency, and once via Jetpack?

projects/plugins/jetpack/modules/concat/class.jetpack-scripts.php
```diff
@@ -0,0 +1,56 @@
...
+
+	/**
+	 * Constructor.
+	 *
+	 * @since 2.6.0
```
jeherve, line 8: For all those `@since` values, you can use the `$$next-version$$` placeholder. It will get replaced by the proper version during the release.

jeherve:
Thanks for the contribution! I left a few comments above. Let me know what you think about them. Overall, I think that could be a really good improvement, but I think it would make sense to extract this from the Jetpack plugin itself, so it can be used elsewhere, by other plugins, whether from Automattic or from third-parties. 
Extracting the logic into a package may also make it easier when adding unit tests.

jeherve:
Closing this for now because of the lack of activity on this. We can always reopen in the future if needed.

PR #30515, opened 2023-05-08
Enable the new Jetpack Forms WP Admin feedback interface

ice9js:
## Proposed changes:

This PR will enable the new Jetpack Forms dashboard for all users. The old page still remains the default and users can opt-in to see the new interface by using the `view` switch in the upper right corner. The switch allows users to toggle between either view at their convenience.

### Other information:

## Jetpack product discussion

N/A

No.

## Testing instructions:

- You should now be able to access `admin.php?page=jetpack-forms` without the need to define any filters to enable it.
- Unless `jetpack_forms_admin_preferred_view` user option is set, classic view should still remain the default.

### DIFF
--- a/projects/packages/forms/changelog/update-feedback-default-view
+++ b/projects/packages/forms/changelog/update-feedback-default-view
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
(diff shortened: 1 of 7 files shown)


CGastrell (dismissed):
Worked fine for me! :shipit:

CGastrell (approved):
Let's try this! :shipit:

Issue #23695, opened 2022-03-30
General: compatibility with PHP 8.1


jeherve:
This primary issue should help us keep track of any issues related to PHP 8.1.

- [x] #23693 
- [x] #22852 
- [x] #23328
- [x] #21621

pyronaur:
@jeherve What's the status on this? Seems like all of the actions in the checklist have been completed, can this be closed?

jeherve:
Yup, this can be closed now!

PR #30517, opened 2023-05-08
PHP 8 Compatibility fixes

markbiek:
## Proposed changes:

These are minor changes as part of our PHP 8.1 compatibility update initiative pdqkMK-UN-p2.

### Other information:

## Jetpack product discussion

* pdqkMK-UN-p2
* #23695 

* No

## Testing instructions:

* These are minor enough that I think visual inspection should be sufficient but please let me know if you need more information.

### DIFF
--- a/projects/packages/stats/src/class-tracking-pixel.php
+++ b/projects/packages/stats/src/class-tracking-pixel.php
@@ -242,7 +242,7 @@ _stq.push([ "clickTrackerInit", "%2$s", "%3$s" ]);',
 		foreach ( $kvs as $k => $v ) {
 			$jskvs[] = "$k:'$v'";
 		}
-		return join( ',', $jskvs );
+		return implode( ',', $jskvs );
 	}
 
 	/**
--- a/projects/packages/my-jetpack/changelog/php8-fixes-may08-1311
+++ b/projects/packages/my-jetpack/changelog/php8-fixes-may08-1311
@@ -1,0 +1,4 @@
+Significance: patch
+Type: changed
+
+PHP8.1 compatibility updates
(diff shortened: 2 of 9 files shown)


projects/packages/stats/src/class-tracking-pixel.php
```diff
@@ -242,7 +242,7 @@ private static function stats_array_to_string( $kvs ) {
 		foreach ( $kvs as $k => $v ) {
 			$jskvs[] = "$k:'$v'";
 		}
-		return join( ',', $jskvs );
+		return implode( ',', $jskvs );
```
jeherve, line 245: I'm curious about this. Is `join` discouraged as of PHP 8? I'm not against that change, I'd much rather we use `implode` everywhere, but I don't understand how that's linked to PHP 8? Is `join` going to be deprecated maybe?

markbiek: I think `join` to `implode` is less a PHP 8-specific thing and more of a "standardize across the codebase" thing. I'm not sure who decided on these but there are a few similar substitutions that you can see in https://github.com/Automattic/wpcom-php-migration/blob/main/rector.php#L57

anomiex: Chances are if we really want to care about this sort of thing we should configure phpcs in the monorepo to make these replacements too.

anomiex (approved):

markbiek:
@anomiex Should I merge this or do you have a process where someone else will merge it?

anomiex:
You can go ahead.

projects/packages/stats/src/class-tracking-pixel.php
markbiek: Working on the phpcs change here: https://github.com/Automattic/jetpack/pull/30601

PR #30522, opened 2023-05-08
Launchpad: update task completion logic

mattwiebe:
The current task completion logic depends on the task being active, due to the limitations of sharing a callback between multiple tasks.

This allows marking a task complete, no matter the system's assumptions about whether it's active or not. Particularly useful in Multisite situations, as tasks are not set up properly outside of the `rest_api_switched_to_blog` call.

## Proposed changes:

* allow forcing a task to be marked complete

### Other information:

## Jetpack product discussion
n/a

## Testing instructions:
try it on your sandbox: `wpcom_mark_launchpad_task_complete( 'id', true )` for an inactive task

* Go to '..'
*


mattwiebe:
This: not good

PR #30523, opened 2023-05-08
PHP 8 Compatibility fixes

markbiek:
## Proposed changes:

These are minor changes as part of our PHP 8.1 compatibility update initiative pdqkMK-UN-p2.

### Other information:

## Jetpack product discussion
* pdqkMK-UN-p2
* #23695 
## Testing instructions:

* These are minor enough that I think visual inspection should be sufficient but please let me know if you need more information.

### DIFF
--- a/projects/packages/sync/src/class-queue.php
+++ b/projects/packages/sync/src/class-queue.php
@@ -377,13 +377,16 @@ class Queue {
 		);
 
 		$items = $wpdb->get_results( $query, OBJECT ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
-		foreach ( $items as $item ) {
-			// @codingStandardsIgnoreStart
-			$item->value = @unserialize( $item->value );
-			// @codingStandardsIgnoreEnd
+		$items_count = is_countable( $items ) ? count( $items ) : 0;
+
+		if ( $items_count > 0 ) {
+			foreach ( $items as $item ) {
+				// @codingStandardsIgnoreStart
+				$item->value = @unserialize( $item->value );
+				// @codingStandardsIgnoreEnd
+			}
 		}
 
-		$items_count = is_countable( $items ) ? count( $items ) : 0;
 		if ( $items_count === 0 ) {
 			$this->delete_checkout_id();
 			return false;
--- a/projects/packages/sync/src/class-queue.php
+++ b/projects/packages/sync/src/class-queue.php
@@ -114,7 +114,7 @@ class Queue {
 			}
 		}
 
-		$rows_added = $wpdb->query( $query . join( ',', $rows ) ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+		$rows_added = $wpdb->query( $query . implode( ',', $rows ) ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared, WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
 
 		if ( count( $items ) !== $rows_added ) {
 			return new WP_Error( 'row_count_mismatch', "The number of rows inserted didn't match the size of the input array" );
@@ -261,7 +261,7 @@ class Queue {
 
 		$items = $this->fetch_items( $buffer_size );
 
-		if ( count( $items ) === 0 ) {
+		if ( ! is_countable( $items ) || count( $items ) === 0 ) {
 			return false;
 		}
 
@@ -342,7 +342,7 @@ class Queue {
 			OBJECT
 		);
 
-		if ( count( $items_with_size ) === 0 ) {
+		if ( ! is_countable( $items_with_size ) || count( $items_with_size ) === 0 ) {
 			return false;
 		}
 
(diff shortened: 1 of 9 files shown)


markbiek:
Note to self: Need to stop using `is_countable` (or add a shim) since we need to support PHP 5.6.

projects/packages/sync/src/class-rest-endpoints.php
```diff
@@ -353,7 +353,7 @@ public static function full_sync_start( $request ) {
 				$modules['users'] = 'initial';
 			} elseif ( is_array( $request->get_param( $module_name ) ) ) {
 				$ids = $request->get_param( $module_name );
-				if ( count( $ids ) > 0 ) {
+				if ( is_countable( $ids ) && count( $ids ) > 0 ) {
```
anomiex: I note this one is redundant. We already checked `is_array()` just above on line 354.

Maybe this could use `!== array()` instead to make the tool happy.

projects/packages/sync/src/class-queue.php
```diff
@@ -375,9 +383,9 @@ public function checkout_with_memory_limit( $max_memory, $max_buffer_size = 500
 			// @codingStandardsIgnoreEnd
 		}
 
-		if ( count( $items ) === 0 ) {
+		$items_count = is_countable( $items ) ? count( $items ) : 0;
```
anomiex: Perhaps you should do something to stop line 380 from raising a warning if `$items` is not an array too?

projects/packages/sync/src/class-sender.php
```diff
@@ -565,6 +565,7 @@ private function fastcgi_finish_request() {
 	 * @return boolean|WP_Error True if this sync sending was successful, error object otherwise.
 	 */
 	public function do_sync_for_queue( $queue ) {
+		$wp_error = null;
```
anomiex: Kind of pointless considering the actual code here, but I guess the tool they're using isn't smart enough to realize it.

But I wonder if it'd be happy if you changed lines 645–647 to be like
```php
$wp_error = $had_wp_error ? array_pop( $processed_item_ids ) : null;
```
to avoid the pointless assignment up here.

projects/packages/sync/src/class-sender.php
```diff
@@ -645,7 +646,7 @@ public function do_sync_for_queue( $queue ) {
 					$wp_error = array_pop( $processed_item_ids );
 				}
 				// Also checkin any items that were skipped.
-				if ( count( $skipped_items_ids ) > 0 ) {
+				if ( is_countable( $skipped_items_ids ) && count( $skipped_items_ids ) > 0 ) {
```
anomiex: Might be more straightforward to do `!== array()` instead to make the thing happy. With the existing code here it'll always be an array.

projects/packages/sync/src/class-sender.php
anomiex: Kind of pointless considering the actual code here, but I guess the tool they're using isn't smart enough to realize it.

But I wonder if it'd be happy if you changed lines 645–647 to be like
```php
$wp_error = $had_wp_error ? array_pop( $processed_item_ids ) : null;
```
to avoid the pointless assignment up here when only one code path uses the variable.

projects/packages/sync/src/modules/class-callables.php
```diff
@@ -389,7 +389,7 @@ public function set_plugin_action_links() {
 				$action_links = array();
 			}
 			$formatted_action_links = null;
-			if ( ! empty( $action_links ) && count( $action_links ) > 0 ) {
+			if ( ! empty( $action_links ) && array() !== $action_links ) {
```
anomiex: `! empty( $action_links )` is already false for the empty array, so both the `count()` check before and the `array() !==` check now are redundant.

For that matter the `! empty()` itself is a bit redundant too, because `isset( $action_links )` will always be true. So this could reduce to just `if ( $action_links ) {`.

anomiex (commented):
Nothing really wrong here other than needing a run of `tools/fixup-project-versions.sh`. But I left some suggestions for improvement inline.

markbiek:
@anomiex Thanks for the feedback. I'll make those adjustments.

projects/packages/sync/src/class-rest-endpoints.php
```diff
@@ -353,7 +353,7 @@ public static function full_sync_start( $request ) {
 				$modules['users'] = 'initial';
 			} elseif ( is_array( $request->get_param( $module_name ) ) ) {
 				$ids = $request->get_param( $module_name );
-				if ( is_countable( $ids ) && count( $ids ) > 0 ) {
+				if ( array() !== $ids && count( $ids ) > 0 ) {
```
anomiex:
```suggestion
				if ( array() !== $ids ) {
```

projects/packages/sync/src/class-sender.php
```diff
@@ -642,11 +641,9 @@ public function do_sync_for_queue( $queue ) {
 			} else {
 				// Detect if the last item ID was an error.
 				$had_wp_error = is_wp_error( end( $processed_item_ids ) );
-				if ( $had_wp_error ) {
-					$wp_error = array_pop( $processed_item_ids );
-				}
+				$wp_error     = $had_wp_error ? array_pop( $processed_item_ids ) : null;
 				// Also checkin any items that were skipped.
-				if ( is_countable( $skipped_items_ids ) && count( $skipped_items_ids ) > 0 ) {
+				if ( array() !== $skipped_items_ids && count( $skipped_items_ids ) > 0 ) {
```
anomiex:
```suggestion
				if ( array() !== $skipped_items_ids ) {
```

nelsonec87 (approved):
LGTM

PR #30524, opened 2023-05-08
Check for `allowedBlocks` === true for payments intro

### LINKED ISSUE #28523: [Earn] Fix the Payments block selector
Originally spotted by @n3f:

> The payments block appears that lets you choose:

> But only the middle button works... Anyone else seeing something similar. (It's the same for me on trunk).

I was trying to point out when did this problem occur, and even after rolling back to mid-december, the problem is still there. It seems unrelated to our paid-newsletter work, but nevertheless - seems like a problem

## Reproduction path

This will be only possible on WPCOM testing flow as the premium content block is only available on wpcom:

1. Sync to wpcom
2. insert `/payment` on a site with enabled payment features
3. Observe that donation and premium content blocks are not able to be selected.

n3f:
Fixes #28523 (again).

The previous fix would not take the case where `settings.allowedBlockTypes` could be `true`.

## Proposed changes:

* One line change to check when all blocks are allowed.

No

## Testing instructions:

* Add a `Payments Button` block
* Selecting a block should work
Or see #28523


millerf (approved):
LGTM 🚢

projects/plugins/jetpack/changelog/earn-fix-payments-intro-blocks-true
```diff
@@ -0,0 +1,5 @@
+Significance: patch
+Type: bugfix
+Comment: one line change -- should be obvious
```
kraftbj, line 3: Approving, but please do put something here that's useful. The changelogs are used outside of the context of the PR itself so, in and of itself, this changelog is not obvious.

kraftbj (approved):

samiff:
cherry-picked to `weekly/branch-2023-05-08` in 65eb3ffcbdfcd49bf30499fe63090c2ea1706e5c

projects/plugins/jetpack/changelog/earn-fix-payments-intro-blocks-true
n3f: Oops, sorry!

PR #30530, opened 2023-05-09
Donations block: prevent crash on currency change

### LINKED ISSUE #30299: Payments block: cannot change currency
### Impacted plugin

Jetpack

### Quick summary

When using the payments block, changing currency breaks the block.

### Steps to reproduce

1. On Simple site (didn't check Atomic/Jetpack sites)
2. In the page editor (could happen also in other editors)
3. Add payments block
4. Pick donations
5. In the block toolbar, change currency to any other from the default USD
    Screenshot 2023-04-26 at 11 40 49
6. Observe crash and error in console

Screenshot 2023-04-26 at 11 40 39

https://user-images.githubusercontent.com/87168/234521391-a9098b08-c2d2-490a-90f1-60a099d0891f.mov

image

Not sure if this warning is related, but its in the console:

image

Should be able to change currency.

In fact, since I'm in Lithuania, I expected the default pre-selected currency to be Euros.They should

### What actually happened

Block crashed.

### Impact

All

### Available workarounds?

Simple

### Logs or notes

- Chrome browser

n3f:
Fixes #30299 (donation block currency changes crashed the block)

## Proposed changes:

* Adds `removeEventListener` to Amount blur.
* Updates the default amounts when the change occurs.

### Other information:

## Testing instructions:

* On sandbox run `bin/jetpack-downloader test jetpack earn/fix/change-donation-block-currency`
* Sandbox the site you want to check on and load the donations block.  Change the currency.  The block should not crash.
* There exist some other issues with the amount component, but these will be addressed in other patches.


n3f:
@millerf, I think that I found the root cause (there was a useEffect that was adding an event handler but never removing it -- I think this handler was firing (it was on blur). But it seems that it "works" now. With it working though I can see an issue with the isInvalid check that will need some attention.  I'll file that issue next.

projects/plugins/jetpack/extensions/shared/currencies.js
```diff
@@ -60,6 +60,21 @@ export function minimumTransactionAmountForCurrency( currency_code ) {
 	return SUPPORTED_CURRENCIES[ currency_code ];
 }
 
+/**
+ * Returns the default amounts for the given currency.
+ *
+ * @param {string} currency_code - three character currency code to get default amounts for
+ * @returns {number[]} Default amounts for the given currency_code
+ */
+export function getDefaultAmountsForCurrency( currency_code ) {
```
n3f: Should we rename this to `getDefaultDonationAmountsForCurrency()`?

projects/plugins/jetpack/extensions/blocks/donations/controls.js
```diff
@@ -75,7 +76,14 @@ const Controls = props => {
 											<MenuItem
 												isSelected={ ccy === currency }
 												onClick={ () => {
-													setAttributes( { currency: ccy } );
+													const defaultAmounts = getDefaultAmountsForCurrency( ccy );
```
millerf: We should move that to its own callback for clarity

projects/plugins/jetpack/extensions/blocks/donations/tabs.js
```diff
@@ -39,7 +39,7 @@ const Tabs = props => {
 				annualDonation: { ...annualDonation, planId: products[ '1 year' ] },
 			} ),
 		} );
-	}, [ oneTimeDonation, monthlyDonation, annualDonation, setAttributes, products ] );
+	}, [ oneTimeDonation, monthlyDonation, annualDonation, products, setAttributes ] );
```
millerf: Do we have to include setAttributes here?! Also this change is quite uneneeded :)

projects/plugins/jetpack/extensions/shared/currencies.js
millerf: Go ahead 👍

projects/plugins/jetpack/extensions/blocks/donations/amount.js
```diff
@@ -54,17 +55,21 @@ const Amount = ( {
...
+				setAmount( formatCurrency( defaultValue, currency, { symbol: '' } ) );
+			}
+		};
+
+		_ref.addEventListener( 'blur', onBlur );
 
-		setAmount( formatCurrency( defaultValue, currency, { symbol: '' } ) );
-	}, [ currency, defaultValue, editedValue, isFocused, setAmount ] );
+		return () => {
+			_ref.removeEventListener( 'blur', onBlur );
```
millerf, line 69: Nice catch

millerf: I wonder if  `_ref` is properly referenced in the clean up 🤔

projects/plugins/jetpack/extensions/blocks/donations/tabs.js
n3f: Ha! I didn't notice that.  But yeah, your PR removed it and the linter wanted it back, so I re-added it.  I'll get this fixed.

projects/plugins/jetpack/extensions/blocks/donations/amount.js
n3f: This was actually another catch from the linter.  Originally I just used the `.current` reference, but the linter correctly pointed out that it could be removed from the scope, but creating  a local const preserves it in this closure.

projects/plugins/jetpack/extensions/blocks/donations/tabs.js
millerf: Yeah, this linter rule kind of unsettles me...

n3f: Same, I'm not a big fan of it.

millerf:
@n3f Can you remove the draft status when you are done?

projects/plugins/jetpack/extensions/blocks/donations/amount.js
```diff
@@ -16,6 +16,7 @@ const Amount = ( {
 	const [ editedValue, setEditedValue ] = useState(
 		formatCurrency( value, currency, { symbol: '' } )
 	);
+	// const [ previousCurrency, setPreviousCurrency ] = useState( currency );
```
millerf: Needs removing

millerf (approved):
Works for me 👍

samiff (approved):
Tested well on my WoA site.

PR #30557, opened 2023-05-09
CRM: Resolves 2891 - clean up logic surrounding Woo status mapping

tbradsha:
Resolves Automattic/zero-bs-crm#2891 - clean up logic surrounding Woo status mapping

## Proposed changes:
This PR cleans up and centralises the Woo order status mapping logic. There is some legacy code (e.g. settings keys) that made it a bit difficult, but we're now in a better place.

It introduces three new functions, which should be roughly understandable by name:
* `is_valid_obj_status()`
* `get_default_status_for_order_obj()`
* `translate_order_status_to_obj_status()`

It refines the "default" order -> object status mapping:
* contacts: use default set in settings (same as before)
* invoices: processing and completed are considered paid (instead of unpaid), draft is draft (instead of unpaid), and others are unpaid (as before)
* transactions: uses order statuses (same as before, but now maps "Draft" order transactions to "Draft" status)

In settings, one will no longer see "Default" if there is no setting saved, but instead will see an accurate representation of the status to which it will map. If the default status isn't a valid one, the dropdown will show a message to select a status.

When syncing from WooCommerce, it will use the following mapping precedence:

1. The status mapping selected in settings (if a valid status)
2. The default mapping
3. The order status (only really applies to transactions)

If the status mapping setting is disabled:
* new contacts use the default mapping status (none passed, but DAL uses default)
* existing contacts maintain their current status
* transactions and invoices use the default mapping status

Note that the changes to `includes/ZeroBSCRM.DAL3.Helpers.php` are irrelevant to this PR; I just chopped some commented code when I was planning to add a function to this file (but ultimately didn't).

### Other information:

## Testing instructions:

There's a lot of behind-the-scenes stuff, but in most cases things should work as before. Read through the changes listed in this PR's description, and make sure they work as expected.

Likewise, make sure that the actual sync process works as expected, both with new and existing objects. 

Before:
image

After:
image

### DIFF
--- a/projects/plugins/crm/modules/woo-sync/includes/class-woo-sync.php
+++ b/projects/plugins/crm/modules/woo-sync/includes/class-woo-sync.php
@@ -274,7 +274,11 @@ class Woo_Sync {
 		} elseif ( $obj_type_id === ZBS_TYPE_TRANSACTION ) {
 			// weird legacy mapping fix
 			if ( $order_status === 'on-hold' ) {
-				$status = __( 'Hold', 'zero-bs-crm' );
+				// transaction statuses aren't translated, as they're user-configurable
+				$status = 'Hold';
+			} elseif ( $order_status === 'checkout-draft' ) {
+				// for lack of a better status
+				$status = 'Hold';
 			} else {
 				// default transaction status is the same as the Woo order status
 				$status = ucfirst( $order_status );
--- a/projects/plugins/crm/modules/woo-sync/includes/class-woo-sync.php
+++ b/projects/plugins/crm/modules/woo-sync/includes/class-woo-sync.php
@@ -248,7 +248,7 @@ class Woo_Sync {
 	 *
 	 * @return str|bool Status string to use for object
 	 */
-	public function get_default_woo_order_status_mapping_for_obj_type( $obj_type_id, $order_status ) {
+	public function get_default_status_for_order_obj( $obj_type_id, $order_status ) {
 		global $zbs;
 
 		$status = false;
@@ -1606,7 +1606,7 @@ class Woo_Sync {
 		$settings = $this->get_settings();
 
 		// get default object status for given Woo order status
-		$default_status = $this->get_default_woo_order_status_mapping_for_obj_type( $obj_type_id, $order_status );
+		$default_status = $this->get_default_status_for_order_obj( $obj_type_id, $order_status );
 
 		// if status mapping is disabled, return default status
 		$is_status_mapping_enabled = ( isset( $settings['enable_woo_status_mapping'] ) ? ( (int) $settings['enable_woo_status_mapping'] === 1 ) : true );
--- a/projects/plugins/crm/modules/woo-sync/includes/class-woo-sync.php
+++ b/projects/plugins/crm/modules/woo-sync/includes/class-woo-sync.php
@@ -272,13 +272,13 @@ class Woo_Sync {
 				$status = __( 'Unpaid', 'zero-bs-crm' );
 			}
 		} elseif ( $obj_type_id === ZBS_TYPE_TRANSACTION ) {
-			// weird legacy mapping fix
+			// note that transaction statuses aren't translated, as they're user-configurable
 			if ( $order_status === 'on-hold' ) {
-				// transaction statuses aren't translated, as they're user-configurable
+				// weird legacy mapping fix
 				$status = 'Hold';
 			} elseif ( $order_status === 'checkout-draft' ) {
 				// for lack of a better status
-				$status = 'Hold';
+				$status = 'Draft';
 			} else {
 				// default transaction status is the same as the Woo order status
 				$status = ucfirst( $order_status );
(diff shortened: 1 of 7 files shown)


coder-karen (commented):
Nice work! From my testing, it tests well.

> transactions: uses order statuses (same as before, but now maps "Draft" order transactions to "Draft" status)

Having not used WooSync too much, when clicking on the order number of a synced Draft transaction it did seem unexpected to then see the transaction status listed as Succeeded:

Screenshot 2023-05-10 at 10 01 58

That was the behaviour before as well, however still seems quite confusing to me. I understand we don't have a 'Draft' option for Transactions though, hence the option to select a status instead of preselecting one. I wonder if it's worth adding some info somewhere (underneath the Draft Transaction Status selector?) to clarify this is the intended default expected behaviour.

**Contact default**

This is more an enhancement suggestion, and I'm totally happy to create a separate issue for this but wanted to share it here too since you've been cleaning up this section :D 

Referring to the default status mapping table: `wp-admin/admin.php?page=zerobscrm-plugin-settings&tab=woosync`:

This uses the status default as you mentioned, however without knowing the settings set-up it isn't clear this default can be set here: `wp-admin/admin.php?page=zerobscrm-plugin-settings&tab=fieldoptions`.  Would it be worth adding a link (`wp-admin/admin.php?page=zerobscrm-plugin-settings&tab=fieldoptions#zbs-default-status`)? Perhaps an info icon next to 'Contact status' with a popup including the link to set the default? Or a link underneath the Contact status heading itself.

tbradsha:
> Having not used WooSync too much, when clicking on the order number of a synced Draft transaction it did seem unexpected to then see the transaction status listed as Succeeded

Indeed, I'll be tackling that issue separately (Automattic/zero-bs-crm#2874) after this one is merged. :^)

I plan to add the "invalid" status to that transaction's dropdown so it'll always show the current status even if it's not one of the preconfigured ones.

> I wonder if it's worth adding some info somewhere (underneath the Draft Transaction Status selector?) to clarify this is the intended default expected behaviour.

I was torn between adding a new "Draft" status to the initial setup or just leaving it as-is. Adding this would only affect new users, and I suspect few people will ever even have a "draft" transaction, so I opted for not adding it. What's your vote?

> This uses the status default as you mentioned, however without knowing the settings set-up it isn't clear this default can be set here

I considered this as well, but decided that since the default only applies until they save the page, it's more likely to cause confusion to have that info. If they were on the status mapping page and decided they wanted something else, it doesn't make sense to send them to another page change the default; instead, they would simply select the new statuses from the dropdowns then save, right?

coder-karen:
> Indeed, I'll be tackling that issue separately

Ah missed that! Nice.

> I was torn between adding a new "Draft" status to the initial setup or just leaving it as-is. Adding this would only affect new users, and I suspect few people will ever even have a "draft" transaction, so I opted for not adding it. What's your vote?

I think (I may be wrong) that checkout-draft and hence a draft transaction may be more common, as it appears to be created when an order is placed but not completed. If stores allow baskets to be saved for awhile, then I think it would make sense for the status on our end to match.

> instead, they would simply select the new statuses from the dropdowns then save, right?

Yes, true. I'm all for helping aid in cross-navigation though :D If there is a subtle way to add a link without adding clutter (even making the title 'Contact status') clickable I'd be for it.

tbradsha:
> I think (I may be wrong) that checkout-draft and hence a draft transaction may be more common, as it appears to be created when an order is placed but not completed. If stores allow baskets to be saved for awhile, then I think it would make sense for the status on our end to match.

I've added "Draft" as a default Transaction status going forward (only new installs will see it, but I think that's okay).

As for a "Default" indicator on Contact statuses, feel free to create a new issue if you want, or we can wait until a user reports confusion. :^)

coder-karen (approved):
> I've added "Draft" as a default Transaction status going forward (only new installs will see it, but I think that's okay).

This looks great!

> As for a "Default" indicator on Contact statuses, feel free to create a new issue if you want, or we can wait until a user reports confusion. :^)

I'll follow up on that one :)

jinnypark:
Hi @tbradsha!

I checked the Woo Sync status without making any changes, and this was the default status mapping on Woo Settings page:

image

Is this as intended? This is different from the screenshot you shared in the testing instructions.

I'm using Bleeding Edge CRM: 
image

tbradsha:
> I checked the Woo Sync status without making any changes, and this was the default status mapping on Woo Settings page
> Is this as intended? This is different from the screenshot you shared in the testing instructions.

Until you save the WooSync settings, the default status for contacts will depend on what is set in CRM Settings → Field Options. For most people that would be "Lead" (not "Customer" as in my screenshot). Does that answer this question?

projects/plugins/crm/modules/woo-sync/includes/class-woo-sync.php
```diff
@@ -228,42 +228,90 @@ public function get_settings() {
...
+		$status = false;
+
+		if ( $obj_type_id === ZBS_TYPE_CONTACT ) {
+			// default contact status is configured in CRM settings
+			$status = $zbs->settings->get( 'defaultstatus' );
+		} elseif ( $obj_type_id === ZBS_TYPE_INVOICE ) {
+			// reasonable default paid mapping based on Woo status descriptions:
+			// https://woocommerce.com/document/managing-orders/#order-statuses
+			$paid_statuses = array(
+				'pending',
```
cleacos, line 263: It seems that this status is incorrect. Pending is not a paid status for WooCommerce.

> Pending payment — Order received, no payment initiated. Awaiting payment (unpaid).

cleacos:
I'm late testing this PR, but it seems it doesn't fix the issue at all. The order `pending` status implies that the transaction is unpaid. The right status to use instead is `completed`.

PR #30562, opened 2023-05-09
Forms fix salesforce variation discoverability

CGastrell:
This PR moves the salesforce form variation definition alongside the default variations.

## Proposed changes:
The Salesforce variation is not discoverable because of the way it was being included. With this change, the variation becomes discoverable from the inserter:
image

The changes also fix a class method by making it public as the action trigger can't execute on private methods.

### Other information:

No

## Testing instructions:
Checkout and install/build:
```
jetpack install packages/forms && jetpack build packages/forms
```

Create a post and use the block inserter to search for the Salesforce variation (salesforce, lead). Inserter should show the variation. Insert on post and fill the Organization ID setting. Test published post by filling the form and confirm the lead is created on the Salesforce instance.

### DIFF
--- a/projects/packages/forms/changelog/forms-fix-salesforce-variation-discoverability
+++ b/projects/packages/forms/changelog/forms-fix-salesforce-variation-discoverability
@@ -1,0 +1,4 @@
+Significance: minor
+Type: fixed
+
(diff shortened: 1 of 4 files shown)


jcheringer (approved):
Works for me! 👍

PR #30599, opened 2023-05-10
PHP 8 Compatibility fixes

markbiek:
## Proposed changes:

These are minor changes as part of our PHP 8.1 compatibility update initiative [pdqkMK-UN-p](https://href.li/?pdqkMK-UN-p2)

### Other information:

## Jetpack product discussion
* pdqkMK-UN-p2
* #23695 
## Testing instructions:

### DIFF
--- a/projects/packages/sync/src/modules/class-users.php
+++ b/projects/packages/sync/src/modules/class-users.php
@@ -737,8 +737,9 @@ class Users extends Module {
 
 		// phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
 		$user_ids = $wpdb->get_col( "SELECT user_id FROM $wpdb->usermeta WHERE meta_key = '{$wpdb->prefix}user_level' AND meta_value > 0 LIMIT " . ( self::MAX_INITIAL_SYNC_USERS + 1 ) );
+		$user_ids_count = is_countable( $user_ids ) ? count( $user_ids ) : 0;
 
-		if ( count( $user_ids ) <= self::MAX_INITIAL_SYNC_USERS ) {
+		if ( $user_ids_count <= self::MAX_INITIAL_SYNC_USERS ) {
 			return $user_ids;
 		} else {
 			return false;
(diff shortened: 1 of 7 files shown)


zinigor (approved):
Looks good to me!

projects/packages/sync/src/modules/class-users.php
```diff
@@ -737,8 +737,9 @@ public function get_initial_sync_user_config() {
 
 		// phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
 		$user_ids = $wpdb->get_col( "SELECT user_id FROM $wpdb->usermeta WHERE meta_key = '{$wpdb->prefix}user_level' AND meta_value > 0 LIMIT " . ( self::MAX_INITIAL_SYNC_USERS + 1 ) );
+		$user_ids_count = is_countable( $user_ids ) ? count( $user_ids ) : 0;
 
-		if ( count( $user_ids ) <= self::MAX_INITIAL_SYNC_USERS ) {
+		if ( $user_ids_count <= self::MAX_INITIAL_SYNC_USERS ) {
```
anomiex, line 742: Meh. `$wpdb->get_col()` always returns an array, so technically this change isn't really needed and is just because rector doesn't know better.

OTOH, if some non-array were able to be returned, the previous code was obviously buggy because it didn't try to handle that case, while the new code is less obviously buggy because it now handles that case but does so incorrectly.

projects/packages/sync/src/replicastore/class-table-checksum.php
```diff
@@ -456,7 +456,8 @@ protected function prepare_filter_values_as_sql( $filter_values = array(), $tabl
 			switch ( $filter['operator'] ) {
 				case 'IN':
 				case 'NOT IN':
-					$values_placeholders = implode( ',', array_fill( 0, count( $filter['values'] ), '%s' ) );
+					$filter_values_count = is_countable( $filter['values'] ) ? count( $filter['values'] ) : 0;
```
anomiex, line 459: This is probably less correct than the previous code. Really there should be some explicit error handling here for the case that `$filter['values']` is not an array or is an empty array.

Back in PHP 7 the old code would have unintentionally worked for non-null scalars, only producing invalid SQL when passed null. Now it'll produce invalid SQL for any non-countable.

PR #30601, opened 2023-05-10
PHPCS rule to prefer implode over join

markbiek:
## Proposed changes:

This adds a phpcs rule to enforce using `implode` instead of `join`.

### Other information:

## Testing instructions:

* Check out this branch.
* Make an update that uses the `join` function and try to commit it.
* Your code should automatically be updated to use `implode` instead of `join`.


anomiex:
Seems to work, as can be seen by all the failures in the linting job. 😀

Two suggestions:

* You linked elsewhere to [a list of other such replacements](https://github.com/Automattic/wpcom-php-migration/blob/main/rector.php#L57) that wpcom is wanting to apply. We should probably do all of those.
* This rule does not fix with phpcbf. Perhaps we should use [MediaWiki.Usage.ForbiddenFunctions](https://github.com/wikimedia/mediawiki-tools-codesniffer/blob/v40.0.0/MediaWiki/Sniffs/Usage/ForbiddenFunctionsSniff.php) instead, which does have a fixer? We already use other MediaWiki phpcs rules so it's already available and just needs activating.
  * It looks like that one does have more replacements built in though. Looking through the code, it looks like the way to disable the built-in fixes (once we update to 41.0.0, which I've just told Renovate to make a PR for) is to set the property to map the function to itself.

markbiek:
@anomiex Could you give me a little more context on this? I'm not sure I follow what you're describing.

> Looking through the code, it looks like the way to disable the built-in fixes (once we update to 41.0.0, which I've just told Renovate to make a PR for) is to set the property to map the function to itself.

I _think_ this is doing what you're asking for but I want to make sure I understand
```
<rule ref="MediaWiki.ForbiddenFunctions">
        <properties>
            <property name="forbidden_functions">
                join=implode
            </property>
        </properties>
    </rule>
```

Assuming that's what you're talking about, I can add the other functions that rector is checking for.

projects/packages/codesniffer/Jetpack/ruleset.xml
```diff
@@ -84,4 +84,32 @@
 		<exclude name="Universal.NamingConventions.NoReservedKeywordParameterNames" />
 	</rule>
 
+	<rule ref="MediaWiki.Usage.ForbiddenFunctions">
+        <properties>
+            <property name="forbidden_functions">
+                join=implode
+				chop=rtrim
```
nelsonec87: indentation?

nelsonec87 (dismissed):
LGTM!

nelsonec87 (dismissed):
LGTM

markbiek:
@anomiex I'm not sure what do do about that "PHP 8.2 WP latest" test failure. Do you have any advice?

anomiex:
The PHP 8.2 failure was because you had the configuration wrong. I fixed it in 7626f8f30014871b512ca9bf3fe98b1d84089b47.

anomiex:
Looks good to me now, but someone other than me should review all the fixes I made.

@markbiek If you're the one who does that review, you can't hit "approve" in GitHub because you're listed as the author of the PR. But if you comment to say my changes are good, then I can hit the button.

markbiek:
> Looks good to me now, but someone other than me should review all the fixes I made.
> 
> @markbiek If you're the one who does that review, you can't hit "approve" in GitHub because you're listed as the author of the PR. But if you comment to say my changes are good, then I can hit the button.

Everything looks good to me!

anomiex (approved):

anomiex:
For the (internal) record, more details have been posted at pdWQjU-j5-p2.

PR #8464, opened 2018-01-05
Move code from plugins install endpoint to new Plugins Library

ockham:
Some prep work for ~#8450~ #8482. De-duplicate all the things!

Move:
* the `Jetpack_Automatic_Install_Skin` helper class to `_inc/lib/class.jetpack-automatic-install-skin.php`
* some static helper methods from `Jetpack_JSON_API_Plugins_Install_Endpoint` to `Jetpack_Plugins` in `_inc/lib/plugins.php`.

To test: Make sure all affected endpoints still work. Note that `Jetpack_Automatic_Install_Skin` is used by a few other endpoints (grep to verify!):
* Plugin installation (from WP.org): `POST` to `/sites/%s/plugins/%s/install` (try `woocommerce`)
* Theme installation (from WP.org): `POST` to `/sites/%s/themes/%s/install` (try `twentyseventeen`) -- *see note at bottom*
* Plugin installation (via upload): `POST` to `/sites/%s/plugins/%s/new` -- use Calypso to test (`wordpress.com/plugins/upload/:site`)
* Theme installation (via upload): `POST` to `/sites/%s/themes/%s/new` -- use Calypso to test (`wordpress.com/themes/upload/:site`) 
  
Note:
* `/sites/%s/themes/%s/install` doesn't work at all, returns a 404. However, that behavior seems to be preexisting!

PR #7843, opened 2017-09-23
Add V1.2 Plugin Endpoints

enejb:
V1.2 of the api 
- returns no errors if the action state is set to what was expected already
- Updated the attributes, removes id, and next autoupdate timestamp
- adds reason for why autoupdate will not work

#### Changes proposed in this Pull Request:

* 

#### Testing instructions:

* See D7358-code

PR #30607, opened 2023-05-10
PHP 8 Compatibility fixes

markbiek:
## Proposed changes:

### Other information:

## Testing instructions:

### DIFF
--- a/projects/plugins/jetpack/json-endpoints/jetpack/class.jetpack-json-api-plugins-install-endpoint.php
+++ b/projects/plugins/jetpack/json-endpoints/jetpack/class.jetpack-json-api-plugins-install-endpoint.php
@@ -90,6 +90,10 @@ class Jetpack_JSON_API_Plugins_Install_Endpoint extends Jetpack_JSON_API_Plugins
 			}
 		}
 
+		if ( ! $result ) {
+			return new WP_Error( 'plugin_install_failed', __( 'Plugin install failed because the result was invalid.', 'jetpack' ) );
+		}
+
 		if ( is_wp_error( $result ) ) {
 			return $result;
 		}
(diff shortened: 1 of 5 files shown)


nelsonec87 (approved):
There are a few build error, but other than that it LGTM

projects/plugins/jetpack/json-endpoints/jetpack/class.jetpack-json-api-plugins-install-endpoint.php
```diff
@@ -90,6 +90,10 @@ protected function install() {
 			}
 		}
 
+		if ( ! $result ) {
+			return new WP_Error( 'plugin_install_failed', __( 'Plugin install failed because the result was invalid.', 'jetpack' ) );
```
anomiex, line 94: ... This whole function is a WTF.

* When the bulk-install functionality was first added (7496bab33dda9dc3d8f7f041293f0d7d0ae1c130) the error handling was already confusing. When `$this->bulk` was false, there was always exactly one plugin so it didn't break when they put some error handling after the loop. And when it was true, no errors were handled (but `$this->log` was updated).
* Then 405b554a343f24a012b117a692f6262459deea80 weirdly put what became line 102 outside the loop as well, when it would have made more sense inside. Even then it had no `! $this->bulk` guard, but no warnings because an empty `$this->plugins` is rejected during validation so the loop would always run at least once. It also oddly changed the ordering of the non-bulk error handling.
* #7843 fixed a little bit of the weirdness in the misplaced after-the-loop error handling, which may have confused the intent going forward.
* #8464 refactored a bunch of stuff, and (unintentionally?) made it so bulk mode could now return an error but only if the last plugin failed.
* Now we reach this change. Which doesn't make things worse, it just continues the confusion.

A better fix would probably have been to just delete line 97 (no longer needed for non-bulk mode, and it running in bulk mode seems to be a bug) and move line 102 inside the loop (it being after was always a bug AFAICT).

PR #30608, opened 2023-05-10
Fix/plugins modify endpoint

markbiek:
## Proposed changes:

The `Plugins_Modify` endpoint references `$args` for a check on [line 170](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/jetpack/json-endpoints/jetpack/class.jetpack-json-api-plugins-modify-endpoint.php#L170) but `$args` is never defined.

This adds a declaration for `$args` in the same fashion we do elsewhere in the endpoint.

### Other information:

## Testing instructions:

Visual inspection.

### DIFF
--- a/projects/plugins/jetpack/json-endpoints/jetpack/class.jetpack-json-api-plugins-modify-endpoint.php
+++ b/projects/plugins/jetpack/json-endpoints/jetpack/class.jetpack-json-api-plugins-modify-endpoint.php
@@ -153,6 +153,8 @@ class Jetpack_JSON_API_Plugins_Modify_Endpoint extends Jetpack_JSON_API_Plugins_
 	 * @return bool|WP_Error
 	 */
 	public function callback( $path = '', $blog_id = 0, $object = null ) {
+		$args = $this->input();
+
 		Jetpack_JSON_API_Endpoint::validate_input( $object );
 		switch ( $this->action ) {
 			case 'delete':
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/json-endpoints/jetpack/class.jetpack-json-api-plugins-modify-endpoint.php
```diff
@@ -153,6 +153,8 @@ class Jetpack_JSON_API_Plugins_Modify_Endpoint extends Jetpack_JSON_API_Plugins_
 	 * @return bool|WP_Error
 	 */
 	public function callback( $path = '', $blog_id = 0, $object = null ) {
+		$args = $this->input();
+
```
daledupreez: Nit: I think it would be easier to read if we move the initialisation to the same area of the code where we check the array values. It might also be worth verifying that `$args` is an array, too.
```php
		$args = $this->input();

		if ( is_array( $args ) && ( isset( $args['autoupdate'] ) || isset( $args['autoupdate_translations'] ) ) ) {
			$this->needed_capabilities = 'update_plugins';
		}
```

daledupreez (dismissed):
Other than the minor code rearrangement I suggested, this looks good to me!

projects/plugins/jetpack/json-endpoints/jetpack/class.jetpack-json-api-plugins-modify-endpoint.php
```diff
@@ -153,6 +153,7 @@ class Jetpack_JSON_API_Plugins_Modify_Endpoint extends Jetpack_JSON_API_Plugins_
 	 * @return bool|WP_Error
 	 */
 	public function callback( $path = '', $blog_id = 0, $object = null ) {
+
```
daledupreez, line 156: Nit: I don't know that we need this additional whitespace in this PR.

daledupreez (approved):
This still looks good to me. Thanks for shifting the code.

PR #30616, opened 2023-05-10
PHP 8 Compatibility fixes

markbiek:
## Proposed changes:

### Other information:

## Jetpack product discussion
* pdqkMK-UN-p2
* #23695 
## Testing instructions:

### DIFF
--- a/projects/plugins/jetpack/changelog/php8-fixes-0510-1552
+++ b/projects/plugins/jetpack/changelog/php8-fixes-0510-1552
@@ -1,0 +1,4 @@
+Significance: patch
+Type: compat
+
+PHP 8 Compatibility fixes
--- a/projects/plugins/jetpack/json-endpoints/class.wpcom-json-api-add-widget-endpoint.php
+++ b/projects/plugins/jetpack/json-endpoints/class.wpcom-json-api-add-widget-endpoint.php
@@ -138,8 +138,7 @@ class WPCOM_JSON_API_Add_Widgets_Endpoint extends WPCOM_JSON_API_Endpoint {
 
 		if ( empty( $args['sidebar'] ) ) {
 			$active_sidebars = Jetpack_Widgets::get_active_sidebars();
-			reset( $active_sidebars );
-			$args['sidebar'] = key( $active_sidebars );
+			$args['sidebar'] = array_key_first( $active_sidebars );
 		}
 
 		return Jetpack_Widgets::activate_widget( $args['id_base'], $args['sidebar'], $args['position'], $args['settings'] );


nelsonec87 (approved):
LGTM since `array_key_first` has a polyfill

jeherve:
> LGTM since `array_key_first` has a polyfill

For reference, it is defined here:
https://github.com/WordPress/WordPress/blob/fe27bb8d0b59dc9261bac4a637256b767c4f4a30/wp-includes/compat.php#L396

Unfortunately, the PHP Compatibility check for PHP_CodeSniffer isn't aware of that polyfill, so that's a blocker for this PR:
https://github.com/Automattic/jetpack/actions/runs/4940798725/jobs/8832803738?pr=30616

@anomiex How do you think we should proceed with cases like this one? Should we move `PHPCompatibility.FunctionUse.NewFunctions.array_key_firstFound` to a notice instead of an error in our phpcs config?

anomiex:
> @anomiex How do you think we should proceed with cases like this one? Should we move `PHPCompatibility.FunctionUse.NewFunctions.array_key_firstFound` to a notice instead of an error in our phpcs config?

I don't think there's a "notice" level in our phpcs config, anything printed will fail the CI.

I'd not want to globally disable the message, since some of the packages we have are intended for possible use outside of a WordPress context. We could disable it in `projects/plugins/jetpack/.phpcs.dir.xml` (and other plugins) safely, as plugins are not such code, and in packages that already depend on other WordPress functions. But for packages that aren't already dependent on being in a WordPress environment, I'd just continue to use code like that here. It works fine in PHP 8.2, nothing here is deprecated.

markbiek:
> We could disable it in projects/plugins/jetpack/.phpcs.dir.xml (and other plugins) safely

@anomiex Can you elaborate on this a bit more? I'm not clear on what we actually need to do here.

anomiex:
If we want to disable this warning for the Jetpack plugin, something like this could be added to [projects/plugins/jetpack/.phpcs.dir.xml](https://github.com/Automattic/jetpack/blob/php8/fixes-0510-1203/projects/plugins/jetpack/.phpcs.dir.xml).
```xml
<!-- WordPress Core provides a polyfill. -->
<rule ref="PHPCompatibility.FunctionUse.NewFunctions.array_key_firstFound">
    <severity>0</severity>
</rule>
```
Probably we'd also have to create a `projects/plugins/jetpack/.phpcs.dir.phpcompatibility.xml` file with the same thing, since this is a PHPCompatibility rule.

We could do the same for other plugins too. For that matter, we could probably create `projects/plugins/.phpcs.dir.xml` and `projects/plugins/.phpcs.dir.phpcompatibility.xml` to do it for all of them at once, assuming none declare compatibility with an old enough version of WP Core that the polyfill might not exist yet.

As for packages, we could do it for the ones that assume a WordPress environment but we wouldn't want to do it for things like `projects/packages/changelogger/` or `projects/packages/ignorefile/` or `projects/packages/phpcs-filter/` that are intended to run in a non-WordPress environment, which is also why we don't want to globally exclude the rule.

markbiek:
@anomiex Thanks for those details. I can see about opening a separate PR to handle that.

In the meantime, are you comfortable with me merging this PR?

anomiex:
You can't merge this PR as-is due to the failure in the PHP Compatibility check. You'd at least need to add a `// phpcs:ignore PHPCompatibility.FunctionUse.NewFunctions.array_key_firstFound -- WP Core provides a polyfill` comment.

Personally I'd just keep the existing code since it will work fine in PHP 8.1 and 8.2, none of the methods are even soft-deprecated. But if you really want to go to the extra work for this polyfill, I won't try to stop you.

markbiek:
I'll just close it. Thanks!

@nelsonec87 FYI. How about we plan on skipping `array_key_first` and `array_key_last` fixes?

PR #30527, opened 2023-05-08
Launchpad: checklists API optimization

mattwiebe:
### Proposed changes:

This PR makes additional improvements to our backend Launchpad checklist API. In particular, we update how we add hooks for active tasks, and how we mark tasks complete. For most of these changes, the behavior of Launchpad should be exactly the same, so we'll want to test to ensure no regressions. 

We are also adding two new two new tasks for newsletters. These are being added here rather than a separate PR https://github.com/Automattic/jetpack/pull/30512. Some of the changes in this PR were spurred by discussion of that PR. See [p1683571462928819-slack-C052XEUUBL4](p1683571462928819-slack-C052XEUUBL4).

## Does this pull request change what data or activity we track or use?

No.

## Testing instructions:

**Setup.** Update your sandbox and then run `bin/jetpack-downloader test jetpack-mu-wpcom-plugin update/launchpad-optimization` to build and sync this branch to your sandbox. Sandbox public api. 

**First, test for regressions**

1. Free flow. 
   - Create a new Free site at https://wordpress.com/setup/free/intro. Sandbox the url. Continue to Launchpad. 
   - Confirm Launchpad loads with correct checklist and statuses. 
   - Go through tasks and confirm they update as expected. Pay special attention to the Edit Design and Launch tasks. 
  
2. Link in Bio flow.
   - Create a new LIB site at https://wordpress.com/setup/link-in-bio. Sandbox the url. Continue to Launchpad. 
   - Confirm Launchpad loads with correct checklist and statuses. 
   - Go through tasks and confirm they update as expected. Pay special attention to the First Post, Edit Links and Launch tasks. 
 
3. Newsletter flow. 
   - Create a new Newsletter site at https://wordpress.com/setup/newsletter. Sandbox the url. Continue to Launchpad. 
   - Confirm Launchpad loads with correct checklist and statuses. 
   - Go through tasks and confirm they update as expected. Pay special attention to the Write First Post task.
  

**Second, test new newsletter tasks.**

3. Newsletter flow. 
   - Checkout this frontend PR https://github.com/Automattic/wp-calypso/pull/76602. Run yarn and yarn start if needed. 
   - Create a new Newsletter site at http://calypso.localhost:3000/setup/newsletter/intro.
   - On the Setup page, click the box for a Paid Newsletter. 
   - Sandbox the url of the new. Continue to Launchpad. 
   - Confirm Launchpad loads with correct checklist and statuses.
[...]

PR #30624, opened 2023-05-10
Launchpad: update disabling logic

mattwiebe:
The previous launchpad completion code baked in an unfortunate assumption: that _all tasks_ must be completed in order to deactivate Launchpad. This is incorrect. Our previous logic, [on the client](https://github.com/Automattic/wp-calypso/blob/bda8d93ffcbbb3a2e20e69cc25c858b6ab209ee7/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts#L469), was to assume that having completed the final task in a list was equivalent to completing Launchpad. 

So this replicates that, and also allows for explicitly marking a task as `isLaunchTask` to say "complete this and Launchpad is done." I recommend that we map this onto existing flows.

## Proposed changes:

* Allows launchpad disabling whenever the last task, or the `isLaunchTask`, is completed.

### Other information:

## Testing instructions:

With this patch applied to your dotcom sandbox, trying completing a flow without first completing all other steps. With this you should be able to. Without it, you shouldn't be able to.


projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -390,7 +390,7 @@ public function add_hooks_for_active_tasks( $task_list_id = null ) {
 				// We only need to know the built completion status if the task has an `add_listener_callback` property.
 				// Small optimization to not run `is_complete_callback` as often.
 				$task = $this->build_task( $task_definition );
-				if ( ! $task['completed'] && is_callable( $task_definition['add_listener_callback'] ) ) {
```
lezama, line 397: we don't want to check `is_callable`?

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -414,7 +414,17 @@ public function mark_task_complete( $task_id ) {
 		$statuses[ $key ] = true;
 		$result           = update_option( 'launchpad_checklist_tasks_statuses', $statuses );
 
-		$this->maybe_disable_launchpad();
+		if ( isset( $task['isLaunchTask'] ) && $task['isLaunchTask'] ) {
```
lezama, line 421:
```suggestion
		if ( ! empty( $task['isLaunchTask'] ) ) {
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
mattwiebe: It turned out that it was already being checked above it, before we even built the task. It was too late to stop the merge train so I decided to clean it up here. [our discussion](https://github.com/Automattic/jetpack/pull/30527#discussion_r1190399533)

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
mattwiebe: I preferred mine, to guard against the kind of guy (I'm that kind of guy) who would add `isLaunchTask=false` in a task.

mattwiebe: Oh right looking at this again I forgot empty also checked for false (yay PHP), I do like that better.

mattwiebe:
I finally circled back to test this and it works as described: completing the last task of the `isLaunchTask` will switch `launchpad_screen` to `off` as previously. Any sites that launch currently are relying on the [client-side check](https://github.com/Automattic/wp-calypso/blob/bda8d93ffcbbb3a2e20e69cc25c858b6ab209ee7/client/landing/stepper/declarative-flow/internals/steps-repository/launchpad/task-helper.ts#L469) which isn't great.

 @lezama or @edanzer can you help get this merged?

samiff:
> trying completing a flow without first completing all other steps. With this you should be able to. Without it, you shouldn't be able to.

As someone not all that familiar with Launchpad, could you explain a bit more on the testing aspect of this PR please.

It's not clear to me for example, if I've completed all launch tasks, should the Launchpad and "Launch your site" button still display, or should launch happen automatically when all tasks are completed?

PR #30634, opened 2023-05-11
Plugin config: update reference links to releases in changelogs

### LINKED ISSUE #30631: CHANGELOG.md diff links are wrong
### Impacted plugin

None / Other

### Quick summary

When you run `jetpack release plugins/<plugin> changelog -s --add-pr-num`, it generates compare url for the version like:
`compare/vx.y.z...vx.y.z`. Which is wrong because the tags are named like `x.y.z` and branches are named like `<plugin>/branch-x.y.z`. So, it creates a wrong reference.

### Steps to reproduce

Check logs for CRM: https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/crm/CHANGELOG.md#570---2023-04-19.

The link for all versions including 5.7.0 is wrong and produces an empty comparison.

It should generate URLs like `compare/x.y.z...x.y.z`. For the above example the link should have been pointing to: https://github.com/Automattic/jetpack-crm/compare/5.6.0...5.7.0

### What actually happened

For the above example the link is pointing to https://github.com/Automattic/jetpack-crm/compare/v5.6.0...v5.7.0.

### Impact

One

### Available workarounds?

_No response_

### Logs or notes

It seems like a problem in the changelogger package.

jeherve:
Fixes #30631

## Proposed changes:

Let's update changelogger link formats when a plugin is set to use a different format for its tags.

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

* Not much to test here, check if the format matches the version expected by the plugin (check the `autotagger` key).

### DIFF
--- a/projects/plugins/backup/changelog/fix-changelog-compare-references
+++ b/projects/plugins/backup/changelog/fix-changelog-compare-references
@@ -1,0 +1,4 @@
+Significance: patch
+Type: changed
+
+General: update link references to releases in changelog.
--- a/projects/plugins/backup/changelog/fix-changelog-compare-references#2
+++ b/projects/plugins/backup/changelog/fix-changelog-compare-references#2
@@ -1,0 +1,5 @@
+Significance: patch
+Type: changed
+Comment: Updated composer.lock.
+
+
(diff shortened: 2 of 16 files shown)


zinigor (approved):
Looks good to me, thanks for the fix!

PR #30638, opened 2023-05-11
CRM: Make sure segment counts are updated when tag changes occur

coder-karen:
Fixes: https://github.com/Automattic/zero-bs-crm/issues/2690

## Proposed changes:

* This PR adds two `compileSegment` calls within the `addUpdateTagObjLinks` function in `ZeroBSCRM.DAL3.php` - when adding and removing tags. These calls work on all segments (if segments include relevant tags), making sure counts are updated.
* The reason for this is that the counts on the 'Manage Segments' page - `wp-admin/admin.php?page=manage-segments` - was showing out of date counts straight after tag changes had occurred. 

### Other information:

## Jetpack product discussion

https://github.com/Automattic/zero-bs-crm/issues/2690

No.

## Testing instructions:

* Using the [Jetpack Beta plugin](https://github.com/Automattic/jetpack-beta) on a test site with this branch applied or testing locally with the branch checked out, create a few contacts as well as a tag.
* On the manage segments page - `wp-admin/admin.php?page=manage-segments` - create a new segment based on the contact fields 'Has Tag', set to the tag you've created. 
* Add that tag to a couple of contacts via the contact list view page: `/wp-admin/admin.php?page=manage-customers`.
* Refresh or re-visit the 'Manage Segments' page, and the number in the 'Contact Count' field should be correct. If there is no contact count, open the list view and add 'Contact Count' to the page.
* Remove some tags or add more, and the contact count should update accordingly.
* Without this PR applied, the count won't update straight away - you can deactivate or install and activate the Advanced Segments plugin and then try refreshing the page again, and that should trigger the update instead.


projects/plugins/crm/includes/ZeroBSCRM.DAL3.php
```diff
@@ -4273,6 +4274,9 @@ public function addUpdateTagObjLinks($args=array()){
             // mode
             if (gettype($mode) != 'string' || !in_array($mode, array('replace','append','remove'))) return false;
 
+			// Retrieving segment information in order to update them with tag changes later.
+			$segments = $zbs->DAL->segments->getSegments( $owner, 1000, 0, true ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
```
cleacos: We don't need to retrieve the segments here. For example, in the `switch` block below, we don't use the $segments list in some cases, so we are performing the SQL query that we are not going to use.

The method `compile_segments_from_tagIDs` will be called once, so why not retrieve the segments there? in that method `compile_segments_from_tagIDs` or maybe `compile_segments_on_tags_update`

cleacos (changes requested):
It doesn't work as expected:

- The bulk action for removing tag doesn't work for segments counter.
- Sometimes, the bulk action for adding tags doesn't work either. I cannot determine the pattern when it works and when not.

projects/plugins/crm/includes/ZeroBSCRM.DAL3.php
coder-karen: Oh, good point! I should have moved it over when creating the helper function. Moved now.

coder-karen:
Thank you for testing this @cleacos 

> It doesn't work as expected:

This appears to be due to the last parameter added to `in_array` (`true`), added as a result of a PHPCS suggestion, which I somehow didn't realize broke the functionality. Removed that and all works well again for me. Does it work for you now?

tbradsha:
> This appears to be due to the last parameter added to `in_array`

Yeah, gotta be careful with this; it's an unsafe change to make unless you're sure the types will always match (same with `===` and `!==` suggestions).

cleacos (approved):
Great! Now it works as expected. 

LGTM!

PR #30587, opened 2023-05-10
AuthorRecommendation: add basic support

heavyweight:
This PR builds upon the scaffolding of the blog.

The block now supports the attribute `recommendations`.
Based on this we render the recommendations for users via PHP.

The backend is not yet wired to this and will be done in other PR and diff. 

1721-gh-Automattic/dotcom-forge#issuecomment-1523158523

## Proposed changes:
Screenshot 2023-05-10 at 14 59 56

* Add attribute recommendations.
* Block rendering in PHP.
* Editor: list subscriptions, and allow subscriptions to be selected. When selected the block attribute recommendations updates accordingly.

## Does this pull request change what data or activity we track or use?
The PR adds the recommendations (sites followed) as attributes of the block

## Jetpack product discussion
1721-gh-Automattic/dotcom-forge
pdDR7T-LK-p2

## Testing instructions:
1. Since the backend is not yet connected, this can be tested using mock data. You can apply what was removed here: https://github.com/Automattic/jetpack/commit/a15be4c760f3cc375c9adb8e6a2d2e905fa20086
2. Run `jetpack watch plugins/jetpack`
3. Add the block in the editor
4. Select a few of the recommendations and save
5. Visit the saved page to see if it is rendering properly

PR #30639, opened 2023-05-11
AuthorRecommendation: add backend support

heavyweight:
Followup PR for https://github.com/Automattic/jetpack/pull/30587

This PR adds jetpack endpoint for retrieving the user's followed sites.
In case the request is handled on wp com env then we use the library to retrieve this, otherwise, we send out a request to get this data.

Since users automatically follow blogs that they manage, this PR adds the option `filter_user_blogs` to filter the results returned by the backend. 

This PR changes the editor file to use the endpoint and work with real data. Also provides some minimalistic error handling.

Fixes 1721-gh-Automattic/dotcom-forge#issuecomment-1523158523

## Proposed changes:
- Add jetpack endpoint
- Connect endpoint with editor
- Wire `remove_user_blogs`

### Other information:

## Jetpack product discussion
pdDR7T-LK-p2
1721-gh-Automattic/dotcom-forge

This block saves the subscribed sites that the user wants to include as block attributes. 
For each recommendation, its `blog_id`, `URL`, `name`, and `site_icon` is persisted.

## Testing instructions:

* Run `bin/jetpack-downloader test jetpack author-recommendation-backend` on your sandbox
* Add the author recommendations block to a page or to your site
* Ensure that your blog subscriptions are displayed properly
* Check which of the subscriptions you want to include/show
* Check out the rendered block on your site


projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-following.php
```diff
@@ -0,0 +1,97 @@
+<?php
+/**
+ * REST API endpoint for the Jetpack author recommendations block.
+ *
+ * @package automattic/jetpack
+ * @since 11.8
```
jeherve:
```suggestion
 * @since $$next-version$$
```

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-following.php
```diff
@@ -0,0 +1,97 @@
...
+	 */
+	public function register_routes() {
+		register_rest_route(
+			$this->namespace,
+			$this->rest_base . '/mine',
+			array(
+				array(
+					'methods'             => WP_REST_Server::READABLE,
+					'callback'            => array( $this, 'get_response' ),
+					'permission_callback' => 'is_user_logged_in',
```
jeherve: Should we be more restrictive here? Folks with less than an author probably wouldn't benefit from this, I think?

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-following.php
```diff
@@ -0,0 +1,97 @@
...
+				),
+				'args' => array(
```
jeherve: I believe the args should be inside the array above.

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-following.php
```diff
@@ -0,0 +1,97 @@
...
+	public function get_response( $request ) {
+		$remove_user_blogs = $request->get_param( 'remove_user_blogs' );
+
+		if ( defined( 'IS_WPCOM' ) && IS_WPCOM ) {
+			require_once WP_CONTENT_DIR . '/lib/wpcom-user-following.php';
+			return get_user_followed_blogs( get_current_user_id(), $remove_user_blogs );
+		}
+
+		$body = Client::wpcom_json_api_request_as_user(
+			'/me/following?remove_user_blogs=' . (int) $remove_user_blogs,
```
jeherve: Since the parameter is not required, maybe we should only add it if it is provided?

jeherve (changes requested):

heavyweight:
@jeherve thanks for the review!
I applied your feedback

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-following.php
```diff
@@ -0,0 +1,105 @@
...
+	 * @var string
+	 */
+	public $rest_base = 'following';
+
+	/**
+	 * WPCOM_REST_API_V2_Endpoint_AI constructor.
+	 */
+	public function __construct() {
+		$this->wpcom_is_wpcom_only_endpoint    = true;
+		$this->wpcom_is_site_specific_endpoint = false;
```
jeherve, line 35: This will stop us from using the endpoint on self-hosted sites; it's not going to work there.

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-following.php
```diff
@@ -0,0 +1,105 @@
...
+
+	/**
+	 * Endpoint base route.
+	 *
+	 * @var string
+	 */
+	public $rest_base = 'following';
+
+	/**
+	 * WPCOM_REST_API_V2_Endpoint_AI constructor.
```
jeherve: You'll probably want to update the comments here and elsewhere.

jeherve (changes requested):

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-following.php
```diff
@@ -0,0 +1,105 @@
...
+	/**
+	 * Gets the sites the user is following
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 */
+	public function get_response( $request ) {
+		$ignore_user_blogs = $request->get_param( 'ignore_user_blogs' );
+
+		if ( defined( 'IS_WPCOM' ) && IS_WPCOM ) {
+			require_once WP_CONTENT_DIR . '/lib/wpcom-user-following.php';
```
arcangelini:
```suggestion
			require_once WP_CONTENT_DIR . '/lib/wpcom-get-user-followed-blogs.php';
```

I think this should be `wpcom-get-user-followed-blogs.php`

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-following.php
arcangelini: Are you sure about this @jeherve ? I thought `wpcom_is_site_specific_endpoint` just appended the site details to the API. So instead of this being `/me/following` it would be `/sites/WPCOM_SITE/me/following`.

projects/plugins/jetpack/extensions/blocks/author-recommendation/style.scss
```diff
@@ -17,6 +17,12 @@
 
         img.site-icon {
             border-radius: 50%;
+			min-width: 36px;
+			max-height: 36px;
+		}
+
+		div {
+            min-width: 36px;
```
arcangelini:
```suggestion
            max-width: 36px;
            max-height: 36px;
        }
        
        div {
            max-width: 36px;
```

arcangelini: Fixes the spacing and sets the logos to an even square. I think this is better for spacing in a table, otherwise the logos can push the text out like this:

Markup 2023-05-18 at 15 54 31

Also may want to apply this styling to the editor because it's a jungle in there haha:

Markup 2023-05-18 at 15 39 45

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-following.php
jeherve: Ah, I may be missing something here. I'll let @fgiannar weigh in, she knows more about this than me :)

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-following.php
```diff
@@ -0,0 +1,106 @@
...
+	 * @var string
+	 */
+	public $rest_base = 'following';
+
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		$this->wpcom_is_wpcom_only_endpoint    = true;
+		$this->wpcom_is_site_specific_endpoint = true;
```
arcangelini: I think this needs to be false, or else it will require `/sites/SITE_ID` to be appended to the endpoint.

arcangelini (dismissed):
Ran into a 404 error with the endpoint but fixed it manually to be able to test it. Everything worked well besides the comments below 👍

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-following.php
heavyweight: yes I think so as well
reverted this change

projects/plugins/jetpack/extensions/blocks/author-recommendation/style.scss
heavyweight: Nice catch. As discussed in the squad channel the site icon can come in sizes bigger than expected.
To handle this I committed the appropriate CSS rules to limit the size on the front-end.

agrullon95:
@heavyweight - Great work, the changes are working on my end. I tested on both Simple and Atomic. 

I'm curious about this site selection behavior. I attached a short video of what I'm seeing and I'm wondering if this is expected.
To summarize, if the user selects a site (or multiple sites) , filters the list by using the search box, then selects a site from the filtered list, should the previously selected site be ignored?

https://github.com/Automattic/jetpack/assets/10482592/2d2981c1-397d-4ba0-8684-2c62882c05fd

heavyweight:
> @heavyweight - Great work, the changes are working on my end. I tested on both Simple and Atomic.
> 
> I'm curious about this site selection behavior. I attached a short video of what I'm seeing and I'm wondering if this is expected. To summarize, if the user selects a site (or multiple sites) , filters the list by using the search box, then selects a site from the filtered list, should the previously selected site be ignored?
> 
>  CleanShot.2023-05-19.at.09.06.35.mp4

Tnx for reporting this.
Pushed a fix for it.

agrullon95 (approved):
Tested and worked for simple and atomic sites.
I'm no longer seeing the selection behavior I described above.

jeherve (approved):
Thing works on my end now 👍

***********

I noticed that the loading item is outside the block borders:

Screenshot 2023-05-22 at 17 56 10

*********

It seems there are some issues with filtering, some items seem to be always included, and some seem to be included twice:

https://github.com/Automattic/jetpack/assets/426388/cd6b398d-b4be-4442-b35b-18c96897e933

*********

I think we could benefit from an explanation label with the setting to hide user blogs, to explain what it does.

******

I'll approve this, since I think this can be merged and we can iterate in follow-up PRs.

heavyweight:
> Thing works on my end now 👍
> 
> I noticed that the loading item is outside the block borders:
> 
> Screenshot 2023-05-22 at 17 56 10
> It seems there are some issues with filtering, some items seem to be always included, and some seem to be included twice:
> 
>  Screen.Recording.2023-05-22.at.17.58.37.mov 
> I think we could benefit from an explanation label with the setting to hide user blogs, to explain what it does.
> 
> I'll approve this, since I think this can be merged and we can iterate in follow-up PRs.

Interesting... I can't seem to reproduce this, but I expect to iterate on the design.
Will try to fix reproduce and fix it in a follow-up PR.

PR #30643, opened 2023-05-11
CRM: Resolves 2880 - clean up date pickers in CRM

tbradsha:
CRM: Resolves Automattic/zero-bs-crm#2880 - clean up date pickers in CRM

## Proposed changes:

The initial issue was to clean up some duplicate JS code used between Client Portal and Woo My Account. The code in question was directly related to the third-party daterangepicker library, so I went ahead and bypassed it with native functions instead. Doing so also affected custom fields of type `date` and quote date fields (due to tightly-wound code).

Note we currently store date fields, adjusted to the UTC timezone, stored in UTS. I adjusted the new date/time functions to support this, though ideally we'd store dates in `YYYY-MM-DD` format instead: pbhBOz-1s2-p2#comment-2188

### Other information:

## Testing instructions:

Everything should work as before, with native date pickers. There's less JS needed on the Client Portal and My Account.

1. Create a contact custom field of type `date`.
2. Ensure viewing/modifying the field works as before
    * Contact view page
    * Contact edit page
    * Client Portal as contact
    * Woo My Account as contact (note that you'll need to enable the field in WooSync settings: `/wp-admin/admin.php?page=zerobscrm-plugin-settings&tab=woosync`
3. Make sure quote dates save/display as before.


coder-karen (commented):
Everything works well in terms of date functionality.

There were some date inconsistencies testing with Jurassic Ninja and the beta tester plugin vs testing locally which I couldn't understand though.

In the Client Portal as the customer, with the PR applied the date format is showing as DD/MM/YYYY using JN. With bleeding edge (using the beta tester plugin) it's MM/DD/YYYY, whereas testing locally it's DD/MM/YYYY in both cases. 

On the edit Quotes page, with the PR applied the format is DD/MM/YYYY (and in trunk locally), whereas in trunk and bleeding edge it is MM/DD/YYYY.

The contact view page was MM/DD/YYYY both in bleeding edge and trunk locally, but on the edit page it was DD/MM/YYYY with the PR applied using JN (and locally).

I couldn't quite work out why it would be different in these two cases, despite clearing browser caches. 

None of that should matter really, as the results with the PR applied are what they should be now... almost always. On JN sometimes the date when viewing the contact is wrong - MM/DD/YYYY  - and I've not worked out what is causing that. Date format and time settings on the site seem to be the same. When this happens, I notice the 'Contact since' date when viewing the contact is also in the wrong format 🤔 

I'm not too sure whether this warrants any more investigation or whether it's something to keep an eye on, maybe JN specific, or just how I've been testing, but wanted to share what I was finding here.

tbradsha:
> There were some date inconsistencies testing with Jurassic Ninja and the beta tester plugin vs testing locally which I couldn't understand though.

The date format for native date pickers uses the browser locale (which often matches the OS locale). As such, I'd expect it should show `DD/MM/YYYY` on your end anywhere you see a native date picker.

> In the Client Portal as the customer, with the PR applied the date format is showing as DD/MM/YYYY using JN.

Good.

> With bleeding edge (using the beta tester plugin) it's MM/DD/YYYY

It sounds like bleeding edge isn't picking up the PR, perhaps? 

> whereas testing locally it's DD/MM/YYYY in both cases.

Not sure what "both cases" here refers to.

> On the edit Quotes page, with the PR applied the format is DD/MM/YYYY

Good.

> (and in trunk locally)

What date format does your local WP site have in WP Settings?

> whereas in trunk and bleeding edge it is MM/DD/YYYY.

I again suspect bleeding edge isn't picking up this PR.

> The contact view page was MM/DD/YYYY both in bleeding edge and trunk locally

This would respect WP settings.

> but on the edit page it was DD/MM/YYYY with the PR applied using JN (and locally).

This would respect the browser settings.

coder-karen:
>  It sounds like bleeding edge isn't picking up the PR, perhaps?

Bleeding edge using the Beta tester plugin should act like trunk when testing locally - any merged PR should be included (but not an open PR like this).

> Not sure what "both cases" here refers to.

Sorry, both cases refers to testing locally with the PR checked out, and testing in trunk.

> What date format does your local WP site have in WP Settings?

I actually thought the date format across all the JN sites and my local development site were all the same at first glance, but it turns out they weren't.  

Locally it was  15 May 2023: j F Y

On JN it's showing May 15, 2023: F j, Y

Both are the top date format options in settings. But when I change my site language on JN from English (US) to English (UK) to match my local site, then press save, the selected date format is now showing as custom, and the date format I use on the local site is showing as the top, unselected option. Changing it to match, everything appears as it should. So everything does look like it should now...

coder-karen (approved):
Works well now that I'm using the same date formats! Looking through the code, everything looks ok to me there too.

PR #30647, opened 2023-05-11
Add WAF tests for unsupported environments

dkmyta:
### Description
This PR introduces some fixes/improvements to code related to unsupported environments and adds tests to validate things are working as expected when certain platform constants are set.

## Proposed changes:

* Replaces the `IS_WPCOM` constant check with the `Host` class `is_wpcom_simple` method in `Waf_Runner::is_supported_environment`
* Adds an action for the `jetpack_get_available_standalone_modules` filter to ensure the callback to remove the `waf` module from the available modules list also applies to the standalone plugin
* Adds a `Waf_Initializer::remove_standalone_module_on_unsupported_environments` method to ensure that when an unsupported environment is detected an array with the `waf` module removed is returned when it is used as the callback for the above
   * Note that the available/active modules array provided by Jetpack is an associated array with module slugs as the keys and versions as the values. The array provided by standalone Protect is a simple array of module slugs. Therefore a separate callback method was required to handle entry removal in each situation.
* Tests were added for the various unsupported environments. 
   * The tests establish a base set of available modules that would be generated by the plugin that employs the WAF package. It then initializes the WAF, verifies that the package properly recognizes and handles the environment, and removes the WAF module from the list of available modules when applicable. The resulting available modules list is what is used for eventual module activation - if the module is not in the list when `Waf_Runner::enable` (or `Brute_Force_Protect::enable`) is called within the package or plugin, the module will not be moved to the active modules list, `Waf_Runner:: is_enabled` (or `Brute_Force_Protection::is_enabled`) methods will return `false`, and no package code reliant on these checks should be run, effectively disabling it.
   * Testing covers the following:
      * Supported environment 
      * WPcom environment
      * Atomic environment
      * VIP environment

### Other information:

## Jetpack product discussion

- 1201069996155224-as-1204565761279419
- p1683666778750069/1683665504.220709-slack-C04E0KTGW9Z

- No

## Testing instructions:

* Checkout this branch
* Run `jetpack test packages/waf php` and verify that all testing succeeds
* Install the Jetpack Beta Tester plugin on a Pressable or Atomic site (using this branch for Jetpack only first, Protect only next, and then Jetpack and Protect) and ensure that the WAF settings are inaccessible and functionality disabled and that Brute force protection settings continue to be accessible and functionality available
   * Ensure no regression in functionality
* Install the Jetpack Beta Tester plugin on a Jurassic Ninja site (using this branch for Jetpack only first, Protect only next, and then Jetpack and Protect) and ensure that the WAF settings are inaccessible and functionality available and that Brute force protection settings are accessible and functionality available
   * Ensure no regression in functionality

### DIFF
--- a/projects/packages/waf/changelog/add-waf-tests-for-unsupported-envs
+++ b/projects/packages/waf/changelog/add-waf-tests-for-unsupported-envs
@@ -1,0 +1,4 @@
+Significance: patch
+Type: added
+
+Integration tests for unsupported environments
--- a/projects/packages/waf/changelog/add-waf-tests-for-unsupported-envs
+++ b/projects/packages/waf/changelog/add-waf-tests-for-unsupported-envs
@@ -1,4 +1,4 @@
 Significance: patch
 Type: added
 
-Integration tests for unsupported environments
+Add integration tests for unsupported environments
--- a/projects/packages/waf/src/class-waf-initializer.php
+++ b/projects/packages/waf/src/class-waf-initializer.php
@@ -30,6 +30,7 @@ class Waf_Initializer {
 	public static function init() {
 		// Do not run in unsupported environments
 		add_action( 'jetpack_get_available_modules', __CLASS__ . '::remove_module_on_unsupported_environments' );
+		add_action( 'jetpack_get_available_standalone_modules', __CLASS__ . '::remove_module_on_unsupported_environments' );
 
 		// Ensure backwards compatibility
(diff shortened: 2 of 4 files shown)


nateweller:
Haven't reviewed in detail yet, but this is an awesome contribution to the waf package, much needed tests 🤩

nateweller (approved):
Beautiful! LGTM :+1:

PR #29322, opened 2023-03-06
required-review: Don't fail on case mismatch for `@singleuser` pseudo-teams

anomiex:
## Proposed changes:

GitHub usernames are case-insensitive but case-preserving. To match that logic, we want `@singleuser` to match `SingleUser` or whatever the actual username is.

This is complicated by the fact that "case-insensitive" can vary depending on what exactly GitHub is doing on their end, what version of Unicode they're using, and so on. So let's just ask GitHub's API for the correct casing of the username and use that instead of trying to work out all those edge cases ourself.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

https://github.com/Automattic/jetpack/pull/29318#issuecomment-1456890325

## Does this pull request change what data or activity we track or use?

No

## Testing instructions:

* Try changing the case of an `@singleuser` entry in `.github/files/required-review.yaml` and using the `test.sh` script to see how it handles an appropriate PR.

PR #30653, opened 2023-05-11
Adds feature where if review requirement not met, then request

gfog-floqast:
Fixes #

## Proposed changes:

* Adds feature where if a review requirement is not met, then request the review. This is beneficial so that a developer does not need to look at the review requirement(s) to determine who to request reviews from in the case that review requirements are not met.

### Other information:

## Testing instructions:

* Create `.github/required-review.yaml` similar to the following:
```
- name: Terraform
  paths:
    - '**/*.tf'
  teams:
    - any-of:
      - all-of:
        - team1
        - team2
      - '@user1'

- name: CODEOWNERS
  paths:
    - 'CODEOWNERS'
  teams:
    - team1
    - '@user2'

- name: .github
  paths:
    - '.github/**'
  teams:
    - all-of:
      - team1
      - team2
```
* Make updates to each of the files, as needed
* Observe that on first pass the necessary teams/users are requested

Tested this by deploying fork to my local github env and using settings above. 
* Observed initial failure on all requirements and observed required reviews were requested:
Screenshot 2023-05-11 at 10 59 30 AM
* Received review from team1 and observed CODEOWNERS check passed:
Screenshot 2023-05-11 at 11 02 43 AM
* Received review from user1 and observed Terraform check passed:
Screenshot 2023-05-11 at 11 03 51 AM
* Received review from team2 and observed .github check passed:
Screenshot 2023-05-11 at 11 04 54 AM
* Additionally received review from user2, which had no affect on status.


projects/github-actions/required-review/CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.1.0] - 2023-05-11
+### Added
+- Added feature where if a review requirement is not met, then request the review.
+
+### Changed
+- Updates misspelling
```
jeherve: Good news, you do not need to fill this in. This changelog will be automatically updated when we next release a new version of this action, based off the different changelog entries in the `changelog/` directory (where you've added your own changelog entry already, thank you).

```suggestion
```

projects/github-actions/required-review/src/request-review.js
```diff
@@ -0,0 +1,66 @@
...
+ * @param {string} user - @ followed by a GitHub user name.
+ */
+async function getUsername( user ) {
+    const octokit = github.getOctokit( core.getInput( 'token', { required: true } ) );
+    try {
+        const res = await octokit.rest.users.getByUsername( { username: user.slice( 1 ) } );
+        return res.data.login
+    } catch ( error ) {
+        throw new WError(
+            // prettier-ignore
```
jeherve: I see you've ignored Prettier here and across the file in general. Could you remove the exception and run it, so the style remains consistent with the other files in the action?

projects/github-actions/required-review/src/request-review.js
```diff
@@ -0,0 +1,66 @@
...
+	const octokit = github.getOctokit( core.getInput( 'token', { required: true } ) );
+	const owner = github.context.payload.repository.owner.login;
+	const repo = github.context.payload.repository.name;
+	const pr = github.context.payload.pull_request.number;
+
+    if ( team.startsWith( '@' ) ) {
+		// Handle @singleuser virtual teams. Fetch the correct username case from GitHub
+		// to avoid having to worry about edge cases and Unicode versions and such.
+        try {
+            const login = await getUsername(team);
```
jeherve: I like that you extracted that here, that's clean. 👍

I wonder if we should make further use of that extraction in the existing logic in `projects/github-actions/required-review/src/team-members.js`?

jeherve (changes requested):
First of all, thank you for the contribution!

I like that addition, I can see how useful it can be. However, since it can generate additional noise for maintainers since it will generate notifications and / or emails when they get added as reviewers, I think this option should be opt-in. I would consequently recommend that this only be available when specifying it in the action configuration.

What do you think of that?

projects/github-actions/required-review/src/request-review.js
gfog-floqast: yup, I can pull this out. Is there a file or location that is preferred for this? ie a helper file, or I can call it `get-username.js` and reference it in both files.

gfog-floqast:
Thanks @jeherve, I'll look into making this feature opt-in

projects/github-actions/required-review/src/request-review.js
jeherve: > or I can call it get-username.js and reference it in both files

This seems like a good approach to me 👍

jeherve (dismissed):
I've pushed a commit to address some linting errors, hopefully that helps a bit.

I think we're still missing a check for the new `request-reviews` action parameter you added, before we try to request reviews.

gfog-floqast:
@jeherve Thanks for addressing those linting errors. Regarding the `// prettier-ignore` [comment](https://github.com/Automattic/jetpack/pull/30653#discussion_r1192526060), is there anything you would like me to do with it? I copied that from the [team-members file](https://github.com/Automattic/jetpack/pull/30653/files#diff-a2d2e2075dd774cef44c3db803766dcf96aeb8486b1580b61deb9ac28c144674L31) so not sure if this needs an update.

jeherve:
> Regarding the `// prettier-ignore` [comment](https://github.com/Automattic/jetpack/pull/30653#discussion_r1192526060), is there anything you would like me to do with it? I copied that from the [team-members file](https://github.com/Automattic/jetpack/pull/30653/files#diff-a2d2e2075dd774cef44c3db803766dcf96aeb8486b1580b61deb9ac28c144674L31) so not sure if this needs an update.

I'll let @anomiex weigh in on that, since he worked on this part of the code in #29322

gfog-floqast:
> missing a check for the new request-reviews action parameter

Can I get more clarification on this, I'm not sure where I should be looking. thx

jeherve:
> Can I get more clarification on this, I'm not sure where I should be looking. thx

Sure thing. Here, you add a review request when needed:
https://github.com/Automattic/jetpack/pull/30653/files#diff-d99dd1c9d319cf9914b2111b19b7bb88771bcb4e4cee56ace33a4a95812bda3eR36

Here, you added a new Action parameter, `request-reviews`, so someone setting up the GitHub action in their own repo can decide to request reviews or not:
https://github.com/Automattic/jetpack/pull/30653/files#diff-b6d548c4bd130b59b0d693c1bd892f31aec433b2abfdd7067dc4f3f9ca5b031bR26

But you do not check for that parameter later on, so that parameter is useless right now. I would recommend that you check if that parameter is set before to call `requestReview` here:
https://github.com/Automattic/jetpack/pull/30653/files#diff-d99dd1c9d319cf9914b2111b19b7bb88771bcb4e4cee56ace33a4a95812bda3eR36

gfog-floqast:
garr - missed it in my commit :D  - It's there now

projects/github-actions/required-review/src/requirement.js
```diff
@@ -31,10 +32,12 @@ function buildReviewerFilter( config, teamConfig, indent ) {
...
-			return printSet(
-				`${ indent }Members of ${ team }:`,
-				reviewers.filter( reviewer => members.includes( reviewer ) )
-			);
+			const reviewSatisfied = reviewers.filter( reviewer => members.includes( reviewer ) );
+			const request = core.getInput('request-reviews')
+			if ( reviewSatisfied.length === 0 && request == 'true' ) {
+				await requestReview( team );
+			}
+			return printSet( `${ indent }Members of ${ team }:`, reviewSatisfied );
```
anomiex: This is not the right place to add the requesting. If the configuration is requiring a review from "teamA or teamB", and someone from teamB already reviewed, there's no need to request a review from teamA. Also putting it here may result in redundant API calls if teamA is listed in multiple requirement blocks, and will almost certainly result in more API calls than are needed.

IMO the best thing to do would be, somewhere in
https://github.com/Automattic/jetpack/blob/8602a933da022394247f3dff851b8bec639ed6c9/projects/github-actions/required-review/src/main.js#L84-L88
add the list of needed teams for `r` to a `Set`. You'll need to figure out how to get just the needed teams, of course. For example, if review from both teamX and ( teamY or userBob ) is required, and there's already a review from teamY, you'd ideally only want to request teamX, not teamY again and not userBob either.

Then, just after
https://github.com/Automattic/jetpack/blob/8602a933da022394247f3dff851b8bec639ed6c9/projects/github-actions/required-review/src/main.js#L93-L96
if the new input is true and the set isn't empty, print a message along the lines of `Requesting reviews from ${ members of the set }` then pass the whole set to requestReviewer. Since the [request reviewers API endpoint](https://docs.github.com/en/rest/pulls/review-requests?apiVersion=2022-11-28#request-reviewers-for-a-pull-request) accepts multiple reviewers in one call, only one call should need to be made.

projects/github-actions/required-review/src/get-username.js
```diff
@@ -0,0 +1,25 @@
+const core = require( '@actions/core' );
+const github = require( '@actions/github' );
+const { WError } = require( 'error' );
+
+/**
+ * Fetch Username based on @user
+ *
+ * @param {string} user - @ followed by a GitHub user name.
+ */
+async function getUsername( user ) {
```
anomiex: Add caching in here please. If called twice for the same `user`, it shouldn't hit the API again.

projects/github-actions/required-review/src/request-review.js
```diff
@@ -0,0 +1,46 @@
...
+ */
+async function requestReviewer( team ) {
+	const octokit = github.getOctokit( core.getInput( 'token', { required: true } ) );
+	const owner = github.context.payload.repository.owner.login;
+	const repo = github.context.payload.repository.name;
+	const pr = github.context.payload.pull_request.number;
+
+	if ( team.startsWith( '@' ) ) {
+		// Handle @singleuser virtual teams. Fetch the correct username case from GitHub
+		// to avoid having to worry about edge cases and Unicode versions and such.
```
anomiex: Is that a concern here? We do it in the other place because we're comparing the name from the config with the name returned on the review. But here I'd hope submitting "someusername" instead of "someUserName" would just do the right thing.

anomiex (dismissed):
> > Regarding the `// prettier-ignore` [comment](https://github.com/Automattic/jetpack/pull/30653#discussion_r1192526060), is there anything you would like me to do with it? I copied that from the [team-members file](https://github.com/Automattic/jetpack/pull/30653/files#diff-a2d2e2075dd774cef44c3db803766dcf96aeb8486b1580b61deb9ac28c144674L31) so not sure if this needs an update.
> 
> I'll let @anomiex weigh in on that, since he worked on this part of the code in #29322

I don't like when prettier decides to wrap a backquoted string over multiple lines, in large part because vim doesn't handle it very well. So I tend to `prettier-ignore` when that happens.

projects/github-actions/required-review/src/requirement.js
gfog-floqast: Thanks for the review @anomiex  After going through the code I went this route due to how reviewers seemed to be grabbed first and then it looked like it worked back to the teams. This route seemed the most direct w/o making major changes to the rest of the code. With that said, I agree 💯 with limiting unneeded API calls and will look into reducing these as you're suggesting.

I'm wondering if it might be reasonable to say that review requests only occur on the first run? On that initial run I think it's safe to say that there would be no reviews therefor all required reviewers should be requested. Then as reviews are received, there shouldn't be the further need to request them again, as you point out. There could be a case where a review was removed (new commit push, etc), but thinking of how codeowners files works, it only requests reviews the first time iirc, which is the functionality I'm looking to introduce. Not sure if this would make the above any more straight forward.

anomiex: If someone pushes a commit to a PR that touches additional files, that might also trigger a need for more reviewers. 

For example, if someone from teamX says "oh, you also need to change this other thing over here" in their review, and that other thing happens to require review from teamZ.

gfog-floqast: yup - great point. 👍🏻

projects/github-actions/required-review/src/request-review.js
gfog-floqast: Tested, and yes `someUserName` works just as well as `someusername`. So guess there's really no reason to factor out [this code](https://github.com/Automattic/jetpack/pull/30653/files#diff-a2d2e2075dd774cef44c3db803766dcf96aeb8486b1580b61deb9ac28c144674L26) as here we can just use `const login = team.slice( 1 );` and not touch `team-members.js`.

If you agree, I'll just put that code back so I'm touching fewer files.

anomiex: Sounds good to me!

gfog-floqast:
Please let me know what you think of this approach. The main issue that I'm running into is accounting for 1) a nested `all-of` inside an `any-of` along with 2) a nested `any-of` inside an `all-of`. I tried creating a way to only add the `alls` in 1) if the `any` was not satisfied, but this then caused issues in 2) where the `all` would not get added if the `any` _was_ satisfied. Without some sort of messy counter I'm not sure how best to handle this situation. 

Currently I'm thinking the most optimal solution would be to add the `all-of` to the list for review request and if it gets an unneeded request that would be better than having a needed request not getting added. I can also add a check to see if a review is already requested and then not request again. This would lower notifications, but may cause issues when a new file needed to be touched that included a reviewer from scenario 1) who would now be required again. 

I'm open to other suggestions of course, I started running in circles so thought best to reach out. This is my first time working with javascript so I may not be familiar with some other tricks that exist in this language for handing a situation like this.

projects/github-actions/required-review/src/requirement.js
```diff
@@ -166,7 +178,7 @@ class Requirement {
 			);
 		}
 
-		this.reviewerFilter = buildReviewerFilter( config, { 'any-of': config.teams }, '  ' );
+		this.reviewerFilter = buildReviewerFilter( config, 'any-of', { 'any-of': config.teams }, '  ' );
```
gfog-floqast: This is suggestion, if y'all agree with this solution I would look to update `{ 'any-of': config.teams }` to just be config.teams and rework the related code.

gfog-floqast:
Please let me know what you think of this approach. The main issue that I'm running into is accounting for 1) a nested `all-of` inside an `any-of` along with 2) a nested `any-of` inside an `all-of`. I tried creating a way to only add the `alls` in 1) if the `any` was not satisfied, but this then caused issues in 2) where the `all` would not get added if the `any` _was_ satisfied. Without some sort of messy counter I'm not sure how best to handle this situation.

Currently I'm thinking the most optimal solution would be to add the `all-of` to the list for review request and if it gets an unneeded request that would be better than having a needed request not getting added. I can also add a check to see if a review is already requested and then not request again. This would lower notifications, but may cause issues when a new file needed to be touched that included a reviewer from scenario 1) who would now be required again.

I'm open to other suggestions of course, I started running in circles so thought best to reach out. This is my first time working with javascript so I may not be familiar with some other tricks that exist in this language for handing a situation like this.

anomiex:
I don't like the globals inside `requirements.js`. I think that trying to keep them straight is probably where you're getting confused.

The functions returned from `buildReviewerFilter()` currently return an array of found reviewers, and when things get to the top if the result is non-empty then the requirement is satisfied. Let's change that so the functions now return two values (as an object with two keys): the array of found reviewers and an array or set of needed teams. 

* The [simple string case](https://github.com/Automattic/jetpack/blob/1a69c56be4e87ecd85e728240727842b15c2405c/projects/github-actions/required-review/src/requirement.js#L33-L37) will either return a non-empty array of found reviewers and an empty array/set of needed teams, or an empty array of found reviewers and a 1-element array/set containing the passed-in team.
* The ["any-of"](https://github.com/Automattic/jetpack/blob/1a69c56be4e87ecd85e728240727842b15c2405c/projects/github-actions/required-review/src/requirement.js#L83-L88) and ["all-of"](https://github.com/Automattic/jetpack/blob/1a69c56be4e87ecd85e728240727842b15c2405c/projects/github-actions/required-review/src/requirement.js#L94-L99) cases would both calculate their arrays of found reviewers as they do now. Then if the found array is empty they will merge the needed-teams returned by all the sub-filters to return, otherwise they'll return an empty array/set for needed-teams.

Then I'd replace `isSatisfied()` with a `needsReviewsFrom()` method that returns the needed-teams set instead of a boolean. [The code in main.js](https://github.com/Automattic/jetpack/blob/a9d73d48c21feb033e209ff7fb1e8a7dd44ed5f4/projects/github-actions/required-review/src/main.js#L82-L90) would then look something like this
```js
} else {
	const neededForRequirement = await r.needsReviewsFrom( reviewers );
	core.endGroup();
	if ( neededForRequirement.size === 0 ) {
		core.info( `Requirement "${ r.name }" is satisfied by the existing reviews.` );
	} else {
		core.error( `Requirement "${ r.name }" is not satisfied by the existing reviews.` );
		// Merge into `neededTeams` set for possible use after the loop.
		neededForRequirement.forEach( neededTeams.add, neededTeams );
	}
}
```
Instead of `ok`, before the loop we'd initialize `const neededTeams = new Set()` and the post-loop check would look for `neededTeams.size === 0`.

gfog-floqast:
@jeherve @anomiex I've made updates to address the comments, please let me know if there are any further changes or additions you would like to see. Thanks.

projects/github-actions/required-review/changelog/request-review-if-not-satisfied
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: added
+
+Request review if requirement is not satisfied
```
anomiex:
```suggestion
Option to request review if requirement is not satisfied.
```

projects/github-actions/required-review/src/request-review.js
```diff
@@ -0,0 +1,42 @@
...
+	const octokit = github.getOctokit( core.getInput( 'token', { required: true } ) );
+	const owner = github.context.payload.repository.owner.login;
+	const repo = github.context.payload.repository.name;
+	const pr = github.context.payload.pull_request.number;
+
+	let userReviews = []
+	let teamReviews = []
+
+	for (var i = 0; i < teams.length; i++) {
+		const t = teams[ i ]
```
anomiex: I'd recommend doing it like this instead
```suggestion
	for ( const t of teams ) {
```

projects/github-actions/required-review/src/requirement.js
```diff
@@ -81,22 +85,42 @@ function buildReviewerFilter( config, teamConfig, indent ) {
...
+			const neededTeams = [];
+			for ( const requirementResult of reviewersAny ) {
+				if ( requirementResult.teamReviewers.length != 0 ) {
+					requirementsMet.push( requirementResult.teamReviewers );
+				};
+				if ( requirementResult.neededTeams.length != 0 ) {
+					neededTeams.push( requirementResult.neededTeams );
+				};
+			};
+			if( requirementsMet.some( a => a.length > 0 ) ){ // If there are requirements met, zero out the needed teams
```
anomiex: The way you have the code above written, every element of `requirementsMet` will have `.length > 0`. So as long as `requirementsMet.length > 0`, we're good here.

projects/github-actions/required-review/src/requirement.js
```diff
@@ -31,10 +32,13 @@ function buildReviewerFilter( config, teamConfig, indent ) {
...
 			const members = await fetchTeamMembers( team );
-			return printSet(
+			teamReviewers = reviewers.filter( reviewer => members.includes( reviewer ) );
+			const neededTeams = teamReviewers.length ? [] : [team];
+			printSet(
 				`${ indent }Members of ${ team }:`,
-				reviewers.filter( reviewer => members.includes( reviewer ) )
+				teamReviewers
 			);
+			return {teamReviewers, neededTeams};
```
anomiex: You changed `printSet` so this will work:
```suggestion
			return printSet(
				`${ indent }Members of ${ team }:`,
				teamReviewers,
				neededTeams
			);
```

projects/github-actions/required-review/src/requirement.js
```diff
@@ -81,22 +85,42 @@ function buildReviewerFilter( config, teamConfig, indent ) {
...
+					requirementsMet.push( requirementResult.teamReviewers );
+				};
+				if ( requirementResult.neededTeams.length != 0 ) {
+					neededTeams.push( requirementResult.neededTeams );
+				};
+			};
+			if ( neededTeams.length !== 0 ) { // If there are needed teams, zero out requirements met
+				return printSet( `${ indent }=>`, [], [ ...new Set( neededTeams.flat( 1 ) ) ] );
+			};
+			return printSet( `${ indent }=>`, [ ...new Set( requirementsMet.flat( 1 ) ) ], [ ...new Set( neededTeams.flat( 1 ) ) ] );
```
anomiex: You know at this point that `neededTeams` is empty, so no need to run it through Set.
```suggestion
			return printSet( `${ indent }=>`, [ ...new Set( requirementsMet.flat( 1 ) ) ], [] );
```

anomiex (commented):
Looks pretty good. I left a few comments inline, and I'll kick off the CI jobs where the linter will have some additional comments.

gfog-floqast:
@anomiex Addressed comments and committed suggestions - I didn't get the lint results before committing suggested changes so I lost those runs if you could please retrigger those I'll make any needed updates.

gfog-floqast:
@jeherve @anomiex could one of y'all re-trigger the workflows for me? please and thank you.

gfog-floqast:
@jeherve Linting errors should be fixed and ready for a final 🤞🏻 run.

projects/github-actions/required-review/src/request-review.js
```diff
@@ -1,27 +1,25 @@
 const core = require( '@actions/core' );
 const github = require( '@actions/github' );
-const getUsername = require( './get-username.js' );
 
 /**
  * Request review from the given team
  *
- * @param {string} team - GitHub team slug, or @ followed by a GitHub user name.
+ * @param {object} teams - GitHub team slug, or @ followed by a GitHub user name.
```
anomiex:
```suggestion
 * @param {string[]} teams - GitHub team slug, or @ followed by a GitHub user name.
```

gfog-floqast:
Thanks @anomiex I committed your suggestion, also pushed up final linting error corrections.

anomiex (approved):
If CI is happy, let's go for it. Thanks!

PR #30654, opened 2023-05-11
[RNMobile] refactor webview pressable

jhnstn:
Refactor the VideoPress embed overlay on Android

## Proposed changes:

* Nest the `Sandbox` component inside the embed `Pressable` component

### Other information:

## Testing instructions:

On Android
* Visit a post with a VideoPress block
* Verify that the WebView does not open when selecting the block
* Tap on the video player
* The video should load in a Native WebView

On iOS
- Verify that the video does not play when selecting a VideoPress block.

### DIFF
--- a/projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
+++ b/projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
@@ -148,17 +148,28 @@ export default function Player( { isSelected, attributes } ) {
 
 	let overlay = ! isSelected && <View style={ style[ 'videopress-player__overlay' ] } />;
 
-	// On Android render an overlay to send the embed markup to a native WebView.
-	if ( isSelected && IS_ANDROID ) {
-		overlay = (
-			<View style={ style[ 'videopress-player__overlay' ] }>
-				<Pressable
-					style={ style[ 'videopress-player__open-embed-button' ] }
-					onPress={ () => {
-						requestEmbedFullscreenPreview( html, title );
-					} }
-				/>
-			</View>
+	let sandbox = (
+		<SandBox
+			html={ html }
+			onWindowEvents={ { message: onSandboxMessage } }
+			viewportProps="user-scalable=0"
+			testID="videopress-player"
+			onLoadEnd={ onLoadEnd }
+		/>
+	);
+
+	// On Android, we need to open the embed preview in a Native WebView
+	if ( IS_ANDROID ) {
+		sandbox = (
+			<Pressable
+				style={ style[ 'videopress-player__open-embed-button' ] }
+				onPress={ () => {
+					requestEmbedFullscreenPreview( html, title );
+				} }
+			>
+				<View style={ style[ 'videopress-player__overlay' ] } />
+				{ sandbox }
+			</Pressable>
 		);
 	}
 
@@ -171,15 +182,7 @@ export default function Player( { isSelected, attributes } ) {
 		<View style={ [ style[ 'videopress-player' ], { aspectRatio } ] }>
 			{ overlay }
 			{ showLoadingOverlay && loadingOverlay }
-			{ html && (
-				<SandBox
-					html={ html }
-					onWindowEvents={ { message: onSandboxMessage } }
-					viewportProps="user-scalable=0"
-					testID="videopress-player"
-					onLoadEnd={ onLoadEnd }
-				/>
-			) }
+			{ html && sandbox }
 		</View>
 	);
 }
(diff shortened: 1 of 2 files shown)


projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
```diff
@@ -148,17 +148,28 @@ export default function Player( { isSelected, attributes } ) {
 
 	let overlay = ! isSelected && <View style={ style[ 'videopress-player__overlay' ] } />;
```
fluiddot: As mentioned by the ESLint CI job, this should be a `const`.

projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
```diff
@@ -148,17 +148,28 @@ export default function Player( { isSelected, attributes } ) {
...
+	// On Android, we need to open the embed preview in a Native WebView
+	if ( IS_ANDROID ) {
+		sandbox = (
+			<Pressable
+				style={ style[ 'videopress-player__open-embed-button' ] }
+				onPress={ () => {
+					requestEmbedFullscreenPreview( html, title );
+				} }
+			>
+				<View style={ style[ 'videopress-player__overlay' ] } />
```
fluiddot, line 168: I understand we add the overlay here to disable touch events in the WebView, is this accurate?

projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
```diff
@@ -171,15 +182,7 @@ export default function Player( { isSelected, attributes } ) {
 		<View style={ [ style[ 'videopress-player' ], { aspectRatio } ] }>
 			{ overlay }
```
fluiddot: Nitpick: If the overlay is only used to prevent touch events get to the player. There's an alternative using the [`pointerEvents` prop](https://reactnative.dev/docs/view#pointerevents).

```
<View
	style={ [ style[ 'videopress-player' ], { aspectRatio } ] }
	pointerEvents={ isSelected ? 'auto' : 'none' }
>
	{ showLoadingOverlay && loadingOverlay }
	{ html && sandbox }
</View>
```

projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
fluiddot: I'm wondering if we could reuse the `overlay` constant defined above here, WDYT?

fluiddot (commented):
I tried to reproduce the issue I experienced in the previous version, where the `Pressable` component wasn't working, but seems it's now working 🎊. I'll hold back my approval until we solve the ESLint issue, but from my side, this is good to go. Thanks @jhnstn for fixing it so quickly 🙇 !

I added a couple of suggestions but feel free to skip them in favor of merging the PR as-is.

projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
jhnstn: Oh interesting, I'll take a look at that prop.

projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
jhnstn: > I understand we add the overlay here to disable touch events in the WebView, is this accurate?

Correct, I noticed that the video sometimes starts playing as the WebView is opening. Which is ironic since we need the web view to probably handle the touch events 😅

projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
jhnstn: fixed in 2d886e2aa3e954fc02ff9a00d35458abe9d424a6

projects/packages/videopress/src/client/block-editor/blocks/video/components/player/index.native.js
jhnstn: fixed in 2d886e2aa3e954fc02ff9a00d35458abe9d424a6

jhnstn: > I'm wondering if we could reuse the overlay constant defined above here, WDYT?

I ended up removing the `overlay` variable so it's just the single `View` now.

fluiddot (approved):
I confirmed what I shared in https://github.com/Automattic/jetpack/pull/30654#pullrequestreview-1424168870. LGTM 🎊 !

PR #30656, opened 2023-05-11
AI Assistant: Parse markdown inside ShowLittleByLittle

oskosk:
Moves markdown parsing to the ShowLittleByLittle intead of doing it in the hook

This was making retry queries be rendered with escaped html as the second query would send html (stored in `attributes.content` to the API

Fixes #

## Proposed changes:

* Moves call to `markdownConverter.render()` into the ShowLittleByLittle component
* Fixes lack of new lines transformation to `<br/>`. Evident when requesting poems of haikus

### Other information:

No

## Testing instructions:

1. Add an AI Assitant block
2. Run a query
3. Run a second query after the first one comes back
4. Expect to see no HTML markup in the second result


retrofox (approved):
This is an excellent enhancement! 🚀

PR #30678, opened 2023-05-12
CRM: Client Portal PHP fatal error when loading endpoints/shortcodes

cleacos:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3003

## Proposed changes:
This PR fixes a PHP fatal error during the Client Portal module initialization. The module was initializing the endpoints too late, via action hook. In some cases, there are third parties (plugins/themes) that could process page content during the WP init process, which implies a call to the `do_shortcode` function, which will initialize the Client Portal module, but the endpoints initialization is enqueued in the WP init process and is not ready to be used, causing that fatal error.

To replicate it, add this code, for example, in the theme `functions.php` file:

```
add_action( 'init', function() {
	do_shortcode( '[jetpackcrm_clientportal]' );	
}, 5);
```

### Other information:

## Jetpack product discussion

N/A

N/A 

## Testing instructions:

- Add the `do_shortcode` code described above to your `functions.php` file.
- Add the `[jetpackcrm_clientportal]` shortcode in a page
- Save the page and try to visit it (I did it in the incognito tab)
- In trunk, you will get the fatal error
- With this branch, everything will work

### DIFF
--- a/projects/plugins/crm/changelog/fix-crm-3003-client-portal-fatal-error-loading-endpoints
+++ b/projects/plugins/crm/changelog/fix-crm-3003-client-portal-fatal-error-loading-endpoints
@@ -1,4 +1,4 @@
 Significance: patch
 Type: fixed
 
-Fix a Client Portal faltal error initializationg the endpoints
+Client Portal: Fix a fatal error initializing endpoints and shortcodes.
--- a/projects/plugins/crm/changelog/fix-crm-3003-client-portal-fatal-error-loading-endpoints
+++ b/projects/plugins/crm/changelog/fix-crm-3003-client-portal-fatal-error-loading-endpoints
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fix a Client Portal faltal error initializationg the endpoints
--- a/projects/plugins/crm/modules/portal/class-client-portal.php
+++ b/projects/plugins/crm/modules/portal/class-client-portal.php
@@ -34,7 +34,8 @@ class Client_Portal {
 		$this->router = new Client_Portal_Router();
 		$this->render = new Client_Portal_Render_Helper( $this );
 
-		$this->init();
+		// Initializes it later. Priority 10
+		add_action( 'init', array( $this, 'init' ) );
 	}
 
 	/**
--- a/projects/plugins/crm/modules/portal/class-client-portal.php
+++ b/projects/plugins/crm/modules/portal/class-client-portal.php
@@ -57,7 +57,7 @@ class Client_Portal {
 		// Custom login redirect hook (this one is in our $router).
 		add_action( 'login_redirect', array( $this->router, 'redirect_contacts_upon_login' ), 10, 3 );
 		// Initializes all endpoints (including the ones from external plugins).
-		add_action( 'init', array( $this, 'init_endpoints' ) );
+		$this->init_endpoints();
 		// this catches failed logins, checks if from our page, then redirs
 		// From mr pippin https://pippinsplugins.com/redirect-to-custom-login-page-on-failed-login/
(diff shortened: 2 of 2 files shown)


tbradsha:
I'm hesitant to merge this without being able to reproduce the original issue. That said, when reviewing let's make sure this change doesn't impact additional endpoints (e.g. Client Portal Pro ones).

cleacos:
> when reviewing let's make sure this change doesn't impact additional endpoints (e.g. Client Portal Pro ones).

I did and checked it with Client Portal Pro activated, and all good. All plugins are loaded before the WP `init` hook, so any WP `action` added from the extension is ready when the Client Portal is initialized.

At that point, we don't need to enqueue the endpoint initialization to the WP `init` hook again.

gogdzl (approved):
It works.
The explanation by @cleacos checks out.

I generally dislike fixing something that we cannot replicate properly, but this code will do no harm.

If the problem arises again we can try another approach.

gogdzl (changes requested):
Actually, Client Portal Pro stopped working with this PR.
Most probably because of load order.

Install Client Portal Pro and check that 'Your files' tab stops working with this PR.

cleacos:
The Client Portal Pro adds those menus on WP `init` but with the default priority `10`.

- If we initialize the Client Portal module later, with priority `10`, it works.

The thing is that for the extension to be able to add the endpoints via the `jpcrm_client_portal_register_endpoint` hook action, it has to be initialized before the Client Portal module. A better approach could be, first, we have to load the module fully and then attach the endpoints, but this is another issue/improvement, not for now.

cleacos:
I'm going to push the option of delaying the module initialization a bit, priority 10. I tested it and it works well.

gogdzl (changes requested):
I don't understand how this would help us.

We are moving a bunch of hooks and actions to a completely different load order, and that got me confused.

I also fail to see how this would fix said problem, because the load order wasn't the original problem, and even if it was it's being called in the same priority as before.

I need more information because I didn't fully understand how this solves the problem.

cleacos:
> We are moving a bunch of hooks and actions to a completely different load order, and that got me confused.

First, we have to  group all of them to be able to add shortcodes and the API endpoints at the same moment to avoid calls to `do_shortcode` without the endpoints initialized:

- `add_shortcode(...)`
- `$this->init_endpoints()`

Then, we have to move that group and initialize it at priority `10` to wait for the Client Portal PRO initialization first (adding the action `add_action - jpcrm_client_portal_register_endpoint`) and then the Client Portal module that will initialize the endpoints and call that action (`do_action - jpcrm_client_portal_register_endpoint`) to add the Client Portal PRO endpoints. That is the reason for moving to priority `10`.

**This is the order currently**

1. init(1): Client Portal initialization
   - add_shortcode
   - add_actions
   - Postpone the endpoints' initialization to init(10). The reason is that it needs Client Portal PRO initialized before it.

2. init(10): Client Portal PRO initialization
   - Add this action `jpcrm_client_portal_register_endpoint` to add its endpoints

3. init(10): Client Portal - init_endpoints()
   - Do the action `jpcrm_client_portal_register_endpoint` to load third parties endpoints (CP PRO)

**This would be the order in this PR:**

1. init(1): Client Portal initialization
   - Group and postpone the whole initialization to init(10).

2. init(10): Client Portal PRO initialization
   - Add this action `jpcrm_client_portal_register_endpoint` to add its endpoints

3. init(10): Client Portal Initialization
   - add_shortcode
   - add_actions
   - init_endpoints
      - Do the action `jpcrm_client_portal_register_endpoint` to load third parties endpoints (CP PRO)

gogdzl (approved):
Ok, thanks for the in-depth explanation!

Ship it!

PR #30682, opened 2023-05-12
Related Posts: Fix context for use with blocks

### LINKED ISSUE #30673: Related Posts: Fix "Display context" feature for use with blocks
When enabling the "Display context (category or tag)" toggle, the preview shows some escaped HTML instead of the proper quote characters. The annotated image below shows:

1. The escaped HTML where opening and closing quotes should be.
2. How it should look, with nicely styled quotes.

Image

Additionally, if the user modifies the "context" via PHP, the editor isn't provided with sufficient useful data to properly preview/render the context. Here's an example when modifying the context to show the post author [as discussed in our support docs](https://jetpack.com/support/related-posts/customize-related-posts/#author). On the front end we see this:

SCR-20230512-nfso

So far, so good. But if we then view a post that uses the block version of Related Posts, we see this:

SCR-20230512-ngrw

Finally, if we switch over to the editor, we see this:

SCR-20230512-nhjl

Based on my testing, it looks like we'll need a separate context for use in the block editor. Likewise, we should avoid applying any global PHP filters to the context as that input isn't useful inside the editor and probably makes more sense to be added directly within the editor so it can be controlled on a block-by-block basis.

a8ck3n:
Fixes #30673 

## Proposed changes:

Adds a second "block context" for use in block rendering. The existing context is still generated for use by the non-block version of the output. With the new context, we also properly set up a link to the category, tag, or comments as appropriate.

With the new context in place, the rendered block should show…

1. Linked up category.
2. Linked up tag.

SCR-20230512-ssje

In the block editor we should see the same:

SCR-20230512-stda

### Other information:

## Jetpack product discussion

pejTkB-yT-p2

No

## Testing instructions:

1. Spin up a test site with this branch.
3. Make sure Related Posts are enabled via Jetpack → Settings → Traffic → Related Posts.
4. Create a new post and add a Related Posts block.
5. Confirm the context (category or tag) is rendered correctly in the editor and on the front-facing page (as annotated in the screenshots). You may need to publish before the placeholders are replaced with actual post content.


projects/plugins/jetpack/modules/related-posts/jetpack-related-posts.php
```diff
@@ -395,12 +395,22 @@ public function render_block_item( $related_post, $block_attributes ) {
...
 		}
 
-		if ( ( $block_attributes['show_context'] ) && ! empty( $related_post['context'] ) ) {
-			$context_tag = sprintf(
-				'<li class="jp-related-posts-i2__post-context">%1$s</li>',
-				esc_html( $related_post['context'] )
-			);
-
+		if ( ( $block_attributes['show_context'] ) && ! empty( $related_post['bcon'] ) ) {
+			// New render logic, ignoring original context.
```
jsnmoon: Could we explain why we're ignoring the original context in this code comment?

Also, could we add some background information on the source of this context? Is it provided by the related posts API endpoint?

projects/plugins/jetpack/modules/related-posts/jetpack-related-posts.php
```diff
@@ -1339,6 +1349,7 @@ public function get_related_post_data_for_post( $post_id, $position, $origin ) {
 				$this->to_utf8( $this->generate_related_post_context( $post->ID ) ),
 				$post->ID
 			),
+			'bcon'     => $this->generate_related_post_context_block( $post->ID ),
```
jsnmoon: Could we use a more descriptive name for this parameter?

projects/plugins/jetpack/modules/related-posts/jetpack-related-posts.php
```diff
@@ -1666,6 +1677,62 @@ protected function filter_non_public_posts( array $related_posts ) {
...
+	 * Order of importance:
+	 *   - First category (Not 'Uncategorized')
+	 *   - First post tag
+	 *   - Number of comments
+	 *
+	 * @param int $post_id - the post ID.
+	 * @uses get_the_category, get_the_terms, get_comments_number, number_format_i18n, __, _n
+	 * @return string
+	 */
+	protected function generate_related_post_context_block( $post_id ) {
```
jsnmoon, line 1708: What's the rationale for making this new function instead of updating the existing `generate_related_post_context` function?

Is it possible to update the original context to include the link data?

jsnmoon (commented):
This works as expected, but I don't understand why we're creating a new `generate_related_post_context_block` when we could update the existing  `generate_related_post_context` function.

I left some comments inline.

projects/plugins/jetpack/modules/related-posts/jetpack-related-posts.php
a8ck3n: > Could we explain why we're ignoring the original context in this code comment?

This is explained a little in the [related issue](https://github.com/Automattic/jetpack/issues/30673) but the summary is that I think we need a second context for use within the block environment to avoid breaking the existing behaviour.

The current behaviour is to pull in the existing context and render it as needed. You'll see that this kinda works as long as the context isn't "filtered" through any PHP snippets but it breaks things if any filtering is applied. See screenshots in linked issue for that.

> Also, could we add some background information on the source of this context? Is it provided by the related posts API endpoint?

The context is built locally in the PHP function `get_related_post_data_for_post()`. That function sets up data needed for display after getting a list of post IDs from the API in `get_related_post_ids()`.

Based on my understanding of the code paths, we have two options:

1. Keep the original context for legacy rendering and add a second context for the block environment (the solution I've coded for here) or…
2. Insert something in the code path so we know if it's a legacy widget or a block and alter the context based on that. This second option has a higher likely hood of breakage though as it's used in different scenarios like in the Customizer.

projects/plugins/jetpack/modules/related-posts/jetpack-related-posts.php
a8ck3n: This is mostly the same as my comment above but the main reasoning is to keep the legacy code path working as is with the ability to filter the results via PHP code snippets.

Then we have the new "block context" for use in the block editor where I don't think we really want a user-filtered value being pushed into the editor. My thinking is the context should work much like the date. If enabled, we generate a date for display. We don't enable any hooks to allow the user to then modify how the date is formatted before we output to screen.

In the legacy code path we have some hooks that allow the user to modify the context as they see fit. They can use this to add a timestamp, author, tag cloud, etc before the final string is added to the page HTML. It feels like something we should preserve but I'm not sure we want to bring it into the block editor.

> Is it possible to update the original context to include the link data?

To actually answer this question, yes, we could update the original context but we'd be modifying the "input" to the user-supplied filter function which could result in breakage, depending on how defensively coded their PHP snippets are.

This is all with the caveat that this is a bit of a legacy scenario. The original context would only be seen (and user-supplied filters run) on sites using themes that haven't been updated for FSE.

projects/plugins/jetpack/modules/related-posts/jetpack-related-posts.php
jsnmoon: Got it; thank you for the context. This approach makes sense to me, though we'll want to update our documentation to reflect this new behavior following a release.

Also, it'd be great to get some of this context into the code as code comments 🙂. I wouldn't mind if we did this in a follow-up PR, though!

projects/plugins/jetpack/modules/related-posts/jetpack-related-posts.php
jsnmoon: I agree with your approach of not altering non-FSE Related Posts widget behavior.

projects/plugins/jetpack/modules/related-posts/jetpack-related-posts.php
jsnmoon: This could be tackled in a follow-up PR 🙂

jsnmoon (dismissed):
The rationale for these changes makes sense to me. I'm a bit concerned about the E2E test breakage, so please confirm that they're unrelated to these changes before merging 👍

projects/plugins/jetpack/modules/related-posts/jetpack-related-posts.php
a8ck3n: Added some comments around the 'context' and 'block_context' to help clarify things. Also added a ReadMe file with some notes.

projects/plugins/jetpack/modules/related-posts/jetpack-related-posts.php
```diff
@@ -1669,6 +1686,62 @@ protected function filter_non_public_posts( array $related_posts ) {
...
+		$categories = get_the_category( $post_id );
+		if ( is_array( $categories ) ) {
+			foreach ( $categories as $category ) {
+				$cat_link = get_category_link( $category );
+				if ( 'uncategorized' !== $category->slug && '' !== trim( $category->name ) ) {
```
jeherve, line 1713: Is the intent here to avoid "Uncategorized", or to avoid the default category (which can be named differently now)?

projects/plugins/jetpack/modules/related-posts/jetpack-related-posts.php
```diff
@@ -395,12 +395,24 @@ public function render_block_item( $related_post, $block_attributes ) {
...
+			// Note: The original 'context' value is not used when rendering the block.
+			// It is still generated and available for the legacy rendering code path though.
+			// See './related-posts.js' for that usage.
+			$context_tag   = '';
+			$block_context = $related_post['block_context'];
+			if ( ! empty( $block_context['link'] ) ) {
+				$context_tag = sprintf(
+					'<li class="jp-related-posts-i2__post-context"><a href="%1$s">%2$s</a></li>',
+					$block_context['link'],
+					$block_context['text']
```
jeherve: Do we need to do some escaping here? It may be overkill, but I'm wondering if it may be useful in some cases.

jeherve (commented):
This looks good to me; I only have 2 questions about the implementation.

projects/plugins/jetpack/modules/related-posts/jetpack-related-posts.php
a8ck3n: To be honest, I was just aiming to be consistent with the existing behaviour used in the non-block rendering.

I do think this is probably the "correct" behaviour though, in that it will only single out posts based on the slug. If the user has renamed their default category (name & slug), those posts will be recognized. I'm not sure I'd argue that we need to keep this behaviour in the new code path, if you think it makes sense to remove the check entirely.

a8ck3n: PS: The function comment calls this out as the original intent.

> First category (Not 'Uncategorized')

projects/plugins/jetpack/modules/related-posts/jetpack-related-posts.php
a8ck3n: My understanding is that it's not needed since both values are pulled from core functions but I don't think it hurts to add it. Incoming!

> most WordPress functions properly prepare the data for output, and additional escaping is not needed.

https://developer.wordpress.org/apis/security/escaping/

a8ck3n: Also, (a) the rest of the function is escaping data without concern for the origin and (b) easier for readers of the function if they don't have to double-check how the data is generated.

projects/plugins/jetpack/extensions/blocks/related-posts/edit.js
```diff
@@ -101,7 +104,8 @@ function RelatedPostsEditItem( props ) {
 			) }
 			{ props.displayContext && (
 				<div className="jp-related-posts-i2__post-context has-small-font-size">
-					{ props.post.context }
+					{ contextHasLink && <a href={ contextLink }>{ contextText }</a> }
+					{ contextHasLink === false && contextText }
```
jeherve, line 109: Should we ensure that `contextText` is available?

projects/plugins/jetpack/modules/related-posts/readme.md
```diff
@@ -0,0 +1,36 @@
...
+
+```
+[{"id":919},{"id":9},{"id":903}]
+```
+
+Note the data used in rendering is all generated locally once the post IDs have been provided.
+
+## Block Source
+
+The source for the Gutenburg block lives separately at:
```
jeherve:
```suggestion
The source for the Gutenberg block lives separately at:
```

projects/plugins/jetpack/modules/related-posts/readme.md
```diff
@@ -0,0 +1,36 @@
...
+
+For the block-based version:
+
+1. Make an API call to get the working post IDs.
+2. Call `get_related_post_data_for_post()` for each ID.
+3. Use generated data to render the block when called.
+
+## Prerequisites
+
+If using a FSE-enabled theme, the site-wide Related Posts setting will be ignored. For other themes, the site-wide setting is honoured as long as a Related Posts block or shortcode is not found on the page. The Related Posts code paths will only output once per page.
```
jeherve:
```suggestion
If using a block theme, the site-wide Related Posts setting will be ignored. For other themes, the site-wide setting is honored as long as a Related Posts block or shortcode is not found on the page. The Related Posts code paths will only output once per page.
```

jeherve (changes requested):
This tests well for me. I left a few last notes below.

jeherve (approved):

PR #30687, opened 2023-05-12
Social: Update the UI logic for checking broken social connections

pablinos:
With the move to using the connection ID rather than the token ID we needed to update the logic in the classic editor so that connections are correctly identified as broken.

Similarly, there has been a longstanding problem with the twitter options section of the publicize-form where it only checks the `enabled` property. This needs to also check the `is_healthy` property to also disable the options if there are no 'healthy' twitter connections.

Fixes #

## Proposed changes:
* Updates the connection check in the twitter options to also check the connection is healthy
* Updates the broken connection check in the classic editor UI to check against the `connection_id`

### Other information:

No

## Testing instructions:

* With a site running Jetpack, not Jetpack Social...
* Create a twitter connection on the site, and either mark twitter as down or break the connection (invalidate the token etc)
* Without this PR, in the block editor, the connection will be shown as broken but you'll still see the twitter thread options
* With this PR, the twitter thread options should be correctly hidden once the test response flags the connection as broken
* Enable the classic editor plugin
* Without this PR, in the editor you will be told the connection is broken, but the connection in the UI will still be enabled
* With this PR, the connection will be disabled.

image

*The classic editor with a broken connection and this fix*

image

*The block editor with a broken connection and this fix - correctly hiding the tweetstorm options*


manzoorwanijk (dismissed):
Works as expected.

gmjuhasz (approved):

PR #30693, opened 2023-05-12
PHP 8 compatibility updates

nelsonec87:
## Proposed changes:

### Other information:

## Jetpack product discussion

* pdqkMK-UN-p2
* #23695 

## Testing instructions:

### DIFF
--- a/projects/plugins/jetpack/modules/tiled-gallery/tiled-gallery/tiled-gallery-rectangular.php
+++ b/projects/plugins/jetpack/modules/tiled-gallery/tiled-gallery/tiled-gallery-rectangular.php
@@ -125,7 +125,7 @@ class Jetpack_Tiled_Gallery_Grouper {
 	 * @return array
 	 */
 	public function get_current_row_size() {
-		$images_left = count( $this->images );
+		$images_left = is_countable( $this->images ) ? count( $this->images ) : 0;
 		if ( $images_left < 3 ) {
 			return array_fill( 0, $images_left, 1 );
 		}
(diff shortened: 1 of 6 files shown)


projects/plugins/jetpack/modules/tiled-gallery/tiled-gallery/tiled-gallery-rectangular.php
```diff
@@ -207,7 +207,8 @@ public function get_grouped_images() {
 	public function apply_content_width( $width ) {
 		foreach ( $this->grouped_images as $row ) {
 			$row->width      = $width;
-			$row->raw_height = 1 / $row->ratio * ( $width - $this->margin * ( count( $row->groups ) - $row->weighted_ratio ) );
+			$group_count     = is_countable( $row->groups ) ? count( $row->groups ) : 0;
+			$row->raw_height = 1 / $row->ratio * ( $width - $this->margin * ( $group_count - $row->weighted_ratio ) );
```
markbiek, line 211: I wonder if there's a different way to handle this. It seems like having a `$group_count` of `0` is going to lead to unexpected results because the `$row->raw_height` will end up being negative.

projects/plugins/jetpack/modules/tiled-gallery/tiled-gallery/tiled-gallery-rectangular.php
```diff
@@ -225,8 +226,9 @@ public function calculate_group_sizes( $row ) {
 		$group_widths_array = array();
 		foreach ( $row->groups as $group ) {
 			$group->height = $row->height;
+			$image_count   = is_countable( $group->images ) ? count( $group->images ) : 0;
```
markbiek, line 229: I worry about the same thing here. If `$image_count` is `0`, then the `$group->raw_width` will end up being negative.

nelsonec87:
Rebased as check was failing

projects/plugins/jetpack/modules/tiled-gallery/tiled-gallery/tiled-gallery-rectangular.php
nelsonec87: Having 
`$row->raw_height = 1 / $row->ratio * ( $width - $this->margin * max(0, $group_count - $row->weighted_ratio ) );`
would prevent the negative (in some cases), but it's hard to say if there would be unwanted consequences since it also depends on other variables like `$width`, `$this->margin`, and `$row->weighted_ratio`. It's a weird one 🤔

The proposed change doesn't change the existing behavior, though.

markbiek (approved):
LGTM!

PR #30699, opened 2023-05-12
PHP 8 compatibility updates

nelsonec87:
## Proposed changes:

### Other information:

## Jetpack product discussion

* pdqkMK-UN-p2
* #23695 

## Testing instructions:

### DIFF
--- a/projects/plugins/jetpack/modules/widgets/gallery.php
+++ b/projects/plugins/jetpack/modules/widgets/gallery.php
@@ -259,7 +259,6 @@ class Jetpack_Gallery_Widget extends WP_Widget {
 	 * @return string String of HTML representing a slideshow gallery
 	 */
 	public function slideshow_widget( $args, $instance ) {
-		$attr = array();
 		global $content_width;
 
 		require_once plugin_dir_path( realpath( __DIR__ . '/../shortcodes/slideshow.php' ) ) . 'slideshow.php';
--- a/projects/plugins/jetpack/modules/widgets/gallery.php
+++ b/projects/plugins/jetpack/modules/widgets/gallery.php
@@ -259,6 +259,7 @@ class Jetpack_Gallery_Widget extends WP_Widget {
 	 * @return string String of HTML representing a slideshow gallery
 	 */
 	public function slideshow_widget( $args, $instance ) {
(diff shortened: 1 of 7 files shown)


projects/plugins/jetpack/modules/widgets/gallery.php
```diff
@@ -259,6 +259,7 @@ public function circle_widget( $args, $instance ) {
 	 * @return string String of HTML representing a slideshow gallery
 	 */
 	public function slideshow_widget( $args, $instance ) {
+		$attr = array();
```
markbiek: I see where we check `$attr` below but I don't see anywhere else where `$attr` gets initialized. I wonder if this is a bug in the original code.

projects/plugins/jetpack/modules/widgets/rsslinks-widget.php
```diff
@@ -221,6 +221,10 @@ public function form( $instance ) {
 	 * @param array  $args Widget arguments.
 	 */
 	private function rss_link( $type, $args ) {
+		$link_text     = null;
+		$rss_type      = null;
+		$subscribe_to  = null;
+		$link_contents = null;
```
markbiek: I'm tempted to move `$link_contents` down to just before line 253.

nelsonec87: I agree it's more readable; done!

projects/plugins/jetpack/modules/widgets/gallery.php
nelsonec87: I saw this weird one... There is even a comment to make phpcs ignore this problem on line 304 and a TODO to fix it.
I thought this was beyond the scope of these changes, but what do you think @markbiek ?

markbiek: I'd ask in the jetpack-developers Slack channel and see what they say. If it is an actual bug, we might be able to handle the fix. Otherwise we can create an issue for the Jetpack team to handle it.

nelsonec87:
Rebased as check was failing

markbiek (approved):
LGTM!

PR #30703, opened 2023-05-13
Debug Helper: Fix Broken Token and IDC Simulator styling

chrisbliss18:
#### Changes proposed in this Pull Request:
At some point in the past, it seems that the debug helper modules had a single shared `style.css` file. Over time, modules split their styling into their own separate CSS files. The Broken Token and IDC Simulator modules continued to use the shared CSS file. The hook name for the Debug Helper page also changed and the code used to conditionally load the CSS file for these modules was not updated to match, resulting in the CSS files not enqueuing for them.

This PR includes a set of fixes that do the following:
- Breaks out the styling for the Broken Token and IDC Simulator modules to their own CSS files.
- Removes the now-unused `inc/css/style.css` file for the Debug Helper modules.
- Fixes the code used to enqueue the CSS file for the Broken Token and IDC Simulator modules so that the module pages are now properly-loading the CSS file
- Updated the styling for the IDC Simulator to use `overflow:auto` rather than `overflow:scroll`.

The changes will affect the following:
* Debug Helper

#### Other information:

#### Does this pull request change what data or activity we track or use?
* No

#### Testing instructions:
* Apply the PR.
* On the Jetpack Debug Jetpack Debug page, enable "Broken token Utilities" and "Identity Crisis Simulation Utility".
* Go to Jetpack Debug > Broken Token and confirm that the buttons under the "Break some stuff" heading are red.


samiff (approved):
Works for me :shipit:

PR #30704, opened 2023-05-13
PHP 8 compatibility updates

nelsonec87:
## Proposed changes:

### Other information:

## Jetpack product discussion

* pdqkMK-UN-p2
* #23695 

## Testing instructions:

### DIFF
--- a/projects/plugins/jetpack/modules/custom-css/csstidy/class.csstidy-print.php
+++ b/projects/plugins/jetpack/modules/custom-css/csstidy/class.csstidy-print.php
@@ -205,7 +205,8 @@ class csstidy_print { // phpcs:ignore
 		}
 
 		if ( ! empty( $this->import ) ) {
-			for ( $i = 0, $size = count( $this->import ); $i < $size; $i++ ) {
+			$import_count = is_countable( $this->import ) ? count( $this->import ) : 0;
+			for ( $i = 0, $size = $import_count; $i < $size; $i++ ) {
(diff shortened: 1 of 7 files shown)


nelsonec87:
Rebased as check was failing

projects/plugins/jetpack/modules/custom-css/csstidy/class.csstidy-print.php
```diff
@@ -205,7 +205,8 @@ public function _print( $plain = false, $default_media = '' ) { // phpcs:ignore
 		}
 
 		if ( ! empty( $this->import ) ) {
-			for ( $i = 0, $size = count( $this->import ); $i < $size; $i++ ) {
+			$import_count = is_countable( $this->import ) ? count( $this->import ) : 0;
+			for ( $i = 0, $size = $import_count; $i < $size; $i++ ) {
```
markbiek: I don't think we need `$size` any more. I think it was only there before because we wanted to avoid calling `count( $this->import )` on every iteration of the loop.

```suggestion
			for ( $i = 0, $import_count; $i < $import_count; $i++ ) {
```

markbiek (changes requested):

projects/plugins/jetpack/modules/custom-css/csstidy/class.csstidy-print.php
nelsonec87: Of course, thank you for catching it!

markbiek (approved):
LGTM!

PR #30705, opened 2023-05-13
PHP 8 compatibility updates

nelsonec87:
## Proposed changes:

### Other information:

## Jetpack product discussion

* pdqkMK-UN-p2
* #23695 

## Testing instructions:

### DIFF
--- a/projects/packages/connection/legacy/class-jetpack-signature.php
+++ b/projects/packages/connection/legacy/class-jetpack-signature.php
@@ -94,7 +94,7 @@ class Jetpack_Signature {
 			// and encoded on the Jetpack side.
 			if ( defined( 'IS_WPCOM' ) && IS_WPCOM ) {
 				// phpcs:ignore WordPress.Security.NonceVerification.Missing
-				if ( empty( $body ) && is_array( $_POST ) && count( $_POST ) > 0 ) {
+				if ( empty( $body ) && is_array( $_POST ) && $_POST !== array() ) {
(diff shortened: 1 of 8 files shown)


projects/packages/connection/legacy/class-jetpack-signature.php
```diff
@@ -94,7 +94,7 @@ public function sign_current_request( $override = array() ) {
 			// and encoded on the Jetpack side.
 			if ( defined( 'IS_WPCOM' ) && IS_WPCOM ) {
 				// phpcs:ignore WordPress.Security.NonceVerification.Missing
-				if ( empty( $body ) && is_array( $_POST ) && count( $_POST ) > 0 ) {
+				if ( empty( $body ) && is_array( $_POST ) && $_POST !== array() ) {
```
jeherve, line 97: Quick question about this change, and similar ones in other PRs: here wouldn't it be more readable to check if the array if not empty?

```suggestion
				if ( empty( $body ) && is_array( $_POST ) && ! empty( $_POST ) ) {
```

nelsonec87: Hi @jeherve!
`rector` rules make these suggestions because of performance reasons: https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#countarraytoemptyarraycomparisonrectorr

nelsonec87:
Rebased because `PR is up to date` check was failing

markbiek (approved):
LGTM!

PR #30709, opened 2023-05-14
E2E tests: Add WPCOM REST API request count as metric to each test

fgiannar:
This PR adds a mechanism for tracking the total number of outgoing requests to WPCOM REST API during each of the E2E tests and sending it to Allure for reporting.

## Proposed changes:

* Adds and activates a new plugin, `e2e-wpcom-request-interceptor`: The purpose of this plugin is to track the number of total requests to the WPCOM REST API and store them in a transient. This transient is reset on every new test and added to the corresponding E2E reports. This is particularly useful in order to identify PRs that might introduce an excessive amount of requests to WPCOM. ( Updated docs as well)
* Resets the aforementioned transient before each test runs
* Sends the count of WPCOM REST API requests to Allure (reporting) after each test runs

### Other information:

## Jetpack product discussion

p9dueE-7dr-p2

No

## Testing instructions:

- Make sure E2E tests don't break
- Current branch only runs pre-connection E2E tests so the corresponding parameter, `Requests to WPCOM API` will be `0`
- Try running locally all the E2E tests and make sure `Requests to WPCOM API` are properly populated. You can enable sandboxing and verify that the reported number of WPCOM API requests, logged in `debug.log` matches the Allure parameter

Screenshot 2023-05-15 at 09 34 31


jeherve (commented):
Let's give it a try! 🚢

sergeymitr (approved):
👍

PR #30712, opened 2023-05-15
Woo Analytics: avoid fatals with third-party plugins

### LINKED ISSUE #11301: WooCommerce Analytics: better fallback when conflicting with plugins changing product data

1. Some plugins change the data returned by `wc_get_product`. 
2. It then causes Fatal Errors on our end, like so:

```
PHP Fatal error:  Uncaught Error: Call to undefined method WC_Product_Package::get_title() in /plugins/jetpack/7.0/modules/woocommerce-analytics/classes/wp-woocommerce-analytics-universal.php:172
Stack trace:
#0 /plugins/jetpack/7.0/modules/woocommerce-analytics/classes/wp-woocommerce-analytics-universal.php(90): Jetpack_WooCommerce_Analytics_Universal->get_product_details(Object(WC_Product_Package))
#1 /wp-includes/class-wp-hook.php(286): Jetpack_WooCommerce_Analytics_Universal->loop_session_events('')
```

We should aim to have a better fallback than this fatal in `get_product_details`.

jeherve:
Fixes #11301

## Proposed changes:

Some plugins make changes to Products and Orders, overwriting methods that we call in Jetpack. Let's check for such methods before we use them on our end, to avoid issues.

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

* Start with a site that's connected to WordPress.com, using a block theme like Twenty Twenty Three, and where you have installed the WooCommerce plugin.
* Install this plugin: https://wordpress.org/plugins/woo-gutenberg-products-block/
* Add the WooCommerce Single Product block to one of your templates in the site editor.

I'm not clear on how to reproduce the exact error after this setup. @katiebethbrown Maybe that's something you could tell us more about, or maybe you could test that patch?

Thank you!


projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-universal.php
```diff
@@ -362,6 +366,14 @@ function ( $payment_gateway ) {
 	public function order_process( $order_id ) {
 		$order = wc_get_order( $order_id );
 
+		if (
+			! $order
+			|| ! $order instanceof WC_Order
+			|| ! $order instanceof WC_Order_Refund
```
samiff: How did you figure out about checking for `instanceof WC_Order_Refund` here? Below I see `$order->get_payment_method();`, but I don't see a `get_payment_method()` method for the `WC_Order_Refund` class unless I'm missing something 🤔 

Otherwise, I wasn't able to reproduce the error even trying a dev build of woocommerce-blocks unfortunately, but the PR changes make sense and are consistent with existing checking in the file.

jeherve: I based this off the possible return values for `wc_get_order`:
https://github.com/woocommerce/woocommerce/blob/cf87fab3c2feb2b116c120107e0083c2a1a0d41b/plugins/woocommerce/includes/wc-order-functions.php#L79

I didn't dig deeper though, and you're right, that method doesn't exist there, so I'll remove it. 👍
1b7f28a8d7f5961f00a86219e24658abef71c4dd

samiff (approved):

PR #30713, opened 2023-05-15
WP Super Cache: use constants in the email dropdown

donnchawp:
This is a tiny PR to change the preload email dropdown so it uses a newly defined constant rather than hard coding the number of emails.

## Proposed changes:
The dropdown used to display 100 as the number of emails but we just introduced the constant WPSC_PRELOAD_POST_COUNT so this PR changes that number to that constant. There won't be any visible changes except 100 will be 10 in the dropdown that appears on the Preload page now.

### Other information:

No

## Testing instructions:
Apply PR
Go to the Preload page of the WP Super Cache settings.
Note the email dropdown says 10 emails rather than 100.

### DIFF
--- a/projects/plugins/super-cache/changelog/update-wp_super_cache_email_dropdown
+++ b/projects/plugins/super-cache/changelog/update-wp_super_cache_email_dropdown
@@ -1,0 +1,4 @@
+Significance: minor
+Type: fixed
+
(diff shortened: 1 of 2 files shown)


thingalon (approved):
Looks good to me :)

PR #30714, opened 2023-05-15
PHP8 compatibility updates

markbiek:
## Proposed changes:

### Other information:

## Testing instructions:

### DIFF
--- a/projects/packages/connection/src/class-error-handler.php
+++ b/projects/packages/connection/src/class-error-handler.php
@@ -278,7 +278,7 @@ class Error_Handler {
 
 		// Let's store a maximum of 5 different user ids for each error code.
 		$error_code_count = is_countable( $stored_errors[ $error_code ] ) ? count( $stored_errors[ $error_code ] ) : 0;
-		if ( count( $error_code_count ) > 5 ) {
+		if ( $error_code_count > 5 ) {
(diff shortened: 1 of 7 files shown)


projects/packages/connection/src/class-error-handler.php
```diff
@@ -277,7 +277,8 @@ public function store_error( \WP_Error $error ) {
 		$stored_errors[ $error_code ][ $user_id ] = $error_array;
 
 		// Let's store a maximum of 5 different user ids for each error code.
-		if ( count( $stored_errors[ $error_code ] ) > 5 ) {
+		$error_code_count = is_countable( $stored_errors[ $error_code ] ) ? count( $stored_errors[ $error_code ] ) : 0;
+		if ( count( $error_code_count ) > 5 ) {
```
nelsonec87: Wouldn't it be `if( $error_code_count > 5 ) { ` ?

nelsonec87 (changes requested):

projects/packages/connection/src/class-error-handler.php
markbiek: Yes, it would. Good catch! 🤦

nelsonec87 (dismissed):
LGTM

nelsonec87 (approved):
LGTM

PR #30715, opened 2023-05-15
PHP8 compatibility updates

markbiek:
## Proposed changes:

### Other information:

## Testing instructions:

### DIFF
--- a/projects/packages/videopress/src/class-stats.php
+++ b/projects/packages/videopress/src/class-stats.php
@@ -69,7 +69,7 @@ class Stats {
 
 		$data = $response['data'];
 
-		if ( ! $data || ! isset( $data['days'] ) || count( $data['days'] ) === 0 ) {
+		if ( ! $data || ! isset( $data['days'] ) || ! is_countable( $data['days'] ) || count( $data['days'] ) === 0 ) {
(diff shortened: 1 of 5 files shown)


projects/packages/videopress/src/class-stats.php
```diff
@@ -69,7 +69,7 @@ public static function get_today_plays() {
 
 		$data = $response['data'];
 
-		if ( ! $data || ! isset( $data['days'] ) || count( $data['days'] ) === 0 ) {
+		if ( ! $data || ! isset( $data['days'] ) || ! is_countable( $data['days'] ) || count( $data['days'] ) === 0 ) {
```
nelsonec87, line 72: I never know when to replace (or not) `count( $var ) === 0` by `$var === array()`

nelsonec87 (dismissed):
LGTM

projects/packages/videopress/src/class-stats.php
markbiek: Yeah, I'm not sure about that either. Mostly, I've only been changing those when rector flags it.

nelsonec87 (approved):
LGTM

PR #30724, opened 2023-05-15
Social: Add image validation for Instagram connections

gmjuhasz:
Instagram connections will work a bit different compared to other Publicize connections. It needs an image to be uploaded, so if a user is on a free plan, the featured image will be used, if they have the Advanced plan the image will be picked from SIG, picked image, featured image, in this order.

To avoid bad requests the image needs to be validated, which is handled by this PR.

## Proposed changes:

* Added restrictions for instagram into our restrictions array
* Created a hook that validates the featured image for a given array of connections
* Added a notice and disabled instagram connection if there is no image, or if it's the featured image and it is invalid

CleanShot 2023-05-15 at 16 09 05 png

### Other information:

## Jetpack product discussion

1204479765121257-as-1204528424246673

no.
## Testing instructions:

* It's a bit hard to test this before we have the instagram connection. The best way is to change `const CONNECTIONS_NEED_MEDIA = [ 'instagram' ];` in `projects/js-packages/publicize-components/src/components/form/index.js` for facebook.
* Add a Facebook connection to your site

On a free plan:
* The connection should be turned off by default. Adding a featured image re-enables it.
* If the featured image if bigger than the limit - which is now FB-s limit and is 4MB, the connection should be disabled as before

On the Advanced plan:
* The connection should be on by default. Disabling SIG in the post, so that you have no featured image, no picked image either should result in the connection being disabled and the notice appears. Adding any of the three should re-enable the connection


projects/js-packages/publicize-components/src/hooks/use-media-restrictions/index.js
```diff
@@ -28,7 +28,7 @@ const reduceVideoLimits = ( prev, current ) => ( {
 const getVideoLimits = enabledConnections =>
 	enabledConnections
 		.map( connection =>
-			RESTRICTIONS[ connection.service_name ]
+			RESTRICTIONS[ connection.service_name ]?.video
 				? RESTRICTIONS[ connection.service_name ].video
 				: DEFAULT_RESTRICTIONS.video
```
manzoorwanijk: Since we have used optional chaining here, we can simplify it as a one-liner
```suggestion
			RESTRICTIONS[ connection.service_name ]?.video || DEFAULT_RESTRICTIONS.video
```

gmjuhasz: Good point!

projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -206,6 +224,25 @@ export default function PublicizeForm( {
...
+									),
+									{
+										br: <br />,
+										fixLink: (
+											<ExternalLink
+												href={ getRedirectUrl( 'jetpack-social-media-support-information' ) }
+											/>
+										),
+									}
+								) }
```
manzoorwanijk: To simplify things here a bit, we can use two separate strings here:
1. `__( 'You need a valid image in your post to share to Instagram.', 'jetpack' )`
2. `__( 'Learn more', 'jetpack' )`

It has a few benefits:
- We will not need to use `createInterpolateElement`
- `<br />` may not be necessary.
- We can style the items better without `<br />`
- Translation of `__( 'Learn more', 'jetpack' )` is already available and thus can be re-used.

```suggestion
								{ __( 'You need a valid image in your post to share to Instagram.', 'jetpack' ) }
								<br />
								<ExternalLink href={ getRedirectUrl( 'jetpack-social-media-support-information' ) }>
									{ __( 'Learn more', 'jetpack' ) }
								</ExternalLink>
```

manzoorwanijk (dismissed):
This looks good to me. Just have an optional suggestion about a minor code improvement.

projects/js-packages/publicize-components/src/components/form/index.js
gmjuhasz: Makes sense. Thanks!

manzoorwanijk (approved):

PR #30732, opened 2023-05-15
Removed unset variable

nelsonec87:
## Proposed changes:

* This PR removes the `$autostart` variable because it's always `true`, since `$attr` is never set.
* This change does not change the behavior of the code.
* More context here p1684153352762479-slack-CDLH4C1UZ

### Other information:

## Testing instructions:

Visual inspection should be fine since it's such a small change.


markbiek (approved):
LGTM!

PR #30741, opened 2023-05-15
AI Assistant: convert HTML to markdown in the request content

### LINKED ISSUE #30739: AI Assistant: do not send html in the content when requesting
When asking for content based on previous ones, the content as part of the payload could contain an HTML format. Let's convert them to markdown.

retrofox:
This PR converts the content provided by blocks (usually enriched via HTML) to markdown before being used to perform the request to get generated content. It fixes the issue that, sometimes, the response contains plain HTML.

**This PR introduces the new [turndown](https://github.com/mixmark-io/turndown) package dependency.**

Fixes https://github.com/Automattic/jetpack/issues/30739

## Proposed changes:

* AI Assistant: convert HTML to markdown in the request content

### Other information:

## Testing instructions:

* Go to the block editor
* Create some content using paragraph blocks with formated context (bold, italic, etc)
* Generate content based on the post (`Continue writing`, for instance)
* Confirm that the content in the prompt contains markdown format

before | after
-------|------
Screenshot 2023-05-15 at 15 35 19|Screenshot 2023-05-15 at 15 34 13


lhkowalski:
I checked the code and it seems to be working fine:

Screen Shot 2023-05-19 at 18 37 47

Given the way we get the content from the blocks, we are still ignoring lists and some rich content that could be converted to markdown and then sent to the completion. This may be a good thing for not supported content like images, but we could improve the way we get the content to be able to include lists and tables, for example.

I think it may be a good idea to have markdown as the standard language for output and **input**. We were using it for output but didn't used for input yet. Maybe when we do it, we can also specify on the prompt that the provided content may contain markdown markup. This could improve the understanding of the prompt content by the model.

lhkowalski (approved):
Code change looks good and the libray used for the convertion is pretty stable.

dhasilva (approved):
LGTM

PR #30743, opened 2023-05-15
VideoPress: Add original core/video attributes to extended attributes definition for v5

### LINKED ISSUE #30571: Adding Subtitles to a 'regular' video block causes Block Recovery
### Impacted plugin

Jetpack

### Quick summary

With https://jetpack.com/support/jetpack-blocks/ activated, I can no longer include subtitles

### Steps to reproduce

1. Create a post with a video (remote or local, doesn't matter)
2. Upload VTT file to WordPress
3. Edit the block to add text tracks 
4. Save the post
5. Refresh the editor

Video should have subtitles. Yay!

### What actually happened

Block Validation Fails :( 

Screenshot 2023-05-09 at 3 41 06 PM

```
Block validation: Block validation failed for `core/video` (Object).

Content generated by `save` function:

<figure class="wp-block-video"><video controls src="https://example.com/video.mp4"></video></figure>

Content retrieved from post body:

<figure class="wp-block-video"><video controls src="https://example.com/video.mp4"><track src="https://example.local/wp-content/uploads/2023/05/video.vtt" label="English" srclang="en" kind="subtitles"/></video></figure>
```

I reproduced this on a clean site without any plugin except Jetpack, as well as with a bunch more plugins, different themes, ec.

The only fix is to disable the blocks. :( 

### Impact

All

### Available workarounds?

_No response_

### Logs or notes

_No response_

dhasilva:
Fixes #30571

## Proposed changes:

* Adds the original `core/video` attributes (this issue particularly needs the `tracks` attribute) to the blocks attributes so the saved content matches the save function

### Other information:

## Testing instructions:

* Do NOT install Jetpack
* Go to the block editor and add a `core/video` block
* Upload subtitles to the videos tracks and save the post
* Hard refresh
* Check that the block is rendered as normal
* Install and activate Jetpack
* Go back to the post in the block editor
* Hard refresh
* Check that the block is rendered as normal, without invalid content
* Check that the subtitles work as expected

Before | After
-|-
 |


projects/plugins/jetpack/extensions/blocks/videopress/editor.js
```diff
@@ -292,6 +292,13 @@ const addVideoPressSupport = ( settings, name ) => {
 				},
 				default: [],
 			},
+			tracks: {
```
retrofox: we should map this attribute with [videoPressTracks](https://github.com/Automattic/jetpack/blob/99937fb86a9287312c5df88dcf9eb4f64184fdc4/projects/plugins/jetpack/extensions/blocks/videopress/editor.js#LL288C4-L288C20) instead of adding a new one.

dhasilva: The thing is, they are not related to VideoPress at all. They are listed as `tracks` originally, before the extension from `core/video`, and since `tracks` is not in the attributes list, the attribute is removed and we end with a `<track ...` tag without the attribute.

dhasilva:
```
		const attributesDefinition = {
			...settings.attributes,
```
also works, but there may be unintended side effects doing this?

dhasilva: The issue here is that the block is registered in general, irrespective of a particular block's data, but overwriting the attributes and removing some of the original ones. When a block does not use a VideoPress video, but a local one, this causes the rendering problem.

I updated the PR to add the original attributes before the new ones, so if an attribute is not used by v5 they are simply ignored, but normal videos do not break.

retrofox (approved):

PR #30744, opened 2023-05-15
Fix action backwards compatibility

CGastrell:
This PR is a fix for a breaking change introduced in #30191 

## Proposed changes:
- Rollback the hook parameter change, make it the fields collection instead of the form instance.
- Change the hook handler introduced on #30191 so it will try to get the form instance from the first field that can return it.

### Other information:

## Jetpack product discussion
p1684177869512529-slack-C02ER2E0Z

No

## Testing instructions:
JPCRM should continue to work as before.

Repeat the tests on #30191 to verify the new feature is still working as expected

### DIFF
--- a/projects/packages/forms/changelog/fix-action-backwards-compat
+++ b/projects/packages/forms/changelog/fix-action-backwards-compat
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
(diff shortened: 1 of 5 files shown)


tbradsha (approved):
The code looks good. I don't have Salesforce access to test for regressions as per steps in #30191, but I suspect we're good.

PR #30748, opened 2023-05-16
Social: Add failsafe for non-int error codes

gmjuhasz:
Fixes https://github.com/Automattic/jetpack/pull/30280#discussion_r1189041464

After introducing some new error codes there were some errors appearing, because of int-string casting issues. We hopefully fixed that in D110540-code, however this change adds a failsafe check, to fallback to the generic error code even if there is some issue with casting

## Proposed changes:

* Added a failsafe

### Other information:

## Jetpack product discussion

1204479765121257-as-1204599712012225

No.
## Testing instructions:

* Mark Twitter as unsupported with the 'publicize-twitter-down' feature flag
* Add a broken connection by removing FB-s permission.
* Load up the editor, the connections should reflect the broken and unsupported states. There should be no error either on jetpack side, nor on WPCOm side.


pablinos (approved):
This is working well for me. I passed a string back like before and all connections were broken, but there were no warnings.

PR #30750, opened 2023-05-16
Related Posts: Add support for author name in block editor

### LINKED ISSUE #30823: Related Posts: Add support for author name in block editor
Currently the only way to surface post authors in the Related Posts block is to use a PHP snippet as described in our [public-facing support page](https://jetpack.com/support/related-posts/customize-related-posts/#author). The problem with this is that it doesn't render correctly in the block editor.

Here's the author added to the legacy widget via PHP:

Image

Here's how the block-based version is rendered when that PHP snippet is enabled:

Image

My proposed solution here is to maintain the existing context for legacy rendering (in themes that do not support FSE) and to manage the author value separately on a per-block basis when using blocks.

See #30682 for work on updating the block context. For this issue, we'll be focused on getting the author toggle set up in the block editor, generating the string for display, and updating the rendered block in both front and backend.

a8ck3n:
Fixes #30823.

## Proposed changes:

1. Adds a toggle to the block editor to "Display author".
2. Generates author string during `render_block()` call.
3. Renders author in resulting block in both front/backend.

The resulting UI should look like this:

SCR-20230520-pvtc

### Other information:

## Jetpack product discussion

pejTkB-yT-p2

No

## Testing instructions:

1. Spin up a test site with this branch.
3. Make sure Related Posts are enabled via Jetpack → Settings → Traffic → Related Posts.
4. Create a new post and add a Related Posts block.
5. Confirm the "Display author" toggle is present and disabled by default.
6. Confirm enabling the toggle updates the block in the editor.
7. Confirm the author is shown on the front-facing page once published.

**Note:** You may need to publish before the placeholders are replaced with actual post content.


projects/plugins/jetpack/modules/related-posts/jetpack-related-posts.php
```diff
@@ -1669,6 +1680,22 @@ protected function filter_non_public_posts( array $related_posts ) {
...
+	 * Generates the author byline for the related post.
+	 *
+	 * @param int $post_id - the post ID.
+	 * @uses get_post_field, get_the_author_meta
+	 * @return string
+	 */
+	protected function generate_related_post_display_author( $post_id ) {
+		$post_author         = get_post_field( 'post_author', $post_id );
+		$author_display_name = get_the_author_meta( 'display_name', $post_author );
+		if ( isset( $author_display_name ) && ! empty( $author_display_name ) ) {
```
kangzj:
```suggestion
		if ( ! empty( $author_display_name ) ) {
```
I guess `$author_display_name` is always set by the point 😄

projects/plugins/jetpack/modules/related-posts/jetpack-related-posts.php
```diff
@@ -439,6 +448,7 @@ public function render_block( $attributes ) {
 		$block_attributes   = array(
 			'headline'        => isset( $attributes['headline'] ) ? $attributes['headline'] : null,
 			'show_thumbnails' => isset( $attributes['displayThumbnails'] ) && $attributes['displayThumbnails'],
+			'show_author'     => isset( $attributes['displayAuthor'] ) ? (bool) $attributes['displayAuthor'] : true,
```
kangzj: It seems author is not shown in the current stable version, I would prefer the behaviour to be consistent with the old block - i.e. set default to `false` 😄 . Please ignore this comment if this is intended.

a8ck3n: Thanks for flagging this! I hadn't realized but the `attributes` value we are passed in the `render_block()` callback only contains settings that do not match the defaults. I must have changed the default to false in the block setup and missed updating it here, which caused things to get out of sync. Now that both the JS and PHP default to false, the correct values are set up for rendering.

kangzj (dismissed):
It works as described. Well done 👍

kangzj (dismissed):

jeherve (dismissed):

jeherve (approved):

PR #30752, opened 2023-05-16
CRM: Company profile - make sure object prefills include the company name

coder-karen:
## Proposed changes:

* This PR changes the prefill name on the company 'view.page.php' for invoices, transactions and tasks to match what is expected: `zbsprefillco`. It also fixes the company tasks UI function - `zeroBSCRM_task_ui_for_co` - to add a case that checks for the company prefill (as the previous case isn't working via the company view page 'add tasks' links).

### Other information:

## Jetpack product discussion

https://github.com/Automattic/zero-bs-crm/issues/3085

No.

## Testing instructions:
* Using the [Jetpack Beta plugin](https://github.com/Automattic/jetpack-beta) on a test site with this branch applied or testing locally with the branch checked out, view the Company page for any company: `http://localhost/wp-admin/admin.php?page=zbs-add-edit&action=view&zbstype=company&zbsid={companyid}`.
* Underneath the 'Documents' heading, click on any of the tabs and the relevant links to 'add new', for example 'Add Invoice', 'Add Transaction' or 'Add Task' if you already have invoices, transactions or tasks linked to that company, or else click on the 'create one' link that is shown (eg.  'This Company does not have any tasks yet. Do you want to create one?')
* All new invoice, transaction or tasks pages should be prefilled with the company name.
* On trunk or the previous release version if testing on a Jurassic Ninja / test site, those links on the 'View Company' page will not prefill the company name.

### DIFF
--- a/projects/plugins/crm/admin/company/view.page.php
+++ b/projects/plugins/crm/admin/company/view.page.php
@@ -592,7 +592,7 @@ item"><?php esc_html_e( 'Tasks', 'zero-bs-crm' ); ?></div><?php } ?>
 								<tbody>
 								<?php
 								// prep link to create a new invoice
-								$new_invoice_url = jpcrm_esc_link( 'create', -1, ZBS_TYPE_INVOICE ) . '&prefillco=' . $company['id'];
+								$new_invoice_url = jpcrm_esc_link( 'create', -1, ZBS_TYPE_INVOICE ) . '&zbsprefillco=' . $company['id'];
 
 								if ( $company_invoice_count_inc_deleted > 0 ) {
 
@@ -708,7 +708,7 @@ item"><?php esc_html_e( 'Tasks', 'zero-bs-crm' ); ?></div><?php } ?>
 									<?php
 
 									// prep link to create a new transaction
-									$new_transaction_url = jpcrm_esc_link( 'create', -1, ZBS_TYPE_TRANSACTION ) . '&prefillco=' . $company['id'];
+									$new_transaction_url = jpcrm_esc_link( 'create', -1, ZBS_TYPE_TRANSACTION ) . '&zbsprefillco=' . $company['id'];
 
 									if ( count( $company['transactions'] ) > 0 ) {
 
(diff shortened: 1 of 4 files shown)


tbradsha (commented):
This works, but I'm a bit unclear as to why we added a nonce to the task prefill?

coder-karen:
> This works, but I'm a bit unclear as to why we added a nonce to the task prefill?

Double checking the usage of nonces (correct me if I'm missing something here) I'd say it's because we're taking the prefill from the URL so want to make sure the URL is genuine (is the one specified in the current request), to avoid anything unexpected being passed in.

`$co_id` will cast anything added to an int so will otherwise just be 0 from what I can see, so even though we're currently ok with the data, if we later did anything else with it we could be using a GET variable which includes a JS script. Though it can at a more basic level just result in the wrong company being auto-filled without the nonce. The nonce is verifying that that the URL given is expected, basically. 

However at the moment if the nonce is wrong it just doesn't add the company name (or contact name), which I think is fine. 

What is your understanding on this?

coder-karen:
After another conversation about nonces in this use case I've removed them here 😄 - The main reason being that when weighing up what issues they could prevent (very limited) vs what issues they could cause (could we create a URL to add tasks elsewhere and cause issues due to not requiring the nonce for example), then it seemed to make sense to remove them.

I did add in a check though to remove the customer / company ID from the hidden input field if it's not a valid ID.

tbradsha:
> Double checking the usage of nonces (correct me if I'm missing something here) I'd say it's because we're taking the prefill from the URL so want to make sure the URL is genuine (is the one specified in the current request), to avoid anything unexpected being passed in.

Issues with potential XSS should be handled with proper input sanitisation and output escaping rather than nonces.

More specifically, nonces are used to prevent cross-site request forgeries (CSRF). As there aren't any changes being made to the server, and it's simply retrieving data (just like an API or anything else), accessing a prefill URL without a nonce is a legitimate use case.

tbradsha (approved):
Saving a task assigned to an invalid contact ID seems to handle mostly properly even in `trunk`. I did see that if a fake contact/company ID is passed in `trunk`, there's a `view` button provided, so I'm assuming that's what prompted this last change.

I did have some concerns with us checking the name before we assign the task:

* Some contacts don't have names, but it looks like `getContactNameWithFallback()` creates a fallback name.
* Companies created through the edit page require a name.

There is an edge case where if companies are created without a name (e.g. via API or something) they'll not prefill, but I think that's okay.

Approved.

PR #30755, opened 2023-05-16
Akismet: include current gmt time in submission

### LINKED ISSUE #30753: Sending a value for comment_date_gmt when submitting feedbacks to Akismet (checks and reports)
When the contact form module interacts with the Akismet API, it would be really helpful if a stable date could be sent as `comment_date_gmt` (ISO 8601). It's important for that date to remain consistent across the original check, and any reports (missed spam or false positive) that might get submitted.

EG: if a feedback is submitted on April 18th at 3:23pm, and gets reported days after, the date we get in the report should always be the original April 18th at 3:23pm.

This would help us (Akismet devs) correlate a specific check and its related reports, when we have to investigate issues.

Thanks for considering it. 🙂

kraftbj:
Fixes #30753

## Proposed changes:

* Akismet can benefit from having the current datetime in ISO 8601 format.

### Other information:

## Jetpack product discussion

n/a

It does add the date to the submission, but this isn't really new information per se.

## Testing instructions:

* Confirm that Akismet sees the new field when submitting a contact form entry using this patch.

### DIFF
--- a/projects/packages/forms/changelog/add-akismet-date-30753
+++ b/projects/packages/forms/changelog/add-akismet-date-30753
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Akismet: include current gmt time to assist in spam detection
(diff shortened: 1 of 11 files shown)


kbrown9:
This tests well for me! The date is included in the submissions to Akismet, including the original check and missed spam/false positive reports.

kbrown9 (approved):

PR #30756, opened 2023-05-16
packages: Add dev keywords for Composer

anomiex:
## Proposed changes:

Modern versions of Composer will automatically suggest passing `--dev` to `composer require` if certain keywords ("dev", "testing", or "static analysis") are set on the package. Let's set those keywords.

And then we may as well set additional keywords on them too since we're doing it.

### Other information:

## Jetpack product discussion

None

No

## Testing instructions:

* Proofread, see if the keywords chosen make sense.
* Not sure the `--dev` recommending can be tested without this being merged and released. 🤷


anomiex:
Although... within the monorepo we still need to have jetpack-autoloader and jetpack-composer-plugin as non-dev so the build can continue to use `--no-dev` and still get the autoloader and jetpack_vendor stuff. Sigh.

jeherve (commented):
I think it makes sense for changelogger and codesniffer, but I don't know if I would suggest `dev` for the autoloader or `composer-plugin`, since those need to be included in production builds for consumer plugins as far as I know.

anomiex:
> but I don't know if I would suggest `dev` for the autoloader or `composer-plugin`, since those need to be included in production builds for consumer plugins as far as I know.

It's a little complicated. We don't actually need the autoloader package itself (`vendor/automattic/jetpack-autoloader/`) in production builds, just the results of it (`vendor/jetpack-autoloader/` and `vendor/composer/jetpack_autoload_*`). Same goes for composer-plugin.

But to get those effects the package has to be installed, so it won't work to have them in `require-dev` while doing `--no-dev` to try to prepare for bundling composer deps. So yeah, it might be better to remove the "dev" keyword for those.

jeherve (approved):

PR #30764, opened 2023-05-16
Forms: Fix hash generation

jcheringer:
## Proposed changes:
* Attempt to fix Forms hash generation causing an issue reported here: p1683922710395059-slack-C0299DMPG

### Other information:

## Jetpack product discussion
N/A

No

## Testing instructions:

* Create a post a add a Form block
* Wrap some fields in the Group block and use the Sidebar to change the Group's block spacing
* Publish the page, go to the published page, fill out the form, and submit it
* You should the the after-submission form summary


jeherve (commented):
Do we want to continue to update the old module in `projects/plugins/jetpack/modules/contact-form/` as well?

CGastrell:
> Do we want to continue to update the old module in `projects/plugins/jetpack/modules/contact-form/` as well?

That is a good question. Initially I'd say "no", but it's also true this is the first release cycle the package is enabled by default and we're still waiting for any hiccups there. Ideally we'd just fix whatever happens, but until we're sure maybe it's safer to deploy to both.

Anyways, I need to review this one and, if everything goes well, we can port the changes back to the plugin. Sounds good @jcheringer ?

jeherve:
> That is a good question. Initially I'd say "no", but it's also true this is the first release cycle the package is enabled by default and we're still waiting for any hiccups there. Ideally we'd just fix whatever happens, but until we're sure maybe it's safer to deploy to both.

Indeed. As long as folks have the option to use the module (with the filter), I think we'll need to keep the 2 codebases in sync.

CGastrell:
I just pushed the same changes for the plugin.

CGastrell (dismissed):
Worked for me, but we should improve this on its own. We shouldn't rely on a hash to pinpoint the submitted form.

jcheringer:
@CGastrell, I had to fix the changelogger error and merge with `trunk`.
I'll need another 👍, please.

CGastrell (approved):
LGTM! :shipit:

PR #30767, opened 2023-05-17
WAF: avoid multisite errors when used by the Protect plugin

### LINKED ISSUE #30766: Jetpack Protect: PHP Fatal error occurs in multisite.
### Impacted plugin

Jetpack

### Quick summary

 If only Jetpack Protect is enabled on the network, the following error occurs.
 
 PHP Fatal error:  Uncaught Error: Undefined constant "Automattic\Jetpack\Waf\Brute_Force_Protection\JETPACK__PLUGIN_FILE" in /path/to/wp/wp-content/plugins/jetpack-protect/jetpack_vendor/automattic/jetpack-waf/src/class-brute-force-protection.php:279

### Steps to reproduce

1. Install jetpack protect.
2. Activate jectpack protect on network or site..
3. Go wp-admin/admin.php?page=jetpack-protect
4. Push start for free button.
5. reload.

_No response_

### What actually happened

_No response_

### Impact

Most (> 50%)

### Available workarounds?

_No response_

### Logs or notes

_No response_

jeherve:
Fixes #30766

## Proposed changes:

The WAF package can be used by both the Jetpack plugin and the Jetpack Protect plugin; we should handle both in the package.

> **Note**
> Ideally we should handle just about any plugin using the package, but I'm not sure looking for the plugin by going several directories up would be a good approach. I would be happy to have a second opinion on this.

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

* Start with a fresh multisite installation.
* Install the Jetpack Protect plugin.
* **do not** network activate the plugin.
* In a subsite of the network, go to the Plugins menu and activate the Protect plugin.
* Notice the error in your logs.
* Apply patch.
* The error should be gone, and the dashboard notice should mention Jetpack Protect instead of Jetpack.

### DIFF
--- a/projects/packages/waf/changelog/fix-error-constant-protect
+++ b/projects/packages/waf/changelog/fix-error-constant-protect
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Multisite: avoid errors when the package is used in the Protect plugin instead of the Jetpack plugin.
--- a/projects/packages/waf/src/class-brute-force-protection.php
+++ b/projects/packages/waf/src/class-brute-force-protection.php
@@ -276,7 +276,16 @@ class Brute_Force_Protection {
 				require_once ABSPATH . '/wp-admin/includes/plugin.php';
 			}
 
-			if ( ! is_plugin_active_for_network( plugin_basename( JETPACK__PLUGIN_FILE ) ) ) {
+			// This warning is only relevant if either Jetpack or Jetpack Protect is active.
+			if ( defined( 'JETPACK__PLUGIN_FILE' ) ) {
+				$plugin_root_file = JETPACK__PLUGIN_FILE;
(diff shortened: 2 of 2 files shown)


nateweller (approved):
> Ideally we should handle just about any plugin using the package, but I'm not sure looking for the plugin by going several directories up would be a good approach. I would be happy to have a second opinion on this.

The WAF package does already have some similar plugin-specific handling related to the Brute Force Protection feature's API ([here](https://github.com/Automattic/jetpack/blob/b3c71ef6a1ecd6ad5845b1e6dc2444649bbd141a/projects/packages/waf/src/class-brute-force-protection.php#L960-L972)), so I don't mind continuing this existing limitation in order to get this fixed up.

That said, we do have intentions to use the WAF package in additional plugins, so at that time we may want to introduce a way for the implementing plugins to provide configuration arguments to the WAF package.

All that to say, LGTM, and thank you for fixing this! :+1:

PR #30769, opened 2023-05-17
Social: Featured image check classic editor

gmjuhasz:
This PR adds the featured image check for the Classic editor to finish up #30724 which added it for Gutenberg. This version has the limitation to only check for the availability, not the size. We can find a way for that in a follow-up

## Proposed changes:

* Added new div and styling for the warning message
* Added logic that validates the featured image if needed on start and on image change.

### Other information:

## Jetpack product discussion

1204479765121257-as-1204618693786023

No.
## Testing instructions:

* Build the package with `jetpack build packages/publicize`.
* Activate the Classic editor, then open up the editor with a Facebook connection. You should see the message, and clicking edit should show the checkbox as unchecked an disabled.
* Add a featured image. Clicking edit should show the checkbox as enabled and checked.
* Clear the image. It should get disabled again.
* Change back the action in the first step. Build. Now the validation should not run.

CleanShot 2023-05-17 at 11 14 25 png

### DIFF
--- a/projects/packages/publicize/composer.json
+++ b/projects/packages/publicize/composer.json
@@ -39,6 +39,10 @@
 		"build-development": [
 			"pnpm run build"
 		],
+		"watch": [
+			"Composer\\Config::disableProcessTimeout",
+			"pnpm run watch"
+		],
 		"build-production": [
 			"pnpm run build-production-concurrently"
 		]
--- a/projects/packages/publicize/src/class-publicize-ui.php
+++ b/projects/packages/publicize/src/class-publicize-ui.php
@@ -342,14 +342,9 @@ jQuery( function($) {
 	box-shadow: 0 1px 3px 0 rgba(0, 0, 0, 0.1);
 }
 .publicize__notice-media-warning {
-	display: block;
-	padding: 1px 12px;
-	margin: 5px 0;
-	border: 1px solid #c3c4c7;
-	border-left-width: 4px;
-	border-left-color: #dba617;
-	border-left-style: solid;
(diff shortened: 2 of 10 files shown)


projects/packages/publicize/src/class-publicize-ui.php
```diff
@@ -341,6 +341,16 @@ public function post_page_metabox_assets() {
...
+.publicize__notice-media-warning {
+	display: block;
+	padding: 1px 12px;
+	margin: 5px 0;
+	border: 1px solid #c3c4c7;
+	border-left-width: 4px;
+	border-left-color: #dba617;
+	border-left-style: solid;
+	box-shadow: 0 1px 1px rgba(0, 0, 0, 0.04);
+}
```
danielpost: You can get most of these styles for free by using the existing `.notice` class. Then, you just need to remove the margin and change the border color.

danielpost (commented):
Overall this works well! If I manually disable Instagram (Facebook for testing purposes), and then delete my featured image and add a new one, the connection gets enabled again. Is it possible to default it to the latest setting instead?

projects/packages/publicize/composer.json
```diff
@@ -39,6 +39,10 @@
 		"build-development": [
 			"pnpm run build"
 		],
+		"watch": [
```
pablinos, line 42: Doesn't have to be part of this PR, but I wanted to watch the package build.

spsiddarthan (dismissed):
This tests well!

spsiddarthan (approved):

PR #30777, opened 2023-05-17
Jetpack Social: Add notice to promote Instagram

danielpost:
## Proposed changes:

This PR adds a notice to the Jetpack Social admin page, as well as the Jetpack Social sidebar, to promote that Instagram is now available.

- On the admin page, the text as well as primary button change depending on whether or not the user has a paid plan.
- Dismissing one notice should prevent the other one from showing.

**Note:** We cannot merge this until we have implemented a proper feature check for Instagram.

### Other information:

## Jetpack product discussion

pdrWKz-XH-p2

No.

## Testing instructions:

* Create a new site with Jetpack Social.
* Open the Social admin page and verify that you don't see a notice.
* Add support for Instagram connections to your site: D111381-code
* Refresh and make sure there notice is there.

* Dismiss the notice, refresh the page and confirm that the notice is missing.
* Reset the notice: `jetpack docker wp option delete jetpack_social_dismissed_notices`
* Refresh and make sure it appears again.
* Buy the Advanced plan and make sure the notice has different text and CTA.

* Create a new post, go to the Jetpack Social sidebar.
* Check that the notice is there and dismissable.
* Make sure the notices match the designs.

Screenshot 2023-05-17 at 16 49 25@2x

Screenshot 2023-05-17 at 16 48 36@2x

### DIFF
--- a/projects/js-packages/publicize-components/src/components/form/index.js
+++ b/projects/js-packages/publicize-components/src/components/form/index.js
@@ -8,11 +8,12 @@
 
 import { getRedirectUrl } from '@automattic/jetpack-components';
 import { getSiteFragment } from '@automattic/jetpack-shared-extension-utils';
-import { PanelRow, Disabled, ExternalLink } from '@wordpress/components';
+import { Button, PanelRow, Disabled, ExternalLink } from '@wordpress/components';
 import { useDispatch, useSelect } from '@wordpress/data';
-import { Fragment, createInterpolateElement, useCallback } from '@wordpress/element';
+import { useState, Fragment, createInterpolateElement, useCallback } from '@wordpress/element';
 import { _n, sprintf, __ } from '@wordpress/i18n';
 import useAttachedMedia from '../../hooks/use-attached-media';
+import useDismissNotice from '../../hooks/use-dismiss-notice';
 import useImageGeneratorConfig from '../../hooks/use-image-generator-config';
 import useSocialMediaConnections from '../../hooks/use-social-media-connections';
 import useSocialMediaMessage from '../../hooks/use-social-media-message';
@@ -56,7 +57,17 @@ export default function PublicizeForm( {
 		useSocialMediaConnections();
 	const { message, updateMessage, maxLength } = useSocialMediaMessage();
 	const { isEnabled: isSocialImageGeneratorEnabledForPost } = useImageGeneratorConfig();
+	const { dismissedNotices, dismissNotice } = useDismissNotice();
 
+	const hasInstagramConnection =
+		connections.filter( connection => connection.service_name === 'instagram' ).length > 0;
+	const [ shouldShowInstagramNotice, setShouldShowInstagramNotice ] = useState(
+		! hasInstagramConnection
+	);
+	const onDismissInstagramNotice = useCallback( () => {
+		dismissNotice( 'instagram' );
+		setShouldShowInstagramNotice( false );
+	}, [ dismissNotice, setShouldShowInstagramNotice ] );
 	const shouldDisableMediaPicker =
 		isSocialImageGeneratorEnabled && isSocialImageGeneratorEnabledForPost;
 	const Wrapper = isPublicizeDisabledBySitePlan ? Disabled : Fragment;
@@ -236,9 +247,24 @@ export default function PublicizeForm( {
 						) }
 				</>
 			) }
-
 			{ ! isPublicizeDisabledBySitePlan && (
 				<Fragment>
+					{ ! dismissedNotices.includes( 'instagram' ) && shouldShowInstagramNotice && (
+						<Notice
+							onDismiss={ onDismissInstagramNotice }
+							type={ 'highlight' }
+							actions={ [
+								<Button href={ connectionsAdminUrl } variant="primary">
+									{ __( 'Connect now', 'jetpack' ) }
+								</Button>,
+								<Button href={ getRedirectUrl( 'jetpack-social-connecting-to-social-networks' ) }>
+									{ __( 'Learn more', 'jetpack' ) }
+								</Button>,
+							] }
+						>
+							{ __( 'You can now share directly to your Instagram account!', 'jetpack' ) }
+						</Notice>
+					) }
 					<PublicizeSettingsButton />
 
 					{ isPublicizeEnabled && connections.some( connection => connection.enabled ) && (
--- a/projects/js-packages/publicize-components/src/components/form/index.js
+++ b/projects/js-packages/publicize-components/src/components/form/index.js
@@ -26,6 +26,7 @@ import PublicizeSettingsButton from '../settings-button';
 import styles from './styles.module.scss';
 
 const CONNECTIONS_NEED_MEDIA = [ 'instagram' ];
+const PUBLICIZE_STORE_ID = 'jetpack/publicize';
 
 const checkConnectionCode = ( connection, code ) =>
 	false === connection.is_healthy && code === ( connection.error_code ?? 'broken' );
@@ -59,11 +60,15 @@ export default function PublicizeForm( {
 	const { isEnabled: isSocialImageGeneratorEnabledForPost } = useImageGeneratorConfig();
 	const { dismissedNotices, dismissNotice } = useDismissNotice();
 
+	const { isInstagramConnectionSupported } = useSelect( select => ( {
+		isInstagramConnectionSupported: select( PUBLICIZE_STORE_ID ).isInstagramConnectionSupported(),
+	} ) );
+
 	const hasInstagramConnection = connections.some(
 		connection => connection.service_name === 'instagram'
 	);
 	const [ shouldShowInstagramNotice, setShouldShowInstagramNotice ] = useState(
-		! hasInstagramConnection
+		! hasInstagramConnection && isInstagramConnectionSupported
 	);
 	const onDismissInstagramNotice = useCallback( () => {
 		dismissNotice( 'instagram' );
--- a/projects/js-packages/publicize-components/src/components/form/index.js
+++ b/projects/js-packages/publicize-components/src/components/form/index.js
@@ -59,8 +59,9 @@ export default function PublicizeForm( {
 	const { isEnabled: isSocialImageGeneratorEnabledForPost } = useImageGeneratorConfig();
 	const { dismissedNotices, dismissNotice } = useDismissNotice();
 
-	const hasInstagramConnection =
-		connections.filter( connection => connection.service_name === 'instagram' ).length > 0;
+	const hasInstagramConnection = connections.some(
+		connection => connection.service_name === 'instagram'
+	);
 	const [ shouldShowInstagramNotice, setShouldShowInstagramNotice ] = useState(
 		! hasInstagramConnection
 	);
--- a/projects/js-packages/publicize-components/src/components/form/index.js
+++ b/projects/js-packages/publicize-components/src/components/form/index.js
@@ -254,10 +254,13 @@ export default function PublicizeForm( {
 							onDismiss={ onDismissInstagramNotice }
 							type={ 'highlight' }
 							actions={ [
-								<Button href={ connectionsAdminUrl } variant="primary">
+								<Button key="connect" href={ connectionsAdminUrl } variant="primary">
 									{ __( 'Connect now', 'jetpack' ) }
 								</Button>,
-								<Button href={ getRedirectUrl( 'jetpack-social-connecting-to-social-networks' ) }>
+								<Button
+									key="learn-more"
+									href={ getRedirectUrl( 'jetpack-social-connecting-to-social-networks' ) }
+								>
 									{ __( 'Learn more', 'jetpack' ) }
 								</Button>,
 							] }
--- a/projects/js-packages/publicize-components/src/components/notice/styles.module.scss
+++ b/projects/js-packages/publicize-components/src/components/notice/styles.module.scss
@@ -1,13 +1,17 @@
 @import '@automattic/jetpack-base-styles/gutenberg-base-styles';
 
 .notice {
-	margin-bottom: 16px;
+	font-size: 13px;
+	line-height: 24px;
+	margin: 16px 0;
+	position: relative;
 }
 
 .notice--default {
 	color: $gray-700;
 }
 
+.notice--highlight,
 .notice--warning,
 .notice--error {
 	padding: 8px 16px;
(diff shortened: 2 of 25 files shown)


projects/js-packages/publicize-components/src/components/notice/styles.module.scss
```diff
@@ -26,3 +30,31 @@
...
+	all: unset;
+	cursor: pointer;
+	position: absolute;
+	top: 12px;
+	right: 16px;
+
+	&:focus:not( :disabled ) {
+		box-shadow: inset 0 0 0 1px var( --wp-components-color-background, #fff ),
+			0 0 0 var( --wp-admin-border-width-focus )
+				var( --wp-components-color-accent, var( --wp-admin-theme-color, #007cba ) );
```
danielpost, line 54: Prettier is the proud author of this piece of indentation art.

projects/js-packages/publicize-components/src/hooks/use-dismiss-notice/index.js
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Hook to handle retrieving dismissed notices and dismissing a notice.
+ *
+ * @returns {DismissNoticeHook} - An object with the dismissed notice hook properties set.
+ */
+export default function useDismissNotice() {
+	const dismissedNotices =
+		getJetpackData()?.social?.dismissedNotices ??
+		window?.jetpackSocialInitialState?.jetpackSettings?.dismissedNotices ??
+		[];
```
danielpost, line 19: This fetches the dismissed notices from the initial state of either 1) Jetpack or 2) Jetpack Social. It feels messy—improvements welcome.

projects/packages/publicize/src/class-publicize-base.php
```diff
@@ -1776,6 +1783,36 @@ public function has_paid_plan( $refresh_from_wpcom = false ) {
...
+	 * Check if we have a Jetpack Social Advanced plan.
+	 *
+	 * @param bool $refresh_from_wpcom Whether to force refresh the plan check.
+	 *
+	 * @return bool True if we have an Advanced plan, false otherwise.
+	 */
+	public function has_advanced_plan( $refresh_from_wpcom = false ) {
+		static $has_advanced_plan = null;
+		if ( ! $has_advanced_plan ) {
+			$has_advanced_plan = Current_Plan::supports( 'social-image-generator', $refresh_from_wpcom );
```
danielpost: Is there a better way to check the Advanced plan that I should be using instead?

projects/packages/publicize/src/class-rest-controller.php
```diff
@@ -204,6 +240,27 @@ public static function get_social_product_info() {
...
 	}
 
+	/**
+	 * Dismisses a notice to prevent it from appearing again.
+	 *
+	 * @param WP_Request $request The request object, which includes the parameters.
+	 * @return WP_REST_Response|WP_Error True if the request was successful, or a WP_Error otherwise.
+	 */
+	public function update_dismissed_notices( $request ) {
+		$notice            = $request->get_param( 'notice' );
```
danielpost, line 250: We can safely assume that `$notice` is set to a valid value (only `instagram` right now), because the schema doesn't allow for anything else.

projects/plugins/social/src/class-rest-settings-controller.php
```diff
@@ -135,13 +135,19 @@ public function update_item( $request ) {
 					}
 					break;
 				case 'show_pricing_page':
-					if ( ! update_option( Jetpack_Social::JETPACK_SOCIAL_SHOW_PRICING_PAGE_OPTION, (int) $params[ $name ] ) ) {
```
danielpost, line 138: `update_option` returns `false` if the option wasn't updated (e.g. if you're setting it to `true` but it's already set to `true`), so we need to check the old value first before updating.

(or, alternatively, just get rid of the `WP_Error` here?)

projects/plugins/social/src/js/components/admin-page/index.jsx
```diff
@@ -16,12 +17,23 @@ import SupportSection from '../support-section';
...
 import Header from './../header';
 import InfoSection from './../info-section';
+import InstagramNotice from './../instagram-notice';
 import AdminPageHeader from './header';
 import './styles.module.scss';
 
 const Admin = () => {
 	const { isUserConnected, isRegistered } = useConnection();
 	const showConnectionCard = ! isRegistered || ! isUserConnected;
+	const [ manuallyShowPricingPage, setManuallyShowPricingPage ] = useState( false );
```
danielpost: Naming things is hard!

gmjuhasz:
I think Videos/reels won't be supported yet. Should we take that out from the text in the meantime? Or say "soon". What do you think @spsiddarthan ?

projects/packages/publicize/src/class-rest-controller.php
```diff
@@ -204,6 +240,27 @@ public static function get_social_product_info() {
...
+		$dismissed_notices = get_option( Publicize::OPTION_JETPACK_SOCIAL_DISMISSED_NOTICES );
+
+		if ( ! is_array( $dismissed_notices ) ) {
+			$dismissed_notices = array();
+		}
+
+		$dismissed_notices = array_unique( array_merge( $dismissed_notices, array( $notice ) ) );
```
gmjuhasz: Could we put a comment on this line to explain why is this needed?

projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -56,7 +57,17 @@ export default function PublicizeForm( {
 		useSocialMediaConnections();
 	const { message, updateMessage, maxLength } = useSocialMediaMessage();
 	const { isEnabled: isSocialImageGeneratorEnabledForPost } = useImageGeneratorConfig();
+	const { dismissedNotices, dismissNotice } = useDismissNotice();
 
+	const hasInstagramConnection =
+		connections.filter( connection => connection.service_name === 'instagram' ).length > 0;
```
gmjuhasz: We could use [some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) here.

```
connections.some( connection => connection.service_name === 'instagram' )
```

gmjuhasz (commented):
I left a few comments, will give a proper test tomorrow

gmjuhasz (commented):
I left a few comments, will give a proper test tomorrow. 

Also, should this be behind a feature flag or are we planning to finish up everything?

gmjuhasz:
I have an issue on a brand new JN site, where I cannot dismiss the pricing table

https://github.com/Automattic/jetpack/assets/36671565/5ddd93cc-9605-4df7-b114-7df838985c02

projects/js-packages/publicize-components/src/components/form/index.js
danielpost: Good idea, done!

projects/packages/publicize/src/class-rest-controller.php
danielpost: I simplified the code a little bit instead, it should be more clear what's happening now.

projects/plugins/social/src/js/components/admin-page/index.jsx
manzoorwanijk: May be `forceDisplayPricingPage`?

projects/plugins/social/src/js/components/admin-page/index.jsx
```diff
@@ -16,12 +17,23 @@ import SupportSection from '../support-section';
...
+
+	const onUpgradeToggle = useCallback(
+		() => setManuallyShowPricingPage( true ),
+		[ setManuallyShowPricingPage ]
+	);
+	const onPricingPageDismiss = useCallback(
+		() => setManuallyShowPricingPage( false ),
+		[ setManuallyShowPricingPage ]
+	);
```
manzoorwanijk: The reference of the state setter from `useState` never changes, so those can be omitted from the dependencies.

projects/plugins/social/src/js/components/admin-page/index.jsx
danielpost: Good idea!

projects/plugins/social/src/js/components/admin-page/index.jsx
danielpost: Ah good to know, fixed!

gmjuhasz (dismissed):
This runs well for me. @pablinos want to give it a test too as its a bigger change?

projects/packages/publicize/src/class-publicize-base.php
pablinos: I think we could use the instagram plan feature now couldn't we?

projects/js-packages/publicize-components/src/hooks/use-dismiss-notice/index.js
pablinos: For some reason I thought we rendered `jetpackSocialInitialState` in the editor too. If so, then perhaps we could just rely on that?

I don't think it's much better, but we could break it up a bit:

```suggestion
	const stateObj                = getJetpackData()?.social ?? window?.jetpackSocialInitialState?.jetpackSettings
	const dismissedNotices = stateObj?.dismissedNotices ?? [];
```

projects/packages/publicize/src/class-publicize-base.php
```diff
@@ -1783,6 +1799,36 @@ public function has_paid_plan( $refresh_from_wpcom = false ) {
...
+	/**
+	 * Check if we have a Jetpack Social Advanced plan.
+	 *
+	 * @param bool $refresh_from_wpcom Whether to force refresh the plan check.
+	 *
+	 * @return bool True if we have an Advanced plan, false otherwise.
+	 */
+	public function has_advanced_plan( $refresh_from_wpcom = false ) {
+		static $has_advanced_plan = null;
+		if ( ! $has_advanced_plan ) {
```
pablinos: I think we'd need to check against `null` as otherwise `false` would keep triggering the plan check.

```suggestion
		if (  null === $has_advanced_plan ) {
```

Should we also check for `refresh_from_wpcom` being true? Otherwise it will always return the cached value.

projects/plugins/social/src/js/components/instagram-notice/index.jsx
```diff
@@ -0,0 +1,87 @@
...
+	'Share featured images directly to your Instagram Business account for free, or share unlimited photos and video reels with Jetpack Social Advanced.',
+	'jetpack-social'
+);
+const paidPlanNoticeText = __(
+	'Enjoy automatically sharing unlimited photos and video reels to Instagram Business with your Jetpack Social Advanced plan!',
+	'jetpack-social'
+);
+
+const InstagramNotice = ( { onUpgrade = () => {} } = {} ) => {
+	const [ showNotice, setShowNotice ] = useState( true );
```
pablinos: Could this be?

```suggestion
	const [ showNotice, setShowNotice ] = useState( dismissedNotices.includes( 'instagram' ) );
```

projects/plugins/social/src/js/components/instagram-notice/index.jsx
```diff
@@ -0,0 +1,87 @@
...
+import { __ } from '@wordpress/i18n';
+import { STORE_ID } from '../../store';
+import styles from './styles.module.scss';
+
+const freePlanNoticeText = __(
+	'Share featured images directly to your Instagram Business account for free, or share unlimited photos and video reels with Jetpack Social Advanced.',
+	'jetpack-social'
+);
+const paidPlanNoticeText = __(
+	'Enjoy automatically sharing unlimited photos and video reels to Instagram Business with your Jetpack Social Advanced plan!',
```
pablinos: I think it's already been mentioned but we won't support video to begin with

projects/packages/publicize/src/class-publicize-base.php
pablinos: I've just seen that this is for displaying different notices, but I think the feature check should be for the enhanced publishing feature, as that will allow you to select an image.

projects/plugins/jetpack/extensions/blocks/related-posts/test/fixtures/jetpack__related-posts.json
```diff
@@ -1,17 +1,17 @@
...
-            "displayDate": false,
-            "displayAuthor": false,
-            "displayThumbnails": true,
-            "displayContext": true,
-            "postsToShow": 4
-        },
-        "innerBlocks": [],
-        "originalContent": ""
-    }
+	{
```
pablinos: I'm not sure how these changes have got in here. Did I mess up the trunk merge?

pablinos (commented):
It's pretty close, I think. Just a few things to check.

projects/js-packages/publicize-components/src/hooks/use-dismiss-notice/index.js
```diff
@@ -0,0 +1,33 @@
...
+ * @property {Function} dismissNotice - Callback used to dismiss a notice.
+ */
+
+/**
+ * Hook to handle retrieving dismissed notices and dismissing a notice.
+ *
+ * @returns {DismissNoticeHook} - An object with the dismissed notice hook properties set.
+ */
+export default function useDismissNotice() {
+	const dismissedNotices =
```
pablinos, line 16: Thinking about it some more, could we move the state into this hook? The handler could update the state before calling the API

projects/js-packages/publicize-components/src/hooks/use-dismiss-notice/index.js
```diff
@@ -0,0 +1,33 @@
...
+		getJetpackData()?.social?.dismissedNotices ??
+		window?.jetpackSocialInitialState?.jetpackSettings?.dismissedNotices ??
+		[];
+
+	const handleDismiss = notice => {
```
pablinos, line 21: I'm never sure of the performance hit/gains but we could move this outside of the hook, so it's not being recreated each render.

If we swap to having the state in here, we may want to wray it in a `useCallback`

pablinos (commented):
A couple more thoughts about the hook.

projects/plugins/social/src/class-rest-settings-controller.php
pablinos: This is madness! Why doesn't it return WP_Error or false for failure and an int for option values updated.

Maybe we should just assume it did update?

projects/plugins/social/src/js/components/instagram-notice/index.jsx
```diff
@@ -0,0 +1,82 @@
...
+} from '@automattic/jetpack-components';
+import { useDismissNotice } from '@automattic/jetpack-publicize-components';
+import { useSelect } from '@wordpress/data';
+import { useState, useCallback } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+import { STORE_ID } from '../../store';
+import styles from './styles.module.scss';
+
+const freePlanNoticeText = __(
+	'Share featured images directly to your Instagram Business account for free, or share unlimited photos and video reels with Jetpack Social Advanced.',
```
pablinos: We should probably get rid of the 'video reels' here. The feature is probably more about being able to pick an imag I think?

pablinos (commented):
Last tweak I think!

projects/plugins/social/src/js/components/instagram-notice/index.jsx
danielpost: Done!

projects/plugins/social/src/class-rest-settings-controller.php
danielpost: Yeah that's what I was considering as well. Done!

pablinos (approved):
Couple of follow up bits and bobs to do, but this is working well for me.

PR #30778, opened 2023-05-17
AI Assistant: Add treatment for error when prompt is not understood

### LINKED ISSUE #30651: AI Assistant: Catch errors on requests that are not understood
The user can type some senseless input like "sidjsadn" that is not undestood by the AI.
We need to flag those responses and treat them in a user-friendly way.

ref p1HpG7-mmO-p2#comment-63128

dhasilva:
Fixes #30651

## Proposed changes:

* Adds a rule so that the AI responds with a specific token when the prompt is not understood
* Adds an initial buffer to check for that error token

This PR is focused more in the mechanism, not much in the UI, as it is being defined and will be improved soon.

### Other information:

## Testing instructions:

* Add an AI Assistant block
* Ask some nonsense like "fdeswjfdaskfj"
* Check that a notice appears asking for a clear input
* Generate some valid content
* Ask some nonsense again
* Check that the notice appears and the previous content is not overwritten


projects/plugins/jetpack/extensions/blocks/ai-assistant/create-prompt.js
```diff
@@ -82,6 +82,7 @@ Rules:
 - Do not include a top level heading by default.
 - Only output generated content ready for publishing.
 - Segment the content into paragraphs as deemed suitable.
+- If you do not understand the request, please prefix your answer with __JETPACK_AI_ERROR__.
```
retrofox, line 85: :smart:

projects/plugins/jetpack/extensions/blocks/ai-assistant/use-suggestions-from-openai.js
```diff
@@ -12,7 +12,7 @@ import { buildPrompt } from './create-prompt';
 import { askJetpack, askQuestion } from './get-suggestion-with-stream';
 import { DEFAULT_PROMPT_TONE } from './tone-dropdown-control';
 
-const debug = debugFactory( 'jetpack:ai-assistant' );
+const debug = debugFactory( 'jetpack-ai-assistant' );
```
retrofox, line 15: I like `:` :-)

dhasilva: That is to make `jetpack-ai-assistant` and `jetpack-ai-assistant:prompt` visible at the same time when filtering by `jetpack-ai-assistant`. We can change it back to `:` and change the other one to `jetpack:ai-assistant:prompt`.

dhasilva: What do you think?

projects/plugins/jetpack/extensions/blocks/ai-assistant/use-suggestions-from-openai.js
```diff
@@ -218,12 +220,35 @@ const useSuggestionsFromOpenAI = ( {
...
 			if ( chunk ) {
-				fullMessage += chunk;
-				setAttributes( {
-					content: fullMessage,
-				} );
-				debug( fullMessage );
-				// debug( chunk );
+				// Check if the first chunks form the error marker by buffering them until we know for sure
+				if ( ! fullMessage ) {
+					initialBuffer += chunk;
```
retrofox: First feedback is `initialBuffer` can be defined in this scope if I'm not wrong. 

```suggestion
					let initialBuffer = chunk;
```

retrofox: The second comment is `fullMessage` enough to check the string in the content?

Also, I have to admit that the double-checking is confusing, but I think what you want to check is when the message starts or partially starts, with the `__JETPACK_AI_ERROR__` string, for instance:

```
__JETPACK_AI_ERROR__ ...
```

```
__JETPACK_AI_ER
```

is that correct?

dhasilva: Yes. The idea behind the buffer is to avoid changing the content to, say `__JETPACK_AI_ER` before finally getting the full error. That is what the `'__JETPACK_AI_ERROR__'.startsWith( initialBuffer )` checks for. It is the "it seems there will be an error, but not sure yet" check.

The inverse check, `initialBuffer.startsWith( '__JETPACK_AI_ERROR__' )` is because the received tokens can make the content go from `__JETPACK_AI_ERRO` to `__JETPACK_AI_ERROR__: N`, for instance, in a single event. If that happens we know the error is there and we can kill the connection and show the error.

projects/plugins/jetpack/extensions/blocks/ai-assistant/use-suggestions-from-openai.js
retrofox: Makes sense 💯

projects/plugins/jetpack/extensions/blocks/ai-assistant/use-suggestions-from-openai.js
dhasilva: Yes, `fullMessage` should be enough. Updating...

retrofox: This process also should be done on the server side.

retrofox (dismissed):
LGTM

image

retrofox (approved):

PR #30782, opened 2023-05-17
AI Assistant: implement JWT token storing

retrofox:
This PR adds a layer to the WJT token library. The idea is to avoid the requests.
It defines the expiration time to limit the lifetime of it.

Fixes #

## Proposed changes:

* AI Assistant: implement JWT token storing

### Other information:

## Testing instructions:

* Go to the block editor
* Create an AI Assistant block
* Request a suggestion
* Confirm, the first time, the token is generated:

image

image

* Perform other requests.
* Confirm that it uses the stored token, avoiding requesting a new one, if the request is made before to 2 minutes

image

* Confirm after 2 minutes, the app requests a new token

image


oskosk (approved):
Works great

PR #27949, opened 2022-12-14
Fix various PHP 8.2 issues

anomiex:
#### Changes proposed in this Pull Request:

Fix various issues that fail in tests when run with PHP 8.2. Skipped Jetpack for now, it has a lot of tests to potentially fix.

See individual commit comments for details.

#### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

#### Jetpack product discussion

#27927

#### Does this pull request change what data or activity we track or use?

No

#### Testing instructions:

* Is CI happy?
* Actually running this with PHP 8.2 would be a pain, since WordPress itself throws deprecation warnings for a bunch of tests.

Issue #27927, opened 2022-12-14
General: compatibility with PHP 8.2


jeherve:
**Note**: For now (December 14, 2022), we would recommend against running WordPress (and Jetpack) on PHP 8.2 since the Core WordPress team is still working on PHP 8.2 compatibility [here](https://core.trac.wordpress.org/ticket/56009).

For a better view of WordPress's PHP compatibility and support, you can refer to [this page](https://make.wordpress.org/core/handbook/references/php-compatibility-and-wordpress-versions/).

- [x] Actions: enable experimental PHP 8.2 run https://github.com/Automattic/jetpack/pull/27986
- [x] Fix deprecation warnings from PHP 8.2 https://github.com/Automattic/jetpack/pull/27949 https://github.com/Automattic/jetpack/pull/27968
- [x] #29167
- [x] #30786
- [x] #31296
- [x] #31438
- [x] #31640
- [x] #31909

---------

For reference: PHP 8.1 compatibility is tracked in #23695.

usman051:
Deprecated: Function Jetpack::is_development_mode is deprecated since version jetpack-8.0! Use \Automattic\Jetpack\Status->is_offline_mode instead. in

jeherve:
@usman051 It would appear that you use a third-party theme or plugin that relies on an old Jetpack function, that is now deprecated, the `is_development_mode` function. I would recommend that you search your site's codebase to find out where that function is defined, and then contact the theme/plugin author to let them know that they should update to use the newer function.

PR #23786, opened 2022-04-05
Use the Publicize composer package in Jetpack. 

spsiddarthan:
We are going to use the Publicize package [from the package's folder](https://github.com/Automattic/jetpack/tree/master/projects/packages/publicize) in Jetpack instead of the plugin's module files.  We are using this package already for Jetpack Social. The goal is to eventually get rid of the Publicize code within the Jetpack Plugin. 

For WPCOM, we are going to continue to use the Jetpack plugin's modules files for now.

However, once this PR is merged and the Publicize package is deployed to dotcom, we will start working on making the WPCOM code use the Publicize package as well. 

#### Changes proposed in this Pull Request:

* Use the config package to load the Publicize package files instead of requiring the Jetpack plugin's module files.

#### Does this pull request change what data or activity we track or use?
- No

#### Testing instructions:
On both dotcom and Jetpack, after checking out this PR.

- Publicize a post after adding a Publicize connection. 
- Republicize a post (make sure you have the jetpack plan Jetpack Security or above)

PR #30786, opened 2023-05-18
PHP Compatibility: fix dynamic property deprecation notices

jeherve:
## Proposed changes:

Follow-up to #27949. Fix more deprecation notices introduced with PHP 8.2:

```
PHP Deprecated:  Creation of dynamic property Automattic\Jetpack\Search\Customizer::$plan is deprecated in /wp-content/plugins/jetpack-dev/jetpack_vendor/automattic/jetpack-search/src/customizer/class-customizer.php on line 24
```

This was reported here:
https://wordpress.org/support/topic/dynamic-property/

### Other information:

## Jetpack product discussion

Primary issue: #27927

* No

## Testing instructions:

* Load this branch on a site running PHP 8.2, and a site running an old version of PHP.
* Ensure each feature touched by this PR still works.

### DIFF
--- a/projects/packages/publicize/changelog/fix-php82-dynamic-properties
+++ b/projects/packages/publicize/changelog/fix-php82-dynamic-properties
@@ -1,4 +1,0 @@
-Significance: patch
-Type: changed
-
-PHP Compatibility: fix dynamic property deprecation notices
--- a/projects/packages/publicize/changelog/fix-php82-dynamic-properties
+++ b/projects/packages/publicize/changelog/fix-php82-dynamic-properties
@@ -1,0 +1,4 @@
+Significance: patch
+Type: changed
+
+PHP Compatibility: fix dynamic property deprecation notices
--- a/projects/packages/publicize/src/class-publicize-ui.php
+++ b/projects/packages/publicize/src/class-publicize-ui.php
@@ -14,6 +14,14 @@ use Automattic\Jetpack\Assets;
  */
 class Publicize_UI {
 	/**
+	 * If Publicize is executing within Jetpack.
+	 *
+	 * @access public
+	 * @var bool
+	 */
+	public $in_jetpack;
+
+	/**
 	 * Contains an instance of class 'Publicize' which loads Keyring, sets up services, etc.
 	 *
 	 * @var Publicize Instance of Publicize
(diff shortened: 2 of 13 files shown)


projects/packages/publicize/src/class-publicize-ui.php
```diff
@@ -13,6 +13,14 @@
  * Only user facing pieces of Publicize are found here.
  */
 class Publicize_UI {
+	/**
+	 * If Publicize is executing within Jetpack.
+	 *
+	 * @access public
```
anomiex: I wonder whether we really need `@access public` when the definition itself already says `public`.

I could see something like `@access private` on some of these if we're saying public for historical reasons but don't really want anyone else to be accessing the property. But `@access public` seems redundant.

projects/packages/publicize/src/class-publicize-ui.php
```diff
@@ -13,6 +13,14 @@
...
+	 * @var bool
+	 */
+	public $in_jetpack;
```
anomiex: I wonder if this is actually needed. The one call I see setting this is at https://github.com/Automattic/jetpack/blob/9a37cb2ded4385c0294a26759e52be395771cdb0/projects/plugins/jetpack/modules/publicize.php#L108 but I don't see anything anywhere actually using the value anymore (the one historical use [appears to have been removed in #23786 a year ago](https://github.com/Automattic/jetpack/pull/23786/files#r849322516)).

projects/packages/publicize/src/class-publicize-ui.php
jeherve: That's a good point. Done in 4327a312e341b398f0e1649c52c90448e34ec773

projects/packages/publicize/src/class-publicize-ui.php
jeherve: That's a good point. I removed it in 599862fcada2f5319934bc7b7e6725d285c7ab50

projects/packages/publicize/changelog/fix-php82-dynamic-properties
anomiex, line 1: No change in this package anymore.

anomiex (approved):

PR #30790, opened 2023-05-18
AI Assistant: add upgrade notice to AI Assistant

### LINKED ISSUE #30736: AI Assistant: Add upgrade banner to block whenever user runs out of requests
Blocked by [#30726](https://github.com/Automattic/jetpack/issues/30726), we need to build the requests counter first. 

Once the user runs out of requests, add a way to let the user upgrade or buy more credits. 

As a first iteration, maybe we can use something similar to what we have in VideoPress, but with a link to the interstitial for Jetpack AI:

Screenshot 2023-05-16 at 11 36 57

nunyvega:
Fixes #30736

## Proposed changes:

* Add upgrade notice to the AI Assistant block. It won't be used yet.
It will be enabled once we have the monetization logic. 

Screenshot 2023-05-18 at 13 23 37

### Other information:

## Testing instructions:

* Apply to your local JP
* Enable it by removing the `false && `
* confirm that it is visible on your editor
* Confirm that it redirects to the Jetpack AI interstitial

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/upgrade-prompt.js
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/upgrade-prompt.js
@@ -1,6 +1,7 @@
 /*
  * External dependencies
  */
+import { isAtomicSite, isSimpleSite } from '@automattic/jetpack-shared-extension-utils';
 import { __ } from '@wordpress/i18n';
 import React from 'react';
 /*
@@ -9,11 +10,15 @@ import React from 'react';
 import { Nudge } from '../../shared/components/upgrade-nudge';
 
 const UpgradePrompt = () => {
+	if ( isAtomicSite() || isSimpleSite() ) {
+		return null;
+	}
+
 	return (
 		<Nudge
 			buttonText={ 'Upgrade' }
-			checkoutUrl={ 'admin.php?page=my-jetpack#/add-jetpack-ai' }
-			className={ 'jetpack-ai' }
+			checkoutUrl={ `${ window?.Jetpack_Editor_Initial_State?.adminUrl }admin.php?page=my-jetpack#/add-jetpack-ai` }
+			className={ 'jetpack-ai-upgrade-banner' }
 			description={ __( 'Free requests used. Upgrade now to keep using Jetpack AI.', 'jetpack' ) }
 			goToCheckoutPage={ () => {} }
 			isRedirecting={ false }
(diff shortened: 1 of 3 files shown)


projects/plugins/jetpack/extensions/blocks/ai-assistant/upgrade-prompt.js
```diff
@@ -0,0 +1,25 @@
...
+/*
+ * Internal dependencies
+ */
+import { Nudge } from '../../shared/components/upgrade-nudge';
+
+const UpgradePrompt = () => {
+	return (
+		<Nudge
+			buttonText={ 'Upgrade' }
+			checkoutUrl={ 'admin.php?page=my-jetpack#/add-jetpack-ai' }
```
retrofox: You can use the `Jetpack_Editor_Initial_State` global object here:

```suggestion
			checkoutUrl={ `${ Jetpack_Editor_Initial_State?.adminUrl } ?page=my-jetpack#/add-jetpack-ai` }
```

Something like this ^

projects/plugins/jetpack/extensions/blocks/ai-assistant/upgrade-prompt.js
```diff
@@ -0,0 +1,25 @@
...
+ */
+import { __ } from '@wordpress/i18n';
+import React from 'react';
+/*
+ * Internal dependencies
+ */
+import { Nudge } from '../../shared/components/upgrade-nudge';
+
+const UpgradePrompt = () => {
+	return (
```
retrofox: We should return `null` when it isn't a Jetpack site (at least for now)

```suggestion
const UpgradePrompt = () => {
	if ( isAtomicSite() || isSimpleSite() ) {
		return null;
	}
	
	return (
```

retrofox: You need to requite the functions in the `External dependencies` block, too.

```es6
import { isAtomicSite, isSimpleSite } from '@automattic/jetpack-shared-extension-utils';
```

projects/plugins/jetpack/extensions/blocks/ai-assistant/upgrade-prompt.js
oskosk: Addressed in 3481fba980f7cab15851c51cf56e401c587bc1d9

oskosk (approved):
LGTM!

oskosk:
LGTM but we need to revisit the styles

PR #30794, opened 2023-05-18
CLI: defend against errors when running composer --version

kraftbj:
When there is stdout from `composer --version` but it does not contain a version string, there is an uncaught fatal by our code attempting to use the first regex match when there is none.

This PR prevents that fatal and adds some error messaging to help self-resolution.

## Proposed changes:

* Check for valid regex matches.

### Other information:

## Jetpack product discussion

p1684416555993969-slack-CDLH4C1UZ

No.

## Testing instructions:

* Have a borked install of Composer.
* Run `pnpm jetpack cli link`

### DIFF
--- a/tools/cli/helpers/checkEnvironment.js
+++ b/tools/cli/helpers/checkEnvironment.js
@@ -1,5 +1,6 @@
 import child_process from 'child_process';
 import fs from 'fs';
+import process from 'node:process';
 import path from 'path';
 import { fileURLToPath } from 'url';
 import chalk from 'chalk';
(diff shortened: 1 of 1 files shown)


tools/cli/helpers/checkEnvironment.js
```diff
@@ -30,6 +31,19 @@ export async function getComposerVersion() {
...
+		console.log(
+			chalk.yellow(
+				`Composer version is not valid! This may cause issues when working with monorepo tooling. Please run "composer --version" to determine if there are any errors.`
+			),
+			chalk.yellow(
+				`Additionally, running "tools/check-development-environment.sh" may help identify other issues impacting your dev environment. The full output of "composer --version" is below:`
+			),
+			composerString
+		);
+		process.exit( 1 );
```
anomiex, line 45: Above we just return if composer isn't installed at all. I wonder if we should do the same here to let the caller handle the error situation?

lezama: it doesn't print the output of "composer --version" form me:

Screenshot 2023-05-18 at 13 15 45

PR #11280, opened 2019-02-05
Display notice for LinkedIn connection so they can reauthenticate

eliorivero:
```
======================================================
  _____          _   _       _   
 |  __ \        | \ | |     | |  
 | |  | | ___   |  \| | ___ | |_ 
 | |  | |/ _ \  | . ` |/ _ \| __|
 | |__| | (_) | | |\  | (_) | |_ 
 |_____/ \___/  |_| \_|\___/ \__|
 |  \/  |                        
 | \  / | ___ _ __ __ _  ___     
 | |\/| |/ _ \ '__/ _` |/ _ \    
 | |  | |  __/ | | (_| |  __/    
 |_|  |_|\___|_|  \__, |\___|    
                   __/ |         
                  |___/          
======================================================
```

Fixes #11145

Requires https://github.com/Automattic/wp-calypso/pull/30504 for changes in Publicize for Gutenberg in WP Admin in the Jetpack site

#### Changes proposed in this Pull Request:

In Classic editor

captura de pantalla 2019-02-06 a la s 16 57 08

Gutenberg in Calypso

captura de pantalla 2019-02-11 a la s 13 03 23

Gutenberg in WP Admin in JP site

captura de pantalla 2019-02-11 a la s 13 03 23

#### Testing instructions:

* Ensure you have an old LinkedIn connection
* Apply D23810-code to your sandbox and use this branch, remember to sandbox your Jetpack site
* Go to Publicize panel in post editing, both in Gutenberg and classic editor
* They should display a notice alerting user that the LinkedIn connection needs reauthentication
* It should display a link to Calypso Sharing settings

#### Proposed changelog entry for your changes:

* LinkedIn is deprecating its API v1 on March 1st so we've moved to API v2. We're showing a message to users that lets them know that they need to reauthenticate.

PR #30795, opened 2023-05-18
Publicize: avoid PHP warning

jeherve:
## Proposed changes:

This should fix the following warnings:

```
PHP Warning:  Undefined variable $item in /usr/local/src/jetpack-monorepo/projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/publicize-connection-test-results.php on line 125
PHP Warning:  Trying to access array offset on value of type null in /usr/local/src/jetpack-monorepo/projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/publicize-connection-test-results.php on line 125
```

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

Start with a site that runs Jetpack, connected to WordPress.com, and where you have activated the Publicize feature under Jetpack > Settings > Social

* Go to Posts > Add New
* Open the Jetpack sidebar
* Check your logs, you should see no warnings.

### DIFF
--- a/projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/publicize-connection-test-results.php
+++ b/projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/publicize-connection-test-results.php
@@ -122,7 +122,11 @@ class WPCOM_REST_API_V2_Endpoint_List_Publicize_Connection_Test_Results extends 
 			}
 		}
 
-		if ( 'linkedin' === $item['id'] && 'must_reauth' === $test_result['connectionTestPassed'] ) {
+		if (
+			isset( $item['id'] )
+			&& 'linkedin' === $item['id']
+			&& 'must_reauth' === $test_result['connectionTestPassed']
+		) {
 			$item['test_success'] = 'must_reauth';
 		}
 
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/publicize-connection-test-results.php
```diff
@@ -122,7 +122,11 @@ public function get_items( $request ) { // phpcs:ignore VariableAnalysis.CodeAna
 			}
 		}
 
-		if ( 'linkedin' === $item['id'] && 'must_reauth' === $test_result['connectionTestPassed'] ) {
+		if (
+			isset( $item['id'] )
```
anomiex, line 126: While this will fix the warning, the logic here still seems weird. Why is it looking at only the last element of `$items`? Should this check be inside the loop just above instead of after it?

anomiex: Looks like I'm not the first to ask that: https://github.com/Automattic/jetpack/pull/11280/files#r257089706

kraftbj (approved):
Approving as it does fix the warning. I agree that we should revisit this to improve the logic, but let's get the warning out of folks logs without delay.

PR #30800, opened 2023-05-18
Stats: add toggle for showing post views in the WordPress.com Reader

creativecoder:
## Proposed changes:

* Adds an option `wpcom_reader_views_enabled` for toggling whether to show post views for the site in the WordPress.com Reader
* Toggle is added to Jetpack Settings > Traffic > Stats (when the Stats module is enabled)
* Option is also made available in via the Settings REST API endpoints

Screenshot 2023-05-24 at 21 42 14

Screenshot 2023-05-18 at 17 23 21

### Other information:

## Jetpack product discussion

PT: pe7F0s-PF-p2

No, but it does control if post views for the site are visible to other users.

### Jetpack sites

* Connect the site to WordPress.com, if it isn't already
* Enable the Stats module, if needed (`/wp-admin/admin.php?page=jetpack_modules`)
* Go to `/wp-admin/admin.php?page=jetpack#/traffic` and see the toggle under Stats settings.
* Check changing the setting and saving, that the option `wpcom_reader_views_enabled` is added/updated

### REST API

- Apply this branch to a wpcom sandbox and sandbox public-api
- Use the [REST API console](https://developer.wordpress.com/docs/api/console/) to query `https://public-api.wordpress.com/rest/v1.4/sites/{site}/settings`
- See that `wpcom_reader_views_enabled` is added to the list of settings

### WP.com sites

Test with https://github.com/Automattic/wp-calypso/pull/77113, which adds the toggle to `/settings/reading/{site}` in Calypso


davemart-in (dismissed):
Tested on JP site. Works as expected. No errors.

CleanShot 2023-05-19 at 09 38 20@2x

Addison-Stavlo (dismissed):
Also tested on simple in conjunction with D111332 and https://github.com/Automattic/wp-calypso/pull/77113 . Looks good, works well!

samiff (commented):
This mostly looks good to me, though I was wondering why the setting toggle was being placed in `Settings > Reading` versus adding it to `Jetpack > Settings > Traffic > Jetpack Stats (section)`?

jeherve:
> This mostly looks good to me, though I was wondering why the setting toggle was being placed in `Settings > Reading` versus adding it to `Jetpack > Settings > Traffic > Jetpack Stats (section)`?

That's also what I would expect when adding a new Stats setting to wp-admin. You could add a similar setting in Settings > Reading in Calypso though, where it would be useful for WordPress.com sites.

creativecoder:
@jeherve @samiff Thanks for the feedback!

When I discussed this with @davemart-in before implementing, our thinking was that there may be other settings related to Reader and that we would want to group those together.

> That's also what I would expect when adding a new Stats setting to wp-admin. You could add a similar setting in Settings > Reading in Calypso though, where it would be useful for WordPress.com sites

That's a good idea. I'll move this to the Jetpack stats section in the Jetpack code, but keep the Calypso setting in Settings > Reading.

creativecoder:
This PS has now been updated, to add the toggle to the Jetpack Stats settings (instead of wp-admin Settings > Reading).

samiff (approved):
Updated toggle worked well for me 👍

anomiex:
I see this was flagged "ready to merge" four days ago. I'll go ahead and click the button so it gets into 12.2.

creativecoder:
> I see this was flagged "ready to merge" four days ago. I'll go ahead and click the button so it gets into 12.2.

Thanks you so much @anomiex ! I just forgot to hit the button before I went on AFK.

PR #30812, opened 2023-05-19
Social previews | Fix multiple issues

manzoorwanijk:
Fixes 1203820933396971-as-1204619418011134/f

## Proposed changes:

* Hides "Your post" section for Twitter preview
* Fix custom message not used for LinkedIn preview
* Fix modal header scrolling issue

### Other information:

## Jetpack product discussion

pdrWKz-Ka-p2

No

## Testing instructions:

* In Jetpack, run `jetpack build plugins/jetpack`
* Start writing a new post with a JT site with Social Advanced plan/features enabled
* Add a Featured Image to the post
* Click Jetpack icon on the top of the sidebar
* Under "Social Previews", click on "Preview"
* Resize the browser window to a smaller height (< 500px)
* Confirm that the modal header is not scrollable
* Confirm that Twitter preview is no longer available
* Confirm that LinkedIn preview reflects the custom message

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2023-05-19 at 1 51 56 PM | Screenshot 2023-05-19 at 1 52 06 PM  | 

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - 0-as-1204619418011134

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -686,8 +686,8 @@ importers:
         specifier: workspace:*
         version: link:../shared-extension-utils
       '@automattic/social-previews':
-        specifier: 2.0.0-beta.6
-        version: 2.0.0-beta.6(@babel/core@7.21.5)(@babel/runtime@7.21.5)(@types/react@18.0.27)(react-dom@18.2.0)(react@18.2.0)
+        specifier: 2.0.0-beta.7
+        version: 2.0.0-beta.7(@babel/core@7.21.5)(@babel/runtime@7.21.5)(@types/react@18.0.27)(react-dom@18.2.0)(react@18.2.0)
       '@wordpress/annotations':
         specifier: 2.32.0
         version: 2.32.0(react@18.2.0)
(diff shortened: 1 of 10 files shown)


pablinos (commented):
I'm guessing it's because I haven't got a connection, but twitter isn't showing up at all on my test site. I think we should always have the different networks showing in the modal as we will want to show people what the post will look like when someone else shares it, even if they're not sharing it to that network themselves.

This could be something that we do in a follow-up though.

There's still something odd going on with the state too. If I select a featured image, open the previews and close it, then select the 'share as a social post' checkbox and open the previews again, the Your post section still shows a link preview.

The modal header etc. all seems fixed though.

pablinos:
> There's still something odd going on with the state too. If I select a featured image, open the previews and close it, then select the 'share as a social post' checkbox and open the previews again, the Your post section still shows a link preview.

Doing some more testing this is specifically a problem with the featured image, but also isn't an issue we're trying to fix with this PR, so perhaps it would be best to treat it as a separate change.

pablinos:
> I'm guessing it's because I haven't got a connection, but twitter isn't showing up at all on my test site. 

Ignore me, I don't think I had the latest changes, and the twitter link preview is showing for me now.

> Doing some more testing this is specifically a problem with the featured image,

I think I've manage to fix this issue with [74711fd](https://github.com/Automattic/jetpack/pull/30812/commits/74711fd6fa5cad484a6881b23f67b2e8ef988c29)

pablinos (approved):
The twitter preview looks a little squashed at mobile sizes, but I think that's something which can be handled in a follow-up

image

Apart from that I think this is good to merge.

manzoorwanijk:
> Doing some more testing this is specifically a problem with the featured image, but also isn't an issue we're trying to fix with this PR, so perhaps it would be best to treat it as a separate change.

That is because of some complicated usage of `useEffect` and `useRef` in those components.

Issue #30771, opened 2023-05-17
AI Assistant: implement error handling when request fails


retrofox:
When the request fails, the app doesn't show any notice about it.

![Image](https://user-images.githubusercontent.com/77539/238916094-4acfcda3-b945-4eb2-8fe0-f4643eb7f1ef.png)

PR #30813, opened 2023-05-19
AI Assistant: catch and show error when requesting jwt

retrofox:
This PR tries to pick the error response of the jwt endpoint from the dotcom side to propagate it to the client site.
Also, on the client side, it tries to catch it to be able to update the UI.

Fixes #
Part of https://github.com/Automattic/jetpack/issues/30771

## Proposed changes:

* AI Assistant: catch and show error when reqeusting jwt

### Other information:

## Testing instructions:

* Go to the block editor
* Ensure the server side returns an error when the app hits the jwt token
* Confirm you see the notice above the block

https://github.com/Automattic/jetpack/assets/77539/4ad6e3cb-0216-4997-99d5-5f54e35ca0e9

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/get-suggestion-with-stream.js
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/get-suggestion-with-stream.js
@@ -83,27 +83,22 @@ export async function requestToken() {
 	const hasWPComFetch = ! window.wpcomFetch;
 	let data;
 
-	try {
-		if ( ! isSimpleSite() && ! isAtomicSite() ) {
-			// Jetpack sites
-			data = await apiFetch( {
-				path: '/jetpack/v4/jetpack-ai-jwt?_cacheBuster=' + Date.now(),
-				credentials: 'same-origin',
-				headers: {
-					'X-WP-Nonce': apiNonce,
-				},
-				method: 'POST',
-			} );
-		} else {
-			// Simple and Atomic sites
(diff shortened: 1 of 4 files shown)


projects/plugins/jetpack/extensions/blocks/ai-assistant/get-suggestion-with-stream.js
```diff
@@ -89,31 +93,38 @@ export async function requestToken() {
...
+				method: 'POST',
+			} );
+		} else {
+			// Simple and Atomic sites
+			data = await apiFetch( {
+				path: '/wpcom/v2/sites/' + siteSuffix + '/jetpack-openai-query/jwt',
+				method: 'POST',
+			} );
+		}
+	} catch ( err ) {
```
oskosk: I don't see how this `catch()` is adding any value. It just catches and throws the same error without any other action.
The result is that the code in `askJetpack` (which is calling `akQuestion`, and thus `getToken`) will still catch the same error generated by `apiFetch` failing

```
	try {
		source = await askQuestion( question );
	} catch ( err ) {
		debug( 'Error', err );
		return source;
	}
```

retrofox: Right. Thanks for the feedback. Updated.

retrofox:
closing in favor of https://github.com/Automattic/jetpack/pull/30984

PR #30815, opened 2023-05-19
CRM: Adding Automation contact triggers

coder-karen:
## Proposed changes:

* This PR adds Automation Contact triggers. The files with triggers will likely need changes as we better understand what data is being passed in and out, but currently all triggers take contact objects (with modified data depending on what was passed into the object, such as in this example: https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Contacts.php#L3007)
* This PR also includes tests

### Other information:

## Jetpack product discussion

https://github.com/Automattic/zero-bs-crm/issues/3097

No.

## Testing instructions:

* To test the objects that would be passed to the actions within each trigger, you can add an `error_log` within the relevant hook such as `zeroBSCRM_IA_EditCustomerWPHook` (within each case), such as `error_log('Firing the contact update: ' . json_encode( $obj ) );`. 
* You can then test updating (status / email / any contact field) / deleting / adding a new contact, and then make sure the $obj from the error log you are expecting is fired.

### DIFF
--- a/projects/plugins/crm/includes/ZeroBSCRM.InternalAutomatorRecipes.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.InternalAutomatorRecipes.php
@@ -55,6 +55,7 @@
 	#} WP Hook tie-ins (for Mike [and 3rd party developers!], mostly)
 	zeroBSCRM_AddInternalAutomatorRecipe('contact.new','zeroBSCRM_IA_NewCustomerWPHook',array());
 	zeroBSCRM_AddInternalAutomatorRecipe('contact.update','zeroBSCRM_IA_EditCustomerWPHook',array());
+	zeroBSCRM_AddInternalAutomatorRecipe( 'contact.status.update', 'zeroBSCRM_IA_EditCustomerWPHook', array() );
 	zeroBSCRM_AddInternalAutomatorRecipe('contact.vitals.update','zeroBSCRM_IA_EditCustomerVitalsWPHook',array());
 	zeroBSCRM_AddInternalAutomatorRecipe('contact.email.update','zeroBSCRM_IA_EditCustomerEmailWPHook',array());
 	zeroBSCRM_AddInternalAutomatorRecipe('contact.delete','zeroBSCRM_IA_DeleteCustomerWPHook',array());
@@ -1288,6 +1289,8 @@
 	function zeroBSCRM_IA_NewCustomerWPHook($obj=array()){
 
 		if (is_array($obj) && isset($obj['id']) && !empty($obj['id'])) {
+
+			do_action( 'jpcrm_automation_contact_new', $obj );
 		
 			do_action( 'jpcrm_after_contact_insert', $obj['id'] );
 
@@ -1297,42 +1300,78 @@
 		}
 
 	}
-   	#} Fires on 'customer.edit' IA 
-	function zeroBSCRM_IA_EditCustomerWPHook($obj=array()){
 
-		if (is_array($obj) && isset($obj['id']) && !empty($obj['id'])) {
-		
-			do_action( 'jpcrm_after_contact_update', $obj['id'] );
+/**
+ * Fires on 'contact.update', 'contact.email.update', and 'contact.status.update IA.
+ *
+ * @param array $obj An array holding contact object data.
+ */
+function zeroBSCRM_IA_EditCustomerWPHook( $obj = array() ) {
 
-			// legacy, use `jpcrm_after_contact_update` from 5.3+
-			do_action( 'zbs_edit_customer', $obj['id'] );
+	if ( is_array( $obj ) && isset( $obj['id'] ) && ! empty( $obj['id'] ) ) {
 
+		do_action( 'jpcrm_after_contact_update', $obj['id'] );
+
+		// legacy, use `jpcrm_after_contact_update` from 5.3+
+		do_action( 'zbs_edit_customer', $obj['id'] );
+
+		// If the contact status has changed.
+		if ( isset( $obj['from'] ) && ! empty( $obj['from'] ) ) {
+			do_action( 'jpcrm_automation_contact_status_update', $obj );
+		} elseif ( isset( $obj['userMeta']['zbsc_email'] ) && $obj['prev_contact']['email'] !== $obj['userMeta']['zbsc_email'] ) {  // If the contact email has changed.
+			do_action( 'jpcrm_automation_contact_email_update', $obj );
+		} else { // Fallback if any other contact field has changed.
+			do_action( 'jpcrm_automation_contact_update', $obj );
 		}
 
 	}
-   	#} Fires on 'customer.vitals.edit' IA 
+}
+
+	#} Fires on 'customer.vitals.edit' IA.
 	function zeroBSCRM_IA_EditCustomerVitalsWPHook($obj=array()){
 
 		if (is_array($obj) && isset($obj['id']) && !empty($obj['id'])) do_action('zbs_edit_customer_vitals', $obj['id']);
 
 	}
-   	#} Fires on 'customer.email.edit' IA 
+
+	/**
+	 * Fires on 'contact.email.update' IA. Now legacy, redirecting to zeroBSCRM_IA_EditCustomerWPHook
+	 *
+	 * @param array $obj An array holding contact object data.
+	 */
 	function zeroBSCRM_IA_EditCustomerEmailWPHook($obj=array()){
+
+		zeroBSCRM_IA_EditCustomerWPHook( $obj );
 
 		if (is_array($obj) && isset($obj['id']) && !empty($obj['id'])) do_action('zbs_edit_customer_email', $obj['id']);
 
 	}
-   	#} Fires on 'customer.delete' IA 
-	function zeroBSCRM_IA_DeleteCustomerWPHook($obj=array()){
 
-		if (is_array($obj) && isset($obj['id']) && !empty($obj['id'])) do_action('zbs_delete_customer', $obj['id']);
+/**
+ * Fires on 'contact.delete' IA.
+ *
+ * @param array $obj An array holding contact object data.
+ */
+function zeroBSCRM_IA_DeleteCustomerWPHook( $obj = array() ) {
 
+	if ( is_array( $obj ) && isset( $obj['id'] ) && ! empty( $obj['id'] ) ) {
+		do_action( 'jpcrm_automation_contact_delete', $obj );
+		// Legacy:
+		do_action( 'zbs_delete_customer', $obj['id'] );
 	}
+}
 
-	#} Fires on 'contact.before.delete' IA
+	/**
+	 * Fires on 'contact.before.delete' IA.
+	 *
+	 * @param array $obj An array holding contact object data.
+	 */
 	function zeroBSCRM_IA_BeforeDeleteCustomerWPHook($obj=array()){
 
-		if (is_array($obj) && isset($obj['id']) && !empty($obj['id'])) do_action('jpcrm_before_delete_contact', $obj);
+	if ( is_array( $obj ) && isset( $obj['id'] ) && ! empty( $obj['id'] ) ) {
+		do_action( 'jpcrm_automation_contact_before_delete', $obj );
+		do_action( 'jpcrm_before_delete_contact', $obj );
+	}
 
 	}
 
--- a/projects/plugins/crm/includes/ZeroBSCRM.InternalAutomatorRecipes.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.InternalAutomatorRecipes.php
@@ -1310,6 +1310,7 @@ function zeroBSCRM_IA_EditCustomerWPHook( $obj = array() ) {
 
 	if ( is_array( $obj ) && isset( $obj['id'] ) && ! empty( $obj['id'] ) ) {
 
+		do_action( 'jpcrm_automation_contact_update', $obj );
 		do_action( 'jpcrm_after_contact_update', $obj['id'] );
 
 		// legacy, use `jpcrm_after_contact_update` from 5.3+
@@ -1318,12 +1319,13 @@ function zeroBSCRM_IA_EditCustomerWPHook( $obj = array() ) {
 		// If the contact status has changed.
 		if ( isset( $obj['from'] ) && ! empty( $obj['from'] ) ) {
 			do_action( 'jpcrm_automation_contact_status_update', $obj );
-		} elseif ( isset( $obj['userMeta']['zbsc_email'] ) && $obj['prev_contact']['email'] !== $obj['userMeta']['zbsc_email'] ) {  // If the contact email has changed.
-			do_action( 'jpcrm_automation_contact_email_update', $obj );
-		} else { // Fallback if any other contact field has changed.
-			do_action( 'jpcrm_automation_contact_update', $obj );
 		}
-
+		// If the contact email has changed.
+		if ( isset( $obj['userMeta']['zbsc_email'] ) && isset( $obj['prev_contact'] ) ) {
+			if ( $obj['prev_contact']['email'] !== $obj['userMeta']['zbsc_email'] ) {
+				do_action( 'jpcrm_automation_contact_email_update', $obj );
+			}
+		}
 	}
 }
 
--- a/projects/plugins/crm/src/automation/commons/triggers/contacts/class-contact-before-deleted.php
+++ b/projects/plugins/crm/src/automation/commons/triggers/contacts/class-contact-before-deleted.php
@@ -19,41 +19,66 @@ class Contact_Before_Deleted extends Base_Trigger {
 	/**
 	 * @var Automation_Workflow The Automation workflow object.
 	 */
-	private $workflow;
+	protected $workflow;
 
-	/**
-	 * Contructs the Contact_Before_Deleted instance.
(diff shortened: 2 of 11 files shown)


projects/plugins/crm/includes/ZeroBSCRM.InternalAutomatorRecipes.php
```diff
@@ -1297,42 +1300,78 @@ function zeroBSCRM_IA_NewCustomerWPHook($obj=array()){
...
+		// legacy, use `jpcrm_after_contact_update` from 5.3+
+		do_action( 'zbs_edit_customer', $obj['id'] );
 
+		// If the contact status has changed.
+		if ( isset( $obj['from'] ) && ! empty( $obj['from'] ) ) {
+			do_action( 'jpcrm_automation_contact_status_update', $obj );
+		} elseif ( isset( $obj['userMeta']['zbsc_email'] ) && $obj['prev_contact']['email'] !== $obj['userMeta']['zbsc_email'] ) {  // If the contact email has changed.
+			do_action( 'jpcrm_automation_contact_email_update', $obj );
+		} else { // Fallback if any other contact field has changed.
+			do_action( 'jpcrm_automation_contact_update', $obj );
```
cleacos: They should be parallel conditions, we have to check the `status`, the `email`, and the `fallback` case. For example, a contact update could be status and email change, so we should trigger both events.

coder-karen: Ah got it, thanks! Added another commit to change that.

projects/plugins/crm/includes/ZeroBSCRM.InternalAutomatorRecipes.php
```diff
@@ -55,6 +55,7 @@
 	#} WP Hook tie-ins (for Mike [and 3rd party developers!], mostly)
 	zeroBSCRM_AddInternalAutomatorRecipe('contact.new','zeroBSCRM_IA_NewCustomerWPHook',array());
 	zeroBSCRM_AddInternalAutomatorRecipe('contact.update','zeroBSCRM_IA_EditCustomerWPHook',array());
+	zeroBSCRM_AddInternalAutomatorRecipe( 'contact.status.update', 'zeroBSCRM_IA_EditCustomerWPHook', array() );
```
kallehauge, line 58: Oh, we're reusing the old trigger system. Can we perhaps simplify this in a more generic WordPress way and add a `do_action()` in the contacts DAL instead?

That way, other parts of the code can react on the action as well (not just our automations)?
Example: we could move `do_action( 'jpcrm_after_contact_update', ...)` to the bottom of `zbsDAL_contacts::addUpdateContact` together with the rest of the actions we need to trigger automations?

That way we don't introduce a pattern of creating duplicate hooks and/or hooks that are automation specific (read: automations are supposed to hook into the system, not be a deeply integrated system inside all code)

projects/plugins/crm/includes/ZeroBSCRM.InternalAutomatorRecipes.php
```diff
@@ -1288,6 +1289,8 @@ function zeroBSCRM_IA_CustomerStatusChangeAutoLog($obj=array()){
 	function zeroBSCRM_IA_NewCustomerWPHook($obj=array()){
 
 		if (is_array($obj) && isset($obj['id']) && !empty($obj['id'])) {
+
+			do_action( 'jpcrm_automation_contact_new', $obj );
```
kallehauge, line 1293: _(See `zeroBSCRM_AddInternalAutomatorRecipe` comment)_

projects/plugins/crm/includes/ZeroBSCRM.InternalAutomatorRecipes.php
```diff
@@ -1297,42 +1300,80 @@ function zeroBSCRM_IA_NewCustomerWPHook($obj=array()){
...
+ * @param array $obj An array holding contact object data.
+ */
+function zeroBSCRM_IA_EditCustomerWPHook( $obj = array() ) {
 
-			// legacy, use `jpcrm_after_contact_update` from 5.3+
-			do_action( 'zbs_edit_customer', $obj['id'] );
+	if ( is_array( $obj ) && isset( $obj['id'] ) && ! empty( $obj['id'] ) ) {
 
-		}
+		do_action( 'jpcrm_automation_contact_update', $obj );
```
kallehauge, line 1313: _(See `zeroBSCRM_AddInternalAutomatorRecipe` comment)_

projects/plugins/crm/src/automation/commons/triggers/contacts/class-contact-before-deleted.php
```diff
@@ -0,0 +1,49 @@
...
+class Contact_Before_Deleted extends Base_Trigger {
+
+	/**
+	 * Contructs the Contact_Before_Deleted instance.
+	 */
+	public function __construct() {
+		$trigger_data = array(
+			'name'        => 'contact_before_delete',
+			'title'       => 'Contact Before Deleted',
+			'description' => 'Triggered just before a contact is deleted.',
```
kallehauge: These looks like they could benefit from translation 😄 

```suggestion
			'title'       => __( 'Contact Before Deleted', 'zero-bs-crm' ),
			'description' => __( 'Triggered just before a contact is deleted.', 'zero-bs-crm' ),
```

projects/plugins/crm/src/automation/commons/triggers/contacts/class-contact-deleted.php
```diff
@@ -0,0 +1,49 @@
...
+	public function __construct() {
+		$trigger_data = array(
+			'name'        => 'contact_delete',
+			'title'       => 'Contact Deleted',
+			'description' => 'Triggered when a contact is deleted.',
+			'category'    => 'contact',
+		);
+
+		parent::__construct( $trigger_data );
+	}
```
kallehauge: It looks like these attributes are data we'll always rely on, so should we perhaps convert them into class properties instead?

projects/plugins/crm/src/automation/commons/triggers/contacts/class-contact-deleted.php
```diff
@@ -0,0 +1,49 @@
...
+	public function init( Automation_Workflow $workflow ) {
+		add_action(
+			'jpcrm_automation_contact_delete',
+			function ( $contact_data ) use ( $workflow ) {
+				$workflow->execute( $this, $contact_data );
+			},
+			10,
+			2
+		);
+	}
```
kallehauge: Could this be simplified in a more traditional WP way by creating a callback method as part of the trigger class instead of relying on parents scope of an anonymous function? 😄 

AFAIK then we could achieve the same by doing something like this: 

```suggestion
	public function init( Automation_Workflow $workflow ) {
		add_action( 'jpcrm_automation_contact_delete', array( $workflow, 'callback' ), 10 );
	}

	public function callback( $contact_data ) {
		// do something with the contact data
	}
```

That way `init()` is only responsible for triggering actions (read: separation of concerns could become clearer this way)

cleacos: > That way init() is only responsible for triggering actions

Here, the init method is not triggering the action, it's adding the action hook to the WP, and then the callback function (embedded function, using closures way), that executes the workflow. 

If we want to extract the closure, it should be:

```
	private $workflow;
	
	public function init( Automation_Workflow $workflow ) {
		$this->workflow = $workflow;
		add_action( 'jpcrm_automation_contact_delete', array( $this, 'execute_workflow' ) );
	}
	
	public function execute_workflow( $contact_data ) {
		if ( $this->workflow ) {
			$this->workflow->execute( $this, $contact_data );
		}
	}
```

The `do_action`, in this case, will have 1 argument, the contact object/data. In this case, the $workflow should be stored in the trigger instance (private property). I'm fine extracting it this way, but I have to update `Base_Trigger`, so triggers implementations don't have to be aware of it

kallehauge: I, personally, find it more like "the WordPress way" to do it the way you just described 😄

projects/plugins/crm/includes/ZeroBSCRM.InternalAutomatorRecipes.php
cleacos: > Oh, we are using the old trigger system. Can we maybe simplify this in a more generic WordPress way and add a do_action() in the contact DAL instead?

That was the debate we had in our last session, whether we end up going our own way or reusing the current, already centralized system until we revise it. But moving away from the old trigger system later and adding the WP hooks directly to the DAL is not a big deal. We don't have a clear picture here, so the decision was to be conservative and use this centralized location to emit the hooks if they already exist. In this case, we need to create a new one because we need the full object/data and the current hook only passes the contact ID. So if we already have the full contact data, we avoid another SQL query that passes the object/data directly.

> This way, other parts of the code can also respond to the action (not just our automation)?

Other parts of the code can react to the WP action, because from the DAL we call this old system to trigger the action, and this system calls `do_action` at the end.

**Summary:**

- I am in favor of creating our own system for automation and centralizing everything we need there. And that was more or less the conclusion, we will make in the end. 
- We started to reuse the old centralized system because it already exists and to move forward with creating/testing the first triggers. 
- We need a centralized system because we need to multicast the same event in different ways and avoid adding all the hooks in the DAL. For instance:
   - A contact was updated. The DAL emmit the action and notifies it to the trigger event, using `do_action` or calling to the new triggering system `$jpcrm_trigger_system->event( 'contact_updated', $contact_old, $contact_updated )`
   - Then, in the new triggering system, we cast the WP actions. One single event could imply multiple actions:
      - Always: `do_action( 'jpcrm_contact_updated', $contact_updated, $contact_old )`
      - If the email was changed: `do_action( 'jpcrm_contact_email_updated', $contact_updated, $old_email )`
      - If the status was changed: `do_action( 'jpcrm_contact_status_updated', $contact_updated, $prev_status )`
      - If the phone was changed: `do_action( 'jpcrm_contact_email_updated', $contact_updated )`
      - If a tag was added: `do_action( 'jpcrm_contact_tag_added', $contact_updated, $tag_added )`
      - If a tag was removed: `do_action( 'jpcrm_contact_tag_removed', $contact_updated, $tag_removed )`
      - If a task was added: `do_action( 'jpcrm_contact_task_added', $contact_updated, $task_added )`
      - If a task was removed: `do_action( 'jpcrm_contact_task_removed', $contact_updated, $task_removed )`
 - That way, we don't overload DAL with more code lines and move the responsibility, and the triggering logic, to another actor, the triggering system. And we will have a class responsible for this task, we are not replacing the WP system, only adding a middle layer to handle events before using the WP hooks.
 - Later, once we implement most of the triggers, we will have to migrate the old triggers to the new ones to avoid duplicated system and unused code. TBH I've not gone deeper into the old system.

projects/plugins/crm/src/automation/commons/triggers/contacts/class-contact-deleted.php
cleacos: In the main branch,[ I replied to this one](https://github.com/Automattic/jetpack/pull/30366#discussion_r1201254426). Really we don't need to call the parent constructor, in fact, we can instance triggers without any property filled. The Automation System will be responsible for accepting triggers or not, for example, we only accept triggers with names.

That said, each trigger implementation should set its own properties directly:

```
public function __construct() {
    $this->name         = 'contact_delete';
    $this->title        = __( 'Contact Deleted', 'zero-bs-crm' );
    $this->description  = __( 'Triggered when a contact is deleted.', 'zero-bs-crm' );
    $this->category     = 'contact';
}
```

or... since the trigger data are immutable (set them programmatically), using the static way that will facilitate access to the trigger data information without the need to instance it:

```
public function __construct() {
    self::$name         = 'contact_delete';
    self::$title        = __( 'Contact Deleted', 'zero-bs-crm' );
    self::$description  = __( 'Triggered when a contact is deleted.', 'zero-bs-crm' );
    self::$category     = 'contact';
} 
```

projects/plugins/crm/src/automation/commons/triggers/contacts/class-contact-deleted.php
cleacos: Karen and me tested this solution, and it works well and easier to test it using the PHPUnit Mock builder

projects/plugins/crm/includes/ZeroBSCRM.InternalAutomatorRecipes.php
cleacos: Karen and I discussed it, and it seems better to implement our centralized triggering system separately from the current one. We will build it from scratch, avoiding patches and hacks, and without the need to understand and adapt to the old system. Later, once we implement most of the core triggers, we could migrate the old one to the new one if it's necessary.

We need a centralized triggering system to be able to emit multiple WP actions from one original event as I exposed in the example above.

projects/plugins/crm/src/automation/commons/triggers/contacts/class-contact-before-deleted.php
```diff
@@ -0,0 +1,84 @@
...
+	 * Execute the workflow.
+	 *
+	 * @param array $contact_data The contact data to be included in the workflow.
+	 * @throws Automation_Exception Throws a 'class not found' or general error.
+	 */
+	public function execute_workflow( $contact_data = null ) {
+		if ( $this->workflow ) {
+			$this->workflow->execute( $this, $contact_data );
+		}
+	}
```
cleacos: We won't need to implement these methods, they are already implemented on Base_Trigger.

projects/plugins/crm/src/automation/commons/triggers/contacts/class-contact-deleted.php
```diff
@@ -0,0 +1,84 @@
...
+	 * Execute the workflow.
+	 *
+	 * @param array $contact_data The contact data to be included in the workflow.
+	 * @throws Automation_Exception Throws a 'class not found' or general error.
+	 */
+	public function execute_workflow( $contact_data = null ) {
+		if ( $this->workflow ) {
+			$this->workflow->execute( $this, $contact_data );
+		}
+	}
```
cleacos: We won't need to implement these methods, they are already implemented on Base_Trigger.

projects/plugins/crm/src/automation/commons/triggers/contacts/class-contact-email-updated.php
```diff
@@ -0,0 +1,84 @@
...
+	 * Execute the workflow.
+	 *
+	 * @param array $contact_data The contact data to be included in the workflow.
+	 * @throws Automation_Exception Throws a 'class not found' or general error.
+	 */
+	public function execute_workflow( $contact_data = null ) {
+		if ( $this->workflow ) {
+			$this->workflow->execute( $this, $contact_data );
+		}
+	}
```
cleacos: We won't need to implement these methods, they are already implemented on Base_Trigger.

projects/plugins/crm/src/automation/commons/triggers/contacts/class-contact-new.php
```diff
@@ -0,0 +1,84 @@
...
+	 * Execute the workflow.
+	 *
+	 * @param array $contact_data The contact data to be included in the workflow.
+	 * @throws Automation_Exception Throws a 'class not found' or general error.
+	 */
+	public function execute_workflow( $contact_data = null ) {
+		if ( $this->workflow ) {
+			$this->workflow->execute( $this, $contact_data );
+		}
+	}
```
cleacos: We won't need to implement these methods, they are already implemented on Base_Trigger.

projects/plugins/crm/src/automation/commons/triggers/contacts/class-contact-status-updated.php
```diff
@@ -0,0 +1,84 @@
...
+	 * Execute the workflow.
+	 *
+	 * @param array $contact_data The contact data to be included in the workflow.
+	 * @throws Automation_Exception Throws a 'class not found' or general error.
+	 */
+	public function execute_workflow( $contact_data = null ) {
+		if ( $this->workflow ) {
+			$this->workflow->execute( $this, $contact_data );
+		}
+	}
```
cleacos: We won't need to implement these methods, they are already implemented on Base_Trigger.

projects/plugins/crm/src/automation/commons/triggers/contacts/class-contact-updated.php
```diff
@@ -0,0 +1,85 @@
...
+	 * Execute the workflow.
+	 *
+	 * @param array $contact_data The contact data to be included in the workflow.
+	 * @throws Automation_Exception Throws a 'class not found' or general error.
+	 */
+	public function execute_workflow( $contact_data = null ) {
+		if ( $this->workflow ) {
+			$this->workflow->execute( $this, $contact_data );
+		}
+	}
```
cleacos: We won't need to implement these methods, they are already implemented on Base_Trigger.

projects/plugins/crm/src/automation/commons/triggers/contacts/class-contact-before-deleted.php
coder-karen: Thanks, followed up with an additional commit (for all contact trigger classes).

cleacos (approved):
LGTM! into the Automation main branch

PR #30817, opened 2023-05-19
CRM: Adding Automation company triggers

coder-karen:
## Proposed changes:

* This PR adds Automation Company triggers. The files with triggers will likely need changes as we better understand what data is being passed in and out, but currently all triggers take company objects (with modified data depending on what was passed into the object, such as in this example: https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Companies.php#L2398)
* This PR also includes tests

### Other information:

## Jetpack product discussion

https://github.com/Automattic/zero-bs-crm/issues/3098

No.

## Testing instructions:

* To test the objects that would be passed to the actions within each trigger, you can add an `error_log` within the relevant hook such as `zeroBSCRM_IA_EditCompanyWPHook` (within each case), such as `error_log('Firing the company update: ' . json_encode( $obj ) );`.
* You can then test updating / deleting / adding a new company, and then make sure the $obj from the error log you are expecting is fired in the tests that call the `do_action` hook.

### DIFF
--- a/projects/plugins/crm/changelog/add-crm-automation-triggers-companies
+++ b/projects/plugins/crm/changelog/add-crm-automation-triggers-companies
@@ -1,0 +1,4 @@
+Significance: patch
+Type: added
+
+Automations: Added company triggers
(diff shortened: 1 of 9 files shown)


cleacos:
The `$workflow` property in the triggers is not necessary.

cleacos (approved):
Test passed! Code looks good.

LGTM!

PR #30820, opened 2023-05-19
AI Assistant: Ignore first chunk on unclear prompt check

dhasilva:
Fixes an error where, somehow, the first response chunk is either not sent, received or generated and the prompt was not undestood, ending up with this content:

image

## Proposed changes:

* Refactors the logic a bit to make it more easily understood
* Ignores the first `__` in the message when checking for the error marker
* Tweaks and moves the prompt

### Other information:

## Testing instructions:

This one is very tricky to test because I was not able to reproduce the problem normally.
* Add an AI Assistant block and ask some nonsense like "sdjsd"
* Check that the notice about trying again is displayed as normal
* Check that subsequent requests, both clear and unclear, are treated correctly

To simulate a state where the first `__` is not received, it would be possible with something like this:
```js
		if ( chunk && chunk !== '__' ) {
			this.fullMessage += chunk;
```


projects/plugins/jetpack/extensions/blocks/ai-assistant/get-suggestion-with-stream.js
```diff
@@ -151,11 +152,21 @@ export class SuggestionsEventSource extends EventSource {
 		if ( chunk ) {
 			this.fullMessage += chunk;
 
-			if ( this.fullMessage.startsWith( '__JETPACK_AI_ERROR__' ) ) {
-				// The error is confirmed
-				this.dispatchEvent( new CustomEvent( 'error_unclear_prompt' ) );
-			} else if ( ! '__JETPACK_AI_ERROR__'.startsWith( this.fullMessage ) ) {
-				// Confirmed to be a valid response
+			if ( ! this.isPromptClear ) {
```
oskosk: Can we move all this error handling to another method in the class?

retrofox (approved):
LGTM. Nice prompt enhancement. :shipit:

PR #30821, opened 2023-05-19
AI Assistant: Update upgrade nudge copy 

oskosk:
Updates the upgrade nudge to match the copy in Figma. gG1smMHbNGas0qwP9y6RwK-fi-3609_17698

#### After

image

#### Before

image

Fixes #

## Proposed changes:

* Updates the copy

### Other information:

No
## Testing instructions:

* Confirm the change makes sense

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/upgrade-prompt.js
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/upgrade-prompt.js
@@ -19,7 +19,10 @@ const UpgradePrompt = () => {
 			buttonText={ 'Upgrade' }
 			checkoutUrl={ `${ window?.Jetpack_Editor_Initial_State?.adminUrl }admin.php?page=my-jetpack#/add-jetpack-ai` }
 			className={ 'jetpack-ai-upgrade-banner' }
-			description={ __( 'Free requests used. Upgrade now to keep using Jetpack AI.', 'jetpack' ) }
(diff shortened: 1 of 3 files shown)


projects/plugins/jetpack/extensions/blocks/ai-assistant/upgrade-prompt.js
```diff
@@ -19,7 +19,10 @@ const UpgradePrompt = () => {
 			buttonText={ 'Upgrade' }
 			checkoutUrl={ `${ window?.Jetpack_Editor_Initial_State?.adminUrl }admin.php?page=my-jetpack#/add-jetpack-ai` }
 			className={ 'jetpack-ai-upgrade-banner' }
-			description={ __( 'Free requests used. Upgrade now to keep using Jetpack AI.', 'jetpack' ) }
+			description={ __(
+				'You have reached the limit of 20 free requests. Upgrade now to keep using Jetpack AI.',
```
lhkowalski, line 23: Just a suggestion for a generic message, still friendly but flexible:

```suggestion
				'You have reached the limit of free requests. Upgrade now to keep using Jetpack AI.',
```

lhkowalski (approved):
Change makes sense to me, wording got better. Since we know we are going to offer `20` as the free limit, I think it's OK to keep it hardcoded for now.

Alternatively, we could use a generic number with this better wording, as the suggestion I left.

PR #30828, opened 2023-05-22
Fix warning in WP Super Cache taxonomy preload loop

donnchawp:
This is a tiny code change to fix a PHP warning in the taxonomy preload loop reported on the forum here: https://wordpress.org/support/topic/wpsc-1-10-0-alpha-warning/

## Proposed changes:
Check that the $details array has more than 0 elements in it, instead of checking that it hasn't got 1 element. If it has 0 elements, the next check on $details[0] generates a warning.

### Other information:

No

## Testing instructions:
None.


haqadn (approved):
Codewise looks good 👍🏼

PR #30829, opened 2023-05-22
Boost: Update Minify to use old Page Optimize file functions

### LINKED ISSUE #30791: Boost: Minify breaks website on specific servers
### Impacted plugin

Boost

### Quick summary

pedxs5-eK-p2#comment-109

Link to the whole debugging thread p1684395317660539-slack-C016BBAFHHS.

To summarize - when testing with Page Optimize, everything works as expected, because it relies on `file_get_contents`. In Boost Minify, however, we moved to WP's File System in https://github.com/Automattic/jetpack/pull/29352 to make the linter happy. I'm not sure why, but WP decides to use `WP_Filesystem_FTPext` instead of `WP_Filesystem_Direct` and this causes the cache service to fail with an error 500.

On Jurassic Ninja test sites, the filesystem instance is of `WP_Filesystem_Direct` and everything works.

### Steps to reproduce

Unsure why this happens, it was discovered while testing Boost 1.9 beta 1.

_No response_

### What actually happened

_No response_

### Impact

One

### Available workarounds?

There is no user impact

Self-hosted

### Logs or notes

_No response_

dilirity:
Fixes #30791.

## Proposed changes:

* Updates Minify concat file functions to the old ones Page Optimize used before the changes introduced in https://github.com/Automattic/jetpack/pull/29352.

### Other information:

## Jetpack product discussion

p1684752302963189/1684746834.766979-slack-C016BBAFHHS

no

### Testing that it breaks properly
* Install the `trunk` version of the plugin;
* Enable CSS or JS concatenation;
* Your WP installation's File System should use `ftpext`. To force it to use it, add `define( 'FS_METHOD', 'ftpext' );` to `wp-config.php`;
* Check any page in the front end and it should be broken.

### Testing that it works properly
* Install this version of the plugin;
* Enable CSS or JS concatenation;
* Open a page and check that its source contains `_static` references and it works as expected.

### DIFF
--- a/projects/plugins/boost/app/lib/minify/Concatenate_JS.php
+++ b/projects/plugins/boost/app/lib/minify/Concatenate_JS.php
@@ -62,8 +62,6 @@ class Concatenate_JS extends WP_Scripts {
 	 * Override for WP_Scripts::do_item() - this is the method that actually outputs the scripts.
 	 */
 	public function do_items( $handles = false, $group = false ) {
-		global $wp_filesystem;
-
 		$handles     = false === $handles ? $this->queue : (array) $handles;
 		$javascripts = array();
 		$siteurl     = apply_filters( 'page_optimize_site_url', $this->base_url );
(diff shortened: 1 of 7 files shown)


haqadn (commented):
While this solves the problem, perhaps we should depend on [WP_Filesystem_Direct](https://developer.wordpress.org/reference/classes/wp_filesystem_direct/get_contents/)?

donnchawp (approved):
Works as expected on my server. I'd love to see us use the FTP Filesystem class to chmod wp-content, then directly mkdir the cache/page_optimize directory, and then fix wp-content again immediately.

thingalon:
> While this solves the problem, perhaps we should depend on [WP_Filesystem_Direct](https://developer.wordpress.org/reference/classes/wp_filesystem_direct/get_contents/)?

I think this is a good idea for a future update. For our first release, we're leaning pretty hard on what Page Optimize already does.

dilirity:
Opened an issue regarding the WP File System - https://github.com/Automattic/jetpack/issues/30861.

thingalon:
@dilirity when we fix something that needs to be copied over to a current beta version (e.g.: 1.9.0), please mark them as "Needs cherry-pick" and @ me on them -- or cherry-pick them directly. :)

thingalon:
Cherry picked to 1.9.0 in e74f85214586b3a0859d43e5caa0100b9e21b0d2

PR #30830, opened 2023-05-22
WP Super Cache: don't delete child taxonomies during preload

donnchawp:
The preload system in WP Super Cache allows a site owner to prime the cache by visiting taxonomy archives and post/page URLs.
The original cached page would have to be deleted before the new page was fetched. Unfortunately, the taxonomy loop deleted a taxonomy, and any sub directories which had the effect of deleting the cached pages for child taxonomies.

This PR fixes that by only deleting files in the taxonomy directory, and then deleting the feed and page directories separately.

ref: https://wordpress.org/support/topic/preload-mode-skips-caching-pages-or-deletes-already-cached-pages/

### Other information:

No

## Testing instructions:
To test this you'll need to create child taxonomies on a test blog.
Enable debugging and when it's done, verify that the cache pages for the child taxonomies still exist.

### DIFF
--- a/projects/plugins/super-cache/wp-cache.php
+++ b/projects/plugins/super-cache/wp-cache.php
@@ -3271,8 +3271,8 @@ function wp_cron_preload_cache() {
 						$dir = get_supercache_dir() . $url_info[ 'path' ];
 						wp_cache_debug( "wp_cron_preload_cache: delete $dir", 5 );
 						wpsc_delete_files( $dir );
-						prune_super_cache( $dir . 'feed/', true );
-						prune_super_cache( $dir . 'page/', true );
+						prune_super_cache( trailingslashit( $dir ) . 'feed/', true );
+						prune_super_cache( trailingslashit( $dir ) . 'page/', true );
 						$fp = @fopen( $permalink_counter_msg, 'w' );
 						if ( $fp ) {
 							@fwrite( $fp, "$taxonomy: $url" );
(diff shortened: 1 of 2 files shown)


projects/plugins/super-cache/wp-cache.php
```diff
@@ -3270,7 +3270,9 @@ function wp_cron_preload_cache() {
 						$url_info = parse_url( $url );
 						$dir = get_supercache_dir() . $url_info[ 'path' ];
 						wp_cache_debug( "wp_cron_preload_cache: delete $dir", 5 );
-						prune_super_cache( $dir );
+						wpsc_delete_files( $dir );
+						prune_super_cache( $dir . 'feed/', true );
+						prune_super_cache( $dir . 'page/', true );
```
dilirity: If the permalink structure doesn't end with a trailing slash, the path will be incorrect resulting in a PHP warning:
```
PHP Warning:  Undefined array key 0 in /usr/local/src/jetpack-monorepo/projects/plugins/super-cache/wp-cache.php on line 3303
```

I've worked on quite a few websites that have `.php` appended at the end of the post URL.

dilirity (changes requested):
I wasn't able to properly test this, but I discovered something small - details below.

projects/plugins/super-cache/wp-cache.php
donnchawp: Thanks, that's a good point. I guess it's better to have //feed/ there.

donnchawp: trailingslashit() helps here of course.

thingalon (approved):
Ok, looks straightforward to me :)

dilirity (approved):

PR #30834, opened 2023-05-22
AI Assistant: set fade effect when just requested completion

retrofox:
Based on the following feedback...

> It feels like the text should fade out a bit or some other visual cue while the thing is processing as an indication that it’s acting on the whole chunk of text.

p1683892486076719/1683890690.532879-slack-C054LN8RNVA

Fixes #

## Proposed changes:

* AI Assistant: set fade effect when just requested completion

### Other information:

## Testing instructions:

* Go to the block editor
* Generate content
* Iterate over the block content. For instance, change the tone, language, etc
* Confirm the block fades the content while the request starts
* Confirm the opacity gets removed once the block starts filling its content with new text.

https://github.com/Automattic/jetpack/assets/77539/8ab61b0d-93a2-4d13-b9a7-c76cdfc277c6

### DIFF
--- a/projects/plugins/jetpack/changelog/update-ai-assistant-tweak-asking-content
+++ b/projects/plugins/jetpack/changelog/update-ai-assistant-tweak-asking-content
@@ -1,0 +1,4 @@
+Significance: minor
+Type: enhancement
+
+AI Assistant: set fade effect when just requested completion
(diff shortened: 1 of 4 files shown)


lhkowalski (approved):
Tests as expected and the code chage is neat.

PR #30841, opened 2023-05-22
WP Super Cache: reset the preload settings when done

donnchawp:
The preload wasn't finishing correctly because the transient setting we used for taxonomies wasn't deleted at the end of the preload loop.

This PR fixes that by calling wpsc_reset_preload_settings(), but I had to move the wp_mail() command out of there as it wouldn't make sense in this context.

### Other information:

No

## Testing instructions:
Create a test site with a couple of posts and tags.
Preload the site and you will notice the preload is stuck at the last post.
The cancel button won't work either, but if you leave it ten minutes it will fix itself as the transient setting expires.
You can also run `delete_transient( 'taxonomy_preload' );` to delete it.
Apply the PR and preload. The preload should stop when it's done.

### DIFF
--- a/projects/plugins/super-cache/wp-cache.php
+++ b/projects/plugins/super-cache/wp-cache.php
@@ -3283,6 +3283,11 @@ function wp_cron_preload_cache() {
 						if ( @file_exists( $cache_path . "stop_preload.txt" ) ) {
 							wp_cache_debug( 'wp_cron_preload_cache: cancelling preload. stop_preload.txt found.', 5 );
 							wpsc_reset_preload_settings();
+
+							if ( $wp_cache_preload_email_me ) {
+								// translators: Home URL of website
+								wp_mail( get_option( 'admin_email' ), sprintf( __( '[%1$s] Cache Preload Stopped', 'wp-super-cache' ), home_url(), '' ), ' ' );
+							}
 							return true;
 						}
 					}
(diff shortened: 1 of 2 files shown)


projects/plugins/super-cache/wp-cache.php
```diff
@@ -3374,6 +3379,11 @@ function wp_cron_preload_cache() {
 			if ( @file_exists( $cache_path . "stop_preload.txt" ) ) {
 				wp_cache_debug( 'wp_cron_preload_cache: cancelling preload. stop_preload.txt found.', 5 );
 				wpsc_reset_preload_settings();
+
+				if ( $wp_cache_preload_email_me ) {
+					// translators: Home URL of website
+					wp_mail( get_option( 'admin_email' ), sprintf( __( '[%1$s] Cache Preload Stopped', 'wp-super-cache' ), home_url(), '' ), ' ' );
+				}
```
dilirity, line 3386: Just a suggestion - you can move this code to a function so it doesn't repeat.

dilirity (approved):
It's working! Cancelling preload no longer gets stuck. I also got the emails, though not sure why one of them was marked as SPAM
image

Maybe it's because it has an empty body?

projects/plugins/super-cache/wp-cache.php
donnchawp: I think I need to put all the email notifications for preloads into a single function. I'll leave it for another PR.

PR #30843, opened 2023-05-22
Update boost score bar to use package api

CodeyGuyDylan:
This updates the Boost plugin to get Boost Scores from the API package instead of the local code

## Proposed changes:
 * Remove functions relating to boost scores locally
 * Import boost-score-api `js-package` and use it instead

### Other information:

## Jetpack product discussion
P2: pc9hqz-1tV-p2

No

## Testing instructions:

1. Checkout this patch or get an ephemeral site with the Jetpack Beta plugin having Jetpack Boost go off this branch
2. If you haven't already, Install the Jetpack Boost plugin (can be free or paid)
3. Go to `/wp-admin/admin.php?page=jetpack-boost`
4. Make sure the Scores load as expected
image


CodeyGuyDylan:
I think a good followup task for this would be to export the `api` object from the package so it can be imported and used for more than just the boost scores in the Boost plugin. What do you think @thingalon?

Also noting here that I plan to followup with another task to self identify the client making the request for boost scores. Discussion starts here: p1684459218186289/1684437025.553649-slack-C016BBAFHHS

robertsreberski (approved):
I've tested code through Jetpack Beta. Jetpack Boost 2.0.0-alpha version has been loaded to the WP instance, and scores have been displayed on the Boost page.

The code looks good!

One thing to precise. Shouldn't Jetpack Beta feature branch be activated for Jetpack Boost, not Jetpack general plugin?

CodeyGuyDylan:
> One thing to precise. Shouldn't Jetpack Beta feature branch be activated for Jetpack Boost, not Jetpack general plugin?

Yes that is correct, I'll update the plan 😄

thingalon (approved):
Changes look ok, seem to work

PR #30850, opened 2023-05-22
Release packages/blaze

anomiex:
## Proposed changes:

It seems that the redux store in js-packages/connection is global rather than per-bundle, so when something is added to that then all the deps that bundle it should get rebuilt.

Looks like blaze is the only package that didn't get a release already, and 🤞 the standalone plugins that bundle it directly (not via packages) look like they all either only load it on their own admin pages or check for Jetpack first.

I think this will cover Jetpack-the-plugin along with any of our standalones too. Not so sure about a site with say Social and VideoPress but not Jetpack-the-plugin; hopefully a real fix gets in before either of those gets a release.

### Other information:

## Jetpack product discussion

p1684753564127059-slack-CBG1CP4EN
p1684777125233179-slack-C01U2KGS2PQ

No

## Testing instructions:

* Nothing to test here. The test will be in cherry-picking this to the Jetpack weekly.

### DIFF
--- a/projects/packages/blaze/CHANGELOG.md
+++ b/projects/packages/blaze/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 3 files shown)


anomiex:
Sigh, I should probably update that check one of these days to account for that edge case.

samiff (approved):

PR #30860, opened 2023-05-22
CRM: Resolves 3104 - Emerald listviews

tbradsha:
Resolves Automattic/zero-bs-crm#3104 - Emerald listviews

Upon merge, this will solve the following:
Automattic/zero-bs-crm#690
Automattic/zero-bs-crm#774
Automattic/zero-bs-crm#1571
Automattic/zero-bs-crm#2996
Automattic/zero-bs-crm#2997

## Proposed changes:

Old:
image

New:
image

Notes:
* doesn't handle multiple segments of same name (pre-existing bug: Automattic/zero-bs-crm#3113)
* listview per-page settings are reset periodically (pre-existing bug: Automattic/zero-bs-crm#3115)
* scrolling issue on shorter listviews (pre-existing bug: Automattic/zero-bs-crm#3116)
* quickfilters sometimes are not available (pre-existing issue: Automattic/zero-bs-crm#1413)
* quickfilters aren't sorted (there currently is no way to differentiate between segment, status, and built-in filters)
* no top pagination (for long results pages)
* no bottom bulk actions dropdown (for long results pages)
* sort and pagination don't use AJAX calls (but quickfilters, tags, and search now do)
* no change in "no results" or "loading" content
* "table options" button should probably use our new button components
* list view options have not been touched
* there's no clear way to determine if a table field is a link
* listview field styles are not curated (see "Name and Avatar" column, for example)

### Other information:

## Testing instructions:

Everything should function as before (or better). Mobile screen sizes are out of scope in this PR. Make sure to build CSS before testing.


gogdzl:
I pinged @ilonagl  to take a look at what we achieved here, but I have one thought:
It seems that other Jetpack products center their content, we should probably do the same.
Below Jetpack Social vs Jetpack CRM.

image

image

gogdzl:
About the code, I reviewed all files and everything seems to be in order, great cleanups!

gogdzl (changes requested):
See comments.

tbradsha:
> It seems that other Jetpack products center their content, we should probably do the same.

Yes, I considered it here but wanted a consistent holistic approach when implemented. I've laid the groundwork to a degree here to allow for it (chopping lots of the parent containers), but I think this should be a separate issue for global changes.

gogdzl (approved):
LGMT!
Nice chops!

PR #30862, opened 2023-05-23
Backup: add connect screen to secondary admins

Initsogar:
## Proposed changes:

* Add connect screen for secondary admins when the site has an active backup plan and has an owner connected.
  * After login, the secondary admin will be able to view the backup dashboard and storage details (if apply).

## Screenshots:
| Before | After |
|---|---|
| image | image |

## Known issues:
* The connect screen is not displaying the VaultPress Backup logo. It may require to modify [`ConnectScreen` component](https://github.com/Automattic/jetpack/blob/436e9244c303dc1bc7ac57bb258e4148998962e6/projects/js-packages/connection/components/connect-screen/basic/index.jsx#L26) to receive the logo, as [`ConnectScreenRequiredPlan`](https://github.com/Automattic/jetpack/blob/436e9244c303dc1bc7ac57bb258e4148998962e6/projects/js-packages/connection/components/connect-screen/required-plan/index.jsx#L14) does. It is better to address this in another PR given that we will be modifying another package.

### Other information:

## Jetpack product discussion

* peaFOp-FV-p2
* p6TEKc-6KA-p2

No

## Testing instructions:

* Spin up a Jurassic Ninja site with just the Jetpack VaultPress Backup plugin.
* Connect the site to Jetpack using your main WordPress.com account and add a Jetpack VaultPress Backup plan.
* Ensure you are able to see the VaultPress Backup dashboard (main page).
* Add a new user to your blog.
* In a incognito window (or a new browser), log in with the new user.
* Navigate to VaultPress Backup page
* You should see a connect screen similar to the one shared above.
* Log in using a different WordPress.com account.
* After login, you should see the backup dashboard as the first user.


projects/packages/backup/src/js/components/Admin/index.js
```diff
@@ -20,26 +20,56 @@ import Backups from '../Backups';
...
+	useSiteHasBackupProduct,
+} from './hooks';
 import NoBackupCapabilities from './no-backup-capabilities';
 import './style.scss';
 import '../masthead/masthead-style.scss';
 
 /* eslint react/react-in-jsx-scope: 0 */
 const Admin = () => {
-	const [ connectionStatus ] = useConnection();
+	const [ connectionStatus, , BackupSecondaryAdminConnectionScreen ] = useConnection();
```
Initsogar, line 34: I think it would be better to make `useConnection()` to return an object instead of an array. It might require some refactoring, so we could address this in another PR.

jwebbdev (approved):
Looks good as an improvement, so 👍 

There are a couple things of note that we may want to follow-up on.
1. The previous screen still displays while loading, so we might want to introduce some loading logic there.
2. After using the connect screen as the secondary user I was dropped into the purchase page, which clearly showed I already had VaultPress.  I'm not sure if that's something we need to handle in what we send / connect package or some other logic.

Screenshot 2023-05-23 at 4 39 09 PM

Initsogar:
@jwebbdev Thanks for the comments! I just added them to the backlog, so we can address them in a new iteration!

PR #30865, opened 2023-05-23
Photon: Fix adding query parameters to Photon URLs

### LINKED ISSUE #29770: Potential double question mark in photon URLs
As identified by @anomiex [here](https://github.com/Automattic/jetpack/pull/29561#pullrequestreview-1361595085), there appears to be a potential bug in Photon's `cdn_url` method [here](https://github.com/Automattic/jetpack/blob/1009e246529375ad28a3ed221eb02574ff24962f/projects/plugins/jetpack/functions.photon.php#L181-L192) and [here](https://github.com/Automattic/jetpack/pull/29561/files/286a2d8acfc892bbeae06b326401bd3259fe2195#diff-1f9bd3de530b97b7d6c7f7704fa26b34c5e5043532c9475eea843e96a540b918).

There are two ways a query parameter can be added to the URL; one via a filter (`jetpack_photon_add_query_string_to_domain`), and the other via `$args`. Neither path takes the other into consideration, leading to a possible pair of `?` symbols in the URL.

donnchawp:
As $photon_url can be manipulated by a filter and then by $args added on to it, there's a chance that "?" is added twice in the `cdn_url` function.

This PR fixes that by checking if "?" is already in the URL, and appending the new parameter with "&".

Fixes #29770

### Other information:

No

## Testing instructions:
I'm not sure how to test it, sorry.


jeherve (dismissed):
> I'm not sure how to test it, sorry.

It would be great if you could add a unit test for this, that'd be helpful I think.

donnchawp:
Added a test, you can run it with:
```
jetpack test packages/image-cdn php
```

haqadn (approved):
👍🏼

PR #30868, opened 2023-05-23
[Newsletters] Fix issue with Premium-content blocks on self-hosted

millerf:
Fixes https://github.com/Automattic/gold/issues/40

## Proposed changes:

* Allow Premium-content blocks on production and do not gate on self-hosted environments

## Jetpack product discussion
p1684310694056649-slack-C01B6KEJ5GE

No

## Testing instructions:
I am not really sure on how to test this...

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-newsletter-issue-with-self-hosted
+++ b/projects/plugins/jetpack/changelog/fix-newsletter-issue-with-self-hosted
@@ -1,0 +1,4 @@
+Significance: patch
+Type: compat
+
+Allow Premium content block to production and prevent potential issue on self-hosted installations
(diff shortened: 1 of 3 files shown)


jeherve (approved):

jeherve:
@millerf I'm afraid we'll need to do a few more changes for this block to be fully available on free self-hosted sites:

- To the stripe nudge: 
https://github.com/Automattic/jetpack/blob/436e9244c303dc1bc7ac57bb258e4148998962e6/projects/plugins/jetpack/extensions/blocks/premium-content/premium-content.php#L102-L103
- When registering the block:
https://github.com/Automattic/jetpack/blob/3a3c4adc1c0913bac224ddb832713bd6101fe1ab/projects/plugins/jetpack/extensions/blocks/premium-content/editor.js#L21

millerf:
@jeherve Thanks for the feedback. I'll follow up on another PR today if possible.

PR #30869, opened 2023-05-23
Add w.org assets

kraftbj:
Add all w.org assets, except for CRM, which already had them from a pre-existing setup. Moving them to match is saved for another PR after confirming how/if they still use the existing automation for theirs.

This would provide an easier way for product folks to pass screenshots to devs and provide a future route toward possible automation.

## Proposed changes:

* Add assets for jetpack

### Other information:

## Jetpack product discussion

p1HpG7-mAr-p2

no

## Testing instructions:

* make sure it doesn't pass to the production mirror repo.


.gitattributes
```diff
@@ -17,6 +17,7 @@
 .gitkeep      production-exclude
 
 # We don't need these files in any production mirror.
-.eslintrc.js  production-exclude
-jsconfig.json production-exclude
-tsconfig.json production-exclude
+.eslintrc.js   production-exclude
+jsconfig.json  production-exclude
+tsconfig.json  production-exclude
+.w.org-assets/ production-exclude
```
anomiex: Because of the way git's attributes work, we have to do it like this:
```suggestion
**/.w.org-assets/** production-exclude
```

anomiex (changes requested):

anomiex (approved):
Build assets look good

PR #30872, opened 2023-05-23
Fix links to the cache directory when a site isn't at "/"

donnchawp:
If a site doesn't live at /, then the debug log link and preload notification status update didn't work properly.

This PR fixes that by using get_home_path() to find the path to the homepage of the site.

### Other information:

No

## Testing instructions:
Install the plugin on a site hosted in a sub directory. Go to the debug settings page and see the link to the debug log is broken and has a duplicate of the directory name in the URL. The notification on the preload won't work either.
Apply the changes in this PR.
The debug log link will be fixed, and the preload notification works again.


dilirity (approved):
:shipit:

PR #30875, opened 2023-05-23
Jetpack: optimize gutenberg initialization

sergeymitr:
## Proposed changes:
Optimize initialization of the `Jetpack_Gutenberg` class:
- Get rid of the `init()` method, use on-demand extension loading instead.
- Hardcode block directories instead of using expensive `glob()` function.

Performance improvements on my machine (Monorepo docker env, wp-admin "All Posts" page with 4 posts):
[`Jetpack_Gutenberg` initialization](https://github.com/Automattic/jetpack/pull/30875/files#diff-028fe0b6a09c6ed2b0a61f21d8db4d21cf8ca35c8904f94f78503619455fe31fL685-L687) before the changes: ~0.22s
With the changes applied: ~0.08s

### Other information:

## Jetpack product discussion
p1HpG7-lOi-p2

No.

## Testing instructions:
1. Connect Jetpack
2. Go to "Settings" and enable VideoPress: `/wp-admin/admin.php?page=jetpack#/settings?term=videopress`
3. Edit or create a new post
4. Try using the "Calendly" block, make sure it shows up and works fine.
5. Try using the VideoPress block, confirm it's the new one (with the fancy icon, see screenshot), and it does NOT say `Enable Jetpack module to continue using VideoPress`
Screen Shot 2023-05-24 at 17 39 04


projects/plugins/jetpack/changelog/update-optimize-jetpack-gutenberg-init
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: other
+
+Optimize Jetpack_Gutenbert initialization.
```
anomiex:
```suggestion
Optimize Jetpack_Gutenberg initialization.
```

projects/plugins/jetpack/class.jetpack-gutenberg.php
```diff
@@ -209,45 +210,13 @@ public static function set_extension_unavailable( $slug, $reason, $details = arr
...
-		/**
-		 * Filter the list of block editor plugins that are available through Jetpack.
-		 *
-		 * @deprecated 7.0.0 Use jetpack_set_available_extensions instead
-		 *
-		 * @since 6.9.0
-		 *
-		 * @param array
-		 */
-		self::$extensions = apply_filters( 'jetpack_set_available_plugins', self::$extensions );
```
anomiex, line 250: There should be a change entry mentioning the removal of these deprecated filters (which would be of "major" significance, not that it matters since Jetpack doesn't use semver).

You could edit the existing entry or make a second one. Either is fine.

sergeymitr: Good point, [fixed](https://github.com/Automattic/jetpack/pull/30875/commits/41277236ea8e3d270d55463f8aba282b3836e436).

zinigor (approved):
Nice change, let's get it in before the code freeze!

PR #5922, opened 2016-12-16
Follow Button: disable and remove from SVN

dereksmart:
@gravityrail noticed a pretty significant performance issue with this Widget: 

> I noticed that the follow button was rendering pretty slowly, so I did a trace on the call to /v1/batch and found that one of the batched requests is /sites/http://foo.com, which actually calls through to the Jetpack site in order to retrieve the site info.
> 
> This is pretty bad from a performance point of view (adds an extra second or two on my Dreamhost blog).  The call is NOT cached by the site (why??) and generates a call from WPCOM to the site for every single pageload.

We're going to pull it from 4.5 and will enable again pending performance updates.

PR #30884, opened 2023-05-24
Follow Button Widget: add localized strings for iframe title

creativecoder:
## Proposed changes:

Adds localized strings to follow button widget iframe. Replicates change in D110939-code

Note that this feature does not appear to be active in Jetpack.

### Other information:

## Jetpack product discussion

See pdqkMK-12A-p2

No.

## Testing instructions:

- Connect site to Jetpack, if needed, and enable the "Extra Sidebar Widgets" module at `/wp-admin/admin.php?page=jetpack_modules`
- Uncomment `add_action( 'widgets_init', 'follow_button_register_widget' );` on line 6 of `projects/plugins/jetpack/modules/widgets/follow-button.php`
- Switch to a legacy theme with widget areas that supports the Customizer, like Dara
- In the Customizer or Widget editor, edit a widget area, and insert a Follow Button block
- Load a page on the site that shows the widget areas
- You should see the a `title` attribute on the iframe for the follow button
- window.wpcomPlatform object should be present on the JS global scope with strings for iframe titles.

### DIFF
--- a/projects/plugins/jetpack/changelog/add-wpcom-platform-js-localized-strings
+++ b/projects/plugins/jetpack/changelog/add-wpcom-platform-js-localized-strings
@@ -1,0 +1,5 @@
+Significance: patch
+Type: other
+Comment: Minor update to keep in sync with wpcom, feature not active
+
+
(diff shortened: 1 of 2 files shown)


jeherve (approved):
> Replicates change in D110939-code
> Note that this feature does not appear to be active in Jetpack.

Indeed. We disabled it in #5922 because of performance issues. This is also why the widget isn't used as is on WordPress.com, and why this kind of manual port is still necessary.

Maybe we should remove it altogether, to be honest. :)

PR #30888, opened 2023-05-24
Social: Fix the check for Social Image Generator for the media picker

pablinos:
The check for Social Image Generator was checking the current post config twice, rather than the config for the post and the feature availability, and so the media picker was never being disabled.

## Proposed changes:

This swaps the `isSocialImageGeneratorEnabled` prop for `isSocialImageGeneratorAvailable` so we check for the feature availability

### Other information:

No

## Testing instructions:

* With a site with the advanced plan, enable SIG in the sidebar
* With this change the media picker should be disabled with a notice. Without it, nothing will happen with the media picker.


manzoorwanijk:
What happens if a user already had a media added and then enabled SIG?

Screenshot 2023-05-24 at 12 17 36 PM

The remove (X) icon on the media is not clickable.

pablinos:
> What happens if a user already had a media added and then enabled SIG?

I think that's fine. We're ignoring that image in those circumstances, which is why it gets greyed out.

manzoorwanijk (dismissed):
Works as expected

spsiddarthan (dismissed):
This works well for me! Tested it both on Social and Jetpack.

spsiddarthan (approved):

PR #30863, opened 2023-05-23
Boost: Fix Minify concat not working when WP is running from a subdirectory on the server

dilirity:
## Proposed changes:

* Fixes JS an CSS concatenation not working on nested WP instances.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

n/a

## Does this pull request change what data or activity we track or use?

no

### Test that it breaks
* Setup the version from `trunk` on a WordPress site, running in a sub-directory on your server;
* Enable JS or CSS concatenation;
* Open the front-page and you should see 404 for concatenated URLs (they contain `_static`).

### Test that it works
* Setup the version from this PR on a WordPress site, running in a sub-directory on your server;
* Enable JS or CSS concatenation;
* Open the front-page and there should be no 404 for concatenated URLs.

PR #30893, opened 2023-05-24
[Boost] Fix minify paths on sites in a subdir.

thingalon:
Related: pedxs5-eK-p2#comment-114

When minifying and concatenating css on a site in a subdirectory, URLs weren't being properly interpreted. URLs like `/wp-includes/*` are relative to WordPress' home URL, and were being treated as relative to the root path of the domain.

#30863 went some way towards fixing that, but only fixed it for JS URLs. This PR updates the fix to apply to both JS and CSS URLs.

## Proposed changes:
* Ensure that enqueued script and css URLs are treated as relative to WordPress home

### Other information:

## Jetpack product discussion
n/a

no

## Testing instructions:

* Use a WordPress site on a subdirectory on any domain.
* Ensure that CSS and JS are properly minfied and concatenated.

### DIFF
--- a/projects/plugins/boost/app/lib/minify/functions-helpers.php
+++ b/projects/plugins/boost/app/lib/minify/functions-helpers.php
@@ -70,6 +70,23 @@ function jetpack_boost_page_optimize_uninstall() {
 }
 
 /**
+ * Convert enqueued home-relative URLs to absolute ones.
+ *
+ * Enqueued script URLs which start with / are relative to WordPress' home URL.
+ * i.e.: "/wp-includes/x.js" should be "WP_HOME/wp-includes/x.js".
+ *
+ * Note: this method uses home_url, so should only be used plugin-side when
+ * generating concatenated URLs.
+ */
+function enqueued_to_absolute_url( $url ) {
+	if ( substr( $url, 0, 1 ) === '/' ) {
+		return home_url( $url );
+	}
+
+	return $url;
+}
+
+/**
  * Get the list of JS slugs to exclude from minification.
  */
 function jetpack_boost_page_optimize_js_exclude_list() {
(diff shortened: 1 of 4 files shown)


dilirity (dismissed):
Good catch! Thanks for fixing this. Code looks great.

:shipit:

projects/plugins/boost/app/lib/minify/functions-helpers.php
```diff
@@ -69,6 +69,23 @@ function jetpack_boost_page_optimize_uninstall() {
...
+/**
+ * Convert enqueued home-relative URLs to absolute ones.
+ *
+ * Enqueued script URLs which start with / are relative to WordPress' home URL.
+ * i.e.: "/wp-includes/x.js" should be "WP_HOME/wp-includes/x.js".
+ *
+ * Note: this method uses home_url, so should only be used plugin-side when
+ * generating concatenated URLs.
+ */
+function enqueued_to_absolute_url( $url ) {
```
brandonpayton: Hi @thingalon, it seems like it would be good to prefix this function name with `jetpack_boost_` to reduce the possibility of naming collisions.

thingalon: Good point, thank you. Fixed in e0b3f355af.

dilirity: 🤦🏻 I missed that. I thought they were namespaced.

brandonpayton (commented):
I haven't tested this directly, but these changes look reasonable to me.

projects/plugins/boost/app/lib/minify/functions-helpers.php
brandonpayton: It's an understandable assumption. :) Namespaces seem common in Jetpack, but the team that created the bits from Page Optimize heavily favors function prefixing over dealing with namespaces.

dilirity (approved):

thingalon:
Cherry-picked to 1.9.0 in fb26b1452f

PR #30894, opened 2023-05-24
CRM: Removing jetpack-forms as a required dependency in composer.json

coder-karen:
## Proposed changes:

* This PR removes jetpack-forms as a required dependency in `composer.json`, which was previously added here - https://github.com/Automattic/jetpack/pull/30749 - in order to fix a Jetpack form compatibility issue with CRM, on WoA and WPcom sites. As there has been a WoA / WPcom release since then, we now no longer need to have the dependency in our `composer.json` file.

### Other information:

## Jetpack product discussion

https://github.com/Automattic/zero-bs-crm/issues/3093

No.

## Testing instructions:

* Proof-read, make sure the dependency has been removed. There's not much else we could test here.


tbradsha (approved):
Thank you!

PR #30902, opened 2023-05-24
Fix: generic user_id user meta value used in get_author function

zaerl:
See https://github.com/Automattic/dotcom-forge/issues/2511 for additional information.

Fixes #https://github.com/Automattic/dotcom-forge/issues/2511

## Proposed changes:

* Add a `$author instanceof WP_User || isset( $author->user_email )` check is moved as the first one in the `get_author` function.

### Other information:

## Testing instructions:

* Go to `http://calypso.localhost:3000/people/team/SITE` and add `danielbachhuber` as a subscriber or a user that has a `user_id: a valid WPCOM user ID` entry in `wp_usermeta`
* Make sure that the user of the list is the correct one.

### DIFF
--- a/projects/plugins/jetpack/tests/php/json-api/test_class.json-api-jetpack-base-endpoint.php
+++ b/projects/plugins/jetpack/tests/php/json-api/test_class.json-api-jetpack-base-endpoint.php
@@ -1,0 +1,131 @@
+<?php
+/**
+ * Jetpack_JSON_API_Endpoint class unit tests.
+ * Run this test with command: jetpack docker phpunit -- --filter=WP_Test_Jetpack_Base_Json_Api_Endpoints
+ *
+ * @package automattic/jetpack
+ * @phpcs:disable Generic.Files.OneObjectStructurePerFile.MultipleFound
+ */
+
+require_once JETPACK__PLUGIN_DIR . 'class.json-api-endpoints.php';
+
+/**
+ * Generic tests for Jetpack_JSON_API_Endpoint.
+ */
+class WP_Test_Jetpack_Base_Json_Api_Endpoints extends WP_UnitTestCase {
+	/**
+	 * A super admin user used for test.
+	 *
+	 * @var int
+	 */
+	private static $super_admin_user_id;
+
+	/**
+	 * Alternative super admin user used for test.
+	 *
+	 * @var int
+	 */
+	private static $super_admin_alt_user_id;
+
+	/**
+	 * Inserts globals needed to initialize the endpoint.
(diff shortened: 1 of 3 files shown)


danielbachhuber (commented):
I can verify that I now appear when I invite myself to a site.

It'd be great to have test coverage on this change! Tests will help make the behavior more explicit, so it doesn't accidentally break in the future.

zaerl:
@danielbachhuber I have added some tests for the `get_author` method that cover this. There wasn't a specific class used for tests of `Jetpack_JSON_API_Endpoint`, so I created a new one. Maybe in future other tests should be put there.

projects/plugins/jetpack/tests/php/json-api/test_class.json-api-jetpack-base-endpoint.php
```diff
@@ -0,0 +1,131 @@
...
+		$this->assertIsObject( $author );
+		$this->assertSame( self::$super_admin_user_id, $author->ID );
+
+		$author = $endpoint->get_author( $user );
+
+		// The user should be the same as the one passed as object to the method.
+		$this->assertIsObject( $author );
+		$this->assertSame( self::$super_admin_user_id, $author->ID );
+
+		delete_user_meta( self::$super_admin_user_id, 'user_id' );
```
danielbachhuber: Should we do this on `tearDown()` so it still runs if the test fails?

danielbachhuber (commented):
Thanks for including the tests!

I think it'd be good to get an opinion from @aerych , @kraftbj, @jeherve, @dereksmart or someone from Jetpack who might be able to verify that this is a safe change to make.

projects/plugins/jetpack/tests/php/json-api/test_class.json-api-jetpack-base-endpoint.php
zaerl: Sure, [done](https://github.com/Automattic/jetpack/pull/30902/files#diff-1fbe386e12e1492c039e58b9ac033f3d8996ed25c799749aa380ccb1f267b189R49-R57).

ouikhuan (approved):
LGTM!

projects/plugins/jetpack/tests/php/json-api/test_class.json-api-jetpack-base-endpoint.php
jeherve, line 1: Thanks for adding those tests!

jeherve (approved):
This looks good to me. 🚢

PR #30909, opened 2023-05-24
 AI Assistant: Change loading icon with Spinner component

### LINKED ISSUE #30716: Ai Assistant: revisit spinner and double check with dark backgrounds
Image

dhasilva:
Fixes #30716

## Proposed changes:

* Changes the loading icon with the Spinner component

The spinner can blend with the background depending on the theme. Since themes do not use CSS variables in a consistent way and we can't use fallback values because of [this](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/jetpack/tools/postcss.config.js#L14), there is not much we can do about that other than setting a background color to the input regardless of theme.

### Other information:

## Testing instructions:

* Add an AI Assistant block and make a request
* Check that the spinner is displayed while the request is underway


projects/plugins/jetpack/extensions/blocks/ai-assistant/editor.scss
```diff
@@ -47,6 +47,16 @@
 	.jetpack-ai-assistant__input-loader-icon {
 		color: var( --wp--preset--color--foreground );
```
retrofox, line 48: <3

retrofox (approved):
LGTM, even in dark themes 😅 

https://github.com/Automattic/jetpack/assets/77539/2860c6dd-06ad-49bf-ba01-6e3b22dbc612

projects/plugins/jetpack/extensions/blocks/ai-assistant/editor.scss
```diff
@@ -47,6 +47,16 @@
...
 	}
+	
+	.jetpack-ai-assistant__input-spinner {
+		margin: 0;
+		width: 24px;
+		height: 24px;
```
renatoagds, line 54: Don't know which is better. But we have the possibility to add the height and width from the props. Not?

renatoagds (approved):

renatoagds:
LGTM

projects/plugins/jetpack/extensions/blocks/ai-assistant/editor.scss
dhasilva: There are no props for this. On Storybook there is only one example, and it uses `style` to set `width` and `height`: https://wordpress.github.io/gutenberg/?path=/story/components-spinner--custom-size

renatoagds: Looking at the code in Gutenberg repo. It has, since it's applied here: https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/spinner/index.tsx

PR #30911, opened 2023-05-24
AI Assistant: handle network errors

### LINKED ISSUE #30867: AI Assistant: when an error is returned, show the try again option
Current behavior:

https://github.com/Automattic/jetpack/assets/16329583/4d78e361-ed8b-4045-a5f0-8c4af027010f

We need to show that the request failed

lhkowalski:
Fixes #30867.

## Proposed changes:

* Add `error_network` event on the `SuggestionsEventSource` class, so anytime the connection to the event source fails, this event gets fired
* Handle the `error_network` event on the suggestions hook to set a friendly message to the user, as well as show the retry option

### Other information:

## Testing instructions:

* This is tricky to test because we need to force a backend error
* The simplest way to do this is to sandbox your `public-api` and add the following filter on your `0-sandbox.php` file:

```
<?php

function force_disable_openai( $result, $blog_id ) {
	return false;
}

add_filter( 'openai_blog_active', 'force_disable_openai', 10, 2 );
```

* The snippet will disable OpenAI for any request on your sandboxed `public-api`
* While sandboxing the API, build this change and go to the block editor
* Open your browser console to inspect the network activity
* Add an AI Assistant block and run some prompt (example: "10 words related to open source software")
* Watch the network activity and confirm that the request to `/jetpack-ai-query` is failing and returning a `500` response (this is what the snippet above forced the API to do)

Screen Shot 2023-05-24 at 15 22 24

* On the block:
   * notice that a friendly error message shows up
   * notice that the Retry option shows up

Screen Shot 2023-05-24 at 15 22 51

* After testing, don't forget to remove the snippet from your `0-sandbox.php` file


projects/plugins/jetpack/extensions/blocks/ai-assistant/get-suggestion-with-stream.js
```diff
@@ -182,4 +184,11 @@ export class SuggestionsEventSource extends EventSource {
 			}
 		}
 	}
+
+	processErrorEvent( e ) {
+		debug( e );
+
+		// Dispatch a generic network error event
+		this.dispatchEvent( new CustomEvent( 'error_network' ) );
```
retrofox, line 192: Super simple silly question. Should it be `network_error` instead?

retrofox (approved):
LGTM

PR #30913, opened 2023-05-24
Launchpad: Add "is enabled" callback support to the task list definition

### LINKED ISSUE #77270: Launchpad: Add callback logic to determine whether a task list is enabled for a site
Add a new `is_enabled` logic to the launchpad endpoint to determine whether the task list is enabled for a site. This should replace the current launchpad_screen value to simplify it. The current launchpad_screen value is a bit confusing.

It will be simpler to avoid "context" for v1 and require callers to specify a task list, as general-purpose "context" support requires more complex algorithms and prioritization.

We should aim to implement this functionality only for the post-launch task list rather than replace the current behavior on the stepper Launchpad that uses launchpad_screen.

valterlorran:
Closes https://github.com/Automattic/wp-calypso/issues/77270

## Proposed changes:

* Add a new `is_enabled` logic to the launchpad endpoint to determine whether the task list is enabled for a site. We should have more flexibility and control over the task list to better define its availability throughout the WPCOM.
* We are aiming to add a new callback called `is_enabled_callback` to the task list definition.

### Other information:

## Testing instructions:

* Run `bin/jetpack-downloader test jetpack-mu-wpcom-plugin add/launchpad-is-enabled-callback` to build and sync this branch to your sandbox.
* Sandbox the public API
* Navigate to https://developer.wordpress.com/docs/api/console/
* Update the two toggles in the top left to `WP REST API` and `wpcom/v2`, but leave the `HTTP` action as `GET`
* Make some requests to the existing task lists to ensure it continue working as expected: `/sites/vlorran45.wordpress.com/launchpad?checklist_slug={TASK_LIST}`
  * link-in-bio
  * build
  * free
  * link-in-bio-tld
* Now, edit the `launchpad.php` file and add the `is_enabled_callback` to one of the task lists:
```php
wpcom_register_launchpad_task_list(
	array(
		'id'       => 'build',
		'title'    => 'Build',
		'is_enabled_callback' => '__return_false',
		'task_ids' => array(
			'setup_general',
			'design_selected',
			'first_post_published',
			'design_edited',
			'site_launched',
		),
	)
);
```
* Make a request again to the task list and verify that the `is_enabled` key is set to `false` in the response.

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
@@ -877,6 +877,29 @@ function wpcom_log_launchpad_being_enabled_for_test_sites( $option, $value ) {
 	);
 }
 
+/**
+ * Checks if a specific launchpad task list is enabled or if the overall launchpad is enabled.
+ *
+ * @param string|false $checklist_slug The slug of the launchpad task list to check.
+ * @return bool True if the task list is enabled, false otherwise.
+ */
+function wpcom_get_launchpad_task_list_is_enabled( $checklist_slug = false ) {
+	// If the checklist slug is provided, check the status of the task list.
+	if ( false !== $checklist_slug ) {
+		$is_enabled = wpcom_launchpad_checklists()->is_task_list_enabled( $checklist_slug );
+
+		// If the status of the task list is known, return its value.
+		if ( null !== $is_enabled ) {
+			return $is_enabled;
+		}
+	}
+
+	// If the status of the task list is not known, check the status of the overall launchpad.
+	// @todo: Remove this fallback once all task lists have been migrated to the new system.
+	// https://github.com/Automattic/wp-calypso/issues/77407
+	return wpcom_launchpad_checklists()->is_launchpad_enabled();
+}
+
 // Unhook our old mu-plugin - this current file is being loaded on 0 priority for `plugins_loaded`.
 if ( class_exists( 'WPCOM_Launchpad' ) ) {
 	remove_action( 'plugins_loaded', array( WPCOM_Launchpad::get_instance(), 'init' ) );
--- a/projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php
@@ -122,7 +122,7 @@ class WPCOM_REST_API_V2_Endpoint_Launchpad extends WP_REST_Controller {
 			'launchpad_screen'   => get_option( 'launchpad_screen' ),
 			'checklist_statuses' => get_option( 'launchpad_checklist_tasks_statuses', array() ),
 			'checklist'          => wpcom_get_launchpad_checklist_by_checklist_slug( $checklist_slug ),
-			'is_enabled'         => wpcom_launchpad_checklists()->is_launchpad_enabled(),
+			'is_enabled'         => wpcom_get_launchpad_task_list_is_enabled( $checklist_slug ),
 		);
 	}
 
--- a/projects/packages/jetpack-mu-wpcom/changelog/add-launchpad-is-enabled-callback
+++ b/projects/packages/jetpack-mu-wpcom/changelog/add-launchpad-is-enabled-callback
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Add a new is_enabled logic to the launchpad endpoint to determine whether the task list is enabled for a site.
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
@@ -129,6 +129,18 @@ class Launchpad_Task_Lists {
 	}
 
 	/**
+	 * Check if a task list is enabled by checking its is_enabled_callback callback.
+	 *
+	 * @param string $id Task List id.
+	 * @return bool|null True if enabled, false if not, null if not found.
+	 */
+	public function is_task_list_enabled( $id ) {
+		$task_list = $this->get_task_list( $id );
+
+		return $this->load_value_from_callback( $task_list, 'is_enabled_callback', null );
+	}
+
+	/**
 	 * Get all registered Launchpad Task Lists.
 	 *
 	 * @return array All registered Launchpad Task Lists.


projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
```diff
@@ -812,6 +812,21 @@ function wpcom_log_launchpad_being_enabled_for_test_sites( $option, $value ) {
 	);
 }
 
+/**
+ * Returns whether or not the Launchpad is enabled for a given checklist.
+ *
+ * @param string $checklist_slug The checklist slug.
+ * @return bool True if the Launchpad is enabled, false otherwise.
+ */
+function wpcom_get_launchpad_is_enabled( $checklist_slug ) {
```
sixhours: Maybe `wpcom_get_launchpad_task_list_is_enabled` for clarity?

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
```diff
@@ -812,6 +812,21 @@ function wpcom_log_launchpad_being_enabled_for_test_sites( $option, $value ) {
 	);
 }
 
+/**
+ * Returns whether or not the Launchpad is enabled for a given checklist.
+ *
+ * @param string $checklist_slug The checklist slug.
+ * @return bool True if the Launchpad is enabled, false otherwise.
```
sixhours: I'm not sure this docblock is accurate the way the code is written. If we don't have an `is_enabled` status for the checklist, we're falling back to `is_launchpad_enabled()`, right? And that checks the value of `launchpad_screen` and is not tied to the provided checklist slug. 

So I wonder if we need to decouple this from `is_launchpad_enabled()` entirely.

projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php
```diff
@@ -122,7 +122,7 @@ public function get_data( $request ) {
 			'launchpad_screen'   => get_option( 'launchpad_screen' ),
 			'checklist_statuses' => get_option( 'launchpad_checklist_tasks_statuses', array() ),
 			'checklist'          => wpcom_get_launchpad_checklist_by_checklist_slug( $checklist_slug ),
-			'is_enabled'         => wpcom_launchpad_checklists()->is_launchpad_enabled(),
+			'is_enabled'         => wpcom_get_launchpad_is_enabled( $checklist_slug ),
```
sixhours: If we change this now, in order to avoid breaking production, do we need to add the `is_enabled_callback` to all the active task lists? Or is everything in production still looking directly at `launchpad_screen`?

valterlorran: We won't need to add the `is_enabled_callback` to the task lists now, as `wpcom_get_launchpad_task_list_is_enabled` only uses the callback if it's defined(and returns something different from null). Otherwise, it will fall back to `wpcom_launchpad_checklists()->is_launchpad_enabled()`, which uses the `launchpad_screen` as you mentioned.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
```diff
@@ -812,6 +812,21 @@ function wpcom_log_launchpad_being_enabled_for_test_sites( $option, $value ) {
...
+ */
+function wpcom_get_launchpad_task_list_is_enabled( $checklist_slug ) {
+	$is_enabled = wpcom_launchpad_checklists()->is_task_list_enabled( $checklist_slug );
+	if ( null !== $is_enabled ) {
+		return $is_enabled;
+	}
+
+	return wpcom_launchpad_checklists()->is_launchpad_enabled();
```
sixhours, line 900: Do you envision removing the fallback to `wpcom_launchpad_checklists()->is_launchpad_enabled()` here once we've refactored all the task lists to use `is_enabled_callback`? If so, maybe adding a note or `@todo` to that effect here would reduce confusion, because I can see us getting tripped up. They have similar names but map to two different values, if that makes sense.

projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php
sixhours: Gotcha, the fallback was tripping me up. 😅 

This works for now, but to keep this function tight, let's plan to refactor the other task lists to use `is_enabled_callback` -- they can point directly to `wpcom_launchpad_checklists()->is_launchpad_enabled()` and remove the need for a fallback in this function. LMK if that makes sense!

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
valterlorran: Yes, I should've made it clear. But I have also been wondering what "is enabled" means because if you look at the `is_launchpad_enabled` definition, we are also checking if the task list has incomplete tasks:
```php
public function is_launchpad_enabled() {
	$launchpad_screen = get_option( 'launchpad_screen' );
	if ( 'full' !== $launchpad_screen ) {
		return false;
	}

	return $this->has_active_tasks();
}
```
Maybe we should also have an `is_complete` state.

valterlorran: Another thought: we could add a new callback that uses the `is_launchpad_enabled` method and add the `is_enabled_callback` to all existing task lists and make it mandatory on the definition. What do you think?

sixhours: > But I have also been wondering what "is enabled" means because if you look at the is_launchpad_enabled definition, we are also checking if the task list has incomplete tasks:

Good point, I think there's been some work done to update that logic in https://github.com/Automattic/jetpack/pull/30624 but it hasn't been reviewed/shipped. We should probably take that over and get it shipped.

projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php
valterlorran: Yes, totally makes sense. I will create a new issue to tackle this refactor.

valterlorran: Here we go: https://github.com/Automattic/wp-calypso/issues/77407

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
sixhours: > Another thought: we could add a new callback that uses the is_launchpad_enabled method and add the is_enabled_callback to all existing task lists and make it mandatory on the definition. What do you think?

Do you mean making `is_launchpad_enabled` the default callback for all task lists unless we specify otherwise? That works!

valterlorran: I meant the same thing you said here https://github.com/Automattic/jetpack/pull/30913#discussion_r1205846448, I only saw it afterwards 😆  
I describe my idea here: https://github.com/Automattic/wp-calypso/issues/77407

sixhours (dismissed):
This is working for me! 👍

projects/plugins/mu-wpcom-plugin/changelog/add-launchpad-is-enabled-callback
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: added
+
+Fix mu-wpcom-plugin lint error
```
sixhours: Can you separate this fix out into a fresh PR? This one is already big and it's confusing to have two changelogs for one PR.

sixhours (changes requested):

sixhours (approved):
:shipit:

PR #30915, opened 2023-05-24
Fix translation for launchpad tasks

daledupreez:
## Proposed changes:

* This PR updates the Launchpad task code to use a `get_title` callback to retrieve task titles, as using inline calls to `__()` would run the translation before the user's locale had been set. More specifically, the PR includes the following specific changes:
   - We now support a new `get_title` callable attribute for task definitions, in addition to the existing `title` attribute, but we prefer the `get_title` callback. This also involved creating a new `load_title()` helper and updating the task validation.
   - Existing launchpad tasks in the package have all been updated to use the new `get_title` approach.

This work was sparked by the following internal discussion and exploration: p1684883448623069-slack-C057AH42XQD

**Note:** The primary APIs in use for the checklist are via our general `public-api.wordpress.com` REST APIs, so we want to make sure we address that in this PR. We also verify that the API works when accessed directly, but that's not our primary use case.

### Other information:

## Jetpack product discussion

This is primarily a WordPress.com change

No.

### For WordPress.com Simple sites

* Apply this patch to your WordPress.com developer sandbox, and ensure requests to `public-api.wordpress.com` are routed to your sandbox
* Open the [WordPress.com developer console](https://developer.wordpress.com/docs/api/console/)
* Update the two toggles in the top left to `WP REST API` and `wpcom/v2`, but leave the HTTP action as `GET`
* In the main text/URL input, type in `/sites/:siteSlug/launchpad?checklist_slug=newsletter&_locale=es`
* Submit the request
* Verify that the items in the `checklist` key have translated `title` values

### For WordPress.com Atomic sites

* Set up a WoA dev site without this patch applied
* While on that site, visit `/wp-admin/users.php`
* Select your user and click on the "Edit" link
* Scroll down and create a new application password for your user -- keep the new password somewhere accessible
* From a local terminal, run the following `curl` command to confirm that you can get the checklist data:
```
curl --user "yourusername:your app password" https://your-site-url.wpcomstaging.com/wp-json/wpcom/v2/launchpad?checklist_slug=newsletter&_locale=es
```
* Verify that you get a response back that is untranslated.
* Apply the changed files from this PR to `wp-content/mu-plugins/vendor/automattic/jetpack-mu-wpcom/src/features/launchpad/` - you can use `rsync` or `sftp` to do so
* Re-run the `curl` command from your local terminal:
```
curl --user "yourusername:your app password" https://your-site-url.wpcomstaging.com/wp-json/wpcom/v2/launchpad?checklist_slug=newsletter&_locale=es
```
* Verify that you get _un-translated_ content back successfully -- we don't include the translated strings AFAIK, so we're testing that we didn't break anything

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
@@ -213,9 +213,15 @@ class Launchpad_Task_Lists {
 	 */
 	private function build_task( $task ) {
 		$built_task                 = array(
-			'id'    => $task['id'],
-			'title' => $task['title'],
+			'id' => $task['id'],
 		);
+
+		$default_title = '';
+		if ( isset( $task['title'] ) ) {
+			$default_title = $task['title'];
+		}
+
+		$built_task['title']        = $this->load_value_from_callback( $task, 'get_title', $default_title );
 		$built_task['completed']    = $this->is_task_complete( $task );
 		$built_task['disabled']     = $this->is_task_disabled( $task );
 		$built_task['subtitle']     = $this->load_subtitle( $task );
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
@@ -212,16 +212,11 @@ class Launchpad_Task_Lists {
 	 * @return Task Task with current state.
 	 */
 	private function build_task( $task ) {
-		$built_task                 = array(
+		$built_task = array(
 			'id' => $task['id'],
 		);
 
-		$default_title = '';
-		if ( isset( $task['title'] ) ) {
-			$default_title = $task['title'];
-		}
-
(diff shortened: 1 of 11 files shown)


projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -240,6 +241,26 @@ private function load_value_from_callback( $task, $callback, $default = '' ) {
 		return $default;
 	}
 
+	/**
```
sixhours, line 244: Minor, but it might be a good idea to explain in the docblock why we're using a callback instead of a string.

sixhours (commented):
This works for me so far!

I think you can `define( 'JETPACK__WPCOM_JSON_API_BASE', 'your.sandbox.url' );` in the Atomic site's `wp-config.php` to sandbox public-api from an WoA dev Atomic site for testing.

valterlorran:
This is working for me as well!
Screen Shot 2023-05-24 at 16 52 49

sixhours:
Tested this again for Atomic and I still get back untranslated content on Atomic, as expected:

Screen Shot 2023-05-30 at 11 51 37 AM

sixhours (dismissed):

sixhours (approved):
Still looks good!

PR #30921, opened 2023-05-24
Update pnpm to v8

anomiex:
## Proposed changes:

Mainly because people have been getting confused when installing the latest pnpm and then it doesn't work.

But 8.3.0 does bring a useful new flag to `pnpm install` that we can use to have CI better avoid people accidentally committing broken peer dependencies.

Also we need a few config changes to maintain expected behavior as v8 made some questionable changes to the defaults.

### Other information:

## Jetpack product discussion

p9dueE-7mw-p2#comment-9725

No

## Testing instructions:

* Everything should continue to work as it did before. There should be no changes to build artifacts.
* If you create a PR reverting d9fa0377bc960c8238ee7f0f5326db7475afe004, CI should catch the problem now.
  * [x] https://github.com/Automattic/jetpack/actions/runs/5073263141/jobs/9112072281#step:4:1834


zinigor (dismissed):
Both the install and the build succeed for me, LGTM!

zinigor (approved):

PR #30914, opened 2023-05-24
Add Jetpack Boost Speed Score package

IanRamosC:
Fixes 1201210512993648-as-1204651717398079

## Proposed changes:

* Add a new package `boost-speed-score` using `jetpack generate`, and respective tests. This package is responsible for fetching the API to calculate the speed score of a site. This package was originally part of the Jetpack Boost plugin, and this PR moves it to its own package. The idea is to reuse this package both on the Jetpack Dashboard, and the Boost plugin itself. Another PR will be created to that on the Boost plugin side after this PR is merged.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

p1HpG7-k9o-p2

## Does this pull request change what data or activity we track or use?

No

## Testing instructions:

* Checkout this branch locally
* Run `jetpack install packages/boost-speed-score` and `jetpack build packages/boost-speed-score`
* From the Boost plugin folder `cd projects/plugins/boost` run `composer require automattic/jetpack-boost-speed-score`
* Open `projects/plugins/boost/app/admin/class-admin.php` and change line 12 from `use Automattic\Jetpack_Boost\Features\Speed_Score\Speed_Score;` to `use Automattic\Jetpack\Boost_Speed_Score\Speed_Score;`
* Run `jetpack install plugins/boost && jetpack build plugins/boost`
* Go to `/wp-admin/admin.php?page=jetpack-boost` and confirm it loads the speed core as expected
* To run unit tests, go to `projects/packages/boost-speed-score` and run `composer test-php`

PR #30927, opened 2023-05-24
Update/boost plugin to use speed score package

IanRamosC:
Fixes 1201210512993648-as-1204651717398080

## Proposed changes:

* This PR udpates the Boost plugin to use the Boost Speed Score package added in #30914. I created this branched off of `add/jetpack-boost-speed-score-package` and I'm currently comparing this branch against trunk, but I'll rebase it as soon as #30914 is deployed.

### Other information:

## Jetpack product discussion

p1HpG7-k9o-p2

Nope

## Testing instructions:

* Checkout this branch locally
* Run `jetpack build packages/boost-speed-score` and `jetpack install packages/boost-speed-score`
* Run `jetpack build plugins/boost` and `jetpack install plugins/boost`
* Alternatively, create a JN (or an Atomic Ephemeral) site with Jetpack Beta installed and with `update/boost-plugin-to-use-speed-score-package` as the selected branch
* Visit `/wp-admin/admin.php?page=jetpack-boost` and make sure Boost works as expected


projects/packages/boost-speed-score/.phpcsignore
```diff
@@ -0,0 +1 @@
+src/class-example.php
```
robertsreberski: We don't have this file from what I see - maybe this rule is redundant?

robertsreberski (commented):
One minor comment, not a blocker

robertsreberski (dismissed):
The migration of code to standalone package looks good! For the future, these kind of changes could be broken down into several PRs instead - one that generates the package, second that moves the code to the package and updates references in Jetpack Boost library.

That way we would separate auto-generated code from the migration code making it easier to process.

CodeyGuyDylan:
> The migration of code to standalone package looks good! For the future, these kind of changes could be broken down into several PRs instead - one that generates the package, second that moves the code to the package and updates references in Jetpack Boost library.

This is actually in separate PR's, this PR contains code from https://github.com/Automattic/jetpack/pull/30914 which has not been merged yet. Once that is merged and this is rebased, the PR should be much smaller 😄

CodeyGuyDylan:
Had some trouble with the rebase but all should be good now. @robertsreberski if you wanted to look at this again, it should reflect more of what is actually changing now

thingalon (dismissed):
Seems to work, and the code changes seem ok

robertsreberski (dismissed):
Thanks @CodeyGuyDylan @IanRamosC , it looks clear now!

CodeyGuyDylan (approved):

PR #30928, opened 2023-05-25
Plugin Search Hints: avoid PHP warnings

jeherve:
## Proposed changes:

Some other plugins may play with the plugins_api_result hook and not return the kind of results we're expecting. Let's avoid warnings in this case.

```
PHP Warning:  array_filter() expects parameter 1 to be array, null given in /wp-content/plugins/jetpack/modules/plugin-search.php on line 409

PHP Warning:  array_unshift() expects parameter 1 to be arr
```

### Other information:

## Jetpack product discussion

* p1685003148609169-slack-CDLH4C1UZ

* No

## Testing instructions:

Test that feature hints still work:

* On a site that's connected to WordPress.com, go to Plugins > Add New
* Search for "cdn"
* See that Jetpack's Photon is suggested as the first result.


anomiex (dismissed):
Meh. This will work, although by [the documentation](https://developer.wordpress.org/reference/hooks/plugins_api_result/) the only thing we're supposed to have to check for is `$result instanceof WP_Error`. Which may have been the actual cause of the problem rather than some other plugin screwing up.

jeherve:
> This will work, although by [the documentation](https://developer.wordpress.org/reference/hooks/plugins_api_result/) the only thing we're supposed to have to check for is `$result instanceof WP_Error`. Which may have been the actual cause of the problem rather than some other plugin screwing up.

Good point. Added in 0c0c23f9d4cd0aeb3c7860cb8eaf638e5da77164

zinigor (approved):

anomiex (approved):

PR #30937, opened 2023-05-25
Social: Add account name to the publicize object

spsiddarthan:
Some services such as Instagram need both the username and the display name of the jetpack social connection. 

## Proposed changes:

* Added publicize account name to the social object. 
### Other information:

## Jetpack product discussion

p1685004756108959/1685004225.849949-slack-C02JJ910CNL

No 

## Testing instructions:

* Add a Facebook connection and an Instagram connection (see D111486-code if you haven't added a connection before. If you already have a connection, please make sure you disconnect and reconnect, since the meta data of the connection has been updated in the phab diff. 
* Console.log the connections in `projects/js-packages/publicize-components/src/hooks/use-social-media-connections/index.js` and confirm that the `account_name` field shows up.

### DIFF
--- a/projects/packages/publicize/src/class-connections-post-field.php
+++ b/projects/packages/publicize/src/class-connections-post-field.php
@@ -94,7 +94,7 @@ class Connections_Post_Field {
 					'readonly'    => true,
 				),
 				'display_name'    => array(
-					'description' => __( 'Usually username of the connected account', 'jetpack-publicize-pkg' ),
+					'description' => __( 'Display name of the connected account', 'jetpack-publicize-pkg' ),
 					'type'        => 'string',
 					'context'     => array( 'view', 'edit' ),
 					'readonly'    => true,
--- a/projects/packages/publicize/src/class-connections-post-field.php
+++ b/projects/packages/publicize/src/class-connections-post-field.php
@@ -94,6 +94,12 @@ class Connections_Post_Field {
 					'readonly'    => true,
 				),
 				'display_name'    => array(
+					'description' => __( 'Usually username of the connected account', 'jetpack-publicize-pkg' ),
+					'type'        => 'string',
+					'context'     => array( 'view', 'edit' ),
+					'readonly'    => true,
+				),
+				'account_name'    => array(
 					'description' => __( 'Username of the connected account', 'jetpack-publicize-pkg' ),
 					'type'        => 'string',
 					'context'     => array( 'view', 'edit' ),
--- a/projects/packages/publicize/src/class-connections-post-field.php
+++ b/projects/packages/publicize/src/class-connections-post-field.php
@@ -99,7 +99,7 @@ class Connections_Post_Field {
 					'context'     => array( 'view', 'edit' ),
 					'readonly'    => true,
 				),
-				'account_name'    => array(
+				'username'        => array(
 					'description' => __( 'Username of the connected account', 'jetpack-publicize-pkg' ),
 					'type'        => 'string',
 					'context'     => array( 'view', 'edit' ),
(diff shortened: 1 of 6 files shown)


projects/packages/publicize/src/class-connections-post-field.php
```diff
@@ -94,6 +94,12 @@ private function post_connection_schema() {
 					'readonly'    => true,
 				),
 				'display_name'    => array(
+					'description' => __( 'Usually username of the connected account', 'jetpack-publicize-pkg' ),
+					'type'        => 'string',
+					'context'     => array( 'view', 'edit' ),
+					'readonly'    => true,
+				),
+				'account_name'    => array(
```
manzoorwanijk: Can we use `username` here to be super explicit?

projects/packages/publicize/src/class-publicize-base.php
```diff
@@ -527,6 +527,23 @@ public function get_display_name( $service_name, $connection ) {
...
 	}
 
+	/**
+	 * Returns the account name for the Connection. For services like Instagram, we need both the username and the account's name.
+	 *
+	 * @param string       $service_name 'facebook', 'linkedin', etc.
+	 * @param object|array $connection The Connection object (WordPress.com) or array (Jetpack).
+	 * @return string
+	 */
+	public function get_account_display_name( $service_name, $connection ) {
```
pablinos: It might be more consistent to not mention `display` here, and if this is the username, then we should probably rename this to `get_username`

projects/plugins/jetpack/_inc/lib/core-api/wpcom-fields/post-fields-publicize-connections.php
```diff
@@ -120,6 +120,12 @@ private function post_connection_schema() {
 					'readonly'    => true,
 				),
 				'display_name'    => array(
+					'description' => __( 'Usually username of the connected account', 'jetpack' ),
```
pablinos: Is this right? Is it normally the username?

pablinos (commented):
Looks good. We just need to decide on what we're calling it!

projects/plugins/jetpack/_inc/lib/core-api/wpcom-fields/post-fields-publicize-connections.php
spsiddarthan: In the case of Facebook, it's the Facebook page's name, isn't it?

pablinos: Yeah, my point being that the display name is usually *not* the username

projects/packages/publicize/src/class-publicize-base.php
```diff
@@ -527,6 +527,23 @@ public function get_display_name( $service_name, $connection ) {
...
+	 * Returns the account name for the Connection. For services like Instagram, we need both the username and the account's name.
+	 *
+	 * @param string       $service_name 'facebook', 'linkedin', etc.
+	 * @param object|array $connection The Connection object (WordPress.com) or array (Jetpack).
+	 * @return string
+	 */
+	public function get_username( $service_name, $connection ) {
+		$cmeta = $this->get_connection_meta( $connection );
+
+		if ( 'instagram-business' === $service_name && isset( $cmeta['connection_data']['meta']['username'] ) ) {
```
pablinos: Shall we make this generic? If the `username` is set then we use that and otherwise fallback to the display name?

manzoorwanijk (approved):
This works great now. Thanks

PR #30940, opened 2023-05-25
[Newsletter] Create default membership product silently in newsletter context

millerf:
Fixes https://github.com/Automattic/gold/issues/49

## Proposed changes:

* Add a separate selector for the Newsletter products that resolves and silences the creation of the default product  

### Other information:

## Testing instructions:

* Create a new site
* Sandbox the site
* Go to "new post" page and open the newsletter 
* **You should not see** any snackbar saying "Successfully created payment plan"
Screenshot 2023-05-25 at 15 34 07

* Jest tests:
```
cd projects/plugins/jetpack && pnpm test-extensions extensions/store/membership-products/test
```
I tried to implement some resolver tests, but I failed to make them work as expected....


projects/plugins/jetpack/extensions/store/membership-products/actions.js
```diff
@@ -32,7 +32,8 @@ export const saveProduct =
 		product,
 		productType = PRODUCT_TYPE_PAYMENT_PLAN,
 		setSelectedProductId = () => {},
-		callback = () => {}
+		callback = () => {},
+		silentProductCreation = false
```
romarioraffington: What do you think about giving this a more descriptive name like `shouldDisplayProductCreationNotice`?

Same feedback for the vars and test cases in the other files as well

millerf: Good idea.Fixed in 2f6d1d26d6

projects/plugins/jetpack/extensions/store/membership-products/resolvers.js
```diff
@@ -81,11 +87,20 @@ const setDefaultProductIfNeeded = ( selectedProductId, setSelectedProductId, sel
 	}
 };
 
+export const getNewsletterProducts = (
+	productType = PRODUCT_TYPE_PAYMENT_PLAN,
+	selectedProductId = 0,
+	setSelectedProductId = () => {}
+) =>
+	// Returns the products, but silences the snack bar if a default product is created
+	getProducts( productType, selectedProductId, setSelectedProductId, false );
```
romarioraffington, line 97: Optional feedback here. Personally, I find params either to read verses having to look up the fn definition

```suggestion
       $shouldDisplayProductCreationNotice = false;
	getProducts( productType, selectedProductId, setSelectedProductId, shouldDisplayProductCreationNotice );
```

romarioraffington (dismissed):

romarioraffington:
Added an optional feedback. It also appears that a required test failing

millerf:
@jeherve I am unsure if those tests are failing because of this PR or if this is the "usual" tests failing...

jeherve:
> I am unsure if those tests are failing because of this PR or if this is the "usual" tests failing...

I can't tell since they're running at the moment, but if the tests are required, then they should not be flaky. If they're failing on your PR and if the failure is not related to your PR, then you would see the same test failing on all other recent PRs in the repo.
Let me know if they fail again after this new run!

millerf:
@jeherve YEs, they are still failing. I also tested on trunk: https://github.com/Automattic/jetpack/pull/31138

jeherve:
That "E2E Tests / Jetpack blocks e2e tests" check isn't required, so you should be able to merge even when they're failing.

They're a bit flaky, which is why we retry them when they run for the mirror copy of `trunk`. You can see the recent runs here:
https://a8c-jetpack-e2e-reports.s3.amazonaws.com/reports/jetpack-production/report/index.html

projects/plugins/jetpack/extensions/store/membership-products/resolvers.js
millerf: I personnally find it too noisy, when one can just look at the definition of `getProducts()` with one click.  But thanks for the suggestion.

millerf: I personnally find it too noisy, when one can just look at the definition of `getProducts()` with one click.  But thanks for the suggestion.

jeherve (commented):
I'm seeing this when testing this on a new WordPress.com simple site, after connecting my site to Stripe, when I'm landing back in the editor:

Screenshot 2023-06-02 at 12 45 43
Screenshot 2023-06-02 at 12 46 38

millerf:
@jeherve Nice catch, I am still wondering why I didn't see it before... 🤔 
I have fixed it (I missed an optional parameter)

https://github.com/Automattic/jetpack/assets/790558/09821c52-8352-41e2-b397-bc4de46ca6b6

anomiex:
I have to ask, what's going on with all these Friday PRs you're flagging for cherry-picking? Why did these not get done earlier in the week, or even better last week before the previously-announced code freeze? When I click through to the linked bugs they were known about back then.

At this point this stuff is going to get basically no testing in the beta before being released.

millerf:
> Why did these not get done earlier in the week, or even better last week before the previously-announced code freeze?

Because most of those were either discovered during beta tests, or were last-minute decisions.

anomiex:
> Because most of those were either discovered during beta tests, or were last-minute decisions.

This PR itself was created on May 25, but it didn't get reviewed until May 30 and still isn't merged. If something is created that late in the month and is wanted in the release, getting it reviewed and merged should have been a higher priority.

The others I've looked at seemed similar to me: an issue was identified fairly early on, but fixing it seemingly wasn't treated as a priority until the end of beta week.

millerf:
> This PR itself was created on May 25,

Yes, that is what I call beta test phase. 

And as you have seen, there were a couple of those PRs lying around. The ones you haven't seen, as were fixed before, had a bigger priority.

What I want to convey is that we have pushed the higher priority first, and the lesser (as this one) for after. 

Now, if this one cannot be cherry-picked, that is fine.

anomiex:
> Now, if this one cannot be cherry-picked, that is fine.

At this point I think that will be for the best. Thanks!

millerf:
@jeherve would you mind having a look at this? It has been stalled for a while... I just rebased.

jeherve (approved):
I believe that works. I've had some issues testing the payment flow behavior on WordPress.com Simple though, so I'd appreciate if someone from Gold could test too!

romarioraffington:
This is already merged but LGTM for the Gold side

PR #30941, opened 2023-05-25
AI Assistant: introduce first approach of default prompt templates

### LINKED ISSUE #30633: Ai Assistant: Add prompt templates that the user has to autocomplete
Whenever we have a post/page with no content, we show this:

Image

We should show templates for prompts. We can add options like:
Write a blog post about
Write a page about
Write a list

Once the user clicks on one of those options, it should add that text to the text field. The user will then need to complete it.

retrofox:
Fixes #30633

## Proposed changes:

* AI Assistant: introduce first approach of default prompt templates

### Other information:

## Testing instructions:

https://github.com/Automattic/jetpack/assets/77539/6204616a-b29a-41d4-a2e0-f84549f85837

https://github.com/Automattic/jetpack/assets/77539/deec321d-3f86-41d4-8d76-c10f7ef59256

### DIFF
--- a/projects/plugins/jetpack/changelog/update-ai-assistant-add-basic-tpls-in-block-toolbar
+++ b/projects/plugins/jetpack/changelog/update-ai-assistant-add-basic-tpls-in-block-toolbar
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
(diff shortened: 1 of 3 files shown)


dhasilva (approved):
LGTM :rocket:

PR #30944, opened 2023-05-25
Jetpack Forms: fix a typo in the template filter

donnchawp:
A filter on the template used for form responses has a small typo in it.
The filter is "jetpack_forms_respone_email_template" but it should be "jetpack_forms_response_email_template"

This PR fixes that, while maintaining the old filter, just in case someone has started using it already.

### Other information:

## Jetpack product discussion
pcsBup-yo-p2#comment-452

No

## Testing instructions:
Apply the changes in this PR.
Embed a contact form in a page and then fill it out.
There shouldn't be any visible changes or PHP errors if the patch works.

You could also use the filter to create a simple template for the response email.
Create a file called wp-content/jp-forms-template.php with the following content:
```
<?php

$template = '
<!doctype html>
<html xmlns="http://www.w3.org/1999/xhtml">
<body>
<!-- title -->
<h1>%1$s</h1>

<!-- response -->
<p>%2$s</p>

<!-- link to responses page -->
<p><a href="%3$s">View Responses</a></p>

<!-- link to edit form -->
<p><a href="%4$s">Edit</a></p>

<!-- footer -->
<p>%5$s</p>
</body>
</html>';
```
Then create an mu-plugin with the following code:
```
<?php

function jp_forms_template( $filename ) {
        return ABSPATH . '/wp-content/jp-forms-template.php';
}

add_filter( 'jetpack_forms_response_email_template', 'jp_forms_template' );
```

Now fill out the contact form and wait for the response email. It should show a basic email with only the response and some metadata.


projects/packages/forms/src/contact-form/class-contact-form.php
```diff
@@ -1692,7 +1692,10 @@ public static function wrap_message_in_html_tags( $title, $response_link, $form_
 		 *
 		 * @param string the filename of the HTML template used for response emails to the form owner.
 		 */
-		require apply_filters( 'jetpack_forms_respone_email_template', __DIR__ . '/templates/email-response.php' );
+		$template_filename = apply_filters( 'jetpack_forms_respone_email_template', __DIR__ . '/templates/email-response.php' );
```
jeherve: I would recommend marking the first filter as deprecated with `apply_filters_deprecated` instead, and having 2 docblocks, one for each filter.

jeherve (changes requested):

projects/packages/forms/src/contact-form/class-contact-form.php
donnchawp: ah, thanks.

donnchawp: The filter wasn't public yet, so I'll just fix it. It'll be public in the next version of the plugin.

CGastrell (dismissed):
Cool! Yeah, being a completely new/unused filter should spare you from the hotfix situation I had last week. 

:shipit:

projects/packages/forms/src/contact-form/class-contact-form.php
```diff
@@ -1688,11 +1688,11 @@ public static function wrap_message_in_html_tags( $title, $response_link, $form_
 		 *
 		 * @module contact-form
 		 *
-		 * @since 0.18.0
+		 * @since 12.2
```
jeherve: Since this is in a package, it uses the package version and not the plugin version. `0.18.0` is okay here.

```suggestion
		 * @since 0.18.0
```

jeherve (changes requested):

projects/packages/forms/src/contact-form/class-contact-form.php
donnchawp: oh, thanks

jeherve (approved):

donnchawp:
I'm going to recreate this PR. For some reason it wants me to update the CRM lock file, which I did on Thursday and landed in a world of trouble with "divergent branches", and I reverted that CRM change. Now it wants me to update it again.
I'll create another branch with this one line fix.

PR #30947, opened 2023-05-25
VideoPress: Add wpcom support to Ajax class

roundhill:
We are looking to share the same `class-ajax.php` code with Jetpack and wp.com. During testing I discovered that on wpcom it was trying to use the `shadow_blog_id` which isn't available on wp.com simple sites.

## Proposed changes:

* This PR adds a helper function to get the appropriate blog id depending on if we are on Jetpack or wp.com.

### Other information:

## Jetpack product discussion

pxWta-1P4-p2

No

## Testing instructions:

* From a Jetpack site, ensure VideoPress is enabled and upload a video in the block editor. It should succeed.
* Apply patch to wp.com (`bin/jetpackddownloader test jetpack update/videopress-ajax-wpcom-support`), apply VideoPress HQ and attempt to add a new video, you should see the edit fields which means the ajax call succeeded.


jgcaruso (dismissed):
Code looks good and tests as expected on WPCOM and Jetpack

projects/packages/videopress/src/class-ajax.php
```diff
@@ -283,4 +279,18 @@ public function wp_ajax_update_transcoding_status() {
...
+	 *
+	 * @return int the blog id
+	 */
+	public function get_videopress_blog_id() {
+		if ( defined( 'IS_WPCOM' ) && IS_WPCOM ) {
+			return get_current_blog_id();
+		} else {
+			$options = Options::get_options();
+			return $options['shadow_blog_id'];
+		}
```
jgcaruso: Since the first if statement returns, you could skip the `else` here and un-indent this code.

sergeymitr (approved):
👍

PR #30958, opened 2023-05-25
[Newsletter] Handle disabled module gracefully

millerf:
## Proposed changes:
* Show the panel to enable the module when the Subscription module is disabled.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
@@ -52,8 +52,13 @@ function NewsletterEditorSettingsPanel( {
 	socialFollowers,
 	emailSubscribers,
 	paidSubscribers,
+	isModuleActive,
 	showMisconfigurationWarning,
 } ) {
+	if ( ! isModuleActive ) {
+		return;
+	}
+
 	return (
 		<PluginDocumentSettingPanel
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
```diff
@@ -139,14 +144,16 @@ function NewsletterPrePublishSettingsPanel( {
...
 		>
-			<NewsletterAccessPrePublishSettings
-				accessLevel={ accessLevel }
-				setPostMeta={ setPostMeta }
-				socialFollowers={ socialFollowers }
-				emailSubscribers={ emailSubscribers }
-				paidSubscribers={ paidSubscribers }
-				showMisconfigurationWarning={ showMisconfigurationWarning }
-			/>
+			{ isModuleActive && (
```
romarioraffington, line 147: Not sure if we need `isModuleActive` check here when we have the check above.

jeherve: I think the GitHub UI makes this a bit confusing. The check above is for another component, but the GitHub UI collapsed lines 65 to 144, so we do not see that.

jeherve (approved):
This looks good and tests well for me. 👍 

Screenshot 2023-05-26 at 17 41 46
Screenshot 2023-05-26 at 17 41 28
Screenshot 2023-05-26 at 17 41 22

PR #30966, opened 2023-05-25
ci: Bump time limit for build

anomiex:
## Proposed changes:

Builds still keep timing out. Let's bump the limit.

### Other information:

## Testing instructions:

*


samiff (approved):
How did #30824 go as far as helping out build times?

anomiex:
> How did #30824 go as far as helping out build times?

It's hard to tell. The graph does look to be a little lower, but there's so much variation in the times.

PR #30969, opened 2023-05-25
build: Skip composer install in CI build if not needed

anomiex:
## Proposed changes:

I probably should have asked myself this question earlier: do we really need to `composer install` all 100-some projects being built? If they're not plugins and don't have any build script, there's no reason to other than for changelogger which the build command itself runs. Might save us a minute or two.

Probably we could skip even more, JS packages that just call through to `pnpm build` probably don't need it either. Might even help to do `--no-dev` for more things, if it turns out most of the time the dev stuff is for tests rather than building. But I'll leave looking into that for a followup.

### Other information:

## Testing instructions:

* Build shouldn't change. Just it should hopefully be faster.

### DIFF
--- a/tools/cli/commands/build.js
+++ b/tools/cli/commands/build.js
@@ -536,7 +536,32 @@ async function buildProject( t ) {
 		await fs.readFile( `${ t.cwd }/composer.json`, { encoding: 'utf8' } )
 	);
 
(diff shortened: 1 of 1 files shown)


samiff (approved):
Let's give it a go :shipit:

PR #30971, opened 2023-05-26
Fix Docker when data/plugins not present

kraftbj:
When trying to start up Docker today, I started hitting an error that caused the `wordpress` container to fail that we couldn't `ln` `wordpress-develop/tests/phpunit/data/plugins/jetpack` because the directory wasn't available. Looking on my filesystem, there is a `data/blocks`, but no `plugins`. Manually adding it resolved the issue.

My proposed solution is to run `mkdir --parents` to ensure the holding directory is present without worrying about an error if it does.

## Proposed changes:

* `mkdir --parents` before attempting the `ln`.

### Other information:

## Jetpack product discussion

no

no

## Testing instructions:

* `jetpack docker down && jetpack docker clean && jetpack docker up`
*  Does it successfully load?

### DIFF
--- a/tools/docker/bin/run.sh
+++ b/tools/docker/bin/run.sh
@@ -96,6 +96,7 @@ if [ "$COMPOSE_PROJECT_NAME" == "jetpack_dev" ] ; then
 	fi
 
 	# Symlink jetpack into wordpress-develop for WP >= 5.6-beta1
+	mkdir --parents /tmp/wordpress-develop/tests/phpunit/data/plugins/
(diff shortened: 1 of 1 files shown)


anomiex:
I'm wary of this change. That directory should have existed since [10 years ago](https://core.trac.wordpress.org/browser/trunk/tests/phpunit/data/plugins?rev=25165), so there's no reason it should need to be created. Did the `svn co` above somehow break without triggering bash's `-e`? What else is missing in that case?

PR #30974, opened 2023-05-26
CRM: Quote template output - making sure new lines are retained

coder-karen:
## Proposed changes:

* This PR fixes an issue whereby new lines were not being preserved in quote templates: https://github.com/Automattic/zero-bs-crm/issues/3123
* The issue was caused in this PR - https://github.com/Automattic/jetpack/pull/30306 - specifically the removal of a helper function that was taking DB saved HTML and outputting into the WP editor using this: ~`html_entity_decode(nl2br(stripslashes($string)));`~ `wpautop(html_entity_decode(stripslashes($string), ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401));`.
* The current code was just returning the string as HTML is escaped prior to output, but the new line (added by ~`nl2br`~ `wpautop`) was not retained , so this PR adds that back.

### Other information:

## Jetpack product discussion

https://github.com/Automattic/zero-bs-crm/issues/3123

No.

## Testing instructions:

* To replicate the issue, in trunk if testing locally (or just testing with CRM 5.8.0),  open up the quotes template page - `/wp-admin/admin.php?page=manage-quote-templates` - and make sure you have a quote template with regular test added on new lines.
* Add a new quote and add that new quote template to the quote. 
* The new lines you added will be removed - unless you manually added HTML in the editor (eg `<p>Test</p>` - which wouldn't be expected behaviour anyway).
* To test the fix, using the [Jetpack Beta plugin](https://github.com/Automattic/jetpack-beta) on a test site with this branch applied or testing locally with the branch checked out, run through those same steps again.
*  The new lines you'd added in the template itself should be retained.


tbradsha (changes requested):
Thanks for linking to the other PR. I did some comparisons and noticed that the old `zeroBSCRM_io_WPEditor_DBToHTML()` used `wpautop()`, which effectively replaces double newlines with a paragraph tag.

By using `nl2br()` in this PR, there are some scenarios where we end up with a double `<br>` instead of `<p>`. For example, if you send an email, you'll get extra lines, and if you use a quote template to generate a quote you'll get extra lines as well.

Likely the best thing to do is to look at the old PR and use `wpautop()` wherever `zeroBSCRM_io_WPEditor_DBToHTML()` was used previously.

coder-karen:
Thanks @tbradsha , yes that seems to do it -  I seem to have been referencing the older output functions from `zeroBSCRM_io_WPEditor_DBToHTML` before instead. Changed via a new commit now.

tbradsha (approved):
Looks good!

PR #30975, opened 2023-05-26
Docs: Making it clearer when a Crew review is needed in pull-request.md

coder-karen:
## Proposed changes:

* This PR adds a little extra information to clarify when a team would need to add the 'Needs Review' label so that Crew can review a PR.

### Other information:

## Jetpack product discussion

n/a

No.

## Testing instructions:

* Proof-read.


jeherve (commented):
It's true we haven't updated the docs since the requirements changed. Thanks for getting this in!

Since we're looking at this, I think it may be worth updating this too:
https://github.com/Automattic/jetpack/blob/7a053bba620ac150b0fe18a7e89de373c8ce640c/projects/github-actions/repo-gardening/src/tasks/check-description/index.js#L517-L521

However, I don't know if I would phrase it like you did. There are only 2 packages that have strict requirements these days: the connection and the sync packages. Crew reviews aren't *needed* for anything else.

However, we're still around if folks want an extra set of eyes, an opinion, any help really.

With that mind, maybe the default in our docs should be to add the Needs Team Review, get that review and merge. And then, we'd mention that if you need an extra review from someone familiar with the codebase, you can add the "needs review" to get that extra review.

What do you think?

coder-karen:
> With that mind, maybe the default in our docs should be to add the Needs Team Review, get that review and merge. And then, we'd mention that if you need an extra review from someone familiar with the codebase, you can add the "needs review" to get that extra review.

> What do you think?

That makes sense to me, and thanks for pointing out that additional file. I've added a new commit with a change to both.

anomiex:
> There are only 2 packages that have strict requirements these days: the connection and the sync packages. 

Even those don't actually need a crew review, as the check was made non-required and was never re-required after the list was trimmed down. We may as well just disable the CI workflow entirly (p1684435201776199/1684417733.819429-slack-CBG1CP4EN).

jeherve (approved):

PR #30983, opened 2023-05-26
Boost: Update Image Guide tracking to use Pixel instead of ajax requests

dilirity:
Fixes https://github.com/Automattic/boost-cloud/issues/234

## Proposed changes:

* Update tracking UI interactions in the front-end for Image Guide, to happen via Pixel tracking, instead of sending requests to the server.

### Other information:

## Jetpack product discussion

https://github.com/Automattic/boost-cloud/issues/234

no

## Testing instructions:

* Setup this version on Boost on a website (you need a valid Jetpack connection);
* Enable Image Guide;
* While logged-in, view a page with images in the front-end;
* Check the network tab while the page is loading - you should see requests to `http://pixel.wp.com/t.gif`.


thingalon (dismissed):
👏 Nice one! This seems to work, and the code changes look sensible to me. It is a bit messy that we have an `analytics` and a `jetpackBoostAnalytics` object side-by-side in the global namespace, but I'm not sure how avoidable that is.

dilirity:
@thingalon good point.  I tested several ways to do this (one of which involved enqueueing a [separate JS file](https://github.com/Automattic/jetpack/blob/trunk/projects/packages/connection/src/js/tracks-callables.js)) and it seems I forgot to remove the `analytics` from the global scope. It doesn't belong there, as it's imported from a package.

thingalon (approved):
Nice! Thank you :)

Issue #30842, opened 2023-05-22
AI Assistant: Show upgrade notice when user runs out of requests


nunyvega:
Modify the upgrade prompt added in #30736 so it is visible once the user runs out of requests.

~This issue is blocked by #30727~

Tasks list

- [x] Create an endpoint to get AI Assistant feature data [30926](https://github.com/Automattic/jetpack/pull/30926)
- [x] Show upgrade banned based on the endpoint response [30936](https://github.com/Automattic/jetpack/pull/30936)
- [ ] Check the feature availability when requesting JWT [30984](https://github.com/Automattic/jetpack/pull/30984)
- [x] Check the feature availability when streaming [30989](https://github.com/Automattic/jetpack/pull/30989)
- [x] Tweak the UI/UX [31004](https://github.com/Automattic/jetpack/pull/31004)

retrofox:
There is a patch D111358-code and a PR https://github.com/Automattic/jetpack/pull/30813 that handles the issue. Anyway, we are still blocked by https://github.com/Automattic/jetpack/issues/30727

PR #30984, opened 2023-05-26
AI Assistant: check feature availability when requesting jwt

retrofox:
We are checking if the site requires an upgrade when the block renders, but it should be done dynamically, considering the requests count changes while the user interacts with the block.

This PR checks the feature availability when the app requests a `jwt`. Taking this path, we reduce the requests performed by the client, avoiding unnecessarily hitting other endpoints.

Regarding the UI/UX, the block shows an error when the site needs an upgrade. We will continue with follow-up tasks

Fixes #
Part of https://github.com/Automattic/jetpack/issues/30842

## Proposed changes:

* AI Assistant: check feature availability when requesting jwt

### Other information:

#### Endpoint changes

It introduces the `require-upgrade` property into the response. You can check it by hitting the `/wpcom/v2/sites/$site/jetpack-ai/ai-assistant-feature` endpoint.

image

* Go to the block editor
* Get an AI Assistant block instance
* Test with a site without the `AI Assistant` feature
* Use the block. You can check the request number by hitting the endpoint mentioned above
* Confirm the error shows up when the site archives the requests limit, and when it doesn't support the AI Assistant feature.

Screenshot 2023-05-26 at 10 45 33

### DIFF
--- a/projects/plugins/jetpack/_inc/lib/class-jetpack-ai-helper.php
+++ b/projects/plugins/jetpack/_inc/lib/class-jetpack-ai-helper.php
@@ -321,11 +321,20 @@ class Jetpack_AI_Helper {
 			$requests_limit = \OpenAI_Limit_Usage::NUM_FREE_REQUESTS_LIMIT;
 			$requests_count = \OpenAI_Request_Count::get_count( $blog_id );
 
+			/*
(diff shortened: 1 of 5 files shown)


nunyvega:
Is this still necessary?

PR #30987, opened 2023-05-26
Social: Include username in publicize connection test results

spsiddarthan:
## Proposed changes:

* Include username in the connection test results schema

### Other information:

## Jetpack product discussion

p1685108345251429/1685009328.266869-slack-C02JJ910CNL

NO
## Testing instructions:

* You can test this only on the Jetpack plugin and not the Jetpack Social plugin, since the latter calls the endpoint via WPCOM. 
* Without this PR applied, you won't see a 'username' being returned by the `connection-test-results` endpoint.
* With this PR applied, you will see a 'username' being returned by the `connection-test-results` endpoint.


pablinos (approved):
This fixes it!

PR #30990, opened 2023-05-26
Block Editor: add connected user's email address to editor context

jeherve:
## Proposed changes:

This data may prove useful to display the current user's email address in the editor to demo subscription-related features.

### Other information:

## Jetpack product discussion

- D110864-code
- p1685112173899369-slack-CBG1CP4EN

* No

## Testing instructions:

* Go to Posts > Add New
* In your browser console, type in `Jetpack_Editor_Initial_State.tracksUserData.email`
    * You should see the current connected user's email address.

### DIFF
--- a/projects/plugins/jetpack/class.jetpack-gutenberg.php
+++ b/projects/plugins/jetpack/class.jetpack-gutenberg.php
@@ -643,6 +643,7 @@ class Jetpack_Gutenberg {
 		if ( defined( 'IS_WPCOM' ) && IS_WPCOM ) {
 			$user                      = wp_get_current_user();
 			$user_data                 = array(
+				'email'    => $user->user_email,
 				'userid'   => $user->ID,
 				'username' => $user->user_login,
 			);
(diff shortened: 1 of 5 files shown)


lezama:
this is exactly what I need! will test it right away

projects/plugins/jetpack/class.jetpack-gutenberg.php
```diff
@@ -643,6 +643,7 @@ public static function enqueue_block_editor_assets() {
 		if ( defined( 'IS_WPCOM' ) && IS_WPCOM ) {
 			$user                      = wp_get_current_user();
 			$user_data                 = array(
+				'email'    => $user->user_email,
```
simison, line 646: Is it worth limiting to add this only when Email/subscribe features are active?

jeherve: I wouldn't think so, to be honest. Some of the other items we populate in that initial state are only used by a few specific blocks, but we do not necessarily limit them to specific modules. 
And since we request user information anyway, adding one more item to the array doesn't seem too expensive to me.

lezama (approved):
tests great!

PR #30999, opened 2023-05-26
AI Assistant: add change tone menu option when post has content

### LINKED ISSUE #30998: AI Assistant: Show Change tone in the menu whenever the post has content
If there's previous content, we should offer the option to change tone, and we currently don't do so:

Image
(The haiku in the screenshot is a paragraph block)

nunyvega:
Fixes #30998

## Proposed changes:

* Display change tone option when the post already has content
Screenshot 2023-05-26 at 20 38 12

### Other information:

No 
## Testing instructions:

* Go to the editor and verify that change tone is there
* Confirm that the existing content is sent and the tone is changed

### DIFF
--- a/projects/plugins/jetpack/changelog/add-assistant-change-tone-when-content
+++ b/projects/plugins/jetpack/changelog/add-assistant-change-tone-when-content
@@ -1,0 +1,4 @@
+Significance: patch
+Type: other
+
+Changing menu of AI assistant, not yet available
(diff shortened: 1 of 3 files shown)


lhkowalski (approved):
Code looks good.

The option was added to the toolbar:

Screen Shot 2023-05-26 at 16 21 13

When the tone option is clicked, the correct content is sent with the prompt:

Screen Shot 2023-05-26 at 16 22 09

The option works as expected and I can also followup on the generated content changing the tone again, with the correct content being sent with the prompt (in this case, the generated content).

PR #31009, opened 2023-05-26
Fix check for falsy values in `getReachForAccessLevelKey()`

### LINKED ISSUE #30986: Newsletter: "undefined" displayed instead of subscribers count
#### Steps to reproduce

1. Start with a site running Jetpack trunk and connected to WordPress.com.
2. Go to Pages > add New, create a new subscribe page with a subscribe block, and add a payment plan there.
3. Go to Posts > Add New, and create a new post.

You'll see the following (note the "undefined") displayed in the different subscriptions panels:

Screenshot 2023-05-26 at 12 39 52
Screenshot 2023-05-26 at 12 38 46

n3f:
Fixes #30986.

## Proposed changes:

* Small bugfix.

### Other information:

## Testing instructions:

* Create a WoA site (see p9o2xV-1r2-p2)
* Connect stripe & create a plan.
* Create a subscribe page (with the subscribe block) and make sure both it and the site are published.
* Verify that you see the issue when you create a new post and change the access level to 'paid subscribers'.
* Build this PR locally and rsync to your newly created dev blog.
* Verify that the issue is gone.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
@@ -189,7 +189,6 @@ function NewsletterAccessSetupNudge( { stripeConnectUrl, isStripeConnected, hasN
 function NewsletterAccessRadioButtons( {
 	onChange,
 	accessLevel,
-	socialFollowers,
 	emailSubscribers,
 	paidSubscribers,
 	hasNewsletterPlans,
@@ -268,7 +267,6 @@ function NewsletterAccessRadioButtons( {
 export function NewsletterAccessDocumentSettings( {
 	accessLevel,
 	setPostMeta,
-	socialFollowers,
 	emailSubscribers,
 	paidSubscribers,
 	showMisconfigurationWarning,
@@ -342,7 +340,6 @@ export function NewsletterAccessDocumentSettings( {
 											<NewsletterAccessRadioButtons
 												onChange={ setPostMetaAndClose( onClose ) }
 												accessLevel={ _accessLevel }
-												socialFollowers={ socialFollowers }
 												emailSubscribers={ emailSubscribers }
 												paidSubscribers={ paidSubscribers }
 												stripeConnectUrl={ stripeConnectUrl }
@@ -378,7 +375,6 @@ export function NewsletterAccessDocumentSettings( {
 export function NewsletterAccessPrePublishSettings( {
 	accessLevel,
 	setPostMeta,
-	socialFollowers,
 	emailSubscribers,
 	paidSubscribers,
 	showMisconfigurationWarning,
@@ -419,7 +415,6 @@ export function NewsletterAccessPrePublishSettings( {
 									<NewsletterAccessRadioButtons
 										onChange={ setPostMeta }
 										accessLevel={ _accessLevel }
-										socialFollowers={ socialFollowers }
 										emailSubscribers={ emailSubscribers }
 										paidSubscribers={ paidSubscribers }
 										stripeConnectUrl={ stripeConnectUrl }
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
@@ -35,7 +35,10 @@ function getReachForAccessLevelKey(
 	paidSubscribers,
 	socialFollowers
 ) {
-	if ( emailSubscribers === null || paidSubscribers === null || socialFollowers === null ) {
+	const isFalsy = value => {
+		return value === null || value === undefined || value === '';
+	};
+	if ( [ emailSubscribers, paidSubscribers, socialFollowers ].filter( isFalsy ).length > 0 ) {
 		return 0;
 	}
 
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
@@ -29,16 +29,7 @@ function Link( { href, children } ) {
 	);
 }
 
-function getReachForAccessLevelKey(
-	accessLevelKey,
-	emailSubscribers,
-	paidSubscribers,
-	socialFollowers
-) {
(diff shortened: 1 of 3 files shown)


n3f:
@Automattic/gold , does this look right?  I actually think a better fix would be this:

```js
function getReachForAccessLevelKey(
	accessLevelKey,
	emailSubscribers,
	paidSubscribers,
	socialFollowers
) {
	if ( ! socialFollowers ) {
		return 0;
	}

	switch ( accessOptions[ accessLevelKey ].key ) {
		case accessOptions.everybody.key:
			return emailSubscribers || 0;
		case accessOptions.subscribers.key:
			return emailSubscribers || 0;
		case accessOptions.paid_subscribers.key:
			return paidSubscribers || 0;
		default:
			return 0;
	}
}
```

But that's not what the current function is doing and I'm not sure if there's a reason for the current behavior...

projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
```diff
@@ -35,17 +35,17 @@ function getReachForAccessLevelKey(
 	paidSubscribers,
 	socialFollowers
 ) {
-	if ( emailSubscribers === null || paidSubscribers === null || socialFollowers === null ) {
+	if ( ! socialFollowers ) {
```
romarioraffington: Interesting, I suspect we could remove `socialFollowers` since it's not used in the fn?

millerf: I agree.

millerf:
@n3f Reading this, I am wondering if the issue does not lie somewhere else... 

The subscriber's counts are retrieved from [this endpoint](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/jetpack/extensions/blocks/subscriptions/api.js#L16).

Part of the code is here:
https://github.com/Automattic/jetpack/blob/0bd8d071588aa00188272b511a5ee74304362214/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php#L172

And I am guessing the core of the issue is this code:
https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php#L176

Notice the `\wpcom_fetch_subs_counts( false )`, which does not include the paid_subscribers count...

Counting the number of paid subscribers is somewhat resource-intensive, so we gated, as NL were not enabled on Jetpack. 

My idea to move forward, and happy to be corrected:
- Keep your changes which are a good gating in case of issues
- Add @romarioraffington 's feedback
- Change this "false" to "true"
- Add a default for paid_subscribers [here](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php#L192) - TBH, I am not sure if that code is ever run (CC @jeherve)


We need to properly test this and check the paid-subscriber count goes up on self-hosted (because I am pretty sure it stick to 0 at the moment)...

We can also add a clean up task to remove that wpcom_fetch_subs_counts's parameter (code 
 fbhepr%2Skers%2Sjcpbz%2Sjc%2Qpbagrag%2Szh%2Qcyhtvaf%2Sfhofpevcgvbaf.cuc%3Se%3Q639q5o09%231428-og ) log, as it is not needed anymore.

projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
n3f: Taking it out affects the behavior.  Not sure why.

n3f:
@millerf

> Change this "false" to "true

That is wrapped in a `is_wpcom()` condition.  It shouldn't be involved here at all.

**TL;DR** I tried all these suggested changes in isolation and they don't work.  I can add them to this PR, but I'm not sure they are needed or effective at this point. See #31094.

projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
n3f: Hmm, working today... Updated PR.

millerf:
@n3f 

> That is wrapped in a is_wpcom() condition. It shouldn't be involved here at all.

It is. Remember the "Jetpack way" is to reuse callbacks.
Front-end calls the (self-hosted) server that in turns calls WPCOM for data.
The "is_wpcom()" part is the part that actually retrieves the data from the DB, the part on the self-hosted servers mainly caches it.
 

> TL;DR I tried all these suggested changes in isolation and they don't work. 

That is because the self-hosted server calls to WPCOM are not sandboxed... Making it work is very complex (remember we started working like that in October last year).
 

> I can add them to this PR, but I'm not sure they are needed or effective at this point. See https://github.com/Automattic/jetpack/pull/31094.
 
An easy way to test it is to subscribe as a paid subscriber to a self-hosted version with your fix. I am pretty sure the count will stay as `undefined/0`...

n3f:
> Making it work is very complex

Well I spent a couple hours trying to test it and I couldn't figure it out.  Should I just merge #31094 into this one and call it good?

As it currently stands, it works and fixes the issue...

millerf:
> > Making it work is very complex
> 
> Well I spent a couple hours trying to test it and I couldn't figure it out. Should I just merge #31094 into this one and call it good?

Did you point the site to your sandbox? I think that was the step missing yesterday: pbOQVh-2zk-p2
But yes, you can merge it...

> 
> As it currently stands, it works and fixes the issue...

I am curious: did you try to subscribe to it as a paid-subscriber?

n3f:
> I am curious: did you try to subscribe to it as a paid-subscriber?

Yes, but something else seems to be borked as well.

| Test | Paid Subscribers |
|--------|--------|
| trunk: No paid subscribers | undefined |
| this pr: No paid subscribers | 0 |
| #31094: no paid subscribers | undefined | 
| trunk: One paid subscriber | undefined |
| this pr: One paid subscriber | 0 |
| #31094: One paid subscriber | undefined |

millerf:
@n3f Follow-up here: https://github.com/Automattic/jetpack/pull/31141
Please close this one if you want, as the other has the proper testing instructions

PR #31016, opened 2023-05-28
Jetpack Forms: fix a typo in the template filter

donnchawp:
Note: this is a duplicate of #30944.
A filter on the template used for form responses has a small typo in it.
The filter is "jetpack_forms_respone_email_template" but it should be "jetpack_forms_response_email_template"

This PR fixes that, while maintaining the old filter, just in case someone has started using it already.

### Other information:

## Jetpack product discussion
pcsBup-yo-p2#comment-452

No

## Testing instructions:
Apply the changes in this PR.
Embed a contact form in a page and then fill it out.
There shouldn't be any visible changes or PHP errors if the patch works.

You could also use the filter to create a simple template for the response email.
Create a file called wp-content/jp-forms-template.php with the following content:
```
<?php

$template = '
<!doctype html>
<html xmlns="http://www.w3.org/1999/xhtml">
<body>
<!-- title -->
<h1>%1$s</h1>

<!-- response -->
<p>%2$s</p>

<!-- link to responses page -->
<p><a href="%3$s">View Responses</a></p>

<!-- link to edit form -->
<p><a href="%4$s">Edit</a></p>

<!-- footer -->
<p>%5$s</p>
</body>
</html>';
```
Then create an mu-plugin with the following code:
```
<?php

function jp_forms_template( $filename ) {
        return ABSPATH . '/wp-content/jp-forms-template.php';
}

add_filter( 'jetpack_forms_response_email_template', 'jp_forms_template' );
```

Now fill out the contact form and wait for the response email. It should show a basic email with only the response and some metadata.

### DIFF
--- a/projects/packages/forms/changelog/fix-template_filter_name
+++ b/projects/packages/forms/changelog/fix-template_filter_name
@@ -1,0 +1,5 @@
+Significance: patch
+Type: fixed
+Comment: A typo in an unreleased filter was fixed.
+
+
--- a/projects/packages/forms/package.json
+++ b/projects/packages/forms/package.json
@@ -1,7 +1,7 @@
 {
 	"private": true,
 	"name": "@automattic/jetpack-forms",
-	"version": "0.19.0",
+	"version": "0.19.1-alpha",
 	"description": "Jetpack Forms",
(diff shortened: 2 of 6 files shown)


haqadn (approved):
👍

PR #31017, opened 2023-05-28
WP Super Cache: preload notification shows last 5 preloaded files now

donnchawp:
When the plugin is preloading pages it displays the last preloaded URL. This patch changes that so it show the last 5 preloaded URLs, with the newest URL coming up at the bottom, and the oldest being removed off the top.

### Other information:

No

## Testing instructions:
Apply patch
Run a preload and  ten seconds or so after it starts,  you will see the last 5 URLs that have been preloaded appear in the admin notice area of the preload page.

BEFORE
Before

AFTER
After


thingalon:
Testing this out, I had a bit of an odd experience:
https://github.com/Automattic/jetpack/assets/1369626/2ad2ac22-3c40-4678-93eb-e0ef78af1fd1

I clicked preload now, and waited for more than 10 seconds. Nothing appeared to happen. So I refreshed the page, and nothing was happening. Tried again, saw the preload in progress, where it reported the same page twice. Nothing further seemed to happen, so I refreshed again to what appeared to be a completed process.

Was I clicking around too impatiently?

donnchawp:
There used to be a notification that preload was going to run in X seconds when you reloaded the page but for some reason it's not working now. So, what you saw is expected, apart from the URL showing up twice, but that might be because you have only one or not many posts in your test site? The JS notification script loads the post URL every 3 seconds. It is supposed to check for repeats though, so it shouldn't have shown the URL twice.
The notification system doesn't actually show a "Preload finished" message. For most sites, with hundreds or thousands of posts, the site owner isn't going to see it anyway.

donnchawp:
@thingalon - I fixed the notification. If the "preload interval" was set to 0 it wouldn't show at all, so I just removed that check.
It used to show a notification on the "next preload" which was a 30 second pause, but as it's only 3 seconds now, I removed it. It only shows a notification on the "Full preload" at the start of a preload.

thingalon:
@donnchawp I've tinkered with making the preload notification more informative using a bit more JavaScript, and storing more state in the preload info txt file. It also uses wp-ajax to fetch the preload status, ensuring that the preload will tick forward even if you are the only visitor on your site.

Please test it and review it - ed9458015f I am worried about breaking something with the change :)

projects/plugins/super-cache/inc/preload-notification.js
```diff
@@ -1,18 +1,92 @@
...
+			},
+			( json ) => {
+				if ( ! json || ! json.success ) {
+					return;
+				}
+
+				update_preload_status( json.data );
+			}
+		);
+	}, CHECK_INTERVAL );
```
dilirity: I'm not a huge fan of relying on the user's connection being fast enough to finish the request in the time specified in `CHECK_INTERVAL`. I know it's an edge case, but it makes my senses tingle :D

dilirity: The request continues even after the preload has finished.

projects/plugins/super-cache/inc/preload-notification.js
```diff
@@ -1,18 +1,92 @@
...
+					)
+				);
+
+				panel.append( p );
+			}
+
+			if ( data.previous ) {
+				const p = jQuery( '<p>' );
+				p.append( jQuery( '<b>' ).text( __( 'Last preload finished:', 'wp-super-cache' ) ) );
+				p.append( jQuery( '<span>' ).text( new Date( data.previous * 1000 ).toLocaleString() ) );
```
dilirity: Missing space after the colon :)
CleanShot 2023-06-05 at 18 45 43@2x

projects/plugins/super-cache/inc/preload-notification.js
```diff
@@ -1,18 +1,92 @@
...
+		if ( ! data || ( ! data.running && ! data.next && ! data.previous ) ) {
+			return;
+		}
+
+		preloadInfoPanel.empty();
+
+		if ( data.running ) {
+			const panel = jQuery( '<div class="notice notice-warning">' );
+
+			panel.append( jQuery( '<b>' ).text( __( 'Preloading', 'wp-super-cache' ) ) );
```
dilirity: Missing paragraph around the `b` tag, causes it's spacing to look off
image

Maybe change it to an `H4`?

projects/plugins/super-cache/inc/preload-notification.js
```diff
@@ -1,18 +1,92 @@
...
+
+			preloadInfoPanel.append( panel );
+		} else if ( data.next || data.previous ) {
+			const panel = jQuery( '<div class="notice notice-info">' );
+
+			if ( data.next ) {
+				const diff = data.next - Math.floor( Date.now() / 1000 );
+				const seconds = diff % 60;
+				const minutes = Math.floor( diff / 60 ) % 60;
+				const hours = Math.floor( diff / 3600 ) % 24;
```
dilirity: I got a weird result while waiting for the preload to start
image

projects/plugins/super-cache/inc/preload-notification.js
dilirity, line 1: Can the JS file be moved to the root `js` directory? Not a big thing, just wanted to avoid mixing file types.

I just noticed that there's another JS file in there.

dilirity (changes requested):
Good ol' jQuery. I'm not sure who to send the feedback to :D

cc @thingalon just in case.

I really like the new UI! No more manually refreshing the page - love it.

GitHub Actions was kind enough to remind us that we can remove one of the files from the linter exclude lists:
image

I've also left some comments.

projects/plugins/super-cache/inc/preload-notification.js
donnchawp: It's 3 seconds, and it used to be 1 second. I don't think it's much to worry about though as it's only for the notification. The AJAX request is fairly light weight too.

donnchawp: I had a fix for that but it was clobbered. The fix was to check for duplicates, but the code @thingalon added has a "running" status, which I could use to stop it.

projects/plugins/super-cache/inc/preload-notification.js
donnchawp: Yes, good idea! The inc directory predates the js directory.

donnchawp:
The "Cancel Preload" button reverts back to "Preload Cache Now". I'm not sure yet what is causing that to happen, but I want to merge #31022 first which will cause quite a few conflicts with this patch.

projects/plugins/super-cache/inc/preload-notification.js
donnchawp: Unfortunately, "running" is set to false when the preload is scheduled to start, so I think we'll have to go back to checking for duplicate URLs again.
Once #31022 is merged I'll use wpsc_is_preload_active() to stop the JS files being enqueued, or do we mind it doing an AJAX request every 3 seconds on the preload settings page?

projects/plugins/super-cache/inc/preload-notification.js
thingalon: Done in 4447abfc82.

projects/plugins/super-cache/inc/preload-notification.js
thingalon: 🤔 Defensive code is a good idea. In 4d41ccdc4f I made it use `setTimeout` which is called after each request completes so each will be 3 seconds after the last is completed.

It will also stop if one of the update requests fails, so that we don't keep polling a broken server. If users need to resume after an error they can refresh the page.

thingalon: In 406ea75214 , I updated the preload code to vary the refresh speed based on whether preload is running (or is scheduled to run soon). If it's running, it refreshes every 3 seconds. If it's idle or the next preload is more than a minute away, it updates every 30 seconds.

projects/plugins/super-cache/inc/preload-notification.js
thingalon: Fixed in d58e6b8d1b. Weirdly, h4 was too far to the left too - so I just kept it simple and added the missing p

projects/plugins/super-cache/inc/preload-notification.js
thingalon: I can't replicate this, but I guess it might happen if the next run time is in the past.

So I've added a guard against that in d2dfe131bb, which guarantees the diff between now and the next-run time never dips below 0.

projects/plugins/super-cache/inc/preload-notification.js
thingalon: Fixed in 232a35aff5

thingalon:
@dilirity thank you for the thorough review. I believe I've addressed all of your points. Please check it out :-)

thingalon:
> The "Cancel Preload" button reverts back to "Preload Cache Now". I'm not sure yet what is causing that to happen, but I want to merge #31022 first which will cause quite a few conflicts with this patch.

It seems to be working to me. How can I reproduce the issue?

donnchawp:
The issue was, $currently_preloaded wasn't set to true because [this line](https://github.com/Automattic/jetpack/pull/31017/files#diff-49e24876fd767aa7d12d7018a180f19e42dc62b47484a137fe4d42fdb31d10f6L1018) was removed. I just added that check in partials/preload.php and it appears to be working again,
However, the "Preloading is currently running" message is still showing, after I cancelled the preload so I need to clean that up.

donnchawp:
The issue there was the idle status was't being set if preload was cancelled during taxonomy preload. Fixed that in https://github.com/Automattic/jetpack/pull/31017/commits/fd89b9c119ccaf89422137a395d90c3bd04daf20 :)

dilirity (approved):
Thanks guys! This is looking great.

:shipit:

PR #31021, opened 2023-05-28
Add Send Email Preview feature

lezama:
Part of https://github.com/Automattic/wp-calypso/issues/76134

## Proposed changes:

* Adds a "Preview" button to the Pre Publish Newsletter Panel.
* The "Preview" button opens a "Send a test email" modal showing the current user email and a "Send" button.

Screenshot 2023-06-20 at 15 46 49

Future iteration will use https://github.com/WordPress/gutenberg/pull/50605

### Other information:

## Jetpack product discussion
peKye1-6z-p2

No

## Testing instructions:

- Build/Run this branch on your test site.
- Create a new post on your self-hosted site.
- Click Publish, in Pre Publish, click "Preview" at the bottom of the Newsletter panel.
- A "Send a test email" modal should open.
- Click on the "Send" button.
- You should receive a test email!


projects/plugins/jetpack/extensions/blocks/subscriptions/preview-email.js
```diff
@@ -0,0 +1,66 @@
...
+							</Button>
+						</>
+					) }
+				</Modal>
+			) }
+		</>
+	);
+};
+
+export default compose( [
```
lezama: no need for this, will move it up

projects/plugins/jetpack/extensions/blocks/subscriptions/email-preview.js
```diff
@@ -0,0 +1,85 @@
+import apiFetch from '@wordpress/api-fetch';
+import { Button, Flex, FlexItem, Modal, TextControl } from '@wordpress/components';
+import { useSelect } from '@wordpress/data';
+import { useState } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+import './email-preview.scss';
+import check from './check.svg';
+import illoShare from './illo-share.svg';
```
simison: Let's rename this, it came from share modal and thus was called "illustration share".

projects/plugins/jetpack/extensions/blocks/subscriptions/email-preview.js
```diff
@@ -0,0 +1,85 @@
...
+		} )
+			.then( () => {
+				// Handle response here
+				setEmailSending( false );
+				setEmailSent( true );
+			} )
+			.catch( () => {
+				setEmailSending( false );
+
+				// Handle error here
```
simison: Snackbar notice would be quick to use here:

https://wordpress.github.io/gutenberg/?path=/story/components-snackbar--default

...but I think a state flag for "error" would be good, and we can toggle some message in place of success message that now appears with `emailSent`.

projects/plugins/jetpack/extensions/blocks/subscriptions/email-preview.js
```diff
@@ -0,0 +1,85 @@
...
+			{ isModalOpen && (
+				<Modal
+					className="jetpack-email-preview"
+					title={ __( 'Send a test email', 'jetpack' ) }
+					onRequestClose={ () => setIsModalOpen( false ) }
+				>
+					{ emailSent ? (
+						<Flex>
+							<FlexItem className="jetpack-email-preview__email-sent">
+								<img className="jetpack-email-preview__check" src={ check } alt="" />
```
simison: Let's use [`@wordpress/icon`](https://wordpress.github.io/gutenberg/?path=/story/components-icon--default) and existing [check icon](https://wordpress.github.io/gutenberg/?path=/story/icons-icon--library).

projects/plugins/jetpack/extensions/blocks/subscriptions/index.js
```diff
@@ -95,5 +95,9 @@ export const settings = {
 };
 
 export const pluginSettings = {
-	render: SubscribePanels,
+	render: () => (
+		<Fragment>
```
simison: You can nowadays replace `<Fragment></Fragment>` with `<></>`

projects/plugins/jetpack/extensions/blocks/subscriptions/email-preview.scss
```diff
@@ -0,0 +1,44 @@
...
+  width: 80px;
+  height: 44px;
+}
+
+.jetpack-email-preview__img {
+  height: 110px;
+}
+
+.jetpack-email-preview__email-sent {
+  color: green;
```
simison: Let's use colour variable.

Would one from core's `@wordpress/base-styles` be close to designed? https://github.com/WordPress/gutenberg/blob/053413e9e7e1053037432a57931e53d7d03c7060/packages/base-styles/_colors.scss#L28

I think there was Jetpack base-styles package as well to pick some colours from.

projects/plugins/jetpack/extensions/blocks/subscriptions/email-preview.scss
```diff
@@ -0,0 +1,44 @@
+.jetpack-email-preview {
+  height: 200px;
+  width: 645px;
+  left: 391px;
+  top: 392px;
```
simison: How does the modal behave without defining these sizes/positions?

If possible without any overrides is best.

We could do `max-width` and `max-height` if we must.

projects/plugins/jetpack/extensions/blocks/subscriptions/email-preview.scss
```diff
@@ -0,0 +1,44 @@
...
+  border-radius: 0;
```
simison: Let's stick to core's version, and share screenshot with @crisbusquets to see if the core version differs from design. Just to avoid component overrides.

The same for all `components-modal__` bits below.

projects/plugins/jetpack/extensions/blocks/subscriptions/email-preview.js
```diff
@@ -0,0 +1,85 @@
+import apiFetch from '@wordpress/api-fetch';
+import { Button, Flex, FlexItem, Modal, TextControl } from '@wordpress/components';
+import { useSelect } from '@wordpress/data';
+import { useState } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+import './email-preview.scss';
+import check from './check.svg';
```
jeherve: Could we use `check` from `@wordpress/icons` here?

projects/plugins/jetpack/extensions/blocks/subscriptions/email-preview.scss
jeherve: > I think there was Jetpack base-styles package as well to pick some colours from.

Yes, that's what I would recommend as well. The `$studio-jetpack-green-` colors could be useful here. Here is an example of how you can use it:
https://github.com/Automattic/jetpack/blob/3a3c4adc1c0913bac224ddb832713bd6101fe1ab/projects/plugins/jetpack/extensions/plugins/publicize/editor.scss#L2-L15

You can read more about it here:
https://github.com/Automattic/jetpack/blob/ade2b4cad370b6feb6153f90f7f1cf8e9dfc9d77/projects/plugins/jetpack/extensions/README.md#colors

projects/plugins/jetpack/extensions/blocks/subscriptions/email-preview.scss
```diff
@@ -0,0 +1,44 @@
...
+}
+
+.jetpack-email-preview__email {
+  .components-text-control__input {
+    height: 44px;
+    width: 266px;
+    border-radius: 3px;
+    border-color: gray;
+    padding: 12px 16px 12px 16px;
+    background-color: lightgray;
```
jeherve: Just like above, you should be able to pull the colors dynamically.

projects/plugins/jetpack/extensions/blocks/subscriptions/email-preview.js
```diff
@@ -0,0 +1,85 @@
...
+				setEmailSending( false );
+
+				// Handle error here
+			} );
+	};
+
+	return (
+		<>
+			<FlexItem>
+				<Button isPrimary onClick={ () => setIsModalOpen( true ) }>
```
jeherve: `isPrimary` is deprecated. You can use `variant="primary"` now.

```suggestion
				<Button variant="primary" onClick={ () => setIsModalOpen( true ) }>
```

projects/plugins/jetpack/extensions/blocks/subscriptions/email-preview.js
```diff
@@ -0,0 +1,85 @@
...
+								<TextControl
+									className="jetpack-email-preview__email"
+									value={ window?.Jetpack_Editor_Initial_State?.tracksUserData?.email }
+									disabled
+								/>
+							</FlexItem>
+							<FlexItem>
+								<Button
+									className="jetpack-email-preview__button"
+									isPrimary
```
jeherve: Same as above.

```suggestion
									variant="primary"
```

projects/plugins/jetpack/extensions/blocks/subscriptions/email-preview.scss
lezama: Thanks!

projects/plugins/jetpack/extensions/blocks/subscriptions/email-preview.scss
simison: Update: we stuck to core version and in [core we're changing to 4px rounding](https://github.com/WordPress/gutenberg/pull/51254).

lezama:
Moved the button to the Newsletter PrePublish panel.

Screenshot 2023-06-20 at 15 38 16

projects/plugins/jetpack/extensions/blocks/subscriptions/email-preview.js
```diff
@@ -0,0 +1,103 @@
...
+									<div className="jetpack-email-preview__sent_text">
+										{ __( 'Email sent successfully', 'jetpack' ) }
+									</div>
+								</HStack>
+							) : (
+								<HStack>
+									<TextControl
+										className="jetpack-email-preview__email"
+										value={ window?.Jetpack_Editor_Initial_State?.tracksUserData?.email }
+										disabled
```
jeherve, line 82: I assume you'll make this customizable in a future PR?

jeherve (approved):
This tests well for me. TIL about `__experimentalHStack` and `__experimentalVStack`!

🚢

projects/plugins/jetpack/extensions/blocks/subscriptions/email-preview.js
lezama: yes, that's the plan but needs further thinking to prevent spam

edanzer (approved):
This is testing as expected, and I'm signing off on that basis. Code changes look right too, but I can see that Jetpack crew already signed off on the code, and they know Jetpack much better than me. 

Note: I did have an issue where /send-email-preview/ endpoint did not work initially. I had to reset permalinks to get it working. But I don't that issue is specific to this PR.

crisbusquets:
> Moved the button to the Newsletter PrePublish panel.
> 
> Screenshot 2023-06-20 at 15 38 16


Hi @lezama !
Can we change it to a link that provides more context? For example:

Screenshot 2023-06-30 at 18 15 06

I wouldn't use a button, the information architecture feels weird —the button component in the pre-publish panel seems to only be used for the two main actions (`Publish` or `Cancel`).

What do you think?

lezama:
I copied the button from the Social Preview panel above so they had similar patterns.

Screenshot 2023-07-03 at 11 38 20

Happy to do whatever, maybe change both 😆

crisbusquets:
> I copied the button from the Social Preview panel above so they had similar patterns.
> 
> Screenshot 2023-07-03 at 11 38 20
> Happy to do whatever, maybe change both 😆

Oh! I wasn't aware this one existed 🥲

Do you know who worked on that? I prefer to use links instead of buttons, but I'd like to know the reason behind the `Social Previews` design. 

Also, are we able to change or add additional text to the Newsletter accordion?

lezama:
> Do you know who worked on that? I prefer to use links instead of buttons, but I'd like to know the reason behind the Social Previews design.

Looks like @manzoorwanijk did the coding. 

> Also, are we able to change or add additional text to the Newsletter accordion?

yes, we can easily add something like the Social one has.

crisbusquets:
> > Do you know who worked on that? I prefer to use links instead of buttons, but I'd like to know the reason behind the Social Previews design.
> 
> Looks like @manzoorwanijk did the coding.
> 
> > Also, are we able to change or add additional text to the Newsletter accordion?
> 
> yes, we can easily add something like the Social one has.

Cool, thanks!

I've reviewed it again, and I'm ok with using the button. A link would be less prominent, and I guess we want people to be aware they can preview the email they're going to send.

Can we add `Preview what this newsletter will look like before sending it to your audience.` on the top of the first radio button?

Screenshot 2023-07-04 at 09 24 22

[Figma file](h8tMpJlCqNFemEerU9owHI-fi-1319_37266) (last frame on the right)

lezama:
the buttons would change who receives/access the newsletter, maybe we should add two texts?

crisbusquets:
> the buttons would change who receives/access the newsletter, maybe we should add two texts?

I don't understand this 😅 Do you mean the radio buttons?

lezama:
it confuses me that the text is not next to the button, it makes me think that the options there are just for the Preview 🤷

crisbusquets:
> it confuses me that the text is not next to the button, it makes me think that the options there are just for the Preview 🤷

Oh, ok. Now I get it 😬

Maybe the other way around is better?

Screenshot 2023-07-04 at 16 36 54

lezama:
I like it better! 🤝

lezama:
I guess we should avoid "your" here too 😁

PR #31022, opened 2023-05-28
WP Super Cache: Simplify the preload loop and improve stopping procedures

donnchawp:
This PR simplifies the preload loop by removing the "taxonomy_preload" transient option recently added.

To do this it has to do more checks on the files generated by the preload which was done in the new wpsc_is_preload_active() function.

The taxonomy preload loop doesn't depend on the site owner preloading "all" posts now either. 

The "start email" had to be rejigged because of the missing transient option.

The cancellation code will work better now and should stop the process more reliably. Sometimes it doesn't stop properly if the job is unscheduled and files not removed.

The preload notification doesn't display the current URL by default now, it lets the JS script do that.

### Other information:

no

## Testing instructions:
Apply the patch. Test the preload by running it, with or without taxonomies. Cancel it and make sure it cancels properly.


projects/plugins/super-cache/wp-cache.php
```diff
@@ -1011,13 +1011,12 @@ function toggleLayer( whichLayer ) {
...
-			$preload_counter = get_option( 'preload_cache_counter' );
-
-			if ( ( is_array( $preload_counter ) && $preload_counter['c'] > 0 ) || get_transient( 'taxonomy_preload' ) ) {
+			if ( wpsc_is_preload_active() ) {
 				$currently_preloading = true;
 				if ( @file_exists( $cache_path . 'preload_permalink.txt' ) ) {
 					$url  = file_get_contents( $cache_path . 'preload_permalink.txt' );
-					$msg .= '<p>' . sprintf( __( '<strong>Page last cached:</strong> %s', 'wp-super-cache' ), '<span id="preload_status">' . $url . '</span>' ) . '</p>';
+					// translators: %s is a URL.
+					$msg .= '<p>' . sprintf( __( '<strong>Page last cached:</strong> %s', 'wp-super-cache' ), '<div><ul id="preload_status"></ul></div>' ) . '</p>';
```
dilirity: The extra div breaks the paragraph. ul is a block item so it's unnecessary.

projects/plugins/super-cache/wp-cache.php
```diff
@@ -3638,32 +3684,38 @@ function wpsc_reset_preload_settings() {
...
-	@fclose( $fp );
+	$stop_preload_flag = $cache_path . 'stop_preload.txt';
+	if ( file_exists( $stop_preload_flag ) && time() - filemtime( $stop_preload_flag ) > 2 ) {
+		wp_cache_debug( 'wpsc_cancel_preload: force cancelled preload' );
+		wpsc_reset_preload_settings(); // reset the preload if we already tried more than 2 seconds ago.
+	} else {
+		// phpcs:ignore -- WordPress.WP.AlternativeFunctions.file_system_read_fopen WordPress.PHP.NoSilencedErrors.Discouraged
+		$fp = @fopen( $stop_preload_flag, 'w' );
+		// phpcs:ignore -- WordPress.WP.AlternativeFunctions.file_system_operations_fclose WordPress.PHP.NoSilencedErrors.Discouraged
+		@fclose( $fp );
```
dilirity: Does this work? I'm not sure why it's necessary to open a file and close it :D

On second look - it looks like this creates the file if it doesn't exist.

projects/plugins/super-cache/wp-cache.php
```diff
@@ -3638,32 +3684,38 @@ function wpsc_reset_preload_settings() {
...
 		wp_cache_debug( 'wpsc_cancel_preload: unscheduling wp_cache_full_preload_hook' );
 		wp_unschedule_event( $next_preload, 'wp_cache_full_preload_hook' );
 	}
 	wp_cache_debug( 'wpsc_cancel_preload: creating stop_preload.txt' );
-	$fp = @fopen( $cache_path . "stop_preload.txt", 'w' );
-	@fclose( $fp );
+	$stop_preload_flag = $cache_path . 'stop_preload.txt';
+	if ( file_exists( $stop_preload_flag ) && time() - filemtime( $stop_preload_flag ) > 2 ) {
+		wp_cache_debug( 'wpsc_cancel_preload: force cancelled preload' );
+		wpsc_reset_preload_settings(); // reset the preload if we already tried more than 2 seconds ago.
```
dilirity: I feel like this comment should be right above the `if` statement. Took me a while to figure out what's going on. Also - why 2 seconds?

projects/plugins/super-cache/wp-cache.php
```diff
@@ -3207,113 +3207,131 @@ function wp_cron_preload_cache() {
...
-					if ( $fp ) {
-						fwrite( $fp, $out );
-						fclose( $fp );
-					}
-					$details = explode( "\n", $out );
-				} else {
-					$details = explode( "\n", file_get_contents( $taxonomy_filename ) );
+				if ( ! $just_started_preloading && $wp_cache_preload_email_me ) {
+					// translators: 1: site url, 2: blank space
+					wp_mail( get_option( 'admin_email' ), sprintf( __( '[%1$s] Cache Preload Started', 'wp-super-cache' ), home_url(), '' ), ' ' );
```
dilirity, line 3325: Why is there a blank space needed here?

projects/plugins/super-cache/wp-cache.php
```diff
@@ -3207,113 +3207,131 @@ function wp_cron_preload_cache() {
...
 			}
 		}
+
+		if ( $preload_more_taxonomies === true ) {
+			wpsc_schedule_next_preload();
+			return true;
+		}
+	} elseif ( $c === 0 && $wp_cache_preload_email_me ) {
+		// translators: Home URL of website, blank space
+		wp_mail( get_option( 'admin_email' ), sprintf( __( '[%1$s] Cache Preload Started', 'wp-super-cache' ), home_url(), '' ), ' ' );
```
dilirity, line 3416: I noticed the email notification code has been added twice. Wouldn't it make more sense to send the notification when the mutex file is created? As I understand, that file is what prevents multiple preloads from happening at the same time.

dilirity (changes requested):
I tried testing the preload on trunk and this branch, but I wasn't able to see much difference compared to how it works/stops. Sorry if I'm missing something.

Because almost the whole `wp_cron_preload_cache` function was modified (judging by GIT), reviewing it was kind of hard. I think we need to do something about separating "make the monorepo happy" changes from the actual update.

I left some comments on what I thought could be improved. Nothing major, just a few small things.

projects/plugins/super-cache/wp-cache.php
donnchawp: Thanks, removed them!

projects/plugins/super-cache/wp-cache.php
donnchawp: It's a flag that stops the preload. It just creates an empty file. :)

projects/plugins/super-cache/wp-cache.php
donnchawp: The 2 second wait is in case someone double tapped the "cancel" button. I'd rather the preload process stop the preload loop than the process serving the admin page.
But if the preload has got stuck or the user really wants to cancel immediately, then cancel it. That might have unforeseen consequences though - if the preload loop has started, then it will cache ten pages, and then schedule another loop.

I think it would be useful to make the preload loop more resilient. One way might be to check for the mutex file before scheduling the next loop. If it's not there it means the site owner cancelled the preload.

donnchawp: That 2 second wait is now removed. When the user cancels the preload, all the settings are reset, and the "stop preload" file are created to make sure that the preload stops. I also added a check for that file to `wpsc_is_preload_active` and use that where I check for that "stop" file.

projects/plugins/super-cache/wp-cache.php
donnchawp: I found that wp_mail() refused to send a blank email in the past. I'm not sure if that has changed since, however.

projects/plugins/super-cache/wp-cache.php
donnchawp: The mutex file is deleted at the end of a loop, and then recreated on the next scheduled loop, so it's impossible to know at the top of the function if this is the first or 10th loop, especially as I got rid of the taxonomy_preload transient option.

projects/plugins/super-cache/wp-cache.php
dilirity: Today I learned - thanks! I didn't know you can't use `wp_mail` with an empty message. I checked the code and it's intended. Sending an empty string results in a `WP_Error`:
```
Message body empty
```

However, I was referring to the blank space in the translators line [here](https://github.com/Automattic/jetpack/pull/31022/files/bfd32c460a106b90d7ea24e1e9af364cd7d729af#diff-49e24876fd767aa7d12d7018a180f19e42dc62b47484a137fe4d42fdb31d10f6R3237). Only the message is translated, the blank space isn't.

projects/plugins/super-cache/wp-cache.php
dilirity: Thanks for elaborating. On a side note, the blank space isn't part of the translation string.

projects/plugins/super-cache/wp-cache.php
```diff
@@ -3624,7 +3632,15 @@ function wpsc_is_preload_active() {
 	$mutex = $cache_path . 'preload_mutex.tmp';
 	if ( file_exists( $mutex ) ) {
 		return true;
+	} else {
+		return false;
 	}
```
dilirity: This prevents code after the if statement to get executed.

projects/plugins/super-cache/wp-cache.php
```diff
@@ -3605,6 +3632,42 @@ function supercache_admin_bar_render() {
...
+ */
+function wpsc_is_preload_active() {
+	global $cache_path;
+
+	$mutex = $cache_path . 'preload_mutex.tmp';
+	if ( file_exists( $mutex ) ) {
+		return true;
+	} else {
+		return false;
+	}
```
dilirity: This prevents any code after the if statement from getting executed.

dilirity (changes requested):
I love how simpler the preload is becoming! Just found one small bug - below.

projects/plugins/super-cache/wp-cache.php
donnchawp: Doh! I'll sort that out.

projects/plugins/super-cache/wp-cache.php
donnchawp: oh, thanks! Will fix that.

projects/plugins/super-cache/wp-cache.php
donnchawp: Thanks, fixed that!

projects/plugins/super-cache/wp-cache.php
donnchawp: Fixed by moving the "stop preload" file check above it.

dilirity (dismissed):
I think we're good to go :shipit:

dilirity (changes requested):
Looks good. However, cancelling a preload that hasn't already started still updates the `last preload finished` value.

EDIT:
Let's get this merged - the `last preload finished` can be fixed in another PR.

dilirity (dismissed):

dilirity (approved):

PR #31029, opened 2023-05-29
CRM: Resolves 3121 - allow unsubscribe flag removal

tbradsha:
Resolves Automattic/zero-bs-crm#3121 - allow unsubscribe flag removal

## Proposed changes:

When using Mail Campaigns, a contact can unsubscribe, but there's no user-facing way to remove that flag. This is a quick patch to allow an admin to remove an unsubscribe flag from a contact.

Note that while the unsubscribe feature is only implemented in Mail Campaigns, the underlying architecture is in core, which is why this is implemented in core. It is a very frequent request (see comments in Automattic/zero-bs-crm-mailcampaigns#60).

I also split out some of the `AJAX.php` content into separate `admin/*/*.ajax.php` files while figuring out load orders.

### Other information:

## Testing instructions:

With the `fix/crm/3121-allow_unsubscribe_flag_removal` branch:

1. Add a `do-not-email` flag to a CRM contact, e.g.:
```
INSERT INTO `wp_zbs_meta` (`zbs_site`, `zbs_team`, `zbs_owner`, `zbsm_objtype`, `zbsm_objid`, `zbsm_key`, `zbsm_val`, `zbsm_created`, `zbsm_lastupdated`) VALUES (1, 1, -1, 1, 2513, 'do-not-email', '1', 1681414091, 1681414091);
```
2. Go to the contact profile.
3. Click on the "Do not email" flag.
4. Confirm the popup. The flag will be removed.
5. Refresh the page to confirm.


gogdzl (approved):
Nice cleanup!
I changed the cursor of the button to be pointer, but other than that great work!

PR #31032, opened 2023-05-29
CRM: Resolves 3129 - catch PHP errors

tbradsha:
Resolves Automattic/zero-bs-crm#3129 - catch PHP errors

## Proposed changes:

This catches two minor PHP notices:

* If loading the listview but there were no tags or quickfilters (e.g. Mail Campaigns listviews) it would try to access a non-existent index.
* If offline, the extensions page tries to access the non-existent `slug` property when processing extensions.

### Other information:

## Testing instructions:

1. Go to `/wp-admin/admin.php?page=manage-campaigns`.
2. Turn off internet and go to `/wp-admin/admin.php?page=zerobscrm-extensions`.

In `trunk`, both of these URLs will cause PHP notices.

In `fix/crm/3129-php_errors`, the notices are no more.

### DIFF
--- a/projects/plugins/crm/changelog/fix-crm-3129-php_errors
+++ b/projects/plugins/crm/changelog/fix-crm-3129-php_errors
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Extensions: catch PHP notice if offline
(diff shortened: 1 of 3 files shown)


cleacos:
> Turn off internet 

Could you provide more info here? for example, if I'm testing it locally...

tbradsha:
> Could you provide more info here? for example, if I'm testing it locally...

e.g. disable your WiFi on your computer, or unplug your router, or block outbound requests. :^)

coder-karen (approved):
This works for me.

I was also getting the I get the `Uncaught TypeError: count(): Argument #1 ($value) must be of type Countable|array, bool given` when online as well, on `/wp-admin/admin.php?page=manage-campaigns` - but both PHP errors are gone with the PR applied.

tbradsha:
The error you got is likely the PHP 8.0 version of the one I mentioned in the issue (I think it went from a notice to a fatal). Glad it's working now, though!

PR #31055, opened 2023-05-30
Launchpad: Add is completed callback to launchpad

### LINKED ISSUE #77494: Launchpad: Move the completion check logic out of the task list availability status
Currently, the logic that checks if the task list is enabled is tied with the completion logic, which means that a task list with the completed status will make the is_enabled flag false. We need to decouple this logic to give more flexibility to Launchpad in other contexts.

- [x] Add a new callback called `is_completed_callaback` to the task list definition.
- [x] Add a new `required_task_ids` attribute to the task list definition.
- [x] Add a general-purpose function that we can use as the default completion callback.
  - [x] Identify any required tasks for the task list
  - [x] If any required tasks exist, the task list is complete if all required tasks are complete
  - [x] If no required tasks exist, we may want to look at tasks flagged with `isLaunchTask` as a secondary check for backwards-compatibility, and have completion be defined by whether all of those tasks are complete
  - [x] If no required tasks and no launch tasks exist, we can check whether all tasks are complete – if they are, then we mark the task list as complete
  - [x] If no required tasks and no launch tasks exist, and we have one or more incomplete tasks, then we can check if the last task was completed, and use that as the deciding factor
    - [x] We may want to have this be controlled by a flag rather than being a default for new task lists, as this would be confusing for task lists where the steps may not be sequential.
    - [x] In general, I think we likely want to keep the task list marked as incomplete by default, as most task list use cases are going to be non-blocking and more focused on user guidance and next steps rather than pushing people to a super-specific goal like site launch.

More context: paYKcK-30Q-p2#comment-2193

valterlorran:
Currently, how we check if the Launchpad is completed is tied to its availability status(enabled/disabled). I believe we need to have a separate status to define whether the Launchpad is completed. So we can have more flexibility in future implementations of the Launchpad in other contexts.

Closes https://github.com/Automattic/wp-calypso/issues/77494

## Proposed changes:

* Add support to a new `is_completed_callback` to the task list.
* Add a default callback to the task lists with the current behavior 
* Add the `required_task_ids` attribute to the task list definition. This attributes defines which tasks need to be completed to mark the task list as completed.
* Add the `require_last_task_completion` flag to the task list. This attribute flags if concluding the last task is enough to mark the task list as completed.

### Other information:

## Jetpack product discussion

paYKcK-30Q-p2

## Testing instructions:

* Run the unit tests with the following command:
```bash
cd projects/packages/jetpack-mu-wpcom; ./vendor/bin/phpunit --filter=Launchpad_Task_Lists_Test; cd -
```
* Now tests your changes on WPCOM, run the following command on your sandbox to apply these changes:
```bash
bin/jetpack-downloader test jetpack-mu-wpcom-plugin add/is-completed-callback-to-launchpad
```
* Create a new free site and go through the launchpad making sure it continues to work as expected.

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
@@ -61,6 +61,11 @@ class Launchpad_Task_Lists {
 			return false;
 		}
 
+		// If no is_completed_callback is set, use the default.
+		if ( ! isset( $task_list['is_completed_callback'] ) ) {
+			$task_list['is_completed_callback'] = 'wpcom_default_launchpad_task_list_completed';
+		}
+
 		$this->task_list_registry[ $task_list['id'] ] = $task_list;
 		return true;
 	}
@@ -174,6 +179,46 @@ class Launchpad_Task_Lists {
 	}
 
 	/**
+	 * Get the required task ids for a given task list.
+	 *
+	 * @param array $task_list_id Task list.
+	 * @return array Required task ids.
+	 */
+	public function get_required_task_ids( $task_list_id ) {
+		$task_list = $this->get_task_list( $task_list_id );
+		if ( ! isset( $task_list['required_task_ids'] ) ) {
+			return array();
+		}
+		return $task_list['required_task_ids'];
+	}
+
+	/**
+	 * Check if the task list requires the last task to be completed in order to consider
+	 * the task list complete.
+	 *
+	 * @param array $task_list_id Task list id.
+	 * @return bool True if the last task must be completed, false if not.
+	 */
+	public function get_require_last_task_completion( $task_list_id ) {
+		$task_list = $this->get_task_list( $task_list_id );
+		if ( ! isset( $task_list['require_last_task_completion'] ) ) {
+			return false;
+		}
+		return $task_list['require_last_task_completion'];
+	}
+
+	/**
+	 * Check if a task list is completed.
+	 *
+	 * @param string $task_list_id Task list id.
+	 * @return bool True if the task list is completed, false if not.
+	 */
+	public function is_task_list_completed( $task_list_id ) {
+		$task_list = $this->get_task_list( $task_list_id );
+		return $this->load_value_from_callback( $task_list, 'is_completed_callback', false );
+	}
+
+	/**
 	 * Get all registered Launchpad Tasks.
 	 *
 	 * @return array All registered Launchpad Tasks.
@@ -248,16 +293,16 @@ class Launchpad_Task_Lists {
 	}
 
 	/**
-	 * Given a task definition and a possible callback, call it and return the value.
+	 * Given a task or task list definition and a possible callback, call it and return the value.
 	 *
-	 * @param Task   $task A task definition.
+	 * @param array  $item     The task or task list definition.
 	 * @param string $callback The callback to attempt to call.
-	 * @param mixed  $default The default value, passed to the callback if it exists.
+	 * @param mixed  $default  The default value, passed to the callback if it exists.
 	 * @return mixed The value returned by the callback, or the default value.
 	 */
-	private function load_value_from_callback( $task, $callback, $default = '' ) {
-		if ( isset( $task[ $callback ] ) && is_callable( $task[ $callback ] ) ) {
-			return call_user_func_array( $task[ $callback ], array( $task, $default ) );
+	private function load_value_from_callback( $item, $callback, $default = '' ) {
+		if ( isset( $item[ $callback ] ) && is_callable( $item[ $callback ] ) ) {
+			return call_user_func_array( $item[ $callback ], array( $item, $default ) );
 		}
 		return $default;
 	}
@@ -373,6 +418,16 @@ class Launchpad_Task_Lists {
 
 		if ( ! isset( $task_list['task_ids'] ) ) {
 			_doing_it_wrong( 'validate_task_list', 'The Launchpad task list being registered requires a "task_ids" attribute', '6.1' );
+			return false;
+		}
+
+		if ( isset( $task_list['required_task_ids'] ) && ! is_array( $task_list['required_task_ids'] ) ) {
+			_doing_it_wrong( 'validate_task_list', 'The required_task_ids attribute must be an array', '6.1' );
+			return false;
+		}
+
+		if ( isset( $task_list['require_last_task_completion'] ) && ! is_bool( $task_list['require_last_task_completion'] ) ) {
+			_doing_it_wrong( 'validate_task_list', 'The require_last_task_completion attribute must be a boolean', '6.1' );
 			return false;
 		}
 
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
@@ -278,6 +278,116 @@ function wpcom_register_launchpad_task_list( $task_list ) {
 }
 
 /**
+ * Retrieves the required tasks from a given set of tasks.
+ *
+ * This function filters the provided tasks array and returns only the tasks that are marked as required,
+ * based on the specified array of required task IDs. It's worth noting that some tasks may not be
+ * visible to the user, so we are going to return all tasks that are marked as required, regardless
+ *
+ * @param array $required_task_ids An array of task IDs that are considered required.
+ * @param array $tasks An array of tasks to filter.
+ * @return array An array containing only the required tasks from the provided task list.
+ */
+function wpcom_launchpad_get_required_tasks( $required_task_ids, $tasks ) {
+	return array_filter(
+		$tasks,
+		function ( $task ) use ( $required_task_ids ) {
+			return in_array( $task['id'], $required_task_ids, true );
+		}
+	);
+}
+
+/**
+ * Get all tasks that are marked as launch tasks.
+ *
+ * @param array $tasks Array of tasks.
+ * @return array Array of launch tasks.
+ */
+function wpcom_launchpad_get_launch_tasks( $tasks ) {
+	return array_filter(
+		$tasks,
+		function ( $task ) {
+			if ( isset( $task['isLaunchTask'] ) ) {
+				return $task['isLaunchTask'];
+			}
+			return false;
+		}
+	);
+}
+
+/**
+ * Check if all the given tasks are completed.
+ *
+ * @param array $tasks Array of tasks to check if they are completed.
+ * @return bool True if all tasks are completed, false otherwise.
+ */
+function wpcom_launchpad_are_all_tasks_completed( $tasks ) {
+	return array_reduce(
+		$tasks,
+		function ( $carry, $launch_task ) {
+			return $carry && $launch_task['completed'];
+		},
+		true
+	);
+}
+
+/**
+ * Determine the completion status of a task list based on its tasks.
+ *
+ * This function identifies any required tasks within the task list. If there are required tasks, the task list is considered complete
+ * only if all required tasks are completed. In the absence of required tasks, a secondary check is performed on tasks flagged with
+ * "isLaunchTask" to determine completion based on whether all of those tasks are complete. If there are no required tasks and no
+ * launch tasks, the task list's completion is determined by checking whether all tasks are complete. However, if there are incomplete
+ * tasks in this scenario, the completion status is based on whether the last task was completed. It is possible to control this behavior
+ * using a flag, as the default approach might be confusing for task lists where the steps may not be sequential. By default, the task
+ * list is marked as incomplete to align with most use cases that focus on user guidance and next steps rather than a specific end goal.
+ *
+ * @param array $task_list An array of tasks within the task list.
+ * @return bool True if the task list is considered complete, false otherwise.
+ */
+function wpcom_default_launchpad_task_list_completed( $task_list ) {
+	$task_list_id      = $task_list['id'];
+	$required_task_ids = wpcom_launchpad_checklists()->get_required_task_ids( $task_list_id );
+	$all_visible_tasks = wpcom_launchpad_checklists()->build( $task_list_id );
+
+	// If there are required tasks, check if they are all completed.
+	if ( ! empty( $required_task_ids ) ) {
+		$required_tasks = wpcom_launchpad_get_required_tasks( $required_task_ids, $all_visible_tasks );
+		return wpcom_launchpad_are_all_tasks_completed( $required_tasks );
+	}
+
+	// If there are no required tasks, check if there are any launch tasks.
+	$launch_tasks = wpcom_launchpad_get_launch_tasks( $all_visible_tasks );
+	if ( ! empty( $launch_tasks ) ) {
+		return wpcom_launchpad_are_all_tasks_completed( $launch_tasks );
+	}
+
+	// If there are no required tasks and no launch tasks, check if all tasks are completed.
+	if ( wpcom_launchpad_are_all_tasks_completed( $all_visible_tasks ) ) {
+		return true;
+	}
+
+	// If there are incomplete tasks, check if the last task was completed.
+	$require_last_task_completion = wpcom_launchpad_checklists()->get_require_last_task_completion( $task_list_id );
+	$last_task                    = end( $all_visible_tasks );
+	if ( $require_last_task_completion && $last_task['completed'] ) {
+		return true;
+	}
+
+	return false;
+}
+
+/**
+ * Callback to determine the completion status of a task list.
+ *
+ * @param string $task_list_id The task list ID.
+ * @return bool True if the task list is considered complete, false otherwise.
+ */
+function wpcom_launchpad_is_task_list_completed( $task_list_id ) {
+	return wpcom_launchpad_checklists()->is_task_list_completed( $task_list_id );
+}
+
+/**
  * Wrapper that registers a launchpad checklist.
  *
  * @param Task $tasks Collection of Task definitions.
--- a/projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-list-validation-test.php
+++ b/projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-list-validation-test.php
@@ -1,0 +1,102 @@
+<?php
+/**
+ * Test class for Launchpad_Task_Lists.
+ *
+ * @package automattic/jetpack-mu-wpcom
+ */
+
+/**
+ * Test class for Launchpad_Task_Lists.
+ *
+ * @coversDefaultClass Launchpad_Task_Lists
+ */
+class Launchpad_Task_List_Validation_Test extends \WorDBless\BaseTestCase {
+	/**
+	 * Data provider for test_validate_task_list.
+	 *
+	 * @return array
+	 */
+	public function provide_validate_task_list_test_cases() {
+		return array(
+			'Valid task list with required task IDs and last task completion' => array(
+				array(
+					'id'                           => 'task_list_1',
+					'task_ids'                     => array( 'task_1', 'task_2' ),
+					'required_task_ids'            => array( 'task_1' ),
+					'require_last_task_completion' => true,
+				),
+				true,
+			),
+			'Valid task list with last task completion'   => array(
+				array(
+					'id'                           => 'task_list_1',
+					'task_ids'                     => array( 'task_1', 'task_2' ),
+					'require_last_task_completion' => true,
+				),
+				true,
+			),
+			'Valid task list with only required task IDs' => array(
+				array(
+					'id'                => 'task_list_1',
+					'task_ids'          => array( 'task_1', 'task_2' ),
+					'required_task_ids' => array( 'task_1' ),
+				),
+				true,
+			),
+			'Valid task list with minimal validation'     => array(
+				array(
+					'id'       => 'task_list_1',
+					'task_ids' => array( 'task_1', 'task_2' ),
+				),
+				true,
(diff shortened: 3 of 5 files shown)


projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -152,6 +157,46 @@ public function get_task( $id ) {
...
+	 *
+	 * @param array $task_list_id Task list.
+	 * @return array|null Required task ids.
+	 */
+	public function get_required_task_ids( $task_list_id ) {
+		$task_list = $this->get_task_list( $task_list_id );
+		if ( ! isset( $task_list['required_task_ids'] ) ) {
+			return array();
+		}
+		return $task_list['required_task_ids'];
```
daledupreez, line 192: As a safety net, can we also add logic to `validate_task_list()` to verify that if `required_task_ids` is present, it is specified as an array?

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -152,6 +157,46 @@ public function get_task( $id ) {
 		return $this->task_registry[ $id ];
 	}
 
+	/**
+	 * Get the required task ids for a given task list.
+	 *
+	 * @param array $task_list_id Task list.
+	 * @return array|null Required task ids.
```
daledupreez: Should this not return `array` in all cases?
```suggestion
	 * @return array Required task ids.
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -152,6 +157,46 @@ public function get_task( $id ) {
...
+	 *
+	 * @param array $task_list_id Task list id.
+	 * @return bool True if the last task must be completed, false if not.
+	 */
+	public function get_require_last_task_completion( $task_list_id ) {
+		$task_list = $this->get_task_list( $task_list_id );
+		if ( ! isset( $task_list['require_last_task_completion'] ) ) {
+			return false;
+		}
+		return $task_list['require_last_task_completion'];
```
daledupreez, line 207: As above, does it make sense to validate that this value is a boolean in `validate_task_list()`?

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -227,16 +272,16 @@ private function build_task( $task ) {
 	}
 
 	/**
-	 * Given a task definition and a possible callback, call it and return the value.
+	 * Given a task or task list definition and a possible callback, call it and return the value.
 	 *
-	 * @param Task   $task A task definition.
-	 * @param string $callback The callback to attempt to call.
-	 * @param mixed  $default The default value, passed to the callback if it exists.
+	 * @param array|Task $item     The task or task list definition.
```
daledupreez: It may be better to use `array` here in all cases, as the `Task` class isn't defined.
```suggestion
	 * @param array $item     The task or task list definition.
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
```diff
@@ -608,6 +608,112 @@ function wpcom_register_launchpad_task_list( $task_list ) {
...
+	return false;
+}
+
+/**
+ * Callback to determine the completion status of a task list.
+ *
+ * @param string $task_list_id The task list ID.
+ * @return bool True if the task list is considered complete, false otherwise.
+ */
+function wpcom_launchpad_is_completed( $task_list_id ) {
```
daledupreez: I think this function should specify that it is for a task list.
```suggestion
function wpcom_launchpad_is_task_list_completed( $task_list_id ) {
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
```diff
@@ -608,6 +608,112 @@ function wpcom_register_launchpad_task_list( $task_list ) {
...
+ * using a flag, as the default approach might be confusing for task lists where the steps may not be sequential. By default, the task
+ * list is marked as incomplete to align with most use cases that focus on user guidance and next steps rather than a specific end goal.
+ *
+ * @param array $task_list An array of tasks within the task list.
+ * @return bool True if the task list is considered complete, false otherwise.
+ */
+function wpcom_default_launchpad_task_list_completed( $task_list ) {
+	$task_list_id      = $task_list['id'];
+	$required_task_ids = wpcom_launchpad_checklists()->get_required_task_ids( $task_list_id );
+	$tasks             = wpcom_launchpad_checklists()->build( $task_list_id );
```
daledupreez: Maybe call this `$all_tasks` to make it clearer what the list represents.

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-lists-test.php
```diff
@@ -22,4 +22,86 @@ public function test_build_creates_no_PHP_warnings() {
...
+			array( array( $incomplete_task, $launch_task_completed ), true ),
+			array( array( $incomplete_task, $completed_task ), true, array( 'require_last_task_completion' => true ) ),
+			array( array( $completed_task, $incomplete_task ), false, array( 'require_last_task_completion' => true ) ),
+			array( array( $completed_task, $incomplete_task ), true, array( 'required_task_ids' => array( 'task_0' ) ) ),
+			array( array( $incomplete_task, $incomplete_task ), false, array( 'required_task_ids' => array( 'task_0' ) ) ),
+			array( array( $incomplete_task, $completed_task ), false, array( 'required_task_ids' => array( 'task_0' ) ) ),
+			array( array( $incomplete_task, $launch_task_completed ), false, array( 'required_task_ids' => array( 'task_0' ) ) ),
+			array( array( $incomplete_task, $incomplete_task ), true, array( 'is_completed_callback' => '__return_true' ) ),
+			array( array( $completed_task, $incomplete_task ), true, array( 'is_completed_callback' => array( $this, 'callback_first_task_completed' ) ) ),
+		);
```
daledupreez: It would be really helpful to describe each test case. I've suggested a few keys, but not all.
```suggestion
		return array(
			// tasks, expected, task_list_options
			'all tasks incomplete should be incomplete using default options' => array( array( $incomplete_task, $incomplete_task ), false ),
			'all tasks complete should be complete using default options' => array( array( $completed_task, $completed_task ), true ),
			'incomplete task and incomplete launch task should be incomplete using default options' => array( array( $incomplete_task, $launch_task ), false ),
			array( array( $incomplete_task, $launch_task_completed ), true ),
			array( array( $incomplete_task, $completed_task ), true, array( 'require_last_task_completion' => true ) ),
			array( array( $completed_task, $incomplete_task ), false, array( 'require_last_task_completion' => true ) ),
			array( array( $completed_task, $incomplete_task ), true, array( 'required_task_ids' => array( 'task_0' ) ) ),
			array( array( $incomplete_task, $incomplete_task ), false, array( 'required_task_ids' => array( 'task_0' ) ) ),
			array( array( $incomplete_task, $completed_task ), false, array( 'required_task_ids' => array( 'task_0' ) ) ),
			array( array( $incomplete_task, $launch_task_completed ), false, array( 'required_task_ids' => array( 'task_0' ) ) ),
			array( array( $incomplete_task, $incomplete_task ), true, array( 'is_completed_callback' => '__return_true' ) ),
			array( array( $completed_task, $incomplete_task ), true, array( 'is_completed_callback' => array( $this, 'callback_first_task_completed' ) ) ),
		);
```

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-lists-test.php
```diff
@@ -22,4 +22,86 @@ public function test_build_creates_no_PHP_warnings() {
...
+	/**
+	 * Data provider for ::test_task_list_is_completed().
+	 */
+	public function provider_test_task_list_is_completed() {
+		$incomplete_task = array();
+		$completed_task  = array(
+			'is_complete_callback' => '__return_true',
+		);
+
+		$launch_task = array(
```
daledupreez: I think it would be easier to understand the setup if you included `incomplete` in the variable name.
```suggestion
		$launch_task_incomplete = array(
```

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-lists-test.php
```diff
@@ -22,4 +22,86 @@ public function test_build_creates_no_PHP_warnings() {
...
+	 * @param array $tasks             Array of tasks to register.
+	 * @param bool  $expected          Expected result of wpcom_launchpad_is_completed().
+	 * @param array $task_list_options Optional. Array of options to pass to wpcom_register_launchpad_task_list().
+	 */
+	public function test_task_list_is_completed( $tasks, $expected, $task_list_options = array() ) {
+		$task_ids = array();
+		foreach ( $tasks as $key => $task ) {
+			$task_id    = 'task_' . $key;
+			$task_ids[] = $task_id;
+			wpcom_register_launchpad_task(
```
daledupreez, line 111: Does it make sense to reset the data for tasks and task lists before each test run? We might see unexpected bleedover/behaviour if we don't.

daledupreez (commented):
I have a number of comments from inspection -- I haven't tested yet.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -355,6 +400,16 @@ public static function validate_task_list( $task_list ) {
 			return false;
 		}
 
+		if ( isset( $task_list['required_task_ids'] ) && ! is_array( $task_list['required_task_ids'] ) ) {
+			_doing_it_wrong( 'validate_task_list', 'The required_task_ids attribute must be an array', '6.1' );
+			return false;
```
daledupreez, line 426: Possible future improvement for the API: have this function (or an equivalent) return `null | WP_Error[]` so all the issues can be identified. _Definitely_ non-blocking and more a reaction to reading this code again!

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
```diff
@@ -608,6 +608,112 @@ function wpcom_register_launchpad_task_list( $task_list ) {
 	return wpcom_launchpad_checklists()->register_task_list( $task_list );
 }
 
+/**
+ * Retrieves the required tasks from a given set of tasks.
+ *
+ * This function filters the provided tasks array and returns only the tasks that are marked as required, based on the specified
+ * array of required task IDs.
```
daledupreez: Nit: It may be worth introducing the newline earlier.
```suggestion
 * This function filters the provided tasks array and returns only the tasks that are marked as required,
 * based on the specified array of required task IDs.
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
```diff
@@ -608,6 +608,112 @@ function wpcom_register_launchpad_task_list( $task_list ) {
...
+ * Get all tasks that are marked as launch tasks.
+ *
+ * @param array $tasks Array of tasks.
+ * @return array Array of launch tasks.
+ */
+function wpcom_launchpad_get_launch_tasks( $tasks ) {
+	return array_filter(
+		$tasks,
+		function ( $task ) {
+			return $task['isLaunchTask'];
```
daledupreez: Just to be extra safe, it may be worth checking if the field is set. Also, `??` was only added in PHP 7, hence the use of `isset()`.
```suggestion
			if ( isset( $task['isLaunchTask'] ) ) {
				return $task['isLaunchTask'];
			}
			return false;
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
```diff
@@ -608,6 +608,112 @@ function wpcom_register_launchpad_task_list( $task_list ) {
...
+ * using a flag, as the default approach might be confusing for task lists where the steps may not be sequential. By default, the task
+ * list is marked as incomplete to align with most use cases that focus on user guidance and next steps rather than a specific end goal.
+ *
+ * @param array $task_list An array of tasks within the task list.
+ * @return bool True if the task list is considered complete, false otherwise.
+ */
+function wpcom_default_launchpad_task_list_completed( $task_list ) {
+	$task_list_id      = $task_list['id'];
+	$required_task_ids = wpcom_launchpad_checklists()->get_required_task_ids( $task_list_id );
+	$all_tasks         = wpcom_launchpad_checklists()->build( $task_list_id );
```
daledupreez: For clarity, it may be well worth calling this variable `$all_visible_tasks` - I was just wondering how we should handle non-visible tasks that are either required or launch tasks, and I discovered that the `build()` function removes all non-visible tasks.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
daledupreez: It may also be worth noting that some tasks may not be visible to the user, so we will only return required tasks that are in the task list, meaning we may be "missing" some tasks.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
daledupreez: Maybe worth handling in a follow-up, but it might be a good idea to validate that all the required task IDs are actually in the task list. 😁

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-list-validation-test.php
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * Test class for Launchpad_Task_Lists.
+ *
+ * @coversDefaultClass Launchpad_Task_Lists
+ */
+class Launchpad_Task_List_Validation_Test extends \WorDBless\BaseTestCase {
+	/**
+	 * Test valid task lists.
+	 */
+	public function test_valid_task_list() {
```
daledupreez: Non-blocking thought: there is a lot of repetition across the validation tests that makes it harder for me to understand the reasoning behind each test case. Personally, I would find it easier to use a data provider with descriptive messages for the test cases and a simple test runner. Something like the following.
```suggestion
	public function provide_validate_task_list_test_cases() {
		return array(
			'Valid task list with required task IDs and last task completion' => array(
				array(
					'id'                           => 'task_list_1',
					'task_ids'                     => array( 'task_1', 'task_2' ),
					'required_task_ids'            => array( 'task_1' ),
					'require_last_task_completion' => true,
				),
				true,
			),
			'Valid task with last task completion' => array(
				array(
					'id'                           => 'task_list_1',
					'task_ids'                     => array( 'task_1', 'task_2' ),
					'require_last_task_completion' => true,
				),
				true,
			),
			...
			'Invalid task list with no id' => array(
				array(
					'task_ids'                     => array( 'task_1', 'task_2' ),
					'required_task_ids'            => array( 'task_1' ),
					'require_last_task_completion' => true,
				),
				false,
			),
			'Invalid task list with no task_ids' => array(
				array(
					'id'                           => 'task_list_1',
					'required_task_ids'            => array( 'task_1' ),
					'require_last_task_completion' => true,
				),
				false
			),
			...
		);
	}

	/**
	 * @dataProvider provide_validate_task_list_test_cases
	 */
	public function test_validate_task_list( $task_list, $expected_result ) {
		$result = Launchpad_Task_Lists::validate_task_list( $task_list );
		
		$this->assertSame( $expected_result, $result );
	}
```

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-lists-test.php
```diff
@@ -22,4 +22,101 @@ public function test_build_creates_no_PHP_warnings() {
 		$this->assertIsArray( $result );
 		$this->assertEmpty( $result );
 	}
+
+	/**
+	 * Callback to check if the first task in a task list is completed.
+	 *
+	 * @param array $task_list Task list.
+	 */
+	public function callback_first_task_completed( $task_list ) {
```
daledupreez: It may be worth adding "is" to the function name to make it clearer what we're doing.
```suggestion
	public function callback_is_first_task_completed( $task_list ) {
```

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-lists-test.php
```diff
@@ -22,4 +22,101 @@ public function test_build_creates_no_PHP_warnings() {
...
+			'with the require_last_task_completion option, the last task being complete should make the task list complete'
+				=> array( array( $incomplete_task, $completed_task ), true, array( 'require_last_task_completion' => true ) ),
+			'with the require_last_task_completion option, the last task being incomplete should make the task list incomplete'
+				=> array( array( $completed_task, $incomplete_task ), false, array( 'require_last_task_completion' => true ) ),
+			'with the required_task_ids option, the task list should be complete if the required task is complete'
+				=> array( array( $completed_task, $incomplete_task ), true, array( 'required_task_ids' => array( 'task_0' ) ) ),
+			'with the required_task_ids option, the task list should be incomplete if the required task is incomplete'
+				=> array( array( $incomplete_task, $incomplete_task ), false, array( 'required_task_ids' => array( 'task_0' ) ) ),
+			'with the required_task_ids option and the last task incomplete, the task list should be incomplete if the required task is incomplete'
+				=> array( array( $incomplete_task, $completed_task ), false, array( 'required_task_ids' => array( 'task_0' ) ) ),
```
daledupreez: The data for this test case doesn't match the description -- we should also have `$incomplete_task` as the final task to match the description. (Though we could also change the description!)
```suggestion
			'with the required_task_ids option and the last task incomplete, the task list should be incomplete if the required task is incomplete'
				=> array( array( $incomplete_task, $completed_task, $incomplete_task ), false, array( 'required_task_ids' => array( 'task_0' ) ) ),
```

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-lists-test.php
```diff
@@ -22,4 +22,101 @@ public function test_build_creates_no_PHP_warnings() {
...
+			'the launch task completion should not mark the task list complete if the required task is incomplete'
+				=> array( array( $incomplete_task, $launch_task_completed ), false, array( 'required_task_ids' => array( 'task_0' ) ) ),
+			'overriding the is_completed_callback should be the only thing that matters'
+				=> array( array( $incomplete_task, $incomplete_task ), true, array( 'is_completed_callback' => '__return_true' ) ),
+			'custom is_completed_callback should be able to check if first task is complete'
+				=> array( array( $completed_task, $incomplete_task ), true, array( 'is_completed_callback' => array( $this, 'callback_first_task_completed' ) ) ),
```
daledupreez: Question: can we use an inline function for the callback? I don't know if that actually helps with overall readability, but it does keep the logic closer to the test case.
```suggestion
				=> array(
		array( $completed_task, $incomplete_task ),
		true,
		array(
			'is_completed_callback' => function ( $task_list ) {
				$first_task = reset( wpcom_launchpad_checklists()->build( $task_list['id'] ) );
				return $first_task['completed'];
			},
		),
	),
```

daledupreez (commented):
I have more feedback from inspection, will try to test in an hour or two.

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-lists-test.php
valterlorran: I think I meant to check that the required_task_ids option overrides the require_last_task_completion. I fixed the test and the description.

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-lists-test.php
```diff
@@ -22,4 +22,107 @@ public function test_build_creates_no_PHP_warnings() {
 		$this->assertIsArray( $result );
 		$this->assertEmpty( $result );
 	}
+
+	/**
+	 * Data provider for ::test_task_list_is_completed().
```
daledupreez: You should be able to use an inline `@see` here to make this link more explicit.
```suggestion
	 * Data provider for {@see test_task_list_is_completed()}.
```

daledupreez (dismissed):
Tested and reviewed after the recent changes, and this LGTM! 🚢

daledupreez (commented):
This got rebased, so I will need to take a closer look tomorrow, as I can't confirm what changed. I _think_ the only change is a unit test fix, but I'll confirm that tomorrow.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
valterlorran: Create the issues so we can tackle them:
* https://github.com/Automattic/wp-calypso/issues/77939
* https://github.com/Automattic/wp-calypso/issues/77942

daledupreez (approved):
The changes still look good to me. I'll get this merged shortly.

PR #31058, opened 2023-05-30
[Boost] Remove invalid apply_filter on minified CSS

thingalon:
Filter `style_loader_tag`'s [second parameter](https://developer.wordpress.org/reference/hooks/style_loader_tag/) is a string containing one tag. When minifying CSS on a site with some plugins installed which responded to `style_loader_tag`, passing in an array of handles where a single string is expected can be problematic.

This PR fixes it by removing the invalid filter call, in line with the [Page Optimize implementation](https://github.com/Automattic/page-optimize/blob/8e4c9ef79c374b25ca0ca41210f2e8feeec0685e/concat-css.php#LL186C1-L186C1).

## Proposed changes:
* Remove an invalid apply_filters call.

### Other information:

## Jetpack product discussion

n/a

no

## Testing instructions:
* Use Boost's minify css alongside a plugin which defines a callback to `style_loader_tag`, such as wp-ocean's companion plugin.
* Ensure that it works properly.

### DIFF
--- a/projects/plugins/boost/app/lib/minify/Concatenate_CSS.php
+++ b/projects/plugins/boost/app/lib/minify/Concatenate_CSS.php
@@ -193,7 +193,6 @@ class Concatenate_CSS extends WP_Styles {
 				}
 
 				$style_tag = apply_filters( 'page_optimize_style_loader_tag', $style_tag, $handles, $href, $media );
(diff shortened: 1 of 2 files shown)


haqadn (approved):
Can confirm the issue and fix 👍🏼

thingalon:
Cherry picked into 1.9.0 - dd808a47a0

PR #31059, opened 2023-05-30
[Boost] Avoid double-prefixing url() values in minified CSS

thingalon:
When minifying CSS for a WordPress site in a subdirectory, the subdirectory name gets added twice to `url()` entries in the CSS when trying to absolute-ify the paths.

This fix prevents that by stripping the site's dir prefix out of the css url() prefix (if present).

## Proposed changes:
* Prevent double-prefixing of CSS paths in WordPress sites inside a subdirectory.

### Other information:

## Jetpack product discussion

n/a

no

## Testing instructions:
* Create a WordPress site in a subdirectory.
* Use a theme which imports its own custom fonts, or otherwise uses url() elements pointing to other resources in the site directory (not external resources)
* Verify that with this patch, these resources are loaded properly.

### DIFF
--- a/projects/plugins/boost/app/lib/minify/functions-service.php
+++ b/projects/plugins/boost/app/lib/minify/functions-service.php
@@ -119,6 +119,20 @@ function jetpack_boost_page_optimize_service_request() {
 }
 
 /**
+ * Strip matching parent paths off a string. Returns $path without $parent_path.
+ */
+function jetpack_boost_strip_parent_path( $parent_path, $path ) {
+	$trimmed_parent = ltrim( $parent_path, '/' );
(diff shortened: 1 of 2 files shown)


haqadn:
I wasn't able to reproduce the issue so cannot confirm if it works. Here is the output I get regardless of testing with `rc` and this branch:

```css
@font-face { font-family: roboto-regular; 
			 src: url(/polarbear/wp-content/themes/oceanwp/assets/css/font/Roboto-Regular.ttf); }
```

haqadn (approved):
👍🏼

thingalon:
Cherry-picked into 1.9.0 - cbc1c8335e

PR #31080, opened 2023-05-30
Publicize: request wpcom for Mastodon feature check

### LINKED ISSUE #31079: Publicize: Mastodon feature is not updated

monsieur-z:
Fixes #31079

## Proposed changes:

In the `publicize` package, the value returned by `has_mastodon_connection_feature` is cached. Switching on and off the Mastodon feature in the backend has therefore no effect in the UI of the Publicize sidebar. This PR fixes it by always requesting the backend when checking for the Mastodon feature.

### Other information:

## Jetpack product discussion

Project thread: pdrWKz-Tm-p2

No.

### Prerequisites
- Note your test site id
- Make sure you know how to toggle the Mastodon feature on and off (D111014-code)
- Check that Social is activated

### Testing
- Create a new post
- Click on the Jetpack icon to show the Publicize sidebar
Screenshot 2023-05-30 at 11 37 03 AM

- If the Mastodon feature is disabled, make sure you don't see the Mastodon icon in the _Social Previews_ section or in the previews modal. If it's enabled, verify you see it.
Screenshot 2023-05-30 at 11 25 52 AM
Screenshot 2023-05-30 at 11 25 57 AM

- Toggle the feature
- Check the presence/absence of the Mastodon icon

### DIFF
--- a/projects/packages/publicize/src/class-publicize-base.php
+++ b/projects/packages/publicize/src/class-publicize-base.php
@@ -1799,7 +1799,7 @@ abstract class Publicize_Base {
 	 * @return bool
 	 */
 	public function has_mastodon_connection_feature() {
-		return Current_Plan::supports( 'social-mastodon-connection' );
+		return Current_Plan::supports( 'social-mastodon-connection', true );
 	}
 
 	/**
--- a/projects/packages/publicize/changelog/fix-publicize-mastodon-feature-cache
+++ b/projects/packages/publicize/changelog/fix-publicize-mastodon-feature-cache
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Always request wpcom for Mastodon feature check
--- a/projects/packages/publicize/package.json
+++ b/projects/packages/publicize/package.json
@@ -1,7 +1,7 @@
 {
 	"private": true,
 	"name": "@automattic/jetpack-publicize",
-	"version": "0.30.0",
+	"version": "0.30.1-alpha",
 	"description": "Publicize makes it easy to share your site’s posts on several social media networks automatically when you publish a new post.",
 	"homepage": "https://github.com/Automattic/jetpack/tree/HEAD/projects/packages/publicize/#readme",
 	"bugs": {


projects/packages/publicize/src/class-publicize-base.php
```diff
@@ -1799,7 +1799,7 @@ public function has_instagram_connection_feature() {
 	 * @return bool
 	 */
 	public function has_mastodon_connection_feature() {
-		return Current_Plan::supports( 'social-mastodon-connection' );
+		return Current_Plan::supports( 'social-mastodon-connection', true );
```
pablinos, line 1802: As discussed, I think it's fine to leave this as is. The plan data will be refreshed at least every 24 hours on the `jetpack_heartbeat`, when the plan changes, or if the dashboard is visited.

It means the feature won't be available straightaway for everyone but within 24 hours it will be.

monsieur-z: What do you think, @kraftbj? We'll need to take that delay into consideration when setting up test sites.

pablinos: You can force the update on a test site by going to the Jetpack dashboard or running `do_action( 'jetpack_heartbeat' )` in the wp-cli.

As the check is run on editor load it's probably best to deal with the caching rather than making lots of unnecessary API calls. We've also previously caused some performance issues on WPCOM/Atomic doing that.

monsieur-z: > You can force the update on a test site by going to the Jetpack dashboard 

Just by visiting the page in WPadmin?

zinigor: What if instead of always refreshing the cache we could refresh it once when actually switching the feature on and off in the backend?

anomiex:
> @[monsieur-z](https://github.com/monsieur-z) monsieur-z added [Status] Needs Cherry-Pick label [May 30, 2023](https://github.com/Automattic/jetpack/pull/31080#event-9380146666)

I think at this point it would be fairly late to cherry-pick this, particularly since I find myself agreeing with @pablinos 's concern over bypassing the cache potentially causing issues for wpcom and the workaround being just to wait a few more hours for the cache to expire.

monsieur-z:
> I think at this point it would be fairly late to cherry-pick this

@anomiex: yes we'll take another approach. I'll close the PR and add more details afterwards.

projects/packages/publicize/src/class-publicize-base.php
pablinos: > Just by visiting the page in WPadmin?

Yes, that's right. It triggers the heartbeat action I think, but definitely refreshes the plan data.

> What if instead of always refreshing the cache we could refresh it once when actually switching the feature on and off in the backend?

We could do that. There is an API endpoint on the site that is pinged as a plan is activated/deactivated, so we could do that for these sites.

I think it is fine to wait. I don't know if the plan data is refreshed after plugin upgrade, but that would be perfect if it is. Otherwise, the features will become available within 24 hours and if someone contacts us asking where they are, we can tell them to visit the Jetpack Dashboard. That's not ideal, but it's better than refreshing the plan data all the time.

monsieur-z:
For the record, this PR was made irrelevant by D112615-code. Mastodon is now enabled on self-hosted sites with a Jetpack version above 12.1.

PR #31084, opened 2023-05-30
Add tests to boost score api package

CodeyGuyDylan:
## Proposed changes:
Adds tests for the boost-score-api `js-package` 

### Other information:

N/A

## Jetpack product discussion

N/A

No

## Testing instructions:

1. You can just make sure tests pass and it should be good to go

If you want to test them locally

2. Checkout this branch
3. Go to the `projects/js-packages/boost-score-api` directory and run `pnpm run test`
4. Make sure all tests pass

### DIFF
--- a/projects/js-packages/boost-score-api/tests/index.test.js
+++ b/projects/js-packages/boost-score-api/tests/index.test.js
@@ -57,16 +57,18 @@ describe( 'didScoresChange', () => {
 	} );
 
 	it( 'should return true if scores changed', () => {
-		mockData.scores.noBoost.desktop = 60;
-		mockData.scores.noBoost.mobile = 50;
+		const changedMockData = Object.assign( {}, mockData );
+		changedMockData.scores.noBoost.desktop = 60;
+		changedMockData.scores.noBoost.mobile = 50;
 
-		expect( didScoresChange( mockData.scores ) ).toBe( true );
+		expect( didScoresChange( changedMockData.scores ) ).toBe( true );
 	} );
 } );
 
 describe( 'getScoreMovementPercentage', () => {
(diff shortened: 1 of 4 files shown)


projects/js-packages/boost-score-api/tests/index.test.js
```diff
@@ -1,8 +1,86 @@
...
+} );
+
+describe( 'didScoresChange', () => {
+	it( 'Should return false if scores did not change', () => {
+		expect( didScoresChange( mockData.scores ) ).toBe( false );
+	} );
+
+	it( 'should return true if scores changed', () => {
+		mockData.scores.noBoost.desktop = 60;
+		mockData.scores.noBoost.mobile = 50;
```
robertsreberski: Maybe we should create and work on a copy of `mockData` object instead? So as not to modify a global object that we'll reuse in other tests. It might just lead to unexpected errors.
```suggestion
		const changedMockData = Object.assign({}, mockData);
		changedMockData.scores.noBoost.desktop = 60;
		changedMockData.scores.noBoost.mobile = 50;
```

Other option I would see is to define `mockData` as `let` and re-define it in `beforeEach` fn.

robertsreberski (commented):
Tests look good, and pass without raising issues. I just left a comment to make sure we don't pollute global-scope objects during specific tests.

projects/js-packages/boost-score-api/tests/index.test.js
CodeyGuyDylan: You're right, this is better. I'll get that updated, thanks!

robertsreberski (approved):
Looks good to me!

PR #31087, opened 2023-05-30
CRM: Resolves 3133 - allow quote sort by status

tbradsha:
Resolves Automattic/zero-bs-crm#3133 - allow quote sort by status

## Proposed changes:

It seems DAL uses table fields to do sorts, with fallbacks for other fields. Since technically quotes don't have a real status and there wasn't a fallback, the query fails. This PR adds a hacky fallback sort:

`CASE WHEN zbsq_accepted > 0 THEN 2 WHEN zbsq_template > 0 THEN 1 ELSE 0 END`

### Other information:

## Testing instructions:

1. Go to quote listview.
2. Create three quotes (one for each status: draft, unaccepted, accepted).
3. Sort by status.

In `trunk` the listview will show an error.

In the `fix/crm/3133-quote_sort_by_status` branch, the listview properly loads:

* Ascending mode: draft, unaccepted, accepted
* Descending mode: accepted, unaccepted, draft

### DIFF
--- a/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Quotes.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Quotes.php
@@ -797,12 +797,12 @@ class zbsDAL_quotes extends zbsDAL_ObjectLayer {
 
             // Mapped sorts
             // This catches listview and other exception sort cases
-            $sort_map = array(
+				$sort_map = array(
+					'customer' => '(SELECT zbsol_objid_to FROM ' . $ZBSCRM_t['objlinks'] . ' WHERE zbsol_objtype_from = ' . ZBS_TYPE_QUOTE . ' AND zbsol_objtype_to = ' . ZBS_TYPE_CONTACT . ' AND zbsol_objid_from = quote.ID)', // phpcs:ignore WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
+					// hack to support quote statuses (for here only, 2 is accepted, 1 is unaccepted, 0 is draft)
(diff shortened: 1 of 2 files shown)


projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Quotes.php
```diff
@@ -797,12 +797,12 @@ public function getQuotes($args=array()){
 
             // Mapped sorts
             // This catches listview and other exception sort cases
-            $sort_map = array(
+				$sort_map = array(
+					'customer' => '(SELECT zbsol_objid_to FROM ' . $ZBSCRM_t['objlinks'] . ' WHERE zbsol_objtype_from = ' . ZBS_TYPE_QUOTE . ' AND zbsol_objtype_to = ' . ZBS_TYPE_CONTACT . ' AND zbsol_objid_from = quote.ID)', // phpcs:ignore WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
+					// hack to support quote statuses (for here only, 2 is accepted, 1 is unaccepted, 0 is draft)
+					'status'   => '(CASE WHEN zbsq_accepted > 0 THEN 2 WHEN zbsq_template > 0 THEN 1 ELSE 0 END)',
```
cleacos, line 803: Interesting, I think it's the first time I see using the CASE function...

2 => Accepted
1 => Not accepted yet
0 => Draft

cleacos (approved):
It works as expeted.

LGTM!

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Quotes.php
tbradsha: Yeah, ideally I'd just use the columns for sort, but the sort direction is built for use with only one column, so this was the best I could come up with for now.

PR #31093, opened 2023-05-30
AI Assistant: Add "Accept" button so the user can accept partially generated content

### LINKED ISSUE #31038: AI Assistant: allow the Done button when request fails in the middle
A request failed when it was almost done and there's no button to accept it

Image

lhkowalski:
Fixes #31038.

## Proposed changes:

* Add Accept button so the user can accept a partially generated content
* The button will be shown anytime the Retry button would appear, but only when there is at least one piece of content to accept
* One of the use cases is when a network error happens

### Other information:

## Testing instructions:

* Go to the block editor and add an AI Assistant block
* Open your browser console on the network activity tab
* Throttle your connection to simulate a slow 3G connection:

Screen Shot 2023-05-30 at 18 05 00

* Write a prompt that will probably result on a long response from the assistant, like "5 paragraphs on [your subject]"
* Execute the prompt and, while the response is being rendered, change your network throttling settings to "offline"; this way we can simulate a network issue
* Notice that an error message is shown by the block, as well as a "Retry" button and the new "Accept" button
* Try clicking the "Accept" button; it will insert the partial content on the editor
* The section below describes the before and after behavior

**Before:**

Screen Shot 2023-05-30 at 17 58 58

* There was no way to use the partial content.

---

**After:**

Screen Shot 2023-05-30 at 17 55 30

* If accepted, the content is inserted regularly:

Screen Shot 2023-05-30 at 17 56 46

* If there is no content before the error, the button will not show up:

Screen Shot 2023-05-30 at 18 03 18

### DIFF
--- a/projects/plugins/jetpack/changelog/update-ai-assistant-add-accept-button-for-partial-content
+++ b/projects/plugins/jetpack/changelog/update-ai-assistant-add-accept-button-for-partial-content
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
+AI Assistant: Add Accept button so the user can accept a partially generated content.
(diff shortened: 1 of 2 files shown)


nunyvega (approved):
Works as advised!

kraftbj:
[monthly/branch-2023-05-29 e8a992ab7e]

PR #31101, opened 2023-05-31
Boost: Update CSS regen notice UX

dilirity:
Addresses https://github.com/Automattic/boost-cloud/issues/217

## Proposed changes:

* Update DB value used to check if regen is necessary, to include what triggered it;
* Update the regenerate Critical CSS notice to include a more descriptive text based on what triggered it;
* Update `Regenerate` button to a CTA when regen is necessary (reverts back to the original after regen is done);
* Add issue to Site Health panel in case a Critical CSS regen is necessary;
* Add issue to Site Health panel in case there was a problem generating Critical CSS.

### Other information:

## Jetpack product discussion

https://github.com/Automattic/boost-cloud/issues/217

no

### Note about Site Health panel issues
* The Site Health menu item does not change the issue count until you visit the page and wait for all checks to finish;
* The same goes for when an error gets fixed - you need to visit the Site Health panel and wait for the checks to finish, so it can properly update its count.
* To observe this - make sure you don't have any issues in your Site Health panel.
* Close the page, and open another WP admin page (the main WP Dashboard for example).
* Now enable WP_DEBUG on your website.
* Refresh the page you were on.
* Note, that the Site Health panel doesn't show that as an issue.
* Now visit the Site Health panel and wait for all checks to finish.
* You should see an issue about debug being visible to visitors.
* Visit another WP admin page and hover over `Tools -> Site Health` to see, that it shows 1 issue.
* Disable WP_DEBUG, but don't visit the Site Health panel.
* Open any other WP Admin page, and note, that it still shows 1 issue.
* Now open the Site Health panel and wait for the checks to finish.
* It should no longer show issues.

### Test regen notice description change, notification bubble, Regenerate link appearance change and Site Health issue
* Setup this version of Boost (free mode);
* Navigate to the Boost dashboard and wait for it to generate Critical CSS;
* After that's done, add a new page to your website;
* Check the Boost dashboard again you should see a notice to regenerate Critical CSS;
* The `Regenerate` link should also change appearance while the notice is visible;

<details>
<summary>Screenshot</summary>
CleanShot 2023-06-02 at 13 00 19@2x
</details>

* There should also be a notification bubble next to the Boost menu item in the sidebar;

<details>
<summary>Screenshot</summary>
CleanShot 2023-06-02 at 13 01 55@2x
</details>

* Add a post;
* Check the Boost dashboard again and the notice text should change accordingly;

<details>
<summary>Screenshot</summary>
CleanShot 2023-06-02 at 13 18 21@2x
</details>

* You also see the notice description change after activating/deactivating a plugin;

<details>
<summary>Screenshot</summary>
CleanShot 2023-06-02 at 13 19 23@2x
</details>

* Doing any of the changes, should also render an issue in the Site Health panel.

<details>
<summary>Screenshot</summary>
CleanShot 2023-06-02 at 13 16 34@2x
</details>

### Test if Site Health panel shows Critical CSS generation failure issue
* Continuing from the above tests...
* While the website is in free mode, setup Boost Developer plugin;
* Make sure you don't have Boost Cloud running locally;
* Change `Critical CSS Mode` to `Cloud CSS`;
* Navigate to the Boost dashboard and wait for Critical CSS to fail;

<details>
<summary>Screenshot</summary>
CleanShot 2023-06-02 at 13 08 47@2x
</details>

* Check the Site Health panel to see if there's an issue about the failure.

<details>
<summary>Screenshot</summary>
CleanShot 2023-06-02 at 13 15 26@2x
</details>

### DIFF
--- a/projects/plugins/boost/app/lib/Boost_Health.php
+++ b/projects/plugins/boost/app/lib/Boost_Health.php
@@ -11,11 +11,11 @@ class Boost_Health {
 
 	public function __construct() {
 		if ( self::critical_css_needs_regeneration() ) {
-			$this->issues[] = 'Outdated Critical CSS';
+			$this->issues[] = __( 'Outdated Critical CSS', 'jetpack-boost' );
 		}
 
 		if ( self::critical_css_has_errors() ) {
-			$this->issues[] = 'Failed to generate Critical CSS';
+			$this->issues[] = __( 'Failed to generate Critical CSS', 'jetpack-boost' );
 		}
 	}
 
--- a/projects/plugins/boost/app/lib/Boost_Health.php
+++ b/projects/plugins/boost/app/lib/Boost_Health.php
@@ -2,6 +2,7 @@
 
 namespace Automattic\Jetpack_Boost\Lib;
 
+use Automattic\Jetpack_Boost\Lib\Critical_CSS\Critical_CSS_State;
 use Automattic\Jetpack_Boost\Modules\Optimizations\Cloud_CSS\Cloud_CSS;
 
 class Boost_Health {
@@ -11,6 +12,10 @@ class Boost_Health {
 	public function __construct() {
 		if ( self::critical_css_needs_regeneration() ) {
 			$this->issues[] = 'Outdated Critical CSS';
+		}
+
+		if ( self::critical_css_has_errors() ) {
+			$this->issues[] = 'Failed to generate Critical CSS';
 		}
 	}
 
@@ -30,5 +35,9 @@ class Boost_Health {
 		$suggest_regenerate = jetpack_boost_ds_get( 'critical_css_suggest_regenerate' );
 
 		return in_array( $suggest_regenerate, Environment_Change_Detector::get_available_env_change_statuses(), true );
+	}
+
+	public static function critical_css_has_errors() {
+		return ( new Critical_CSS_State() )->has_errors();
 	}
 }
--- a/projects/plugins/boost/app/lib/Boost_Health.php
+++ b/projects/plugins/boost/app/lib/Boost_Health.php
@@ -5,10 +5,23 @@ namespace Automattic\Jetpack_Boost\Lib;
 use Automattic\Jetpack_Boost\Lib\Critical_CSS\Critical_CSS_State;
 use Automattic\Jetpack_Boost\Modules\Optimizations\Cloud_CSS\Cloud_CSS;
 
+/**
+ * Class Boost_Health
+ *
+ * Represents the health of the Jetpack Boost plugin.
+ */
 class Boost_Health {
 
+	/**
+	 * @var array List of issues affecting the health of the plugin.
+	 */
 	private $issues = array();
 
+	/**
+	 * Boost_Health constructor.
+	 *
+	 * Initializes the Boost_Health object and checks for any health issues.
+	 */
 	public function __construct() {
 		if ( self::critical_css_needs_regeneration() ) {
 			$this->issues[] = __( 'Outdated Critical CSS', 'jetpack-boost' );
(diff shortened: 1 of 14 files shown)


projects/plugins/boost/wp-js-data-sync.php
```diff
@@ -128,7 +128,7 @@ function jetpack_boost_initialize_datasync() {
  * Register Data Sync Stores
  */
 jetpack_boost_register_option( 'critical_css_state', $critical_css_state_schema );
-jetpack_boost_register_option( 'critical_css_suggest_regenerate', Schema::as_boolean()->fallback( false ) );
+jetpack_boost_register_option( 'critical_css_suggest_regenerate', Schema::as_string()->fallback( '' ) );
```
thingalon: What do you think of `Schema::as_enum( array( ... allowed values ) )->nullable()`? That way `null` is considered as no notice needed, then there is a set of allowed values. I believe `nullable` also auto-falls back to `null` if an invalid value is supplied.

Then you can make the typescript type `z.coerce.enum( [ ... allowed types ... ] ).nullable()` too. :)

thingalon (dismissed):
This works! I've added one suggestion, but it's not a requirement. :)

projects/plugins/boost/wp-js-data-sync.php
dilirity: Unfortunately, I couldn't find a way to make zod coerce enums.

> z.enum is a Zod-native way to declare a schema with a fixed set of allowable string values.

I couldn't get around null getting cast as string each time.

I was, however, able to find a way to validate against a specific list of items:
https://github.com/Automattic/jetpack/pull/31101/commits/f22b18e341fce41dde8fb6bd72036f3685d2ee0d#diff-f843d9a2e27c8f8afc64e85cf20ecff360f4fcda5045f7a35865d43e392b1b96R23-R35

The reason for `1` being in the list as a string, is because older versions of Boost stored a boolean and `1`, was a way of telling the UI a regen was needed.

dilirity:
@thingalon converting to draft, as there are more things left to do on https://github.com/Automattic/boost-cloud/issues/217 and I didn't want to create separate PRs for everything.

dilirity:
Ready for review.

projects/plugins/boost/wp-js-data-sync.php
thingalon: But if you make the `enum` `nullable()` then it won't coerce null to a string? See: d5cf219602

Can you double-check it seems ok to you?

projects/plugins/boost/app/lib/Boost_Health.php
```diff
@@ -0,0 +1,43 @@
...
+
+	private $issues = array();
+
+	public function __construct() {
+		if ( self::critical_css_needs_regeneration() ) {
+			$this->issues[] = 'Outdated Critical CSS';
+		}
+
+		if ( self::critical_css_has_errors() ) {
+			$this->issues[] = 'Failed to generate Critical CSS';
```
thingalon: untranslated strings.

projects/plugins/boost/app/lib/Boost_Health.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+namespace Automattic\Jetpack_Boost\Lib;
+
+use Automattic\Jetpack_Boost\Lib\Critical_CSS\Critical_CSS_State;
+use Automattic\Jetpack_Boost\Modules\Optimizations\Cloud_CSS\Cloud_CSS;
+
+class Boost_Health {
```
thingalon, line 13: Pls ad a class comment describing its purpose, as it wouldn't be immediately clear without prior context.

thingalon (changes requested):
This is lovely: 
image

Looks pretty good, except for two nitpicks.

projects/plugins/boost/app/lib/Boost_Health.php
dilirity: 😅 Fixed in [0dec583](https://github.com/Automattic/jetpack/pull/31101/commits/0dec5839e3a1c469cb1020436b2df42b6fea1989).

projects/plugins/boost/app/lib/Boost_Health.php
dilirity: Good catch - thanks. I've added comment in [eea340e](https://github.com/Automattic/jetpack/pull/31101/commits/eea340eb2cc0d529de9a45e59c41664cb1009aed).

projects/plugins/boost/wp-js-data-sync.php
dilirity: Thank you for this! I was unable to do it, probably because I was trying to coerce an enum and didn't think to include `nullable`. Your solution is much simpler and worked as expected during my tests.

thingalon (approved):

PR #31111, opened 2023-05-31
Add tracks event for boost score error

CodeyGuyDylan:
## Proposed changes:

Add track event that is recorded when the boost speed score API returns an error

### Other information:

## Jetpack product discussion

P2: p8oabR-1e4-p2#comment-7341

Yes, new event named `jetpack_boost_speed_score_error` that triggers when a speed score request returns an error

## Testing instructions:

**This will have to be done locally as 2 lines of code need to be edited**

1. Checkout these changes locally
2. Go to the file `projects/plugins/jetpack/_inc/client/at-a-glance/boost/index.jsx` and update line 134 to the following (this will force the speed scores to update and not use the cached results)
```jsx
if ( ! latestSpeedScores && calculateDaysSince( latestSpeedScores.timestamp * 1000 ) < 21 ) {
```
3. Go to the file `projects/packages/boost-speed-score/src/class-speed-score.php` and update line 94 to the following to simulate an error
```php
if ( isset( $params['url'] ) ) {
```
4. Without Boost installed, go to `/wp-admin/admin.php?page=jetpack#/dashboard` and wait for the Boost module to load. After loading it should look like this
image
5. In Tracks Vigilante, you should see this event
image

### DIFF
--- a/projects/plugins/jetpack/_inc/client/at-a-glance/boost/index.jsx
+++ b/projects/plugins/jetpack/_inc/client/at-a-glance/boost/index.jsx
@@ -97,7 +97,7 @@ const DashBoost = ( {
 			setIsLoading( false );
 		} catch ( err ) {
 			analytics.tracks.recordEvent( 'jetpack_boost_speed_score_error', {
-				feature: 'boost',
+				feature: BOOST_PLUGIN_SLUG,
 				position: 'at-a-glance',
 				error: err,
 			} );
--- a/projects/plugins/jetpack/_inc/client/at-a-glance/boost/index.jsx
+++ b/projects/plugins/jetpack/_inc/client/at-a-glance/boost/index.jsx
@@ -96,6 +96,12 @@ const DashBoost = ( {
 			setDaysSinceTested( 0 );
 			setIsLoading( false );
 		} catch ( err ) {
+			analytics.tracks.recordEvent( 'jetpack_boost_speed_score_error', {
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/_inc/client/at-a-glance/boost/index.jsx
```diff
@@ -96,6 +96,12 @@ const DashBoost = ( {
 			setDaysSinceTested( 0 );
 			setIsLoading( false );
 		} catch ( err ) {
+			analytics.tracks.recordEvent( 'jetpack_boost_speed_score_error', {
+				feature: 'boost',
```
jboland88: Would there be any reason to use the same constant `BOOST_PLUGIN_SLUG` for the feature here?

jboland88 (dismissed):
This tested well for me, I was able to get the error and the tracks event.

One small comment about the feature label for the event, but non-blocking.

projects/plugins/jetpack/_inc/client/at-a-glance/boost/index.jsx
CodeyGuyDylan: You're right, I think that's better. I'll go ahead and update that so you can re-approve

jboland88 (approved):
Re-approving after updating the feature attribute of the event 👍

anomiex:
Cherry-picked in 19376f8c0d

PR #31114, opened 2023-05-31
tools: Fix deploy-to-svn.sh stable tag handling

anomiex:
## Proposed changes:

When we update the stable tag in w.org SVN, we usually forget to bring that back to the monorepo. This can then too easily result in the stable tag getting accidentally downgraded on a future deploy.

We already handle this correctly in [the wp-svn-autopublish action](https://github.com/Automattic/jetpack/blob/614c751e23dcb7326e68564e21192088024294bb/.github/files/gh-wp-svn-autopublish/files/wp-svn-autopublish.sh#L72-L93), let's use the same logic here to ensure that the tag in `trunk/readme.txt` is never changed by the script.

### Other information:

## Jetpack product discussion

Relevant to p1685470667664939-slack-C0299DMPG

No

## Testing instructions:

* You might comment out lines 136–138, then try running the script for some existing tag.

### DIFF
--- a/tools/deploy-to-svn.sh
+++ b/tools/deploy-to-svn.sh
@@ -173,13 +173,26 @@ fi
 
 cd "$DIR"
 
-STABLE_TAG="$(sed -n -E -e 's/^Stable tag: +([^ ]+) *$/\1/p' trunk/readme.txt)"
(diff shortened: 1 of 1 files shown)


samiff (approved):
Tested well for me 👍 Thanks for fixing this.

PR #31121, opened 2023-05-31
Launchpad: Refactor launchpad task completion for future tasks

sixhours:
Related to Automattic/wp-calypso#77278

## Proposed changes:

* Based on @daledupreez suggestions in paYKcK-30u-p2, this PR attempts to put these ideas into play for real. :)
* Decouple task definitions from task registrations.
* Decouple task list definitions from task list registrations.
* Put task definitions and their callbacks in their own file.
* Add a `wpcom_get_launchpad_task_list` function that fetches a single registered task list in lieu of fetching all task lists at once.
* Add a `has_task_lists` method on the `Launchpad_Task_Lists` singleton that checks to see if we have already registered any task lists.
* Refactor `wpcom_mark_launchpad_task_complete`, allowing us to mark a task as complete by directly interfacing with the task definition and the site options table rather than waiting for the task to be registered.
* Fire a Tracks event when marking a task as complete with the new function.
* Initialize deprecated `add_listener_callback`s for existing tasks.

### Other information:

## Testing instructions:

* Apply this patch to your sandbox and sandbox `public-api.wordpress.com`
* Visit the dev console and select WP REST API, `wpcom/v2`
* Make a GET request to `/sites/[siteSlug]/launchpad?checklist_slug=[slug]`, replacing [slug] with the following task list slugs:
- newsletter
- link-in-bio
- link-in-bio-tld
- build
- write
- start-writing
- design-first
- free
- keep-building
- videopress
* Make sure the correct tasks are still returned for each list based on the definitions in `launchpad.php`
* Make sure you can still mark tasks as complete when viewing the Launchpad at `/setup/[flow]/launchpad?siteSlug=[siteSlug]`

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -249,7 +249,7 @@ function wpcom_mark_launchpad_task_complete( $task_id ) {
 /**
  * Initialize the Launchpad task listener callbacks.
  *
- * @param array $task_definitions
+ * @param array $task_definitions The tasks to initialize.
  */
 function wpcom_launchpad_init_listeners( $task_definitions ) {
 	require_once WP_CONTENT_DIR . '/lib/log2logstash/log2logstash.php';
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -38,7 +38,7 @@ function wpcom_launchpad_get_task_definitions() {
 			'badge_text_callback'  => 'wpcom_get_domain_upsell_badge_text',
 		),
 		'first_post_published'            => array(
-			'get_title'            => function () {
+			'get_title'             => function () {
 				return __( 'Write your first post', 'jetpack-mu-wpcom' );
 			},
 			'add_listener_callback' => function () {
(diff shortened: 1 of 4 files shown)


projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -0,0 +1,518 @@
...
+ */
+function wpcom_launchpad_init_listeners( $task_definitions ) {
+	require_once WP_CONTENT_DIR . '/lib/log2logstash/log2logstash.php';
+
+	foreach ( $task_definitions as $task_id => $task_definition ) {
+		if ( isset( $task_definition['add_listener_callback'] ) && is_callable( $task_definition['add_listener_callback'] ) ) {
+			$task_data = array_merge( $task_definition, array( 'id' => $task_id ) );
+
+			try {
+				call_user_func( $task_definition['add_listener_callback'], $task_data ); // Current callbacks expect the built, registered task for the second parameter, which won't work in this case.
```
andres-blanco, line 262: Did you think about having the `add_listener_callback` return `true` if successful and a wp_error if not? That way we could avoid the generic `try/catch` and get a more detailed error if something went wrong.

FYI: I finished reviewing the code, I am going to run it now :)

andres-blanco:
I am seeing the wrong results for the `newsletter` task list
| Expected  | Actual result |
| image | image  |

andres-blanco:
The launchpad is still working as expected. Tasks get marked as complete and launching the site works 👍

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
sixhours: I hadn't considered that, and I like the idea! But I think doing so would involve refactoring all the existing callbacks to return the expected bool/error, and I see potential for disrupting those production flows in the process and drawing out our timeline. What do you think about adding it to the cleanup list as a nice-to-have once we get the Launchpad in Customer Home?

sixhours:
> I am seeing the wrong results for the newsletter task list

I see the same list on production; some of those tasks are tied to `wpcom_has_goal_paid_subscribers` which my test site does not have, so I think that's expected. 

Screen Shot 2023-06-05 at 3 34 04 PM

Screen Shot 2023-06-05 at 3 37 46 PM

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
andres-blanco: yes, I agree with that.

nelsonec87 (approved):
LGTM! I was able to see and mark tasks as done as per testing instructions!

andres-blanco (approved):

PR #30759, opened 2023-05-16
[RNMobile] Disable VideoPress settings for not belonged videos

fluiddot:
## Proposed changes:

* Following the changes introduced in https://github.com/Automattic/jetpack/pull/30438 and https://github.com/Automattic/jetpack/pull/30443, disable Details, Privacy, and Rating settings if the video doesn't belong to the site.
* Render a warning message when the video doesn't belong to the site. I used the same design of the warning we display on the [`ContrastChecker` component](https://github.com/WordPress/gutenberg/blob/f0266c22cdbab064482c5e57d6a63a29b232cf2d/packages/block-editor/src/components/contrast-checker/index.native.js#L24-L59).

<details><summary><strong>Click here to show Android screenshots</strong></summary>

| Light mode | Dark mode |
|--------|--------|
|   |  | 
|  |  | 

</details>

<details><summary><strong>Click here to show iOS screenshots</strong></summary>

| Light mode | Dark mode |
|--------|--------|
|   |  | 
|   |  |
</details>

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

N/A

## Does this pull request change what data or activity we track or use?

N/A

## Testing instructions:

* Open a post/page.
* Add a VideoPress block.
* Add a video.
* Open block settings and set title and description.
* Observe that it works as expected.
* Tap on the "Privacy and Rating" option.
* Change rating, privacy, "Allow download", and "Show video sharing menu" settings.
* Observe that it works as expected.
* Add another VideoPress block
* Add a video that doesn't belong to the site. E.g. insert the video from an URL - https://videopress.com/v/NGYE1Hqy.
* Open the block settings.
* Observe that the title and description settings are disabled and that a warning message is displayed.
* Tap on the "Privacy and Rating" option.
* Observe that settings are disabled and a warning message is displayed.

PR #31134, opened 2023-06-01
[RNMobile] Show "No title"/"No description" placeholder for not belonged videos

fluiddot:
Follow up on https://github.com/Automattic/jetpack/pull/30759.

## Proposed changes:

* For not belonged videos that have an empty title and/or description, instead of displaying the placeholder `Add title`/`Add description`, show `No title`/`No description` to avoid confusion.

| Empty title and description | With title and description |
|--------|--------|
|  |  | 

### Other information:

## Jetpack product discussion

N/A

N/A

## Testing instructions:

- Open a post/page.
- Add a VideoPress block.
- Add a video that doesn't belong to the site and has an empty title and description (e.g. insert the video from an URL - https://videopress.com/v/XBqrRFxJ).
- Open the block settings.
- Observe that the title and description settings are disabled.
- Observe that the title field displays a placeholder with the text `No title`.
- Observe that the description field displays a placeholder with the text `No description`.
- Replace the video with another one that doesn't belong to the site, but has a title and description. (e.g. insert the video from an URL - https://videopress.com/v/E6So1Wpf).
- Observe that the title and description settings are disabled.
- Observe that the title field displays the title of the video.
- Observe that the description field displays the description of the video.

### DIFF
--- a/projects/packages/videopress/changelog/rnmobile-videopress-no-description-placeholder
+++ b/projects/packages/videopress/changelog/rnmobile-videopress-no-description-placeholder
@@ -1,0 +1,4 @@
+Significance: patch
+Type: added
+
+RNMobile: Display no title and no description placeholder for not belonged videos
--- a/projects/packages/videopress/package.json
+++ b/projects/packages/videopress/package.json
@@ -1,7 +1,7 @@
 {
 	"private": true,
 	"name": "@automattic/jetpack-videopress",
-	"version": "0.14.5",
+	"version": "0.14.6-alpha",
 	"description": "VideoPress package",
 	"homepage": "https://github.com/Automattic/jetpack/tree/HEAD/projects/packages/videopress/#readme",
 	"bugs": {
--- a/projects/packages/videopress/src/class-package-version.php
+++ b/projects/packages/videopress/src/class-package-version.php
@@ -11,7 +11,7 @@ namespace Automattic\Jetpack\VideoPress;
  * The Package_Version class.
  */
 class Package_Version {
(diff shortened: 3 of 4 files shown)


SiobhyB (dismissed):
@fluiddot, LGTM! Thank you for spotting and following up on this 🙇‍♀️ 

What do you think about also updating the title placeholder? It'd be a edge case, but an `add title` placeholder does display when there's no title, which would be confusing if it can't actually be edited.

https://github.com/Automattic/jetpack/blob/87262ba1bbe42444260f05f152615687d3efdbb5/projects/packages/videopress/src/client/block-editor/blocks/video/components/details-panel/index.native.js#L31

fluiddot:
> What do you think about also updating the title placeholder? It'd be a edge case, but an `add title` placeholder does display when there's no title, which would be confusing if it can't actually be edited.

Yeah, I could apply the same approach to that field too. I'll push an update with this suggestion, thanks!

fluiddot:
> > What do you think about also updating the title placeholder? It'd be a edge case, but an `add title` placeholder does display when there's no title, which would be confusing if it can't actually be edited.
> 
> Yeah, I could apply the same approach to that field too. I'll push an update with this suggestion, thanks!

@SiobhyB I applied this suggestion in https://github.com/Automattic/jetpack/pull/31134/commits/9759f0ddf149078555bc91536c18ca55c020fc94. Let me know if you could take a look at the PR, thanks 🙇 !

jhnstn (approved):
Looks good :shipit:

PR #31145, opened 2023-06-01
Jetpack Boost: Add a proxy to the Image Guide so it can load remote images

### LINKED ISSUE #29896: Boost: Show estimated saving for cross origin images
Currently, the image guide cannot determine the size and potential saving of images that aren't hosted on the same site. A workaround is possible by proxying the image through the site's backend like we do for CSS assets while generating critical CSS.

image

donnchawp:
The Image Guide provides information about the images used on a web page but it cannot load remote images and show file size information about them.

This PR adds a proxy server and makes AJAX calls to it to fetch the remote image through the home server of the page.

Props to @thingalon for all his help with this.

Fixes #29896

## Proposed changes:
* It modifies MeasurableImage, adding a third parameter, a proxy callback, to the constructor.
* That callback is used to fetch remote images, returning only the size, which is handled in fetchFileWeight.
* Similarly, getMeasurableImages has a third parameter in the same way, passing it down to MeasurableImage.
* We needed to pass the ajax URL to prepareAdminAjaxRequest, as it would be used in the frontend.
* The proxy function is fetchWeightUsingProxy, which is used in the initialize() function when creating a list of MeasurableImage objects.
* Remove the notice that images on a different domain can't be loaded.

### Other information:

No

## Testing instructions:
Apply PR
Embed an image from another domain in a test post.
View the post with Image Guide enabled and it will show the filesize of that image.

## Screenshots
before
after

### DIFF
--- a/projects/js-packages/image-guide/src/MeasurableImage.ts
+++ b/projects/js-packages/image-guide/src/MeasurableImage.ts
@@ -11,7 +11,7 @@ export class MeasurableImage {
 	readonly node: HTMLElement | HTMLImageElement;
 	private getURLCallback: SourceCallbackFn;
 
-	public fetch = fetch;
+	public fetch = window.fetch;
 
 	/**
 	 * Constructor.
--- a/projects/js-packages/image-guide/src/MeasurableImage.ts
+++ b/projects/js-packages/image-guide/src/MeasurableImage.ts
@@ -99,18 +99,8 @@ export class MeasurableImage {
 			console.log( `Can't get image size for ${ url } likely due to a CORS error.` );
 			return -1;
 		}
-		let size = '';
-		if ( response.headers.get( 'content-type' ).indexOf( 'text/html' ) !== -1 ) {
-			size = await response.text();
-		} else {
-			size = response.headers.get( 'content-length' );
-		}
 
-		if ( size ) {
-			return parseInt( size, 10 ) / 1024;
-		}
-
-		return -1;
+		return parseInt( response.headers.get( 'content-length' ), 10 ) / 1024;
 	}
 
 	/**
--- a/projects/js-packages/image-guide/src/MeasurableImage.ts
+++ b/projects/js-packages/image-guide/src/MeasurableImage.ts
@@ -18,7 +18,7 @@ export class MeasurableImage {
 	 *
 	 * @param {HTMLElement | HTMLImageElement} node -  The DOM Element that contains the image.
 	 * @param {SourceCallbackFn} getURL             -  A function that takes in the node and returns the URL of the image.
-	 * @param {(input: string, init?: Array) => Promise<Response>} fetchFn                               -  A function that fetches a URL and returns a Promise.
+	 * @param {(input: string, init?: Array) => Promise<Response>} fetchFn -  A function that fetches a URL and returns a Promise.
 	 */
 	constructor( node: HTMLElement | HTMLImageElement, getURL: SourceCallbackFn, fetchFn = fetch ) {
 		this.node = node;
--- a/projects/js-packages/image-guide/src/MeasurableImage.ts
+++ b/projects/js-packages/image-guide/src/MeasurableImage.ts
@@ -2,6 +2,8 @@ export type SourceCallbackFn = ( node: HTMLElement ) => string | null;
 export type Dimensions = { width: number; height: number };
 export type Weight = { weight: number };
 
+type FetchFn = ( input: URL | RequestInfo, init?: RequestInit ) => Promise< Response >;
+
 /**
  * A class that represents a DOM Element that
  * has an image that should be measured and
@@ -16,11 +18,15 @@ export class MeasurableImage {
 	/**
 	 * Constructor.
 	 *
-	 * @param {HTMLElement | HTMLImageElement} node                        -  The DOM Element that contains the image.
-	 * @param {SourceCallbackFn} getURL                                    -  A function that takes in the node and returns the URL of the image.
-	 * @param {(input: string, init?: Array) => Promise<Response>} fetchFn -  A function that fetches a URL and returns a Promise.
+	 * @param {HTMLElement | HTMLImageElement} node -  The DOM Element that contains the image.
+	 * @param {SourceCallbackFn} getURL             -  A function that takes in the node and returns the URL of the image.
+	 * @param {FetchFn} fetchFn                     -  A function that fetches a URL and returns a Promise.
 	 */
-	constructor( node: HTMLElement | HTMLImageElement, getURL: SourceCallbackFn, fetchFn = fetch ) {
+	constructor(
+		node: HTMLElement | HTMLImageElement,
+		getURL: SourceCallbackFn,
+		fetchFn: FetchFn = fetch
+	) {
 		this.node = node;
 		this.getURLCallback = getURL;
 		this.fetch = fetchFn;
--- a/projects/js-packages/image-guide/src/MeasurableImage.ts
+++ b/projects/js-packages/image-guide/src/MeasurableImage.ts
@@ -11,15 +11,19 @@ export class MeasurableImage {
 	readonly node: HTMLElement | HTMLImageElement;
 	private getURLCallback: SourceCallbackFn;
 
+	public fetch = fetch;
+
 	/**
 	 * Constructor.
 	 *
 	 * @param {HTMLElement | HTMLImageElement} node -  The DOM Element that contains the image.
 	 * @param {SourceCallbackFn} getURL             -  A function that takes in the node and returns the URL of the image.
+	 * @param {(input: string, init?: Array) => Promise<Response>} fetchFn                               -  A function that fetches a URL and returns a Promise.
 	 */
-	constructor( node: HTMLElement | HTMLImageElement, getURL: SourceCallbackFn ) {
+	constructor( node: HTMLElement | HTMLImageElement, getURL: SourceCallbackFn, fetchFn = fetch ) {
 		this.node = node;
 		this.getURLCallback = getURL;
+		this.fetch = fetchFn;
 	}
 
 	public getURL() {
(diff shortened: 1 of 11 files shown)


projects/js-packages/image-guide/src/MeasurableImage.ts
```diff
@@ -83,14 +87,19 @@ export class MeasurableImage {
...
 	private async fetchFileWeight( url: string ) {
-		const response = await fetch( url, { method: 'HEAD', mode: 'no-cors' } );
+		const response = await this.fetch( url );
 		if ( ! response.url ) {
 			// eslint-disable-next-line no-console
 			console.log( `Can't get image size for ${ url } likely due to a CORS error.` );
 			return -1;
 		}
+		let size = '';
+		if ( response.headers.get( 'content-type' ).indexOf( 'text/html' ) !== -1 ) {
```
thingalon: On my test site, I found that a test webp image I used was not returning a content-type header at all, causing this line to crash with the following error:

```
Cannot read properties of null (reading 'indexOf')
```

projects/js-packages/image-guide/src/MeasurableImage.ts
```diff
@@ -83,14 +87,19 @@ export class MeasurableImage {
...
+			size = await response.text();
+		} else {
+			size = response.headers.get( 'content-length' );
+		}
```
thingalon: This class accepts an alternate `fetch` function, but this function treats fetch as potentially special -- not fetching the image, but returning just its size. That feels a bit off to me, like it's being treated differently by different parts of code.

When I suggested a fetch function, I had imagined that it would behave just like fetch with super-powers to bypass CORS restrictions. i.e.: fetch HEAD to get the headers, etc.

Limiting the proxy to just return file-size makes a load of sense - it's both more secure, and lighter weight than returning the full body. So I'm on the fence about what to suggest here. 

IMHO we should either:
1. Instead of accepting a fetch function, we could have the MeasurableImage class accept a getImageSize function that it calls instead of using its own logic if supplied. That way, we could put the special-case logic in the getImageSize call, and *expect* just a size in the body, or
2. Make the fetch function behave like fetch, and just fetch the HEAD.

What do you think?

projects/js-packages/image-guide/src/MeasurableImage.ts
```diff
@@ -11,15 +11,19 @@ export class MeasurableImage {
 	readonly node: HTMLElement | HTMLImageElement;
 	private getURLCallback: SourceCallbackFn;
 
+	public fetch = fetch;
```
thingalon: On review, this does feel a bit odd. Do you think it's worth saying `fetch = window.fetch` just to add extra context for future devs reading the code?

projects/js-packages/image-guide/src/MeasurableImage.ts
```diff
@@ -11,15 +11,19 @@ export class MeasurableImage {
...
+
 	/**
 	 * Constructor.
 	 *
 	 * @param {HTMLElement | HTMLImageElement} node -  The DOM Element that contains the image.
 	 * @param {SourceCallbackFn} getURL             -  A function that takes in the node and returns the URL of the image.
+	 * @param {(input: string, init?: Array) => Promise<Response>} fetchFn                               -  A function that fetches a URL and returns a Promise.
```
thingalon: Why does the linter like you more than me? If I step out of line, the thing screams at me. :) 

The internal spacing to line up the descriptions on this comment is kinda weird. Minor nitpick, I know.

projects/js-packages/image-guide/src/MeasurableImage.ts
```diff
@@ -11,15 +11,19 @@ export class MeasurableImage {
...
 	 */
-	constructor( node: HTMLElement | HTMLImageElement, getURL: SourceCallbackFn ) {
+	constructor( node: HTMLElement | HTMLImageElement, getURL: SourceCallbackFn, fetchFn = fetch ) {
```
thingalon: One thing I'm fussy about: I like to specify types for all arguments. As a pro-tip, you can define a type at the top of the file like:

```
type FetchFn = ( input: URL | RequestInfo, init?: RequestInit ) => Promise<Response>;
```

Then whenever you specify the type -- both in the param list here, and in the @param comment above, you can just say `FetchFn`.

projects/plugins/boost/app/modules/image-guide/Image_Guide_Proxy.php
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Add an ajax endpoint to proxy external CSS files.
+ */
+class Image_Guide_Proxy {
+	const NONCE_ACTION = 'jb-ig-proxy-nonce';
+
+	public static function init() {
+		$instance = new self();
+		add_action( 'wp_ajax_boost_proxy_ig', array( $instance, 'handle_ig_proxy' ) );
```
thingalon: Why instantiate at all? Seems like we can just make `handle_ig_proxy` static and avoid instantiating :)

projects/plugins/boost/app/modules/image-guide/Image_Guide_Proxy.php
```diff
@@ -0,0 +1,47 @@
...
+		// Validate URL and fetch.
+		$proxy_url = filter_var( wp_unslash( isset( $_POST['proxy_url'] ) ? $_POST['proxy_url'] : null ), FILTER_VALIDATE_URL );
+		if ( ! wp_http_validate_url( $proxy_url ) ) {
+			die( 'Invalid URL' );
+		}
+
+		$response = wp_remote_get( $proxy_url );
+		if ( is_wp_error( $response ) ) {
+			// TODO: Nicer error handling.
+			die( 'error' );
```
thingalon: For an endpoint that only returns image or a text, I think maybe just `die`? `wp_die` I think returns HTML, right?

thingalon (dismissed):
Nice work. I have some suggestions, though - see inline

projects/plugins/boost/app/modules/image-guide/Image_Guide_Proxy.php
dilirity: Tagging along, there's a TODO left in there.

projects/js-packages/image-guide/src/MeasurableImage.ts
donnchawp: Thanks, good idea.

projects/js-packages/image-guide/src/MeasurableImage.ts
donnchawp: The linter did change that line, but it somehow wasn't caught by Git, but it also only changed line 21, and didn't complain about lines 19-20, which I will fix and line up.

projects/js-packages/image-guide/src/MeasurableImage.ts
donnchawp: Thanks, I wasn't sure what type to put in there.

projects/plugins/boost/app/modules/image-guide/Image_Guide_Proxy.php
donnchawp: Good point about wp_die(). That code was copied from CSS_proxy.php as it returns error code 400. We might need to fix that there.

I removed the "TODO" comment, that was copied from the CSS proxy too.

projects/plugins/boost/app/modules/image-guide/Image_Guide_Proxy.php
donnchawp: Good idea, done!

projects/js-packages/image-guide/src/MeasurableImage.ts
donnchawp: I realised after working on this for a while that I was returning the content length in the ajax request body, and so it wouldn't be a drop-in replacement for fetch. It's impossible to set the content-length header of the ajax request to be the length of the proxied image. You have to pass the content length in the body.

While I modified the function so it returns the entire header array, I think I should just go with a drop-in getImageSize function instead.

projects/plugins/boost/app/modules/image-guide/Image_Guide_Proxy.php
thingalon: Nice. Thank you :) 

This is a totally optional thing, but did you know you can avoid having to spell out the full namespace in the hook name by using `__NAMESPACE__` like this example:

https://github.com/Automattic/jetpack/blob/3c05799e6811f48b1352a1cd38a669272e4a92cf/projects/plugins/boost/jetpack-boost.php#L149

It's a trick I like to use.

projects/js-packages/image-guide/src/MeasurableImage.ts
donnchawp: I added more checks so this won't happen.

projects/js-packages/image-guide/src/MeasurableImage.ts
donnchawp: I moved the processing of the proxy into fetchWeightUsingProxy, the function used to replace fetch. It creates new headers and then returns the content length as part of the response.

projects/plugins/boost/app/modules/image-guide/Image_Guide_Proxy.php
```diff
@@ -9,14 +9,13 @@ class Image_Guide_Proxy {
 	const NONCE_ACTION = 'jb-ig-proxy-nonce';
 
 	public static function init() {
-		$instance = new self();
-		add_action( 'wp_ajax_boost_proxy_ig', array( $instance, 'handle_ig_proxy' ) );
+		add_action( 'wp_ajax_boost_proxy_ig', 'Automattic\Jetpack_Boost\Modules\Image_Guide\Image_Guide_Proxy::handle_proxy' );
```
dilirity: Small note - I think this can be simplified further by doing
```php
add_action( 'wp_ajax_boost_proxy_ig' , array( __CLASS__, 'handle_proxy' ) );
```

That way, you won't need to update the namespace inline when we eventually change it.

projects/plugins/boost/app/modules/image-guide/Image_Guide_Proxy.php
```diff
@@ -18,28 +18,27 @@ public static function init() {
...
 		$response = wp_remote_get( $proxy_url );
 		if ( is_wp_error( $response ) ) {
-			die( 'error' );
+			wp_send_json_error( 'error', 400 );
 		}
 
-		header( 'Content-type: text/html' );
-		echo (int) $response['headers']['content-length'];
-
+		header( 'Content-type: application/json' );
```
dilirity: You don't need to exclusively send the headers. `wp_send_json` does that for [you and also includes the charset](https://github.com/WordPress/wordpress-develop/blob/6.2/src/wp-includes/functions.php#L4412-L4417).

dilirity (changes requested):
A few small improvements I could find. Hope you don't mind ^^

projects/plugins/boost/app/modules/image-guide/Image_Guide_Proxy.php
donnchawp: Thanks, I did not know that!

projects/plugins/boost/app/modules/image-guide/Image_Guide_Proxy.php
donnchawp: Thank you, that's useful!

dilirity (approved):

PR #31150, opened 2023-06-01
Update monorepo to Composer 2.5 and PHP 8.2

anomiex:
## Proposed changes:

Composer 2.5 has been out for a while now, and doesn't even have an API version update in the lockfile.

PHP 8.0 security support ends in November, and we've been running CI tests with 8.1 and 8.2 for a while. Let's update that too.

### Other information:

## Jetpack product discussion

p9dueE-7mw-p2#comment-9725

No

## Testing instructions:

* Does everything still work? In particular, look for any of PHP 8.2's many added deprecation warnings.


anomiex:
We'll need to un-require the "PHP tests: PHP 8.0 WP trunk" and require the new "PHP tests: PHP 8.2 WP trunk" check before merging this. Otherwise, seems good for review to me.

anomiex:
P.S. Internal announcement draft is at pdWQjU-jV-p2

.github/files/generate-ci-matrix.php
```diff
@@ -65,10 +65,16 @@
...
+	$phpver = $versions['PHP_VERSION'];
+	if ( $wp === 'previous' ) {
+		$phpver = '8.0';
+	}
 	$matrix[] = array(
-		'name'    => "PHP tests: PHP {$versions['PHP_VERSION']} WP $wp",
+		'name'    => "PHP tests: PHP {$phpver} WP $wp",
 		'script'  => 'test-php',
-		'php'     => $versions['PHP_VERSION'],
+		'php'     => $phpver,
```
zinigor, line 77: Is this the reason why the PR is still waiting for the PHP 8.0 results on `trunk`?

anomiex: Kind of.

We run our tests against the previous and trunk versions of WordPress, to make sure our code doesn't break against them. We don't much care which version of PHP we use for those, so we just use whatever is the default version from versions.sh. So when we _change_ the version in versions.sh, we need to have GitHub stop requiring "PHP tests: PHP $OLDVER WP previous" and "PHP tests: PHP $OLDVER WP trunk" that will no longer be run, and then have it start requiring "PHP tests: PHP $NEWVER WP previous" and "PHP tests: PHP $NEWVER WP trunk" instead after the changeover PR is merged.

Except in this case we hack the code to still use 8.0 instead of 8.2 for "WP previous" because WP 6.1 throws a bunch of deprecation warnings with later PHP versions. So we'll only have to update GitHub's expectation for the "WP trunk" test.

But if you want to pick a specific line to blame, you should actually blame the `'name'` two lines up. If we were to omit the `{$phpver}` from that line then the check name wouldn't change when changing PHP versions.

zinigor (approved):

PR #31152, opened 2023-06-01
AI Assistant: replace tone button icon

retrofox:
Fixes #

## Proposed changes:

* AI Assistant: replace tone button of the block toolbar with audio icon

### Other information:

## Testing instructions:

before | after
-------|-------
image | Screenshot 2023-06-05 at 17 09 51


ederrengifo:
I have concerns about this change because people might associate this with audio and not "tone." What is the context of this idea? Is it to look for a better association with "tone of voice."?

Sadly, our icon set is quite limited, but here are a few other options that I think might reduce that risk (although I still think the megaphone is the best available option):

icons for tone of voice

oskosk:
Even if sometimes it looks that it makes sense to grab an icon from the library, it is confusing for those who already know that icon from others pieces of UI in WordPress, where the icon actually means what it was meant to represent (audio here). We should ask Design for custom icons for our toolbars.

retrofox:
> We should ask Design for custom icons for our toolbars.

[The idea](p1HpG7-mHr-p2#comment-63549) came from designers.

ederrengifo:
We explored this new icon if we can use a custom one for the block:

new-tone icon


Feedback is welcome! But if we all think this option works, here the [Figma asset](https://www.figma.com/file/gG1smMHbNGas0qwP9y6RwK/Jetpack-AI-Assistant?type=design&node-id=4269%3A20656&t=DjTcOzhvSBHrigkG-1)

dhasilva (approved):
LGTM!

PR #31154, opened 2023-06-01
CRM: Resolves 3114 and 3115 - listview settings bugs

tbradsha:
Resolves Automattic/zero-bs-crm#3114 and Automattic/zero-bs-crm#3115 - listview settings bugs

## Proposed changes:

This PR addresses two issues:

1. PHP errors when saving page settings (e.g. listview settings or contact edit metabox settings)
2. listview settings resets

For the first item, there was a `for` loop that was trying to cycle through key values on `null` in some instances. Adding a non-null condition fixed this.

The second issue was far more complex. Essentially anytime a module (e.g. WooSync or MailPoet) loads its settings, it does a brute force resave of all loaded settings, including "screen options" (per-page view settings). While it's a bit overkill, it generally works fine.

However, screen options were implemented to save in a per-user fashion, and the brute force save has no concept of setting ownership, so when resaving, the setting owner is changed to `-1` (no owner), and then the page can't find its screen options anymore.

Given that:

* Per-user settings were never fully implemented through the system,
* It hasn't ever worked,
* A change is unlikely to impact many installs, and
* A core fix wouldn't suffice, as extensions also implement the same brute force update...

I just went ahead and made screen option settings global until a future date when we might consider per-user settings.

### Other information:

## Testing instructions:

1. Go to a listview: `/wp-admin/admin.php?page=manage-customers`
2. Change per-page settings.
3. Go to the WooSync Hub: `/wp-admin/admin.php?page=woo-sync-hub`
4. Go back to the listview and look at its settings.

In `trunk`, one gets PHP errors after step 2, and per-page settings are reset to default (20).

In the `fix/crm/3114-screenopt_save_issues` branch, there are no errors after step 2 and the per-page settings are not affected by modules/extensions.

### DIFF
--- a/projects/plugins/crm/includes/wh.config.lib.php
+++ b/projects/plugins/crm/includes/wh.config.lib.php
@@ -65,33 +65,11 @@
 		#} Checks through defaults + existing and adds defaults where unset
 		function validateAndUpdate(){
 
-			global $zbs;
-
-			// DAL1 save:
-			if (!$zbs->isDAL2()){
-				
-				$defaultsAdded = 0;
-				foreach ($this->settingsDefault as $key => $val){ 
-					if (!isset($this->settings[$key])) {
-						$this->settings[$key] = $val;
-						$defaultsAdded++;
-					}
-				}
-				
-				if ($defaultsAdded > 0) $this->saveToDB();
-
-			} else {
-
-				// DAL 2
-				
 				foreach ($this->settingsDefault as $key => $val){ 
 					if (!isset($this->settings[$key])) {
 						$this->update($key,$val);
 					}
 				}
-				
-
-			}
 
 		}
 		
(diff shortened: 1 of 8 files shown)


projects/plugins/crm/includes/wh.config.lib.php
```diff
@@ -65,33 +65,11 @@ function __construct($config=array()) {
...
-						$defaultsAdded++;
-					}
-				}
-				
-				if ($defaultsAdded > 0) $this->saveToDB();
-
-			} else {
-
-				// DAL 2
-
```
cleacos, line 86: 👍

projects/plugins/crm/includes/wh.config.lib.php
```diff
@@ -159,22 +137,8 @@ function get($key,$freshFromDB=false){
...
-			if (!$freshFromDB && $zbs->isDAL2()){
-
-				// normal way
-				if (isset($this->settings[$key]))
-					return $this->settings[$key];
-				else
-					return false;
-
-			} else {
-
```
cleacos, line 173: 👍 removing DAL2 code

projects/plugins/crm/includes/wh.config.lib.php
```diff
@@ -184,22 +148,9 @@ function update($key,$val=''){
...
-			if (!$zbs->isDAL2()){
-
-				#} Don't even check existence as I guess it doesn't matter?
-				$this->settings[$key] = $val;	
-				
-				#} Save down
-				$this->saveToDB();
-
-			} else {
-
```
cleacos, line 197: 👍 removing DAL1 code

cleacos (approved):
It works as expected.

LGTM!

PR #31155, opened 2023-06-01
CRM: Resolves 3138 - hide resources and support pages on white label installs

tbradsha:
Resolves Automattic/zero-bs-crm#3138 - hide resources and support pages on white label installs

## Proposed changes:

On white label, we were sometimes showing pages we shouldn't. For example, we see Resources:

image

And we see Support and Review:
image

Likewise, on a regular install we don't see support if using Full/CRM Only menu mode:
image

This PR removes Resources and the middle column of the profile menu from white label installs, and adds support to regular installs when using Full/CRM Only menu mode.

**Note:** beware of Automattic/zero-bs-crm#3137 if using an older white label install to test the "before" state.

### Other information:

## Testing instructions:

1. Add a built copy of this PR to your local `app`'s `/rebrandr/_build/core-versions/` and `/rebrandr/_build/templates/`. You can download that from here: https://betadownload.jetpack.me/data/zero-bs-crm/fix_crm_3138-hide_resources_page_on_white_label/zero-bs-crm-dev.zip
2. Run `app` and go to `/rebrandr-app` to set up a template and build the white-label CRM.
3. Download the CRM bundle zip.
4. Upload the downloaded CRM zip (which is nested in a parent zip) to a test site.
5. Ensure the menus are as described.

### DIFF
--- a/projects/plugins/crm/changelog/fix-crm-3138-hide_resources_page_on_white_label
+++ b/projects/plugins/crm/changelog/fix-crm-3138-hide_resources_page_on_white_label
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+White label: JPCRM support and resources pages no longer show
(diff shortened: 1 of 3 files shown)


gogdzl (approved):
It works as advertised.
Nice job!

PR #31168, opened 2023-06-02
Jetpack Social | Fix state when adding a new connection

manzoorwanijk:
Fixes 1203820938137329-as-1204692223892765/f

## Proposed changes:

* Disable a connection by default if no shares left.
* Imrove `useDismissNotice` hook to create a local re-usable state of dismissed notices
* Ensure that Instagram notice is displayed only if there is no IG account connected.
* Update Social admin page notice to get rid of local `useState` in favour of `dismissedNotices` from `useDismissNotice`

### Other information:

### To test "Disable a connection by default if no shares left."

* Point `public-api.wordpress.com` to your sandbox
* In your sandbox, update `PUBLICIZE_MONTHLY_LIMIT` to `0` in `publicize/publicize-wpcom.php`
* On a site pointing to your sandbox with no Jetpack Social plan, create a new post
* Open Jetpack plugin sidebar
* Confirm that under "Share this post" section, you see the notice `You have 0 shares remaining in the next 30 days`
* Click on "Connect an account" link to add a new connection in a new tab
* Add another social media account
* Goto the post edit page again and wait for the social media accounts list to refresh
* Confirm that the newly added account is _disabled_ by default
* Now set `PUBLICIZE_MONTHLY_LIMIT` back to `30`
* Reload the post edit page
* Add another social media account as above
* Goto the post edit page again and wait for the social media accounts list to refresh
* Confirm that the newly added account is _enabled_ by default

### To test "Ensure that Instagram notice is displayed only if there is no IG account connected."

* Add Jetpack Social Advance plan to your site
* Ensure that you have D111103-code applied to your sandbox 
* Add Instagram blog sticker to the site as mentioned in D111103-code
* Run `jetpack docker wp option delete jetpack_social_dismissed_notices` to delete the dismiss notice option to ensure that Instagram notice appears
* Create a new post
* Open Jetpack plugin sidebar
* Confirm that under "Share this post" section, you see the notice `You can now share directly to your Instagram account!`
* Add an Instagram connection to the site.
* Goto the post edit page again and wait for the social media accounts list to refresh
* Confirm that the above notice is hidden
* Disconnect the Instagram account
* Goto the post edit page again and wait for the social media accounts list to refresh
* Confirm that the notice appears again
* Click on the dismiss icon for the notice
* Confirm that the notice is hidden
* Reload the page to verify that the notice is no longer visible
* Run `jetpack docker wp option get jetpack_social_dismissed_notices` and verify that `'instagram'` is added to the dismissed notices option
* Run `jetpack docker wp option delete jetpack_social_dismissed_notices` again 
* Goto Jetpack Social admin page - `/wp-admin/admin.php?page=jetpack-social`
* Confirm that you see the "Instagram is now available in Jetpack Social" notice
* Click on its dismiss icon
* Confirm that the notice is hidden
* Reload the page
* Confirm that the notice is hidden

| Before adding a new connection | Before adding a new connection |
|--------|--------|
| Screenshot 2023-06-02 at 6 56 06 PM | Screenshot 2023-06-02 at 6 57 05 PM |

### DIFF
--- a/projects/js-packages/publicize-components/src/components/form/index.js
+++ b/projects/js-packages/publicize-components/src/components/form/index.js
@@ -10,7 +10,7 @@ import { getRedirectUrl } from '@automattic/jetpack-components';
 import { getSiteFragment } from '@automattic/jetpack-shared-extension-utils';
 import { Button, PanelRow, Disabled, ExternalLink } from '@wordpress/components';
 import { useDispatch, useSelect } from '@wordpress/data';
-import { useState, Fragment, createInterpolateElement, useCallback } from '@wordpress/element';
+import { Fragment, createInterpolateElement, useCallback } from '@wordpress/element';
 import { _n, sprintf, __ } from '@wordpress/i18n';
 import useAttachedMedia from '../../hooks/use-attached-media';
 import useDismissNotice from '../../hooks/use-dismiss-notice';
@@ -67,13 +67,15 @@ export default function PublicizeForm( {
 	const hasInstagramConnection = connections.some(
 		connection => connection.service_name === 'instagram-business'
 	);
-	const [ shouldShowInstagramNotice, setShouldShowInstagramNotice ] = useState(
-		! hasInstagramConnection && isInstagramConnectionSupported
-	);
+
+	const shouldShowInstagramNotice =
+		! hasInstagramConnection &&
+		isInstagramConnectionSupported &&
+		! dismissedNotices.includes( 'instagram' );
+
 	const onDismissInstagramNotice = useCallback( () => {
 		dismissNotice( 'instagram' );
-		setShouldShowInstagramNotice( false );
-	}, [ dismissNotice, setShouldShowInstagramNotice ] );
+	}, [ dismissNotice ] );
 	const shouldDisableMediaPicker =
 		isSocialImageGeneratorAvailable && isSocialImageGeneratorEnabledForPost;
 	const Wrapper = isPublicizeDisabledBySitePlan ? Disabled : Fragment;
@@ -255,7 +257,7 @@ export default function PublicizeForm( {
 			) }
 			{ ! isPublicizeDisabledBySitePlan && (
 				<Fragment>
-					{ ! dismissedNotices.includes( 'instagram' ) && shouldShowInstagramNotice && (
+					{ shouldShowInstagramNotice && (
 						<Notice
 							onDismiss={ onDismissInstagramNotice }
 							type={ 'highlight' }
--- a/projects/js-packages/publicize-components/src/store/effects.js
+++ b/projects/js-packages/publicize-components/src/store/effects.js
@@ -13,9 +13,9 @@ import { SUPPORTED_CONTAINER_BLOCKS } from '../components/twitter';
  */
 export async function refreshConnectionTestResults() {
 	try {
+		const jetpackSocialData = getJetpackData()?.social || {};
 		const connectionRefreshPath =
-			getJetpackData()?.social?.connectionRefreshPath ??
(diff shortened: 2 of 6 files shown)


projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -67,13 +67,15 @@ export default function PublicizeForm( {
 	const hasInstagramConnection = connections.some(
 		connection => connection.service_name === 'instagram-business'
 	);
-	const [ shouldShowInstagramNotice, setShouldShowInstagramNotice ] = useState(
-		! hasInstagramConnection && isInstagramConnectionSupported
-	);
```
manzoorwanijk, line 72: It's always good to avoid `useState` for the initial values that can change. The state can get quickly out of sync.

projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -67,13 +67,15 @@ export default function PublicizeForm( {
...
+	const shouldShowInstagramNotice =
+		! hasInstagramConnection &&
+		isInstagramConnectionSupported &&
+		! dismissedNotices.includes( 'instagram' );
+
 	const onDismissInstagramNotice = useCallback( () => {
 		dismissNotice( 'instagram' );
-		setShouldShowInstagramNotice( false );
-	}, [ dismissNotice, setShouldShowInstagramNotice ] );
+	}, [ dismissNotice ] );
```
pablinos, line 78: This is much better. I was wondering if `useDismissNotice` could take a notice name, so that the `dismissNotice` callback returned wouldn't need to be wrapped in a `useCallback` like this.

Perhaps it's future improvement.

projects/js-packages/publicize-components/src/store/effects.js
```diff
@@ -24,7 +24,7 @@ export async function refreshConnectionTestResults() {
 		const connections = [];
 		const defaults = {
 			done: false,
-			enabled: true,
+			enabled: Boolean( jetpackSocialData.sharesData?.shares_remaining ),
```
pablinos, line 27: Ah a simple solution! I was wondering if we might be able to get this default coming back from the server too, but it seems we always use the default when we haven't seen the connection before.

pablinos (approved):
This looks good. There's two other bits which I think we could do as a different change. The notice to add an image is still shown when you are out of shares:

image

I think it would be good to remove that.

The Connect Now button on the Instagram notice, opens in the same tab, but it should probably open in a new one like we do for the other manage connections links.

projects/js-packages/publicize-components/src/components/form/index.js
manzoorwanijk: > I was wondering if `useDismissNotice` could take a notice name

That API is not flexible given the rules of React hooks. Sometimes you may want to dismiss multiple notices by passing the value at call time (to the callback)

> so that the `dismissNotice` callback returned wouldn't need to be wrapped in a `useCallback` like this.

That doesn't affect whether the callback should be wrapped in `useCallback`. It will still need to be used.

projects/js-packages/publicize-components/src/store/effects.js
manzoorwanijk: Yeah, I agree that having a flag passed from the server is better because we can disable it for any other reasons in the future.

manzoorwanijk:
> The notice to add an image is still shown when you are out of shares

We have a separate task for that - 1203820938137329-as-1204692223892751/f

> The Connect Now button on the Instagram notice, opens in the same tab, but it should probably open in a new one like we do for the other manage connections links.

I did notice that but I thought to do it in a separate PR to keep the PRs granular. Thanks

PR #31169, opened 2023-06-02
Jetpack Sync: Enforce args size limits per action

fgiannar:
Adds args size checks before attempting to add a certain action to the Sync queue.

If the arguments' size turns out to be larger than the max defined, the corresponding action will not be added to the Sync queue.

## Proposed changes:

* `Sync\Defaults`: Introduce a new static property `$default_max_args_size_per_action` that holds the max serialized data size per Sync action.
* `Sync\Listener`: Introduce a new protected method, `check_data_size_by_action` for checking the args size before attempting to add a certain action to the Sync queue.
* `WP_Test_Jetpack_Sync_Listener`: Add 3 new unit tests

### Other information:

## Jetpack product discussion

[p9F6qB-cii-p2](p9dueE-7rF-p2)

No

## Testing instructions:

- Set `Defaults::$default_max_args_size_per_action['updated_option']['default']` = 1; 
- Update the `blogname`
- Confirm the action will not be synced
- Revert the `Defaults::$default_max_args_size_per_action['updated_option']['default']` change
- Update the `blogname` once more
- Confirm the action is synced

Make sure to checkout the added unit tests as well.


projects/packages/sync/src/class-listener.php
```diff
@@ -480,4 +484,56 @@ public function get_request_url() {
...
+				}
+			}
+		}
+
+		$max_args_size = isset( $max_args_size ) ? $max_args_size : $max_args_size_per_action['default'];
+		try {
+			$serialized_args = serialize( $args ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.serialize_serialize
+		} catch ( Exception $e ) {
+			// Possible case would be a callable returning an anonymous function.
+			return true;
```
fgiannar: @bisko Should we maybe return `false` here? Do we allow that?

fgiannar: Just an FYI, after discussing this we decided to return `false` instead since we don't allow callables returning an anonymous function

zinigor:
We have discussed this one a bit in person recently with @fgiannar , and I'm OK with how it's going to work, but it needs an update. Please feel free to put the needs review label back when the PR is up to date with trunk.

PR #31170, opened 2023-06-02
Subscribers Page: Add `Subscribers` link to the menu

ivan-ottinger:
This PR adds the Subscribers link to the menu in these places:

- WPCOM & Atomic: Users menu
- Self-hosted Jetpack: Jetpack menu

It only adds them when the user locale is set to English.

### Other information:

- [X] Have you written new tests for your changes, if applicable?

## Jetpack product discussion
p1686580459461379-slack-C02TCEHP3HA

No

## Testing instructions:
Since the way menus work in Calypso ([more info](PCYsg-vFk-p2), this patch has to be deployed on your sandbox, and `public-api` needs to be sandboxed. 

I found it very difficult to properly test this for Atomic & self-hosted, as the API endpoint seemed to think that every site I gave it was a Simple Site. 

I resorted to unit testing, as the issue is probably with my setup. 

If you get this working please tell me how 👍 

To run the tests:

``` jetpack docker phpunit -- --filter=Admin_Menu ```


jeherve:
Should we make this available to self-hosted sites as well, via a new item under the Jetpack menu maybe?

ivan-ottinger:
> Should we make this available to self-hosted sites as well, via a new item under the Jetpack menu maybe?

Thank you for the comment, Jeremy!

I actually created a TODO item in the related GH task last week: https://github.com/Automattic/wp-calypso/issues/77531, just didn't get to it yet. 🙂 

By the "Jetpack" menu, do you mean this one?

Markup on 2023-06-05 at 15:15:56

Here's the project thread for context: pdDOJh-1TE-p2.

jeherve:
> By the "Jetpack" menu, do you mean this one?

Yes, this one on WoA sites, and on self-hosted sites.

TimBroddin:
Added to Jetpack. I still have to test this.

ivan-ottinger:
Hey @jeherve 👋🏼 Just to double-check: If we understand correctly, there's currently a daily deploy to WPCOM Simple (p58i-eqs-p2). Is that correct?

Does that mean that once this PR is merged, it will be deployed to WPCOM Simple the next day?

Also, does this follow UTC time (e.g. everything merged prior the end of the day in UTC will get deployed the next day)? Thank you.

jeherve:
> If we understand correctly, there's currently a daily deploy to WPCOM Simple (p58i-eqs-p2). Is that correct?

That's correct.

> Does that mean that once this PR is merged, it will be deployed to WPCOM Simple the next day?

It may even be deployed on the same day, if you merge it before deployment happens.

> Also, does this follow UTC time (e.g. everything merged prior the end of the day in UTC will get deployed the next day)?

There is no code freeze for those daily deployments, so anything you merge before the release lead starts the daily deployment (usually after 14:00 UTC).

phcp (dismissed):
Code looks good and the tests are passing as expected. I also tried but couldn't find other ways to test it.

jeherve (dismissed):
This tests well for me, it should be good to  merge. I must say I was a bit curious about the language restriction. Is there some background behind this? If we wanted to limit access, I would have expected it to limit access based on the user's language, not the site language.

TimBroddin:
> This tests well for me, it should be good to merge. I must say I was a bit curious about the language restriction. Is there some background behind this? If we wanted to limit access, I would have expected it to limit access based on the user's language, not the site language.

You're 100% right, this should be user's language. I'll update the PR.

TimBroddin:
I've updated the PR! We'll be pushing this only to English users if we don't have all translations when we merge this. Otherwise I'll remove this check.

yansern (approved):
Re-ran test, ~now no more failing tests~ (update: still failing). Re-approving.

Greenlight to merge: p1686869690550299/1686869270.217619-slack-CBG1CP4EN

PR #31157, opened 2023-06-01
CRM: Resolves 3143 - additional Emerald polish

tbradsha:
Resolves Automattic/zero-bs-crm#3143 - additional Emerald polish

## Proposed changes:

* Gave the top left menu a `max-height` to prevent issues on white label
* Fixed highlights on secondary menu
* Fixed lots of padding/margin stuff on various screen sizes (e.g. pbhBOz-3op-p2 and more)
* More choppage, including a few unused files

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

Check and see if there are any inconsistent left/right margins.

PR #31174, opened 2023-06-02
CRM: Resolves 3091 - clean up user roles

tbradsha:
Resolves Automattic/zero-bs-crm#3091 - clean up user roles

## Proposed changes:

In the linked issue, a user cited our docs stating that the Mail Manager should not be able to edit quotes or invoices:

https://kb.jetpackcrm.com/knowledge-base/setting-up-your-team/

Looking at the roles, I confirmed that confirmed that we were lax in most of our permissions; most roles could view and edit most objects. I cleaned up the code and tried to standardise the role definitions as follows:

* CRM Admin: full access to everything
* WP Admin: full access to everything
* Contact Manager: view and edit all objects, but not settings
* Quote Manager: view and edit contacts and quotes
* Invoice Manager: view and edit contacts, invoices, and transactions (as transactions are directly linked to invoices)
* Transaction Manager: view and edit contacts and transactions
* CRM Customer: custom role with no special capabilities (used for Client Portal)
* Mail Manager: view all objects (in listview), and edit contacts and tasks

Note that all team member roles can edit contacts, as that permission is deeply engrained in the system checks that are not directly related to contacts.

Prior to this PR, anyone could add a new object (even if they could edit it). This adds a basic check to prevent that (it's not pretty at this stage, but no one should run into it accidentally).

I also restricted display of the Dashboard and Task Scheduler according to their proper permissions.

Finally, I co-opted an old migration so that the roles would be properly refreshed (deleted and added) in WordPress.

I have not updated the KB to reflect the above, but we'll need to do so if this gets merged.

### Other information:

## Testing instructions:

Essentially make a user and switch its role (I used a private browser) to make sure things are properly restricted per my outline above.

### DIFF
--- a/projects/plugins/crm/includes/ZeroBSCRM.Core.Page.Controller.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.Core.Page.Controller.php
@@ -94,7 +94,9 @@ function zeroBSCRM_pages_admin_addedit_page( $type = 'contact', $action = 'new',
 			if ( zeroBSCRM_permsObjType( $obj_type_id ) ) {
 				zeroBSCRM_pages_admin_addedit_page_generic( $id, $action );
 			} else {
+				echo '<div style="margin-left: 20px">';
 				echo esc_html( sprintf( __( 'You do not have permission to edit this %s.', 'zero-bs-crm' ), $zbs->DAL->typeStr( $obj_type_id ) ) ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase,WordPress.WP.I18n.MissingTranslatorsComment
+				echo '</div>';
 			}
 			break;
 
--- a/projects/plugins/crm/includes/ZeroBSCRM.Core.Page.Controller.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.Core.Page.Controller.php
@@ -66,51 +66,36 @@ function zeroBSCRM_pages_admin_addedit() {
 // } This is a slow general move to new UI (and new DB) and moving away from the custom post add / edit links
 function zeroBSCRM_pages_admin_addedit_page( $type = 'contact', $action = 'new', $id = -1 ) {
 
-	// } learn script (for this page)
+	global $zbs;
+	$obj_type_id = $zbs->DAL->objTypeID( $type ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
 
-	?>
+	switch ( $obj_type_id ) {
 
-	<script type="text/javascript">
-
-		jQuery(function($){
-
-			jQuery('.learn')
-			.popup({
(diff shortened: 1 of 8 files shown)


projects/plugins/crm/includes/ZeroBSCRM.Core.Page.Controller.php
```diff
@@ -66,51 +66,36 @@ function zeroBSCRM_pages_admin_addedit() {
...
-
+	global $zbs;
+	$obj_type_id = $zbs->DAL->objTypeID( $type ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
 
-	<?php
-	switch ( $type ) {
+	switch ( $obj_type_id ) {
 
-		case 'contact':
+		case ZBS_TYPE_CONTACT:
```
coder-karen, line 74: Nice move using the predefined object names.

projects/plugins/crm/includes/ZeroBSCRM.Core.Page.Controller.php
```diff
@@ -66,51 +66,36 @@ function zeroBSCRM_pages_admin_addedit() {
...
+		case ZBS_TYPE_QUOTE:
+		case ZBS_TYPE_INVOICE:
+		case ZBS_TYPE_TRANSACTION:
+		case ZBS_TYPE_EVENT:
+		case ZBS_TYPE_FORM:
+		case ZBS_TYPE_QUOTETEMPLATE:
+			if ( zeroBSCRM_permsObjType( $obj_type_id ) ) {
+				zeroBSCRM_pages_admin_addedit_page_generic( $id, $action );
+			} else {
+				echo esc_html( sprintf( __( 'You do not have permission to edit this %s.', 'zero-bs-crm' ), $zbs->DAL->typeStr( $obj_type_id ) ) ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase,WordPress.WP.I18n.MissingTranslatorsComment
```
coder-karen, line 98: Looks like there are some spacing issues here on smaller screens:
Screenshot 2023-06-04 at 16 02 02
Screenshot 2023-06-04 at 16 02 16

coder-karen (commented):
Works well!! Left a couple of comments related to the 'you do not have permissions error'.

projects/plugins/crm/includes/ZeroBSCRM.Core.Page.Controller.php
tbradsha: The more general issue with margin/padding is already addressed in #31157, but I've pushed the error message slightly to the right in 35ffd86734 so it lines up with the page title.

coder-karen: Looks good!

coder-karen (approved):

PR #31177, opened 2023-06-02
CRM: Resolves 3130 - centralise browser autocomplete disable

tbradsha:
Resolves Automattic/zero-bs-crm#3130 - centralise browser autocomplete disable

## Proposed changes:

Copied from the issue:
>Browsers try to be smart and autocomplete/autosuggest content in input fields. We often don't want that, and the way around it is to give the `autocomplete` attribute a random value (since `autocomplete="off"` is basically unsupported by Chrome). Throughout the code we have variations of this code. I've created `jpcrm_disable_browser_autocomplete()` to handle this, and we should replace all such instances in our code (search `autocomplete=` for examples).

This PR updates all existing inputs that disable autocomplete to use `jpcrm_disable_browser_autocomplete()`. You can search the code with this:

`autocomplete=`

You can search for references to `autocomplete` that don't use that function with this (requires regex engine with negative lookahead):

`autocomplete=(?!.*?jpcrm_disable_browser_autocomplete)`

The above will return one reference in JS, to which I added a rudimentary `Math.random()` string (there's also some JS autokey code in the same file around line 1815 that does some stuff, but I wouldn't rely on it).

Edit: I also noticed that the transactions page had some fields ("date paid" and "date completed") that didn't match the new Emerald edit field style. In order to avoid a merge conflict, I fixed that in this PR (9dc8769).

### Other information:

## Testing instructions:

Ideally nothing changed; this is all about making things consistent for the developer.

### DIFF
--- a/projects/plugins/crm/js/ZeroBSCRM.admin.invoicebuilder.js
+++ b/projects/plugins/crm/js/ZeroBSCRM.admin.invoicebuilder.js
@@ -435,7 +435,7 @@ function zbscrm_JS_draw_invoice_top_right_form( res ) {
 		html +=
 			'<input type="text" name="zbsi_ref" id="ref" class="form-control widetext" placeholder="" value="' +
 			jpcrm.esc_attr( res.invoiceObj.id_override ) +
-			'" autocomplete="zbsinv" />';
+			'" autocomplete="' + Math.random() + '" />';
 	}
 	html += '</td>';
 	html += '</tr>';
--- a/projects/plugins/crm/admin/email/main.page.php
+++ b/projects/plugins/crm/admin/email/main.page.php
@@ -450,7 +450,7 @@ function zeroBSCRM_pages_admin_sendmail() {
 
 		}
 		?>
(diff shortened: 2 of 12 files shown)


projects/plugins/crm/js/ZeroBSCRM.admin.invoicebuilder.js
```diff
@@ -435,7 +435,7 @@ function zbscrm_JS_draw_invoice_top_right_form( res ) {
 		html +=
 			'<input type="text" name="zbsi_ref" id="ref" class="form-control widetext" placeholder="" value="' +
 			jpcrm.esc_attr( res.invoiceObj.id_override ) +
-			'" autocomplete="zbsinv" />';
+			'" autocomplete="' + Math.random() + '" />';
```
cleacos, line 438: Wondering why here we don't use the function

cleacos: > The above will return one reference in JS, to which I added a rudimentary Math.random() string (there's also some JS autokey code in the same file around line 1815 that does some stuff, but I wouldn't rely on it).

^^OK

cleacos (approved):
The code looks good.

LGTM!

PR #31180, opened 2023-06-02
Actions: add new action to run k6 tests against staging sites

samiff:
## Proposed changes:

Add some k6 tests to generate traffic on predefined staging/test sites.

### Other information:

## Jetpack product discussion

Internal PT: p9dueE-6Re-p2

No.

## Testing instructions:

The backend tests will require the proper secrets, but you can see an example action run here: https://github.com/samiff/test/actions/runs/5225021803/jobs/9433951842

## Todo:

- [x] Add new secrets to repo.
- [x] Add Slack alerting for failures (also consider this for the action that updates the staging sites).


jeherve:
TIL about k6. That seems super useful!

anomiex:
It's not clear to me what exactly is the intention here.

The summary says "to generate traffic", but looking at the example run the traffic generated seems minimal (and I'm wary that a non-minimal level of traffic might bring a repeat of p1682539226654069-slack-C02AVAR9B).

The actual code seems to be more of a run of some quick E2E tests. Which is not a bad thing, but we'd probably need to add something (e.g. an entry [here](https://github.com/Automattic/jetpack/blob/add/k6-for-jp-staging-sites/.github/workflows/slack-workflow-failed.yml#L7-L12)) to notify us if it fails. And I wonder whether it'd be better to integrate these tests with whatever tests KitKat runs or wants to run if testing is the purpose.

anomiex:
I'm going to bump this out of "Needs Review" since in the team meeting yesterday we talked about wanting to add Slack notification of some sort on failures.

samiff:
> we talked about wanting to add Slack notification of some sort on failures.

Thanks @anomiex I've used the existing `slack-workflow-failed.yml` to add alerts for `Update Jetpack Staging Test Sites` and `k6 Tests for Jetpack Staging Test Sites`. That seems like it should work for what we need here.

As mentioned in our call, we just want a small amount of baseline traffic on these test sites so we can leverage that in a Grafana dashboard. Just hoping to catch any obvious PHP fatals or large deviances in other graph thresholds we can setup.

anomiex (approved):
Sure, let's try it.

PR #31183, opened 2023-06-05
IDC: fix connection screen for broken connection

sergeymitr:
Solving an edge case for missing connection screen when the site is technically in IDC, but actually disconnected.

When Jetpack site falls into IDC (`jetpack_sync_error_idc` option is set), we do not display Jetpack Dashboard and show the IDC screen instead.

However, if Jetpack connection is broken to a point where we consider the site no longer connected (e.g. blog token is missing), but the `jetpack_sync_error_idc` option hasn't been cleared, we do the following:
- do not show the connection screen because there should be the IDC screen (`jetpack_sync_error_idc` option is present)
- do not show the IDC screen because we don't consider the site connected (blog token is missing)

The option `jetpack_sync_error_idc` is irrelevant if we force the user to reconnect, as it'll be cleared during connection process anyway.

**Note:** this PR does not fix the WoA problem that can be reproduced the same way, as it's complicated by WoA-specific setup.
See details in the linked PR.

## Proposed changes:
* Do not hide Jetpack Dashboard (including connection screen) if site is not connected, no matter if there's IDC or not.

### Other information:

No.

## Testing instructions:
1. Connect Jetpack
2. Use "IDC Simulator" to force IDC (fill `jetpack_sync_error_idc` option with values)
3. Use "Broken Token" to remove the blog token
4. Go to "Jetpack -> Jetpack" and confirm that connection screen shows up.
5. Go through the connection flow, confirm it

### DIFF
--- a/projects/plugins/jetpack/_inc/lib/admin-pages/class.jetpack-admin-page.php
+++ b/projects/plugins/jetpack/_inc/lib/admin-pages/class.jetpack-admin-page.php
@@ -74,7 +74,7 @@ abstract class Jetpack_Admin_Page {
 		$this->jetpack = $jetpack;
 
 		self::$block_page_rendering_for_idc = (
-			Identity_Crisis::validate_sync_error_idc_option() && ! Jetpack_Options::get_option( 'safe_mode_confirmed' )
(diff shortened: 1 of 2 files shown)


bindlegirl (approved):
LGTM

anomiex:
Cherry-picked into 12.2 in 955d129584

PR #31186, opened 2023-06-05
Launchpad: Support task list filtering of visible task

### LINKED ISSUE #77297: Launchpad: Support task list filtering of visible task
We can toggle a task's visibility only by the task level with the `is_visible_callback` callback. But having this functionality on the task list level would be helpful. The idea is to add the possibility of passing a filter to the task list definition to allow us to control the visibility of certain tasks inside the task list.

**Note**: We need to ensure that the task list completion verification will consider this filtering to mark the task list as completed.

valterlorran:
Currently, we can filter tasks by the task definition, but it would be hard to filter tasks in different task lists. We add support for filtering tasks on the task list level to allow us to manage the tasks for experiments or different contexts.

Closes https://github.com/Automattic/wp-calypso/issues/77297

## Proposed changes:

* Add a new callback to allow filtering tasks on the task list level.

### Other information:

## Testing instructions:

* The callback is not being used yet. I added some unit tests to cover it:
```
cd projects/packages/jetpack-mu-wpcom; ./vendor/bin/phpunit --filter=La
unchpad_Task_Lists_Test; cd -
```
* Make sure that this change doesn't break the current behavior.
  * Sandbox the public-api
  * Apply these changes to your sandbox: https://github.com/Automattic/jetpack/pull/31186#issuecomment-1576416273
  * Create a new free site, and go through the Launchpad.

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/changelog/add-launchpad-is-visible-callback
+++ b/projects/packages/jetpack-mu-wpcom/changelog/add-launchpad-is-visible-callback
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Support task list filtering of visible task
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
@@ -197,9 +197,12 @@ class Launchpad_Task_Lists {
 			return $tasks_for_task_list;
 		}
 
+		// Filter the task list's task ids to only include visible tasks if a callback is provided.
+		$task_ids = $this->load_value_from_callback( $task_list, 'visible_tasks_callback', $task_list['task_ids'] );
+
 		// Takes a registered task list, looks at its associated task ids,
 		// and returns a collection of associated tasks.
-		foreach ( $task_list['task_ids'] as $task_id ) {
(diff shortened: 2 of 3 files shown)


sixhours:
I think this is intended to address issues like the one I've run into in #31191, where you have a single task assigned to multiple task lists, and the conditions for showing that task depend on the list context, right?

Using that as an example, we could do something like:

```
$task_statuses = get_option( 'launchpad_checklist_tasks_statuses', array() );
return array_filter(
	$task_ids,
	function ( $task_id ) {
              // Only show design_edited if it hasn't been marked as completed.
              if ( 'design_edited' === $task_id ) {
                    return ! isset( $task_statuses[$task_id] ) || ! $task_statuses[$task_id] ? $task_id : null;
              }
              
              // All other tasks.
              return $task_id;
	}
);
```

sixhours (dismissed):
This makes sense to me and I think it will help us include tasks on multiple lists under different conditions. I wasn't able to run the tests locally; I think there's something weird with my configuration.

valterlorran:
I think the checks will pass once https://github.com/Automattic/jetpack/pull/31055 is merged.

andres-blanco (approved):
Tests are passing and the launchpad is still working. 

Nitpick: the 3rd parameter to the `_doing_it_wrong` call should be the current WP/Jetpack version. but since this code should only be run internally I think it's not a blocker.

PR #31190, opened 2023-06-05
e2e tests: always dismiss the sharing modal added by ETK

jeherve:
## Proposed changes:

We do not need to test this modal in Jetpack, and we do not want it to interfer with the rest of the tests

### Other information:

## Jetpack product discussion

p1685976419147559-slack-C03QNBQKG73

No

## Testing instructions:

Check CI

### DIFF
--- a/tools/e2e-commons/bin/e2e-env.sh
+++ b/tools/e2e-commons/bin/e2e-env.sh
@@ -72,6 +72,9 @@ configure_wp_env() {
 	$BASE_CMD wp option set permalink_structure ""
 	$BASE_CMD wp jetpack module deactivate sso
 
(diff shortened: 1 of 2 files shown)


adimoldovan:
@jeherve You could also set this in the env setup script for a less verbose approach: https://github.com/Automattic/jetpack/blob/dc763089ecb92769f695d14e23e40f4bcc7ef765/tools/e2e-commons/bin/e2e-env.sh#L56

jeherve:
> You could also set this in the env setup script for a less verbose approach:

I like that, it seems simpler. I just pushed 5898a218c12deba4ef1042940dd341e629d62a65, how does that look?

adimoldovan (approved):
Looks good!

PR #29736, opened 2023-03-27
Plans: Fix fatal on WPCOM due to lack of Jetpack::get_available_modules

pablinos:
The WPCOM implementation of `Jetpack` doesn't have the `get_available_modules` method. This change swaps to using the Modules package in the Plans package, which should be fine.

Fixes #

## Proposed changes:

*

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* Go to '..'
*

PR #29678, opened 2023-03-23
Jetpack: register v6-video-frame-poster beta extension

retrofox:
This PR register the `v6-video-frame-poster` beta extension to be used to hide the video frame poster feature while developing it.

Part of https://github.com/Automattic/jetpack/issues/29278
Fixes #

## Proposed changes:

* Jetpack: register v6-video-frame-poster beta extension

### Other information:

- [ ] Have you written new tests for your changes, if applicable?

## Testing instructions:

* Test in Simple, Atomic, and Simple sites
* Go to the block editor
* Open the dev console
* With the [beta extension](https://github.com/Automattic/jetpack/tree/trunk/projects/plugins/jetpack/extensions#beta-extensions) disabled, confirm the `Jetpack_Editor_Initial_State.available_blocks['v6-video-frame-poster']` is not in `true`

```
Jetpack_Editor_Initial_State.available_blocks['v6-video-frame-poster']
```

image

* With the `beta` extension enabled, confirm the `Jetpack_Editor_Initial_State.available_blocks['v6-video-frame-poster']` is in `true`:

image

PR #25347, opened 2022-08-01
Publicize/Plans: Use plans package in publicize to determine if a feature is available. 

spsiddarthan:
In the Jetpack Social plugin, we are going to check if a paid feature is available.  We are going to use the 'Current_Plan' package. 

Once this PR is merged, I will work on deprecating `Jetpack_Plan` in the Jetpack Plugin in a new PR and use the `Current_Plan` package instead. 

#### Changes proposed in this Pull Request:

* Made the Current_Plans package usage
* Add a param `refresh_plan_from_wpcom` to the `supports` method which will enable the method to optionally force refresh the plans cache before getting the plans data.

#### Jetpack product discussion

pdWQjU-4i-p2

#### Does this pull request change what data or activity we track or use?

No

#### Testing instructions:

* You can error_log($show_nudge)  and see that's it's true
* Buy the Jetpack Social Basic plan by visiting this link https://wordpress.com/checkout/following-jambu.jurassic.ninja/jetpack_social_monthly), replacing following-jambu.jurassic.ninja with your Jetpack Site name
* You can error_log($show_nudge)  and see that's it's false
* Make sure you test this with the Jetpack Plugin deactivated (Jetpack Social plugin should be activated)

PR #23789, opened 2022-04-05
Update/jetpack plans ref

mkm119:
This PR aims to delete Jetpack_Plan and Jetpack_Plans classes from Jetpack plugin directory, and update all references to these classes with the new Plans package introduced [here](https://github.com/Automattic/jetpack/pull/23503)

Fixes #

#### Changes proposed in this Pull Request:

*

#### Testing instructions:

* Make sure all the tests are passing
* Further instructions will be added shortly

PR #31207, opened 2023-06-06
Plans: update to remove Jetpack-the-plugin specificities

### LINKED ISSUE #30288: Jetpack Plugin: Swap to using Plans\Current_Plan for plan checks
We started doing making the change to using the Plans package in [this PR](https://github.com/Automattic/jetpack/pull/23789). We're already using it for our plan checks in Jetpack Social, and have implemented [the heartbeat to refresh the plan data](https://github.com/Automattic/jetpack/blob/d8e48e7e9ec0c05b631a2ccd847de3b11b9dcc5f/projects/plugins/social/src/class-jetpack-social.php#L192) but it might be good to implement Configure for the package, so this is standardised.

Initially, it would be good to retire Jetpack_Plan, so the Jetpack plugin is using the package.

After that, we should migrate the other plugins to use it too. [VideoPress](https://github.com/Automattic/jetpack/blob/d8e48e7e9ec0c05b631a2ccd847de3b11b9dcc5f/projects/packages/videopress/src/class-options.php#L57) and [Protect](https://github.com/Automattic/jetpack/blob/d8e48e7e9ec0c05b631a2ccd847de3b11b9dcc5f/projects/plugins/protect/src/class-plan.php#L101) look to be already using it, but Search is using their own mechanism and I'm not sure about Backup etc. but we can create follow-up tasks for each plugin.

jeherve:
Fixes #30288

## Proposed changes:

This is another try at #23789.  This PR brings the `Current_Plan` up to date so it can completely replace `Jetpack_Plan` in the future.

- We add VideoPress blocks that were added in #29678
- We pull Modules data from package instead of using the Jetpack class. See #29736 for a previous discussion about this.
- We do not rely on `Jetpack_Options` directly

The `Current_Plan` isn't widely used yet, but these updates should help us make it usable in a follow-up PR. 
For reference, it is already used for Publicize, see #25347.

### Other information:

## Jetpack product discussion

* pdWQjU-4i-p2
* p9dueE-7pi-p2
* 1202726932197503-as-1202726932197515

* No

## Testing instructions:

Right now, not much should change in the current behavior. A good way to test this PR would be to use the Social plugin:

1. Install the Social plugin, without purchasing a plan, and visit the Social menu page to see the prompt to upgrade.
2. Upgrade and see that the state is updated in the menu page.

### DIFF
--- a/projects/packages/plans/changelog/update-current-plan
+++ b/projects/packages/plans/changelog/update-current-plan
@@ -1,0 +1,4 @@
+Significance: patch
+Type: changed
+
+Remove Jetpack-the-plugin dependencies from Current Plan class.
--- a/projects/packages/plans/package.json
+++ b/projects/packages/plans/package.json
@@ -1,7 +1,7 @@
 {
 	"private": true,
 	"name": "@automattic/jetpack-plans",
-	"version": "0.3.1",
+	"version": "0.3.2-alpha",
 	"description": "Fetch information about Jetpack Plans from wpcom",
 	"homepage": "https://github.com/Automattic/jetpack/tree/HEAD/projects/packages/plans/#readme",
 	"bugs": {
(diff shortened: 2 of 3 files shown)


fgiannar (approved):
LGMT!

PR #31211, opened 2023-06-06
AI Assistant: improve the process of getting post content

### LINKED ISSUE #30619: AI Assistant: Change previous content detection to include blocks like the list block
When detecting content above the AI Assistant block, it checks for `content` block attribute, which does not work for lists or images with captions.

To reproduce:
- Add a list
- Add an AI Assistant block
- Add a paragraph
- Check that the AI Assistant block does not detect previous content
- Move the AI Assistant block below the paragraph
- Check that the AI Assistant block now detects the paragraph content

retrofox:
This PR improves the process of getting post content.
Currently, the approach relies on the `content` block attribute, which doesn't scale properly.
In this PR, the process has two steps:

* Serialize the blocks
* Transform the serialized content to markdown.

The second step removes the serialized part on the fly since, in the end, they are comments from the HTML perspective.

Fixes https://github.com/Automattic/jetpack/issues/30619

## Proposed changes:

* AI Assistant: improve the process of getting post content

### Other information:

## Testing instructions:

* Go to the block editor
* Add content to the post by using different block types: headers, paragraphs, lists, event images (set the image caption, for instance)
* Request an action to will consider the post content, for instance, translate:
* Confirm the prompt contains all content independently of the block type
* Confirm that the prompt can handle images

https://github.com/Automattic/jetpack/assets/77539/6f6333c9-8e58-48d5-bf57-0cf4286367e0

image


dhasilva (approved):
:brain: 
Very nice work! It reads the content as expected. The only issue I found is that it does not regenerate tables, but that is to be expected for now. LGTM!

PR #31212, opened 2023-06-06
Fix: Redirecting from setup page to dashboard

bindlegirl:
If user tries to connect Jetpack but clicks browser back button before approving the user 
connection, they will get redirected to the Jetpack dashboard but the header will be 
missing and the URL will still show /setup page.

## Proposed changes:

* Adding redirection from /setup to /dashboard for sites with site connection regardless of 
the user connection.

### Other information:

## Jetpack product discussion

1191179647901802-as-1204540829524155/f

no
## Testing instructions:

* Go to dashboard and click Set up Jetpack
* Before approving the user connection click browser back button
* You should go back to the dashboard but the Jetpack logo will be missing and URL should 
end with /setup
* Disconnect Jetpack
* Load this PR and try same steps again
* Dashboard should look OK and URL should show /dashboard instead of /setup

### DIFF
--- a/projects/plugins/jetpack/_inc/client/main.jsx
+++ b/projects/plugins/jetpack/_inc/client/main.jsx
@@ -504,7 +504,6 @@ class Main extends React.Component {
 			case '/connect-user':
 			case '/connect-user-setup':
 			case '/woo-setup':
-			case '/setup':
 				pageComponent = (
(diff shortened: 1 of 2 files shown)


sergeymitr (approved):
Works great 👍

PR #31213, opened 2023-06-06
Plans: use the Jetpack Plans package in the Jetpack plugin, and deprecate existing class

jeherve:
Fixes #31213

## Proposed changes:

In this PR, we start using the package in the Jetpack plugin, and deprecate the existing class. No features should be impacted by the change, everything should work just like before.

### Other information:

## Jetpack product discussion

* pdWQjU-4i-p2
* p9dueE-7pi-p2
* 1202726932197503-as-1202726932197515

* No

## Testing instructions:

Lots of things to test here. I would also recommend tailing the site's logs as you test, to ensure no calls of `Jetpack_Plan` (now returning deprecation notices) remains.

* Start by connecting your site to WordPress.com.
* Check all the features where files are touched.


projects/plugins/jetpack/class.jetpack-plan.php
```diff
@@ -2,39 +2,45 @@
...
-	private static $active_plan_cache;
-
 	/**
 	 * The name of the option that will store the site's plan.
 	 *
+	 * @deprecated $$next-version$$ use Automattic\Jetpack\Current_Plan::PLAN_OPTION
+	 *
 	 * @var string
 	 */
 	const PLAN_OPTION = 'jetpack_active_plan';
```
anomiex: Worth doing like this?
```suggestion
	const PLAN_OPTION = Current_Plan::PLAN_OPTION;
```
and the same for the other constants below?

projects/packages/plans/src/class-current-plan.php
```diff
@@ -41,6 +41,7 @@ class Current_Plan {
 				'jetpack_free',
 			),
 			'supports' => array(
+				'advanced-seo',
```
anomiex, line 44: I was wondering what this was for, until I saw how putting this here simplifies the logic in `Jetpack_SEO_Utils::is_enabled_jetpack_seo()` where before the plan was only checked for 'advanced-seo' on wpcom while on non-wpcom it just always returned true.

Also confirmed that wpcom doesn't get to the point of looking for this here, since it [exits early](https://github.com/Automattic/jetpack/blob/1caf07bac4f4740d55c817169a2be62241d641da/projects/packages/plans/src/class-current-plan.php#L352) instead.

Mentioning it in this comment in case someone doing git archaeology looking for why this got added lands here.

anomiex (dismissed):
Looks ok, diffing trunk `projects/plugins/jetpack/class.jetpack-plan.php` and `projects/packages/plans/src/class-current-plan.php` shows minimal differences and some quick poking around with this on wpcom didn't blow up.

Not sure what's going on with all the failing E2E tests.

projects/plugins/jetpack/class.jetpack-plan.php
jeherve: Sure thing! b7ad7fc23979d88a0d7da164219b094cb677b47e

projects/packages/status/src/class-modules.php
```diff
@@ -441,7 +442,7 @@ public function activate( $module, $exit = true, $redirect = true ) {
 				}
 			}
 
-			if ( class_exists( 'Jetpack_Plan' ) && ! \Jetpack_Plan::supports( $module ) ) {
+			if ( Jetpack_Plan::supports( $module ) ) {
```
anomiex: Tyop. Seems to be what's making the E2E tests fail.
```suggestion
			if ( ! Jetpack_Plan::supports( $module ) ) {
```

anomiex (commented):
Glad I decided to poke at the E2E failures, found a bug.

Also this needs a version bump now.

projects/packages/status/src/class-modules.php
jeherve: Good catch!

anomiex (approved):

PR #18579, opened 2021-01-28
Build: Fix more always-changing stuff

anomiex:
#### Changes proposed in this Pull Request:

* Use a constant build directory in the action because the path winds up
    in vendor/composer/installed.json for all the included monorepo packages.
* Set COMPOSER_ROOT_VERSION in build-production-php
    
    Without this (or .version or .extra.branch-version in composer.json),
    composer pulls the git commit sha for a 'reference' field in
    vendor/composer/InstalledVersions.php. And unlike when it does so for
    monorepo packages, it doesn't even check for the most recent commit that
    actually touched projects/plugins/jetpack/.

#### Jetpack product discussion

None.

#### Does this pull request change what data or activity we track or use?

No.

#### Testing instructions:

* Make a PR depending on this one that doesn't touch anything in projects. The build artifacts from this PR and that PR should be the same, no differences inside Jetpack's vendor directory.

#### Proposed changelog entry for your changes:

* None needed.

PR #31220, opened 2023-06-06
Added project versioning suffixes.

zinigor:
This PR adds conditions to the build step that add suffixes to `alpha` versions based on the GitHub Action run ID. This will allow for the mirror repos to have package versions that are versioned with consecutive numbers.

## Proposed changes:
* Adds alpha suffixes to package versions.

### Other information:

## Testing instructions:

I have tested the changes on [my fork of the repo](https://github.com/zinigor/jetpack/actions). You can re-run the build and download the artifacts to check the versions.


tools/cli/commands/build.js
```diff
@@ -809,6 +809,14 @@ async function buildProject( t ) {
...
 								let massagedVer = ctxPkg.version;
 								massagedVer = `^${ massagedVer }`;
+								if ( process.env.GITHUB_RUN_NUMBER && massagedVer.endsWith( 'alpha' ) ) {
+									massagedVer =
+										massagedVer +
+										'.' +
+										process.env.GITHUB_RUN_NUMBER +
+										'.' +
+										process.env.GITHUB_RUN_ATTEMPT;
+								}
```
anomiex: This is not necessary, as this comes from the data structure that has already been munged by the code added at line 942 below.

Further, I'm not sure we even _want_ the munged version here. It makes unnecessary diffs in composer.json for everything.

```suggestion
```

tools/cli/commands/build.js
```diff
@@ -849,6 +857,14 @@ async function buildProject( t ) {
...
```
anomiex: Same as above.

```suggestion
```

tools/cli/commands/build.js
```diff
@@ -923,6 +939,14 @@ async function buildProject( t ) {
...
 		if ( match && match[ 1 ] ) {
 			projectVersionNumber = match[ 1 ].replace( /[[\]]/g, '' );
+			if ( process.env.GITHUB_RUN_NUMBER && projectVersionNumber.endsWith( 'alpha' ) ) {
+				projectVersionNumber =
+					projectVersionNumber +
+					'.' +
+					process.env.GITHUB_RUN_NUMBER +
+					'.' +
+					process.env.GITHUB_RUN_ATTEMPT;
+			}
```
anomiex: As I mentioned above, I'm not sure we really want to add the run number to `projectVersionNumber` since it has the side effect of changing all the composer.json deps. Possibly we should do this instead:

* A few lines down (around line 959), do like this to set the appended version in a separate variable.
  ```js
  let projectRunVersionNumber = projectVersionNumber;
  if ( process.env.GITHUB_RUN_NUMBER && projectVersionNumber.endsWith( 'alpha' ) ) {
      projectRunVersionNumber += '.' + process.env.GITHUB_RUN_NUMBER + '.' + process.env.GITHUB_RUN_ATTEMPT;
  }
  ```
* Add another member into the `t.ctx.versions` object holding this new variable, so we'd have both the existing `version: projectVersionNumber` and also something like `runversion: projectRunVersionNumber`.
* Line 579 would then change to use the new "runversion" instead of `version`.
  ```js
  versions[ t.ctx.versions[ dep ].name ] = t.ctx.versions[ dep ].runversion;
  ```

anomiex (commented):
Works as intended, but some suggestions for improvement inline.

I note this also undoes some of the work we put in in PRs like #18579 and #23867 in avoiding diffs being created for every plugin for every build due to run-specific data being injected.

<details>
<summary>We could avoid that by doing something like `git log -1 --format=%ct ${ t.cwd }` to get the timestamp of the last commit touching the package instead of using the run number. Maybe only for slugs starting with `packages/` since those are the only ones where it matters. Think that's worth the effort?</summary>

More specifically, that could look like this.
```js
let projectRunVersionNumber = projectVersionNumber;
if ( t.project.startsWith( 'packages/' ) && projectVersionNumber.endsWith( 'alpha' ) ) {
    const ts = (
        await t.execa( 'git', [ 'log', '-1', '--format=%ct', '.' ], {
            cwd: t.cwd,
            stdio: [ null, 'pipe', null ],
        } )
    ).stdout;
    if ( ts.match( /^\d+$/ ) ) {
        projectRunVersionNumber += '.' + ts;
    }
}
```
</details>

(Yeah, I know, using `GITHUB_RUN_NUMBER` was my idea in the first place. I can have bad ideas too! 😀)

tools/cli/commands/build.js
zinigor: Thank you, I have implemented your suggestions in the latest commit.

tools/cli/commands/build.js
```diff
@@ -921,17 +920,20 @@ async function buildProject( t ) {
...
-					'.' +
-					process.env.GITHUB_RUN_ATTEMPT;
+			projectRunVersionNumber = projectVersionNumber = match[ 1 ].replace( /[[\]]/g, '' );
+			if ( t.project.startsWith( 'packages/' ) && projectVersionNumber.endsWith( 'alpha' ) ) {
+				const ts = (
+					await t.execa( 'git', [ 'log', '-1', '--format=%ct', '.' ], {
+						cwd: t.cwd,
+						stdio: [ null, 'pipe', null ],
+					} )
+				 ).stdout;
```
anomiex: Weird indent here.
```suggestion
				).stdout;
```

tools/cli/commands/build.js
```diff
@@ -919,10 +920,21 @@ async function buildProject( t ) {
...
 		if ( match && match[ 1 ] ) {
-			projectVersionNumber = match[ 1 ].replace( /[[\]]/g, '' );
+			projectRunVersionNumber = projectVersionNumber = match[ 1 ].replace( /[[\]]/g, '' );
+			if ( t.project.startsWith( 'packages/' ) && projectVersionNumber.endsWith( 'alpha' ) ) {
+				const ts = (
+					await t.execa( 'git', [ 'log', '-1', '--format=%ct', '.' ], {
+						cwd: t.cwd,
+						stdio: [ null, 'pipe', null ],
+					} )
+				).stdout;
```
anomiex: Apparently prettier is being stupid. Sigh.
```suggestion
				 ).stdout;
```

tools/cli/commands/build.js
anomiex: It seems Prettier has decided to be stupid here and insist on this space.

tools/cli/commands/build.js
```diff
@@ -919,10 +920,21 @@ async function buildProject( t ) {
 		crlfDelay: Infinity,
 	} );
 
-	rl.on( 'line', line => {
+	rl.on( 'line', async line => {
```
anomiex: Using `async` here doesn't work right. When the `await` happens `rl` continues processing lines, which can result in a later version line of the changelog getting processed, incorrectly overwriting `projectVersionNumber` with a wrong version before the `git` command finishes.

Probably the best thing to do here would be to move lines 927–937 outside of the event handler entirely, probably to around what's currently line 947.

anomiex (changes requested):

anomiex (approved):
Comparing the artifacts between this PR and the corresponding trunk run shows the expected changes now. :ship:

PR #31226, opened 2023-06-06
ci: Fix checkout logic in pr-is-up-to-date workflow

anomiex:
## Proposed changes:

We need to check out trunk for pushes to tags and to PRs.

But for pushes to trunk, we need to check out the actual commit rather than whatever is currently trunk to avoid skipping commits.

### Other information:

## Testing instructions:

* Check that it worked for this PR, and in particular that the "checkout" step checked out trunk.
* Merge to trunk, then check that it worked for that too and that the "checkout" step used a SHA rather than "trunk".

### DIFF
--- a/.github/workflows/pr-is-up-to-date.yml
+++ b/.github/workflows/pr-is-up-to-date.yml
@@ -14,12 +14,18 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 5  # 2021-03-23: The run on push to the tag might take a minute or two.
     steps:
-      - name: Checkout trunk
+
+      # We basically have two workflows in one here, one for pushes to trunk and one for PRs and pushes to tags.
+      # The reason we don't separate them into two is because GitHub's UI would then always be showing a skipped job
+      # in the PR check list, which is kind of annoying.
+
+      # First, the "PR or tag" job.
+
+      - name: Checkout trunk for tag push or PR
         uses: actions/checkout@v3
+        if: github.event_name != 'push' || github.ref != 'refs/heads/trunk'
         with:
           ref: trunk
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-is-up-to-date.yml
```diff
@@ -14,12 +14,18 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 5  # 2021-03-23: The run on push to the tag might take a minute or two.
     steps:
-      - name: Checkout trunk
+
+      # We basically have two workflows in one here, one for pushes to trunk and one for PRs and pushes to tags.
+      # The reason we don't separate them into two is because GitHub's UI would then always be showing a skipped job
+      # in the PR check list, which is kind of annoying.
```
anomiex, line 20: Yeah, this is kind of weird, but it has been the case since #19352.

.github/workflows/pr-is-up-to-date.yml
```diff
@@ -14,12 +14,18 @@ jobs:
...
+
+      # First, the "PR or tag" job.
+
+      - name: Checkout trunk for tag push or PR
         uses: actions/checkout@v3
+        if: github.event_name != 'push' || github.ref != 'refs/heads/trunk'
         with:
           ref: trunk
-          # The "Check whether the tag needs updating for trunk commit" needs the previous commit for diffing.
-          fetch-depth: 2
```
zinigor, line 22: I'm assuming that without this here we'll be checking out the entire repo. Does this mean that deepen is no longer needed?

anomiex: No, without this here it will default to `fetch-depth: 1`. Then the deepen step will deepen it if necessary.

zinigor (approved):

PR #31242, opened 2023-06-07
PHP8 compatibility updates.

markbiek:
## Proposed changes:

Updates for PHP 8 compatibility.

### Other information:

## Testing instructions:

Visual inspection

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/business-hours/business-hours.php
+++ b/projects/plugins/jetpack/extensions/blocks/business-hours/business-hours.php
@@ -158,7 +158,7 @@ function render( $attributes ) {
 				gmdate( $time_format, $opening ),
 				gmdate( $time_format, $closing )
 			);
-			if ( $key + 1 < count( $day['hours'] ) ) {
+			if ( is_countable( $day['hours'] ) && $key + 1 < count( $day['hours'] ) ) {
 				$days_hours .= ', ';
 			}
 		}
--- a/projects/plugins/jetpack/extensions/blocks/business-hours/business-hours.php
+++ b/projects/plugins/jetpack/extensions/blocks/business-hours/business-hours.php
@@ -147,20 +147,20 @@ function render( $attributes ) {
 		$content   .= '<dd class="' . esc_attr( $day['name'] ) . '">';
 		$days_hours = '';
 
-		foreach ( $day['hours'] as $key => $hour ) {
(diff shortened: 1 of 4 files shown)


projects/plugins/jetpack/extensions/blocks/business-hours/business-hours.php
```diff
@@ -158,7 +158,7 @@ function render( $attributes ) {
 				gmdate( $time_format, $opening ),
 				gmdate( $time_format, $closing )
 			);
-			if ( $key + 1 < count( $day['hours'] ) ) {
+			if ( is_countable( $day['hours'] ) && $key + 1 < count( $day['hours'] ) ) {
```
anomiex: If it's not a numerically-indexed array, other code here is also going to do the wrong thing. Probably we should rework the loop to do something like this
```php
		foreach ( $day['hours'] as $hour ) {
			$opening = strtotime( $hour['opening'] );
			$closing = strtotime( $hour['closing'] );
			if ( ! $opening || ! $closing ) {
				continue;
			}
			if ( $days_hours !== '' ) {
				$days_hours .= ', ';
			}
			$days_hours .= sprintf(
				'%1$s - %2$s',
				gmdate( $time_format, $opening ),
				gmdate( $time_format, $closing )
			);
		}
```

projects/plugins/jetpack/extensions/blocks/podcast-player/podcast-player.php
```diff
@@ -104,7 +104,7 @@ function render_block( $attributes, $content ) {
 		return render_error( __( 'Your podcast URL is invalid and couldn\'t be embedded. Please double check your URL.', 'jetpack' ) );
 	}
 
-	if ( isset( $attributes['selectedEpisodes'] ) && count( $attributes['selectedEpisodes'] ) ) {
+	if ( isset( $attributes['selectedEpisodes'] ) && is_countable( $attributes['selectedEpisodes'] ) && count( $attributes['selectedEpisodes'] ) ) {
```
anomiex: Testing `count( $attributes['selectedEpisodes'] )` is basically checking for it not being an empty array (or for being a non-array in PHP 7, but in that case the `array_map()` just below would break). So we could simplify this:
```suggestion
	if ( ! empty( $attributes['selectedEpisodes'] ) ) {
```

anomiex (approved):

PR #31243, opened 2023-06-07
PHP8 compatibility updates.

markbiek:
## Proposed changes:

Updates for PHP 8 compatibility.

### Other information:

## Testing instructions:

Visual inspection

### DIFF
--- a/projects/plugins/jetpack/class.json-api.php
+++ b/projects/plugins/jetpack/class.json-api.php
@@ -175,8 +175,7 @@ class WPCOM_JSON_API {
 	 */
 	public static function init( $method = null, $url = null, $post_body = null ) {
 		if ( ! self::$self ) {
-			$class      = function_exists( 'get_called_class' ) ? static::class : __CLASS__; // phpcs:ignore PHPCompatibility.PHP.NewFunctions.get_called_classFound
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/class.json-api.php
```diff
@@ -175,7 +175,7 @@ class WPCOM_JSON_API {
 	 */
 	public static function init( $method = null, $url = null, $post_body = null ) {
 		if ( ! self::$self ) {
-			$class      = function_exists( 'get_called_class' ) ? get_called_class() : __CLASS__; // phpcs:ignore PHPCompatibility.PHP.NewFunctions.get_called_classFound
+			$class      = function_exists( 'get_called_class' ) ? static::class : __CLASS__; // phpcs:ignore PHPCompatibility.PHP.NewFunctions.get_called_classFound
 			self::$self = new $class( $method, $url, $post_body );
```
anomiex: This replacement results in code that really doesn't make any sense.

But since we dropped PHP 5.2 support years ago, we can just do it like this.
```suggestion
			self::$self = new static( $method, $url, $post_body );
```

anomiex (approved):

PR #31245, opened 2023-06-07
Launchpad: Add helper function to get the task statuses from site options

sixhours:
## Proposed changes:

* It looks like we will use this line of code frequently given the new method of marking tasks as complete, so create a function to return the task statuses from site options to avoid repetition.
* Update existing instances of the line of code to use the function instead.
* No visual/functional changes.

### Other information:

## Testing instructions:

* Apply this patch to your sandbox and sandbox the API
* Create a new site from `/start` and ensure the Launchpad still works as expected.
* You can double-check that you're getting the expected results from the developer console, too.


andres-blanco (commented):
Looks like the option `launchpad_checklist_tasks_statuses` is being used [in other places](https://github.com/search?q=repo%3AAutomattic%2Fjetpack%20launchpad_checklist_tasks_statuses&type=code).
 Should those be updated to?

sixhours:
> Looks like the option launchpad_checklist_tasks_statuses is being used [in other places](https://github.com/search?q=repo%3AAutomattic%2Fjetpack%20launchpad_checklist_tasks_statuses&type=code).
Should those be updated to?

Hmm, good point. I hadn't even seen one of those files. I dunno, now that I'm looking at it, do you think it's worth adding this helper at all? Or does it add an unnecessary layer of abstraction?

andres-blanco:
> > Looks like the option launchpad_checklist_tasks_statuses is being used [in other places](https://github.com/search?q=repo%3AAutomattic%2Fjetpack%20launchpad_checklist_tasks_statuses&type=code).
> > Should those be updated to?
> 
> Hmm, good point. I hadn't even seen one of those files. I dunno, now that I'm looking at it, do you think it's worth adding this helper at all? Or does it add an unnecessary layer of abstraction?

It would be good to have it in the long run, but right now it's not really necessary. I think we can leave this as cleanup task for later.

sixhours:
Closing, I don't think this is particularly useful!

PR #29180, opened 2023-02-27
General: update code detected by Rector for PHP 8.1

kraftbj:
- Add rector config file that matches what WP.com is using
- _inc: Running JP through Reactor for PHP 8.1 compat

WordPress.com is preparing to move to PHP 8.1 and flagged various things in Jetpack that may need to be updated to prevent issues. This is an intial PR to add the settings WP.com is using and updating _inc files as an initial group.

## Proposed changes:

* Various.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

* p5TWut-OG-p2
* #23695

## Does this pull request change what data or activity we track or use?

No.

## Testing instructions:

* pending

PR #31250, opened 2023-06-07
PHP8 compatibility updates.

markbiek:
## Proposed changes:

Updates for PHP 8 compatibility. This is a fresh start for the work begun in #29180.

### Other information:

## Testing instructions:

Visual inspection

### DIFF
--- a/projects/plugins/jetpack/_inc/lib/class-jetpack-tweetstorm-helper.php
+++ b/projects/plugins/jetpack/_inc/lib/class-jetpack-tweetstorm-helper.php
@@ -392,7 +392,12 @@ class Jetpack_Tweetstorm_Helper {
 		} else {
 			$lines = array( $block['text'] );
 		}
-		$line_total = count( $lines );
+
+		if ( is_countable( $lines ) ) {
+			$line_total = count( $lines );
+		} else {
+			$line_total = 0;
+		}
 
 		// Keep track of how many characters from this block we've allocated to tweets.
 		$current_character_count = 0;
@@ -443,8 +448,13 @@ class Jetpack_Tweetstorm_Helper {
 			}
 
 			// The line is too long for a single tweet, so split it by sentences, or linebreaks.
-			$sentences      = preg_split( '/(?|(?<!\.\.\.)(?<=[.?!]|\.\)|\.["\'])(\s+)(?=[\p{L}\'"\(])|(\n+))/u', $line_text, -1, PREG_SPLIT_DELIM_CAPTURE );
-			$sentence_total = count( $sentences );
+			$sentences = preg_split( '/(?|(?<!\.\.\.)(?<=[.?!]|\.\)|\.["\'])(\s+)(?=[\p{L}\'"\(])|(\n+))/u', $line_text, -1, PREG_SPLIT_DELIM_CAPTURE );
+
+			if ( is_countable( $sentences ) ) {
+				$sentence_total = count( $sentences );
+			} else {
+				$sentence_total = 0;
+			}
 
 			// preg_split() puts the blank space between sentences into a seperate entry in the result,
 			// so we need to step through the result array by two, and append the blank space when needed.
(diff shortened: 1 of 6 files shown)


projects/plugins/jetpack/_inc/lib/class-jetpack-tweetstorm-helper.php
```diff
@@ -392,7 +392,12 @@ private static function add_text_to_tweets( $block ) {
 		} else {
 			$lines = array( $block['text'] );
 		}
-		$line_total = count( $lines );
+
+		if ( is_countable( $lines ) ) {
```
anomiex: Given lines 391 and 393, it seems unlikely it will ever not be a countable in a version of PHP where `count()` cares about being passed a non-countable. PHP 8 will throw or fatal from the `explode()` rather than returning null or false, while in earlier versions it might but `count()` at most only logs a deprecation warning.

But OTOH this doesn't really hurt anything, especially if reduced to be just one line instead of five.

projects/plugins/jetpack/_inc/lib/class-jetpack-tweetstorm-helper.php
```diff
@@ -392,7 +392,12 @@ private static function add_text_to_tweets( $block ) {
...
+			$line_total = count( $lines );
+		} else {
+			$line_total = 0;
+		}
```
anomiex: Why not
```suggestion
		$line_total = is_countable( $lines ) ? count( $lines ) : 0;
```
like it did in previous PRs? Same for similar cases below.

projects/plugins/jetpack/_inc/lib/class-jetpack-tweetstorm-helper.php
```diff
@@ -899,7 +916,7 @@ private static function get_boundary( $block, $offset ) {
 
 		$block_def = self::get_block_definition( $block['name'] );
 
-		if ( isset( $block_def['content'] ) && count( $block_def['content'] ) > 0 ) {
+		if ( isset( $block_def['content'] ) && is_countable( $block_def['content'] ) && count( $block_def['content'] ) > 0 ) {
```
anomiex: `$block_def` can only ever be null or come from an entry in the private `self::$supported_blocks`, which all have `'content'` either not set or set to an array. There's not even any filter that might add other kinds of values. So we should be safe to do like this (since `empty( array() )` returns true):
```suggestion
		if ( ! empty( $block_def['content'] ) ) {
```

projects/plugins/jetpack/_inc/lib/class-jetpack-tweetstorm-helper.php
```diff
@@ -1325,7 +1342,11 @@ function ( $tweet ) {
 					$tweet['blocks'] = array();
 				} else {
 					// Remove the parts of the block data that the editor doesn't need.
-					$block_count = count( $tweet['blocks'] );
+					if ( is_countable( $tweet['blocks'] ) ) {
```
anomiex: Another case where it would be hard for it to ever not be a countable, since line 1341 has already verified that either `$tweet['blocks'][0]['attributes']` or `$tweet['blocks'][0]['clientId']` is set. So unless `$tweet['blocks']` is some kind of non-countable ArrayAccess or something else extremely weird...

projects/plugins/jetpack/_inc/lib/class.media-summary.php
```diff
@@ -264,7 +264,7 @@ public static function get( $post_id, $blog_id = 0, $args = array() ) {
 				$return['secure']['image'] = self::ssl_img( $return['image'] );
 				++$return['count']['image'];
 
-				if ( $number_of_paragraphs <= 2 && 1 === count( $extract['image'] ) ) {
+				if ( $number_of_paragraphs <= 2 && is_countable( $extract['image'] ) && 1 === count( $extract['image'] ) ) {
```
anomiex, line 267: I note line 263 would also have problems if it's somehow not the expected array, although that line will only warn while this in PHP 8 will throw.

projects/plugins/jetpack/_inc/lib/class-jetpack-tweetstorm-helper.php
markbiek: I had gotten feedback from other places about not using ternaries. But if y'all don't object to them here, I'll switch it!

anomiex: We're fine with ternaries. Personally I wish WordPress core's coding standards would stop disallowing the Elvis operator (`?:`) so we could use that where appropriate too.

anomiex (approved):
Seems fine to me. I'd have been ok with leaving some of the `is_countable`s as such too.

PR #31251, opened 2023-06-07
tools: Add deploy-assets-to-svn script

anomiex:
## Proposed changes:

PR #30869 added directories for w.org SVN's assets to each plugin. This adds a script to help deploy those properly to SVN.

The new script prepares and pushes an SVN commit that does the following:

* Syncs the plugin's `.w.org-assets/` dir to SVN `assets/`
* Copies the plugin's `readme.txt` to SVN trunk, taking care not to change the stable tag.
* Copies the new `readme.txt` into the stable tag too, since that's where w.org will actually load it from, taking care not to change the stable tag or changelog.

### Other information:

No

## Testing instructions:

* Run the script.
* Commit some changes to readme and/or assets for a plugin, then run the script again.

### DIFF
--- a/tools/deploy-assets-to-svn.sh
+++ b/tools/deploy-assets-to-svn.sh
@@ -1,0 +1,225 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+BASE=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)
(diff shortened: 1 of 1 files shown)


zinigor (approved):
I haven't actually tested the script, I'd suggest we merge it and actually battle test it with real assets. But the code looks sane to me.

PR #31255, opened 2023-06-07
AI Assistant: Check if the block is in the block editor context before enabling Accept title action

### LINKED ISSUE #31192: AI Assistant: Fix "Accept title" action on P2, site editor and widgets
ref pe4Cmx-1km-p2#comment-651

On P2, when using "Accept title", the block is removed but the title is gone. That happens because the title of a P2 post is defined in its content.

It is necessary to check and fix or disable for:

- P2 post
- P2 comment
- Site editor
- Widgets

dhasilva:
Fixes #31192

## Proposed changes:

* Adds an initial state object with the current screen to determine if the block is inserted in the block editor
* Only shows "Accept title" when the block is in the block editor

### Other information:

## Testing instructions:

Block editor:
* Go to the block editor
* Add some content (tip: use the AI Assistant block to generate a haiku)
* Add an AI Assistant block
* Click on Write with AI... > Generate a post title
* Check that the `Accept title` button is displayed when the title is generated and clicking it applies the title as normal

Site editor
* Go to the site editor (wp-admin > Appearance > Editor)
* Repeat the process of the block editor
* Check that the `Accept title` button is replaced by the `Accept` button and clicking it adds the title inline like other content

P2
* Set up a P2 test site: https://github.com/Automattic/p2#set-up-a-sandboxed-test-site
* Apply this PR to your sandbox and enable the redirect on your hosts file
* In your sandboxed P2 test site, create a post
* Repeat the previous process of creating a title
* Check that the `Accept title` button is replaced by the `Accept` button and clicking it adds the title inline like other content

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/ai-assistant.php
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/ai-assistant.php
@@ -33,7 +33,7 @@ add_action( 'init', __NAMESPACE__ . '\register_block' );
  *
  * @return array|null
  */
-function initial_state() {
+function get_initial_state() {
 	$base = null;
 
 	if ( function_exists( 'get_current_screen' ) ) {
@@ -52,9 +52,9 @@ function initial_state() {
  * @return string
  */
 function render_initial_state() {
-	$initial_state = initial_state();
+	$initial_state = get_initial_state();
 
-	if ( ! $initial_state['currentScreen'] ) {
+	if ( empty( $initial_state['currentScreen'] ) ) {
 		return null;
 	}
 
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/ai-assistant.php
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/ai-assistant.php
@@ -29,6 +29,54 @@ function register_block() {
 add_action( 'init', __NAMESPACE__ . '\register_block' );
 
 /**
+ * Get the initial state data for hydrating the React UI.
+ *
+ * @return array|null
+ */
+function initial_state() {
+	$base = null;
+
+	if ( function_exists( 'get_current_screen' ) ) {
+		$screen = get_current_screen();
+		$base   = $screen->base;
+	}
+
+	return array(
+		'currentScreen' => $base,
+	);
+}
+
+/**
+ * Render the initial state into a JavaScript variable.
+ *
+ * @return string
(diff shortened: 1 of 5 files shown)


projects/plugins/jetpack/extensions/blocks/ai-assistant/ai-assistant.php
```diff
@@ -28,6 +28,54 @@ function register_block() {
...
+ * @return string
+ */
+function render_initial_state() {
+	$initial_state = get_initial_state();
+
+	if ( empty( $initial_state['currentScreen'] ) ) {
+		return null;
+	}
+
+	return 'var jetpackAIAssistantInitialState=JSON.parse(decodeURIComponent("' . rawurlencode( wp_json_encode( $initial_state ) ) . '"));';
```
retrofox: Can we store these data in the `Jetpack_Editor_Initial_State` global object? For instance, [here we expose AI Assistant feature data](https://github.com/Automattic/jetpack/blob/c6e4509b63b6f074abcfd713a35f2d2d13993615/projects/plugins/jetpack/class.jetpack-gutenberg.php#L713).

https://github.com/Automattic/jetpack/blob/878ece2673a0d54f749a2e0436aaa8597380829a/projects/plugins/jetpack/class.jetpack-gutenberg.php#L713

image

projects/plugins/jetpack/extensions/blocks/ai-assistant/ai-control.js
```diff
@@ -28,6 +28,8 @@ import origamiPlane from './icons/origami-plane';
 import { isUserConnected } from './lib/connection';
 import UpgradePrompt from './upgrade-prompt';
 
+const isInBlockEditor = window?.Jetpack_Editor_Initial_State?.screenBase === 'post';
```
retrofox, line 31: I don't know the API, so wondering how reliable this condition is. I was taking a look at the [is_block_editor()](https://developer.wordpress.org/reference/classes/wp_screen/is_block_editor/) method. Just FYI.

retrofox (approved):
LGTM. Nice one. TIL about WP_Screen 🤦

projects/plugins/jetpack/extensions/blocks/ai-assistant/ai-control.js
dhasilva: Is block editor returns `true` in the site editor.

Issue #30738, opened 2023-05-15
AI Assistant streaming endpoint tracking


nunyvega:
Make sure in the new streaming endpoint, we’re tracking properly what we were tracking before, in the previous endpoint

lhkowalski:
I went through the backend code, following the line of execution for a given chat completion request, and taking note of any log or tracking record that we may do on the request.

A summary of the flow and the log events:

**For the old version:**

1. File `class-wpcom-rest-api-v2-endpoint-ai.php` - API entrypoint
   - No logs or event tracking
1. File `class-jetpack-ai-helper.php`
   - `mark_post_as_ai_assisted` - add metadata to a post counting the number of times some AI request got processed coming from the post
1. File `lib/openai/openai.php` - method `request_gpt_completion`
   - possible `gpt3_completion_error`
   - possible `gpt3_bad_word` and related moderation logs, if the prompt contains bad words
   - estimated costs for `openai_gpt3`
   - possbile `gpt3_completion`

**For the new, Streamed Version:**

1. File `endpoints/jetpack-openai-query.php` - API entrypoint
   - No logs or event tracking
1. File `class-jetpack-ai-helper.php`
   - It's not used anymore, so the `mark_post_as_ai_assisted` call is missing
   - something to add on the new code
1. File `lib/openai/openai.php` - method `request_chat_completion`
   - The method changed, but both methods track the same logs and events:
   - possible `gpt3_completion_error`
   - possible `gpt3_bad_word` and related moderation logs, if the prompt contains bad words
   - estimated costs for `openai_gpt3`
   - possible `gpt3_completion`

So, by comparing the flows on the backend, the only thing that is different is the tracking of the number of AI assisted calls being performed on a given post. I don't think it's being used internally for any kind of control, but we can add it so we can fetch insights about it in the future.

Tomorrow I will check the frontend code as well.

lhkowalski:
On the frontend code, there was no tracking before the change to streaming. The code for streaming the response included a tracks event for `jetpack_ai_gpt3_completion` so we started to track it after the change. No action needed on this side.

lhkowalski:
Edit: I was seeing it as low priority but seems like the post_id goes all the way down to the logstash defaults array, so it gets registered on all the important logs that we do. This is done for WPCOM sites.

Here are the action items to comply with this:

- [ ] include code to mark a given post as AI assisted, if possible
- [x] send the post_id being edited on the completion request, so it can be logged on the backend - PR: #30765
[...]

PR #31262, opened 2023-06-07
AI Assistant: Request WPCOM JWT endpoint as user, not blog

lhkowalski:
Part of #30738. Test with D113109-code.

## Proposed changes:

* Jetpack sites handle the JWT token request on `jetpack/v4/jetpack-ai-jwt` and forward it to `wpcom/v2` `/sites/$site/jetpack-openai-query/jwt`, so we can stablish the right chain of authentication
* Before this change, the request is being performed as the `blog`, so no user info is sent on the request
* After this change, the same user that requested the Jetpack endpoint is forwarded to the WPCOM request, using `wpcom_json_api_request_as_user`

### Other information:

## Testing instructions:

* Apply D113109-code to your sandbox
* Make sure your test site points to the sandbox when requesting the `public-api` (case you test using a JN site; locally everything is set, as long as you are sandboxing the `public-api` on your machine)
* Go to the block editor, open your browser console on the network tab and add an "AI Assistant" block
* Inspect the response from the `/jetpack-ai-jwt` request; you will find a token in the response
* Copy the token, go to https://jwt.io/#debugger-io and paste the token on the debugger
* Notice the `user_id` property
   * before this change, it would be set to `0` since the request is performed as the blog
   * after this change, it should be set with the user ID of the WPCOM account connected on your test Jetpack site (run v1.1 `/me` request to discover your account ID), since now the request is performed as the user
* The block should work normally

### DIFF
--- a/projects/plugins/jetpack/_inc/lib/class.core-rest-api-endpoints.php
+++ b/projects/plugins/jetpack/_inc/lib/class.core-rest-api-endpoints.php
@@ -775,7 +775,7 @@ class Jetpack_Core_Json_Api_Endpoints {
 	public static function get_openai_jwt() {
 		$blog_id = \Jetpack_Options::get_option( 'id' );
 
-		$response = \Automattic\Jetpack\Connection\Client::wpcom_json_api_request_as_blog(
+		$response = \Automattic\Jetpack\Connection\Client::wpcom_json_api_request_as_user(
 			"/sites/$blog_id/jetpack-openai-query/jwt",
 			'2',
 			array(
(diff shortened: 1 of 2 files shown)


retrofox:
> Jetpack sites handle the JWT token request on jetpack/v4/jetpack-ai-jwt and forward it to wpcom/v2 /sites/$site/jetpack-openai-query/jwt, so we can stablish the right chain of authentication

Why don't we have the same endpoint on the Jetpack site? I mean `wpcom/v2 /jetpack-openai-query/jwt`?

I know this topic is out of the PR scope, just wondering.

retrofox (approved):
This is awesome. Thanks for the testing instructions.

image

PR #31266, opened 2023-06-08
Testing comment reminder: update to include WoA testing details.

jeherve:
## Proposed changes:

Let's provide information about testing on WoA as well, to invite folks to test on that environment as well.

> **Note**
> I opted to suggest the Jetpack Beta plugin option first, since that doesn't require a WoA dev blog, and requires less technical skills I think. I did mention `jetpack rsync` as well.

### Other information:

## Jetpack product discussion

* See p9dueE-7pi-p2#comment-9797

* No

## Testing instructions:

See this test PR for an example:
https://github.com/Automattic/jetpack/pull/31267#issuecomment-1582390568

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -146,10 +146,11 @@ jobs:
             - To test on Simple, run the following command on your sandbox:
             ${ commands }
 
-            *Interested in more testing tips?*
+            *Interested in more tips and information?*
 
             - In your local development environment, use the \`jetpack rsync\` command to sync your changes to a WoA dev blog.
-            - Find out more about testing on WordPress.com Simple here: p9dueE-5Nn-p2`;
+            - Read more about our development workflow here: PCYsg-eg0-p2
+            - Figure out **when your changes will be shipped to customers** here: PCYsg-eg5-p2`;
             await github.rest.issues.updateComment( {
               owner: context.repo.owner,
               repo: context.repo.repo,
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -140,9 +140,16 @@ jobs:
               `;
             }, '' );
             const commentBody = `${ TEST_COMMENT_INDICATOR }
-            Are you an Automattician? You can now test your Pull Request on WordPress.com. On your sandbox, run
+            Are you an Automattician? Please test your changes on WordPress.com.
+
+            - To test on WoA, install [the Beta plugin](https://jetpack.com/download-jetpack-beta/), go to Jetpack > Jetpack Beta > Jetpack and enable the `\${ BRANCH_NAME }`\ branch.
+            - To test on Simple, run the following command on your sandbox:
             ${ commands }
-            to get started. More details: p9dueE-5Nn-p2`;
+
+            *Interested in more testing tips?*
+
+            - In your local development environment, use the `\jetpack rsync\` command to sync your changes to a WoA dev blog.
+            - Find out more about testing on WordPress.com Simple here: p9dueE-5Nn-p2`;
             await github.rest.issues.updateComment( {
               owner: context.repo.owner,
               repo: context.repo.repo,
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -142,13 +142,13 @@ jobs:
             const commentBody = `${ TEST_COMMENT_INDICATOR }
             Are you an Automattician? Please test your changes on WordPress.com.
 
-            - To test on WoA, install [the Beta plugin](https://jetpack.com/download-jetpack-beta/), go to Jetpack > Jetpack Beta > Jetpack and enable the `\${ BRANCH_NAME }`\ branch.
+            - To test on WoA, install [the Beta plugin](https://jetpack.com/download-jetpack-beta/), go to Jetpack > Jetpack Beta > Jetpack and enable the \`${ BRANCH_NAME }\` branch.
             - To test on Simple, run the following command on your sandbox:
             ${ commands }
 
             *Interested in more testing tips?*
 
-            - In your local development environment, use the `\jetpack rsync`\ command to sync your changes to a WoA dev blog.
+            - In your local development environment, use the \`jetpack rsync\` command to sync your changes to a WoA dev blog.
             - Find out more about testing on WordPress.com Simple here: p9dueE-5Nn-p2`;
             await github.rest.issues.updateComment( {
               owner: context.repo.owner,
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -148,7 +148,7 @@ jobs:
 
             *Interested in more testing tips?*
 
-            - In your local development environment, use the `\jetpack rsync\` command to sync your changes to a WoA dev blog.
+            - In your local development environment, use the `\jetpack rsync`\ command to sync your changes to a WoA dev blog.
             - Find out more about testing on WordPress.com Simple here: p9dueE-5Nn-p2`;
             await github.rest.issues.updateComment( {
               owner: context.repo.owner,
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -140,7 +140,7 @@ jobs:
               `;
             }, '' );
             const commentBody = `${ TEST_COMMENT_INDICATOR }
-            Are you an Automattician? Please test your changes on WordPress.com.
+            Are you an Automattician? Please test your changes on all WordPress.com environments to help mitigate accidental explosions.
 
             - To test on WoA, install [the Beta plugin](https://jetpack.com/download-jetpack-beta/), go to Jetpack > Jetpack Beta > Jetpack and enable the \`${ BRANCH_NAME }\` branch.
             - To test on Simple, run the following command on your sandbox:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -140,9 +140,16 @@ jobs:
               `;
             }, '' );
             const commentBody = `${ TEST_COMMENT_INDICATOR }
-            Are you an Automattician? You can now test your Pull Request on WordPress.com. On your sandbox, run
+            Are you an Automattician? Please test your changes on WordPress.com.
```
danielbachhuber:
```suggestion
            Are you an Automattician? Please test your changes on all WordPress.com environments to help mitigate accidental explosions.
```

.github/workflows/build.yml
```diff
@@ -140,9 +140,16 @@ jobs:
...
+
+            - To test on WoA, install [the Beta plugin](https://jetpack.com/download-jetpack-beta/), go to Jetpack > Jetpack Beta > Jetpack and enable the \`${ BRANCH_NAME }\` branch.
```
danielbachhuber:
```suggestion
            - To test on a WoA site: Create a new Business site and take it Atomic, install and activate [the Jetpack Beta plugin](https://jetpack.com/download-jetpack-beta/), go to Jetpack > Jetpack Beta > Jetpack and enable the \`${ BRANCH_NAME }\` branch.
```

.github/workflows/build.yml
```diff
@@ -140,9 +140,16 @@ jobs:
...
+            - To test on Simple, run the following command on your sandbox:
             ${ commands }
-            to get started. More details: p9dueE-5Nn-p2`;
+
```
danielbachhuber, line 148: Can we include some detail here about when/how the code will be deployed after merge?

I've found it difficult in the past to figure out when and how my code will end up in production.

danielbachhuber, line 148:
```suggestion
            Note: Because the command applies the entire Jetpack `trunk`, you may end up with a lot of changed files in your sandbox. Make sure to clean all of them up once you're done with your testing.

```

Hopefully this applies cleanly after the last command.

danielbachhuber (commented):
Already much better! I left a few suggestions and had a question.

.github/workflows/build.yml
jeherve: >  Create a new Business site and take it Atomic

I worry this may be a bit cryptic for some. What do you think about 

```suggestion
            - To test on WoA, go to the Plugins menu on a WordPress.com Simple site. Click on the "Upload" button and follow the upgrade flow to be able to upload, install, and activate [the Jetpack plugin](https://jetpack.com/download-jetpack-beta/). Once the plugin is active, go to Jetpack > Jetpack Beta > Jetpack and enable the \`${ BRANCH_NAME }\` branch.
```

danielbachhuber: @jeherve I think that works too 👍

.github/workflows/build.yml
anomiex: For this, and maybe the detailed installation instructions above too, it might be better to just include a "p9dueE-5Nn-p2"-style doc ref instead of making this comment extremely detailed. Plus it's easier to edit a doc than to have to commit changes to the monorepo every time something changes.

danielbachhuber: As a WordPress.com developer, it would be helpful to have this information easily accessible, instead of hidden behind a link.

jeherve: > it might be better to just include a "p9dueE-5Nn-p2"-style doc ref instead of making this comment extremely detailed. 

That's also what I had in mind. I worry that if the comment gets too long, folks will ignore it.

I was updating some of our FieldGuide pages with that in mind, and would suggest something like this:

```diff
diff --git a/.github/workflows/build.yml b/.github/workflows/build.yml
index 8afb3754a6..cdf40a472d 100644
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -146,10 +146,11 @@ jobs:
             - To test on Simple, run the following command on your sandbox:
             ${ commands }
 
-            *Interested in more testing tips?*
+            *Interested in more tips and information?*
 
             - In your local development environment, use the \`jetpack rsync\` command to sync your changes to a WoA dev blog.
-            - Find out more about testing on WordPress.com Simple here: p9dueE-5Nn-p2`;
+            - Read more about our development workflow here: PCYsg-eg0-p2
+            - Figure out **when your changes will be shipped to customers** here: PCYsg-eg5-p2`;
             await github.rest.issues.updateComment( {
               owner: context.repo.owner,
               repo: context.repo.repo,

```

What do you think?

jeherve: I think this would be covered with the above:
https://github.com/Automattic/jetpack/pull/31266/files#r1223229356

danielbachhuber: > Figure out **when your changes will be shipped to customers** here: PCYsg-eg5-p2

This doesn't actually say _when_, though:

image

Nor does PCYsg-Jjm-p2. And `#jetpack-releases` isn't linked in the first document.

Atomic makes it pretty clear, and this clarity is helpful:

image

anomiex: It says as close to when as we have information available. It's not like we have a specific time to do any of it, particularly since it's not always a person in the same timezone being on the rotation to do it.

Plus some of what's there has changed pretty recently, and might do so again in the near future. As I mentioned, it's easier to update that doc page (which will have to be updated anyway) than to have to make more commits in here.

jeherve: The good news is, we can update and iterate on that page as needed. I added the link to the Slack channel just now, for example. I did not add at what time deployments happen each day since I do not think there is any official agreed upon deployment hours yet. We can add those later when that's defined.

Edit: looks like I posted at the same time as @anomiex, my bad, didn't see your comment until after I posted mine.

anomiex: P.S. I note the Atomic screenshot says when "update alerts" happen, but not how soon they act on those alerts.

.github/workflows/build.yml
```diff
@@ -140,9 +140,16 @@ jobs:
               `;
             }, '' );
             const commentBody = `${ TEST_COMMENT_INDICATOR }
-            Are you an Automattician? You can now test your Pull Request on WordPress.com. On your sandbox, run
+            Are you an Automattician? Please test your changes on all WordPress.com environments to help mitigate accidental explosions.
+
+            - To test on WoA, go to the Plugins menu on a WordPress.com Simple site. Click on the "Upload" button and follow the upgrade flow to be able to upload, install, and activate [the Jetpack plugin](https://jetpack.com/download-jetpack-beta/). Once the plugin is active, go to Jetpack > Jetpack Beta > Jetpack and enable the \`${ BRANCH_NAME }\` branch.
```
anomiex, line 145: This strikes me as a bit long. Personally I'd go back to what you had earlier, and include a doc ref for even more verbose details in case someone doesn't know how to add a plugin yet.

.github/workflows/build.yml
```diff
@@ -140,9 +140,16 @@ jobs:
...
+            - To test on Simple, run the following command on your sandbox:
             ${ commands }
-            to get started. More details: p9dueE-5Nn-p2`;
+
+            *Interested in more testing tips?*
+
+            - In your local development environment, use the \`jetpack rsync\` command to sync your changes to a WoA dev blog.
```
anomiex, line 151: It can also work with a sandbox, or a JN site or the like too. Reference to PCYsg-eg0-p2 might be appropriate. For that matter, that page could possibly use some cleaning up.

.github/workflows/build.yml
danielbachhuber: We can agree to disagree, then. I think hiding information away in documents makes it less accessible, particularly when the documents themselves get out of date.

Putting the information in the GitHub bot message:
1. Makes the information more accessible and obvious.
2. Helps ensure it will be corrected sooner when it's out of date (because folks are looking at it every day).

.github/workflows/build.yml
jeherve: Done in 76cd497d2317007b86fec2cb280b8ab6affb3348

.github/workflows/build.yml
jeherve: 76cd497d2317007b86fec2cb280b8ab6affb3348 should cover this.

jeherve: I committed the above in 76cd497d2317007b86fec2cb280b8ab6affb3348. We can see how this change is received, and iterate in follow-up PRs based on the feedback we'll get.

.github/workflows/build.yml
jeherve: It is a bit long indeed. It may, however, stand out a bit more like this. In the spirit of compromising, I think it is a good middleground based on [the conversation](https://github.com/Automattic/jetpack/pull/31266#discussion_r1223501147) we had with @danielbachhuber above.

We can iterate in follow-up PRs based on the feedback we'll get from everyone once this is merged.

anomiex: Seems like it'll be TL;DR to me. How many people do we really expect to get this message, be able to follow it, but not know how to install a plugin?

jeherve: I think that can be the case for folks used to Jetpack or WordPress.com Simple development (or getting started with those ), but not familiar with the Atomic infrastructure.

zinigor (approved):
Thank you for iterating over the messaging, it looks good to me!

PR #31271, opened 2023-06-08
AI Assistant: introduce block Message component

### LINKED ISSUE #31105: AI Assistant: add Footer component
Image

### LINKED ISSUE #31106: AI Assistant: make more evident user needs to accept generated content
> If you don’t hover over the blocks, you have no idea that you need to accept it (we received this feedback twice in the CFT)

p1685525644508259-slack-C054LN8RNVA

### LINKED ISSUE #31244: AI Assistant: make it clear that generated content needs to be accepted to be visible on the frontend
As an alternative to #31100, we need to make it clear to users that the generated content will not be visible on the frontend just after being generated, it needs to be accepted so it becomes blocks that are then visible.

Check design iterations to see how far we went on this, and also reference other AI products to gather ideas.

Slack Discussion: p1686086042667839-slack-C054LN8RNVA

retrofox:
Fixes https://github.com/Automattic/jetpack/issues/31105
Fixes https://github.com/Automattic/jetpack/issues/31244
Fixes https://github.com/Automattic/jetpack/issues/31106

## Proposed changes:

* AI Assistant: introduce block Message area

### Other information:

## Testing instructions:

* Go to the block editor
* Use the AI Assistant
* Check when the content is generated, confirm you see the message:

image


ederrengifo:
I think it is valuable to use this area to clarify potential confusion in the UX related to the block, but maybe:
- We only use it for specific use cases; for example, from the list you share, I think only the last one addresses a major potential confusion, the other ones seem to add more noise than clarification, especially if the users sees that every time. OR
- We include it only for the first use of the block or when we notice the user cannot perform
 an action, like too much time without accepting a suggestion. There are opportunities for the block to read context there.
 
 I personally like the latter, but I would love to hear @keoshi inputs, maybe we keep it as a low priority improvement?
 
 Edit: Also, we shouldn't discard this to be used to communicate potential malfunction and link the support page.

retrofox:
> * We only use it for specific use cases; for example, from the list you share, I think only the last one...

I do agree

> * We include it only for the first use of the block ...

I think it could be easily doable by adding a flag/attribute.

> I personally like the latter, but I would love to hear @keoshi inputs, maybe we keep it as a low priority improvement?

Sure, but just in case, this PR tries to tackle the "I Assistant: make it clear that generated content needs to be accepted to be visible on the frontend" issue, with should be addressed this cycle/week. 

> Edit: Also, we shouldn't discard this to be used to communicate potential malfunction and link the support page.

Agree.

dhasilva (approved):
LGTM

PR #31274, opened 2023-06-08
AI Assistant: avoid to send the same AI request

retrofox:
This PR disables the `Send` button when the user prompt has been used before. The idea is avoiding to send the same request twice.

Fixes #

## Proposed changes:

* AI Assistant: avoid to send the same AI request

### Other information:

## Testing instructions:

* Go to the block editor
* Create a new AI Assistant block instance
* Type a user prompt
* Send it
* Once the generated text is done:
  * Confirm the `Send` button is disabled
  * Confirm the button enables/disables depending on the current prompt is the same as the sent one

https://github.com/Automattic/jetpack/assets/77539/34c6dd0e-d03c-4f96-89c1-b82a32f99679

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/ai-control.js
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/ai-control.js
@@ -11,7 +11,7 @@ import {
 	ToolbarGroup,
 	Spinner,
 } from '@wordpress/components';
-import { forwardRef, useImperativeHandle, useRef } from '@wordpress/element';
+import { forwardRef, useImperativeHandle, useRef, useState } from '@wordpress/element';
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/ai-assistant/ai-control.js
```diff
@@ -97,6 +99,21 @@ const AIControl = forwardRef(
 			[]
 		);
 
+		/*
+		 * Ready to generate state
+		 * - User is connected
+		 * - User has entered at least 3 characters
```
dhasilva, line 105: A legitimate prompt can have only one character: p1685992984122109/1685992930.893179-slack-C054LN8RNVA

PR #31279, opened 2023-06-08
Device Detection: Prevent checking the platform when the user agent is not set

srtfisher:
Prevent checking the platform when the user agent is not set. 

Fixes an PHP 8.1 error when `strpos()` is used and argument 1 is null:

```
PHP Deprecated:  strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated in /Users/srtfisher/broadway/www/client/wp-content/mu-plugins/jetpack-12.0/jetpack_vendor/automattic/jetpack-device-detection/src/class-user-agent-info.php on line 235
```

Primarily came up during unit testing when the user agent is not set.

## Proposed changes:
* Prevents the platform from being checked when the user agent is not set. 

### Other information:

## Testing instructions:

n/a

### DIFF
--- a/projects/packages/device-detection/changelog/patch-1
+++ b/projects/packages/device-detection/changelog/patch-1
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Catch PHP notice if User Agent is not available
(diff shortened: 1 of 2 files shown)


anomiex (approved):
Thanks for the contribution!

PR #31281, opened 2023-06-08
Launchpad: Add domain upsell task to keep building list

### LINKED ISSUE #77812: Launchpad: Create "Customize your domain" task
This task should only show for users who are on a free plan and haven't purchased a domain.

This task should point to `/domains` and is considered complete when a user purchases and activates a domain for their site.

andres-blanco:
Closes https://github.com/Automattic/wp-calypso/issues/77812

## Proposed changes:

Adds the `domain_upsell` task to the `keep-building` list. 

### Other information:

## Testing instructions:

* Apply this patch to your sandbox and sandbox `public-api.wordpress.com`.
* Visit the dev console and select WP REST API, `wpcom/v2`.
* You'll need to get a free site and purchase a domain (make sure to cancel the subscription after testing)
* Make a GET request to the site with the domain you just created `/sites/[siteSlug]/launchpad?checklist_slug=keep-building`.
* Verify you see the `domain_upsell` task as completed.
* Make the same GET request but use a free site without any domain purchases.
* Verify you do see the `domain_upsell` task marked as incomplete.
* Make the same GET request but use a site with a plan.
* Verify you don't see `domain_upsell` task.

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -36,6 +36,7 @@ function wpcom_launchpad_get_task_definitions() {
 			},
 			'is_complete_callback' => 'wpcom_is_domain_upsell_completed',
 			'badge_text_callback'  => 'wpcom_get_domain_upsell_badge_text',
+			'is_visible_callback'  => 'wpcom_is_domain_upsell_visible',
 		),
 		'first_post_published'            => array(
 			'get_title'             => function () {
@@ -359,6 +360,29 @@ function wpcom_is_domain_upsell_completed( $task, $default ) {
 function wpcom_get_domain_upsell_badge_text() {
 	// Never run `wpcom_is_checklist_task_complete` within a is_complete_callback unless you are fond of infinite loops.
 	return wpcom_is_checklist_task_complete( 'domain_upsell' ) ? '' : __( 'Upgrade plan', 'jetpack-mu-wpcom' );
+}
+
+/**
+ * Determines whether or not domain upsell task should be visible.
+ *
+ * @return bool True if user is on a free plan with no custom domain name.
+ */
+function wpcom_is_domain_upsell_visible() {
+	if ( ! function_exists( 'wpcom_get_site_purchases' ) ) {
+		return false;
+	}
+
+	// Check if the site has any domain purchases.
+	$site_purchases = wpcom_get_site_purchases();
+
+	$domain_purchases = array_filter(
+		$site_purchases,
+		function ( $site_purchase ) {
+			return in_array( $site_purchase->product_type, array( 'domain_map', 'domain_reg' ), true );
+		}
+	);
+
+	return ! empty( $domain_purchases );
 }
 
 /**
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -368,6 +368,14 @@ function wpcom_get_domain_upsell_badge_text() {
  * @return bool True if user is on a free plan with no custom domain name.
  */
 function wpcom_is_domain_upsell_visible() {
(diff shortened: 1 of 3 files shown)


projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -355,6 +356,37 @@ function wpcom_get_domain_upsell_badge_text() {
...
 
+/**
+ * Determines whether or not domain upsell task should be visible.
+ *
+ * @return bool True if user is on a free plan with no custom domain name.
+ */
+function wpcom_is_domain_upsell_visible() {
+	if ( ! class_exists( '\A8C\Billingdaddy\Container' ) ) {
+		return false;
+	}
```
daledupreez: This means the task will only be visible when we're on WordPress.com Simple sites. I think it would be more robust for us to rely on the data returned from `wpcom_get_site_purchases()` and check for items with `$purchase->product_type === 'bundle'`.

andres-blanco: The visibility condition that I was looking for is: `User is on a free plan with no custom domain name.`, I've added this check since I don't think it's possible to be on a free plan on a non-simple site. 

Maybe I am missing something?

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -355,6 +356,38 @@ function wpcom_get_domain_upsell_badge_text() {
 	return wpcom_is_checklist_task_complete( 'domain_upsell' ) ? '' : __( 'Upgrade plan', 'jetpack-mu-wpcom' );
 }
 
+/**
+ * Determines whether or not domain upsell task should be visible.
+ *
+ * @return bool True if user is on a free plan with no custom domain name.
+ */
+function wpcom_is_domain_upsell_visible() {
```
daledupreez: Nit: it may be cleaner to call this something like `wpcom_is_domain_upsell_task_visible()`, as we have lots of domain upsells, so it would be better to avoid any possible confusion or ambiguity.
```suggestion
function wpcom_is_domain_upsell_task_visible() {
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -355,6 +356,38 @@ function wpcom_get_domain_upsell_badge_text() {
...
+	$site_purchases = wpcom_get_site_purchases();
+
+	$bundle_purchases = array_filter(
+		$site_purchases,
+		function ( $site_purchase ) {
+			return $site_purchase->product_type === 'bundle';
+		}
+	);
+
+	$is_free_plan = empty( $bundle_purchases );
```
daledupreez: Non-blocking: It may be worth using this as an early return condition instead.
```suggestion
	// If we have a paid plan, we don't want to show the domain upsell task.
	if ( ! empty( $bundle_purchases ) ) {
		return false;
	}
```

daledupreez (dismissed):
Looks good to me, though I do have two suggestions you may want to fold in.

sixhours (dismissed):
This is working as described!

My only question is whether someone who sees this task initially (users on the Free plan) should continue to see it after completion. Right now the task will be removed from the list entirely. It feels like it should still be shown, just checked off like the other tasks. 

What do you think?

andres-blanco:
> This is working as described!
> 
> My only question is whether someone who sees this task initially (users on the Free plan) should continue to see it after completion. Right now the task will be removed from the list entirely. It feels like it should still be shown, just checked off like the other tasks.
> 
> What do you think?

Yes, definitely. It changes a little bit the initial logic. Now it will be:
* The task is visible if the site is on a free plan
* The task is completed if the site is on a paid plan with the `custom-domain` feature (this was the previous logic) OR if the site has a `domain_map` | `domain_reg` product.

I will update this PR to adjust to those rules.

andres-blanco:
@sixhours I updated the logic and the testing instructions!

sixhours (approved):
Screen Shot 2023-06-13 at 2 14 43 PM

Looks good! :shipit:

PR #31282, opened 2023-06-08
AI Assistant: block query requests when required upgrade

### LINKED ISSUE #31198: AI Assistant: block client requests when the site achieves the requests limit
Once the site achieves the requests limit and requires an upgrade, we should block the client from hitting the endpoint for at least one minute. We should show let the users know by displaying a message.

retrofox:
This PR checks if the site requires an upgrade before hitting the query endpoint.
This PR makes evident that we need a better data handling, probably by using Redux or useContext() React hook.

Fixes https://github.com/Automattic/jetpack/issues/31198

## Proposed changes:

* AI Assistant: block query requests when required upgrade

### Other information:

## Testing instructions:

* Go to the block editor
* Perform a query request
* Confirm it works as expected
* Test with a site that requires an upgrade
* When you see the upgrade banner, request a new query, for instance, by using Tone or Translate.
* Confirm the query request never takes off
* Confirm you see the error

Screenshot 2023-06-08 at 16 40 27
Screenshot 2023-06-08 at 16 40 02

### DIFF
--- a/projects/plugins/jetpack/changelog/update-ai-assistant-block-requests-when-limit-achieved
+++ b/projects/plugins/jetpack/changelog/update-ai-assistant-block-requests-when-limit-achieved
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
+AI Assistant: block query requests when required upgrade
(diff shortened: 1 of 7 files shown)


dhasilva (approved):
LGTM

PR #31286, opened 2023-06-08
AI Assistant: Move caret to end of generated content when accepting

### LINKED ISSUE #31236: AI Assistant: Move caret to end of content after accepting it
After accepting the generated content, the caret stays at the start of the content. Moving it to the end would make the editing flow a bit easier.
It should be taken in consideration that the content can end in something other than a paragraph and setting the caret is not always possible.

ref p1686094525662569-slack-C054LN8RNVA

dhasilva:
Fixes #31236

## Proposed changes:

* When accepting a content, gets the last editable element and moves the caret to its end

### Other information:

## Testing instructions:

* Go to the block editor
* Add an AI Assistant block
* Generate some content
* Accept it
* Check that the keyboard caret is moved to the end of the content

Before

https://github.com/Automattic/jetpack/assets/8486249/44f5fc29-1d94-46e5-8314-212b52220433

After

https://github.com/Automattic/jetpack/assets/8486249/f49376ee-eb3a-4d67-b041-5aaaf44b703d

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/edit.js
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/edit.js
@@ -139,11 +139,40 @@ export default function AIAssistantEdit( { attributes, setAttributes, clientId }
 	// Content is loaded
 	const contentIsLoaded = !! attributes.content;
 
-	const handleAcceptContent = () => {
-		replaceBlocks(
-			clientId,
-			rawHandler( { HTML: markdownConverter.render( attributes.content ) } )
-		);
+	const getLastEditableElement = newContentBlocks => {
+		let lastEditableElement = null;
+
+		newContentBlocks.forEach( block => {
+			const element = document.querySelector( `.wp-block[data-block="${ block.clientId }"]` );
+			if ( element.contentEditable === 'true' ) {
+				lastEditableElement = element;
+			}
+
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/ai-assistant/edit.js
```diff
@@ -139,11 +139,40 @@ export default function AIAssistantEdit( { attributes, setAttributes, clientId }
...
-		replaceBlocks(
-			clientId,
-			rawHandler( { HTML: markdownConverter.render( attributes.content ) } )
-		);
+	const getLastEditableElement = newContentBlocks => {
+		let lastEditableElement = null;
+
+		newContentBlocks.forEach( block => {
+			const element = document.querySelector( `.wp-block[data-block="${ block.clientId }"]` );
+			if ( element.contentEditable === 'true' ) {
```
retrofox, line 151: Is it possible that `element` is undefined? Maybe `element.contentEditable` can trigger an exception.

retrofox (dismissed):
I agree with you (partially, haha). It's important that we pay some attention to the approach of inspecting DOM elements in Gutenberg.
I still wonder whether the issue is worth it :-/

Anyway, Nice work!!! 🚀 

I left a comment that I think we should take in consideration.

retrofox (approved):
🚢

PR #31295, opened 2023-06-09
Beta: fix PHP 8.1 deprecation warning for array manipulation

jeherve:
## Proposed changes:

This should avoid warnings like this one on sites with PHP 8.1+:

```
PHP Deprecated:  key(): Calling key() on an object is deprecated
```

### Other information:

## Jetpack product discussion

* Primary issue: #23695

* No

## Testing instructions:

* On a site running PHP 8.1+, visit the Jetpack Beta admin pages.
* Ensure that you can still search for branches and switch to a new branch.

### DIFF
--- a/projects/plugins/beta/src/admin/plugin-manage.template.php
+++ b/projects/plugins/beta/src/admin/plugin-manage.template.php
@@ -194,9 +194,10 @@ if ( is_plugin_active( $plugin->plugin_file() ) ) {
 		</div>
 		<div id="section-pr">
 			<?php
-			end( $manifest->pr );
-			$last = key( $manifest->pr );
-			foreach ( $manifest->pr as $k => $pr ) {
+			$pr_list = (array) $manifest->pr;
+			end( $pr_list );
+			$last = key( $pr_list );
+			foreach ( $pr_list as $k => $pr ) {
 				$branch = $plugin->source_info( 'pr', $pr->branch );
 				if ( $branch && ! is_wp_error( $branch ) ) {
 					// Add spaces around the branch name for historical reasons.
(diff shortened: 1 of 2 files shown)


projects/plugins/beta/src/admin/plugin-manage.template.php
```diff
@@ -194,9 +194,10 @@ class="is-primary jp-form-button dops-button is-primary is-compact jptracks"
 		</div>
 		<div id="section-pr">
 			<?php
-			end( $manifest->pr );
-			$last = key( $manifest->pr );
-			foreach ( $manifest->pr as $k => $pr ) {
+			$pr_list = (array) $manifest->pr;
+			end( $pr_list );
+			$last = key( $pr_list );
+			foreach ( $pr_list as $k => $pr ) {
```
anomiex, line 200: If we don't care about being able to run with WordPress < 5.9 on PHP < 7.3, we could just do
```suggestion
			$last = array_key_last( (array) $manifest->pr );
			foreach ( $manifest->pr as $k => $pr ) {
```
(with the appropriate `phpcs:ignore` or adjustment to `.phpcs.dir.xml` to let phpcompatibility know about the WP Core ployfill of course)

anomiex (approved):

projects/plugins/beta/src/admin/plugin-manage.template.php
jeherve: Since the beta plugin is publicly available, I think it needs to remain PHP 5.6+ compatible.

anomiex: https://github.com/WordPress/WordPress/blob/41396a44f520f973b9ea481083981385e9b005a6/wp-includes/compat.php#L404-L416

jeherve: Ah, I didn't know about the polyfill! 👍

PR #31318, opened 2023-06-12
Spike: Jetpack Sync: Rework Queue class to support different backend storage engines.

bisko:
## Proposed changes:

This PR reworks the `Queue` class to extract the logic interfacing with the database to a separate class.

This is a preparation to moving towards a dedicated table for Sync events in the near future. 

This PR is a Spike and will be merged in multiple follow-up PRs:

* #32089
* #32111
* #32135

### Other information:
N/a

## Jetpack product discussion

p9dueE-7e6-p2

No

## Testing instructions:

Testing instructions are supplied in subsequent PRs


projects/packages/sync/src/class-queue.php
```diff
@@ -46,6 +78,180 @@ public function __construct( $id ) {
 		$this->id           = str_replace( '-', '_', $id ); // Necessary to ensure we don't have ID collisions in the SQL.
 		$this->row_iterator = 0;
 		$this->random_int   = wp_rand( 1, 1000000 );
+
+		/**
+		 * If the Dedicated table is not force-disabled (status = -1), we can try to initialize it.
+		 */
+		if ( ! self::dedicated_table_disabled() ) {
+			$this->maybe_initialize_dedicated_sync_table();
+			$this->queue_storage = new Queue_Storage_Table( $this->id );
```
fgiannar, line 90: I believe we need to only set this if `$this->maybe_initialize_dedicated_sync_table();` returns `true`

projects/packages/sync/src/sync-queue/class-queue-storage-table.php
```diff
@@ -0,0 +1,471 @@
...
+			return false;
+		}
+
+		// Try to read from the table
+
+		// Ignore the interpolated table name
+		// phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		$query = $wpdb->query( "SELECT count(`ID`) FROM {$this->table_name}" );
+
+		if ( ! $query ) {
```
fgiannar:
```suggestion
		if ( false === $query ) {
```

As it might return 0 results

projects/packages/sync/src/sync-queue/class-queue-storage-table.php
```diff
@@ -0,0 +1,471 @@
...
+
+	/**
+	 * Drop the dedicated table as part of cleanup.
+	 *
+	 * @return bool If the table is cleared. It's using `is_dedicated_table_healthy` to check.
+	 */
+	public function drop_table() {
+		global $wpdb;
+
+		// TODO do we need to check if the table is healthy or not before dropping it?
```
fgiannar: I'd suggest keeping it simple here, by only returning `$wpdb->query( "DROP TABLE {$this->table_name}" );`

projects/packages/sync/src/sync-queue/class-queue-storage-table.php
```diff
@@ -0,0 +1,471 @@
...
+
+		$charset_collate = $wpdb->get_charset_collate();
+
+		$table_definition = "CREATE TABLE {$this->table_name} (
+			  `ID` bigint(20) NOT NULL AUTO_INCREMENT,
+			  `queue_id` varchar(50) NOT NULL,
+			  `event_id` varchar(100) NOT NULL,
+			  `event_payload` text NOT NULL,
+			  `timestamp` timestamp NOT NULL DEFAULT current_timestamp(),
+			  `object_type` varchar(100) DEFAULT NULL,
```
fgiannar: Do we need the `object_type` and `parent_object_id ` here for now?

fgiannar (changes requested):
First, quick review by only looking at the code and focusing on the switch logic. Will continue testing and keep reviewing here

projects/packages/sync/src/sync-queue/class-queue-storage-table.php
bisko: Right, added them to test out things, but not needed for now.

bisko:
Forgot to close this one, as we merged all the smaller tasks quite a while ago and there's no need to keep this open.

PR #31303, opened 2023-06-09
Blocks: Add a Tock block

pablinos:
This is adding a Tock block based on the [Tock Widget plugin](https://wordpress.org/plugins/tock-widget/) 

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion
p1HpG7-n0W-p2

## Does this pull request change what data or activity we track or use?

No

## Testing instructions:
With a test site with beta blocks enabled:

* In the editor, insert a Tock block
* Add a URL slug like `roister` or `etch`
* Click Set URL
* Preview the post and check the button works as expected

PR #31338, opened 2023-06-12
Tock Block: Update toolbar and popover

pablinos:
## Proposed changes:

We decided to move the settings popover to the toolbar, instead of underneath the block, as well as change the button to text rather than an icon.

This also includes a couple of grammatical tweaks to the helper and description text.

### Other information:

## Jetpack product discussion

p1HpG7-n0W-p2#comment-63921

No

## Testing instructions:

This is the same as #31303

*Before*

image

*After*
image


projects/plugins/jetpack/extensions/blocks/tock/edit.js
```diff
@@ -44,42 +48,60 @@ const UrlPopover = ( { tockUrl, setEditedUrl, popoverAnchor, setUrl, cancel } )
 		}
 	};
 
+	const closeIfNotToggle = () => {
```
pablinos, line 51: This is all a bit messy, but we essentially have to implement our own dropdown so that we can call our own reset code on close, but only when the popover is closed by losing focus.

jeherve (approved):

pablinos:
Closing in favour of #31355

PR #31346, opened 2023-06-13
AI Assistant: Add and use ImproveToolbarDropdownMenu in block toolbar

retrofox:
This PR adds and uses the `<ImproveToolbarDropdownMenu />` component.

Fixes #

## Proposed changes:

* AI Assistant: Add and use ImproveDropdown in block toolbar

### Other information:

## Testing instructions:

* Go to the block editor
* Create and use the AI Assistant block
* Confirm the Improve dropdown works as expected
  * `Make longer`, `Make shorter`,  `Summarize`
  * Confirm when it's editing the post title, the `Summarize` option is not listed

image


projects/plugins/jetpack/extensions/blocks/ai-assistant/components/improve-dropdown/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+			} }
+		>
+			{ () => {
+				// Exclude quick edits from the list.
+				const quickActionsListFiltered = quickActionsList.filter(
+					quickAction => ! exclude.includes( quickAction.key )
+				);
+
+				return (
+					<MenuGroup label={ __( 'Improve', 'jetpack' ) }>
```
renatoagds: IMO we should give more context like we do in `Write with AI` ... Something like `Improve based on previous context`.

retrofox: Let's don't add a menu group label for now? Same as now.

retrofox: image

dhasilva (approved):
LGTM

PR #31355, opened 2023-06-13
Tock Block: Implement settings with a dropdown

pablinos:
This is an alternative version of #31338. 

## Proposed changes:

On top of the changes in #31338, this PR:

* Swaps to using the Dropdown component to make the implementation of the popover more standard
* Adds some extra logic to reset the input when the block or dropdown loses focus
* Removes the notice in the editor to prevent things jumping around.

### Other information:

## Jetpack product discussion
p1HpG7-n0W-p2#comment-63988

No.

## Testing instructions:

This is the same as #31303

You will still see a notice in the front end if you have `edit_posts` permissions

image

I think this is safe as it won't be appearing and disappearing.


samiff (dismissed):
Tests as advertised 👍 

Just a couple thoughts:

1. "Adds some extra logic to reset the input when the block or dropdown loses focus" -- Was there background on this particular change? There's no obvious feedback that input is cleared when it's not explicitly saved, so I could see that potentially tripping some users up.
2. Is there a plan for a support document to link to for this block?
3. Entering an invalid business name is possible (e.g., `test`). And when clicking the Tock button on the frontend, things don't fail too gracefully `www.exploretock.com refused to connect.`. Not sure if there is a great way around that, but thought I'd mention.

pablinos:
Thanks @samiff! I added an extra change not having seen your review, but to address your comments.

> "Adds some extra logic to reset the input when the block or dropdown loses focus"

This is actually to fix a problem with that behaviour, where I was resetting state in some circumstances, but not in others. There was never feedback about this, but I was mimicking the link/URL popover for links on text. Unless you confirm the change with 'enter' or clicking on the button, it resets.

> Is there a plan for a support document to link to for this block?

That is a good point! I'm not sure of the status on that. I'll find out.

> Entering an invalid business name is possible (e.g., test). 

Yes, this is not the best experience. I've got a vague idea of how we might get around that, by doing a request to `exploretock.com/businessname` and checking the status code, but ideally we'd have an API to call or some way of the widget checking for an invalid state. I was thinking this should be a known issue that we could try and fix in a follow up.

gmjuhasz (approved):
This works well, and code looks good to me!

PR #31356, opened 2023-06-13
Instant Search: Add a style-able class to the wrapper divs for the filter options

georgestephanis:
This can allow sites to hide filters if desired, or to make them bold, or change the color of a specific category or post type that they wish to emphasize or the like.

I didn't add anything to line 198 as I wasn't quite certain how that was used or what "groups" it renders, so I didn't want to tempt fate and mess with it, but a similar logic could likely extend to add a differentiating class there based on `group.value`

There's nothing visible that this should change or be testable, apart from some classes rendering in the markup for sites to be able to style based off of.

## Proposed changes:

* Add classes to the individual divs of each filter on Instant Search.

### Other information:

## Jetpack product discussion
Discussed the impetus for this in Slack with @robfelty, threw this together as a proof of concept of what I was hoping for.

Nope

## Testing instructions:

* Use the inspector to ensure that the generated markup includes classes on the wrapper divs in the sidebar for Filtering the search results.

### DIFF
--- a/projects/packages/search/src/instant-search/components/search-filter.jsx
+++ b/projects/packages/search/src/instant-search/components/search-filter.jsx
@@ -69,7 +69,7 @@ class SearchFilter extends Component {
 	renderDate = ( { key_as_string: key, doc_count: count } ) => {
 		const { locale = 'en-US' } = this.props;
 		return (
-			<div>
+			<div className={ `jetpack-instant-search__search-filter__dates-${ key }` }>
 				<input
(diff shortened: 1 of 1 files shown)


robfelty (commented):
This seems like a reasonable idea to me. I would also like some input from team-red though, as they know the front-end side of Jetpack Search better. I am adding them as reviewers.

jsnmoon (approved):
Also looks reasonable to me 👍

PR #30543, opened 2023-05-09
My Jetpack: Add scaffolding for a Jetpack AI interstitial

oskosk:
Kick-off for #30553

Introduces route, interstitial and product class for a Jetpack AI product

Available under wp-admin/admin.php?page=my-jetpack#/add-jetpack-ai

Stlil needs work.

image

## Proposed changes:

* Introduces product class file for Jetpack AI
* Adds route `#/add-jetpack-ai`

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Does this pull request change what data or activity we track or use?
No. Or yes. It Doesn't add a new event, but it also another path for tracking the event

## Testing instructions:

* Go to `wp-admin/admin.php?page=my-jetpack#/add-jetpack-ai`
* Conirm you see a new interstitial

PR #31368, opened 2023-06-14
My Jetpack: Add support for Stats product

### LINKED ISSUE #31342: Paid Stats: Add Stats Product for My Jetpack with mocked WPCOM product
Add Stats product in https://github.com/Automattic/jetpack/tree/52b0b88b8588e3b53702b8b2b2a61bb28159e458/projects/packages/my-jetpack/src/products

We could reference Jetpack AI as it's also sold as a Jetpack module 

https://github.com/Automattic/jetpack/blob/52b0b88b8588e3b53702b8b2b2a61bb28159e458/projects/packages/my-jetpack/src/products/class-jetpack-ai.php

We could take a look at Akismet as well. However, I don't think the PWYW is implemented on site tho.

https://github.com/Automattic/jetpack/blob/52b0b88b8588e3b53702b8b2b2a61bb28159e458/projects/packages/my-jetpack/src/products/class-anti-spam.php

As the WPCOM product might not be ready yet, we could mock the response for now. More information: pbNhbs-6YI-p2.

kangzj:
Fixes #31342 

## Proposed changes:

- Add Stats product support to My Jetpack
- Changed Security and Starter to use `Wpcom_Products::get_site_current_purchases` to reduce network requests

Next step:

- Replace the mocked WPCOM pricing with real pricing when ready

### Other information:

## Jetpack product discussion
n/a

no

## Testing instructions:

* Build Jetpack `jetpack build plugins/jetpack`
* Open `/wp-admin/admin.php?page=my-jetpack`
* Open console in browser dev tools
* Ensure `myJetpackInitialState.products.items.stats` looks like the following

image


projects/packages/my-jetpack/src/products/class-module-product.php
```diff
@@ -131,4 +133,64 @@ public static function deactivate() {
 		return Jetpack::deactivate_module( static::$module_name );
 	}
 
+	/**
+	 * Get Products that support this product
+	 *
+	 * @return array
+	 */
+	public static function get_supporting_products() {
```
dognose24: Could we use another function name like `get_containing_plans` to avoid getting confused with [get_supported_products](https://github.com/Automattic/jetpack/blob/trunk/projects/packages/my-jetpack/src/products/class-product.php#L329)?

projects/packages/my-jetpack/src/products/class-security.php
```diff
@@ -235,6 +178,15 @@ public static function get_supported_products() {
 		return array( 'backup', 'scan', 'anti-spam' );
 	}
 
+	/**
+	 * Return all the products that supports the Stats.
```
dognose24: This annotation does not conform to the functionality of this file.

projects/packages/my-jetpack/src/products/class-starter.php
```diff
@@ -226,6 +172,15 @@ public static function get_supported_products() {
 		return array( 'backup', 'anti-spam' );
 	}
 
+	/**
+	 * Return all the products it contains.
```
dognose24: This annotation seems to be the same as the function `get_supported_products` above.

dognose24 (commented):
These changes look solid and work as described. 👍🏼 I just left a few comments for naming and annotations.

projects/packages/my-jetpack/src/products/class-module-product.php
```diff
@@ -131,4 +133,64 @@ public static function deactivate() {
...
+				if ( in_array( $purchase->product_slug, static::get_supporting_products(), true ) ) {
+					return true;
+				}
+			}
+		}
+		return false;
+	}
+
+	/**
+	 * Hits the wpcom api to check stats status.
```
jsnmoon:
```suggestion
	 * Hits the wpcom api to check purchase status.
```

projects/packages/my-jetpack/src/products/class-security.php
jsnmoon:
```suggestion
	 * Return all the products that supports the security product.
```

projects/packages/my-jetpack/src/products/class-module-product.php
jsnmoon: What additional use cases does `get_supporting_products` cover when `get_supported_products` already exists?

projects/packages/my-jetpack/src/products/class-module-product.php
```diff
@@ -131,4 +133,64 @@ public static function deactivate() {
...
+	 *
+	 * @todo Maybe add caching.
+	 *
+	 * @return Object|WP_Error
+	 */
+	protected static function get_state_from_wpcom() {
```
jsnmoon: Nice 👍

jsnmoon (commented):
Testing instructions worked as expected; I left some observations below.

I also looped in @Automattic/jetpack-agora for their thoughts on how we're approaching these changes for My Jetpack.

projects/packages/my-jetpack/src/products/class-module-product.php
kangzj: It's so hard to pick a right name 😅 

The existing `get_supported_products` is to get the product supported by the current product. For example: https://github.com/Automattic/jetpack/pull/31368/files#diff-96a3f7c8a02c743fc1ef8f8382674d897195688c05c45ef1d90b7422d0be465fR178

The `get_supporting_products` function I added is on the other hand is to get what products support the current product if it makes sense.

kangzj: > Could we use another function name like get_containing_plans to avoid getting confused with get_supported_products?

I guess I'll probably just go with `get_products`. What do you think? @dognose24

projects/packages/my-jetpack/src/products/class-starter.php
kangzj:
```suggestion
	 * Return all the products that support the current product.
```

projects/packages/my-jetpack/src/products/class-module-product.php
```diff
@@ -131,4 +133,64 @@ public static function deactivate() {
...
+	 * Get Products that support the current product
+	 *
+	 * @return array
+	 */
+	public static function get_products() {
+		return array();
+	}
+
+	/**
+	 * Checks whether the current plan (or purchases) of the site already supports the product
```
jsnmoon:
```suggestion
	 * Checks whether the site already supports this product through an existing plan or purchase
```

projects/packages/my-jetpack/src/products/class-module-product.php
```diff
@@ -131,4 +133,64 @@ public static function deactivate() {
 		return Jetpack::deactivate_module( static::$module_name );
 	}
 
+	/**
+	 * Get Products that support the current product
```
jsnmoon:
```suggestion
	 * Get products that include the current product
```

projects/packages/my-jetpack/src/products/class-module-product.php
jsnmoon: > > What additional use cases does `get_supporting_products` cover when `get_supported_products` already exists?

> It's so hard to pick a right name 😅 
> 
>The existing `get_supported_products` is to get the product supported by the current product. For example: https://github.com/Automattic/jetpack/pull/31368/files#diff-96a3f7c8a02c743fc1ef8f8382674d897195688c05c45ef1d90b7422d0be465fR178
>
>The `get_supporting_products` function I added is on the other hand is to get what products support the current product if it makes sense.

Moving this discussion to a separate thread for better organization 🙂 

Two questions:
- Which products must specify a non-empty return value for this new function? I noticed we defined them for Jetpack Starter and Jetpack Security.
- What's the reason for defining this function instead of adding `stats` to Jetpack Complete's `get_supported_products`? Is it because Jetpack Complete is currently not defined as a product within My Jetpack?

I find it strange to redundantly define a parent-child relationship in two places: `get_supported_products` and `get_supporting_products`.

Take Jetpack Security, for example, which defines `array( 'backup', 'scan', 'anti-spam' )` as its "supported products". With this new `get_supporting_products`, wouldn't we also need to modify `backup`, `scan`, and `anti-spam` products to specify `array( $itself, 'jetpack_security' )` as the return value for their `get_supporting_products`?

If this is the case, then it'll be trivially easy to introduce regressions by updating the output of `get_supporting_products` of the child product but not the output of `get_supported_products` of the parent product.

jsnmoon (commented):
I left a few nitty suggestions and a request for clarification around `get_supported_products`.

Note that we may need to tweak this as we adopt our latest design iteration (p1HpG7-nar-p2), simplifying the purchase flow even further by removing the $10 commercial plan.

jsnmoon (dismissed):
A similar PR for Jetpack AI can be found [here](https://github.com/Automattic/jetpack/pull/30543). I'm okay with shipping this change iteratively 🙂

projects/packages/my-jetpack/src/products/class-module-product.php
kangzj: > Which products must specify a non-empty return value for this new function? I noticed we defined them for Jetpack Starter and Jetpack Security.

From the refactoring perspective, these Starter, Security and Stats are all Module Products. By extracting `has_required_plan` and `get_state_from_wpcom` to `Module_Product`, we managed to remove duplication and reduce API calls, but in the same time we'd need a method or member implemented in child classes to identify what products support the current products. Otherwise, `has_required_plan` wouldn't be able to be generalized.

A few of the products other than Module Products would better to be changed as well, but I want to limit the scope for the PR. And if it works well, we could go ahead and refactor those.

> What's the reason for defining this function instead of adding stats to Jetpack Complete's get_supported_products? Is it because Jetpack Complete is currently not defined as a product within My Jetpack?

I guess this is question about what a product is in My Jetpack. In my point of view, the products defined in My Jetpack are ones that we want to show Interstitial cards and also are "physical" products, meaning they are either modules, plugins or both. I don't think they are 'plans' or wpcom products.

It's bit mixed up for Security and Starter tho. I guess in the two cases, wpcom products become My Jetpack products. And that's the reason why `get_supported_products` is defined and used.

I'd like to loop in @Automattic/jetpack-agora for more information. Thanks team.

kangzj: @jsnmoon I think I'll revert the changes extracting `has_required_plan`, as I don't think it's actually working for Security or Starter. I'll only promote `get_state_from_wpcom` instead.

jsnmoon: Sounds like a plan 👍

kangzj:
@jsnmoon I reverted some changes and could use another review now. Thanks.

projects/packages/my-jetpack/src/products/class-security.php
```diff
@@ -162,45 +160,13 @@ public static function is_active() {
...
-		);
-		if ( 200 !== wp_remote_retrieve_response_code( $response ) ) {
-			return new WP_Error( 'purchases_state_fetch_failed' );
-		}
-
-		$body   = wp_remote_retrieve_body( $response );
-		$status = json_decode( $body );
-		return $status;
-	}
-
```
jsnmoon, line 196: This is a good change but slightly outside the scope of the current PR, which is to initialize the stats product in My Jetpack.

Let's go ahead and ship this since `get_state_from_wpcom`'s and `get_site_current_purchases` are identical, but it'd be nice to split these refactors into a separate PR in the future.

jsnmoon (dismissed):
Looks good 👍

jsnmoon (approved):
Pushed a trunk merge to resolve conflicts.

kangzj:
@jsnmoon thanks for resolving the conflicts 👍 

> Let's go ahead and ship this since get_state_from_wpcom's and get_site_current_purchases are identical, but it'd be nice to split these refactors into a separate PR in the future.

I agree we could use another PR for the refactoring of other modules. I did that only because the code is identical. Still it's better to change that in another PR.

Thanks again 👍

PR #31374, opened 2023-06-14
Fix for newsletters when not in strict WP post context

millerf:
Fixes https://github.com/Automattic/wp-calypso/issues/78171
Fixes https://github.com/Automattic/gold/issues/68

## Proposed changes:

* Prevent the newsletter column to be displayed on CPT

No

### P2
* Sandbox public-api.
* Sync this code to your sandbox
```
bin/jetpack-downloader test jetpack fix/nl-column-on-products
```
* Notice in `wp-admin/edit.php` there is no Newsletter column
* Notice in `wp-admin/post-new.php` there is no Newsletter panel
  
### From a new instance

* Create a WOO Express site: https://woocommerce.com/start/?freetrial=true
* Upgrade the plan to exit free trial
* Access SSH via **Settings > Hosting configuration**
* Install jetpack beta
```
cd htdocs/wp-content/plugins
wget https://github.com/Automattic/jetpack-beta/releases/latest/download/jetpack-beta.zip
unzip jetpack-beta.zip
wp plugin activate jetpack-beta
```
* Modify `Jetpack` branch to modify this one:
Screenshot 2023-06-14 at 12 22 42
* Go to the product

### Already setup
* go to https://woo-bravely-insightful-stranger.wpcomstaging.com/wp-admin/edit.php?post_type=product
* Notice there is **no** Newsletter column
*  go to https://woo-bravely-insightful-stranger.wpcomstaging.com/wp-admin/edit.php
* Notice there is Newsletter column 
* [optional] you can revert to Jetpack "bleeding edge" via `Jetpack beta` and notice the Newsletter column


romarioraffington (dismissed):

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -127,6 +135,11 @@ function is_wpcom() {
  * @return array An array of column names.
  */
 function register_newsletter_access_column( $columns ) {
+
+	if ( get_post_type() !== 'post' ) {
```
jeherve: Instead of adding that check, I would recommend hooking into `manage_post_columns` instead of `manage_posts_columns`; that should help.

For more info, see https://developer.wordpress.org/reference/hooks/manage_posts_columns/ vs. https://developer.wordpress.org/reference/hooks/manage_post_type_posts_columns/

jeherve (dismissed):

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
millerf: You mean `manage_post_posts_columns` ? I didn't know about that hook. Nice!

samiff (approved):
Tests well :shipit:

PR #31353, opened 2023-06-13
Update/hide customizer and custom css on block based themes

rcrdortiz:
Fixes https://github.com/Automattic/wp-calypso/issues/71130

## Proposed changes:

* We remove the `Customize` and `Additional CSS` submenus when the theme is block-based.
* We disable the Safe CSS free trial since it is returning a Custom CSS comment that makes the Additional CSS submenu visible.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [x] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [x] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* Sandbox public API.
* Copy the branch in WPCom and use a block based theme.
* The `Customize` and `Additional CSS` submenus should be hidden.
* Switch to a non-block based theme.
* The submenus should now be visible.
* Use this branch in an Atomic site and repeat the steps.
* Use this branch in a JN site and make sure that it works just as it did before applying this branch.

PR #31376, opened 2023-06-14
Enhance/block themes hide customizer

rcrdortiz:
Follow up for https://github.com/Automattic/jetpack/pull/31353
## Proposed changes:

* Removed the customizer submenu on self-hosted sites when using a block-based theme.

### Other information:

## Testing instructions:

* Create a JN site or self host
* Add this Jetpack branch
* When using a block based theme you shouldn't see the customize submenu
* When the theme isn't block based you should see the customize submenu.

### DIFF
--- a/projects/plugins/jetpack/changelog/enhance-block-themes-hide-customizer
+++ b/projects/plugins/jetpack/changelog/enhance-block-themes-hide-customizer
@@ -1,0 +1,4 @@
+Significance: minor
+Type: enhancement
+
+The Customize menu is removed on self hosted sites that are using a block based theme.
--- a/projects/plugins/jetpack/changelog/enhance-block-themes-hide-customizer
+++ b/projects/plugins/jetpack/changelog/enhance-block-themes-hide-customizer
@@ -1,4 +1,4 @@
 Significance: minor
 Type: enhancement
 
-The Customize menu is removed on self hosted sites that are using a block based theme.
+Removed the "Appearance > Customize" entry from the WordPress.com menu when the site uses a block-based theme.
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/changelog/enhance-block-themes-hide-customizer
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: enhancement
+
+The Customize menu is removed on self hosted sites that are using a block based theme.
```
mmtr: Nitpick: This message will be visible in the public changelog of the next Jetpack release. Since this change will affect Jetpack users, it'd be good to use a more clear language since they might don't know what a "self hosted site" is. I think we should also specify that this change only affects the WordPress.com menu (and not the WP Admin menu).

Maybe something like this?

```suggestion
Removed the "Appearance > Customize" entry from the WordPress.com menu when the site uses a block-based theme.
```

projects/plugins/jetpack/modules/masterbar/admin-menu/class-jetpack-admin-menu.php
```diff
@@ -205,6 +205,13 @@ public function add_appearance_menu() {
 
 		add_menu_page( esc_attr__( 'Appearance', 'jetpack' ), __( 'Appearance', 'jetpack' ), 'switch_themes', $themes_url, null, 'dashicons-admin-appearance', 60 );
 		add_submenu_page( $themes_url, esc_attr__( 'Themes', 'jetpack' ), __( 'Themes', 'jetpack' ), 'switch_themes', 'https://wordpress.com/themes/' . $this->domain );
+
+		if ( function_exists( 'wp_is_block_theme' ) && wp_is_block_theme() ) {
+			if ( class_exists( 'Jetpack_Custom_CSS' ) && empty( Jetpack_Custom_CSS::get_css() ) ) {
```
mmtr: The Additional CSS submenu is never exposed to Jetpack sites in Calypso, so I think we can always remove the Customize menu regardless of whether there is custom CSS or not.

mmtr (commented):
The code looks correct, but I cannot get this to work. The Customize menu is still there for me:

Screenshot 2023-06-15 at 09 55 17

I double checked Jetpack Beta has activated the branch of this PR, the site uses a block theme, and that there is no custom CSS.

projects/plugins/jetpack/changelog/enhance-block-themes-hide-customizer
rcrdortiz: Sure

projects/plugins/jetpack/modules/masterbar/admin-menu/class-jetpack-admin-menu.php
rcrdortiz: Done

mmtr (approved):
Works great now!

PR #31378, opened 2023-06-14
AI Feature: update block content once AI response is ready

### LINKED ISSUE #31362: AI Feature: generate content from AI Assistant menu
Let's start to generate content from the AI Assistant toolbar.

retrofox:
This PR implements the part that allows requesting and updating the block text content by hitting the AI endpoint.

https://github.com/Automattic/jetpack/assets/77539/4d29c63b-28f4-41e0-8835-015a80fd7745

### Significant changes

* **It introduces the new `useSuggestionsFromAI()` custom hook.** 
  Using the current `useSuggestionsFromOpenAI()` hook gets tough. For instance, it updates the block content by setting the block attributes.
That makes the ability to handle other blocks tricky to scale. Ideally, from my PoV, the hook should take over the data layer, meaning that it should accept the request data (prompt, for instance) and return only the generated suggestion.
Everything else should be handled out of the hook.

* It plugs the `AiAssistantDropdown` dropdown (UI) with the data layer via the mentioned hook
* Multiple selection edition is impleemnted
* Edit selected is not implemented.

### Disclaimer

* This implementation relies on the `done` event to update the block content. We do want to see the `suggestion` event instead, but it will be addressed in a follow-up.
* Error handling is not implemented
* Checking the product feature availability is not implanted.

Fixes https://github.com/Automattic/jetpack/issues/31362

## Proposed changes:

* AI Feature: update block content once AI response is ready

### Other information:

## Testing instructions:

* Go to the block editor
* Select a paraphrase/heading block
* Change the content by using the AI Assistant dropdown
* Wait...
* Confirm the content is updated

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
@@ -3,7 +3,7 @@
  */
 import { InspectorControls, BlockControls } from '@wordpress/block-editor';
 import { createHigherOrderComponent } from '@wordpress/compose';
-import { useCallback } from '@wordpress/element';
+import { useCallback, useState } from '@wordpress/element';
 import React from 'react';
 /**
  * Internal dependencies
@@ -13,7 +13,8 @@ import AiAssistantDropdown, {
 	AiAssistantSuggestionProp,
 } from '../../components/ai-assistant-controls';
 import AiAssistantPanel from '../../components/ai-assistant-panel';
-import { buildPrompt } from '../../lib/prompt';
+import useSuggestionsFromAI from '../../hooks/use-ai-suggestions-from-ai';
+import { PromptItemProps, buildPrompt } from '../../lib/prompt';
 
 /*
  * Extend the withAIAssistant function of the block
@@ -21,18 +22,49 @@ import { buildPrompt } from '../../lib/prompt';
  */
 export const withAIAssistant = createHigherOrderComponent(
 	BlockEdit => props => {
-		const content = props?.attributes?.content;
+		const { setAttributes } = props;
+		const [ prompt, setPrompt ] = useState< Array< PromptItemProps > >( [] );
+
+		/*
+		 * Pick the content from the block attribute from now.
+		 * @todo: it doesn't scale well, we need to find a better way to get the content.
+		 */
+		const content: string = props?.attributes?.content;
+
+		/**
+		 * Set the content of the block.
+		 *
+		 * @param {string} newContent - The new content of the block.
+		 * @returns {void}
+		 */
+		const setContent = useCallback(
+			( newContent: string ) => {
+				/*
+				 * Update the content of the block
+				 * by calling the setAttributes function,
+				 * updating the `content` attribute.
+				 * It doesn't scale for other blocks.
+				 * @todo: find a better way to update the content.
+				 */
+				setAttributes( { content: newContent } );
+			},
+			[ setAttributes ]
+		);
+
+		useSuggestionsFromAI( { content, prompt, onDone: setContent } );
+
 		const requestSuggestion = useCallback(
 			(
 				suggestion: AiAssistantSuggestionProp,
 				options: AiAssistantDropdownOnChangeOptionsArgProps
 			) => {
-				// eslint-disable-next-line @typescript-eslint/no-unused-vars
-				const prompt = buildPrompt( {
-					type: suggestion,
-					generatedContent: content,
-					options,
-				} );
+				setPrompt(
+					buildPrompt( {
+						type: suggestion,
+						generatedContent: content,
+						options,
+					} )
+				);
 			},
 			[ content ]
 		);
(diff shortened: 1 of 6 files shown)


projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-ai-suggestions-from-ai/index.ts
lhkowalski, line 1: Maybe we should call the directory `use-suggestions-from-ai`, dropping the extra `ai` after `use`.

projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-ai-suggestions-from-ai/index.ts
```diff
@@ -0,0 +1,152 @@
...
+	// Request suggestions automatically when ready.
+	useEffect( () => {
+		if ( ! readyToRequest ) {
+			return;
+		}
+
+		// Trigger the request.
+		request();
+
+		// Close the conenction when unmounting.
```
lhkowalski:
```suggestion
		// Close the connection when unmounting.
```

projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-ai-suggestions-from-ai/index.ts
```diff
@@ -0,0 +1,152 @@
...
+import { useSelect } from '@wordpress/data';
+import { store as editorStore } from '@wordpress/editor';
+import { useCallback, useEffect, useRef } from '@wordpress/element';
+/**
+ * Types
+ */
+import { SuggestionsEventSource, askQuestion } from '../../lib/suggestions';
+import type { PromptItemProps } from '../../lib/prompt';
+
+type useSuggestionsFromAIOptions = {
```
lhkowalski: I thinks it's neater to have types with UpperCamelCase:

```suggestion
type UseSuggestionsFromAIOptions = {
```

Same for the other new types.

projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-ai-suggestions-from-ai/index.ts
retrofox: yes, 100% agree. updating...

projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
```diff
@@ -13,26 +13,58 @@ import AiAssistantDropdown, {
...
  */
 export const withAIAssistant = createHigherOrderComponent(
 	BlockEdit => props => {
-		const content = props?.attributes?.content;
+		const { setAttributes } = props;
+		const [ prompt, setPrompt ] = useState< Array< PromptItemProps > >( [] );
+
+		/*
+		 * Pick the content from the block attribute from now.
+		 * @todo: it doesn't scale well, we need to find a better way to get the content.
```
dhasilva, line 30: If we send the block's code this should fix itself, I hope. Maybe we should try to start working on the Gutenberg syntax this week, at least some research and initial tests

projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-suggestions-from-ai/index.ts
```diff
@@ -0,0 +1,152 @@
...
+	// Collect data
+	const { postId, postTitle } = useSelect( select => {
+		return {
+			postId: select( editorStore ).getCurrentPostId(),
+			postTitle: select( editorStore ).getEditedPostAttribute( 'title' ),
+		};
+	}, [] );
+
+	// Store the event source in a ref, so we can handle it if needed.
+	const source = useRef< SuggestionsEventSource | undefined >( undefined );
```
dhasilva, line 84: Minor nitpick: Why `undefined` instead of `null`? Assigning something to `undefined` seems wrong to me.

projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
retrofox: yes, that's the next action to take in favor of Gutenberg syntax IMO.
The most important part, today, is to be aware that we'll have to handle other blocks depending on their anatomy.
Now, we rely on the block attribute.

projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-suggestions-from-ai/index.ts
retrofox: why? 😅 To me `null` is almost the same that `undefined`, but with the huge difference that React checks the `null` value when the component shouldn't render anything.

lhkowalski (approved):
My first attempt was using the Simplify prompt, but it's actually broken. Notice that the context was not sent, so the assitant was lost:

Screen Shot 2023-06-14 at 12 02 42

It's probably related to what prompt type we should use, and how the prompt is being built, not necessarily related to your change.

I also got this `Stream closed unexpectedly` debug message, that may be related to your implementation.

---

I tried using other actions and they worked:

Screen Shot 2023-06-14 at 12 03 34

I got the `Stream closed unexpectedly` debug message again.

---

Overall, the implementation works good. Having the content of the paragraph was nice because CMD+Z works on it :D

We probably we need to iterate over it a bit, since the new hook was not battle-tested yet, but I think we are in a good direction.

retrofox:
> I got the `Stream closed unexpectedly` debug message again.

Yes, that's something we should improve. I'd like to stop the connection when the block is removed, for instance.

dhasilva (approved):
LGTM

There is an issue with setting the right content, let's leave it for a follow-up

PR #31385, opened 2023-06-14
Tock Block: Hardcode the HTML code for PHP rendering

pablinos:
On WPCOM we strip data attributes to prevent XSS vulnerabilities from unsanitized user data being injected in to the HTML.

For the Tock block to function properly it needs these attrbutes, so we can hardcode the HTML for the embed in the PHP render_block function.

### Other information:

## Testing instructions:

* Insert a Tock block in a post on a WordPress.com simple site
* Publish and view the post. Notice that the button isn't styled correctly
* Check out this PR and sandbox the site
* Reload the post and check that it fixes the issue

#### Before
image

#### After
image


samiff (approved):
Tested well 👍

PR #31386, opened 2023-06-14
Jetpack: Fix newsletter editor sidebar panel in WP 6.1

anomiex:
## Proposed changes:

PR #30958 added some code to return early from the sidebar panels when the subscriptions module is disabled. Unfortunately it did this in a way that only works with React 18, and WordPress 6.1 (which we still support) comes with React 17.

Fixing this for React 17 is simple enough, we just have to return `null` instead of `undefined` in the early return.

### Other information:

## Jetpack product discussion

p1686754280648659-slack-CDD9LQRSN
p1686759008790089-slack-C01U2KGS2PQ

No

## Testing instructions:

* Create a testing site with Jetpack trunk, WordPress 6.1.3, and without the Gutenberg plugin. Open the post editor and verify the presence of red error bars complaining about "The "jetpack-subscriptions" plugin has encountered an error and cannot be rendered".
* Update Jetpack to this PR. Verify the red error bars are no longer present.

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-newsletter-panel-with-react-17
+++ b/projects/plugins/jetpack/changelog/fix-newsletter-panel-with-react-17
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
+Newsletters: Fix sidebar panel in WordPress 6.1.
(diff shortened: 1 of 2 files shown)


jeherve (approved):
This fixes it. 👍

samiff:
cherry-picked in 164b2260900743765a0942fa28b04e2cc8955876 for Jetpack 12.2.1 release

PR #31387, opened 2023-06-14
AI Feature: fix generating prompt when requesting suggestion

retrofox:
The current `buildPrompt()` doesn't scale very well with the AI Feature requirements. It expects some parameters that may not make sense when trying to build the prompt. This is why I've created a getPrompt() helper that is a wrapper of `buildPrompt()`.
It isn't a flawless implementation but makes the implementation clearer. Also, it opens the door to continue iterating over it.
Almost all changes are part of consts and types reorganization.

Fixes #

## Proposed changes:

* AI Feature: fix generating prompt when requesting suggestion

### Other information:

## Testing instructions:

* Go to the block editor
* Create paragraph/heading block
* Change the content by using the AI Assistant dropdown
* Check the prompt
* Confirm the generated code is expected 
* Don't forget to test the AI Assistant block. It should work as expected too.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
@@ -10,11 +10,10 @@ import React from 'react';
  */
 import AiAssistantDropdown, {
 	AiAssistantDropdownOnChangeOptionsArgProps,
-	AiAssistantSuggestionProp,
 } from '../../components/ai-assistant-controls';
 import AiAssistantPanel from '../../components/ai-assistant-panel';
 import useSuggestionsFromAI from '../../hooks/use-suggestions-from-ai';
-import { PromptItemProps, buildPrompt } from '../../lib/prompt';
+import { PromptItemProps, PromptTypeProp, buildPrompt } from '../../lib/prompt';
 
 /*
  * Extend the withAIAssistant function of the block
@@ -54,10 +53,7 @@ export const withAIAssistant = createHigherOrderComponent(
 		useSuggestionsFromAI( { content, prompt, onDone: setContent } );
 
 		const requestSuggestion = useCallback(
-			(
-				suggestion: AiAssistantSuggestionProp,
-				options: AiAssistantDropdownOnChangeOptionsArgProps
-			) => {
+			( suggestion: PromptTypeProp, options: AiAssistantDropdownOnChangeOptionsArgProps ) => {
 				setPrompt(
 					buildPrompt( {
 						type: suggestion,
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
@@ -53,7 +53,7 @@ export const withAIAssistant = createHigherOrderComponent(
 		useSuggestionsFromAI( { content, prompt, onDone: setContent } );
 
 		const requestSuggestion = useCallback(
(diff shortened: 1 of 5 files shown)


projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
```diff
@@ -54,15 +53,11 @@ export const withAIAssistant = createHigherOrderComponent(
 		useSuggestionsFromAI( { content, prompt, onDone: setContent } );
 
 		const requestSuggestion = useCallback(
-			(
-				suggestion: AiAssistantSuggestionProp,
-				options: AiAssistantDropdownOnChangeOptionsArgProps
-			) => {
+			( promptTupe: PromptTypeProp, options: AiAssistantDropdownOnChangeOptionsArgProps ) => {
```
lhkowalski:
```suggestion
			( promptType: PromptTypeProp, options: AiAssistantDropdownOnChangeOptionsArgProps ) => {
```

projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
```diff
@@ -54,15 +53,11 @@ export const withAIAssistant = createHigherOrderComponent(
...
 				setPrompt(
-					buildPrompt( {
-						type: suggestion,
-						generatedContent: content,
-						options,
+					getPrompt( promptTupe, {
```
lhkowalski:
```suggestion
					getPrompt( promptType, {
```

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
```diff
@@ -314,3 +334,28 @@ export function buildPrompt( {
 
 	return prompt;
 }
+
+type GetPromptOptionsProps = {
+	content: string;
+};
+
+/**
+ * Retunr a prompt based on the type and options
```
lhkowalski:
```suggestion
 * Returns a prompt based on the type and options
```

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
```diff
@@ -314,3 +334,28 @@ export function buildPrompt( {
...
+ * @returns {Array< PromptItemProps >}      The prompt.
+ */
+export function getPrompt(
+	type: PromptTypeProp,
+	options: GetPromptOptionsProps
+): Array< PromptItemProps > {
+	return buildPrompt( {
+		type,
+		generatedContent: options.content,
+		postContentAbove: options.content,
```
lhkowalski, line 356: Maybe we need two different option properties for "content" and "content above". I see this being the same for some prompts, but if we plan to use `getPrompt` everywhere, we will need both options, right?

lhkowalski:
Code looks good, just some typos. Going to test the change now.

lhkowalski:
When changing the tone of a paragraph, I get a `Rewrite your last answer with a undefined tone` prompt:

Screen Shot 2023-06-14 at 15 13 21

---

It's working correctly for the AI Assistant change tone option, though.

lhkowalski:
It's now working:

Screen Shot 2023-06-14 at 15 36 03

lhkowalski (approved):
Nice improvements!

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
retrofox: we don't need both options from the PoV of the getPrompts() function. However, and depending on the prompt type, it uses `generatedContent` or `postContentAbove`.
I'm afraid it will require a bigger refactoring.

PR #31392, opened 2023-06-14
[Paid Newsletters] Add the "Lower transaction fees" Post Publish Panel

### LINKED ISSUE #31302: [Paid Newsletters] Add "Upgrade Plan" upsell to PostPublishPanel
#### Tracking 
Add "Upgrade Plan" tracking
(leftover from https://github.com/Automattic/jetpack/issues/31287)

#### Design
Screenshot 2023-06-09 at 3 47 17 PM

#### Edit

**_[edit]_** Text was slightly changed (see [this comment](https://github.com/Automattic/jetpack/issues/31302#issuecomment-1607268654))

romarioraffington:
Fixes #31302

## Proposed changes:

Implement the design in https://github.com/Automattic/jetpack/issues/31302

### Other information:

https://github.com/Automattic/jetpack/assets/7076981/7839ecfc-ac89-40f0-9fb5-378f8cf0a6c0

No

## Testing instructions:

1. Publish a newsletter post on WPCOM
2. Notice the post-publish panel titled "Lower transaction fees" is displayed when the WPCOM's commission is more than `0`
3. Publish a newsletter post on a Jetpack site
4. Notice the post-publish panel titled "Lower transaction fees" is never displayed


projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
```diff
@@ -285,6 +285,36 @@ function NewsletterPostPublishSettingsPanel( {
...
+			<PluginPostPublishPanel
+				initialOpen
+				className="paid-newsletters-post-publish-panel"
+				title={ __( 'Lower transaction fees', 'jetpack' ) }
+				icon={ <JetpackLogo showText={ false } height={ 16 } logoColor="#1E1E1E" /> }
+			>
+				<PanelRow>
+					<p>
+						{ __(
+							'With your current plan, the transaction fee for payments is XX% (+ Stripe fees). Upgrade to {another plan} to lower it.',
```
romarioraffington: There is some complexity here.

1. Could be a WPCOM site
2. Could be a Jetpack site
3. The commission and the next plan to upgrade to depends on 1 and 2 above.
4. The upgrade link would be different based on the type of site and plan

@millerf / @simison  I wanted to highlight the complexity. Completing this may take some time due to the edge cases. 

If any knows how to do the following, it would greatly help:

1. Determining the current plan
2. Determining the next plan to upgrade to
3. Determining the upgrade link

I will continue to investigate

millerf: I'll leave it to @crisbusquets and @simison to define the task better....

crisbusquets: Hi @romarioraffington and @millerf !

From a product/design point of view, this new module in the post-publish panel exists because it can help us get people to upgrade their plan in order to lower their fees.

The `Upgrade plan` button can open https://wordpress.com/plans/. It would be great if the post-panel itself tells them what plan they should upgrade to, so once they arrive at the `Plans` page, they'll know what to choose.

If we can't provide this information, it will be confusing to simply redirect them to https://wordpress.com/plans/, because there they won't find any information regarding fees :(

As for the technical questions, I can't answer them 😅

millerf: I think @simison worked on an endpoint to retrieve the commissions [here](https://github.com/Automattic/wp-calypso/blob/fb1d49bc854b8d0fca0e10265d2d0debd71df047/client/my-sites/earn/home.tsx#L548). More info in the "Earnings" row of : p81Rsd-1kp-p2



You should be able to use that endpoint.

romarioraffington: Thanks for the feedback everyone. @millerf , this panel will be rendered on JP sites as well. I'll take a look at the "Earnings" row but IIRC that was specific to Calypso and WPCOM. I'll double-check if the endpoint also takes into consideration JP sites.

> The Upgrade plan button can open https://wordpress.com/plans/. It would be great if the post-panel itself tells them what plan they should upgrade to, so once they arrive at the Plans page, they'll know what to choose.

I think this works for WPCOM sites but may be confusing for Jetpack users

millerf: >  I'll double-check if the endpoint also takes into consideration JP sites.

Yes it does. I checked on Calypso.

millerf: But I am not sure if we can access it without forwarding it through JP or something. Did you try?

simison: - Commission fee is available in the API, we can pull that in. You can start from WP sites and extend to JP separately if that helps.
- When commission fee is 0%, don't show upgrade nudge at all.
- You can leave figuring out the exact plan for another PR; you can simplify _"Upgrade to {another plan} to lower it."_ to _"Upgrade to lower it."_ meanwhile. We should share implementation for plan recommendation [with Earn page.](https://github.com/Automattic/wp-calypso/pull/77748#discussion_r1214371005)
- The plans grid currently doesn't have commission fees, but that'll change in https://github.com/Automattic/wp-calypso/pull/77953
- Jetpack plans grid doesn't have commission fees either, so until that's sorted I'd actually just leave the nudge out from JP sites? Otherwise the flow from nudge to plans feels broken for JP sites. See pbNhbs-6Ju-p2/#comment-15788 for adding them.

romarioraffington: > You can start from WP sites and extend to JP separately if that helps. 
>
> Jetpack plans grid doesn't have commission fees either, so until that's sorted I'd actually just leave the nudge out from JP sites? 
>
> You can leave figuring out the exact plan for another PR; you can simplify "Upgrade to {another plan} to lower it." to "Upgrade to lower it." meanwhile.  

Sounds good to me!

I created a new task to track the JP support. See https://github.com/Automattic/jetpack/issues/31424. 

To further expound on my previous points, my understanding is that JP users would be redirected to upgrade via a site similar to https://cloud.jetpack.com/features/comparison while a WPCOM user would be redirected to `https://wordpress.com/plans/[site].` JP plans don't match perfectly with a WPCOM plan, and I suspect some logic would have to be created ( [I suspect this already exists](https://github.com/Automattic/jetpack/blob/ecb8f7d24fa4dde0831f84b0f4d7b6af9a086c72/projects/js-packages/shared-extension-utils/src/plan-utils.js#L20) ) to figure out the next plan a user should be upgraded to, based on the type of site ( JP or WPCOM  ) and the current plan. I wasn't sure if this was a consideration when this design was created, and I wanted to highlight the scope before diving in.


> Jetpack plans grid doesn't have commission fees either, so until that's sorted I'd actually just leave the nudge out from JP sites? Otherwise the flow from nudge to plans feels broken for JP sites. See pbNhbs-6Ju-p2/#comment-15788 for adding them.

Sounds good to me.

> We should share implementation for plan recommendation [with Earn page.](https://github.com/Automattic/wp-calypso/pull/77748#discussion_r1214371005)
>
> The plans grid currently doesn't have commission fees, but that'll change in https://github.com/Automattic/wp-calypso/pull/77953

💯 .  The "Upgrade plan" in this PR also links to  `/plans/{siteSlug}`. Based on the comment [here](https://github.com/Automattic/wp-calypso/pull/77748#discussion_r1214371005), should I wait for the Earn page to be updated before merging this?

projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
```diff
@@ -189,6 +190,7 @@ function NewsletterPostPublishSettingsPanel( {
 	isModuleActive,
 	showMisconfigurationWarning,
 } ) {
+	const isJetpackSite = ! window.wpcomFetch;
```
jeherve, line 193: Maybe we can use `isSimpleSite` and `isAtomicSite` from `@automattic/jetpack-shared-extension-utils`, to remain consistent with other parts of the block bundle?

projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
```diff
@@ -285,6 +287,13 @@ function NewsletterPostPublishSettingsPanel( {
 					</div>
 				</PluginPostPublishPanel>
 			) }
+
+			{ /*
+			     Only show the following panel on WPCOM sites since the
+			     Fees and "Upgrade plan" links displayed are WPCOM specific.
```
jeherve, line 293: Could we do something similar for Jetpack sites? afaik, the process is the same there, paid plans give you access to different commission fees?

projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
```diff
@@ -285,6 +287,13 @@ function NewsletterPostPublishSettingsPanel( {
...
+			  */ }
+
+			{ ! isJetpackSite && <NewsletterPostPublishPanelUpgradeNude /> }
```
jeherve, line 296: Do we need to display it when someone hasn't restricted the post to subscribers and / or hasn't even set things up for payments? This way they wouldn't get confused by a message that doesn't apply to them.

projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
```diff
@@ -352,3 +365,58 @@ export function NewsletterAccessPrePublishSettings( {
...
+			className="paid-newsletters-post-publish-panel"
+			title={ __( 'Lower transaction fees', 'jetpack' ) }
+			icon={ <JetpackLogo showText={ false } height={ 16 } logoColor="#1E1E1E" /> }
+		>
+			<PanelRow>
+				<p>
+					{ sprintf(
+						/* translators: %s is the commission based on the current WordPress.com plan */
+						__(
+							'With your current plan, the transaction fee for payments is %s (+ Stripe fees). Upgrade to to lower it.',
```
jeherve, line 401: Would it make sense to have a link to an explanation somewhere in there, showing what this commission fee means, where it applies. Since that's a new panel, it's not immediately clear to me that this is about paid newsletters.

jeherve (changes requested):

projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
romarioraffington: We could! I had a few concerns here that I've yet to look into:

1. Would the memberships/earnings endpoint return the correct commission for JP sites? 
2. What would be the upgrade URL for JP sites?

projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
romarioraffington: Great point, Jeremy. I'll only show this panel when:

1. The Paid subscribers option is selected
2. Stripe is connected
3. Newsletter paid plan has been created

projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
romarioraffington: Great point!

@crisbusquets  and @simison any thoughts on this?

romarioraffington:
Putting this on hold as requested by Fab. 

See p1687187852858469/1687187149.097789-slack-C01B6KEJ5GE

projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
crisbusquets: > Would it make sense to have a link to an explanation somewhere in there, showing what this commission fee means, where it applies. Since that's a new panel, it's not immediately clear to me that this is about paid newsletters.

Hi!

We currently have a similar text on the Earn page, and the `Stripe fees` link is: https://stripe.com/en-gb-es/pricing.

Question: Does the URL depend on the country or browser language (I'm in Spain, but the browser is in English)?

Screenshot 2023-06-26 at 09 43 10

projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
millerf: Just noting there that checking for "The Paid subscribers option is selected" is enough, as it can only be selected when 2. and 3. are already true

projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
millerf: @crisbusquets Any strong feeling about changing the text from the design:


Screenshot 2023-06-26 at 09 53 32


To something closer to the one on Earn page (basically removing the need for "another plan" that could be tricky...

crisbusquets: > @crisbusquets Any strong feeling about changing the text from the design:
> 
> Screenshot 2023-06-26 at 09 53 32
> To something closer to the one on Earn page (basically removing the need for "another plan" that could be tricky...

No strong feeling at all :) Both can work. Thanks for asking!

millerf: Ok, adapting the scope of the task, then.

PR #30970, opened 2023-05-25
DRAFT: Jetpack-mu-com: Add newsletter subscribe modal

edanzer:
Addresss 2236-gh-Automattic/dotcom-forge

## Proposed changes:

This PR adds a subscribe popup on the frontend of newsletter sites. The code is added via the jetpack-mu-wpcom package/plugin. 

This is still a draft. I'll fill out the PR description more when it is ready for testing.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* Go to '..'
*

Issue #30578, opened 2023-05-10
Subscribe block: improve confirmation


simison:
Via p9Jlb4-787-p2

@crisbusquets wrote:

> The alerts are not distinct enough, as they appear as a regular paragraph. I think it would be better to implement something like this
>
> ![Image](https://github.com/Automattic/jetpack/assets/87168/0bfb6772-0fed-4d24-b20e-4de516f2489c)

Jeremy shared:

> A visual checkmark would be good, with some sort of smaller and simpler text message (maybe in a colored box?) like “You’re all set! Check your inbox to confirm your subcription.” And without the page refresh would be great.

See further design convo p9Jlb4-787-p2#comment-7736 (Todo: copy it to here without internal references)

We could use [Interactivity API](https://make.wordpress.org/core/2023/03/30/proposal-the-interactivity-api-a-better-developer-experience-in-building-interactive-blocks/) to make the block more dynamic.

PR #31393, opened 2023-06-15
Newsletters: Add subscriber modal

edanzer:
Addresses: 2236-gh-Automattic/dotcom-forge
Replaces this PR: https://github.com/Automattic/jetpack/pull/30970

## Proposed changes:

**Important for any testers: We've now put this behind a feature flag using the jetpack_subscriptions_modal_enabled filter. The most important thing to review and/or test is that the feature flag is correctly set to false, and will prevent the modal or modal related features or settings from loading when set to false.**

* Add a subscribe popup modal to frontend that allows readers to subscribe. This works by adding a Subscribe Modal template part to the site. This template part should only be created once. This template part contains the Subscriptions Block, and we load the entire template part inside the popup modal. 
* Add an option to enable the modal, and add the associated toggle on settings screen.
* Add the same new option to the settings endpoint - this change to the settings endpoint is not used in this PR, but will be used to add a separate option in calypso later for WP.com users.

**Limited initial exposure.**
As currently written the subscribe modal in **this feature should have very limited initial exposure**. 
   - First, even to see the option to turn on this feature, a site must have newsletter as site_intent (only done on [WP.com](http://wp.com/)) or have the Lettre theme active, and must have a block theme enabled (which likely excludes the majority of self-hosted jetpack sites). This is all just to see the option. 
   - Second, for the feature to become active, users must find the setting and actively enable the feature. And the setting for this is pretty hard to find, for better or worse.
   - In short, most users won't see the setting, and only those who enable the setting will see the modal. We should not have a situation where lots of users suddenly start seeing subscribe modals.
   - **If you are reviewing this PR, may be put special attention on confirm that I'm right and have correctly implemented all the checks above**. As long as these are all true, we have a lot of leeway and breathing space with this feature. 

### Outstanding questions
- We need to update the copy on the modal. It currently suggests it is only for protected posts. This needs design input.
- This PR is already large, and I'd like to avoid also updating the success behavior/notification when someone subscribes. But that part of the experience is currently not fluid. We could leave that as is, add it here, or update that via a separate PR before deploying this one. 
- This isn't working on Simple sites yet - I hadn't realized that we hide jetpack settings on simple sites. So still discussing solutions to that. 
- I have not deployed code before to the core jetpack plugin, so I'll welcome input on any aspect of this PR from seasoned Jetpack contributors. 
- I have not added tests. If possible, I'd prefer to add/update tests for this in a follow, again just given how big the PR is already.

**Modal looks like this:**

Screenshot on 2023-06-20 at 11-16-42

**New option to enable the modal looks like this:**

Markup 2023-06-15 at 12 21 43

### Other information:

## Jetpack product discussion

See 2236-gh-Automattic/dotcom-forge

Possibly? We are allowing our users to collect email addresses from their readers.

## Testing instructions:

**Test on Simple Sites**
1. **Setup: Test Site**. You will need a test site that either has newsletter as the site_intent, or has the Lettre theme active. You can create one by starting at https://wordpress.com/setup/newsletter/intro if needed.
2. **Setup: Backend diff**. You'll need checkout this diff in your sandbox. It includes some extra code needed to enable Jetpack subscription functionality on WordPress.com: D114167-code
3. **Setup: Apply Jetpack PR**. In your sandbox run `bin/jetpack-downloader test jetpack add/subscriber-modal` to load this brand in your sandbox. Be sure to keep the backend diff from step 2 in place.
4. **Setup: Sandbox**. Be sure to sandbox public-api and the domain of your test site. 
5. **Setup: Add Filter**. As a form of feature flag, we've disabled this entire feature using a filter. To enable, go into your sandbox mu-plugins, and add the following line to your 0-sandbox.php file: 
   - `add_filter( 'jetpack_subscriptions_modal_enabled', '__return_true', 20 );`
7. **Setup: Calypso Dev + turn on option.** You need to enable the modal via a setting in Calypso. We've merged the PR for this, but that is also behind a feature flag. Checkout the trunk of calypso, run yarn start, go to the your site's admin are using http://calypso.localhost:3000/, navigate to Settings > Reading, scroll down to the Newletter settings, and enable the toggle for the subscriber modal.
8. **Setup: Add single post.** The modal will only show on a single post. If your test site does not have one, create one. 
9. **Test**: Go to front end of your website, go to a single post, and scroll. Shortly after you stop scrolling, you should see the subscribe modal popup. Confirm the modal loads and looks like the screenshot above. 
10. **Test**: Confirm that once you close the modal, it no longer loads (note: it will load again if you reload the page). 
11. **Test**: Enter an email and confirm you are subscribed, and the the behavior of the success notice is sane. Note: this is already a large PR, and I would prefer not to also update the success notification behavior of the subscribe modal. 

**Test on Atomic Site**
1. **Setup: WoA site**. You will need a WoA site. This site will need to either have newsletter as site_intent, or have the Lettre theme active.
2. **Setup: Jetpack beta tester**. Install and activate [the Jetpack beta plugin](https://jetpack.com/download-jetpack-beta/) on your site. Go to Jetpack > Jetpack Beta > Jetpack and enable the add/subscriber-modal branch.
3. **Setup: Add Filter**. As a form of feature flag, we've put this feature behind a filter. For the feature to show, you will need to use add_filter with the 'jetpack_subscriptions_modal_enabled' to return true. 
4. **Test**: Login to your site and go to Jetpack settings > Discussion. Enable Subscriptions and enable the Subscribe modal. Confirm that this toggle works, and that the update value persists (ie, try refreshing the page). 
5. **Test**: Go to front end of your website, and scroll. Shortly after you stop scrolling, you should see the subscribe modal popup. Confirm the modal loads and looks like the screenshot above. 
6. **Test**: Confirm that once you close the modal, it no longer loads (note: it will load again if you reload the page). 
7. **Test**: Enter an email and confirm you are subscribed, and the the behavior of the success notice is sane. Note: this is already a large PR, and I would prefer not to also update the success notification behavior of the subscribe modal. 

**Test on self-hosted Jetpack site (ie, just use your existing jetpack dev environment)**
1. **Setup: Normal Jetpack dev environment**. Use your normal Jetpack dev environment. Checkout this branch and run any needed build commands to build jetpack locally.
2. **Setup: Jurassic Tube.** This functionality requires a connected site, so you'll need to set up a jurassic tube site and 
3. **Setup: Lettre theme.** Activate the Lettre theme on your dev site. 
4. **Setup: Add Filter**. As a form of feature flag, we've put this feature behind a filter. For the feature to show, you will need to use add_filter with the 'jetpack_subscriptions_modal_enabled' to return true. 
5. **Test**: Login to your site and go to Jetpack settings > Discussion. Enable Subscriptions and enable the Subscribe modal. Confirm that this toggle works, and that the update value persists (ie, try refreshing the page). 
6. **Test**: Go to front end of your website, and scroll. Shortly after you stop scrolling, you should see the subscribe modal popup. Confirm the modal loads and looks like the screenshot above. 
7. **Test**: Confirm that once you close the modal, it no longer loads (note: it will load again if you reload the page). 
8. **Test**: Enter an email and confirm you are subscribed, and the the behavior of the success notice is sane. Note: this is already a large PR, and I would prefer not to also update the success notification behavior of the subscribe modal.


projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
```diff
@@ -0,0 +1,196 @@
+<?php
```
edanzer: I should probably namespace and/or move code in this file into a class. At the very least, I will remove the wpcom prefixes.

crisbusquets:
Hi @edanzer !

Great work, and thanks for sharing the video in Slack too :)

**Modal behavior**
Would it be possible that the modal only appears in posts? Now that I see it, I find it odd that it's on the homepage, where there are already two subscriber blocks. 

**Copy**
The copy will be ok if we can change this behavior. If it's not possible, I'd change the copy to:

Heading
`Discover more from {SiteName}`

Text
`Subscribe to the newsletter to keep reading and get access to the full archive.`

Link to close (in any copywriting option we use)
`Continue reading`

[Figma file](h8tMpJlCqNFemEerU9owHI-fi-522_9013) and screenshot: 

Screenshot 2023-06-16 at 09 31 42


**Success screen**
I'd use this, it's not ideal but it might work for now: https://github.com/Automattic/jetpack/issues/30578

[Figma file](h8tMpJlCqNFemEerU9owHI-fi-522_9013) and screenshot:

Screenshot 2023-06-16 at 09 31 26


**Questions**
- Is it possible to close de modal when clicking outside it?
- The modal uses the typography on the design, not the one that the user has in the theme, right?

🚀

simison:
> The modal uses the typography on the design, not the one that the user has in the theme, right?

That's an interesting one; the subscribe block does inherit theme styles, why not do the same with paragraph and heading?

projects/plugins/jetpack/_inc/client/discussion/subscriptions.jsx
```diff
@@ -134,6 +139,17 @@ class SubscriptionsComponent extends React.Component {
...
+							<ToggleControl
+								checked={ isSubscriptionsActive && this.props.getOptionValue( 'sm_enabled' ) }
+								disabled={
+									! isSubscriptionsActive ||
+									unavailableInOfflineMode ||
+									this.props.isSavingAnyOption( [ 'subscriptions' ] )
+								}
+								toggling={ this.props.isSavingAnyOption( [ 'sm_enabled' ] ) }
+								onChange={ this.handleSubscribeModalToggleChange }
+								label={ __( 'Enable the popup Subscribe Modal', 'jetpack' ) }
```
simison: Copy feels a bit clunky, but I don't have alternatives to suggest now. Non blocking merge.

projects/plugins/jetpack/_inc/client/discussion/subscriptions.jsx
```diff
@@ -134,6 +139,17 @@ class SubscriptionsComponent extends React.Component {
...
+							/>
```
simison: Can we have a link to edit the modal in the focused template part editor somewhere near this toggle? Ok in follow-up PR as well.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.css
```diff
@@ -0,0 +1,45 @@
+#wpcom-subscribe-modal {
```
simison: Nit; can be `jetpack-` prefix but no biggie if wpcom.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.css
```diff
@@ -0,0 +1,45 @@
...
+#wpcom-subscribe-modal.open {
+	background-color: rgba(0,0,0,0.2);
+	visibility: visible;
+}
+
+#wpcom-subscribe-modal .modal-content {
+	position: relative;
+	visibility: hidden;
+    top: 100%;
+	background-color: #fefefe;
```
simison, line 23: Some mixed tabs/spaces white spacing here and elsewhere in the file.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.js
```diff
@@ -0,0 +1,27 @@
+window.onload = function () {
+	const modal = document.getElementById( 'wpcom-subscribe-modal' );
```
simison: Let's bail out early if the element isn't found for any reason.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.js
```diff
@@ -0,0 +1,27 @@
+window.onload = function () {
```
simison: There's a GB package for this: https://github.com/WordPress/gutenberg/tree/trunk/packages/dom-ready#readme

It has a bit more complete version of onload:

https://github.com/WordPress/gutenberg/blob/trunk/packages/dom-ready/src/index.js#L31-L45

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.css
```diff
@@ -0,0 +1,45 @@
...
+#wpcom-subscribe-modal.open .modal-content {
+	top: 0;
+	visibility: visible;
+}
+
+#wpcom-subscribe-modal #close {
+	text-align: center;
+}
+
+#wpcom-subscribe-modal #close a {
```
simison: We might need to get more defensive with styles here in case we want theme styles not leak in? Worth testing with dozen different themes! [See plugin](https://wordpress.org/plugins/back-to-the-theme/)

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
simison: Jetpack prefixes are prolly enough? I'm not sure how fond we're using namespaces in Jetpack. @jeherve ?

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
```diff
@@ -0,0 +1,196 @@
...
+/**
+ * Subscribe modal popup feature
+ *
+ * @package automattic/jetpack-mu-wpcom
+ * @since 2.2.1
```
simison: Package name needs updating/removing

Needs some `$` variable here that gets automatically replaced with right version later I think.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
```diff
@@ -0,0 +1,196 @@
...
+
+		?>
+			<div id="wpcom-subscribe-modal">
+				<div class="modal-content">
+					<?php
+					foreach ( $blocks as $block ) {
+						echo wp_kses( render_block( $block ), $allowed_html );
+					}
+					?>
+					<div id="close"><a href="#">Close</a></div>
```
simison: "Close" needs wrapping in translation function.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
```diff
@@ -0,0 +1,196 @@
...
+add_action( 'wp_enqueue_scripts', 'wpcom_enqueue_subscribe_modal_assets' );
+
+/**
+ * Returns true if we should enable Newsletter content.
+ * This is currently limited to lettre theme or newsletter sites.
+ * We could open it to all themes or site intents.
+ *
+ * @return bool
+ */
+function should_enable_subscriber_modal() {
```
simison: We can also limit to WP.com on first release?

Rest of the limits won't make sense for Jetpack sites, but for .com they could be reasonable start. And then release a bit wider next up?

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
```diff
@@ -0,0 +1,196 @@
...
+	return 'subscribe-modal';
+}
+
+/**
+ * Returns the title for the Subcribe template.
+ *
+ * @return string
+ */
+function get_subscribe_template_title() {
+	return 'Subscribe Modal';
```
simison: Translation function

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
```diff
@@ -0,0 +1,196 @@
...
+/**
+ * Returns the initial content of the Subscribe Modal template.
+ * This can then be edited by the user.
+ *
+ * @return string
+ */
+function get_subscribe_template_content() {
+	return '<!-- wp:group {"style":{"spacing":{"padding":{"top":"var:preset|spacing|40","right":"var:preset|spacing|40","bottom":"var:preset|spacing|40","left":"var:preset|spacing|40"}}},"layout":{"type":"constrained"}} -->
+	<div class="wp-block-group" style="padding-top:var(--wp--preset--spacing--40);padding-right:var(--wp--preset--spacing--40);padding-bottom:var(--wp--preset--spacing--40);padding-left:var(--wp--preset--spacing--40)"><!-- wp:heading {"textAlign":"center"} -->
+	<h2 class="wp-block-heading has-text-align-center">This post is for subscribers</h2>
```
simison: Translation function.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
```diff
@@ -0,0 +1,196 @@
...
+ * @since 2.2.1
+ */
+
+/**
+ * Adds a Subscribe Modal template part that can be edited in the editor.
+ * This is later loaded in a pop up modal for Newsletter sites.
+ *
+ * @return void
+ */
+function wpcom_create_subscribe_template() {
```
simison: Do we need any function checks and bail out, or everything we need is supported in WP versions that Jetpack supports? Assume Gutenberg plugin gets disabled.

simison:
> Success screen
> I'd use this, it's not ideal but it might work for now: https://github.com/Automattic/jetpack/issues/30578

This success screen update is a good separate generic update to the block as well; and I'd also consider separate PR for it.

projects/plugins/jetpack/modules/subscriptions.php
```diff
@@ -461,6 +476,29 @@ public function subscription_comment_subscribe_setting() {
...
+
+		<p class="description">
+			<input type="checkbox" name="sm_enabled" id="jetpack-subscribe-modal" value="1" <?php checked( $sm_enabled, 1 ); ?> />
+			<?php
+			echo wp_kses(
+				__(
+					'Show a popup subscribe modal to readers.',
+					'jetpack'
+				)
+			);
```
jeherve: I don't know that we need `wp_kses` here. It could just be 

```suggestion
			esc_html_e( 'Show a popup subscribe modal to readers.', 'jetpack' );
```

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.css
jeherve: I would agree with the prefix. I would also suggest, if possible, that we use a class for all those changes if possible, so site owners have it a bit easier if they want to customize the look of the pop-up, by just targeting the (more specific) ID.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
jeherve: +1. That would give us

```suggestion
 * @since $$next-version$$
```

For more information:
https://github.com/Automattic/jetpack/blob/2aac5857f66d07bfda1e08da7db299b9aaf63db9/projects/packages/README.md#package-version-annotations

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
jeherve: A namespace would be okay. We use them in a lot of places in the codebase.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
jeherve: Maybe we could add a filter in here, so folks wanting to test can use the filter to toggle on the feature?

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
```diff
@@ -0,0 +1,196 @@
...
+		$subscribe_template = reset(
+			array_filter(
+				$posts,
+				function ( $post ) {
+					return $post->post_name === get_subscribe_template_slug();
+				}
+			)
+		);
+
+		$blocks = parse_blocks( $subscribe_template->post_content );
```
jeherve: Should we bail here on sites that won't have / support template parts?

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
```diff
@@ -0,0 +1,196 @@
...
+add_action( 'wp_footer', 'wpcom_add_subscribe_modal_to_frontend' );
+
+/**
+ * Enqueues JS to load modal.
+ *
+ * @return void
+ */
+function wpcom_enqueue_subscribe_modal_assets() {
+	if ( should_enable_subscriber_modal() && ! is_admin() ) {
+		wp_enqueue_style( 'subscribe-modal-css', plugins_url( 'subscribe-modal.css', __FILE__ ), array(), JETPACK__VERSION );
```
jeherve: You could add this new stylesheet to `concatenated_style_handles` in `class.jetpack.php` so the contents of the file can be added to the concatenated `jetpack.css` that's loaded for all connected Jetpack sites.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
edanzer: Following what seems to be convention in the module folders, I did not add a namespace, but did update the file to be a class, so all method names are now encapsulated within that class.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.css
edanzer: I've updated this from id to class, and from `wpcom-` to `jetpack-`.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.css
edanzer: Fixed!

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.js
edanzer: Added this.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.js
edanzer: Added this.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
edanzer: I've updated this!

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
edanzer: Added.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
edanzer: Is there any weirdness about trying to use a translation function inside of block markup?

projects/plugins/jetpack/modules/subscriptions.php
edanzer: Yes, this is correct. A bit of copy/past going on here. Fixed.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
edanzer: Good thought. I believe all FSE themes should support template parts? Assuming yes, I've added a check with `wp_is_block_theme()` here: 

https://github.com/Automattic/jetpack/pull/31393/files#diff-bcd4c897206a1f66102ef44f0e272985d67be37c91dfacbbd6458d46d66ac72eR188

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
edanzer: I've added this. To confirm, I still enqueue the new stylesheet like normal too? Based on what I'm seeing, that seems to be case.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
edanzer: Good question. I have added a check for wp_is_block_theme() based on separate feedback below. As is, we're now checking if it's a block theme, if the modal option is turned on, and if (the site intent is newsletter or the theme is lettre). Off the top, I'd think that would be enough checks, though I'm happy to add others if we think doing so is warranted.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
edanzer: @jeherve - interesting thought. Can you clarify or maybe link to an example of how that might work in jetpack for a situation like this? Not sure if we'd do this via, say, constant in wp-config or an actual WP filter or some other mechanism?

jeherve: Basically something like this:

```diff
diff --git a/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php b/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
index 332abe2165..b77837781c 100644
--- a/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
+++ b/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
@@ -187,7 +187,16 @@ class Jetpack_Subscribe_Modal {
 		$is_modal_enabled   = get_option( 'sm_enabled' ) || false;
 		$is_fse_theme       = wp_is_block_theme();
 
-		return ( $is_lettre || $is_newsletter_site ) && $is_fse_theme && $is_modal_enabled;
+		$should_enable_subscriber_modal = ( $is_lettre || $is_newsletter_site ) && $is_fse_theme && $is_modal_enabled;
+
+		/**
+		 * Allows force-enabling or disabling the Subscriptions modal.
+		 *
+		 * @since $$next-version$$
+		 *
+		 * @param bool $should_enable_subscriber_modal Whether the Subscriptions modal should be enabled.
+		 */
+		return (bool) apply_filters( 'jetpack_subscriptions_modal_enabled', $should_enable_subscriber_modal );
 	}

```

You would then be able to force-enable the modal by adding the following in a functionality plugin:

```php
add_filter( 'jetpack_subscriptions_modal_enabled', '__return_true' );
```

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
jeherve: Yes, you continue to enqueue the stylesheet as usual. We take care of dequeuing it when the concatenated version has to be enqueued instead.

edanzer:
@crisbusquets 

I've addressed most of your feedback, and updated the screenshot in the PR description for how the modal looks now. 

> Would it be possible that the modal only appears in posts?

This has been done!

> The copy will be ok if we can change this behavior. If it's not possible, I'd change the copy to:

Even though we only show on posts, I've gone ahead and updated the copy to what you suggest above anyways. The way the modal works, we don't (for now) prevent users from reading - ie, it is not limited to subscribers. The new copy you suggested above reads nicely and seems a bit more versatile. 

> Success screen/behavior

If it's ok, I'm going to try to tackle this in a separate PR that updates the subscribe block itself. 

> Is it possible to close de modal when clicking outside it?

Yes, the modal will close if you click anywhere outside of it!

> The modal uses the typography on the design, not the one that the user has in the theme, right?

It is actually using the theme fonts right now. This includes not just font family, but also certain things like font weight, which have limited options in the block editor. I'm open to changing that. Two notes. First, it may makes sense to keep fonts consistent for each site - ie, a site that all all blocky serif fonts may want those in the modal. Second, we have limited ability to add custom fonts within the core blocks in the block editor. It's possible we'd need to override these via separate CSS, which just means users could not edit the font family short of writing their own custom css. 

> Figma

For some reason, the figma file links you provided don't work. Something's not quite working with the dynamic links. I'm trying to keep accurate, but I'm not quite sure how to find the figma files without the direct links above.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
edanzer: Note: I've actually moved the text/markup for the Close button inside the template part now too, so same question will apply on that.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -0,0 +1,236 @@
...
+	 * Returns the initial content of the Subscribe Modal template.
+	 * This can then be edited by the user.
+	 *
+	 * @return string
+	 */
+	public function get_subscribe_template_content() {
+		return '<!-- wp:group {"style":{"spacing":{"padding":{"top":"50px","bottom":"50px","left":"20px","right":"20px"}}},"layout":{"type":"constrained"}} -->
+		<div class="wp-block-group" style="padding-top:50px;padding-right:20px;padding-bottom:50px;padding-left:20px"><!-- wp:group {"style":{"dimensions":{"minHeight":"0px"},"spacing":{"blockGap":"8px"}},"layout":{"type":"flex","flexWrap":"wrap","justifyContent":"center"}} -->
+		<div class="wp-block-group" style="min-height:0px"><!-- wp:heading {"style":{"typography":{"fontStyle":"normal","fontWeight":"600","fontSize":"26px"},"layout":{"selfStretch":"fit","flexSize":null}}} -->
+		<h2 class="wp-block-heading" style="font-size:26px;font-style:normal;font-weight:600">Discover more from</h2>
```
lezama: should we make these translatable?

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -0,0 +1,236 @@
...
+		}
+
+		return self::$instance;
+	}
+
+	/**
+	 * Jetpack_Subscribe_Modal class constructor.
+	 */
+	public function __construct() {
+		add_action( 'wp_loaded', array( $this, 'wpcom_create_subscribe_template' ) );
```
lezama: should we move the `if $this->should_enable_subscriber_modal()` check up here?

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -0,0 +1,236 @@
...
+	 * Adds modal with Subscribe Modal content.
+	 *
+	 * @return void
+	 */
+	public function wpcom_add_subscribe_modal_to_frontend() {
+		if (
+			$this->should_enable_subscriber_modal() &&
+			! is_admin() &&
+			is_single() &&
+			'post' == get_post_type()
```
lezama:
```suggestion
			'post' === get_post_type()
```

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
edanzer: Yes, I'd posted a question above. Think there's any weirdness in trying to translate content that is ultimately inserted as post content in the block editor?

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
edanzer: That makes sense to me. Might as well do the check early as possible and run as little unneeded code as possible. I'll update that.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
edanzer: Good catch, will update.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -0,0 +1,236 @@
...
+		) {
+			$posts = get_posts(
+				array(
+					'post_type'   => 'wp_template_part',
+					'post_status' => 'publish',
+					'numberposts' => -1,
```
lezama: I think you can pass 

```php
'name' => $this->get_subscribe_template_slug(),
``` 

so you don't need to do the `reset` call bellow

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -0,0 +1,236 @@
...
+			$subscribe_template = reset(
+				array_filter(
+					$posts,
+					function ( $post ) {
+						return $post->post_name === $this->get_subscribe_template_slug();
+					}
+				)
+			);
+
+			$blocks = parse_blocks( $subscribe_template->post_content );
```
lezama: I think if we do `apply_filters('the_content', $post->post_content);` we can avoid doing the loop and `wp_kses`

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
edanzer: FYI - text for this has been moved inside the subscribe modal template part. So any translation will need to be handled there.

projects/plugins/jetpack/_inc/client/discussion/subscriptions.jsx
edanzer: Yup, I will handle this in a follow up PR.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.css
edanzer: Handy tool! I did try this. The first thing I found the modal wasn't loading at all on most themes due to a JS error with the wp object not being available. So I pushed a fix for that - just requiring wp-components when enqueuing.

I then used the tool, and below is the result. Obviously the style varies quite a bit. This is intentional for now, and reflects how WordPress is supposed to work in the block editor / FSE era. In short, the modal picks up default theme styling (fonts, colors, etc). We'd need to decide if we want to override those. My sense is probably not. 

The one thing that does need fixing is there are several cases where the font is dark on the button, and thus not visible. Given buttons are generally darker colors, it could make sense to force a white font on the button? 

screencapture-erick-jurassic-tube-wp-admin-tools-php-2023-06-27-11_15_10

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
edanzer: @jeherve - Thanks for the details. I've added this. I find myself wondering if we should also add a filter for users to add their own modal content, vs just using the new subscribe modal template part.  I'd do that later as a separate PR.

As a side note, a lot of my recent dev work has been on WordPress.com, and in that environment we think a lot less about leveraging these WP apis to make code behavior customizable by other devs. So this was a great general lesson to be thinking much more about that when working on Jetpack or other code that will be delivered via WordPress.org.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
edanzer: Since we weren't sure, I just tested this. It works. Having the translation methods as part of the markup here doesn't negatively affect creating the template. So I've pushed the change adding translations.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
edanzer: I updated this. I looked at moving the other checks for loading the frontend modal markup and JS, but the nature of those checks is that they need to happen later in the WordPress loading process.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
edanzer: Hey @lezama - Couple questions on this one...

First...  Basically, we're taking some content from an editor template, content which is formatted in block format, and outputting outside the normal context, in a custom modal. My understanding is that you'd still need to correctly parse/convert the block markup to regular html mark up. Would the method you're describing somehow do that as well?

Second... I may be rusty on my filters, but apply_filter() I think just adds a new filter. Not sure how that would resolve the main issue just above. Did you mean to suggest add_filter()? Even in that case, I wouldn't think add_filter( 'the_content', $my_content ) wouldn't work here either. That filters the post content, but doesn't solve the issue of converting block markup to html markup.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
edanzer: Updated see https://github.com/Automattic/jetpack/pull/31393#discussion_r1244160182

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
edanzer: Good call @lezama! Updated this. While I was at it, I also added a check to confirm we found posts in this query before we proceed to try adding the template modal.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -0,0 +1,251 @@
...
+	 *
+	 * @return void
+	 */
+	public function wpcom_add_subscribe_modal_to_frontend() {
+		if ( $this->is_front_end_single_post() ) {
+			$posts = get_posts(
+				array(
+					'post_type'   => 'wp_template_part',
+					'post_status' => 'publish',
+					'numberposts' => -1,
```
lezama: why not  `'numberposts'   => 1,`?

edanzer: While it shouldn't happen, I had an instance today where this query returned more than one post, and the right post was the second / more recent one. Somehow, and old version was still there matching the query. That's why the combo of -1 here (which is any matches) and end() just below, which grabs the most recent match.

edanzer:
FYI - I'm looking into the failing PHP unit tests now. Looks like those are failing because add a whitelist option but didn't updated tests to include that.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
lezama: oh, we could add `order` to the query then.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -183,7 +183,16 @@ public function should_enable_subscriber_modal() {
...
+		$should_enable_subscriber_modal = ( $is_lettre || $is_newsletter_site ) && $is_fse_theme && $is_modal_enabled;
+
+		/**
+		* Allows force-enabling or disabling the Subscriptions modal.
+		*
+		* @since $$next-version$$
+		*
+		* @param bool $should_enable_subscriber_modal Whether the Subscriptions modal should be enabled.
+		*/
+		return (bool) apply_filters( 'jetpack_subscriptions_modal_enabled', $should_enable_subscriber_modal );
```
lezama: perfect

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.php
simison: I think it's fine if they can modify the modal via template editor and no need to modify via code.

edanzer:
I just tested the new 'jetpack_subscriptions_modal_enabled' that we've added as a feature flag. I also updated testing instructions for others to set that now.

projects/plugins/jetpack/_inc/client/discussion/subscriptions.jsx
jeherve: I would agree. The capitalization was what jumped at me. Maybe this could at least be:

```suggestion
								label={ __( 'Enable the popup subscription modal', 'jetpack' ) }
```

projects/plugins/jetpack/_inc/client/discussion/subscriptions.jsx
```diff
@@ -134,6 +139,17 @@ class SubscriptionsComponent extends React.Component {
 									'jetpack'
 								) }
 							/>
+							<ToggleControl
```
jeherve: I think it may be best to hide this with the feature flag. Otherwise folks will toggle that option, expect something to happen, and reach out to support when nothing happens.

You can check #31479 for an example of a PR that adds the necessary tooling to pull the feature flag from PHP into the dashboard here.

projects/packages/sync/src/class-package-version.php
```diff
@@ -12,7 +12,7 @@
  */
 class Package_Version {
 
-	const PACKAGE_VERSION = '1.50.2-alpha';
+	const PACKAGE_VERSION = '1.51.1-alpha';
```
jeherve: Looks like you'll need to update this PR with latest trunk once more, and run the script to update the versions.

projects/plugins/jetpack/modules/subscriptions.php
```diff
@@ -348,6 +348,21 @@ public function configure() {
 			'stc_enabled'
 		);
 
+		/** Enable Subscribe Modal */
+
+		add_settings_field(
```
jeherve: I think this should also be hidden behind the feature flag.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -0,0 +1,261 @@
...
+
+			?>
+				<div class="jetpack-subscribe-modal">
+					<div class="jetpack-subscribe-modal__modal-content">
+						<?php
+						foreach ( $blocks as $block ) {
+							echo wp_kses( render_block( $block ), $allowed_html );
+						}
+						?>
+						<Dashicon icon="minus" className="jp-idc__idc-screen__card-action-separator" />
```
jeherve: This isn't valid HTML, so it won't be displayed properly I'm afraid.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -0,0 +1,261 @@
...
+		}
+	}
+
+	/**
+	 * Adds a Subscribe Modal template part that can be edited in the editor.
+	 * This is later loaded in a pop up modal for Newsletter sites.
+	 *
+	 * @return void
+	 */
+	public function wpcom_create_subscribe_template() {
```
jeherve: I wonder if this could be triggered when the toggle is enabled in the Jetpack Settings. This may prove better for performance, since later on we wouldn't have to check for an existing post on each `wp_loaded`. What do you think?

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -0,0 +1,261 @@
...
+
+	/**
+	 * Enqueues JS to load modal.
+	 *
+	 * @return void
+	 */
+	public function wpcom_enqueue_subscribe_modal_assets() {
+		if ( $this->is_front_end_single_post() ) {
+			wp_enqueue_style( 'subscribe-modal-css', plugins_url( 'subscribe-modal.css', __FILE__ ), array(), JETPACK__VERSION );
+			wp_enqueue_script( 'subscribe-modal-js', plugins_url( 'subscribe-modal.js', __FILE__ ), array( 'wp-components' ), JETPACK__VERSION, true );
```
jeherve: Do we need `wp-components` to be loaded here? Could we do without it, to avoid loading too many additional things on the frontend?

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -0,0 +1,261 @@
...
+	 * Returns true if we should enable Newsletter content.
+	 * This is currently limited to lettre theme or newsletter sites.
+	 * We could open it to all themes or site intents.
+	 *
+	 * @return bool
+	 */
+	public function should_enable_subscriber_modal() {
+		$is_lettre          = get_option( 'stylesheet' ) === 'lettre';
+		$is_newsletter_site = get_option( 'site_intent' ) === 'newsletter';
+		$is_modal_enabled   = get_option( 'sm_enabled' ) || false;
```
jeherve: Instead of `|| false` here, you should be able to provide a default option in your `get_option` call:

```suggestion
		$is_modal_enabled   = get_option( 'sm_enabled', false );
```

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -0,0 +1,261 @@
...
+	 * Returns true if we we are on frontend of single post.
+	 * Note: Because of how WordPress works, this function will
+	 * only return the correct value after a certain point in the
+	 * WordPress loading process. You cannot, for example, use this
+	 * method in the class contructor method - it will always return false.
+	 *
+	 * @return bool
+	 */
+	public function is_front_end_single_post() {
+		return ! is_admin() && is_single() && 'post' === get_post_type();
```
jeherve: You should be able to use `is_singular` here:

```suggestion
		return ! is_admin() && is_singular( 'post' );
```

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -0,0 +1,261 @@
...
+	 * Returns the initial content of the Subscribe Modal template.
+	 * This can then be edited by the user.
+	 *
+	 * @return string
+	 */
+	public function get_subscribe_template_content() {
+		return '<!-- wp:group {"style":{"spacing":{"padding":{"top":"50px","bottom":"50px","left":"20px","right":"20px"}}},"layout":{"type":"constrained"}} -->
+		<div class="wp-block-group" style="padding-top:50px;padding-right:20px;padding-bottom:50px;padding-left:20px"><!-- wp:group {"style":{"dimensions":{"minHeight":"0px"},"spacing":{"blockGap":"8px"}},"layout":{"type":"flex","flexWrap":"wrap","justifyContent":"center"}} -->
+		<div class="wp-block-group" style="min-height:0px"><!-- wp:heading {"style":{"typography":{"fontStyle":"normal","fontWeight":"600","fontSize":"26px"},"layout":{"selfStretch":"fit","flexSize":null}}} -->
+		<h2 class="wp-block-heading" style="font-size:26px;font-style:normal;font-weight:600">' . __( 'Disover more from', 'jetpack' ) . '</h2>
```
jeherve:
```suggestion
		<h2 class="wp-block-heading" style="font-size:26px;font-style:normal;font-weight:600">' . __( 'Discover more from', 'jetpack' ) . '</h2>
```

jeherve: I should add that this may prove difficult to translate for translators, since the site title is outside of the string. In languages where the site title would come before "from", it will be a problem.
It would be nice to have the site title part of that string instead.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -0,0 +1,261 @@
...
+		
+		<!-- wp:site-title {"level":2,"textAlign":"center","style":{"typography":{"fontStyle":"normal","fontWeight":"600","lineHeight":"1.2","fontSize":"26px"}}} /--></div>
+		<!-- /wp:group -->
+		
+		<!-- wp:paragraph {"align":"center","style":{"typography":{"fontSize":"15px"},"spacing":{"margin":{"top":"4px","bottom":"0px"}}}} -->
+		<p class="has-text-align-center" style="margin-top:4px;margin-bottom:0px;font-size:15px">' . __( 'Subscribe to the newsletter to keep reading', 'jetpack' ) . '</p>
+		<!-- /wp:paragraph -->
+		
+		<!-- wp:paragraph {"align":"center","style":{"typography":{"fontSize":"15px"},"spacing":{"margin":{"bottom":"20px","top":"0px"}}}} -->
+		<p class="has-text-align-center" style="margin-top:0px;margin-bottom:20px;font-size:15px">and get access to the full archive.</p>
```
jeherve: We'll probably want to make translatable.

jeherve (dismissed):
Nice work so far, thanks for sticking with it in this massive PR :)

I gave it a quick try and left some notes below.

*********

I'm seeing a number of notices when the modal is displayed:

```
PHP Notice:  Undefined index:  in /wp-includes/blocks/template-part.php on line 238
PHP Warning:  Cannot modify header information - headers already sent by (output started at /wp-includes/blocks/template-part.php:238) in /wp-includes/pluggable.php on line 1427
```

projects/plugins/jetpack/_inc/client/discussion/subscriptions.jsx
lezama: Erick is probably following the pattern of the two previous settings.

Screenshot 2023-07-05 at 18 12 52

lezama:
Thanks @jeherve! I will be working on your feedback since Erick is AFK this week.

> I'm seeing a number of notices when the modal is displayed:

I am not getting these, anything I should do in particular to reproduce?

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
simison: Same note about ease of translating https://github.com/Automattic/jetpack/pull/31393#discussion_r1253363857

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
simison: This would be important to solve — can be a follow-up PR.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
simison: It might still be important to do the check outside the toggle; thinking situations where we add the option by default onto the site when site is created.

simison:
In my standalone Jetpack with Gutenberg disabled I'm getting this visually, but I still need to try see if something didn't just build correctly.

image

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
simison: Actually we shouldn't need it, looking at `subscribe-modal.js`, but we do need `domReady` which is a tiny script.

jeherve:
> I am not getting these, anything I should do in particular to reproduce?

I got them on a fresh JN site with current WP and no Gutenberg. I didn't do anything special I think.

lezama:
@scruffian what do you think of [this way of adding a template part](https://github.com/Automattic/jetpack/pull/31393/files#diff-bcd4c897206a1f66102ef44f0e272985d67be37c91dfacbbd6458d46d66ac72eR52). Anything we could do differently?

Also any ideas on why the template part wouldn't show up on my test site full site editor 😅 ?

projects/plugins/jetpack/_inc/client/discussion/subscriptions.jsx
lezama: [Done!](https://github.com/Automattic/jetpack/pull/31393/commits/e5134959dabfce6e9e36bb6893b8e1c4c1feac24)

projects/plugins/jetpack/modules/subscriptions.php
lezama: [done](https://github.com/Automattic/jetpack/commit/41c5af8c58e3698130b7906748bdacffcb8ccbdb), where these show up?

scruffian:
> @scruffian what do you think of [this way of adding a template part](https://github.com/Automattic/jetpack/pull/31393/files#diff-bcd4c897206a1f66102ef44f0e272985d67be37c91dfacbbd6458d46d66ac72eR52). Anything we could do differently?

I think you should look into doing this via the API - there's an enpoint at `/wp-json/wp/v2/templates?_locale=user`, which you should be able to post to, to create a new template. Let me know if you need more help with that :)

scruffian:
Another option is adding templates via a filter, like I prototype here: https://github.com/WordPress/create-block-theme/pull/124

This avoids creating stuff in a user's database unless they really need it.

lezama:
After @simison and @scruffian feedback I cleaned up the code inspired by [Ben's suggestion](https://github.com/Automattic/jetpack/pull/31393#issuecomment-1625386473). I will follow up with a PR to make the template part show up in the site editor.

@jeherve I think I went through all your feedback, except one thing, the mess with the composer.lock files and package versions. I would love your help with that 🙏

oskosk:
The changelogger check is failing. When this happens, a rebase against trunk (`git merge trunk`) and the command shown in the files view for this PR should be run, followed by a commit, and a new push. I'm doing it for you now.

image

oskosk:
Pushed rebase in 67f117c3f2832008cc87837e5212deda67a49202 and versions update in 8fe478c1a2c64e2b6d7c1fc9af58b1700dcba41a

edanzer:
The new tests written for this PR were failing due to latest changes. I've updated them to reflect the current logic for the feature flag and for should_load_subscriber_modal().

Valecia2023:
circular-form.jpg-photo-message

projects/plugins/jetpack/_inc/client/discussion/subscriptions.jsx
```diff
@@ -134,6 +141,27 @@ class SubscriptionsComponent extends React.Component {
...
+										disabled={
+											! isSubscriptionsActive ||
+											unavailableInOfflineMode ||
+											this.props.isSavingAnyOption( [ 'subscriptions' ] )
+										}
+										toggling={ this.props.isSavingAnyOption( [ 'sm_enabled' ] ) }
+										onChange={ this.handleSubscribeModalToggleChange }
+										label={ __( 'Enable subscriber modal', 'jetpack' ) }
+									/>
+									<p className="jp-form-setting-explanation">
```
simison, line 157: I used separate `p` here because `help={}` prop just doesn't look great at all for a longer text. I copied the pattern from elsewhere in Jetpack settings:

**help prop**
Screenshot 2023-07-13 at 11 38 34

**paragraph**
Screenshot 2023-07-13 at 11 44 01

simison (dismissed):
This is in good shape now! I've tested a bunch. Since this is behind feature flags, we can safely merge and continue polishing (and releasing soon!) in follow-up PRs.

Screenshot 2023-07-13 at 11 44 01

image

In follow-up PRs we should:
- Add "edit modal" link next to the toggle
- Tweak logic when the modal shows up; site owners, already subscribed, if the post is paid/subscribe walled already in which case they see the same content inline already, etc.

simison (approved):

projects/plugins/jetpack/modules/subscriptions.php
```diff
@@ -461,6 +479,22 @@ public function subscription_comment_subscribe_setting() {
...
+	/**
+	 * Subscribe Modal Toggle.
+	 */
+	public function subscribe_modal_setting() {
+
+		$sm_enabled = get_option( 'sm_enabled', 1 );
+		?>
+
+		<p class="description">
+			<input type="checkbox" name="sm_enabled" id="jetpack-subscribe-modal" value="1" <?php checked( $sm_enabled, 1 ); ?> />
```
jeherve, line 491: Coming back to this today; shouldn't the default / fallback here be `0` everywhere, since we do not enable the modal by default?

PR #31401, opened 2023-06-15
AI Assistant: Separate prompt text from relevant content for extensions

dhasilva:
Fixes #

## Proposed changes:

This PR separates the prompt text and rules from what content should be relevant. This makes it possible to reuse the same prompts in the block and the extensions with different relevant contents.

### Other information:

## Testing instructions:

* Go to the block editor
* Add an AI Assistant block
* Check all the options, they should work as normal (check the results and the logs)
* Add a new paragraph or heading
* Try the options in the extended block
* They should use the correct content (the block content) and work as intended


retrofox (approved):
LGTM. We need to keep polishing the API. Ideally, from my PoV, it should be agnostic of the requester (block, extension, whatever). Nice work.

PR #31402, opened 2023-06-15
Fix jetpack iframe wpcom path for proper staticizing url generation

jgcaruso:
This PR updates the path that is used for enqueueing the `jetpack-iframe-embed.js` script when run on wordpress.com to match pattern used by other scripts (see fbhepr%2Skers%2Sjcpbz%2Sjc%2Qpbagrag%2Szh%2Qcyhtvaf%2Seyg.cuc%3Se%3Q7410s8nn%2319-og `rlt.php`). This script was being enqueued with `http` even though the page was requested on `https`. It was the only script that behaved this way.

Alternatively, the path could just be hardcoded with `https://s0.wp.com` like `videopress-iframe-api.js` does in https://github.com/Automattic/jetpack/blob/dfc7588bcc0f034b7aa53e87c9777c956f3988c3/projects/packages/videopress/src/class-initializer.php#L398

| before | after |
| -- | -- |
| image | image |

Fixes #

## Proposed changes:

* updates the path that is used for enqueueing the `jetpack-iframe-embed.js` script when run on wordpress.com

### Other information:

## Jetpack product discussion

pMz3w-hKd-p2

## Testing instructions:

* Apply this patch to a wpcom sandbox
* Skip the `WPCOM_SANDBOXED` early return in `wp-content/mu-plugins/optimizations.php` by adding `false &&` to the condition (or just comment out the code)
* sandbox `jgcactivateissue06132023.wordpress.com`
* visit the following url `https://jgcactivateissue06132023.wordpress.com/?iframe=true&theme_preview=true&hide_banners=true&preview=true&do_preview_no_interactions=false&concat_js=no`
* view page source
* search for `jetpack-iframe-embed-js`
* the url for the script should be `https://s0.wp.com` not `http`
* remove the patch
* refresh the page source, you will see the script has the wrong protocol now (original issue)

### DIFF
--- a/projects/plugins/jetpack/_inc/lib/class.jetpack-iframe-embed.php
+++ b/projects/plugins/jetpack/_inc/lib/class.jetpack-iframe-embed.php
@@ -39,7 +39,7 @@ class Jetpack_Iframe_Embed {
 		if ( defined( 'IS_WPCOM' ) && IS_WPCOM ) {
 			wp_enqueue_script(
 				'jetpack-iframe-embed',
-				WPMU_PLUGIN_URL . '/jetpack-iframe-embed/jetpack-iframe-embed.js',
+				'/wp-content/mu-plugins/jetpack-iframe-embed/jetpack-iframe-embed.js',
 				array( 'jquery' ),
 				$ver,
 				false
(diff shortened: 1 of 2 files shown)


anomiex (approved):
Works as advertised.

PR #31407, opened 2023-06-15
ci: Checkout package trees for build

anomiex:
## Proposed changes:

In #31220 we added logic to the build to append a timestamp to package alpha versions, derived from the last git commit to the package.

But in CI we use a shallow checkout, so the "last git commit" is the grafted commit. To get the correct timestamp we need to deepen the checkout until `git log -1` returns a non-grafted commit for each package that has any change entries.

On the plus side, we can use `--filter=blob:none` to fetch only the directory tree metadata and not all the file contents.

### Other information:

## Jetpack product discussion

None

No

## Testing instructions:

* Check the build artifact for this PR. Assuming no later PRs have been merged touching packages/connection, the version recorded in the Jetpack plugin's `vendor/composer/jetpack_autoload_classmap.php` for `Automattic\\Jetpack\\Connection\\Client` should be "1.53.0.0-alpha1686840636" (based on when dfc7588bcc0f034b7aa53e87c9777c956f3988c3 was committed).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,6 +64,31 @@ jobs:
             const data = await checkTestReminderComment( github, context, core );
             return data;
 
+      # We need the tree (but not the blob) for packages that will have -alpha version numbers so the timestamp appending works right.
+      - name: Deepen tree for packages
+        env:
+          CHANGED: ${{ steps.changed.outputs.projects }}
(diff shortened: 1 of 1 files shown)


zinigor (approved):
TIL about grafted commits! Looks good to me!

PR #31413, opened 2023-06-16
Custom CSS: update Settings UI to recommend Global Styles over Custom CSS for Block-Based Themes

### LINKED ISSUE #71130: Remove Additional CSS menu for Block Themes
# Description 

Until now, "Custom CSS" feature has been the last feature in good old Customizer that FSE sites have access to:

Screenshot 2022-12-13 at 14 30 47

Screenshot 2022-12-13 at 14 31 10

Now that Gutenberg has ["custom CSS" panel directly in the site editor](https://github.com/WordPress/gutenberg/pull/46141), we should be able to deprecate the WP.com Custom CSS feature (at least for FSE sites) and rely on core implementation instead:

Screenshot 2022-12-13 at 14 30 30

Furthermore, core's custom CSS should probably be gated to paid plans just like our own Jetpack version has been.

There's also ["per block CSS" coming up](https://github.com/WordPress/gutenberg/issues/44412).

----
## T-rex work

 * P2: [Removing customizer menu items]  pekYwv-1N2-p2

rcrdortiz:
Fixes https://github.com/Automattic/wp-calypso/issues/71130

## Proposed changes:

This PR introduces a new feature to the Jetpack plugin to detect if the WordPress site has the core site editor feature enabled. This information is stored in the global state tree and can be accessed via the getIsCoreSiteEditorEnabled function. The detection mechanism is used to adjust the Jetpack settings, particularly under Writing settings.

Moreover, the PR makes a contextual adjustment to the user interface. The Jetpack > Settings > Writing > Custom CSS toggle is now conditionally rendered based on whether the site uses a block-based theme or not. For sites using a block-based theme, this toggle is removed, aligning with the concept that such themes inherently offer greater flexibility for customization and thus the additional CSS settings might not be necessary.

This PR involves changes in multiple files, including state reducers, the writing settings component, and theme enhancements, to ensure the changes are effectively integrated.

This is classified as a minor enhancement due to its impact on improving the user interface and experience when interacting with block-based themes.

The change should be transparent to the end-user and help streamline the interface by removing potentially redundant controls for sites using block-based themes.

See some screenshots in #31601. 

### Other information:

## Jetpack product discussion

See discussion in the PR below, and in #31601.

* Yes, it adds a Tracks event when you click on the link to access Global Styles, matching the existing link to the customizer.

## Testing instructions:

1. Start with a site connected to WordPress.com, either self-hosted on WoA, with a classic theme like Twenty Twenty One.
2. Go to Jetpack > Settings > Writing
    * You should be able to turn the Custom CSS feature on and off.
    * Leave the feature on.
3. Now go to Appearance > Themes and switch to the Twenty Twenty Three theme.
4. Go back to Jetpack > Settings > Writing.
    * You should still have the option to turn the feature off, but a notice now recommends Global styles instead of the Customizer.
5. Turn the feature off
    * The toggle should disappear. A notice should remain, explaining that Global Styles are the way to go now.


projects/plugins/jetpack/_inc/client/writing/theme-enhancements.jsx
```diff
@@ -63,8 +63,8 @@ class ThemeEnhancements extends React.Component {
 	/**
 	 * Update state so toggles are updated.
 	 *
-	 * @param {string} optionName option slug
-	 * @param {string} module module slug
+	 * @param {string} optionName - option slug
+	 * @param {string} module - module slug
```
rcrdortiz, line 67: All these unrelated changes are because of the linter.

mmtr:
I don't see the Custom CSS toggle after activating Twenty Seventeen (a classic theme). Not sure if I'm missing something 🤔 





https://github.com/Automattic/jetpack/assets/1233880/739f5aa2-a29c-49bf-9950-224f385c92ac

rcrdortiz:
> I don't see the Custom CSS toggle after activating Twenty Seventeen (a classic theme). Not sure if I'm missing something 🤔
> 
>  Screen.Recording.2023-06-16.at.15.46.10.mov

Hmmm that's strange, I'm able to reproduce it when I use the Jetpack beta plugin, but not when I sync the code manually. I'll have a look at it and get back to you.

mmtr:
> I don't see the Custom CSS toggle after activating Twenty Seventeen (a classic theme).

Seems to be a caching issue (the toggle shows up after waiting a few minutes). I don't think this is a blocker, because probably not many users needs this toggle immediately after switching the theme, so we can investigate it in a follow-up.

This should be good to go as long as we revert https://github.com/Automattic/jetpack/pull/31413/commits/df38f48b62564826e1731a2f1a5c7bd13cd585f7

rcrdortiz:
> > I don't see the Custom CSS toggle after activating Twenty Seventeen (a classic theme).
> 
> Seems to be a caching issue (the toggle shows up after waiting a few minutes). I don't think this is a blocker, because probably not many users needs this toggle immediately after switching the theme, so we can investigate it in a follow-up.
> 
> This should be good to go as long as we revert [df38f48](https://github.com/Automattic/jetpack/commit/df38f48b62564826e1731a2f1a5c7bd13cd585f7)

Removed.

jeherve (dismissed):
I think you're missing a piece, where `is_core_site_editor_enabled` is added to `Jetpack_Redux_State_Helper`?

jeherve:
> I think you're missing a piece, where `is_core_site_editor_enabled` is added to `Jetpack_Redux_State_Helper`?

@rcrdortiz reached out to me, and it turns out I had misread. Sorry about that! I had expected the theme information to be fetched from the local initial state, not from site state which is fetched from the WordPress.com REST API.

I think it would make sense to fetch that information locally, we would avoid any potentially outdated information while sync is happening, right after updating your theme. Something like this:

```diff
diff --git a/projects/plugins/jetpack/_inc/client/state/initial-state/reducer.js b/projects/plugins/jetpack/_inc/client/state/initial-state/reducer.js
index 76b908919b..bf4c6df6ac 100644
--- a/projects/plugins/jetpack/_inc/client/state/initial-state/reducer.js
+++ b/projects/plugins/jetpack/_inc/client/state/initial-state/reducer.js
@@ -372,6 +372,16 @@ export function currentThemeSupports( state, feature ) {
 	return get( state.jetpack.initialState.themeData, [ 'support', feature ], false );
 }
 
+/**
+ * Check that the current theme is a block theme.
+ *
+ * @param {object} state - Global state tree.
+ * @returns {boolean} True if the current theme is a block theme, false otherwise.
+ */
+export function currentThemeIsBlockTheme( state ) {
+	return get( state.jetpack.initialState.themeData, [ 'isBlockTheme' ], false );
+}
+
 /**
  * Check if backups UI should be displayed.
  *
diff --git a/projects/plugins/jetpack/_inc/client/state/site/reducer.js b/projects/plugins/jetpack/_inc/client/state/site/reducer.js
index 5f11b04283..711582a3c8 100644
--- a/projects/plugins/jetpack/_inc/client/state/site/reducer.js
+++ b/projects/plugins/jetpack/_inc/client/state/site/reducer.js
@@ -532,16 +532,6 @@ export function getSiteID( state ) {
 	return get( state.jetpack.siteData, [ 'data', 'ID' ] );
 }
 
-/**
- * Returns if the site has the core site editor feature enabled.
- *
- * @param {object} state - Global state tree
- * @returns {boolean} True if the site has the core site editor feature enabled, false otherwise.
- */
-export function getIsCoreSiteEditorEnabled( state ) {
-	return get( state.jetpack.siteData, [ 'data', 'is_core_site_editor_enabled' ] );
-}
-
 /**
  * Returns plugins that use the Jetpack connection
  *
diff --git a/projects/plugins/jetpack/_inc/client/writing/index.jsx b/projects/plugins/jetpack/_inc/client/writing/index.jsx
index 7ddc4a3a3b..689b00fe78 100644
--- a/projects/plugins/jetpack/_inc/client/writing/index.jsx
+++ b/projects/plugins/jetpack/_inc/client/writing/index.jsx
@@ -13,7 +13,6 @@ import { userCanManageModules, userCanEditPosts, isAtomicSite } from 'state/init
 import { isModuleActivated, getModuleOverride, getModule } from 'state/modules';
 import { isModuleFound } from 'state/search';
 import { getSettings } from 'state/settings';
-import { getIsCoreSiteEditorEnabled } from 'state/site';
 import Composing from './composing';
 import CustomContentTypes from './custom-content-types';
 import { Masterbar } from './masterbar';
@@ -123,7 +122,6 @@ export default connect( state => {
 		userCanManageModules: userCanManageModules( state ),
 		isModuleFound: module_name => isModuleFound( state, module_name ),
 		connectUrl: getConnectUrl( state ),
-		isBlockBasedThemeActive: getIsCoreSiteEditorEnabled( state ),
 		getModuleOverride: module_name => getModuleOverride( state, module_name ),
 	};
 } )( Writing );
diff --git a/projects/plugins/jetpack/_inc/client/writing/theme-enhancements.jsx b/projects/plugins/jetpack/_inc/client/writing/theme-enhancements.jsx
index 0d489379a8..046ffd57dc 100644
--- a/projects/plugins/jetpack/_inc/client/writing/theme-enhancements.jsx
+++ b/projects/plugins/jetpack/_inc/client/writing/theme-enhancements.jsx
@@ -9,7 +9,7 @@ import SettingsGroup from 'components/settings-group';
 import analytics from 'lib/analytics';
 import React from 'react';
 import { connect } from 'react-redux';
-import { currentThemeSupports } from 'state/initial-state';
+import { currentThemeSupports, currentThemeIsBlockTheme } from 'state/initial-state';
 import { getModule } from 'state/modules';
 import { isModuleFound } from 'state/search';
 
@@ -194,7 +194,7 @@ class ThemeEnhancements extends React.Component {
 						) }
 					</SettingsGroup>
 				) }
-				{ ! this.props.isBlockBasedThemeActive && foundCustomCSS && (
+				{ ! this.props.isBlockThemeActive && foundCustomCSS && (
 					<SettingsGroup
 						module={ { module: customCSS.module } }
 						support={ {
@@ -240,6 +240,7 @@ class ThemeEnhancements extends React.Component {
 export default connect( state => {
 	return {
 		module: module_name => getModule( state, module_name ),
+		isBlockThemeActive: currentThemeIsBlockTheme( state ),
 		isInfiniteScrollSupported: currentThemeSupports( state, 'infinite-scroll' ),
 		isModuleFound: module_name => isModuleFound( state, module_name ),
 	};
diff --git a/projects/plugins/jetpack/_inc/lib/admin-pages/class-jetpack-redux-state-helper.php b/projects/plugins/jetpack/_inc/lib/admin-pages/class-jetpack-redux-state-helper.php
index be4335f43e..731b9b99af 100644
--- a/projects/plugins/jetpack/_inc/lib/admin-pages/class-jetpack-redux-state-helper.php
+++ b/projects/plugins/jetpack/_inc/lib/admin-pages/class-jetpack-redux-state-helper.php
@@ -193,9 +193,10 @@ class Jetpack_Redux_State_Helper {
 				'latestBoostSpeedScores'     => $speed_score_history->latest(),
 			),
 			'themeData'                   => array(
-				'name'      => $current_theme->get( 'Name' ),
-				'hasUpdate' => (bool) get_theme_update_available( $current_theme ),
-				'support'   => array(
+				'name'         => $current_theme->get( 'Name' ),
+				'hasUpdate'    => (bool) get_theme_update_available( $current_theme ),
+				'isBlockTheme' => (bool) $current_theme->is_block_theme(),
+				'support'      => array(
 					'infinite-scroll' => current_theme_supports( 'infinite-scroll' ) || in_array( $current_theme->get_stylesheet(), $inf_scr_support_themes, true ),
 					'widgets'         => current_theme_supports( 'widgets' ),
 					'webfonts'        => (

```

jeherve:
> the PR makes a contextual adjustment to the user interface. The Jetpack > Settings > Writing > Custom CSS toggle is now conditionally rendered based on whether the site uses a block-based theme or not. For sites using a block-based theme, this toggle is removed, aligning with the concept that such themes inherently offer greater flexibility for customization and thus the additional CSS settings might not be necessary.

I wonder if we could take a different approach, and only display a notice. In some scenarios, custom CSS can still be useful when using a block theme (I personally use it on a site of mine at the moment running Twenty Twenty Three, so that's only my personal experience, feel free to disagree! :) )

Once Gutenberg's global CSS feature will no longer be classified as experimental, we could display a notice when folks have CSS stored in Custom CSS (not SCSS though, just regular CSS), inviting them to move that CSS to a new location before to disable Jetpack's Custom CSS feature. Until then, I think custom CSS still has its use with block themes.

Copons:
> Once Gutenberg's global CSS feature will no longer be classified as experimental

@jeherve Just chiming in to say that it's not an experimental feature anymore as of [WP 6.2](https://wordpress.org/documentation/wordpress-version/version-6-2/#custom-css). 🙂

jeherve:
> Just chiming in to say that it's not an experimental feature anymore as of [WP 6.2](https://wordpress.org/documentation/wordpress-version/version-6-2/#custom-css). 🙂

Sweet, I couldn't find it at first, but finally figured out how to get there!

rcrdortiz:
Thanks @jeherve the changes worked great!

jeherve (commented):
This works as expected for me, so should be good to merge! 

Before we do that though and following on my remark above, I would be curious to have your take on an alternative, that I pushed in #31601, where we would display a notice and allow folks to turn off the feature if they do not use it.
Let me know what you think!

rcrdortiz:
> This works as expected for me, so should be good to merge!
> 
> Before we do that though and following on my remark above, I would be curious to have your take on an alternative, that I pushed in #31601, where we would display a notice and allow folks to turn off the feature if they do not use it. Let me know what you think!

It makes sense to give the user the information about what's happening with the toggle, I'm going to add just one small detail to your alternative, I'll provide a diff for it.

jeherve:
> It makes sense to give the user the information about what's happening with the toggle, I'm going to add just one small detail to your alternative, I'll provide a diff for it.

Thank you! Feel free to commit to my PR directly if you'd like!

mmtr:
I tested with a site that uses a block theme and doesn't have any Additional CSS code entered in the Customizer. However the toggle still shows up.

Screenshot 2023-06-28 at 17 35 54

Would it be possible to hide it in these cases (or automatically turn off the `custom-css` module) and show the button that links to the site editor?

I'm also unsure about the copy. The "Hurray!" sounds way more informal than the copy in the rest of the sections. 

---

Also, a minor thing when using classic themes. Seems there is more separation that needed between the text and the toggle. I guess it'd look better with a lower margin-top in the toggle (or a lower margin-bottom in the text):

Screenshot 2023-06-28 at 17 42 31

jeherve:
> Would it be possible to hide it in these cases (or automatically turn off the custom-css module) and show the button that links to the site editor?

I don't think we can. This would effectively block someone from using the Custom CSS module; if they were to activate the module via another way (like via the old modules list or via CLI) it would get automatically disabled right away, before they get the chance to add in their custom CSS.

I don't know if many folks will be in that situation though. The module is not automatically enabled, so folks who have it enabled on their site today enabled it on purpose at some point in the past, so they're likely to have custom CSS filled in in my opinion.

If we opted to force disable it that way anyway, it may end up being expensive since we would have to call `wp_get_custom_css()` to get that information every time we load the dashboard.

> I'm also unsure about the copy. The "Hurray!" sounds way more informal than the copy in the rest of the sections.

Yeah, I could see how it's not as consistent with the rest of the interface. Do you have any suggestions there?

> a minor thing when using classic themes. Seems there is more separation that needed between the text and the toggle. I guess it'd look better with a lower margin-top in the toggle (or a lower margin-bottom in the text):

Good catch, I had forgotten to consider that use-case. I removed the extra padding in fc853ceb67680a4956c68c37a9c5bda9f5f0917a

image

rcrdortiz:
> I'm also unsure about the copy. The "Hurray!" sounds way more informal than the copy in the rest of the sections.

The copy is the same one that we are currently displaying on the customizer. I think it makes sense for them to be the same in order to avoid confusion. Though I do see your point about it being informal (I also thought the customizer copy is slightly informal), maybe it could make sense to change the copy in the customizer too.

mmtr:
> The module is not automatically enabled, so folks who have it enabled on their site today enabled it on purpose at some point in the past, so they're likely to have custom CSS filled in in my opinion.

Yeah, I didn't know that was the case for Jetpack sites. But I think it's automatically enabled for WoA sites (see fbhepr%2Skers%2Sjcpbz%2Sjc%2Qpbagrag%2Syvo%2Sngbzvp%2Spynff%2Qwrgcnpx%2Qznantre.cuc%3Se%3Q0ppqr343%2328-og), so at least for WP.com Atomic sites we'll likely have more users without custom CSS code added.

Maybe the proper fix for WoA sites would be to keep the module disabled during the AT transfer if there is a block theme active.

> I could see how it's not as consistent with the rest of the interface. Do you have any suggestions there?

Maybe "Your site has a block theme that allows you to apply custom CSS from the Site Editor. [Learn more](https://wordpress.org/documentation/article/styles-overview/#applying-custom-css)"

> The copy is the same one that we are currently displaying on the customizer. I think it makes sense for them to be the same in order to avoid confusion.

IMO the notice shouldn't aim at being consistent with the Customizer. It should be consistent with the rest of the interface where it's currently displayed.

---

Also, another thing I just noted. The tooltip we display when hovering the "i" icon is no longer accurate when the module is disabled for a site with a block theme:

Screenshot 2023-06-29 at 09 33 06

Perhaps we could remove it?

mmtr:
Anyway, all the feedback I mentioned above can be addressed in follow-ups.

mmtr (approved):

jeherve:
> Maybe the proper fix for WoA sites would be to keep the module disabled during the AT transfer if there is a block theme active.

Yes, that seems like a good idea. I don't think automatically enabling Custom CSS makes sense in today's world anymore.

> all the feedback I mentioned above can be addressed in follow-ups.

I'll submit a follow-up PR for those changes, they all make sense to me. 👍

jeherve:
> I'll submit a follow-up PR for those changes, they all make sense to me. 👍

Done in #31634.

PR #31416, opened 2023-06-16
AI Assistant: collect, store and send prompt history

retrofox:
This PR aims to collect, store, and pass the prompt history to the AI endpoint.

* Register a new promptHistory attribute
* Combine the prompt generated by the library with the history
* Send the history when hitting the endpoint 
* Limit the number of prompt items to 20 (hardcoded for now)
* It tries to catch [422](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) which happens when the URL is too long. If so, it removes items from the history to be able to retry the request

### Follow-up issues

* Compute the prompt size depending on tokens and URL length
* Keep iterating over the Prompt list. The API is still tough to use.

Fixes #

## Proposed changes:

* AI Assistant: collect, store and send prompt history

### Other information:

## Testing instructions:

* Go to the block editor
* Start to generate content.
* Use the de console to see the prompt data

Screenshot 2023-06-16 at 13 36 34

* Keep iterating over the same block instance. You can check the block attribute:
image

* Finally, confirm the block works as expected.
* Check block generation options.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-suggestions-from-openai/index.js
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-suggestions-from-openai/index.js
@@ -195,6 +195,14 @@ const useSuggestionsFromOpenAI = ( {
 				content: assistantResponse,
 			} );
 
+			/*
+			 * Limit the prompt history to 20 items.
+			 * @todo: limit the prompt based on tokens.
+			 */
+			if ( updatedPromptHistory.length > 20 ) {
+				updatedPromptHistory.splice( 0, updatedPromptHistory.length - 20 );
+			}
+
 			stopSuggestion();
 
 			updateBlockAttributes( clientId, {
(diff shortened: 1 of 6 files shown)


projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-suggestions-from-openai/index.js
```diff
@@ -14,7 +14,8 @@ import { buildPromptForBlock } from '../../lib/prompt';
 import { askJetpack, askQuestion } from '../../lib/suggestions';
 import { getContentFromBlocks, getPartialContentToBlock } from '../../lib/utils/block-content';
 
-const debug = debugFactory( 'jetpack-ai-assistant:event:fullMessage' );
+const debug = debugFactory( 'jetpack-ai-assistant:event' );
+const debugPromt = debugFactory( 'jetpack-ai-assistant:prompt' );
```
dhasilva:
```suggestion
const debugPrompt = debugFactory( 'jetpack-ai-assistant:prompt' );
```
Small typo, would need an update on all calls

dhasilva (approved):
LGTM

PR #31418, opened 2023-06-16
CRM: API include tags in the contact endpoints responses

cleacos:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3154

## Proposed changes:
This PR adds the field `withTags` to allow the API contacts endpoint to retrieve contacts including tags:

- `/zbs_api/customers/`
- `/zbs_api/customer_search/`

### Other information:

## Jetpack product discussion
N/A

N/A

## Testing instructions:

- Use an application to make API queries
- Get/copy your JPCRM API keys from your installation: `api_key` and `api_secret`
- Make these queries:

GET `/zbs_api/customer_search?api_key=you_api_key&api_secret=your_api_secret&email=contact_email&tags=1`
- **In trunk,** you won't get the tags in the JSON response
- **In this branch,** you will get the contact tags.

POST `/zbs_api/customers?api_key=you_api_key&api_secret=your_api_secret` and in the body request, put this JSON content:

```
{
	"tags": true,
	"perpage": 3
}
```

- **In trunk,** you won't get the tags in the JSON response
- **In this branch,** you will get the contact tags.

### DIFF
--- a/projects/plugins/crm/api/customers.php
+++ b/projects/plugins/crm/api/customers.php
@@ -26,6 +26,8 @@ if ( ! defined( 'ZEROBSCRM_PATH' ) ) {
 // Ultimately this should be switched to GET, but the docs have it as POST, so best to wait for a rewrite
 // Also seems to mostly be a duplicate of customer_search, other than not being able to search by email...
 
+// phpcs:disabled WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
+
 global $zbs;
 
 $json_params     = file_get_contents( 'php://input' );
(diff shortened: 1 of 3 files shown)


cleacos:
Auto-merge enabled

projects/plugins/crm/api/customers.php
```diff
@@ -65,6 +67,11 @@
 	$companyID = (int) $customer_params['company'];
 }
 
+$withTags = false;
+if ( isset( $customer_params['tags'] ) ) {
+	$withTags = true;
+}
```
kallehauge: Technically, if someone provides "false" as the parameter value, this would still include tags because we just look for its existence
```suggestion
$withTags = false;
if ( isset( $customer_params['tags'] ) && $customer_params['tags'] ) {
	$withTags = true;
}
```

kallehauge (changes requested):
Just the one remark. Other than that, then the PR tests well and looks good 😄

projects/plugins/crm/api/customers.php
cleacos: Yes! It'll work with any value distinct to `0`, `false` or `null`

Also, this way works too:

```
if ( isset( $customer_params['tags'] ) ) {
	$withTags = (bool) $customer_params['tags'];
}
```

kallehauge: Either solution is 💯 with me as long as we respect "falsy values" since that's how engineers would probably understand it 😄

kallehauge (approved):
LGTM 👍

PR #31419, opened 2023-06-16
CRM: Fix learn menu buttons overlapping

gogdzl:
## Proposed changes:

This PR fixes an alignment issue in the CRM Learn menu.
It is a simple fix wrapping its contents when it is not big enough to fit all buttons.
This fix looks good enough for the moment, a more elaborate fix should be done when fixing the CRM for mobile (small screen) devices.

### Other information:

## Jetpack product discussion
https://github.com/Automattic/zero-bs-crm/issues/3147

No.

## Testing instructions:

* Open the Edit Invoice page
* Make the window small enough to make the buttons overlap. (This should happen near 1519px wide)

In `trunk` the buttons will overlap.
image

In `fix/crm/3147-learn-menu-buttons-overlapping` the buttons will wrap.
image


cleacos:
What about if we include the View contact button along with the other buttons instead of allowing text-wrap? Keeping them aligned.

Medium-small screen:
image

Mobile screen:
image


```
<div class="jpcrm-learn-menu-subdiv-25">
   <a target="_blank" style="margin-left:6px;" class="zbs-invoice-quicknav-contact jpcrm-button" href="https://your-site.com/wp-admin/admin.php?page=zbs-add-edit&amp;action=edit&amp;zbstype=contact&amp;zbsid=1">View contact</a>
   <a href="https://your-site.com/wp-admin/admin.php?page=zerobscrm-plugin-settings&amp;tab=invbuilder" class="jpcrm-button white-bg font-14px" title="Invoice settings">Invoice settings</a>
   <a href="https://your-site.com/wp-admin/admin.php?page=zerobscrm-plugin-settings&amp;tab=bizinfo" class="jpcrm-button white-bg font-14px" title="Business settings">Business settings</a>				
</div>
```

gogdzl:
> What about if we include the View contact button along with the other buttons instead of allowing text-wrap? 

Can't be done that easily because the buttons are located in two different divs, and with two different parents.

I just noticed another problem in another page. We will have to put this on pause for a bit.

cleacos:
Diego suggests postponing this PR, and not including it in the release. We agree.

gogdzl:
TIL

**What's flex-grow, flex-shrink, and flex-basis?**

In CSS Flexbox, `flex-grow`, `flex-shrink`, and `flex-basis` are properties that control the sizing and distribution of flex items within a flex container.

- `flex-grow`: It defines the ability of a flex item to grow and occupy additional space in the flex container. It accepts a non-negative unitless value. By default, all flex items have a `flex-grow` value of `0`, meaning they won't grow beyond their initial size. When multiple flex items have a positive `flex-grow` value, the available space is distributed among them proportionally based on their respective `flex-grow` values. For example, if one item has a `flex-grow` value of `2`, and another has `flex-grow` set to `1`, the first item will grow twice as much as the second item when there is extra space.

- `flex-shrink`: It determines the ability of a flex item to shrink when the available space in the flex container is limited. It also accepts a non-negative unitless value. By default, flex items have a `flex-shrink` value of `1`, allowing them to shrink proportionally. When the total size of all flex items exceeds the size of the flex container, the shrinking behavior is determined by their respective `flex-shrink` values. For instance, if one item has a `flex-shrink` value of `3`, and another has `flex-shrink` set to `1`, the first item will shrink three times as much as the second item.

- `flex-basis`: It specifies the initial size of a flex item before any remaining space is distributed. It accepts various CSS length values, such as pixels, percentages, or `auto`. By default, flex items have a `flex-basis` value of `auto`, which means their initial size is determined by their content or other layout properties. You can set a specific length value for `flex-basis` to establish the initial size of a flex item.

Together, these properties allow you to create flexible and responsive layouts. By adjusting the values of `flex-grow`, `flex-shrink`, and `flex-basis`, you can control how flex items grow, shrink, and occupy space within a flex container, adapting to different screen sizes and content proportions.

gogdzl:
I think this solves all our current problems!

Could you try out the new code?

kallehauge (changes requested):
There's a very minor spacing issue with the button in what is still considered a reasonable full width screen size.

AFAIK then adding 10px margin right to either the button or the container div is sufficient.
The only caveat by doing so is that we're messing with the full width alignment compared to the main content, but we're not hitting it 100% to begin with, so I would personally consider that feature parity and call it a day for now 😅 

## Screenshots

### Current look

Screenshot 2023-07-12 at 12 28 34

### Adding 10px margin right

Screenshot 2023-07-12 at 12 28 26

---

### Full width difference

#### With 10px margin right

Screenshot 2023-07-12 at 12 33 42

#### Without 10px margin right

Screenshot 2023-07-12 at 12 28 41

gogdzl:
I've made changes so that margins are kept even at smaller resolutions.
The changes broke mobile resolutions, so I added some extra css to address that and now everything should look the same as before.

kallehauge (approved):
LGTM! 🚀 🎸

PR #31420, opened 2023-06-16
Fix link to Nginx instructions in Super Cache

Dan-Q:
## Proposed changes:

When used on Nginx, the Settings > Advanced page Cache Delivery Method setting contains a link to Nginx configuration instructions, but the link is outdated.

image

The existing link points to a removed codex page, which contains a link to the right page... but the anchor is lost. This change corrects the link.

This PR was originally opened at https://github.com/Automattic/wp-super-cache/pull/846, but a robot told me it should be here

### White box

Examine the old and new URLs and see what I mean.

### Black box

1. Check out this branch into a WP installation on Nginx.
2. Visit Settings > WP Super Cache > Advanced > Cache Delivery Method and click the "here" link (incidentally, it might be nice to not use a "here" link, they're bad for accessibility: perhaps we should have a sentence like "Nginx is not officially supported, but [documentation for configuring Nginx for static caching] is available"
3. See the right page

No.


jeherve (changes requested):
Although you're only editing an existing link, your changes now trigger PHPCS errors. We would need to fix before we can merge this PR. I'll do that now to save you the trouble.

jeherve (approved):

jeherve:
Should be good to merge now. Thanks for the contribution!

dilirity:
PR was stuck so I merged with latest trunk.

EDIT: Disabling auto-merge to make sure all checks pass before we merge.

PR #31421, opened 2023-06-16
AI Assistant: Separate rules prompt message and refactor title props check

dhasilva:
Fixes #
This PR is aimed at making the responses better, with less "Sure! Here it is" kind of answers. It also improves the title handling, disabling the Markdown rule for that case.

## Proposed changes:

* IMPORTANT: Changes the AI context
* Separates the rules from the `system` context message to a separate `user` message
* Removes `isGeneratingTitle` prop
* Refer to `const`s instead of strings when dealing with prompt types
* Tweaks the prompt/rules

### Other information:

## Testing instructions:

* Use the AI Assistant block and check all cases. It should work as normal with better results


projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
```diff
@@ -130,21 +175,21 @@ export const buildPromptTemplate = ( {
...
-	isGeneratingTitle?: boolean;
+	isTitlePrompt?: boolean;
 } ): Array< PromptItemProps > => {
 	if ( ! request && ! relevantContent ) {
 		throw new Error( 'You must provide either a request or content' );
 	}
 
 	// Add initial system prompt.
-	const messages = [ getInitialSystemPrompt( { rules } ) ];
+	const messages = [ getInitialSystemPrompt() ];
```
retrofox, line 192: Why did you remove the ability to extend the system rules?

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
```diff
@@ -44,50 +44,95 @@ export type PromptItemProps = {
...
-- If you cannot generate a meaningful response to a user’s request, reply with “__JETPACK_AI_ERROR__“. This term should only be used in this context, it is used to generate user facing errors.
-`;
+	let formattedRules = '';
+	formattedRules = promptRules.map( rule => `- ${ rule }.` ).join( '\n' );
+
+	const prompt = `Strictly follow these rules:
+${ formattedRules }`;
 
-	return { role: 'system', content: prompt };
+	return { role: 'user', content: prompt };
```
retrofox: What's the motivation behind replacing the role here? These rules do seem to belong to the `system` role.

image

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
dhasilva: Because I separated the rules in another message, so they were moved [here](https://github.com/Automattic/jetpack/pull/31421/files#diff-49fc00a7ab90da5972cc0277af7a647b8059ae846c1952515f7a1cad8be20652R215).

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
dhasilva: It seems to respect the rules better if they are in a `user` message. Did you notice any change?

retrofox: I dont :-(

retrofox (approved):
I tested the generated responses and they work as usual.
There is an improvement in removing the markdown when generating the post title.
I got the `Sure!...` undesired text as part of the response :-/

PR #31425, opened 2023-06-19
Component: Add Pricing Slider component

### LINKED ISSUE #31345: Paid Stats / Components: Pricing Slider
Please refer to the implementation on Akismet to accelerate the development: https://akismet.com/plan/personal/

Typical props would include: `defaultValue`, `value`, `minStep`, `minValue`, `maxValue`, `onChange`, `sliderColor` etc.

Image

dognose24:
Fixes #31345 

## Proposed changes:

* Add component `Pricing Slider` with third-party library [ReactSlider](https://zillow.github.io/react-slider/#reactslider).
* Add storybook use cases for operations.

### Other information:

## Jetpack product discussion

p1HpG7-nar-p2

N/A.

## Testing instructions:

* Run the local storybook with `cd projects/js-packages/storybook && npm run storybook:dev`.
* Ensure the `Pricing Slider` component under `JS Packages` works well.

截圖 2023-06-19 下午12 04 03


projects/js-packages/components/components/pricing-slider/index.tsx
```diff
@@ -0,0 +1,53 @@
+import classNames from 'classnames';
+import React from 'react';
+import ReactSlider from 'react-slider';
+import { PricingSliderProps } from './types';
+import './style.scss';
```
kangzj, line 5: Can we please change it to use CSS module: https://github.com/Automattic/jetpack/blob/49b12cd171952f0ab65c5a2d92901b65644de073/projects/js-packages/components/components/action-button/index.jsx#L13

kangzj:
I fixed the build with c90c302 and I think we should use CSS module just like other components to avoid CSS conflicts. Commented inline.

It otherwise works well for me! You are a legend @dognose24 👍

projects/js-packages/components/components/pricing-slider/style.scss
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// Base styles
+.pricing-slider {
+	width: 100%;
+    height: $thumb-height;
+
+    .pricing-slider__track {
+        @include adjust-track-rail-styles;
+        top: calc(($thumb-height / 2) - ($track-height / 2));
```
jsnmoon: I think we need to fix the indentation style in this file, which uses both spaces and tabs.

projects/js-packages/components/components/pricing-slider/style.scss
```diff
@@ -0,0 +1,55 @@
...
+        letter-spacing: -0.02em;
+        white-space: nowrap;
+        cursor: pointer;
+
+        &.pricing-slider__thumb--is-active {
+            // On focus styling
+            outline: none;
+        }
+    }
+}
```
jsnmoon: Missing an empty line at the end.
```suggestion
}

```

projects/js-packages/components/components/pricing-slider/types.ts
```diff
@@ -0,0 +1,49 @@
...
+	/**
+	 * Callback called only after moving a thumb has ended. The callback will only be called if the action resulted in a change.
+	 * The function will be called with two arguments, the first being the result value(s) the second being thumb index.
+	 */
+	onAfterChange?: ( value: number ) => void;
+
+	/**
+	 * Node to render on the slider.
+	 */
+	renderThumb?: React.ReactNode;
```
jsnmoon: This needs to be updated to resolve a typing error.

jsnmoon: Pushed up an attempted fix with [332a5ea](https://github.com/Automattic/jetpack/pull/31425/commits/332a5ea8b2074d57ef0e7739c86ee49d2afbda09).

projects/js-packages/components/components/pricing-slider/style.scss
jsnmoon: Fixed in [2c43d33](https://github.com/Automattic/jetpack/pull/31425/commits/2c43d3362265d02461b4129abf4f87f093c47df9).

projects/js-packages/components/components/pricing-slider/style.scss
jsnmoon: Fixed in [2c43d33](https://github.com/Automattic/jetpack/pull/31425/commits/2c43d3362265d02461b4129abf4f87f093c47df9).

jsnmoon (dismissed):
Looks great! I pushed up two commits to resolve my review nits and the build failure. 🚢

projects/js-packages/components/components/pricing-slider/index.tsx
kangzj: `react-slider` appends `-0` and `-1` to for slider colours, so it's hard to leverage CSS modules. We agreed to change the class names with good / long BEM class names to avoid any conflicts.

dognose24:
> use CSS module just like other components to avoid CSS conflicts

@kangzj I agree that the CSS conflict could affect more than we imagine. However, CSS modules would not be adaptable to the library since the mechanism of dynamically determining class names for tracks (`${trackClassName}-${i}`).

I updated class names with more detailed BEM styles to avoid CSS conflict. [9cd3acf](https://github.com/Automattic/jetpack/pull/31425/commits/9cd3acfad066505968b7dce772a213e63aabc7b9)

cc @jsnmoon

kangzj (dismissed):
It works well for me. Great job! 👍

Let's try to fix the build errors and then we could ship 🐑

projects/packages/backup/webpack.config.js
```diff
@@ -17,6 +17,10 @@ module.exports = [
 		},
 		resolve: {
 			...jetpackWebpackConfig.resolve,
+			modules: [
+				'node_modules',
+				path.resolve( __dirname, './node_modules' ), // For core-js
```
anomiex: This should not be necessary. I'm guessing this is trying to work around the missing peer dep but with an inaccurate comment?

Please remove this (here and everywhere else you added it) when trying the fix I suggest in the top-level comment.

pnpm-lock.yaml
anomiex, line 1: You've apparently done something weird with the dependency updates here, there are way too many package versions being bumped here.

I recommend starting over here by making sure any dependencies you're adding on existing packages match those used already in use in the monorepo, do a trunk merge, then `git checkout origin/trunk pnpm-lock.yaml && pnpm install` to update this afresh.

Although you may want to wait until I finish getting #31308 merged before doing that, which I'm working on right now.

anomiex (dismissed):
It looks like the react-slider library you're trying to use has an undeclared peer dependency on `@babel/runtime`, similar to [this bug in Gutenberg](https://href.li/?https://github.com/WordPress/gutenberg/issues/41343). An upstream bug should be filed for that, and then you could do something like [this](https://github.com/Automattic/jetpack/blob/95b8fab93bde8973940e5fc397582673d8a9ef6e/.pnpmfile.cjs#L46-L54) (including a comment linking to the upstream bug) to work around the bug if you don't want to wait for upstream to release a fix.

dognose24:
@anomiex Thanks for your direction a lot! 🙇🏼‍♂️

> An upstream bug should be filed for that

There was an [upstream issue reported](https://github.com/zillow/react-slider/issues/284), but it seemed to be ignored by the author.

> then you could do something like [this](https://github.com/Automattic/jetpack/blob/95b8fab93bde8973940e5fc397582673d8a9ef6e/.pnpmfile.cjs#L46-L54) (including a comment linking to the upstream bug) to work around the bug

So, could we apply the workaround for now? 🙏🏼

anomiex:
> There was an [upstream issue reported](https://github.com/zillow/react-slider/issues/284), but it seemed to be ignored by the author.

It doesn't look ignored to me so much as that the person there found that, in a hoisted environment, adding the missing peer dep worked around it so they closed it.

A bug report explicitly pointing out that it's a missing peer dependency and that the workaround of just adding the dependency won't work with pnpm with hoisting disabled or with yarn's PnP would be good to try. Feel free to crib from the Gutenberg bug I linked.

> So, could we apply the workaround for now? 🙏🏼

Yes, but please do file the upstream bug for the comment to link to.

projects/packages/backup/webpack.config.js
jsnmoon: These changes have been reverted.

pnpm-lock.yaml
jsnmoon: This has been resolved.

jsnmoon:
This pull request should now be ready for another round of reviews.

kangzj (approved):
LGTM 👍

PR #31432, opened 2023-06-19
Jetpack plugin: add srcset to related posts thumbs

### LINKED ISSUE #13106: SrcSet for relatedposts thumbnails
According to your description about the thumbnails for related posts
(https://jetpack.com/support/related-posts/#details-on-related-post-thumbnails),
as well as my own findings, it seems you are only delivering a single image for
the related posts thumbnails.

Wouldn't it make sense to deliver multiple images in a srcset, and let the
clients determine which image they wanted to use?

sgomes:
Fixes #13106

## Proposed changes:

In an effort to move away from the `devicepx` script towards native functionality, this PR adds a `srcset` to related posts thumbnails (from the Jetpack module feature, not the block). The browser will automatically pick the most appropriate image from the options in the `srcset`, depending on the screen pixel ratio and the current zoom value.

* Add a `srcset` attribute to related post thumbnail API responses (`sizes` is not needed, since we're using pixel ratio multipliers instead of widths)
* Add the ability for the related posts script to set `srcset` (and `sizes`) if present in the API response

### Other information:

## Jetpack product discussion
None.

No.

## Testing instructions:
- Enable the "Related Posts" feature for your test blog (`Settings -> Traffic -> Show related content after posts`).
- Enable related posts thumbnails (`Show a thumbnail image where available` in the same place).
- Open a post that has related posts with thumbnails.
- Verify that the thumbnails include a `srcset` with different pixel ratios (up to 4x, if the source image is large enough).
- Optionally, validate that the variants that get fetched on page load change as you change zoom values and/or try different pixel ratio screens.

### DIFF
--- a/projects/plugins/jetpack/changelog/update-add-srcset-to-related-posts
+++ b/projects/plugins/jetpack/changelog/update-add-srcset-to-related-posts
@@ -1,0 +1,4 @@
+Significance: minor
+Type: enhancement
+
+Related posts: added srcset to thumbnails
--- a/projects/plugins/jetpack/modules/related-posts/jetpack-related-posts.php
+++ b/projects/plugins/jetpack/modules/related-posts/jetpack-related-posts.php
@@ -1501,7 +1501,9 @@ EOT;
 			);
 
 			if ( is_array( $post_image ) ) {
-				$img_url = $post_image['src'];
+				$img_url    = $post_image['src'];
+				$src_width  = $post_image['src_width'];
+				$src_height = $post_image['src_height'];
 			} elseif ( class_exists( 'Jetpack_Media_Summary' ) ) {
 				$media = Jetpack_Media_Summary::get( $post_id );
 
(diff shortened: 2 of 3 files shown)


jeherve (approved):
This seems to test well for me!

image


🚢

sgomes:
Thank you for the super quick review and approval, @jeherve! 👍

samiff:
While rolling out Jetpack 12.3-a.7 ([Slack](p1687278018237129-slack-C01U2KGS2PQ)) with this change to WoA sites, we picked up on an uptick of the following Warning:

```
PHP Warning:  Undefined array key "src_width" in /wordpress/plugins/jetpack/12.3-a.7/modules/related-posts/jetpack-related-posts.php on line 1505
```

@sgomes Would you be able to take a look at this so we can hopefully remove the noise from our logging?

sgomes:
Thank you, @samiff! 👍

I've added a bunch of checks in #31480, which should hopefully fix not only the new warnings, but some apparently pre-existing ones as well.

PR #31435, opened 2023-06-19
Boost: Fix image analysis report not rendering

### LINKED ISSUE #31434: Boost Image Size Analysis: occasionally not loading items
There are cases where the issues don't get loaded. They are fetched from the API, the UI just doesn't render them.

Image is from the Homepage tab.

CleanShot 2023-06-15 at 17 21 00@2x

dilirity:
Fixes #31434

## Proposed changes:

* Allow Image Size Analysis report items to not have an `edit_url`, which was causing the whole list of items to not render. Websites without front pages don't have an edit_url for their front-page.
* This needs expanding, as it just solves the problem with a missing property. Ideally, we would skip an invalid item from the batch of items for the page, and show the rest. I've opened an additional issue for that - https://github.com/Automattic/jetpack/issues/31436

### Other information:

## Jetpack product discussion

n/a

no

## Testing instructions:

**You need to have the Image Size Analysis feature enabled**

* Don't set a front-page for your website - leave it to `Your latest posts`;
* Make sure you have a large image on the front-page (you can add a featured image to the first post);
* Generate a report;

### Test that it breaks
* Open the Homepage tab of the report UI;
* It should show 10 placeholder images.

### Test that it works
* Apply the patch;
* The UI should properly show the front page as having an issue.

### DIFF
--- a/projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/store/zod-types.ts
+++ b/projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/store/zod-types.ts
@@ -25,9 +25,9 @@ export const ImageData = z.object( {
 	} ),
 	page: z.object( {
 		id: z.number(),
-		url: z.string().url().nullable(),
+		url: z.string().url(),
 		title: z.string(),
-		edit_url: z.string().url().nullable(),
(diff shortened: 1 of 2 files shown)


thingalon (approved):
👏

PR #31296, opened 2023-06-09
Tiled Galleries: fix dynamic property deprecation notices

jeherve:
Follow-up from 30786

## Proposed changes:

Fix more deprecation notices introduced with PHP 8.2:

```
PHP Deprecated:  Creation of dynamic property Jetpack_Tiled_Gallery_Rectangular_Item::$img_src is deprecated in modules/tiled-gallery/tiled-gallery/tiled-gallery-item.php on line 56
PHP Deprecated:  Creation of dynamic property Jetpack_Tiled_Gallery_Rectangular_Item::$img_src_grayscale is deprecated in modules/tiled-gallery/tiled-gallery/tiled-gallery-item.php on line 134
PHP Deprecated:  Creation of dynamic property Jetpack_Tiled_Gallery_Rectangular_Item::$size is deprecated in modules/tiled-gallery/tiled-gallery/tiled-gallery-item.php on line 136
```

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

Primary issue: #27927

## Does this pull request change what data or activity we track or use?

* No

## Testing instructions:

On a site connected to WordPress.com:

* Go to Posts > Add New and add a Tiled Gallery block.
* Ensure such galleries are displayed properly on the frontend.

PR #31438, opened 2023-06-19
Masterbar: avoid php warning with PHP 8.2

jeherve:
## Proposed changes:

- Follow-up to #30786 and #31296

### Other information:

## Jetpack product discussion

- Primary issue: 27927

* No

## Testing instructions:

* On a site running PHP 8.2, go to Jetpack > Settings > Writing and enable the WordPress.com toolbar (masterbar module).
* Check your logs; you should not see any warnings.

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-php82-dynamic-properties-masterbar
+++ b/projects/plugins/jetpack/changelog/fix-php82-dynamic-properties-masterbar
@@ -1,0 +1,4 @@
+Significance: patch
+Type: compat
+
+WordPress.com Toolbar: avoid PHP warnings with PHP 8.2.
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/changelog/fix-php82-dynamic-properties-masterbar
```diff
@@ -0,0 +1,4 @@
+Significance: patch
```
anomiex, line 1: Offtopic: This sort of PR is why semver probably doesn't make sense for Jetpack-the-plugin. Is this random field on this class part of a public API? If so, making it private should be 'major' significance. But chances are it's not supposed to be.

Anyway, I see nothing outside this class using it in the monorepo, or in wpdirectory, or in a8c internal codebases.

anomiex (approved):

PR #31447, opened 2023-06-19
AI Assistant: Stop event stream request when block gets deleted

### LINKED ISSUE #31390: AI Assistant: close source connection when the block is removed
When the AI Assistant block instance is removed from the block editor, the client keeps hitting the endpoint.

https://github.com/Automattic/jetpack/assets/77539/628f786f-602e-4c5c-bec6-517cdb771f76

Let's handle it closing the connection.

lhkowalski:
Fixes #31390.

## Proposed changes:

* Add a cleanup function to close the event stream source when the component gets unmounted, so the block stops the completion event stream when it's deleted

### Other information:

## Testing instructions:

* Go to the block editor
* Open your browser console to inspect the debug messages of the block (remember to enable it)
* Add an AI Assistant block
* Ask for some completion that may take a few seconds to complete (for example: "give me 5 long paragraphs about the history of software development")
* Wait for the completion to start, then go the three-dot menu of the block and click "Remove AI Assistant (Experimen...":

Screen Shot 2023-06-19 at 19 40 59

* Watch the debug logs on the console
* **Before the fix**, the completion will process until the end, and we can see the debug messages getting logged even after the block is gone (what is not expected):

https://github.com/Automattic/jetpack/assets/77539/628f786f-602e-4c5c-bec6-517cdb771f76

* **After the fix**, the completion will stop immediately (a `Cleaning things up...` debug message will be the last thing):

Screen Shot 2023-06-19 at 19 45 02


projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-suggestions-from-openai/index.js
```diff
@@ -93,6 +93,19 @@ const useSuggestionsFromOpenAI = ( {
...
 	}, [ loading ] );
 
+	useEffect( () => {
+		/*
+		 * Returning a cleanup function that will stop
+		 * the suggestion if it's still rolling.
+		 */
+		return () => {
+			debug( 'Cleaning things up...' );
+			if ( source?.current ) {
```
retrofox:
```suggestion
			if ( source?.current ) {
				debug( 'Cleaning things up...' );
```

projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-suggestions-from-openai/index.js
```diff
@@ -93,6 +93,19 @@ const useSuggestionsFromOpenAI = ( {
 		setIsLoadingCategories( loading );
```
retrofox, line 93: Can you put the clean-up code in the same hook instance? I think adding the proper comments can make it simpler.

projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-suggestions-from-openai/index.js
```diff
@@ -93,6 +93,19 @@ const useSuggestionsFromOpenAI = ( {
...
 
+	useEffect( () => {
+		/*
+		 * Returning a cleanup function that will stop
+		 * the suggestion if it's still rolling.
+		 */
+		return () => {
+			debug( 'Cleaning things up...' );
+			if ( source?.current ) {
+				source?.current?.close();
```
retrofox, line 102: There is a [stopSuggestion()](https://github.com/Automattic/jetpack/blob/64168db046efdd088c3eb60251b743201e7e820d/projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-suggestions-from-openai/index.js#L342-L352) function that closes the connection and also sets other states. We could reuse it.

projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-suggestions-from-openai/index.js
lhkowalski: Good catch.

projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-suggestions-from-openai/index.js
lhkowalski: Yes, doing it.

projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-suggestions-from-openai/index.js
lhkowalski: I tried to reuse it on a couple different ways:

- by calling it directly; I needed to add it to the `dependencies` array of the `useEffect` hook, and it resulted on the `useEffect` being called every render since the function `stopSuggestion` was redefined on every render, causing the effect to run every time and closing the suggestion early
- by moving it's logic to a function saved by `useCallback`, with the same results

Did I missed something?

retrofox: If I'm not wrong, the hook dependency is altered by the same hook itself. In this case, `loading` dependency is updated by the stopSuggestion() fn.

The fix for it is creating a new hook instance, as we did before, where it doesn't have that dependency:

```es6
useEffect( () => {
  stopSuggestion();
}, [] );
```

Something like ⬆️

retrofox (approved):
LGTM.
I added a suggestion to use the [stopSuggestion()](https://github.com/Automattic/jetpack/blob/64168db046efdd088c3eb60251b743201e7e820d/projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-suggestions-from-openai/index.js#L342-L352) function.
Sorry to add friction here.
Feel free to address it now, or we can iterate in follow-up tasks. 🚢

PR #31449, opened 2023-06-20
AI Extension: improve prompt when using the AI Assistant in extended blocks

retrofox:
This PR polishes the process that builds the prompt only for the AI Extension. It doesn't change the prompt for the AI Assistant block. For this:

* Add a few basic helpers functions to build prompt items
* Tries to preserve the language and tone when it's desired.
* Renames (again) some helper prompt functions

Fixes #

## Proposed changes:

* AI Extension: improve prompt when using the AI Assistant in extended blocks

### Other information:

## Testing instructions:

* Go to the block editor
* Create a paragraph/heading block instance 
* Change, again and again, the content of the block. The app will try to preserve the language, for instance
  1. Create content in English (English)
  2. Translate to another language (Spanish)
  3. Change the tone (Spanish)
  4. Summarize (Spanish)
The trace mentioned above doesn't work properly in trunk. The implementation suggested by this PR improves the process. 

https://github.com/Automattic/jetpack/assets/77539/4a8664ce-97cb-48bd-a363-21e2fdc5b6e5

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
@@ -24,7 +24,7 @@ import { PromptItemProps, PromptTypeProp, getPrompt } from '../../lib/prompt';
 export const withAIAssistant = createHigherOrderComponent(
 	BlockEdit => props => {
 		const { clientId } = props;
-		const [ storedPrompt, storePromptInLocalState ] = useState< Array< PromptItemProps > >( [] );
+		const [ storedPrompt, setStoredPrompt ] = useState< Array< PromptItemProps > >( [] );
 
 		const { updateBlockAttributes } = useDispatch( blockEditorStore );
 
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
@@ -58,7 +58,7 @@ export const withAIAssistant = createHigherOrderComponent(
 
 		const requestSuggestion = useCallback(
 			( promptType: PromptTypeProp, options: AiAssistantDropdownOnChangeOptionsArgProps ) => {
-				storePromptInLocalState(
+				setStoredPrompt(
 					getPrompt( promptType, {
 						...options,
 						content,
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
@@ -15,7 +15,7 @@ import AiAssistantDropdown, {
 } from '../../components/ai-assistant-controls';
 import AiAssistantPanel from '../../components/ai-assistant-panel';
 import useSuggestionsFromAI from '../../hooks/use-suggestions-from-ai';
-import { PromptItemProps, PromptTypeProp, buildPromptForExtensions } from '../../lib/prompt';
+import { PromptItemProps, PromptTypeProp, getPrompt } from '../../lib/prompt';
 
 /*
  * Extend the withAIAssistant function of the block
@@ -24,7 +24,7 @@ import { PromptItemProps, PromptTypeProp, buildPromptForExtensions } from '../..
 export const withAIAssistant = createHigherOrderComponent(
 	BlockEdit => props => {
 		const { clientId } = props;
-		const [ prompt, setPrompt ] = useState< Array< PromptItemProps > >( [] );
+		const [ storedPrompt, storePromptInLocalState ] = useState< Array< PromptItemProps > >( [] );
 
 		const { updateBlockAttributes } = useDispatch( blockEditorStore );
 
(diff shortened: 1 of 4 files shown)


projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
```diff
@@ -24,7 +24,7 @@ import { PromptItemProps, PromptTypeProp, buildPromptForExtensions } from '../..
 export const withAIAssistant = createHigherOrderComponent(
 	BlockEdit => props => {
 		const { clientId } = props;
-		const [ prompt, setPrompt ] = useState< Array< PromptItemProps > >( [] );
+		const [ storedPrompt, storePromptInLocalState ] = useState< Array< PromptItemProps > >( [] );
```
dhasilva:
```suggestion
		const [ storedPrompt, setStoredPrompt ] = useState< Array< PromptItemProps > >( [] );
```
To keep with convention

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
```diff
@@ -104,6 +104,87 @@ ${ postTitle?.length ? `- Current title: ${ postTitle }\n` : '' }${
...
+}
+
+function getSimplifyPrompt( {
+	content,
+	role = 'user',
+}: PromptOptionsProps ): Array< PromptItemProps > {
+	return [
+		{
+			role,
+			content: `Simplify the following text, using words and phrases that are easier to understand. Write the content in the same language as the original content:\n\n${ content }`,
```
dhasilva, line 133: Why not a rule to avoid repeating it?

retrofox: because sometimes we don't want to preserve the language and the tone. For instance, when changing the language and/or the tone :-D

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
```diff
@@ -315,30 +403,63 @@ export function buildPromptForBlock( {
...
-
 /**
  * Returns a prompt based on the type and options
  *
  * @param {PromptTypeProp} type           - The type of prompt.
  * @param {GetPromptOptionsProps} options - The prompt options.
  * @returns {Array< PromptItemProps >}      The prompt.
  */
-export function buildPromptForExtensions(
+export function getPrompt(
```
dhasilva, line 413: This function is only used for extensions and is independent from all the other functions. `getPrompt` in this case seems misleading. Maybe this should be moved to `/extensions/ai-assistant/lib`?

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
dhasilva: https://github.com/Automattic/jetpack/pull/31421/files#diff-49fc00a7ab90da5972cc0277af7a647b8059ae846c1952515f7a1cad8be20652R113

An idea going forward is to filter the default rules, so when a rule should be applied to _most_ prompts we could add it to the default rules and just turn it off when not needed.

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
```diff
@@ -104,6 +104,87 @@ ${ postTitle?.length ? `- Current title: ${ postTitle }\n` : '' }${
...
+}
+
+function getExpandPrompt( {
+	content,
+	role = 'user',
+}: PromptOptionsProps ): Array< PromptItemProps > {
+	return [
+		{
+			role,
+			content: `Expand the following text to about double its size approximately. Write the content in the same language as the original content:\n\n${ content }`,
```
dhasilva:
```suggestion
			content: `Expand the following text to about double its size. Write the content in the same language as the original content:\n\n${ content }`,
```
The "about" should be enough for the intended meaning.

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
```diff
@@ -315,30 +403,63 @@ export function buildPromptForBlock( {
...
-export function buildPromptForExtensions(
+export function getPrompt(
 	type: PromptTypeProp,
-	options: GetPromptOptionsProps
+	options: PromptOptionsProps
 ): Array< PromptItemProps > {
-	const promptText = promptTextFor( type, false, options );
+	const context =
+		'You are an excellent polyglot ghostwriter. ' +
+		'Your task is to help to the user to create and modify content based on their requests.';
```
dhasilva:
```suggestion
		'Your task is to help the user to create and modify content based on their requests.';
```

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
retrofox: It can be a nice enhancement 👍

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
retrofox: yes, let me do it in a follow-up please

dhasilva (approved):
LGTM. We'll have to think how to make these changes compatible with the conversation history in the block later, but the results in the extension are better.

retrofox:
> LGTM. We'll have to think how to make these changes compatible with the conversation history in the block later, but the results in the extension are better.

Let's create an issue for it.

PR #31450, opened 2023-06-20
 Properly Gate Newsletter Subscriptions

romarioraffington:
Fixes https://github.com/Automattic/gold/issues/64

## Proposed changes:

Properly Gate Newsletter Subscriptions. Currently, any user with any paid membership plan can access a "Paid Subscribers" newsletter gated post.

The fix can be found below. Note that in the last seconds of the video, I refreshed the browser several times to verify that I wouldn't get access to a Paid Newsletter gated post after subscribing to a premium content plan.

https://github.com/Automattic/jetpack/assets/7076981/6a570ba4-f92f-4858-9aaa-462d417e9c91

### Other information:

No

## Testing instructions:

* Subscribe via the premium content block in incognito
* Open a paid subscriber newsletter gated post in the same browser but a different tab to reuse the saved cookie
* Verify that subscribing to premium content doesn't give access to the newsletter-gated post by refreshing the newsletter-gated post.

**Note:** In the past, users could choose a newsletter paid plan as a paid plan option for premium content.  If your premium content block currently uses a newsletter-paid plan as an option, I suggest changing the plan to a non-newsletter plan before testing this PR.


romarioraffington:
Looks like a bunch of tests broke, I will look into this tomorrow

projects/plugins/jetpack/modules/memberships/class-jetpack-memberships.php
```diff
@@ -487,7 +501,7 @@ public static function user_can_view_post() {
 
 		require_once JETPACK__PLUGIN_DIR . 'extensions/blocks/premium-content/_inc/subscription-service/include.php';
 		$paywall       = \Automattic\Jetpack\Extensions\Premium_Content\subscription_service();
-		$can_view_post = $paywall->visitor_can_view_content( self::get_all_plans_id_jetpack_recurring_payments(), $post_access_level );
+		$can_view_post = $paywall->visitor_can_view_content( self::get_all_plans_id_jetpack_recurring_payments( self::$plan_type_newsletter ), $post_access_level );
```
millerf: What is your idea behind adding this parameter? get_all_plans_id_jetpack_recurring_payments() seems to only be used for Newsletters (Premium content block is using a different approach: fbhepr%2Skers%2Sjcpbz%2Sjc%2Qpbagrag%2Szh%2Qcyhtvaf%2Swrgcnpx%2Qcyhtva%2Sfgntvat%2Srkgrafvbaf%2Soybpxf%2Scerzvhz%2Qpbagrag%2S_vap%2Snpprff%2Qpurpx.cuc%3Se%3Qno0p29np%2385-og by getting the plan from the Block context.

The way you implemented it is fine but adds some complexity when it is not needed I think.

What about renaming `get_all_plans_id_jetpack_recurring_payments` to something more meaningful in the NL context instead?

Keep in mind we will revisit this as well once we unlock the different user tiers...

millerf (commented):
Added a small comment.

This would also need a small unit test that would add some more context...

millerf:
> Looks like a bunch of tests broke, I will look into this tomorrow

Some linting issue for some reason

```
Strict Standards: Static function Automattic\Jetpack\My_Jetpack\Product::get_name() should not be abstract in /tmp/wordpress-latest/src/wp-content/plugins/jetpack/jetpack_vendor/automattic/jetpack-my-jetpack/src/products/class-product.php on line 154
```

projects/plugins/jetpack/modules/memberships/class-jetpack-memberships.php
romarioraffington: > What is your idea behind adding this parameter? get_all_plans_id_jetpack_recurring_payments() seems to only be used for Newsletters (Premium content block is using a different approach:

@millerf I thought premium content was using non-newsletter plans, like [here](https://github.com/Automattic/jetpack/pull/31450/files#diff-d31988b4a8aab0212b66d79f7ff48aa585b26d40f59dabfe642f3be01c0cc47aL39). After reading the fn comment I see "newsletter" being mentioned but the function is in the premium-content folder

millerf: > I thought premium content was using non-newsletter plans, like here


Two things here:

First of all, I made a mistake, and I did not see your use of `plan_type_non_newsletter`. 
I did not see it because 
[class-wpcom-offline-subscription-service.php](https://github.com/Automattic/jetpack/pull/31450/files#diff-d31988b4a8aab0212b66d79f7ff48aa585b26d40f59dabfe642f3be01c0cc47a) actually gates sending newsletter (see here: fbhepr%2Skers%2Sjcpbz%2Snflap%2Qwbof%2Svapyhqrf%2Sfhofpevcgvba%2Qrznvyf%2Sfhofpevcgvba%2Qrznvyf.cuc%3Se%3Q123p982o%23406-og

But you are right, Premium content uses non-newsletters plans.
 
My understanding about `Premium_Content\subscription_service()` is that it is actually only gating Newsletters (both access to post and email sending). 
The different classes (Offline, Online, Token) only depends on the environment (Async jobs, WPCOM or Jetpack  - in that order, I think). 

That's my understanding, I might be wrong... There will soon be an opportunity to rename everything and makes sense of all that...

millerf: > but the function is in the premium-content folder

Yes, I am not exactly sure why it is there.

millerf: Oh. It is actually used here: https://github.com/Automattic/jetpack/blob/5e92e1c63281227afe235030ae4f7957646f7021/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/access-check.php#L98

To gate access to the Premium content block. But  `get_all_plans_id_jetpack_recurring_payments` is not used...

millerf: From the history:

https://github.com/Automattic/jetpack/commit/94ddae8284758db97529fad30752b9e11a556bb1#diff-bbd4ff6d14a8af458b4095fb0207187272efbf16d8e29301714d2a5c452c40e2L94

It looks like we piggy backed on the gating of Premium content block and extended it:

Screenshot 2023-06-20 at 16 42 00

So it seems all the different classes were here already (except `Online`).

And gating premium content on WPCOM when sending emails is used with `acces_check::current_visitor_can_access` again....

romarioraffington: Fixed in [f8cae35](https://github.com/Automattic/jetpack/pull/31450/commits/f8cae3533d66378c88f64d544371a4e6956d7903)

millerf (approved):
LGTM

PR #31456, opened 2023-06-20
Social Review Prompt: Ensure it is still shown with Jetpack Active

pablinos:
We had to revert #29910 in #30101 because the review prompt state was stamping on the rest of the initial state object that's used by Social/Publicize.

## Proposed changes:

This change adds the Post Publish panel with the review prompt to the publicize-components package and merges the required state in the Social plugin, so that if both plugins are active the review prompt is still shown in Jetpack.

### Other information:

## Jetpack product discussion

1198218726984184-as-1204393860935466/f

No.

## Testing instructions:
* The testing is the same as #27917 but for both Jetpack and Social plugins
* Check that you see the review prompt with Social active, but not with just Jetpack active
* Check that the prompt can be dismissed
* Check that the link takes you to review Social.


jeherve (dismissed):
It seems the tests are failing at the moment:
https://github.com/Automattic/jetpack/actions/runs/5327939776/jobs/9651984469?pr=31456

pablinos:
> It seems the tests are failing at the moment:

Yes, I didn't manage to solve the conflicts properly and it was some package version problem. I think it should be ok now, if you'd like to take another look @jeherve?

projects/plugins/social/src/class-jetpack-social.php
```diff
@@ -286,23 +288,30 @@ public function is_supported_post() {
...
 			class_exists( 'Jetpack' ) ||
-			! $this->is_supported_post()
+			! $this->should_enqueue_block_editor_scripts()
 		) {
 			return;
 		}
 
 		Assets::register_script(
-			'jetpack-social-editor',
+			'jetpack-blocks-editor',
```
jboland88: Nitpick: I assume this name `jetpack-blocks-editor` is the same as the main blocks editor script so that the extra data can be appended without needing to differentiate which one was loaded. AFAIK, it's good to have these handles be unique. I think it works in this case, but could be good to use a different handle.

It took me a while to remember that we check to see if Jetpack is loaded above, so I was confused how the name was the same with no conflict at first :)

projects/plugins/social/src/class-jetpack-social.php
```diff
@@ -342,14 +349,35 @@ class_exists( 'Jetpack' ) ||
...
+		}
+
+		$review_state = array(
+			'reviewRequestDismissed'   => self::is_review_request_dismissed(),
+			'dismissReviewRequestPath' => '/jetpack/v4/social/review-dismiss',
+		);
+
+		wp_add_inline_script(
+			'jetpack-blocks-editor',
+			sprintf( 'Object.assign( window.Jetpack_Editor_Initial_State.social, %s )', wp_json_encode( $review_state ) ),
```
jboland88, line 376: This is clever :)

jboland88 (dismissed):
This tested well and LGTM, nice solution.

I had one nitpick about registering the scripts with unique handle names, but it's not a blocker.

projects/plugins/social/src/class-jetpack-social.php
pablinos: Yeah it's so the inline script will get attached to the editor script whether we're loading the Social version or the Jetpack one.

Perhaps I can change the handle that we add the inline script to based on whether Jetpack is enabled.

projects/plugins/social/src/class-jetpack-social.php
pablinos: ...or hacky :)

projects/js-packages/publicize-components/src/components/post-publish-review-prompt/index.js
```diff
@@ -0,0 +1,57 @@
...
+import { PluginPostPublishPanel } from '@wordpress/edit-post';
+import { useCallback, useState } from '@wordpress/element';
+import usePublicizeConfig from '../../hooks/use-publicize-config';
+import { usePostStartedPublishing } from '../../hooks/use-saving-post';
+import useSocialMediaConnections from '../../hooks/use-social-media-connections';
+import ReviewPrompt from '../review-prompt';
+
+const PostPublishReviewPropmpt = () => {
+	const [ isReviewRequestDismissed, setIsReviewRequestDismissed ] = useState(
+		getJetpackData()?.social?.reviewRequestDismissed ?? true
```
jeherve, line 13: Since we do not have `reviewRequestDismissed` in the Jetpack plugin, this will always end up returning `true`. Is this okay?
https://github.com/Automattic/jetpack/blob/540064cf9bf3ac2618147e26b71c2c61f83bb3c2/projects/plugins/jetpack/class.jetpack-gutenberg.php#L726

Edit: and now I understand why you renamed Social's bundle below, you end up making that info available to the Jetpack plugin when the Social plugin is also available. I should have read better, my bad!

projects/js-packages/publicize-components/src/components/post-publish-review-prompt/index.js
```diff
@@ -0,0 +1,57 @@
...
+	usePostStartedPublishing( () => {
+		setShouldReviewRequestShow(
+			! isPostAlreadyShared && isPublicizeEnabled && hasEnabledConnections
+		);
+	}, [ isPostAlreadyShared, hasEnabledConnections, isPublicizeEnabled ] );
+
+	// Handle when the review request is dismissed
+	const handleReviewDismiss = useCallback( () => {
+		const reviewRequestDismissUpdatePath =
+			getJetpackData()?.social?.dismissReviewRequestPath ?? null;
```
jeherve, line 30: Same as above.

jboland88 (approved):
Thanks for the update,
I did not re-run the full test plan with the latest change, but I specifically tested to make sure that the additional properties for the review request were added to the initial state for social in the editor with and without the Jetpack plugin active 👍 

LGTM!

projects/js-packages/publicize-components/src/components/post-publish-review-prompt/index.js
pablinos: Ah I missed this comment somehow. Some weird GH UI stuff.

That's right, but I've ended up changing it back and then conditionally using the Social or Jetpack script name based on the active plugin.

PR #27961, opened 2022-12-15
Revue block: Part 1 of retiring the block - adding a placeholder message.

coder-karen:
Fixes #27806

#### Changes proposed in this Pull Request:

* This is the first part of retiring the Revue block. This involves outputting a message within the block in the editor, and on the front-end for admins only, explaining that Revue is shutting down.

#### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [x] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

#### Jetpack product discussion

Slack conversation: p1671036856312609-slack-C01U2KGS2PQ

Some further information about Revue shutting down here: p58i-dGf-p2 

#### Does this pull request change what data or activity we track or use?

No.

#### Testing instructions:

* To test, you will need to add a Revue block before applying this PR (as the PR also removes the ability to find the Revue block in the block inserter tool). It is ok if you don't have a Revue account - you should still see a block in the editor with a form to add you Revue username, and the front-end should not show anything in that case.
* Apply this PR. The editor should now show a message and link related to Revue shutting down, suggesting the block should be removed and adding a CTA to migrate subscribers via WordPress.com. On the front-end, if you are logged in as an admin you should see an info notice with similar messaging. View the same post / page whilst not logged in or not as an admin and you shouldn't see that notice.

PR #31463, opened 2023-06-20
Blocks: remove retired Revue block and add standardized way to display fallback block content

jeherve:
## Proposed changes:

The block was retired in #27961, when the service shut down back in December of last year. It's now been 6 months, we can consequently safely remove the block altogether.

As part of this PR, I'm introducing 2 new things:

1. A list of deprecated blocks, where we'll add any other blocks that we stop supporting in the future.
2. A function hooked into the `render_block` hook to display a custom message if someone happens to still be using a deprecated block. That custom message will only be displayed for editors+; site visitors will see nothing.

This setup will allow us to keep the frontend clean of any old code that may not be functional anymore. In the editor, however, we'll let WP Core handle things and display the usual message showing the block is not supported:

image

Here is how things will look like on the frontend:

**Logged out**

image

**Logged in**

image

### Other information:

## Jetpack product discussion

Internal reference: 1202723222637401-as-1203700397893924

* No

## Testing instructions:

2 main things to test here:

1. you should not find any more references to Revue in our codebase, and should not be able to insert a new Revue block on your site.
2. Existing Revue blocks have returned a warning in the past 6 months, now they'll display in 3 different ways:
    1. In the editor, you'll see an error if you still haven't removed them.
    2. On the frontend when logged in as an editor+, you'll see a message inviting you to remove the block.
    3. On the frontend when logged out, you wil see nothing.

To test this, you could try adding a Revue block to your site by pasting the following into the code editor view of the block editor:

```
<!-- wp:jetpack/revue {"revueUsername":"jeherve","emailLabel":"Email address or something"} -->
<div class="wp-block-jetpack-revue">
<!-- wp:jetpack/button {"element":"button","text":"Subscribe"} /-->
<a class="wp-block-jetpack-revue__fallback" href="https://www.getrevue.co/profile/jeherve">https://www.getrevue.co/profile/jeherve</a></div>
<!-- /wp:jetpack/revue -->
```


chrisbliss18 (approved):
After running `jetpack build` on this branch, I can confirm that the only references I found were in readme and changelog files. I also found a reference in `Jetpack_Gutenberg::deprecated_blocks`, but then I realized that is intentionally part of this update. :)

I rolled back to an old version of Jetpack to add a proper Revue block. After doing so, I was able to confirm all three of the editor and frontend changes when looking at content with the Revue block.

Thanks for the manual HTML code. That was useful for testing two of the items, but I wanted to confirm 2.i, and I needed a proper block for that.

Great work on the changes. I think your approach makes a lot of sense, and I like that there's now code in place for doing this again in the future if/when we need to.

Approved.

P.S. Thanks for the great PR description.

PR #31464, opened 2023-06-20
Akismet: update Jetpack replacement logo styles

bluefuton:
## Proposed changes:

A couple of years ago, we added code to replace the Akismet logo with the Jetpack logo in the Akismet plugin. This meant that the Jetpack logo would display instead if the user had a Jetpack subscription.

Screenshot 2023-06-20 at 16 37 00

The product at the time looked a little different:
* It was named "Jetpack Anti-Spam" with no reference to Akismet
* There was a plan to build separate configuration tools for Anti-Spam in preference to using the Akismet plugin

In 2023, we call the product "Jetpack Akismet Anti Spam" officially, so we're upfront that Akismet is powering this service. As far as I know we also have no plans to build a separate admin tool for Anti-Spam.

This PR starts showing the Jetpack Akismet Anti-Spam logo to customers of that product, rather than the generic Jetpack logo.

Screenshot 2023-07-06 at 3 32 44 PM

### Other information:

No.

## Testing instructions:
On a Jetpack development site, try signing up for a Jetpack Anti-Spam subscription and then visit `/wp-admin/admin.php?page=akismet-key-config`. You should always see the Jetpack Akismet Anti-Spam logo in the masthead.

On a fresh Jetpack development site, don't connect Jetpack and try activating the Akismet plugin via the Plugins page and visiting `/wp-admin/admin.php?page=akismet-key-config`. You should see the normal Akismet logo.


bluefuton:
I've discussed this with @robertbpugh and we're going to try it with the Jetpack Akismet Anti-Spam logo instead (thanks @ballio2000 for preparing the assets!).

Product=akismet-antispam, Type=logo-horizontal, Theme=positive

Product=akismet-antispam, Type=logo-horizontal, Theme=positive

bluefuton:
This is how the Anti-spam logo now looks on desktop and mobile:

Screenshot 2023-07-06 at 3 32 44 PM

Screenshot 2023-07-06 at 3 32 51 PM

bluefuton:
Todo: 
- [x] test with Akismet pre-5.2

bluefuton:
In pre-refresh Akismet versions, it looks like this (v5.1 pictured):

Screenshot 2023-07-07 at 10 15 58 AM

jeherve (approved):
This works in my tests and should be good to merge, but I wonder if in a follow-up PR, we could take that opportunity to fix what types of plans we look for before we overwrite the logo. We currently only look for Anti-spam plans, while I think we should also look for bundles that include Anti-spam, like the security bundle.

PR #31470, opened 2023-06-20
AI Extension: populate the prompt with previous messages

retrofox:
This PR populates the prompt with the last messages when the user generates changes by using the AI Extensions. For this:

* It slightly changes the hook `useSuggestionsFromAI()` API:
  The `request()` method accepts a prompt argument
* The `getPrompt()` accepts `prevMessages` options to populate the fresh prompt.
* `withAIAssistant()` collects and stores the previous messages in a local state.
  Data is not persistent. We don't want to touche the core blocks extending the attributes.

Fixes #

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to the block editor
* Create some content by using a paragraph/heading block
* Start to edit its content by using the AI Extension
* Take a look at the logs
* Confirm the prompt collects previous messages

image

image

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/components/ai-assistant-controls/index.tsx
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/components/ai-assistant-controls/index.tsx
@@ -125,7 +125,7 @@ export default function AiAssistantDropdown( {
 							iconPosition="left"
 							key={ `key-${ quickAction.key }` }
 							onClick={ () => {
-								onChange( quickAction.aiSuggestion, { contentType: 'generated' } );
+								onChange( quickAction.aiSuggestion );
 								closeDropdown();
 							} }
 							isSelected={ key === quickAction.key }
@@ -136,14 +136,14 @@ export default function AiAssistantDropdown( {
 
 					<ToneDropdownMenu
 						onChange={ tone => {
-							onChange( PROMPT_TYPE_CHANGE_TONE, { tone, contentType: 'generated' } );
+							onChange( PROMPT_TYPE_CHANGE_TONE, { tone } );
 							closeDropdown();
 						} }
 					/>
 
 					<I18nMenuDropdown
 						onChange={ language => {
-							onChange( PROMPT_TYPE_CHANGE_LANGUAGE, { language, contentType: 'generated' } );
+							onChange( PROMPT_TYPE_CHANGE_LANGUAGE, { language } );
 							closeDropdown();
 						} }
 					/>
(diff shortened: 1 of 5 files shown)


projects/plugins/jetpack/extensions/blocks/ai-assistant/components/ai-assistant-controls/index.tsx
```diff
@@ -125,7 +125,7 @@ export default function AiAssistantDropdown( {
 							iconPosition="left"
 							key={ `key-${ quickAction.key }` }
 							onClick={ () => {
-								onChange( quickAction.aiSuggestion, { contentType: 'generated' } );
```
retrofox, line 128: we don't need `options` arg here

projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
```diff
@@ -54,18 +64,63 @@ export const withAIAssistant = createHigherOrderComponent(
...
+					 * - Preserve the first item of the array (`system` role )
+					 * - Keep the last 4 messages.
+					 */
+
+					// Pick the first item of the array.
+					const firstItem = prevPrompt.messages.shift();
+
+					const messages: Array< PromptItemProps > = [
+						firstItem, // first item (`system` by default)
+						...prevPrompt.messages.splice( -3 ), // last 3 items
```
lhkowalski, line 81: I know it's an arbitrary number to prevent the history from growing too much, just need to make sure the responses will not become strange due to "truncating" the history. It's a matter of test, test, test.

lhkowalski:
Just tested the code it's very insteresting to see the history of the chat on it:

### First

Screen Shot 2023-06-21 at 12 07 02

### Second

Screen Shot 2023-06-21 at 12 07 37

### Third

Screen Shot 2023-06-21 at 12 08 03

### Fourth

Screen Shot 2023-06-21 at 12 18 20

This seems to have a good impact on the results, because every change I requested worked the way it should, respecting the languages and tone of the text.

My only concern is about repeating the content that we already have as `[assistant]` on the `[user]` message, but I know it's not related to this PR functionality and we can tweak it if we find a better way to do it, as we learn about engineering good prompts.

lhkowalski (approved):
Testing as expected. Code looks good. The decision about keeping last 3 messages is a good one, seems to give enough context to the assistant and keeps the interaction on a reasonable size (max of 6 messages in my tests).

PR #31471, opened 2023-06-20
Launchpad: Reuse completion method across all plugin codebase

andres-blanco:
## Proposed changes:

The idea behind this PR is to centralise the task completion logic, including updates to the `'launchpad_checklist_tasks_statuses'` option and triggering any relevant Tracks events, while preserving the existing `wpcom_mark_launchpad_task_complete()` API. 
Right now we have multiple code paths updating the option and triggering events, which is prone to error and can cause confusion.

From an implementation perspective, there are a number of key changes:
* We're introducing a new `wpcom_launchpad_update_task_status()` function that accepts an array with task IDs as keys and booleans as values to represent the task completion status. The task IDs can be id_map values or the "raw" task ID - the function is responsible for identifying the right underlying option key to store.
  - This code is responsible for updating the `'launchpad_checklist_tasks_statuses'` option and calling `wpcom_launchpad_track_completed_task()`.
  - With the centralisation, we are also updating our storage mechanism to only store truthy values, and to enforce that newly stored values are `true`.
  - The function returns an array of successful task IDs as keys with their new values as array values. This includes ignoring unknown task IDs, returning an empty array for no successful operations, and returning the requested task ID as a key even when that task is mapped to a different underlying task slug in the option via an `id_map`
* We're introducing a new `wpcom_mark_launchpad_task_incomplete()` function to support marking a task as incomplete.
* The existing REST API to update the option has been refactored to call `wpcom_launchpad_update_task_status()` directly.
* The existing `wpcom_mark_launchpad_task_complete` function has been refactored to call `wpcom_launchpad_update_task_status()`.

### Other information:

No

### Run local tests

* Download this branch to your local environment, run `composer phpunit` in the plugin root and verify test passes.
 
### Test the actual API

* Checkout this branch on your sandbox.
* Sandbox `public-api`.
* Create a new site with the build intent.
* Publish your site using the Stepper Launchpad and verify it works as expected.
* After that go to your site home and complete some of the tasks in the build checklist and verify it works as expected.

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -514,13 +514,14 @@ function wpcom_update_launchpad_task_status( $original_statuses ) {
 
 	$statuses = array();
 	foreach ( $original_statuses as $task_id => $value ) {
-		// Filter out non-existent tasks taking into account the id_map.
+		// Find out wether the task id sent is the actual task id or the id_map.
 		$actual_task_id = ( isset( $reverse_id_map[ $task_id ] ) ) ? $reverse_id_map[ $task_id ] : $task_id;
 		if ( ! isset( $task_definitions[ $actual_task_id ] ) ) {
+			// Bail if no task definition was found.
 			continue;
 		}
 
-		// Find the task id that should be used to store the status.
+		// If the task has an id_map, use that, otherwise use the actual task id as the key.
 		$task                        = $task_definitions[ $actual_task_id ];
 		$stored_task_id              = isset( $task['id_map'] ) ? $task['id_map'] : $actual_task_id;
 		$statuses[ $stored_task_id ] = $value;
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -512,14 +512,18 @@ function wpcom_update_launchpad_task_status( $original_statuses ) {
 	$task_definitions = wpcom_launchpad_get_task_definitions();
 	$reverse_id_map   = wpcom_launchpad_get_reverse_id_mappings();
 
-	// Filter out non-existent tasks taking into account the id_map.
 	$statuses = array();
 	foreach ( $original_statuses as $task_id => $value ) {
+		// Filter out non-existent tasks taking into account the id_map.
 		$actual_task_id = ( isset( $reverse_id_map[ $task_id ] ) ) ? $reverse_id_map[ $task_id ] : $task_id;
 		if ( ! isset( $task_definitions[ $actual_task_id ] ) ) {
 			continue;
 		}
-		$statuses[ $task_id ] = $value;
+
+		// Find the task id that should be used to store the status.
+		$task                        = $task_definitions[ $actual_task_id ];
+		$stored_task_id              = isset( $task['id_map'] ) ? $task['id_map'] : $actual_task_id;
+		$statuses[ $stored_task_id ] = $value;
 	}
 
 	$old_values = (array) get_option( 'launchpad_checklist_tasks_statuses', array() );
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -471,66 +471,97 @@ function wpcom_launchpad_get_reverse_id_mappings() {
 /**
  * Mark a task as complete.
  *
- * @param string|string[] $task_ids The task IDs that should be marked as complete.
+ * @param string $task_id The task ID that should be marked as complete.
  * @return bool True if the task was marked as complete, false otherwise.
  */
-function wpcom_mark_launchpad_task_complete( $task_ids ) {
-	$task_ids = is_array( $task_ids ) ? $task_ids : array( $task_ids );
-	$values   = array_fill_keys( $task_ids, true );
-
-	$result = wpcom_update_launchpad_task_status( $values );
-
-	// Record the completion event in Tracks.
-	foreach ( $task_ids as $key => $value ) {
-		wpcom_launchpad_track_completed_task( $key );
+function wpcom_mark_launchpad_task_complete( $task_id ) {
+	if ( empty( $task_id ) ) {
+		return false;
 	}
 
-	return $result;
+	$result = wpcom_launchpad_update_task_status( array( $task_id => true ) );
+
+	return isset( $result[ $task_id ] ) && $result[ $task_id ];
 }
 
 /**
- * Mark a task as incomplete. This method is here for backwards compatibility of
- * the Rest API that allowed marking a task as incomplete.
+ * Mark a task as incomplete.
  *
- * @param string|string[] $task_ids The task IDs that should be marked as complete.
- * @return bool True if the task was marked as complete, false otherwise.
+ * @param string $task_id The task ID that should be marked as incomplete.
+ * @return bool True if the task was marked as incomplete, false otherwise.
  */
-function wpcom_mark_launchpad_task_incomplete( $task_ids ) {
-	$task_ids = is_array( $task_ids ) ? $task_ids : array( $task_ids );
-	$values   = array_fill_keys( $task_ids, false );
+function wpcom_mark_launchpad_task_incomplete( $task_id ) {
+	if ( empty( $task_id ) ) {
+		return false;
+	}
 
-	return wpcom_update_launchpad_task_status( $values );
+	$result = wpcom_launchpad_update_task_status( array( $task_id => false ) );
+
+	return isset( $result[ $task_id ] ) && ! $result[ $task_id ];
 }
 
 /**
  * Updates task/s statuses. If a non-existent task is passed, it will be ignored.
  *
- * @param array $original_statuses Array of mappings [ task_id: string => status: bool ].
- * @return bool True if the option where the statuses are stored was updated, false otherwise.
+ * @param bool[] $new_statuses Array of mappings [ task_id: string => new_status: bool ].
+ * @return bool[] Return the new values of the requested statuses with the requested task ID as the key. This will be an empty array if the option update fails, and any ignored status values will not be returned.
  */
-function wpcom_update_launchpad_task_status( $original_statuses ) {
+function wpcom_launchpad_update_task_status( $new_statuses ) {
 	$task_definitions = wpcom_launchpad_get_task_definitions();
 	$reverse_id_map   = wpcom_launchpad_get_reverse_id_mappings();
 
-	$statuses = array();
-	foreach ( $original_statuses as $task_id => $value ) {
-		// Find out wether the task id sent is the actual task id or the id_map.
-		$actual_task_id = ( isset( $reverse_id_map[ $task_id ] ) ) ? $reverse_id_map[ $task_id ] : $task_id;
-		if ( ! isset( $task_definitions[ $actual_task_id ] ) ) {
-			// Bail if no task definition was found.
+	$statuses_to_update = array();
+	$response_statuses  = array();
+	$option_map         = array();
+
+	foreach ( $new_statuses as $requested_task_id => $new_status ) {
+		// Work out what the underlying task ID/option is.
+		if ( isset( $task_definitions[ $requested_task_id ] ) ) {
+			// Check if the requested task has an id_map.
+			if ( isset( $task_definitions[ $requested_task_id ]['id_map'] ) ) {
+				$resolved_task_id = $task_definitions[ $requested_task_id ]['id_map'];
+			} else {
+				$resolved_task_id = $requested_task_id;
+			}
+		} elseif ( isset( $reverse_id_map[ $requested_task_id ] ) ) {
+			$resolved_task_id = $reverse_id_map[ $requested_task_id ];
+		} else {
 			continue;
 		}
 
-		// If the task has an id_map, use that, otherwise use the actual task id as the key.
-		$task                        = $task_definitions[ $actual_task_id ];
-		$stored_task_id              = isset( $task['id_map'] ) ? $task['id_map'] : $actual_task_id;
-		$statuses[ $stored_task_id ] = $value;
+		$new_status = (bool) $new_status;
+
+		$statuses_to_update[ $resolved_task_id ] = $new_status;
+		$response_statuses[ $requested_task_id ] = $new_status;
+		$option_map[ $resolved_task_id ]         = $requested_task_id;
 	}
 
 	$old_values = (array) get_option( 'launchpad_checklist_tasks_statuses', array() );
-	$new_values = array_merge( $old_values, $statuses );
+	// Only store truthy values.
+	$new_values = array_filter(
+		array_merge( $old_values, $statuses_to_update )
+	);
 
-	return update_option( 'launchpad_checklist_tasks_statuses', $new_values );
+	$update_result = update_option( 'launchpad_checklist_tasks_statuses', $new_values );
+	if ( ! $update_result ) {
+		return array();
+	}
+
+	// Track task completion for newly completed tasks.
+	$maybe_newly_completed_tasks = array_filter( $statuses_to_update );
+
+	foreach ( $maybe_newly_completed_tasks as $task_id => $task_status ) {
+		if ( isset( $old_values[ $task_id ] ) ) {
+			// Task was already complete - no need to mark as complete again.
+			continue;
+		}
+		// Use the requested task ID for completion tracking.
+		$requested_task_id = $option_map[ $task_id ];
+
+		wpcom_launchpad_track_completed_task( $requested_task_id );
+	}
+
+	return $response_statuses;
 }
 
 /**
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -507,6 +507,10 @@ function wpcom_mark_launchpad_task_incomplete( $task_id ) {
  * @return bool[] Return the new values of the requested statuses with the requested task ID as the key. This will be an empty array if the option update fails, and any ignored status values will not be returned.
  */
 function wpcom_launchpad_update_task_status( $new_statuses ) {
+	if ( ! is_array( $new_statuses ) ) {
+		return array();
+	}
+
 	$task_definitions = wpcom_launchpad_get_task_definitions();
 	$reverse_id_map   = wpcom_launchpad_get_reverse_id_mappings();
 
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -524,7 +524,8 @@ function wpcom_launchpad_update_task_status( $new_statuses ) {
 				$resolved_task_id = $requested_task_id;
 			}
 		} elseif ( isset( $reverse_id_map[ $requested_task_id ] ) ) {
-			$resolved_task_id = $reverse_id_map[ $requested_task_id ];
+			// We have some tasks that are only an id_map, but not a standalone task ID.
+			$resolved_task_id = $requested_task_id;
 		} else {
 			continue;
 		}
@@ -542,7 +543,18 @@ function wpcom_launchpad_update_task_status( $new_statuses ) {
 		array_merge( $old_values, $statuses_to_update )
 	);
 
-	$update_result = update_option( 'launchpad_checklist_tasks_statuses', $new_values );
+	// Check for a no-op where we are not actually writing anything.
+	if ( $new_values === $old_values ) {
+		return $response_statuses;
+	}
+
+	if ( empty( $new_values ) ) {
+		// If the new value is empty, but we had values before, we need to delete the option.
+		$update_result = delete_option( 'launchpad_checklist_tasks_statuses' );
+	} else {
+		$update_result = update_option( 'launchpad_checklist_tasks_statuses', $new_values );
+	}
+
 	if ( ! $update_result ) {
 		return array();
 	}
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -480,22 +480,10 @@ function wpcom_mark_launchpad_task_complete( $task_ids ) {
 
 	$result = wpcom_update_launchpad_task_status( $values );
 
-<<<<<<< HEAD
 	// Record the completion event in Tracks.
-	wpcom_launchpad_track_completed_task( $key );
-=======
 	foreach ( $task_ids as $key => $value ) {
-		if ( defined( 'IS_WPCOM' ) && IS_WPCOM ) {
-			require_lib( 'tracks/client' );
-
-			tracks_record_event(
-				wp_get_current_user(),
-				'launchpad_mark_task_completed',
-				array( 'task_id' => $key )
-			);
-		}
+		wpcom_launchpad_track_completed_task( $key );
 	}
->>>>>>> c51e097aa6 (reuse completion method across all plugin codebase)
 
 	return $result;
 }
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -451,33 +451,93 @@ function wpcom_launchpad_track_completed_task( $task_id, $extra_props = array() 
 }
 
 /**
- * Mark a task as complete.
+ * Returns a list of mappings from 'id_map' to the actual task id.
  *
- * @param string $task_id The task ID.
- * @return bool True if the task was marked as complete, false otherwise.
+ * @return array
  */
-function wpcom_mark_launchpad_task_complete( $task_id ) {
+function wpcom_launchpad_get_reverse_id_mappings() {
 	$task_definitions = wpcom_launchpad_get_task_definitions();
 
-	// If the task ID isn't defined, return false.
-	if ( ! isset( $task_definitions[ $task_id ] ) ) {
-		return false;
+	$mapping = array();
+	foreach ( $task_definitions as $task_id => $value ) {
+		if ( ! isset( $value['id_map'] ) ) {
+			continue;
+		}
+		$mapping[ $value['id_map'] ] = $task_id;
 	}
 
-	// If the task has an id_map, use that instead.
-	$key = $task_id;
-	if ( isset( $task_definitions[ $task_id ]['id_map'] ) ) {
-		$key = $task_definitions[ $task_id ]['id_map'];
-	}
+	return $mapping;
+}
+/**
+ * Mark a task as complete.
+ *
+ * @param string|string[] $task_ids The task IDs that should be marked as complete.
+ * @return bool True if the task was marked as complete, false otherwise.
+ */
+function wpcom_mark_launchpad_task_complete( $task_ids ) {
+	$task_ids = is_array( $task_ids ) ? $task_ids : array( $task_ids );
+	$values   = array_fill_keys( $task_ids, true );
 
-	$statuses         = get_option( 'launchpad_checklist_tasks_statuses', array() );
-	$statuses[ $key ] = true;
-	$result           = update_option( 'launchpad_checklist_tasks_statuses', $statuses );
+	$result = wpcom_update_launchpad_task_status( $values );
 
+<<<<<<< HEAD
 	// Record the completion event in Tracks.
 	wpcom_launchpad_track_completed_task( $key );
+=======
+	foreach ( $task_ids as $key => $value ) {
+		if ( defined( 'IS_WPCOM' ) && IS_WPCOM ) {
+			require_lib( 'tracks/client' );
+
+			tracks_record_event(
+				wp_get_current_user(),
+				'launchpad_mark_task_completed',
+				array( 'task_id' => $key )
+			);
+		}
+	}
+>>>>>>> c51e097aa6 (reuse completion method across all plugin codebase)
 
 	return $result;
+}
+
+/**
+ * Mark a task as incomplete. This method is here for backwards compatibility of
+ * the Rest API that allowed marking a task as incomplete.
+ *
+ * @param string|string[] $task_ids The task IDs that should be marked as complete.
+ * @return bool True if the task was marked as complete, false otherwise.
+ */
+function wpcom_mark_launchpad_task_incomplete( $task_ids ) {
+	$task_ids = is_array( $task_ids ) ? $task_ids : array( $task_ids );
+	$values   = array_fill_keys( $task_ids, false );
+
+	return wpcom_update_launchpad_task_status( $values );
+}
+
+/**
+ * Updates task/s statuses. If a non-existent task is passed, it will be ignored.
+ *
+ * @param array $original_statuses Array of mappings [ task_id: string => status: bool ].
+ * @return bool True if the option where the statuses are stored was updated, false otherwise.
+ */
+function wpcom_update_launchpad_task_status( $original_statuses ) {
+	$task_definitions = wpcom_launchpad_get_task_definitions();
+	$reverse_id_map   = wpcom_launchpad_get_reverse_id_mappings();
+
+	// Filter out non-existent tasks taking into account the id_map.
+	$statuses = array();
+	foreach ( $original_statuses as $task_id => $value ) {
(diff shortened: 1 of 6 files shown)


sixhours:
Big thanks for tackling this Andy. Let's get the launch out of the way, then we can make testing this and implementing it a top priority for cleanup.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -394,35 +394,88 @@ function wpcom_launchpad_track_completed_task( $task_id, $extra_props = array()
...
-	$statuses         = get_option( 'launchpad_checklist_tasks_statuses', array() );
-	$statuses[ $key ] = true;
-	$result           = update_option( 'launchpad_checklist_tasks_statuses', $statuses );
+	$result = wpcom_update_launchpad_task_status( $values );
 
 	// Record the completion event in Tracks.
-	wpcom_launchpad_track_completed_task( $key );
+	foreach ( $task_ids as $key => $value ) {
+		wpcom_launchpad_track_completed_task( $key );
+	}
```
daledupreez: The result of the `array_fill_keys()` call is being put into `$values`, so I think we should be looping over the values in `$task_ids`.
```suggestion
	foreach ( $task_ids as $task_id ) {
		wpcom_launchpad_track_completed_task( $task_id );
	}
```

I also have a concern that we're performing validation of the tasks in `wpcom_update_launchpad_task_status()`, which means we might log the completion here despite the task not being processed. Can we move this logic to `wpcom_update_launchpad_task_status()` so we track completion when it is actually done?

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -394,35 +394,88 @@ function wpcom_launchpad_track_completed_task( $task_id, $extra_props = array()
...
+	return wpcom_update_launchpad_task_status( $values );
+}
+
+/**
+ * Updates task/s statuses. If a non-existent task is passed, it will be ignored.
+ *
+ * @param array $original_statuses Array of mappings [ task_id: string => status: bool ].
+ * @return bool True if the option where the statuses are stored was updated, false otherwise.
+ */
+function wpcom_update_launchpad_task_status( $original_statuses ) {
```
daledupreez: Nit: can we use `wpcom_launchpad_` as a prefix for this new function? Maybe `wpcom_launchpad_update_task_status()`?
Also, can we call the parameter `$new_statuses` to reflect the purpose?
```suggestion
function wpcom_launchpad_update_task_status( $original_statuses ) {
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -394,35 +394,88 @@ function wpcom_launchpad_track_completed_task( $task_id, $extra_props = array()
...
+		}
+
+		// If the task has an id_map, use that, otherwise use the actual task id as the key.
+		$task                        = $task_definitions[ $actual_task_id ];
+		$stored_task_id              = isset( $task['id_map'] ) ? $task['id_map'] : $actual_task_id;
+		$statuses[ $stored_task_id ] = $value;
+	}
+
+	$old_values = (array) get_option( 'launchpad_checklist_tasks_statuses', array() );
+	$new_values = array_merge( $old_values, $statuses );
```
daledupreez: As per my previous comments, it may be worth adding the tracking logic in this block of code. I think we also want to remove the values when the new value is false - should we filter out non-true values?
```suggestion
	$old_status_values = (array) get_option( 'launchpad_checklist_tasks_statuses', array() );
	$statuses = array();
	foreach ( $new_statuses as $task_id => $value ) {
		// Find out wether the task id sent is the actual task id or the id_map.
		$actual_task_id = ( isset( $reverse_id_map[ $task_id ] ) ) ? $reverse_id_map[ $task_id ] : $task_id;
		if ( ! isset( $task_definitions[ $actual_task_id ] ) ) {
			// Bail if no task definition was found.
			continue;
		}

		// If the task has an id_map, use that, otherwise use the actual task id as the key.
		$task                        = $task_definitions[ $actual_task_id ];
		$stored_task_id              = isset( $task['id_map'] ) ? $task['id_map'] : $actual_task_id;
		$statuses[ $stored_task_id ] =(bool) $value;

		// If we're completing a task for the first time, track that completion via the supplied task ID.
		if ( $value && empty( $old_status_values[ $stored_task_id ] ) ) {
			wpcom_launchpad_track_completed_task( $task_id );
		}
	}

	$new_values = array_merge( $old_values, $statuses );
	// Consider adding a filter for non-truthy values
	$new_values = array_filter( $new_values );
```

projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php
```diff
@@ -139,19 +139,23 @@ public function update_site_options( $request ) {
...
-					$launchpad_checklist_tasks_statuses_option = array_merge( $launchpad_checklist_tasks_statuses_option, $value );
-
-					foreach ( $value as $task => $task_value ) {
-						if ( $task_value ) {
-							// If we're marking a task as complete, the value should be truthy, so fire the Tracks event.
-							wpcom_launchpad_track_completed_task( $task );
+					$complete_values   = array();
+					$incomplete_values = array();
+
+					foreach ( $value as $task_id => $task_status ) {
```
daledupreez: Side note: should we verify that we have an array before we call the `foreach`?

projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php
```diff
@@ -139,19 +139,23 @@ public function update_site_options( $request ) {
...
 
-					if ( update_option( 'launchpad_checklist_tasks_statuses', $launchpad_checklist_tasks_statuses_option ) ) {
+					$updated_complete_tasks   = wpcom_mark_launchpad_task_complete( $complete_values );
+					$updated_incomplete_tasks = wpcom_mark_launchpad_task_incomplete( $incomplete_values );
+					if ( $updated_complete_tasks || $updated_incomplete_tasks ) {
 						$updated[ $key ] = $value;
 					}
+
 					// This will check if we have completed all the tasks and disable Launchpad if so.
 					wpcom_launchpad_checklists()->maybe_disable_launchpad();
```
daledupreez: Another side note: I think we should rename this function and the related functions to specify that they are for the fullscreen launchpad: i.e. `maybe_disable_fullscreen_launchpad()`
But that's for another PR.

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-definitions-test.php
```diff
@@ -0,0 +1,104 @@
...
+
+		wpcom_mark_launchpad_task_incomplete(
+			array(
+				'test_task2',
+				'test_task3',
+			)
+		);
+		$options = get_option( 'launchpad_checklist_tasks_statuses' );
+		$this->assertTrue( isset( $options['test_task2'] ) && ! $options['test_task2'] );
+		$this->assertTrue( isset( $options['test_task3'] ) && ! $options['test_task3'] );
```
daledupreez: I thought the existing behaviour is that we only set values in the task list to be `true`, and don't set any falsey values.

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-definitions-test.php
```diff
@@ -0,0 +1,104 @@
...
+				return array(
+					'test_task1'            => array(),
+					'test_task2'            => array(),
+					'test_task3'            => array(),
+					'test_task_with_id_map' => array(
+						'id_map' => 'test_task_id_map',
+					),
+				);
+			}
+		);
```
daledupreez, line 35: Should we clear out the `launchpad_checklist_task_statuses` option in this function between tests?
```suggestion
			delete_option( 'launchpad_checklist_tasks_statuses' );
		);
```

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-definitions-test.php
```diff
@@ -0,0 +1,104 @@
...
+
+		add_filter(
+			'wpcom_launchpad_extended_task_definitions',
+			function () {
+				return array(
+					'test_task1'            => array(),
+					'test_task2'            => array(),
+					'test_task3'            => array(),
+					'test_task_with_id_map' => array(
+						'id_map' => 'test_task_id_map',
```
daledupreez, line 31: This setup confuses me -- from my reading of the code, we should be requiring `'test_task_id_map'` to exist, but the unit test is passing, so I feel either the code, the test, or I am missing something!

daledupreez (commented):
I have a number of comments that should hopefully solidify this code.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
andres-blanco: I've fixed this but I didn't add the `array_filter` in the end to prevent having to unset an existing key if the task is being marked as incomplete. [This comment is also relevant](https://github.com/Automattic/jetpack/pull/31471#discussion_r1281973163)

projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php
andres-blanco: The `checklist_statuses` can only receive an array as value, it's guarded by the API definition: https://github.com/Automattic/jetpack/blob/1f0ccb72cdef7bacb3f0b112c249f5cf8b8ee4ea/projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php#L53

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-definitions-test.php
andres-blanco: The [current API behaviour](https://github.com/Automattic/jetpack/blob/trunk/projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php#L143) puts whatever is sent into the option, that's why I've done it this way.

Do you feel this should be changed?

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-definitions-test.php
andres-blanco: We are using worddbless that clears most of the WP state after each test runs: https://github.com/Automattic/wordbless/blob/0.4.2/src/BaseTestCase.php#L29. So it shouldn't be necessary.

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-definitions-test.php
andres-blanco: I think the confusion came from the key and the id map being really similar. I updated those to be really different.

The idea is to test that even if we mark a task complete using the array key (in this case `test_task_with_id_map`) it will be stored in the option using the id map (`THIS_IS_THE_MAP`).

projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php
andres-blanco: I've created this issue to track this: https://github.com/Automattic/wp-calypso/issues/80140

andres-blanco: This is already done & merged.

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-definitions-test.php
daledupreez: Yes, I think we shouldn't bloat the options data if we can avoid it, as the options data is aggressively cached, and we don't need the falsey values pushed into both the options table and the cache.

projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php
daledupreez: That makes sense! Resolving.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -424,34 +424,92 @@ function wpcom_launchpad_track_completed_task( $task_id, $extra_props = array()
...
+			// Bail if no task definition was found.
+			continue;
+		}
 
-	$statuses         = get_option( 'launchpad_checklist_tasks_statuses', array() );
-	$statuses[ $key ] = true;
-	$result           = update_option( 'launchpad_checklist_tasks_statuses', $statuses );
+		// If the task has an id_map, use that, otherwise use the actual task id as the key.
+		$task                        = $task_definitions[ $actual_task_id ];
+		$stored_task_id              = isset( $task['id_map'] ) ? $task['id_map'] : $actual_task_id;
```
daledupreez: I am not sure why we're not just using `$actual_task_id` here. This either needs a comment to explain why the second lookup is necessary, or we need to use the already-resolved `$actual_task_id`.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -424,34 +424,92 @@ function wpcom_launchpad_track_completed_task( $task_id, $extra_props = array()
...
-	}
+	$statuses               = array();
+	$original_status_values = (array) get_option( 'launchpad_checklist_tasks_statuses', array() );
 
-	// If the task has an id_map, use that instead.
-	$key = $task_id;
-	if ( isset( $task_definitions[ $task_id ]['id_map'] ) ) {
-		$key = $task_definitions[ $task_id ]['id_map'];
-	}
+	foreach ( $new_statuses as $task_id => $value ) {
```
daledupreez: Nit: it may be worth using `$new_value` instead of `$value` to be extra-explicit about what the value represents.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -424,34 +424,92 @@ function wpcom_launchpad_track_completed_task( $task_id, $extra_props = array()
...
+			$value &&
+			( empty( $original_status_values[ $stored_task_id ] ) || $original_status_values[ $stored_task_id ] === false )
+		) {
+			wpcom_launchpad_track_completed_task( $actual_task_id );
+		}
+	}
 
-	// Record the completion event in Tracks.
-	wpcom_launchpad_track_completed_task( $key );
+	$new_values = array_merge( $original_status_values, $statuses );
```
daledupreez: As noted in [this comment](https://github.com/Automattic/jetpack/pull/31471#discussion_r1285744413), I think we should avoid bloating the options data, so I think we should only save true values and filter out anything that isn't `true`.

projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php
```diff
@@ -139,19 +139,23 @@ public function update_site_options( $request ) {
...
+							$incomplete_values[] = $task_id;
 						}
 					}
 
-					if ( update_option( 'launchpad_checklist_tasks_statuses', $launchpad_checklist_tasks_statuses_option ) ) {
+					$updated_complete_tasks   = wpcom_mark_launchpad_task_complete( $complete_values );
+					$updated_incomplete_tasks = wpcom_mark_launchpad_task_incomplete( $incomplete_values );
+					if ( $updated_complete_tasks || $updated_incomplete_tasks ) {
 						$updated[ $key ] = $value;
 					}
```
daledupreez: Thought: do we want to return the specific items that were updated? By simply returning `$value`, we may be giving an incorrect picture of what actions were actually taken. This isn't necessarily blocking, but our additional validation means that the return values may not match the changes we're making.

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-definitions-test.php
```diff
@@ -0,0 +1,114 @@
...
+
+	/**
+	 * Tests wether the correct task ID is stored in the 'launchpad_checklist_tasks_statuses' option.
+	 * When tasks have an 'id_map' property, the 'id_map' value should be used as the task ID.
+	 */
+	public function test_correct_task_id_is_stored() {
+		wpcom_mark_launchpad_task_complete( 'test_task_with_id_map' );
+		$options = get_option( 'launchpad_checklist_tasks_statuses' );
+		$this->assertFalse( isset( $options['test_task_with_id_map'] ) );
+		$this->assertTrue( isset( $options['THIS_IS_THE_MAP'] ) );
```
daledupreez: I am really confused by this test and why it passes. Based on the following lines in `wpcom_launchpad_update_task_status()`, we should only mark tasks as complete when we have a defined task for the mapped task ID:
```php
		if ( ! isset( $task_definitions[ $actual_task_id ] ) ) {
			// Bail if no task definition was found.
			continue;
		}
```

In the test setup, there is no task with an ID of `THIS_IS_THE_MAP`, so I am confused as to how this works or should work. Should we have separate test cases to handle the following situations:
 1. Task is mapped to a valid task ID -- we should mark the underlying task as complete
 2. Task is mapped to an invalid task ID -- we should _not_ mark the underlying task as complete

projects/packages/jetpack-mu-wpcom/tests/php/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad-test.php
```diff
@@ -89,7 +89,7 @@ public function test_update_checklist_statuses() {
 
 		$values = array(
 			'domain_upsell_deferred' => true,
-			'site_launched'          => true,
+			'site_launched'          => false,
```
daledupreez, line 118: Could you explain why you made this change?

andres-blanco: Since I modified the code that was being used to store this values I thought changing this was a good opportunity to exercise both the `true` and `false` code paths.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
andres-blanco: TLDR; I've made this change to prevent changing the API response when being called with an `id_map` instead of a `task_id`.

This is what I meant this variables to represent:
* `actual_task_id` is the key in the task definition array. 
* `stored_task_id` is the key in the `launchpad_checklist_tasks_statuses` _option_.

[This preexisting test](https://github.com/Automattic/jetpack/blob/trunk/projects/packages/jetpack-mu-wpcom/tests/php/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad-test.php#L91) implied that the Rest API might be called with `id_maps` directly. This use case was supported by the API endpoint because it was storing directly whatever it was being sent, but now, since we are making everything go through the `wpcom_mark_launchpad_task_complete` function I had to make this changes.
The idea behind this code is that task_ids and id_maps can be used interchangeably, but the value stored in the option is always the `id_map`.
 
If we used the `actual_task_id` as key then that test will need to be changed [here](https://github.com/Automattic/jetpack/blob/trunk/projects/packages/jetpack-mu-wpcom/tests/php/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad-test.php#L106) to expect `domain_upsell` key instead of `domain_upsell_deferred`. While making this change is easier it would be changing the rest API response.

Writing this made me realize that changing the API response might not be such a bad idea, mostly because it should only be being used by WPCOM and it would make the code clearear. But I would need a confidence check on doing that. What do you think?

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-definitions-test.php
andres-blanco: Depending on the outcome of [this conversation](https://github.com/Automattic/jetpack/pull/31471#discussion_r1285871674) this might change.

projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php
andres-blanco: This is also related to the outcome of [this other conversation](https://github.com/Automattic/jetpack/pull/31471#discussion_r1285871674)

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-definitions-test.php
andres-blanco: I'll mark this as resolved because it's being discussed [here](https://github.com/Automattic/jetpack/pull/31471#discussion_r1285761913)

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
daledupreez: Apologies for the delay in getting back to you on this point.

At this stage, I am not sure what we are looking up to find `$actual_task_id`. As far as I am concerned, there are two relevant values: the requested task id, and the task option key. In the logic we have, we would end up ignoring the requested task ID in cases where another task maps to that ID, but the request was for the underlying task. (A current example of this is for `first_post_published` and `first_post_published_newsletter`.

So I think we should either use the requested task ID or the stored task ID as the returned key for that item. I would strongly lean towards using the requested task ID, as the internal storage should be irrelevant to the client.

To lay this out in a more concrete way, how does the following look?

```suggestion
	foreach ( $new_statuses as $requested_task_id => $new_value ) {
		// If the requested task is defined, resolve the right underlying task ID.
		if ( isset( $task_definitions[ $requested_task_id ] ) ) {
			// If the task is defined, we either need to use the `id_map`, or the task itself.
			$resolved_task_id = isset( $task_definitions[ $requested_task_id ]['id_map'] ) ? $task_definitions[ $requested_task_id ]['id_map'] : $requested_task_id;
		} elseif ( isset( $reverse_id_map[ $requested_task_id ] ) ) {
			// Otherwise we need to check if the task only has an ID map.
			// NOTE: This may not be possible, or might not be something we want to allow via the API. I think we only want clients to make API calls for true external tasks.
			$resolved_task_id = $reverse_id_map[ $requested_task_id ];
		} else {
			continue;
		}

		// $resolved_task_id is what we should use for storage, and $requested_task_id is what we'd return to the client.
```

projects/packages/jetpack-mu-wpcom/tests/php/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad-test.php
daledupreez: :+1:

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -443,34 +443,94 @@ function wpcom_launchpad_track_completed_task( $task_id, $extra_props = array()
...
+
+	return wpcom_launchpad_update_task_status( $values );
+}
+
+/**
+ * Updates task/s statuses. If a non-existent task is passed, it will be ignored.
+ * It will trigger the completion event in Tracks for the tasks that were marked as complete.
+ *
+ * @param array $new_statuses Array of mappings [ task_id: string => status: bool ].
+ * @return bool True if the option where the statuses are stored was updated, false otherwise.
```
daledupreez: Might it make sense for this function to return an array of boolean values with the task IDs as the keys? It's going to be hard to clearly map a set of actions back to a single boolean. (This is not blocking, BTW.)

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -443,34 +443,94 @@ function wpcom_launchpad_track_completed_task( $task_id, $extra_props = array()
...
+ */
+function wpcom_launchpad_update_task_status( $new_statuses ) {
```
daledupreez, line 509: We may want to validate that `$new_statuses` is an array.
```suggestion
function wpcom_launchpad_update_task_status( $new_statuses ) {
	if ( ! is_array( $new_statuses ) ) {
		return false;
	}
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -443,34 +443,94 @@ function wpcom_launchpad_track_completed_task( $task_id, $extra_props = array()
...
+			continue;
+		}
 
-	$statuses         = get_option( 'launchpad_checklist_tasks_statuses', array() );
-	$statuses[ $key ] = true;
-	$result           = update_option( 'launchpad_checklist_tasks_statuses', $statuses );
+		// If the task has an id_map, use that, otherwise use the actual task id as the key.
+		$task                        = $task_definitions[ $actual_task_id ];
+		$stored_task_id              = isset( $task['id_map'] ) ? $task['id_map'] : $actual_task_id;
+		$statuses[ $stored_task_id ] = $new_value;
```
daledupreez: Should we ensure that we're only storing booleans?
```suggestion
		$statuses[ $stored_task_id ] = (bool) $new_value;
```

daledupreez (commented):
Dropped some quick comments based on another look at the code, but I think we need to resolve the task ID handling thread before we can make too much progress.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
andres-blanco: I kept this because this was what the endpoint was returning. If we are changing that, this is a good change! Probably makes sense to make it in a separate PR so it's easier to test.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
andres-blanco: I agree with the proposed changes, I think this sums them up:
* you can call the function with a task id (a key in the `$task_definitions` array) or with a id_map. 
* The task id will take precedence over id_map.
* `$resolved_task_id` is going to be used for storage

The only doubt I have with the suggested code is in this part
image
Suppose someone calls the function with the task `domain_upsell_deferred` (which is an id_map). `$resolved_task` is going to be defined in line `534` as `domain_upsell`. But if someone calls the function with the actual task id (`domain_upsell`) `$resolved_task_id` is going to have the `domain_upsell_deferred` value. 
In the end the function will change it's storage location for the same task based on wether a task_id or an id_map is provided. Shouldn't line 534 be: `$resolved_task_id = $requested_task_id`?

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
andres-blanco: yes, nothing else should be stored there.

sixhours:
Just noting I've been reviewing and trying to wrap my head around this (the id/id_map definitely adds confusion, but it's unavoidable) and it all seems to work in testing.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
daledupreez: I am going to think out loud here, because there are three cases we need to get right, _and_ because I changed this code while I was writing the unit tests:

**1. Simple task ID without a map**
For this case, we find the task in `$task_definitions[ $requested_task_id ]` and we don't have an `id_map`, so we use the task ID.

**2. Task ID for a task that has an `id_map`**
For this case, we find the task in `$task_definitions[ $requested_task_id ]`, but we _do_ have an `id_map`, so we use the `id_map` value for the underlying option. In your example above, this means we would receive `domain_upsell_deferred` in the request, and resolve to `domain_upsell`.

**3. We only get an ID map where no task exists with that task ID**
For this case, we don't find the task in `$task_definitions[ $requested_task_id ]`, but we do find an entry in `$reverse_id_map[ $requested_task_id ]`. This means that one or more tasks has/have this task as an `id_map`, so we use `$requested_task_id`, because we want the `id_map` value for the underlying option. I made this change in https://github.com/Automattic/jetpack/pull/31471/commits/3005453109bc4120f19a07f06eb77d0da1b8ec2d.

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-definitions-test.php
daledupreez: I am going to mark this conversation as resolved. We don't need a task for the `id_map` to work.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
daledupreez: I modified the API to handle only one incoming task, so this question is moot.

projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php
daledupreez: My refactors should take care of this issue as well.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
daledupreez: Missed this previously, but added in https://github.com/Automattic/jetpack/pull/31471/commits/a442f12cf4eb3d0a616ee1b0bb6e449ac09e37c9.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
daledupreez: I have also added some explicit test cases in https://github.com/Automattic/jetpack/pull/31471/commits/e12bf3c2ba3492d7dc1cf0ec451d72a2587ad0f1.

projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php
```diff
@@ -155,19 +155,8 @@ public function update_site_options( $request ) {
...
-					$launchpad_checklist_tasks_statuses_option = (array) get_option( 'launchpad_checklist_tasks_statuses', array() );
-					$launchpad_checklist_tasks_statuses_option = array_merge( $launchpad_checklist_tasks_statuses_option, $value );
-
-					foreach ( $value as $task => $task_value ) {
-						if ( $task_value ) {
-							// If we're marking a task as complete, the value should be truthy, so fire the Tracks event.
-							wpcom_launchpad_track_completed_task( $task );
-						}
-					}
+					$updated[ $key ] = wpcom_launchpad_update_task_status( $value );
```
andres-blanco, line 158: I like this much better than my previous approach ❤️

andres-blanco:
This is looking good to me. I can't approve because I am the original author, but it's ready to ship.

daledupreez (approved):
Despite taking this PR over from @andres-blanco, he's still listed as the owner, so I am approving based on his approval comment [here](https://github.com/Automattic/jetpack/pull/31471#issuecomment-1682828343).

PR #31479, opened 2023-06-21
Blaze: introduce Blaze module

### LINKED ISSUE #29294: Blaze showing up by default on all my sites with Jetpack
### Impacted plugin

Jetpack

### Quick summary

Blaze is now a default and showing up in Posts

### Steps to reproduce

1. Show all posts
2. Mouse over posts to Edit, View, etc and now new option Blaze

This should not be turned on by default, it should have to be selected by Admin

### What actually happened

_No response_

### Impact

All

### Available workarounds?

Yes, difficult to implement

Self-hosted

### Logs or notes

It is simple to disable but must be done on every site with a filter

jeherve:
Fixes #29294

> **Warning**
> This depends on #30103 and #31485 being merged.

## Proposed changes:

Instead of loading Blaze automatically and only providing a code snippet to disable it, let's make it a module (automatically enabled upon connection to WordPress.com) that can be disabled via the modules as well as the Settings page.

This will be especially important once #30103 will be merged and the new dashboard will be enabled, since the Blaze feature  will then have its own menu in the dashboard, and will thus become more prominent in WordPress dashboards.

Since the feature becomes a module, we do not need to include it in the Config package anymore. It will be initialized when the module is activated.

> **Note**
> The new module will be automatically activated when new sites connect to WordPress.com or when sites are updated to the next version of Jetpack. However, on your test site using the Beta plugin or on your local development site, the module will not be automatically updated. You'll consequently have the feature disabled when you start testing.

### UI Changes

This PR introduces UI changes in 2 different interfaces:

- The old modules page at `wp-admin/admin.php?page=jetpack_modules`, where you'll see a new "Blaze" item.
- A new card in Jetpack > Settings > Traffic

### Screenshots

Here are some screenshots of the different use-cases you may find in Jetpack > Settings > Traffic:

**Offline mode**
Screenshot 2023-06-21 at 12 43 35

**Feature disabled on the site**

 for example via `add_filter( 'jetpack_blaze_enabled', '__return_false' );`

image

**Inactive**

image

**Active with the new Dashboard enabled**

via `add_filter( 'jetpack_blaze_dashboard_enable', '__return_true' );`

image

**Active with the new Dashboard not active yet**

This should be the default, does not require any filter:

image

**Site connected, but user not connected**
Screenshot 2023-06-21 at 12 44 52

### Other information:

## Jetpack product discussion

Related PRs:

* https://github.com/Automattic/wp-calypso/pull/78637
* 1416-gh-Automattic/wpcomsh

* No

## Testing instructions:

Here are some basic tests:

* Go to Jetpack > Settings > Traffic.
* Enable the Blaze feature.
* Go to Posts > All Posts, and notice the Blaze CTA links when hovering the published posts.
* Check that when clicking on Blaze you're redirected to the Blaze support docs.
* Disable the module.
* Go back to Posts > All Posts: the links should be gone.

For more advanced tests, I would recommend playing with the filters I mentioned above.

### DIFF
--- a/projects/packages/blaze/README.md
+++ b/projects/packages/blaze/README.md
@@ -11,32 +11,11 @@ Use composer to add the package to your project:
 composer add automattic/jetpack-blaze
 ```
 
-### Initializing the features
-
-#### Using the Config Package
-
-The JITMs can be enabled using the Config package:
-
-```
-use Automattic/Jetpack/Blaze;
-
-add_action( 'plugins_loaded', 'configure_blaze', 1 );
-
-function configure_blaze() {
-    $config = new Config();
-    $config->ensure( 'blaze' );
-}
-```
-
-#### Hook it
-Or can initialize it on the `admin_init` hook:
-
-```php
-add_action( 'admin_init', array( '\Automattic\Jetpack\Blaze', 'configure' ) );
-```
+### Initializing the feature
 
 #### Direct invocation
-Or directly invoke with a method call: 
+You can directly invoke the feature with a method call: 
+
 ```php
 use Automattic\Jetpack\Blaze;
 Blaze::init();
--- a/projects/packages/blaze/changelog/update-blaze-module
+++ b/projects/packages/blaze/changelog/update-blaze-module
@@ -1,0 +1,4 @@
+Significance: patch
+Type: changed
+
+Blaze can now be loaded as a module, instead of relying on the Config package.
--- a/projects/packages/config/changelog/update-blaze-module
+++ b/projects/packages/config/changelog/update-blaze-module
@@ -1,0 +1,4 @@
+Significance: patch
+Type: changed
+
+Blaze can now be loaded as a module, instead of relying on the Config package.
--- a/projects/packages/config/src/class-config.php
+++ b/projects/packages/config/src/class-config.php
@@ -12,7 +12,6 @@ namespace Automattic\Jetpack;
  * contain the package classes shown below. The consumer plugin
  * must require the corresponding packages to use these features.
  */
-use Automattic\Jetpack\Blaze as Blaze;
 use Automattic\Jetpack\Connection\Manager;
 use Automattic\Jetpack\Connection\Plugin;
 use Automattic\Jetpack\Import\Main as Import_Main;
@@ -56,7 +55,6 @@ class Config {
 		'videopress'      => false,
 		'stats'           => false,
 		'stats_admin'     => false,
-		'blaze'           => false,
 		'yoast_promo'     => false,
 		'import'          => false,
 	);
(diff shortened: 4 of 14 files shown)


samiff (dismissed):
This tested well for me 👍 

My only question was regarding the Calypso side of things, is there plan to take into account the module/filter setting there (wordpress.com/posts/example.com)? Or should that always show the "Promote with Blaze" option?

jeherve:
> My only question was regarding the Calypso side of things, is there plan to take into account the module/filter setting there (wordpress.com/posts/example.com)? Or should that always show the "Promote with Blaze" option?

That's a good point. I had initially thought that this should always appear, given that the Advertising menu in there always appears, but after your account I'm starting to have my doubts. Maybe the whole thing should disappear then?

If it does, we have 2 options to do that:

1. We could update the eligibility endpoint with D114507-code to consider a site non-eligible to Blaze when the module is disabled (it would also require some changes in this PR, see below) since we would want to still allow you to turn the module on when your site is not eligible yet). **I'm not a big fan of this approach since I think a site can be considered *eligible* for Blaze when the module is off.**

```diff
diff --git a/projects/plugins/jetpack/_inc/client/traffic/blaze.jsx b/projects/plugins/jetpack/_inc/client/traffic/blaze.jsx
index 3befd68308..acc51f70fb 100644
--- a/projects/plugins/jetpack/_inc/client/traffic/blaze.jsx
+++ b/projects/plugins/jetpack/_inc/client/traffic/blaze.jsx
@@ -57,15 +57,6 @@ function Blaze( props ) {
 	};
 
 	const blazeToggle = () => {
-		if ( ! blazeAvailable ) {
-			return (
-				<ToggleControl
-					disabled={ true }
-					label={ __( 'Blaze is not available on your site.', 'jetpack' ) }
-				/>
-			);
-		}
-
 		return (
 			<ModuleToggle
 				slug="blaze"

```

2. We could merge this PR as is, and then update Calypso to add a check for the module on top of the eligibility check. **That would be my preferred approach personally.** Here is PR for it: https://github.com/Automattic/wp-calypso/pull/78637

What do you think?

jeherve:
Coming back to this discussion, I committed 06f80b09ef186b164adf83447afa6834703ee889 to hide the card on WoA sites, where we'll want the feature to be available by default, like the WordPress.com toolbar. In parallel to that, I've opened 1416-gh-Automattic/wpcomsh to ensure the feature is available on WoA sites.

samiff (approved):
> We could merge this PR as is, and then update Calypso to add a check for the module on top of the eligibility check

I like this approach more 👍 

>  I committed 06f80b0 to hide the card on WoA sites

Makes sense and tested well for me.

PR #31480, opened 2023-06-21
Jetpack plugin: fix warnings in related posts

sgomes:
This PR fixes some warnings in the Related Posts feature, following the work in #31432.

Fixes the issue raised in https://github.com/Automattic/jetpack/pull/31432#issuecomment-1599254960 .

## Proposed changes:
* Add some checks to prevent the undefined array key warning for `src_width` in related posts
* Add further checks to prevent other undefined array key warnings in the same module

### Other information:

## Jetpack product discussion
None.

No.

## Testing instructions:
I'm not entirely sure how to reproduce the conditions that lead to the warning; for that to happen, there needs to exist a `Jetpack_PostImages` image that has a defined URL, but no defined `src_width`. I'm not entirely sure how to accomplish that. @samiff : Do you happen to know how?

Assuming you're able to reproduce the above, you can test by following the same instructions as in #31432, pasted below:
>- Enable the "Related Posts" feature for your test blog (`Settings -> Traffic -> Show related content after posts`).
>- Enable related posts thumbnails (`Show a thumbnail image where available` in the same place).
>- Open a post that has related posts with thumbnails.
>- Verify that the thumbnails include a `srcset` with different pixel ratios (up to 4x, if the source image is large enough).
>- Optionally, validate that the variants that get fetched on page load change as you change zoom values and/or try different pixel ratio screens.

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-related-posts-undefined-array-key-warnings
+++ b/projects/plugins/jetpack/changelog/fix-related-posts-undefined-array-key-warnings
@@ -1,0 +1,5 @@
+Significance: patch
+Type: bugfix
+Comment: Small fix to a recent enhancement.
+
+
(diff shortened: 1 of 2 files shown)


jeherve (approved):

sgomes:
Thank you for the review, @jeherve! 👍

PR #31487, opened 2023-06-21
Super Cache: Type check $posts to prevent warning in gzip output stream

### LINKED ISSUE #31486: Super Cache: $posts in theme causes "cannot decode raw data" in safari
### Impacted plugin

Super Cache

### Quick summary

We are happy users of wp-super-cache and have been for some time. Recently we discovered that sometimes pages would not render in Safari, but instead show "cannot decode raw data". The problem happens when visiting a page for that ought to be cached, but currently is not cached yet.

We traced the issue back to wp-super-cache, in combination with using the variable $posts in theme code to store a WP_Query (or anything that is not an array, for that matter). This causes wp_cache_post_id to show a warning which is added in plain text to the otherwise gzipped content. Safari can't decode the gzipped content (because of the extra content at the end) which causes the "cannot decode raw data" error.

Pull request is incoming.

### Steps to reproduce

- Install wordpress with supercache
- Create a page with slug "test"
- Add `$posts = new WP_query();` to functions.php
- Delete all cached pages
- Visit page with slug "test" in safari

Page should be displayed

### What actually happened

Safari displays an error "cannot decode raw data".

### Impact

Some (< 50%)

### Available workarounds?

Yes, easy to implement

_No response_

### Logs or notes

_No response_

septemberservices:
Fixes #31486

## Proposed changes:
* Type check $posts in wp_cache_post_id() to prevent a warning breaking the gzipped content in some browsers

No.

## Testing instructions:

* Install wordpress with supercache
* Create a page with slug "test"
* Add $posts = new WP_query(); to functions.php
* Delete all cached pages
* Visit page with slug "test" in safari

### DIFF
--- a/projects/plugins/super-cache/wp-cache-phase2.php
+++ b/projects/plugins/super-cache/wp-cache-phase2.php
@@ -3370,7 +3370,7 @@ function wp_cache_post_id() {
 	if ( $comment_post_ID > 0 ) {
 		return $comment_post_ID;
 	}
-	if ( is_singular() && ! empty( $posts ) ) {
+	if ( is_singular() && ! empty( $posts ) && is_array($posts) ) {
 		return $posts[0]->ID;
 	}
 	if ( isset( $_GET['p'] ) && $_GET['p'] > 0 ) {
--- a/projects/plugins/super-cache/wp-cache-phase2.php
+++ b/projects/plugins/super-cache/wp-cache-phase2.php
@@ -3370,7 +3370,7 @@ function wp_cache_post_id() {
 	if ( $comment_post_ID > 0 ) {
 		return $comment_post_ID;
 	}
-	if ( is_singular() && ! empty( $posts ) && is_array($posts) ) {
+	if ( is_singular() && ! empty( $posts ) && is_array( $posts ) ) {
 		return $posts[0]->ID;
 	}
 	if ( isset( $_GET['p'] ) && $_GET['p'] > 0 ) {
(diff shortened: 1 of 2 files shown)


septemberservices:

No.

projects/plugins/super-cache/wp-cache-phase2.php
```diff
@@ -3370,7 +3370,7 @@ function wp_cache_post_id() {
 	if ( $comment_post_ID > 0 ) {
 		return $comment_post_ID;
 	}
-	if ( is_singular() && ! empty( $posts ) ) {
+	if ( is_singular() && ! empty( $posts ) && is_array($posts) ) {
```
jeherve: You'll most likely need to add some spacing here to avoid linting errors:

```suggestion
	if ( is_singular() && ! empty( $posts ) && is_array( $posts ) ) {
```

jeherve (changes requested):
Thanks for your contribution. I've started the checks that will give you automated feedback about your PR, but wanted to highlight a tiny issue right away.

projects/plugins/super-cache/wp-cache-phase2.php
septemberservices: It seems your commit has been added to my repository, resolving the problem. Is that correct?

zinigor (approved):
Looks good to me, doesn't affect the logic, should be fine.

zinigor:
@jeherve do you mind taking another look? I don't want to dismiss your review without your approval.

jeherve (approved):

PR #31496, opened 2023-06-21
AI Extension: extend core list item core block

retrofox:
Fixes #

## Proposed changes:

* AI Extension: extend core list item core block

### Other information:

## Testing instructions:

* Go to the block editor
* Create content with list composition
* Confirm it's possible to change the content by using the AI extension tools

https://github.com/Automattic/jetpack/assets/77539/71463922-d157-4089-95dd-d32384502ccd


dhasilva (approved):
LGTM

The multi-block option only changes the first block, probably because the list item is a child of the list block and that is not the case with the paragraph or heading blocks. Something for a follow-up.

PR #31497, opened 2023-06-21
AI Extension: handle errors from extended block actions

### LINKED ISSUE #31382: AI Assistant: show errors in the UI when something fails
We need to show the errors in the UI when something fails.
We can offer a message similar to what we do for the AI Assistant block, but we also should keep in mind that we are extending other blocks. It could be intrusive.

lhkowalski:
Fixes #31382.

## Proposed changes:

* Add event handlers for the error events emitted by `SuggestionsEventSource`
* Create custom onError callback to display the error on the editor, [using native editor notice support](https://developer.wordpress.org/block-editor/how-to-guides/notices/#notices-in-the-block-editor)

### Other information:

## Testing instructions:

* To test this, you need to force some errors on the backend; refer to the "How to force errors on the backend" section below to see how to do it
* With the backend changed to force an error return, go to the block editor
* Write some content on a paragraph
* Click the AI Assistant extension button and select some action:

Screen Shot 2023-06-21 at 19 31 38

* Since you are forcing an error on the backend, you will see a notice on the top of the editor:

Screen Shot 2023-06-22 at 15 22 07

* Change the `status` code returned on the backend error to the following values and repeat the test
   * 422
   * 429
   * 500
   * 503
* Each `status` code has it's own message on the frontend, so you will see different messages on each occasion

### How to force errors on the backend

* Sandbox your `public-api`
* If you are running this branch on JN, don't forget to point it to your sandbox
* Update the `jetpack-ai-query.php` file on your sandbox to add the following snippet:

```
return new \WP_Error( 'error_jetpack_ai', __( 'Generic error message.' ), [ 'status' => 503 ] );
```

* Add it as the first line of the function:

Screen Shot 2023-06-21 at 19 29 33

* You will test changing the `status` value, so keep the file open and ready to edit


projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
```diff
@@ -142,6 +146,13 @@ export const withAIAssistant = createHigherOrderComponent(
 
 		return (
 			<>
+				{ suggestionError && (
+					<ErrorNotice
+						message={ suggestionError.message }
+						onRemove={ () => setSuggestionError( null ) }
```
retrofox: Could we consider avoiding the arrow function here?

```suggestion
						onRemove={ setSuggestionError }
```

projects/plugins/jetpack/extensions/blocks/ai-assistant/components/error-notice/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * React component to render a notice with a message.
+ *
+ * @param {ErrorNoticeProps} props - The props for the component.
+ * @returns {React.ReactElement } The Notice component.
+ */
+export default function ErrorNotice( { message, onRemove }: ErrorNoticeProps ): React.ReactElement {
+	return (
+		<Notice className="jetpack-ai-assistant__error-notice" status="warning" onRemove={ onRemove }>
```
retrofox: `status` could be a component prop, with a `warning` value as default. just a comment.

projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-suggestions-from-ai/index.ts
```diff
@@ -125,12 +149,68 @@ export default function useSuggestionsFromAI( {
...
+						onError( {
+							code: 'error_network',
+							message: __(
+								'It was not possible to process your request. Mind trying again?',
+								'jetpack'
+							),
+							status: 'info',
+						} );
+					} );
+				}
```
retrofox, line 207: Could we expose this errors list in an isolated function based on the error code?
We can do that in a follow-up, no problem at all.

retrofox (changes requested):
The notice is anchored at the top of the component:

image

Let's anchor it at the top of the post page. For instance:

image

The [Notice doc](https://developer.wordpress.org/block-editor/reference-guides/components/notice/) page has some cases of good and bad practices.

Btw, probably we'd like to implement the notices by using the [withNotice](https://developer.wordpress.org/block-editor/reference-guides/components/with-notices/) HOC.




https://developer.wordpress.org/block-editor/reference-guides/components/with-notices/

projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-suggestions-from-ai/index.ts
lhkowalski: Yeah, I think it's a good idea since we are using the same errors on the block as well, so we could reuse them on 2 places.

projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
lhkowalski: Going to create a small `clearSuggestionError` function to be used here.

projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-suggestions-from-ai/index.ts
lhkowalski: Going to leave this for a followup so we can move forward with basic error handling.

projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
lhkowalski: Was not necessary since using the native notice support made me remove the error state inside the hook.

projects/plugins/jetpack/extensions/blocks/ai-assistant/components/error-notice/index.tsx
lhkowalski: I made it refer to `error.status`, that may be `error` or `info` so far.

retrofox (dismissed):
Looks good to me! I believe this error handling is a great choice when compared to other alternatives available 🚀

retrofox (approved):
LGTM

PR #31499, opened 2023-06-21
Fix send email preview endpoint

lezama:
Fixes https://github.com/Automattic/jetpack/pull/31021#pullrequestreview-1491718360

## Proposed changes:

* When a site uses "plain" permalinks the API endpoint path is passed via a `rest_route` query arg, in the `proxy_request_to_wpcom_as_user` this one was being passed to wpcom and wpcom was trying to use it to serve the endpoint. This removes the `rest_route` query arg when proxing api requests.

### Other information:

No
## Testing instructions:

* Set your site permalinks to "plain"
* Try the send an email preview feature using the newsletter prepublish panel.
* The request should succeed.

### DIFF
--- a/projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/trait-wpcom-rest-api-proxy-request-trait.php
+++ b/projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/trait-wpcom-rest-api-proxy-request-trait.php
@@ -23,7 +23,7 @@ trait WPCOM_REST_API_Proxy_Request_Trait {
 	public function proxy_request_to_wpcom_as_user( $request, $path = '' ) {
 		$blog_id = \Jetpack_Options::get_option( 'id' );
 		$path    = '/sites/' . rawurldecode( $blog_id ) . rawurldecode( $this->rest_base ) . ( $path ? '/' . rawurldecode( $path ) : '' );
-		$api_url = add_query_arg( $request->get_query_params(), $path );
+		$api_url = remove_query_arg( 'rest_route', add_query_arg( $request->get_query_params(), $path ) );
 
 		$request_options = array(
 			'headers' => array(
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/trait-wpcom-rest-api-proxy-request-trait.php
```diff
@@ -23,7 +23,7 @@ trait WPCOM_REST_API_Proxy_Request_Trait {
 	public function proxy_request_to_wpcom_as_user( $request, $path = '' ) {
 		$blog_id = \Jetpack_Options::get_option( 'id' );
 		$path    = '/sites/' . rawurldecode( $blog_id ) . rawurldecode( $this->rest_base ) . ( $path ? '/' . rawurldecode( $path ) : '' );
-		$api_url = add_query_arg( $request->get_query_params(), $path );
+		$api_url = remove_query_arg( 'rest_route', add_query_arg( $request->get_query_params(), $path ) );
```
jeherve: I was wondering if we could be a bit more verbose about the change, so we remember why we did that a few months from now?

Something like this?

```suggestion
		$query_params = $request->get_query_params();

		/*
		 * A rest_route parameter can be added when using plain permalinks.
		 * It is not necessary to pass them to WordPress.com,
		 * and may even cause issues with some endpoints.
		 * Let's remove it.
		 */
		if ( isset( $query_params['rest_route'] ) ) {
			unset( $query_params['rest_route'] );
		}
		$api_url = add_query_arg( $query_params, $path );
```

What do you think?

lezama: Agreed a 100% 🤝

jeherve (approved):

edanzer (approved):
Good find on this! This is testing as expect. The requests we tested yesterday that were failing with not-nice permalinks are now working. And code looks straightforward. 

Given that this change touches most rest requests (ie, wide effect on critical functionality), I'd still want at least one other review/approval on this, ideally from Jetpack crew, but looks like we already have that too.

lezama:
Thank you both!

PR #31504, opened 2023-06-22
Boost: exclude small images from Image Size Analysis

### LINKED ISSUE #31429: Boost: Image Size Analysis - ignore images under certain dimensions
Pixel tracking gets added to the Image Size Analysis reports.

donnchawp:
Image Guide excludes small images by checking the width and height is over 64 pixels which I have replicated here.
This will avoid doing analysis on small images like trackers.

Fixes #31429 

## Proposed changes:
* Add a new isBigImage() function
* Add that function as a filter on the list of images returned by getMeasurableImages()

### Other information:

No

## Testing instructions:
* Install Jetpack Beta to install Boost "Bleeding Edge".
* Upgrade to a premium subscription
* Enable image size analysis by defining the constant JETPACK_BOOST_IMAGE_SIZE_ANALYSIS.
* Go to the Boost page and enable and run "Image Size Analysis".
* The report shouldn't include any small images.


donnchawp:
It works, but on the wrong Image Guide, not the report analysis.

dilirity (commented):
This works great - thank you! No more tracking pixels in the reports!

While code reviewing, I noticed, that we use `getSizeOnPage` which uses `getBoundingClientRect` and that returns the size in the viewport.

I tested adding a large image (1707x2560), and resized it to 38x57. The image gets reported (which I didn't expect), though I didn't expect the size recommendations to be wrong:

CleanShot 2023-06-22 at 19 38 12@2x

I also noticed, that the Image Guide doesn't add a "report bubble" in the front-end for this image, which is probably to be expected as it's way small...

CleanShot 2023-06-22 at 19 39 47@2x

Shouldn't we sync the analyzer and image guide?

thingalon:
> Shouldn't we sync the analyzer and image guide?

Yes, any change made to the Premium Image Size Analyzer should also apply to the free one. Since this change adds in a filter step inside MeasurableImage, it seems like we should remove it from `app/modules/image-guide/src/index.ts` too, right?

> I tested adding a large image (1707x2560), and resized it to 38x57. The image gets reported (which I didn't expect), though I didn't expect the size recommendations to be wrong:

🤔 That doesn't seem right. I feel like it shouldn't have been in your results, because we check the image size on the page, not the file size. Can you run getBoundingRect on the image element in your JavaScript console to check what it says?

dilirity:
> Can you run getBoundingRect on the image element in your JavaScript console to check what it says?

CleanShot 2023-06-23 at 15 02 51@2x


This is the HTML of the image on the page:

```html
<img decoding="async" src="http://jetpack-boost.test/wp-content/uploads/2023/06/fernando-paredes-murillo-Xc6gtOwSMSA-unsplash-scaled.jpg" alt="" class="wp-image-931" width="38" height="57" srcset="http://jetpack-boost.test/wp-content/uploads/2023/06/fernando-paredes-murillo-Xc6gtOwSMSA-unsplash-scaled.jpg 1707w, http://jetpack-boost.test/wp-content/uploads/2023/06/fernando-paredes-murillo-Xc6gtOwSMSA-unsplash-200x300.jpg 200w, http://jetpack-boost.test/wp-content/uploads/2023/06/fernando-paredes-murillo-Xc6gtOwSMSA-unsplash-683x1024.jpg 683w, http://jetpack-boost.test/wp-content/uploads/2023/06/fernando-paredes-murillo-Xc6gtOwSMSA-unsplash-768x1152.jpg 768w, http://jetpack-boost.test/wp-content/uploads/2023/06/fernando-paredes-murillo-Xc6gtOwSMSA-unsplash-1024x1536.jpg 1024w, http://jetpack-boost.test/wp-content/uploads/2023/06/fernando-paredes-murillo-Xc6gtOwSMSA-unsplash-1365x2048.jpg 1365w" sizes="(max-width: 38px) 100vw, 38px">
```

For some reason, it's actually reporting the 200x300 version of the image - not the full size one.

thingalon:
I just had a huddle with @dilirity, and he has pointed out to me that I've asked for the wrong thing. I asked that we discard images based on their size on the page, but that is wrong. Sorry @donnchawp but we have to take this in a slightly different direction.

@dilirity's [example above](https://github.com/Automattic/jetpack/pull/31504#pullrequestreview-1493622257) has an image that appears tiny on the page, but is using a large (in pixels) image file to fill that space. We should include it in the report, because it is a performance issue.

The goal is to exclude _tracking pixels_ and other junk that users don't think of as an "image" on the page.

Given that, I think we should change direction -- and reject images based on how large the image file is (in pixels) so that tracking images (typically 1x1 pixels) get thrown away, but a huge image crammed into a tiny space on the page will be reported. In other words, we should base it on the file data, not the client rect.

I don't know if all tracking pixels are 1x1, so I think a fair rule to follow is: throw away any image whose **file** is less than 10 pixels wide, or less than 10 pixels tall.

I think that Boost's existing check (in `app/modules/image-guide/src/index.ts`) which rejects based on client rect is OK for the free in-browser version because we can't fit the UI elements for reporting images in a tiny space. So that can remain as-is. We just want to add a new check to reject images based on their image file dimensions to the Image Guide Package.

donnchawp:
Both free and upgraded guides now use isImageBig() to filter out small images, so there's no need for the discard function in index.ts.

An image is "big" if it is at least 65 pixels high and wide. That excludes Gravatar images too.

thingalon (approved):
Thanks for sticking with this and getting it done, @donnchawp !

PR #31515, opened 2023-06-22
 AI Assistant: Add a specific delimiter for content in the prompts

dhasilva:
Fixes #

## Proposed changes:

* Following a tip from the DeepLearning.ai course, this PR adds a clear delimiter for the content to be acted upon. This should reduce the occasions where parts of the prompt are repeated or its structure leaks, like this:

* It also replaces the `system` message with the content to a `user` message, to fix this problem:

### Other information:

## Testing instructions:

* Test the AI Assistant block many times. The responses should not contain phrases like "The specific relevant content for this request"

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
@@ -86,11 +86,6 @@ type PromptOptionsProps = {
 	content: string;
 
 	/*
-	 * The delimited content.
-	 */
-	delimitedContent: string;
-
-	/*
 	 * The language to translate to. Optional.
 	 */
 	language?: string;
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
```diff
@@ -83,6 +85,11 @@ type PromptOptionsProps = {
 	 */
 	content: string;
 
+	/*
+	 * The delimited content.
+	 */
+	delimitedContent: string;
```
retrofox: Can we avoid adding new props to the helper functions?
Could you add the delimiter changes straight into the fn?

retrofox:
```jsx
content = `${ delimiter }${ content?.replaceAll(
		delimiter,
		''
	) }${ delimiter }`;
```

Something like this

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
```diff
@@ -104,14 +106,20 @@ type PromptOptionsProps = {
 	prevMessages?: Array< PromptItemProps >;
 };
 
+function getDelimitedContent( content: string ): string {
```
retrofox, line 109: 💯

retrofox (approved):
Very nice enhancement. 👏 🚢

PR #31518, opened 2023-06-22
 AI Assistant: Change delimiter and remove it from responses

dhasilva:
Follow-up to #31515

## Proposed changes:

* Changes the delimiter to ```` because `####` has meaning in Markdown
* Removes any delimiter from the response as it is streamed because sometimes the AI repeats them in the response
* Fixes the JETPACK_AI_ERROR check as sometimes more tokens are streamed together
* Fixes a typo

### Other information:

## Testing instructions:

* Add some text in Chinese
* In the extension, try to translate it to French. This is the most consistent way I could find to generate errors, as sometimes the AI asks for English text for some reason.
* Check that no JETPACK_AI_ERROR is printed and the responses do not have the delimiter in them

Before | After
-|-
 |

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-suggestions-from-ai/index.ts
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-suggestions-from-ai/index.ts
@@ -8,8 +8,8 @@ import debugFactory from 'debug';
 /**
  * Types
  */
+import { PromptItemProps, delimiter } from '../../lib/prompt';
 import { SuggestionsEventSource, askQuestion } from '../../lib/suggestions';
-import type { PromptItemProps } from '../../lib/prompt';
 
 const debug = debugFactory( 'jetpack-ai-assistant:prompt' );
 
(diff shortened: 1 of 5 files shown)


projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-suggestions-from-ai/index.ts
```diff
@@ -88,7 +88,10 @@ export default function useSuggestionsFromAI( {
 	 * @returns {void}
 	 */
 	const handleSuggestion = useCallback(
-		( event: CustomEvent ) => onSuggestion( event?.detail ),
+		( event: CustomEvent ) => {
+			// Remove the delimiter from the suggestion.
+			onSuggestion( event?.detail?.replaceAll( delimiter, '' ) );
```
retrofox: do we need to remove it from the AI response? What happens if the content contains the same chars that the delimiter?

dhasilva: The `before` screenshot has the delimiter in the AI response.

>What happens if the content contains the same chars that the delimiter?

It would be lost, that is why I changed it from `####`, as 4 backticks should be very rare. We could change it to something like `!?;` or something.

retrofox (approved):
LGTM :shipit:

PR #31527, opened 2023-06-22
CRM: add check for finfo_open

gogdzl:
## Proposed changes:
* This PR removes the `finfo_open` dependency when uploading files. When `finfo_open` is not available the CRM resorts to using `mime_content_type`.

### Other information:

## Jetpack product discussion
https://github.com/Automattic/zero-bs-crm/issues/3082

No.

## Testing instructions:
* Remove/Disable the `fileinfo` PHP extension
* Open up any contact view page (e.g. `wp-admin/admin.php?page=zbs-add-edit&action=view&zbstype=contact&zbsid=2`)
* Scroll down to `Documents` and open the `Files` tab
* Upload any file

In `trunk` a PHP error will appear: `Fatal error: Uncaught Error: Call to undefined function finfo_open()
in /Users/diegorodrigues/git/avengers/jetpack/projects/plugins/crm/includes/ZeroBSCRM.FileUploads.php on line 920`.

In `fix/crm/3082-add-check-for-fifo-open` the file will be uploaded with no errors.

### DIFF
--- a/projects/plugins/crm/includes/ZeroBSCRM.GeneralFuncs.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.GeneralFuncs.php
@@ -1634,3 +1634,21 @@ function jpcrm_generate_pdf( $html, $pdf_filename ) {
 function jpcrm_disable_browser_autocomplete() {
 	return time() . wp_rand( 0, 1000000 );
 }
+
+/**
+ * Retrieves the MIME type of a file.
+ *
+ * If the FileInfo extension is available, it uses finfo_file to determine the MIME type.
+ * Otherwise, it falls back to mime_content_type for MIME type detection.
+ *
+ * @param string $file_path The path to the file.
+ * @return string|false The MIME type of the file, or false if the MIME type cannot be determined.
+ */
+function jpcrm_get_mimetype( $file_path ) {
+	if ( function_exists( 'finfo_file' ) ) {
+		$file_info = finfo_open( FILEINFO_MIME_TYPE );
+		return finfo_file( $file_info, $file_path );
+	}
+
+	return mime_content_type( $file_path );
+}
--- a/projects/plugins/crm/includes/ZeroBSCRM.GeneralFuncs.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.GeneralFuncs.php
@@ -1639,16 +1639,96 @@ function jpcrm_disable_browser_autocomplete() {
  * Retrieves the MIME type of a file.
  *
  * If the FileInfo extension is available, it uses finfo_file to determine the MIME type.
- * Otherwise, it falls back to mime_content_type for MIME type detection.
+ * Otherwise, if the mime_content_type function is available, it falls back to mime_content_type.
+ * If neither are available it uses a file signature/MIME type map as a fallback.
  *
  * @param string $file_path The path to the file.
  * @return string|false The MIME type of the file, or false if the MIME type cannot be determined.
  */
 function jpcrm_get_mimetype( $file_path ) {
-	if ( function_exists( 'finfo_file' ) ) {
+	if ( function_exists( 'finfo_file' ) && function_exists( 'finfo_open' ) ) {
 		$file_info = finfo_open( FILEINFO_MIME_TYPE );
 		return finfo_file( $file_info, $file_path );
 	}
 
-	return mime_content_type( $file_path );
+	if ( function_exists( 'mime_content_type' ) ) {
+		return mime_content_type( $file_path );
+	}
+
+	$signature_to_mime = array(
+		'89504e47'             => 'image/png', // PNG
+		'ffd8ffe0'             => 'image/jpeg', // JPEG
+		'47494638'             => 'image/gif', // GIF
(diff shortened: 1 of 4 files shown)


projects/plugins/crm/includes/ZeroBSCRM.GeneralFuncs.php
```diff
@@ -1634,3 +1634,21 @@ function jpcrm_generate_pdf( $html, $pdf_filename ) {
...
+ * @param string $file_path The path to the file.
+ * @return string|false The MIME type of the file, or false if the MIME type cannot be determined.
+ */
+function jpcrm_get_mimetype( $file_path ) {
+	if ( function_exists( 'finfo_file' ) ) {
+		$file_info = finfo_open( FILEINFO_MIME_TYPE );
+		return finfo_file( $file_info, $file_path );
+	}
+
+	return mime_content_type( $file_path );
```
cleacos: What is the difference between using `mime_content_type` and `finfo_file`? Why don't we use directly `mime_content_type` always? Does `mime_content_type` only works with Apache HTTP servers?

gogdzl: Both are not available for all platforms, but I do suspect that `mime_content_type` won't present an issue. Having that said, it is better to try both options, their result should always be the same.

If we still get an error, we should use a fallback for this function.

cleacos: Could we add an extra fallback? Check if `mine_content_type` doesn't exist and return something generic

Or maybe test if this solution works. It's a polyfill function as the final fallback:

```
if(!function_exists('mime_content_type')) {

    function mime_content_type($filename) {

        $mime_types = array(

            'txt' => 'text/plain',
            'htm' => 'text/html',
            'html' => 'text/html',
            'php' => 'text/html',
            'css' => 'text/css',
            'js' => 'application/javascript',
            'json' => 'application/json',
            'xml' => 'application/xml',
            'swf' => 'application/x-shockwave-flash',
            'flv' => 'video/x-flv',

            // images
            'png' => 'image/png',
            'jpe' => 'image/jpeg',
            'jpeg' => 'image/jpeg',
            'jpg' => 'image/jpeg',
            'gif' => 'image/gif',
            'bmp' => 'image/bmp',
            'ico' => 'image/vnd.microsoft.icon',
            'tiff' => 'image/tiff',
            'tif' => 'image/tiff',
            'svg' => 'image/svg+xml',
            'svgz' => 'image/svg+xml',

            // archives
            'zip' => 'application/zip',
            'rar' => 'application/x-rar-compressed',
            'exe' => 'application/x-msdownload',
            'msi' => 'application/x-msdownload',
            'cab' => 'application/vnd.ms-cab-compressed',

            // audio/video
            'mp3' => 'audio/mpeg',
            'qt' => 'video/quicktime',
            'mov' => 'video/quicktime',

            // adobe
            'pdf' => 'application/pdf',
            'psd' => 'image/vnd.adobe.photoshop',
            'ai' => 'application/postscript',
            'eps' => 'application/postscript',
            'ps' => 'application/postscript',

            // ms office
            'doc' => 'application/msword',
            'rtf' => 'application/rtf',
            'xls' => 'application/vnd.ms-excel',
            'ppt' => 'application/vnd.ms-powerpoint',

            // open office
            'odt' => 'application/vnd.oasis.opendocument.text',
            'ods' => 'application/vnd.oasis.opendocument.spreadsheet',
        );

        $ext = strtolower(array_pop(explode('.',$filename)));
        if (array_key_exists($ext, $mime_types)) {
            return $mime_types[$ext];
        }
        elseif (function_exists('finfo_open')) {
            $finfo = finfo_open(FILEINFO_MIME);
            $mimetype = finfo_file($finfo, $filename);
            finfo_close($finfo);
            return $mimetype;
        }
        else {
            return 'application/octet-stream';
        }
    }
}
```

gogdzl: We do not have access to the file extension in that function. The fallback is much more complicated, but I made one in f0f9a4a that should be good enough. If we want anything better than the fallback I made we should be using a third party lib, or just accepting the file by just checking the extension while giving the admin a warning or making the admin use a new settings option.

cleacos: > by just checking the extension while giving the admin a warning or making the admin use a new settings option.
Maybe that is the "easy" solution for this edge case where the site doesn't support those functions. Letting them know about it and the potential risk: "We have checked that your PHP installation doesn't have this extension installed. Please install it or deactivate the file type security checker, this action will imply some risks".

In the GH issue, I suggested going with something like that, an alert, and letting the user take an action, allowing a basic check (extension).

About the fallback code, how could we test it to check if it's working?

gogdzl: > About the fallback code, how could we test it to check if it's working?

Just comment out the first part of the function and the fallback should be used.

cleacos (approved):

PR #31534, opened 2023-06-23
[Boost] Fix UI on older versions of Safari

thingalon:
Fixes user complaints about white-screen of death on older versions of Safari. We relied on `structuredClone` which isn't always 100% available on older browsers.

## Proposed changes:
* Fall back to using json.serialize/parse when structuredClone isn't available.

### Other information:

## Testing instructions:

* Use browserstack to test on older versions of safari.
* Make sure the boost dashboard loads and seems ok


dilirity (approved):
:shipit:

thingalon:
Cherry-picked to 1.9.2 in edd3dd3b97

PR #31537, opened 2023-06-23
Sharing Buttons: Add a Nextdoor button

pablinos:
This adds a sharing button for Nextdoor. It's a simple link with the post title and URL.

### Other information:

## Jetpack product discussion

1201611340747558-as-1203666305041653/f

No.

## Testing instructions:
* Check this PR out on a sandbox and sandbox the API
* On this branch on a dev site, go to `/wp-admin/admin.php?page=jetpack#/sharing`
* Enable Sharing Buttons, and click the link to configure the buttons
* Enable the Nextdoor button ( the logo won't show currently )
* Save changes and load a post on the dev site.
* The Nextdoor button should be displayed and link through to share the post on Nextdoor

image

### DIFF
--- a/projects/plugins/jetpack/modules/sharedaddy/sharing-sources.php
+++ b/projects/plugins/jetpack/modules/sharedaddy/sharing-sources.php
@@ -3198,3 +3198,76 @@ class Share_Mastodon extends Sharing_Source {
 		);
 	}
 }
+
+/**
+ * Nextdoor sharing service.
+ */
+class Share_Nextdoor extends Sharing_Source {
+	/**
+	 * Service short name.
+	 *
+	 * @var string
+	 */
+	public $shortname = 'nextdoor';
+
+	/**
+	 * Service icon font code.
+	 *
+	 * @var string
+	 */
+	public $icon = '\f10c';
+
+	/**
+	 * Constructor.
+	 *
+	 * @param int   $id       Sharing source ID.
+	 * @param array $settings Sharing settings.
+	 */
+	public function __construct( $id, array $settings ) { // phpcs:ignore Generic.CodeAnalysis.UselessOverridingMethod.Found
+		parent::__construct( $id, $settings );
+	}
+
+	/**
+	 * Service name.
+	 *
+	 * @return string
+	 */
+	public function get_name() {
+		return __( 'Nextdoor', 'jetpack' );
+	}
+
+	/**
+	 * Get the markup of the sharing button.
+	 *
+	 * @param WP_Post $post Post object.
+	 *
+	 * @return string
+	 */
+	public function get_display( $post ) {
+		return $this->get_link(
+			$this->get_process_request_url( $post->ID ),
+			_x( 'Nextdoor', 'share to', 'jetpack' ),
+			__( 'Click to share on Nextdoor', 'jetpack' ),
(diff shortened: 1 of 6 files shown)


pablinos:
With the deployment of this, it could be that the share button becomes available to configure but can't be displayed on the Jetpack site (because the site has not been upgraded, or the new version hasn't been released yet). I'm not sure how we've handled this in the past. What's the usual procedure here @jeherve?

jeherve (commented):
I pushed a few commits to help. What do you think about also updating the look of the icon-only button to use the brand's green, like so:

Screenshot 2023-06-23 at 10 42 46

```diff
diff --git a/projects/plugins/jetpack/modules/sharedaddy/sharing.css b/projects/plugins/jetpack/modules/sharedaddy/sharing.css
index 6580a7ff52..79d5a590ff 100644
--- a/projects/plugins/jetpack/modules/sharedaddy/sharing.css
+++ b/projects/plugins/jetpack/modules/sharedaddy/sharing.css
@@ -691,6 +691,11 @@ body .sd-social-icon .sd-content li.share-custom a span {
 	color: #fff !important;
 }
 
+.sd-social-icon .sd-content ul li[class*='share-'].share-nextdoor a.sd-button {
+	background: #8ED500;
+	color: #fff !important;
+}
+
 /**
  * Screen Reader Text for "Icon Only" option
  */

```

jeherve:
> With the deployment of this, it could be that the share button becomes available to configure but can't be displayed on the Jetpack site (because the site has not been upgraded, or the new version hasn't been released yet). I'm not sure how we've handled this in the past.

Do you mean that the button would be available to add in Calypso, while not yet available on the site? I don't think that should happen. Or maybe I'm missing something?

jeherve:
Noting that we will need both https://github.com/Automattic/wp-calypso/pull/78562 and https://github.com/Automattic/wp-calypso/pull/78565 on the Calypso side, to match this PR.

pablinos:
> Do you mean that the button would be available to add in Calypso, while not yet available on the site? I don't think that should happen

Yes it seems to me, that this will adjust the available services on the WPCOM API, which would get deployed the day after merging. Then in Calypso you could add a Nextdoor button, but it wouldn't be available on the Jetpack site until the next release.

projects/plugins/jetpack/modules/sharedaddy/sharing-sources.php
```diff
@@ -3217,16 +3217,6 @@ class Share_Nextdoor extends Sharing_Source {
 	 */
 	public $icon = '\f10c';
 
-	/**
```
pablinos: Ah nice, I didn't spot that! I copied this from the WhatsApp one. The dangers of copy/paste...

pablinos:
>  What do you think about also updating the look of the icon-only button to use the brand's green

Yes, let's do that! Added in 3a6378c

jeherve (approved):
This seems to test well for me, it should be good to merge!

> Yes it seems to me, that this will adjust the available services on the WPCOM API, which would get deployed the day after merging. Then in Calypso you could add a Nextdoor button, but it wouldn't be available on the Jetpack site until the next release.

I don't think it should be an issue, the button will not appear in Calypso until the site has the button available.

PR #31538, opened 2023-06-23
[Boost] Fix image guide proxy

thingalon:
The image guide was using `response.url` to check if a given Response worked, which isn't correct. This PR updates it to use `response.ok` which should verify the response is valid (aka 200 response).

The image guide proxy work in https://github.com/Automattic/jetpack/pull/31145 generates a manufactured response without a `url` set, causing it to always fail.

## Proposed changes:
* Use `response.ok` to check that the response received by the image guide is ok

### Other information:

## Testing instructions:
* Use the image guide to check the size of an image on an external URL
* Verify it works with this patch, doesn't without it.

### DIFF
--- a/projects/js-packages/image-guide/changelog/fix-image-guide-proxy
+++ b/projects/js-packages/image-guide/changelog/fix-image-guide-proxy
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Check response.url, not response.ok to verify a response worked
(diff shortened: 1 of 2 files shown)


dilirity (approved):
Good catch on this! :shipit: 

Tested and it properly returns the image size for remote images.

CleanShot 2023-06-23 at 10 49 05@2x

thingalon:
Cherry picked to 1.9.2 in ea942e0262

PR #31551, opened 2023-06-23
Added an initial checkout of Jetpack Inspect.

zinigor:
This PR adds an initial checkout of Jetpack Inspect with a scaffold for a new project. @thingalon  I'd appreciate if you or anyone on your team could take a look because this is also using Svelte.

## Proposed changes:
* Introduces an existing plugin into the Monorepo.

### Other information:

## Jetpack product discussion
* p1HpG7-nkd-p2
* p1HpG7-gKS-p2 

No.

## Testing instructions:

* TBD


projects/plugins/inspect/.distignore
anomiex, line 1: I don't think this file is needed in the monorepo. We instead use `production-exclude` entries in `.gitattributes` to avoid copying the relevant files into the mirror repo in the first place.

But some of this probably could use moving into `.gitattributes`.

projects/plugins/inspect/.editorconfig
anomiex, line 1: This appears to be redundant to [the copy in the monorepo root](https://github.com/Automattic/jetpack/blob/1452eef76cd1912bc3d2880f8596e73e2daf5d09/.editorconfig).

P.S. Probably not a matter for this PR, but I see Boost has its own redundant copy of this file too.

projects/plugins/inspect/.eslintrc.cjs
anomiex, line 1: Someday it would probably be nice to bring Boost's eslint ruleset more in line with the rest of the monorepo. Probably not a matter for this PR though.

Ideally something like this would work
```
module.exports = {
    extends: [ require.resolve( 'jetpack-js-tools/eslintrc/svelte' ) ],
    rules: {
        // Enforce the use of the jetpack-inspect textdomain.
        '@wordpress/i18n-text-domain': [
            'error',
            {
                allowedTextDomain: 'jetpack-inspect',
            },
        ],
    },
};
```
but in practice it seems to get confused in a few places.

projects/plugins/inspect/.gitignore
```diff
@@ -0,0 +1,10 @@
+.vscode
+*.swp
```
anomiex: If we're going to include a `*.swp` ignore, we should probably do it in the monorepo root instead of here.

projects/plugins/inspect/.gitignore
```diff
@@ -0,0 +1,10 @@
...
+node_modules
+public/build
+dist
+
+/vendor/
+/app-ui/build
+/composer.lock
```
anomiex: We normally check in composer.lock for plugins. Although offhand I don't recall why.

projects/plugins/inspect/bin/zip.sh
anomiex, line 1: I suspect this will not be needed in the monorepo. It seems intended to generate a plugin zip for a release, while in the monorepo the mirror repo serves that purpose.

If it is, it'll need editing to use pnpm instead of npm.

projects/plugins/inspect/composer.json
```diff
@@ -0,0 +1,65 @@
...
+		}
+	},
+	"authors": [
+		{
+			"name": "pyronaur",
+			"email": "hi@pyronaur.com"
+		}
+	],
+	"require": {
+		"automattic/jetpack-autoloader": "^2.11",
```
anomiex: Regarding all of these warnings, `@dev` would be preferable.

projects/plugins/inspect/composer.json
```diff
@@ -0,0 +1,65 @@
...
+		}
+	},
+	"scripts": {
+		"phpunit": [
+			"./vendor/phpunit/phpunit/phpunit --colors=always"
+		],
+		"test-php": [
+			"@composer phpunit"
+		],
+		"build-development": "echo 'Add your build step to composer.json, please!'"
```
anomiex: Pretty sure it'll need to do a build step here.

projects/plugins/inspect/composer.json
```diff
@@ -0,0 +1,65 @@
...
+			"url": "../../packages/*",
+			"options": {
+				"monorepo": true
+			}
+		}
+	],
+	"minimum-stability": "dev",
+	"prefer-stable": true,
+	"extra": {
+		"mirror-repo": "Automattic/jetpack-inspect",
```
anomiex, line 61: Currently that's the old private repo. What's the eventual plan?

projects/plugins/inspect/inspect.php
anomiex, line 1: Remove this in favor of `projects/plugins/inspect/jetpack-inspect.php`.

projects/plugins/inspect/jetpack-inspect.php
```diff
@@ -0,0 +1,74 @@
+<?php
```
anomiex, line 1: phpcs is confused here. Moving the `require_once` on line 14 down to below all the `use`s should fix it.

projects/plugins/inspect/package-lock.json
anomiex, line 1: This is a lock file for `npm`. We don't need or want it in the monorepo.

projects/plugins/inspect/package.json
```diff
@@ -0,0 +1,49 @@
...
+		"@wordpress/i18n": "4.35.0",
+		"postcss": "8.4.21",
+		"rollup": "2.79.1",
+		"rollup-plugin-livereload": "2.0.5",
+		"rollup-plugin-postcss": "4.0.2",
+		"rollup-plugin-svelte": "7.1.0",
+		"rollup-plugin-terser": "7.0.2",
+		"svelte-preprocess": "5.0.3",
+		"typescript": "5.0.4"
+	}
```
anomiex: Two changes here:
* `rollup-plugin-terser` should be replaced by `@rollup/plugin-terser`
* Adding in the optional peer dep on `tslib` avoids an extra entry in pnpm-lock.

```suggestion
	"devDependencies": {
		"@babel/core": "7.21.5",
		"@babel/preset-env": "7.21.5",
		"@rollup/plugin-alias": "5.0.0",
		"@rollup/plugin-commonjs": "19.0.2",
		"@rollup/plugin-node-resolve": "13.3.0",
		"@rollup/plugin-terser": "0.4.3",
		"@rollup/plugin-typescript": "8.3.3",
		"@wordpress/i18n": "4.35.0",
		"postcss": "8.4.21",
		"rollup": "2.79.1",
		"rollup-plugin-livereload": "2.0.5",
		"rollup-plugin-postcss": "4.0.2",
		"rollup-plugin-svelte": "7.1.0",
		"svelte-preprocess": "5.0.3",
		"tslib": "2.5.0",
		"typescript": "5.0.4"
	}
```

projects/plugins/inspect/rollup.config.js
```diff
@@ -0,0 +1,96 @@
+import alias from '@rollup/plugin-alias';
+import commonjs from '@rollup/plugin-commonjs';
+import resolve from '@rollup/plugin-node-resolve';
+import typescript from '@rollup/plugin-typescript';
+import livereload from 'rollup-plugin-livereload';
+import postcss from 'rollup-plugin-postcss';
+import svelte from 'rollup-plugin-svelte';
+import { terser } from 'rollup-plugin-terser';
```
anomiex: To go along with the switch to `@rollup/plugin-terser`,
```suggestion
import terser from '@rollup/plugin-terser';
```

projects/plugins/inspect/app-ui/src/Dashboard/Form.svelte
```diff
@@ -0,0 +1,181 @@
+<script type="ts">
+	import { createEventDispatcher } from "svelte";
+	import { slide } from "svelte/transition";
+	import { maybeStringify } from "../../../packages/Async_Option/scripts/utils";
+	import { API } from "../Options";
+	import { EntryData } from "../utils/ZodSchema";
+	import { createPersistentStore } from "../utils/peristentStore";
+	import FormError from "./FormError.svelte";
+	import type { LogEntry } from "@src/utils/ZodSchema";
```
anomiex: Missed a `@src`?

projects/plugins/inspect/packages/Async_Option/scripts/Options.ts
```diff
@@ -0,0 +1,99 @@
+// @TODO: need feedback on this: 👇
+import { writable } from 'svelte/store';
+import type { AsyncOptions as AO } from '@async-options/types';
```
anomiex: Missed an `@async-options`?

projects/plugins/inspect/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+	"exclude": [ "node_modules/*", "app-ui/build/*" ],
+	"compilerOptions": {
+		"strict": true,
+		"types": [],
+		"baseUrl": ".",
+		"paths": {
+			"@src/": [ "./app-ui/src/" ],
+			"@src/*": [ "./app-ui/src/*" ],
+			"@async-options/*": [ "./packages/Async_Option/scripts/*" ]
+		}
```
anomiex: May as well remove this since it looks like you got rid of almost all the uses of these in favor of importing the paths directly.

projects/plugins/inspect/rollup.config.js
```diff
@@ -0,0 +1,96 @@
...
+		name: 'app',
+		file: 'app-ui/build/jetpack-inspect.js',
+	},
+	plugins: [
+		alias( {
+			entries: [
+				// If you add a new top-level-folder besides src which you want to use, add it here
+				{ find: /^@src(\/|$)/, replacement: `${ __dirname }/app-ui/src/` },
+			],
+		} ),
```
anomiex: May as well remove this since it looks like you got rid of almost all the uses of these in favor of importing the paths directly.

That would also allow removing the dep on `@rollup/plugin-alias` from package.json.

anomiex (commented):
I see the code here seems to not be compatible with PHP < 7.4. If we want to go with that instead of fixing it, it might be useful to look at the following that CRM does to only support 7.2+:

* [tools/parallel-lint.sh#L30-L33](https://github.com/Automattic/jetpack/blob/1452eef76cd1912bc3d2880f8596e73e2daf5d09/tools/parallel-lint.sh#L30-L33)
* [projects/plugins/crm/.phpcs.dir.phpcompatibility.xml](https://github.com/Automattic/jetpack/blob/add/project-for-jetpack-inspect/projects/plugins/crm/.phpcs.dir.phpcompatibility.xml) (and also [projects/plugins/crm/.phpcs.dir.xml#L34-L36](https://github.com/Automattic/jetpack/blob/f75404e10240cc853005180cbbe3ee8b92635b2c/projects/plugins/crm/.phpcs.dir.xml#L34-L36))
* [projects/plugins/crm/tests/action-skip-test-php.sh](https://github.com/Automattic/jetpack/blob/add/project-for-jetpack-inspect/projects/plugins/crm/tests/action-skip-test-php.sh) (and [projects/plugins/crm/composer.json#L36](https://github.com/Automattic/jetpack/blob/1452eef76cd1912bc3d2880f8596e73e2daf5d09/projects/plugins/crm/composer.json#L36))

projects/plugins/inspect/.gitignore
zinigor: I can't find it anywhere now, but we went with some guideline telling us that we should check in lock files for all projects other than libraries that aren't getting shipped on their own.

projects/plugins/inspect/composer.json
zinigor: I didn't notice it was private, I thought we could open it and use as a mirror repo, since we're open sourcing the plugin anyway.

anomiex: If everything in the repo's history is ok to make public, that would be fine. In that case we might [import the history of the repo](https://github.com/Automattic/jetpack/blob/trunk/docs/monorepo.md#importing-an-existing-repo) too.

OTOH, if old commits there have data we'd want to keep private (I haven't checked), I think we could still rename the private repo and create a fresh mirror in its place.

thingalon:
@pyronaur is on sabbatical until 2023-08-15. He wrote this cool plugin, and I'm pretty sure he'd be happy to help port it to the monorepo when it lands back. :) 

Anyway, to help out I dove in and got the Svelte side compiling for you. But I can't commit it due to reasons explained below. Here is a gist of the changes I made to make Svelte compile all of the javascript happily: https://gist.github.com/thingalon/9833fe117c114d586219e5c2bd607edc

**Why can't I commit this fix myself?**

There's both a `projects/plugins/inspect/README.txt` AND a `projects/plugins/inspect/readme.txt` file in this PR. It's causing my git setup to go bananas:

image

My local clone is broken, and I can't stop `readme.txt` and `README.txt` from fighting on my filesystem. It insists there are unstashed changes, and won't let me switch branch, pull, merge, anything until that's resolved. If I try to `git reset` or `git restore` or even `git add` either readme file, git changes its mind about which version of the file is a problem and throws a tantrum.

I'm going to have to nuke, reclone and rebuild the monorepo. 😭

zinigor:
Yikes @thingalon ! Sorry for making you get up and nuke it from the orbit!

I have pushed a new commit that removes one of the readme files, I hope git can manage to correctly create the working dir state now. Thank you so much for your help with this!

anomiex:
> There's both a `projects/plugins/inspect/README.txt` AND a `projects/plugins/inspect/readme.txt` file in this PR. It's causing my git setup to go bananas:

TL;DR: Mac OS uses a case-insensitive file system be default. I thought git [was supposed to be a bit smarter about it](https://www.git-scm.com/docs/git-config/2.14.6#Documentation/git-config.txt-coreignoreCase), but I guess not.

> If I try to `git reset` or `git restore` or even `git add` either readme file, git changes its mind about which version of the file is a problem and throws a tantrum.
> 
> I'm going to have to nuke, reclone and rebuild the monorepo. sob

I wonder if `git checkout -f trunk` would have worked.

Now I'm tempted to (1) see if I can make a docker container with a case-insensitive file system to try this and (2) figure out some way to have pre-push and/or CI warn us Linux users if the push is going to cause this issue.

thingalon:
Ok, I just merged my gist in 8c1ea7e49a, and it now compiles the Svelte into an app bundle for you :) 

Just run `pnpm run build-js` in the inspect directory and you have a healthy bundle ready to go!

projects/plugins/inspect/package.json
```diff
@@ -16,17 +16,19 @@
...
-		"build-production-js": "echo 'Not implemented.'",
+		"build-production-js": "rollup -c rollup.config.js",
 		"clean": "true"
 	},
 	"engines": {
 		"node": "^18.13.0",
 		"yarn": "use pnpm instead - see docs/yarn-upgrade.md"
 	},
 	"dependencies": {
+		"node-sass": "^9.0.0",
```
anomiex: `node-sass` was deprecated three years ago and should not be needed. `sass` should suffice.

projects/plugins/inspect/app/Monitor.php
```diff
@@ -50,21 +92,24 @@ protected function match_request_filter( $url ): bool {
 			return true;
 		}
 
-		// https://example.com/?foo=bar will match "*example[s].com*
+		// https://example.com/?foo=bar will match "*example[s].com*.
 		if ( str_contains( $filter, '*' ) || ( str_contains( $filter, '[' ) && str_contains( $filter, ']' ) ) ) {
```
anomiex, line 96: It's ok to ignore `PHPCompatibility.FunctionUse.NewFunctions.str_containsFound`, WordPress core has had a polyfill for it since 5.9 and the plugin's readme requires 6.1+.

Feel free to add it with a comment to that effect in `.phpcs.dir.phpcompatibility.xml`.

projects/plugins/inspect/.phpcs.dir.phpcompatibility.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0"?>
+<ruleset>
+	<rule ref="PHPCompatibility.Constants.NewConstants">
```
anomiex: I'm wary of ignoring all of `PHPCompatibility.Constants.NewConstants` and `PHPCompatibility.Operators.NewOperators`. If there are new constants or operators added in 8.0+, we'd still want those warnings.

`PHPCompatibility.Syntax.NewFunctionCallTrailingComma` is probably ok, from the name this sniff seems likely to only apply to the one feature added in 7.3. OTOH, it would be just as easy to follow the rest and specify `PHPCompatibility.Syntax.NewFunctionCallTrailingComma.FoundInFunctionCall` fully for that too.

anomiex: Made #31711 that should make this sort of thing easier to get right.

anomiex: That got merged, we should use it here now.

projects/plugins/inspect/app/Monitor.php
zinigor: Thanks, I added it in one of the last commits!

projects/plugins/inspect/.phpcs.dir.phpcompatibility.xml
zinigor: Done, thank you!

anomiex (commented):
Code structure seems ok now. [Mirror repo situation](https://github.com/Automattic/jetpack/pull/31551/files/1452eef76cd1912bc3d2880f8596e73e2daf5d09#r1242221758) still needs figuring out though.

And then adding it to the betadownload site once we confirm what the mirror repo name will be.

zinigor:
It looks like a prettify run has also changed some files on merging trunk. But nevertheless, I have implemented the needed changes to the jetpack-inspect plugin repository, can you take a look again please?

projects/packages/forms/changelog/add-project-for-jetpack-inspect
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: added
+
+Added a new project for Jetpack Inspect.
```
anomiex: Change entry should reflect the effect on _this_ project.
```suggestion
Significance: patch
Type: added
Comment: Prettier reformatted some code. No change to the code itself.

```
Of course, doing that will then take the version back down to 0.19.8-alpha, so all the bumps below will get undone too.

projects/packages/jetpack-mu-wpcom/changelog/add-project-for-jetpack-inspect
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: added
+
+Added a new project for Jetpack Inspect.
```
anomiex: Change entry should reflect the effect on _this_ project.
```suggestion
Significance: patch
Type: added
Comment: Prettier reformatted some code. No change to the code itself.

```

projects/plugins/jetpack/changelog/add-project-for-jetpack-inspect
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: other
+
+Added a new project for Jetpack Inspect.
```
anomiex: Change entry should reflect the effect on _this_ project.
```suggestion
Significance: patch
Type: other
Comment: Prettier reformatted some code. No change to the code itself.

```

anomiex (commented):
> It looks like a prettify run has also changed some files on merging trunk.

I hate when that happens. I try to always remember to do like `HUSKY=0 git merge` to stop it from doing that. Unfortunately I don't think there's a way to have the hook detect that it's a merge commit.

projects/plugins/inspect/composer.json
```diff
@@ -0,0 +1,56 @@
...
+		"changelogger": {
+			"link-template": "https://github.com/Automattic/jetpack-inspect/compare/v${old}...v${new}",
+			"versioning": "semver"
+		},
+		"release-branch-prefix": "inspect"
```
anomiex: Need to configure a few more things here
```suggestion
		"autorelease": true,
		"autotagger": true,
		"mirror-repo": "Automattic/jetpack-inspect",
		"changelogger": {
			"link-template": "https://github.com/Automattic/jetpack-inspect/compare/v${old}...v${new}",
			"versioning": "semver"
		},
		"release-branch-prefix": "inspect",
		"beta-plugin-slug": "jetpack-inspect"
```

projects/plugins/inspect/composer.json
```diff
@@ -0,0 +1,60 @@
...
+		"autorelease": true,
+		"autotagger": true,
+		"mirror-repo": "Automattic/jetpack-inspect",
+		"changelogger": {
+			"link-template": "https://github.com/Automattic/jetpack-inspect/compare/v${old}...v${new}",
+			"versioning": "semver"
+		},
+		"release-branch-prefix": "inspect",
+		"beta-plugin-slug": "jetpack-inspect",
+		"wp-svn-autopublish": true
```
anomiex: On second thought, shouldn't have that until it's ready to publish
```suggestion
		"beta-plugin-slug": "jetpack-inspect"
```

anomiex (dismissed):

anomiex (approved):

PR #31563, opened 2023-06-23
AI Extension: extend the block when the edit post store is undefined (P2)

retrofox:
The edit post store is not defined in P2 sites. We can't rely on it to check if there are hidden blocks. This PR checks if the store and the function to detect hidden blocks are defined. If so, it relies on it. Otherwise, the app won't check the hidden blocks. Follow-up PR to handle this for P2 sites is coming on.
Fixes #

## Proposed changes:

* AI Extension: extend the block when the edit post store is undefined (P2)

### Other information:

## Testing instructions:

* Test with a Simple site
* Confirm the AI Extension depends on the AI block availability
* Confirm the extension is visible for P2 sites

### DIFF
--- a/projects/plugins/jetpack/changelog/update-ai-extension-handle-hidden-in-p2
+++ b/projects/plugins/jetpack/changelog/update-ai-extension-handle-hidden-in-p2
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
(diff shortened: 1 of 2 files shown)


dhasilva (approved):
Working on P2

lhkowalski (approved):
Working on Jetpack site.

PR #31565, opened 2023-06-25
Publicize Components: Refactor the media validation

pablinos:
We need to change the way we validate the media to be shared. Previously
we checked the lowest requirements for all enabled connections, but
given the way that Instagram requires an image we were forcing the media
to be valid for Instagram, which is currently only a JPG.

The idea around this change is to disable connections based on whether the selected media is valid or not, rather than removing the media. Later we can provide options to convert the image or video to fit within the restrictions, but for now we'll disable theme with an appropriate message.

## Proposed changes:
This change reworks the validation so we do it on a per connection basis
and disable the individual connections if the selected media is invalid.

### Other information:

No.

## Testing instructions:

Check the tests pass:
```bash
jetpack test js-packages/publicize-components js
```

* Create a test site with Social sharing enabled
* Add one or more social connections
* Optionally enable Instagram on the site and add a test instagram connection - credentials can be provided for this
* Add and remove media, enable and disable Social Image Generator
* In each state check that connections for which the media is invalid is disabled and an appropriate notice is displayed. For example, with no image Instagram can't be shared to, an image larger than 4MB can't be shared to Facebook etc.
* Save or publish the post and check the metadata. For any disabled connections there should exist a key of `_wpas_skip_` with the connection ID


spsiddarthan (approved):
This looks good and tests well!

spsiddarthan (changes requested):
Testing it some more for some other combinations, don't merge this yet!

spsiddarthan (approved):
This looks good, now the connections are getting disabled correctly when the requirement is not met and the connections that should be skipped are not stamped on.

samiff:
cherry-picked to `monthly/branch-2023-06-26` in cd614e300936864687cc0b927b4631567b728790

PR #31572, opened 2023-06-26
tools: Clean up class-jetpack-phpcs-exclude-filter.php

anomiex:
## Proposed changes:

* Remove some leftover code from before we switched the exclude file to json.
* Use `$this->filterBaseDir` (from the parent class) instead of `$this->basedir` as the former works better if you supply paths on the command line.

### Other information:

## Jetpack product discussion

None

No

## Testing instructions:

* Does `composer phpcs:lint:required` still check just the non-excluded files?


zinigor (approved):
It's probably fine, judging by how `filterBaseDir` is processed inside PHPCS code.

PR #31573, opened 2023-06-26
phpcs: DRY in .phpcs.dir.phpcompatibility.xml

anomiex:
## Proposed changes:

For cases where we want to ignore a PHPCompatibility rule, we need to ignore it in both `.phpcs.dir.xml` and `.phpcs.dir.phpcompatibility.xml`.

But there's no need to do it via copy-paste. We can have `.phpcs.dir.xml` load `.phpcs.dir.phpcompatibility.xml` so we only have to have the ignores in one place.

### Other information:

## Jetpack product discussion

None

No

## Testing instructions:

* Is CI happy?


zinigor (approved):
TIL you can do that with PHPCS rulesets, nice!

PR #31577, opened 2023-06-26
Tock Block: avoid PHP warning when restaurant name isn't set yet

jeherve:
## Proposed changes:

If you create the block but do not fill in a restaurant name, you'll get a PHP warning when viewing the post:

```
PHP Warning:  Undefined array key "url" in /usr/local/src/jetpack-monorepo/projects/plugins/jetpack/extensions/blocks/tock/tock.php on line 60
```

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

* Enable Beta blocks on your site.
* Go to Posts > Add New
* Add a Tock block
* Do not enter a restaurant name.
* Publish your post.
* Visit the post on the frontend. If you've added other blocks, they should be displayed with no issues. You should only see a warning instead of the Tock block, when logged in as an admin.

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-tock-warning
+++ b/projects/plugins/jetpack/changelog/fix-tock-warning
@@ -1,0 +1,4 @@
+Significance: patch
+Type: other
+
+Tock Block: avoid PHP warning when restaurant name isn't set.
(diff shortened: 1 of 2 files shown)


anomiex (approved):
Looks reasonable. There's a slight behavior change in that the block itself doesn't appear anymore when viewing as an admin (just the warning), but since it doesn't seem to function in that case anyway maybe that's ok.

And if we did still want to display the block, it might be less confusing for the warning to go above the rest of the block rather than below it.

PR #31583, opened 2023-06-26
Update postcss-loader configs

anomiex:
## Proposed changes:

A few years ago postcss-loader added the ability to specify plugins as an array of plugin objects rather than as an object mapping plugin names to configs.

Taking advantage of this lets us not have to use `NODE_PATH` for webpack invocations all over the place to allow postcss-loader to find its plugins.

### Other information:

## Jetpack product discussion

None

No

## Testing instructions:

* Is the build happy?
* Generated assets should not be changed.


zinigor (approved):
Code looks good to me, I have tested the affected plugins: builds seem to be working fine, the plugins themselves look fine on the site too.

PR #31584, opened 2023-06-26
AI Extension: avoid error notice stacking

### LINKED ISSUE #31569: AI Extension: Clean previous error before to add a new one
The app currently has a problem where it does not clear previous errors, resulting in multiple error messages appearing when something fails.

Image

Let's try to clean up the previous message.

lhkowalski:
Fixes #31569

## Proposed changes:

* Start adding a notice ID on the notices generated by the extension; if we use the same ID, any previous notice is replaced with the new one
* Use `ai-assistant` as the ID

### Other information:

## Jetpack product discussion

Discussion: p1687558335821679-slack-C054LN8RNVA

## Testing instructions:

* To test this, you need to force some errors on the backend; refer to the "How to force errors on the backend" section below to see how to do it
* With the backend changed to force an error return, go to the block editor
* Write some content on a paragraph
* Click the AI Assistant extension button and select some action:

Screen Shot 2023-06-26 at 15 57 35

* Since you are forcing an error on the backend, you will see a notice on the top of the editor:

Screen Shot 2023-06-26 at 15 58 10

* Execute another action
* Since you are still forcing backend errors, another notice will be generated
* Before the fix, a second notice would be showed on top of the first; any new notice would be stacked upon the previous ones

image

* After the fix, the new notice will replace the existing one, in the case there was one; only notices generated by the AI Extension will be replaced:

_Request 1:_
 
Screen Shot 2023-06-26 at 15 54 48

_Request 2:_

Screen Shot 2023-06-26 at 15 55 17

* Experiment changing the `status` code returned on the backend error to the following values and repeat the test
   * 422
   * 429
   * 500
   * 503
* Each `status` code has it's own message on the frontend, so you will see different messages on each occasion, and the new message will replace the existing one

### How to force errors on the backend

* Sandbox your `public-api`
* If you are running this branch on JN, don't forget to point it to your sandbox
* Update the `jetpack-ai-query.php` file on your sandbox to add the following snippet:

```
return new \WP_Error( 'error_jetpack_ai', __( 'Generic error message.' ), [ 'status' => 503 ] );
```

* Add it as the first line of the function:

Screen Shot 2023-06-21 at 19 29 33

* You will test changing the `status` value, so keep the file open and ready to edit

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-ai-extension-avoid-error-notice-stacking
+++ b/projects/plugins/jetpack/changelog/fix-ai-extension-avoid-error-notice-stacking
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
+AI Extension: Use ID on error notices to prevent stacking multiple notices.
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/ai-assistant/with-ai-assistant.tsx
@@ -32,6 +32,13 @@ type StoredPromptProps = {
 };
 
 /*
+ * An identifier to use on the extension error notices,
+ * so a existing notice with the same ID gets replaced
+ * by a new one, avoiding the stacking of notices.
+ */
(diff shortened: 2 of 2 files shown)


retrofox (approved):
LGTM!

image

I have encountered multiple errors, but the app is only displaying one.

PR #31561, opened 2023-06-23
AI Assistant: Fix empty content on P2

dhasilva:
Fixes #31555
Depends on #31563

## Proposed changes:

* Changes the `'core/block-editor'` constant to a string. It seems that P2 does not load correctly using the import

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* Test locally
* build and `rsync` the changes to your sandbox
* Sandbox a P2 site
* In a post or comment, use the extension
* Check that the content is sent

PR #31585, opened 2023-06-26
AI Assistant: Replace imported store constants with strings

### LINKED ISSUE #31582: AI Assistant: do not use imported const to define store IDs
It appears that the store constants are not defined for P2 sites, for some unknown reason.
As a temporary solution, we have replaced them with strings directly. However, we should update them to strings completely to avoid any potential issues in the future.
Furthermore, it would be beneficial to investigate why these constants are not defined in the P2 context.

dhasilva:
Fixes #31582

## Proposed changes:

* Removes imported stores and uses strings directly
ref: https://github.com/Automattic/jetpack/pull/31561

### Other information:

## Testing instructions:

* Use the AI Assistant block and extensions
* There should be no regressions

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-ai-assistant-store-ids
+++ b/projects/plugins/jetpack/changelog/fix-ai-assistant-store-ids
@@ -1,0 +1,4 @@
+Significance: patch
+Type: other
+
+AI Assistant: Replace imported store constants with strings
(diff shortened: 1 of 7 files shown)


retrofox (approved):
LGTM! Thanks for it 💯

lhkowalski (approved):
Tested on a Jetpack site, covering the assistant and extension flows, and nothing broken came up. 🚀

PR #31592, opened 2023-06-26
Jetpack: backport 12.3-beta release

samiff:
## Proposed changes:

- Jetpack: backport 12.3-beta release

### Other information:

## Jetpack product discussion
n/a

No.

## Testing instructions:

Proofread.

### DIFF
--- a/projects/plugins/jetpack/readme.txt
+++ b/projects/plugins/jetpack/readme.txt
@@ -293,40 +293,104 @@ Jetpack Backup can do a full website migration to a new host, migrate theme file
 
 
 == Changelog ==
-### 12.3-a.7 - 2023-06-19
+### 12.3-beta - 2023-06-26
 #### Enhancements
 - AI Assistant: add and use ImproveToolbarDropdownMenu in block toolbar.
+- AI Assistant: add a specific delimiter for content in the prompts.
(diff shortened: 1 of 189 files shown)


projects/plugins/jetpack/readme.txt
```diff
@@ -293,40 +293,104 @@ Jetpack Backup can do a full website migration to a new host, migrate theme file
 
 
 == Changelog ==
-### 12.3-a.7 - 2023-06-19
+### 12.3-beta - 2023-06-26
 #### Enhancements
 - AI Assistant: add and use ImproveToolbarDropdownMenu in block toolbar.
+- AI Assistant: add a specific delimiter for content in the prompts.
```
jeherve, line 299: I don't know if we need all those entries in the final readme.txt; that's a bit overwhelming. I would rather see groups of related changes that are easier to parse for users.

jeherve (approved):
Approving, but I think it would make sense to summarize the AI entries in the final readme.txt, in a follow-up maybe.

projects/plugins/jetpack/readme.txt
samiff: https://github.com/Automattic/jetpack/pull/31600#issuecomment-1609690430

PR #31596, opened 2023-06-27
AI Extension: add a way to cancel the AI request

### LINKED ISSUE #31590: AI Extension: Implement a way to cancel the current suggesting process.
At the moment, it's not feasible to cancel the request process within the AI Extension. Let's brainstorm solutions to address this issue.

retrofox:
This PR introduces a functionality to cancel a request in the AI Extension dropdown. Here's an overview:

* We've integrated a cancelRequest() helper function into the suggestion hook to handle request cancellation.
* We've optimized the onClick event of the toolbar button to enable the cancellation process.
* We've utilized the isActive state to represent the ongoing request status.

Fixes issue #31590

This PR implements a way to cancel the request when the user uses the AI Extension dropdown. In summary:

* It adds a `cancelRequest()` helper to the suggestion hook
* propagates the helper between the components 
* Improve the onClick event of the toolbar button to be able to cancel the request
* It uses the `isActive` state to denote the requesting process.

Fixes https://github.com/Automattic/jetpack/issues/31590

## Proposed changes:

* AI Extension: add a way to cancel the AI request

### Other information:

## Testing instructions:

1. Navigate to the block editor.
2. Create a Paragraph and add some content.
3. Use the AI dropdown to request content.
4. Verify that you can cancel the ongoing request by re-clicking the toolbar button.

https://github.com/Automattic/jetpack/assets/77539/cd150142-afcd-434d-bd3d-22784b9b7df3

### DIFF
--- a/projects/plugins/jetpack/changelog/update-ai-extension-be-able-to-cancel-request
+++ b/projects/plugins/jetpack/changelog/update-ai-extension-be-able-to-cancel-request
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
+AI Extension: add a way to cancel the AI request
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/components/ai-assistant-controls/index.tsx
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/components/ai-assistant-controls/index.tsx
@@ -119,6 +119,8 @@ type AiAssistantControlComponentProps = {
 	onChange: ( item: PromptTypeProp, options?: AiAssistantDropdownOnChangeOptionsArgProps ) => void;
 
 	onReplace: () => void;
+
+	onCancel: () => void;
 };
 
 export default function AiAssistantDropdown( {
@@ -129,6 +131,7 @@ export default function AiAssistantDropdown( {
 	disabled,
 	onChange,
 	onReplace,
+	onCancel,
 }: AiAssistantControlComponentProps ) {
 	const quickActionsListFiltered = quickActionsList.filter(
 		quickAction => ! exclude.includes( quickAction.key )
@@ -150,7 +153,13 @@ export default function AiAssistantDropdown( {
 							[ `is-${ requestingState }` ]: true,
 						} ) }
 						showTooltip
-						onClick={ onToggle }
+						onClick={ () => {
+							if ( ! isProcessingRequest ) {
+								onToggle();
+							}
+
+							onCancel?.();
+						} }
 						aria-haspopup="true"
 						aria-expanded={ isOpen }
 						label={ toolbarLabel }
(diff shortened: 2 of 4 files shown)


keoshi:
As mentioned in Slack, I think this is an acceptable solution for the moment. It is hidden, but totally ok to leave it as a pro user feature that people will have to explore and find for themselves. The animation draws the eye in and gives it a nice affordance, so I'm happy about that.

For a future iteration, I'd like to hear from our experts @jasmussen @jay on whether there's an existing visual mechanism (or component, or pattern) we could make use of here to send tips (i.e.: temporary notices) to our users letting them know of power features such as this one.

_For additional context_: what we're trying to communicate here is that users can click the AI button a second time to cancel the prompt.

jasmussen:
> It is hidden, but totally ok to leave it as a pro user feature that people will have to explore and find for themselves. The animation draws the eye in and gives it a nice affordance, so I'm happy about that.

I'd echo this, if I ever found a need to stop the process I'd probably intuit clicking that toggle too.

> For a future iteration, I'd like to hear from our experts @jasmussen @jay on whether there's an existing visual mechanism (or component, or pattern) we could make use of here to send tips (i.e.: temporary notices) to our users letting them know of power features such as this one.

For the moment, there is not. We used to have pointers, but they didn't work very well. The closest we have now are the snackbar and the notice which exists in a gray minimal version; this notice you see in the split inspector informing you where your settings went. So as far as existing notices, you might best fire a snackbar initially or in context of a particular action. So long as you fire it contextually and not all the time, it can probably be a easy good solution.

There are some cursory explorations around showing a dot indicator, for example in context of [broken](https://github.com/WordPress/gutenberg/issues/51980#issuecomment-1610827401) or incomplete blocks, or around [responsive editing](https://github.com/WordPress/gutenberg/issues/19909), but no concrete plans at the moment.

retrofox:
Thanks, folks, for the feedback. 

To begin with, we have made significant changes to the implementation from the user interface perspective. Now, when the user selects an option from the dropdown, the app replaces the extended block (paragraph, heading) to an AI Assistant instance. Sorry about that

Anyway, I'd like to add some comments.

>  The closest we have now are the snackbar and the notice which exists in a gray minimal version;

We explored and talked about this idea. The basic conclusion was a Notice seems to be like a response from an action that the user could do:

Save the post -> The notice shows a message
Something went wrong -> A warning notice shows up

In this scenario, it's important to inform the user that they have the option to cancel the request process by clicking on the active toolbar button. It's like we should find out a way to inform about it ahead of time.

jasmussen:
Yep, I'll defer to Filipe on the importance of highlighting the stop ability. You could even change the AI icon to a "Stop" icon (square), or add a button when it's working, though the latter would cause a layout shift.

But just to be clear I was referring to [the snackbar](https://wordpress.github.io/gutenberg/?path=/story/components-snackbar--default), not the [notice](https://wordpress.github.io/gutenberg/?path=/story/components-notice--default).

retrofox:
> But just to be clear I was referring to [the snackbar](https://wordpress.github.io/gutenberg/?path=/story/components-snackbar--default), not the [notice](https://wordpress.github.io/gutenberg/?path=/story/components-notice--default).

Correct, the bar that shows up when the user saves the post, for instance. This bar is always anchored in the same place, right (bottom, left)

retrofox:
> You could even change the AI icon to a "Stop" icon (square)

~that would be the coup de grace~ If we change the icon, we lose the requesting animation..._ 😬

PR #31603, opened 2023-06-27
AI Extension: transform to AI Assistant block when requesting a suggestion

### LINKED ISSUE #31599: AI Extensions: transform to block for all dropdown options
Instead of updating the extended block, we'd like to:

* Transform the block for any options that the extensions show
* Trigger on the fly the request, but now in the new AI Assistant block instance

Image

1. Transform to AI Assistant block instance
2. Trigger the request on the fly. `Simplify` in the example image above.

### LINKED ISSUE #31819: AI Assistant: Previous text selection is kept after new response
ref p8oabR-1eP-p2#comment-7397

Suggestion: If the selected text is exactly the same, find its new position and select it. Otherwise deselect everything.

retrofox:
With the changes added to this PR, the extended blocks are transformed into AI Assistant block instances. When it happens, the block performs an auto-request based on the original content of the block.

Fixes https://github.com/Automattic/jetpack/issues/31599
Fixes https://github.com/Automattic/jetpack/issues/31819

## Proposed changes:

* AI Extension: transform to AI Assistant block when requesting a suggestion

### Other information:

## Testing instructions:

* Go to 'the block editor
* Create some context using a Paragraph block
* Change its content by using the AI Extension dropdown
* Confirm the core block transforms to an AI Assistant block
* Confirm it automatically requests the suggestion
* Confirm it continues working as usual
* Confirm it doesn't perform the automatic request when the block is transformed using the `Ask AI Assistant` of the Transform block tool.

https://github.com/Automattic/jetpack/assets/77539/f31bdc84-b4e1-470c-a884-7f93e2a739bd

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/edit.js
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/edit.js
@@ -110,6 +110,24 @@ export default function AIAssistantEdit( { attributes, setAttributes, clientId }
 		requireUpgrade,
 	} );
 
+	// Auto request if the block has the autoRequestPrompt attribute
+	useEffect( () => {
+		if ( ! attributes?.autoRequestPrompt?.type ) {
+			return;
+		}
+
+		const { type, options } = attributes.autoRequestPrompt;
+		setAttributes( { autoRequestPrompt: undefined } );
+
+		/*
+		 * Add a small delay to race condition with the block editor.
+		 * Todo: Let's clean this up when we have a better way to handle this.
+		 */
+		setTimeout( () => {
+			getSuggestionFromOpenAI( type, options );
+		}, 200 );
+	}, [ attributes.autoRequestPrompt, getSuggestionFromOpenAI, setAttributes ] );
+
 	useEffect( () => {
 		if ( errorData ) {
 			setErrorDismissed( false );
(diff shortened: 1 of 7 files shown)


lhkowalski:
I was able to test it, using all the features, and it seems to work just fine:

- the action does what it is supposed to do (all of them)
- the assistant box shows up correctly
- the request starts automatically when it needs to start (all actions except the “Ask …” one)
- the result is showed correctly

The small issues that I found are all related to mixed content from outside the paragraph being included on the response, due to the prompts providing the post content as context. This does not happen always, so like @retrofox said before, it’s a matter of improving the prompt building.

The actions work well, though, well enough to do what they are supposed to do.

projects/plugins/jetpack/extensions/blocks/ai-assistant/edit.js
```diff
@@ -91,6 +92,30 @@ export default function AIAssistantEdit( { attributes, setAttributes, clientId }
 		requireUpgrade,
 	} );
 
+	/*
+	 * Auto request the prompt if we detect
+	 * it was previsulay defined in the local storage.
```
lhkowalski:
```suggestion
	 * it was previously defined in the local storage.
```

lhkowalski:
Tested on a Simple site and worked as expected.

lhkowalski:
Tested on a p2 site, both on the content editor and on the comment box, and it's working nicely.

dhasilva (approved):
LGTM. One minor issue: When transforming two or more headings, only the first one remain as a header, while the others are converted to paragraphs.

PR #31607, opened 2023-06-27
Update storybook config to work around some bugs

anomiex:
## Proposed changes:

There are two module resolution bugs to work around here to be able to get rid of the use of `NODE_PATH`.

* Hack to pass a full path to the module's base for framework.name.
* Override an option for addon-docs to set a full path where the default relies on hoisting to resolve.

Also we need to clean up a few cases where `NODE_PATH` was hiding some missing deps in packages, and we may as well clean up the places using `@storybook/addon-docs` where they should be using `@storybook/blocks` too.

### Other information:

## Jetpack product discussion

None

No

## Testing instructions:

* Is CI happy?
* Check that the stories in the storybook still function as they did before, either running `pnpm storybook:dev` in `projects/js-packages/storybook` or by uploading the artifact to a webpage somewhere and accessing it that way.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -316,6 +316,9 @@ importers:
       resize-observer-polyfill:
         specifier: 1.5.1
         version: 1.5.1
+      ts-dedent:
+        specifier: 2.2.0
+        version: 2.2.0
       typescript:
         specifier: 5.0.4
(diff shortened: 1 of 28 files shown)


sergeymitr (approved):
Works great 👍

PR #31613, opened 2023-06-28
AI Extensions: Check for null and undefined blocks

nunyvega:
Fixes https://github.com/Automattic/p2/issues/6150 , where  `is select( 'core/block-editor' ).getBlocksByClientId( clientIds ); `returns an array with null values, and is breaking lists and certain following blocks on the p2 comments box

## Proposed changes:

* Add a safeguard check for null and undefined values.

### Other information:

## Jetpack product discussion

p1687946664807469-slack-C054LN8RNVA

No

## Testing instructions:

* Apply this patch in your sandbox
* Point a p2 page to your sandbox and confirm that the error does not occur anymore
* Test it on a simple site in the editor

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/utils/block-content.ts
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/utils/block-content.ts
@@ -84,7 +84,10 @@ export function getTextContentFromSelectedBlocks(): GetTextContentFromBlocksProp
 		count: blocks.length,
 		clientIds,
 		content: blocks
-			? blocks.map( block => getBlockTextContent( block.clientId ) ).join( HTML_JOIN_CHARACTERS )
+			? blocks
+					.filter( block => block !== null && block !== undefined ) // Safeguard against null or undefined blocks
+					.map( block => getBlockTextContent( block.clientId ) )
+					.join( HTML_JOIN_CHARACTERS )
 			: '',
 	};
 }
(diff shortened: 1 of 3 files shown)


projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/utils/block-content.ts
```diff
@@ -84,7 +84,10 @@ export function getTextContentFromSelectedBlocks(): GetTextContentFromBlocksProp
 		count: blocks.length,
 		clientIds,
 		content: blocks
-			? blocks.map( block => getBlockTextContent( block.clientId ) ).join( HTML_JOIN_CHARACTERS )
+			? blocks
+					.filter( block => block !== null && block !== undefined ) // Safeguard against null or undefined blocks
```
retrofox, line 88: Should we make the condition stronger, checking, for instance, the name object?
```suggestion
					.filter( block => !! block?.name ) // Safeguard against null or undefined blocks
```

retrofox (approved):
Left a suggestion. Feel free to ignore it. Nice fix!

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/utils/block-content.ts
dhasilva: `block !== null && block !== undefined` is the same as `block != null`

samiff:
cherry-picked to `monthly/branch-2023-06-26` in 39badef8fa34c37a8daf055e13851dbc883ce8fb

PR #31617, opened 2023-06-28
Blaze: ensure the Dashboard menu is only inserted once.

jeherve:
## Proposed changes:

This fixes issues on WoA sites and WordPress.com Simple sites, where when the new dashboard is enabled, the navigation menu (under Tools > Advertising) is added by the package, while the old Calypso link is also added, via the masterbar.

To solve this problem, we rely on the package, and only the package, to register the menu. **We do point WoA and simple site owners to the Calypso advertising page (and not the wp-admin advertising page) at all times, since they have nav-unification on and are logged in to WordPress. com at all times.**

This PR also changes the order of the new menu added by the package, to ensure it's always added at the top of the Tools menu.

### Other information:

## Jetpack product discussion

* p1687955577164529/1687894794.905929-slack-C053FAERCDP

* No

## Testing instructions:

You'll want to test this on all three site types (WoA, Simple, self-hosted).

- On WoA sites, you should see a Tools > Advertising menu item pointing you to Calypso at all times.
- On Simple sites, you should see a Tools > Advertising menu item pointing you to Calypso at all times.
- On self-hosted sites, you should not see any Tools > Advertising menu item by default.
    - If you enable the Blaze feature under Jetpack > Settings > Traffic, you should still see no new menu item.
    - If you enable the Blaze dashboard with `add_filter( 'jetpack_blaze_dashboard_enable', '__return_true' );`, you should see a new menu item, pointing you to the wp-admin advertising page.

### DIFF
--- a/projects/plugins/jetpack/modules/masterbar/admin-menu/class-atomic-admin-menu.php
+++ b/projects/plugins/jetpack/modules/masterbar/admin-menu/class-atomic-admin-menu.php
@@ -382,7 +382,7 @@ class Atomic_Admin_Menu extends Admin_Menu {
 		// performance settings already have a link to Page Optimize settings page.
 		$this->hide_submenu_page( 'options-general.php', 'page-optimize' );
 
-		if ( Blaze::should_initialize() ) {
+		if ( Blaze::should_initialize() && ! Blaze::is_dashboard_enabled() ) {
 			add_submenu_page( 'tools.php', esc_attr__( 'Advertising', 'jetpack' ), __( 'Advertising', 'jetpack' ), 'manage_options', 'https://wordpress.com/advertising/' . $this->domain, null, 1 );
 		}
 	}
--- a/projects/plugins/jetpack/modules/masterbar/admin-menu/class-atomic-admin-menu.php
+++ b/projects/plugins/jetpack/modules/masterbar/admin-menu/class-atomic-admin-menu.php
@@ -7,7 +7,6 @@
 
 namespace Automattic\Jetpack\Dashboard_Customizations;
 
-use Automattic\Jetpack\Blaze;
 use Automattic\Jetpack\Connection\Client;
 use Automattic\Jetpack\Current_Plan as Jetpack_Plan;
 use Automattic\Jetpack\Status;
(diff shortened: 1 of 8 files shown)


projects/plugins/jetpack/modules/masterbar/admin-menu/class-atomic-admin-menu.php
```diff
@@ -381,10 +380,6 @@ public function add_options_menu() {
 		// would conflict with our own Settings > Performance that links to Calypso, so we hide it it since the Calypso
 		// performance settings already have a link to Page Optimize settings page.
 		$this->hide_submenu_page( 'options-general.php', 'page-optimize' );
-
-		if ( Blaze::should_initialize() ) {
```
sbarbosa, line 385: I checked and this code is now inside the Blaze package. This means that the Advertising menu will appear only if the Blaze module is enabled. Right?

I think it is already like this right now, you are just moving this code to the blaze package.

sbarbosa (dismissed):
I tested in simple/atomic and self-hosted sites, and the menu option always pointed to the correct place. (wpcom for simple/atomic, and new dashboard for self-hosted).

The code looks good 👍

sbarbosa:
One more question @jeherve, after we merged/deployed this change. Would it be a way to test the Blaze Dashboard on atomic sites? 
I just want to know how can we proceed with the testing after this. Thanks!

projects/plugins/jetpack/modules/masterbar/admin-menu/class-atomic-admin-menu.php
jeherve: That's correct, yes.

jeherve:
> Would it be a way to test the Blaze Dashboard on atomic sites?
> I just want to know how can we proceed with the testing after this. Thanks!

That's a very good point, and something I should have addressed in this PR. Since the menu is not registered on WordPress.com sites, you cannot and should not be able to get to it. Right now you still can, via the quick links in the posts list, or via the post-publish panel. I'll get this fixed.

sbarbosa (approved):

PR #31618, opened 2023-06-28
Connection: restore invalid connection owner

sergeymitr:
## Proposed changes:
* Trigger an error if connection owner ID stored by Jetpack is invalid
* Display the "Restore Connection" notice to let the user fix the issue

### Other information:

## Jetpack product discussion
1204904295387809-as-1204774058674945

No.

### Standalone
_If the error doesn't show up, try removing the transient:
`wp option delete _transient_jetpack_connection_error_reporting_gate_invalid_connection_owner`_

1. Connect a Jetpack site.
2. Use "Broken Token" to randomize connection owner.
3. Go to "Jetpack -> Dashboard", you should see the "Your connection with WordPress.com seems to be broken." notice.
4. Click "Restore Connection", you should see the "Reconnecting Jetpack" notice on the top right, and user connection screen.
5. Click "Connect your user account", finish the connection flow.
6. Use "Broken Token" to confirm that connection owner ID is now correct.

### WoA
Pretty much the same as standalone, but you'll need to:
1. Rebuild/upload the WPCOMSH counterpart PR.
2. Use Jetpack Beta to checkout the branch.


bindlegirl:
I tested on JN so far and I have a question about My Jetpack. Is this OK: 

Screenshot 2023-06-29 at 13 36 54

Do we want to remove the Connected as XX part since the user connection is broken? The token is still there so is this message expected?

Another thing I found was the error message gets cached and it disappears once I visit My Jetpack page. The connection is still broken but I see no warning except on My Jetpack ( `Jetpack Stats needs a user connection to WordPress.com to be able to work.` ). Once I try reconnecting I get the `This WordPress.com account is already connected to another user on this site. Please login to another WordPress.com account to complete the connection.` in Calypso.

Not sure if any of this has more to do with the way I'm testing and breaking the Primary connection. I will try testing more on a clean JN site, just wondering if you can reproduce any of this.

sergeymitr:
Hi @bindlegirl, thank you for such a thorough testing!

1. I adjusted "My Jetpack" to hide the "Also connected" like if connection owner is invalid. However, if the user it still connected, most of the functionality will still work, so we don't really need to hide that. Besides, if we make them connect again, connection flow will fail because they're already connected.
2. I also made "My Jetpack" hide the `Jetpack Stats needs a user connection...` if the user is connected, even if connection owner is invalid. The button we ask users to click will lead them to the broken connection flow as they're already connected.
3. Really good catch about disappearing error messages! Turns out all errors get removed upon successful `/jetpack/v4/site` request, so I had to adjust that to only remove `xmlrpc` and `rest' errors, keeping `connection` ones intact.

Please take another look when you get a few minutes.

bindlegirl:
Ah, thanks for reproducing these. Great find with the `jetpack/v4/site` requests!

I rechecked it and noticed some small stuff.

1. Stats page shows no error. Is that expected?
2. Search page looks like this:
Screenshot 2023-06-30 at 13 37 02

3. After clicking the Restore connection and then closing the User connection (ie. not clicking the connect button) I see this in My Jetpack:

Screenshot 2023-06-30 at 13 42 32

The double message does go away on refresh or moving to a different page, so this is really a minor thing. Just a bit dramatic 😉 

4. Stats card button (Fix connection) on My Jetpack will load the wrong connection flow and lead to that already connected error.

sergeymitr:
> 1. Stats page shows no error. Is that expected?
> 2. Search page looks like this:

It shouldn't be that way 🙂 
But those are not caused to this PR, it behaves that way for other connection errors as well.
We'll need to look into it separately.

> 3. After clicking the Restore connection and then closing the User connection (ie. not clicking the connect button) I see this in My Jetpack:

I wasn't able to reproduce this. When you click "Restore":
1. It tries [to find an invalid token](https://github.com/Automattic/jetpack/blob/d236fec553d2abbc503d61d891a78f79f55ba393/projects/packages/connection/src/class-manager.php#L1733), but they are all fine.
2. Since it couldn't repair the connection, it [disconnects the site](https://github.com/Automattic/jetpack/blob/d236fec553d2abbc503d61d891a78f79f55ba393/projects/packages/connection/src/class-manager.php#L1716C40-L1716C40), [triggering error cleanup](https://github.com/Automattic/jetpack/blob/853c03dbb38dd97be5db5617537adcfceb3c4b7d/projects/packages/connection/src/class-error-handler.php#L144).
4. It [registers the site](https://github.com/Automattic/jetpack/blob/d236fec553d2abbc503d61d891a78f79f55ba393/projects/packages/connection/src/class-manager.php#L1718) again, also [triggering the error cleanup](https://github.com/Automattic/jetpack/blob/853c03dbb38dd97be5db5617537adcfceb3c4b7d/projects/packages/connection/src/class-error-handler.php#L142).
5. Site connection is established, so users get shown the user connection screen. If they don't proceed, the site stays in site-only connection mode.

So I'm not sure why the error showed up after clicking "Restore", it should've been removed.
Connection owner also [gets removed](https://github.com/Automattic/jetpack/blob/d236fec553d2abbc503d61d891a78f79f55ba393/projects/packages/connection/src/class-manager.php#L1634), so the error shouldn't get triggered again.

Can you please try to reproduce that again and check the "Broken Token" and "Connection Errors" tools to see what's going on there?
If that's a JN site, could you share the credentials so I could try to debug it?

> 4. Stats card button (Fix connection) on My Jetpack will load the wrong connection flow and lead to that already connected error.

There are a few ways to approach this, and we'll need to consult with My Jetpack devs and possibly Design too.
This issue wasn't introduced in this PR and has been there for quite a while, so let's deal with it separately.

bindlegirl (dismissed):
Everything looks good to me now. And, I tried breaking it 😆 Thanks for addressing these issues @sergeymitr

sergeymitr:
Hi @bindlegirl.
I had to merge the trunk so the review got dismissed. Can you please re-approve it when you get a minute?

bindlegirl (dismissed):
Done

bindlegirl (approved):

PR #31622, opened 2023-06-28
tools: Add checks for filename collisions on case-insensitive systems

anomiex:
## Proposed changes:

In the pre-push hook, we check the commit being pushed for collisions, as checking out the commit for review will confuse git on Mac OS and other case-insensitive systems.

In the mirror build, we check the output as collisions will likely cause problems if the package/plugin/whatever is used on a case-insensitive system.

### Other information:

## Jetpack product discussion

https://github.com/Automattic/jetpack/pull/31551#issuecomment-1611425157

no

## Testing instructions:

* Make a commit that adds some files that are the same as other files except for case (e.g. add a `readme.md` alongside `README.md`, or a `Src` dir (with at least one file inside) alongside `src`. Try to push, it should be rejected with an appropriate message.
* Make a commit adding a file or directory that will conflict with a built file, e.g. `projects/plugins/jetpack/modules/Module-Headings.php`. Push and create a PR, the build should fail with an appropriate error somewhere in the log.
  * [x] https://github.com/Automattic/jetpack/actions/runs/5405081542/jobs/9820136470?pr=31625#step:9:5181

### DIFF
--- a/tools/cli/commands/build.js
+++ b/tools/cli/commands/build.js
@@ -525,6 +525,40 @@ async function copyFileAtomic( src, dest ) {
 }
 
 /**
+ * Check for filename collisions.
+ *
+ * @param {string} basedir - Base directory.
+ * @returns {string[]} Colliding file names.
+ */
+async function checkCollisions( basedir ) {
+	// @todo Once we require Node 20.1+, use the new `recursive` option to `fs.readdir` instead of manually recursing here.
+	// Doing `const files = await fs.readdir( basedir, { recursive: true } );` should suffice.
+	const files = [];
+	const ls = async dir => {
+		for ( const file of await fs.readdir( dir, { withFileTypes: true } ) ) {
+			const path = npath.join( dir, file.name );
+			files.push( npath.relative( basedir, path ) );
+			if ( file.isDirectory() ) {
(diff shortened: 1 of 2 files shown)


tools/cli/commands/build.js
```diff
@@ -524,6 +524,40 @@ async function copyFileAtomic( src, dest ) {
...
+async function checkCollisions( basedir ) {
+	// @todo Once we require Node 20.1+, use the new `recursive` option to `fs.readdir` instead of manually recursing here.
+	// Doing `const files = await fs.readdir( basedir, { recursive: true } );` should suffice.
+	const files = [];
+	const ls = async dir => {
+		for ( const file of await fs.readdir( dir, { withFileTypes: true } ) ) {
+			const path = npath.join( dir, file.name );
+			files.push( npath.relative( basedir, path ) );
+			if ( file.isDirectory() ) {
+				await ls( path );
```
zinigor, line 542: ~What if there are other folders in that path too? How would we recurse into them?~
Disregard please, I thought for a second that `ls` here was a system call, instead it's the call to the same `ls` that we're defining here, all good.

zinigor (approved):
Works nicely!

PR #31627, opened 2023-06-28
JetpackFooter: add About, Privacy and Terms links

### LINKED ISSUE #31598: My Jetpack: Apply new footer
ref p1687874518864269-slack-C02TQF5VAJD

Follow-up to https://github.com/Automattic/jetpack/pull/30279

During an accessibility review it was noticed that My Jetpack does not use the new footer with the About, Privacy and Terms links.
Since My Jetpack can be available without the Jetpack plugin installed, it is necessary to check if the plugin is available and only show those first links if not.

monsieur-z:
Fixes #31598

## Proposed changes:

The `JetpackFooter` component is used in the Jetpack flagship plugin and many other Jetpack plugins. In the former, it contains links to about, TOS, and privacy pages that are not included in the latter, though they can provide value in those instances as well. This PR integrates the _About_, _Terms_, and _Privacy_ links into the `JetpackFooter` component directly.

It also removes from the props the URL to which the Automattic logo links to since all instances using `JetpackFooter` would pass the same URL.

In more detail:
- it makes the _About_, _Terms_, and _Privacy_ links part of `JetpackFooter`
- it passes click handlers to these links as props  of `JetpackFooter`
- it removes the prop `a8cLogoHref` from `JetpackFooter`
- when applicable, it links the _Privacy_ and `An Automattic Airline` links to internal pages
- makes the Jetpack logo on the _My Jetpack_ footer match the size of the same logo in other sections

### Other information:

## Jetpack product discussion

p1687876101340969-slack-C01264051NE
p1687874518864269-slack-C02TQF5VAJD

No.

## Testing instructions:

- Make sure your site is not connected to Jetpack
- Visit the _Jetpack_ section
- In the footer, check the following links:
  - Jetpack: `https://jetpack.com/`
  - About: `https://jetpack.com/about/`
  - Privacy: `https://automattic.com/privacy/`
  - Terms: `https://wordpress.com/tos/`
  - An Automattic Airline: `https://automattic.com/about/`
- Visit the _My Jetpack_ section
  - In the footer, check the following links:
  - Jetpack: `https://jetpack.com/`
  - About: `https://jetpack.com/about/`
  - Privacy: `https://automattic.com/privacy/`
  - Terms: `https://wordpress.com/tos/`
  - An Automattic Airline: `https://automattic.com/about/`
- Connect your site to Jetpack
- In the footer, check the following links:
  - Jetpack: `https://jetpack.com/`
  - About: `https://jetpack.com/about/`
  - Privacy: `/wp-admin/admin.php?page=jetpack#/privacy`
  - Terms: `https://wordpress.com/tos/`
  - An Automattic Airline: `/wp-admin/admin.php?page=jetpack_about`


projects/js-packages/components/components/admin-page/index.tsx
```diff
@@ -43,11 +42,7 @@ const AdminPage: React.FC< AdminPageProps > = ( {
 			{ showFooter && (
 				<Container horizontalSpacing={ 5 }>
 					<Col>
-						<JetpackFooter
-							moduleName={ moduleName }
-							a8cLogoHref={ a8cLogoHref }
-							moduleNameHref={ moduleNameHref }
-						/>
+						<JetpackFooter moduleName={ moduleName } moduleNameHref={ moduleNameHref } />
```
monsieur-z, line 45: `href` of the Automattic logo is set in `JetpackFooter` directly.

projects/js-packages/components/components/jetpack-footer/index.tsx
```diff
@@ -1,12 +1,15 @@
-import { __ } from '@wordpress/i18n';
+import { useSelect } from '@wordpress/data';
+import { __, _x } from '@wordpress/i18n';
 import { Icon, external } from '@wordpress/icons';
 import classnames from 'classnames';
 import React from 'react';
+import { getRedirectUrl } from '../..';
+import { STORE_ID as CONNECTION_STORE_ID } from '../../../../js-packages/connection/state/store';
```
monsieur-z, line 7: Importing from `@automattic/jetpack-connection` directly made the build fail because of some circular dependency. This import seems fine, but let me know if there's a better approach. Hardcoding the store id seemed error-prone.

projects/js-packages/components/components/jetpack-footer/index.tsx
```diff
@@ -19,17 +22,72 @@ const JetpackIcon: React.FC = () => (
...
+			return {
+				connectedPlugins: connectionStatus?.getConnectedPlugins(),
+				...connectionStatus.getConnectionStatus(),
+			};
+		},
+		[ CONNECTION_STORE_ID ]
+	);
+
+	const areAdminLinksEnabled =
+		siteAdminUrl &&
```
monsieur-z, line 55: Ideally, `siteAdminUrl` would be retrieved from an initial state, as is `connectedPlugins`. I haven't found a way of retrieving basic site information that works across packages yet.

andrii-lysenko (dismissed):
Looks great and works as described 👍

projects/js-packages/components/components/jetpack-footer/index.tsx
jeherve: I wonder if this may prove problematic in the version of [the package that we ship to npm](https://www.npmjs.com/package/@automattic/jetpack-components).

Couldn't we bring in that data as a new component prop, to keep the component a bit simpler?

> Hardcoding the store id seemed error-prone.

That'd be my next suggestion, with a comment in the original file mentioning that it's copied there, but I'm not a regular component maintainer so I'll let others weigh in on this. :)

jeherve (commented):
There seems to be a bug? When switching to this branch on a connected site that only runs Jetpack and Jetpack Beta, the About link is now external in the Jetpack React dashboard.

andrii-lysenko:
@jeherve could you please clarify? 
> When switching to this branch on a connected site that only runs Jetpack and Jetpack Beta, the About link is now external in the Jetpack React dashboard.

You mean on `?page=jetpack#/dashboard`? But instead we should be redirected to `wp-admin/admin.php?page=jetpack_about` right?

jeherve:
> You mean on `?page=jetpack#/dashboard`? But instead we should be redirected to `wp-admin/admin.php?page=jetpack_about` right?

Yes, exactly!

andrii-lysenko:
@jeherve the `about` link should be fixed now. As per `STORE_ID` import I don't see a problem there, and passing it as a parameter also might be not ideal as one of the renders of `JetpackFooter` is from `AdminPage` which is also in same package. We would then need to pass it through additional component. However, I'm also not an expert in components and may be there's some other ways.

bindlegirl:
The logged-in About link on the Jetpack dashboard (and other) pages shows up as an external link, although it's not. However, it loads the About page in a new tab, which differs from the Terms and Airline pages.

There are inconsistencies between the MyJetpack footer and other pages. This is what I see:

1. Logged out About link on the Jetpack page will load https://jetpack.com/about/?site=nutritious-fairy.jurassic.ninja and MyJetpack https://jetpack.com/about
2. Logged in About link on the Jetpack page will load /wp-admin/admin.php?page=jetpack_about and MyJetpack will load https://jetpack.com/about/

Privacy on connected MyJetpack is still an external link.

I'm unsure if aligning these should be a part of another PR.

kraftbj:
> I'm unsure if aligning these should be a part of another PR.

Yeah, let's land this and open an issue to track the variations.

andrii-lysenko:
@bindlegirl I think the idea of shipping as is, but, creating a follow up issue sounds as a good idea to me.
@jeherve do you want to take a second look?

bindlegirl:
Sounds good to me 👍
Other than noted details, the PR looks good

bindlegirl (dismissed):

jeherve (approved):

projects/js-packages/components/components/jetpack-footer/index.tsx
manzoorwanijk: We now have `getAdminUrl` and other such utils available in `@automattic/jetpack-script-data`

```ts
import { getAdminUrl, getJetpackAdminPageUrl } from '@automattic/jetpack-script-data';

const aboutPageUrl = getAdminUrl( 'admin.php?page=jetpack_about' );

const privacyPageUrl = getJetpackAdminPageUrl( '#/privacy' );

```

projects/js-packages/components/components/jetpack-footer/index.tsx
manzoorwanijk: Now that the connection package js is loaded externally and not bundled into any plugin after #38877, this import here can be problematic and result in duplicate store registration.

We should fix this in a better way. A few options we have:

- Let the consumer component pass the needed connection related information
- Make the component composable, which means the consumer component passes in the final output without this component here knowing anything about the connection data

projects/js-packages/components/components/jetpack-footer/index.tsx
manzoorwanijk: Likewise, you can use the same package to get the list of connected plugins and the status.

```ts
import { getScriptData } from '@automattic/jetpack-script-data';

const { connectedPlugins, connectionStatus } = getScriptData().connection;

```

projects/js-packages/components/components/jetpack-footer/index.tsx
manzoorwanijk: Another solution would be to use the data from the scrip-data package as explained in [this thread](https://github.com/Automattic/jetpack/pull/31627#discussion_r1251026389).

manzoorwanijk: That said, composability is still better because that will not make the component dependent upon anything related to jetpack connection or its logic.

PR #31630, opened 2023-06-29
Changelogger: Allow comments in non-interactive mode

psealock:
## Proposed changes:

When using Jetpack Changelogger in non-interactive mode, I can create a changelog entry without issue. When I provide a comment (using significance `patch`) instead of an entry, the following error occurs:

```
Entry must be specified in non-interactive mode.
```
Here is the raw command using Jetpack Changelogger v3.3.0

```
composer exec -- changelogger "add" "-f" update-add_gitignore_to_package" "-s" "patch" "-t" "added" "-c" "Minimal change that was missed during bootstrap, no changelog necessary." "-n"
```

### Other information:

## Testing instructions:

1. Navigate to any project, for example:
```
cd projects/packages/changelogger
```
2. Execute a non interactive changefile addition with a comment
```
composer exec -- changelogger add -f file-name -s patch -t added -c "this is a comment" -n
```
3. The command should succeed. Note it fails when on `trunk`


anomiex:
`jetpack changelog` is a nodejs wrapper around changelogger that works a little more conveniently in the context of the monorepo. You can test your change by running the raw command you quoted in the directory of a project.

I note that the raw command you quoted works correctly if you add `-e ''` to explicitly specify that you intend an empty entry.

sergeymitr:
Looks like unit tests are failing now, and the package needs a version bump. Can you please take a look?
Thanks!

psealock:
Thanks for the help @anomiex, I have updated the test instructions.

>I note that the raw command you quoted works correctly if you add -e '' to explicitly specify that you intend an empty entry.

I can confirm adding `-e ''` works in the Jetpack repository, [but `pnpm` is swallowing empty strings](https://github.com/pnpm/pnpm/issues/6594) in WooCommerce, ie `-e ''` gets passed to composer as `"-e"` only. In the meantime, could we proceed with the changes proposed here?

anomiex:
> I can confirm adding `-e ''` works in the Jetpack repository, [but `pnpm` is swallowing empty strings](https://github.com/pnpm/pnpm/issues/6594) in WooCommerce, ie `-e ''` gets passed to composer as `"-e"` only.

That's unfortunate. The long option name with `=` (like `--entry=''`) can be used to work around that.

> In the meantime, could we proceed with the changes proposed here?

If you want to proceed, you'll need to update the PR to fix the CI errors.

psealock:
>If you want to proceed, you'll need to update the PR to fix the CI errors.

Thanks @anomiex, I've addressed the failing test. `E2E Tests / Jetpack blocks e2e tests` is still failing, but I don't think its related to these changes.

projects/packages/changelogger/src/AddCommand.php
```diff
@@ -285,7 +286,7 @@ function ( $v ) {
 					return 1;
 				}
 			} else {
-				if ( null === $entry ) {
+				if ( null === $entry && $comment === '' ) {
 					$output->writeln( '<error>Entry must be specified in non-interactive mode.</>' );
```
anomiex, line 289: This error message is kind of misleading now. When significance is patch and a comment is specified, the entry doesn't have to be specified.

anomiex (commented):
Personally the more I think about it the more I'm not really liking this PR. It's not a bug that the parameter is always required even when the empty string is an allowed value. I think I'd rather clear up your confusion by adding a line to the error message:
```php
$output->writeln( '<error>Entry must be specified in non-interactive mode.</>' );
if ( 'patch' === $significance ) {
	$output->writeln( '<info>If you want to have an empty entry for this change, pass the empty string as the entry (like <comment>--entry=</>) and also please provide a comment (using <comment>--comment</>).</>' );
}
```

If you still want to go ahead with this, please add a test (in `tests/php/tests/src/AddCommandTest.php` function `provideExecute`) for the new behavior (omitting `--entry` while providing `--comment` not resulting in an error). For that matter, if you take my suggestion then we should still add a test for that extra line being present when appropriate (and, ideally, not present when not).

projects/packages/changelogger/src/AddCommand.php
```diff
@@ -287,6 +287,9 @@ function ( $v ) {
 			} else {
 				if ( null === $entry ) {
 					$output->writeln( '<error>Entry must be specified in non-interactive mode.</>' );
+					if ( null === $entry && 'patch' === $significance ) {
```
anomiex: We already checked `null === $entry` on line 288 which wraps this, no need to repeat it here.
```suggestion
					if ( 'patch' === $significance ) {
```

projects/packages/changelogger/tests/php/tests/src/AddCommandTest.php
```diff
@@ -399,6 +399,19 @@ public function provideExecute() {
...
+				array(
+					'--significance' => 'patch',
+					'--type'         => 'fixed',
+				),
+				array( 'interactive' => false ),
+				array(),
+				1,
+				null,
+				array(
+					'/Entry must be specified in non-interactive mode./\nIf you want to have an empty entry for this change, pass the empty string as the entry (like --entry=) and also please provide a comment (using --comment)./',
```
anomiex: Either like this
```suggestion
					'/Entry must be specified in non-interactive mode\.\nIf you want to have an empty entry for this change, pass the empty string as the entry \(like --entry=\) and also please provide a comment \(using --comment\)\./',
```
or like this
```suggestion
					'/Entry must be specified in non-interactive mode\./',
					'/If you want to have an empty entry for this change, pass the empty string as the entry \(like --entry=\) and also please provide a comment \(using --comment\)\./',
```

projects/packages/changelogger/changelog/fix-changelogger-comment-non-interactive
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fixed providing a comment in non-interactive mode. Previously, this would cause an error
```
anomiex: This needs updating too. Maybe like this?
```suggestion
Significance: patch
Type: added

When omitting `--entry` for `changelogger add` in non-interactive mode for a patch-significance change, provide a hint on how to successfully create an empty entry.

psealock:
Thanks for the suggestions @anomiex, I had meant to get around to polishing those items up. This is now ready for review, cheers

anomiex (approved):
LGTM! :ship:

PR #31631, opened 2023-06-29
[Boost] Change _static dir for minify

thingalon:
Fixes jpop-issue 8371

Users on the Atomic platform who have uninstalled Page Optimize get 404 errors when loading minified CSS and JS via Boost's Minify JS and CSS features. This occurs because they both use the same base path (`/_static/`) for minified content, and Atomic has inbuilt rules for that path.

This PR fixes this in two ways:
1. It changes Jetpack's minify path base to `/_jb_static/` to remove that collision, and
2. It makes the path configurable by defining a constant `JETPACK_BOOST_STATIC_PREFIX`.

## Proposed changes:
* Changes Jetpack's minify path base to `/_jb_static/` to remove that collision, and
* Make the path configurable by defining a constant `JETPACK_BOOST_STATIC_PREFIX`.

### Other information:

## Jetpack product discussion

n/a

no

## Testing instructions:
* Turn on CSS and JS minify
* Ensure that assets correctly load from `/_jb_static/`.
* Define an alternate base (include the slashes!), like so: `define( 'JETPACK_BOOST_STATIC_PREFIX', '/.my-path/' );`
* Check that your minified content is now served correctly from the new path.

### DIFF
--- a/projects/plugins/boost/app/lib/minify/functions-helpers.php
+++ b/projects/plugins/boost/app/lib/minify/functions-helpers.php
@@ -260,7 +260,7 @@ function jetpack_boost_page_optimize_cache_bust_mtime( $path, $siteurl ) {
 }
 
 /**
- * Detects requests within the `/_static/` directory, and serves minified content.
+ * Detects requests within the `/_jb_static/` directory, and serves minified content.
  *
  * @return void
  */
@@ -270,7 +270,7 @@ function jetpack_boost_minify_serve_concatenated() {
 	if ( isset( $_SERVER['REQUEST_URI'] ) ) {
 		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
 		$request_path = explode( '?', wp_unslash( $_SERVER['REQUEST_URI'] ) )[0];
-		if ( '/_static/' === substr( $request_path, -9, 9 ) ) {
+		if ( '/_jb_static/' === substr( $request_path, -9, 9 ) ) {
 			require_once __DIR__ . '/functions-service.php';
 			jetpack_boost_page_optimize_service_request();
 			exit;
--- a/projects/plugins/boost/app/lib/minify/functions-helpers.php
+++ b/projects/plugins/boost/app/lib/minify/functions-helpers.php
@@ -260,6 +260,14 @@ function jetpack_boost_page_optimize_cache_bust_mtime( $path, $siteurl ) {
 }
 
 /**
+ * Get the URL prefix for static minify/concat resources. Defaults to /jb_static/, but can be
+ * overridden by defining JETPACK_BOOST_STATIC_PREFIX. Please include leading and trailing slashes.
+ */
+function jetpack_boost_get_static_prefix() {
+	return defined( 'JETPACK_BOOST_STATIC_PREFIX' ) ? JETPACK_BOOST_STATIC_PREFIX : '/_jb_static/';
+}
+
+/**
  * Detects requests within the `/_jb_static/` directory, and serves minified content.
  *
  * @return void
@@ -270,7 +278,8 @@ function jetpack_boost_minify_serve_concatenated() {
 	if ( isset( $_SERVER['REQUEST_URI'] ) ) {
 		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
 		$request_path = explode( '?', wp_unslash( $_SERVER['REQUEST_URI'] ) )[0];
-		if ( '/_jb_static/' === substr( $request_path, -9, 9 ) ) {
+		$prefix       = jetpack_boost_get_static_prefix();
+		if ( $prefix === substr( $request_path, -strlen( $prefix ), strlen( $prefix ) ) ) {
 			require_once __DIR__ . '/functions-service.php';
 			jetpack_boost_page_optimize_service_request();
 			exit;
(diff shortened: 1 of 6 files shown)


projects/plugins/boost/app/lib/minify/functions-service.php
```diff
@@ -199,7 +199,7 @@ function jetpack_boost_page_optimize_build_output() {
 	// We can't assume that the root serves the same content as the subdir.
 	$subdir_path_prefix = '';
 	$request_path       = $utils->parse_url( $request_uri, PHP_URL_PATH );
-	$_static_index      = strpos( $request_path, '/_static/' );
+	$_static_index      = strpos( $request_path, '/_jb_static/' );
```
brandonpayton: Should `jetpack_boost_get_static_prefix()` be used here rather than the string literal?

projects/plugins/boost/jetpack-boost.php
```diff
@@ -68,7 +68,9 @@
 if ( isset( $_SERVER['REQUEST_URI'] ) ) {
 	// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
 	$request_path = explode( '?', wp_unslash( $_SERVER['REQUEST_URI'] ) )[0];
-	if ( '/_static/' === substr( $request_path, -9, 9 ) ) {
+
+	$static_prefix = defined( 'JETPACK_BOOST_STATIC_PREFIX' ) ? JETPACK_BOOST_STATIC_PREFIX : '/_jb_static/';
```
brandonpayton, line 73: Is there a reason to avoid using `jetpack_boost_get_static_prefix()` here?

brandonpayton (commented):
Earlier we discussed using a filter to change the path prefix, and I'm glad you switched to using a constant instead. A constant seems more sensible in this case because we want to be able to set at a platform level and have little interest (I think) in allowing third party plugins to change the prefix. It's also likely faster to access the constant.

I left a couple of questions, but overall, this seems good to me.

projects/plugins/boost/app/lib/minify/functions-helpers.php
```diff
@@ -260,7 +260,15 @@ function jetpack_boost_page_optimize_cache_bust_mtime( $path, $siteurl ) {
 }
 
 /**
- * Detects requests within the `/_static/` directory, and serves minified content.
+ * Get the URL prefix for static minify/concat resources. Defaults to /jb_static/, but can be
+ * overridden by defining JETPACK_BOOST_STATIC_PREFIX. Please include leading and trailing slashes.
```
dilirity: Shouldn't the code make sure all the slashes are in place? I feel this will reduce human error.

projects/plugins/boost/app/lib/minify/functions-service.php
dilirity: I noticed that this isn't passed through the "filter" function, but is hard-coded. I even tested in a nested WP instance - everything worked.

Seeing it hard-coded seems wrong, but I can't find anything wrong with it.

projects/plugins/boost/jetpack-boost.php
```diff
@@ -68,7 +68,9 @@
...
+	if ( $static_prefix === substr( $request_path, -strlen( $static_prefix ) ) ) {
```
dilirity, line 74: Nice thinking changing `-9, 9` to `-strlen`!

dilirity (dismissed):
I tested the changes and everything looks good and seems to be working properly.

Left some comments, since I wasn't sure if things should be left as is, but am okay with them staying like that.

projects/plugins/boost/app/lib/minify/functions-service.php
thingalon: Thank you; good catch. Fixed in fc5405d979

projects/plugins/boost/jetpack-boost.php
thingalon: Yes. Jetpack Boost doesn't load code from modules which aren't enabled by default, so it would be unusual to load the Minify CSS/JS modules this early in the plugin life-cycle.

However, this particular bit of code wants to make a decision early about letting the minify service handle a given request.

So it made sense in this one case to handle it in-line before loading anything extra.

I added a comment in 809d477186 to explain this oddity, as future devs might not realize why.

projects/plugins/boost/app/lib/minify/functions-helpers.php
thingalon: You're right. I was avoiding it because in my head, we didn't want to add overhead for this check -- but the overhead would be negligible and I was optimizing the wrong thing.

Changed in e5aea543be.

projects/plugins/boost/app/lib/minify/functions-service.php
thingalon: Thank you; good catch. Fixed in https://github.com/Automattic/jetpack/commit/fc5405d97972255e0ae5a507cd5b1e0e461ee83e

thingalon:
@dilirity, @brandonpayton thank you for your reviews. You both picked up on things I missed, and this is better for it.

I think it's ready to go now, so please re-:eyes:. I would like to roll this into Jetpack Boost 1.9.2 beta so I can release it on Monday.

> Earlier we discussed using a filter to change the path prefix, and I'm glad you switched to using a constant instead. A constant seems more sensible in this case because we want to be able to set at a platform level and have little interest (I think) in allowing third party plugins to change the prefix. It's also likely faster to access the constant.

I realized that filters are not available when we don't load WordPress. So a constant is necessary for the service php file too.

dilirity (approved):
Looks good to me. :shipit:

thingalon:
Cherry picked to 1.9.2 in c42a661eba.

PR #31633, opened 2023-06-29
CRM: Automations module

cleacos:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3127

Some of the required base work to load the Automations module was already done in #31853, but this aims to finish the implementation with expected CRM logic and quality of life additions like e.g. conditions.

## Proposed changes:

* Load the module - on demand - by the `jpcrm_load_modules` filter instead of directly in the `jpcrm-automations-init.php` file.
* Update the namespace to be different from the "automations package" in `projects/plugins/crm/src`
* Dynamically fetch the namespace for module functions used in filters _(we would have to add `\Modules\` anyway, so this just makes it easier for future potential changes)_
* Add knowledge base and external URLs for automations
* Use `admin_zerobs_manage_options` permission to access the admin page
* We are hiding the UI for now since we don't have any "system automations", so we have to add a `jetpack_crm_automations_load_ui` filter to make sure the Automation extensions can enable it.
* Added `@since $$next-version$$` to DocBlocks comply with WordPress best practices
* Define automation module constants

### Other information:

## Jetpack product discussion

N/A

No

## Testing instructions:

1. Add the following filter somewhere in your codebase (e.g. 0-sandbox.php): `add_filter( 'jetpack_crm_feature_flag_automations', '__return_true' );`
2. Run `jetpack build plugins/crm`
3. Visit `/wp-admin/admin.php?page=jpcrm-automations` and verify that you see a very basic automation admin page.
4. Disable the feature flag filter added in step 1 and verify that the Automation page does not show up in the menu bar and that the page is not accessible.


projects/plugins/crm/admin/automations/main.page.php
```diff
@@ -12,10 +12,6 @@
 
 defined( 'ZEROBSCRM_PATH' ) || exit;
 
-if ( ! apply_filters( 'jetpack_crm_feature_flag_automations', false ) ) {
-	return;
-}
-
```
kallehauge, line 18: I'm removing this because it didn't have any value/impact besides running the filter one more time, unnecessarily.

_It's not needed since the logic that registers the page isn't added unless the feature flag is enabled._

gogdzl (approved):
Simple code that gets the job done.
LGTM!

PR #31634, opened 2023-06-29
Custom CSS: adjust wording in dashboard card

jeherve:
## Proposed changes:

Following up from #31413, based on feedback in https://github.com/Automattic/jetpack/pull/31413#issuecomment-1612567000

1. Update the wording of the notice recommending global styles.
2. Conditionally display the custom css support info (text and link). If a site uses a block theme, we don't want to say too much about the module to avoid confusing people. Let's only display that info for folks using classic themes.

**Block theme, custom CSS module inactive**

Screenshot 2023-06-29 at 12 41 30

**Block theme, custom CSS module active**

Screenshot 2023-06-29 at 12 41 22

**Classic theme**

Screenshot 2023-06-29 at 12 40 57

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

1. Start with a site connected to WordPress.com, either self-hosted on WoA, with a classic theme like Twenty Twenty One.
4. Go to Jetpack > Settings > Writing
    * Check wording in the Custom CSS card.
    * You should see the "i" icon and info about the Custom CSS module.
5. Now go to Appearance > Themes and switch to the Twenty Twenty Three theme.
6. Go back to Jetpack > Settings > Writing.
    * Check wording again.
    * The "i" icon should be gone now.

### DIFF
--- a/projects/plugins/jetpack/_inc/client/writing/custom-css.jsx
+++ b/projects/plugins/jetpack/_inc/client/writing/custom-css.jsx
@@ -51,7 +51,7 @@ function CustomCss( props ) {
 				<div className="jp-custom-css-site-editor__text">
 					{ createInterpolateElement(
 						__(
-							'Hurray! Your theme supports site editing with blocks. <a>Tell me more.</a>',
+							'Your site has a block theme that allows you to apply custom CSS from the Site Editor. <a>Learn more.</a>',
 							'jetpack'
 						),
 						{
--- a/projects/plugins/jetpack/_inc/client/writing/custom-css.jsx
+++ b/projects/plugins/jetpack/_inc/client/writing/custom-css.jsx
@@ -57,7 +57,7 @@ function CustomCss( props ) {
 						{
 							a: (
 								<ExternalLink
-									href="https://wordpress.org/documentation/article/site-editor/"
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/_inc/client/writing/custom-css.jsx
```diff
@@ -51,7 +51,7 @@ function CustomCss( props ) {
 				<div className="jp-custom-css-site-editor__text">
 					{ createInterpolateElement(
 						__(
-							'Hurray! Your theme supports site editing with blocks. <a>Tell me more.</a>',
+							'Your site has a block theme that allows you to apply custom CSS from the Site Editor. <a>Learn more.</a>',
```
mmtr, line 54: What do you think about linking to https://wordpress.org/documentation/article/styles-overview/#applying-custom-css instead?

mmtr (commented):
Thanks! It works well for me. Just left a minor comment, but this is pretty much ready to go.

projects/plugins/jetpack/_inc/client/writing/custom-css.jsx
jeherve: I like the idea! Done in 718870107629423b9f452036f91d569cfe31fcaa

mmtr (approved):

PR #31635, opened 2023-06-29
Subscriptions: do not display block when user isn't connected

### LINKED ISSUE #26178: Subscriptions: Investigate potential fatal
### Impacted plugin

Jetpack

### Steps to Reproduce

* Unverified, but setup a local testing site with a tunnel. While tunneling with the public domain, set up the site inclusive of the subscriptions module.
* Then, access the site via https://localhost
* Open the post editor and start to write a post.

No fatals.

### What actually happened

```
[12-Sep-2022 18:55:45 UTC] PHP Fatal error:  Uncaught Error: Class "Jetpack_Subscriptions_Widget" not found in /usr/local/src/jetpack-monorepo/projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/subscribers.php:67
```

### Browser

_No response_

### Other information

My guess is something with Developer mode disabling the subscriptions module, but either a block always expects it or doesn't realize it won't ba available, and the rest endpoint class doesn't sufficiently ensure the class is available before trying to use it.

### Platform (Simple, Atomic, or both?)

_No response_

### Reproducibility

Consistent

### Severity

One

### Available workarounds?

_No response_

### Workaround details

_No response_

jeherve:
Fixes #26178

## Proposed changes:

Jetpack's Subscribe block relies on the Subscriptions module, which requires a user connection. When only the site is connected, but not the user, we should not attempt to register / display the block.

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

* Start with a site that's fully connected to WordPress.com.
* Go to Jetpack > settings and enable Subscriptions.
* Go to Appearance > Themes and swicth to the Twenty Twenty One theme
* Go to Appearance > Widgets and add a Subscribe block to a widget area.
* Go to Jetpack > Dashboard, and scroll down the connections area.
* Disconnect your site from WordPress.com.
* Start reconnecting your site, but do not complete the connection, do not connect your user to WordPress.com.
* Try viewing a page where the widget can be found.
    * The widget should not be displayed, and there should be no errors in your logs.

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-fatal-subscriptions
+++ b/projects/plugins/jetpack/changelog/fix-fatal-subscriptions
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
+Subscriptions: avoid Fatal Error when site is connected to WordPress.com, but user is not.
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -8,6 +8,7 @@
 namespace Automattic\Jetpack\Extensions\Subscriptions;
 
 use Automattic\Jetpack\Blocks;
+use Automattic\Jetpack\Connection\Manager as Connection_Manager;
 use Automattic\Jetpack\Extensions\Premium_Content\Subscription_Service\Token_Subscription_Service;
 use Automattic\Jetpack\Status;
 use Jetpack;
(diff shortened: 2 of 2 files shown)


sergeymitr (approved):
Tested a few scenarios of connection being broken in strange ways, works great!

samiff:
cherry-picked to `monthly/branch-2023-06-26` in 68adef52b62ec8a9726ce7b4c589fac3bfa80bdb

PR #31636, opened 2023-06-29
Sharing: deprecate Skype sharing button

### LINKED ISSUE #31624: Sharing Buttons: Skype sharing button not working due to expired SSL cert
**To Reproduce:** 

- On a Jetpack connected site, enable sharing buttons in `Jetpack > Settings > Sharing` then click on the "Configure your sharing buttons" link.
- Add the Skype sharing button.
- On the frontend, try sharing a post with the Skype sharing button.

**Result:**

- The popup loads but several resources on the Skype side of things are failing due to an expired SSL cert such as `https://sdk.cdn.skype.com/0-1-12/js/share.js`

Reported via internal testing: p8oabR-1eP-p2#comment-7390

Skype sharing button originally requested in #3170

jeherve:
Fixes #31624

## Proposed changes:

The button is no longer supported by Skype, we can consequently deprecate it and allow site owners to remove it from their site.

This PR also makes a small change to the message we display for deprecated buttons. Instead of saying "service has shut down", we're a bit more specific, since some services (like Skype) may still be available while their sharing button is no longer supported.

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

* Start with a site running `trunk` or the latest stable version of Jetpack.
* Go to Jetpack > Settings > Social and enable sharing buttons.
* Go to Settings > Sharing and add a Skype sharing button to your site.
* Check your site's frontend; the button should appear.
* Now switch to this branch.
* Check your site's frontend again
    * the button will appear with a message when you're logged in
    * the button will not appear when you're logged out.


projects/plugins/jetpack/modules/sharedaddy/sharing-sources.php
```diff
@@ -521,13 +521,13 @@ public function display_preview( $echo = true, $force_smart = false, $button_sty
 			esc_attr(
 				$is_deprecated
 					/* translators: %1$s is the name of a deprecated Sharing Service like "Google+" */
-					? sprintf( __( 'The %1$s service has shut down. This sharing button is not displayed to your visitors and should be removed.', 'jetpack' ), $this->get_name() )
+					? sprintf( __( 'The %1$s sharing service is no longer supported. This sharing button is not displayed to your visitors and should be removed.', 'jetpack' ), $this->get_name() )
```
anomiex: I wonder whether "is no longer supported" is the best wording. People seeing this might think we dropped the support versus the service shut down or dropped sharing support.

Maybe we could have a method on the deprecated service to return a reason, defaulting to the old "The %1$s service has shut down." but allowing the Skype class to return something like "Skype has discontinued sharing buttons" instead?

anomiex (commented):
Works as advertised, but I'm a little concerned over the new wording.

projects/plugins/jetpack/modules/sharedaddy/sharing-sources.php
jeherve: > I wonder whether "is no longer supported" is the best wording.

Yeah, I can see how this could be a bit confusing.

I'm a bit on the fence with adding a new method just to handle different wording though. Since we do not use that class that often, maybe we could come up with a more neutral / general wording that would work for everyone?

anomiex: Who can we ask for help with that wording? I don't have any ideas beyond something kind of vague like "The %1$s service has shut down or discontinued support for sharing buttons".

sdixon194: I think "is no longer supported" would be fine - while I see the confusion that some users, may have, at the end of the day it's still just no longer supported regardless of which end is responsible.

But to expand on Brad's idea, maybe simply "The %1$s sharing service is no longer supported or has been discontinued by the provider." would do it?

jeherve: Either would work for me. For the title attribute we can afford to be long. It's probably trickier for the actual display, which has to be a lot shorter (currently `%1$s is no longer supported`):

image

What if we just said "is deprecated"?

anomiex: I don't think "is deprecated" is any better than "is no longer supported". Maybe worse if fewer blog admins even know what "deprecated" means, how common is the word outside of software development?

OTOH, if the tooltip says "The %1$s service has shut down or discontinued support for sharing buttons" then the button itself saying "is no longer supported" or the like seems less likely to cause lasting confusion.

jeherve: 👍 I pushed c140cd2b53555fc4bcb3e8e47d2d1d247ff8fc47, so "is no longer supported" remains.

projects/plugins/jetpack/modules/sharedaddy/sharing.php
```diff
@@ -293,7 +293,7 @@ public function output_service( $id, $service, $show_dropdown = false ) { // php
 
 		if ( $service->is_deprecated() ) {
 			/* translators: %1$s is the name of a deprecated Sharing Service like "Google+" */
-			$title     = sprintf( __( 'The %1$s service has shut down. This sharing button is not displayed to your visitors and should be removed.', 'jetpack' ), $service->get_name() );
+			$title     = sprintf( __( 'The %1$s sharing service is no longer supported. This sharing button is not displayed to your visitors and should be removed.', 'jetpack' ), $service->get_name() );
```
anomiex:
```suggestion
			$title     = sprintf( __( 'The %1$s sharing service has shut down or discontinued support for sharing buttons. This sharing button is not displayed to your visitors and should be removed.', 'jetpack' ), $service->get_name() );
```

anomiex (approved):

PR #31641, opened 2023-06-29
Blaze: Fixes a 404 error in the dashboard select posts page

sbarbosa:
## Proposed changes:

This fixes an issue in the Blaze dashboard - Select post to promote page, related to a 404 in the endpoint `blaze-app/sites/{SITE_ID}/wordads/dsp/api/v1/wpcom/sites/{SITE_ID}/blaze/posts`.

I am adding the missing controller inside the Blaze package, that will be in charge of redirecting that call to the WPCOM API.

### Other information:

## Jetpack product discussion

N/A

No

## Testing instructions:

On a self-hosted site, enable the Blaze Dashboard with `add_filter( 'jetpack_blaze_dashboard_enable', '__return_true' );`

* Go to Tools->Advertising
* Click on the Promote button next to the Jetpack Blaze header
* Click the Get Started button (if the get started page is shown for your user)
* Verify that the `Select post to promote` page is shown and that your posts eligible for Blaze are displayed.

Screenshot 2023-06-29 at 13 36 24


projects/packages/blaze/src/class-dashboard-rest-controller.php
```diff
@@ -52,6 +52,17 @@ public function register_rest_routes() {
 			)
 		);
 
+		// WordAds DSP API Posts routes
+		register_rest_route(
+			static::$namespace,
+			sprintf( '/sites/%d/wordads/dsp/api/v1/wpcom/sites/%d/blaze/posts(\?.*)?', $site_id, $site_id ),
```
jeherve: Maybe we can use a numbered placeholder here to avoid having to repeat `$site_id` twice?

```suggestion
			sprintf( '/sites/%1$d/wordads/dsp/api/v1/wpcom/sites/%1$d/blaze/posts(\?.*)?', $site_id ),
```

jeherve: Is there any way we could be more restrictive of the content we allow with `(\?.*)?`? What kind of info do we expect here?

sbarbosa: That is to make the regex works for sites using plain permalinks. I am not storing that regex patch in a variable.
The actual parse of the parameters happens on the function `build_subpath_with_query_strings`, and in there, we call the `stripslashes_deep` method to clean it up.

For that specific method, the query params that we expect are for the page, title, sorting, etc.

sbarbosa: I will try that approach, it sounds better that way. 👍

sbarbosa: Maybe it's a config problem, but in the last PR I had to append `(\?.*)?` that to the end of the regex, because for sites using plain permalinks, all the request from the client that had query string weren't working.

That was because the path that the controller was matching against, had the query param on it (this doesn't happen for sites using the other configurations).

If there is a better way to handle that, let me know. I can remove that extra `(\?.*)?` and leave it like it was before.

jeherve: Let's address this in a follow-up PR. 👍

jeherve (dismissed):

sbarbosa:
@jeherve, I noticed that another of the endpoints was incorrectly configured. 

The `wordads/dsp/api/v1/woo/countries` was redirected to an incorrect WP endpoint: `wordads/dsp/api/v1/countries`. (I missed the /woo in the redirection).

I pushed a commit to fix that endpoint too. To test this, you need to click on Promote on one post, make a change in the campaign and add click **Add new card** in the Payment section. That will trigger the `/wordads/dsp/api/v1/woo/countries` request to Jetpack.

jeherve (approved):

timur987 (approved):
Thanks @sbarbosa!

The code looks great and works like a charm!

PR #31642, opened 2023-06-29
actions: Remove bitrotted issues-to-projects.yml workflow

anomiex:
## Proposed changes:

This workflow has apparently been broken since October 1, 2022, as v1 of the action that it uses broke on that date and we never upgraded to v2.

Looking deeper into it, even if we were to do that upgrade now it still wouldn't do anything useful:

* The "Assign Boost issues to Boost board" job would try to assign issues to [a GH project that no longer exists][1].
* The "Assign PRs where JARVIS is requested as reviewer to the JARVIS board" job will never run as [the team it looks for no longer exists][2].

Since it seems no one has missed this in the 9 months since it broke, let's just delete it.

[1]: https://github.com/orgs/Automattic/projects/322
[2]: https://github.com/orgs/Automattic/teams/jarvis-jetpack-quality

### Other information:

## Jetpack product discussion

None

No

## Testing instructions:

* Nothing to test


thingalon (approved):
Thank you for picking up on this and clearing out

PR #31643, opened 2023-06-29
AI Assistant: Support upgrade prompt for Simple and Atomic sites

dhasilva:
Part of https://github.com/orgs/Automattic/projects/667/views/1?pane=issue&itemId=32070782

Depends on D114991-code. Upgrade prompts will only be visible after D114991-code deployment.

## Proposed changes:

* Uses the `useUpgradeFlow` hook for non-Jetpack websites
* Removes the condition on being a Jetpack site for the `$require_upgrade` flag

### Other information:

## Testing instructions:

* On a JN site, send 20 requests for the AI Assistant
* Check that the upgrade flow works and the user is initially redirected to a price page
* On a Simple or Atomic site, check that the upgrade prompt is displayed after 20 requests and the user is redirected directly to the checkout page


projects/plugins/jetpack/extensions/blocks/ai-assistant/components/upgrade-prompt/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+import useAutosaveAndRedirect from '../../../../shared/use-autosave-and-redirect';
+import useUpgradeFlow from '../../../../shared/use-upgrade-flow';
+
+const useAIAssistantUpgradeFlow = () => {
+	// Direct checkout flow for atomic and simple sites, as they do not have My Jetpack.
+	const [ directCheckoutUrl, goToCheckoutPage, isRedirecting ] =
+		useUpgradeFlow( 'jetpack_ai_monthly' );
+
+	// My Jetpack checkout flow for Jetpack sites.
+	const myJetpackCheckoutUrl = `${ window?.Jetpack_Editor_Initial_State?.adminUrl }admin.php?page=my-jetpack#/add-jetpack-ai`;
```
retrofox: Is `myJetpackCheckoutUrl` the checkout URL?

projects/plugins/jetpack/extensions/blocks/ai-assistant/components/upgrade-prompt/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+	const { autosaveAndRedirect, isRedirecting: isRedirectingMyJetpack } =
+		useAutosaveAndRedirect( myJetpackCheckoutUrl );
```
retrofox: Is it possible it redirects to this URL even for dotcom sites?

dhasilva: Nope, the page is `my-jetpack`, it does not exist on dotcom sites. If we register another one somehow, that can be reused just like we do on Jetpack by just changing the URL

projects/plugins/jetpack/extensions/blocks/ai-assistant/components/upgrade-prompt/index.tsx
dhasilva: Technically it is the URL of the pricing page before the checkout page. It is named as such because the `Nudge` component uses a `checkoutUrl` prop for the link.

lhkowalski:
Tested on a Simple site.

When I reached the limit, I got the notices from the extension and also the upgrade prompt on the AI Assistant:

Screen Shot 2023-06-30 at 12 41 24

Screen Shot 2023-06-30 at 12 42 17

I also got the `site-require-upgrade = true` from the usage info endpoint:

Screen Shot 2023-06-30 at 12 40 23

The link on the upgrade prompt is correct for Simple sites and lead me to the checkout page with the right plan:

Screen Shot 2023-06-30 at 12 43 35

After the purchase, I got a confirmation:

Screen Shot 2023-06-30 at 12 45 19

The button lead me to the WP.COM home for the site:

Screen Shot 2023-06-30 at 12 45 41

After the upgrade, I could continue using the assistant and the extension. The usage endpoint also returns that the site does not requires an upgrade anymore:

Screen Shot 2023-06-30 at 12 47 17

---

I will test on a Jetpack site as well to make sure it's working the way it should there.

lhkowalski:
Tested on my Jetpack test site and the upgrade flow seems to work as it usually worked.

- The link:

Screen Shot 2023-06-30 at 15 28 38

- The interstitial:

Screen Shot 2023-06-30 at 15 31 40

- The block working as it should after the upgrade:

Screen Shot 2023-06-30 at 15 34 24

dhasilva:
`DO NOT MERGE` - Needs to be merged together with D114991-code once we decide how to proceed with P2 and decide on a date

lhkowalski:
I tested on Atomic sites and the upgrade button is leading to the right place: the checkout page.

Screen Shot 2023-07-12 at 16 43 00

lhkowalski (dismissed):
Code looks good and works as expected. Nice move on porting it to TS.

projects/plugins/jetpack/_inc/lib/class-jetpack-ai-helper.php
```diff
@@ -327,11 +327,7 @@ public static function get_ai_assistance_feature() {
 			 * should support site-type handling.
 			 * @todo: research how to do it.
```
retrofox: can we update these comments?

lhkowalski (approved):
Updated the comments as @retrofox requested.

PR #31645, opened 2023-06-29
Prevent race conditions during dedicated syncs & plugin update routines

WPprodigy:
## Proposed changes:

(this is a proof of concept right now, wanting to open up a discussion about it)

Relying on SELECT queries from the database for important locks does not work very well at high scale. Read queries can come from replica databases with a somewhat stale value (or during times of trouble the query could altogether fail). 

The best way to truly prevent race conditions and stampedes is to utilize `wp_cache_add()` when an external object cache is in use. This function should only return `true` if it was able to save the lock & the lock did not exist beforehand. With this you can ensure one request and one request only is the "claimer" of the lock. The other patterns of GETTING, then CHECKING, then SETTING are always going to be super prone to race conditions under heavy traffic. Whereas `wp_cache_add()` does it all in one go.

I've applied changes to both the dedicated sync and plugin update logic to utilize this safer version of locking - which is especially important since both of these bits of functionality run on all frontend requests - highly increasing the likelihood of stampedes.

## Questions:

I'm still a bit uncertain about freeing this new optional cache lock in the dedicated sync portion, as it appears the lock is released before the actual syncing even takes place: https://github.com/Automattic/jetpack/blob/dfb6431ed5af7a0956e8d4c27ba9d72728b421a0/projects/packages/sync/src/class-sender.php#L401-L404

 So it's not really gating things to only one sync at a time because of this? I'm curious for the reasoning here?

## Testing instructions:
Ensure dedicated sync and plugin update routines are still able to complete per usual.

### DIFF
--- a/projects/packages/sync/changelog/fix-race-conditions-with-db-locks
+++ b/projects/packages/sync/changelog/fix-race-conditions-with-db-locks
@@ -1,0 +1,4 @@
+Significance: patch
+Type: added
+
+Jetpack & Jetpack Sync: Added cache check when trying to spawn dedicated sync or update JETPACK__VERSION to avoid additional requests to the DB if external cache is available.
--- a/projects/packages/sync/src/class-dedicated-sender.php
+++ b/projects/packages/sync/src/class-dedicated-sender.php
@@ -220,6 +220,13 @@ class Dedicated_Sender {
 	public static function try_lock_spawn_request() {
 		$current_microtime = (string) microtime( true );
 
+		if ( wp_using_ext_object_cache() ) {
+			if ( true !== wp_cache_add( self::DEDICATED_SYNC_REQUEST_LOCK_OPTION_NAME, $current_microtime, 'jetpack', self::DEDICATED_SYNC_REQUEST_LOCK_TIMEOUT ) ) {
+				// Cache lock has been claimed already.
+				return false;
+			}
+		}
+
 		$current_lock_value = \Jetpack_Options::get_raw_option( self::DEDICATED_SYNC_REQUEST_LOCK_OPTION_NAME, null );
 
 		if ( ! empty( $current_lock_value ) ) {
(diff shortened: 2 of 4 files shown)


bisko:
@WPprodigy hey! Thanks for submitting that PR! When we were discussing possible mitigations and checks when using external caching solution, I was envisioning just what you made here.

I tested this and it's working as expected with and without the object cache, which is awesome.

I am going to send it over to @Automattic/ground-control for review on the changes in `projects/plugins/jetpack/class.jetpack.php` as I am not sure if they're needed or how they might interfere with the main workings of Jetpack the Plugin.

My initial assessment on that change is that we don't need it, unless the site is VERY high traffic and the requests to the high-traffic areas aren't cached.

But at the moment I am not sure how this works on the VIP infrastructure, so I am leaving it to @Automattic/ground-control for their thoughts.

As for the Dedicated Sync changes - I am happy with them and would gladly accept/merge this PR as-is.

To answer your question:

> So it's not really gating things to only one sync at a time because of this? I'm curious for the reasoning here?

The locks there are two-fold. First we have basic Sync locking where we lock anyway when a Sync send is underway. This is shared between normal Sync and Dedicated Sync. 

The additional locking that happens in the Dedicated Sync request that you're modifying here is there ONLY for the Dedicated Sync request to the site. With the idea to have only one request running at a time and if by chance we spawn multiple of them, only one of them continues processing, instead of all of them.

The need for a lock on the Dedicated Sync request specifically is that we can get into a race-condition on high traffic sites, where multiple requests try to "lock" and spawn the request and we can easily get over some limit of simultaneous self-requests to sites, which we have seen in the past on WordPress.com Business sites. 

If we somehow manage to spawn multiple Dedicated Sync requests, we want the "extra" ones to quit as soon as possible, so that's why we check this lock quickly and lightly, before moving on in the shared Sync code where we do much more additional checks and DB requests for them.

zinigor:
The changes to `class.jetpack.php` look good to me because they don't change any logic. We just change the method of locking from a transient to a cache entry with the same timeout.

bisko (dismissed):
LGTM!

bisko (approved):

PR #31654, opened 2023-06-30
Adds connected user information to the Blaze dashboard initial load

sbarbosa:
This PR changes the initial user information that the Blaze Dashboard uses to load the application. That information will be used by the internal state of the Dashboard, primarily for tracks and analytics.

## Proposed changes:

* Changes the information we are returning in the initial load of the page, to include the connected user info.

### Other information:

## Jetpack product discussion

N/A

This module is not open for users yet, we will need to include the Jetpack Blaze privacy information in the future when we open the gate. 

For now, the Dashboard is using the same tracks used in Wordpress.com Calypso (Advertising page).

## Testing instructions:

On a self-hosted site, enable the Blaze Dashboard with `add_filter( 'jetpack_blaze_dashboard_enable', '__return_true' );`

* Go to Tools->Advertising
* Verify that the Jetpack Blaze dashboard loads correctly

Screenshot 2023-06-30 at 11 00 35


projects/packages/blaze/src/class-dashboard-config-data.php
```diff
@@ -81,6 +81,31 @@ public function get_data() {
 		);
 	}
 
+	/**
+	 * Gets the WordPress.com user's identity, if connected.
+	 *
+	 * @return array|bool
+	 */
+	protected function get_connected_user_identity() {
```
jeherve, line 89: Could we use `Jetpack_Tracks_Client::get_connected_user_tracks_identity()` here?

sbarbosa: Sure! I will make the change and test it. Thanks for the suggestion.

sbarbosa: I double-checked, and I still need to rename some of the attributes returned by the `get_connected_user_tracks_identity` method. I still need a method that acts as an intermediary and creates the dummy data (if a user is not connected), and add/rename the missing attributes.
The app is expecting the user to have the `ID` attribute, and the `localeSlug`.

What I can do here is use the `get_connected_user_tracks_identity` method instead of the `( new Manager() )->get_connected_user_data()`.

What do you think about that change?

jeherve: >  still need to rename some of the attributes returned by the get_connected_user_tracks_identity method.

Ah, in this case I suppose it's fine as it is today. 👍

jeherve (approved):

timur987 (approved):
Thanks @sbarbosa!

LGTM and works as advertised! 👍

PR #31665, opened 2023-07-03
Social: Move settings to a modal from the editor sidebar

pablinos:
This moves the Social Image Generator settings to a modal from the block editor sidebar. Eventually we should be able to move all the social settings for the post to a modal window.

CleanShot 2023-07-19 at 14 55 21 png

Fixes #

## Proposed changes:

* Adds the new modal, where the Template picker and the custom header props live from now
* Saves on modal save

### Other information:

## Jetpack product discussion

1204930070946424-as-1204411047112864

No.
## Testing instructions:

* Open up the Editor with the Advanced plan, there should be only a Settings button now under the SIG panel
* Clicking it opens up the modal, where there is a preview on top, and settings below
* Play around with the settings, it should get reflected in the modal preview
  * Saving the modal will change the preview in the editor to the new one instantly
  * Saving the post and refreshing should result in seeing the most up to date version


projects/js-packages/publicize-components/src/components/media-picker/index.js
```diff
@@ -4,10 +4,18 @@ import { ResponsiveWrapper, Spinner, VisuallyHidden } from '@wordpress/component
...
 import { Icon, closeSmall } from '@wordpress/icons';
+import classNames from 'classnames';
 import { isVideo } from '../../hooks/use-media-restrictions';
 import VideoPreview from '../video-preview';
 import styles from './styles.module.scss';
 
+// This is to cope with the problem describeed here:
+// https://github.com/WordPress/gutenberg/blob/ebad47952fc94ce4324e989691dde2d3d7689f45/packages/block-editor/src/components/inserter/media-tab/media-tab.js#L122
+const clickHandler = open => e => {
+	e.target.focus();
```
pablinos: I think we might want to change this to `e.currentTarget.focus()` I noticed that when an image is selected, clicking on it to change it means the target is the image and not the button.

pablinos (commented):
The changelog/linter check is failing. Probably something related to package versions.

The media picker for the custom image is unnecessarily huge! 
image

We also seem to use the square thumbnail for certain images, and that seems odd to me. Could we a version of the image that maintains the aspect ratio and constrain it somehow?

This also makes me wonder if we should always display the picker and move to the two column layout that we discussed.

gmjuhasz:
This is weird, it's not this huge for me.. Taking a look!

Update: It's working fine for me. Could it be that you tested with not the latest changes?

pablinos:
I think it depends on the image. Is it using the thumbnail for you? It does that if the original image is small.

I could also have not got the new changes. I'll test it when I'm back, unless @spsiddarthan can check it? But I guess we've missed code freeze so it can probably wait.

gmjuhasz:
> I think it depends on the image. Is it using the thumbnail for you?

It could depend on the image, but even the picker itself is not that big for me, which is weird. Also the image should never be bigger with the container with the CSS changes I did hopefully
CleanShot 2023-07-25 at 15 26 03 png

spsiddarthan:
I will give this a test!

spsiddarthan:
For some reason, I get an error when I open the editor, I checked on both social and jetpack plugins on JN with Jetpack beta. 
Screenshot 2023-07-27 at 13 41 43


@gmjuhasz, do you see this too?

gmjuhasz:
> For some reason, I get an error when I open the editor, I checked on both social and jetpack plugins on JN with Jetpack beta.

@spsiddarthan I did not see it, but I have a feeling that it is connected to the modified structure with the other change, as this is not rebased yet, so using the new object structure would result an error with `dismissedNotices` not being an array. I'll rebase it

projects/js-packages/publicize-components/src/components/media-picker/index.js
spsiddarthan: Perhaps this is a chang we still need to make @gmjuhasz ?

spsiddarthan (commented):
This tests well for me, however, I am not 100% sure about how to check if the thumbnail is being loaded correctly. It looks like this for me for a standard rectangular image. Sems ok to me, but perhaps @pablinos  can confirm when he is back? 

It looks good otherwise and I am leaning more toward merging it, pending Paul's confirmation when he is back. 

Screenshot 2023-07-27 at 15 14 44

projects/js-packages/publicize-components/src/components/media-picker/index.js
gmjuhasz: Yes! On it now.

pablinos:
That's exactly the problem @spsiddarthan, a huge grey border around the thumbnail image. I'll let you have the image I'm using @gmjuhasz, as it's still happening for me with everything updated.

spsiddarthan (approved):

PR #30088, opened 2023-04-13
Jetpack Forms: Update the styling on response emails

donnchawp:
Jetpack Forms sends out lots of emails to website owners and they are very basic right now. This PR will add filters to change content more easily and a new, more attractive default style to those emails.

Filters added:
* jetpack_forms_response_email - fired when creating the HTML of the response. An array is passed to it of response field label => value arrays and return a string with formatted HTML if it is to be used.
* jetpack_forms_response_email_title - The title of the email message shown in bold in the email body. Default: "You got a new response!"
* jetpack_forms_response_email_footer - an array of meta information about the response such as time, IP, source URL, and user information.
* jetpack_forms_respone_email_template - fired when wrapping the message in HTML tags. Pass it a string containing the boiler plate open and closing tags of a HTML page with %1$s for the message title, %2$s for the response, %3$s for the response link, %4$s for the link to the form on the website, %5$s for the meta data footer.

This allows website owners to customize the response emails sent from their website.

For context: pcsBup-yo-p2

Emails before:
image

After:

## Does this pull request change what data or activity we track or use?
No

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion
pcsBup-yo-p2

## Testing instructions:
Apply patch
Add a contact form to a website and fill it in
Check your inbox.
The email will look much the same as before but you'll see two extra parts:
* A header "You got a new response!" message
* A footer "View Response" message

PR #31667, opened 2023-07-03
Contact Form: remove default title in feedback emails

### LINKED ISSUE #31516: Enhancement: Remove the 'You got a new response' title in contact form emails
### Impacted plugin

Jetpack

### What

Contact form emails contain an unnecessary title

### How

At some point recently the emails sent by the Jetpack contact form have a 'You got a new response' title added to the email body in a large font. As an informational bit of text this is not needed, but if you then reply to the message it gets included in the email and just looks very odd.

image

Each time I have to manually edit it out. Could we consider not having the title?

jeherve:
Fixes #31516

## Proposed changes:

Following the feedback we received after adding that title in #30088, let's default to an empty title, while still allowing folks to add their own custom title via the `jetpack_forms_response_email_title` filter.

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

* Add a new contact form to one of your pages.
* Visit that page while logged out, and submit a form.
* Check your inbox for the notification email for that submitted form.
    * You should not see any "You got a new response!" title at the top of the email.

### DIFF
--- a/projects/packages/forms/src/contact-form/class-contact-form.php
+++ b/projects/packages/forms/src/contact-form/class-contact-form.php
@@ -1399,7 +1399,7 @@ class Contact_Form extends Contact_Form_Shortcode {
 		 *
 		 * @param string the title of the email
 		 */
-		$title   = apply_filters( 'jetpack_forms_response_email_title', __( 'You got a new response!', 'jetpack-forms' ) );
+		$title   = apply_filters( 'jetpack_forms_response_email_title', '' );
 		$message = self::get_compiled_form_for_email( $post_id, $this );
 
 		if ( is_user_logged_in() ) {
--- a/projects/packages/forms/src/contact-form/class-contact-form.php
+++ b/projects/packages/forms/src/contact-form/class-contact-form.php
@@ -1689,7 +1689,7 @@ class Contact_Form extends Contact_Form_Shortcode {
 				'',
 				$template
 			),
-			$title,
+			( ! empty( $title ) ? '<h1>' . $title . '</h1>' : '' ),
 			$body,
 			'',
 			'',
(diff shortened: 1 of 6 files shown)


anomiex (dismissed):
Seems to work, renders fine in gmail.

I note the HTML includes an empty `<h1></h1>` from https://github.com/Automattic/jetpack/blob/da0dd4c5843e9c26baa17e6cab7346040ac41cb8/projects/plugins/jetpack/modules/contact-form/grunion-response-email-template.php#L20. Do we care?

jeherve:
That's a good point. Maybe we should care. I pushed d202ed6151837f31ecdd7d781446ec8363fdc0e2 to address that.

projects/packages/forms/src/contact-form/class-contact-form.php
```diff
@@ -1689,7 +1689,7 @@ public static function wrap_message_in_html_tags( $title, $body, $footer ) {
 				'',
 				$template
 			),
-			$title,
+			( ! empty( $title ) ? '<h1>' . $title . '</h1>' : '' ),
```
anomiex: No need for `empty` since it's always going to be set. Also having `! empty() ? ... : ...` is really kind of confusing.

As for the replacement, I'm most inclined to do like this
```suggestion
			( $title !== '' ? '<h1>' . $title . '</h1>' : '' ),
```
but then in case of people doing `__return_false` or something incorrect like that for the filter we may want to add a string cast on line 1402.

jeherve: 👍 Done in b8021145bd1a638e430306fa01200b2b03fdc6e7 and 6cd95724451f749541d89f44049f659f56c87502

anomiex (approved):

PR #31670, opened 2023-07-03
Boost: cut off image filenames in the IG analysis report at 20 characters to avoid overflow

### LINKED ISSUE #31430: Boost: Image Size Analysis - Make sure images with long names don't break the UI.
Long image names overlap with other UI elements.

image

donnchawp:
The Image Guide Size Analysis report lists images, but if the titles of any image are very long they will overflow and break the UI. This PR fixes that by cutting the filename at 20 characters and adding "..." at the end.

Fixes #31430 

### Other information:

No

## Testing instructions:

1. On a test site create a post with an image with a filename that is over twenty characters long.
2. Install Jetpack Beta to install Boost. Look for the "fix/overflowing_filenames_image_guide_analysis" feature branch.
3. Upgrade to a premium subscription
4. Enable image size analysis by defining the constant JETPACK_BOOST_IMAGE_SIZE_ANALYSIS.
5. Go to the Boost page and enable and run "Image Size Analysis".
6. The report will show the filename of the image cut off at 20 characters and "..." at the end of it.

### DIFF
--- a/projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/recommendations/components/RowTitle.svelte
+++ b/projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/recommendations/components/RowTitle.svelte
@@ -1,11 +1,11 @@
 <script lang="ts">
+	import { truncateString } from '../../../../utils/truncate-string';
 	export let title: string;
 	export let url: string;
 	const maxTitleLength = 20;
 
 	const urlWithoutProtocol = url.replace( /^https?:\/\/(www\.)?/, '' );
-	const croppedTitle =
-		title.length > maxTitleLength ? title.slice( 0, maxTitleLength ) + '...' : title;
+	const croppedTitle = truncateString( title, maxTitleLength );
 </script>
 
 <b>{croppedTitle}</b>
--- a/projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/recommendations/components/RowTitle.svelte
+++ b/projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/recommendations/components/RowTitle.svelte
@@ -1,11 +1,14 @@
 <script lang="ts">
 	export let title: string;
 	export let url: string;
+	const maxTitleLength = 20;
 
 	const urlWithoutProtocol = url.replace( /^https?:\/\/(www\.)?/, '' );
+	const croppedTitle =
+		title.length > maxTitleLength ? title.slice( 0, maxTitleLength ) + '...' : title;
 </script>
 
-<b>{title}</b>
+<b>{croppedTitle}</b>
 <a href={url} target="_blank">
 	{urlWithoutProtocol}
 </a>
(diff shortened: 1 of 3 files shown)


dilirity (changes requested):
Thanks for addressing this!

I think this would be better in a util function (`projects/plugins/boost/app/assets/src/js/utils`).

Then the change to the template would just be:

```js
<b>{addEllipsisToString( title, 20 )}</b>
```

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/recommendations/components/RowTitle.svelte
```diff
@@ -1,11 +1,11 @@
 <script lang="ts">
+	import { truncateString } from '../../../../utils/truncate-string';
 	export let title: string;
 	export let url: string;
 	const maxTitleLength = 20;
 
 	const urlWithoutProtocol = url.replace( /^https?:\/\/(www\.)?/, '' );
-	const croppedTitle =
-		title.length > maxTitleLength ? title.slice( 0, maxTitleLength ) + '...' : title;
+	const croppedTitle = truncateString( title, maxTitleLength );
```
dilirity, line 8: Since the default is 20, you can get rid of `maxTitleLength`.

projects/plugins/boost/app/assets/src/js/utils/truncate-string.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Returns a string truncated to a given length.
+ * If truncated, the string will end with an ellipsis.
+ *
+ * @param {string} str       string to truncate
+ * @param {number} maxLength maximum length of the string
+ */
+export function truncateString( str: string, maxLength?: number ): string {
+	if ( ! maxLength ) {
+		maxLength = 20;
+	}
```
dilirity: You can just do:

```js
truncateString( str: string, maxLength = 20 )
```

instead.

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/recommendations/components/RowTitle.svelte
donnchawp: That's tempting, but it should probably stay, just in case we change the default value in truncateString and suddenly reports look awful again.

dilirity: Ahaa! Great thinking!

projects/plugins/boost/app/assets/src/js/utils/truncate-string.ts
donnchawp: Done, thanks.

dilirity (approved):

thingalon:
I think 20 characters is too short in many circumstances. Can we [do this with CSS instead](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow)? 

Wrote up a new ticket to switch to CSS name length limiting. https://github.com/Automattic/boost-cloud/issues/300

PR #31675, opened 2023-07-04
Fix/boost e2e

thingalon:
I've noticed our E2e tests have been failing lately. This is an attempt to fix it. The issue I found in testing was a redirection loop --> prices failed to load, so it kept on redirecting back adn forth between the getting started page and the dashboard.

## Proposed changes:

* Don't skip the getting started page if the pricing info fails to load - getting started is a must-use connection point now.

### Other information:

## Testing instructions:
* make sure you can use the plugin ok
* Make sure the plugin still works even if loading pricing information fails
* Make sure e2e tests pass


dilirity (approved):
> Make sure the plugin still works even if loading pricing information fails

I wasn't sure how to test this, but the plugin works as expected. Tests no longer fail and the code looks good.

I say we :shipit:

PR #31678, opened 2023-07-04
Newsletter panel: update copy for activating subscribers feature

simison:
- Remove "yours" which can sound a bit off for multi-author sites
- Condense
- Add "sending" to imply emails instead of other kinds of sharing

**Before**

**After**

> In order to send posts to your subscribers, activate the Subscriptions feature.
>
> Activate Subscriptions
> 
> Learn more about Subscriptions

## Proposed changes:

*

### Other information:

## Testing instructions:

* On Jetpack site with subscriptions module disabled
* Go to post editor, and open Newsletter panel

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
@@ -38,7 +38,7 @@ const SubscriptionsPanelPlaceholder = ( { children } ) => {
 		<Flex align="center" gap={ 4 } direction="column" style={ { alignItems: 'center' } }>
 			<FlexItem>
 				{ __(
-					"In order to send posts to your subscribers, activate the Subscriptions feature.",
+					'In order to send posts to your subscribers, activate the Subscriptions feature.',
 					'jetpack'
 				) }
 			</FlexItem>
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
@@ -45,7 +45,7 @@ const SubscriptionsPanelPlaceholder = ( { children } ) => {
 			<FlexItem>{ children }</FlexItem>
 			<FlexItem>
 				<ExternalLink href="https://jetpack.com/support/subscriptions/">
-					{ __( 'Learn more', 'jetpack' ) }
+					{ __( 'Learn more about Subscriptions', 'jetpack' ) }
 				</ExternalLink>
 			</FlexItem>
 		</Flex>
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
```diff
@@ -38,14 +38,14 @@ const SubscriptionsPanelPlaceholder = ( { children } ) => {
 		<Flex align="center" gap={ 4 } direction="column" style={ { alignItems: 'center' } }>
 			<FlexItem>
 				{ __(
-					"In order to share your posts with your subscribers, you'll need to activate the Subscriptions feature.",
+					"In order to share posts as emails with your subscribers, activate the Subscriptions feature.",
```
simison: @crisbusquets how do you feel about this? Alternative I considered but felt like the "share posts as emails" would be good to be early in case people don't read more than few words when skimming.

_Activate the Subscriptions feature to share posts as emails with your subscribers._

Overall also condenced the text, and added "emails" because sharing could also just mean social sharing or other kind of sharing.

crisbusquets: @simison what about:
`In order to send posts to your subscribers, activate the Subscriptions feature.` ?

I think `send` implies `emails`.

@mtias suggested reducing the _Learn more_ link as well. What about:
Original: `Learn more about the Subscriptions feature.`
Proposal: `Learn more about Subscriptions`

simison: > In order to send posts to your subscribers, activate the Subscriptions feature.

I like it! 🚢 

> Original: Learn more about the Subscriptions feature.
> Proposal: Learn more about Subscriptions

I shortened this already to "Learn more" earlier, because it was indeed lengthy and 3rd repetition of "subscriptions" after description and CTA so reading them all together felt clunky. Lemme know if you prefer your proposal and I'm happy to update.

simison: I'll update to your version of "learn more" after your comment on Slack:

> I thought about that too, but as far as I know, it's not recommended to use learn more, read more, and others. They don't provide enough context.
> I used this other "learn more" as a reference:
>
>

lezama:
@crisbusquets Is the button here not left aligned (compared to the Send Email Preview button) on purpose?

crisbusquets:
> @crisbusquets Is the button here not left aligned (compared to the Send Email Preview button) on purpose?

Oh, I didn't see that :( thanks for the ping, @lezama!

We should align all the buttons to maintain some consistency. I'd align this to the left to match the other two.

jeherve (dismissed):

simison:
> We should align all the buttons to maintain some consistency. I'd align this to the left to match the other two.

Yup! I'll merge this and follow-up in another PR.

@jeherve oops, I dismissed your review with small lint-fix commit.

jeherve (approved):

PR #31684, opened 2023-07-04
Fix sonar qube issues

### LINKED ISSUE #31683: [Flagship plugin]: Sonar Qube: Refactor suggested issues

andrii-lysenko:
Fixes #31683

## Proposed changes:

* I did run a [Sonar Qube](https://docs.sonarqube.org/latest/) over Jetpack plugin's related packages in the Monorepo and fixed most important in my opinion. Many issues are non relative, or very low significance, which made fixing them not worth the risk in my opinion.

### Other information:

## Testing instructions:

* Go to '..'
*


projects/js-packages/components/components/boost-score-bar/index.tsx
```diff
@@ -81,7 +81,7 @@ export const BoostScoreBar: FunctionComponent< BoostScoreBarProps > = ( {
 						<div className="jb-score-bar__score">{ score }</div>
 					) }
 
-					{ showPrevScores && prevScore && prevScore < score && (
+					{ showPrevScores && !! prevScore && prevScore < score && (
```
andrii-lysenko, line 84: In rare occasion, it could render 0.

projects/js-packages/components/tools/jp-redirect/index.ts
```diff
@@ -38,7 +38,7 @@ export default function getRedirectUrl( source: string, args: GetRedirectUrlArgs
 		queryVars.source = encodeURIComponent( source );
 	}
 
-	Object.keys( args ).map( argName => {
+	Object.keys( args ).forEach( argName => {
```
andrii-lysenko: There's no need for `map` as it will create new array, while we don't need it.

projects/js-packages/connection/components/disconnect-survey/index.jsx
```diff
@@ -129,6 +129,7 @@ const DisconnectSurvey = props => {
 		return (
 			<SurveyChoice
 				id={ customOption.id }
+				key={ customOption.id }
```
andrii-lysenko, line 132: `key` helps React with better rendering preformance

projects/packages/videopress/src/client/hooks/use-uploader.js
```diff
@@ -95,7 +95,7 @@ export const uploadVideo = ( { file, onProgress, onSuccess, onError, data } ) =>
 
 			req._xhr.open( req._method, req._url, true );
 			// Set the headers again, reopening the xhr resets them.
-			Object.keys( req._headers ).map( function ( headerName ) {
+			Object.keys( req._headers ).forEach( function ( headerName ) {
```
andrii-lysenko: There's no need for map as it will create new array, while we don't need it.

projects/packages/videopress/src/client/hooks/use-uploader.js
```diff
@@ -116,7 +116,7 @@ export const uploadVideo = ( { file, onProgress, onSuccess, onError, data } ) =>
 				if ( jwtsForKeys[ maybeUploadkey ] ) {
 					req.setHeader( 'x-videopress-upload-token', jwtsForKeys[ maybeUploadkey ] );
 				} else if ( 'HEAD' === method ) {
-					return getJWT( maybeUploadkey ).then( responseData => {
+					return getJWT().then( responseData => {
```
andrii-lysenko, line 119: `getJWT` doesn't accept attributes.

projects/packages/waf/src/class-waf-runtime.php
```diff
@@ -222,7 +222,7 @@ public function match_targets( $transforms, $targets, $match_operator, $match_va
 				$v['value'] = $this->transforms->$t( $v['value'] );
 			}
 		}
-
+		unset( $v );
```
andrii-lysenko, line 225: Not an expert, but suggested unset value after `foreach ( $values as &$v )`

projects/js-packages/components/tools/jp-redirect/index.ts
anomiex: You could go a step further and turn it into `for ( const argName of Object.keys( args ) ) {`.

monsieur-z (dismissed):
Code changes look sensible 👍

bindlegirl:
LGTM but pinging @anomiex for a second opinion.

anomiex (dismissed):
LGTM

samiff (approved):

PR #31696, opened 2023-07-05
Jetpack Dashboard: add new "Earn" section

jeherve:
## Proposed changes:

This PR adds a new "Earn" section to the Jetpack Settings at Jetpack > Settings. That section currently includes: 

- the "Ads" section, previously found at Jetpack > Settings > Traffic
- the contents of the existing `https://wordpress.com/earn/` page, but could / should be updated in the future.

**Screenshot**

- When the user is not connected:

image

- When the user is connected:

image

### Other information:

## Jetpack product discussion

* p1688481163954019-slack-C0299DMPG

* Yes, all clicks in the new dashboard screen are tracked, like in other areas of the dashboard.

## Testing instructions:

* Go to Jetpack > Settings > Earn
* Check contents of the new page.
    * Try this with different plans and statuses.


projects/plugins/jetpack/_inc/client/admin.js
```diff
@@ -125,6 +128,8 @@ export function getRouteName( path ) {
 			return _x( 'Settings', 'Navigation item.', 'jetpack' );
 		case '/discussion':
 			return _x( 'Discussion', 'Navigation item.', 'jetpack' );
+		case '/earn':
```
simison, line 131: FYI we will rename the section in WP.com soon to "Monetize" in upcoming projects.

We can rename it simultaneously across Jetpack and WP.com then!

projects/plugins/jetpack/_inc/client/components/settings-card/index.jsx
```diff
@@ -302,6 +303,36 @@ export const SettingsCard = props => {
...
+				}
+
+				return props.hasConnectedOwner ? (
+					<JetpackBanner
+						callToAction={ upgradeLabel }
+						title={ __(
+							'Start accepting PayPal payments for physical products, digital goods, or donations.',
+							'jetpack'
+						) }
+						plan={ getJetpackProductUpsellByFeature( FEATURE_SIMPLE_PAYMENTS_JETPACK ) }
```
simison, line 318: Is this feature slug correct? It's the same as for paypal feature below.

simison: Ignore, I misread the code a bit!

projects/plugins/jetpack/_inc/client/components/settings-card/index.jsx
```diff
@@ -302,6 +303,36 @@ export const SettingsCard = props => {
...
+						feature={ feature }
+						href={ props.simplePaymentsUpgradeUrl }
+						rna
+					/>
+				) : (
+					<JetpackBanner
+						callToAction={ connectLabel }
+						title={ __(
+							'Connect your WordPress.com account to upgrade and access PayPay features in your editor.',
```
simison: PayPay -> PayPal

projects/plugins/jetpack/_inc/client/earn/index.jsx
```diff
@@ -0,0 +1,202 @@
...
+					title={ __( 'Collect payments', 'jetpack' ) }
+					supportLink={ getRedirectUrl( 'jetpack-support-jetpack-blocks-payments-block' ) }
+					infoLink={ getRedirectUrl( 'wpcom-earn-payments', {
+						site: siteRawUrl,
+					} ) }
+					infoDescription={ __(
+						'Let visitors pay for digital goods and services or make quick, pre-set donations by enabling the Payment Button block.',
+						'jetpack'
+					) }
+					buttonText={ __( 'Enable Payment Button', 'jetpack' ) }
```
simison: I'm not sure about Jetpack copy style but you use a lot of "Title Case" instead of "Sentence case". :-) I'd expect more latter e.g. at WP.com

simison (dismissed):
This is a great start! Tests well.

We can keep iterating on copy, design, layout etc.

For future iterations, we are brainstorming some copy around these features for plans, which can help here too: 
pdgrnI-2xq-p2/#comment-4144

We're also re-designing Earn section in WP.com, which will surely inform copy here too.

One small thing that can be fixed in this PR or in a follow up is this "upgrade" button's alignment problem on mobile version:

Screenshot 2023-07-12 at 23 12 39

simison:
FYI, I rebased as I wanted to test against latest `trunk` myself.

projects/plugins/jetpack/_inc/client/components/settings-card/index.jsx
jeherve: 🤦 
201003dbf89e5542182fe47833a7a8539ddc9931

projects/plugins/jetpack/_inc/client/earn/index.jsx
jeherve: I'm not quite sure in this case, since those are feature names and not just names. I'm happy to make the change though. c9b1196b6432e5ccabb1165f17a02bc73134387d

jeherve:
> One small thing that can be fixed in this PR or in a follow up is this "upgrade" button's alignment problem on mobile version:

That's a good point, good catch. That's something I would address in a separate PR though, since that button is common and used in other places of the dashboard.

bindlegirl:
The Upgrade button on "Start accepting PayPal payments for physical products, digital goods, or donations." loads the Jetpack dashboard instead of offering an upgrade/checkout page. Is that expected?

jeherve:
> The Upgrade button on "Start accepting PayPal payments for physical products, digital goods, or donations." loads the Jetpack dashboard instead of offering an upgrade/checkout page. Is that expected?

That's my bad. Since there is no Simple Payments plan, in 4462cfbef2d8a134ec72652682c04c899039e2e2 I updated the link to point to the security bundle instead, which includes access to simple payments.

Good catch, thank you!

bindlegirl (approved):

simison (approved):

PR #31700, opened 2023-07-05
Social: Refactor SIG preview component

gmjuhasz:
This change refactors the preview component a bit, so the values can come from props, instead of the hooks. It is also possible to turn off the debounce, as in the editor we won't need it anymore after introducing the modal in #31665.

### Other information:

## Testing instructions:

* Create a new post with the Advanced plan and SIG turned on.
 * Changing the following should be reflected on the preview
  * Changing the title
  * Changing the description
  * Template
  * Image type, or image
  * Debouncing should work on text changes as before

* In `projects/js-packages/publicize-components/src/components/social-image-generator/panel/index.js` update the GeneratedImagePreview, and add properties to check that they overwrite the hook values. For example if `customText` is provided, it should be there all the time


spsiddarthan (commented):
> It is also possible to turn off the debounce, as in the editor we won't need it anymore after introducing the modal in https://github.com/Automattic/jetpack/pull/31665.

Are we turning off the debounce because every time you need to see the preview you need to open the modal?

> In projects/js-packages/publicize-components/src/components/social-image-generator/panel/index.js update the GeneratedImagePreview, and add properties to check that they overwrite the hook values. For example if customText is provided, it should be there all the time

Pretty sure, I am doing this wrong like this `<GeneratedImagePreview customText={ 'some text' } />
`, can you tell me how to do this bit correctly, pass an object I think? 

This tests well per the instructions, just needs a rebase.

gmjuhasz:
> Are we turning off the debounce because every time you need to see the preview you need to open the modal?

So after we have the modal, we will have the preview in 2 places, one in the modal, and one in the editor as before. In the modal we will need the debounce, because they will set the custom text there. On the other hand, in the editor we want the preview to appear instantly after they are closing/saving the modal, and not wait the 1.5 seconds for the preview to update, if that makes sense.

> Pretty sure, I am doing this wrong like this <GeneratedImagePreview customText={ 'some text' } /> , can you tell me how to do this bit correctly, pass an object I think?

Weird, that should work! For example doing this should make the preview show `mycustomtext` all the time:
`<GeneratedImagePreview shouldDebounce={ false } customText="mycustomtext" />`.
I just tried it and it worked for me, are you watching the the files? If it's not working for you let's pair up and debug it!

spsiddarthan (approved):
LGTM and thanks for explaining

PR #31711, opened 2023-07-05
codesniffer: Add "compat" rulesets for per-dir configs

anomiex:
## Proposed changes:

One annoying thing about the per-dir configs is that we can't set PHPCompatibility's `testVersion` per directory. We work around that by instead excluding the relevant rules in the config.

Unfortunately it's easy for people to screw this up, e.g. ignoring all of `PHPCompatibility.FunctionDeclarations.NewReturnTypeDeclarations` when some like `.mixedFound` shouldn't be.

So let's make some centralized rulesets that ignore only the rules that need ignoring per PHP version so they can be used in the per-directory configs. It looks like PHPCompatibility has pretty comprehensive tests based on fixture files, so we should be able to get a pretty comprehensive list of the rules by running phpcs over those files.

### Other information:

## Jetpack product discussion

Thought of this while writing https://github.com/Automattic/jetpack/pull/31551#discussion_r1253452977

No

## Testing instructions:

* Is CI happy?


sdixon194:
Wanted to make sure the e2e failure wasn't from this PR but doesn't seem to be related.

sdixon194 (approved):

PR #31718, opened 2023-07-06
Subscriptions: add a Subscribers menu to the dashboard

jeherve:
## Proposed changes:

The menu will only appear on self-hosted sites, since WordPress.com sites (simple and WoA) already get a Subscribers menu under "Users".

~That menu will either link to Calypso (by default), or to the Stats screen when the Stats module is enabled and Odyssey stats are enabled.~
That menu will always link to Calypso, and will appear if your site is connected to WordPress.com, not offline, and when the Subscriptions module is active.

### Other information:

## Jetpack product discussion

* p1688642967595329/1688481163.954019-slack-C0299DMPG
* pbNhbs-7bW-p2
* pbNhbs-7F3-p2#comment-17295

* No

## Testing instructions:

Test this on a self-hosted site, connected to WordPress.com, with the Subscriptions module active.

* You should see a Subscribers menu under Jetpack.
* ~That menu should lead you to a different place whether the Stats module is active or not.~
* ~When linking to wp-admin, it should open the Subscribers tab of the Stats screen.~
* That menu should lead to the relevant Subscribers page in Calypso


TimBroddin (approved):
I tested this PR.

After enabling subscribers & refreshing, the menu item appears:

CleanShot 2023-07-12 at 09 26 18@2x

Clicking it with the Stats module active brings me to the subscribers stats page. One thing I notice is the Stats item is highlighted instead of the Subscribers item. 

CleanShot 2023-07-12 at 09 27 54@2x

When the Jetpack Stats module is deactivated, clicking this link takes you to:

CleanShot 2023-07-12 at 09 33 03@2x

One last oddity I noticed is that when you opt-out of the new stats experience, clicking the link takes you to the stats, but there's no mention of subscribers on that (old) page. 

CleanShot 2023-07-12 at 09 38 13@2x

Both these problems can probably be addressed in follow-up PR's, so I'm approving this PR to unblock this.

TimBroddin (dismissed):
Approving once again as Apex (hopefully)

jeherve:
> ne thing I notice is the Stats item is highlighted instead of the Subscribers item.

Ah, good catch! I wonder if it may be worth registering a new menu item just to solve that.

> When the Jetpack Stats module is deactivated, clicking this link takes you to:

That's indeed the expected behaviour.

> One last oddity I noticed is that when you opt-out of the new stats experience, clicking the link takes you to the stats, but there's no mention of subscribers on that (old) page.

Is it possible that you clicked on the link right from the Jetpack React dashboard, after clicking on the opt-out toggle? In this case, since the dashboard page does not refresh, the link does not get the chance to refresh. 

To fix that I would have to force a refresh of the page when you click that toggle, but I do not know if it is worth it?

simison:
Why is there a connection between Stats module and linking to Subscribers management? I'd expect Subscribers module needing to be enabled.

jeherve:
> Why is there a connection between Stats module and linking to Subscribers management?

Today, you can access information about your Jetpack subscribers in 4 places:

1. In Calypso, in the new `/subscribers` page.
2. In Calypso, in the old `people/subscribers` -> this page is being deprecated.
3. In Calypso, in `stats/subscribers/` -> **that page is only reachable if the Stats module is enabled on your site**
4. In wp-admin, in `wp-admin/admin.php?page=stats#!/stats/subscribers/` (this is the same content as the above item) -> **that page is only reachable if the Stats module is enabled on your site, and if you have enabled the new Stats experience in your Stats settings**

This PR introduces a new menu item, and that menu item will link to 1) or 4) depending on your site settings.

That whole setup isn't simple or straightforward, but that's what we have :)

> I'd expect Subscribers module needing to be enabled.

Yes, that code will only run and a Subscribers menu item will only be added when the Subscriptions module is enabled.

simison:
Wow, thanks for elaborating those details! It's quite something indeed.

projects/plugins/jetpack/modules/subscriptions.php
```diff
@@ -1031,6 +1040,64 @@ public function register_post_meta() {
...
+				__( 'Subscribers', 'jetpack' ),
+				'view_stats',
+				'stats#!/stats/subscribers/' . $domain,
+				array( $stats_dashboard, 'render' )
+			);
+			add_action( "load-$hook", array( $stats_dashboard, 'admin_init' ) );
+		} else {
+			add_submenu_page(
+				'jetpack',
+				esc_attr__( 'Subscribers', 'jetpack' ),
```
lezama: is `esc_attr` necessary?  Should we do it in [the other call](https://github.com/Automattic/jetpack/pull/31718/files#diff-776cd74bed19fb81f284b9dbfd8258e57d5fb0a5db07c0e1db94e47dbb0d6182R1083) too?

jeherve: Good call. Done in ce5ad18308fa65561f29f684b2ffe3461e9c45ae

zinigor (commented):
I got it to work, but here I found a small thing that I would like to nitpick about. Do you see how the Stats sidebar entry is highlighted even though I am on the Subscribers tab?
изображение
I would say it's minor, but it's a very annoying thing.
Besides that everything's working great!

mikestottuk (changes requested):
The code looks good. I also see the Stats side menu getting highlighted. If we can fix that then great this can probably make 12.4.

I have also added #32007 to look to bring the subscription management to be able to be managed within wp-admin. 

I'm not sure on the technicalities (i.e. does it have to be part of Stats) or can it be its own page similar to stats. So even if they disable Stats (i.e. the tracking pixel) they'd still see the Dash (with some stats) for their subscribers?

If you link to here `wp-admin/admin.php?page=stats#!/stats/subscribers/{domain}` for Subscribers you still miss the ability to do anything with those subscribers like you can via this link.

`https://wordpress.com/subscribers/{domain}`

Such as import, view their details etc, which #32007 looks to start to solve. 

subscribers page

cc @jboland88

jeherve:
@zinigor Do you think you could check again when you get the chance? I'm curious to see if my changes helped.

Thank you!

zinigor:
Thanks for making the change @jeherve ! It works for me, but with a very weird twist, take a look at the video to see what I mean. It looks like the link changes to active only when I click it the second time for some reason:
[Screencast from 2023-07-27 19-33-29.webm](https://github.com/Automattic/jetpack/assets/374293/13638adf-51df-4a89-befb-d5c567ce673c)

jeherve:
@zinigor Thanks for bearing with me on this. Following conversation on p2, we've decided to always link to Calypso. That's fine by me as it simplifies things quite a bit, and avoids the workarounds I had previously set up. So this should be simpler to review now :)

simison:
@jeherve I think this link should link to the same place as this sidebar:

image

And we could have additional link to Email & Subscriber stats in that place, too.

jeherve:
> I think this link should link to the same place as this sidebar

Agreed. Done in #32393.

> And we could have additional link to Email & Subscriber stats in that place, too.

I agree that would be nice, but I'm not quite sure how this should appear. The 2 links under each other may not look too nice?

zinigor (approved):
Now it works as expected, and the only really tiny corner case is that when you disable newsletters on Jetpack side, the sidebar link is still there until you refresh the page. But that's not that significant.

zinigor:
@mikemayhem3030 this one is now waiting for your approval. Can you please take another look?

kraftbj:
Updated the label to reflect a requested review from @mikemayhem3030 . Good to merge with that.

mikestottuk (approved):
Tested it from a UX perspective and it looks to link fine. 

My only observation is around the persistence of the subscribers area, if you disable the setting, you can still see subscribers under stats, and also on WP.com but that's a different issue IMO and isn't a blocker for this PR.

jeherve:
>  persistence of the subscribers area, if you disable the setting, you can still see subscribers under stats

That's something that would need to be fixed in Calypso. I logged it here: https://github.com/Automattic/wp-calypso/issues/80572

PR #31720, opened 2023-07-06
Subscriptions: update link to Subscribers page in Jetpack dashboard

jeherve:
## Proposed changes:

We currently link to an old Calypso page when folks want to see their subscribers after enabling the Subscriptions. We can do better than that.

In the past few months, 2 initiatives introduced 2 new screens that can be better than the existing one:

1. In Calypso, the `/subscribers` page includes more info and options than the existing `/people/subscribers` page.
2. In wp-admin (when the Stats feature is enabled, and when the new "Odyssey" stats experience is enabled), the Stats page includes a "Subscribers" page with extra info.

In this PR, we'll switch from the legacy to one of the 2 new options above, depending on availability.

I'm also updating the wording, but I'm not sure it's good enough at the moment:

image

### Other information:

## Jetpack product discussion

* p1688643479440219/1688481163.954019-slack-C0299DMPG
* pbNhbs-7bW-p2

* The tracking links have been updated to take the new link options into account.

## Testing instructions:

* Your site will need to be connected to WordPress.com, and the Subscriptions module must be enabled.
* Go to Jetpack > Settings > Discussion
* Ensure the link to view subscribers works. It should be updated depending on your site settings
    * Is the Stats module enabled (you can disable it under Jetpack > Settings > Module)
    * Is the new Stats experience enabled (you can check that under Jetpack > Settings > Traffic


TimBroddin (approved):
Works as intended!

projects/plugins/jetpack/_inc/client/discussion/subscriptions.jsx
```diff
@@ -56,26 +60,42 @@ class SubscriptionsComponent extends React.Component {
...
-					href={ getRedirectUrl( 'calypso-people-email-followers', {
+					onClick={ this.trackViewSubsClick }
+					href={ getRedirectUrl( 'calypso-subscribers', {
 						site: this.props.siteRawUrl,
 					} ) }
 					target="_blank"
 					rel="noopener noreferrer"
 				>
-					{ __( 'View your Email Subscribers', 'jetpack' ) }
+					{ __( 'Manage your Subscribers in the Subscribers dashboard', 'jetpack' ) }
```
simison: This could just be "Manage all subscribers"? No Title Case, sentence case.

To the point, shorter, and not repeating words.

simison (dismissed):

TimBroddin (dismissed):
Approving once again as Apex

projects/plugins/jetpack/_inc/client/discussion/subscriptions.jsx
jeherve: Sure thing, that works for me. Done in 943fd0f14957c0aeac0489e8d12c1bf0b60d9495

simison (dismissed):

samiff (approved):

Issue #31725, opened 2023-07-06
Jetpack: Several features - Only register customizer sections if current theme is not a block theme


oskosk:
Following up on https://github.com/Automattic/wp-calypso/issues/71130#issuecomment-1617982587, I'd like to make changes to rapidly address the issue with a temporary solution:

Only register customizer sections for [Social links](https://github.com/Automattic/jetpack/blob/2e69b7fe9f9eba8496932296fe022f698ccc1e8a/projects/plugins/jetpack/modules/theme-tools/social-links.php#L134), [Jetpack Search](https://github.com/Automattic/jetpack/blob/2e69b7fe9f9eba8496932296fe022f698ccc1e8a/projects/packages/search/src/customizer/class-customizer.php#L38), [Testimonials](https://github.com/Automattic/jetpack/blob/2e69b7fe9f9eba8496932296fe022f698ccc1e8a/projects/plugins/jetpack/modules/custom-post-types/testimonial.php#L532), [Related posts](https://github.com/Automattic/jetpack/blob/2e69b7fe9f9eba8496932296fe022f698ccc1e8a/projects/plugins/jetpack/modules/related-posts/class.related-posts-customize.php#L48), [Featured content](https://github.com/Automattic/jetpack/blob/2e69b7fe9f9eba8496932296fe022f698ccc1e8a/projects/plugins/jetpack/modules/theme-tools/featured-content.php#L511), and [Portfolios](https://github.com/Automattic/jetpack/blob/2e69b7fe9f9eba8496932296fe022f698ccc1e8a/projects/plugins/jetpack/modules/custom-post-types/portfolios.php#L83), if the current theme is a modern block theme.

Tasks:

* [x] #31838
* [x] #31730
* [x] #31731
* [x] #31754
* [x] #31755
* [x] #31779

oskosk:
All of these were addressed and merged. All of them except the change to social links have been deployed to WordPress.com today. The Social Links change will make it there tomorrow

PR #31730, opened 2023-07-06
Testimonials: Make the testimonials customizer extension only be shown if current theme is not a block theme

oskosk:
Fixes partially #31725

## Proposed changes:

* Check if current active theme is a block theme before attempting to add a section to the customizer

### Other information:

## Jetpack product discussion
https://github.com/Automattic/wp-calypso/issues/71130#issuecomment-1617982587

No

## Testing instructions:
* Checkout this branch
* Enable testimonials in Jetpack -> Settings -> Writing
* If you're using a modern theme expect to see no Customizer menu item under Appearance

### DIFF
--- a/projects/plugins/jetpack/modules/custom-post-types/testimonial.php
+++ b/projects/plugins/jetpack/modules/custom-post-types/testimonial.php
@@ -84,7 +84,9 @@ class Jetpack_Testimonial {
 		add_filter( 'enter_title_here', array( $this, 'change_default_title' ) );
 		add_filter( sprintf( 'manage_%s_posts_columns', self::CUSTOM_POST_TYPE ), array( $this, 'edit_title_column_label' ) );
 		add_filter( 'post_updated_messages', array( $this, 'updated_messages' ) );
-		add_action( 'customize_register', array( $this, 'customize_register' ) );
+		if ( ! wp_is_block_theme() ) {
+			add_action( 'customize_register', array( $this, 'customize_register' ) );
+		}
 
 		// Only add the 'Customize' sub-menu if the theme supports it.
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/modules/custom-post-types/testimonial.php
```diff
@@ -84,7 +84,9 @@ public function maybe_register_cpt() {
 		add_filter( 'enter_title_here', array( $this, 'change_default_title' ) );
 		add_filter( sprintf( 'manage_%s_posts_columns', self::CUSTOM_POST_TYPE ), array( $this, 'edit_title_column_label' ) );
 		add_filter( 'post_updated_messages', array( $this, 'updated_messages' ) );
-		add_action( 'customize_register', array( $this, 'customize_register' ) );
+		if ( ! function_exists( 'wp_is_block_theme' ) || ( function_exists( 'wp_is_block_theme' ) && ! wp_is_block_theme() ) ) {
```
jeherve: Since `wp_is_block_theme` was introduced in 5.9, I think we can do without the check if the function exists.

oskosk: Not sure I got you.You mean we can drop the `function_exists` check or just rely on its existence for assuming a modern installation?

jeherve: Sorry, I should have been clearer. I meant dropping the `function_exists` check.

oskosk: Cool. I'll revert the last commit

samiff (approved):

PR #31733, opened 2023-07-07
[Boost] Clean up 'no results' display

### LINKED ISSUE #31679: Boost: Image Size Analysis - don't load placeholders when no issues are found
### Impacted plugin

Boost

### Quick summary

If no issues are found, placeholders are loaded and groups aren't clickable.

image

### Steps to reproduce

1. Make sure there are no image issues on your site (you can do this by either removing images from pages or making pages/posts private);
2. Request a report;
3. Check the UI - it should show placeholders (same as image above).

Maybe we should show something else instead.

### What actually happened

_No response_

### Impact

One

### Available workarounds?

There is no user impact

_No response_

### Logs or notes

_No response_

thingalon:
Fixes #31679

When there are no results, show an explicit message stating so - instead of broken placeholders.
Also fixes incorrect language on report button.

## Proposed changes:
* When there are no results, show an explicit message stating so - instead of broken placeholders.
* Also fixes incorrect language on report button.

### Other information:

## Jetpack product discussion

n/a

no

## Testing instructions:

* Create an empty report
*View the UI.


dilirity (approved):
I fixed a bug where the ay reported issues weren't listed in the table and instead it was showing the "no issues found" label.

I've also updated the "last updated" header to show up if there are no issues, so the user knows when the last report was run.

:shipit:

PR #31734, opened 2023-07-07
[Boost] Conditional Image Guide links

thingalon:
Make the categories in the image guide only clickable if there are issues to show there. Linking to an empty page seems weird.

## Proposed changes:

* Make the categories in the image guide only clickable if there are issues to show there. Linking to an empty page seems weird.

### Other information:

## Jetpack product discussion

n/a

no

## Testing instructions:

* Generate an image guide report
* Ensure that any category with nothing to show doesn't have a clickable link.


dilirity (approved):
Good catch on this. I totally missed it when I added the links 😅 

:shipit:

PR #31738, opened 2023-07-07
[Boost] Fix E2E tests yet again

thingalon:
Fixes e2e tests

Switches from using `page.$eval` to using `page.evaluate` with `document.querySelector` instead. The code was finding a parent element, then waiting for the child to appear, then using the old reference to the parent element to access the child.

The problem is, with modern JS toolkits you sometimes end up rewriting parts of the DOM and the old parent may not be the parent that's in the DOM after other things chnage.

so this should be more reliable.

## Proposed changes:
* Switches from using `page.$eval` to using `page.evaluate` with `document.querySelector` instead.

### Other information:

## Jetpack product discussion

n/a

no

## Testing instructions:

* let the tests run, ensure they're happy

### DIFF
--- a/projects/plugins/boost/changelog/boost-fix-e2e-yet-again
+++ b/projects/plugins/boost/changelog/boost-fix-e2e-yet-again
@@ -1,0 +1,5 @@
+Significance: patch
+Type: fixed
+Comment: Fixed e2e test flakiness
+
+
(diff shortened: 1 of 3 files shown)


dilirity (approved):
:shipit:

PR #31740, opened 2023-07-07
Social: Refactor the TemplatePicker component

gmjuhasz:
This change refactors our current `TemplatePicker` component, so that the inner logic can be used separately, decoupled from the modal. This is needed so we can use the picker in the SIG modal for #31665 .

## Proposed changes:

* Created new component for the picker itself. It has a `onTemplateSelected` prop, which can be used by the parent component to commit the changes, or do something with them.
* Used this new component inside our old one, which was renamed to `TemplatePickerModal`.

### Other information:

## Jetpack product discussion

1204930070946424-as-1204986292229845

No.
## Testing instructions:

* Open a new post with the Advanced plan.
  * Check that the template picker works as before:
    * It should open up correctly, and select the first - default - template.
    * Changing a template and saving should be reflected on the preview.
    * Re-opening the picker should show the selected template.
    * Publishing the post with SIG should use the correct image.
 * On the admin page, the template selector and changing default templates should work too.


projects/js-packages/publicize-components/src/components/social-image-generator/template-picker-modal/styles.module.scss
```diff
@@ -0,0 +1,8 @@
+.footer {
+	display: flex;
+	justify-content: flex-end;
+
+	> * + * {
```
pablinos, line 5: I know this is copied over from the other CSS file, but I personally don't like magic selectors like this. Could we use
```suggestion
	> :not(:first-child) {
```

to make it a bit clearer?

projects/js-packages/publicize-components/src/components/social-image-generator/template-picker/index.js
```diff
@@ -1,80 +1,58 @@
...
-	}, [ onSelect, setIsOpen, selectedTemplate ] );
-
 	// Add memoized callback function for each template.
 	const TEMPLATES = useMemo(
 		() =>
 			TEMPLATES_DATA.map( template => ( {
 				...template,
-				onSelect: () => setSelectedTemplate( template.name ),
+				onSelect: () => {
+					setSelectedTemplate( template.name );
```
pablinos: Do we need this internal state anymore? The modal also has it, and I guess the new SIG settings modal would have it too.

projects/js-packages/publicize-components/src/components/social-image-generator/template-picker/index.js
```diff
@@ -1,80 +1,58 @@
...
+					if ( onTemplateSelected ) {
```
pablinos: I wonder if there is a more succinct way to do this, especially if we remove the internal state.

Perhaps we could do something like:

```js
onSelect: () => onSelectTemplate?.( template.name )
```

I don't know if that's better or not!

projects/js-packages/publicize-components/src/components/social-image-generator/template-picker/index.js
```diff
@@ -1,80 +1,58 @@
...
-							{ __( 'Save', 'jetpack' ) }
-						</Button>
-					</div>
-				</Modal>
-			) }
-		</ThemeProvider>
+		<div className={ styles.templates }>
+			{ TEMPLATES.map( template => (
+				<button
+					onClick={ template.onSelect }
```
pablinos: We might even be able to define the callback here then, so we don't have to set `onSelect` on each of the templates first.

pablinos (commented):
This is working well for me. I've added a few comments to the code, and I think there are some versions that need updating after the weekly release, but it's pretty close.

projects/js-packages/publicize-components/src/components/social-image-generator/template-picker-modal/styles.module.scss
gmjuhasz: Yeah sounds good. When we have the time we could clear up all these across social

projects/js-packages/publicize-components/src/components/social-image-generator/template-picker/index.js
gmjuhasz: Yeah probably we can get rid of it. It makes the TemplatePicker completely stateless though, so it always needs a parent that handles the state changes and rerenders it

projects/js-packages/publicize-components/src/components/social-image-generator/template-picker/index.js
```diff
@@ -1,7 +1,6 @@
-import { useState, useMemo } from '@wordpress/element';
 import { sprintf, __ } from '@wordpress/i18n';
 import classnames from 'classnames';
-import React from 'react';
+import React, { useCallback } from 'react';
```
pablinos: I think this should come from `@wordpress/element`. I've seen problems where we end up importing multiple versions of React when we don't use the wrapped version. 

Similarly, we probably shouldn't importing `React` here. Is that for the JSDoc?

projects/js-packages/publicize-components/src/components/social-image-generator/template-picker/index.js
```diff
@@ -16,31 +15,23 @@ import TEMPLATES_DATA from './templates.js';
...
-					setSelectedTemplate( template.name );
-					if ( onTemplateSelected ) {
-						onTemplateSelected( template.name );
-					}
-				},
-			} ) ),
-		[ setSelectedTemplate, onTemplateSelected ]
+	const onTemplateClicked = useCallback(
+		event => {
+			const templateName = event.target.id;
```
pablinos, line 22: Are we using a const to make it explicit that we're looking for the template's name?

projects/js-packages/publicize-components/src/components/social-image-generator/template-picker-modal/styles.module.scss
pablinos: Oh do we use this a lot? I think maybe this is a question of personal choice. The con to this is that it's a fair bit longer. Do you have a preference?

projects/js-packages/publicize-components/src/components/social-image-generator/template-picker/index.js
gmjuhasz: Yes for the JSDoc type

projects/js-packages/publicize-components/src/components/social-image-generator/template-picker/index.js
gmjuhasz: I think it makes it more readable, than just `onTemplateSelected?.( event.target.id );`. I could change it to an inline version too if you prefer that

projects/js-packages/publicize-components/src/components/social-image-generator/template-picker/index.js
pablinos: Seems like this should be solved in the `@wordpress/element` package, but [I notice this is handled elsewhere using a `@typedef`](https://github.com/WordPress/gutenberg/blob/e7b7f01b272c955ca51efb2f602a7ea16406f084/packages/block-editor/src/components/block-context/index.js#L6) although that is still importing from 'react'. What do you think doing it this way?

projects/js-packages/publicize-components/src/components/social-image-generator/template-picker/index.js
pablinos: I don't have strong feelings really. I was more curious. It's fine to leave it as is.

pablinos:
Just a thought about the ReactNode definition and we should probably get the E2E tests passing (or at least some of them!), but otherwise I think this is good.

pablinos (approved):
I think this is good now. I like that the component is now much simpler.

PR #31761, opened 2023-07-07
Protect: Add standalone brute force protection support

dkmyta:
#### Description:
* Adds access to the Firewall page for all environments.
* On environments where the request firewall is not supported (Atomic, WPCOM), updates the Firewall page to only render the brute force protection settings.

## Proposed changes:

* [ Fix ] Prevent module activation/deactivation attempts if module already in the state requested (i.e. don't try and deactivate the `waf` module if it isn't currently active).
* [ Fix ] Add compatibility and post-upgrade hooks during WAF initialization regardless of environment support (i.e. still migrate the brute force protection options).
* [ Enhancement ] Allow all users to access the Firewall page, but only show brute force protection settings for users on environments that do not support the WAF.
   
### Other information:

## Jetpack product discussion

- 1201069996155224-as-1204967124526888

- No

## Testing instructions:

* Run this branch via Jurassic Ninja or Jurassic Tube
* Validate no regression on the Firewall screen for supported environments
* Add `define( 'DISABLE_JETPACK_WAF', true )` to your wp-config
* Validate only brute force protection settings are available on the Firewall screen.

## Screenshots:

Standalone Active:

Screen Shot 2023-07-14 at 11 22 44-fullpage

Standalone Inactive:

Screen Shot 2023-07-14 at 11 22 56-fullpage


nateweller:
Taking this one over from @dkmyta as he sets off for parental leave!

Brute force protection should be available in all environments, it is just the firewall/request blocking component of the WAF package that is redundant on Atomic sites (such as those hosted on Pressable).

Therefore, the `is_brute_force_supported` can either be removed or always return `true`. We just need to ensure the now-more-aptly-named `is_waf_supported` blocks don't include anything required for brute-only sites.

Updates to follow :+1:

bhoop (dismissed):
Followed the test instructions and tried to break things as many ways as I could but couldn't 👍

nateweller (approved):

PR #31708, opened 2023-07-05
AI Assistant Gutenberg experiment

dhasilva:
Fixes #

## Proposed changes:

*

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* Add an AI Assistant block
* Ask for a table, a list or any combination, e.g. "generate a currency conversion table between BRL, ARS and SOL, followed by a list with the currency names"
* Only works with tables and lists!

PR #31766, opened 2023-07-08
AI Assistant: Add Gutenberg syntax parsing

dhasilva:
Extracted from #31708

## Proposed changes:

* Adds the Gutenberg parsing

### Other information:

## Testing instructions:

* Add an AI Assistant block
* Enable Gutenberg syntax
* Send a request
* Check that the blocks are parsed and inserted as children blocks


retrofox (approved):
Tested in Simple, Atomic, and Jetpack sites and locally dev-env.
It works as expected.
There are several things to polish but considering that the feature is hidden behind a global const, it's save to take off.
🚀

PR #31767, opened 2023-07-09
AI Assistant: Only append flag if there's no error

renatoagds:
Fixes #

## Proposed changes:

* Move ai playground flag to be inserted only if there's no error.

### Other information:

## Testing instructions:

* Insert block
* Force some error
* Check if doesn't generate some fatal if had some error.


oskosk (approved):
LGTM

PR #31780, opened 2023-07-10
Release Script: Check for GH CLI Login

sdixon194:
## Proposed changes:

If we're not logged into the GH CLI, the script will fail when it checks the status of the build. This checks to make sure you're logged in and prompts you to do so before continuing. 

### Other information:

## Jetpack product discussion

n/a

no
## Testing instructions:

If you're logged in, you can use "gh auth logout" and then run tools/release-plugin.sh jetpack to make sure it asks you to log in. Throw an exit statement on line 51 if you want to test further without running through the whole script.

### DIFF
--- a/tools/release-plugin.sh
+++ b/tools/release-plugin.sh
@@ -49,6 +49,7 @@ if ! gh auth status &> /dev/null; then
 	gh auth login
 fi
 
+# Get the options passed and parse them.
 while getopts "h" opt; do
 	case ${opt} in
 		h)
--- a/tools/release-plugin.sh
+++ b/tools/release-plugin.sh
@@ -42,7 +42,13 @@ if ! version_compare "$GH_VER" "2.21.2"; then
 	die "Your version of the GH CLI is out of date. Please upgrade your version$WITH and start again"
 fi
 
-# Get the options passed and parse them.
+# Make sure we're signed into the GitHub CLI.
+if ! gh auth status &> /dev/null; then
+	yellow "You are not signed into the GitHub CLI."
+	proceed_p "Sign in to the GitHub CLI?"
+	gh auth login
+fi
+
 while getopts "h" opt; do
 	case ${opt} in
 		h)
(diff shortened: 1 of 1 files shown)


tools/release-plugin.sh
```diff
@@ -42,7 +42,13 @@ if ! version_compare "$GH_VER" "2.21.2"; then
 	die "Your version of the GH CLI is out of date. Please upgrade your version$WITH and start again"
 fi
 
-# Get the options passed and parse them.
```
anomiex, line 45: Why delete this comment though?

zinigor (dismissed):

tools/release-plugin.sh
```diff
@@ -42,7 +42,13 @@ if ! version_compare "$GH_VER" "2.21.2"; then
...
+# Make sure we're signed into the GitHub CLI.
+if ! gh auth status &> /dev/null; then
+	yellow "You are not signed into the GitHub CLI."
+	proceed_p "Sign in to the GitHub CLI?"
+	gh auth login
+fi
+
 while getopts "h" opt; do
```
zinigor, line 53:
```suggestion
# Get the options passed and parse them.
while getopts "h" opt; do
```

tools/release-plugin.sh
zinigor: Restored the comment, thank you!

tools/release-plugin.sh
```diff
@@ -42,6 +42,13 @@ if ! version_compare "$GH_VER" "2.21.2"; then
 	die "Your version of the GH CLI is out of date. Please upgrade your version$WITH and start again"
 fi
 
+# Make sure we're signed into the GitHub CLI.
+if ! gh auth status &> /dev/null; then
```
anomiex: I wonder if we should do like this
```suggestion
if ! gh auth status --hostname github.com &> /dev/null; then
```
since we only care about GH itself here and not any enterprise things.

anomiex (dismissed):
Makes sense to me. One suggestion inline.

sdixon194:
Applied the suggestion, just needs another ✅

anomiex (approved):

PR #31781, opened 2023-07-10
crm: Remove composer `.config.platform.php` setting

anomiex:
## Proposed changes:

This configuration setting is a blunt instrument, causing composer to install dependencies for a specific version of PHP regardless of what is actually being run. It looks like it was added at some point pre-monorepo when they were bundling a bunch of oauth libs, before (still pre-monorepo) they switched to downloading them at runtime. As things currently stand, this doesn't seem to make any difference to the composer packages that do get bundled.

On the other hand, it does cause some issues in our CI: versions of symfony and codeception compatible with 7.2 are brought in, which raise deprecation warnings with 8.1 and 8.2.

Since it doesn't seem to be needed anymore, let's drop it. We already have some checks in CI that will flag if the non-dev deps accidentally drop support for any version of PHP that we run tests with. In fact, we also need to update those checks to not run if the plugin _doesn't_ support the PHP version.

### Other information:

## Jetpack product discussion

None

No

## Testing instructions:

* Is CI happy? Did tests actually run for 7.2+?
  * [x] 5.6: [Skipped setup](https://github.com/Automattic/jetpack/actions/runs/5509116754/jobs/10041386104#step:8:489), [skipped test](https://github.com/Automattic/jetpack/actions/runs/5509116754/jobs/10041386104#step:9:5847)
  * [x] 7.2: [Setup worked](https://github.com/Automattic/jetpack/actions/runs/5509116754/jobs/10041387016#step:8:452), [tests passed](https://github.com/Automattic/jetpack/actions/runs/5509116754/jobs/10041387016#step:9:6260).
  * [x] 8.2: [Setup worked](https://github.com/Automattic/jetpack/actions/runs/5509116754/jobs/10041388525#step:8:342), [tests passed](https://github.com/Automattic/jetpack/actions/runs/5509116754/jobs/10041388525#step:9:6686).
* Verify no change to build artifact.


anomiex:
Heh, I thought [that](https://github.com/Automattic/jetpack/actions/runs/5508744089/jobs/10040453470?pr=31781#step:8:520) might happen. The mentioned check for non-dev deps dropping support triggered for PHP 5.6 and 7.0, we need to not run it if the `skip-test-php` script says to skip. 😀

zinigor (approved):

PR #31792, opened 2023-07-10
[Boost] Fix handling of broken background-image urls in the Image Guide

### LINKED ISSUE #31788: Boost: Image guide is broken in trunk
### Impacted plugin

Boost

### Quick summary

Image guide is not functioning in trunk when Jetpack is also installed.

### Steps to reproduce

1. Install Boost trunk and Jetpack (e.g. using JN). Activate both with free plans.
2. Enable the image guide feature.
3. Create a post using an image block with a Pexels image. Publish it.
4. Visit the post.

Image guide should work.

### What actually happened

Image guide does not show any UI on the image.

In the browser console, I see
```
Uncaught (in promise) TypeError: t is null
```
with reference to what appears to be code generated from https://github.com/Automattic/jetpack/blob/9d4b8438027a44f34be36c0b49849f8737f479ce/projects/plugins/boost/app/modules/image-guide/src/index.ts#L134

### Impact

All

### Available workarounds?

Self-hosted

### Logs or notes

Bisect points to https://github.com/Automattic/jetpack/pull/31504 as the first commit with the problem. I note that commit dropped a `.filter( image => image !== null )`. After some subsequent refactors, that should probably be restored to this function: https://github.com/Automattic/jetpack/blob/9d4b8438027a44f34be36c0b49849f8737f479ce/projects/js-packages/image-guide/src/find-image-elements.ts#L58-L90
The doc block for that function claims it returns `MeasurableImage[]`, but as written some elements of the array can be null.

thingalon:
Fixes #31788.

It looks like a filter to remove `null` MeasurableImage objects was removed. That is needed to remove images that were rejected either for being SVGs (ignored by the Image Guide), or for having malformed URLs in a background-image CSS style.

This PR reinstates the filter, preventing bad background-images from breaking the Image Guide.

## Proposed changes:
* Filter out null ImageMeasurements from the Image Guide.

### Other information:

## Jetpack product discussion
n/a

no

## Testing instructions:

* Rig your site to render a bad or ignored background-image URL via custom CSS. e.g.: `h1 { background-image: url(/.svg); }` ought to do it.
* Make sure the Image Guide still functions.

### DIFF
--- a/projects/js-packages/image-guide/changelog/fix-missing-null-filter
+++ b/projects/js-packages/image-guide/changelog/fix-missing-null-filter
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Image Guide: Fix for broken background-images causing the image guide not to load
--- a/projects/js-packages/image-guide/package.json
+++ b/projects/js-packages/image-guide/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "@automattic/jetpack-image-guide",
-	"version": "0.5.0",
+	"version": "0.5.1-alpha",
 	"description": "Go through the dom to analyze image size on screen vs actual file size.",
 	"homepage": "https://github.com/Automattic/jetpack/tree/HEAD/projects/js-packages/image-guide/#readme",
 	"type": "module",
--- a/projects/js-packages/image-guide/src/find-image-elements.ts
+++ b/projects/js-packages/image-guide/src/find-image-elements.ts
@@ -86,7 +86,7 @@ export async function getMeasurableImages(
 		return null;
 	} );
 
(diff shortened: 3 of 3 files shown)


haqadn (approved):
The code changes look good. However, I did not see any changes in behavior.

Is the issue a total breakage of image-guide? In that case, I couldn't reproduce the issue.

Is the issue the element still detecting an image when it's null? The issue may still exist. 


With the patch
image

From rc:
image

thingalon:
cherry-picked to 2.0.0 in 7b184df5f70ed6b212a01a5a6d9261e8c74de5c6

PR #31795, opened 2023-07-11
[Donation Block] Default to the Stripe Currency on Mount and Display a  ✔️ next to the selected currency

romarioraffington:
Fixes https://github.com/Automattic/gold/issues/39

## Proposed changes:

*  Display a  ✔️ next to the selected currency
* Default to the Stripe currency on mount. 

### Other information:

No

## Testing instructions:

* Sync changes to WPCOM
* Change the Stripe account Currency connected to the site to any currency besides USD or you can change the default value in `donations.php` for the `currency` attribute to another currency
* Add the Donation block to a post
* Notice the Stripe default currency is selected. In most cases, this will be USD
* Select a currency and notice that  the ✔️ icon is to the right of the currency
* Save the post and notice no errors are thrown
* Change the currency to another currency and save the post and notice no errors are thrown
* Refresh the page and notice that the user-selected currency isn't overridden by the Stripe currency


millerf:
@romarioraffington , as explained in the [task description](https://github.com/Automattic/gold/issues/39), there is already a way [to retrieve the current Stripe currency from the data store](https://github.com/Automattic/wp-calypso/blob/7a62b7d389734da64256173b6c8f6183ef1d3ee8/client/state/memberships/settings/selectors.js#L13).
I might be mistaken, but you should be able to use it without having to reinvent the wheel with `getConnectedAccountDefaultCurrency`. That is adding a lot of complexity...

Please test with `getconnectedAccountDefaultCurrencyForSiteId()` and remove the unecessary code...

romarioraffington:
> @romarioraffington , as explained in the [task description](https://github.com/Automattic/gold/issues/39), there is already a way [to retrieve the current Stripe currency from the data store](https://github.com/Automattic/wp-calypso/blob/7a62b7d389734da64256173b6c8f6183ef1d3ee8/client/state/memberships/settings/selectors.js#L13). I might be mistaken, but you should be able to use it without having to reinvent the wheel with `getConnectedAccountDefaultCurrency`. That is adding a lot of complexity...
> 
> Please test with `getconnectedAccountDefaultCurrencyForSiteId()` and remove the unecessary code...

The code you linked is in the Calypso repo. There isn't a way to do this in Jetpack as the moment 😄

millerf:
> The code you linked is in the Calypso repo. There isn't a way to do this in Jetpack as the moment 😄

Yes 😅
At least it gives some ideas on where to get the data.

millerf (changes requested):
I tried to test it on https://bestmemberships.wordpress.com/ and it fails. I have a console error when adding a Donation block...

I am not sure if some changes were not uploaded...

https://github.com/Automattic/jetpack/assets/790558/aa6fa955-2147-4d93-b5f1-08c4e25b8d5b

projects/plugins/jetpack/extensions/blocks/donations/edit.js
```diff
@@ -10,18 +10,36 @@ import Tabs from './tabs';
...
+	const stripeCurrency = useSelect( select => {
+		const defaultStripeCurrency = select(
+			MEMBERSHIPS_PRODUCTS_STORE
+		).getConnectedAccountDefaultCurrency();
+
+		return defaultStripeCurrency ? defaultStripeCurrency : currency;
+	} );
+
+	useEffect( () => {
+		if ( ! hasCurrencyBeenUpdatedByUser ) {
```
millerf: It feels like `hasCurrencyBeenUpdatedByUser` could be deduced by "the current currency is  not null and the current curtrency is different from the stripe currency"

```
hasCurrencyBeenUpdatedByUser = currency!=null && currency!=stripeCurrency

```

So it might not be needed to save it as an extra attribute... What do you think?

millerf (changes requested):
It looks good appart from the extra attributes that seems unecessary.
I haven't tested yet.

projects/plugins/jetpack/extensions/blocks/donations/edit.js
romarioraffington: When the currency was `null` the Donation block threw an error similar to the one in the video above. I could try again, however.

romarioraffington: See the error that is thrown based on the feedback above 

Screenshot 2023-07-20 at 1 21 06 PM

millerf: > When the currency was null the Donation block threw an error similar to the one in the video above.

Well I am just saying to change the current version to 
```
	if ( ! currency!=null && currency!=stripeCurrency ) {
```

And remove `hasCurrencyBeenUpdatedByUser`

millerf: Does it make more sense?

romarioraffington: > > When the currency was null the Donation block threw an error similar to the one in the video above.
> 
> Well I am not saying to change the current version to
> 
> ```
> 	if ( ! currency!=null && currency!=stripeCurrency ) {
> ```
> 
> And remove `hasCurrencyBeenUpdatedByUser`

Okay, if you're not saying to do the above, what are you suggesting? I've tried the above and it causes the same error.

millerf: > Okay, if you're not saying to do the above, what are you suggesting?

There was a typo. Fixed now.

millerf: See https://github.com/Automattic/jetpack/pull/31795/commits/fd1eef7d70b1ce82d1c16a1d8802124542056d45 and further commits.

millerf (dismissed):

millerf (dismissed):
LGTM

millerf (dismissed):

millerf (approved):

PR #31801, opened 2023-07-11
[Boost] Fix potential savings display

### LINKED ISSUE #31746: Boost: ISA - some images show incorrect `potential size` suggestions
### Impacted plugin

Boost

### Quick summary

CleanShot 2023-07-07 at 17 34 56@2x

As you can see from the image, the potential size optimization doesn't make sense: `5 KB` -> `5 KB`. The image against which the size is checked, is not the original file size - it's one of the URLs from the srcset (the full size image is 1707x2560).

The code for the image:

```html
<figure class="wp-block-image size-full is-resized"><img src="http://jetpack-boost.test/wp-content/uploads/2023/06/fernando-paredes-murillo-Xc6gtOwSMSA-unsplash-scaled.jpg" alt="" class="wp-image-931" width="38" height="57"/></figure>
```

The image was added to the page as a full size image. Then resized to match the ratio to a size, just under the [65 by 65](https://github.com/Automattic/jetpack/blob/trunk/projects/js-packages/image-guide/src/MeasurableImage.ts#L133) threshold we have for tiny images.

CleanShot 2023-07-07 at 17 35 51@2x

Here's the original image:
`https://github.com/Automattic/jetpack/assets/11799079/974ded44-e58f-4249-8605-0866a8ec9696`

### Steps to reproduce

1. Add the above image to a page (full size);
2. Then resize it to dimensions be low 65x65;
3. Run an Image Size Analysis;
4. Check to see if the image is in the list of issues.

Image should be there, but potential size suggestion should be correct.

### What actually happened

Potential size suggestion doesn't appear to be correct.

### Impact

One

### Available workarounds?

There is no user impact

_No response_

### Logs or notes

_No response_

thingalon:
Fixes #31746

## Proposed changes:
* Correctly use the `potential`(savings) field.

### Other information:

## Testing instructions:

* Run ISA report
* View the results
* Ensure the potential size differences seem right(ish)

### DIFF
--- a/projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/recommendations/row-types/ImageSizeRow.svelte
+++ b/projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/recommendations/row-types/ImageSizeRow.svelte
@@ -1,7 +1,6 @@
 <script lang="ts">
 	import { __ } from '@wordpress/i18n';
 	import Button from '../../../../elements/Button.svelte';
-	import { getImageSizeDifferencePercent } from '../../../../utils/get-image-size-difference-percent';
 	import Device from '../components/Device.svelte';
 	import Pill from '../components/Pill.svelte';
 	import RowTitle from '../components/RowTitle.svelte';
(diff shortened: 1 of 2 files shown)


donnchawp (approved):
It works as expected!

thingalon:
Cherry-picked to 2.0.0 in 7909e132ab

PR #31808, opened 2023-07-11
Social: Add reel validation for Instagram connections

gmjuhasz:
This change adds the validation for Instagram Reels, introducing a new logic for aspect ratio checks.

## Proposed changes:

* Added restrictions for instagram-business
* Added logic to check aspect ratio
* Updated messaging to mention `photo or video` on the Advanced plan.

### Other information:

## Testing instructions:

First test the messaging update.

* With both Classic editor and Gutenberg check the following:
  * With the Advanced plan on the site, the error with no media for instagram connections should mention videos 
  * Without the Advanced plan it should say only photos

Test the Instagram reel validation.

* Try uploading a video bigger than FHD, so wider than 1920px, it should result in an error.
* You can test the aspect ratio validation too by changing the values in `restrictions.js`
* Uploading videos that match the criteria should work.


spsiddarthan:
After adding a video to the sidebar, if I click on share as a post, I got this error. Are you getting it too? If not ping me on slack, please, and I will take another look. 


```
TypeError: Cannot read properties of undefined (reading 'min')
    at getVideoValidationError (webpack-internal:///../../js-packages/publicize-components/src/hooks/use-media-restrictions/index.js:72:27)
    at getValidationError (webpack-internal:///../../js-packages/publicize-components/src/hooks/use-media-restrictions/index.js:104:26)
    at eval (webpack-internal:///../../js-packages/publicize-components/src/hooks/use-media-restrictions/index.js:127:21)
    at Array.reduce (<anonymous>)
    at eval (webpack-internal:///../../js-packages/publicize-components/src/hooks/use-media-restrictions/index.js:122:79)
    at updateMemo (https://sidtest.jurassic.tube/wp-includes/js/dist/vendor/react-dom.js?ver=18.2.0:17256:21)
    at Object.useMemo (https://sidtest.jurassic.tube/wp-includes/js/dist/vendor/react-dom.js?ver=18.2.0:17896:18)
    at useMemo (https://sidtest.jurassic.tube/wp-includes/js/dist/vendor/react.js?ver=18.2.0:1640:23)
    at useMediaRestrictions (webpack-internal:///../../js-packages/publicize-components/src/hooks/use-media-restrictions/index.js:121:69)
    at PublicizeForm (webpack-internal:///../../js-packages/publicize-components/src/components/form/index.js:158:102)
    
    ```

**Review by spsiddarthan — DISMISSED** [review]
It tests well and the code looks good to me. 

In the case of the classic editor and the advanced plan, can we upload media? The option didn't show up for me which makes me think if we don't support it for the advanced plan in the classic editor, correct me if I am wrong there. If we don't support it, the messaging for the classic editor-advanced plan combination should perhaps not mention video.  It is currently "You need a valid image or video in your post to share to Instagram"

I am approving this and the code looks good, but tagging @pablinos for a second look as well in case I missed some front-end gotchas.

**Comment by pablinos** [comment]
> The option didn't show up for me which makes me think if we don't support it for the advanced plan in the classic editor, correct me if I am wrong there.

You're right, you can't attach media in the classic editor. Instagram support is just via the Featured image there. We could be that specific in the classic editor and say that the post must have a featured image to be shared to Instagram

File: projects/js-packages/publicize-components/src/components/form/index.js [hunk 1]
```diff
@@ -163,6 +163,26 @@ export default function PublicizeForm( {
...
+			<Notice type={ 'warning' }>
+				{ __( 'You need a valid image or video in your post to share to Instagram.', 'jetpack' ) }
+				<br />
+				<ExternalLink href={ getRedirectUrl( 'jetpack-social-media-support-information' ) }>
+					{ __( 'Learn more', 'jetpack' ) }
+				</ExternalLink>
+			</Notice>
+		) : (
+			<Notice type={ 'warning' }>
+				{ __( 'You need a valid image in your post to share to Instagram.', 'jetpack' ) }
```
pablinos: I know this isn't strictly related to this change but there's something about the 'in your post' phrasing that doesn't seem quite right to me. It's more 'attached' but that doesn't seem to cover both the featured image and the attached media.

Like I mentioned in my other comment, without the advanced plan we could be specific and say that you need a featured image to share to Instagram.

projects/js-packages/publicize-components/src/hooks/use-media-restrictions/index.js
```diff
@@ -1,11 +1,13 @@
 import { useRef, useMemo } from '@wordpress/element';
-import { DEFAULT_RESTRICTIONS, RESTRICTIONS } from './restrictions';
+import { DEFAULT_RESTRICTIONS, GLOBAL_MAX_SIZE, RESTRICTIONS } from './restrictions';
```
pablinos, line 2: I can't remember why but I think this was previously removed as it wasn't needed. It might have been that we set it on all the restrictions, in which case it probably does make sense to double check here.

projects/js-packages/publicize-components/src/hooks/use-media-restrictions/restrictions.js
```diff
@@ -117,10 +122,20 @@ export const RESTRICTIONS = {
 	},
 	[ 'instagram-business' ]: {
 		requiresMedia: true,
-		allowedMediaTypes: [ 'image/jpg', 'image/jpeg' ],
+		allowedMediaTypes: [ 'image/jpg', 'image/jpeg' ].concat( [ MP4, MOV, VIDEOPRESS ] ),
```
pablinos: We can probably concot this ourselves. 
```suggestion
		allowedMediaTypes: [ 'image/jpg', 'image/jpeg', MP4, MOV, VIDEOPRESS ],
```

projects/js-packages/publicize-components/src/hooks/use-media-restrictions/index.js
```diff
@@ -31,13 +33,22 @@ const getImageValidationError = ( sizeInMb, maxImageSize ) =>
...
+	const {
+		minSize = 0,
+		maxSize = GLOBAL_MAX_SIZE,
+		minLength = 0,
+		maxLength = GLOBAL_MAX_SIZE,
+		maxWidth = GLOBAL_MAX_SIZE,
+		aspectRatio = DEFAULT_RESTRICTIONS.video.aspectRatio,
+	} = videoLimits;
+
+	if ( ! sizeInMb || sizeInMb > maxSize || sizeInMb < minSize || width > maxWidth ) {
```
pablinos: As discussed, maybe the width restriction should be grouped with the aspect ratio, and be a DIMENSIONS_ERROR or something

pablinos (commented):
This looks good to me, just a few minor comments.

I also notice that we don't check the mimetype of the featured image in the Classic Editor. I think we discovered that was tricky previously, and probably isn't for this change anyway!

projects/js-packages/publicize-components/src/components/form/index.js
gmjuhasz: I changed it for the scenario when they do not have the advanced plan.

>I know this isn't strictly related to this change but there's something about the 'in your post' phrasing that doesn't seem quite right to me. It's more 'attached' but that doesn't seem to cover both the featured image and the attached media.

This is tricky, becuase they need either the featured image, attached image/video, or SIG turned on. I don't know if I can come up with a better and not longer explanation than the current one

projects/js-packages/publicize-components/src/hooks/use-media-restrictions/index.js
gmjuhasz: Yeah, I put it back, because the with the inner objects in restrictions, the default restrictions does not always do the job. Using default values when extracting from `videoLimits` makes sure it's enough

projects/packages/publicize/src/js/classic-editor-connections.js
```diff
@@ -37,11 +37,11 @@ const validateFeaturedMedia = ( $, connectionsNeedValidation ) => {
 
 	const connectionNeedsMediaString = isEnhancedPublishingEnabled
 		? /* translators: %s is the link to the media upload best practices. */ __(
-				'You need a valid image or video in your post to share to Instagram. <a href="%s" rel="noopener noreferrer" target="_blank">Learn more</a>.',
+				'You need a featured image to share to Instagram - or use Gutenberg editor for more advanced media features! <a href="%s" rel="noopener noreferrer" target="_blank">Learn more</a>.',
```
spsiddarthan: Is the '-' there on purpose? 

Should we say "You need a featured image to share to Instagram. Use the block editor for more advanced media features"

spsiddarthan (dismissed):

projects/packages/publicize/src/js/classic-editor-connections.js
gmjuhasz: It was on purpose, I was not sure what's the best way to say this, I like your version too.

pablinos: We should refer to it as 'the block editor' rather than Gutenberg.

projects/js-packages/publicize-components/src/hooks/use-media-restrictions/index.js
pablinos: Fair enough. Why is that? Is it because we haven't added the defaults to each of the services?

projects/js-packages/publicize-components/src/components/form/index.js
pablinos: Yes, agreed, it's hard to make it succinct. What about something like:

"To share to Instagram, pick an image/video below, a featured image, or enable Social Image Generator"

or maybe just:

"To share to Instagram, add an image/video, or enable Social Image Generator"

Is there a way we could let them know in a tooltip that an image can be added in the media picker or as a featured image? In which case we could maybe make it:

"To share to Instagram, add an image or video to your post"

And lean on the fact that enabling Social Image Generator adds an image to the post.

I realise that 'add' is probably similarly confusing because it sounds like you can put an image in the post content, so we maybe need to be more explicit. but 'in' sounds like it's only in the post content to me.

We can probably go with something imperfect for now and adjust it later!

projects/js-packages/publicize-components/src/hooks/use-media-restrictions/index.js
gmjuhasz: It's because we use the Defaults, when there is nothing for a service, but if we have restrictions for let's say Facebook, which have no aspect ratio limitations, that won't get in from the Defaults. We could write a function that deep merges by keys too, but I think this should do it for now.

projects/packages/publicize/src/js/classic-editor-connections.js
gmjuhasz: Updated!

projects/js-packages/publicize-components/src/components/form/index.js
gmjuhasz: > "To share to Instagram, add an image/video, or enable Social Image Generator"

I think this should do it, for anything more we should discuss it further in our call, or bring in design

spsiddarthan (dismissed):
LGTM!

pablinos (approved):
Still looks good!

PR #31809, opened 2023-07-11
Generate query string for the JITM request

gcsecsey:
Fixes https://github.com/Automattic/dotcom-forge/issues/2891

## Proposed changes:

* generate a query string for the `query_string` parameter

### Other information:

## Testing instructions:

* Apply this version on a WoA site
* Go to `/plugins`
* Search for a plugin where JITMs should be shown, eg. `woocommerce`, `apple pay`, or `elementor`
* Check that JITMs are shown on top of the search results


projects/packages/jitm/src/class-rest-api-endpoints.php
```diff
@@ -55,7 +55,13 @@ public static function get_jitm_message( $request ) {
 			return array();
 		}
 
-		return $jitm->get_messages( $request['message_path'], urldecode_deep( $request['query'] ), 'true' === $request['full_jp_logo_exists'] ? true : false );
+		// add the search term to the query params if it exists
+		$query = $request['query'];
+		if ( ! empty( $request['s'] ) ) {
```
epeicher, line 60: What happens if `$query['s']` exists? Do you think we should add an additional condition here like `&& empty( $query['s'] )`?

epeicher (approved):
I have tested it on a WoA site, and it works as advertised 🎉 . The changes make sense to me but I will leave someone from Jetpack team to review them.

CleanShot 2023-07-13 at 12 45 59@2x

Additionally, I have left a non-blocking question

projects/packages/jitm/src/class-rest-api-endpoints.php
gcsecsey: That's a good point, and we could add the additional condition. However, I think it's not likely that the same request will contain `s` as both a parameter and as part of the query parameter, eg. `?s=woo&query[s]=elementor`.

As far as I could see in other uses, sending everything within `query` is the usual method:
https://github.com/Automattic/jetpack/blob/b0065a944af9f9d5df076f886d3f7ab8374c0541/projects/packages/jitm/src/js/jetpack-jitm.js#L261-L265

It seems that only WPCOM passes the query params as they are when proxying the request [here](fbhepr%2Skers%2Sjcpbz%2Schoyvp.ncv%2Serfg%2Sjcpbz%2Qwfba%2Qraqcbvagf%2Spynff.jcpbz%2Qwfba%2Qncv%2Qwrgcnpx%2Qoybt%2Qraqcbvag.cuc%3Se%3Q8737954p%23831-og).

epeicher: > it's not likely that the same request will contain s as both a parameter and as part of the query parameter

Fair enough, I guess we don't need to cover an unlikely use case and synchronizing `query` and the `s` parameter seems to be correct in all cases

samiff (approved):
Tested well on my end 👍

PR #31822, opened 2023-07-11
AI Assistant: do not pre-populate inner blocks if Gutenberg syntax is not enabled

retrofox:
Fixes #

## Proposed changes:

* AI Assistant: do not pre-populate inner blocks if Gutenberg syntax is not enabled

### Other information:

## Testing instructions:

* Go to the block editor
* Create a Paragraph block
* Transform the block to an AI Assistant block instance

image

### BEFORE
* Confirm it creates a classic block instance

image

### AFTER
* It transforms the block into an AI Assistant instance

**Disclaimer**
If you try to use Gutenberg Syntax to an AI Assistant block containing blocks defined by Markdown syntax, it will trigger the same issue. Let's address something in a follow-up since Gutenberg syntax is beta.

Screenshot 2023-07-11 at 15 51 19

### DIFF
--- a/projects/plugins/jetpack/changelog/update-fix-issue-populating-inner-blocks-from-content
+++ b/projects/plugins/jetpack/changelog/update-fix-issue-populating-inner-blocks-from-content
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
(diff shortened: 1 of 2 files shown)


dhasilva (approved):
LGTM, it does fix the issue

PR #31832, opened 2023-07-12
Social: Update the screenshots to reflect the current UI

pablinos:
This change updates the screenshot assets that are used on the .org plugin listing page. The UI has changed since they were first created so this brings them in line with the current state.

### Other information:

No.

## Testing instructions:
Just check that the new screenshots are replacing the equivalent one


gmjuhasz (approved):
Looks good to me!

PR #31834, opened 2023-07-12
Launchpad: Refactored task list validation

### LINKED ISSUE #77939: Launchpad: Improve the task list validation function
Make the task list validation function return `null | WP_Error[]` so all the issues can be identified.

More context: https://github.com/Automattic/jetpack/pull/31055#discussion_r1221168911

markbiek:
Fixes [77939](https://github.com/Automattic/wp-calypso/issues/77939)

## Proposed changes:

Updated launchpad task list validation return types to be more useful. Previously, we just returned a simple bool. Now we return a WP_Error where appropriate.

### Other information:

## Testing instructions:

* `cd` into the `jetpack-mu-wpcom` directory.
* Run the feature tests (you may need to do `composer install` first).
```
./vendor/bin/phpunit -c ./phpunit.xml.dist --testsuite=features
```
* Tests should pass.

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
@@ -452,12 +452,6 @@ class Launchpad_Task_Lists {
 			$error_messages[] = $msg;
 		}
 
-		if ( isset( $task_list['visible_tasks_callback'] ) && ! is_callable( $task_list['visible_tasks_callback'] ) ) {
-			$msg = 'The visible_tasks_callback attribute must be callable';
-			_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
-			$error_messages[] = $msg;
-		}
-
 		if ( isset( $task_list['require_last_task_completion'] ) && ! is_bool( $task_list['require_last_task_completion'] ) ) {
 			$msg = 'The require_last_task_completion attribute must be a boolean';
 			_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
@@ -57,7 +57,7 @@ class Launchpad_Task_Lists {
 	 * @return bool True if successfully registered, false if not.
 	 */
 	public function register_task_list( $task_list = array() ) {
-		if ( ! $this->validate_task_list( $task_list ) ) {
+		if ( self::validate_task_list( $task_list ) !== null ) {
 			return false;
 		}
 
@@ -407,45 +407,53 @@ class Launchpad_Task_Lists {
 	 *
 	 * @param Task_List $task_list Task List.
 	 *
-	 * @return bool True if valid, false if not.
+	 * @return null | WP_Error Null if valid, WP_Error if not.
 	 */
 	public static function validate_task_list( $task_list ) {
+		$error_code = 'validate_task_list';
+
 		if ( ! is_array( $task_list ) ) {
-			return false;
+			return new WP_Error( $error_code, 'Invalid task list' );
 		}
 
 		if ( ! isset( $task_list['id'] ) ) {
-			_doing_it_wrong( 'validate_task_list', 'The Launchpad task list being registered requires a "id" attribute', '6.1' );
-			return false;
+			$msg = 'The Launchpad task list being registered requires a "id" attribute';
+			_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
+			return new WP_Error( $error_code, $msg );
 		}
 
 		if ( ! isset( $task_list['task_ids'] ) ) {
-			_doing_it_wrong( 'validate_task_list', 'The Launchpad task list being registered requires a "task_ids" attribute', '6.1' );
-			return false;
+			$msg = 'The Launchpad task list being registered requires a "task_ids" attribute';
+			_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
+			return new WP_Error( $error_code, $msg );
 		}
 
 		if ( isset( $task_list['visible_tasks_callback'] ) && ! is_callable( $task_list['visible_tasks_callback'] ) ) {
-			_doing_it_wrong( 'validate_task_list', 'The visible_tasks_callback attribute must be callable', '6.1' );
-			return false;
+			$msg = 'The visible_tasks_callback attribute must be callable';
+			_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
+			return new WP_Error( $error_code, $msg );
 		}
 
 		if ( isset( $task_list['required_task_ids'] ) && ! is_array( $task_list['required_task_ids'] ) ) {
-			_doing_it_wrong( 'validate_task_list', 'The required_task_ids attribute must be an array', '6.1' );
-			return false;
+			$msg = 'The required_task_ids attribute must be an array';
+			_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
+			return new WP_Error( $error_code, $msg );
 		}
 
 		// If we have required tasks, make sure they all exist in the array of `task_ids`.
 		if ( isset( $task_list['required_task_ids'] ) && array_intersect( $task_list['required_task_ids'], $task_list['task_ids'] ) !== $task_list['required_task_ids'] ) {
-			_doing_it_wrong( 'validate_task_list', 'The required_task_ids must be a subset of the task_ids', '6.1' );
-			return false;
+			$msg = 'The required_task_ids must be a subset of the task_ids';
+			_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
+			return new WP_Error( $error_code, $msg );
 		}
 
 		if ( isset( $task_list['require_last_task_completion'] ) && ! is_bool( $task_list['require_last_task_completion'] ) ) {
-			_doing_it_wrong( 'validate_task_list', 'The require_last_task_completion attribute must be a boolean', '6.1' );
-			return false;
+			$msg = 'The require_last_task_completion attribute must be a boolean';
+			_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
+			return new WP_Error( $error_code, $msg );
 		}
 
-		return true;
+		return null;
 	}
 
 	/**
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
@@ -407,7 +407,7 @@ class Launchpad_Task_Lists {
 	 *
 	 * @param Task_List $task_list Task List.
 	 *
-	 * @return null | WP_Error Null if valid, WP_Error if not.
+	 * @return null|WP_Error Null if valid, WP_Error if not.
 	 */
 	public static function validate_task_list( $task_list ) {
 		$error_code = 'validate_task_list';
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
@@ -414,50 +414,54 @@ class Launchpad_Task_Lists {
 		$error_messages = array();
 
 		if ( ! is_array( $task_list ) ) {
-			// If we don't have an array, skip later validation.
-			return new WP_Error( $error_code, 'Invalid task list' );
+			// Ensure we have a valid task list array.
+			$msg = 'Invalid task list';
+			_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
+			return new WP_Error( $error_code, $msg );
 		}
 
 		if ( ! isset( $task_list['id'] ) ) {
+			// Ensure we have an id.
 			$msg = 'The Launchpad task list being registered requires a "id" attribute';
 			_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
 			$error_messages[] = $msg;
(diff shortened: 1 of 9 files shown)


projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -57,7 +57,7 @@ public static function get_instance() {
 	 * @return bool True if successfully registered, false if not.
 	 */
 	public function register_task_list( $task_list = array() ) {
-		if ( ! $this->validate_task_list( $task_list ) ) {
+		if ( self::validate_task_list( $task_list ) !== null ) {
```
daledupreez, line 60: Good catch.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -407,45 +407,53 @@ public function is_task_id_complete( $task_id ) {
 	 *
 	 * @param Task_List $task_list Task List.
 	 *
-	 * @return bool True if valid, false if not.
+	 * @return null | WP_Error Null if valid, WP_Error if not.
```
daledupreez: Nit: I believe this should be `null|WP_Error`.
```suggestion
	 * @return null|WP_Error Null if valid, WP_Error if not.
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -407,45 +407,53 @@ public function is_task_id_complete( $task_id ) {
...
 		if ( isset( $task_list['require_last_task_completion'] ) && ! is_bool( $task_list['require_last_task_completion'] ) ) {
-			_doing_it_wrong( 'validate_task_list', 'The require_last_task_completion attribute must be a boolean', '6.1' );
-			return false;
+			$msg = 'The require_last_task_completion attribute must be a boolean';
+			_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
+			return new WP_Error( $error_code, $msg );
 		}
 
-		return true;
+		return null;
```
daledupreez: Structural question: do we want to only return the first error, or all errors? My thought is that it would be more developer-friendly to return all known errors for simpler debugging when things go wrong. With this change and the existing code, a developer could end up tripping over multiple configuration/setup issues that each require a round trip test to find the next problem.

My thinking was that we could either use a single `WP_Error`, or an array of `WP_Error` instances, but that should make it easier to identify all errors. Something like the following would be more along those lines:
```suggestion
	 * @return null|WP_Error Null if valid, WP_Error if not.
	 */
	public static function validate_task_list( $task_list ) {
		$error_code     = 'validate_task_list';
		$error_messages = array();

		if ( ! is_array( $task_list ) ) {
			// If we don't have an array, skip later validation.
			return new WP_Error( $error_code, 'Invalid task list' );
		}

		if ( ! isset( $task_list['id'] ) ) {
			$msg = 'The Launchpad task list being registered requires a "id" attribute';
			_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
			$error_messages[] = $msg;
		}

		if ( ! isset( $task_list['task_ids'] ) ) {
			$msg = 'The Launchpad task list being registered requires a "task_ids" attribute';
			_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
			$error_messages[] = $msg;
		}

		if ( isset( $task_list['visible_tasks_callback'] ) && ! is_callable( $task_list['visible_tasks_callback'] ) ) {
			$msg = 'The visible_tasks_callback attribute must be callable';
			_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
			$error_messages[] = $msg;
		}

		if ( isset( $task_list['required_task_ids'] ) && ! is_array( $task_list['required_task_ids'] ) ) {
			$msg = 'The required_task_ids attribute must be an array';
			_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
			$error_messages[] = $msg;
		}

		// If we have required tasks, make sure they all exist in the array of `task_ids`.
		if ( isset( $task_list['required_task_ids'] ) && array_intersect( $task_list['required_task_ids'], $task_list['task_ids'] ) !== $task_list['required_task_ids'] ) {
			$msg = 'The required_task_ids must be a subset of the task_ids';
			_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
			$error_messages[] = $msg;
		}

		if ( isset( $task_list['require_last_task_completion'] ) && ! is_bool( $task_list['require_last_task_completion'] ) ) {
			$msg = 'The require_last_task_completion attribute must be a boolean';
			_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
			$error_messages[] = $msg;
		}

		if ( array() !== $error_messages ) {
			$wp_error = new WP_Error();
			$error_message = array_shift( $error_messages );
			while ( ! empty( $error_message ) ) {
				$wp_error->add( $error_code, $error_message );
			}
			
			return $wp_error;
		}

		return null;
```

markbiek: That makes a lot of sense to me! I didn't realize you could add multiple error messages/codes to a single instance of `WP_Error`.

markbiek: Worth noting though that this approach only works for half of the error conditions. Some of those checks require immediately returning or you get worse errors further down.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -407,45 +407,65 @@ public function is_task_id_complete( $task_id ) {
...
-			return false;
+			// If we don't have an array, skip later validation.
+			return new WP_Error( $error_code, 'Invalid task list' );
 		}
 
 		if ( ! isset( $task_list['id'] ) ) {
-			_doing_it_wrong( 'validate_task_list', 'The Launchpad task list being registered requires a "id" attribute', '6.1' );
-			return false;
+			$msg = 'The Launchpad task list being registered requires a "id" attribute';
+			_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
```
sixhours, line 419: Do we need to keep _doing_it_wrong since we're now returning a WP_Error? I admit, I read over the function docs and I'm not entirely sure what it signifies 😆

markbiek: Yes! We decided [here](https://github.com/Automattic/wp-calypso/issues/77939#issuecomment-1631136186) to keep logging it with `_doing_it_wrong` for now.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -407,45 +407,65 @@ public function is_task_id_complete( $task_id ) {
...
+			return new WP_Error( $error_code, $msg );
 		}
 
-		if ( isset( $task_list['visible_tasks_callback'] ) && ! is_callable( $task_list['visible_tasks_callback'] ) ) {
-			_doing_it_wrong( 'validate_task_list', 'The visible_tasks_callback attribute must be callable', '6.1' );
-			return false;
+		if ( isset( $task_list['required_task_ids'] ) && ! is_array( $task_list['required_task_ids'] ) ) {
+			$msg = 'The required_task_ids attribute must be an array';
+			_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
+			return new WP_Error( $error_code, $msg );
```
daledupreez: Instead of returning early for this check, might it make sense to handle this check and the later check for `required_task_ids` to be structured so we don't run the secondary validation when we don't have an array? I can see that if the value is set and not an array, the later code will throw an exception, but I'd prefer to update the code so we can perform as much validation as possible.

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-list-validation-test.php
```diff
@@ -91,11 +107,17 @@ public function provide_validate_task_list_test_cases() {
...
-	public function test_validate_task_list( $task_list, $expected_result ) {
+	public function test_validate_task_list( $task_list, $expected_result, $expected_is_wp_error ) {
 		$result = Launchpad_Task_Lists::validate_task_list( $task_list );
 
+		if ( $expected_is_wp_error ) {
+			$this->assertTrue( is_wp_error( $result ) );
+			return;
+		}
+
 		$this->assertSame( $expected_result, $result );
```
daledupreez: Should we not do this in the opposite order so we always check `$expected_result`?
```suggestion
		$this->assertSame( $expected_result, $result );

		if ( $expected_is_wp_error ) {
			$this->assertTrue( is_wp_error( $result ) );
		}
```

markbiek: The issue with that is `assertSame` will fail because `$expected_result` and `$result` are technically different `WP_Error` objects.

markbiek: _But_, I can probably switch from using `assertSame` to checking the actual error message.

daledupreez: Maybe we can write this as follows, where we can actually drop the `$expected_is_wp_error` argument:
```suggestion
		if ( is_wp_error( $expected_result ) ) {
			$this->assertTrue( is_wp_error( $result ) );
			$this->assertSame( $expected_result->get_error_code(), $result->get_error_code() );
			$this->assertSame( $expected_result->get_error_message(), $result->get_error_message() );
		} else {
			$this->assertSame( $expected_result, $result );
		}
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
markbiek: @daledupreez I'm not sure I'm understanding this, in particular what you mean by "secondary validation". Could you give me any more details?

daledupreez: Apologies for the lack of clarity! This is kind of what I was thinking - the basic idea is that we perform the `required_task_ids` checks in one code block where we know whether we have a valid array or not to work with.
```php
		if ( isset( $task_list['required_task_ids'] ) ) {
			// Ensure we have a valid array.
			if ( ! is_array( $task_list['required_task_ids'] ) ) {
				$msg = 'The required_task_ids attribute must be an array';
				_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
				$error_messages[] = $msg;
			// Ensure all required tasks actually exist in the task list - we need the value to be an array for this to work.
			} elseif ( array_intersect( $task_list['required_task_ids'], $task_list['task_ids'] ) !== $task_list['required_task_ids'] ) {
				$msg = 'The required_task_ids must be a subset of the task_ids';
				_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
				$error_messages[] = $msg;
			}
		}

		if ( isset( $task_list['visible_tasks_callback'] ) && ! is_callable( $task_list['visible_tasks_callback'] ) ) {
			$msg = 'The visible_tasks_callback attribute must be callable';
			_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
			$error_messages[] = $msg;
		}
```

markbiek: Ah, I see what you're going for! Thanks for clarifying!

markbiek:
@daledupreez I think this is ready for another review when you have some time.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -407,45 +407,68 @@ public function is_task_id_complete( $task_id ) {
...
 
 		if ( ! isset( $task_list['task_ids'] ) ) {
-			_doing_it_wrong( 'validate_task_list', 'The Launchpad task list being registered requires a "task_ids" attribute', '6.1' );
-			return false;
+			// Ensure we have task_ids.
+			$msg = 'The Launchpad task list being registered requires a "task_ids" attribute';
+			_doing_it_wrong( 'validate_task_list', esc_html( $msg ), '6.1' );
+			$error_messages[] = $msg;
+		} elseif ( isset( $task_list['required_task_ids'] ) ) {
+				// Ensure we have a valid array.
```
daledupreez, line 436: Nit: the whitespace looks overly indented.
```suggestion
			// Ensure we have a valid array.
```

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-list-validation-test.php
```diff
@@ -25,54 +32,54 @@ public function provide_validate_task_list_test_cases() {
...
+				new WP_Error( 'invalid-task-list', 'The Launchpad task list being registered requires a "task_ids" attribute' ),
 			),
 			'Invalid task list with invalid required_task_ids' => array(
 				array(
 					'id'                => 'task_list_1',
 					'task_ids'          => array( 'task_1', 'task_2' ),
 					'required_task_ids' => 'task_1',
 				),
-				false,
+				new WP_Error( 'invalid-task-list', 'The required_task_ids attribute must be an array' ),
```
daledupreez, line 82: It looks like we're missing a test case for the case where `required_task_ids` isn't a subset of `task_ids`. It's not blocking, but I think we should do that now.

daledupreez (approved):
This is looking good to me, though I do think we should add the unit test case I mentioned.

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-list-validation-test.php
markbiek: I've added the test in https://github.com/Automattic/jetpack/pull/31992

PR #31200, opened 2023-06-05
 Do not display newsletter plans in the Premium content and Recurring payment blocks

romarioraffington:
Fixes https://github.com/Automattic/jetpack/issues/31199

## Proposed changes:

 Do not display newsletter plans in the Premium content and Recurring payment blocks

### Before
||  |
|--------|--------|
| Recurring payments| Screenshot 2023-06-05 at 3 23 07 PM  |
| Premium Content| Screenshot 2023-06-05 at 3 22 51 PM|
| Earn plans|Screenshot 2023-05-30 at 2 40 27 PM|

### After

||  |
|--------|--------|
| Recurring payments| Screenshot 2023-06-05 at 3 21 14 PM |
| Premium Content| Screenshot 2023-06-05 at 3 19 59 PM |
| Earn plans|Screenshot 2023-05-30 at 2 40 27 PM|

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [x] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [x] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Does this pull request change what data or activity we track or use?

No

## Testing instructions:

* Sync changes to WPCOM
* Go to the Earn payment plans page. `https://wordpress.com/earn/payments-plans/{site_url}`
* On the Earn payment plans page, add a payment plan with newsletter enabled and one without newsletter enabled. If the "Email newly published posts to your customers." slider is enabled, the plan will be marked as a newsletter
* Add the Recurring payments and Premium Content block to the post
* Verify that only the non-newsletter plans are displayed in the block

PR #31835, opened 2023-07-12
[Earn] Fix potential regression when selecting plans

millerf:
Fix potential regression from
https://github.com/Automattic/jetpack/pull/31200

### DIFF
--- a/projects/plugins/jetpack/changelog/add-remove-regression
+++ b/projects/plugins/jetpack/changelog/add-remove-regression
@@ -1,4 +1,4 @@
 Significance: patch
 Type: bugfix
 
-Fix regression from https://github.com/Automattic/jetpack/pull/31200/files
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/changelog/add-remove-regression
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
+Fix regression from https://github.com/Automattic/jetpack/pull/31200/files
```
romarioraffington: I think we can make this more user friendly

romarioraffington (dismissed):

projects/plugins/jetpack/changelog/add-remove-regression
millerf: Done

romarioraffington (approved):

PR #31838, opened 2023-07-12
Social Links: Prevent their use in block themes.

pablinos:
We want to make sure that the customizer is not being used for block themes. This change updates the check before initialising the Social Links feature, to make sure that a block theme is not active.

This shouldn't be a problem as block themes are unlikely to declare supoort for Social Links, but in case another plugin does, this ensures the feature isn't enabled.

Related #31725

### Other information:

No

## Testing instructions:
* Create a Jetpack test site, enable Publicize, add a connection and add the following code in place that will get loaded:
```
add_action( 'after_setup_theme', function () {
	add_theme_support( 'social-links', array(
		'facebook', 'twitter', 'linkedin', 'tumblr',
	) );
} );

```
* Install a classic theme like twenty-nineteen
* Head to the Customizer and you should see a Connect menu option
* Activate a block theme such as twenty-twenty-three
* Without this patch the Connect menu will still be in the Customizer
* With this patch it won't be displayed

#### Before 
image

#### After (with a block theme)
image


oskosk (approved):
Tests as described, thank you!

PR #31839, opened 2023-07-12
CRM: Show the email sender log if it has notes

cleacos:
Fixes: https://github.com/Automattic/zero-bs-crm/issues/3176

## Proposed changes:

This PR adds the possibility to show only system logs if they have notes. For debugging and issue investigation is useful to be able to filter logs and show only logs with notes. Show only log with notes is the default behavior. The `getCronLogs` method is only used by Mail Campaign.

### Other information:

## Jetpack product discussion
N/A

N/A

## Testing instructions:
- You need the Mail Campaign extension installed and activated. And [with this PR/branch applied.](https://github.com/Automattic/zero-bs-crm-mailcampaigns/pull/335) to be able to test it well
- Run some mail campaigns if you don't have any done. If you have a testing site with an old mail campaign it'd be ideal, it will already have many logs, so you can check it better.
- [Check the testing instructions for the MC PR](https://github.com/Automattic/zero-bs-crm-mailcampaigns/pull/335)

### DIFF
--- a/projects/plugins/crm/admin/contact/view.page.php
+++ b/projects/plugins/crm/admin/contact/view.page.php
@@ -41,7 +41,6 @@ function jpcrm_render_contact_view_page( $id = -1 ) {
 				'withOwner'                  => true,
 				'withValues'                 => true,
 				'withExternalSourcesGrouped' => true,
-				'withDND'                    => true,
 
 				// but we limit to the top 20 (quotes, invs, trans etc.)
 				// note that this means we have to add calls to specific_obj_type_count_for_assignee, but it protects against contacts with 1000 objs etc.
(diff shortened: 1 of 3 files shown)


gogdzl (dismissed):
Simple. It works!

gogdzl:
On a second thought, this is not the 'jpcrm' way to do it.
I think the most correct way would be to use the `$wheres` array, as in [974909e](https://github.com/Automattic/jetpack/pull/31839/commits/974909e2e3c5eaad9fb66e2e886e2f52095715e3)

Feel free to undo it if you judge that's not the best approach.

cleacos:
Yes, better, thanks. It works too. Go!

gogdzl (approved):

Issue #31741, opened 2023-07-07
Enhancement: Boost ISA - add tracking events


dilirity:
### Impacted plugin

Boost

### What

Add tracking events when interacting with the Image Size Analysis.

### How

- See full report/View report in progress button;
- Analyze again link
- Group links
- Clicking the "Edit page" or "Fix on page" buttons.
- expanding collapsing issues?
- total issues found per report (individual groups also).

thingalon:
Search for `recordBoostEvent` in the codebase.

thingalon:
Things worth tracking: 
- Number of issues found (track from wpcom)
- Requests for an image analysis (track from wpcom) in the API endpoint for kicking off a report.
- Clicking edit page (do it in javascript - `recordBoostEvent`)

dilirity:
This was done in #31840.

PR #31840, opened 2023-07-12
Boost: add tracking events to ISA report

donnchawp:
Record events when the "edit page" and "fix on page" buttons on the ISA report are clicked.

For more, see #31741

## Proposed changes:
* Add a new function called recordBoostEventAndRedirect to record an event and redirect to a new URL.
* Use that function on the two previously mentioned buttons.

### Other information:

No

## Testing instructions:
* Apply patch
* You might want to add a console.log to `recordBoostEventAndRedirect` to check that it's working.
* View the ISA report and click on the edit and fix buttons on any image.


projects/plugins/boost/app/assets/src/js/utils/analytics.ts
```diff
@@ -73,3 +73,13 @@ function addBoostProps( props: TracksEventProperties ): TracksEventProperties {
...
+
+export function recordBoostEventAndRedirect(
+	url: string,
+	eventName: string,
+	eventProp: TracksEventProperties
+) {
+	recordBoostEvent( eventName, eventProp ).then( () => {
+		window.location.href = url;
+	} );
+}
```
thingalon: I think this would be easier to read as an `async` function. 

```suggestion
export async function recordBoostEventAndRedirect(
	url: string,
	eventName: string,
	eventProp: TracksEventProperties
) {
	await recordBoostEvent( eventName, eventProp );
	window.location.href = url;
}
```

This also has the added benefit that if `recordBoostEvent` rejects/throws, then so will this function. So callers can know for sure that it will either do what it says on the tin, or will throw an error. :)

thingalon (changes requested):
Nice; thanks for picking this up. Looks like this will do the job, but I have one recommendation.

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/MultiProgress.svelte
```diff
@@ -34,6 +34,8 @@
 					isLink={hasIssues}
 					class="jb-navigator-link"
 					to="/image-size-analysis/{group}/1"
+					trackEvent="clicked_isa_settings_group"
+					trackEventProps="group"
```
dilirity: I think this should be the group name. Maybe you meant to write `{group}` instead of `"group"` 😅

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/MultiProgress.svelte
```diff
@@ -50,6 +52,8 @@
...
```
dilirity: Using string `group` instead of group slug.

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/MultiProgress.svelte
```diff
@@ -64,6 +68,8 @@
 						isLink={hasIssues}
 						class="jb-navigator-link"
 						to="/image-size-analysis/{group}/1"
+						trackEvent="clicked_isa_settings_group"
+						trackEventProps="group"
```
dilirity: Using string `group` instead of group slug.

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/RecommendationsMeta.svelte
```diff
@@ -53,6 +54,10 @@
 	 * Start a new image analysis job.
 	 */
 	async function onStartAnalysis() {
+		const $event_name =
+			$isaSummary.status === ISAStatus.Completed ? 'clicked_restart_isa' : 'clicked_start_isa';
+		recordBoostEvent( $event_name, {} );
+
```
dilirity: I think explicitly adding individual events to the buttons themselves is better. Calling the function doesn't mean that a click event invoked it.

I don't have a strong preference on this. Just an observation.

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/RecommendationsMeta.svelte
```diff
@@ -122,7 +127,15 @@
 	<!-- Show a button to view the report if it's in progress or completed. -->
 	{#if [ ISAStatus.Queued, ISAStatus.Completed ].includes( $isaSummary.status ) && ! requestingReport}
 		<div class="button-area">
-			<Button href="#image-size-analysis/all/1" disabled={requestingReport}>
+			<Button
+				disabled={requestingReport}
+				on:click={() =>
+					recordBoostEventAndRedirect(
+						'#image-size-analysis/all/1',
+						'clicked_view_isa_report',
```
dilirity: Maybe this should be more specific to what it's tied to? If we were to add another place to check reports from, this name wouldn't be unique enough.

projects/plugins/boost/app/assets/src/js/utils/analytics.ts
```diff
@@ -73,3 +73,12 @@ function addBoostProps( props: TracksEventProperties ): TracksEventProperties {
 
 	return { ...defaultProps, ...props };
 }
+
+export async function recordBoostEventAndRedirect(
+	url: string,
+	eventName: string,
+	eventProp: TracksEventProperties
```
dilirity: I noticed that a lot of places where the function is called, it's with an empty object. Maybe that should be the default value of `eventProp`?

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/recommendations/row-types/ImageSizeRow.svelte
```diff
@@ -137,7 +138,16 @@
...
 			{#if details.page.edit_url}
 				<div class="jb-actions">
-					<Button width="auto" href={details.page.edit_url} fill>
+					<Button
+						width="auto"
+						fill
+						on:click={() =>
+							recordBoostEventAndRedirect(
+								details.page.edit_url,
+								'clicked_fix_on_page_on_isa_report',
```
dilirity, line 151: Thinking about this - maybe we should include the type of device as well? That way, we could later filter based on device type to see what's fixed most by users.

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/recommendations/row-types/TableRowHover.svelte
```diff
@@ -9,7 +11,14 @@
 
 	{#if edit_url}
 		<div class="button-container">
-			<Button small fill href={edit_url}>Edit Page</Button>
+			<Button
+				small
+				fill
+				on:click={() =>
+					recordBoostEventAndRedirect( edit_url, 'clicked_edit_page_on_isa_report', {} )}
```
dilirity: Same as the above suggestion for device type. I think it's a good idea to include that in here.

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/MultiProgress.svelte
```diff
@@ -34,6 +34,8 @@
 					isLink={hasIssues}
 					class="jb-navigator-link"
 					to="/image-size-analysis/{group}/1"
+					trackEvent="clicked_isa_settings_group"
```
dilirity: What if we name this `clicked_isa_summary_group` instead? With the upcoming architecture changes to the UI, who knows if the summary would be on the main Settings page :D

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/RecommendationsMeta.svelte
```diff
@@ -53,6 +54,10 @@
 	 * Start a new image analysis job.
 	 */
 	async function onStartAnalysis() {
+		const $event_name =
+			$isaSummary.status === ISAStatus.Completed ? 'clicked_restart_isa' : 'clicked_start_isa';
```
dilirity: Maybe include the summary part in the names as well?

dilirity (changes requested):
Great work on this! I hope you don't mind, but I added a few suggestions in here as I think they would help us in the long run 😸 

Most notably, the naming of events. I myself struggle with this, but I think overall, it would be good if we included more context. That way, we could have more accurate track queries.

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/MultiProgress.svelte
donnchawp: Thanks, you're right!

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/RecommendationsMeta.svelte
donnchawp: True. I'll fix this.

projects/plugins/boost/app/assets/src/js/utils/analytics.ts
donnchawp: It calls recordBoostEvent() immediately where the second parameter is a TracksEventProperties object. I'd rather leave it that way.

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/recommendations/row-types/ImageSizeRow.svelte
donnchawp: I added the device type in both links in https://github.com/Automattic/jetpack/pull/31840/commits/10a3840e278ad162b23784ab38a400062dd443d7 but when I console_log it Firefox doesn't let me expand the propType when the edit post page appears. It doesn't appear to let me do that either before this change so who knows what's happening there? Maybe the post editor is interfering with it?

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/recommendations/row-types/TableRowHover.svelte
donnchawp: See above :)

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/MultiProgress.svelte
donnchawp: Changed it to "clicked_isa_group_on_summary_page" to make is similar to "clicked_view_isa_report_on_summary_page"

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/RecommendationsMeta.svelte
donnchawp: Added "_on_summary_page" to both event names in https://github.com/Automattic/jetpack/pull/31840/commits/1279bb1c7bcee1804b4b13bf484ee8c32c60eed0

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/MultiProgress.svelte
```diff
@@ -34,6 +34,8 @@
 					isLink={hasIssues}
 					class="jb-navigator-link"
 					to="/image-size-analysis/{group}/1"
+					trackEvent="clicked_isa_group_on_summary_page"
+					trackEventProps={group}
```
thingalon, line 38: I don't understand why this uses a different tracks event to the others on this page. Is there a distinction we need to track?

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/RecommendationsMeta.svelte
```diff
@@ -67,6 +68,15 @@
 		}
 	}
 
+	function handleAnalyzeClick() {
+		const $event_name =
+			$isaSummary.status === ISAStatus.Completed
+				? 'clicked_restart_isa_on_summary_page'
+				: 'clicked_start_isa_on_summary_page';
+		recordBoostEvent( $event_name, {} );
+		return onStartAnalysis();
```
thingalon: Renamed `onStartAnalysis` to just `startAnalysis` in 2ae83afece. I don't mind which convention we use; `handleX` or `onX`, but `on...` implies it's an event handler, and it no longer is. It's just being called from an event handler. :)

projects/plugins/boost/app/assets/src/js/utils/analytics.ts
thingalon: I won't be proscriptive over it, but I agree with @dilirity - it seems that we have a lot of events with no props, so setting a default value of `{}` makes sense to me.

thingalon (dismissed):
This should do the trick, thanks donncha. A question inline about the reasoning behind using different event names, but I'm not too concerned.

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/MultiProgress.svelte
donnchawp: Thanks, I forgot to update the others!

donnchawp: Updated in https://github.com/Automattic/jetpack/pull/31840/commits/803e1fed91e75047afac16bd34faed181bf0e5f4

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/RecommendationsMeta.svelte
donnchawp: Thanks!

projects/plugins/boost/app/assets/src/js/utils/analytics.ts
donnchawp: Done!

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/RecommendationsMeta.svelte
donnchawp: Changed that to "clicked_view_isa_report_on_summary_page"

dilirity (approved):

haqadn:
Cherry-picked b2a5b7f07f4

PR #31843, opened 2023-07-12
Live Branches: Don't allow selecting unbuilt plugins

anomiex:
## Proposed changes:

Jurassic Ninja now has an API endpoint that can list which plugins are built for a particular branch. Use that to disable checkboxes for plugins that have not been built.

### Other information:

## Jetpack product discussion

p1680802118831319-slack-C043B54G6V7

No

## Testing instructions:

* Install the updated version of the script (e.g. with [this link](https://github.com/Automattic/jetpack/raw/update/live-branches-check-built-plugins/tools/jetpack-live-branches/jetpack-live-branches.user.js)).
* Check that Live Branches on this PR shows a "No plugins have been built for this PR." message. The refresh link should refresh to that same message.
* Go to an open PR that did build some (but not all) plugins. Check that Live Branches works, and only allows selecting the plugins that were actually built.

### DIFF
--- a/tools/jetpack-live-branches/jetpack-live-branches.user.js
+++ b/tools/jetpack-live-branches/jetpack-live-branches.user.js
@@ -33,7 +33,7 @@
 		}
 
 		#jetpack-live-branches label.disabled {
-			color: rgba(0,0,0,0.5);
+			color: var( --color-fg-muted, #7d8590 );
 		}
 	`;
 
--- a/tools/jetpack-live-branches/jetpack-live-branches.user.js
+++ b/tools/jetpack-live-branches/jetpack-live-branches.user.js
@@ -1,7 +1,7 @@
 // ==UserScript==
 // @name         Jetpack Live Branches
 // @namespace    https://wordpress.com/
-// @version      1.28
+// @version      1.29
 // @description  Adds links to PRs pointing to Jurassic Ninja sites for live-testing a changeset
 // @grant        GM_xmlhttpRequest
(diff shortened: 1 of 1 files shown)


tools/jetpack-live-branches/jetpack-live-branches.user.js
```diff
@@ -18,6 +18,25 @@
...
+			display: flex;
+			flex-wrap: wrap;
+		}
+
+		#jetpack-live-branches label {
+			font-weight: inherit;
+		}
+
+		#jetpack-live-branches label.disabled {
+			color: rgba(0,0,0,0.5);
```
bindlegirl:
```suggestion
			color:#7d8590;
```

I would suggest a different color choice to work better with dark mode because it looks strange this way:

Screenshot 2023-07-13 at 14 48 01

anomiex: Good idea, but it looks like we can go one step further and try to follow GitHub's own theming with something like
```suggestion
			color: var( --color-fg-muted, #7d8590 );
```
Not sure whether we should fall back to the current light theme value (#656d76) or dark theme value (#7d8590) should GH ever rename their variable. Or maybe halfway in between (#717983)?

bindlegirl: I tried the #7d8590 with a light and dark theme, and it works well either way. It's a small thing to change if we find things don't work.

bindlegirl (approved):

PR #31846, opened 2023-07-12
CRM: Fix for Multiline items in the top-right menu

gogdzl:
## Proposed changes:
* This PR enhances the alignment of the top right menu items. Previously, when the text was too long, it would wrap under the icon, but now it respects the alignment and wraps appropriately within its own space.

### Other information:

## Jetpack product discussion
https://github.com/Automattic/zero-bs-crm/issues/3152

No.

## Testing instructions:

* Change the language to spanish

In `trunk` you should get something like this:
image

In `fix/crm/3152-multiline-names-in-menu` you should see the items alignment as the screenshot below:
image

### DIFF
--- a/projects/plugins/crm/includes/ZeroBSCRM.Core.Menus.Top.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.Core.Menus.Top.php
@@ -269,8 +269,6 @@ function zeroBSCRM_admin_top_menu( $branding = 'zero-bs-crm', $page = 'dash' ) {
 			});
 		})
 		</script>
-		<style>
-		</style>
 		<!---  // mobile only menu -->
 	<jpcrm-top-menu id="jpcrm-top-menu">
 		<div class="logo-cube <?php echo esc_attr( $admin_menu_state ); ?>">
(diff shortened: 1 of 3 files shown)


projects/plugins/crm/includes/ZeroBSCRM.Core.Menus.Top.php
```diff
@@ -269,6 +269,8 @@ function zeroBSCRM_admin_top_menu( $branding = 'zero-bs-crm', $page = 'dash' ) {
 			});
 		})
 		</script>
+		<style>
+		</style>
```
cleacos: Do we need this empty style element?

cleacos:
```suggestion
```

cleacos (dismissed):
It works as expected. I checked all the links too.

image

I left a minor comment to remove an empty <style> tag. Other than that, LGTM!

projects/plugins/crm/includes/ZeroBSCRM.Core.Menus.Top.php
gogdzl: Leftover from tests! Thanks for pointing that out!

cleacos (approved):
re-LGTM!

PR #31853, opened 2023-07-13
CRM: Create Automations UI root

robertf4:
## Proposed changes:

Fixes https://github.com/Automattic/zero-bs-crm/issues/3179

This PR adds the root of a React app for Automations and creates a new Automations module to load and serve it.

image

**It aim to do so by:**

* Register "Automations" as a core module
* Add feature flag to prevent the UI to be exposed.
* Finish the CRM specific `AdminPage` component
  * We also correct an older dummy test for the OBW component that was used to test if React builds worked and was tested.
* Add a new `<AutomationAdmin>` component that will be the main component for the automations view.

**Caveats:**

* The RNA container component only supports the narrow Jetpack view - not the ~1551px one that was recently used in Stats and CRM. I will create a separate PR to address this issue in the RNA package.

### Other information:

## Jetpack product discussion

p1HpG7-mCz-p2

No

## Testing instructions:

1. Add the following to a mu-plugin (e.g. `0-sandbox.php`): `add_filter( 'jetpack_crm_feature_flag_automations', '__return_true' );`
2. Run `pnpm build` in the CRM folder.
3. Visit `/wp-admin/admin.php?page=jpcrm-automations` and verify that you see a page with an "Automations" header like the screenshot above.
4. In the CRM folder run `pnpm run test` and verify that all Automations tests pass.
5. Verify that Automations shows up in the menu bar.
6. Disable the feature flag filter and verify that Automations does not show up in the menu bar and that the page is not accessible.


projects/plugins/crm/modules/automations/jpcrm-automations-init.php
```diff
@@ -0,0 +1,118 @@
...
+ * Register the admin page for the Automations UI
+ *
+ * @return void
+ */
+function register_page() {
+	global $zbs;
+
+	add_submenu_page(
+		$zbs->slugs['automations'],
+		__( 'Welcome to Jetpack CRM Automations', 'zero-bs-crm' ), // TODO: change this
```
robertf4: I put the TODO here as I'm not sure what this copy should be, so this is just a reasonable placeholder.

projects/plugins/crm/src/js/components/automations/styles.scss
```diff
@@ -0,0 +1,9 @@
+@import '../../../../sass/_color_studio.scss';
+
+.jpcrm-automations-admin__container {
+    padding: 34px;
+}
+
+.jpcrm-automations-admin__subheader {
+    color: $studio-gray-60;
+}
```
kallehauge: Missing new line 🆕 📏

projects/plugins/crm/modules/automations/jpcrm-automations-init.php
```diff
@@ -0,0 +1,118 @@
+<?php
```
kallehauge, line 1: **Request**: I assume that this was converted from a class to function due to "the magical logic that loads the init file", but would it be cleaner if we separated this into two files?

1. `class-admin-page.php`: Contains an `Admin_Page` class with 90% of the code from this file.  
2. `jpcrm-automations-init.php` Basically just calls `Admin_Page->init()` together with the feature flag (there might be some magical loading logic I'm forgetting).

By doing the above, we can also init other classes when we e.g. start to bootstrap the automations engine and it doesn't have to be mixed together with admin page related code

projects/plugins/crm/modules/automations/jpcrm-automations-init.php
```diff
@@ -0,0 +1,118 @@
...
+	global $zbs;
+
+	add_submenu_page(
+		$zbs->slugs['automations'],
+		__( 'Welcome to Jetpack CRM Automations', 'zero-bs-crm' ), // TODO: verify or get better copy
+		'',
+		'manage_options',
+		'jpcrm-automations',
+		'Automattic\Jetpack_CRM\Automations\render_page'
+	);
```
kallehauge: **Observation**: The `Jetpack CRM` menu item isn't "active" while we're on this page. 

I also believe that `$zbs->slugs['automations']` can/should be used instead of the hardcoded value.

```suggestion
	add_submenu_page(
		$zbs->slugs['home'],
		__( 'Welcome to Jetpack CRM Automations', 'zero-bs-crm' ), // TODO: verify or get better copy
		'',
		'manage_options',
		$zbs->slugs['automations'],
		'Automattic\Jetpack_CRM\Automations\render_page'
	);
```

The above example might not work, and if it doesn't, then it's 💯 to "just" register a task on the React navigation router issue so we can continue to debug it there

Screenshot 2023-07-13 at 22 36 24

projects/plugins/crm/src/js/components/automations/index.tsx
```diff
@@ -0,0 +1,14 @@
+import { __ } from '@wordpress/i18n';
+
+import './styles.scss';
+
+export const AutomationsAdmin = () => {
```
kallehauge: The file name and component name doesn't match.

Referencing it as "Admin" or "Page" is otherwise 👍 with me

projects/plugins/crm/src/js/components/automations/index.tsx
```diff
@@ -0,0 +1,14 @@
...
+	return (
+		<div className="jpcrm-automations-admin__container">
```
kallehauge: Should we perhaps reuse the `<AdminPage>` RNA component since it makes sure we show a footer etc? 😄

If we use the `AdminPage` component that already exist in CRM (which wraps the RNA component) we should be able to only define e.g. the footer in 1 location.

**References**:

* `projects/plugins/crm/src/js/components/admin-page/index.tsx`
* `projects/js-packages/components/components/admin-page/index.tsx`

projects/plugins/crm/src/js/components/automations/view.tsx
```diff
@@ -0,0 +1,23 @@
...
+ * Render function
+ */
+const render = () => {
+	const container = document.getElementById( 'jetpack-crm-automations-root' );
+
+	if ( null === container ) {
+		return;
+	}
+
+	const component = <AutomationsAdmin />;
```
kallehauge: AFAIK then `ThemeProvider` should make sure that we have support/access to all Jetpack colors (e.g.: `--jp-black-80`)

```suggestion
	// @todo: Remove fallback when we drop support for WP 6.1
	const component = (
		<ThemeProvider>
			<AutomationsAdmin />
		</ThemeProvider>
	);
```

kallehauge (changes requested):

projects/plugins/crm/modules/automations/jpcrm-automations-init.php
kallehauge: The existing routes just say e.g. `Quotes`.
Maybe it's about time we update it to prefix it with e.g.: `Jetpack CRM: `, so in this case it would just be `Jetpack CRM: Automations`.

The main objective here is to make it easy for folks to navigate their browser tabs and doesn't have to apply to any SEO related rules

projects/plugins/crm/modules/automations/jpcrm-automations-init.php
robertf4: I made it functions instead of a class because there is no state here. So in the onboarding wizard there is a singleton pattern and accessing the singleton causes all of the init logic to run, but it is all functions with no state so I decided to just make this functions.

Until there is state I'd prefer to keep things as functions just so the semantics of the code match the functionailty, but I can certainly begin splitting out this logic into more files.

projects/plugins/crm/src/js/components/automations/index.tsx
robertf4: I switched the file name.

projects/plugins/crm/src/js/components/automations/index.tsx
robertf4: In this case we already have the header and footer:

image

Those are both existing functionality written in PHP and added to the page by CRM, so if we switch to the RNA component then we would have to recreate it in React and also change CRM to not render this header and footer for the automations page. This could be something worth doing, especially if we want to start making more pages in React, but I think it's a project on its own.

projects/plugins/crm/modules/automations/jpcrm-automations-init.php
robertf4: Once I got the menu to work I found that this title is what is used in the menu:

image

So I figured it is best as Automations.

projects/plugins/crm/modules/automations/jpcrm-automations-init.php
robertf4: I split out all the admin page init logic into `admin-page-init.php`.

projects/plugins/crm/modules/automations/jpcrm-automations-init.php
robertf4: In investigating this I found that CRM has a whole complex system for building the menu in `ZeroBSCRM.Core.Menus.WP.php`. I was able to get the menu to work by integrating the Automations page into this system.

To accomplish this, the rendering of the page had to be moved to `admin/automations/main.page.php` while the rest of the init logic is in `admin-page-init.php` due to the order that certain hooks are called.

Initially I didn't like the separation, but this pattern does make more sense in the overall context of CRM code than what I was doing. It also makes sense for the logic in `admin-page-init.php` to remain in the module as it is responsible for adding the initial state which will require using business logic that is likely to be shared with other code in the module.

projects/plugins/crm/src/js/components/automations/index.tsx
robertf4: For reference, this is what it looks like when I use `AdminPage` without making any other changes to CRM. The headers and footers are doubled.

image

projects/plugins/crm/src/js/components/automations/view.tsx
robertf4: I'm not exactly sure what `ThemeProvider` is giving us as I still need to have `@import '../../../../sass/_color_studio.scss';` in the scss file in order for it to compile, but I added it anyway.

projects/plugins/crm/src/js/components/automations/index.tsx
kallehauge: The admin component accepts `showHeader` and `showFooter` props to hide them. I've gone ahead and made a small refactor so it's reusable for future admin pages as well.

projects/plugins/crm/src/js/components/automations/view.tsx
kallehauge: `<ThemeProvider>` makes sure that the `<AutomationsAdmin>` component can use the CSS variables Jetpack exposes:

Screenshot 2023-07-18 at 15 41 28

kallehauge (approved):

PR #31856, opened 2023-07-13
Add a login link to newsletter paywall

### LINKED ISSUE #31857: Add login links to Subscriber paywall
The paywall / subscriber-only wall for newsletters doesn't include a login link, making it much harder to login if you already are a subscriber or paid subscriber.

Let's add the link:

Image

simison:
Resolves https://github.com/Automattic/jetpack/issues/31857

TODO:

- [ ] Better looking email prompt
- [ ] ~Splitting code into separate files in a way that makes sense for subscribe block and paywall separately~. I looked into this but concluded it isn't worth it at least yet; we're adding small amount of JS for the link, but duplicating the rest would actually mean _more_ JS in total for anyone with paywall showing up. That might change if we add some styles/JS into the mix for asking the email, but it's also a good follow-up PR.

## Proposed changes:

* Cleans out translations so that we send full sentences for translations; partial ones are kind of impossible to translate in many languages. I could split that to its own PR though.
* Adds "login" links to the paywall

image

### Other information:

## Testing instructions:

* Have subscriptions module enabled in Jetpack, or use WP.com simple site
* Connect Stripe and set payments from WP.com `/earn` page or use links at post editor's "Newsletter" sidepanel
* Create two new posts; one for subscribers only, another for paid subscribers only.
* Create a new free subscriber at the site, and paid subscriber
* Try login for both posts with both users

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -832,7 +832,7 @@ function get_locked_content_placeholder_text( $newsletter_access_level ) {
 			<!-- wp:jetpack/subscriptions {"borderRadius":50,"borderColor":"primary","className":"is-style-compact"} /-->
 
 			<!-- wp:paragraph {"align":"center"} -->
-			<p class="has-text-align-center">' . esc_html( $access_question ) . ' <a href="#">' . esc_html__( 'Login', 'jetpack' ) . '</a></p>
+			<p class="has-text-align-center">' . esc_html( $access_question ) . ' <a href="#" class="jetpack-subscriber-paywall-login">' . esc_html__( 'Login', 'jetpack' ) . '</a></p>
 			<!-- /wp:paragraph -->
 
 			</div>
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -809,29 +809,32 @@ function maybe_gate_existing_comments( $comment ) {
  * @return string
  */
 function get_locked_content_placeholder_text( $newsletter_access_level ) {
-	// Default to subscribers
-	$access_level = __( 'subscribers', 'jetpack' );
+	// Only display paid texts when Stripe is connected and the post is marked for paid subscribers
+	$is_paid_post = $newsletter_access_level === 'paid_subscribers'
+		&& ! empty( Jetpack_Memberships::get_connected_account_id() );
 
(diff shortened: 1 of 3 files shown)


crisbusquets:
Hi @simison !

Can we use this issue to update the paywall design, so it matches de modal @edanzer was developing?

Screenshot 2023-07-13 at 10 43 20

Figma file (h8tMpJlCqNFemEerU9owHI-fi-3483_150189)

simison:
@crisbusquets  Yeah, sounds good!

I'll try craft something neat looking for the email input but feel free to share quick design here as well.

Ideas:

- Swap paywall content inline to just show input+button block but with button label "Login".
    - Plus some link to "subscribe" instead to return to previous state?
    - image
- A modal same size/look as the payments modal, with the same inputs as above.

Doesn't need to be perfect on first attempt; just good enough. Function over form for this one, because the production experience is poor.

crisbusquets:
> * A modal same size/look as the payments modal, with the same inputs as above.

Just to make sure I understand: this `sign in` link will trigger the flow where the user introduces their email, receive a code, introduce the code and then they're logged in?

If that's the case, I would open that model when they click the `sign-in` link in the paywall.

What do you think?

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -809,29 +809,32 @@ function maybe_gate_existing_comments( $comment ) {
...
-			<h2 class="wp-block-heading has-text-align-center" style="margin-bottom:var(--wp--preset--spacing--60);font-size:24px">' .
+			<div class="wp-block-group has-border-color has-primary-border-color" style="border-width:1px;border-radius:4px;padding-top:var(--wp--preset--spacing--80);padding-right:var(--wp--preset--spacing--80);padding-bottom:var(--wp--preset--spacing--80);padding-left:var(--wp--preset--spacing--80)">
 
-			/* translators: Newsletter access. Possible values are "paid newsletters" and "subscribers" */
-			sprintf( esc_html__( 'This post is for %s', 'jetpack' ), $access_level )
-
-			. '</h2>
+			<!-- wp:heading {"textAlign":"center","style":{"typography":{"fontSize":"24px"},"spacing":{"margin":{"bottom":"var:preset|spacing|60"}}}} -->
+			<h2 class="wp-block-heading has-text-align-center" style="margin-bottom:var(--wp--preset--spacing--60);font-size:24px">' . esc_html( $access_heading ) . '</h2>
```
simison, line 829: I did number of style edits here, but I moved these to a separate PR to keep things easier to review. That should get merged first: https://github.com/Automattic/jetpack/pull/31880

lezama:
Working on a simple version for simple sites here: https://github.com/Automattic/jetpack/pull/32849

PR #31863, opened 2023-07-13
Boost: Fix speed score auto refreshing on page load

### LINKED ISSUE #31396: Boost: Page Speed scores refresh on each page load
### Impacted plugin

Boost

### Quick summary

Tested with free version - 1.9.1.

After a page speed score gets generated, refreshing the page generates another page speed score refresh.

### Steps to reproduce

1. Setup Boost 1.9.1 (free)
2. Wait for Page Speed scores to generate
3. Refresh the page

Page Speed scores should only refresh if they are stale - not on every page refresh.

### What actually happened

Page Speed scores refresh on each page load.

### Impact

Some (< 50%)

### Available workarounds?

Self-hosted

### Logs or notes

_No response_

dilirity:
Fixes #31396

### Notes

My idea was to create a helper variable, that holds the state of the config after initial page load. The variable would then be used to prevent auto refreshes. I'm not entirely happy with the solution, though it seems to work. I would love to implement a proper one.

## Proposed changes:

* Prevent speed scores from automatically refreshing each time the Boost settings page is loaded.

### Other information:

## Jetpack product discussion

n/a

no

### Test if it's broken

* Follow the "Steps to reproduce" section in the ticket - https://github.com/Automattic/jetpack/issues/31396.

### Test that it works

* Setup this version of Boost
* Open the Boost settings page and make sure the speed scores aren't automatically refreshed.

### DIFF
--- a/projects/plugins/boost/app/assets/src/js/pages/settings/sections/Score.svelte
+++ b/projects/plugins/boost/app/assets/src/js/pages/settings/sections/Score.svelte
@@ -35,36 +35,43 @@
 		isStale: false,
 	};
 
-	refreshScore( false );
+	// Load the speed score. Will be cached in the plugin.
+	loadScore();
 
-	/**
-	 * The configuration that led to latest speed score.
-	 */
+	// Flat list of which modules are active; useful for tracking changes in state.
 	$: activeModules = Object.entries( $modulesState ).reduce( ( acc, [ key, value ] ) => {
 		if ( key !== 'image_guide' && key !== 'image_size_analysis' ) {
 			acc.push( value.active );
 		}
 		return acc;
 	}, [] );
-	$: lastCreatedState = $criticalCssState.created;
-	$: currentScoreConfigString = JSON.stringify( {
-		modules: activeModules,
-		criticalCss: {
-			created: lastCreatedState,
-		},
-	} );
 
-	let setinitialScoreConfigString = false;
-	$: initialScoreConfigString = () => {
-		if ( ! setinitialScoreConfigString ) {
-			setinitialScoreConfigString = true;
-			return currentScoreConfigString;
+	// Keep a string describing the settings that may affect the score. We'll use it to
+	// work out when we need a new speed score.
+	$: currentScoreConfigString = JSON.stringify( [ activeModules, $criticalCssState.created ] );
+
+	// State we had the last time we requested a speed score -- or when the page first loaded.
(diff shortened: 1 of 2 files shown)


projects/plugins/boost/app/assets/src/js/pages/settings/sections/Score.svelte
```diff
@@ -54,6 +54,16 @@
...
 
+	let setinitialScoreConfigString = false;
+	$: initialScoreConfigString = () => {
+		if ( ! setinitialScoreConfigString ) {
+			setinitialScoreConfigString = true;
+			return currentScoreConfigString;
+		}
+
+		return false;
+	};
```
thingalon: I feel like this could be expressed with:

```
$: initialScoreConfigString = initialScoreConfigString ?? currentScoreConfigString
```

However, I don't think that's the best fix here. I think you've stumbled upon layers of broken code that look like _fancy_ code, so it's hard to see how broken it all is.

Originally, `currentScoreConfigString` was a fancy way to trigger a refresh whenever any setting changed. We used a Svelte trick using derived stores to call a function whenever the value changed. But that was hard to understand so it got refactored in a way that broke it -- but left `currentScoreConfigString` as a vestigial dangling bit of code.

At this point:
- `$: activeModules` and `$: lastCreatedState` are only used to populate `currentScoreConfigString`.
- `$: initialScoreConfigString` is only used to check if `currentScoreConfigString` is different from when the page first loaded.

I think if you change a setting, wait for a speed score, then change it *back* you won't trigger a re-speed-score. It's gotten a bit silly.

Fundamentally, `$:` is a way to re-run a piece of code every time a value it relies on changes. So surely we could just do something like: 

`$: debounceRefresh(), $activeModules, $$criticalCssState.created`

Would that work here?

thingalon (dismissed):
This works, but it's paving over pre-existing problems with a complicated solution. It would be simpler if we tear out the problems first.

projects/plugins/boost/app/assets/src/js/pages/settings/sections/Score.svelte
dilirity: Refactored in [2c5c3da](https://github.com/Automattic/jetpack/pull/31863/commits/2c5c3daab315c43db45e13986ec5bcae798c4110).

haqadn (approved):
Thank you for cleaning it up! 👍🏼

haqadn:
Cherry-picked 52ee604d

PR #31869, opened 2023-07-13
Top posts widget: add srcset to thumbnails

sgomes:
## Proposed changes:

In an effort to move away from the `devicepx` script towards native functionality, this PR adds a `srcset` to thumbnails in the top posts widget. The browser will automatically pick the most appropriate image from the options in the `srcset`, depending on the screen pixel ratio and the current zoom value.

* Add a `srcset` attribute to thumbnails in the top posts widget
* Add `loading=lazy` to images in the top posts widget
* Extract `srcset`-generation code in related posts, and promote it to `Jetpack_PostImages`
* Refactor related posts to use the extracted code

### Other information:

## Jetpack product discussion
None.

No.

## Testing instructions:
To ensure that top posts work correctly:
- Add a "Top Posts" widget to your test blog
- Configure it to display as an image list or image grid
- Load a test page and ensure that thumbnails display correctly
- Find a thumbnail whose source is a post image (not a gravatar or blavatar), and ensure that it has a `srcset` with different pixel ratios (up to 4x, if the source image is large enough).

To ensure that related posts continue to work correctly after the refactor:
- Enable the "Related Posts" feature for your test blog (`Settings -> Traffic -> Show related content after posts`).
- Enable related posts thumbnails (`Show a thumbnail image where available` in the same place).
- Open a post that has related posts with thumbnails.
- Smoke-test for any breakage.
- Verify that the thumbnails include a `srcset` with different pixel ratios (up to 4x, if the source image is large enough).

### DIFF
--- a/projects/plugins/jetpack/class.jetpack-post-images.php
+++ b/projects/plugins/jetpack/class.jetpack-post-images.php
@@ -726,6 +726,47 @@ class Jetpack_PostImages {
 	}
 
 	/**
+	 * Takes an image and base pixel dimensions and returns a srcset for the
+	 * resized and cropped images, based on a fixed set of multipliers.
+	 *
+	 * @param  array $image Array containing details of the image.
+	 * @param  int   $base_width Base image width (i.e., the width at 1x).
+	 * @param  int   $base_height Base image height (i.e., the height at 1x).
+	 * @return string The srcset for the image.
+	 */
+	public static function generate_cropped_srcset( $image, $base_width, $base_height ) {
+		$srcset = '';
+
+		if ( ! is_array( $image ) || empty( $image['src'] ) || empty( $image['src_width'] ) ) {
+			return $srcset;
+		}
+
+		$multipliers   = array( 1, 2, 3, 4 );
+		$srcset_values = array();
+		foreach ( $multipliers as $multiplier ) {
+			// Forcefully cast to int, in case we ever add decimal multipliers.
+			$srcset_width  = (int) ( $base_width * $multiplier );
+			$srcset_height = (int) ( $base_height * $multiplier );
+			if ( $srcset_width < 1 || $srcset_width > $image['src_width'] ) {
+				break;
+			}
+
+			$srcset_url      = self::fit_image_url(
+				$image['src'],
+				$srcset_width,
+				$srcset_height
+			);
+			$srcset_values[] = "{$srcset_url} {$multiplier}x";
+		}
+
+		if ( count( $srcset_values ) > 1 ) {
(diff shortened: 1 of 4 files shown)


projects/plugins/jetpack/class.jetpack-post-images.php
```diff
@@ -725,6 +725,47 @@ public static function get_images( $post_id, $args = array() ) {
...
+			if ( $srcset_width < 1 || $srcset_width > $image['src_width'] ) {
+				break;
+			}
+
+			$srcset_url      = self::fit_image_url(
+				$image['src'],
+				$srcset_width,
+				$srcset_height
+			);
+			$srcset_values[] = "{$srcset_url} {$multiplier}x";
```
bindlegirl, line 759:
```suggestion
			$srcset_values[] = "{$srcset_url} {$srcset_width}w {$multiplier}x";
```
Would this work better? I've noticed the without the `w` property browser picks a larger image from srcset (400 instead of 200 in my test).

sgomes: Hey @bindlegirl, thank you for taking a look! 👍

The syntax I used should be correct; we've already used it in other places, and it should match [the spec](https://html.spec.whatwg.org/multipage/images.html#srcset-attribute). There are two ways of defining a `srcset`: via widths, or via density multipliers. The former is probably more common, but I opted for the latter here. I've never seen them mixed as in your suggested change, and that would be invalid syntax according to the spec, if I'm reading it correctly.

As for why your browser retrieved the 400px image, is it possible that you're using a 2x (or > 1x and <= 2x, really) density display, or have an equivalent level of zoom? If so, then that would be the expected behaviour.

What does your browser indicate in `window.devicePixelRatio`, for any windows in the same screen and with the same level of zoom as the one you tested? In my case I get `2`, both in my built-in screen and in my external screen, which is why I get the 2x versions as well.

bindlegirl: Thanks for clarifying. You are right, what I suggested above is incorrect.

I expected the multiplier and width would come up with pretty much the same reasonable choice when picking a size. And yes, my screen is using a 2x density display. 
I wouldn't even think about it if I didn't have Jetpack Boost enabled on my test site, where a green dot (approving the selected image size) went yellow when testing this PR which confused me.

bindlegirl (approved):
Thanks for the clarification about width vs multiplier.

sgomes:
Thank you for the review, @bindlegirl, and thank you for going ahead and merging this, @sdixon194! 👍

PR #31876, opened 2023-07-13
Subscribe Modal: Fix header styling for long site names

edanzer:
## Proposed changes:

* Fixes a minor style issue in the new subscribe modal: when the site title is long enough for the heading to wrap, it appears left aligned. After this change, it stays centered. 
* Removes the lock icon from the top of the modal. We're not actually locking anything with this modal.
* Where possible, moves the new styles inside the block/site editor rather than using a custom css file.
* Removes a few other css lines that were having no impact.
* UPDATE: In response to feedback, also removed underline from Continue Reading link. Because it was added after, that change isn't visible in the 'after' screenshot below. 

modal styles

### Other information:

## Jetpack product discussion

None.

No.

## Testing instructions:

**To test via simple sites:**
1) Setup: 
   - Run `bin/jetpack-downloader test jetpack fix/subscribe-modal-title-styles` to load this branch in your sandbox.
   - Be sure to sandbox public-api and the domain of your test site. 
   - You'll need to enable the feature flag by adding `add_filter( 'jetpack_subscriptions_modal_enabled', '__return_true', 20 );` to wp-content/mu-plugins/0-sandbox.php
   - Enable the modal on your test site by going to Settings > Reading and toggling the subcribe modal option on. 
2) Go to any single blog post, and scroll, the modal should load. If your site title is not long enough to make the modal header wrap, just right click, open dev tools / inspector, and manually update the name to something longer.
3) Confirm that when the header wraps, it still appears centered. 

**To test in a self hosted jetpack environemnt:**
1) Setup: 
  - You'll need to set up a Jetpack dev environment and a jurassic tube test site
  - You'll need to enable the feature flag by. You can add `add_filter( 'jetpack_subscriptions_modal_enabled', '__return_true', 20 );` to tools/docker/mu-plugins/0-sandbox.php. You may need to rebuild WordPress. 
   - If you haven't already, Go to Jetpack > Settings > Discussion, scroll to subscription options and enable the modal.
2) Go to any single blog post, and scroll, the modal should load. If your site title is not long enough to make the modal header wrap, just right click, open dev tools / inspector, and manually update the name to something longer.
3) Confirm that when the header wraps, it still appears centered.

### DIFF
--- a/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.css
+++ b/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.css
@@ -29,26 +29,12 @@
 	transition: all 0.4s;
 }
 
-.jetpack-subscribe-modal__modal-content .wp-block-group {
-    justify-content: center;
-	gap: 0;
-}
-
 .jetpack-subscribe-modal__modal-content > .wp-block-group {
     max-width: 450px;
 	margin: 0 auto;
 }
 
-.jetpack-subscribe-modal__modal-content h2 {
-    margin: 4px;
-	text-align: center;
-}
-
 .jetpack-subscribe-modal.open .jetpack-subscribe-modal__modal-content {
 	top: 0;
 	visibility: visible;
-}
-
-.jetpack-subscribe-modal__close {
-	cursor: pointer;
 }
--- a/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.css
+++ b/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.css
@@ -34,6 +34,10 @@
 	margin: 0 auto;
 }
 
+.jetpack-subscribe-modal__modal-content > .wp-block-group a {
+    text-decoration: none;
+}
+
 .jetpack-subscribe-modal.open .jetpack-subscribe-modal__modal-content {
 	top: 0;
 	visibility: visible;
--- a/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.css
+++ b/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.css
@@ -52,6 +52,7 @@
 
 .jetpack-subscribe-modal__modal-content h2 {
     margin: 4px;
+	text-align: center;
 }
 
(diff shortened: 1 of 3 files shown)


projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.css
```diff
@@ -52,6 +52,7 @@
 
 .jetpack-subscribe-modal__modal-content h2 {
     margin: 4px;
+	text-align: center;
```
simison: Could we apply this to the block markup itself?

Otherwise we break the editing experience from modal to the template.

The same is true for the margin; let's instead use group block with padding values.

simison:
Some block markup in https://github.com/Automattic/jetpack/pull/31880 that you could copy to modal and it should fix this issue as well.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.css
edanzer: I've moved this into the block markup.

edanzer:
Since the original PR, based on feedback from @simison, I've moved some styling into the block editor vs the custom stylesheet. I tried moving more of the css in, but couldn't get the block styling right. So for now, any css in the custom css file seems like it needs to stay there. But anything that can be moved into the site editor (and thus edited by users) has been.

crisbusquets:
Hi @edanzer !

I almost had a brain meltdown, but managed to test it. Here's what I see (I modified the site's name using inspector, to make it long enough):

Screenshot 2023-07-20 at 11 18 07

This is probably about another PR, but I noticed that the `Continue reading` has two `text-decoration: underline` and it looks really thick (I created an issue https://github.com/Automattic/jetpack/issues/31985):

https://github.com/Automattic/jetpack/assets/25607244/a66e3e6c-6003-4162-bcc6-0685b97ad80a

edanzer:
@crisbusquets - 

> This is probably about another PR, but I noticed that the Continue reading has two text-decoration: underline and it looks really thick

That was easy and relevant enough to include here, so I've added that too in [this commit](https://github.com/Automattic/jetpack/pull/31876/commits/a95952cbae6cfb565d090d31e906eeeab119958d).

lezama (approved):

PR #31880, opened 2023-07-13
Subscriptions: update visual appearance of the locked content wall

simison:
## Proposed changes:

* Update visual appearance by tweaking block template
* Update translated sentences to not be translated in separate parts; much easier to translate full sentence.
* Add svg lock icon

**Before**

Screenshot 2023-07-13 at 23 18 23

**After**

Screenshot 2023-08-02 at 13 05 10

Mobile version could be better but that's a matter of optimizing subscribe block, not this paywall:

Screenshot 2023-08-02 at 13 05 16

### Other information:

## Testing instructions:

- Create a new post, and mark it "subscribers only"
- As logged out user, see the post at the site


projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -809,30 +809,36 @@ function maybe_gate_existing_comments( $comment ) {
...
-			<div class="wp-block-group has-border-color has-primary-border-color" style="border-width:1px;border-radius:4px;padding-top:var(--wp--preset--spacing--80);padding-right:var(--wp--preset--spacing--80);padding-bottom:var(--wp--preset--spacing--80);padding-left:var(--wp--preset--spacing--80)"><!-- wp:heading {"textAlign":"center","style":{"typography":{"fontSize":"24px"},"spacing":{"margin":{"bottom":"var:preset|spacing|60"}}}} -->
-			<h2 class="wp-block-heading has-text-align-center" style="margin-bottom:var(--wp--preset--spacing--60);font-size:24px">' .
+	// translators: %s is the name of the site.
+	$subscribe_text = sprintf( esc_html__( 'Subscribe to "%s" to keep reading and get access to the full archive.', 'jetpack' ), get_bloginfo( 'name' ) );
 
-			/* translators: Newsletter access. Possible values are "paid newsletters" and "subscribers" */
-			sprintf( esc_html__( 'This post is for %s', 'jetpack' ), $access_level )
```
simison, line 832: This string was problematic for translating. We should avoid partial strings.

simison:
@crisbusquets the sub-heading copy feels a bit off for paid subscribers. It does work for regular subscribers, tho!

In the screenshots I have `"` quotes but I removed them in the code just now; without them translations in some languages are hard, but I figured translators can add them in those languages themselves. Would probably lead to better/smoother language if we wouldn't have the site name here, though?

Note that I'm not adding "sign in" link in this PR, that's [for another PR](https://github.com/Automattic/jetpack/pull/31856).

crisbusquets:
Hi @simison !

I tested it on a WoA site, and the lock SVG throws a 404 error:

Screenshot 2023-07-17 at 12 11 17

On mobile, I'd change the `padding` from `60/70px` to `32px`. Here's a comparision:


padding

crisbusquets:
> @crisbusquets the sub-heading copy feels a bit off for paid subscribers. It does work for regular subscribers, tho!

What do you find strange? The archive part?

> Would probably lead to better/smoother language if we wouldn't have the site name here, though?

I added the site's name as a way to remind the user what they're subscribing to, and to reinforce the newsletter/site name. I like that you removed the `""`, it makes it easier to read :)

jeherve (commented):
I'm getting a 404 on the SVG file when trying to apply the patch to WordPress.com Simple, but that's because for some reason the `.svg` file doesn't get copied:

image

For some reason it doesn't get copied with using `jetpack rsync` either. Are you able to test this on WordPress.com Simple yourself?

simison:
> >the sub-heading copy feels a bit off for paid subscribers. It does work for regular subscribers, tho!
>
> What do you find strange? The archive part?

@crisbusquets  Hum, I actually don't remember what I thought. Should we use "full paid archives" for paid posts?

crisbusquets:
> @crisbusquets Hum, I actually don't remember what I thought. Should we use "full paid archives" for paid posts?

@simison Maybe we can phrase it differently:
- Regular subscribers: `Subscribe to {Site Name} to keep reading and receive all new posts in your inbox`
- Paid subscribers: `Subscribe to {Site Name} to keep reading, receive all new posts in your inbox and have access to the full archives`

Paid Subscribers are the only ones who'll have access to ALL posts, isn't it?

simison:
>I'm getting a 404 on the SVG file when trying to apply the patch to WordPress.com Simple, but that's because for some reason the .svg file doesn't get copied:

Weird! I was testing on localhost so far. Instead of using the `` tag (and image block), we can just inline the SVG and it'll just work. Tiny bit less performance-friendly but quite minor, so I wouldn't want to spend much time debugging it.

simison:
Oops... 

I let Github do the default merge instead of usual rebase I do myself, so things went a bit awry: 

image

I fixed the history but labels are a mess.

lezama (dismissed):

projects/plugins/jetpack/extensions/blocks/subscriptions/index.js
```diff
@@ -6,7 +6,9 @@ import { getIconColor } from '../../shared/block-icons';
 import attributes from './attributes';
 import deprecated from './deprecated';
 import edit from './edit';
+import lock from './lock.svg';
```
simison: Creative. 😅 We'd still have to know the hash of the file since it gets rewritten with new hash each time (for cache busting).

Instead of odd import in JS to use the file in PHP, I considered:

- Using [copy plugin](https://webpack.js.org/plugins/copy-webpack-plugin/) and just copy all `**/*.svg` files as-is to destination in:
    ```js
	{
		from: '**/*.svg',
		to: './',
		context: path.join( __dirname, '../extensions' ),
	},
    ```
    https://github.com/Automattic/jetpack/blob/84fbaa1bfda0641fad6177d2a8346d1a7144d4c4/projects/plugins/jetpack/tools/webpack.config.extensions.js#L187-L194

- Using above plugin to just copy this one SVG file
    ```js
	{
		from: 'subscriptions/lock.svg',
		to: './',
		context: path.join( __dirname, '../extensions' ),
	},
    ```
- Loading the file contents in PHP directly with 
    ```php
	$filesystem = new WP_Filesystem_Direct( true );
	$lock_svg_path = plugins_url( 'extensions/blocks/subscriptions/lock.svg', JETPACK__PLUGIN_FILE );
	$lock_svg = $filesystem->get_contents( $lock_svg_path );
    ```

...but concluded to just move the file to existing `images` folder where things work as-is I think.

cc @jeherve if you had opinion on how to handle svgs in blocks that aren't imported in JS/CSS.

simison:
I updated the subheading copy to latest iteration, and applied new padding styles to the group block that are more fitting across desktop & mobile:

Screenshot 2023-08-02 at 12 57 42
Screenshot 2023-08-02 at 12 57 34

@crisbusquets I'm trying to keep the paywall markup as such that it can be edited in block editor (i.e. no external CSS), so that we could one day turn it into something customers can edit on their own. It also tends to work better with theme styles that way. Block editor doesn't support separate spacing for desktop/mobile yet, so I'm applying the same padding for both desktop/mobile now. Seems to work quite well — feel free to grab below block markup and make any adjustments in the editor!

Screenshot 2023-08-02 at 13 02 50


```html
<!-- wp:group {"style":{"border":{"width":"1px","radius":"4px"},"spacing":{"padding":{"top":"var:preset|spacing|70","bottom":"var:preset|spacing|70","left":"32px","right":"32px"}}},"borderColor":"primary","className":"jetpack-subscribe-paywall","layout":{"type":"constrained","contentSize":"400px"}} -->
<div class="wp-block-group jetpack-subscribe-paywall has-border-color has-primary-border-color" style="border-width:1px;border-radius:4px;padding-top:var(--wp--preset--spacing--70);padding-right:32px;padding-bottom:var(--wp--preset--spacing--70);padding-left:32px"><!-- wp:image {"align":"center","width":24,"height":24,"sizeSlug":"large","linkDestination":"none"} -->
<figure class="wp-block-image aligncenter size-large is-resized"><img src="#" alt="" style="width:24px;height:24px" width="24" height="24"/></figure>
<!-- /wp:image -->

<!-- wp:heading {"textAlign":"center","style":{"typography":{"fontStyle":"normal","fontWeight":"600","fontSize":"24px"},"layout":{"selfStretch":"fit"}}} -->
<h2 class="wp-block-heading has-text-align-center" style="font-size:24px;font-style:normal;font-weight:600">This post is for subscribers</h2>
<!-- /wp:heading -->

<!-- wp:paragraph {"align":"center","style":{"typography":{"fontSize":"14px"},"spacing":{"margin":{"top":"10px","bottom":"10px"}}}} -->
<p class="has-text-align-center" style="margin-top:10px;margin-bottom:10px;font-size:14px">Subscribe to %s to keep reading and receive all new posts in your inbox.</p>
<!-- /wp:paragraph -->

<!-- wp:jetpack/subscriptions {"borderRadius":50,"borderColor":"primary","className":"is-style-compact"} /--></div>
<!-- /wp:group -->
```

projects/plugins/jetpack/extensions/blocks/subscriptions/index.js
lezama: > Creative. 😅

Not sure it even worked!

> ...but concluded to just move the file to existing images folder where things work as-is I think.

😂👍

simison: >Not sure it even worked!

We would've needed to pass the new filename+path to frontend somehow which is too much for so little gain.

lezama (approved):

PR #31886, opened 2023-07-14
CRM: Add hash router to Automations admin

robertf4:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3185

## Proposed changes:

This PR proposes using React Router as the client side routing solution for Jetpack CRM.

I selected React Router, and specifically a [Hash Router](https://reactrouter.com/en/main/routers/create-hash-router), as this is currently used in the main Jetpack plugin so other Jetpack developers should be familiar with this.

While the Jetpack plugin uses React Router 5, I chose to use React Router 6 as it is the most updated version and adds [a data API which may be useful to us](https://reactrouter.com/en/main/routers/picking-a-router#data-apis). Just for example it includes data loading functionality which serve the same purpose as the data components pattern while significantly reducing the boilerplate associated with that pattern. While I have not fully evaluated this API yet, I think it is a good idea to use the latest version and keep that option open.

The routing itself is pretty straightforward. New routes are added in `automations-admin/index.tsx`. I have included a few routes with hyperlinks to navigate between them just for the purpose of demonstration, but these should not be considered the final routes. When the relevant functionality is actually developed then we should reconsider these routes and ensure they are what we want. Note that this also applies to `AutomationsAdminContainer`—this component can be reevaluated as development progresses.

### Other information:

No

## Testing instructions:

1. Add `add_filter( 'jetpack_crm_feature_flag_automations', '__return_true' );` (e.g. to your `0-sandbox.php` file)
2. Go to `/wp-admin/admin.php?page=jpcrm-automations#/automations`.
3. Use the hyperlinks on the page to visit the edit and add pages. Verify that the URL changes to `#/automations/edit` and `#/automations/add`. Verify that the browser forward/back buttons work with this.
4. Visit any unspecified route such as `#/` or `#/automations/invalid`, and verify that you are redirected to `#/automations`


robertf4:
Note that I have currently set this PR to "DO NOT MERGE" as it is pointed to `add/crm/automations-admin-route` instead of `trunk`.

gogdzl:
Code checks out, looks good, but I'm not able to open the page using the instructions.

image

kallehauge:
> Code checks out, looks good, but I'm not able to open the page using the instructions.

The instructions needed the `add_filter( 'jetpack_crm_feature_flag_automations', '__return_true' );` section which probably is what was missing here 😄 

Would you mind following them again @gogdzl?

kallehauge (approved):
@gogdzl gave 👍 to how the code looks, and I've done manual testing to address his question, so I'm approving 🚀

PR #31887, opened 2023-07-14
Boost: Fix cloud CSS regeneration when module is toggled on

haqadn:
Fixes Automattic/boost-cloud#275

## Proposed changes:
* When a module is toggled, wait until the status is synced before dispatching an event
* On purchase success page, wait until the status is synced before starting to generate cloud CSS

### Other information:

## Jetpack product discussion

None

None

## Testing instructions:
* Keep the network tab on browser devtools open
* Upgrade boost, on the purchase success page, ensure request to the endpoint `/wp-json/jetpack-boost/v1/critical-css/start` isn't a 404.
* On boost settings page, toggle off and on cloud CSS module and make sure the cloud CSS job starts with a successful request to `/wp-json/jetpack-boost/v1/critical-css/start`.

### DIFF
--- a/projects/plugins/boost/app/assets/src/js/pages/purchase-success/PurchaseSuccess.svelte
+++ b/projects/plugins/boost/app/assets/src/js/pages/purchase-success/PurchaseSuccess.svelte
@@ -6,7 +6,7 @@
 	import ReactComponent from '../../elements/ReactComponent.svelte';
 	import TemplatedString from '../../elements/TemplatedString.svelte';
 	import { regenerateCriticalCss } from '../../stores/critical-css-state';
-	import { modulesState } from '../../stores/modules';
+	import { modulesState, modulesStatePending } from '../../stores/modules';
 	import Logo from '../../svg/jetpack-green.svg';
 	import externalLinkTemplateVar from '../../utils/external-link-template-var';
 
@@ -15,9 +15,17 @@
 	// svelte-ignore unused-export-let - Ignored values supplied by svelte-navigator.
 	export let location, navigate;
 
-	onMount( async () => {
+	onMount( () => {
+		// Enable cloud css module on upgrade.
 		$modulesState.cloud_css.active = true;
-		await regenerateCriticalCss();
+
+		// Ones the module is enabled and synced, regenerate critical css.
+		const unsubscribe = modulesStatePending.subscribe( pending => {
+			if ( ! pending ) {
+				regenerateCriticalCss();
+				unsubscribe();
+			}
+		} );
 	} );
 </script>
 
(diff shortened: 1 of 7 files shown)


projects/plugins/boost/app/assets/src/js/pages/purchase-success/PurchaseSuccess.svelte
```diff
@@ -15,9 +15,17 @@
...
 		$modulesState.cloud_css.active = true;
-		await regenerateCriticalCss();
+
+		// Ones the module is enabled and synced, regenerate critical css.
+		const unsubscribe = modulesStatePending.subscribe( pending => {
+			if ( ! pending ) {
+				regenerateCriticalCss();
+				unsubscribe();
+			}
+		} );
```
thingalon: Wouldn't it make more sense to make the plugin back-end automatically request fresh Cloud CSS when that module is enabled? We have an action on the back end `jetpack_boost_module_status_updated` which seems like it would be good for this purpose.

Then we don't have to do complex awaiting on the front end.

thingalon (changes requested):
I think there's a cleaner way to do this.

projects/plugins/boost/app/assets/src/js/pages/purchase-success/PurchaseSuccess.svelte
haqadn: I have addressed this in 2831905

projects/plugins/boost/app/assets/src/js/pages/settings/elements/Module.svelte
```diff
@@ -13,9 +13,18 @@
...
+		// Toggle the module.
+		$modulesState[ slug ].active = toggledState;
+
+		// Attach a listener to the modulesStatePending store to dispatch an event after the pending state is resolved.
+		const unsubscribe = modulesStatePending.subscribe( isPending => {
+			if ( ! isPending ) {
+				dispatch( eventName );
+				unsubscribe();
+			}
+		} );
```
thingalon: This code is more complex than it needs to be. It's in a svelte component, so we can just use a `$:` block to react to changes in a stored value - including pending state.

In d5d5518368 I replaced this subscribe-unsubscribe stuff with just a separate block that watches for changes to the store and emits when it's different and not pending.

It also means that if saving the new state of the toggle fails, it won't trigger a dispatch falsely indicating the state changed.

Take a look, see if you're happy with it.

thingalon (approved):
Looks good, and seems to work. Though I made a couple of improvements:

- Improved the way that pending state is handled (see inline comment), and
- Fixed code sniffer issue 3db6242770

haqadn:
Cherry-picked 0f557853b

PR #31894, opened 2023-07-14
Fix Forms dropdown required validation

jcheringer:
Fixes pdWQjU-n2-p2

## Proposed changes:

* Fix Forms dropdown required validation

### Other information:

## Jetpack product discussion
N/A

No

## Testing instructions:

* Create a Form and add a Dropdown field
* Make the dropdown required and publish the page
* Go to the published page and try to submit the form without selecting a value for the  dropdown
* You should see a validation error message and the form shouldn't be submitted

### DIFF
--- a/projects/packages/forms/src/contact-form/class-contact-form-field.php
+++ b/projects/packages/forms/src/contact-form/class-contact-form-field.php
@@ -204,7 +204,6 @@ class Contact_Form_Field extends Contact_Form_Shortcode {
 	 * Validates the form input
 	 */
 	public function validate() {
-		error_log( print_r( $this, true ) );
 		// If it's not required, there's nothing to validate
 		if ( ! $this->get_attribute( 'required' ) ) {
 			return;
@@ -717,7 +716,7 @@ class Contact_Form_Field extends Contact_Form_Shortcode {
 	 */
 	public function render_select_field( $id, $label, $value, $class, $required, $required_field_text ) {
 		$field  = $this->render_label( 'select', $id, $label, $required, $required_field_text );
-		$field .= "\t<select name='" . esc_attr( $id ) . "' id='" . esc_attr( $id ) . "' " . $class . ( $required ? "required aria-required='true'" : '' ) . "aria-hidden='true' tabindex='-1'>\n";
+		$field .= "\t<select name='" . esc_attr( $id ) . "' id='" . esc_attr( $id ) . "' " . $class . ( $required ? "required aria-required='true'" : '' ) . ">\n";
 
 		if ( $this->get_attribute( 'togglelabel' ) ) {
 			$field .= "\t\t<option value=''>" . $this->get_attribute( 'togglelabel' ) . "</option>\n";
--- a/projects/packages/forms/src/contact-form/class-contact-form-field.php
+++ b/projects/packages/forms/src/contact-form/class-contact-form-field.php
@@ -717,7 +717,7 @@ class Contact_Form_Field extends Contact_Form_Shortcode {
 	 */
 	public function render_select_field( $id, $label, $value, $class, $required, $required_field_text ) {
 		$field  = $this->render_label( 'select', $id, $label, $required, $required_field_text );
-		$field .= "\t<select name='" . esc_attr( $id ) . "' id='" . esc_attr( $id ) . "' " . $class . ( $required ? "required aria-required='true'" : '' ) . ">\n";
+		$field .= "\t<select name='" . esc_attr( $id ) . "' id='" . esc_attr( $id ) . "' " . $class . ( $required ? "required aria-required='true'" : '' ) . "aria-hidden='true' tabindex='-1'>\n";
 
 		if ( $this->get_attribute( 'togglelabel' ) ) {
 			$field .= "\t\t<option value=''>" . $this->get_attribute( 'togglelabel' ) . "</option>\n";
--- a/projects/packages/forms/src/contact-form/class-contact-form-field.php
+++ b/projects/packages/forms/src/contact-form/class-contact-form-field.php
@@ -204,6 +204,7 @@ class Contact_Form_Field extends Contact_Form_Shortcode {
 	 * Validates the form input
 	 */
 	public function validate() {
+		error_log( print_r( $this, true ) );
 		// If it's not required, there's nothing to validate
 		if ( ! $this->get_attribute( 'required' ) ) {
 			return;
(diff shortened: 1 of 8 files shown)


projects/packages/forms/src/contact-form/class-contact-form-field.php
```diff
@@ -719,7 +720,7 @@ public function render_select_field( $id, $label, $value, $class, $required, $re
 		$field .= "\t<select name='" . esc_attr( $id ) . "' id='" . esc_attr( $id ) . "' " . $class . ( $required ? "required aria-required='true'" : '' ) . ">\n";
```
anomiex, line 719: You should probably add `tabindex='-1'` in here. Otherwise tabbing through the form is confusing as it focuses this hidden field before the visible dropdown.

Also, do we need to do anything to prevent screen readers from reading it?

anomiex (commented):
Code in [projects/plugins/jetpack/modules/contact-form/grunion-contact-form.php](https://github.com/Automattic/jetpack/blob/fix/forms-dropdown-validation/projects/plugins/jetpack/modules/contact-form/grunion-contact-form.php#L5025-L5029) and [projects/plugins/jetpack/modules/contact-form/css/grunion.css](https://github.com/Automattic/jetpack/blob/fix/forms-dropdown-validation/projects/plugins/jetpack/modules/contact-form/css/grunion.css) likely needs the same changes.

jcheringer:
> Code in [projects/plugins/jetpack/modules/contact-form/grunion-contact-form.php](https://github.com/Automattic/jetpack/blob/fix/forms-dropdown-validation/projects/plugins/jetpack/modules/contact-form/grunion-contact-form.php#L5025-L5029) and [projects/plugins/jetpack/modules/contact-form/css/grunion.css](https://github.com/Automattic/jetpack/blob/fix/forms-dropdown-validation/projects/plugins/jetpack/modules/contact-form/css/grunion.css) likely needs the same changes.

Is there any placing not using the Forms package by default?
It has been quite some time since we released the Forms package, the idea would be to stop maintaining the forms in both places at some point.

projects/packages/forms/src/contact-form/class-contact-form-field.php
jcheringer: Well spotted! ✅

projects/packages/forms/src/contact-form/class-contact-form-field.php
```diff
@@ -717,7 +717,7 @@ public function render_checkbox_multiple_field( $id, $label, $value, $class, $re
 	 */
 	public function render_select_field( $id, $label, $value, $class, $required, $required_field_text ) {
 		$field  = $this->render_label( 'select', $id, $label, $required, $required_field_text );
-		$field .= "\t<select name='" . esc_attr( $id ) . "' id='" . esc_attr( $id ) . "' " . $class . ( $required ? "required aria-required='true'" : '' ) . ">\n";
+		$field .= "\t<select name='" . esc_attr( $id ) . "' id='" . esc_attr( $id ) . "' " . $class . ( $required ? "required aria-required='true'" : '' ) . "aria-hidden='true' tabindex='-1'>\n";
```
anomiex: Looks like there will be a space missing, we'll get `aria-required='true'aria-hidden='true'`. With the way the rest of the string is built, we probably want to do like this
```suggestion
		$field .= "\t<select name='" . esc_attr( $id ) . "' id='" . esc_attr( $id ) . "' " . $class . ( $required ? "required aria-required='true' " : '' ) . "aria-hidden='true' tabindex='-1'>\n";
```

projects/packages/forms/src/contact-form/class-contact-form-field.php
```diff
@@ -204,6 +204,7 @@ public function is_error() {
 	 * Validates the form input
 	 */
 	public function validate() {
+		error_log( print_r( $this, true ) );
```
anomiex: The CI sniff is correct, we don't want this in here.
```suggestion
```

projects/packages/forms/src/contact-form/class-contact-form-field.php
anomiex: Also, thinking about this a bit more... Perhaps we shouldn't put the tabindex and aria-hidden here in the HTML after all, as that will break if someone is using NoScript or the like. Perhaps instead we should add the attributes in [the JS that hides the field](https://github.com/Automattic/jetpack/blob/fix/forms-dropdown-validation/projects/packages/forms/src/contact-form/js/dropdown.js).

Probably the added CSS should also only apply after the JS has run, e.g. by having the JS add a class or make the selector like `.contact-form .contact-form-dropdown[aria-hidden="true"]` or something.

anomiex (commented):
> It has been quite some time since we released the Forms package, the idea would be to stop maintaining the forms in both places at some point.

IMO getting rid of the old copy would be great, but as long as both copies exist we should keep them both up.

projects/packages/forms/src/contact-form/class-contact-form-field.php
jcheringer: ✅

projects/packages/forms/src/contact-form/class-contact-form-field.php
jcheringer: Whoops! 😬

projects/packages/forms/src/contact-form/js/dropdown.js
```diff
@@ -19,5 +19,8 @@ jQuery( function ( $ ) {
 				'ui-selectmenu-menu': 'contact-form-dropdown__menu',
 			},
 		} );
+
+		$( '.contact-form .contact-form-dropdown' ).attr( 'aria-hidden', true );
+		$( '.contact-form .contact-form-dropdown' ).prop( 'tabindex', -1 );
```
anomiex: Might be more efficient to chain the calls instead of repeatedly evaluating the selector
```suggestion
		} )
		.attr( 'aria-hidden', true )
		.prop( 'tabindex', -1 );
```

projects/plugins/jetpack/modules/contact-form/js/dropdown.js
```diff
@@ -19,5 +19,8 @@ jQuery( function ( $ ) {
 				'ui-selectmenu-menu': 'contact-form-dropdown__menu',
 			},
 		} );
+
+		$( '.contact-form .contact-form-dropdown' ).attr( 'aria-hidden', true );
+		$( '.contact-form .contact-form-dropdown' ).prop( 'tabindex', -1 );
```
anomiex: Same
```suggestion
		} )
		.attr( 'aria-hidden', true )
		.prop( 'tabindex', -1 );
```

anomiex (approved):
LGTM!

PR #31905, opened 2023-07-15
Launchpad: Renamed keep-building tasklist

nelsonec87:
Related to https://github.com/Automattic/wp-calypso/pull/79461

## Proposed changes:

* Add a copy of the `keep-building` task list with new name `intent-write` to follow the naming conventions we want to switch to.
* Don't rename the gate function `wpcom_launchpad_is_keep_building_enabled` since this is also used on the WP.com side; renaming it will break the My Home views on WP.com.
* This ensures that Atomic sites can still access the old checklist during the "in-between" phase where jetpack-mu-wpcom is synced to WP.com Simple sites but not Atomic ones.

## Testing instructions:

See https://github.com/Automattic/wp-calypso/pull/79461

No.

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/changelog/rename-launchpad-tasklist
+++ b/projects/packages/jetpack-mu-wpcom/changelog/rename-launchpad-tasklist
@@ -1,0 +1,4 @@
+Significance: minor
+Type: changed
+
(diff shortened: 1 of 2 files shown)


markbiek (approved):
Working as expected on Simple and Atomic!

PR #31907, opened 2023-07-17
Add was_migration_trial flag for wpcom sites

zaerl:
This PR adds the `was_migration_trial` flag to the site data we return from the site REST APIs. This value will always be false for normal Jetpack sites, as the Migration trial is only available from WordPress.com.

Fixes https://github.com/Automattic/dotcom-forge/issues/2977

## Proposed changes:
1. Add the `was_migration_trial` flag to the site data
2. Add tests for `ecommerce` and `migration` trials

### Other information:

## Testing instructions:

* `jetpack docker phpunit -- --filter=WP_Test_Jetpack_Site_Json_Api_Endpoints --debug --verbose`
* Follow instruction found [here](https://github.com/Automattic/jetpack/pull/31907#issuecomment-1637606726)
* Install `add/was-migration-trial-flag-for-wpcom-sites` beta
* Open https://developer.wordpress.com/docs/api/console/
* Test `/sites/__ID__` and check for `was_ecommerce_trial` and `was_migration_trial` fields

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -807,16 +807,16 @@
         },
         {
             "name": "phpcsstandards/phpcsutils",
-            "version": "1.0.7",
+            "version": "1.0.8",
             "source": {
                 "type": "git",
                 "url": "https://github.com/PHPCSStandards/PHPCSUtils.git",
-                "reference": "886a728571ab2faca3642c96954886b078b32195"
+                "reference": "69465cab9d12454e5e7767b9041af0cd8cd13be7"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/PHPCSStandards/PHPCSUtils/zipball/886a728571ab2faca3642c96954886b078b32195",
-                "reference": "886a728571ab2faca3642c96954886b078b32195",
+                "url": "https://api.github.com/repos/PHPCSStandards/PHPCSUtils/zipball/69465cab9d12454e5e7767b9041af0cd8cd13be7",
+                "reference": "69465cab9d12454e5e7767b9041af0cd8cd13be7",
                 "shasum": ""
             },
             "require": {
@@ -876,7 +876,7 @@
                 "issues": "https://github.com/PHPCSStandards/PHPCSUtils/issues",
                 "source": "https://github.com/PHPCSStandards/PHPCSUtils"
             },
-            "time": "2023-07-10T13:56:07+00:00"
+            "time": "2023-07-16T21:39:41+00:00"
         },
         {
             "name": "sirbrillig/phpcs-changed",
--- a/projects/plugins/jetpack/json-endpoints/class.wpcom-json-api-get-site-endpoint.php
+++ b/projects/plugins/jetpack/json-endpoints/class.wpcom-json-api-get-site-endpoint.php
@@ -279,8 +279,8 @@ class WPCOM_JSON_API_GET_Site_Endpoint extends WPCOM_JSON_API_Endpoint {
 	 * @var array $jetpack_enabled_trials
 	 */
 	public static $jetpack_enabled_trials = array(
-		'ecommerce',
-		'migration',
+		'ecommerce' => 'was_ecommerce_trial',
+		'migration' => 'was_migration_trial',
 	);
 
 	/**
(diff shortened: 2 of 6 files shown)


composer.lock
vishnugopal, line 1: This really shouldn't change in this PR right?

projects/plugins/jetpack/json-endpoints/class.wpcom-json-api-get-site-endpoint.php
```diff
@@ -579,7 +591,10 @@ protected function render_response_key( $key, &$response, $is_user_logged_in ) {
 				$response[ $key ] = $this->site->get_p2_thumbnail_elements();
 				break;
 			case 'was_ecommerce_trial':
-				$response[ $key ] = $this->site->was_ecommerce_trial();
+				$response[ $key ] = $this->site->was_trial( self::$jetpack_enabled_trials[0] );
```
vishnugopal: Small nit: This `[0]` `[1]` indexing looks a bit odd to me, and what if folks add the third trial in between in the list?

Can we make `$jetpack_enabled_trials` a proper dictionary with keys being constants and check for that?

projects/plugins/jetpack/tests/php/json-api/test-class.json-api-jetpack-site-endpoints.php
```diff
@@ -53,7 +63,44 @@ public function test_get_site() {
...
+		$response = $endpoint->callback( '', $blog_id );
+
+		$this->assertTrue( $response['jetpack'] );
+		$this->assertTrue( $response['jetpack_connection'] );
+
+		$options = (array) $response['options'];
+		$this->assertArrayHasKey( 'jetpack_connection_active_plugins', $options );
+	}
+
+	/**
```
vishnugopal, line 77: This comment needs revision.

projects/plugins/jetpack/tests/php/json-api/test-class.json-api-jetpack-site-endpoints.php
```diff
@@ -53,7 +63,44 @@ public function test_get_site() {
...
+
+		$admin = self::factory()->user->create_and_get(
+			array(
+				'role' => 'administrator',
+			)
+		);
+
+		wp_set_current_user( $admin->ID );
+
+		$endpoint = $this->create_get_site_endpoint();
```
vishnugopal, line 97: Is it possible to add a truthy test here as well? I'm still learning the ropes here but I believe it's possible to create a site with a migration trial now.

vishnugopal (commented):
Some minor nits!

composer.lock
zaerl: It's automatically changed once installed. If I remember well, the change will be pruned during deploy.

projects/plugins/jetpack/json-endpoints/class.wpcom-json-api-get-site-endpoint.php
zaerl: Ok, sure.

composer.lock
vishnugopal: Ack understood, thanks!

projects/plugins/jetpack/tests/php/json-api/test-class.json-api-jetpack-site-endpoints.php
zaerl: It checks `false` because usually in Jetpack unit tests, all the functions that are wpcom-specific return `false` and nothing more. In this case, `has_blog_sticker`.

But I can change the function and make it more smart.

zaerl:
@vishnugopal 
1. Changed `WPCOM_JSON_API_GET_Site_Endpoint::$jetpack_enabled_trials` to an associative array
2. The unit test now is more descriptive and has the `@cover` parameter
3. Changed the mock function to be more "smart" and do not simply return `false`. It uses the options system; some arbitrary data are saved during the tests. So that we can `assertTrue/False`. This will be used automatically in the future once new trials are added.

vishnugopal:
@zaerl Thank you! I'm still in the middle of setting up Jetpack again, so I can't test this PR functionality. Will do once I've set it up.

vishnugopal:
Works now, approving!
Screenshot 2023-07-18 at 3 01 34 PM

vishnugopal (approved):

PR #23966, opened 2022-04-15
CLI: Enforce monorepo composer version

sdixon194:
Take 2 of the previous PR that was reverted because it would fatal if composer wasn't installed. This updated version checks that a version of composer is returned first and bails if it doesn't receive one.

Using monorepo tooling with the wrong composer version can cause composer.lock issues which don't get surfaced until CI checks them. This PR will enforce the composer version being used with the CLI to make sure devs are using the version we need to prevent broken PRs. This PR also lays some groundwork in case we want to do additional kinds of checks in the future.

Check seems performant enough to run on every usage of the CLI, but if it's a problem, we could change when it's used. 
#### Changes proposed in this Pull Request:

Checks the dev environment's composer version against the monorepo required version before running any `jetpack` CLI command, along with instructions on how to fix. *

#### Jetpack product discussion

[N/a](p1649860244940259-slack-C01U2KGS2PQ)
#### Does this pull request change what data or activity we track or use?

No
#### Testing instructions:

* Pull down this PR
* Run `composer self-update` to update composer to the latest version (which isn't supported)
* Try running a CLI command, like `jetpack cli status`
* The command should fail and tell you how to fix your composer version.
* Run the fix suggested.
* Try the command again, it should work fine.*

PR #31918, opened 2023-07-17
cli: Check node, php, and pnpm versions too

anomiex:
## Proposed changes:

Have the CLI check node, PHP, and pnpm versions on startup as well as the Composer version.

And also better handle the case where the program is not installed.

Also fix the handler for "forgot to `pnpm install`", it wasn't working since #23966 added some unguarded `import`s in the `bin/jetpack.js` entry point.

### Other information:

## Jetpack product discussion

p1688649938321989-slack-C05AZRB6RNZ

No

## Testing instructions:

* Install wrong versions of various tools (or change `.github/versions.sh` to indicate current versions are wrong) and run the Jetpack CLI. Observe expected error messages.
* Remove some tools entirely (or change your PATH such that they won't be found). This should be reported in a useful manner.
* Remove `tools/cli/node_modules/` and try to run the CLI. Error message should be helpful.

### DIFF
--- a/tools/cli/bin/jetpack.js
+++ b/tools/cli/bin/jetpack.js
@@ -13,27 +13,23 @@ async function guardedImport( path ) {
 	try {
 		return await import( path );
 	} catch ( error ) {
-		const identity = v => v;
-		const chalk = await import( 'chalk' ).then(
-			m => m.default,
-			() => ( {} )
-		);
+		const bold =
+			( await import( 'chalk' ).then(
+				m => m.default?.stderr?.bold,
+				() => null
+			) ) || ( v => v );
 
+		console.error( error );
+		console.error( '' );
 		if ( error.code === 'ERR_MODULE_NOT_FOUND' ) {
 			// if pnpm install hasn't been run, the import() will fail here.
 			console.error(
-				'Something is missing from your install. Please run `pnpm install` and try again.'
-			);
-
-			// Print the original error's message too, as sometimes the error isn't fixed by `pnpm install`.
-			console.error(
-				( chalk.grey || identity )( `The original error message was ${ error.message }` )
+				bold(
+					'*** Something is missing from your install. Please run `pnpm install` and try again. ***'
+				)
 			);
 		} else {
-			console.error( error );
-			console.error(
-				( chalk.bold || identity )( 'Something unexpected happened. See error above.' )
-			);
+			console.error( bold( '*** Something unexpected happened. See error above. ***' ) );
 		}
 		process.exit( 1 );
 	}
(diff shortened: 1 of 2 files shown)


tools/cli/bin/jetpack.js
```diff
@@ -2,24 +2,52 @@
...
+		const identity = v => v;
+		const chalk = await import( 'chalk' ).then(
+			m => m.default,
+			() => ( {} )
+		);
+
+		if ( error.code === 'ERR_MODULE_NOT_FOUND' ) {
+			// if pnpm install hasn't been run, the import() will fail here.
+			console.error(
+				'Something is missing from your install. Please run `pnpm install` and try again.'
```
zinigor: Since at this point we're certain that the `pnpm install` should be run, let's output this text after the original error message? I know for certain that people tend to read the last line only, I know I do at least. That should get most people through the problem without having to deal with the original message.

tools/cli/bin/jetpack.js
```diff
@@ -2,24 +2,52 @@
...
+			);
+
+			// Print the original error's message too, as sometimes the error isn't fixed by `pnpm install`.
+			console.error(
+				( chalk.grey || identity )( `The original error message was ${ error.message }` )
```
zinigor: This looks like very poor formatting when it happens in my testing:

```
The original error message was Cannot find package 'chalk' imported from /home/zinigor/workspace/jetpack/tools/cli/helpers/checkEnvironment.js
```
Can we quote the original message somehow? Even if we just put a colon and a newline on the end of our text?

tools/cli/bin/jetpack.js
```diff
@@ -29,34 +57,21 @@ try {
...
-		console.error(
-			( chalk.grey || identity )( `The original error message was ${ error.message }` )
-		);
-	} else {
-		console.error( error );
-		console.error(
-			( chalk.bold || identity )( 'Something unexpected happened. See error above.' )
-		);
+	if ( ! ( await compareToolVersions() ) && ! process.env.JETPACK_CLI_IGNORE_VERSION_CHECKS ) {
+		console.error( '(Ignore the above errors by setting JETPACK_CLI_IGNORE_VERSION_CHECKS=1)' );
```
zinigor: I don't know if we should show the errors anyway if the user has decided to ignore them? But that's more of a question than a request for change.

zinigor (changes requested):
Works well for me, but I'm asking for a few minor changes if you don't mind. Thanks!

tools/cli/bin/jetpack.js
anomiex: If we're going to print it before the prompt to `pnpm install`, we may as well just print it without elaboration.

tools/cli/bin/jetpack.js
anomiex: > I know for certain that people tend to read the last line only, I know I do at least.

So, _some_ people.

tools/cli/bin/jetpack.js
anomiex: Personally I'd rather they be there to remind the user about this even if they don't want to fix them right now. But maybe "ignore" is a poor word.

zinigor (approved):

PR #31930, opened 2023-07-18
Comments: add filter for comment title tag

### LINKED ISSUE #31913: Comments: add option to customize the comment reply heading level
### Impacted plugin

Jetpack

### What

Allow the user to choose the tag for the comment section, currently hardcoded as H3

### How

Hi, 

I'm working to improve the accessibility of my pages and facing an issue as the section fo the comments uses H3 tag and in my pages I do not have any H2 tag. I would like if you could allow the user to choose in the setting which tag to be used. 
In Jetpack -> Settings -> Discussions
you can cusomize the "Comment form introduction" but not the tag that it is hardcoded as H3. I would like to use a normal paragraph <p> where to apply my style

jeherve:
Fixes #31913

## Proposed changes:

Let's add a new filter for folks who would rather not use `h3` to display the comment title tag on their site.

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

* Go to Jetpack > Settings > Discussion and enable Jetpack Comments.
* View a post on the frontend, note how the comment title looks like.
* In a functionality plugin, add the following:

```php
add_filter(
     'jetpack_comment_reply_title_tag',
     function () {
         return 'h4';
     }
);
```

* View the post again; the comment form should look the same, but the comment title should be a little smaller and the heading tag should now be `h4`.


projects/plugins/jetpack/modules/comments/comments.php
```diff
@@ -387,17 +387,43 @@ public function comment_form_after() {
...
-			<?php if ( true === $show_greeting ) : ?>
-				<h3 id="reply-title" class="comment-reply-title"><?php comment_form_title( esc_html( $params['greeting'] ), esc_html( $params['greeting_reply'] ) ); ?>
-					<small><?php cancel_comment_reply_link( esc_html__( 'Cancel reply', 'jetpack' ) ); ?></small>
-				</h3>
-			<?php endif; ?>
+			<?php
+			if ( true === $show_greeting ) :
+				printf(
+					'<%1$s id="reply-title" class="comment-reply-title">',
+					esc_attr( $comment_reply_title_tag )
```
zinigor: Since we're replacing a tag name, not a property value, should this be `esc_html` instead? I tried searching for a bit, but couldn't find the difference between the two besides the filter they are calling. Maybe you know which one is more appropriate here.

jeherve: Both would work here, really, since there is no specific one for tag names. `esc_attr` is appropriate for use inside attributes, `esc_html` will remove all HTML entirely so is good for strings. In this specific case I do not think it really matters, but happy to switch to `esc_html`. 👍

jeherve: Done f3eab8cb21437ea40c3d3ea4e9893a0507025948

zinigor (approved):

PR #31931, opened 2023-07-18
Boost: Add ISA scanned count

haqadn:
Fixes Automattic/boost-cloud#319

## Proposed changes:

* Include the number of pages scanned if we are unable to scan all the pages

### Other information:

## Jetpack product discussion
None

No

## Testing instructions:
* Create a site with more than 100 pages.
* Make sure you see the updated messages

Screenshot 2023-07-18 at 2 54 13 PM
Screenshot 2023-07-18 at 2 57 34 PM

### DIFF
--- a/projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/RecommendationsMeta.svelte
+++ b/projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/RecommendationsMeta.svelte
@@ -12,6 +12,8 @@
 		initializeIsaSummary,
 		isaSummary,
 		ISAStatus,
+		totalPagesCount,
+		scannedPagesCount,
 	} from './store/isa-summary';
 
 	onMount( () => {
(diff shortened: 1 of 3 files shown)


projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/RecommendationsMeta.svelte
```diff
@@ -49,6 +51,29 @@
 		( $isaSummary?.status === ISAStatus.Queued &&
 			__( 'Give us a few minutes while we go through your images…', 'jetpack-boost' ) );
 
+	$: foundIssuesMessage =
+		$totalPagesCount === 100
```
dilirity: Are we only showing the detailed message if there are exactly 100 scanned pages? Shouldn't it be "show it on 100 and above"? That way, when we decide to up the limit, we won't need to tweak the UI to allow it.

dilirity (commented):
Looks good. Though I'm not sure on the wording (though I can't come up with something better myself 😅).

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/RecommendationsMeta.svelte
haqadn: The `totalPagesCount` currently shows the number of pages we are aiming to scan. My thoughts here is that, if we lift the limit to a higher number, i.e. virtually unlimited, we don't want to continue showing the number of pages scanned. For now, we only know that we are rate-limiting at exactly 100. If the check was `totalPagesCount > 100` it would not be clear if there is a rate limit or not.

But I do know that this is not ideal, and I am open to suggestions. I wish we had a way to detect if we enforced a limit. May be we should update the response from wpcom?

cc: @thingalon

thingalon: I don't see why we need to detect whether the internal page limit was hit or not. Wouldn't it be better to just say the number of pages that we scan, and ensure our documentation covers what our current limits are?

thingalon (approved):

haqadn:
Cherry-picked c02a8c3a994

PR #31941, opened 2023-07-18
Boost: Fix ISA details page not loading if an image from the batch was not ok

dilirity:
This was discovered while investigating https://github.com/Automattic/boost-cloud/issues/321 There were instances where the detailed report page wouldn't load, with several error 500 from the `wp-json/jetpack-boost-ds/image-size-analysis/set` endpoint. I couldn't find a consistent way of reproducing this.

<details><summary>screenshot</summary>

</details>

## Proposed changes:

* Fixes ISA details page not loading.
* Also fixes the 'bad entry' listing item not showing, due to getting filtered out because of a missing status.

### Other information:

## Jetpack product discussion

n/a

no

## Testing instructions:

During testing, I noticed that an SVG image didn't have width in the database (the value was empty). Since the schema requires a number, in order to consistently reproduce this and break it, you can just delete the value of `fileSize_weight` (from the meta table `boost_report_issue_meta`) for an image.

### Test that it breaks

* Setup Boost with ISA and make sure you have some pages with images;
* Break an image following the instructions above;
* Request a report;
* Check the network tab, to make sure there's an error 500 for the `set` endpoint;
* Wait for the UI to finish polling;
* No issues will load.

### Test that it works

* Apply the fix;
* Request a report;
* Check the network tab, to make sure the `set` endpoint returns 200;
* The UI should properly load and display a "broken item" element.

CleanShot 2023-07-18 at 16 45 05

### DIFF
--- a/projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/store/zod-types.ts
+++ b/projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/store/zod-types.ts
@@ -37,6 +37,7 @@ export const ImageData = z
 	.catch( {
 		id: '',
 		type: 'bad_entry',
+		status: 'active', // We still want to show the UI for this.
 	} );
 
 export type ImageDataType = z.infer< typeof ImageData >;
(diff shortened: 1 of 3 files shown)


thingalon (approved):
Changes look good 👍

haqadn:
Cherry-picked 01330b2a

PR #31950, opened 2023-07-18
e2e tests: Remove use of `pm2`

anomiex:
## Proposed changes:

Getting rid of `pm2` lets us drop several abandoned JS deps, some with known security vunlerabilities.

Also the various unused `tunnel:logs` and `tunnel:write-logs` scripts have been removed, for the same reason.

### Other information:

## Jetpack product discussion

p1689361107348039-slack-C03QNBQKG73

No

## Testing instructions:

* Are the E2E tests running and happy? Does the job output look sensible?


anomiex:
Works ok, but there's the possibility of a bit of a race if the tunnel is slow to open. I should handle that better, have `tunnel:up` exit after opening the tunnel like it used to.

zinigor (approved):
Nice! I wasn't able to test locally because I don't know how to set up the same environment on my machine, but it's great to see a big dependency go away replaced by two functions!

PR #31956, opened 2023-07-19
Whitelist `jetpack_memberships_site_subscriber` post meta

romarioraffington:
## Proposed changes:

Sync `jetpack_memberships_site_subscriber` post meta to WPCOM as we rely on this post meta for business logic

### Other information:

No

## Testing instructions:

* Create a newsletter plan on a JP site that has a JP connection
* Verify that the post and `jetpack_memberships_site_subscriber` post meta is synced to WPCOM

### DIFF
--- a/projects/plugins/jetpack/changelog/add-whitelist-site-subscriber-post-meta
+++ b/projects/plugins/jetpack/changelog/add-whitelist-site-subscriber-post-meta
@@ -1,0 +1,4 @@
+Significance: minor
+Type: bugfix
+
+Whitelist jetpack_memberships_site_subscriber post meta to allow for syncing to WordPress.com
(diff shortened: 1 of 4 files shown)


jeherve (commented):
Since we're making that change, should this be a new constant in `projects/plugins/jetpack/extensions/blocks/subscriptions/constants.php`

millerf:
I really struggle to see how we have been able to test all previous times on JN without this on. I believe this is a red herring...

n3f (commented):
I think a better solution is probably the one outlined [here](p1689712281078529/1689701094.306599-slack-CBG1CP4EN).  It seems that if you add it to the class-jetpack-memberships file it should get taken care of with the existing meta that's getting populated.

```php
	private static function get_plan_property_mapping() {
		$meta_prefix = 'jetpack_memberships_';
		$properties  = array(
			'price'    => array(
				'meta' => $meta_prefix . 'price',
			),
			'currency' => array(
				'meta' => $meta_prefix . 'currency',
			),
			'is_newsletter' => array(
				'meta' => $meta_prefix . 'site_subscriber',
			),
		);
		return $properties;
	}
```

romarioraffington:
Thanks Brent, added your suggestion

millerf (approved):
LGTM

millerf:
I verified:
- paid newsletters are not sent to paid subscribers on Atomic (and I think, also standalone Jetpack).
- unloggued subscribers cannot log via the subscribe block onAtomic and Jetpack

PR #31962, opened 2023-07-19
Jetpack: Have webpack embed social-logos.woff2 in CSS

anomiex:
## Proposed changes:

Having to manually construct the data URI caused some breakage recently. Let's have Webpack do the work instead.

### Other information:

## Jetpack product discussion

p9dueE-7DU-p2#comment-9905

No

## Testing instructions:

* Built `_inc/social-logos/social-logos.min.css` looks ok?
* Check that the stuff using this CSS rule still works.
  * Use the built artifact, not a local build, to ensure everything is in the artifact correctly.
  * Check all supported browsers to ensure nothing has a problem with the mime type in the data URI being "font/woff2" rather than "application/octet-stream" like it was before.


projects/plugins/jetpack/tools/webpack.config.css.js
```diff
@@ -64,12 +64,27 @@ const sharedWebpackConfig = {
...
+					{
+						resourceQuery: /inline/,
+						type: 'asset/inline',
+						/* If we need to force application/octet-stream, add this:
+						generator: {
+							dataUrl: {
+								mimetype: 'application/octet-stream',
+							},
+						},
+						*/
```
anomiex, line 78: This should either be removed or uncommented before merging.

anomiex:
Closing this to clean up lists, it can be reopened if that would be useful.

PR #31965, opened 2023-07-19
My Jetpack: register jetpack-ai-jwt endpoint

### LINKED ISSUE #31938: AI Client: new registration location for AI endpoints
Currently, some endpoints are registered within the Jetpack plugin, a setup that hinders our goal of achieving complete independence for the client.
We need to identify a more suitable location for registering these endpoints to promote the autonomy of the AI Client.

retrofox:
This PR registers the `jetpack-ai-jwt` endpoint in the `my-jetpack/v1` namespace.
Currently, the [endpoint is registered by the Jetpack plugin](https://github.com/Automattic/jetpack/blob/7785086b2b4337f6db1bef8b1d2731154ba250d9/projects/plugins/jetpack/_inc/lib/class.core-rest-api-endpoints.php#L80), meaning it's only available when it's activated on the site which can be considered a blocker for those projects that are independent of the plugin and want to use it.

Moving the endpoint to My Jetpack guarantees its availability at least in the Jetpack environment. Almost all Jetpack products depend on My Jetpack. For instance, it would be possible to hit the endpoint from VideoPress or Social self-hosted plugins.

_**But it's worth mentioning that it could be considered a temporary solution, too.**_

Fixes https://github.com/Automattic/jetpack/issues/31938

## Proposed changes:

* My Jetpack: register jetpack-ai-jwt endpoint

## Testing instructions:

Since the endpoint is not being used yet, the testing steps are quite simple.

* Confirm the endpoint register in the `my-jetpack/v1` namespace. You can list all endpoints by using the `wp-json` URL. For instance, in your local dev-env:

Screenshot 2023-07-19 at 11 59 56

* You can try to hit the endpoint by using the dev console tool.
* Ensure replacing `<your-site-here>` for the site hostname:

```
fetch("http://<your-site-here>/wp-json/my-jetpack/v1/jetpack-ai-jwt?_cacheBuster=1689778884847&_locale=user", {
  "headers": {
    "accept": "application/json, */*;q=0.1",
    "accept-language": "en,es;q=0.9,es-419;q=0.8,it;q=0.7,pt;q=0.6,fr;q=0.5,nb;q=0.4,pl;q=0.3",
    "sec-ch-ua": "\"Not.A/Brand\";v=\"8\", \"Chromium\";v=\"114\", \"Google Chrome\";v=\"114\"",
    "sec-ch-ua-mobile": "?0",
    "sec-ch-ua-platform": "\"macOS\"",
    "sec-fetch-dest": "empty",
    "sec-fetch-mode": "cors",
    "sec-fetch-site": "same-origin",
    "x-wp-api-fetch-from-editor": "true",
    "x-wp-nonce": "e8e40ece26"
  },
  "referrer": "http://localhost/wp-admin/post.php?post=1559&action=edit",
  "referrerPolicy": "strict-origin-when-cross-origin",
  "body": null,
  "method": "POST",
  "mode": "cors",
  "credentials": "include"
});
```
* run that command
* Open the network tab
* Confirm you see the token and post ID

image

### DIFF
--- a/projects/packages/my-jetpack/changelog/update-my-jetpack=-register-ai-jwt-endpoint
+++ b/projects/packages/my-jetpack/changelog/update-my-jetpack=-register-ai-jwt-endpoint
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+My Jetpack: register jetpack-ai-jwt endpoint
--- a/projects/packages/my-jetpack/composer.json
+++ b/projects/packages/my-jetpack/composer.json
@@ -74,7 +74,7 @@
 			"link-template": "https://github.com/Automattic/jetpack-my-jetpack/compare/${old}...${new}"
 		},
 		"branch-alias": {
-			"dev-trunk": "3.1.x-dev"
+			"dev-trunk": "3.2.x-dev"
 		},
 		"version-constants": {
 			"::PACKAGE_VERSION": "src/class-initializer.php"
--- a/projects/packages/my-jetpack/package.json
+++ b/projects/packages/my-jetpack/package.json
@@ -1,7 +1,7 @@
 {
 	"private": true,
 	"name": "@automattic/jetpack-my-jetpack",
-	"version": "3.1.3",
+	"version": "3.2.0-alpha",
 	"description": "WP Admin page with information and configuration shared among all Jetpack stand-alone plugins",
 	"homepage": "https://github.com/Automattic/jetpack/tree/HEAD/projects/packages/my-jetpack/#readme",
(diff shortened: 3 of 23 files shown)


lhkowalski (approved):
Code looks good. The endpoint is registered:

Screen Shot 2023-07-19 at 15 20 28

It also works as expected to get a valid token:

Screen Shot 2023-07-19 at 15 22 56

This is a good move to have the endpoint available on standalone plugins, so we can start to think about using the completions even on products that are not tied to the main Jetpack plugin.

PR #31966, opened 2023-07-19
Launchpad: New update_about_page task

### LINKED ISSUE #78951: Launchpad: Create "Update About Me page" task
Similar to the "Add new page" task (#77295), we need a task for updating the about me page.

1. Task should be visible if the site has an About Me page.
2. Task should be marked complete once the About Me page has been modified.

markbiek:
Fixes https://github.com/Automattic/wp-calypso/issues/78951

## Proposed changes:

This adds a new 'Update About page' Launchpad task on `/home`. The task is only visible if the site has an 'About' page that was headstarted with the theme. In the event that there isn't an 'About' page, we fall back to the 'Add new page' and 'Edit page' tasks.

### Other information:

### Testing Setup

* Apply this PR to Jetpack on your sandbox and sandbox the public-api.
* Run the calypso branch from [79069](https://github.com/Automattic/wp-calypso/pull/79069) locally or use the calypso.live link. This ensures the new Launchpad task will be clickable

### English site testing

* Make sure you WordPress.com account locale is set to English.
* Go to `/start` and create a new site. Using the Free plan is fine.
* Select the "Promote myself or business" goals/intent.
* Select a theme that headstarts an 'About' page (Barnsbury23 is a good example, see 31077-pb/#js for other possibilities).
* Launch the site.
* When you reach the dashboard, you should see an 'Update About page' task.
image
* Clicking the task should take you to the edit view for the 'About' page. 
* Make an edit to the page and publish it. The task should be marked complete when you return to the dashboard.

### Non-English site testing

**NOTE: The task will not appear until the 'Update your About page' string has been translated.**

* Make sure you WordPress.com account locale is set to another language. I've been testing with Spanish (`es`).
* Go to `/start` and create a new site. Using the Free plan is fine.
* Select the "Promote myself or business" goals/intent.
* Select a theme that headstarts an 'About' page (Barnsbury23 is a good example, see 31077-pb/#js for other possibilities).
* Launch the site.
* When you reach the dashboard, you should see an 'Update your About page' task.
* Clicking the task should take you to the edit view for the 'About' page in the correct language. 
* Make an edit to the page and publish it. The task should be marked complete when you return to the dashboard.

### Testing the 'Add a new page' task

* On either of the two test sites you created above, go to the dashboard and click the 'Add new page' task.
image
* Create a new page of any type (other than 'About') and publish it.
* When you return to the dashboard, the 'Add new page' task should be marked complete and the 'Edit page' task _should not_ be visible. The 'Update About' task should still be there.
image

### Testing a site without an 'About' page
* Go to `/start` and create a new site. Using the Free plan is fine.
* Select the "Promote myself or business" goals/intent.
* Select a theme that _does not_ headstart an 'About' page (Russell is a good example).
* Launch the site.
* When you return to the dashboard, you should see the 'Add new page' task but _not_ the 'Update About page' task.
image
* Click the 'Add new page' task.
* Create a new page of any type (other than 'About') and publish it.
* When you return to the dashboard, the 'Add new page' task should be marked complete and the 'Edit page' task _should_ be visible.
image


projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -752,16 +765,130 @@ function wpcom_add_new_page_check( $post_id, $post ) {
 }
 add_action( 'wp_insert_post', 'wpcom_add_new_page_check', 10, 3 );
 
+/**
+ * Determine if the site has an 'About' page. We do this by loading the `en` annotation for the site and theme. We then check if there is a post with the title 'About'. If there is, we get the `hs_old_id` and check to make sure a corresponding post exists in the user's locale.
```
daledupreez: Two comments on the comment:
1. It may be clearer to use `page` instead of `post` to be super-explicit about what we're discussing. I know we're calling post-related functions, but we're talking about pages. 🙃 
2. It may be worth splitting the comment into multiple lines -- it's really long as-is.

```suggestion
 * Determine if the site has an 'About' page.
 * We do this by loading the `en` annotation for the site and theme.
 * We then check if there is a page with the title 'About'.
 * If there is, we get the `hs_old_id` value for the Headstart content.
 * We then find all pages with a matching `hs_old_id` value in post_meta.
 * We then return the first returned page's ID.
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -752,16 +765,130 @@ function wpcom_add_new_page_check( $post_id, $post ) {
...
+	// Return null if there aren't any hs_old_ids to check.
+	if ( empty( $headstart_about_page_hs_old_ids ) ) {
+		return null;
+	}
+
+	$filters     = array(
+		'post_type'    => 'page',
+		'meta_key'     => '_hs_old_id',
+		'meta_compare' => 'IN',
+		'meta_value'   => $headstart_about_page_hs_old_ids,
```
daledupreez, line 878: Thought: should we limit the result to one row? And might we want to specify an order, just to be super-safe?

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -752,16 +765,130 @@ function wpcom_add_new_page_check( $post_id, $post ) {
 }
 add_action( 'wp_insert_post', 'wpcom_add_new_page_check', 10, 3 );
 
+/**
+ * Determine if the site has an 'About' page. We do this by loading the `en` annotation for the site and theme. We then check if there is a post with the title 'About'. If there is, we get the `hs_old_id` and check to make sure a corresponding post exists in the user's locale.
+ *
+ * @return int|null The post ID of the 'About' page if it exists, null otherwise.
+ */
+function wpcom_get_site_about_page_id() {
+	$annotation = wpcom_get_theme_annotation( get_stylesheet() );
```
daledupreez, line 832: The logic in this function feels like it could be relatively heavy/slow, as we may be reading annotation content from disk, and then performing a database query to find the relevant pages. Might it make sense to cache this result and push the computation into a secondary function? The only edge case we'd need to handle would be situations where we're running this code before Headstart has been run for the site, which should be unlikely. (I think we may also set an option while Headstart is running, so we may be able to skip caching the result in that case.)

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -752,16 +765,130 @@ function wpcom_add_new_page_check( $post_id, $post ) {
...
+
+	// Ensure that Headstart posts don't mark this as complete
+	if ( defined( 'HEADSTART' ) && HEADSTART ) {
+		return;
+	}
+
+	// We only care about published pages. Pages added via the API are not published by default.
+	if ( $post->post_status !== 'publish' ) {
+		return;
+	}
```
daledupreez: Might it make sense to defer the check that looks at the database option? All the others are super-quick in-memory comparisons, and they don't depend on the option which may need to be loaded from the database (though it could easily be cached too).

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -752,16 +765,130 @@ function wpcom_add_new_page_check( $post_id, $post ) {
...
+
+	// If the page is not the about page, ignore it.
+	if ( $post->post_title !== 'About' ) {
```
daledupreez: Does this work if we're in a different locale? It may be worth noting that as a comment if it's true, as this looks like it will break for any translated content.

markbiek: Good catch, this is actually a bug. What we really need to do here is make sure that the page being updated has an option matching one of the `hs_old_id`s that denotes if the page is an 'About' page.

Once I've added the caching you suggested about, we should be able to leverage that here to check the page being updated. It's probably worth adding some caching here as well to ensure we're not pulling the same page options over and over.

sixhours:
I ran through this and it works as described! I noticed two strings aren't translated in our system yet, I don't think that's a blocker but good to note:

Screenshot 2023-07-20 at 10 58 29 AM
Screenshot 2023-07-20 at 11 15 36 AM

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -291,6 +291,7 @@ private function build_task( $task ) {
 		$built_task['subtitle']     = $this->load_subtitle( $task );
 		$built_task['badge_text']   = $this->load_value_from_callback( $task, 'badge_text_callback' );
 		$built_task['isLaunchTask'] = isset( $task['isLaunchTask'] ) ? $task['isLaunchTask'] : false;
+		$built_task['extra_data']   = $this->load_extra_data( $task );
```
daledupreez: Rather than assigning this in all cases, can we only return it when we have extra data? That may be easier if the task returns an array or null, but it means we only return the data for `extra_data` over the wire when we have something useful to include. Also, `array()` serialised to JSON doesn't always play nicely with client expectations for JSON objects vs arrays.
```suggestion
		$extra_data                 = $this->load_extra_data( $task );
		if ( is_array( $extra_data ) ) {
			$built_task['extra_data'] = $extra_data;
		}
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -310,6 +311,22 @@ private function load_value_from_callback( $item, $callback, $default = '' ) {
...
+	 * @param Task $task A task definition.
+	 * @return array The extra data for the task.
+	 */
+	private function load_extra_data( $task ) {
+		$extra_data = $this->load_value_from_callback( $task, 'extra_data_callback' );
+		if ( ! empty( $extra_data ) ) {
+			return $extra_data;
+		}
+
+		return array();
```
daledupreez: As discussed above, I think we should update the return type to be `array|null` so we can avoid returning unnecessary data over the wire. We should also enforce that we're actually getting an array back.
```suggestion
	 * @return array|null The extra data for the task.
	 */
	private function load_extra_data( $task ) {
		$extra_data = $this->load_value_from_callback( $task, 'extra_data_callback' );
		if ( is_array( $extra_data ) && array() !== $extra_data ) {
			return $extra_data;
		}

		return null;
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -759,16 +772,151 @@ function wpcom_add_new_page_check( $post_id, $post ) {
...
+ * We do this by loading the `en` annotation for the site and theme.
+ * We then check if there is a post with the title 'About'. If there is, we get the `hs_old_id` and check to make sure a corresponding post exists in the user's locale.
+ *
+ * @return int|null The page ID of the 'About' page if it exists, null otherwise.
+ */
+function wpcom_get_site_about_page_id() {
+	// First, attempt to get the page ID from the cache.
+	$blog_id                 = get_current_blog_id();
+	$about_page_id_cache_key = "wpcom_get_site_about_page_id_{$blog_id}";
+	$about_page_id           = wp_cache_get( $about_page_id_cache_key, 'wpcom' );
```
daledupreez: We should use a much more verbose cache group name that's specific to the package. I would propose something like `'jetpack_mu_wpcom'`. By default, caching is also site-specific, so I think we can also simplify the cache key to something like `'wpcom_about_page_id'`.
```suggestion
	$about_page_id_cache_key = 'wpcom_about_page_id';
	$about_page_id           = wp_cache_get( $about_page_id_cache_key, 'jetpack_mu_wpcom' );
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -759,16 +772,151 @@ function wpcom_add_new_page_check( $post_id, $post ) {
...
+		},
+		$headstart_about_pages
+	);
+
+	// Return null if there aren't any hs_old_ids to check.
+	if ( empty( $headstart_about_page_hs_old_ids ) ) {
+		return null;
+	}
+
+	$filters     = array(
```
daledupreez, line 871: Based on the docs [here](https://developer.wordpress.org/reference/classes/WP_Query/parse_query/), we may also want to add `'fields' => 'ids'` below so we only get the post IDs returned.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -759,16 +772,151 @@ function wpcom_add_new_page_check( $post_id, $post ) {
...
+	// Return null if we couldn't find any published pages matching the hs_old_ids.
+	if ( empty( $about_pages ) ) {
+		return null;
+	}
+
+	// Cache to hs_old_ids so we can validate the post being updated later.
+	wp_cache_set( "headstart_about_page_hs_old_ids_{$blog_id}", $headstart_about_page_hs_old_ids, 'wpcom' );
+
+	// Cache the about page id so we don't have to look it up from scratch every time.
+	wp_cache_set( $about_page_id_cache_key, $about_pages[0]->ID, 'wpcom' );
```
daledupreez: I have a concern here around the caching -- as things stand, we're only going to cache the value when we find a valid about page, but we have a number of `return null` exit conditions above.

To keep things simple, I think it may be easiest if you move the identification logic to a separate function, which would look like the following:
```php
function wpcom_get_site_about_page_id() {
	// First, attempt to get the page ID from the cache.
	$blog_id                 = get_current_blog_id();
	$about_page_id_cache_key = "wpcom_get_site_about_page_id_{$blog_id}";
	$about_page_id           = wp_cache_get( $about_page_id_cache_key, 'jetpack_mu_wpcom' );

	if ( false !== $about_page_id ) {
		// We should decide how we want to cache cases where we don't have an about page.
		// I've used -1 for now, but we could also use 0. I think `null` might run into implementation details in the caching implementation.
		$about_page_id = (int) $about_page_id;
		if ( $about_page_id > 0 ) {
			return $about_page_id;
		}
		return null;
	}

	// Call the current function implementation without any caching logic.
	$about_page_id = wpcom_find_site_about_page_id(); // could also be wpcom_identify_site_about_page_id() or similar

	if ( null === $about_page_id ) {
		$value_to_cache = -1;
	} else {
		$value_to_cache = $about_page_id;
	}

	// Use wp_cache_add() to avoid cache stampedes, and specify an expiration time - I've used an hour here.
	wp_cache_add( $about_page_id_cache_key, $value_to_cache, 'jetpack_mu_wpcom', 3600 );

	return $about_page_id;
}
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -759,16 +772,151 @@ function wpcom_add_new_page_check( $post_id, $post ) {
...
+	}
+
+	// Don't do anything if the task is already complete.
+	if ( wpcom_is_task_option_completed( array( 'id' => 'update_about_page' ) ) ) {
+		return;
+	}
+
+	// If the page is not the previously located About page, ignore it.
+	$blog_id       = get_current_blog_id();
+	$about_page_id = wp_cache_get( "wpcom_get_site_about_page_id_{$blog_id}", 'wpcom' );
```
daledupreez: We shouldn't use the cache here -- we should call `wpcom_get_site_about_page_id()` so we have one code block in charge of the cache retrieval and updates, and that allows us to compute the value if it hasn't been cached. Given that, we should probably change this logic slightly.
```suggestion
	$about_page_id = wpcom_get_site_about_page_id();
	if ( null === $about_page_id || $post->ID !== $about_page_id ) {
		return;
	}
```

daledupreez (commented):
I have some feedback based on inspection.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -310,6 +315,22 @@ private function load_value_from_callback( $item, $callback, $default = '' ) {
 		return $default;
 	}
 
+	/**
+	 * Loads any extra data for a task, calling the 'extra_data_callback' callback if it exists,
+	 * or an empty array if there is no callback.
```
daledupreez: With the most recent update, we return `null` instead of an empty array.
```suggestion
	 * or null if there is no callback or the callback returns an empty array.
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -759,16 +772,176 @@ function wpcom_add_new_page_check( $post_id, $post ) {
 }
 add_action( 'wp_insert_post', 'wpcom_add_new_page_check', 10, 3 );
 
+/**
+ * Return the about page id, if any.
+ *
+ * This fucntion will retrieve the page from the cache whenever possible.
+ *
+ * @return int The page ID of the 'About' page if it exists, null otherwise.
```
daledupreez: Two nits: we have a typo and a type mismatch.
```suggestion
 * This function will retrieve the page from the cache whenever possible.
 *
 * @return int|null The page ID of the 'About' page if it exists, null otherwise.
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -759,16 +772,176 @@ function wpcom_add_new_page_check( $post_id, $post ) {
...
+		$about_page_id = (int) $about_page_id;
+		if ( $about_page_id > 0 ) {
+			return $about_page_id;
+		}
+
+		return null;
+	}
+
+	// Call the current function implementation without any caching logic.
+	$about_page_id = wpcom_find_site_about_page_id(); // could also be wpcom_identify_site_about_page_id() or similar
```
daledupreez: Nit: we can remove the trailing comment here.
```suggestion
	$about_page_id = wpcom_find_site_about_page_id();
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -759,16 +772,176 @@ function wpcom_add_new_page_check( $post_id, $post ) {
...
+	}
+
+	$filters     = array(
+		'post_type'    => 'page',
+		'numberposts'  => 1,
+		'fields'       => 'ids',
+		'orderby'      => 'meta_value',
+		'meta_key'     => '_hs_old_id',
+		'meta_compare' => 'IN',
+		'meta_value'   => $headstart_about_page_hs_old_ids,
```
daledupreez, line 878: Late-breaking thought: do we want to limit this to published posts? I am not sure we do, as modifying an unpublished page is still a good thing.
If we want to add the filter, we'd want to add `'post_status' => 'publish'` to the filter.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -759,16 +772,176 @@ function wpcom_add_new_page_check( $post_id, $post ) {
...
+	);
+	$about_pages = get_posts( $filters );
+
+	// Return null if we couldn't find any published pages matching the hs_old_ids.
+	if ( empty( $about_pages ) ) {
+		return null;
+	}
+
+	// Cache to hs_old_ids so we can validate the post being updated later.
+	wp_cache_set( 'headstart_about_page_hs_old_ids', $headstart_about_page_hs_old_ids, 'jetpack_mu_wpcom' );
```
daledupreez: We should remove these lines as we're not fetching data from this cache.
```suggestion
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -759,16 +772,176 @@ function wpcom_add_new_page_check( $post_id, $post ) {
...
+
+	// Return the id of the first About page.
+	return $about_pages[0];
```
daledupreez: Minor thought: it may be worth forcing the returned value to be an integer, just to be _super-defensive_, and ensuring our upstream code can use strict equality comparisons.
```suggestion
	// Return the id of the first About page.
	return (int) $about_pages[0];
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -238,6 +238,19 @@ function wpcom_launchpad_get_task_definitions() {
 			'is_complete_callback' => 'wpcom_is_task_option_completed',
 		),
 
+		'update_about_page'               => array(
+			'get_title'            => function () {
+				return __( 'Update About page', 'jetpack-mu-wpcom' );
```
daledupreez: After seeing this in the UI, I think it may be better phrased as `Update your About page`.
```suggestion
				return __( 'Update your About page', 'jetpack-mu-wpcom' );
```

As you can see from the screenshot below, we use "your site" in three other tasks, hence me feeling this as being incongruous.

Screenshot 2023-07-24 at 13 39 40

daledupreez (dismissed):
Tested this with the Quadrat theme (which has theme-specific Headstart content), and with English and Dutch clients (with the cache cleared manually for the Dutch client), to make sure the logic is working. Everything looks good to me from a functional perspective, but there are a few minor items to clean up. We'll also need to make a decision on how we want to handle the localisation here. Should we add a check for the translated string in the enabling check?

### Screenshots

These were from the back end being tested in conjunction with https://github.com/Automattic/wp-calypso/pull/79069:

Screenshot 2023-07-24 at 13 39 40
Screenshot 2023-07-24 at 13 51 18

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -322,7 +322,7 @@ private function load_value_from_callback( $item, $callback, $default = '' ) {
 
 	/**
 	 * Loads any extra data for a task, calling the 'extra_data_callback' callback if it exists,
-	 * or an empty array if there is no callback.
+	 * or an null if there is no callback or the callback returns an empty array.
```
daledupreez: Nit: we can drop the `an` here.
```suggestion
	 * or null if there is no callback or the callback returns an empty array.
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -905,11 +902,6 @@ function wpcom_update_about_page_check( $post_id, $post ) {
 		return;
 	}
 
-	// We only care about published pages. Pages added via the API are not published by default.
-	if ( $post->post_status !== 'publish' ) {
-		return;
-	}
```
daledupreez: Could you explain why we're running the check for all post statuses? Do we want to count any basic changes that are not published as completing the task? I want to make sure I understand the intent here, as I think this may have impact on the performance for saving drafts across WordPress.com sites.

markbiek: > Late-breaking thought: do we want to limit this to published posts? I am not sure we do, as modifying an unpublished page is still a good thing. If we want to add the filter, we'd want to add 'post_status' => 'publish' to the filter.

Did I misunderstand your comment about not limiting to only published posts? The implication to me was that applied in this location _and_ the call to `get_posts`.

daledupreez: My thinking wasn't that comprehensive! But when I saw this check removed, I thought it through a bit, and I think I have the following thoughts:
* When we're looking for an About page to update, we are wanting to find a post created by Headstart, so we're not too worried about what publish state it is in to start with.
* When we're trying to detect task completion, we want to know that the user updated that previously identified post. I was thinking that we specifically want them to publish their changes, not just tinker with the content and trigger an auto-save or save without a publish. I suppose we could accept deleting the post, too, as I think that's a valid (and mostly terminal!) update!

I hope that clarifies my thinking, @markbiek!

markbiek: That makes total sense. Sorry, I think I was thinking about it too hard 😝 . I've added this check back in.

markbiek:
@daledupreez When you have a moment, I think this is ready for another review. I've added a change to hide the task on non-English locales until the title is translated.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -315,6 +320,22 @@ private function load_value_from_callback( $item, $callback, $default = '' ) {
 		return $default;
 	}
 
+	/**
+	 * Loads any extra data for a task, calling the 'extra_data_callback' callback if it exists,
+	 * or null if there is no callback or the callback returns an empty array.
```
daledupreez: Reading this fully, it feels like we should break up the sentences.
```suggestion
	 * Loads any extra data for a task, calling the 'extra_data_callback' callback to get the data if the callback is defined.
	 * Returns null if there is no callback or the callback returns an empty array or a non-array.
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -766,16 +779,214 @@ function wpcom_add_new_page_check( $post_id, $post ) {
...
+		'numberposts'  => 1,
+		'fields'       => 'ids',
+		'orderby'      => 'meta_value',
+		'meta_key'     => '_hs_old_id',
+		'meta_compare' => 'IN',
+		'meta_value'   => $headstart_about_page_hs_old_ids,
+	);
+	$about_pages = get_posts( $filters );
+
+	// Return null if we couldn't find any published pages matching the hs_old_ids.
```
daledupreez: Nit: as per our earlier discussion, we're not restricting this to published pages.
```suggestion
	// Return null if we couldn't find any pages matching the hs_old_ids.
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -766,16 +779,214 @@ function wpcom_add_new_page_check( $post_id, $post ) {
...
+}
+
+/**
+ * Check to see if a string has been translated.
+ *
+ * @param string $string The string to check.
+ * @param string $domain The text domain to use.
+ * @return bool True if the string has been translated, false otherwise.
+ */
+function wpcom_launchpad_has_translation( $string, $domain = 'jetpack_mu_wpcom' ) {
```
daledupreez: The domain we're using is `'jetpack-mu-wpcom'`.
```suggestion
function wpcom_launchpad_has_translation( $string, $domain = 'jetpack-mu-wpcom' ) {
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -766,16 +779,214 @@ function wpcom_add_new_page_check( $post_id, $post ) {
...
+	// The task isn't visible if the get_title callback isn't defined or isn't callable.
+	if ( ! isset( $task_definitions['update_about_page']['get_title'] ) || ! is_callable( $task_definitions['update_about_page']['get_title'] ) ) {
+		return false;
+	}
+
+	// The task isn't visible if we're in a non-English locale and the task title hasn't been translated.
+	$update_about_page_title = $task_definitions['update_about_page']['get_title']();
+	if ( ! wpcom_launchpad_has_translation( $update_about_page_title ) ) {
+		return false;
+	}
```
daledupreez: The function in `$task_definitions['update_about_page']['get_title']` is already calling `__()`, so this check isn't 100% correct. Instead, I think we should check using the raw string. Furthermore, I don't think we need to check the task and callback are fully valid, because we don't need to call them.
```suggestion
	// The task isn't visible if the task title hasn't been translated into the current locale.
	if ( ! wpcom_launchpad_has_translation( 'Update your About page', 'jetpack-mu-wpcom' ) ) {
		return false;
	}
```

daledupreez (commented):
The changes look good from inspection, but I'd prefer us to get another round of testing in before we land this. I think we're close to end of day for most of @Automattic/serenity, so I am not sure if anyone else can take a look today.

sixhours (approved):
I went through the testing instructions with English and French and all worked as described!

PR #31970, opened 2023-07-19
Launchpad: Add count properties to task

valterlorran:
Related to https://github.com/Automattic/wp-calypso/issues/79611

## Proposed changes:

Some of the new tasks for the Newsletter launchpad require repetition in the task to mark it as complete. For example, we'll have the "Write 3 posts" task, which requires the user to write 3 posts to complete the task. The properties added in this PR will provide the data so we can use show these numbers in the interface, as the image below suggests.

Screenshot 2023-07-19 at 08 42 38

* Adds the `target_repetitions ` property: This property is used to define the number of repetitions to reach to complete the task.
* Adds the `repetition_count_callback` property: This callback will be used to calculate how many times the task was repeated.

### Other information:

## Testing instructions:

* The properties added in this PR are not being used yet, but I added some unit tests. Run:
```bash
cd projects/packages/jetpack-mu-wpcom; ./vendor/bin/phpunit --filter=Launchpad_Task_Lists_Test; cd -
```

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
@@ -292,6 +292,11 @@ class Launchpad_Task_Lists {
 		$built_task['badge_text']   = $this->load_value_from_callback( $task, 'badge_text_callback' );
 		$built_task['isLaunchTask'] = isset( $task['isLaunchTask'] ) ? $task['isLaunchTask'] : false;
 
+		if ( isset( $task['target_repetitions'] ) ) {
+			$built_task['target_repetitions'] = $task['target_repetitions'];
+			$built_task['repetition_count']   = $this->load_repetition_count( $task );
+		}
+
 		return $built_task;
 	}
 
@@ -348,6 +353,16 @@ class Launchpad_Task_Lists {
 			$task['subtitle'];
 		}
 		return '';
+	}
+
+	/**
+	 * Loads the repetition count for a task, calling the callback if it exists.
+	 *
+	 * @param Task $task A task definition.
+	 * @return int|null The repetition count for the task.
+	 */
+	private function load_repetition_count( $task ) {
+		return $this->load_value_from_callback( $task, 'repetition_count_callback' );
 	}
 
 	/**
@@ -610,6 +625,14 @@ class Launchpad_Task_Lists {
 
 		if ( ! $has_valid_title ) {
 			_doing_it_wrong( 'validate_task', 'The Launchpad task being registered requires a "title" attribute or a "get_title" callback', '6.2' );
+			return false;
+		}
+
+		$has_any_repetition_properties  = isset( $task['target_repetitions'] ) || isset( $task['repetition_count_callback'] );
+		$has_both_repetition_properties = isset( $task['target_repetitions'] ) && isset( $task['repetition_count_callback'] );
+
+		if ( $has_any_repetition_properties && ! $has_both_repetition_properties ) {
+			_doing_it_wrong( 'validate_task', 'The Launchpad task being registered requires both a "target_repetitions" attribute and a "repetition_count_callback" callback', '6.3' );
 			return false;
 		}
 
--- a/projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-lists-test.php
+++ b/projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-lists-test.php
@@ -185,4 +185,73 @@ class Launchpad_Task_Lists_Test extends \WorDBless\BaseTestCase {
 		);
 		$this->assertEquals( $expected, wpcom_launchpad_is_task_list_completed( 'task-list-test' ) );
 	}
+
+	/**
+	 * Test the validation of the repetition props.
+	 * Both target_repetitions and repetition_count_callback must be set if either property is set.
+	 */
+	public function test_repetition_counting_task_definition() {
+		$valid_task = array(
+			'id'                        => 'task_0',
+			'title'                     => 'task_0',
+			'target_repetitions'        => 2,
+			'repetition_count_callback' => function () {
+				return 1;
+			},
+		);
+		$this->assertTrue( Launchpad_Task_Lists::validate_task( $valid_task ) );
+
+		$invalid_task_without_repetition_count_callback = array(
+			'id'                 => 'task_0',
+			'title'              => 'task_0',
+			'target_repetitions' => 2,
+		);
+		$this->assertFalse( Launchpad_Task_Lists::validate_task( $invalid_task_without_repetition_count_callback ) );
+
+		$invalid_task_without_target_repetitions = array(
+			'id'                        => 'task_0',
+			'title'                     => 'task_0',
+			'repetition_count_callback' => function () {
+				return 1;
+			},
+		);
+		$this->assertFalse( Launchpad_Task_Lists::validate_task( $invalid_task_without_target_repetitions ) );
+	}
+
+	/**
+	 * Test the values for target_repetitions and repetition_count_callback.
+	 */
+	public function test_repetition_counting_task_list() {
+		wpcom_register_launchpad_task(
+			array(
+				'id'                        => 'task-0',
+				'title'                     => 'Task 0',
+				'target_repetitions'        => 2,
+				'repetition_count_callback' => function () {
+					return 1;
+				},
+			)
+		);
+
+		wpcom_register_launchpad_task(
+			array(
+				'id'    => 'task-1',
+				'title' => 'Task 1',
+			)
+		);
+
+		wpcom_launchpad_checklists()->unregister_task_list( 'task-list-test' );
+		wpcom_register_launchpad_task_list(
+			array(
+				'id'       => 'task-list-test',
+				'title'    => 'Simple testing task list',
+				'task_ids' => array( 'task-0', 'task-1' ),
+			)
+		);
+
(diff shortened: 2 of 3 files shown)


projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -590,6 +604,13 @@ public static function validate_task( $task ) {
 			return false;
 		}
 
+		$has_both_properties_for_counting = ( isset( $task['minimum_repetition_to_complete'] ) || isset( $task['repetition_count_callback'] ) ) && ! ( isset( $task['minimum_repetition_to_complete'] ) && isset( $task['repetition_count_callback'] ) );
+
+		if ( $has_both_properties_for_counting ) {
```
andres-blanco: Shouldn't this be `if ( ! $has_both_properties_for_counting ) {`?

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -292,6 +292,10 @@ private function build_task( $task ) {
 		$built_task['badge_text']   = $this->load_value_from_callback( $task, 'badge_text_callback' );
 		$built_task['isLaunchTask'] = isset( $task['isLaunchTask'] ) ? $task['isLaunchTask'] : false;
 
+		$repetition_count                 = $this->load_repetition_count( $task );
+		$built_task['repetition_count']   = $repetition_count ? $repetition_count : null;
+		$built_task['minimum_repetition'] = isset( $task['minimum_repetition_to_complete'] ) ? $task['minimum_repetition_to_complete'] : null;
```
daledupreez: Minor naming thought: 'minimum_repetition' doesn't clearly communicate that it's a total, target, or maximum count. Might `target_repetitions` be more accurate? We could use that for both properties.
```suggestion
		$built_task['target_repetitions'] = isset( $task['target_repetitions'] ) ? $task['target_repetitions'] : null;
```

daledupreez: As written, this means we won't set the value to `0` when that's the correct value.

Might it not be better to set `repetition_count` conditionally based on `target_repetitions`/`minimum_repetition`?
```suggestion
		if ( isset( $task['target_repetitions' ] ) {
			$built_task['target_repetitions'] = $task['target_repetitions'];

			// Note: this logic assumes that we return a number from load_repetition_count()
			$repetition_count               = $this->load_repetition_count( $task );
			$built_task['repetition_count'] = $repetition_count;
		} else {
			// Not sure if we need this
			$built_task['target_repetitions'] = null;
			$built_task['repetition_count']   = null;
		}
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
valterlorran: Thanks for catching it, the condition is correct but it was poorly named. I renamed it to: `missing_repepetition_property`, do you think it is clear now?

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -292,6 +292,15 @@ private function build_task( $task ) {
 		$built_task['badge_text']   = $this->load_value_from_callback( $task, 'badge_text_callback' );
 		$built_task['isLaunchTask'] = isset( $task['isLaunchTask'] ) ? $task['isLaunchTask'] : false;
 
+		if ( isset( $task['target_repetitions'] ) ) {
+			$built_task['target_repetitions'] = $task['target_repetitions'];
+			$repetition_count                 = $this->load_repetition_count( $task );
+			$built_task['repetition_count']   = $repetition_count;
```
daledupreez: Now that I see this written out, I don't think we need the variable.
```suggestion
			$built_task['repetition_count']   = $this->load_repetition_count( $task );
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -590,6 +609,13 @@ public static function validate_task( $task ) {
 			return false;
 		}
 
+		$missing_repetition_property = ( isset( $task['target_repetitions'] ) || isset( $task['repetition_count_callback'] ) ) && ! ( isset( $task['target_repetitions'] ) && isset( $task['repetition_count_callback'] ) );
+
+		if ( $missing_repetition_property ) {
```
daledupreez: This logic is pretty complicated. Might it be clearer if we split the logic up into two distinct checks?
```suggestion
		$has_any_repetition_properties  = isset( $task['target_repetitions'] ) || isset( $task['repetition_count_callback'] );
		$has_both_repetition_properties = isset( $task['target_repetitions'] ) && isset( $task['repetition_count_callback'] );

		if ( $has_any_repetition_properties && ! $has_both_repetition_properties ) {
```

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-lists-test.php
```diff
@@ -177,4 +177,64 @@ public function test_task_list_is_completed( $tasks, $expected, $task_list_optio
 		);
 		$this->assertEquals( $expected, wpcom_launchpad_is_task_list_completed( 'task-list-test' ) );
 	}
+
+	/**
+	 * Test the validation of the repetition props. Both properties(target_repetitions & repetition_count_callback)
+	 * must be set if one is set.
```
daledupreez: Nit: I think we can simplify the description a bit.
```suggestion
	 * Test the validation of the repetition props.
	 *  Both target_repetitions and repetition_count_callback must be set if either property is set.
```

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-lists-test.php
```diff
@@ -177,4 +177,64 @@ public function test_task_list_is_completed( $tasks, $expected, $task_list_optio
...
+			},
+		);
+		$this->assertTrue( Launchpad_Task_Lists::validate_task( $valid_task_list ) );
+
+		$valid_task_list = array(
+			'id'                 => 'task_0',
+			'title'              => 'task_0',
+			'target_repetitions' => 2,
+		);
+		$this->assertFalse( Launchpad_Task_Lists::validate_task( $valid_task_list ) );
```
daledupreez: Can we also add a test case for the other field being missing? We should also update the variable name. 😁 
```suggestion
		$invalid_task_list = array(
			'id'                 => 'task_0',
			'title'              => 'task_0',
			'target_repetitions' => 2,
		);
		$this->assertFalse( Launchpad_Task_Lists::validate_task( $invalid_task_list ) );

		$invalid_task_list = array(
			'id'                 => 'task_0',
			'title'              => 'task_0',
			'repetition_count_callback' => function () {
				return 1;
			},
		);
		$this->assertFalse( Launchpad_Task_Lists::validate_task( $invalid_task_list ) );
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -292,6 +292,15 @@ private function build_task( $task ) {
...
 		$built_task['isLaunchTask'] = isset( $task['isLaunchTask'] ) ? $task['isLaunchTask'] : false;
 
+		if ( isset( $task['target_repetitions'] ) ) {
+			$built_task['target_repetitions'] = $task['target_repetitions'];
+			$repetition_count                 = $this->load_repetition_count( $task );
+			$built_task['repetition_count']   = $repetition_count;
+		} else {
+			$built_task['target_repetitions'] = null;
+			$built_task['repetition_count']   = null;
+		}
```
daledupreez: I tested the API and this means we always write these values to the returned JSON payload. Should we only add them when they're actually in use? I think it would be better not to return them unless we actually need them.
```suggestion
		}
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
andres-blanco: Another more concise way of putting it might be:
```
$missing_repetition_property = isset( $task['target_repetitions'] ) xor isset( $task['repetition_count_callback'] );
```
Which will return true if either target_repetitions or repetition_count_callback is set, but not both.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
andres-blanco: yes, the other name was confusing. Thanks ❤️

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
valterlorran: TIL; This is pretty cool!

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
valterlorran: Yes, I prefer not to return it.

daledupreez (approved):
LGTM! I've run the unit tests, and mocked up adding the data for an existing task, and everything works via the launchpad REST API on my sandbox.

However, please update the main description for the PR to reflect the naming changes we made - it's gotten a bit stale.

PR #31983, opened 2023-07-20
Boost: Image guide/more realistic size estimates

thingalon:
Make more realistic estimates of image file size savings when shrinking images, taking mimetype and typical compression ratios into account.

## Proposed changes:
* Update Image Guide to add a method to fetch the mimetype of an image
* Update image guide to accept a mimetype when estimating the size of an image
* Update Boost to use the new Image Guide method updates.

### Other information:

## Jetpack product discussion
n/a

no

## Testing instructions:

* Try out the image guide.
* Look at its size estimates.
* See they are generally more realistic. :)

### DIFF
--- a/projects/js-packages/image-guide/changelog/image-guide-more-realistic-size-estimates
+++ b/projects/js-packages/image-guide/changelog/image-guide-more-realistic-size-estimates
@@ -1,0 +1,4 @@
+Significance: patch
+Type: added
+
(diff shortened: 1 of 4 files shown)


thingalon:
After discussing this with adnan, it is honestly not a better estimate. :-/ will need to tinker with the idea a bit more.

PR #31997, opened 2023-07-20
Jetpack: Replace some bad `plugin_basename()` calls

anomiex:
## Proposed changes:

WordPress's `plugin_basename()` function is supposed to be passed a full path (e.g. `__FILE__`), which it transforms if possible into a path relative to `WP_PLUGIN_DIR` or `WPMU_PLUGIN_DIR`.

We have a few places that are not using it for that purpose. Instead they're passing a relative path, which will normally just return the input but occasionally in the middle of Atomic updates can return a completely incorrect path. What we really want to do in these places is append the relative path to `__DIR__`.

### Other information:

## Jetpack product discussion

p1689195516466559-slack-C034JEXD1RD

No

## Testing instructions:

* Make sure these modules still work, I guess.

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-bad-plugin_basename-calls
+++ b/projects/plugins/jetpack/changelog/fix-bad-plugin_basename-calls
@@ -1,0 +1,4 @@
+Significance: patch
+Type: other
+
+Fix some `require` paths to not uselessly call `plugin_basename()`.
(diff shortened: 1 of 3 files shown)


anomiex:
FYI, the Atomic errors mentioned are the ones that look like this:
```
PHP Fatal error:  Uncaught Error: Failed opening required 'full-site-editingclasses/wp-google-analytics-utils.php' (include_path='/:.') in /wordpress/plugins/jetpack/12.4-a.7/modules/google-analytics/wp-google-analytics.php:25
```
Presumably somehow during the update of the full-site-editing plugin `$wp_plugin_paths` wound up having a mapping from the string "full-site-editing" to the empty string where it's supposed to be mapping from one full path to another.

samiff (approved):

zinigor (approved):
Smoke testing shows that require statements work fine!

PR #31998, opened 2023-07-20
Form block: rename newsletter sign-up variation

### LINKED ISSUE #31989: Rename the /Newsletter block (Jetpack Contact Form variant) to be /Signup to avoid confusion.
Today, we have 4 Jetpack blocks that will pop up when you type in "newsletter" in the inserter:

- The Subscribe block.
- The Premium Content block.
- The Mailchimp block.
- The Form block's Newsletter sign-up variation.

The first three have newsletter set as a keyword, while the 4th has Newsletter in its title, so Gutenberg gives it a bigger priority out of the box.

We could certainly rename it and remove the "Newsletter" name from it, to avoid that. I don't know how it should be called. It would not impact existing blocks that have been added in the past.

Maybe it could be just "Sign-up", instead of "Newsletter Sign-up"?

jeherve:
Fixes #31989

## Proposed changes:

We have multiple blocks answering to the "newsletter" keyword nowadays. Since this variation is not a full-fledged newsletter sign-up on its own, let's avoid calling it "Newsletter".

**After the change**

Screenshot 2023-07-20 at 18 39 57
Screenshot 2023-07-20 at 18 39 46

### Other information:

## Jetpack product discussion

* pbNhbs-7sw-p2

* No

## Testing instructions:

> **Note**
> This variation is not available on WordPress.com simple, so you will not be able to test there.

On a site that's connected to WordPress.com, go to Posts > Add New, and search for "Newsletter". The Form Block variation should not come up. It should, however, come up if you search for "Sign-up".

### DIFF
--- a/projects/packages/forms/changelog/update-newsletter-sign-up-variation-name
+++ b/projects/packages/forms/changelog/update-newsletter-sign-up-variation-name
@@ -1,0 +1,4 @@
+Significance: patch
+Type: changed
+
+Update the name of the Newsletter Sign-up Variation.
--- a/projects/packages/forms/package.json
+++ b/projects/packages/forms/package.json
@@ -1,7 +1,7 @@
 {
 	"private": true,
 	"name": "@automattic/jetpack-forms",
-	"version": "0.19.7",
+	"version": "0.19.8-alpha",
 	"description": "Jetpack Forms",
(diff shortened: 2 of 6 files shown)


mikestottuk (commented):
LGTM - the simplest change. Increasing usage of the block can follow (i.e. make it more obvious "Subscribe" is for Newsletter.

mikestottuk (approved):
LGTM

PR #32002, opened 2023-07-20
Subscriptions: Remove feature flag for subcribe modal

edanzer:
## Proposed changes:

* REMOVE the feature flag for the subscribe modal. 
* As an extra safety check, ADD a check for whether we're in the WordPress.com environment. This will allow us to test this feature on WordPress.com sites without exposing it yet to self-hosted Jetpack sites. You can still force the feature to be active on self-hosted jetpack by adding this filter: `add_filter( 'jetpack_subscriptions_modal_enabled', '__return_true', 20 );`
* The two changes above means this feature will be exposed to (a) WordPress.com sites that (b) have newsletter as site intent OR have the lettre theme active. Even in these conditions, the modal will still be disabled by default - you will just be able to enable it via normal settings rather than needing to add a filter / feature flag. 
   
### Other information:

Possibly. It means users can now have subscribers, which mean collecting and storing the email for each subscriber.

## Testing instructions:

1) Test the modal CANNOT be loaded on self-hosted Jetpack without adding a filter/feature flag.
   - Test in your Jetpack dev environment with Jurassic tube test site configured. 
   - If you have tested the Subscribe Modal before, you may need to undo specific changes. First, remove any filters you added to enable the feature flag before. Second, if you enabled the Subscribe modal before, disable it.
   - Activate the Lettre theme. Before adding the is_wpcom() method in this PR, the only way to see this modal on a self-hosted site was to have Lettre theme active. 
   - Go to Jetpack > Settings > Discussion > Newsletters and enable the option to "Let visitors subscribe to new posts and comments via email. This will enable the Jetpack subscriptions module in general. Then refresh the page. 
   - Confirm you do not see the option for "Enable subscriber modal"

2) Test that the modal CAN be enabled Simple sites. 
   - You will need a a WordPress.com simple site with site_intent of newsletter, or Lettre theme active.
   - Run `bin/jetpack-downloader test jetpack upddate/remove-subscribe-modal-feature-flag` to load this branch in your sandbox. Be sure to sandbox public-api and the domain of your test site. 
   - You will need to access WordPress.com via a local calypso dev environment (there is currently a feature flag such that the setting to enable the modal on WordPress.com only shows in a calypso dev environment). 
   - Go to Settings > Reading > Newsletters, confirm the option to enable the modal shows, and enable it. 
   - On the frontend, go to to any single post on the frontend, scroll, and confirm the modal does not load after scrolling.
   - On the admin side, go to Jetpack > Settings > Discussion > Newsletters and confirm the option to enable the modal does not show. 
   - As an extra check, return to Settings > Reading > Newsletters, disable the modal, and confirm it no longer loads on the frontend.
   
jetpack-subscribe-flag

### DIFF
--- a/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
+++ b/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
@@ -255,15 +255,6 @@ HTML;
 		}
 		return false;
 	}
-
-	/**
-	 * Check if we're in WordPress.com.
-	 *
-	 * @return bool
-	 */
-	public static function is_wpcom() {
-		return defined( 'IS_WPCOM' ) && IS_WPCOM;
-	}
 }
 
 add_filter(
--- a/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
+++ b/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
@@ -257,25 +257,6 @@ add_filter(
 	)
 );
 
-/*
- * The following line is being added temporarily as a feature flag.
- *
- * It disables the subscribe modal feature using the
- * jetpack_subscriptions_modal_enabled filter. If you want to test
- * this feature, you'll need to override the line below by adding
- *
- * add_filter( 'jetpack_subscriptions_modal_enabled', '__return_true', 20 );
- *
- * to your test site. When we are ready for full release of this
- * feature, we will remove this line, but will leave the
- * jetpack_subscriptions_modal_enabled filter in place.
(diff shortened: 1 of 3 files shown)


edanzer:
I need to resolve some test failures in this PR. Test conditions have changed with removal of the feature flag.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -204,6 +204,9 @@ public function get_subscribe_template_content() {
 	 * @return bool
 	 */
 	public static function should_load_subscriber_modal() {
+		if ( ! self::is_wpcom() ) {
```
lezama: you can use this helper:

```php
use Automattic\Jetpack\Status\Host;

( new Host() )->is_wpcom_platform()
```

lezama (approved):

PR #32003, opened 2023-07-20
CRM: Fix error 200 using api connector

gogdzl:
## Proposed changes:
This PR fixes a problem with new api connector sites giving an error 200 while testing the connection.

It simply updates our API key prefix to `jpcrm_`, while keeping compatibility with the previous one (`zbscrm_`). 

### Other information:

## Jetpack product discussion
https://github.com/Automattic/zero-bs-crm/issues/3224

No.

## Testing instructions:
- Set up a API Connector on an external site
- Set up a CRM site with API module enabled
- On the CRM site, disable frontend in CRM Settings.
- Copy over CRM site's API secrets/keys, etc. Re-save permalinks
- Test the API connector connection and find 200 error (without this PR)
- On the CRM site, enable frontend in CRM settings
- Test the API connector connection and find that it works.

With this PR it should work even with the frontend disabled.

### DIFF
--- a/projects/plugins/crm/includes/ZeroBSCRM.AdminPages.Checks.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.AdminPages.Checks.php
@@ -170,7 +170,7 @@ function zeroBSCRM_isAPIRequest() {
 	// below is more reliable as QUERY_STRING will always be set for API requests.
 
 	// lazy, non-wp way of doing this
-	if ( isset( $_SERVER['QUERY_STRING'] ) && strpos( '#' . $_SERVER['QUERY_STRING'], 'api_key=zbscrm_' ) > 0 ) {
+	if ( isset( $_SERVER['QUERY_STRING'] ) && ( strpos( '#' . sanitize_text_field( wp_unslash( $_SERVER['QUERY_STRING'] ) ), 'api_key=zbscrm_' ) > 0 || strpos( '#' . sanitize_text_field( wp_unslash( $_SERVER['QUERY_STRING'] ) ), 'api_key=jpcrm_' ) ) ) {
 		return true;
 	}
 
--- a/projects/plugins/crm/changelog/fix-crm-3224-error-200-using-api-connector
+++ b/projects/plugins/crm/changelog/fix-crm-3224-error-200-using-api-connector
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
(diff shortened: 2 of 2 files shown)


projects/plugins/crm/includes/ZeroBSCRM.AdminPages.Checks.php
```diff
@@ -170,7 +170,7 @@ function zeroBSCRM_isAPIRequest() {
 	// below is more reliable as QUERY_STRING will always be set for API requests.
 
 	// lazy, non-wp way of doing this
-	if ( isset( $_SERVER['QUERY_STRING'] ) && strpos( '#' . $_SERVER['QUERY_STRING'], 'api_key=zbscrm_' ) > 0 ) {
+	if ( isset( $_SERVER['QUERY_STRING'] ) && ( strpos( '#' . sanitize_text_field( wp_unslash( $_SERVER['QUERY_STRING'] ) ), 'api_key=zbscrm_' ) > 0 || strpos( '#' . sanitize_text_field( wp_unslash( $_SERVER['QUERY_STRING'] ) ), 'api_key=jpcrm_' ) ) ) {
```
cleacos, line 173: **Legacy code - for reference:** Wondering why we are not using $_GET to check `api_key` and then if it starts with `jpcrm_` or `zbscrm_`.

cleacos (approved):
It works as expected.

LGTM!

projects/plugins/crm/includes/ZeroBSCRM.AdminPages.Checks.php
tbradsha: Which is somewhat how we handle it in `jpcrm_is_api_request_authorised()`...

Based on the comments, it seems maybe it was trying to address some rewrite/permalinks oddities. But ultimately, a legacy oddity.

gogdzl: This function only checks whether the input resembles an API request and is more lightweight compared to `jpcrm_is_api_request_authorised`.

tbradsha: Indeed, I'm primarily referring to the overly-complex checks in `zeroBSCRM_isAPIRequest()` vs. `jpcrm_is_api_request_authorised()`. But the reason is simply "because it's legacy".

PR #32004, opened 2023-07-20
AI Client: fix completion restart issue when changing tabs

### LINKED ISSUE #31994: Jetpack AI Assistant: Navigating to another tab stops content generation
## Summary 
While testing Jetpack AI on a personal site, I noticed that if I were navigating away while the text generation was underway, it would stop generating text. 

I also noticed that when I was back on the tab, it started producing content from the beginning instead of resuming where it stopped. 

Ideally, it should keep producing content even if I navigate to another tab, as that's how most people might be familiar with when using other tools. :)

## Steps to reproduce
1. On a new post editor, insert the AI block
2. Ask it to write something long "Write a long blog post about fishing"
3. Navigate to another tab just as the text generation starts
4. Navigate back to the editor

## What I expected to happen
* Either the text generation continues in the background
* Or it resumes where it paused when I navigated to another tab

## What happened instead
* Text generation stopped when I switched to another tab
* Text generation started from the beginning (in addition to the existing text in the block) when I switched to the editor tab

## Tested on
Both WordPress.com Simple and self-hosted WordPress.org sites.

Slack convo: p1689860483061449-slack-C054LN8RNVA

lhkowalski:
Fixes #31994.

## Proposed changes:

* Stop using the [fetchEventSource](https://www.npmjs.com/package/@microsoft/fetch-event-source) built-in smart handling of connections ([look for `openWhenHidden` here](https://github.com/Azure/fetch-event-source/blob/main/src/fetch.ts)) so the completion will not stop and restart when changing tabs

### Other information:

## Testing instructions:

Follow the steps to reproduce the issue:

1. On a new post editor, insert the AI block
2. Ask it to write something long, for example "Write a long blog post about fishing"
3. Navigate to another tab when the text generation starts
4. Navigate back to the editor

Before the fix:

* Text generation stops when I switch to another tab
* Text generation starts from the beginning (in addition to the existing text in the block) when I switch back to the editor tab; a new request to the backend can be found on the network logs

After the fix:

* The text generation continues in the background and the second request to the backend does not happen

### DIFF
--- a/projects/js-packages/ai-client/changelog/update-ai-assistant-fix-completion-restart-when-changing-tabs
+++ b/projects/js-packages/ai-client/changelog/update-ai-assistant-fix-completion-restart-when-changing-tabs
@@ -1,0 +1,4 @@
+Significance: patch
+Type: changed
+
+AI Client: stop using smart document visibility handling on the fetchEventSource library, so it does not restart the completion when changing tabs.
--- a/projects/js-packages/ai-client/src/suggestions-event-source/index.ts
+++ b/projects/js-packages/ai-client/src/suggestions-event-source/index.ts
@@ -93,6 +93,7 @@ export default class SuggestionsEventSource extends EventTarget {
 		}
 
 		await fetchEventSource( url, {
+			openWhenHidden: true,
 			method: 'POST',
 			headers: {
 				'Content-type': 'application/json',


retrofox (approved):
LGTM 🚢

PR #32005, opened 2023-07-21
IDC: URL secret to detect multi-domain sites

sergeymitr:
## Proposed changes:
* Introduce URL secret to detect multi-domain sites.

### Other information:

## Jetpack product discussion
p9dueE-7BD-p2#comment-9897

No.

## Testing instructions:
1. Install the "Jetpack Debug" plugin, activate "IDC Simulation" and "REST API Tester".
2. Connect Jetpack.
3. Go to "Jetpack Debug -> IDC Simulation", enter a Spoof URL and enable "IDC Simulation". Click "Send Remote Request".
4. You should see the IDC screen. Run `wp option get jetpack_options` and confirm that it contains the URL Secret value:
```
'identity_crisis_url_secret' =>
  array (
    'secret' => 'random_string',
    'expires_at' => 1690162076,
  ),
```
5. Go to "Jetpack Debug -> "REST API Tester" and send a GET request to `jetpack/v4/identity-crisis/fetch-url-secret`. Confirm that you get the 403 error:
```
{
    "code": "invalid_user_permission_identity_crisis",
    "message": "You do not have the correct user permissions to perform this action.",
    "data": {
        "status": 403
    }
}
```
6. Use 3122a-pb in WPCOM sandbox via `wpsh` to test the `jetpack.idcUrlValidation` and `/jetpack/v4/identity-crisis/fetch-url-secret` endpoints.
Confirm that the secret in both API responses matches, and expiration timestamp (only returned by the "fetch" endpoint) is approximately 5 minutes in the future.


projects/packages/identity-crisis/src/class-rest-endpoints.php
```diff
@@ -62,6 +63,28 @@ public static function initialize_rest_api() {
 				),
 			)
 		);
+
+		// Generate URL verification secret.
+		register_rest_route(
```
fgiannar: I believe this route is redundant since we have agreed that generating the secret will be handled by the existing XML-RPC endpoint, right?

projects/packages/identity-crisis/src/class-url-secret.php
```diff
@@ -0,0 +1,162 @@
...
+	 *
+	 * @return array
+	 * phpcs:ignore Squiz.Commenting.FunctionCommentThrowTag -- The exception is being caught, false positive.
+	 */
+	public static function add_secret_to_url_validation_response( array $response ) {
+		try {
+			$secret = new URL_Secret();
+
+			if ( ! $secret->create() ) {
+				throw new Exception( esc_html__( 'Unable to create an URL secret.', 'jetpack-idc' ), 'unable_to_create_url_secret' );
```
fgiannar: I think we could make this cleaner by avoiding this approach of throwing Exception inside a try/catch block.
Also, optionally we could consider that if we fail to generate the secret we could just return the original `$response` without the error

projects/packages/identity-crisis/src/class-url-secret.php
```diff
@@ -0,0 +1,162 @@
...
+		}
+	}
+
+	/**
+	 * Fetch the URL secret from the database.
+	 *
+	 * @return array|null
+	 */
+	private function fetch() {
+		$data = Jetpack_Options::get_option( static::OPTION_KEY );
```
fgiannar, line 60: How come we don't store the secret as a transient with expiration 5mins?

projects/packages/identity-crisis/src/class-url-secret.php
```diff
@@ -0,0 +1,162 @@
...
+
+namespace Automattic\Jetpack\IdentityCrisis;
+
+use Jetpack_Options;
+use WP_Error;
+
+/**
+ * IDC URL secret functionality.
+ */
+class URL_Secret {
```
fgiannar, line 16: Minor but why the `URL_` prefix? I think plain `Secret` could also work. Also we could add a DocBlock here explaining it's purpose, eg: "A short lived secret used to verify whether an IDC is coming from the same vs a different Jetpack site"

projects/packages/identity-crisis/src/class-exception.php
```diff
@@ -0,0 +1,13 @@
...
+ *
+ * @package  automattic/jetpack-identity-crisis
+ */
+
+namespace Automattic\Jetpack\IdentityCrisis;
+
+/**
+ * Exception class for the Identity Crisis package.
+ */
+class Exception extends \Exception {}
```
fgiannar, line 13: This class feels a bit redundant. Why do we need it?

projects/packages/identity-crisis/src/class-url-secret.php
sergeymitr: I opted against them because they are by definition unreliable: _["it is possible for the transient to not be available before the expiration time"](https://developer.wordpress.org/apis/transients/)_.
So I decided to use options instead, and manually invalidate them according to the expiration timestamp.

projects/packages/identity-crisis/src/class-exception.php
sergeymitr: We use it to differentiate the IDC exceptions from all other kinds of exceptions we wouldn't want to accidentally handle ([here](https://github.com/Automattic/jetpack/blob/80ed2c122e610b90c03fd7d28c59e80df99018e5/projects/packages/identity-crisis/src/class-rest-endpoints.php#L207) and [here](https://github.com/Automattic/jetpack/blob/1f3358423b960f6cf3ebf9b828ccb2d32d607477/projects/packages/identity-crisis/src/class-url-secret.php#L143)).

projects/packages/identity-crisis/src/class-url-secret.php
sergeymitr: The secret is not multi-purpose (e.g. WP nonce), and used solely for a very specific task. I believe the name should also be more specific than simple `Secret`.
Updated phpdoc: https://github.com/Automattic/jetpack/pull/32005/commits/1f3358423b960f6cf3ebf9b828ccb2d32d607477

projects/packages/identity-crisis/src/class-rest-endpoints.php
sergeymitr: That's a leftover from when we planned to use two separate endpoints, thanks for catching that.
Removed: https://github.com/Automattic/jetpack/pull/32005/commits/80ed2c122e610b90c03fd7d28c59e80df99018e5

projects/packages/identity-crisis/src/class-url-secret.php
sergeymitr: The idea is to log the error in Kibana, we don't need to throw it for the user.
We could instead save it into Tracks/Hue, but it would be a bit harder to lookup.

And the idea behind throwing the exception is to only handle those exceptions in a single place (the `catch` block) instead of copying the error handling.

projects/packages/identity-crisis/src/class-rest-endpoints.php
```diff
@@ -184,4 +196,46 @@ public static function identity_crisis_mitigation_permission_check() {
 		return new WP_Error( 'invalid_user_permission_identity_crisis', $error_msg, array( 'status' => rest_authorization_required_code() ) );
 	}
 
+	/**
+	 * Endpoint for fetching the existing secret.
+	 *
+	 * @return WP_Error|\WP_REST_Response
+	 */
+	public static function fetch_url_secret() {
+		try {
```
fgiannar: I think we don't need a `try/catch` here, right?

projects/packages/identity-crisis/src/class-url-secret.php
fgiannar: Agreed, my logic was that we are already under the `Automattic\Jetpack\IdentityCrisis` namespace so the `Secret` is essentially a functionality tight to IDC

projects/packages/identity-crisis/src/class-url-secret.php
fgiannar: I'm fine with returning the error but I strongly suggest we don't throw inside the `try/catch` block. How about something like:

```
$secret = new URL_Secret();

try {
    $secret->create();
 }
 catch ( Exception $e ) {
    $response['url_secret_error'] = new WP_Error( $e->getCode(), $e->getMessage() );
}

if ( $secret->exists() ) {
	$response['url_secret'] = $secret->get_secret();
}

return $response;
```

projects/packages/identity-crisis/src/class-url-secret.php
```diff
@@ -0,0 +1,163 @@
...
+	/**
+	 * Adds `url_secret` to the `jetpack.idcUrlValidation` URL validation endpoint.
+	 * Adds `url_secret_error` in case of an error.
+	 *
+	 * @param array $response The endpoint response that we're modifying.
+	 *
+	 * @return array
+	 * phpcs:ignore Squiz.Commenting.FunctionCommentThrowTag -- The exception is being caught, false positive.
+	 */
+	public static function add_secret_to_url_validation_response( array $response ) {
```
fgiannar: I could see this method inside `class-identity-crisis.php` instead of here. I think it makes more sense there, leaving the current class only handle fetching and storing IDC secrets.

projects/packages/identity-crisis/src/class-rest-endpoints.php
sergeymitr: Right, there used to be an exception thrown in `URL_Secret` constructor, so it's no longer needed.
Thanks for noticing that, removed: https://github.com/Automattic/jetpack/pull/32005/commits/5c37ab2eb4d605ba9376f165a21df2b084f8f5cf

projects/packages/identity-crisis/src/class-url-secret.php
sergeymitr: Removed: https://github.com/Automattic/jetpack/pull/32005/commits/b65f1919756900972c93eb9fa6c6cb6b2938dfb0

projects/packages/identity-crisis/src/class-url-secret.php
sergeymitr: Good point, done: https://github.com/Automattic/jetpack/pull/32005/commits/1a7294b8dfbf5d4a13d3da68df97814804976584

projects/packages/identity-crisis/src/class-rest-endpoints.php
```diff
@@ -62,6 +63,17 @@ public static function initialize_rest_api() {
 				),
 			)
 		);
+
+		// Fetch URL verification secret.
+		register_rest_route(
+			'jetpack/v4',
+			'/identity-crisis/fetch-url-secret',
```
fgiannar: Not a blocker: Do you think we could make the route more RESTfull by avoiding the verb (`fetch`) in its name? aka `GET /jetpack/v4/identity-crisis/url-secret`

sergeymitr: Good point, updated: https://github.com/Automattic/jetpack/pull/32005/commits/bb6c956dd49a6cff5bd8005c14e2474404fa25bb
@bindlegirl please note that the fetching endpoint URL has changed to `/jetpack/v4/identity-crisis/url-secret`

bindlegirl (approved):
Tests worked as described

PR #32016, opened 2023-07-21
Jetpack Boost: Add uPlot score history graph

sejas:
- Fixes https://github.com/Automattic/boost-cloud/issues/284

## Proposed changes:

It adds a Historical Performance section with a chart to Jetpack Boost page `/wp-admin/admin.php?page=jetpack-boost`. You need to be Boost premium to load some data.

* Add `uPlot` library to `package.json`
* Add new components `BoostScoreGraph`, `PerformanceHistory`
* Add new storybook stories
* Integrate `PerformanceHistory` with svelte to show it in Jetpack Boost page.
* Add jetpack endpoint `/wp-json/jetpack-boost/v1/speed-scores-history`

### Other information:

## Testing instructions:

It's necessary to have purchased a site with Jetpack Boost premium a few days ago. Or you will need to override the data returned by the endpoint.

* Install Jetpack Beta on your site
* Set this branch `add/uplot-boost-graph` for Jetpack Boost.

Screenshot 2023-08-18 at 14 37 34

* Go to `Jetpack > Boost` - `/wp-admin/admin.php?page=jetpack-boost`
* Active the premium plan. The corn runs daily for all premium sites, if you just purchased the premium plan, you will need to wait to see data from the API. 
* Observe the new graph.
* Observe a network request to `/wp-json/jetpack-boost/v1/speed-scores-history`

**Storybook**
* Execute the storybook `cd projects/js-packages/storybook && npm run storybook:dev`
* Access http://localhost:50240/?path=/docs/plugins-boost-performance-history--docs
* Observe the new `PerformanceHistory` component

## Screenshot

jetpack-boost-chart

Storybook:
jetpack-boost-with-tooltip


sejas:
This PR has grown more than I initially expected.
It installs `uPlot` dependency and adds new files on `js-packages/components`, `js-packages/boost-score-api` and `plugins/boost`.
It has been easier to develop and test all these changes in the same branch, but I'm happy to split it into three different PRs if it helps reviewers.

projects/js-packages/components/components/boost-score-graph/get-date-format.ts
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Returns a formatted date based on the provided template and locale.
+ *
+ * @param {string} template - The template used to format the date.
+ * @param {Date} date - The date object to be formatted.
+ * @param {string} [locale='en'] - The locale code specifying the language and region to be used for formatting. Default 'en'.
+ * @returns {string} The formatted date as a string.
+ */
+export default function getDateFormat( template: string, date: Date, locale = 'en' ): string {
```
thingalon, line 12: It looks like this function [is passed to uPlot as its `fmtDate` option](https://github.com/Automattic/jetpack/blob/f22ccc144f4153e09ed1e9eee19064449b23c7fb/projects/js-packages/components/components/boost-score-graph/uplot-line-chart.tsx#L74). Based on [the only docs I could find](https://github.com/leeoniya/uPlot/blob/master/dist/uPlot.d.ts#L144) for this option, it looks like `fmtDate` lets you supply your own function for formatting a date using a template.

I'm confused, and have some questions:

1. Have I understood the role of the `fmtDate` function correctly?
2. Where do the date format strings come from? Are they decided by uPlot, or do we supply them somewhere?
3. If we supply the date strings, can we supply format strings compatible with [WordPress' date functions](https://github.com/WordPress/gutenberg/blob/80f2c2ee0780e90e2bf809b0569f70276fae4a7f/packages/date/README.md) and use those instead of custom code?
4. If uPlot supplies the format functions, do we know a complete list of templates they might send us? It looks like this switches between a selection of pre-set templates, instead of understanding the `{M}`, `{D}`, etc markers. Is that a concern?

I think it would be a good idea to explain in the function comment that this is specifically aimed at uPlot and its role.

projects/js-packages/components/components/boost-score-graph/get-period-date-format.ts
```diff
@@ -0,0 +1,58 @@
+/**
+ * Returns a formatted date based on the provided period and locale.
+ *
+ * @param {string | undefined} period - The period for which the date is formatted. Available options are:
```
thingalon: Missing available options. :)

projects/js-packages/components/components/boost-score-graph/uplot-line-chart.tsx
```diff
@@ -0,0 +1,148 @@
...
+ * @param {uPlot.AlignedData} root0.data - The data for the uPlot chart.
+ * @param {string} root0.period - The period to be used for the chart.
+ * @returns {React.Element} The JSX element representing the UplotLineChart component.
+ */
+export default function UplotLineChart( { data, period }: UplotChartProps ) {
+	const uplot = useRef< uPlot | null >( null );
+	const uplotContainer = useRef( null );
+
+	const [ options ] = useState< uPlot.Options >(
+		useMemo( () => {
```
thingalon: I'm new-ish to React, so this may be a silly question - but what is the purpose of using `useMemo` in the parameters to `useState`? Can't you just pass an initializer function to useState directly?

(Also, I assume you're using state here because options may need to be modified later?)

projects/packages/boost-speed-score/src/class-speed-score-graph-history-request.php
```diff
@@ -0,0 +1,141 @@
...
+
+namespace Automattic\Jetpack\Boost_Speed_Score;
+
+use Automattic\Jetpack\Boost_Core\Lib\Boost_API;
+use Automattic\Jetpack\Boost_Core\Lib\Cacheable;
+
+/**
+ * Class Speed_Score_Graph_History_Request
+ */
+class Speed_Score_Graph_History_Request extends Cacheable {
```
thingalon, line 16: 👍 Nice one; diving into an unfamiliar codebase and picking up on tools like Cacheable.

thingalon (commented):
Nice work so far diving into unfamiliar code. I've left some questions for bits I don't understand.

I also see that the svelte code duplicates some stuff. I know this is a work in progress - so I didn't review those bits in detail. Let me know if you need help on that side of things.

projects/js-packages/components/components/boost-score-graph/get-date-format.ts
sejas: > Have I understood the role of the fmtDate function correctly?

@thingalon, yes you are right!

`uPlot.fmtDate`  is the default date formatted. We can overwrite it to display our custom locale instead of the one provided by the browser. We can also receive the format date string from uPlot and output what we want instead, and overwriting the uPlot's decision.

> Where do the date format strings come from? Are they decided by uPlot, or do we supply them somewhere?

uPlot decides what format to display in the cart, if your range is just a day, it will add hours and minutes, if you range is a year, it will display only the month names. So overwritting this function gives us the control to display the part of the date we want.

> If we supply the date strings, can we supply format strings compatible with [WordPress' date functions](https://github.com/WordPress/gutenberg/blob/80f2c2ee0780e90e2bf809b0569f70276fae4a7f/packages/date/README.md) and use those instead of custom code?

I'll review the link. From what I know we cannot define the format string just interpret them.

> If uPlot supplies the format functions, do we know a complete list of templates they might send us? It looks like this switches between a selection of pre-set templates, instead of understanding the {M}, {D}, etc markers. Is that a concern?

Yes, it does supply the format function and the format template. `uPlot.fmtDate` default function is also customizable but less useful for us, because we would need to pass the values for each locale. Here is a demo view-source:https://leeoniya.github.io/uPlot/demos/months-ru.html


I'll take a deeper look this week to see if uPlot provides a better option for our use case.
The original implementation of this function, `getPeriodDateFormat` and `useResize` come from Calypso: https://github.com/Automattic/wp-calypso/blob/360a4c9a8265a9e5abf5037196fc41945ec58f4b/packages/components/src/chart-uplot/lib/get-date-format.ts

projects/js-packages/components/components/boost-score-graph/uplot-line-chart.tsx
sejas: Thanks for pointing this out.
I removed the `useState` in favor of only using `useMemo` because on our case we won't modify the options at all:
https://github.com/Automattic/jetpack/pull/32016/commits/216b854112ea9fb3be1b05b89b3af7da6f1a996b

projects/js-packages/components/components/boost-score-graph/get-period-date-format.ts
sejas: I removed `get-period-date-format.ts` as it's not useful in our case 👍 .
It helps to display the date on the legend when the mouse is hover.

sejas:
> Nice work so far diving into unfamiliar code. I've left some questions for bits I don't understand.
> 
> I also see that the svelte code duplicates some stuff. I know this is a work in progress - so I didn't review those bits in detail. Let me know if you need help on that side of things.

@thingalon Thank you for the early code review, it helped me a lot to clean the code. I'll be working and learning about uPlot in the next days.

haqadn:
Looks like the trunk merge didn't go very clean

dilirity:
I just started reviewing this, and deployed it to my atomic site (that's the only test site I have that's older than a few days) - https://unrealduck.wpcomstaging.com/

For some reason, the graph is messed up:

CleanShot 2023-08-21 at 12 56 46

dilirity:
@haqadn that fixed it - thanks.

projects/js-packages/components/components/boost-score-graph/get-date-format.ts
dilirity: @sejas hey! Is this convo still relevant?

projects/packages/boost-speed-score/src/class-speed-score.php
```diff
@@ -131,6 +141,62 @@ public function dispatch_speed_score_request( $request ) {
...
 	}
 
+	/**
+	 * Verify and normalize the parametesr for score grpah history request.
+	 *
+	 * @param \WP_REST_Request $request The request object.
+	 *
+	 * @return array|\WP_Error An error to return or an array with the parameters.
+	 */
+	private function process_score_graph_history_params( $request ) {
```
dilirity: If the `args` [are specified](https://developer.wordpress.org/rest-api/extending-the-rest-api/adding-custom-endpoints/#arguments) when registering the rest route, this method would not be necessary.

dilirity: I just realized the code was probably copied over from `process_url_arg`.

projects/js-packages/components/components/boost-score-graph/tooltip.tsx
```diff
@@ -0,0 +1,87 @@
+import { getScoreLetter } from '@automattic/jetpack-boost-score-api';
+import { dateI18n } from '@wordpress/date';
+import { __, sprintf } from '@wordpress/i18n';
+import { type FunctionComponent } from 'react';
+import './style-tooltip.scss';
+import { Period } from '.';
+
+export const Tooltip: FunctionComponent = ( { period }: { period: Period } ) => {
```
dilirity, line 8: I'm probably not used to the syntax, but - why not just expect for `period: Period` here?

```js
{ period }: { period: Period }
```

dilirity (changes requested):
> It has been easier to develop and test all these changes in the same branch, but I'm happy to split it into three different PRs if it helps reviewers.

I probably should have noticed this comment before reviewing it 😅

@pyronaur noticed, that the graph doesn't start with the first available date, but goes back a month (even though there isn't any data).

CleanShot 2023-08-21 at 16 54 59

I agree, that it should probably start with the first available data, and show +1 month forward. We should show -1 back until we have data available. Once we have more than a month, we could probably change the start date. Not sure about this. We can tweak as we go.

jsnmoon (dismissed):
@sejas: This PR contains changes across multiple packages and plugins. Let's split this PR into multiple PRs to reduce the likelihood of regressions and make reviewing easier. I suspect @Automattic/heart-of-gold will need to assert some sort of technical ownership of these new components, so it would also be a good opportunity for some domain knowledge sharing.

Here's what I think would work well for getting everything merged ASAP:

# uPlot PRs

Each of these PRs will have to land sequentially but can be proposed in parallel by setting the target branch as the preceding feature branch to expedite shipping.

(For example, the first PR on branch `first-pr` would target `trunk`; the second PR on branch `second-pr` would target `first-pr`, and so on. Using branch merges instead of rebases would make this workflow go smoothly.)

1. Adding the uPlot dependency to `@automattic/jetpack-components`.
2. Adding the uPlot line chart, corresponding Storybook stories, and unit tests for `getDateFormat` to `@automattic/jetpack-components`. Note that I don't think those unit tests have been added yet.
3. Adding the uPlot tooltip functionality and corresponding storybook to `@automattic/jetpack-components`.

# API endpoint PR

This PR can be spun up in parallel to the uPlot PRs.

- Add and register Speed_Score_Graph_History_Request to the Speed Score class.

# Integration PR

This PR can be spun up in parallel to the preceding PRs. However, I recommend against this unless you've nailed both the uPlot component and the endpoint changes beyond a shadow of a doubt. If you need to change the component property or endpoint schemas in a prior PR, you may need to sync those changes to this integration PR (which can be a hassle).

- Integrate uPlot into the Boost plugin via the uPlot line chart component from `@automattic/jetpack-components`, complete with Storybook stories.

# Feedback

What do you think?

projects/js-packages/components/components/boost-score-graph/tooltip.tsx
haqadn: If we had done `{ period: Period }`, it would not be considered as the type. It would rename the `period` property to `Period` for the context of this function.

haqadn:
> @sejas: This PR contains changes across multiple packages and plugins. Let's split this PR into multiple PRs to reduce the likelihood of regressions and make reviewing easier. I suspect @Automattic/heart-of-gold will need to assert some sort of technical ownership of these new components, so it would also be a good opportunity for some domain knowledge sharing.
> 
> Here's what I think would work well for getting everything merged ASAP:
> 
> # uPlot PRs
> Each of these PRs will have to land sequentially but can be proposed in parallel by setting the target branch as the preceding feature branch to expedite shipping.
> 
> (For example, the first PR on branch `first-pr` would target `trunk`; the second PR on branch `second-pr` would target `first-pr`, and so on. Using branch merges instead of rebases would make this workflow go smoothly.)
> 
> 1. Adding the uPlot dependency to `@automattic/jetpack-components`.
> 2. Adding the uPlot line chart, corresponding Storybook stories, and unit tests for `getDateFormat` to `@automattic/jetpack-components`. Note that I don't think those unit tests have been added yet.
> 3. Adding the uPlot tooltip functionality and corresponding storybook to `@automattic/jetpack-components`.
> 
> # API endpoint PR
> This PR can be spun up in parallel to the uPlot PRs.
> 
> * Add and register Speed_Score_Graph_History_Request to the Speed Score class.
> 
> # Integration PR
> This PR can be spun up in parallel to the preceding PRs. However, I recommend against this unless you've nailed both the uPlot component and the endpoint changes beyond a shadow of a doubt. If you need to change the component property or endpoint schemas in a prior PR, you may need to sync those changes to this integration PR (which can be a hassle).
> 
> * Integrate uPlot into the Boost plugin via the uPlot line chart component from `@automattic/jetpack-components`, complete with Storybook stories.
> 
> # Feedback
> What do you think?

Thank you for suggesting a breakdown of the PR. Looking forward, we will make sure the features are being developed in smaller PRs and in more iterations. This will help ensure the progress is on track aside from making the PRs easier to review.

However, since this has already been reviewed, breaking it down may actually increase the time to finish this and will require re-reviewing. So, we have decided to merge this after the current reviews are addressed. We will iterate on this feature even after this is merged and polish it further.

Do you think this is a bad idea? Let us know if you feel strongly about breaking it down. 

@jsnmoon

projects/js-packages/components/components/boost-score-graph/tooltip.tsx
dilirity: Oh, I meant as in `( period: Period )` - just one parameter.

jsnmoon:
> Looking forward, we will make sure the features are being developed in smaller PRs and in more iterations. This will help ensure the progress is on track aside from making the PRs easier to review.
> 
> However, since this has already been reviewed, breaking it down may actually increase the time to finish this and will require re-reviewing. So, we have decided to merge this after the current reviews are addressed. We will iterate on this feature even after this is merged and polish it further.
> 
> Do you think this is a bad idea? Let us know if you feel strongly about breaking it down. (@haqadn)

It's not ideal, but that sounds like a reasonable compromise in this instance.

Let's find ways to avoid creating large, long-lived PRs like this in the future. Such PRs can become problematic for DevEx (e.g., poor knowledge transfer from creator to reviewers) and create inappropriate expectations for ICs (e.g., "my PR should be perfect before shipping"). Small and quick iterations, as mentioned in your comment, sounds good to me.

projects/js-packages/components/components/boost-score-graph/tooltip.tsx
haqadn: I see what you mean now. Yes, that would look much nicer. However, since this is a React component, all arguments are passed as object properties.

projects/packages/boost-speed-score/src/class-speed-score.php
haqadn: Addressed in https://github.com/Automattic/jetpack/pull/32016/commits/caaf333a060641e976bd194e3f5ddd49b6025eb5

haqadn:
> @pyronaur noticed, that the graph doesn't start with the first available date, but goes back a month (even though there isn't any data).

https://github.com/Automattic/jetpack/pull/32016/commits/f6ed985ed29cd7c62e8cf65fbe8eab31fdd5b00e addressed this. @dilirity

dilirity (approved):
Thanks for addressing the feedback @haqadn!

I think we can finally merge this 🥳 

:shipit: 

EDIT:
Just noticed that this test is failing, but I see it's not marked as required - https://github.com/Automattic/jetpack/pull/32016/checks?check_run_id=16165615669

PR #32017, opened 2023-07-21
pre-commit hook: Clean up PHP linting

anomiex:
## Proposed changes:

* Run `composer phpcs:fix` instead of running `vendor/bin/phpcbf` directly.
* Run `composer phpcs:lint` rather than `composer phpcs:lint:errors`. Our CI checks will run the former, so there's not much point to ignoring the warnings for pre-commit.
* Pass the list of files to the functions instead of using the global, like we do everywhere else.

### Other information:

## Jetpack product discussion

p1689956907073589-slack-CBG1CP4EN

No

## Testing instructions:

* Add something like `$foobar  = 42;` to `jetpack.php` or some other non-excluded php file. Try to commit. The double-space (Generic.Formatting.MultipleStatementAlignment.IncorrectWarning) should get fixed, but then the unused variable warning (VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable) should prevent the commit.

### DIFF
--- a/tools/js-tools/git-hooks/pre-commit-hook.js
+++ b/tools/js-tools/git-hooks/pre-commit-hook.js
@@ -354,9 +354,11 @@ function runPHPLinter( toLintFiles ) {
 
 /**
  * Runs PHPCS against checked PHP files. Exits if the check fails.
+ *
+ * @param {Array} toLintFiles - List of files to lint
  */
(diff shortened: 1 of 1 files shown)


samiff (approved):
Tested as expected and makes sense 👍

PR #32032, opened 2023-07-24
Social: Added upsell links to the sharing section of the Jetpack plugin

spsiddarthan:
Fixes 1205003617340866-as-1205091696786991

## Proposed changes:

* Adde upsell links in the Jetpack Social section of the Sharing screen

### Other information:

Yes, if the upsell links are clicked, they will be tracked

## Testing instructions:

* Go to the sharing screen of the Jetpack plugin and make sure that you don't have a paid jetpack social plan. 
* When you're on the free plan, you will see a link like this asking you to upgrade. 
Screenshot 2023-07-24 at 19 44 36

* Click on the link to upgrade, finish the checkout, and come back to the sharing screen to see the advanced plan's upsell link which will look as shown in the screenshot below
Screenshot 2023-07-25 at 17 57 15

* Finish the upgrade and ensure that no upsell links show up when you return to the sharing screen

### DIFF
--- a/projects/plugins/jetpack/_inc/client/sharing/publicize.jsx
+++ b/projects/plugins/jetpack/_inc/client/sharing/publicize.jsx
@@ -24,6 +24,7 @@ export const Publicize = withModuleSettingsFormHelpers(
 				isLinked = this.props.isLinked,
 				isOfflineMode = this.props.isOfflineMode,
 				siteRawUrl = this.props.siteRawUrl,
+				siteAdminUrl = this.props.siteAdminUrl,
 				isActive = this.props.getOptionValue( 'publicize' ),
 				hasSocialBasicFeatures = this.props.hasSocialBasicFeatures,
 				hasSocialAdvancedFeatures = this.props.hasSocialAdvancedFeatures,
@@ -38,11 +39,8 @@ export const Publicize = withModuleSettingsFormHelpers(
 				isActive &&
 				! hasSocialAdvancedFeatures;
 
-			const jetpackSocialText = __(
-				'Connect your website to the social media networks you use and share your content across all your social accounts with a single click. When you publish a post, it will appear on all connected accounts.',
-				'jetpack'
-			);
-
+			// We need to strip off the trailing slash for the pricing modal to open correctly.
+			const redirectUrl = encodeURIComponent( siteAdminUrl.replace( /\/$/, '' ) );
 			const configCard = () => {
 				if ( unavailableInOfflineMode ) {
 					return;
@@ -88,9 +86,14 @@ export const Publicize = withModuleSettingsFormHelpers(
 								link: getRedirectUrl( 'jetpack-support-publicize' ),
 							} }
 						>
-							<p>{ jetpackSocialText }</p>
+							<p>
+								{ __(
+									'Connect your website to the social media networks you use and share your content across all your social accounts with a single click. When you publish a post, it will appear on all connected accounts.',
+									'jetpack'
+								) }
+							</p>
 							{ showUpgradeLink && (
-								<React.Fragment>
+								<>
 									<p>
 										{ ! hasSocialBasicFeatures
 											? createInterpolateElement(
@@ -105,9 +108,7 @@ export const Publicize = withModuleSettingsFormHelpers(
 																	'jetpack-plugin-admin-page-sharings-screen',
 																	{
 																		site: siteRawUrl,
-																		query:
-																			'redirect_to=' +
-																			encodeURIComponent( window.location.href.split( '#' )[ 0 ] ),
+																		query: 'redirect_to=' + redirectUrl,
 																	}
 																) }
 															/>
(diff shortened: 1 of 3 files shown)


projects/plugins/jetpack/_inc/client/sharing/publicize.jsx
```diff
@@ -24,8 +24,35 @@ export const Publicize = withModuleSettingsFormHelpers(
...
+				'jetpack'
+			);
+
+			const jetpackSocialBasicUpgradeText = __( 'Upgrade to a Jetpack Social plan', 'jetpack' );
+			const jetpackSocialAdvancedText = __(
+				'Upgrade to the Jetpack Social Advanced plan',
+				'jetpack'
+			);
+
+			const jetpackSocialBasicUpgradeTextSuffix = __(
```
gmjuhasz: I wonder if we shouldn't split this sentence like this, as it looks good code wise but could be challenging for translators to translate and keep the meaning intact. I think creating two interpolate elements, one for advanced and one for the other would be easier for them like [here](https://github.com/Automattic/jetpack/blob/trunk/projects/js-packages/publicize-components/src/components/form/index.js#L130)

projects/plugins/jetpack/_inc/client/sharing/publicize.jsx
```diff
@@ -71,12 +88,56 @@ export const Publicize = withModuleSettingsFormHelpers(
...
+										{ ! hasSocialBasicFeatures
+											? createInterpolateElement(
+													__(
+														'<moreInfo>Upgrade to a Jetpack Social plan</moreInfo> to get unlimited shares and advanced media sharing options.',
+														'jetpack'
+													),
+													{
+														moreInfo: (
+															<a
+																href={ getRedirectUrl(
```
jeherve, line 107: Since these links link outside of wp-admin, I would recommend using ExternalLink here.

projects/plugins/jetpack/_inc/client/sharing/publicize.jsx
```diff
@@ -71,12 +88,56 @@ export const Publicize = withModuleSettingsFormHelpers(
...
+																	'jetpack-plugin-admin-page-sharings-screen',
+																	{
+																		site: siteRawUrl,
+																		query:
+																			'redirect_to=' +
+																			encodeURIComponent( window.location.href.split( '#' )[ 0 ] ),
```
jeherve: Could we use `siteAdminUrl` for those links?

projects/plugins/jetpack/_inc/client/sharing/publicize.jsx
```diff
@@ -71,12 +88,56 @@ export const Publicize = withModuleSettingsFormHelpers(
...
 						>
-							<p>
-								{ __(
-									'Connect your website to the social media networks you use and share your content across all your social accounts with a single click. When you publish a post, it will appear on all connected accounts.',
-									'jetpack'
-								) }
-							</p>
+							<p>{ jetpackSocialText }</p>
+							{ showUpgradeLink && (
+								<React.Fragment>
```
jeherve: `<>` would work as well here.

projects/plugins/jetpack/_inc/client/sharing/publicize.jsx
```diff
@@ -24,8 +25,24 @@ export const Publicize = withModuleSettingsFormHelpers(
...
 				userCanManageModules = this.props.userCanManageModules;
 
+			const showUpgradeLink =
+				! isAtomicSite &&
+				activeFeatures &&
+				activeFeatures.length > 0 &&
+				isActive &&
+				! hasSocialAdvancedFeatures;
+
+			const jetpackSocialText = __(
```
jeherve: Since we only use that string once, maybe we do not need to extract it into its own constant; we could display it in place right away.

projects/plugins/jetpack/_inc/client/sharing/publicize.jsx
spsiddarthan: I thought of using the `ExternalLink`, but it doesn't accept the `target` param, which means the link always opens in a new tab and I think case we would like for it to open n the same. Do you think we can do anything else to work around that?

projects/plugins/jetpack/_inc/client/sharing/publicize.jsx
spsiddarthan: I have done that, but I needed to strip off the last slash in it. This is because, on the pricing page, the final URL looks like this `https://cloud.jetpack.com/pricing/?site=sidtest.jurassic.tube&view=products&redirect_to=<redirect-url>#jetpack_social_advanced_yearly` We need the `#jetpack_social_advanced_yearly` bit in the end to open the pricing modal. 

If I did not strip off the trailing slash from siteAdminUrl, `jetpack_social_advanced_yearly` is seen as a part of the redirect URL and the pricing modal doesn't open up.

spsiddarthan:
Thanks for your review, @jeherve! I think have I addressed them and responded to 1 item regarding the `ExternalLink`. Let me know your thoughts

projects/plugins/jetpack/_inc/client/sharing/publicize.jsx
jeherve: Let's leave it like that yet. 👍

jeherve (approved):
Tests well for me! 🚢

gmjuhasz (approved):
Looks good! 👍

PR #32033, opened 2023-07-24
Socia: Refactor notice dismiss functionality to support time periods

gmjuhasz:
This change updates the current functionality for the notice dismiss feature, so it can be given a timestamp to send it away for some time rather than forever.

## Proposed changes:

* Added a new parameter to the endpoint that is the reappearance time

### Other information:

## Testing instructions:

* Open up your local site with this PR. If you already dismissed the Instagram notice run `jetpack docker wp option delete jetpack_social_dismissed_notices`
* Open the Admin page or the editor, and dismiss the notice.
  * Run `jetpack docker wp option get jetpack_social_dismissed_notices`, it should return an object with instagram in it, and the value should be 0.
  * Refresh the page, the notice should not come back.
* Clear the option again with the command provided above.
* In `projects/js-packages/publicize-components/src/components/form/index.js` change `dismissNotice( 'instagram' )` to `dismissNotice( 'instagram', 1 )`.
  * This should dismiss the notice for 1 minute. Run the command above to check the actual value in the option, it should be the Unix timestamp of the current time + one minute.
  * Refresh the page before the minute elapsed, the notice should not appear.
  * Refresh after 1 minute, the notice should come back again. Clicking close now should send it away for another minute


projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -69,9 +69,7 @@ export default function PublicizeForm( {
 	);
 
 	const shouldShowInstagramNotice =
-		! hasInstagramConnection &&
-		isInstagramConnectionSupported &&
-		! dismissedNotices.includes( 'instagram' );
+		! hasInstagramConnection && isInstagramConnectionSupported && shouldShowNotice( 'instagram' );
```
spsiddarthan: With the new notices that we plan to add, shall we define a constant for all the notices in this file? Something like `const INSTAGRAM_NOTICE = 'instagram'` and then use that above?

projects/js-packages/publicize-components/src/hooks/use-dismiss-notice/index.js
```diff
@@ -2,6 +2,13 @@ import { getJetpackData } from '@automattic/jetpack-shared-extension-utils';
 import apiFetch from '@wordpress/api-fetch';
 import { useCallback, useMemo, useState } from 'react';
 
+const calculateReappearanceTime = minutes => {
```
spsiddarthan: Like we chatted earlier, let's change this to seconds and the param could passed could be `secondsToSnooze`

projects/js-packages/publicize-components/src/hooks/use-dismiss-notice/index.js
```diff
@@ -2,6 +2,13 @@ import { getJetpackData } from '@automattic/jetpack-shared-extension-utils';
...
+	if ( minutes === -1 ) {
```
spsiddarthan: May be add a comment here saying -1 means never show the notice again.

spsiddarthan (commented):
This tests well, Geri! Just had a few comments.

projects/js-packages/publicize-components/src/components/form/index.js
gmjuhasz: Good idea. I'll do that

projects/js-packages/publicize-components/src/hooks/use-dismiss-notice/index.js
gmjuhasz: I added the comment on the `dismissNotice` function, because that's what you call, this one is only an internal function used by `dismissNotice`, but I can comment here too!

spsiddarthan (dismissed):
LGTM

spsiddarthan (approved):

PR #32043, opened 2023-07-24
Subscriptions: Change conditions for displaying modal

edanzer:
Fixes #

## Proposed changes:

* Adds extra checks and conditions for displaying (or not) the subscribe modal. Collectively these all avoid a situation where a user keeps re-seeing the subscribe modal even though they have subscribed or dismissed it.
* Do not show if user has already subscribed (`jetpack_blog_subscribe_` cookie is set).
* Do not show if `subscribe` query parameter is set (set when user first submits the subscribe form)
* If user dismisses the cookie, we add 1-day `jetpack-subscribe-modal-dismissed` cookie, and do not re-show the modal while that cookie exists.

### Other information:

## Testing instructions:

1) Setup
   - Load this branch in your local Jetpack environment - you will need a jurassic tube connected site. 
   - Add the `add_filter( 'jetpack_subscriptions_modal_enabled', '__return_true', 20 );` filter. The feature is behind a feature flag, and this will enable it. You can add this to tools/docker/mu-plugins/0-sandbox.php, or temporarily in any file that loads on your instance. 
   - Go to Jetpack > Settings > Discussion > Newsletters and enable the "Enable subscriber modal". If you do not see that option enable "Let visitors subscribe to new posts and comments via email" and refresh the page.

2) Test dismissing the modal. 
   - Go to any single post on the front end, scroll, and confirm the modal shows. 
   - Dismiss the modal, refresh the page, and confirm the modal does not show again.
   - Open Dev Tools > Applications > Cookies, and confirm the `jetpack-subscribe-modal-dismissed` is set. 
   
3) Test the subscribe query parameter.
   - Delete the `jetpack-subscribe-modal-dismissed` cookie from the last test so the modal will show again. 
   - Add `?subscribe=anythinghere` to the url and refresh the page.
   - Confirm the modal never show.

4) Test subscribing to the blog.
   - Remove the subscribe query parameter from the last test.
   - Visit any single post, scroll down, add a comment, check the boxes to subscribe to the blog, and submit.
   - Once page reloads, confirm you see the `jetpack_blog_subscribe_...` cookie. 
   - If needed, delete the `jetpack-subscribe-modal-dismissed` cookie again. 
   - Refresh the page and confirm the modal never shows.

### DIFF
--- a/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.js
+++ b/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.js
@@ -2,12 +2,14 @@ const { domReady } = wp;
 
 domReady( function () {
 	const modal = document.getElementsByClassName( 'jetpack-subscribe-modal' )[ 0 ];
-	const close = document.getElementsByClassName( 'jetpack-subscribe-modal__close' )[ 0 ];
 
 	if ( ! modal ) {
 		return;
 	}
 
+	const close = document.getElementsByClassName( 'jetpack-subscribe-modal__close' )[ 0 ];
+	const modalDismissedCookie = 'jetpack_subscribe_modal_dismissed';
+	const hasModalDismissedCookie = document.cookie.indexOf( modalDismissedCookie ) > -1;
 	let hasLoaded = false;
 	let isScrolling;
 
@@ -15,7 +17,7 @@ domReady( function () {
 		window.clearTimeout( isScrolling );
 
 		isScrolling = setTimeout( function () {
-			if ( ! hasLoaded ) {
+			if ( ! hasLoaded && ! hasModalDismissedCookie ) {
 				modal.classList.toggle( 'open' );
 				hasLoaded = true;
 			}
@@ -24,11 +26,19 @@ domReady( function () {
 
 	close.onclick = function () {
 		modal.classList.toggle( 'open' );
+		setModalDismissedCookie();
 	};
 
 	window.onclick = function ( event ) {
 		if ( event.target === modal ) {
 			modal.style.display = 'none';
+			setModalDismissedCookie();
 		}
 	};
+
+	function setModalDismissedCookie() {
+		// Expires in 1 day
+		const expires = new Date( Date.now() + 86400 * 1000 ).toUTCString();
+		document.cookie = `${ modalDismissedCookie }=true; expires=${ expires };`;
+	}
 } );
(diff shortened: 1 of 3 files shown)


edanzer:
I actually wonder if I should update to use this method in place of some of my current checks here: 
https://github.com/Automattic/jetpack/blob/5fdf5d7f2f534e1bc062eb5c37dc42c3f9b9d257/projects/plugins/jetpack/modules/subscriptions/views.php#L556

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.js
```diff
@@ -2,20 +2,22 @@ const { domReady } = wp;
...
 	const modal = document.getElementsByClassName( 'jetpack-subscribe-modal' )[ 0 ];
-	const close = document.getElementsByClassName( 'jetpack-subscribe-modal__close' )[ 0 ];
 
 	if ( ! modal ) {
 		return;
 	}
 
+	const close = document.getElementsByClassName( 'jetpack-subscribe-modal__close' )[ 0 ];
+	const modalDismissedCookie = 'jetpack_subscribe_modal_dismissed';
+	const hasModalDismissedCookie = document.cookie.indexOf( modalDismissedCookie ) > -1;
```
simison: I think `document.cookie` always exists, but I wanted to mention in case we need some protection, e.g. `document.cookie?.indexOf()`

edanzer: I've added a check for this. For some reason, adding the optional chaining operator was giving an eslint error.

edanzer:
Updated: I've added a check based on @simison's comment above, added an extra conditional check using the existing is_current_user_subscribed() method in views.php, and rebased.

lezama:
should we use this [cookie state class](https://github.com/Automattic/jetpack/blob/6040a756958a58ff13ca536810be259d0147dc54/projects/packages/status/src/class-cookiestate.php) somehow?

edanzer:
> should we use this [cookie state class](https://github.com/Automattic/jetpack/blob/6040a756958a58ff13ca536810be259d0147dc54/projects/packages/status/src/class-cookiestate.php) somehow?

Hey @lezama - Good question. Just spent some time looking at this. Interesting construct and seems like it could plausibly be used, but...
   - For setting cookies, we're only doing that once here, for a client-side-only 'dismissed' cookie, so the CookieState class is not as relevant for that. 
   - What could be interesting is if information about the user and user subscription status was stored using CookieState, you could key on that for checks and conditionals. But best I can tell, that information doesn't seem to be stored via CookieState. So we'd need to refactor other existing methods to store information that way first. If I'm understanding that right, it may not be worth adding that initially.

lezama (approved):

PR #32044, opened 2023-07-24
Prevent use of JP_Memberships to get connected account ID

millerf:
No change in feature. Just use DB based on blog id when on WPCOM

Duplicate from D116913-code


n3f (approved):
WFM

PR #32052, opened 2023-07-25
Introduce 'jetpack_woocommerce_analytics_event_props' filter

### LINKED ISSUE #32050: Enhancement: Introduce filter to append properties to WooCommerce analytics Tracks events
### Impacted plugin

Jetpack

### What

I would like to introduce additional properties to WooCommerce analytics Tracks events (`woocommerceanalytics`)

### How

_No response_

manospsyx:
Fixes #32050
- #32050 

**Related Project Thread**
peapX7-2Vz-p2#comment-3798

## Proposed changes:

Introduces a filter to modify/append WooCommerce analytics event properties. Will be utilized to add a Woo Express `host` parameter. 

I considered adding a `host` key directly, but decided to take the filter approach to make it easier for others to add props in the future.

### Other information:

## Testing instructions:

N/A

### DIFF
--- a/projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-universal.php
+++ b/projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-universal.php
@@ -143,6 +143,15 @@ class Jetpack_WooCommerce_Analytics_Universal {
 		}
 		$product_details = $this->get_product_details( $product );
 
+		/**
+		 * Allow defining custom event properties in WooCommerce Analytics.
+		 *
+		 * @module woocommerce-analytics
+		 *
+		 * @since $$next-version$$
+		 *
+		 * @param array $all_props Array of event props to be filtered.
+		 */
 		$all_props = apply_filters(
 			'jetpack_woocommerce_analytics_event_props',
 			array_merge(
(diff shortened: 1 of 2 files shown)


PanosSynetos (dismissed):
Thanks @manospsyx - The filter is helpful. I don't see any reason not to introduce this

Let's see what Jetpack Crew has to say

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-universal.php
```diff
@@ -143,9 +143,12 @@ public function process_event_properties( $event_name, $product_id, $properties
 		}
 		$product_details = $this->get_product_details( $product );
 
-		$all_props = array_merge(
-			$properties,
-			$this->get_common_properties()
+		$all_props = apply_filters(
```
jeherve, line 155: I would recommend adding a docblock to that filter, so our [Jetpack codex](https://developer.jetpack.com/) can be automatically updated with some information about the filter. Something like this:
```suggestion
		/**
		 * Allow defining custom even properties in WooCommerce Analytics.
		 *
		 * @module woocommerce-analytics
		 *
		 * @since $$next-version$$
		 *
		 * @param array $all_props Array of standard event props.
		 */
		$all_props = apply_filters(
```

jeherve (changes requested):

jeherve (approved):

PanosSynetos (commented):
👌

PR #31914, opened 2023-07-17
Integrate Pexels to Gutenberg media sidebar

rcrdortiz:
Part of https://github.com/Automattic/wp-calypso/issues/71232

## Proposed changes:

* Added Pexels to Gutenberg's Media Sidebar

### Pexels image selector
Captura de Pantalla 2023-07-18 a las 15 07 11

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* Go to a Jetpack-connected site.
* Edit or create a new Post.
* Open the Media Sidebar
* You should see the Pexels integration.

PR #32058, opened 2023-07-25
External Sources in Media Inserter: Google Photos media inserter aware of authentication

rcrdortiz:
Fixes https://github.com/Automattic/dotcom-forge/issues/3124

## Proposed changes:

* Google Photos is now aware of the connection status when the user authorizes Google Photos through the media block.
* Google Photos now shows a non-empty tooltip.
* Restored Pexels search results to 10.
* Fixed an error in how `with-media` is setting the isAuthenticated state.

https://github.com/Automattic/jetpack/assets/1989914/7c506362-ee30-4fce-bb50-c6d44de2ff0c

### Other information:

## Testing instructions:

Keep in mind that this branch hasn't been created from trunk and branches off the branch where we [add pexels to the sidebar](https://github.com/Automattic/jetpack/pull/31914).

* Go to a Jetpack-connected site.
* Edit or create a new Post.
* Make sure you already have the Google Photos service isn't connected.
* Open the Media Sidebar.
* You should not see the Google Photos integration.
* Insert an image block and connect to Google Photos.
* Open the Media Sidebar.
* You should now see the Google Photos integration listed and working.


projects/plugins/jetpack/extensions/shared/external-media/store/actions.ts
```diff
@@ -0,0 +1,15 @@
+export const SET_AUTHENTICATED = 'SET_AUTHENTICATED';
+
+export type SetAuthenticatedAction = {
+	type: typeof SET_AUTHENTICATED;
+	payload: boolean;
+};
+
+export default {
+	setAuthenticated: ( isAuthenticated: boolean ) => {
```
mmtr: This should also accept a source param, since every source handle the authentication separately.

Otherwise, I think users could use the image block to insert a Pexels image which will call `setAuthenticated( true )` which will cause the `isGooglePhotosConnected` function to return true even if the connection to Google Photos has not been established.

mmtr (commented):
I'm unable to test this. There are several JS errors that prevent blocks from being added to the editor – I guess a rebase might fix it.

Screenshot 2023-07-25 at 17 21 14

rcrdortiz:
> I'm unable to test this. There are several JS errors that prevent blocks from being added to the editor – I guess a rebase might fix it.
> 
> Screenshot 2023-07-25 at 17 21 14

Not related to rebase, I had a commit that needed to be pushed to the repo.

projects/plugins/jetpack/extensions/shared/external-media/media-service/index.ts
```diff
@@ -177,24 +195,46 @@ export const getPexelsMediaCategory = () =>
...
+ * Checks if a given MediaSource is authenticated in the store.
  *
- * @param {ConnectedMediaSourceCallback} isConnectedCallback
- * @returns {void}
+ * @param {MediaSource} source - MediaSource to check.
+ * @returns {boolean} True if the MediaSource is authenticated false otherwise.
  */
-export const isGooglePhotosConnected = ( isConnectedCallback: ConnectedMediaSourceCallback ) =>
-	isMediaSourceConnected( MediaSource.GooglePhotos, isConnectedCallback );
+const isMediaSourceAuthenticated = ( source: MediaSource ) =>
```
mmtr: Nitpick (non-blocking): It's not clear to me what's the difference between `isMediaSourceConnected` and `isMediaSourceAuthenticated`, since the authentication is the same as a connection in the context of a external media source.

projects/plugins/jetpack/extensions/shared/external-media/sources/with-media.js
```diff
@@ -8,8 +8,10 @@ import { UP, DOWN, LEFT, RIGHT } from '@wordpress/keycodes';
 import classnames from 'classnames';
 import { uniqBy } from 'lodash';
 import { PATH_RECENT } from '../constants';
+import { authenticateMediaSource } from '../media-service';
+import { MediaSource } from '../media-service/types';
 
-export default function withMedia() {
+export default function withMedia( mediaSource = MediaSource.Unknown ) {
```
mmtr, line 14: :chefkiss:

projects/plugins/jetpack/extensions/shared/external-media/sources/with-media.js
```diff
@@ -153,8 +159,6 @@ export default function withMedia() {
 				const path = this.getRequestUrl( url );
 				const method = 'GET';
 
-				this.setAuthenticated( true );
```
mmtr, line 156: Nice catch!

mmtr (commented):
Can we exclude videos from Google Photos? They are inserted as images which is confusing:

Screenshot 2023-07-27 at 11 10 39

projects/plugins/jetpack/extensions/shared/external-media/media-service/index.ts
rcrdortiz: The `isMediaSourceConnected` use case checks if we don't get an error from calling the API, this is needed since `isMediaSourceAuthenticated` only is initialized/available when a media block is inserted. On the first use case, we're not checking if it's authorized. The JSDoc states that this checks if the MediaSource is authenticated in the store.

How do you think we could make it clearer, adding a comment or changing some naming, or anything else?

rcrdortiz: Changed to a name that's a bit clearer about the intent of the use case.

rcrdortiz:
> Can we exclude videos from Google Photos? They are inserted as images which is confusing:
> 
> Screenshot 2023-07-27 at 11 10 39

Nice catch! In theory, the Inserter should be capable of inserting videos if we specify the correct media type when we build the Media Category, I think it can't be mixed though so I've filtered the videos out.

mmtr (approved):
💥

PR #32067, opened 2023-07-25
Launchpad: Check the Newsletter type on the enabled function

valterlorran:
Related to:
* D116904-code
* https://github.com/Automattic/wp-calypso/issues/79788

## Proposed changes:

We decided to move most of the logic that validates if the user can access the Newsletter task lists to Jetpack.

* Adds the check for the type of Newsletter(Free or Paid) on the functions that check whether the launchpad is enabled.
* Checks the site intent to ensure we only display the task list to sites with the newsletter intent.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* Apply the patch D116904-code on your sandbox
* Then, use the command linked on the [comment](https://github.com/Automattic/jetpack/pull/32067#issuecomment-1650256431) to apply this changes to your sandbox
* Follow the testing instructions on D116904-code

PR #32068, opened 2023-07-25
Launchpad: Add site goals option to Jetpack sync

valterlorran:
Related to:
* https://github.com/Automattic/jetpack/pull/32067
* D117113-code

## Proposed changes:

* Adds the site_goals option to the Jetpack sync, so it can be accessible from Atomic sites. This option will be used to determine if the site has a Free or Paid Newsletter. This way, we can decide which launchpad we display for the user since we have two separate task lists for each case(free or paid).

### Other information:

## Testing instructions:

* Follow the instructions on D117113-code


sixhours (approved):
This is working for me!

PR #32071, opened 2023-07-25
Memberships: Ensure endpoint handles source param for wpcom

edanzer:
**Important: This PR will have a [companion backend diff/PR](D117077-code) that will need to be merged first to WP.com.**

## Proposed changes:

This PR updates the memberships endpoint handling of the source param, which is used to determine where a user is returned to when they come back from connecting stripe. For WP.com sites, we were not passing the source param to the method that gets the stripe connect url. Now we are. 

Note this PR should not change how this endpoint works for self-hosted Jetpack sites, only for requests made from WordPress.com sites.

### Other information:

## Jetpack product discussion

None. 

No.

## Testing instructions:

Please test via this [Calypso frontend PR](https://github.com/Automattic/wp-calypso/pull/79671), which uses the source parameter. Testing instructions are there.


lezama (approved):
Makes sense after the backend change

PR #32081, opened 2023-07-26
My Jetpack: rename the namespace of the JWT endpoint

retrofox:
On this PR:

* Rename the namespace of the JWT endpoint from `my-jetpack/v1` to `jetpack/v4`
* Register the endpoint only if it isn't registered. Currently, [it's registered by the Jetpack plugin](https://github.com/Automattic/jetpack/blob/96f64ddaa59d36b244814311017cd4ebd40d348f/projects/plugins/jetpack/_inc/lib/class.core-rest-api-endpoints.php#L81).

The idea behind these changes is to be able to have always the JWT endpoint available, at least for all projects that rely on My Jetpack.

_About avoiding an endpoint double registration. I tested registering the same endpoint twice. It doesn't break the app, but still, it's better to check it._

Fixes #

## Proposed changes:

* My Jetpack: rename the namespace of the JWT endpoint, and register it only when it isn't already registered

## Testing instructions:

* Take a testing site
* Do not enable any Jetpack plugin/product. No Jetpack plugin and other plugins that depend on My Jetpack
* Confirm the `jetpack/v4/jetpack-ai-jwt` endpoint is not available
* Active the Jetpack plugin
* Confirm now it's available
* Active another product that depends on My Jetpack, for instance, the Social standalone plugin
* Confirm the endpoint is available 
* Deactivate the Jetpack plugin 
* Confirm that the endpoint is still available

You can hit the `<your-testing-site>/wp-json/jetpack/v4/` to check if the endpoint is registered.

Screenshot 2023-07-26 at 11 59 37

### DIFF
--- a/projects/packages/my-jetpack/changelog/update-rename-jwt-endpoint-namespace
+++ b/projects/packages/my-jetpack/changelog/update-rename-jwt-endpoint-namespace
@@ -1,0 +1,4 @@
+Significance: patch
+Type: changed
+
+My Jetpack: Rename the namespace of the JWT endpoint, and register it only when it isn't already registered
--- a/projects/packages/my-jetpack/package.json
+++ b/projects/packages/my-jetpack/package.json
@@ -1,7 +1,7 @@
 {
 	"private": true,
 	"name": "@automattic/jetpack-my-jetpack",
-	"version": "3.2.0",
+	"version": "3.2.1-alpha",
 	"description": "WP Admin page with information and configuration shared among all Jetpack stand-alone plugins",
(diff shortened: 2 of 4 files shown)


lhkowalski (approved):
It's working as described.

- The endpoint is not registered without the plugin:

Screen Shot 2023-07-26 at 15 20 04

- With Jetpack plugin, it's registered:

Screen Shot 2023-07-26 at 15 22 19

- With Jetpack + VideoPress plugin, it's also registered:

Screen Shot 2023-07-26 at 15 23 16

- With just VideoPress:

Screen Shot 2023-07-26 at 15 24 27

---

A question: did we consider creating a new namespace only for AI, for example `jetpack-ai/v1/*`?

PR #32084, opened 2023-07-26
VideoPress: handle uploading video files when dropping in the editor canvas

### LINKED ISSUE #30584: VideoPress: Add a VideoPress video v6 block when user drops video file in the editor
Currently dropping a file creates a `core/video` block even if v6 is available, immediately showing a prompt for conversion.

p1683658923915969-slack-C02LT75D3

retrofox:
This PR implements the proper block transform in the VideoPress video block to be able to handle the uploading process when the user drops off the video in the editor canvas 

Fixes #30584

## Proposed changes:

* VideoPress: handle uploading video files when dropping in the editor canvas

### Other information:

## Testing instructions:

* Go to the block editor
* Test with `Jetpack plugin` and/or `VideoPress plugin` activated.
* Drop off some files in the editor canvas
* Confirm the app creates a VideoPress video instance ONLY for video files
* Confirm the videos are uploaded as usual

In the example below, I'm dropping four files where only two are video files

https://github.com/Automattic/jetpack/assets/77539/807114ce-5056-404e-a918-5bc77d5a97cf

### DIFF
--- a/projects/packages/videopress/src/client/block-editor/utils/video.ts
+++ b/projects/packages/videopress/src/client/block-editor/utils/video.ts
@@ -12,3 +12,13 @@ export function isVideoFile( file: File ): boolean {
 
 	return file.type.startsWith( 'video/' );
 }
+
+/**
+ * Filter an array of files to only include video files.
+ *
+ * @param {File[]} files - Array of files to filter.
+ * @returns {File[]}       Array of video files.
+ */
+export function filterVideoFiles( files: File[] ): File[] {
+	return files.filter( isVideoFile );
+}
--- a/projects/packages/videopress/src/client/block-editor/utils/video.ts
+++ b/projects/packages/videopress/src/client/block-editor/utils/video.ts
@@ -1,0 +1,14 @@
+/**
+ * Pretty basic helper to check if a file is a video
+ * based on its mime type.
+ *
+ * @param {File} file - File to check.
+ * @returns {boolean}   Whether the file is a video.
+ */
+export function isVideoFile( file: File ): boolean {
+	if ( ! file?.type ) {
+		return false;
+	}
+
+	return file.type.startsWith( 'video/' );
+}
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -2129,6 +2129,9 @@ importers:
       '@wordpress/api-fetch':
         specifier: 6.34.0
         version: 6.34.0
+      '@wordpress/blob':
+        specifier: 3.37.0
+        version: 3.37.0
       '@wordpress/block-editor':
         specifier: 12.5.0
         version: 12.5.0(@types/react-dom@18.2.6)(@types/react@18.2.14)(react-dom@18.2.0)(react@18.2.0)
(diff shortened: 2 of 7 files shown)


projects/packages/videopress/src/client/block-editor/utils/video.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @param {File} file - File to check.
+ * @returns {boolean}   Whether the file is a video.
+ */
+export function isVideoFile( file: File ): boolean {
+	if ( ! file?.type ) {
+		return false;
+	}
+
+	return file.type.startsWith( 'video/' );
```
dhasilva, line 13:
```suggestion
	const allowedMimeTypes = Object.values(
		window?.Jetpack_Editor_Initial_State?.allowedMimeTypes ?? {}
	).filter( type => type.startsWith( 'video/' ) );
	return allowedMimeTypes.includes( file.type );
```

dhasilva: To filter only the allowed mimetypes

retrofox: Nice suggestion. We need to ensure the right functionality also in the VideoPress standalone plugin.

dhasilva: ref p1690387124731989/1690369223.294339-slack-C03TA48NSUX
For future reference, `Jetpack_Editor_Initial_State` is only available with Jetpack, not when using only VP.

dhasilva (approved):
LGTM :rocket:

lhkowalski (approved):
Code looks good. Tested on a Simple site and it works as expected.

The inserted block, uploading:

Screen Shot 2023-07-26 at 13 20 49

The final result after the upload:

Screen Shot 2023-07-26 at 13 21 05

The code of the inserted block:

Screen Shot 2023-07-26 at 13 21 46

Imran92:
Hey folks! 👋 I am working on a relevant issue. From the video example attached above in the PR detail, I can see that even though four files were being dragged and dropped in the editor, only the Video files were getting transformed to blocks and the rest of the files are not even there (got removed) on the editor. So I was curious is it an intentional behavior?

Thanks!

@retrofox @dhasilva

oskosk:
@Imran92, it was an unintended side effect of how the block is handling the drop by using an effect 

image

Imran92:
Thanks so much for the very promp reply @oskosk ! <3

Basically I was working on this issue - https://github.com/Automattic/jetpack/issues/40469,  and after some digging I found that this was causing it. Before starting to think/work on a fix, I just wanted to confirm if it was intentional due to some reason I wasn't aware of.

Thanks again!

Imran92:
I’ve submitted a potential fix for the issue [here](https://github.com/Automattic/jetpack/pull/42312). Apologies for adding you all as reviewers to my PR and for the extra pings you're receiving, really sorry for that. As I'm not too familiar with the Jetpack codebase, having some of you take a look would give me more confidence that I’m not unintentionally breaking anything.

Thank you for your time and feedback!

PR #32087, opened 2023-07-26
Social: Add upsell nudge for Basic users

gmjuhasz:
We are not promoting the Advanced plan much as of now. This change adds a notice to the editor that should appear once every 3 months after dismissed.

## Proposed changes:

* Added new prop to the `usePublicizeConfig` hook to check if user is on the Basic plan 77f930f6514953690d3f53fb0a620fb72ab240ba
* Added the new upsell notice to the editor 0b7119f59f4cbbdcc6c5f1a0e616c82890dc8602

### Other information:

## Jetpack product discussion

1205003617340866-as-1205091696786993

No.
## Testing instructions:

* If you have no plans, or have the Advanced plan, you should not see the new notice in the editor. Check this.
* Buy the Basic plan. Now you should see the upsell in the editor below the custom message field.
  * Click the External Link, it should lead you to the Jetpack Cloud pricing page with the Advanced Social modal open
  * Click the close button on the notice. It should dismiss it, and should not come back on refresh.
  * Run `jetpack docker wp option get jetpack_social_dismissed_notices` and check the value of `advanced-upgrade-nudge`. It should be a timestamp, and the value should be 30 days from now. https://www.unixtimestamp.com/
  * Running `jetpack docker wp option delete jetpack_social_dismissed_notices` should bring back the notice. __Note: It brings back the Instagram notice too.__
    * Closing either of the two should work separately not dismissing the other.


projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -108,6 +111,11 @@ export default function PublicizeForm( {
 		[ autosave, isEditedPostDirty ]
 	);
 
+	const onAdvancedNudgeDismiss = useCallback(
+		() => dismissNotice( 'advanced-upgrade-nudge', MONTH_IN_SECONDS ),
```
spsiddarthan: Let's make this `3 * MONTH_IN_SECONDS`

spsiddarthan (commented):
I tested this and it works well except for one corner use-case. Let's increase the time to 3 months as well to snooze the notice. 

- I did not dismiss the advanced plan upsell notice, but instead bought the complete plan. When I cancelled the complete plan and refreshed the plan data, I coudln't see the notice pop back. Can you try this as well and see if we are able to replicate it?  I did this on ninja, so I am not able to check the options. 
- We need to make sure that the notice doesn't show up on simple sites after sandboxing this patch. I will check that as well before the final review, but please check that as well if you haven't already.

gmjuhasz:
> * I did not dismiss the advanced plan upsell notice, but instead bought the complete plan. When I cancelled the complete plan and refreshed the plan data, I coudln't see the notice pop back. Can you try this as well and see if we are able to replicate it?  I did this on ninja, so I am not able to check the options.

It should only come back if you buy the Basic plan again, as we only so this to Basic plan users.

> * We need to make sure that the notice doesn't show up on simple sites after sandboxing this patch. I will check that as well before the final review, but please check that as well if you haven't already.

I'll take another look at that too!

spsiddarthan:
I was just thinking about this some more and to avoid an overload of notices in the editor, I think we should be careful about the number of notices we show in the editor.

Now that we are storing the Instagram notice dismiss time, shall we use that to delay the time at which the advanced plan upsell is shown? 

- If both the Instagram and the upsell notice for the advanced plan have to be shown, we only show the Instagram notice. 

- If only the social advanced plan notice has to be shown, we check if the Instagram notice was dismissed at least a month ago and the social advanced plan notice was never shown before or  the social advanced plan notice was shown  3 months ago, something like `showAdvancedPlanUpsell = !showInstagramNotice && ( upsellNeverShownBefore || upsellShown90DaysAgo ) && instagramNoticeDismissed30daysAgo`

I know it's a bit complex, let me know your thoughts!

spsiddarthan:
> If both the Instagram and the upsell notice for the advanced plan have to be shown, we only show the Instagram notice.

Or we could simply show the `showAdvancedPlanUpsell` for the basic plan users and don't show them the Instagram notice at all to them.

Both approaches are fine I guess.

gmjuhasz:
Interesting thoughts.
I don't think we should treat the one logic based on the other as they are giving different information. Instagram notice is just a basic FYI that there is a new feature, while the upsell is nudging them to consider upgrading. Only showing the upsell would mean that we do not notify them about Instagram, so they could miss that, as the connection screen is not typically in a basic publishing user flow.

On the other hand I see the point of the notice overwhelming and I agree with that. With the next spike we are planning to only show the upsell if they shared 3 posts already. This would minimize the chance for them to see both at the same time. 
If this is a concern, we could not merge this before that, what do you think of that?

spsiddarthan:
>This would minimize the chance for them to see both at the same time.
If this is a concern, we could not merge this before that, what do you think of that?

Perfect, let's do that then!

projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -43,13 +44,15 @@ const checkConnectionCode = ( connection, code ) =>
...
  * @param {string} props.adminUrl                         - URL af the plugin's admin page to redirect to after a plan upgrade
+ * @param {boolean} props.hasBasicPlan                    - Whether the site has a basic plan
  * @returns {object}                                      - Publicize form component.
  */
 export default function PublicizeForm( {
 	isPublicizeEnabled,
 	isPublicizeDisabledBySitePlan,
 	numberOfSharesRemaining = null,
 	isEnhancedPublishingEnabled = false,
+	hasBasicPlan = false,
```
pablinos: I think it would be good to discuss this a bit. We spent some time making sure that we only had feature checks and not plan checks, although we're specifically wanting to promote the advanced plan to basic plan customers so maybe this is ok?

projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -338,6 +346,21 @@ export default function PublicizeForm( {
 							/>
 						</>
 					) }
+					{ hasBasicPlan && shouldShowNotice( NOTICES.advancedUpgrade ) && (
```
pablinos: I think there's a good chance you've considered this and it's why this task is blocked, but would it be possible to move the plan/feature check to the backend? Then `shouldShowNotice` would be false, and we wouldn't need the month check etc.

pablinos (commented):
I think it's probably a bigger refactor, but it seems like we're going to implement the same logic on the server and the frontend. We only get if the notice has been dismissed coming through from the backend, but perhaps that should be a combination of the display criteria (whether enough time has passed, the feature availability etc.) and whether it has been dismissed.

Essentially I think [`useDismissNotice`](https://github.com/Automattic/jetpack/blob/0fe80d04930c9a3cea6baaef29df73d7466e0d28/projects/js-packages/publicize-components/src/hooks/use-dismiss-notice/index.js#L36) should maybe be `useNotice` or something, and would perhaps look at notices to show rather than the dismissed ones.

spsiddarthan (approved):
Tests well against the patch I created!

PR #32089, opened 2023-07-26
Jetpack Sync: Extract Sync Queue storage handling to an external class.

bisko:
This is a spin-off of #31318

## Proposed changes:

Extract storage of Sync Queue actions to a separate class. This is to make the merge of Custom Table work easier and in smaller batches, to avoid reviewing and testing a huge PR.

### Other information:

## Jetpack product discussion

p9dueE-7e6-p2

No

## Testing instructions:

* Apply PR
* Run the Docker PHPUnit tests - `jetpack docker phpunit` and make sure they pass.

### DIFF
--- a/projects/packages/sync/src/class-queue.php
+++ b/projects/packages/sync/src/class-queue.php
@@ -7,6 +7,7 @@
 
 namespace Automattic\Jetpack\Sync;
 
+use Automattic\Jetpack\Sync\Queue\Queue_Storage_Options;
 use WP_Error;
 
 /**
@@ -38,6 +39,15 @@ class Queue {
 	public $random_int;
 
 	/**
+	 * Queue Storage instance where we'll store the queue items.
+	 *
+	 * For now it's only the Options table. To be updated to include the Custom table in future updates.
+	 *
+	 * @var Queue_Storage_Options|null
+	 */
+	public $queue_storage = null;
+
+	/**
 	 * Queue constructor.
 	 *
 	 * @param string $id Name of the queue.
@@ -46,6 +56,9 @@ class Queue {
 		$this->id           = str_replace( '-', '_', $id ); // Necessary to ensure we don't have ID collisions in the SQL.
 		$this->row_iterator = 0;
 		$this->random_int   = wp_rand( 1, 1000000 );
+
+		// Initialize the storage with the Options table backend. To be changed in subsequent updates to include the logic to switch to Custom Table.
+		$this->queue_storage = new Queue_Storage_Options( $this->id );
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


projects/packages/sync/src/class-queue.php
```diff
@@ -54,34 +67,27 @@ public function __construct( $id ) {
 	 * @param object $item Event object to add to queue.
 	 */
 	public function add( $item ) {
```
fgiannar, line 71: This method didn't return so far. I checked and believe it's ok to do so, so in this case lets add a corresponding docblock on the return value or keep the previous logic.

projects/packages/sync/src/class-queue.php
```diff
@@ -228,15 +181,7 @@ public function size() {
 	 * @return bool
 	 */
 	public function has_any_items() {
```
fgiannar, line 185: I'd suggest we keep the `SELECT exists` logic here as it's a more performant query. EXISTS can stop as soon as the logical test proves true, but COUNT(*) must count every row, even after it knows one row has passed the test.

projects/packages/sync/src/class-queue.php
```diff
@@ -251,7 +196,8 @@ public function checkout( $buffer_size ) {
 			return new WP_Error( 'unclosed_buffer', 'There is an unclosed buffer' );
 		}
 
-		$buffer_id = uniqid();
+		// TODO check if adding a prefix is going to be a problem
+		$buffer_id = uniqid( '', true );
```
fgiannar, line 202: Nice! 

>  If called without $more_entropy = true, uniqid() calls sleep for at least 1 microsecond. uniqid($prefix, $more_entropy = true) will prevent function from sleeping.

https://blog.kevingomez.fr/til/2015/07/26/why-is-uniqid-slow/

projects/packages/sync/src/sync-queue/class-queue-storage-options.php
```diff
@@ -0,0 +1,286 @@
...
+
+		return $rows_added;
+	}
+
+	/**
+	 * Return $max_count items from the queue, including their value string length.
+	 *
+	 * @param int $max_count How many items to fetch from the queue.
+	 *
+	 * @return array|object|\stdClass[]|null
```
fgiannar:
```suggestion
	 * @return \stdClass[]|null
```

projects/packages/sync/src/sync-queue/class-queue-storage-options.php
```diff
@@ -0,0 +1,286 @@
...
+
+		return $items;
+	}
+
+	/**
+	 * Fetches items with specific IDs from the Queue.
+	 *
+	 * @param array $items_ids Items IDs to fetch from the queue.
+	 *
+	 * @return array|object|\stdClass[]|null
```
fgiannar:
```suggestion
	 * @return \stdClass[]|null
```

projects/packages/sync/src/class-queue.php
```diff
@@ -367,34 +302,37 @@ public function checkout_with_memory_limit( $max_memory, $max_buffer_size = 500
...
-			$max_item_id = $item_with_size->id;
+			$item_ids_to_fetch[] = $item_with_size->id;
 		}
 
-		$query = $wpdb->prepare(
-			"SELECT option_name AS id, option_value AS value FROM $wpdb->options WHERE option_name >= %s and option_name <= %s ORDER BY option_name ASC",
-			$min_item_id,
-			$max_item_id
-		);
+		$current_items = $this->queue_storage->fetch_items_by_ids( $item_ids_to_fetch );
```
fgiannar, line 310: I do have some performance concerns about replacing the `WHERE option_name >= %s and option_name <= %s` with `WHERE IN` for at least 500 items. Any reasons we didn't keep the original logic here?

projects/packages/sync/src/sync-queue/class-queue-storage-options.php
```diff
@@ -0,0 +1,286 @@
...
+	 * Delete items with specific IDs from the queue.
+	 *
+	 * @param array $ids IDs of the items to remove from the queue.
+	 *
+	 * @return bool|int|\mysqli_result|resource|null
+	 */
+	public function delete_items_by_ids( $ids ) {
+		global $wpdb;
+		// TODO check if it's working properly - no need to delete all options in the table if the params are not right
+		$ids_placeholders = implode( ', ', array_fill( 0, count( $ids ), '%s' ) );
```
fgiannar, line 276: Maybe add a check here for ids to ensure it's a non empty array

fgiannar (changes requested):
Great work, Biser!

Left a few inline comments, nothing major though.

It would be nice to introduce an interface when we add the table storage backend for the Queue :)

projects/packages/sync/src/class-queue.php
bisko: One of the benefits of a capable IDE. PHPStorm mentioned that the parameter is a much better option.

projects/packages/sync/src/class-queue.php
bisko: There should not be an issue, as we're calculating memory limitations beforehand.

In this case we first calculate the size and items that we want to include in the batch and then fetch only the items we need.

If we see issues, we can always `array_chunk` the IDs and fetch them in smaller batches.

fgiannar (approved):
LGTM

PR #32095, opened 2023-07-26
Refactor/subscriber counts into redux store

chrismccluskey:
Closes 88-gh-Automattic/gold

## Proposed changes:

* Fetches subscriber counts from data store instead of passing it to the child component as props from parents.

### Other information:

## Testing instructions:

* On an installation of WordPress, make sure the Jetpack Beta plugin is installed, and then activate this branch.
* Check this patch out to sun on sandbox.
* Ensure subscribe and public-api are pointed to sandbox.
* Go to [the earn settings page](https://wordpress.com/earn/payments/) and ensure Stripe is setup in test mode.
* Create a new page or post and add the subscribe block.
* Publish the page and try subscribing. Confirm the block works as expected.
* Return to edit the page with the subscription button. Click on the "Newsletter Visibility" drop-down and select "paid subscribers". You should not see a message indicating that you don't have any subscribers. Instead, you should see a message saying this will be sent to your _n_ subscribers.


projects/plugins/jetpack/extensions/store/membership-products/reducer.js
```diff
@@ -23,6 +30,34 @@ export default function reducer( state = DEFAULT_STATE, action ) {
...
+					...state.counts,
+					emailSubscribers: action.socialFollowers,
+				},
+			};
+		case 'SET_PAID_SUBSCRIBER_COUNT':
+			return {
+				...state,
+				counts: {
+					...state.counts,
+					paidSubscribers: action.socialFollowers,
```
millerf: Might be an error there
```suggestion
					paidSubscribers: action.paidSubscribers,
```

projects/plugins/jetpack/extensions/store/membership-products/reducer.js
```diff
@@ -23,6 +30,34 @@ export default function reducer( state = DEFAULT_STATE, action ) {
...
+					...state.counts,
+					socialFollowers: action.socialFollowers,
+				},
+			};
+		case 'SET_EMAIL_SUBSCRIBER_COUNT':
+			return {
+				...state,
+				counts: {
+					...state.counts,
+					emailSubscribers: action.socialFollowers,
```
millerf: might be an error there

```suggestion
					emailSubscribers: action.emailSubscribers,
```

projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
```diff
@@ -115,18 +99,16 @@ const NewsletterDisabledPanels = () => (
...
-	emailSubscribers,
-	paidSubscribers,
-	isModuleActive,
-	showMisconfigurationWarning,
-	showPreviewModal,
-} ) {
+function NewsletterPrePublishSettingsPanel( { setPostMeta, isModuleActive, showPreviewModal } ) {
 	const { tracks } = useAnalytics();
 	const { changeStatus, isLoadingModules, isChangingStatus } = useModuleStatus( name );
+	const { accessLevel } = useSelect( select => {
```
millerf: Wouldn't writing it this way be a bit more readable? 

```suggestion
	const accessLevel = useSelect( select =>  select( 'jetpack/membership-products').getAccessLevel() );
```

Instead of wrapping it in an object to extract it straight away?

projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
```diff
@@ -194,12 +169,20 @@ function NewsletterPrePublishSettingsPanel( {
...
+	const { accessLevel, showMisconfigurationWarning } = useSelect( select => {
+		const { getAccessLevel, getShowMisconfigurationWarning } = select(
+			'jetpack/membership-products'
+		);
+
+		return {
+			accessLevel: getAccessLevel(),
+			showMisconfigurationWarning: getShowMisconfigurationWarning(),
+		};
+	} );
```
millerf: Same comment than previously...

I would write it:

```
const accessLevel = useSelect( select =>  select( 'jetpack/membership-products').getAccessLevel() );
const showMisconfigurationWarning = useSelect( select =>  select( 'jetpack/membership-products').getShowMisconfigurationWarning() );
```

projects/plugins/jetpack/extensions/store/membership-products/resolvers.js
```diff
@@ -61,6 +73,41 @@ const mapAPIResponseToMembershipProductsStoreData = ( response, registry, dispat
...
+
+	return response;
+};
+
+const mapSubscriberCountsAPIResponseToMembershipProductsStoreData = (
+	response,
+	registry,
+	dispatch
+) => {
+	dispatch( setSocialFollowerCount( response.counts.social_followers ) );
```
millerf: Does each dispatch() triggers a new component refresh?

projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
chrismccluskey: Yes, this is much more readable. I'll make these changes, thanks!

projects/plugins/jetpack/extensions/store/membership-products/resolvers.js
chrismccluskey: I'm not sure if it does. I'll need to check this and see. From memory, there were no visuals which might indicate a refresh, although it's possible one still may have happened.

projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
millerf: [Object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) is pretty useful and powerful in JS

projects/plugins/jetpack/extensions/store/membership-products/resolvers.js
millerf: It is fine. Let's keep it that way...

projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
```diff
@@ -323,12 +314,8 @@ export default function SubscribePanels() {
 		if ( ! isModuleActive ) {
 			return;
 		}
-		getSubscriberCounts( counts => {
-			setEmailSubscribers( counts.email_subscribers );
-			setSocialFollowers( counts.social_followers );
-			setPaidSubscribers( counts.paid_subscribers );
-		} );
-	}, [ isModuleActive ] );
+		getSubscriberCounts();
```
millerf: That should not be needed as it should be handled by resolvers?

millerf (approved):
LGTM. The code looks a bit more clean!

I added a couple of commits:
- e7e1842f16 to move getShowMisconfigurationWarning to its own utility fucntion
- 278a6c0300 I tried to also remove setPostMeta, but on changing the panel is not refreshed (it would need some extra work to pass it in the data store), so I reverted it. Not needed really.
- c6e3f496ee to prevent calling getSubscriberCounts() out of context. There was no need to have an event for each of the separate counts. Now it make things easier to read.

Nice work anyway for the first React PR in a while.

jeherve:
This is introducing a regression. When the Subscriptions module is not active, the `wpcom/v2/subscribers/counts` is not available so we should not attempt to make a request. Before this PR, we waited until the module was active to make a request ; now we get a 404 as soon as we load the post editor:

image

I assume that eventually we'll get rid of `api.js` entirely, to avoid the duplication?

millerf:
> This is introducing a regression. When the Subscriptions module is not active, the `wpcom/v2/subscribers/counts` is not available so we should not attempt to make a request. Before this PR, we waited until the module was active to make a request ; now we get a 404 as soon as we load the post editor:
> 
> image
> I assume that eventually we'll get rid of `api.js` entirely, to avoid the duplication?

Interesting. I am not sure why the data is called... AFAIK it should not get triggered... I'll look into it.
Thanks for finding this out..

millerf:
> I assume that eventually we'll get rid of api.js entirely, to avoid the duplication?

Yes that is a good idea :)

PR #32096, opened 2023-07-26
Jetpack AI: Support different upgrade prompts and free limits

lhkowalski:
## Proposed changes:

* Add support for different upgrade prompts, based on the value of the new `upgrade-prompt-type` flag from the `ai-assistant-feature` endpoint
* Add a different prompt for VIP sites, without the button and the call to action to follow the upgrade path

Screen Shot 2023-07-27 at 14 16 15

* The D116636-code adds a getter to read the number of free requests available for a given site, so we are changing the Jetpack AI Helper to use it, because this will allow us to set a different limit based on any information linked to the `$blog_id`(site type, for example)

### Other information:

## Jetpack product discussion

p1690380406050879-slack-C054LN8RNVA

pbtFPC-5uC-p2

## Testing instructions:

* Apply the changes from this PR to your sandbox
* Follow the instructions described on D116636-code to mark your test site as VIP (you don't need to follow the testing instructions from there, just the VIP flagging)
* On the API Console, check the result from the `wpcom/v2` `/sites/$wpcom_site/jetpack-ai/ai-assistant-feature` endpoint; it should contain the new field `upgrade-type`, and since you flagged your site as VIP, you will see `upgrade-type: "vip"` on the response:

Screen Shot 2023-07-31 at 18 00 18

* To test the new upgrade prompt for VIP, we need to simulate that the site reached the limit to be able to see the upgrade prompt; since the limit is 1000, it's hard to test by hand, so we are going to:
   * edit the code for the `jetpack-ai-query.php` file on your sandbox
   * add the following snippet as the first line of the `run_query` function and save the file:

```
return new \WP_Error( 'error_quota_exceeded', 'You exceeded your current quota', array( 'status' => 429 ) );
```

* We also need to point our test site to our sandbox, using the `JETPACK__SANDBOX_DOMAIN` constant; this is necessary for Jurassic Ninja sites or for Atomic sites and can be done using the `Jetpack Constants` page created by Jetpack Beta
* Finally, we also need to enable global http access on the sandbox so requests coming from your JN or Atomic test site are served by it; details here: paJDYF-WR-p2
* With the quota simulation snippet in place, the testing site running this branch and pointing to your sandbox, go to the block editor of your test site, add an AI Assistant block and execute some action (example: `a haiku about the stars`)
* Since the backend will respond with a quota exceeded response, you will see the upgrade banner on the block:

Screen Shot 2023-07-31 at 18 24 50

* Don't forget to remove the snippet from your sandbox
* Don't forget to disable global HTTP access on your sandbox as well

### DIFF
--- a/projects/plugins/jetpack/_inc/lib/class-jetpack-ai-helper.php
+++ b/projects/plugins/jetpack/_inc/lib/class-jetpack-ai-helper.php
@@ -321,10 +321,11 @@ class Jetpack_AI_Helper {
 			$requests_limit = \OpenAI_Limit_Usage::get_free_requests_limit( $blog_id );
 			$requests_count = \OpenAI_Request_Count::get_count( $blog_id );
 
-			/*
-			 * Check if the site requires an upgrade.
-			 */
+			// Check if the site requires an upgrade.
 			$require_upgrade = $is_over_limit && ! $has_ai_assistant_feature;
+
+			// Determine the upgrade type
+			$upgrade_type = wpcom_is_vip( $blog_id ) ? 'vip' : 'default';
 
 			return array(
 				'has-feature'          => $has_ai_assistant_feature,
@@ -332,7 +333,7 @@ class Jetpack_AI_Helper {
 				'requests-count'       => $requests_count,
 				'requests-limit'       => $requests_limit,
 				'site-require-upgrade' => $require_upgrade,
-				'upgrade-prompt-type'  => wpcom_is_vip( $blog_id ) ? 'vip' : 'default',
+				'upgrade-type'         => $upgrade_type,
 			);
 		}
 
--- a/projects/plugins/jetpack/_inc/lib/class-jetpack-ai-helper.php
+++ b/projects/plugins/jetpack/_inc/lib/class-jetpack-ai-helper.php
@@ -318,7 +318,7 @@ class Jetpack_AI_Helper {
 
 			$blog_id        = get_current_blog_id();
 			$is_over_limit  = \OpenAI_Limit_Usage::is_blog_over_request_limit( $blog_id );
-			$requests_limit = \OpenAI_Limit_Usage::NUM_FREE_REQUESTS_LIMIT;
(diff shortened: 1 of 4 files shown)


retrofox (dismissed):
LGTM 🪨 

~# 🚨 IMPORTANT 🚨~
Ship D116636-code before merging it to ensure the method is available.

Patch deployed  r272069-wpcom

projects/plugins/jetpack/_inc/lib/class-jetpack-ai-helper.php
```diff
@@ -332,6 +332,7 @@ public static function get_ai_assistance_feature() {
 				'requests-count'       => $requests_count,
 				'requests-limit'       => $requests_limit,
 				'site-require-upgrade' => $require_upgrade,
+				'upgrade-prompt-type'  => wpcom_is_vip( $blog_id ) ? 'vip' : 'default',
```
retrofox: I'm not totally convinced about the parameter name :-/. Can we add something about the site type? Maybe just `is-vip`?
```suggestion
				'is-vip'  => wpcom_is_vip( $blog_id ),
```

retrofox: I so... you'll have to update the rest of the code

lhkowalski: I thought about sending the site type, but it seemed a bit off, kind of unrelated to the purpose of the endpoint. The choice for `upgrade-prompt-type` was a way to flag different types of prompts, like `vip`, `default`, `atomic-trial`, `jetpack-bonus-trial` or whatever we can come up with.

Maybe the `vip` mention is too specific, and we could call it `no-upgrade-button` option, for example.

What do you think?

retrofox: [I've just renamed the field](9262a55798d9d53b1091e5900824eb1ae352f28c) to `upgrade-type`. Also, defined the type for it: `vip`, `defautl`

nunyvega (approved):
Works as advised!

PR #32098, opened 2023-07-26
Donation block preserves invalid amounts across tabs

### LINKED ISSUE #30669: Donations block :: invalid amount "sticks" for all tabs
### Impacted plugin

Jetpack

### Quick summary

Fixing #30530 uncovered an error If an amount sets the `isInvalid` flag, all the tabs show the invalid value.

Kapture 2023-05-11 at 17 48 40

### Steps to reproduce

1. Add a donations block
2. Set the amount below the minimum value for  a currency
3. Try changing the tabs.

Should show the correct value for the selected tab.

### What actually happened

Continued to show the invalid amount.

### Impact

All

### Available workarounds?

Yes, difficult to implement

_No response_

### Logs or notes

_No response_

n3f:
Fixes #30669.

## Proposed changes:
* Adds effect handler when the Amount component value changes.

This change effectively "forgets" the invalid amount when the tab is switched.  This approach might need additional consideration.

### Other information:

## Testing instructions:

* Get the code into your testing environment. (For WPCOM run `bin/jetpack-downloader test jetpack  earn/donation/unstick-invalid-amounts`)
* Load a donations block
* Set currency to USD
* Set an amount to less than $0.50.
* Change the tab
* View that the amount changes.
* Also notice that switching back to "error" tab the value is the last valid amount.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/donations/amount.js
+++ b/projects/plugins/jetpack/extensions/blocks/donations/amount.js
@@ -78,6 +78,11 @@ const Amount = ( {
 		setEditedValue( formatCurrency( value, currency, { symbol: '' } ) );
 	}, [ currency, isFocused, isInvalid, setAmount, value ] );
 
+	useEffect( () => {
+		setAmount( formatCurrency( value, currency, { symbol: '' } ) );
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [ value ] );
+
 	return (
 		<div
 			className={ classnames( 'donations__amount', className, {
--- a/projects/plugins/jetpack/changelog/earn-donation-unstick-invalid-amounts
+++ b/projects/plugins/jetpack/changelog/earn-donation-unstick-invalid-amounts
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
+Fix tab switching by adding a value handler for the Amount component.


millerf:
@n3f For the record

```
 $ bin/jetpack-downloader test earn/donation/unstick-invalid-amounts
Unrecognized plugin earn/donation/unstick-invalid-amounts
```

millerf:
Oh. It was missing `jetpack`. I fixed the description.

projects/plugins/jetpack/extensions/blocks/donations/amount.js
```diff
@@ -78,6 +78,11 @@ const Amount = ( {
 		setEditedValue( formatCurrency( value, currency, { symbol: '' } ) );
 	}, [ currency, isFocused, isInvalid, setAmount, value ] );
 
+	useEffect( () => {
+		setAmount( formatCurrency( value, currency, { symbol: '' } ) );
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [ value ] );
```
millerf, line 84: I think removing the dependencies would run this when the component is mounted, therefore achieving the expected effect, and you could remove the `// eslint-disable-next-line react-hooks/exhaustive-deps`.

```suggestion
	} );
```

millerf (approved):
It works!

I added a small comment to prevent ignoring linting...

projects/plugins/jetpack/extensions/blocks/donations/amount.js
n3f: This doesn't actually fix the issue! (I know, right!) The problem here is the Amount component doesn't have the normal react lifecycle.  It's actually the same component on each tab, so the initial effects only apply when the block is first loaded.  When you change the tab it's not recreating the Amount it's just changing the amounts shown.  I tried the exhaustive list too (to remove the line), but that creates an infinite loop.  We really only want to run this when the value has changed, I can apply this change:

```diff
diff --git a/projects/plugins/jetpack/extensions/blocks/donations/amount.js b/projects/plugins/jetpack/extensions/blocks/donations/amount.js
index 7479565180..1dd8dd8397 100644
--- a/projects/plugins/jetpack/extensions/blocks/donations/amount.js
+++ b/projects/plugins/jetpack/extensions/blocks/donations/amount.js
@@ -4,6 +4,14 @@ import { useCallback, useEffect, useRef, useState } from '@wordpress/element';
 import classnames from 'classnames';
 import { minimumTransactionAmountForCurrency, parseAmount } from '../../shared/currencies';
 
+function usePreviousValue( value ) {
+       const ref = useRef();
+       useEffect( () => {
+               ref.current = value;
+       } );
+       return ref.current;
+}
+
 const Amount = ( {
        className = null,
        currency = null,
@@ -19,6 +27,7 @@ const Amount = ( {
        const [ isFocused, setIsFocused ] = useState( false );
        const [ isInvalid, setIsInvalid ] = useState( false );
        const richTextRef = useRef( null );
+       const previousValue = usePreviousValue( value );
 
        const setAmount = useCallback(
                amount => {
@@ -79,9 +88,10 @@ const Amount = ( {
        }, [ currency, isFocused, isInvalid, setAmount, value ] );
 
        useEffect( () => {
-               setAmount( formatCurrency( value, currency, { symbol: '' } ) );
-               // eslint-disable-next-line react-hooks/exhaustive-deps
-       }, [ value ] );
+               if ( previousValue !== value ) {
+                       setAmount( formatCurrency( value, currency, { symbol: '' } ) );
+               }
+       }, [ currency, previousValue, setAmount, value ] );
 
        return (
                <div
```

What do you think?

millerf: > This doesn't actually fix the issue! (I know, right!) 

I was sure I tried and it was working...  Damn... 

> What do you think?

I am not that familiar with `useRef()`... So I trust you if you think it works...
Otherwise leave the actual one with the es-lint comment. I think it reads better than this version...

Also, there are projects in motion to refactor this whole block and merge it with the Team51 one: p81Rsd-1n8-p2
and pewaj1-kC-p2#comment-252, so it does not matter too much if we keep the disable es-lint comment :) ...

I also think this issue is such a small one it is not worth spending more time on it....

PR #32101, opened 2023-07-27
Update Paywall Block styles

lezama:
## Proposed changes:

* Add similar styles to the "More" Block

### Other information:

## Jetpack product discussion
p2-peKye1-fd

No

## Testing instructions:

* Add define( 'JETPACK_BLOCKS_VARIATION', 'beta' ); to your setup – I use mu-plugins/0-sandbox.php
* Check that the new Paywall block shows like this:

Screenshot 2023-07-26 at 22 44 12

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/paywall/edit.js
+++ b/projects/plugins/jetpack/extensions/blocks/paywall/edit.js
@@ -1,19 +1,15 @@
-import { BlockIcon } from '@wordpress/block-editor';
-import { Placeholder } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 import './editor.scss';
-import icon from './icon';
 
 function PaywallEdit( { className } ) {
+	const text = __( 'Paywall', 'jetpack' );
+	const style = {
+		width: `${ text.length + 1.2 }em`,
+	};
+
 	return (
 		<div className={ className }>
-			<Placeholder
-				label={ __( 'Paywall', 'jetpack' ) }
-				instructions={ __( 'Instructions go here.', 'jetpack' ) }
-				icon={ <BlockIcon icon={ icon } /> }
-			>
-				{ __( 'User input goes here?', 'jetpack' ) }
-			</Placeholder>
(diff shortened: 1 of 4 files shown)


simison (dismissed):
Great!

Screenshot 2023-07-27 at 11 34 30

image

projects/plugins/jetpack/extensions/blocks/paywall/edit.js
```diff
@@ -1,19 +1,15 @@
-import { BlockIcon } from '@wordpress/block-editor';
-import { Placeholder } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 import './editor.scss';
-import icon from './icon';
 
 function PaywallEdit( { className } ) {
+	const text = __( 'Paywall', 'jetpack' );
```
simison, line 5: We'll probably want to have something that is instructive, e.g. "Paid content below" or "Divides paid and free content", but you can keep "Paywall" until we have final copy ready to send for translating.

projects/plugins/jetpack/extensions/blocks/paywall/index.js
```diff
@@ -21,26 +21,9 @@ export const settings = {
 	category: 'grow',
```
simison: Let's move to `earn` category where other payment related stuff is. Can be in different PR.

simison (approved):
image

projects/plugins/jetpack/extensions/blocks/paywall/index.js
simison: Moved in https://github.com/Automattic/jetpack/pull/32101/commits/b6c45991b22bd7027a3aa3983153ad4477fda66a

simison:
@lezama  I noticed the gap between dash and the label looks a bit better in core blocks:

Screenshot 2023-07-28 at 18 17 43

I'd follow the "page break" styles here. The "more" block has label that one can edit, so it's different.

I also tested that dark background works nicely 👍 

Screenshot 2023-07-28 at 18 20 47

PR #32106, opened 2023-07-27
WP Super Cache: fix bug when dynamic caching and gzip compression are enabled

### LINKED ISSUE #32074: Super Cache: fix dynamic caching and compressed content
When dynamic caching is enabled it stops the plugin creating a gzipped cache file. Unfortunately, the gzip header is still added to the accompanying meta file, if compression is enabled.
This results in content negotiation errors on the blog and compression must be disabled.

Two ways to fix it:
1. Remove the gzip header. Pages will be served uncompressed to the site visitor. A simple fix.
2. Keep the gzip header. The page is still cached uncompressed but when serving it, compress the page after running through the dynamic content filter.

I think we should do 2 because of how important file size is to visitor experience and SEO.
Steps:
1. When serving a dynamic cache page, check the gzip header is there and compression is enabled.
2. Compress page with gzip.
3. Serve page by sending headers and page content.

donnchawp:
The function "wp_cache_serve_cache_file()" serves cached content to web site visitors. It can serve both wp-cache and supercache files. Serving supercache files to anonymous users works fine. However, if compression and dynamic caching are enabled logged in/known users will see a "content negotiation" error when they view a cached page.

This happened because the system that builds cache file does not create a compressed file if dynamic caching is enabled. However, the "Content-Encoding: gzip" header was still added to the associated meta file. This resulted in the browser being sent that header, followed by uncompressed output. The Content-Length was incorrect too.

Fixes #32074 

## Proposed changes:
* Modify some of the debug logging to make it clearer what's happening.
* Add extra debug logging.
* Fix the Content-Length header when serving dynamic content.
* Compress the cached content after applying the `wpsc_cachedata` filter to run the dynamic content.

### Other information:

No

## Testing instructions:
* Apply PR to test site.
* Make sure that dynamic caching and "Compress pages so they’re served more quickly to visitors." are enabled on the Advanced settings page.
* Visit a page on your blog as a logged in user. Reload the page.
* Visit a page in a private/incognito browser.
* Repeat tests with dynamic caching disabled.
* Repeat tests with compression disabled.

To see the problem in action, test trunk in the same way. You'll get a "Content Encoding Error" when you load a cached page as a "known user", aka logged in user.

### DIFF
--- a/projects/plugins/super-cache/wp-cache-phase2.php
+++ b/projects/plugins/super-cache/wp-cache-phase2.php
@@ -158,7 +158,9 @@ function wp_cache_serve_cache_file() {
 
 	extract( wp_super_cache_init() ); // $key, $cache_filename, $meta_file, $cache_file, $meta_pathname
 
-	// Look for wp-cache file+meta for the current URL
+	// Look for wp-cache file + meta file for the current URL
+	// If we can't find them, we will look for supercache html files. These files don't have any meta data
+	// which is why the code below does more work setting up the headers, etc.
 	if (
 		! defined( 'WPSC_SUPERCACHE_ONLY' ) &&
 		(
--- a/projects/plugins/super-cache/wp-cache-phase2.php
+++ b/projects/plugins/super-cache/wp-cache-phase2.php
@@ -158,6 +158,7 @@ function wp_cache_serve_cache_file() {
 
 	extract( wp_super_cache_init() ); // $key, $cache_filename, $meta_file, $cache_file, $meta_pathname
 
+	// Look for wp-cache file+meta for the current URL
 	if (
 		! defined( 'WPSC_SUPERCACHE_ONLY' ) &&
 		(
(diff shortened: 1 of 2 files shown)


thingalon:
I'm seeing some odd behaviour in testing, and I'm not sure it's intentional:

When I load a page with dynamic caching enabled, I see the page load correctly with dynamic content the first time - but on subsequent loads, it reverts to showing the replaceable tags without inserting the dynamic content.

I'm seeing that behaviour with or without this PR, so I'm not sure if I'm missing something here.

projects/plugins/super-cache/wp-cache-phase2.php
```diff
@@ -315,54 +317,71 @@ function wp_cache_serve_cache_file() {
 
 	$cache_file = do_cacheaction( 'wp_cache_served_cache_file', $cache_file );
 	// Sometimes the gzip headers are lost. Make sure html returned isn't compressed!
-	if ( $cache_compression && $wp_cache_gzip_encoding && ! in_array( 'Content-Encoding: ' . $wp_cache_gzip_encoding, $meta['headers'] ) ) {
-		$ungzip = true;
-		wp_cache_debug( 'GZIP headers not found. Force uncompressed output.', 1 );
-	} else {
-		$ungzip = false;
+	$do_not_serve_gzip_data = true;
```
thingalon, line 320: I like how much clearer this variable name is.

thingalon (approved):
Nice! I've tested this, and it fixes the issue.

The changes look sensible -- and I particularly appreciate extra comments, and code changes that make it easier to follow the code. Nice one @donnchawp :)

PR #30079, opened 2023-04-13
Blocks: update CLI to only input strings in descriptions

jeherve:
See #26792

## Proposed changes:

Let's not encourage the use of non-strings in block descriptions; only strings should be used from now on.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

* N/A

## Does this pull request change what data or activity we track or use?

* No

## Testing instructions:

* Not much to test here, but you could try to run `jetpack docker wp jetpack scaffold block something` and check the files being created.

PR #32110, opened 2023-07-27
Blocks: use placeholder instead of random version in CLI template

jeherve:
## Proposed changes:

Since we have the `$$next-version$$` that can be used whenever we need to provide a version number, let's use that instead of a hardcoded version that will never be up-to-date for too long.

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

Not much to test here, but you could try to run `jetpack docker wp jetpack scaffold block "thing"`.


projects/plugins/jetpack/wp-cli-templates/block-register-php.mustache
```diff
@@ -2,7 +2,7 @@
 /**
  * {{ title }} Block.
  *
- * @since 12.1
+ * @since $$next-version$$
```
anomiex: The problem is that when the next release happens this will be replaced like every other instance of `$$next-version$$`. You already tried this in #30079, for example, and it got replaced in #30127.

Can you do something like `{{ next-version-literal }}` here in the template and have the code in whatever is supplying the value for `{{ title }}` above also supply this literal string for that variable?

jeherve: heh, I had forgotten that I had already done that. Your idea sounds good. done in 0ca1f959a0b4b9d1c302f92d5787eb0d5c9db074

projects/plugins/jetpack/class.jetpack-cli.php
```diff
@@ -2010,6 +2010,7 @@ public function block( $args, $assoc_args ) {
 			"$path/$slug.php"     => self::render_block_file(
 				'block-register-php',
 				array(
+					'nextVersion'      => '$$next-version$$',
```
anomiex: Have to disguise it somehow so replace-next-version-tag.sh doesn't see it here. There are various ways you might do that, e.g.
```suggestion
					'nextVersion'      => "\x24\x24next-version$$", // Escapes to hide the string from tools/replace-next-version-tag.sh
```

zinigor (approved):

PR #32111, opened 2023-07-27
Jetpack Sync: Add logic to switch to Custom Table if a configuration flag is enabled.

bisko:
This is a spin-off of #31318 and continuation of #32089

## Proposed changes:

This PR adds the Custom table logic, hidden behind a feature flag. 

It doesn't contain any of the table initialization logic, which will be added in a subsequent PR, spun off of this one.

This PR is essentially just adding the code that will be used later, as it cannot be enabled (easily :) ) at the moment.

The PR depends on #32089 and needs to have it merged before this one can be merged. We'll also need to update the changelog and package versions afterwards.

### Other information:

## Jetpack product discussion

p9dueE-7e6-p2

No

## Testing instructions:

* Apply PR

### DIFF
--- a/projects/packages/sync/src/class-queue.php
+++ b/projects/packages/sync/src/class-queue.php
@@ -8,6 +8,7 @@
 namespace Automattic\Jetpack\Sync;
 
 use Automattic\Jetpack\Sync\Queue\Queue_Storage_Options;
+use Automattic\Jetpack\Sync\Queue\Queue_Storage_Table;
 use WP_Error;
 
 /**
@@ -41,9 +42,9 @@ class Queue {
 	/**
 	 * Queue Storage instance where we'll store the queue items.
 	 *
-	 * For now it's only the Options table. To be updated to include the Custom table in future updates.
+	 * For now, it's only the Options table. To be updated to include the Custom table in future updates.
 	 *
-	 * @var Queue_Storage_Options|null
+	 * @var Queue_Storage_Options|Queue_Storage_Table|null
 	 */
 	public $queue_storage = null;
 
@@ -57,8 +58,15 @@ class Queue {
 		$this->row_iterator = 0;
 		$this->random_int   = wp_rand( 1, 1000000 );
 
-		// Initialize the storage with the Options table backend. To be changed in subsequent updates to include the logic to switch to Custom Table.
-		$this->queue_storage = new Queue_Storage_Options( $this->id );
+		/**
+		 * If the Custom queue table is enabled - let's use it as a backend. Otherwise, fall back to the Options table.
+		 */
+		if ( Settings::is_custom_queue_table_enabled() ) {
+			$this->queue_storage = new Queue_Storage_Table( $this->id );
+		} else {
+			// Initialize the storage with the Options table backend. To be changed in subsequent updates to include the logic to switch to Custom Table.
+			$this->queue_storage = new Queue_Storage_Options( $this->id );
+		}
 	}
 
 	/**
--- a/projects/packages/sync/src/class-queue.php
+++ b/projects/packages/sync/src/class-queue.php
@@ -8,7 +8,6 @@
 namespace Automattic\Jetpack\Sync;
 
 use Automattic\Jetpack\Sync\Queue\Queue_Storage_Options;
-use Automattic\Jetpack\Sync\Queue\Queue_Storage_Table;
 use WP_Error;
 
 /**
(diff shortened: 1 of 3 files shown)


projects/packages/sync/src/sync-queue/class-queue-storage-table.php
```diff
@@ -0,0 +1,459 @@
...
+		$this->table_name = $wpdb->prefix . $this->table_name_no_prefix;
+	}
+
+	/**
+	 * Creates the new table and updates the options to work with
+	 * the new table if it was created successfully.
+	 *
+	 * @return void
+	 */
+	public function create_table() {
```
fgiannar: Should we make this a private method as it will only be run by the migration script (follow up PR)?

projects/packages/sync/src/sync-queue/class-queue-storage-table.php
```diff
@@ -0,0 +1,459 @@
...
+		 */
+		\dbDelta( $table_definition );
+	}
+
+	/**
+	 * Check if the Custom table actually exists.
+	 *
+	 * @return bool
+	 */
+	public function custom_table_exists() {
```
fgiannar: Should we make this a private method as it will only be run by the migration script (follow up PR)?

projects/packages/sync/src/sync-queue/class-queue-storage-table.php
```diff
@@ -0,0 +1,459 @@
...
+		}
+
+		return true;
+	}
+	/**
+	 * Check if the table is healthy, and we can read and write from/to it.
+	 *
+	 * @return bool If the custom table is available, and we can read and write from/to it.
+	 */
+	public function is_custom_table_healthy() {
```
fgiannar: Should we make this a private method as it will only be run by the migration script (follow up PR)?

projects/packages/sync/src/sync-queue/class-queue-storage-table.php
```diff
@@ -0,0 +1,459 @@
...
+			return false;
+		}
+
+		return true;
+	}
+
+	/**
+	 * Drop the custom table as part of cleanup.
+	 *
+	 * @return bool If the table is cleared. It's using `is_custom_table_healthy` to check.
```
fgiannar:
```suggestion
	 * @return bool If the table is cleared.
```

projects/packages/sync/src/sync-queue/class-queue-storage-table.php
```diff
@@ -0,0 +1,459 @@
...
+		if ( empty( $result ) || count( $result ) !== 1 || $result[0] !== $this->table_name ) {
+			return false;
+		}
+
+		return true;
+	}
+	/**
+	 * Check if the table is healthy, and we can read and write from/to it.
+	 *
+	 * @return bool If the custom table is available, and we can read and write from/to it.
```
fgiannar, line 122: The write part is not covered here, right? This might even be redundant when we introduce the migrations PR

projects/packages/sync/src/class-queue.php
```diff
@@ -57,8 +58,15 @@ public function __construct( $id ) {
 		$this->row_iterator = 0;
 		$this->random_int   = wp_rand( 1, 1000000 );
 
-		// Initialize the storage with the Options table backend. To be changed in subsequent updates to include the logic to switch to Custom Table.
-		$this->queue_storage = new Queue_Storage_Options( $this->id );
+		/**
+		 * If the Custom queue table is enabled - let's use it as a backend. Otherwise, fall back to the Options table.
+		 */
+		if ( Settings::is_custom_queue_table_enabled() ) {
```
fgiannar: I think it's better to introduce this change in the follow up PR that will enable switching to/from options/custom table upon updating the feature flag.

fgiannar (changes requested):
Nice work, not easy to test though therefore I suggest we introduce the instantiation of the `Queue_Storage_Table` in a follow up PR (after we introduce the migrations) to be on the safe side.

Also as I suggested in the Queue refactor PR that is already merged, it would be nice if we introduced an interface, shared by both storage classes, for defining the public methods they both need to implement.

projects/packages/sync/src/sync-queue/class-queue-storage-table.php
bisko: Yes, the write part is not covered in the health check. It will be fully added in the migrations PR.

projects/packages/sync/src/class-queue.php
bisko: Yes, good point. Reverted the changes in `class-queue.php` as a whole, so we can add them back in the migrations PR.

fgiannar (approved):
Let's ship so we can move forward with the migrations PR and we can fully test this in the actual PR that will enable switching from/to options/custom table. Noting that unit tests will be also added in a follow up PR as agreed.

PR #32112, opened 2023-07-27
Social: Prepare for new dismissable notices

gmjuhasz:
This change makes life easier for the followup tasks on the new notices.

## Proposed changes:

* Added a check to see if user on Basic plan
* Cleaned up `use-dismiss-notice` a bit, so the Notice types are inside the hook

### Other information:

## Testing instructions:

* Sanity check the code.
* You could import and `console.log` out the `hasBasicPlan` value, it should only be true if you have the Basic plan.
* Dismissing Instagram notice should work as before


spsiddarthan:
This works well!

spsiddarthan (approved):

PR #32119, opened 2023-07-27
Newsletters: Add Paywall block functionality

lezama:
## Proposed changes:
* On gated newsletters; display the content before jetpack/paywall block

Editor:
Screenshot 2023-07-27 at 14 21 51

Non subscribers:
Screenshot 2023-07-27 at 14 21 21

Subscribers:
Screenshot 2023-07-27 at 14 21 28

### Other information:

## Jetpack product discussion
p2-peKye1-fd
No

## Testing instructions:

* Add define( 'JETPACK_BLOCKS_VARIATION', 'beta' ); to your setup
* Create or Edit a post and add the paywall block
* Test different combinations Newsletter Visibility and different types of users 😱

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -97,8 +97,8 @@ function register_block() {
 		}
 	);
 
-	// Hide the content
-	add_action( 'the_content', __NAMESPACE__ . '\maybe_get_locked_content' );
+	// Hide the content – Priority 8 makes it run before do_blocks gets called for the content
+	add_filter( 'the_content', __NAMESPACE__ . '\add_paywall', 8 );
 
 	// Close comments on the front-end
 	add_filter( 'comments_open', __NAMESPACE__ . '\maybe_close_comments', 10, 2 );
@@ -752,7 +752,8 @@ function jetpack_filter_excerpt_for_newsletter( $excerpt, $post = null ) {
  *
  * @return string
  */
-function maybe_get_locked_content( $the_content ) {
+function add_paywall( $the_content ) {
+	$block_name = 'jetpack/paywall';
 	require_once JETPACK__PLUGIN_DIR . 'modules/memberships/class-jetpack-memberships.php';
 
 	if ( Jetpack_Memberships::user_can_view_post() ) {
@@ -760,7 +761,13 @@ function maybe_get_locked_content( $the_content ) {
 	}
 
 	$post_access_level = Jetpack_Memberships::get_post_access_level();
-	return get_locked_content_placeholder_text( $post_access_level );
+	$paywalled_content = get_paywall_blocks( $post_access_level );
+
+	if ( has_block( $block_name ) ) {
+		$paywalled_content = strstr( $the_content, '<!-- wp:' . $block_name . ' /-->', true ) . $paywalled_content;
+	}
+
+	return $paywalled_content;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


edanzer:
Just tested this and reviewed the code.

1) **Cannot see behind pawall.** For me, I can't seem to get past the paywall. I go to the post, enter my email to subscribe, go and confirm via email, end up on the subscription management page, manually go back to the browser tab with the post, refresh, and still see the paywall. I've tried this in a fresh browser with multiple emails. I had been wondering if something else was needed - ie, creating an account and logging in to WP.com, but sounds like we don't think that should be needed. For now, not sure what else to try on this.

2) **Paywall block on pages.** I made a mistake initially and tried adding this post to a page. The newsletter visibility settings are not available on pages yet. And I think the block - as it stands in this PR - relies on those. If so, we may want to hide the block from pages as it seems to have no effect there. _We can handle this in a separate PR._ 

3) **Paywall block vs subscriptions.** I noticed the code to apply the paywall is inside the subscription block (ie, extensions/blocks/subscription). I'm wondering it should live someplace more general? Ideally, I think we'd want (later if not now) the Paywall block to be based on other things beyond subscription status. For example, I might want to ask a user to buy a plan (any plan, not just a paid subscription) or to pay for this one post. In such a case, access behind the paywall is unrelated to subscription status.

simison:
> Paywall block vs subscriptions. I noticed the code to apply the paywall is inside the subscription block (ie, extensions/blocks/subscription). I'm wondering it should live someplace more general?

Yes, I think we should take a look at code and move things around a bit to make it clearer and easier to work, but outside this PR.

> Cannot see behind pawall. For me, I can't seem to get past the paywall. 

I've tested and things worked fine. Let's test more these flows outside this PR though, as there might be some caching issues unrelated to this PR.

> If so, we may want to hide the block from pages as it seems to have no effect there. We can handle this in a separate PR.

Yeah let's make sure to hide block in other contexts in the inserters:
- Site editor / template editors
- block Widget editor
- Page editor

Post editor is the only place where the block should be available.

simison (approved):
Tests well, tested:

- Paid posts
- Subscribers only posts
- Adding more/next page tags into mix to see it works well
- Tested in web, and reader
- Tested using admins account, non subscribed, subscribed
- Did not test paid subscriber access specifically but it should work the same as non-subscribed/subscribed case and you didn't change anything about that logic.

Unrelated to this PR and good follow ups:

- [Some style polish](https://github.com/Automattic/jetpack/pull/32101#issuecomment-1655878080)
- We should disallow the block as child-block, just like "next page" block:

    Screenshot 2023-07-28 at 18 23 03
    Screenshot 2023-07-28 at 18 22 44

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -752,15 +752,22 @@ function jetpack_filter_excerpt_for_newsletter( $excerpt, $post = null ) {
...
 
 	if ( Jetpack_Memberships::user_can_view_post() ) {
 		return $the_content;
 	}
 
 	$post_access_level = Jetpack_Memberships::get_post_access_level();
-	return get_locked_content_placeholder_text( $post_access_level );
+	$paywalled_content = get_paywall_blocks( $post_access_level );
+
+	if ( has_block( $block_name ) ) {
```
simison, line 766: Checked that `has_block` has been available since WP 5.0 so we're good not to check if it exists. :-) 

https://github.com/WordPress/WordPress/blob/dcc82c39ccfc4d0aef8751b317724fea0c27b40f/wp-includes/blocks.php#L632-L641

PR #30036, opened 2023-04-11
VideoPress: render VideoPress video block 100% dynamically

retrofox:
Related to https://github.com/Automattic/jetpack/issues/29987
Fixes #

## Proposed changes:

* VideoPress: render VideoPress video block 100% dynamically

### Other information:

- [x] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

In terms of functionality, there should be no changes when using the VideoPress video block between the Jetpack stable version and this branch. Thus

* Create a JN site with the Jetpack beta plugin installed
* Start using the Jetpack stable release (the default one)
* Connect the site
* Enable the VideoPress video module
* Go to the block editor and create a VideoPress video block
* Customize the instance (autoplay, control, loop, etc.)
* Set block caption, too
* Save the post
* Check the block in the front-end

* Switch to this branch by using the Jetpack beta plugin

Screen Shot 2023-04-11 at 16 56 49

* Enable `beta` extensions

Screen Shot 2023-04-11 at 16 58 52

* Go back to the block editor. Same post.
* Hard-refresh
* Confirm you are in the proper branch, taking a look at the Preview On Hover toggle

Screen Shot 2023-04-11 at 17 00 19

* It should not be any change (no block validation errors, for instance)
* Test in the front-end
* Confirm the previous settings are preserved (autoplay, control, loop, etc)

* Go to the block editor
* Enable Preview On Hover
* Save the post
* Go to the front-end
* Confirm it works as expected 

Repeat the test but for Simple and Atomic sites

PR #32123, opened 2023-07-27
VideoPress: Remove HTML support

dhasilva:
Fixes #

## Proposed changes:

* The VideoPress v6 block is [dynamic](https://github.com/Automattic/jetpack/pull/30036), so it does not have a `save` function and cannot be edited as HTML in the block editor. This PR just removes that option by explicitly setting `html` as false in the block's `supports` object.

### Other information:

## Testing instructions:

* Add a VideoPress block
* Check that the option to `Edit as HTML` is not available anymore

Before | After
-|-
|  |  |
|


oskosk (approved):
LGTM! Thank you

PR #32125, opened 2023-07-27
Earn/remove/connected account id option

n3f:
Part of https://github.com/Automattic/gold/issues/59. Requires D117364-code to be deployed for WPCOM tests to pass.

### Other information:

## Testing instructions:

* See p81Rsd-1q4-p2 (TL;DR - create a jetpack ninja site using this branch and sandbox with the last WPCOM commit (D117364-code) and verify that it all works...

### DIFF
--- a/projects/packages/sync/src/class-defaults.php
+++ b/projects/packages/sync/src/class-defaults.php
@@ -63,7 +63,6 @@ class Defaults {
 		'image_default_link_type',
 		'infinite_scroll',
 		'infinite_scroll_google_analytics',
-		'jetpack-memberships-connected-account-id',
 		'jetpack-twitter-cards-site-tag',
 		'jetpack_activated',
 		'jetpack_allowed_xsite_search_ids',
--- a/projects/packages/sync/src/class-defaults.php
+++ b/projects/packages/sync/src/class-defaults.php
@@ -63,6 +63,7 @@ class Defaults {
 		'image_default_link_type',
 		'infinite_scroll',
 		'infinite_scroll_google_analytics',
+		'jetpack-memberships-has-connected-account',
 		'jetpack-twitter-cards-site-tag',
 		'jetpack_activated',
 		'jetpack_allowed_xsite_search_ids',
--- a/projects/plugins/jetpack/modules/memberships/class-jetpack-memberships.php
+++ b/projects/plugins/jetpack/modules/memberships/class-jetpack-memberships.php
@@ -33,7 +33,7 @@ class Jetpack_Memberships {
 	 *
 	 * @var string
 	 */
-	public static $connected_account_id_option_name = 'jetpack-memberships-connected-account-id';
+	public static $has_connected_account_option_name = 'jetpack-memberships-has-connected-account';
 
 	/**
 	 * Post meta that will store the level of access for newsletters
@@ -282,7 +282,7 @@ class Jetpack_Memberships {
 	 */
 	public function should_render_button_preview( $block ) {
 		$user_can_edit              = static::user_can_edit();
-		$requires_stripe_connection = ! static::get_connected_account_id();
+		$requires_stripe_connection = ! static::has_connected_account();
 
 		$jetpack_ready = ! self::is_enabled_jetpack_recurring_payments();
 
@@ -428,10 +428,10 @@ class Jetpack_Memberships {
 	/**
 	 * Get the id of the connected payment acount (Stripe etc).
 	 *
-	 * @return int|void
+	 * @return bool
 	 */
-	public static function get_connected_account_id() {
-		return get_option( self::$connected_account_id_option_name );
+	public static function has_connected_account() {
+		return get_option( self::$has_connected_account_option_name, false ) ? true : false;
 	}
 
 	/**
--- a/projects/plugins/jetpack/modules/memberships/class-jetpack-memberships.php
+++ b/projects/plugins/jetpack/modules/memberships/class-jetpack-memberships.php
@@ -29,7 +29,7 @@ class Jetpack_Memberships {
 	 */
 	public static $post_type_plan = 'jp_mem_plan';
 
-    /**
+	/**
 	 * Option that will store currently set up account (Stripe etc) id for memberships.
 	 *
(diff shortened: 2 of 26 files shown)


projects/packages/sync/src/class-defaults.php
```diff
@@ -63,7 +63,6 @@ class Defaults {
 		'image_default_link_type',
 		'infinite_scroll',
 		'infinite_scroll_google_analytics',
-		'jetpack-memberships-connected-account-id',
```
millerf, line 66: This needs to be replaced with `JP_Memerships::$has_connected_account_option_name` value... So it can be synced...

projects/plugins/jetpack/tests/php/sync/test_class.jetpack-sync-options.php
```diff
@@ -214,7 +214,6 @@ public function test_sync_default_options() {
 			'site_segment'                                 => 'pineapple',
 			'site_vertical'                                => 'pineapple',
 			'jetpack_excluded_extensions'                  => 'pineapple',
-			'jetpack-memberships-connected-account-id'     => '340',
```
millerf, line 217: This needs to be replaced with `JP_Memerships::$has_connected_account_option_name` value... So it can be synced...

projects/plugins/jetpack/modules/memberships/class-jetpack-memberships.php
```diff
@@ -431,7 +443,16 @@ public static function get_blog_id() {
 	 * @return bool
 	 */
 	public static function has_connected_account() {
-		return get_option( self::$has_connected_account_option_name, false ) ? true : false;
+
+		// This is the primary solution.
+		$has_option = get_option( self::$has_connected_account_option_name, false ) ? true : false;
```
millerf, line 448: This is the right approach 👍

millerf, line 448:
```suggestion
		$has_option = get_option( self::$has_connected_account_option_name, false );
```

Also... :)

n3f: I prefer ensuring that it's true or false instead of 0, 1, null or whatever could be put in there...  That's probably just my java background though.  I can change if you feel strongly about it, but my vote is to keep it.

millerf: My JS background would write it has:
```
$has_option = !!get_option( self::$has_connected_account_option_name, false );
```
😬

Joke aside, `get_option` will return `false` if not set, and a boolean if it was set... So not sure of your intent there...

n3f: `get_option()` might return a boolean if it's set, but it might not depending on what the data is...

```
 * @return mixed Value of the option. A value of any type may be returned, including
 *               scalar (string, boolean, float, integer), null, array, object.
 *               Scalar and null values will be returned as strings as long as they originate
 *               from a database stored option value. If there is no option in the database,
 *               boolean `false` is returned.
```

millerf: Yes but you set it as a boolean!

n3f: Doesn't stop the site admin from changing it to something else...

millerf: I see... But if the admin changes to something else we can't even use it as it is no more a source of truth 😅

millerf: But I see your point... Let's keep it then

n3f: Thanks!

millerf (dismissed):
LGTM as it is the same than the WPCOM diff: D117364-code

projects/plugins/jetpack/modules/memberships/class-jetpack-memberships.php
```diff
@@ -428,10 +440,19 @@ public static function get_blog_id() {
...
 	 *
-	 * @return int|void
+	 * @return bool
 	 */
-	public static function get_connected_account_id() {
-		return get_option( self::$connected_account_id_option_name );
+	public static function has_connected_account() {
+
+		// This is the primary solution.
+		$has_option = get_option( self::$has_connected_account_option_name, false ) ? true : false;
```
pottedmeat, line 448: Instead of using `false` for the fallback, could you use null and then return as long as `! is_null( $has_option )`?

projects/plugins/jetpack/modules/memberships/class-jetpack-memberships.php
```diff
@@ -428,10 +440,19 @@ public static function get_blog_id() {
...
+		if ( $has_option ) {
+			return true;
+		}
+
+		// This is the fallback solution.
+		// TODO: Remove this once the has_connected_account_option is migrated to all sites.
+		return get_option( 'jetpack-memberships-connected-account-id', false ) ? true : false;
 	}
```
n3f, line 456: @pottedmeat 

```suggestion
	public static function has_connected_account() {

		// This is the primary solution.
		$has_option = get_option( self::$has_connected_account_option_name, null );
		if ( ! is_null( $has_option ) ) {
			return true;
		}

		// This is the fallback solution.
		// TODO: Remove this once the has_connected_account_option is migrated to all sites.
		return get_option( 'jetpack-memberships-connected-account-id', false ) ? true : false;
	}
```

Is this what you were thinking?  I don't like the ambiguity with what the option _could be_.  I think I would rather be more explicit. Also it seems that if the option was set as false this would still return true, but that's probably my implementations fault.

pottedmeat: You'd still need to do: `return $has_option ? true : false;` within that if block.

pottedmeat (dismissed):

samiff (approved):

PR #32127, opened 2023-07-28
 AI Client: remove unused image library

retrofox:
Since this lib is not being used, let's clean it up for now. We can add new libs when it's required.

Fixes #

## Proposed changes:

*  AI Client: remove unused image library

### Other information:

## Testing instructions:

* Just confirm that the building process works as expected.
* Confirm the client still provides the resources, by using a Jetpack site and some AI features.


nunyvega (approved):
Looks good!

PR #32128, opened 2023-07-28
Social: Add upsell notice for the admin page

gmjuhasz:
This change adds the upsell notice for Basic plan users. It can be dismissed for 3 months.
### Other information:

## Jetpack product discussion

1205003617340866-as-1205091696786995

No.
## Testing instructions:

* Go to the Social Admin page.
Check the steps on  #32087


spsiddarthan (changes requested):
It works well, but I noticed that the pricing page checkout is a generic one and not a site-specific one. That's because the pricing page redirect URL is `https://cloud.jetpack.com/pricing#jetpack_social_advanced_yearly` and it doesn't specify the site name in it. We have the same issue with https://github.com/Automattic/jetpack/pull/32087 as well. This means the purchase is effectively a siteless-checkout. We need to make it a site based checkout. 


We need to add redirect links specific to the site, like we show for the basic plan uprade or the ones I added in https://github.com/Automattic/jetpack/pull/32032. We need to add redirect links for these.

spsiddarthan:
Approving after chatting with Geri on Slack, we will fix the redirect links in a follow-up.

spsiddarthan (approved):

PR #32099, opened 2023-07-26
CRM: Automations Quotes triggers

gogdzl:
## Proposed changes:

* This PR adds all quotes triggers to Automation with their respective tests.
  * New quote
  * Quote accepted
  * Quote updated
  * Quote's status updated
  * Quote deleted

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion
https://github.com/Automattic/zero-bs-crm/issues/3244
https://github.com/Automattic/zero-bs-crm/issues/3245
https://github.com/Automattic/zero-bs-crm/issues/3246

## Does this pull request change what data or activity we track or use?
No.

## Testing instructions:

* Run tests by navigating to the CRM directory in your local installation after running `jetpack build plugins/crm`, and then run `composer phpunit -- --testdox --testsuite automation` (note this will currently run all tests as well with a new commit adding tests to this PR).
* You can test manually the objects being passed to the actions within each trigger by being creative (e.g. using the error_log).

PR #32133, opened 2023-07-28
CRM Automations: Bootstrap system

kallehauge:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3263

## Proposed changes:

* Add a new `Bootstrap` class that is solely responsible for:
  * Initialise the automation engine
  * Register "core triggers/conditions/actions" and allow 3. parties to hook into it 
* Remove the need to inject `Automation_Engine` into workflow constructs
  * The intent here is to make it a smoother experience programmatically registering workflows.
  * This also means that we had to add new set/get logger methods to a workflow
* Update existing workflow tests to inject the engine via `Automation_Workflow::set_engine`
* Add a new test that ensures that we throw a fatal error if someone tries to execute a workflow without injecting the engine.
* Add a `is_admin` condition around the automation bootstrap logic so we do not bootstrap all of it on front-end page loads.
  * I've added a `@todo` condition since I believe this could be a general issue for CRM.

### Other information:

## Jetpack product discussion

pbhBOz-2ES-p2

No.

## Testing instructions:

* Add the below snippet somewhere in your code (e.g. `0-sandbox.php).
* _(Optional) Enable error log and tail it to verify that we do not throw any errors._
* Go to `/wp-admin/admin.php?page=zbs-add-edit&action=edit&zbstype=contact and add a new contact
  * The status should be anything else than `Customer` to verify the workflow worked.
* Verify that the status was set to `Customer` by the workflow.

```php
// Enable the automation system.

// Programmatically define a workflow that will always set a contacts status to "Customer" when they are first created.
add_filter(
	'jpcrm_automation_workflows',
	function ( $workflows ) {
			[
				'id'           => 'testing',
				'name'         => __( 'New contact: Set status to "Customer"', 'zero-bs-crm' ),
				'description'  => __( 'This automation will change the status of a contact to "Customer" when they are created.', 'zero-bs-crm' ),
				'category'     => 'contact',
				'triggers'     => [
				],
				'initial_step' => [
					'slug'            => 'contact_status_condition',
					'class_name'      => \Automattic\Jetpack\CRM\Automation\Actions\Update_Contact_Status::class,
					'attributes'      => [
						'new_status' => 'Customer',
					],
					'next_step_true'  => null,
					'next_step_false' => null,
				],
			]
		);

		$workflows[ $workflow->get_id() ] = $workflow;

		return $workflows;
	}
);
```


cleacos:
We will need these changes for our current working code. Before merging this one, it'd like to finish the review of the [Quote triggers PR](https://github.com/Automattic/jetpack/pull/32099), and merge it.

projects/plugins/crm/src/automation/class-automation-boostrap.php
```diff
@@ -0,0 +1,160 @@
...
+		 * @since $$next-version$$
+		 *
+		 * @param string[] $var A list of condition classes.
+		 */
+		$conditions = apply_filters( 'jpcrm_automation_conditions', $conditions );
+
+		foreach ( $conditions as $condition ) {
+			try {
+				$this->engine->register_step( $condition::get_slug(), $condition );
+			} catch ( \Exception $e ) {
```
cleacos, line 125: Future note: The step register will be the same method used for registering triggers, without the need of calling `get_slug`. Only one parameter.

projects/plugins/crm/src/automation/class-automation-workflow.php
```diff
@@ -12,7 +12,7 @@
  */
 class Automation_Workflow {
 
-	/** @var int */
+	/** @var int|string */
```
cleacos, line 15: This should be int... is the row ID in the database.

projects/plugins/crm/src/automation/class-automation-workflow.php
```diff
@@ -42,28 +42,27 @@ class Automation_Workflow {
...
-		$this->automation_engine = $automation_engine;
-		$this->logger            = $automation_engine->get_logger() ?? Automation_Logger::instance();
 	}
 
 	/**
 	 * Get the id of this workflow
 	 *
-	 * @return int
+	 * This will either be a string if the workflow is registered in the codebase,
+	 * or an integer if it is a custom workflow stored in the database.
```
cleacos, line 119: OK. The id was added for the database row. Normally I try to follow the strict type way and avoid mixed types. Do you have a use case where we need the ID for hardcoded workflows? `$name` could be enough? If so, any format for it?

projects/plugins/crm/src/automation/class-automation-workflow.php
```diff
@@ -248,18 +247,51 @@ public function add_trigger( string $string ) {
...
+	 *
+	 * @return Automation_Engine
+	 *
+	 * @throws Workflow_Exception Throws an exception if there is no engine instance.
+	 */
+	protected function get_engine(): Automation_Engine {
+		if ( ! $this->automation_engine instanceof Automation_Engine ) {
+			throw new Workflow_Exception(
+				/* Translators: %s The ID of the workflow. */
+				sprintf( __( '[%s] Cannot run workflow logic without an engine instance', 'zero-bs-crm' ), $this->get_id() ),
```
cleacos: Should we translate the technical log?

cleacos: If the output is for the end user, I'd say yes, absolutely, but if it is for debugging purposes I'd keep it in English.

cleacos:
I've updated some tests that were still using the Automation_Workflow constructor with the Automation_Engine parameter.

projects/plugins/crm/src/automation/class-automation-boostrap.php
kallehauge: 👍  I noticed it as well, but the PR seemed to get out of scope if I corrected it here 😄

cleacos: Yes, sure! It's not for this PR.

cleacos: Issue added to the project board

cleacos:
I've updated the testing code to add in `0-sandbox.php` file:

```
add_filter(
	'jpcrm_automation_workflows',
	function ( $workflows ) {
			[
				'id'           => 'testing',
				'name'         => __( 'New contact: Set status to "Customer"', 'zero-bs-crm' ),
				'description'  => __( 'This automation will change the status of a contact to "Customer" when they are created.', 'zero-bs-crm' ),
				'category'     => 'contact',
				'triggers'     => [
					`jpcrm/contact_created`,
				],
				'initial_step' => [
					'class_name'      => \Automattic\Jetpack\CRM\Automation\Actions\Update_Contact_Status::class,
					'attributes'      => [
						'new_status' => 'Customer',
					],
					'next_step'  => null,
				],
			]
		);

		$workflows[ $workflow->get_id() ] = $workflow;

		return $workflows;
	}
);
```

Currently, it doesn't work because of the hook name change in the Contact_Created trigger, the listen_to_event should listen to `jpcrm_automation_contact_new`, but it's going to be changed by the Event Manager PR. Changing it manually the test works as expected, on Contact creation it is set as Customer.

**Note:** For the steps, we can use the slug name or the class name, both work independently. The class_name has preferent over the slug when the step is executed, it'll try to instance the class name passed, and the fallback is the slug name.

cleacos:
Other than that, the code looks good and works as expected. Let's merge it ASAP to be able to continue making progress with real use cases.

Let's wait for the Event Manager code to be merged in trunk, and then update this PR.

projects/plugins/crm/src/automation/class-automation-workflow.php
kallehauge: Honestly, then it's still a bit TBD depending on how we end up storing attributes for "internal workflows". E.g. if we ever added something like the example below.

```php
add_filter(
	'jpcrm_automation_workflows',
	function ( $workflows ) {
		// Contact Created: Force contact status to be "New".
			[
				'id'           => 'testing',
				'name'         => __( 'New contact: Set status to "Customer"', 'zero-bs-crm' ),
				'description'  => __( 'This automation will change the status of a contact to "Customer" when they are created.', 'zero-bs-crm' ),
				'category'     => 'contact',
				'triggers'     => [
				],
				'initial_step' => [
					'slug'            => 'contact_status_condition',
					'class_name'      => \Automattic\Jetpack\CRM\Automation\Actions\Update_Contact_Status::class,
					'attributes'      => [
						'new_status' => 'Customer',
					],
					'next_step_true'  => null,
					'next_step_false' => null,
				],
			]
		);

		$workflows[ $workflow->get_id() ] = $workflow;

		return $workflows;
	}
);
```

In this case, we'd have to rely on the manually defined ID to identify which workflow we're potentially updating attributes for via our APIs - and we cannot make it a number since it doesn't actually exist in the database.

We could end up storing everything in the database and in that case we'd be able to be strict about

projects/plugins/crm/src/automation/class-automation-workflow.php
kallehauge: Summary from 1:1 conversation: We're sticking with "machine friendly" messages (read: we won't translate them).
The UI that handles automations should provide guidance instead.

projects/plugins/crm/src/automation/class-automation-workflow.php
cleacos: Fair enough. It's fine for me

projects/plugins/crm/src/automation/class-automation-workflow.php
cleacos: We could create another task for removing the translation from any technical log output.

projects/plugins/crm/src/automation/class-automation-boostrap.php
```diff
@@ -0,0 +1,185 @@
...
+ *
+ * @package Automattic\Jetpack\CRM
+ */
+
+namespace Automattic\Jetpack\CRM\Automation;
+
+/**
+ * Bootstrap the Jetpack CRM Automation engine
+ *
+ * @package Automattic\Jetpack\CRM\Automation
```
cleacos: It seems that `@package` should be used in the file header.

gogdzl (approved):
It works!

Just as a note, the PR instructions do not work out of the box. It seems that the Contact_Created trigger is listening to the wrong event, I had to change from `jpcrm_automation_contact_created` to `jpcrm_automation_contact_new` in `src/automation/commons/triggers/contacts/class-contact-created.php`.

I don't know if we have any task to change event names, but I thought it is worth mentioning here.

PR #32135, opened 2023-07-28
Jetpack Sync: Add Custom table initialization and migration logic

bisko:
This is a spin-off of #31318 and continuation of #32089 and #32111

## Proposed changes:

This PR adds the Custom table initialization and migration logic.

The PR depends on #32089 and #32111 and needs to have both merged before this one can be merged. We'll also need to update the changelog and package versions afterwards.

Additional details will be added later.

### Other information:

## Jetpack product discussion

p9dueE-7e6-p2

No

## Testing instructions:

* Apply PR
* More tests need to be added to support the switching logic. 
* Manual testing instructions from [PHPUnit tests](https://github.com/Automattic/jetpack/blob/ec40e3e62a94e2b5036adbe4b621930b4d94a572/projects/plugins/jetpack/tests/php/sync/test_class.jetpack-sync-queue-dedicated-table.php#L334) from the Spike PR: #31318


projects/packages/sync/src/sync-queue/class-queue-storage-table.php
```diff
@@ -456,4 +476,190 @@ public function delete_items_by_ids( $ids ) {
...
+		);
+	}
+
+	/**
+	 * Migrates the existing Sync events from the Custom table to the Options table
+	 *
+	 * @return void
+	 */
+	public static function migrate_from_custom_table_to_options_table() {
+		global $wpdb;
```
fgiannar, line 605: Do you think we should delete any leftover options first before populating them from the custom table?

fgiannar, line 606: Do you think we should delete any leftover options first before populating them from the custom table?

projects/packages/sync/src/sync-queue/class-queue-storage-table.php
```diff
@@ -456,4 +476,190 @@ public function delete_items_by_ids( $ids ) {
...
+		return true;
+	}
+
+	/**
+	 * Migrates the existing Sync events from the options table to the Custom table
+	 *
+	 * @return void
+	 */
+	public static function migrate_from_options_table_to_custom_table() {
+		global $wpdb;
```
fgiannar, line 510: Do you think we should delete any leftover rows from the custom table first before populating it from the options?

projects/packages/sync/src/sync-queue/class-queue-storage-table.php
```diff
@@ -456,4 +476,190 @@ public function delete_items_by_ids( $ids ) {
 			)
 		);
 	}
+
+	/**
+	 * Table initialization
+	 */
+	public static function initialize_custom_sync_table() {
```
fgiannar, line 484: Would it make sense to rename this to `enable_custom_sync_table` and include the migration logic here? Because `migrate_from_options_table_to_custom_table` is currently public but it should always run after ensuring that the table exists and is healthy.

projects/packages/sync/src/sync-queue/class-queue-storage-table.php
```diff
@@ -456,4 +476,190 @@ public function delete_items_by_ids( $ids ) {
...
+		 * Initialize an instance of the class with a test name, so we can use table prefix and then test if the table is healthy.
+		 */
+		$custom_table_instance = new Queue_Storage_Table( 'test_queue' );
+
+		// Check if the table exists
+		if ( ! $custom_table_instance->custom_table_exists() ) {
+			$custom_table_instance->create_table();
+		}
+
+		if ( ! $custom_table_instance->is_custom_table_healthy() ) {
```
fgiannar: Maybe we should use `immediate-send` here to send the error to WPCOM?

fgiannar (changes requested):
Added a few comments from code review. I'm still a bit confused about how to manually test this though.

As I won't be able to fully test today and I don't want to block you hopefully @chrisbliss18 can help here if time allows today. Otherwise, I'll test first thing tomorrow morning but pls update the manual testing instructions to be more clear in the meantime. Many thanks!

projects/packages/sync/src/sync-queue/class-queue-storage-table.php
bisko: It is public so it can be ran separately from external source, as it's used in #32275

projects/packages/sync/src/sync-queue/class-queue-storage-table.php
bisko: Yes, all errors should be sent with `immediate-send` to WPCOM and the logic needs to cover all broken cases. Not added here yet, as those are TBD. 

Not resolving the conversation as it's still a TODO item.

chrisbliss18: `is_custom_table_healthy()` can now return `WP_Error` in addition to bool. This conditional isn't handling the `WP_Error`.

chrisbliss18 (dismissed):
I was able to successfully execute the testing instructions excluding the final instruction. I discussed the final instruction with you directly, and the linked unit tests are too outdated to properly test the current code. So I skipped those tests.

I added a comment about properly handling a returned `WP_Error` from the call to `is_custom_table_healthy()`. If that is fixed, I don't see any other issues.

projects/packages/sync/src/sync-queue/class-queue-storage-table.php
bisko: I am leaving that logic for the follow-up PR, as we discussed earlier today - to do a two-step migration to avoid any leftover data from parallel requests being stuck in the table.

If we clean up the table in the migration code, we'll delete valid data and the clean up should happen outside of this method.

projects/packages/sync/src/sync-queue/class-queue-storage-table.php
bisko: See answer for `migrate_from_options_table_to_custom_table` below.

projects/packages/sync/src/sync-queue/class-queue-storage-table.php
bisko: Thanks for catching that. Updated the logic.

bisko: Added a TODO in the code to handle in the subsequent PR when we send errors to WPCOM.

fgiannar (approved):
LGTM!

PR #32137, opened 2023-07-28
Paywall: limit to top level on posts, add icon and keywords

simison:
Based on findings in https://github.com/Automattic/jetpack/pull/32119#pullrequestreview-1552370730 

- Limit to top-level only, no inner blocks to avoid issues with visual rendering. We might relax this later but needs more attention before doing so to avoid issues like these:

    

- Use icon from "next page" block — we should invent our own but this is a pretty nice temporary one!

    image

- Add first keywords

## Proposed changes:

*

### Other information:

## Testing instructions:

- Add `define( 'JETPACK_BLOCKS_VARIATION', 'beta' );` to your setup
- Create or Edit a post and add the paywall block
- Check icon
- Test adding as inner block (not allowed), and as top level block (allows)
- Test adding in post editor, page editor

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/paywall/index.js
+++ b/projects/plugins/jetpack/extensions/blocks/paywall/index.js
@@ -1,8 +1,8 @@
 import { __ } from '@wordpress/i18n';
+import { pageBreak as icon } from '@wordpress/icons';
 import { getIconColor } from '../../shared/block-icons';
 import attributes from './attributes';
 import edit from './edit';
-import icon from './icon';
 
 /**
  * Style dependencies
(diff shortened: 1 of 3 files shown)


projects/plugins/jetpack/extensions/blocks/paywall/index.js
```diff
@@ -19,19 +19,27 @@ export const settings = {
...
+		_x( 'subscribe', 'block search term', 'jetpack' ),
+		_x( 'subscription', 'block search term', 'jetpack' ),
+		_x( 'subscriptions', 'block search term', 'jetpack' ),
+	],
 	supports: {
 		customClassName: false,
 		html: false,
 		multiple: false,
 	},
+	parent: [ 'core/post-content' ],
```
simison, line 38: Took from https://github.com/WordPress/gutenberg/blob/4b4914b7e4ccdeb6393e95eee4666c5e62a92af3/packages/block-library/src/nextpage/block.json#L9

lezama:
> Use icon from "next page" block — we should invent our own but this is a pretty nice temporary one!

nice!

projects/plugins/jetpack/extensions/blocks/paywall/index.js
lezama: perfect

lezama (approved):

projects/plugins/jetpack/extensions/blocks/paywall/index.js
simison: Note that this just limits to root of the document, but we still need to restrict to just post-editor and exclude custom post types.

This hook seems good next step: https://developer.wordpress.org/reference/hooks/allowed_block_types_all/

PR #32139, opened 2023-07-28
CRM: Event Manager System - Contact manager

cleacos:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3128

## Proposed changes:

This PR adds the Event Manager system, it's in charge of handling JPCRM events and notifying them via the WP hook system.

This is the first part of the Event Manager system, adding the capabilities for managing contact events. In other PR we will add the rest of the managers (they were removed from this to simplify the PR)

### Other information:

## Testing instructions:
- Check the code
- In `projects/plugins/crm` run `composer dumpautoload`
- Run the unit tests: `composer phpunit -- --testdox --testsuite event-manager``
- Also, you can test it with JPCRM:

Add some actions to your sandbox file: `wp-content/mu-plugins/0-sandbox.php`

```
add_action( 'jpcrm_contact_updated', function( $contact, $old_contact ) {
    error_log( "\nContact Updated!! [{$contact['id']}]", 3, __DIR__ . '/automation.log' );
}, 10, 2 );

add_action( 'jpcrm_contact_created', function( $contact ) {
    error_log( "\nContact Created!! [{$contact['id']}] ", 3, __DIR__ . '/automation.log' );
} );

add_action( 'jpcrm_contact_status_updated', function( $contact, $old_value ) {
    error_log( "\nContact Status updated!! [{$contact['id']}]. Status: $old_value => {$contact['status']}", 3, __DIR__ . '/automation.log' );
}, 10, 2 );

add_action( 'jpcrm_contact_deleted', function( $contact_id ) {
    error_log( "\nContact Deleted!! [$contact_id]", 3, __DIR__ . '/automation.log' );
```
- Now go to JPCRM dashboard, add a contact, update it, and remove it. You should see the log output in a file called `automation.log` in the `mu-plugin` folder.


projects/plugins/crm/src/event-manager/class-events-manager.php
```diff
@@ -0,0 +1,73 @@
...
+namespace Automattic\Jetpack\CRM\Event_Manager;
+
+/**
+ * Events_Manager class.
+ *
+ * @since $$next-version$$
+ */
+class Events_Manager {
+
+	/** @var null The Events_Manager instance. */
```
gogdzl: Same comment as here: https://github.com/Automattic/jetpack/pull/32185#discussion_r1281131513
I think we should make everything WP coding standards compliant.
(There are more places to fix in this PR, but I'm highlighting just this one as an example)

projects/plugins/crm/src/event-manager/class-events-manager.php
```diff
@@ -0,0 +1,73 @@
...
+		return self::$instance;
+	}
+
+	/**
+	 * Set the Events_Manager instance for testing purposes.
+	 *
+	 * @param Events_Manager $instance The Events_Manager instance.
+	 * @return void
+	 */
+	public static function set_instance( Events_Manager $instance ) {
```
gogdzl: This class seems to follow the singleton design pattern, why would we have a public function to change the instance?

projects/plugins/crm/tests/php/event-manager/class-event-manager-faker.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+namespace Automattic\Jetpack\CRM\Event_Manager\Tests;
+
+class Event_Manager_Faker {
```
gogdzl, line 5: Why don't we use just one faker for all tests?
This one has some duplication when compared to Automation_Faker.

projects/plugins/crm/src/event-manager/class-events-manager.php
cleacos: Mainly because of testing purposes (typically, a singleton is not ideal for testing). Testing other parts of the system, we could need to inject a mock object to check something there. 

But yes, to avoid confusion we could remove it and implement the `set` function if needed... so I'll remove it.

projects/plugins/crm/tests/php/event-manager/class-event-manager-faker.php
cleacos: It doesn't have the same expected data structure.

projects/plugins/crm/src/event-manager/class-events-manager.php
cleacos: Okay

projects/plugins/crm/src/event-manager/class-events-manager.php
```diff
@@ -0,0 +1,73 @@
...
+	public function set_active( bool $active ) {
+		$this->active = $active;
+	}
+
+	/**
+	 * Get whether or not the event manager is active.
+	 *
+	 * @return bool
+	 */
+	public function is_active(): bool {
```
gogdzl: I don't see this being used anywhere in the system. IMHO, we should remove this and address in a separate PR, or use this flag where it is needed and add tests.

gogdzl (changes requested):
I left some comments inline. Overall, I like the event manager, and I like the modularity it will bring to the CRM. Well done!

projects/plugins/crm/src/event-manager/class-events-manager.php
cleacos: True. I started to add it after talking about the possibility of switching off the notifications and in consequence the automation (i.e. on migrations). But I didn't continue with it. And yes, better if we address it in a separate PR. Thanks.

projects/plugins/crm/src/event-manager/class-events-manager.php
cleacos: Fixed here: https://github.com/Automattic/jetpack/pull/32139/commits/9cf34431dd081cc89a1bda3b4cd991b1760a5c23

projects/plugins/crm/src/event-manager/class-events-manager.php
cleacos: Fixed here: https://github.com/Automattic/jetpack/pull/32139/commits/b9aa4866c2006b72e52da6e2f4eafee33b697709

projects/plugins/crm/tests/php/event-manager/class-event-manager-faker.php
cleacos: I've removed unused fake data copied from the automation tests. Here: https://github.com/Automattic/jetpack/pull/32139/commits/5eedc6f2badda6a902fe63e2241b182e2a39fba9

cleacos:
I've fixed more PHPDocs issues to meet the WPCS. It's ready for review again @gogdzl

gogdzl (approved):
LGTM!

PR #32144, opened 2023-07-31
Image Guide: Update tiny image check to treat all images as small if there's an error

### LINKED ISSUE #32037: Boost: Image Guide (free) showing up on adminbar avatar
### Impacted plugin

Boost

### Quick summary

CleanShot 2023-07-24 at 18 08 22@2x

### Steps to reproduce

1. Setup latest (2.0.0) Boost;
2. Enable Image Guide;
3. Check any page in the front-end.

Bubble shouldn't show up for avatars.

### What actually happened

_No response_

### Impact

Some (< 50%)

### Available workarounds?

_No response_

### Logs or notes

_No response_

dilirity:
Probably fixes https://github.com/Automattic/jetpack/issues/32037

## Proposed changes:

* Update "is tiny image" check, to treat images who's dimensions couldn't be fetched as tiny.

### Other information:

## Jetpack product discussion

n/a

no

## Testing instructions:

I didn't find a consistent way of testing this. I threw an Error inside the [try block](https://github.com/Automattic/jetpack/blob/6855c6f46cfe9ff409a0c3ec54270c8a964666ed/projects/js-packages/image-guide/src/MeasurableImage.ts#L132-L138) to simulate a "unable to fetch dimensions" case.

### DIFF
--- a/projects/js-packages/image-guide/src/MeasurableImage.ts
+++ b/projects/js-packages/image-guide/src/MeasurableImage.ts
@@ -123,7 +123,7 @@ export class MeasurableImage {
 	}
 
 	/**
-	 * Checks if the image is too small and should be ignored. Will return false on images
+	 * Checks if the image is too small and should be ignored. Will return true on images
 	 * that don't load at all - we can't establish they're tiny!
 	 *
 	 * @returns {boolean} - if the image is smaller than 65 pixels width and height return true
--- a/projects/js-packages/image-guide/src/MeasurableImage.ts
+++ b/projects/js-packages/image-guide/src/MeasurableImage.ts
@@ -134,7 +134,7 @@ export class MeasurableImage {
 			const dimensions = await this.fetchFileDimensions( this.getURL() );
 			return dimensions.width < minSize || dimensions.height < minSize;
 		} catch ( err ) {
-			return false;
+			return true;
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


projects/js-packages/image-guide/src/MeasurableImage.ts
```diff
@@ -134,7 +134,7 @@ export class MeasurableImage {
 			const dimensions = await this.fetchFileDimensions( this.getURL() );
 			return dimensions.width < minSize || dimensions.height < minSize;
 		} catch ( err ) {
-			return false;
+			return true;
```
thingalon, line 137: This function has a comment on it that says what it returns when an image fails to load. If we're changing the behaviour we should update the comment too.

thingalon (changes requested):
Comment needs updating.

However I think this fix is fine. `fetchFileDimensions` uses an `img` tag to load the image and report on its dimensions (width, height). If it fails to load in an `img` tag, we likely shouldn't be showing it as an issue in the image guide.

projects/js-packages/image-guide/src/MeasurableImage.ts
dilirity: Done in 99ca2a5.

thingalon (approved):
🚢  :)

PR #32146, opened 2023-07-31
WP.com settings menu: add newsletter & podcasting sections

simison:
Goes together with 
- Newsletters https://github.com/Automattic/wp-calypso/pull/79986
- Podcasting https://github.com/Automattic/wp-calypso/pull/79996

## Proposed changes:
* Add a new submenus under "Settings" for WP.com for newsletters and podcasting.
* No changes to Jetpack sites.

### Other information:

## Testing instructions:
- Apply the change at WP.com sandbox as [per instructions below](https://github.com/Automattic/jetpack/pull/32146#issuecomment-1658220786)
- Enable filters in your sandbox:
   ```php
   add_filter( 'calypso_use_newsletter_settings', '__return_true' );
   ```
   ```php
   add_filter( 'calypso_use_podcasting_settings', '__return_true' );
   ```
- See under "Settings" in Calypso

Screenshot 2023-07-31 at 15 21 07


projects/plugins/jetpack/modules/masterbar/admin-menu/class-admin-menu.php
```diff
@@ -396,6 +396,34 @@ public function add_options_menu() {
...
+		if (
+			/**
+			 * Filter to enable the Newsletter Settings section in Calypso UI.
+			 *
+			 * @since $$next-version$$
+			 * @module masterbar
+			 *
+			 * @param bool false Enable newsletter setting section? Default to false.
+			 */
+			apply_filters( 'calypso_use_newsletter_settings', '__return_false' )
```
jeherve: Are the filters in there to offer more control as to when you'll enable those menu items?

jeherve (commented):
This looks good to me, but I'm not sure I understand the need for the filters. If we only mean to enable those on WordPress.com simple sites at first, maybe we can place the menu items in the `WPcom_Admin_Menu` class instead?

projects/plugins/jetpack/modules/masterbar/admin-menu/class-admin-menu.php
simison: Correct, since the UIs in calypso are in different PRs (links in the description) and deploying Jetpack PRs on .com isn't super straightforward/fast, I wanted to have easier time enabling things freely on .com when I wanted.

simison:
> WordPress.com simple sites at first, maybe we can place the menu items in the WPcom_Admin_Menu class instead?

I do want these navigation items appear at both Simple and Atomic sites simultaneously.

jeherve (dismissed):
While this works on WordPress.com simple sites, the menus end up activated by default on WoA sites. That's probably something we'll want to fix.

simison:
>I'm not sure I understand the need for the filters. 

To elaborate on filters — this way I'll have full control on timing when to make these visible in Calypso. I can deploy to .com when I need, but Jetpack release is a bit slower process.

I'll try dig what's going on with WoA 👍

projects/plugins/jetpack/modules/masterbar/admin-menu/class-admin-menu.php
mashikag: @simison , I think you may have mixed up the second parameter of `apply_filters` with the second argument of the `add_filter` call. The second argument of `apply_filters` is meant to be a value that is passed to the filter for processing, in case where the filter does not exist the value is returned back. In this case the '__return_false' is returned which passes the condition.
Using boolean `false` instead should fix your issue:
```php
apply_filters( 'calypso_use_newsletter_settings', false )
```

mashikag (dismissed):

jeherve (approved):
This looks good to me now. 👍

PR #32154, opened 2023-07-31
Introduce VideoPress content gating

pgk:
Fixes https://github.com/Automattic/greenhouse/issues/1808

## Proposed changes:

* Alters the behavior of private videos when embedded within blocks/posts that restrict content to subscribers.
* ~Introduces a way for blocks to control the unauthorized message the video player will display.~
* The video iframe and it's files will not be available if a logged in user has no paid subscription, even if logged in.

### Other information:

## Jetpack product discussion

pxWta-1VK-p2

no

## Testing instructions:

* You need the Jetpack plugin to test this, as all Earn blocks are Jetpack-plugin-only.
* Do a full rebuild: `pnpm jetpack build plugins/jetpack --deps`
* Refer to pxWta-1VK-p2 on what to sandbox and what not.
* You will need to setup paid subscriptions and paid newsletter subscriptions.
* You will need 1 simple user, 1 newsletter sub user and 1 paid sub user.
* You will need to upgrade to a plan that includes more than one VideoPress videos. (because you are sandboxing store, you might need to do that again)
* On your test site, ensure you can access the paid content block.
* If you haven't, connect to stripe. If you sandboxed store, you can choose to use a "test" connection when on stripe's site.
* You are connected to test stripe now.

### Paid content block

* Go ahead and create a new post and add a "Paid Content" block.
* Assign a subscription plan.
* On the Subscriber view, create a VideoPress block and upload a video. Set privacy to private.
* Save.
* View as admin/editor. You can see the video.
* Try to view as a non-paid subscriber. You should not be able to view.
* View as a paid subscriber. You can view content and video.

### Newsletter

* Enable Newsletters (Jetpack -> Settings -> Newsletters). Create a newsletter subscription.
* Create a new post and set access to paid subscribers only.
* Add a videopress video, set to private. Save and publish.
* View as non-subscriber. You cant access anything.
* View as subscriber. You can see the video.

### Paywall

* Enable Newsletters (Jetpack -> Settings -> Newsletters). Create a newsletter subscription.
* Create post and upload a Video. Set it to public.
* Add a paywall block and upload another video. Set it to private.
* Publish post.
* View as guest and subscriber. You only see the video if a subscriber.


projects/packages/videopress/src/class-ajax.php
```diff
@@ -115,15 +116,144 @@ public function wp_ajax_videopress_get_playback_jwt() {
...
+		global $post;
+		$post_to_check = get_post( $embedded_post_id );
+
+		if ( empty( $post_to_check ) ) {
+			$restriction_details = $this->default_video_restriction_details( false );
+			return $this->filter_video_restriction_details( $restriction_details );
+		}
+
+		// phpcs:disable WordPress.WP.GlobalVariablesOverride.Prohibited
+		$post = $post_to_check;
```
thedebian: Should it somehow reset the postdata afterwards ?
Looking up the documentation, it seems to be something along this line :
`$post = setup_postdata( $post_to_check );`
And later on
`wp_reset_postdata();`

projects/packages/videopress/src/class-initializer.php
```diff
@@ -309,13 +312,23 @@ public static function render_videopress_video_block( $block_attributes, $conten
 			);
 		}
 
+		// Get premium content from block context
+		$premium_block_plan_id    = isset( $block->context['premium-content/planId'] ) ? intval( $block->context['premium-content/planId'] ) : 0;
+		$is_premium_content_child = isset( $block->context['isPremiumContentChild'] ) ? (bool) $block->context['isPremiumContentChild'] : false;
+		$maybe_premium_script     = '';
+		if ( $is_premium_content_child ) {
+			$script_content       = "if ( ! window.__guidsToPlanIds ) { window.__guidsToPlanIds = {}; }; window.__guidsToPlanIds['$guid'] = $premium_block_plan_id;";
```
thedebian: $guid should probably be esc_js-ed (maybe $premium_block_plan_id also, even though it's casted now, we never know what it'll be in the future :D )

pgk: Good idea! done

projects/packages/videopress/src/class-ajax.php
pgk: Since this is a wp_ajax call, I think it is OK as the request is specifically about the specific post_id. On the other hand it doesn't hurt to cleanup.

roundhill:
> View as a paid subscriber. You can view content and video.

This is not working for me. I see a token in the query string, but then the console has an error `'Token is not achievable'`

Screenshot 2023-09-13 at 5 56 07 PM

What am I missing? :D

projects/packages/videopress/src/class-access-control.php
```diff
@@ -0,0 +1,367 @@
...
+		return isset( $this->guids_to_subscriptions[ $guid ] ) ? $this->guids_to_subscriptions[ $guid ] : false;
+	}
+
+	/**
+	 * Get the singleton instance.
+	 *
+	 * @return AJAX
+	 */
+	public static function instance() {
+		if ( self::$instance === null ) {
```
roundhill: Yoda-flip this?

projects/packages/videopress/src/class-access-control.php
```diff
@@ -0,0 +1,367 @@
...
+		}
+
+		$default_auth        = $this->get_default_user_capability_for_post( $post_to_check );
+		$restriction_details = $this->default_video_restriction_details( $default_auth );
+
+		if ( $this->jetpack_memberships_available() ) {
+			$memberships_can_view_post         = \Jetpack_Memberships::user_can_view_post( $embedded_post_id );
+			$restriction_details               = $this->get_subscriber_only_restriction_details( $default_auth );
+			$restriction_details['can_access'] = $memberships_can_view_post;
+			return $this->check_block_level_access(
```
roundhill: This `return` could be removed since the same call is made outside the if statement with the same vars.

projects/packages/videopress/src/class-access-control.php
```diff
@@ -0,0 +1,367 @@
...
+	private function check_block_level_access( $restriction_details, $guid, $embedded_post_id, $selected_plan_id ) {
+		if ( $this->jetpack_subscriptions_available() && $selected_plan_id > 0 ) {
+			$restriction_details = $this->get_subscriber_only_restriction_details( $restriction_details['can_access'] );
+			$paywall             = \Automattic\Jetpack\Extensions\Premium_Content\subscription_service();
+
+			// Only paid subscribers should be granted access to the premium content.
+			$access_level                      = \Automattic\Jetpack\Extensions\Premium_Content\Subscription_Service\Token_Subscription_Service::POST_ACCESS_LEVEL_PAID_SUBSCRIBERS;
+			$can_view                          = $paywall->visitor_can_view_content( array( $selected_plan_id ), $access_level );
+			$restriction_details['can_access'] = $can_view;
+			return $this->filter_video_restriction_details(
```
roundhill: Same duplicate return statements here.

projects/packages/videopress/src/class-access-control.php
```diff
@@ -0,0 +1,367 @@
...
+	 */
+	private function check_block_level_access( $restriction_details, $guid, $embedded_post_id, $selected_plan_id ) {
+		if ( $this->jetpack_subscriptions_available() && $selected_plan_id > 0 ) {
+			$restriction_details = $this->get_subscriber_only_restriction_details( $restriction_details['can_access'] );
+			$paywall             = \Automattic\Jetpack\Extensions\Premium_Content\subscription_service();
+
+			// Only paid subscribers should be granted access to the premium content.
+			$access_level                      = \Automattic\Jetpack\Extensions\Premium_Content\Subscription_Service\Token_Subscription_Service::POST_ACCESS_LEVEL_PAID_SUBSCRIBERS;
+			$can_view                          = $paywall->visitor_can_view_content( array( $selected_plan_id ), $access_level );
+			$restriction_details['can_access'] = $can_view;
```
roundhill: Would this be a good place to ensure the video is marked as private, since we know we are dealing with subscriber content here? We could use the `update_video_restriction_details_on_wpcom` function?

roundhill (changes requested):
Overall working well, found a few things!

roundhill:
Dropping notes on how to test this on wp.com:

* Apply D122060-code to your sandbox.
* Run `bin/jetpack-downloader test jetpack add/videopress-content-gating`
* Sandbox your dev site and `public-api`, `subscribe`, `video` subdomains
* Apply additional configurations detailed here to your sandbox: pxWta-1VK#comment-6259-p2

projects/packages/videopress/src/class-access-control.php
pgk: @roundhill do you mean force-setting the privacy setting to private?

roundhill: Yeah, I was thinking if we make it to this check maybe we could force the video to be set to private so that the JWTs  get requested?

roundhill (dismissed):
Latest changes LGTM!

simison:
Needs a rebase.

projects/packages/videopress/src/class-access-control.php
```diff
@@ -0,0 +1,355 @@
...
+	private function jetpack_memberships_available() {
+		return class_exists( '\Jetpack_Memberships' );
+	}
+
+	/**
+	 * Determines if Jetpack Subscriptions are available.
+	 *
+	 * @return bool
+	 */
+	private function jetpack_subscriptions_available() {
```
jeherve, line 86: This function relies on checking for existing functions. I wonder if it should rely on existing utilities managing editor extension availability, as they're defined in `Jetpack_Gutenberg`, such as `is_available()`?

Should it also take module availability into account, and check if the Subscriptions module is active for example?

projects/packages/videopress/src/class-access-control.php
```diff
@@ -0,0 +1,355 @@
...
+	 *
+	 * @param array  $video_restriction_details The restriction details.
+	 * @param string $guid the video guid.
+	 * @param int    $embedded_post_id the post id.
+	 * @param int    $selected_plan_id the selected plan id if applicable.
+	 *
+	 * @return array
+	 */
+	private function filter_video_restriction_details( $video_restriction_details, $guid, $embedded_post_id, $selected_plan_id ) {
+		return (array) apply_filters( 'videopress_video_restriction_details', $video_restriction_details, $guid, $embedded_post_id, $selected_plan_id );
```
jeherve, line 227: I would recommend adding docblocks to the different filters you add in here, so our future selves know how they can be used, and so the [Jetpack Codex](https://developer.jetpack.com/) remains up to date.

projects/packages/videopress/src/class-access-control.php
```diff
@@ -0,0 +1,355 @@
...
+		if ( function_exists( '\Automattic\Jetpack\Extensions\Premium_Content\subscription_service' ) ) {
+			return true;
+		}
+
+		if ( ! defined( 'JETPACK__PLUGIN_DIR' ) ) {
+			return false;
+		}
+
+		$subscription_service_file_path = JETPACK__PLUGIN_DIR . 'extensions/blocks/premium-content/_inc/subscription-service/include.php';
+		if ( ! file_exists( $subscription_service_file_path ) ) {
```
jeherve, line 101: Are there situations where it would not exist? I'm not sure I understand what this checks?

projects/packages/videopress/src/class-access-control.php
```diff
@@ -0,0 +1,355 @@
...
+	 *
+	 * @return bool
+	 **/
+	private function get_default_user_capability_for_post( $post_to_check ) {
+		global $post;
+		$old_post = $post;
+
+		// phpcs:disable WordPress.WP.GlobalVariablesOverride.Prohibited
+		$post         = $post_to_check;
+		$default_auth = current_user_can( 'read' );
```
jeherve: Would it be an option to use `read_post` here instead of relying on editing the `$post` global?

https://github.com/WordPress/wordpress-develop/blob/a55dcf46b8d3ce93affed8fbeb9c6a2645157f3f/src/wp-includes/capabilities.php#L278

jeherve (commented):
Just a few questions, but I'll let the folks on team Zap look at this, they know more about Jetpack Memberships than I do!

simison:
> but I'll let the folks on team Zap look at this, they know more about Jetpack Memberships than I do!

That would be @Automattic/gold for memberships and I believe the teams collaborated on this.

projects/packages/videopress/src/class-access-control.php
```diff
@@ -0,0 +1,355 @@
...
+		}
+
+		$subscription_service_file_path = JETPACK__PLUGIN_DIR . 'extensions/blocks/premium-content/_inc/subscription-service/include.php';
+		if ( ! file_exists( $subscription_service_file_path ) ) {
+			return false;
+		}
+
+		require_once $subscription_service_file_path;
+
+		return function_exists( '\Automattic\Jetpack\Extensions\Premium_Content\subscription_service' );
```
millerf, line 107: That would be always false as https://github.com/Automattic/jetpack/pull/32154/files/16f380bd348c72695508a82d8501c9ee82a49959#diff-dc09df9d5422f6d9a1e4bdab0d4b0747bcd6bd5f5d4a561e6c07111e50afcc31R87 already checks that

projects/packages/videopress/src/class-access-control.php
millerf: It looks like they are checking if premium content block is loaded?

pgk: @jeherve @millerf yes, to my understanding there are situations it would not exist. Since all gating/subscription classes are part of only the Jetpack plugin, any other flavor (specifically Jetpack VideoPress plugin) would not contain those.

projects/packages/videopress/src/class-access-control.php
pgk: Works with read_post! Thanks

projects/packages/videopress/src/class-access-control.php
pgk: My reasons for checks I think I covered in the other https://github.com/Automattic/jetpack/pull/32154/files/16f380bd348c72695508a82d8501c9ee82a49959#r1354597036. 

> Should it also take module availability into account, and check if the Subscriptions module is active for example?

Maybe, let me check and report back

projects/packages/videopress/src/class-access-control.php
pgk: @millerf I do a require_once in the meantime for the file containing that function, so that check might return true.

projects/packages/videopress/src/class-access-control.php
pgk: after some thought, we are going to address this separately.

pgk:
@roundhill @jeherve @millerf ready for a final rereview

projects/packages/videopress/src/class-access-control.php
```diff
@@ -0,0 +1,399 @@
+<?php // phpcs:ignore WordPress.Files.FileName.InvalidClassFileName
```
jeherve: Nit-pick, but you can actually remove this phpcs exception at this point, your file name is valid. 👍

```suggestion
<?php
```

jeherve (commented):
This tests well for me on WordPress.com Simple, and looks good. I'll let others chime in with their tests. 👍

projects/packages/videopress/src/client/lib/token-bridge/index.ts
```diff
@@ -24,11 +24,43 @@ const { videopressAjax } = window;
...
+}
+
+/**
+ * Check if the guid has an associated subscriptionPlanId.
+ *
+ * @param {VideoGUID} guid - The guid.
+ * @returns {Promise} promise.
+ */
+function getSubscriberPlanIdIfExists( guid: VideoGUID ): Promise< number > {
+	return new Promise( function ( resolve ) {
```
millerf, line 53: TIL PHP has Promises ❤️

millerf: I think it would make things way clearer if ready() was returning a premise instead of a function. You're mixing callbacks and promises and it never gets good. What about this approach:

```

       
async function getSubscriberPlanIdIfExists( guid: VideoGUID ) : Promise<number> {  
     
     // Can also be defined outside
     const ready = new Promise( ( resolve ) => document.readyState != "loading" ? 
       resolve() : 
       document.addEventListener("DOMContentLoaded", resolve ) );
    
     await ready();
     if ( ! window.__guidsToPlanIds ) {
       return 0;
     }
     return window.__guidsToPlanIds[ guid ] || 0;
}

```
(syntax might be off as I wrote everything here directly)


Writing this, the last if actually useless:
```
   
async function getSubscriberPlanIdIfExists( guid: VideoGUID ) : Promise<number> {  
     
     // Can also be defined outside
     const ready = new Promise( ( resolve ) => document.readyState != "loading" ? 
       resolve() : 
       document.addEventListener("DOMContentLoaded", resolve ) );
    
     await ready();
     return window.__guidsToPlanIds[ guid ] || 0;
}
```

or, in short

```
const ready = new Promise( ( resolve ) => document.readyState != "loading" ? 
       resolve() : 
       document.addEventListener("DOMContentLoaded", resolve ) );
       
const getSubscriberPlanIdIfExists = async guid: VideoGUID => await ready().map( () =>  window.__guidsToPlanIds[ guid ] || 0 );      
```
(syntax might be off as I wrote everything here directly)

projects/packages/videopress/src/class-access-control.php
```diff
@@ -0,0 +1,399 @@
...
+			$restriction_details = $this->default_video_restriction_details( false );
+			return $this->filter_video_restriction_details( $restriction_details, $guid, $embedded_post_id, $selected_plan_id );
+		}
+
+		$default_auth        = $this->get_default_user_capability_for_post( $post_to_check );
+		$restriction_details = $this->default_video_restriction_details( $default_auth );
+
+		if ( $this->jetpack_memberships_available() ) {
+			$memberships_can_view_post         = \Jetpack_Memberships::user_can_view_post( $embedded_post_id );
+			$restriction_details               = $this->get_subscriber_only_restriction_details( $default_auth );
```
millerf, line 150: This is already defined 2 lines above...

projects/packages/videopress/src/class-access-control.php
```diff
@@ -0,0 +1,399 @@
...
+				break;
+			case VIDEOPRESS_PRIVACY::SITE_DEFAULT:
+			default:
+				$is_videopress_private_for_site = Data::get_videopress_videos_private_for_site();
+				$is_user_authed                 = true;
+				if ( $is_videopress_private_for_site ) {
+					$restriction_details = $this->build_restriction_details( $guid, $embedded_post_id, $selected_plan_id );
+					$is_user_authed      = $restriction_details['can_access'];
+				}
+				break;
```
millerf: unnecessary `break;`

projects/packages/videopress/src/class-access-control.php
```diff
@@ -0,0 +1,399 @@
...
+	 * @param string $guid    The video guid that needs updated privacy details.
+	 * @param array  $details The details.
+	 */
+	private function update_video_restriction_details_on_wpcom( $guid, $details = array() ) {
+		$video_blog_id   = $this->get_videopress_blog_id();
+		$args            = array(
+			'headers' => array( 'content-type' => 'application/json' ),
+			'method'  => 'POST',
+		);
+		$default_details = array(
```
millerf: Would it make sense to use
```
 $this->default_video_restriction_details( false );
 ```
 ?

Comment (millerf) [reply projects/packages/videopress/src/class-access-control.php]: You will get translations for free.

File: projects/packages/videopress/src/class-access-control.php [hunk 17]
```diff
@@ -0,0 +1,399 @@
...
+	 *
+	 * @return int the blog id
+	 */
+	public function get_videopress_blog_id() {
+		if ( defined( 'IS_WPCOM' ) && IS_WPCOM ) {
+			return get_current_blog_id();
+		}
+
+		$options = Options::get_options();
+		return $options['shadow_blog_id'];
```
millerf: I think there is a helper function used now : `\Jetpack_Options::get_option( 'id' );` (it returns the cache site id)

projects/packages/videopress/src/class-access-control.php
```diff
@@ -0,0 +1,399 @@
...
+			'version'              => '1',
+			'provider'             => 'auth',
+			'title'                => 'Unauthorized',
+			'unauthorized_message' => 'Unauthorized',
+		);
+
+		$body = array_merge( $default_details, $details );
+
+		$endpoint = "sites/{$video_blog_id}/media/videopress-privacy-details/{$guid}";
+		$result   = Client::wpcom_json_api_request_as_blog( $endpoint, 'v2', $args, wp_json_encode( $body ), 'wpcom' );
```
millerf: As this is called every time a user restrictions is checked, shouldn't it be linked to the user and use
```
wpcom_json_api_request_as_user
```

projects/packages/videopress/src/class-access-control.php
```diff
@@ -0,0 +1,399 @@
...
+				$is_videopress_private_for_site = Data::get_videopress_videos_private_for_site();
+				$is_user_authed                 = true;
+				if ( $is_videopress_private_for_site ) {
+					$restriction_details = $this->build_restriction_details( $guid, $embedded_post_id, $selected_plan_id );
+					$is_user_authed      = $restriction_details['can_access'];
+				}
+				break;
+		}
+
+		$this->update_video_restriction_details_on_wpcom( $guid, $restriction_details );
```
millerf: I am struggling to understand why video  restriction details get updated on WPCOM every time `is_current_user_authed_for_video` is called...
At least shouldn't there be some cache (at least for the current request) to prevent multiple calls?

projects/packages/videopress/src/class-initializer.php
```diff
@@ -304,14 +307,26 @@ public static function render_videopress_video_block( $block_attributes, $conten
...
 		}
 
+		// Get premium content from block context
+		$premium_block_plan_id    = isset( $block->context['premium-content/planId'] ) ? intval( $block->context['premium-content/planId'] ) : 0;
+		$is_premium_content_child = isset( $block->context['isPremiumContentChild'] ) ? (bool) $block->context['isPremiumContentChild'] : false;
+		$maybe_premium_script     = '';
+		if ( $is_premium_content_child ) {
+			Access_Control::instance()->set_guid_subscription( $guid, $premium_block_plan_id );
+			$escaped_guid         = esc_js( $guid );
+			$script_content       = "if ( ! window.__guidsToPlanIds ) { window.__guidsToPlanIds = {}; }; window.__guidsToPlanIds['$escaped_guid'] = $premium_block_plan_id;";
```
millerf, line 317: That can simplified as well

```
$script_content = "window.__guidsToPlanIds = { '$escaped_guid': $premium_block_plan_id, ...window.__guidsToPlanIds };"

```

See this screenshot (totally legit)

Screenshot 2023-10-27 at 12 44 50

millerf (commented):
I left a few comments to maybe improve some JS.
Overall it looks good. 

I am only concerned about 
```
$this->update_video_restriction_details_on_wpcom( $guid, $restriction_details );
```
calling WPCOM everytime `is_current_user_authed_for_video` is called... But there might be some specifics to your logic that I might be missing. Might be worth some static cache for the time of the request?

Otherwise there is no issue with the Memberships interaction.

projects/packages/videopress/src/class-access-control.php
pgk: @millerf I thought about doing it on save but decided against it. Since the privacy setting is impacted by parent blocks, filters etc, it seems like the most "late-bound" place to do it. 

I will go ahead and remove the call entirely for this version of the feature. The impact if removed will be that subscriber-only videos will display a generic "video is private" error when not viewed within the site.

projects/packages/videopress/src/class-initializer.php
pgk: I was trying to be conservative with es6 features on this, since I am not sure where this will run.

projects/packages/videopress/src/class-access-control.php
pgk: removing this code entirely so things are more simple for now.

projects/packages/videopress/src/client/lib/token-bridge/index.ts
pgk: nice idea! Refactoring

millerf:
Looks good to me once rebased 👍

thedebian:
@millerf Can you approve the PR if it's ok now ?

millerf (approved):

PR #32158, opened 2023-07-31
CRM: Automation, rename trigger hooks and classes

cleacos:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3266

## Proposed changes:

This PR renames the hook and trigger class names. Changing `_new` by the verb `_created`.

### Other information:

## Jetpack product discussion
N/A

N/A

## Testing instructions:
- Check the code
- Run the unit tests and check they all pass. Remember to update the autoloader running `composer dumpautoload`

### DIFF
--- a/projects/plugins/crm/src/automation/commons/triggers/companies/class-company-created.php
+++ b/projects/plugins/crm/src/automation/commons/triggers/companies/class-company-created.php
@@ -1,51 +1,45 @@
 <?php
 /**
- * Jetpack CRM Automation Contact_New trigger.
+ * Jetpack CRM Automation Company_Created trigger.
  *
  * @package automattic/jetpack-crm
  */
 
 namespace Automattic\Jetpack\CRM\Automation\Triggers;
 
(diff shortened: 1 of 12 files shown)


projects/plugins/crm/src/automation/commons/triggers/companies/class-company-created.php
```diff
@@ -10,15 +10,15 @@
...
 /**
- * Adds the Company_New class.
+ * Adds the Company_Created class.
  */
-class Company_New extends Base_Trigger {
+class Company_Created extends Base_Trigger {
 
 	/** Get the slug name of the trigger
 	 * @return string
 	 */
```
gogdzl, line 19: Just a reminder, we should open an issue to cleanup these comments to comply with WP Standards.

gogdzl (approved):
Everything seems in order, the only thing I've noticed is that we should do a tidy up in our comments. But as far as this PR goes, LGTM!

PR #32161, opened 2023-07-31
AI Assistant: Add function calling feature

dhasilva:
Fixes #
Requires D117481-code

## Proposed changes:

* Adds support for `options.functions`, that should be an array of functions as described [here](https://levelup.gitconnected.com/openai-function-calling-feature-f846579c9c69).

### Other information:

## Testing instructions:

* Apply D117481-code and sandbox your public API
Test it locally by adding a function:
* On `components/ai-control/index.js`, add this declaration or something similar:
```js
const functions = [
	{
		name: 'getCurrentWeather',
		description: 'Returns the current weather in a given city of a country',
		parameters: {
			type: 'object',
			properties: {
				city: {
					type: 'string',
				},
				country: {
					type: 'string',
				},
			},
			required: [ 'city' ],
		},
		implementation: ( { city, country } ) => {
			return { city, country, temperature: 20, unit: 'C', description: 'Sunny' };
		},
	},
];
```
* In the `onClick` event of the `Send request` button (line 207 counting the snippet above), include the functions in the handler:
```js
onClick={ () => handleGetSuggestion( 'userPrompt', { functions } ) }
```
* Build jetpack plugin
* Go to the block editor
* Add an AI Assistant block
* Ask for the weather in your city and country
* There should be two requests. The first one sending the prompt and the functions object, the second one sending the update chat history with the result of the function.

First | Second
-|-
 | 

* The AI should respond with the weather.

* Check that there are no regressions.


projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-suggestions-from-openai/index.js
```diff
@@ -232,7 +240,57 @@ const useSuggestionsFromOpenAI = ( {
...
+
+			source.current = await askQuestion( prompt, {
+				postId,
+				requireUpgrade,
+				feature: attributes?.useGpt4 ? 'ai-assistant-experimental' : null,
+				functions: options.functions,
+			} );
+
+			// Add the listeners back
+			source?.current?.addEventListener( 'functionDone', onFunctionDone );
```
retrofox: The event-name shape isn't consistent with the rest of the events.
```suggestion
			source?.current?.addEventListener( 'function_done', onFunctionDone );
```

retrofox (approved):
LGTM. Nice feature. I've tested following the instructions, but not for beginners. 

image

🚀 🚀 🚀

PR #32167, opened 2023-08-01
CRM - Automation: Add Data Types and Data Transformers

kallehauge:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3262

We currently have two challenges:

1. Not all hooks share the same data despite being for the same "data type". To combat this, we've implemented data types that will unify the data, so steps can always rely on the same structure.
    * This also opens up for an easier way for transformers to talk together (see more below).
2. We currently don't have a way for e.g. a contact specific step to be triggered by invoice events. To combat this, we've implemented our first "transformer" and all of the underlying logic required to start adding more of them.

_Aside: Updating 90% of tests with data types in this PR reviled quite a few false-positive automated tests. Some of the tests was easier to refactor to actual integration tests as well which mostly reviled further assumption errors. The overall message: we probably have a few more false positives, so going over our tests ones more would make sense._

## Proposed changes:

* Add new data types for all objects (`contact`, `invoice`, and `transaction` have received the most attention, but all tests are satisfied at the moment).
* Add a data transformer that allows invoice triggers to work with contact specific steps.
    * We still need to introduce even more transformers for better cross support. The "Invoice => Contact" transformer in this PR serves as an example of how we can keep expanding moving forward.
* Move our `Automation_Workflow::execute_workflow` logic to the engine
  * This makes sense since we keep referencing engine specific methods and this PR introduces new `Data_Type_*` and `Data_Transformer_*` logic that would otherwise have had to be injected into the workflow. 
  * This could use another round of refactoring, but it would increase the already humongous size of the PR.
* Update all Triggers and Actions to have a static `get_data_type()` method (this is different from category which should be used for front-end categorization). 
* Update _**ALL**_ existing conditions/actions and tests to support data types.
* The following pattern exists in many different places: `$this->get_attributes()['field']`. We should probably great a `get_attribute( $key )` method.

#### Follow-up tasks

* We're missing automated tests for our data types and transformers.
* We need to implement more transformers for better cross-object support.
* Conditions and actions have a new, optional, parameter (`?Data_Type_Base $previous_data = null`). This was added to _theoretically_ support our transitional scenarios, but this could probably be implemented much cleaner.
    * I know that @gogdzl is curious about how we pass values to them in real workflows as well, so that's probably another automated test we need to make sure it works.
    * _**Edit**: Thinking about it with a fresh set of eyes, then perhaps the Data_Type instance could contain a "previous state" instead of specifically passing a second argument just for the few instances where we need it._
* Create PR to store data types on the workflow so we do not have to re-fetch them.

#### Removals

**New_Contact action**

I haven't spent time updating the `New_Contact` action tests since I'm unsure how it fits into our system ([they can be found here if we want to re-implement them](https://github.com/Automattic/jetpack/commit/0dddeabdc7c951638ab6fa89f01fb2a0c8873bdc#diff-827e397b80973ee6b3f74a81d669a5232d527f87cca3b9299aeca7aba2c65a1bR196-R286)).
TL;DR: It's not like it's unheard of to create a contact as an action, but contacts tends to be very dynamic, and we do not have any data types or sources/plans to add dynamic value fields like this.

### Other information:

## Jetpack product discussion

pbhBOz-2ES-p2

No.

## Testing instructions:

* In theory, all of our automated tests should be the only testing we need here.


cleacos:
The action code (workflow configuration) for testing should be updated:

```
add_filter(
	'jpcrm_automation_workflows',
	function ( $workflows ) {
		// New updated: Update contact status to "Customer".
			[
				'id'           => 'testing2',
				'name'         => __( 'Invoice status updated: Set status to "Customer"', 'zero-bs-crm' ),
				'description'  => __( 'This is just a test.', 'zero-bs-crm' ),
				'category'     => 'contact',
				'triggers'     => [
					\Automattic\Jetpack\CRM\Automation\Triggers\Invoice_Updated::get_slug(),
				],
				'initial_step' => [
					'class_name'      => \Automattic\Jetpack\CRM\Automation\Actions\Update_Contact_Status::class,
					'attributes'      => [
						'new_status' => 'Customer',
					],
					'next_step'  => null,
				],
			]
		);

		$workflows[ $workflow->get_id() ] = $workflow;

		return $workflows;
	}
);
```

The initial step is an action, so it should have `next_step` (removed `next_step_true` and `next_step_false`). Also, since the class name is provided, the slug is not needed (`contact_status_condition`)

projects/plugins/crm/src/automation/class-automation-engine.php
```diff
@@ -179,6 +298,98 @@ public function init_workflows() {
...
+				/** @var Step $step */
+				$step = new $step_class( $step_data );
+
+				if ( isset( $step_data['attributes'] ) && is_array( $step_data['attributes'] ) ) {
+					$step->set_attributes( $step_data['attributes'] );
+				}
+
+				$this->get_logger()->log( '[' . $step->get_slug() . '] Executing step. Type: ' . $step->get_data_type() );
+
+				$hook_data = $this->maybe_transform_data_type( $hook_data, $step::get_data_type() );
```
cleacos: We should preserve the original data type throughout the workflow and adapt it (Input Adapter) to the right type at each step.

cleacos: The reason is that a step below in the workflow could need the original type object, for example:

1. On a new invoice created, the trigger receives an invoice object and execute a workflow.
2. Contact action: Set the status:
    - Adapt the input data (Invoice) and pass the contact object to the action.
    - Execute the action: set status to `Customer`
4. Send invoice email action: send to the contact the invoice
    - It needs the invoice.
    - From the invoice, it could get the Contact data.

cleacos:
After understanding the problem that this PR solves. Thank you for dealing with it @kallehauge on you faced with the problem.

I have some thoughts... dropping here:

- I still think that the Adapter concept fits better here since we will adapt the input just before the step execution. The original data type should be kept, we are not going to transform it through the workflow.
- The original object type should be kept while it's traveling through the workflow, and only adapt the input with the right object type before the step execution and pass the expected object, which is linked to the original object (i.e. an invoice has a contact linked).
- Maybe it is a good opportunity to start implementing some JPCRM classes (the right way) Contact, Invoice...  It will simplify things a little, the Event Manager could send the object and then, in the workflow, adapt the input, for instance, from an invoice to get the contact, and for example, it could just be as simple as `$invoice->get_contact()` and return the Contact object directly.

cleacos:
~~I was going to push a PR for adding the `version` property to the `Automation_Workflow`, but it's worth it... Could you add that property to this PR? We are modifying the Automation_Workflow class in this PR, so it's OK to introduce that small change.~~

~~The idea behind having a version for the workflow stored is that we will probably change the structure in the future, for example, to support multiple flows, so we will need to implement the pointer way.~~

Forget it, I just added [the PR](https://github.com/Automattic/jetpack/pull/32280) for it.

projects/plugins/crm/src/automation/data-transformers/class-data-transformer-invoice-to-contact.php
```diff
@@ -0,0 +1,72 @@
...
+use Automattic\Jetpack\CRM\Automation\Data_Types\Data_Type_Base;
+use Automattic\Jetpack\CRM\Automation\Data_Types\Data_Type_Contact;
+use Automattic\Jetpack\CRM\Automation\Data_Types\Data_Type_Invoice;
+
+/**
+ * CRM Invoice to CRM Contact Transformer class.
+ *
+ * @since $$next-version$$
+ *
+ * @package Automattic\Jetpack\CRM\Automation
```
cleacos: Should we use @package in classes too? Or only in the file header as per WP Coding standard expects.

cleacos: There are more classes in this PR with the @package tag added to their PHPDoc.

cleacos:
> We're missing automated tests for our data types and transformers.

Task created: https://github.com/Automattic/zero-bs-crm/issues/3228

cleacos:
> We need to implement more transformers for better cross-object support.

We have to implement them as we go, or implement the most evident and well know cases. We already have the task for creating a real use case, which should cover a complex workflow. Once the real case is designed it will pull transformers and data types required.

cleacos:
> Create PR to store data types on the workflow so we do not have to re-fetch them.

I created the issue: https://github.com/Automattic/zero-bs-crm/issues/3284

Clarification: this task is about keeping the `$trigger_data` through all the workflow execution, and only transforming the data before each step execution, so the original data goes through the workflow.

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Contacts.php
```diff
@@ -3251,7 +3251,7 @@ public function addUpdateContact($args=array()){
                             'customerExtraMeta'=>$confirmedExtraMeta #} This is the "extraMeta" passed (as saved)
                         ));
 
-						$dataArr['id'] = $newID; // phpcs:ignore Generic.WhiteSpace.ScopeIndent.Incorrect
+						$dataArr['ID'] = $newID; // phpcs:ignore Generic.WhiteSpace.ScopeIndent.Incorrect
```
cleacos, line 3254: We have to test if this change could affect the normal operation of Jetpack CRM

projects/plugins/crm/src/automation/commons/actions/contacts/class-add-contact-log.php
```diff
@@ -88,12 +89,23 @@ public static function get_allowed_triggers(): ?array {
...
-	public function execute( array $contact_data = array() ) { // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
+	public function execute( $data, $previous_data = null ) { // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
 		global $zbs;
 
-		$zbs->DAL->contacts->zeroBS_addUpdateObjLog( $this->attributes ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+		$zbs->DAL->logs->addUpdateLog( // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+			array(
+				'data' => array(
+					'objtype'   => ZBS_TYPE_CONTACT,
+					'objid'     => $data['id'],
```
cleacos, line 102: `ID` or `id`?

projects/plugins/crm/src/automation/data-types/class-data-type-contact.php
```diff
@@ -0,0 +1,113 @@
...
+	 * with a single version of the formatting (the most recent one).
+	 *
+	 * @since $$next-version$$
+	 *
+	 * @param mixed $entity The data we want to potentially prepare.
+	 * @return array The unified data.
+	 */
+	public function unify_data( $entity ): array {
+		if ( ! is_array( $entity ) ) {
+			return $entity;
```
cleacos, line 102: If it's not an array, it cannot return something different to an array

cleacos: Fixed here: https://github.com/Automattic/jetpack/pull/32167/commits/176656a51f6dd3574cb91a9d01ca9289611b096b

projects/plugins/crm/src/automation/data-types/class-data-type-invoice.php
```diff
@@ -0,0 +1,115 @@
...
+	 *
+	 * @param mixed $entity The invoice entity data.
+	 * @return void
+	 *
+	 * @throws \Automattic\Jetpack\CRM\Automation\Data_Type_Exception If the entity is not valid.
+	 */
+	public function __construct( $entity ) {
+		$entity = $this->unify_data( $entity );
+
+		parent::__construct( $entity );
```
cleacos, line 41: It seems that the call to the parent constructor should be before unify_data(). The Data_Type_Base constructor performs some data validations.

projects/plugins/crm/src/automation/commons/actions/contacts/class-add-contact-log.php
cleacos: Okay, at this point should be `id`

projects/plugins/crm/src/automation/data-types/class-data-type-invoice.php
cleacos: All good. The data validation should happen after `unify_data`

cleacos (commented):
I've adapted the`Quote_Status_Changed` condition to work with the Data_Type and Transformers. Tests were failing here.

cleacos (approved):
The code looks good. Now all tests passed.

LGTM!

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Contacts.php
cleacos: All is good here. It's only used for notifications and converted by `tidy_contact` call from `ID` to `id`

PR #32174, opened 2023-08-01
Init Super Cache 2.0.1-alpha

haqadn:
## Proposed changes:

* Init Super Cache 2.0.1-alpha

### Other information:

## Jetpack product discussion

No

No

## Testing instructions:

None


dilirity (commented):
I'm a bit confused by having 2.0.0-beta and 2.0.1-alpha. Why is there a difference?

haqadn:
> I'm a bit confused by having 2.0.0-beta and 2.0.1-alpha. Why is there a difference?

2.0.0 is the version we are in the process of releasing, and we have released the beta. `2.0.1-alpha` is the newly initiated version which is the next release.

PR #32177, opened 2023-08-01
Add Calypso path to launchpad task data

daledupreez:
## Proposed changes:

The core change in this PR involves returning a new `calypso_path` string for tasks where we know what Calypso page we want to take users to for that task. The goal of the change is to simplify new task creation by ensuring that we don't need separate Calypso PRs to add specific handling for each new task. Instead, we can simply generate the right path on the server and add generic handling to the client code so that we only have one place to change code for tasks.

The implementation is handled by the following changes:
 * The generic `load_value_from_callback()` helper now takes an additional `$data` argument, which defaults to `array()`, and allows callers to pass in context-specific data, which is then passed to the callback.
 * Tasks can specify a new `get_calypso_path` callback, which accepts three arguments, `$task`, `$default`, and `$data`, where `$data['site_slug']` will contain the site slug and `$data['site_slug_encoded']` will contain the URL-encoded site slug
    - Note that we use `WPCOM_Masterbar::get_calypso_site_slug()` to compute the site slug when that class and function exist, but we fall back on `Automattic\Jetpack\Status->get_site_suffix()`, and then process `site_url()`. The fallbacks here are borrowed from https://github.com/Automattic/wc-calypso-bridge/blob/85664e2c7836b2ddc29e99871ec2c5dc4015bcc8/class-wc-calypso-bridge.php#L227-L251
 * The task list code validates that the returned path is an absolute path of the form `/some/path`

A separate patch in Calypso will update the client-side handling.

### Other information:

## Jetpack product discussion

No product discussion as this is specifically intended for WordPress.com sites.

No changes to data or tracking.

## Testing instructions:

* Apply the changes in this patch to your WordPress.com development system
* Navigate to a site with a launchpad shown in Customer Home
* Verify that the tasks return valid `calypso_path` values where expected

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/changelog/add-calypso-path-data-to-launchpad-tasks
+++ b/projects/packages/jetpack-mu-wpcom/changelog/add-calypso-path-data-to-launchpad-tasks
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Launchpad: add calypso_url property to tasks where we know the Calypso page we want to show
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
@@ -30,6 +30,13 @@ class Launchpad_Task_Lists {
 	private $task_registry = array();
 
 	/**
+	 * Internal reference for the current site slug.
+	 *
+	 * @var string|null
+	 */
+	private $site_slug = null;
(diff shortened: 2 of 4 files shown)


valterlorran (dismissed):
LGTM! I loved this ideia!
Screen Shot 2023-08-01 at 09 26 50

daledupreez:
Can I request a quick set of eyes on this PR? I rebased to pick up the `manage_subscribers` task.

valterlorran (approved):
Still LGTM!

Issue #32124, opened 2023-07-27
Newsletters: Paywall block TODO


lezama:
peKye1-fd-p2

```[tasklist]
### Tasks
- [x] Check the Block Preview in the editor
- [ ] https://github.com/Automattic/jetpack/pull/32182
- [ ] https://github.com/Automattic/jetpack/pull/32190
- [ ] https://github.com/Automattic/jetpack/pull/32201
- [ ] https://github.com/Automattic/wp-calypso/issues/79976
- [ ] https://github.com/Automattic/jetpack/issues/32334
- [x] Add block transform – p2-p9MPsk-3Zr#comment-27764
- [ ] We should ensure excerpts, social shares etc continue functioning without leaking paywalled content. – jetpack_filter_excerpt_for_newsletter 👀
- [x] Test Emails
- [x] Test Reader – RSS — app
- [x] We need to consider how this will work in post editor within the native apps, or “fail gracefully” meanwhile. See Project pcdRpT-3rt-p2
- [x] Release
```

simison:
> Test Emails
> Test Reader – RSS


Will need some styling added, [issue for the reader](https://github.com/Automattic/wp-calypso/issues/79976).

There should be fieldguide docs and example PRs for both reader and emails.

PR #32182, opened 2023-08-01
Paywall Block: register only if Subscription module

lezama:
Related #32124

## Proposed changes:
Register Paywall block only if Subscription module is activated 

### Other information:

NO

## Testing instructions:

* Turn subscription module on/off and check if the block is available.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/paywall/paywall.php
+++ b/projects/plugins/jetpack/extensions/blocks/paywall/paywall.php
@@ -21,6 +21,10 @@ const BLOCK_NAME   = 'jetpack/' . FEATURE_NAME;
  * registration if we need to.
  */
 function register_block() {
+	// check if the subscription module is enabled
+	if ( ! class_exists( '\Jetpack_Memberships' ) ) {
+		return;
+	}
 	Blocks::jetpack_register_block(
 		BLOCK_NAME,
 		array( 'render_callback' => __NAMESPACE__ . '\load_assets' )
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/paywall/paywall.php
```diff
@@ -21,6 +21,10 @@
  * registration if we need to.
  */
 function register_block() {
+	// check if the subscription module is enabled
+	if ( ! class_exists( '\Jetpack_Memberships' ) ) {
```
simison, line 27: This checks if the function is there, but isn't here a different function to check if the module is active? I think you'd need both.

image

simison: Let's do these same checks here too:

https://github.com/Automattic/jetpack/blob/30422d3ebd5c495e0ca13c13874e161623090e96/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php#L63-L77

lezama: oops, sorry for the waste of time, I copied from somewhere and didn't test, should be good now!

projects/plugins/jetpack/extensions/blocks/paywall/paywall.php
```diff
@@ -21,6 +21,14 @@
  * registration if we need to.
  */
 function register_block() {
+	// check if the subscription module is enabled
```
simison: Nit: the comment kinda just repeats what was pretty clear code already. :-)

simison (dismissed):
Works well and looks good now!

edanzer (approved):
Code changes look good, and this tests as expected.

PR #32185, opened 2023-08-01
CRM: Tidy up PHPDocs to comply with WP Coding Standard

cleacos:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3267

## Proposed changes:

This PR simply fixes some PHPDocs that didn't comply with the WP Coding Standard. It adds blank lines, spaces, and punctuation.

### Other information:

## Jetpack product discussion
N/A

N/A

## Testing instructions:

- Check the PHPDocs updated


projects/plugins/crm/src/automation/class-automation-engine.php
```diff
@@ -14,23 +14,23 @@
  */
 class Automation_Engine {
 
-	/** @var Automation_Engine Instance singleton */
+	/** @var Automation_Engine Instance singleton. */
```
gogdzl: WP Coding standards expect a summary and description in different lines, in the form:
```
/**
 * Summary.
 *
 * @var type $var Description.
 */
```
Source: https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#2-1-class-members

This applies to all class members in this PR.

projects/plugins/crm/src/automation/class-base-condition.php
```diff
@@ -8,17 +8,17 @@
 namespace Automattic\Jetpack\CRM\Automation;
 
 /**
- * Base Condition Step
+ * Base Condition Step.
  *
  * @inheritDoc
  */
```
gogdzl, line 16: We could add a ` * @since $$next-version$$` tag here. (and also to all other classes)

projects/plugins/crm/src/automation/class-automation-engine.php
```diff
@@ -241,7 +241,7 @@ public function get_logger(): ?Automation_Logger {
 	}
 
 	/**
-	 * Get the registered triggers
+	 * Get the registered triggers.
```
gogdzl, line 299: We are missing most of our `@return` tags.
In this case it would be something like:
```
 * @return array An associative array containing the registered triggers,
 *               where the keys represent the trigger names and the values represent
 *               their corresponding slugs.
```

gogdzl (changes requested):
I left some inline comments on possible improvements!
Thanks for tackling this huge task!

cleacos:
Thank you for the quick review... I'll fix them

projects/plugins/crm/src/automation/class-base-condition.php
kallehauge: I haven't checked docs or code but I think inheritDoc should be wrapped in `{}`: `{@inheritDoc}`

cleacos:
Thank you for the review @gogdzl. I'm fixing them and more... this is going to take more time than I expected :)

cleacos:
Done! I hope all are fixed now. @gogdzl

projects/plugins/crm/src/automation/class-automation-engine.php
cleacos: Fixed! Thanks

projects/plugins/crm/src/automation/class-automation-engine.php
cleacos: Fixed!

projects/plugins/crm/src/automation/class-base-condition.php
cleacos: Yes, it seems that PHPDocumentor 2 accepted it, but the new versions require to be wrapped in `{}` 

Fixed.

gogdzl (dismissed):
Wow! Looks good! Nice job =D

projects/plugins/crm/src/automation/class-automation-engine.php
```diff
@@ -8,32 +8,42 @@
...
  *
- * @package Automattic\Jetpack\CRM\Automation
+ * @since $$next-version$$
  */
 class Automation_Engine {
 
-	/** @var Automation_Engine Instance singleton */
+	/**
+	 * @var Automation_Engine Instance singleton.
+	 */
```
kallehauge: This is the format shown [in the WP CS](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#2-1-class-members):

```php
/**
 * Summary.
 *
 * @since x.x.x
 * @var type $var Description.
 */
```

```suggestion
	/**
	 * A singleton instance of the engine.
	 *
	 * @since $$next-version$$
	 * @var Automation_Engine
	 */
```

projects/plugins/crm/src/automation/class-automation-workflow.php
```diff
@@ -218,45 +242,49 @@ private function get_step_class( array $step_data ): string {
 	}
 
 	/**
-	 * Turn on the workflow
+	 * Turn on the workflow.
 	 */
```
kallehauge: > Every function, hook, class, and method should have a corresponding @\since version associated with it...
> [source](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#since-section-changelogs)

So we should add `@since $$next-version$$` to all methods in all classes (and interfaces) as well

projects/plugins/crm/src/automation/class-automation-exception.php
```diff
@@ -9,6 +9,8 @@
 
 /**
  * Adds the Automation_Exception class.
+ *
+ * @since $$next-version$$
  */
 class Automation_Exception extends \Exception {
 	const STEP_CLASS_NOT_FOUND    = 10;
```
kallehauge: [The WP CS specifies](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#since-section-changelogs) that constants should also have a DocBlock (and that it includes a since tag as well):

```php
/**
 * Summary.
 *
 * @since x.x.x
 * @var type $var Description.
 */
const NAME = value;
```

The docs does say that the `@var` description may be omitted which I feel like we should do because it doesn't make a lot of sense 

```suggestion
	/**
	 * Step class not found exception code.
	 * 
	 * @since $$next-version$$
	 * @var int
	 */
	const STEP_CLASS_NOT_FOUND    = 10;
```

projects/plugins/crm/src/automation/class-automation-engine.php
```diff
@@ -203,19 +211,19 @@ public function get_registered_step( $step_name, array $step_data = array() ): S
...
 	}
 
 	/**
-	 * Get trigger instance
+	 * Get trigger instance.
 	 *
 	 * @param string $trigger_slug The name of the trigger slug with which to retrieve the trigger class.
-	 * @return string
+	 * @return string The name of the trigger class.
 	 * @throws Automation_Exception Throws an exception if the step class does not exist.
```
kallehauge, line 272: I'm not 100% sure since the WP Docs doesn't say anything but I feel like WP Core (the code) tends to add a new line above `@throws`

kallehauge (changes requested):

projects/plugins/crm/src/automation/class-base-action.php
```diff
@@ -8,9 +8,10 @@
 namespace Automattic\Jetpack\CRM\Automation;
 
 /**
- * Base Action Step
+ * Base Action Step.
  *
- * @inheritDoc
+ * @since $$next-version$$
+ * {@inheritDoc}
```
kallehauge, line 15: The parent docs doesn't really contain anything so `{@inheritDoc}` doesn't seem to provide any value in these cases.

projects/plugins/crm/src/automation/class-automation-engine.php
```diff
@@ -8,32 +8,42 @@
...
 
-	/** @var Automation_Engine Instance singleton */
+	/**
+	 * @var Automation_Engine Instance singleton.
+	 */
 	private static $instance = null;
 
-	/** @var array triggers map name => classname */
+	/**
+	 * @var array triggers map name => classname.
```
kallehauge: This isn't really part of the WP CS but my own preference. If we know that an array is a list of the same type of "thing", then we might as well define it so IDE's can better understand it - and may even support auto complete/suggestion
```suggestion
	 * @var string[] triggers map name => classname.
```

projects/plugins/crm/src/automation/class-automation-engine.php
cleacos: I've seen it in different formats, before, together, with a blank line...

projects/plugins/crm/src/automation/class-base-action.php
cleacos: `Base_Step`? it has the parent docs.

cleacos:
Fixed! more PHPDocs tweaks

kallehauge (approved):
LGTM! 👍

PR #32190, opened 2023-08-01
Paywall: restrict to posts only

lezama:
Related https://github.com/Automattic/jetpack/issues/32124

## Proposed changes:
* Unregister the Paywall block from non `post` editors as the subscription module only works with posts right now.

### Other information:

No

## Testing instructions:

* Check that only when editing posts you can pick the Paywall block

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/paywall/editor.js
+++ b/projects/plugins/jetpack/extensions/blocks/paywall/editor.js
@@ -1,9 +1,10 @@
+import { unregisterBlockType } from '@wordpress/blocks';
+import { subscribe, select } from '@wordpress/data';
 import registerJetpackBlock from '../../shared/register-jetpack-block';
 import { name, settings } from '.';
 
-/* global wp */
-const unsubscribe = wp.data.subscribe( () => {
-	const postType = wp.data.select( 'core/editor' ).getCurrentPostType();
+const unsubscribe = subscribe( () => {
+	const postType = select( 'core/editor' ).getCurrentPostType();
 
 	// If postType is still not available, simply return and wait for the next call.
 	if ( postType === null ) {
@@ -12,7 +13,7 @@ const unsubscribe = wp.data.subscribe( () => {
 	unsubscribe();
 	// If postType is defined and not 'post', unregister the block.
 	if ( postType && postType !== 'post' ) {
-		wp.blocks.unregisterBlockType( 'jetpack/' + name );
+		unregisterBlockType( 'jetpack/' + name );
 	}
 } );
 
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/paywall/editor.js
```diff
@@ -1,4 +1,19 @@
 import registerJetpackBlock from '../../shared/register-jetpack-block';
 import { name, settings } from '.';
 
+/* global wp */
```
simison: Let's use the normal imports from packages:

```js
import { subscribe, select } from '@wordpress/data';
import { unregisterBlockType } from '@wordpress/blocks';
```

projects/plugins/jetpack/extensions/blocks/paywall/editor.js
```diff
@@ -1,4 +1,20 @@
...
+	if ( postType === null ) {
+		return;
+	}
+	unsubscribe();
+	// If postType is defined and not 'post', unregister the block.
+	if ( postType && postType !== 'post' ) {
+		unregisterBlockType( 'jetpack/' + name );
+	}
+} );
+
```
edanzer, line 19: This tests as expected - I don't see the Paywall block on pages. Small question - looks like we using recursion here and I was wondering if you could explain the logic? I think subscribe adds a listener that will run any time state changes. Since it already runs each time state changes, I'm not understanding why we'd need the recursive call.

lezama: it's not recursive, it calls `unsubscribe`, so we don't try to unregister multiple times!

edanzer: So from separate conversation, the WP subscribe() method returns unsubscribe, which can be called later to undo the subscribe() call (ie, remove the listener so it doesn't keep re-running each time state is updated). 

We call it here once we have a post type, since we do not need to continue running this code after that point. 

Seems like a bit of an odd construct to have unsubcribe() returned from subscribe() and also called within subscribe(), but also sounds like this is not an uncommon pattern with this particular method.

edanzer (approved):
Tests as expected.

And resolved my separate question/convo about the behavior of unsubscribe() here. Looks like we'll still need to rebase. If you want another test after that, let me know!

projects/plugins/jetpack/extensions/blocks/paywall/editor.js
lezama: 🙌 thanks for the 🕵️ work

PR #32196, opened 2023-08-01
Widgets: refactor Internet Defense League widget to work locally

jeherve:
## Proposed changes:

The Internet Defense League widget is pretty old at this point, and the service behind it (Internet Defense League) doesn't exist anymore. The campaigns do not exist anymore, and the website now redirects to the company who first came up with the Internet Defense League.

The widget isn't offered to new Jetpack users today, since #22676. You can only add the widget if you use the Classic Widgets plugin.
Still, site owners who had added the widget some time in the past can still view and modify it. Since the campaign website doesn't exist anymore, we must refactor the widget so it still works:

- We remove the campaign script entirely, as well as the logic to select a campaign.
- We load the images locally instead of loading them from the remote site.
- We redirect to https://www.fightforthefuture.org/ instead of the old domain.

### Other information:

## Jetpack product discussion

See p1690903216438169-slack-CBG1CP4EN

* No

## Testing instructions:

* Go to Plugins > Add New, install and activate the Classic Widgets plugin.
* Go to Appearance > Themes and switch to a non-block theme, like Twenty Twenty.
* Go to Appearance > Widgets, and add an Internet Defense Widget.
* Ensure it is displayed properly on the frontend.


zinigor (commented):
I can see that this is mostly removing code that will not get used anymore, and it looks fine. However, what do you think about leaving methods behind with a deprecation notice instead of removing them completely? I mean, it will probably be redundant, but just to be super extra cautious+ :)

jeherve:
> I can see that this is mostly removing code that will not get used anymore, and it looks fine. However, what do you think about leaving methods behind with a deprecation notice instead of removing them completely? I mean, it will probably be redundant, but just to be super extra cautious+ :)

I don't think many folks out there are using those methods, but it doesn't hurt to leave them around for a few releases. Better be safe than sorry! Done in 70a1177bd96d72bcffa141345c2e36c48e84e53d

zinigor (approved):
Thank you for making the changes!

zinigor:
Ouch, you will need to rebase this now.

PR #32200, opened 2023-08-01
Launchpad: dismiss api endpoint

andres-blanco:
Context: paYKcK-3hW-p2#comment-2358 

## Proposed changes:

Updates the `/sites/[YOUR_SITE]/launchpad` API endpoint to query and update the visibility of a task list.

An option called `wpcom_launchpad_config` was added to track general config values for the launchpad in general. This PR handles the launchpad visibility using the `task_list_dismissed_status` key inside that option. All entries in this array follow the form `list-id => true|false` indicating whether the task list with id `list-id` is dismissed.
If the list id is not present in the array stored in `task_list_dismissed_status` then it's assumed to not be dismissed to maintain backward compatibility.

This new option isn't being added to JP sync right now because it's not needed for the dismiss endpoint, but it can be easily added on [this method](https://github.com/Automattic/jetpack/blob/trunk/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php#L669) when necessity arises.

### Other information:

## Jetpack product discussion

p9Jlb4-7CP-p2#comment-8725

No

## Testing instructions:

* Apply the changes in this PR to your sandbox
* Go to dev API console and prepare it to run queries on `WP Rest API > wpcom/v2`
* Make a GET to `/sites/[YOUR_SITE]/launchpad?checklist_slug=intent-build`
* Verify that there's an `is_dismissed` key in the result and that it's set to `false`
* Make a POST to `/sites/[YOUR_SITE]/launchpad` with the following value in the `is_checklist_dismissed` param: `{"slug":"intent-build","is_dismissed":true}` (you can check this out to see how to do this from the dev console: p1690917561303459-slack-C0Q664T29)
* Make a GET to `/sites/[YOUR_SITE]/launchpad?checklist_slug=intent-build`
* Verify that there's an `is_dismissed` key in the result and that it's set to `true`

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
@@ -159,10 +159,10 @@ class Launchpad_Task_Lists {
 	 * @return bool|null True if visible, false if not.
 	 */
 	public function is_task_list_visible( $id ) {
-		$task_list_status = get_option( 'wpcom_launchpad_task_list_visibility', array() );
+		$task_list_visibility = $this->get_task_list_visibility_status();
 
 		// Return true if the task list is not in the visibility option or if it is and the value is true.
-		return ! isset( $task_list_status[ $id ] ) || $task_list_status[ $id ];
+		return ! isset( $task_list_visibility[ $id ] ) || $task_list_visibility[ $id ];
 	}
 
 	/**
@@ -172,9 +172,22 @@ class Launchpad_Task_Lists {
 	 * @param bool   $is_visible True if visible, false if not.
 	 */
 	public function set_task_list_visibility( $id, $is_visible ) {
-		$task_list_status        = get_option( 'wpcom_launchpad_task_list_visibility', array() );
-		$task_list_status[ $id ] = $is_visible;
-		update_option( 'wpcom_launchpad_task_list_visibility', $task_list_status );
+		$task_list_visibility        = $this->get_task_list_visibility_status();
+		$task_list_visibility[ $id ] = $is_visible;
+
+		$launchpad_config                         = get_option( 'wpcom_launchpad_config', array() );
+		$launchpad_config['task_list_visibility'] = $task_list_visibility;
+		update_option( 'wpcom_launchpad_config', $launchpad_config );
+	}
+
+	/**
+	 * Get the task list visibility status for a site.
+	 *
+	 * @return array
+	 */
+	protected function get_task_list_visibility_status() {
+		$launchpad_config = get_option( 'wpcom_launchpad_config', array() );
+		return isset( $launchpad_config['task_list_visibility'] ) ? $launchpad_config['task_list_visibility'] : array();
 	}
 
 	/**
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
@@ -172,8 +172,14 @@ class Launchpad_Task_Lists {
 	 * @param bool   $is_dismissed True if dismissed, false if not.
 	 */
 	public function set_task_list_dismissed( $id, $is_dismissed ) {
-		$task_list_dismissed_status        = $this->get_task_list_dismissed_status();
-		$task_list_dismissed_status[ $id ] = $is_dismissed;
+		$task_list_dismissed_status = $this->get_task_list_dismissed_status();
+		$is_dismissed               = (bool) $is_dismissed;
+
+		if ( $is_dismissed ) {
+			$task_list_dismissed_status[ $id ] = true;
+		} else {
+			unset( $task_list_dismissed_status[ $id ] );
+		}
 
 		$launchpad_config                               = get_option( 'wpcom_launchpad_config', array() );
 		$launchpad_config['task_list_dismissed_status'] = $task_list_dismissed_status;
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
@@ -172,6 +172,11 @@ class Launchpad_Task_Lists {
 	 * @param bool   $is_dismissed True if dismissed, false if not.
 	 */
 	public function set_task_list_dismissed( $id, $is_dismissed ) {
+		$task_list = $this->get_task_list( $id );
+		if ( empty( $id ) || empty( $task_list ) ) {
+			return;
+		}
+
 		$task_list_dismissed_status = $this->get_task_list_dismissed_status();
 		$is_dismissed               = (bool) $is_dismissed;
 
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
@@ -162,7 +162,7 @@ class Launchpad_Task_Lists {
 		$task_list_dismissed_status = $this->get_task_list_dismissed_status();
 
 		// Return true if the task list is on the dismissed status array and its value is true.
-		return isset( $task_list_dismissed_status[ $id ] ) && $task_list_dismissed_status[ $id ];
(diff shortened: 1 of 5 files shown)


daledupreez:
I haven't had a chance to review or test in detail, but I had an immediate thought about these flags: do we want to set visibility based on context? Right now, a task list is either visible or invisible, but that's mostly because we're only thinking of task lists in Customer Home. But there's nothing in this logic that captures that we're working with Customer Home (or any other specific context). I don't know the answer, as this introduces another layer of complexity, and it's not clear to me that we'd expect the same task list to appear in (1) multiple contexts, though we're exploring this when skipping the full-screen launchpad, or (2) in multiple _dismissible_ contexts. But I think we should consider it and make a deliberate choice.

daledupreez:
As another thought, might it make sense for us to store this information in an explicit `is_dismissed` value in the API data?

andres-blanco:
> I haven't had a chance to review or test in detail, but I had an immediate thought about these flags: do we want to set visibility based on context? Right now, a task list is either visible or invisible, but that's mostly because we're only thinking of task lists in Customer Home. But there's nothing in this logic that captures that we're working with Customer Home (or any other specific context). I don't know the answer, as this introduces another layer of complexity, and it's not clear to me that we'd expect the same task list to appear in (1) multiple contexts, though we're exploring this when skipping the full-screen launchpad, or (2) in multiple _dismissible_ contexts. But I think we should consider it and make a deliberate choice.

In my mind adding a contextual visibility shields little gains and it does introduce some complexity. If on some context we want to show the tasklist disregarding this value we can always ignore it on Calypso. 
Can you think about some specific use case where storing the context is useful?

> As another thought, might it make sense for us to store this information in an explicit is_dismissed value in the API data?

I am not sure I understand what you mean. Right now the stored value is [being returned](https://github.com/Automattic/jetpack/pull/32200/files#diff-49fd3d5edac66c43fefb584a72f3229786f117afe4efa6645341962dcf9eea84R141) when calling `/sites/[YOUR_SITE]/launchpad?checklist_slug=[LIST_SLUG]`.
Do you mean to change the `is_visible` name to `is_dismissed`? Do you think it should be on a separate endpoint?

Thanks for the feedback ❤️

daledupreez:
> In my mind adding a contextual visibility shields little gains and it does introduce some complexity. If on some context we want to show the tasklist disregarding this value we can always ignore it on Calypso.
Can you think about some specific use case where storing the context is useful?

I think that's fair - I hadn't come up with a use case, I was just making sure we'd actively decided against that approach. I think this will overcomplicate things in a way that's unlikely to benefit anyone. Most importantly, if you saw a tasklist in a specific context and dismissed it, showing the same task list in a different context is only going to be annoying to the user.

> Do you mean to change the `is_visible` name to `is_dismissed`? Do you think it should be on a separate endpoint?

I was thinking that calling it `is_dismissed` more clearly represents the user's intent/actions. We _might_ have situations in Calypso where we ignore `is_dismissed` and that makes sense, but it would be far trickier to comment on why you're ignoring `is_visible = false`!

andres-blanco:
> I was thinking that calling it is_dismissed more clearly represents the user's intent/actions. We might have situations in Calypso where we ignore is_dismissed and that makes sense, but it would be far trickier to comment on why you're ignoring is_visible = false!

Ignoring `is_visible == false` made me chuckle 🤣 . I will update it to `is_dismissed`.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -152,6 +152,44 @@ public function is_task_list_enabled( $id ) {
...
+	}
+
+	/**
+	 * Get the task list visibility status for a site.
+	 *
+	 * @return array
+	 */
+	protected function get_task_list_dismissed_status() {
+		$launchpad_config = get_option( 'wpcom_launchpad_config', array() );
+		return isset( $launchpad_config['task_list_dismissed_status'] ) ? $launchpad_config['task_list_dismissed_status'] : array();
```
daledupreez: Just to be extra-defensive, we may want to verify that we have an array in the option first.
```suggestion
		return isset( $launchpad_config['task_list_dismissed_status'] ) && is_array( $launchpad_config['task_list_dismissed_status'] ) ? $launchpad_config['task_list_dismissed_status'] : array();
		// Given the complexity, this may be worth splitting up into an early return check.
		if ( ! isset( $launchpad_config['task_list_dismissed_status'] ) || ! is_array( $launchpad_config['task_list_dismissed_status'] ) ) {
			return array();
		}

		return $launchpad_config['task_list_dismissed_status'];
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -152,6 +152,44 @@ public function is_task_list_enabled( $id ) {
...
+
+	/**
+	 * Set wether a task list is dismissed or not for a site.
+	 *
+	 * @param string $id Task List id.
+	 * @param bool   $is_dismissed True if dismissed, false if not.
+	 */
+	public function set_task_list_dismissed( $id, $is_dismissed ) {
+		$task_list_dismissed_status        = $this->get_task_list_dismissed_status();
+		$task_list_dismissed_status[ $id ] = $is_dismissed;
```
daledupreez: I think we should only store cases where the `$is_dismissed` value is true so our option size doesn't get too large. I also think we need to cast the supplied `$is_dismissed` value to a boolean to make sure we're storing valid data.
```suggestion
		$is_dismissed                      = (bool) $is_dismissed;
		if ( $is_dismissed ) {
			$task_list_dismissed_status[ $id ] = $is_dismissed;
		} else {
			unset( $task_list_dismissed_status[ $id ] );
		}
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -152,6 +152,44 @@ public function is_task_list_enabled( $id ) {
...
+	}
+
+	/**
+	 * Set wether a task list is dismissed or not for a site.
+	 *
+	 * @param string $id Task List id.
+	 * @param bool   $is_dismissed True if dismissed, false if not.
+	 */
+	public function set_task_list_dismissed( $id, $is_dismissed ) {
+		$task_list_dismissed_status        = $this->get_task_list_dismissed_status();
```
daledupreez: We need to make sure we're getting a valid `$id` value, which suggests we may need a helper function for that.
```suggestion
	public function set_task_list_dismissed( $id, $is_dismissed ) {
		if ( empty( $id ) || ! $this->has_task_list( $id ) ) {
			return;
		}
 
		$task_list_dismissed_status  = $this->get_task_list_dismissed_status();
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -152,6 +152,44 @@ public function is_task_list_enabled( $id ) {
...
+	 * Check if a task list was dismissed by the user.
+	 *
+	 * @param string $id Task List id.
+	 * @return bool|null True if dismissed, false if not.
+	 */
+	public function is_task_list_dismissed( $id ) {
+		$task_list_dismissed_status = $this->get_task_list_dismissed_status();
+
+		// Return true if the task list is on the dismissed status array and its value is true.
+		return isset( $task_list_dismissed_status[ $id ] ) && $task_list_dismissed_status[ $id ];
```
daledupreez: Do we want an explicit `true` check for the value?
```suggestion
		return isset( $task_list_dismissed_status[ $id ] ) && true === $task_list_dismissed_status[ $id ];
```

daledupreez (commented):
Based on testing, this is working as described, but I think we do need some hardening and additional validation, as well as reducing our option footprint.

andres-blanco:
@daledupreez I've just made the suggested changes. 
I wasn't so strict in the checks before because all the usage is behind the API endpoint that's well guarded. Nonetheless this are great suggestions to future-proofs the functions, so thanks for that :)

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
andres-blanco: I've added this check just in case someone fiddles with this option value from outside the tasklist api.

daledupreez (approved):
I've retested with the hardening and I think we're good to go! 🚢 

Thanks for working through my rounds of feedback!

PR #32208, opened 2023-08-01
CI: install Woo plugin for Jetpack integration tests

samiff:
## Proposed changes:

**WIP**

The end goal is to have CI run WooCommerce/Jetpack integration tests. Those handful of tests are currently gated by the presence of `JETPACK_TEST_WOOCOMMERCE=1`. More discussion in PR comments.

Also see #32331 which is working on the unit tests and allowing local runs from Jetpack CLI.

### Other information:

## Jetpack product discussion
Internal: p1690561196958079-slack-CBG1CP4EN

No.

## Testing instructions:

TBD.


samiff:
The end goal is to have CI run WooCommerce/Jetpack integration tests. I'm guessing that the tests that depend upon `JETPACK_TEST_WOOCOMMERCE=1` haven't been utilized for some time now (e.g., some of the require [paths](https://github.com/Automattic/jetpack/blob/baafa0aeb303362fda77b978fb87cb1167edb9d9/projects/plugins/jetpack/tests/php/sync/test_class.jetpack-sync-woocommerce.php#L36) hint at a previous WooCommerce repo structure -- that one is now in `tests/legacy`).

For current and future Woo/Jetpack integration tests, I _think_ we'd want a built copy of the WooCommerce plugin, but I could be wrong about that as I'm not sure how these tests were being run previously. Things will bail early for unbuilt versions:

- https://github.com/woocommerce/woocommerce/blob/34a8273268efb26a30c0ec4da80e2c51ab8b68cf/plugins/woocommerce/woocommerce.php#L27-L29
- https://github.com/woocommerce/woocommerce/blob/4ff45b55c5058e5e83c42176ae593f145856ed00/plugins/woocommerce/src/Autoloader.php#L30-L35

If we do want a built version, that goes a bit beyond the initial scope of just git cloning things over and would need to check with team on the best way to accomplish that (i.e. differing pnpm/node versions, how long it takes to build, etc.)

zinigor:
That's a good point - building it would add a lot to the process. But can we instead checkout the current stable tag from the WordPress.org SVN? For example, currently parsing [readme.txt](https://plugins.svn.wordpress.org/woocommerce/trunk/readme.txt) gives us 7.9.0 as a stable version, we can just clone it from SVN:
https://plugins.svn.wordpress.org/woocommerce/tags/7.9.0/

zinigor:
Sorry, didn't mean to close it, missed the button.

.github/files/setup-wordpress-env.sh
```diff
@@ -129,6 +129,28 @@ for PLUGIN in projects/plugins/*/composer.json; do
 	echo "::endgroup::"
 done
 
+# Install WooCommerce plugin used for some Jetpack integration tests.
+# Todo: check the require paths being used by the Woo tests, they may need updating since the Woo repo structure changed.
+# Todo: what to check so this only runs when Jetpack will be tested?
+# Todo: what PHP_VERSION variable is considered "default"?
+if [[ "$WP_BRANCH" == "latest" && "$PHP_VERSION" == "8.2" ]]; then
```
zinigor: Actually, now that I think of it - if we're using the latest WordPress here from trunk, maybe we can also use WooCommerce from trunk. One caveat is that we'll be introducing another source of possible blockers.

zinigor: No, actually trunk is pretty stable. Judging by the [revision log](https://plugins.trac.wordpress.org/log/woocommerce/trunk), trunk mostly gets updated to RC versions, which may be even better for us - we'll test against WooCommerce code getting ready to be shipped.

samiff: > But can we instead checkout the current stable tag from the WordPress.org SVN?

@zinigor The problem with grabbing the prebuilt Woo plugin from svn/zip is that it won't contain the [/tests](https://github.com/woocommerce/woocommerce/tree/trunk/plugins/woocommerce/tests) directory. Some of the Sync integration tests use those files for [example](https://github.com/Automattic/jetpack/blob/afe1444b9cbad70f595d1d5f64999248b86e7b79/projects/plugins/jetpack/tests/php/sync/test_class.jetpack-sync-woocommerce.php#L26).

I wonder how bad of an idea it would be grab the build for the latest version, and then additionally grab the matching `/tests` and combine the two 🤔

zinigor: I think that's still a better solution than building WooCommerce :)

.github/files/setup-wordpress-env.sh
```diff
@@ -129,6 +129,28 @@ for PLUGIN in projects/plugins/*/composer.json; do
...
+	echo "::group::Installing plugin WooCommerce into WordPress"
+	cd "/tmp"
+	git clone --depth=1 https://github.com/woocommerce/woocommerce.git
+	cd "/tmp/woocommerce"
```
sdixon194: Are we already in the `tmp` folder here? Can we just `cd /woocommerce`?

.github/files/setup-wordpress-env.sh
```diff
@@ -129,6 +129,28 @@ for PLUGIN in projects/plugins/*/composer.json; do
...
+
+	# Todo: figure out how to build Woo or an alternative.
+	start_time=$(date +%s)
+	pnpm run --filter=woocommerce build # Will fail due to pnpm/node version mismatch, also will want to see how long build takes.
+	end_time=$(date +%s)
+	elapsed_time=$((end_time - start_time))
+	echo "Elapsed install time: $elapsed_time seconds"
+	# Copy the built Woo plugin with included '/tests' dir to WordPress.
+	cp -r "/tmp/woocommerce/plugins/woocommerce" "/tmp/wordpress-$WP_BRANCH/src/wp-content/plugins"
```
sdixon194: Same here, are we already in `tmp`?

.github/files/setup-wordpress-env.sh
samiff: @zinigor I've given that a try and updated the PR. So there is now the built Woo plugin directory with the `/tests` dir added manually. Things are not so "plug & play" though 😅 

You can see the recent test failures [here](https://github.com/Automattic/jetpack/actions/runs/5766034900/job/15633214216?pr=32208). Just looking at the first one there:

```
1) WP_Test_Jetpack::test_partner_codes_are_added_to_authorize_url with data set "subsidiary_code" ('subsidiary', 'jetpack_partner_subsidiary_id', 'subsidiaryId')
  Error: Call to undefined method Automattic\Jetpack\Status\Host::get_calypso_env()
  
  /tmp/wordpress-latest/src/wp-content/plugins/jetpack/class.jetpack.php:4557
  /tmp/wordpress-latest/src/wp-includes/class-wp-hook.php:308
  /tmp/wordpress-latest/src/wp-includes/plugin.php:205
  /tmp/wordpress-latest/src/wp-content/plugins/woocommerce/vendor/automattic/jetpack-connection/src/class-manager.php:1858
  /tmp/wordpress-latest/src/wp-content/plugins/jetpack/class.jetpack.php:4514
  /tmp/wordpress-latest/src/wp-content/plugins/jetpack/tests/php/general/test_class.jetpack.php:1041
```

Makes me wonder if there is going to be an issue with the Jetpack packages used by Woo?

samiff:
@chrisbliss18 This PR is currently grabbing the latest Woo build, and manually adding the [woocommerce/plugins/woocommerce/tests](https://github.com/woocommerce/woocommerce/tree/trunk/plugins/woocommerce/tests) folder.

You can see the test results under `PHP tests: PHP 8.2 WP latest`: https://github.com/Automattic/jetpack/actions/runs/5813114307/job/15759977108?pr=32208

Based on your recent comment [here](https://github.com/Automattic/jetpack/pull/32331#issuecomment-1669983008), would you be able to help with this PR as far as getting the integration tests passing? Thought I'd check with you before trying to poke at them.

chrisbliss18:
> @chrisbliss18 This PR is currently grabbing the latest Woo build, and manually adding the [woocommerce/plugins/woocommerce/tests](https://github.com/woocommerce/woocommerce/tree/trunk/plugins/woocommerce/tests) folder.
> 
> You can see the test results under `PHP tests: PHP 8.2 WP latest`: https://github.com/Automattic/jetpack/actions/runs/5813114307/job/15759977108?pr=32208
> 
> Based on your recent comment [here](https://github.com/Automattic/jetpack/pull/32331#issuecomment-1669983008), would you be able to help with this PR as far as getting the integration tests passing? Thought I'd check with you before trying to poke at them.

@samiff I know how to fix a couple of the failing tests up. As seen [here](https://github.com/Automattic/jetpack/pull/32331/files#diff-9b04bcdb7eb76565eda3fe38b35d256c629272bd89da150c47ebd8f986222267L204) in my changes, a couple of the tests just need the int changed to a string. I confirmed that a string is the correct type for those values.

As for the other failing tests, I wasn't able to fix them up quickly, so I just cheaped out to skip them as I worked on the other changes. I can dig back in and find a way to get them to work as expected however.

Would you like me to push changes to this PR to fix up these tests?

samiff:
> Would you like me to push changes to this PR to fix up these tests?

That'd be great! Like you said, we can always swing back to take a look at the other tests 👍

.github/files/setup-wordpress-env.sh
```diff
@@ -133,6 +133,43 @@ for PLUGIN in projects/plugins/*/composer.json; do
 	echo "::endgroup::"
 done
 
+# Install WooCommerce plugin used for some Jetpack integration tests.
+# Todo: check the require paths being used by the Woo tests, they may need updating since the Woo repo structure changed.
+# Todo: what to check so this only runs when Jetpack will be tested?
```
anomiex: First you'd want to update https://github.com/Automattic/jetpack/blob/38d1aa2ea475761eabb98666329f1feeaadd94ad/.github/workflows/tests.yml#L87-L89 to add
```yaml
        env:
          CHANGED: ${{ steps.changed.outputs.projects }}
```
to pass the data into the script via the environment.

Then you could do like this to check for Jetpack-the-plugin
```sh
if [[ "$WP_BRANCH" == "latest" && "$PHP_VERSION" == "8.2" ]] &&
    jq --argjson changed "$CHANGED" -ne '$changed["plugins/jetpack"] // false' > /dev/null
then
```

.github/files/setup-wordpress-env.sh
```diff
@@ -133,6 +133,43 @@ for PLUGIN in projects/plugins/*/composer.json; do
...
+# Todo: what PHP_VERSION variable is considered "default"?
```
anomiex: Currently the best way to determine that is probably something like
```sh
DEFAULT_PHP_VERSION=$(bash -c 'source .github/versions.sh; echo $PHP_VERSION')
```

.github/files/setup-wordpress-env.sh
```diff
@@ -141,4 +178,6 @@ sed -i "s/yourusernamehere/root/" wp-tests-config.php
 sed -i "s/yourpasswordhere/root/" wp-tests-config.php
 sed -i "s/localhost/127.0.0.1/" wp-tests-config.php
 
+echo "define( 'JETPACK_AUTOLOAD_DEV', true );" | tee -a wp-tests-config.php
```
anomiex:
```suggestion
# If WooCommerce is installed, be sure we get the monorepo versions rather than the versions distributed with that.
echo "define( 'JETPACK_AUTOLOAD_DEV', true );" >> wp-tests-config.php
```

projects/plugins/jetpack/tests/php/modules/google-analytics/test-class.jetpack-google-amp-analytics.php
```diff
@@ -36,7 +36,7 @@ public static function set_up_before_class() {
 
 		self::$woo_enabled = true;
 
-		$woo_tests_dir = __DIR__ . '/../../../../woocommerce/tests';
+		$woo_tests_dir = __DIR__ . '/../../../../../woocommerce/tests';
```
anomiex: I note the bootstrap file defines `JETPACK_WOOCOMMERCE_INSTALL_DIR`, maybe we should just use that?

Same at https://github.com/Automattic/jetpack/blob/38d1aa2ea475761eabb98666329f1feeaadd94ad/projects/plugins/jetpack/tests/php/sync/test_class.jetpack-sync-woocommerce.php#L26

projects/plugins/jetpack/tests/action-test-php.sh
```diff
@@ -2,6 +2,11 @@
 
 set -eo pipefail
 
+# Conditionally allow WooCommerce/Jetpack integration tests to run.
```
anomiex, line 5:
```suggestion
# Conditionally allow WooCommerce/Jetpack integration tests to run.
# Unlike WP_MULTISITE or LEGACY_FULL_SYNC, this only enables some tests, it doesn't change the behavior of any.
```
Assuming that's accurate, of course.

.github/files/setup-wordpress-env.sh
samiff: Awesome, be91bef

.github/files/setup-wordpress-env.sh
samiff: Thanks, wasn't sure if we were doing something similar elsewhere already 👍 a5075c9

projects/plugins/jetpack/tests/php/modules/google-analytics/test-class.jetpack-google-amp-analytics.php
samiff: That would seem okay to me, I know @chrisbliss18 was going to be pulling over some tests changes so I won't poke it just this moment.

.github/files/setup-wordpress-env.sh
```diff
@@ -128,6 +128,43 @@ for PLUGIN in projects/plugins/*/composer.json; do
 	echo "::endgroup::"
 done
 
+# Install WooCommerce plugin used for some Jetpack integration tests.
+#DEFAULT_PHP_VERSION=$(bash -c 'source .github/versions.sh; echo $PHP_VERSION')
+if [[ "$WITH_WOOCOMMERCE" == true ]] &&
+	jq --argjson changed "$CHANGED" -ne '$changed["plugins/jetpack"] // false' > /dev/null
```
fgiannar: Not sure about this line, leftover from previous commit. TODO: The goal is to only consider Jetpack changed and nothing else when `$WITH_WOOCOMMERCE` is `true`

projects/plugins/jetpack/tests/action-test-php.sh
```diff
@@ -2,18 +2,27 @@
...
-echo "::endgroup::"
-
-if [[ "$WP_BRANCH" == "trunk" ]]; then
-	echo "::group::Jetpack multisite tests"
-	WP_MULTISITE=1 phpunit -c tests/php.multisite.xml
+# Conditionally allow WooCommerce/Jetpack integration tests to run.
+# Unlike WP_MULTISITE or LEGACY_FULL_SYNC, this only enables some tests, it doesn't change the behavior of any.
+if [[ "$WITH_WOOCOMMERCE" == true ]]; then
+	export JETPACK_TEST_WOOCOMMERCE=1
+	echo "::group::Jetpack tests"
```
fgiannar:
```suggestion
	echo "::group::Jetpack WooCommerce tests"
```

projects/plugins/jetpack/tests/action-test-php.sh
```diff
@@ -2,18 +2,27 @@
...
-echo "::group::Jetpack tests"
-phpunit
-echo "::endgroup::"
-
-if [[ "$WP_BRANCH" == "trunk" ]]; then
-	echo "::group::Jetpack multisite tests"
-	WP_MULTISITE=1 phpunit -c tests/php.multisite.xml
+# Conditionally allow WooCommerce/Jetpack integration tests to run.
+# Unlike WP_MULTISITE or LEGACY_FULL_SYNC, this only enables some tests, it doesn't change the behavior of any.
+if [[ "$WITH_WOOCOMMERCE" == true ]]; then
```
fgiannar, line 6: Is this enough here? Maybe we should check for the dir too, as there's a chance `$WITH_WOOCOMMERCE` to be truthful but an error occurred while downloading Woo

.github/files/generate-ci-matrix.php
```diff
@@ -49,6 +49,9 @@
 
 	// {string} A valid artifact name for any generated artifacts. If not given, will be derived from the name.
 	'artifact'            => null,
+
+	// {bool} Whether to install WooCommerce.
+	'with-woocommerce'    => false,
```
fgiannar, line 54: Consider changing this to a string instead with the WooCommerce version to check out?

anomiex: If that's something like "latest" and "trunk", that could be good. Fetching what the actual latest version is in here might be a bit much though.

.github/files/setup-wordpress-env.sh
```diff
@@ -129,10 +129,10 @@ for PLUGIN in projects/plugins/*/composer.json; do
 done
 
 # Install WooCommerce plugin used for some Jetpack integration tests.
-# Todo: check the require paths being used by the Woo tests, they may need updating since the Woo repo structure changed.
-# Todo: what to check so this only runs when Jetpack will be tested?
-# Todo: what PHP_VERSION variable is considered "default"?
-if [[ "$WP_BRANCH" == "latest" && "$PHP_VERSION" == "8.2" ]]; then
+#DEFAULT_PHP_VERSION=$(bash -c 'source .github/versions.sh; echo $PHP_VERSION')
```
anomiex: No need for this anymore.
```suggestion
```

.github/files/setup-wordpress-env.sh
anomiex: This line is checking _if_ Jetpack is considered changed, and only installing WooCommerce if it is.

We can probably drop it in favor of having the workflow skip the step entirely when no plugins wind up being changed.

.github/files/setup-wordpress-env.sh
```diff
@@ -128,6 +128,43 @@ for PLUGIN in projects/plugins/*/composer.json; do
...
+if [[ "$WITH_WOOCOMMERCE" == true ]] &&
+	jq --argjson changed "$CHANGED" -ne '$changed["plugins/jetpack"] // false' > /dev/null
+then
+	echo "::group::Installing plugin WooCommerce into WordPress"
+
+	WOO_REPO_URL="https://github.com/woocommerce/woocommerce"
+	WOO_GH_API_URL="https://api.github.com/repos/woocommerce/woocommerce/releases/latest"
+	WOO_LATEST_TAG=""
+	WOO_DL_URL=""
```
anomiex: No need for these, they get unconditionally reset 3 lines down.
```suggestion
```

.github/files/setup-wordpress-env.sh
```diff
@@ -128,6 +128,43 @@ for PLUGIN in projects/plugins/*/composer.json; do
...
+
+	RESPONSE=$(curl -s "$WOO_GH_API_URL")
```
anomiex: Let's add some more curl options:
* `-S` to have it print a message on error
* `-L` to have it follow redirects
* `--fail` to have it return a failure status on error, instead of letting the `jq` on the next line choke on a probably-non-JSON response.

```suggestion
	RESPONSE=$(curl -sSL --fail "$WOO_GH_API_URL")
```

.github/files/setup-wordpress-env.sh
```diff
@@ -128,6 +128,43 @@ for PLUGIN in projects/plugins/*/composer.json; do
...
+	WOO_LATEST_TAG=$(jq -r ".tag_name" <<< "$RESPONSE")
+	WOO_DL_URL=$(jq -r ".assets[0].browser_download_url" <<< "$RESPONSE")
```
anomiex: It doesn't make any real difference, but in general it's more useful to use single quotes here
```suggestion
	WOO_LATEST_TAG=$(jq -r '.tag_name' <<< "$RESPONSE")
	WOO_DL_URL=$(jq -r '.assets[0].browser_download_url' <<< "$RESPONSE")
```

.github/files/setup-wordpress-env.sh
```diff
@@ -128,6 +128,43 @@ for PLUGIN in projects/plugins/*/composer.json; do
...
+
+	if [[ -n "$WOO_LATEST_TAG" && -n "$WOO_DL_URL" ]]; then
+		cd "/tmp"
+		echo "Fetching latest WooCommerce tag: $WOO_LATEST_TAG"
+
+		# Download the built Woo plugin.
+		curl -s -L "$WOO_DL_URL" -o "woocommerce.zip"
```
anomiex:
```suggestion
		curl -sS -L --fail "$WOO_DL_URL" -o "woocommerce.zip"
```

projects/plugins/jetpack/tests/action-test-php.sh
```diff
@@ -2,18 +2,27 @@
...
 
-echo "::group::Jetpack tests"
-phpunit
-echo "::endgroup::"
-
-if [[ "$WP_BRANCH" == "trunk" ]]; then
-	echo "::group::Jetpack multisite tests"
-	WP_MULTISITE=1 phpunit -c tests/php.multisite.xml
+# Conditionally allow WooCommerce/Jetpack integration tests to run.
+# Unlike WP_MULTISITE or LEGACY_FULL_SYNC, this only enables some tests, it doesn't change the behavior of any.
```
anomiex: Not accurate after all
```suggestion
```

projects/plugins/jetpack/tests/action-test-php.sh
anomiex: IMO this is fine. If an error occurred while downloading Woo, it should have errored out then.

projects/plugins/jetpack/tests/action-test-php.sh
anomiex: :+1:

projects/plugins/jetpack/tests/action-test-php.sh
```diff
@@ -2,18 +2,27 @@
...
+	export JETPACK_TEST_WOOCOMMERCE=1
+	echo "::group::Jetpack tests"
+	phpunit --group=woocommerce
 	echo "::endgroup::"
-fi
-
-if [[ "$WP_BRANCH" == "latest" && "$PHP_VERSION" == "7.0" ]]; then
-	echo "::group::Jetpack Legacy Full Sync tests"
-	LEGACY_FULL_SYNC=1 phpunit --group=legacy-full-sync
+else
```
anomiex: Personally, I might just do like
```
	echo "::endgroup::"
        exit 0
fi

```
and not have to re-indent the rest of the file. But this works.

projects/plugins/jetpack/tests/php/bootstrap.php
```diff
@@ -101,6 +101,49 @@ function _manually_load_plugin() {
 }
 
 function _manually_install_woocommerce() {
+	$woo_tests_dir = JETPACK_WOOCOMMERCE_INSTALL_DIR . '/tests';
+	// This is taken from WooCommerce's bootstrap.php file
```
anomiex: This is a lot of stuff to be requiring. Any time one of these files gets removed, or another has to be added, we'd have to update this list. Is there a better way?

.github/workflows/tests.yml
```diff
@@ -85,6 +86,8 @@ jobs:
           echo "any-plugins=${ANY_PLUGINS}" >> "$GITHUB_OUTPUT"
```
anomiex, line 93: Just above here is probably where we want to limit to just Jetpack when testing Woo. Perhaps something like this
```
          CHANGED="$(EXTRA=test .github/files/list-changed-projects.sh)"

          # Only test certain plugins in combination with WC
          if [[ "$WITH_WOOCOMMERCE" == true ]]; then
            CHANGED=$( jq 'with_entries( select( .key == "plugins/jetpack" ) )' <<<"$CHANGED" )
          fi

          ANY_PLUGINS="$(jq --argjson changed "$CHANGED" -n '$changed | with_entries( select( .key | startswith( "plugins/" ) ) ) | any')"
          echo "projects=${CHANGED}" >> "$GITHUB_OUTPUT"
          echo "any-plugins=${ANY_PLUGINS}" >> "$GITHUB_OUTPUT"
```
(unfortunately GitHub won't let me make that a suggestion)

projects/plugins/jetpack/tests/action-test-php.sh
fgiannar: Done

.github/workflows/tests.yml
fgiannar: Awesome, thanks Brad! Done

projects/plugins/jetpack/tests/php/bootstrap.php
fgiannar: I'm trying to find one 😓 

Currently I'm experimenting with loading WooCommerce's `bootstrap.php` but there are various errors which I'm not sure how we should handle.

fgiannar: I ended up keeping the existing approach (which I agree is not ideal to be clear) plus a few hacks to allow those tests to successfully run 😒

.github/files/generate-ci-matrix.php
fgiannar: I guess we can keep it as is for now, as it's simpler like this and iterate later if we decide we need more versions for some reason.

.github/workflows/tests.yml
fgiannar: I get some errors with this syntax currently. I plan to debug and fix after making the unit tests pass

projects/plugins/jetpack/tests/php/modules/google-analytics/test-class.jetpack-google-amp-analytics.php
fgiannar: I believe this whole class is outdated. `WC()->session->get( 'wc_ga_events' )` returns null and I [couldn't find any references ](https://github.com/search?q=repo%3AAutomattic%2Fwoocommerce+wc_ga_events&type=code)to `wc_ga_events` in the Woo repo 🤷

.github/workflows/tests.yml
anomiex: I had one typo in there (needs to compare with true, d'oh) that was making the added bit run when it shouldn't have. Edited the comment.

Also in 19e16aaf0c577fb0115a486f57e000b99bd3cea0 you added an extraneous `--argjson` which is part of why everything broke.

projects/plugins/jetpack/tests/php/sync/test_class.jetpack-sync-woocommerce.php
```diff
@@ -2,63 +2,21 @@
...
+ *
+ * @group woocommerce
  */
 class WP_Test_Jetpack_Sync_WooCommerce extends WP_Test_Jetpack_Sync_Base {
+	use WooCommerceTestTrait;
+
 	protected $post;
 	protected $callable_module;
 	protected $full_sync;
 	public static $woo_enabled;
```
anomiex: We might want to move this `$woo_enabled` field declaration into the trait too, so every class using it doesn't have to declare it itself.

.github/workflows/tests.yml
```diff
@@ -80,11 +81,19 @@ jobs:
         id: changed
         run: |
           CHANGED="$(EXTRA=test .github/files/list-changed-projects.sh)"
+
+          # Only test certain plugins in combination with WC
+          if [[ "$WITH_WOOCOMMERCE" == true ]]; then
+            CHANGED=$( jq 'with_entries( select( .key == "plugins/jetpack" ) )' <<<"$CHANGED" )
```
anomiex: This should fix the error:
```suggestion
            CHANGED=$( jq -c 'with_entries( select( .key == "plugins/jetpack" ) )' <<<"$CHANGED" )
```
(the problem was that the command we're using on line 91 can't handle if the variable has newlines)

.github/workflows/tests.yml
```diff
@@ -80,11 +81,19 @@ jobs:
         id: changed
         run: |
           CHANGED="$(EXTRA=test .github/files/list-changed-projects.sh)"
+
+          # Only test certain plugins in combination with WC
+          if [[ "$WITH_WOOCOMMERCE" == true ]]; then
+            CHANGED=$( jq -c 'with_entries( select( .key == "plugins/jetpack" ) )' <<<"$CHANGED" )
```
anomiex, line 88: Just because I like the logs to be clear about what's happening,
```suggestion
            echo "Testing with WooCommerce, filtering for projects that have WooCommerce-specific tests."
            CHANGED=$( jq -c 'with_entries( select( .key == "plugins/jetpack" ) )' <<<"$CHANGED" )
```

projects/plugins/jetpack/tests/php/modules/google-analytics/test-class.jetpack-google-amp-analytics.php
```diff
@@ -4,45 +4,29 @@
...
  * Class WP_Test_Jetpack_Google_AMP_Analytics
+ *
+ * @group woocommerce
  */
 class WP_Test_Jetpack_Google_AMP_Analytics extends WP_UnitTestCase {
-
 	/**
-	 * A WooCommerce product.
+	 * Using the WooCommerceTestTrait to include WooCommerce related dependencies for the unit tests.
+	 * Noting that we'd need to add coocommerce as group in otder for the tests
```
fgiannar:
```suggestion
	 * Noting that we'd need to add woocommerce as group in order for the tests
```

projects/plugins/jetpack/tests/php/modules/google-analytics/test-class.jetpack-google-amp-analytics.php
```diff
@@ -4,45 +4,29 @@
  *
  * @package automattic/jetpack
  */
-
+require_once __DIR__ . '/../../trait-woo-tests.php';
 /**
  * Class WP_Test_Jetpack_Google_AMP_Analytics
+ *
+ * @group woocommerce
```
fgiannar:
```suggestion
```

anomiex (commented):
Looks good to me from a monorepo perspective, the right test suites are being run in the right checks. I'll leave final approval to you @fgiannar once you're happy with the set of tests being run in the new check and everything. 🙂

projects/plugins/jetpack/tests/php/sync/test_class.jetpack-sync-woocommerce.php
```diff
@@ -274,10 +235,10 @@ public function test_full_sync_order_items() {
 		$order2_item  = reset( $order2_items ); // first item from order2
 		wc_update_order_item_meta( $order2_item->get_id(), '_line_subtotal', 20 );
 
-		$this->full_sync->start();
+		$this->full_sync->start( array( 'woocommerce' => true ) );
 		$this->sender->do_full_sync();
 
-		$full_sync_order_items = $this->server_event_storage->get_most_recent_event( 'jetpack_full_sync_woocommerce_order_items' );
+		$full_sync_order_items = $this->server_event_storage->get_all_events( 'jetpack_full_sync_woocommerce' );
```
fgiannar:
```suggestion
		$full_sync_order_items = $this->server_event_storage-> get_most_recent_event( 'jetpack_full_sync_woocommerce_order_items' );
```

projects/plugins/jetpack/tests/php/sync/test_class.jetpack-sync-woocommerce.php
```diff
@@ -274,10 +235,10 @@ public function test_full_sync_order_items() {
 		$order2_item  = reset( $order2_items ); // first item from order2
 		wc_update_order_item_meta( $order2_item->get_id(), '_line_subtotal', 20 );
 
-		$this->full_sync->start();
+		$this->full_sync->start( array( 'woocommerce' => true ) );
 		$this->sender->do_full_sync();
 
-		$full_sync_order_items = $this->server_event_storage->get_most_recent_event( 'jetpack_full_sync_woocommerce_order_items' );
+		$full_sync_order_items = $this->server_event_storage-> get_most_recent_event( 'jetpack_full_sync_woocommerce_order_items' );
```
fgiannar:
```suggestion
		$full_sync_order_items = $this->server_event_storage->get_most_recent_event( 'jetpack_full_sync_woocommerce_order_items' );
```

fgiannar (approved):
LGTM

PR #32218, opened 2023-08-02
Hovercards: switch to new URL

johngodley:
Updates the Gravatar hovercard URL to use the updated location. Some i18n strings are also supplied, which are used by the new hovercard library.

The option description has been updated.

image

## Proposed changes:
* Update a URL
* Add some i18n strings
* Change option description

### Other information:

No.

## Testing instructions:
* Enable hovercards from settings > discussion
* Visit a post with comments, hover over an avatar, and confirm a popup is shown

### DIFF
--- a/projects/plugins/jetpack/modules/gravatar-hovercards.php
+++ b/projects/plugins/jetpack/modules/gravatar-hovercards.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * Module Name: Gravatar Hovercards
- * Module Description: Enable pop-up business cards over commenters’ Gravatars.
+ * Module Description: Enable profile hovercards on comment avatars.
  * Sort Order: 11
  * Recommendation Order: 13
  * First Introduced: 1.1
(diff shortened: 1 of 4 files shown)


projects/plugins/jetpack/modules/gravatar-hovercards.php
```diff
@@ -274,6 +274,16 @@ function grofiles_attach_cards() {
...
 		}
 		wp_localize_script( 'wpgroho', 'WPGroHo', compact( 'my_hash' ) );
+		wp_localize_script(
+			'grofiles-cards',
+			'GravatarHovercardsi18n',
+			array(
+				'Edit your profile' => __( 'Edit your profile', 'jetpack' ),
+				'View profile'      => __( 'View profile', 'jetpack' ),
+				'Sorry, we are unable to load this Gravatar profile.' => __( 'Sorry, we are unable to load this Gravatar profile.', 'jetpack' ),
+				'Sorry, we are unable to load this Gravatar profile. Please check your internet connection.' => __( 'Sorry, we are unable to load this Gravatar profile. Please check your internet connection.', 'jetpack' ),
```
jeherve, line 284: Shouldn't the keys be JavaScript variables instead?

johngodley: No, it's a map of English text to translation.

PR #32219, opened 2023-08-02
Boost: Integrate performance history graph

haqadn:
## Proposed changes:

* Add performance history graph to a panel
* Cleanup and refactor the svelte component

### Other information:

## Jetpack product discussion
None

None

## Testing instructions:
Collapse/expand the panel and refresh the page to ensure the status sticks across browser reloads

### DIFF
--- a/projects/js-packages/boost-score-api/package.json
+++ b/projects/js-packages/boost-score-api/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "@automattic/jetpack-boost-score-api",
-	"version": "0.1.5",
+	"version": "0.1.6-alpha",
 	"description": "A package to get the Jetpack Boost score of a site",
 	"homepage": "https://github.com/Automattic/jetpack/tree/HEAD/projects/js-packages/boost-score-api/#readme",
 	"bugs": {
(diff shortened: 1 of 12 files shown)


sejas (approved):
@haqadn , Awesome! Thanks for the refactor 🙏  and adding the toggle interaction.

Working as expected.

Screen Capture on 2023-08-02 at 16-37-04

thingalon:
Nice; I can see you're making the integration much more Boost-ey. 👍

This is not a blocker to merging this into the parent branch -- but I saw an error when I tested it out:

image

It turned out that error was caused by me having the Boost Developer plugin enabled. We'll need to take care of that side of things. :-)

haqadn:
> Nice; I can see you're making the integration much more Boost-ey. 👍
> 
> This is not a blocker to merging this into the parent branch -- but I saw an error when I tested it out:
> 
> image
> 
> It turned out that error was caused by me having the Boost Developer plugin enabled. We'll need to take care of that side of things. :-)

Shouldn't we leave the error in place to have some visual feedback for the user?

thingalon:
> Shouldn't we leave the error in place to have some visual feedback for the user?

If there is a problem we absolutely should show an error. In this case I would prefer a more meaningful error - but that isn't what I was getting at.

I meant that we should update the Boost Developer plugin to make it possible to use / test this graph feature with our local envs. :)

PR #32233, opened 2023-08-02
Launchpad: Add path for newsletter tasks

valterlorran:
## Proposed changes:

After we changed how to redirect the user when clicking on the tasks, some tasks were missing the `get_calypso_path` prop introduced on https://github.com/Automattic/jetpack/pull/32177.

* Adds the `get_calypso_path` props with the appropriate path to the "Write 3 posts" and "Enable subscribers modal" tasks

### Other information:

#### Testing on Simple
* Apply this patch to your sandbox using the command in the comment below
* Create a new Newsletter site and go through the flow until you land on the Customer Home page
* Click on the task "Write 3 posts". You should be redirected to the `/post/:siteSlug`
* Click on the task "Enable subscribers modal". You should be redirected to `/settings/reading/:siteSlug#newsletter-settings`

#### Testing on Atomic
* Create a newsletter site from `/setup/newsletter/intro`; you'll need to publish at least one test post to "launch" the site.
* Upgrade the site to a Business plan
* Install and activate the Jetpack Beta plugin (this should move the site to the Atomic infrastructure)
* Go to Manage under the WordPress.com Features part of the beta plugin
* Activate the `add/path-for-newsletter-tasks` feature branch
* SSH into your Atomic site and add the `define( 'JETPACK_MU_WPCOM_LOAD_VIA_BETA_PLUGIN', true );` constant to your wp-config.php
* While you're in there, add the filter to enable the subscriptions modal: `add_filter( 'jetpack_subscriptions_modal_enabled', '__return_true', 20 );` I put it in `wp-content/mu-plugins/index.php`
* Go to `/home/siteSlug` to see the Newsletter task list.
* You should see the "Enable subscribers modal" task, and clicking on it should take you to `https://YOUR-SITE/wp-admin/admin.php?page=jetpack#/discussion`

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
@@ -424,12 +424,32 @@ class Launchpad_Task_Lists {
 			return null;
 		}
 
-		// Require that the string start with `/`, but don't allow `//`.
-		if ( '/' !== substr( $calypso_path, 0, 1 ) || '/' === substr( $calypso_path, 1, 1 ) ) {
+		if ( ! $this->is_valid_url_or_path( $calypso_path ) ) {
 			return null;
 		}
 
 		return $calypso_path;
+	}
+
+	/**
+	 * Checks if a string is a valid URL or path.
+	 *
+	 * @param string $input The string to check.
+	 * @return boolean
+	 */
+	private function is_valid_url_or_path( $input ) {
+		// Validate as a full URL
+		$url_info = wp_parse_url( $input );
+		if ( $url_info !== false && isset( $url_info['scheme'] ) ) {
+			return true;
+		}
+
+		// Require that the string start with a slash, but not two slashes
+		if ( '/' === substr( $input, 0, 1 ) && '/' !== substr( $input, 1, 1 ) ) {
+			return true;
+		}
+
+		return false;
 	}
 
 	/**
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
@@ -424,7 +424,7 @@ class Launchpad_Task_Lists {
 			return null;
 		}
 
-		if ( ! $this->is_valid_url_or_path( $calypso_path ) ) {
+		if ( ! $this->is_valid_admin_url_or_absolute_path( $calypso_path ) ) {
 			return null;
 		}
 
@@ -432,12 +432,12 @@ class Launchpad_Task_Lists {
 	}
 
 	/**
-	 * Checks if a string is a valid URL or path.
+	 * Checks if a string is a valid admin URL or an absolute path.
 	 *
 	 * @param string $input The string to check.
 	 * @return boolean
 	 */
-	private function is_valid_url_or_path( $input ) {
+	private function is_valid_admin_url_or_absolute_path( $input ) {
 		// Checks if the string is URL starting with the admin URL
 		if ( strpos( $input, admin_url() ) === 0 ) {
 			return true;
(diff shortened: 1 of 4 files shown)


projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -343,13 +343,22 @@ function wpcom_launchpad_get_task_definitions() {
 			},
 			'is_complete_callback' => 'wpcom_is_task_option_completed',
 			'is_visible_callback'  => 'wpcom_is_enable_subscribers_modal_visible',
+			'get_calypso_path'     => function ( $task, $default, $data ) {
+				if ( ( new Automattic\Jetpack\Status\Host() )->is_atomic_platform() ) {
+					return 'https://' . $data['site_slug_encoded'] . '/wp-admin/admin.php?page=jetpack#/discussion';
```
daledupreez: I think we should use `admin_url()` for this use case rather than building it dynamically. ([docs](https://developer.wordpress.org/reference/functions/admin_url/))
```suggestion
					return admin_url( '/admin.php?page=jetpack#/discussion' );
```

valterlorran: Thanks! I fixed it. However, one of the rules you added prevents from having a full URL on the `get_calypso_path` prop. Do you think we can remove this rule?

daledupreez: Hmmm... Maybe we can allow URLs that start with `admin_url()`?

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -424,14 +424,33 @@ private function load_calypso_path( $task ) {
...
 		return $calypso_path;
 	}
 
+	/**
+	 * Checks if a string is a valid URL or path.
+	 *
+	 * @param string $input The string to check.
+	 * @return boolean
+	 */
+	private function is_valid_url_or_path( $input ) {
```
daledupreez: Minor thought: to be extra helpful to our future selves, it may be worth using an even more explicit function name to indicate that we allow admin URLs and absolute paths?
```suggestion
	/**
	 * Checks if a string is a valid admin URL or an absolute path.
	 *
	 * @param string $input The string to check.
	 * @return boolean
	 */
	private function is_valid_admin_url_or_absolute_path( $input ) {
```

projects/packages/jetpack-mu-wpcom/tests/php/features/launchpad/class-launchpad-task-lists-test.php
```diff
@@ -434,9 +434,9 @@ function ( $extended_task_definitions ) use ( $task ) {
 	 */
 	public function provide_get_calypso_path_validation_test_cases() {
 		return array(
-			'Full URL is rejected'                       => array(
-				'https://example.com/invalid-full-url',
-				null,
+			'Full URL should be valid'                   => array(
+				'http://example.org/wp-admin/admin.php?page=jetpack#/discussion',
+				'http://example.org/wp-admin/admin.php?page=jetpack#/discussion',
```
daledupreez: I think we should keep the existing test and add a separate, better-named test for admin URLs.
```suggestion
			'External absolute URL should be invalid'                   => array(
				'https://example.com/invalid-full-url',
				null,
			),
			'Admin URL should be valid'                   => array(
				'http://example.org/wp-admin/admin.php?page=jetpack#/discussion',
				'http://example.org/wp-admin/admin.php?page=jetpack#/discussion',
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -343,6 +343,12 @@ function wpcom_launchpad_get_task_definitions() {
 			},
 			'is_complete_callback' => 'wpcom_is_task_option_completed',
 			'is_visible_callback'  => 'wpcom_is_enable_subscribers_modal_visible',
+			'get_calypso_path'     => function ( $task, $default, $data ) {
+				if ( ( new Automattic\Jetpack\Status\Host() )->is_atomic_platform() ) {
+					return admin_url( '/admin.php?page=jetpack#/discussion' );
```
daledupreez, line 348: Nit: I think it's more common to drop the leading `/`, but this still seems to work.
```suggestion
					return admin_url( 'admin.php?page=jetpack#/discussion' );
```

daledupreez (approved):
LGTM! I also ensured that both branches of the site subscribers task were working via manually flipping that condition.

PR #32245, opened 2023-08-02
Launchpad: Add the "Add your about page" task

markbiek:
Fixes #79795

## Proposed changes:

* This PR adds the "Add your About page" task and its completion logic. This task is part of the Newsletter task lists and should be marked as complete when the user creates an "About" page. To achieve it, we had to add a new post meta to store the category selected by the user when they access `/page/:siteSlug`.
See: https://github.com/Automattic/wp-calypso/pull/80186

### Other information:

## Testing instructions:

* Apply this patch to your sandbox.
* Follow the instructions on https://github.com/Automattic/wp-calypso/pull/80186
* The task "Add your About page" should be marked as complete


valterlorran:
Mark had the following TODO added:
```php
function wpcom_is_update_about_page_task_visible() {
	// TODO - Need to make this visible in the case that add_about_page has happened. We need to get that page id somehow.

```
However, I'm not sure if we want to show the "Update the about page" if the user completed the task "Add your about page" task.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -1227,12 +1237,65 @@ function wpcom_launchpad_has_translation( $string, $domain = 'jetpack-mu-wpcom'
 	return $translated_string !== $string;
 }
 
+/**
+ * Determine `add_new_page` task visibility. The task is visible if there is no 'About' page on the site.
+ *
+ * @return bool True if we should show the task, false otherwise.
+ */
+function wpcom_launchpad_is_add_about_page_visible() {
+	return ! wpcom_is_update_about_page_task_visible();
```
daledupreez: Given that the `_wp_layout_category` meta field is getting added in the Editor Toolkit plugin, we should probably add a check here for the existence/registration of that meta field.
```suggestion
	return ! wpcom_is_update_about_page_task_visible() && registered_meta_key_exists( 'post', '_wp_layout_category', 'page' );
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -1227,12 +1237,65 @@ function wpcom_launchpad_has_translation( $string, $domain = 'jetpack-mu-wpcom'
...
+	// If the page is the previously located About page, ignore it.
+	$about_page_id = wpcom_get_site_about_page_id();
+	if ( null !== $about_page_id && $post->ID === $about_page_id ) {
+		return;
+	}
+
+	$is_about_page = 'about' === get_post_meta( $post->ID, '_wp_layout_category', true );
+	if ( ! $is_about_page ) {
+		return;
+	}
```
daledupreez: I think we need to confirm the data model/structure and naming for the layout category meta field before we can ship this PR. I _think_ we want to use a `_wpcom_` prefix, and I also think we should store an array of category values.

Separately, we also need to confirm that the post_meta key is registered -- it's possible that the Editor Toolkit plugin is disabled on some Atomic sites.
```suggestion
	// Don't do anything if we don't have the page category post_meta field registered.
	if ( ! registered_meta_key_exists( 'post', '_wpcom_layout_category', 'page' ) ) {
		return;
	}

	// Don't do anything if the task is already complete.
	if ( wpcom_is_task_option_completed( array( 'id' => 'add_about_page' ) ) ) {
		return;
	}

	// If the page is the previously located About page, ignore it.
	$about_page_id = wpcom_get_site_about_page_id();
	if ( null !== $about_page_id && $post->ID === $about_page_id ) {
		return;
	}

	$stored_page_categories = get_post_meta( $post->ID, '_wpcom_page_layout_category', true );
	if ( ! is_array( $stored_page_categories ) || ! in_array( 'about', $stored_page_categories, true ) ) {
		return;
	}
```

andres-blanco:
> Mark had the following TODO added:
> 
> ```
> function wpcom_is_update_about_page_task_visible() {
> 	// TODO - Need to make this visible in the case that add_about_page has happened. We need to get that page id somehow.
> ```
> 
> However, I'm not sure if we want to show the "Update the about page" if the user completed the task "Add your about page" task.

did you found out what this was about?

valterlorran:
@andres-blanco, I talked with @markbiek. He added the TODO because he was wondering if we should show the "Update the about me task" when you complete the "Add your about page" task. But we agreed that the "Add your about page" task is enough to get the user started with the "About" page.

andres-blanco (approved):
I've followed the test instructions and works as described.

PR #32246, opened 2023-08-02
AI Assistant: Fix AI Control toggle behavior

### LINKED ISSUE #32243: AI Extension: toggle AiExtensionBar doesn't work
Clicking on the Toolbar button doesn't hide the assistant bar

https://user-images.githubusercontent.com/77539/257899809-d5d92f94-d866-40bf-92d1-fcc3104f8a9f.mov

lhkowalski:
Fixes #32243.

The bug was happening because the click to toggle the component off was generating two events, and the handling of the two events was generating conflicting results:

* when you clicked the toolbar button to close the popover, the popover would lose it's focus and fire the `close` event, hidding the component
* after it, we would handle the toolbar button click, changing the component back to visible again

## Proposed changes:

* Stop handling the `close` event from the popover that contains the AI Control, so it stops hiding when it loses focus
* Let the extension toolbar button be in charge of toggling the popover, as the only way to show and hide it
* Do not change the behavior when the whole Jetpack Form block is unselected; in this cases we still hide the popover, as before

### Other information:

## Testing instructions:

* Go to block editor
* Insert a Jetpack Form block
* Click the AI Assistant extension button to show the AI Control component:

Screen Shot 2023-08-02 at 14 49 32

* Click the button again and observe the behavior:
   * before this change, this would not hide the component (see issue report)
   * after this change, the AI Control will be hidden
* Test multiple clicks to toggle the component:
   * it should show/hide it as expected
* With the AI Control visible, click on the Jetpack Form block itself:
   * before this change, this would hide the popover
   * after this change, the popover will still be there; it will only disapear when the toolbar extension button is clicked again
   * this was part of the solution for the bug, and I think it actually improves the experience while using the extension
* With the component visible, click outside the Jetpack Form block so it gets unselected
   * before this change, this would hide the popover
   * after this change, this will also hide the popover (no change on the behavior)

### DIFF
--- a/projects/plugins/jetpack/changelog/update-jetpack-form-ai-control-toggle
+++ b/projects/plugins/jetpack/changelog/update-jetpack-form-ai-control-toggle
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+AI Extension: Stop handling popover close event to prevent the AI Control from hiding when it loses focus.
--- a/projects/plugins/jetpack/changelog/update-jetpack-form-ai-control-toggle
+++ b/projects/plugins/jetpack/changelog/update-jetpack-form-ai-control-toggle
@@ -1,4 +1,4 @@
 Significance: patch
-Type: fixed
+Type: bugfix
 
 AI Extension: Stop handling popover close event to prevent the AI Control from hiding when it loses focus.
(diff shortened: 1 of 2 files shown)


dhasilva (approved):
Nice!

PR #32248, opened 2023-08-02
Chatbot package: Add initial functionality.

dereksmart:
## Proposed changes:

Includes base functionality for rendering the odie widget (chatbot) from wpcom and initializes an actual chatbot. 

## Jetpack product discussion

TBD

No

## Testing instructions:

This branch will be kept up to date on top of https://github.com/Automattic/jetpack/pull/32264 so it can be tested with the starter plugin.


jeherve (commented):
Do you think we could merge a first PR with just the package skeleton, and then you can work on each class in follow-up PRs? It would make reviewing PRs easier imo.

dereksmart:
> Do you think we could merge a first PR with just the package skeleton, and then you can work on each class in follow-up PRs? It would make reviewing PRs easier imo.

💯 @jeherve let's go for that here - #32260

PR #32253, opened 2023-08-03
[Boost] Remove dependence on Critical CSS nonce

thingalon:
We've had a number of users report issues generating Critical CSS. It appears their Critical CSS nonce is undefined, causing errors.

This works around the problem - but does not get down to the bottom of how users are in this state. We are investigating further.

## Proposed changes:
* Remove Critical CSS regen nonce requirement - it isn't actually useful for security.

### Other information:

## Jetpack product discussion

n/a

no
## Testing instructions:

* Test critical css regeneration


haqadn:
Abandoned. Replacement #32502

PR #32257, opened 2023-08-03
Use shared component for editor panel icons

simison:
Use a shared component as Jetpack icon for block editor panels.

Pretty minor but seemed like repeating `showText={ false } height={ 16 } logoColor="#1E1E1E"` everywhere was a bit unnecessary.

Screenshot 2023-08-03 at 12 51 53

### Other information:

## Testing instructions:

* Test all affected panels and ensure icons render fine. They seem all to be in post editor's post settings, pre-publish, or post-publish panels.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -920,6 +920,9 @@ importers:
       '@automattic/jetpack-analytics':
         specifier: workspace:*
         version: link:../analytics
+      '@automattic/jetpack-components':
+        specifier: workspace:*
+        version: link:../components
       '@automattic/jetpack-connection':
         specifier: workspace:*
         version: link:../connection
--- a/projects/js-packages/shared-extension-utils/changelog/update-jetpack-editor-panel-icon
+++ b/projects/js-packages/shared-extension-utils/changelog/update-jetpack-editor-panel-icon
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Add shared block editor logo component.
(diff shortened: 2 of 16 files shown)


jeherve (changes requested):
This looks good to me, it makes sense. I wonder if it would be useful to place `JetpackEditorPanelLogo` in `@automattic/jetpack-shared-extension-utils` instead, so we can use it in standalone plugins as well? Here is what would be a good candidate, for example:
https://github.com/Automattic/jetpack/blob/199afea61f7bc40e15951546faf2d4721622c4ee/projects/plugins/social/src/js/editor.js#L98

simison:
Thanks! Good idea, implemented in https://github.com/Automattic/jetpack/pull/32257/commits/42f005d892bc5c76e27aa157fd69d3c74b67edbd

jeherve (changes requested):
It tests well. It does seem like we've lost some margin styling in the process:

```css
.components-panel__icon {
    color: #757575;
    margin: -2px 0 -2px 6px
}
```

**Before**

Screenshot 2023-08-03 at 16 23 33
image

**After**

Screenshot 2023-08-03 at 16 23 38
image

I think it should be something worth fixing.

simison:
> It tests well. It does seem like we've lost some margin styling in the process:

Huh; I don't understand what happened? The markup should stay exactly the same, we're passing just SVG to a prop that expects an icon and should style its spacing on its own... so why do we need to pass this thing a left margin for it to work?

Did in https://github.com/Automattic/jetpack/pull/32257/commits/cc6a62cec183b2286dbcc624ef65aa8fd965da4f however.

lezama (approved):

lezama:
> Did in https://github.com/Automattic/jetpack/commit/cc6a62cec183b2286dbcc624ef65aa8fd965da4f however.

Looks good now!

Screenshot 2023-08-07 at 09 24 39

simison:
I'm just bothered that I don't understand why that worked. 😅

jeherve (approved):
This tests well for me now! 🚢

PR #32263, opened 2023-08-03
Newsletters: Add Paywall block settings

lezama:
Add Newsletter settings panel to Paywall Block settings

Screenshot 2023-08-04 at 11 49 34

### Other information:

## Testing instructions:

* `define( 'JETPACK_BLOCKS_VARIATION', 'beta' );`
* Add a paywall block and make sure Newsletter visibility can be managed from block settings.


projects/plugins/jetpack/extensions/shared/memberships/edit.js
```diff
@@ -22,24 +20,6 @@ export const useAccessLevel = postType => {
 	return accessLevel;
 };
 
-export const accessOptions = {
```
simison, line 25: Anything that isn't static feels less of a "constant" to me so access options in this shared file felt better, but not a strong opinion.

lezama: you mean is not static because of the translations?

simison: Yes, but I'm just being a semantic. :-)

lezama: I like semantic too but I need it in a small file so the view bundle doesn't grow

simison: You reason is better! 👍 :-)

projects/plugins/jetpack/extensions/blocks/paywall/edit.js
```diff
@@ -29,12 +36,27 @@ function PaywallEdit( { className } ) {
...
+				<span style={ style }>
+					{ text }
+					<Icon icon={ arrowDown } size={ 16 } />
+				</span>
+			</div>
+			<InspectorControls>
+				<PanelBody
+					className="jetpack-subscribe-newsletters-panel"
+					title={ __( 'Newsletter visibility', 'jetpack' ) }
+					icon={ <JetpackLogo showText={ false } height={ 16 } logoColor="#1E1E1E" /> }
```
simison: Let's use the new shared component from https://github.com/Automattic/jetpack/pull/32257

simison (dismissed):
Tests well! 

Just one comment about sharing a component after rebase.

projects/plugins/jetpack/extensions/blocks/paywall/edit.js
lezama: good call

simison (approved):

PR #32265, opened 2023-08-03
Newsletter: visibility settings – remove notice

lezama:
Opened this PR for further discussion. Didn't even test.

Discussed with @crisbusquets that this notice makes little sense 🤔,
image

What do you all think?

Fixes #

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/projects/plugins/jetpack/changelog/update-newsletter-remove-notice
+++ b/projects/plugins/jetpack/changelog/update-newsletter-remove-notice
@@ -1,0 +1,4 @@
+Significance: minor
+Type: other
+
+Newsletter: remove notice
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/settings.js
@@ -4,7 +4,7 @@ import { useInstanceId } from '@wordpress/compose';
 import { useSelect } from '@wordpress/data';
 import { PostVisibilityCheck, store as editorStore } from '@wordpress/editor';
 import { createInterpolateElement } from '@wordpress/element';
-import { __, sprintf } from '@wordpress/i18n';
+import { __ } from '@wordpress/i18n';
 import {
 	META_NAME_FOR_POST_LEVEL_ACCESS_SETTINGS,
 	accessOptions,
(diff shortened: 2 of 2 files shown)


edanzer:
Is there any background conversation / context to check for this?

I'm assuming the idea is that, since the user has selected to send it to 'Anyone Subscribed', they know they're sending it to everyone, and showing the number of people is redundant? 

For my part, I tend to be very conservative about sending emails to readers/subscribers/customers. They can't be reversed. So for me, having a big notice that I'm about to send an email to, say, 1347 subscribers seems like something I'd want to see. It's a bit like a cautionary notice before publishing, since publishing in this case also equals hitting 'send' on emails.

lezama:
> 1347 subscribers seems like something I'd want to see

I agree with this sentiment, it's good to see that number somewhere.
I don't like the copy too much, took me many milliseconds to parse that "*also*" refers to "on top of posting".

crisbusquets:
> > 1347 subscribers seems like something I'd want to see
> 
> I agree with this sentiment, it's good to see that number somewhere. I don't like the copy too much, took me many milliseconds to parse that "_also_" refers to "on top of posting".

Yeah, maybe we can improve the copy to make it more clear?

Or maybe we can go back to this, and use the label to convey all the info:
- Anyone subscribed (2,019)
- Paid subscribers only (1,027)

I totally understand @edanzer's feedback, I do the same. I use Mailerlite, and the "audience selection" appears after you've prepared the email. And I think Substack doesn't even show a number.

simison:
> I totally understand @edanzer's feedback, I do the same. I use Mailerlite, and the "audience selection" appears after you've prepared the email. And I think Substack doesn't even show a number.

They might've removed the number at some point as I remember seeing permanent "n subscribers" under each label, or then I'm mixing with Beehiiv or other services.

Anyway, I feel like it's an useful number to show, but our version is a bit too elaborate and reads less as notice rather than helpful info. @crisbusquets any thoughts between these solutions?

Screenshot 2023-08-04 at 12 22 43

Screenshot 2023-08-04 at 12 25 46

Screenshot 2023-08-04 at 12 25 02

crisbusquets:
> > I totally understand @edanzer's feedback, I do the same. I use Mailerlite, and the "audience selection" appears after you've prepared the email. And I think Substack doesn't even show a number.
> 
> They might've removed the number at some point as I remember seeing permanent "n subscribers" under each label, or then I'm mixing with Beehiiv or other services.
> 
> Anyway, I feel like it's an useful number to show, but our version is a bit too elaborate and reads less as notice rather than helpful info. @crisbusquets any thoughts between these solutions?
> 
> Screenshot 2023-08-04 at 12 22 43 Screenshot 2023-08-04 at 12 25 46 Screenshot 2023-08-04 at 12 25 02

I think this is the right one:
1. It requires less vertical space
2. You can see the number of subscribers without having to click anything

I know we're discussing this in a P2, but what number will we show under `Everyone`? `Anyone subscribers` already include everyone that's subscribed.

I'm wondering if we should have this instead:
- Everyone
- Anyone subscribed (70)
- Paid subscribers only (30)

image

lezama:
I would simplify copies:

Anyone subscribed => All subscribers

Paid subscribers only => Paid subscribers

crisbusquets:
> I would simplify copies:
> 
> Anyone subscribed => All subscribers
> 
> Paid subscribers only => Paid subscribers

Works for me. This is what we had at some point (at least this is how it appears on the Newsletter LP screenshot).

lezama:
no longer makes sense

PR #32270, opened 2023-08-03
Fix/subscriber call when module not available

millerf:
Fixes regression in https://github.com/Automattic/jetpack/pull/32095#issuecomment-1663901576

## Proposed changes:

* Instead of passing isModuleActive and render null in the component, I render the component only if isModuleActive   
* Remove unecessary useEffect

### Other information:

## Testing instructions:

* See in https://github.com/Automattic/jetpack/pull/32095
OR
* Create a new JN instance
* Connect Jetpack, but do not enable the NL module
* Create a post
* Notice there is no calls to the "subscribers" endpoint and no errors in the editor
Screenshot 2023-08-04 at 12 51 40

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-subscriber-call-when-module-not-available
+++ b/projects/plugins/jetpack/changelog/fix-subscriber-call-when-module-not-available
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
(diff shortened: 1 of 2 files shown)


chrismccluskey (approved):
LGTM!

PR #32258, opened 2023-08-03
AI Control: introduce AiStatusIndicator component

retrofox:
This PR is a suggestion to improve the state style that we show when the app performs a request. It's based on the @keoshi design shared p1687347197727109/1687302104.930509-slack-C054LN8RNVA.

Also, it updates the AI Control component replacing the current loading component with this one:

Screenshot 2023-08-03 at 18 46 02

I've added slightly different states depending on the request status. Let's ping @ederrengifo @madebynoam for feedback here. Please feel free to suggest whatever you consider appropriate.

Fixes #

## Proposed changes:

* AI Control: introduce AiStatusIndicator component

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

Test with storybook

```
cd projects/js-packages/storybook/
```

```
pnpm run storybook:dev
````

https://github.com/Automattic/jetpack/assets/77539/e55fa58c-c048-4b02-92ce-bceb8a54859d

Here is how it works in the Jetpack Form AI composition feature

https://github.com/Automattic/jetpack/assets/77539/9fb4ebd6-605d-4784-9cf2-7619aa8429ee

PR #32272, opened 2023-08-03
AI Extension: show AiStatusIndicator in the assistant bar

retrofox:
This PR integrates finally the AiAssistantStatus component with the AiControl properly in the Jetpack Form Ai Asisstant feature. Follow-up https://github.com/Automattic/jetpack/pull/32258

Fixes #

## Proposed changes:

* AI Extension: show AiStatusIndicator into the assistant bar

### Other information:

## Testing instructions:

* Go to the block editor
* Create a Jetpack Form block instance
* Request a form
* Confirm the AiStatusInidicator component

https://github.com/Automattic/jetpack/assets/77539/259a5262-eb1f-464a-bfe6-26aa49682434

### DIFF
--- a/projects/plugins/jetpack/changelog/update-jetpack-form-ai-update-ai-status-indicator
+++ b/projects/plugins/jetpack/changelog/update-jetpack-form-ai-update-ai-status-indicator
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
+AI Extension: show AiStatusIndicator into the assistant bar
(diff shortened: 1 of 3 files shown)


lhkowalski:
Working for me:

Screen Shot 2023-08-03 at 15 41 38

The only status that I could not get to show is the error, even forcing an unclear prompt. There is a way for us to see it? Or not, in the form context so far?

Also, the CSS override of colors seems odd to me considering that the component itself already set it. Maybe we could make it stop overriding with gray in the first place, so we don't need the CSS change.

What do you think?

retrofox:
> The only status that I could not get to show is the error, even forcing an unclear prompt. 

image https://github.com/Automattic/jetpack/pull/32272/commits/64583e749c9f726e8278c3cf12c0ba0a2a61104e

> 
> Also, the CSS override of colors seems odd to me considering that the component itself already set it. Maybe we could make it stop overriding with gray in the first place, so we don't need the CSS change.

That's correct. We need to dig into something about the Package and how it provides the styles. For now, I think i's kinda fine to accept. We need to know why the Jetpack plugin doesn't get the styles from there.

lhkowalski (approved):
Got the orange on the error status as well:

Screen Shot 2023-08-03 at 16 35 13

Well done!

PR #32273, opened 2023-08-03
Remove unused api.js

millerf:
Fixes https://github.com/Automattic/jetpack/pull/32095#issuecomment-1663901576

## Proposed changes:

* Use store to retrieve the subscribers instead of a API call.
* Clean up some files and reuse some logic 

### Other information:

## Testing instructions:

* Go to your favorite testing site 
* Add a subscribe button to your post
* See this in the panel. 🚀
Screenshot 2023-08-04 at 13 53 15


projects/plugins/jetpack/extensions/blocks/subscriptions/edit.js
```diff
@@ -72,6 +72,31 @@ export function SubscriptionEdit( props ) {
 	} = props;
 	const { isLoadingModules, isChangingStatus, isModuleActive, changeStatus } =
 		useModuleStatus( name );
+
+	const { subscriberCount, subscriberCountString } = useSelect( select => {
```
chrismccluskey, line 95: I'm not sure why, but my local build of this branch as well as a build on JN (and using this branch directly with Jetpack Beta) both seem to think that `useSelect` is not a function, preventing this from working. Specifically, this happens when adding a subscribe button or viewing a page with a subscribe button. The result (along with the console log) is that the block isn't rendered. 

I see that we're importing it above, but it doesn't seem to be working. Do we want to use `withSelect` here? I see that it's already imported although I'm not sure of the difference between the two currently. I tried `withSelect` locally and it seemed to work as I would expect `useSelect` to work here.

chrismccluskey (changes requested):
I left a comment on one small change. I'm also seeing that some check annotations in unchanged files seem to be failing (see "Unchanged files with check annotations"). I'm not entirely sure what I'm looking at here but thought I'd point that out.

projects/plugins/jetpack/extensions/blocks/subscriptions/edit.js
millerf: > I see that we're importing it above, but it doesn't seem to be working. Do we want to use withSelect here? I see that it's already imported although I'm not sure of the difference between the two currently. I tried withSelect locally and it seemed to work as I would expect useSelect to work here.

There might be a conflict using both approaches at the same time... But [it is not in the doc](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-data/#useregistry)...

millerf: @chrismccluskey The issue ended up being `useSelect` was imported from the wrong package (`WP/elements` instead of `WP/data`).

But in the meantime I tried the `withSelect` approach that could not be used because of `isModuleActive` being retrieved from `useModuleStatus`. But you can have a look at the implementation on d0dd537da4

chrismccluskey (approved):
This works now, LGTM!

jeherve (approved):
LGTM! 🚢

PR #32275, opened 2023-08-03
Jetpack Sync: Add switching logic between Custom table and Options table

bisko:
This is a spin-off of #31318 and continuation of #32089, #32111 and #32135

## Proposed changes:

This PR builds on top of the migration and initialization logic introduced in #32135 and hooks into the option update for `custom_queue_table_enabled` to either initialize, switch over to the table and migrate the data. 

The PR depends on #32089, #32111 and #32135 and needs to have all of them merged before this one can be merged. We'll also need to update the changelog and package versions afterwards.

To be further described

### Other information:

## Jetpack product discussion

p9dueE-7e6-p2

No

## Testing instructions:

* Apply PR

* All the steps in the testing instructions comment in p9dueE-7e6-p2


chrisbliss18 (dismissed):
I was able to successfully complete all testing steps.

I validated that Sync was operating as expected while running these changes.

While running unit tests, I did receive a number of warnings, errors, and a failure; however, I ran the same tests on `trunk` and confirmed that the same warnings, errors, and failure were present there. So I don't believe that these changes introduce any new issues related to existing unit test coverage.

I ran the testing instructions from `p9dueE-7e6-p2`. I was able to successfully complete those tests and left a comment with my results.

Thanks for the changes and your various testing instructions. :)

Approved.

projects/packages/sync/src/class-settings.php
```diff
@@ -200,6 +201,84 @@ public static function update_settings( $new_settings ) {
...
+			 * Custom table migration logic.
+			 *
+			 * This needs to happen before the option is updated, to avoid race conditions where we update the option,
+			 * but haven't yet created the table or can't create it.
+			 *
+			 * On high-traffic sites this can lead to Sync trying to write in a non-existent table.
+			 *
+			 * So to avoid this, we're going to first try to initialize everything and then update the option.
+			 */
+			if ( 'custom_queue_table_enabled' === $setting ) {
```
fgiannar, line 214: As discussed in Slack, it might be worth checking if an import is currently in progress and skip updating the setting in this case. I was also thinking if it would make sense to check if the site is "out of Sync" while migrating and think if want to add special handling for cases with very long queues

fgiannar (commented):
Amazing work, Biser!

I've tested switching storage type and handling errors and works as expected. I created a corresponding diff on WPCOM that adds special logging for `storage_type` and a corresponding dashboard to monitor performance / errors and it all looks great so far 👍 

Left an inline comment about handling some edge cases, like long running imports and out_of_sync sites with very large queues.

**Update** I was able to reproduce a Fatal (table does not exist) with the custom table enabled upon deactivating Jetpack protect with Jetpack-the-plugin still active.

projects/packages/sync/src/class-settings.php
bisko: Let's leave this for a separate PR, as we need to have a more proper solution for it that will over-complicate the current PR.

bisko:
> Update I was able to reproduce a Fatal (table does not exist) with the custom table enabled upon deactivating Jetpack protect with Jetpack-the-plugin still active.

We're reverting the `DROP TABLE` on `uninstall` for now, as it has caused some unintended consequences and we want to address those properly in a follow-up PR, as it's affecting not only this functionality, but Sync's functionality as a whole.

fgiannar (approved):
We should be fine to merge this as is for now, as long as we ensure the table cleanup logic is added back before the next JP release (AT sites always have JP active, so this is fine for now)

Any edge cases will be handled during next week as we gradually roll out the new functionality to AT sites.

PR #32287, opened 2023-08-04
Social: Fix race condition in mobile app's post request

spsiddarthan:
Fixes 1205142973434174-as-1205190774917545

We will rely on the `wp_after_insert_post` instead. When `rest_api_inserted_post` exists with it, there is a race condition which causes posts from being publicized before the metadata is updated.

More information here: p1691156301965779/1691150316.598199-slack-C02JJ910CNL
## Proposed changes:

* Remove the rest_api_inserted_post hook 

### Other information:

## Jetpack product discussion

p1691156301965779/1691150316.598199-slack-C02JJ910CNL

No
## Testing instructions:

* Follow the testing instructions on D118215-code

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-mobile-sharing-toggling
+++ b/projects/plugins/jetpack/changelog/fix-mobile-sharing-toggling
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
+bugfix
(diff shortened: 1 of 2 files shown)


pablinos (commented):
This seems reasonable to me as we know that the `async_publicize_post` method will be called on the other hook, but only if Publicize has been properly initialised. This code seems to assume it is as otherwise `$GLOBALS['publicize_ui']` would be null, but it would be worth checking. I think we load all the Jetpack modules on `plugins_loaded` and Publicize is never disabled on wpcom, so we should be good.

gmjuhasz (approved):
Publicizing worked for me in the mentioned scenarios!

PR #32289, opened 2023-08-04
CRM Automations: Add types

robertf4:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3206

## Proposed changes:

Adds TypeScript types for the Automations module.

### Other information:

## Jetpack product discussion

pbhBOz-3ss-p2

No

## Testing instructions:

There is no functional testing with this PR. Simply inspect the code and verify that it models the types given in [src/automation](https://github.com/Automattic/jetpack/tree/trunk/projects/plugins/crm/src/automation).

### DIFF
--- a/projects/plugins/crm/src/js/components/automations-admin/types.tsx
+++ b/projects/plugins/crm/src/js/components/automations-admin/types.tsx
@@ -1,0 +1,6 @@
+export type Trigger = {
+	slug: string;
+	title: string;
+	description?: string;
+	category: string; // TODO: these can be a type
+};
--- a/projects/plugins/crm/src/js/components/automations-admin/types.tsx
+++ b/projects/plugins/crm/src/js/components/automations-admin/types.tsx
@@ -6,7 +6,7 @@ export type Trigger = {
 };
 
 export type Step = {
-	attributes: Step[];
+	attributes: string[];
 	nextStep?: Step;
 	slug: string;
 	title: string;
--- a/projects/plugins/crm/src/js/components/automations-admin/types.tsx
+++ b/projects/plugins/crm/src/js/components/automations-admin/types.tsx
@@ -26,7 +26,7 @@ export type Workflow = {
 	description: string;
 	category: Category;
 	triggers: Trigger[];
-	initial_step: Step;
+	initial_step: Step[];
 	active: boolean;
 	version: number;
 	added: string;
--- a/projects/plugins/crm/src/js/components/automations-admin/types.tsx
+++ b/projects/plugins/crm/src/js/components/automations-admin/types.tsx
@@ -29,6 +29,7 @@ export type Workflow = {
 	initial_step: Step;
 	active: boolean;
 	version: number;
+	added: string;
 };
 
 export type Type = 'contacts';
--- a/projects/plugins/crm/src/js/components/automations-admin/types.tsx
+++ b/projects/plugins/crm/src/js/components/automations-admin/types.tsx
@@ -2,5 +2,35 @@ export type Trigger = {
 	slug: string;
 	title: string;
 	description?: string;
-	category: string; // TODO: these can be a type
+	category: Category;
 };
+
+export type Step = {
+	attributes: Step[];
+	nextStep?: Step;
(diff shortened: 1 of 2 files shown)


zinigor:
Can you please merge trunk so we can merge this one?

robertf4:
Rebased against trunk.

projects/plugins/crm/src/js/components/automations-admin/types.tsx
```diff
@@ -0,0 +1,37 @@
...
+	description: string;
+	category: Category;
+	triggers: Trigger[];
+	initial_step: Step;
+	active: boolean;
+	version: number;
+	added: string;
+};
+
+export type Type = 'contacts';
```
gogdzl, line 35: Why is `Type` defined as `'contacts'` ?

projects/plugins/crm/src/js/components/automations-admin/types.tsx
```diff
@@ -0,0 +1,37 @@
...
+export type Workflow = {
+	id: number;
+	name: string;
+	description: string;
+	category: Category;
+	triggers: Trigger[];
+	initial_step: Step;
+	active: boolean;
+	version: number;
+	added: string;
```
gogdzl, line 32: Did we add a task to add `added` in our PHP code?

gogdzl (changes requested):

projects/plugins/crm/src/js/components/automations-admin/types.tsx
robertf4: The only type I could find in the code was `contacts` so I just stuck with that and planned to add others once they are known.

projects/plugins/crm/src/js/components/automations-admin/types.tsx
robertf4: I briefly looked for such a task and did not see one, but I believe you were going to create one if it didn't exist.

projects/plugins/crm/src/js/components/automations-admin/types.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export type Condition = Step;
+
+export type Workflow = {
+	id: number;
+	name: string;
+	description: string;
+	category: Category;
+	triggers: Trigger[];
+	initial_step: Step;
```
gogdzl: Our initial step is being defined as an array, would this be a problem or is it ok to always send `initial_step[0]`?

```
	/**
	 * The workflow initial step.
	 *
	 * @since $$next-version$$
	 * @var array
	 */
	public $initial_step;

	/**
	 * Get the initial step of this workflow.
	 *
	 * @since $$next-version$$
	 *
	 * @return array
	 */
	public function get_initial_step(): array {
		return $this->initial_step;
	}
```

gogdzl (dismissed):
Looks good to me, just check my last comment and if there is no problem merge it!

projects/plugins/crm/src/js/components/automations-admin/types.tsx
robertf4: Good catch, I changed this to be an array.

gogdzl (approved):

projects/plugins/crm/src/js/components/automations-admin/types.tsx
cleacos: Mmm here really the initial step is a Step instance. In the Workflow is defined as an array (as raw data) because is the way we are going to store it in the database (as JSON), but in this context, it could be wrapped as a Step instance.

projects/plugins/crm/src/js/components/automations-admin/types.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export type Condition = Step;
+
+export type Workflow = {
+	id: number;
+	name: string;
+	description: string;
+	category: Category;
+	triggers: Trigger[];
+	initial_step: Step[];
```
cleacos, line 29: `initial_step` is a `Step` type.

PR #32290, opened 2023-08-04
Update Release Script to handle Backporting

sdixon194:
## Proposed changes:

Updates the release script to handle backporting and starting a new cycle.

I don't know if other teams start new cycles, or just Jetpack (I don't think we even do a new cycle for the wpcom-mu-plugin right?)

### Other information:

## Jetpack product discussion

N/a

No
## Testing instructions:

Changes make sense? Can test it out next week for the weekly release.

### DIFF
--- a/.github/BACKPORT_RELEASE_CHANGES.md
+++ b/.github/BACKPORT_RELEASE_CHANGES.md
@@ -1,0 +1,29 @@
+<!--- This template is used for backporting changes made during a plugin release back to trunk. -->
+
+## Proposed changes:
+<!--- Explain what functional changes your PR includes -->
+Backports changes from PLUGIN_RELEASE X.Y
+
+### Other information:
+
+- [ ] Have you written new tests for your changes, if applicable?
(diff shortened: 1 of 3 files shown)


samiff (commented):
Think this in a good place to try for next release. Two thoughts:

1. Could `.github/BACKPORT_RELEASE_CHANGES.md` go into the `.github/files` dir somewhere instead? Just to help keep that root .github dir tidy.
2. Not sure if it'd be super helpful, but I wonder if we'd want to replace any of the project versions being updated in PR title/body, or commit message. There is `tools/plugin-version.sh` for example that can read out a plugin version.

sdixon194:
@samiff Something we could talk about as well is eschewing the template altogether and just write the body and title in the `gh` command. I think the only downside would be getting the annoying flag that the template hasn't been used correctly.

sdixon194:
Took a crack at just replacing the content in the .md file to include the version stuff dynamically.

If that looks good lmk and I can test it out for cutting the weekly today!

kraftbj (approved):

PR #32301, opened 2023-08-04
Publicize: Prevent meta updates on published posts

pablinos:
We've seen instances of posts being updated after publish but before the jobs have run to share them. To fix this we can prevent the metadata from being updated when a published post is saved.

### Other information:

## Jetpack product discussion
p1688069222155869-slack-CDD9LQRSN

No.

## Testing instructions:
* With a test site that has Publicize connections...
* Publish a post with the connections enabled
* Update the post and save it
* Without this patch the skip meta will be added to the post. With it, it won't
* Add multiple connections to the site. Skip one on publish
* Update the post. The skip meta for the other connections will be added without this change, but the meta will remain the same with it.


spsiddarthan (commented):
This works as expected. Seeing the comment that we shouldn't save the post meta when the post status is published, I was wondering what would happen to reshare ( formerly called republicize ), but that appears to work as well. 

I will test this some more today and tomorrow, and see if I am missing anything else. It's looking good thus far and I think we can definitely get it in for the next release.

spsiddarthan (dismissed):
This looks good!

spsiddarthan (approved):

PR #32305, opened 2023-08-05
Social: Update connection toggles to be button switches

oskosk:
Updates the toggles to match designs from p1HpG7-jqQ-p2

<table>
<tr>
 <td>image

 <td>
image

<tr>
 <td>Before
 <td>After
</table>

## Proposed changes:

* Updates the toggle components to be image switches
* Moves the toggles before the limits notice

### Other information:

## Jetpack product discussion
p1HpG7-jqQ-p2

No

## Testing instructions:

* With Social enabled and a few connections active, visit the editor
* Open the Jetpack plugin sidebar
* Confirm the connection buttons look like in the screenshots
* Click the add connection button
* Confirm you're taken to add a new connection


projects/js-packages/publicize-components/src/components/connection-icon/index.jsx
```diff
@@ -5,23 +5,37 @@ import PropTypes from 'prop-types';
...
-				{ hasDisplayPicture && <img src={ profilePicture } alt={ label } onError={ onError } /> }
-				<SocialServiceIcon
-					serviceName={ 'instagram-business' === serviceName ? 'instagram' : serviceName }
-					className="jetpack-publicize-gutenberg-social-icon"
-					invert={ 'tumblr' === serviceName }
-				/>
-			</div>
-			<span className="jetpack-publicize-connection-label-copy">{ label }</span>
-		</label>
+		<div
```
oskosk, line 24: this just removes the label that was there for the toggle. And also adds some attrs to the div to be compliant with the a11n rules

projects/js-packages/publicize-components/src/components/connection-icon/style.scss
```diff
@@ -2,51 +2,43 @@
 	margin-right: 10px;
 }
 
-.jetpack-publicize-connection-label {
```
oskosk, line 5: This whole diff looks weird but is just moving the rules outside `.jetpack-publicize-connection-label` as the label is no longer needed

projects/js-packages/publicize-components/src/components/connection-toggle/style.scss
```diff
@@ -7,8 +7,14 @@
...
 	// Unchecked state.
-	&.is-not-checked .jetpack-gutenberg-social-icon {
-		fill: $gray-300;
+	&.is-not-checked {
+		.components-connection-icon__picture img {
+			filter: saturate(0);
+		}
+		.jetpack-gutenberg-social-icon {
+			fill: $gray-300;
+			filter: saturate(0);
```
oskosk, line 16: I think this second `filter: saturate(0);` shouldn't be here. Does nothing

spsiddarthan:
Thanks for getting this started, @oskosk! 

Screenshot 2023-08-07 at 19 10 21

I added all the possible jetpack social connections, and they showed up vertically stocked and there are some free spaces to the right of it. 

We can take this on in our next sprint and improve on it, should you not have the time to go ahead with it.

oskosk:
> I added all the possible jetpack social connections

Sorry @spsiddarthan I think i forgot to commit a chat making the `<li>` be `display: inline` apparently

spsiddarthan:
Thanks for the input @oskosk! 

I pushed a change on the `<ul>` to have a `display:inherit`, since `<li> being display: inline` didn't work for me. I am no CSS expert though, so I will get that reviewed by our team. 

I noticed something which may be an issue from a UX perspective (video below). I published a post and wanted to reshare (formerly, republicize) a post. The Share Post button was disabled and I had to enable the connections to enable the share post button. 

But with too many disabled elements, I wasn't sure what to do, took me a while to realize that I should toggle the connection back on. 

With the existing designs on the trunk branch with the toggles next to the connection and labels, I felt like it was more apparent to the user on what they should do, but that could also be me just being used to the old designs. What are your thoughts about that? cc-ing  @pablinos and @gmjuhasz  as well. 


https://github.com/Automattic/jetpack/assets/6594561/0d255d31-17d2-406f-a4c4-8d62f8530103

I thought of a couple of follow-up improvements I have in mind for us to make: 
- Get rid of the 'Connect an account` external link, given that the plus icon next to the connection icons lets you do the same. 
- Move the notice for the advanced plan upsell below the `share_post` button.

pablinos:
>     Get rid of the 'Connect an account` external link, given that the plus icon next to the connection icons lets you do the same.
>    Move the notice for the advanced plan upsell below the share_post button.

Yes, good plan. I think we could do those as part of this PR. It would help to reduce the distance between the connection buttons/icons and the Share Post button too.

We could also change the message above the connections for a published post, or when no connections are enabled?

pablinos:
> Move the notice for the advanced plan upsell below the share_post button.

I'm going to leave this for now as the button isn't part of the form and it would be a bit more of a refactor. I feel like we might need to look at the notices in general. It's cluttering the UI a bit at the moment.

pablinos:
I've tweaked things and I think it's working pretty well now. I've had to tinker with the CSS a bit to get the vertical spacing working. It's still not perfect, but probably better than it is with the current toggles.

One thing with this is that now there's no way to differentiate between a broken/disabled connection and one that is turned off. We could add some sort of `!` icon in the top left or something to show there's a problem.

Also I've added the account name as a title for the connections, so that you get a tooltip on hover. I'm not sure how we should handle this for mobile.

projects/js-packages/publicize-components/src/components/connection-icon/index.jsx
pablinos: I've added the label back in as a title attribute, so that the accounts can be distinguished by hovering and getting a tooltip of the account name. We'll need to have a think about that for mobile.

projects/js-packages/publicize-components/src/components/connection-toggle/style.scss
pablinos: The whole rule wasn't doing anything as the class name was wrong, but without this filter, the filter on the avatar causes it to be rendered on top of the network icon. We also only render the icon when the avatar isn't available for any reason, so this means the icon can also be toggled.

spsiddarthan (approved):
Looks good to me!

projects/js-packages/publicize-components/jest-globals.js
```diff
@@ -10,3 +10,11 @@ if ( ! window.matchMedia ) {
 		dispatchEvent: jest.fn(),
 	} );
 }
+
```
gmjuhasz, line 13: Do we need to leave this in?

pablinos: Do you mean the initial state? If so, then yes, all the tests started failing without it. It seems like the connection is being used differently. We might be able to mock something else, but I noticed that Backup and a couple of other places had taken this approach.

oskosk:
You rock folks. Thanks @pablinos @spsiddarthan @gmjuhasz

projects/js-packages/publicize-components/src/components/connection-toggle/index.jsx
```diff
@@ -1,33 +1,25 @@
...
+				serviceName={ serviceName }
 				profilePicture={ profilePicture }
 			/>
-			<FormToggle
-				id={ id }
-				className={ className }
-				checked={ checked }
-				onChange={ onChange }
-				disabled={ disabled }
-			/>
```
manzoorwanijk, line 30: @pablinos This is where that bug of connections not disabled when limits are reached was introduced.

Issue #19897, opened 2021-05-20
Repo Gardening: extra label detection to a config file


jeherve:
The Repo Gardening action's `addLabels` task currently relies on detecting paths that are hardcoded in the action:
https://github.com/Automattic/jetpack/blob/a5f167cbc51792848f4fa53e57801bd709527bfa/projects/github-actions/repo-gardening/src/tasks/add-labels/index.js#L81-L91

While this works, it also means that the task cannot be used by third-parties unless their file structure matches ours. That is not ideal. 

Let's aim to have those paths defined in a config file, maybe a `.json` file that would live in the `.github` directory of any repo using the action and that task, and where we would define a list of label names and matching patterns that could be used by the action.

Noting that this config file would have to support more complex patterns, such as the ones we use here:
https://github.com/Automattic/jetpack/blob/a5f167cbc51792848f4fa53e57801bd709527bfa/projects/github-actions/repo-gardening/src/tasks/add-labels/index.js#L104-L123

************

Related PR: #19891

anomiex:
For the Required Review action I went with yaml data, which can be passed directly from the workflow or stored as a file in the repo. See https://github.com/Automattic/action-required-review/blob/8c79b3706c8e0022be1dbb106c0f8475d07d9c9c/src/main.js#L14-L37. The nice thing about yaml is that it can easily include comments, and the quoting is easier for humans to deal with.

> Noting that this config file would have to support more complex patterns, such as the ones we use here:

Off the top of my head, maybe something like this?
```yaml
# Mappings apply to any capturing group with the corresponding name, and map matched values to label values.
# The "*unmatched*" key is special, indicating how to handle entries not in the map.
#  - literal: Use the matched value unchanged. This should probably be the default.
#  - transform: Capitalize dash-separated words then join with spaces.
#  - error: Report an error.
# If no mapping is specified for a group name, it winds up defaulting to "literal".
mappings:
  ptype:
    editor-extensions: Block
    github-actions: Action
    packages: Package
    plugins: Plugin
    js-packages: JS Package
    '*unmatched*': error
  name:
    sharedaddy: Sharing
    shortcodes: Shortcodes / Embeds
    # ...
    '*unmatched*': transform

# Patterns map JavaScript-style regular expressions to labels.
# Named capturing groups in the regex are passed through corresponding mappings (above) before being substituted into labels.
# Note slashes don't need to be escaped in the regexes, and regexes are implicitly anchored at the start.
patterns:
  # Projects
[...]

PR #32306, opened 2023-08-06
Repo Gardening: Accept add_label input for passing custom path/label matching

dereksmart:
## Proposed changes:

Add add_labels input to addLabels task for setting custom path: label matching directly in the workflow like so: 
```
  with:
    tasks: 'addLabels'
    add_labels: '[
        {"path": "projects/your-project/", "label": "[Project] Your Project"},
        {"path": "somepath/", "label": "Some label"}
      ]'
```

This enables folks to not have to update the source in the monorepo for their own matching rules. 

## Jetpack product discussion

N/A

No

## Testing instructions:

You can test this by pointing the build to `uses: automattic/action-repo-gardening@ai-services`, which is [this branch](https://github.com/Automattic/action-repo-gardening/tree/ai-services/src/tasks) I created and hacked on in the action's mirror repo. Doing this allowed me to test it on both the ai-services repo as well as my personal fork. 

Note: the ai-services repo is currently running the ai-services branch and it seems to be fully functional. I plan to point it to trunk/stable again once this functionality is included. 
Check out the diffs that introduced it: 
- 221-gh-Automattic/ai-services
- 224-gh-Automattic/ai-services

### DIFF
--- a/projects/github-actions/repo-gardening/README.md
+++ b/projects/github-actions/repo-gardening/README.md
@@ -75,6 +75,7 @@ The action relies on the following parameters.
 
 - (Required) `github_token` is a GitHub Access Token used to access GitHub's API. The user account associated with the token is the one that will be seen as posting the checkDescription comment, adding and removing labels, and so on. If omitted, the standard token for the github-actions bot will be used.
 - (Optional) `tasks` allows for running selected tasks instead of the full suite. The value is a comma-separated list of task identifiers. You can find the list of the different tasks (and what event it's attached to) in `src/index.js`.
+- (Optional) `add_labels`. Pass custom labels to add. Defaults to an empty string. Only applies for the [addLabel](src/tasks/add-labels/readme.md) task.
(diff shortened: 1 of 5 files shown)


jeherve (approved):
This tests well for me. 

We were discussing something like this in #19897. I think this is a good first pass, and if needed we can expand on that in the future, to extract the paths/labels to a separate file where it may be easier to manage labels when you start having a lot of them.

dereksmart:
Thanks! 

> to a separate file where it may be easier to manage labels

Wonder what you have in mind for easier? The way I see it - whether it's the workflow file or another file, it's still editing one file.

jeherve:
> Wonder what you have in mind for easier? The way I see it - whether it's the workflow file or another file, it's still editing one file.

You're right, it may not be that different after all.
I like the separate file idea since it only contains the path/labels and nothing else, it's focussed on one thing and you do not have to worry about anything else. If it gets quite long it's still easy to parse. But I can see how splitting things between 2 files would make other things more difficult.

dereksmart:
Either way it should be pretty simple to modify handling of the add_labels input. Like if the string ends with `.json` assume it's a file to try to parse.

PR #32317, opened 2023-08-07
Social: Move Redux store to publicize-components

gmjuhasz:
This change moves the store over from Social to the publicize-components package.

## Proposed changes:

* Moved the store over into `social-store` in 80a1fde04c6a42035e06acb4817f556f413960ad
* Updated the imports and all the usage to use the store from publicize-components in 736029122ffbd1c1da49c1e2418d68cc1048620f
* Removed the old store in acd2b38cf5795db40f2ff5085c2be5e4aa16e454

### Other information:

## Jetpack product discussion

1205142973434174-as-1205212271186925

No.
## Testing instructions:

* Review the changes, it is easier to do by following the commits.
* Check that Jetpack Social works with the plugin enabled, and with it disabled and Jetpack enabled
  * SIG settings should be saved correctly on changes
  * Publishing to social media should work
  * Check on Simple and Atomic sites that it's working too


projects/js-packages/publicize-components/src/types/types.ts
```diff
@@ -0,0 +1,54 @@
+type JetpackSettingsSelectors = {
```
spsiddarthan, line 1: Is this a pre-existing file as well?

spsiddarthan (approved):
Works great and splendid work in breaking things down by commits!

I had just one question and tested for the different combinations mentioned.

projects/js-packages/publicize-components/src/types/types.ts
gmjuhasz: Yes! It came from the other folder where I removed it from.

PR #32325, opened 2023-08-07
get_post_access_level(): Allow $post_id argument

mreishus:
While working on the WordPress.com reader, I realized we needed to keep posts tagged as paid content out of the cache. This ensures users won't accidentally see content they aren't supposed to. The challenge? I wanted to leverage `get_post_access_level()`, but our caching logic runs outside the typical post rendering loop. Given that `get_post_access_level()` relies on `get_the_ID()`, this posed a problem.

To handle this, I've introduced an optional `$post_id` parameter to `get_post_access_level()`. This lets me determine the access level of posts, even when the global `$post` isn't set. Additionally, the classic way of just letting it use `get_the_ID()` still works perfectly.

## Proposed changes:

* `get_post_access_level()`: Now accepts an optional `$post_id` argument. If it's left out, it'll stick to the tried-and-true method of using `get_the_ID()`.

### Other information:

No

## Testing instructions:

* Access control on posts should continue to work.

### DIFF
--- a/projects/plugins/jetpack/changelog/update-get_post_access_level
+++ b/projects/plugins/jetpack/changelog/update-get_post_access_level
@@ -1,0 +1,5 @@
+Significance: patch
+Type: other
+Comment: Adding an optional parameter to get_post_access_level() does not seem very significant.
+
+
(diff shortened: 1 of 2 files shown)


millerf (approved):
LGTM

jeherve (approved):

PR #32329, opened 2023-08-08
[Super Cache] Fix gunzip warning

thingalon:
Fixes: pc9hqz-1WA-p2#comment-1472

In the latest Super Cache beta, if you do not have compression turned on you can sometimes get warnings as it tries to gunzip content that isn't already gzipped.

This PR addresses it by not attempting to gunzip content if compression is turned off. It also adds a warning quashing @ to the gunzip call just in case.

## Proposed changes:
* Don't attempt to uncompress data if the cache isn't compressed
* Quash warnings from gunzip.

### Other information:

## Jetpack product discussion

n/a

no

## Testing instructions:

* Test various configurations with gunzip on and off
* Make sure the cache works correctly and never outputs a warning.


donnchawp (changes requested):
I was able to reproduce the error when gzip support was disabled, since it was trying to uncompress an uncompressed file.

We don't need the $cache_compression check. That will cause the code to fall through to the last "else" section, which will serve the wp-cache cache file straight to the browser.

The reason being is $wp_cache_gzip_encoding. That's the output of gzip_accepted() which checks if:
1. PHP is compressing output
2. The browser doesn't accept encoded (gzip) output.

If either of those is true, then $wp_cache_gzip_encoding is false. That is, the plugin *must* serve uncompressed data to the browser.

Hiding the output of gzuncompress() should be sufficient.

I think we could surround the $uncompressed section of code with a `if ( $cache_compression ) {}` check. But for the moment, it's probably safer to simply hide the output of fzuncompress(). Most of the time, browsers will accept gzipped data.

The same could be done in the other section of the if() on $meta['dynamic'].

thingalon:
Thanks for explaining @donnchawp - on reading your explanation, I agree with your approach here.

donnchawp (approved):
Works as expected now.

donnchawp:
ff6bfb42b29a1bcb4ccb0906e880a0e510dee99c

PR #31423, opened 2023-06-17
Sync: initialize Dedicated Sync endpoint earlier

sergeymitr:
## Proposed changes:
Initialize dedicated sync endpoint earlier in the `init` hook to avoid cookie conflicts.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion
p1686285707645099-slack-CBG1CP4EN

## Does this pull request change what data or activity we track or use?
No.

## Testing instructions:
Make sure the code makes sense.

Do not checkout the code just yet, the issue may be a bit hard to reproduce.
1. Create a JN site with Woo, WCPay, Jetpack, and Smooth Generator.
2. Connect Jetpack, generated a few products.
3. Go to "WooCommerce -> Settings -> Accounts & Privacy" and enable "Allow customers to log into an existing account during checkout"
4. Create a new user.
5. In a private browser window, add a few products to the cart, go to checkout and login right there.
6. The products will probably still be in the cart as expected. Go to Cart and remove them. Then add a few products to the cart manually and remove them again. Make sure your cart is empty.
7. Close the window to clean the session, and repeat 6-7 a few times, and eventually you should lose the products in your card after logging in.
8. Checkout the PR and try a few more times, confirm the issue got fixed.

PR #32331, opened 2023-08-08
Fix: WooCommerce integration unit tests

chrisbliss18:
#### Changes proposed in this Pull Request:
There are two primary goals for this PR:
1) Fix the ability to run WooCommerce integration unit tests. It seems that disuse of this feature resulted in the breakdown of the ability to run these tests over time.
2) Add a unit test that will catch future possible regressions such as the one added by [this change](https://github.com/Automattic/jetpack/pull/31423). Moving this action too early on the `init` hook resulted in Sync failing to sync WooCommerce content.

The changes will affect the following:
* Jetpack

#### Other information:

#### Jetpack product discussion
* p9dueE-7zI-p2

#### Does this pull request change what data or activity we track or use?
* No

#### Testing instructions:
1. Check out this PR's branch: `fix/jetpack-woocommerce-unit-tests`.
1. Remove the WooCommerce plugin from inside the Jetpack repo if you currently have a copy.
1. Download the 8.0.1 [release of WooCommerce](https://downloads.wordpress.org/plugin/woocommerce.8.0.1.zip) and move it to `projects/plugins/woocommerce` in the Jetpack repo.
1. Make a shallow clone of the WooCommerce repo: `git clone --depth 1 --branch 8.0.1 https://github.com/woocommerce/woocommerce`.
1. Move the `woocommerce/plugins/woocommerce/tests` directory from the cloned repo to `projects/plugins/woocommerce/` in the Jetpack repo.
1. Once you have the WooCommerce repo in place, run `jetpack docker phpunit-woocommerce -- --filter WP_Test_Jetpack_Sync_WooCommerce` to run just the Jetpack Sync + WooCommerce specific tests. You should see that 12 tests are executed with no failures and only 1 risky test.


chrisbliss18:
@bisko Please review the `test_full_sync_order_items()` method in `projects/plugins/jetpack/tests/php/sync/test_class.jetpack-sync-woocommerce.php`. I wasn't able to fix up these tests so that they succeeded. Would you be able to look at them at some point so that they function again?

jeherve (commented):
> Run jetpack docker phpunit-woocommerce and see if the given error message is helpful enough to know how to update your repo as needed to run the unit tests.
> Once you have the WooCommerce repo in place [...]

This was a bit confusing to me. I wouldn't expect to have to temporarily clone a full version of the WooCommerce repo just to run the tests. I also couldn't get it to work, as I assume the WooCommerce plugin has to be built too:

```
Error in bootstrap script: Error:
Class "WC_Install" not found
#0 /usr/local/src/jetpack-monorepo/projects/plugins/jetpack/tests/php/bootstrap.php(105): include()
```

Could we instead do a shallow clone in a temporary directory when one runs the tests?

-------

Could you also bump the versions with `tools/fixup-project-versions.sh`?

chrisbliss18:
@jeherve Thanks for reviewing my changes. You're right about the complexity of the issue. After talking it over with my team, I'm going to simplify this PR to only focus on the new unit test that checks for the Sync regression. We're likely to also change how that test is approached.

For now, I'm changing this from "needs review" to "in progress" as I fix up the PR.

And thanks for the reminder to run `tools/fixup-project-versions.sh`. I'll get that taken care of as well as appropriate with my new changes.

chrisbliss18:
Updated PR to do the following:
1. Remove the changes to `Automattic\Jetpack\Sync\Actions`.
1. Fixed logic issues in `projects/packages/sync/src/modules/class-module.php` that break compatibility with non-core tables.
1. Updated `projects/packages/sync/src/modules/class-woocommerce.php` to allow it to work properly with full sync.
1. Updated the WooCommerce unit test code to make use of use a different format of the `woocommerce` files. Details can be found in the updated testing instructions for this PR.
1. Updated the WooCommerce unit tests to fix the `test_full_sync_order_items` tests. Many changes had to be made to allow these tests to be used. By default, WooCommerce changes are not tracked via fully sync. The `full_sync_limits` setting and the `full_sync_config` when calling `$full_sync->start()` both had to be modified to manually add `woocommerce`. I'm unsure if this is an issue in full sync or a purposeful design decision.
1. Added `test_dedicated_sync_order_items` tests to serve as a foundation to test for the `init` hook regression. To make these tests work properly to test for the regression, they need to run before the `init` hook runs so that it can detect if the `Automattic\Jetpack\Sync\Actions::add_dedicated_sync_sender_init()` function runs before WooCommerce registers its post types, taxonomies, etc. I'm open to suggestions on how to properly test for this.
1. Apply changes from running `tools/fixup-project-versions.sh`.

fgiannar:
With https://github.com/Automattic/jetpack/pull/32208 merged, the current PR is redundant. Closing.

PR #32335, opened 2023-08-08
Jetpack Sync: Drop queue custom table on sender uninstall

fgiannar:
Up to now, `Sync\Sender::uninstall` is the method used by consumer plugins on their deactivation hooks.
For this reason, the logic to remove the custom queue table is added there.
In the future, Sync should be able to detect whether this is the last Jetpack plugin being uninstalled and consumers should stop calling `Sync\Sender::uninstall` in their deactivation hooks, as there could be other Jetpack plugins present using Sync. However, this is not under the current scope of this PR

## Proposed changes:

* `Sync\Sender`: Update `uninstall` method to drop the queue custom table

### Other information:

## Jetpack product discussion

p9dueE-7e6-p2

No

## Testing instructions:

Code review mainly for now.
We'll be able to fully test this manually after the full queue custom table functionality is shipped.

### DIFF
--- a/projects/packages/sync/src/class-sender.php
+++ b/projects/packages/sync/src/class-sender.php
@@ -9,6 +9,7 @@ namespace Automattic\Jetpack\Sync;
 
 use Automattic\Jetpack\Connection\Manager;
 use Automattic\Jetpack\Constants;
+use Automattic\Jetpack\Sync\Queue\Queue_Storage_Table;
 use WP_Error;
 
 /**
@@ -997,5 +998,10 @@ class Sender {
 		// Clear the sync cron.
 		wp_clear_scheduled_hook( 'jetpack_sync_cron' );
 		wp_clear_scheduled_hook( 'jetpack_sync_full_cron' );
+
+		// Drom Sync queue custom table.
+		// Instantiate table storage, so we can drop the custom table. Queue ID is just a placeholder here.
+		$queue_table_storage = new Queue_Storage_Table( 'test_queue' );
+		$queue_table_storage->drop_table();
 	}
 }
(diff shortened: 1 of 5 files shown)


projects/packages/sync/src/class-sender.php
```diff
@@ -997,5 +998,10 @@ public function uninstall() {
 		// Clear the sync cron.
 		wp_clear_scheduled_hook( 'jetpack_sync_cron' );
 		wp_clear_scheduled_hook( 'jetpack_sync_full_cron' );
+
+		// Drom Sync queue custom table.
```
fgiannar:
```suggestion
		// Drop Sync queue custom table.
```

projects/packages/sync/src/class-sender.php
```diff
@@ -997,5 +998,10 @@ public function uninstall() {
 		// Clear the sync cron.
 		wp_clear_scheduled_hook( 'jetpack_sync_cron' );
 		wp_clear_scheduled_hook( 'jetpack_sync_full_cron' );
+
+		// Drop Sync queue custom table.
+		// Instantiate table storage, so we can drop the custom table. Queue ID is just a placeholder here.
+		$queue_table_storage = new Queue_Storage_Table( 'test_queue' );
```
bisko, line 1004: We should also reset the option `custom_queue_table_enabled` to `0` before deleting it, as we want to be able to restore it with the upgrade process if the plugin is re-installed.

If the option is already `1`, we won't go through the init process.

fgiannar: I believe this should be already covered by `Settings::reset_data();` that is part of `Sender::reset_data`

bisko (approved):

PR #32342, opened 2023-08-08
Social: Add media conversion settings endpoints

spsiddarthan:
Fixes 1205142973434174-as-1205074394227482

## Proposed changes:

* Added endpoints to get/update auto conversion settings.

### Other information:

## Jetpack product discussion

1205142973434174-as-1205074394227482

No
## Testing instructions:

* cd projects/packages/publicize
* composer run test-php
* Sanity check: Boot a JN site up on this branch and make sure publicize works correctly and updating SIG settings continue to work correctly.

### DIFF
--- a/projects/packages/publicize/src/auto-conversion-settings/class-rest-settings-controller.php
+++ b/projects/packages/publicize/src/auto-conversion-settings/class-rest-settings-controller.php
@@ -74,11 +74,11 @@ class REST_Settings_Controller extends WP_REST_Controller {
 		$settings = new Settings();
 
 		if ( isset( $request['image'] ) ) {
-			$settings->enable_or_disable( 'image', $request['image'] );
+			$settings->set_enabled( 'image', $request['image'] );
 		}
 
 		if ( isset( $request['video'] ) ) {
-			$settings->enable_or_disable( 'video', $request['video'] );
+			$settings->set_enabled( 'video', $request['video'] );
 		}
 
 		return rest_ensure_response( $this->get_auto_coversion_settings() );
@@ -90,7 +90,7 @@ class REST_Settings_Controller extends WP_REST_Controller {
 	 * @return bool|WP_Error True if user can manage options.
 	 */
 	public function settings_permissions_callback() {
-		if ( ! current_user_can( 'manage_options' ) ) {
+		if ( ! current_user_can( 'edit_posts' ) ) {
 			return new WP_Error(
 				'rest_forbidden_context',
 				__( 'Sorry, you are not allowed to access this endpoint.', 'jetpack-publicize-pkg' ),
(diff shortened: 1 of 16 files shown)


projects/packages/publicize/src/auto-conversion-settings/class-settings.php
```diff
@@ -0,0 +1,118 @@
...
+		}
+
+		return false;
+	}
+
+	/**
+	 * Enable or disable Auto Conversion.
+	 *
+	 * @param bool $key Whether video or image.
+	 * @param bool $value True to enable SIG, false to disable.
```
gmjuhasz: I think this wanted to be Auto conversion and not SIG

projects/packages/publicize/src/auto-conversion-settings/class-settings.php
```diff
@@ -0,0 +1,118 @@
...
+	 * @return bool True if the setting was updated successfully, false otherwise.
+	 */
+	public function enable_or_disable( $key, $value ) {
```
gmjuhasz: I am not sure I like this function name. Since we will use this in this namespace anyway, could it be `set_enabled` or `set_enabled_for` like with SIG? I feel that having enable and disable in the function name makes it harder to understand what the value `true` does for example

Would be used like this below
```
$settings->set_enabled( 'video', true );
$settings->set_enabled( 'image', false );
```

projects/packages/publicize/tests/php/test-auto-conversion-controller.php
gmjuhasz, line 1: Nice tests!

projects/packages/publicize/src/auto-conversion-settings/class-rest-settings-controller.php
```diff
@@ -0,0 +1,130 @@
...
+		return rest_ensure_response( $this->get_auto_coversion_settings() );
+	}
+
+	/**
+	 * Check the permissions for accessing and updating the settings endpoint.
+	 *
+	 * @return bool|WP_Error True if user can manage options.
+	 */
+	public function settings_permissions_callback() {
+		if ( ! current_user_can( 'manage_options' ) ) {
```
gmjuhasz: Is this the good permission for us? Just asking because of Jeremys findings p1692177377724379-slack-C02JJ910CNL. We might need to give this to editor role too, what do you think @pablinos ?

projects/packages/publicize/src/auto-conversion-settings/class-settings.php
spsiddarthan: Good point!

projects/packages/publicize/tests/php/test-auto-conversion-controller.php
spsiddarthan: Thank you! :)

projects/packages/publicize/src/auto-conversion-settings/class-rest-settings-controller.php
spsiddarthan: I have changed this to 'edit_posts'

gmjuhasz (approved):

PR #32343, opened 2023-08-08
AI Assistant: Add styles to forms

### LINKED ISSUE #32151: AI Extension: keep improving prompt to handle styles

dhasilva:
Fixes #32151

## Proposed changes:

* Tweaks the prompt to add styles

### Other information:

## Testing instructions:

* Test the form as normal, there should be no regressions
* Ask for styles on new forms and on existing forms

To help testing: "Make a pizza ordering form with dark yellow background and red text", "make the borders rounded".


CGastrell:
This is working quite nicely! Of course I drove the AI crazy with little requests, there are still options it cannot understand, but I'd say this is a good iteration.

2 things that might not be related with this PR:
- focus: once the form is built I don't get the focus back on the AI assistant input. Would be comfy to get it there as it invites to keep interacting with the AI with subsequent modification requests
- Not sure this is only on my local: the first prompt always gets the "Your request was unclear. Mind trying again?". Then I do try again and it works, but the notification doesn't go away

dhasilva:
About the focus, that would be a nice improvement indeed. Adding a card about that. Thanks!

As for the `Mind trying again?`, it happens here very rarely when the OpenAI servers are overloaded. But `Your request was unclear.` does not happen. Mind sharing your initial prompt, please? The debug lines (`localStorage.setItem('debug', '*');`)
 would help as well to determine if the problem is the problem is in the prompt or on how we are checking for errors. I'll create a card about the persistent notification.

CGastrell:
What do you know, can't make it happen again now. I think I was running with `jetpack watch plugins/jetpack` which can bring my mac to its knees.

> (`localStorage.setItem('debug', '*');`)

I'm familiar with debug, yet I didn't get any messages from this (even after reload).

Anyways, as I said, couldn't trigger that message again, I'll be on the lookout for it though. Thanks for the tips!

PR #32350, opened 2023-08-08
AI Client: Handle token fetching errors from inside the SuggestionEventSource

lhkowalski:
Errors that happen when fetching the JWT were not handled properly, resulting on an invalid state for the application when it happens.

## Proposed changes:

* Move the token fetching logic to the SuggestionEventSource so we can handle exceptions there and dispatch the proper error event when it happens
* Start dispatching the `ERROR_REQUEST` with the `ERROR_NETWORK` details when some problem happens on the JWT fetching operation
* Fix typo on `handleNetworkError` callback name

### Other information:

## Testing instructions:

* Go to the block editor and open your browser console on the network tab
* Insert a Jetpack Form block on the editor
* On the browser console, change the network throttling setting to "Offline" so we simulate a connection problem
* On the AI Extension, ask for some form (for example, "a post-meetup feedback form")
* Before this change, the extension would be stuck on the requesting state, because the JWT request would fail and the error would not be handled
* After this change, you will see an "It was not possible to process your request. Mind trying again?" message and the orange state on the assistant icon, indicating an error:

Screenshot 2023-08-08 at 15 41 29


projects/js-packages/ai-client/src/hooks/use-ai-suggestions/index.ts
```diff
@@ -282,8 +282,8 @@ export default function useAiSuggestions( {
 
 				eventSource.addEventListener( 'done', handleDone );
 			} catch ( e ) {
-				// eslint-disable-next-line no-console
-				console.error( e );
+				debug( 'Error outside the event source object: %o', e );
+				handleNetwotkError();
```
jsnmoon: Typo nit: `handleNetwotkError` should be `handleNetworkError`

jsnmoon (dismissed):
These changes appear reasonable and work as described in the testing instructions. I left a  nitty comment, but that shouldn't block merging this work.

I also noticed we're failing e2e's, but they appear unrelated to our changes.

Caveat: I'm new to the `ai-client` package code base and might be missing some context.

> [...] the `useAiSuggestions` hook is called without an error handler callback [...] [an error] handler is only attached [...] to the event souce object [...] [not to] the `handleError` callback of the hook

I assume this refers to our prior lack of error handling in the try/catch clause in `use-ai-suggestions`. Is this correct?

projects/js-packages/ai-client/src/data-flow/use-ai-context.ts
```diff
@@ -60,6 +60,20 @@ export default function useAiContext( {
...
+	 */
+	useEffect( () => {
+		if ( ! requestingError ) {
+			return;
+		}
+
+		if ( onError ) {
+			onError( requestingError );
+		}
+	}, [ requestingError ] );
```
retrofox: imho, we should avoid as much as possible propagating the error through an effect hook if we could rely on listening events, better.

retrofox (changes requested):
I've researched how to improve the error handling considering the goal of relying on 100% on dispatching events.

Before proceeding, it's important to note that the `offline` error may not be something we need to worry about, as the app already takes care of it when you try to save the post, for instance.

image

Although, if we do want to handle this error at the block level, probably we can listen to the `offline` event of the Window object. Something like the following:

```es6
// Listen the offline event to dispatch an error.
useEffect( () => {
	window.addEventListener( 'offline', offlineError );
	return () => {
		window.removeEventListener( 'offline', offlineError );
	};
}, [] );
```

But again, the app will show two notices for the same error:

image

projects/js-packages/ai-client/src/hooks/use-ai-suggestions/index.ts
lhkowalski: Fixed. Thanks for spotting it!

projects/js-packages/ai-client/src/data-flow/use-ai-context.ts
lhkowalski: Moved to the event approach that we discussed.

lhkowalski:
> the app already takes care of it when you try to save the post

@retrofox I think it's important to handle them on the extension as well because it leaves the application on a hanged, `processing` state that you can't get off unless you delete the block and start again.

I implemented the event dispatch idea that you suggested and the result is a cleaner code with cleaner error handling logic. Thank you for it.

I am fixing some other issues that I found and then I will ask for another review on it.

projects/js-packages/ai-client/src/suggestions-event-source/index.ts
```diff
@@ -126,6 +130,18 @@ export default class SuggestionsEventSource extends EventTarget {
...
+			try {
+				debug( 'Token was not provided, requesting one...' );
+				token = ( await requestJwt() ).token;
+			} catch ( err ) {
+				this.processErrorEvent( err );
+
+				return;
+			}
+		}
+
```
retrofox: Can we move this block just below [here](https://github.com/Automattic/jetpack/blob/2f250047eed1287139b64109ab9e18416bd898e0/projects/js-packages/ai-client/src/suggestions-event-source/index.ts#L87).
Let's give it more priority to check the token. Without the token doesn't make sense to continue.

retrofox (dismissed):
I've added a comment that I think we should do on this PR. Beyond of this, the PR us flawless, again. Nice job! 🚀 🚀 🚀 

Ping me for a fresh ✅

projects/js-packages/ai-client/src/suggestions-event-source/index.ts
lhkowalski: Thank you for the suggestion, just updated the code.

retrofox (approved):

PR #32353, opened 2023-08-08
Removed PHP 5.6 from WordPress latest and added it to previous.

zinigor:
This PR removes PHP 5.6 tests from WordPress 6.3 matrix because WordPress 6.3 does not allow 5.6 anymore.

## Proposed changes:
* Removes 5.6 from the `latest` column.
* Adds 5.6 to the `previous` column.
* Adds 8.2 to the `trunk` column.

### Other information:
- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

See #31714

## Testing instructions:

* Go to '..'
*

PR #32356, opened 2023-08-09
Subscriptions: Add check before showing subscribe modal

edanzer:
## Proposed changes:

This PR adds another check before displaying the Subscribe Modal. We were checking for the 'subscribe' query parameter. It is set when a user submits the subscribe form, and we don't want to show the modal again in that case. But in the WordPress.com environment, we set a 'blogsub' query param instead of a 'subscribe'. We weren't checking for that one. With this PR, we are.

### Other information:

## Jetpack product discussion

None.

No.

## Testing instructions:

1) Setup: 
   - You will need a WordPress.com simple site that has newsletter as the site_intent, or has the Lettre theme active. If you need one, you can set up a new site at https://wordpress.com/setup/newsletter/intro.
   - This site will also need a blog post, so be sure to publish one.
   - On your test site, go to Settings > Reading > Newsletter and enable the subscribe modal.
   - Run the bin command provide by Github below to load this branch in your sandbox. Sandbox public-api.wordpress.com and the domain of your test site.

2) Test: 
   - Go to a browser where you are not logged in to WordPress.com and open up a single blog post on your test site.
   - Scroll and pause. You should see the subscribe modal appear.
   - Enter any email and submit. When you do, the page will refresh/reload.
   - Scroll and pause - this time confirm the modal does not appear.

### DIFF
--- a/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
+++ b/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
@@ -219,6 +219,7 @@ HTML;
 
 		// Don't show if one of subscribe query params is set.
 		// They are set when user submits the subscribe form.
+		// The nonce is checked elsewhere before redirect back to this page with query params.
 		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
 		if ( isset( $_GET['subscribe'] ) || isset( $_GET['blogsub'] ) ) {
 			return false;
--- a/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
+++ b/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
@@ -217,10 +217,10 @@ HTML;
 			return false;
 		}
 
-		// Don't show if subscribe query param is set.
-		// It is set when user submits the subscribe form.
+		// Don't show if one of subscribe query params is set.
+		// They are set when user submits the subscribe form.
 		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
-		if ( isset( $_GET['subscribe'] ) ) {
+		if ( isset( $_GET['subscribe'] ) || isset( $_GET['blogsub'] ) ) {
 			return false;
 		}
 
--- a/projects/plugins/jetpack/changelog/update-add-check-on-subscribe-modal
+++ b/projects/plugins/jetpack/changelog/update-add-check-on-subscribe-modal
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
+Subscriptions: Add check when showing subscribe modal


simison (dismissed):
Thanks!

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -217,10 +217,10 @@ public function should_user_see_modal() {
 			return false;
 		}
 
-		// Don't show if subscribe query param is set.
-		// It is set when user submits the subscribe form.
+		// Don't show if one of subscribe query params is set.
+		// They are set when user submits the subscribe form.
```
simison, line 221: Comment could elaborate a bit more when each gets set, but fine as is, too.

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -217,10 +217,10 @@ public function should_user_see_modal() {
...
 		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
-		if ( isset( $_GET['subscribe'] ) ) {
+		if ( isset( $_GET['subscribe'] ) || isset( $_GET['blogsub'] ) ) {
```
simison, line 224: Tested well with Jetpack self hosted, which gave URLs like`?subscribe=pending#subscribe-blog-3`

Looking at the code  `blogsub` is coming from the wp.com simple backend for paid subscriptions only? If I don't have paid subscriptions configured, we use `subscribe`.

simison:
Rebased in case that would help with random test failures, but there wasn't much to rebase. Some convo at Crew channel indicates this is widespread and unrelated.

Update: another rebase should fix it, due to https://github.com/Automattic/jetpack/pull/32353

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -217,10 +217,12 @@ public function should_user_see_modal() {
...
 
-		// Don't show if subscribe query param is set.
-		// It is set when user submits the subscribe form.
+		// Don't show if one of subscribe query params is set.
+		// They are set when user submits the subscribe form.
+		// Subscribe is set on self hosted sites.
+		// Blogsub is set in some situations on WordPress.com.
 		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
-		if ( isset( $_GET['subscribe'] ) ) {
+		if ( isset( $_GET['subscribe'] ) || isset( $_GET['blogsub'] ) ) {
```
jeherve, line 224: Since we're touching that check, I wonder if you take the opportunity to add a comment to the phpcs:ignore, explaining why it's ok to ignore the need for a nonce here?

simison: Nice. +1  I actually reviewed that as well and checked that it indeed is ok to ignore, because we check the nonce elsewhere and these are just final redirected URLs after operations that happen elsewhere.

simison (dismissed):

simison (dismissed):

simison (approved):

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
jeherve: Right, I meant adding a comment to the phpcs:ignore, to indicate that we've actually checked. Here is an example:
https://github.com/Automattic/jetpack/blob/d51bb3b103a60f5e3d911aa042cb4bb388f894a6/projects/plugins/jetpack/class.jetpack-network.php#L77

simison: Does it affect anything to have it on top, or does that comment appear somewhere if done in that specific format? 

I added comment in https://github.com/Automattic/jetpack/pull/32356/commits/bbb487a3df4ff7bc5cd43f267ddbecceecd8efa9 — but happy to change format if it helps :-)

jeherve: It's just fine the way you did it, no worries :)

PR #27102, opened 2022-10-26
SSO: properly disable "Match by email" by default

sergeymitr:
#### Changes proposed in this Pull Request:
Jetpack SSO has "match by email" disabled by default in the UI.
This PR brings the code in accordance to that.

The code treats the option `jetpack_sso_match_by_email` as `true` if it does not exist.
That means that even though the UI shows "match by email" disabled by default, it's in fact enabled.
To properly disable it, an admin would need to:
1. Enable "Match by email" to create the option and set it to `true`.
2. Disable "Match by email" to set the option to `false`.

#### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?

#### Does this pull request change what data or activity we track or use?
No.

#### Testing instructions:
1. Create a new site and enable SSO. Do not touch the "Match by email" setting, leave it disabled.
2. Create a user with email matching another WPCOM account (let's call them "user 2").
3. Use the "user 2" account to log into WP-admin for the first time via SSO. You should see the "We couldn't find your account" error.
4. Go to "Jetpack -> Settings -> SSO" and enable the "Match by email" setting.
5. Try to log in as "user 2" via SSO again, you should succeed.
6. Go to "Jetpack -> Dashboard" and disconnect "user 2" from WPCOM.
7. Go to "Jetpack -> Settings -> SSO" and disable "Match by email".
8. Log out of WPCOM and try to login via SSO. You should see the "We couldn't find your account" error.
9. Add the following code to the `wp-config.php`: `define( 'WPCC_MATCH_BY_EMAIL', true );`
10. Try to login as "user2" via SSO again. You should succeed.

PR #32370, opened 2023-08-09
Jetpack SSO: enable 'Match by Email' by default

sergeymitr:
## Proposed changes:
This PR changes the default behavior of "Match by Email" functionality. This option has historically been enabled by default since the beginning of times. In #27102 we made the default be `off` for consistency with its representation in the UI, and it stayed like this for the last 8 months. But there were consequences. See p9GBOq-5Re-p2. With this update, the option will be reactivated by default for all users who have never modified this particular setting.

**Before:** Match by Email disabled by default
**After:** Match by Email is enabled by default, and only disabled if an admin manually disables it.

### Other information:

## Jetpack product discussion
p9GBOq-5Re-p2
#27102
p1690992724021369-slack-CDD9LQRSN
p9o2xV-2zY-p2

## Testing instructions:
1. Create a new site and enable SSO in "Jetpack -> Settings -> Security". Do not touch the "Match by email" setting, leave it enabled.
2. Create a user with email matching another WPCOM account (let's call them "user 2").
4. Use the "user 2" account to log into WP-admin for the first time via SSO, you should succeed.
5. Go to "Jetpack -> Dashboard" and disconnect "user 2" from WPCOM.
6. Go to "Jetpack -> Settings -> Security" and disable the "Match by email" setting.
7. Try to log in as "user 2" via SSO again. You should see the "We couldn't find your account" error.
9. Go to "Jetpack -> Settings -> Security" and enable "Match by email".
10. Disconnect "user 2" from WPCOM.
11. Logout in Jetpack and WPCOM as "user 2" and try to login via SSO, you should succeed.


oskosk (approved):
Tests well and LGTM

PR #32379, opened 2023-08-09
CRM: Fix error 219 when using date range segments

gogdzl:
This PR addresses the bug causing error 219 to arise when using date ranges within segments.

## Proposed changes:

* This PR is a bugfix.

### Other information:

## Jetpack product discussion
https://github.com/Automattic/zero-bs-crm/issues/3276

No.

## Testing instructions:

1. Activate Advanced Segments plugin
2. Create contacts with custom date fields of various dates
3. Create a new segment (`wp-admin/admin.php?page=manage-segments`) with a condition 'in date range' and select a range that includes custom field values you added in step 2 
4. Click `Preview Segment`
5. Try to save the segment

You should be able to preview and save date range segments.

### DIFF
--- a/projects/plugins/crm/changelog/fix-crm-3276-error-219-when-using-date-range-segments
+++ b/projects/plugins/crm/changelog/fix-crm-3276-error-219-when-using-date-range-segments
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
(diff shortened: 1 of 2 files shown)


cleacos (approved):
It works as expected.

LGTM!

PR #32389, opened 2023-08-10
CRM: Add workflow row

robertf4:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3208

## Proposed changes:

This PR adds a workflow row component.

### Other information:

## Jetpack product discussion

No

## Testing instructions:

1. Review the code.
2. Put `add_filter( 'jetpack_crm_feature_flag_automations', '__return_true' );` in an mu-plugin.
3. Visit `/wp-admin/admin.php?page=jpcrm-automations#/automations/test-workflow-row`. Verify that the visuals match the figma designs. Note that this PR only creates the row component. The table and table headers will be created in the next PR, so the current table is just a testing placeholder that can be ignored.


projects/plugins/crm/.eslintrc.js
```diff
@@ -7,5 +7,6 @@ module.exports = {
 				allowedTextDomain: 'zero-bs-crm',
 			},
 		],
+		'jsdoc/require-jsdoc': 'off',
```
robertf4, line 10: I disabled this because the rule is inconsistent in lamba functions vs function declarations, and there was a conversation in Slack recently regarding this.

gogdzl (dismissed):
Code looks good, and the UI visuals match the Figma file!

I'd like to mention that I'm not really a fan of how the checkboxes appear when they're checked. However, I think this might be beyond the scope of this PR:
image

robertf4:
> Code looks good, and the UI visuals match the Figma file!
> 
> I'd like to mention that I'm not really a fan of how the checkboxes appear when they're checked. However, I think this might be beyond the scope of this PR: image

We can definitely look at this in a later PR. Right now it's just a copy of how it's done in the main Jetpack plugin.

gogdzl (dismissed):

gogdzl (approved):

PR #32393, opened 2023-08-10
Dashboard: update subscribers link to always point to Calypso

jeherve:
## Proposed changes:

Let's always point folks towards the Calypso page when they need to have more information about their subscribers. We previously pointed to the Subscriber Stats page in wp-admin when the Stats module was active, but that page does not include as much information as the Calypso page for now.

### Other information:

## Jetpack product discussion

See https://github.com/Automattic/jetpack/pull/31718#issuecomment-1671992894

* No

## Testing instructions:

* Go to Jetpack > Settings > Newsletter
* Check the link at the bottom the Subscriptions card when the feature is active; it should point to Calypso.

### DIFF
--- a/projects/plugins/jetpack/_inc/client/newsletter/index.jsx
+++ b/projects/plugins/jetpack/_inc/client/newsletter/index.jsx
@@ -13,7 +13,7 @@ import SubscriptionsSettings from './subscriptions-settings';
  * @returns {React.Component} Newsletter settings component.
  */
 function Subscriptions( props ) {
-	const { active, isModuleFound, searchTerm, siteAdminUrl, siteRawUrl } = props;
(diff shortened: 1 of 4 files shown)


simison (approved):
Code looks good and works well; tested with Jetpack site.

image

I think this is good to merge as is, but would be nice to also add copy for adding subscribers which is possible on .com interface.

simison:
BTW I noticed that the "new tab" icon is right-aligned, should be next to the link:

Screenshot 2023-08-10 at 12 35 06

jeherve:
> I noticed that the "new tab" icon is right-aligned, should be next to the link:

That's the current design for those cards, so it's working as intended. I agree it looks a bit off though :)

PR #32405, opened 2023-08-10
AI Form Extension: Remove error notice on new request

### LINKED ISSUE #32381: AI Extension: Error notification does not go away on new request
ref https://github.com/Automattic/jetpack/pull/32343#issuecomment-1671967563

> Not sure this is only on my local: the first prompt always gets the "Your request was unclear. Mind trying again?". Then I do try again and it works, but the notification doesn't go away

The notification can only be removed manually. We need to make it go away after a new request is sent.

dhasilva:
Fixes #32381

## Proposed changes:

* Removes any notice with the `ai_assistant` id when sending a new request

### Other information:

## Testing instructions:

* Add a form
* Cause an error (like forcing it on code or turning off network temporarily)
* Check that the error notice is displayed
* Sendo another request
* Check that the error notice is removed

### DIFF
--- a/projects/js-packages/ai-client/changelog/update-ai-assistant-close-notice
+++ b/projects/js-packages/ai-client/changelog/update-ai-assistant-close-notice
@@ -1,0 +1,4 @@
+Significance: patch
+Type: removed
+
+AI Client: Remove redundant switch case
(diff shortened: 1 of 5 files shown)


lhkowalski (approved):
Code looks nice. Good job on cleaning the `switch` code.

Tested it by forcing a generic error on the first request:

Screenshot 2023-08-10 at 17 49 51

Then I removed the error cause and did another request, and it removed the existing notice before start completing the new request:

Screenshot 2023-08-10 at 17 50 03

🚀

PR #32409, opened 2023-08-10
Newsletters: Render subscriptions block on emails

lezama:
## Proposed changes:

Improve how the Subscriptions block renders on emails.

*  Add a specific rendering method tailored for email display.
*  Use block attributes or global styles if defined.

POST:
Screenshot 2023-08-10 at 18 22 12

EMAIL:
Screenshot 2023-08-10 at 18 22 57

(Pardon my French, that's a separate bug :))

### Other information:

No

## Testing instructions:

* Follow the instructions in D118655-code
* Try updating the block attributes. These should reflect on the emails.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -407,11 +407,90 @@ function get_element_styles_from_attributes( $attributes ) {
 		$email_field_styles   .= $style;
 	}
 
+	if ( ! jetpack_is_frontend() ) {
+		$background_color_style = get_attribute_color( 'buttonBackgroundColor', $attributes, '#113AF5' /* default lettre theme color */ );
+		$text_color_style       = get_attribute_color( 'textColor', $attributes, '#FFFFFF' );
+		$submit_button_styles  .= sprintf( ' background-color: %s; color: %s;', $background_color_style, $text_color_style );
+	}
+
 	return array(
 		'email_field'           => $email_field_styles,
 		'submit_button'         => $submit_button_styles,
 		'submit_button_wrapper' => $submit_button_wrapper_styles,
 	);
+}
+
+/**
+ * Retrieve the resolved color for a given attribute.
+ *
+ * @param string $attribute_name The name of the attribute to resolve.
+ * @param array  $attributes     An array of all attributes.
+ * @param string $default_color  A fallback color in case no color can be resolved.
+ *
+ * @return string Returns the resolved color or the default color if no color is found.
+ */
+function get_attribute_color( $attribute_name, $attributes, $default_color ) {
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -724,6 +807,36 @@ class="<?php echo esc_attr( $classes['submit_button'] ); ?>"
...
+ * @return string
+ */
+function render_for_email( $data, $styles ) {
+	$submit_button_wrapper_style = ! empty( $styles['submit_button_wrapper'] ) ? 'style="' . esc_attr( $styles['submit_button_wrapper'] ) . '"' : '';
+	$button_text                 = wp_kses(
+		html_entity_decode( $data['submit_button_text'], ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401 ),
+		Jetpack_Subscriptions_Widget::$allowed_html_tags_for_submit_button
+	);
+
+	$html = '<div ' . wp_kses_data( $data['wrapper_attributes'] ) . '>
```
lezama, line 824: @simison what do you think about the html here? I tried to keep it as close to the frontend one as possible so it leverage the custom attributes, but we could keep it simpler. If you see the screenshot, it adds an unexpected small margin.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -407,13 +407,92 @@ function get_element_styles_from_attributes( $attributes ) {
 		$email_field_styles   .= $style;
 	}
 
+	if ( ! jetpack_is_frontend() ) {
+		$background_color_style = get_attribute_color( 'buttonBackgroundColor', $attributes, '#113AF5' /* default lettre theme color */ );
+		$text_color_style       = get_attribute_color( 'textColor', $attributes, '#FFFFFF' );
+		$submit_button_styles  .= sprintf( ' background-color: %s; color: %s;', $background_color_style, $text_color_style );
+	}
+
```
edanzer, line 415: Is this above to enforce styling in emails?

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
edanzer: This markup works for me, and seems to relay styling changes from the block well. The left padding/margin is notable in emails. And it seems like there's a lack of space above/below button (I assume I could fix that by updating blocks/styling in the post itself). 

For the left padding, if we can fix that without otherwise breaking how you convey styles from the block to email, that would probably be worth it.

edanzer (approved):
I tested this alongside the phab diff. Testing as expected. Styles are relayed well from the block to email. I added a small comment about the block markup and the extra left padding. But I wouldn't consider that a blocker in any case.

simison:
Rebased :-)

PR #32412, opened 2023-08-10
Refactor/add products v1 to v2 proxied endpoint

chrismccluskey:
Fixes 106-gh-automattic/gold
Blocks https://github.com/Automattic/wp-calypso/pull/80205
Blocks D118667-code

## Proposed changes:

* Adds update and delete API endpoints for V2
* Proxies these requests to WPCOM from the Jetpack site
* Handles these requests if IS_WPCOM.

### Other information:

## Testing instructions:

This change involves WPCOM and requires access to a WPCOM sandbox to test.

There are three components to this issue:
- This patch for Jetpack which adds the "update" and "delete" endpoints for memberships plans to the backend. This patch could be tested by itself if it is downloaded to sandbox, public-api is directed to sandbox and the endpoints are called with the api developer console. 
- Calypso patch https://github.com/Automattic/wp-calypso/pull/80205 which calls the newly added "update" and "delete" endpoints from calypso. This patch could be tested by itself to see that endpoints were hit, but requires the Jetpack patch to function correctly - otherwise those endpoints would result in 404s, because they are defined in the Jetpack.
- WPCOM D118667-code which removes (now unused) the v1.1 endpoints from WPCOM. This patch could be tested by itself by patching sandbox, directing public-api to sandbox and using the api developer console to confirm the v1.1 endpoints for membership products no longer exist. Instead, this test plan uses this to help assert the v1.1 endpoints aren't being called because they no longer exist.

### Testing Non-Proxied Flow

Currently, the only flow available on production.

#### Testing from Calypso Localhost / Jetpack Source
1. Patch D118667-code and sandbox public API.
2. Checkout this branch locally `git fetch && git checkout origin refactor/add-products-v1-to-v2-proxied-endpoint`
3. Build from source with `pnpm jetpack build plugins/jetpack`
4. Sync to your D118667-patched sandbox with `pnpm jetpack rsync jetpack`
5. Make sure the public api is pointing to your sandbox
6. Clone the `wp-calypso` repo if you haven't already, and checkout `refactor/memberships-api-calls-to-v2`
7. Build calypso with `yarn` and then `yarn start`.
8. Open http://calypso.localhost:3000 in a browser and wait for calypso to finish building.
9. Once calypso loads, append `/earn/payment-plans` to the end of the url and hit enter. If you get "You are not authorized to view this page" click "switch site" and select your Jetpack site.
8. You should now be at the calypso.localhost version of `/earn/payment-plans/<your jetpack domain>`. If this is a new Jetpack site, you'll be greeted with an empty list of payment plans.
9. Click to add a new payment plan. The plan should add successfully.
10. Click to edit the created payment plan, make some adjustments and click save. The plan should save successfully and your edits should persist if you refresh the page.
11. Click to delete the created payment plan. The plan should delete, and if you refresh the page the plan should not re-appear in the list of payment plans.
12. Return to your Jetpack site and create a new post, this time with a payment button. Click the icon to select the payment plan and you should see a list of payment plans. Click add new payment plan and fill out the form, clicking save when done. The plan should exist when you return to the calypso.localhost page and view your payment plans.

#### Testing with Calypso Live / Jurassic Ninja

It's possible to test with Jurassic Ninja, but please note that currently no AJAX/XHR requests are expected to go to through to the Jurassic Ninja site's backend - instead all AJAX/XHR requests go to public-api directly from the frontend, meaning that the WPCOM sandbox will be processing the backend requests. Mostly, this prevents needing to build Jetpack and Calypso by source.

1. Patch D118667-code and sandbox public API.
2. Use the release branch in Jetpack Beta on your Jetpack site. In the upper-right corner of the wp-admin interface of your Jetpack site, click Jetpack Beta, then the first option, then "Manage". Paste the following into the Release branch option and click "Activate" when the result is found for: refactor/add-products-v1-to-v2-proxied-endpoint.
3. Download the patched version of Jetpack to your sandbox with `bin/jetpack-downloader test jetpack refactor/add-products-v1-to-v2-proxied-endpoint`.
4. On your Jetpack site, connect Jetpack to your WPCOM account, and add a new post with the "Subscribe" payment block.
5. Follow the prompts to setup a payment plan. Making note of if and when you are transferred to wordpress.com. If you see this, do not continue on wordpress.com - instead, continue to the next step, where we will use the patched Calypso to create a new payment plan.
6. Open calypso via the Calypso live links on https://github.com/Automattic/wp-calypso/pull/80205
7. Once calypso loads, append `/earn/payment-plans` to the end of the url and hit enter. If you get "You are not authorized to view this page" click "switch site" and select your Jetpack site.
8. You should now be at the calypso.live version of `/earn/payment-plans/<your jetpack domain>`. If this is a new Jetpack site, you'll be greeted with an empty list of payment plans.
9. Click to add a new payment plan. The plan should add successfully.
10. Click to edit the created payment plan, make some adjustments and click save. The plan should save successfully and your edits should persist if you refresh the page.
11. Click to delete the created payment plan. The plan should delete, and if you refresh the page the plan should not re-appear in the list of payment plans.
12. Return to your Jetpack site and create a new post, this time with a payment button. Click the icon to select the payment plan and you should see a list of payment plans. Click add new payment plan and fill out the form, clicking save when done. The plan should exist when you return to the calypso live page and view your payment plans.


chrismccluskey:
I don't believe that the failures in the "E2E Tests / Jetpack post-connection e2e tests (pull_request)" check are a result of these changes.

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/memberships.php
```diff
@@ -246,6 +287,130 @@ public function create_products( $request ) {
...
+
+			$result = $product->delete();
+			if ( is_wp_error( $result ) ) {
+				return array( 'error' => $result->get_error_message() );
+			}
+			return array( 'deleted' => true );
+		} else {
+			$blog_id  = Jetpack_Options::get_option( 'id' );
+			$response = Client::wpcom_json_api_request_as_user(
+				"/sites/$blog_id/{$this->rest_base}/product/$product_id/delete",
```
millerf: Here you should use "/sites/$blog_id/{$this->rest_base}/product/$product_id" and DELETE method, as defined https://github.com/Automattic/wp-calypso/pull/80205/files#diff-5e4abdd9e35ec1230c13cf42dcc2723cc52fd5025ec8e0be5370fcd736eed451R135

Also, it seems like you don't defined the DELETE endpoint here. I will try now, but not sure how this diff could work as is with the Calypso fix 🤔

millerf (changes requested):
I can see a few issues here. What prevents me accept this diff is:
-  it is missing the products list endpoint. See my comment on the Calypso diff.
- the definition for the product delete endpoint seems to also be missing (there is only the callback)

I also see a few issues in the "Jetpack" part of the "create" and "delete" endpoints here. I added some separate comments...

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/memberships.php
```diff
@@ -246,6 +287,130 @@ public function create_products( $request ) {
...
+			}
+
+			$updated_product = $product->update( $payload );
+			if ( is_wp_error( $updated_product ) ) {
+				return array( 'error' => $updated_product->get_error_message() );
+			}
+			return array( 'product' => $updated_product->to_array() );
+		} else {
+			$payload = array(
+				'type'     => $request['type'],
```
millerf: This should probably have all the possible parameters (`title`, etc...) as well...

millerf:
@chrismccluskey 

Side note about your testing instructions. If you test those on a Jurassic Ninja instance (or any self-hosted solution), sandboxing `public-api.wordpress.com` will not help testing: the instance will reach out to production, which is probably why it seems to work even with the issues I mentioned earlier.

### To test a self-hosted instance with your sandbox

The instance needs to route all "proxied" requests to your sandbox instead of `public-api.wordpress.com`.

- For that you need to go to `Jetpack constants`panel
Screenshot 2023-08-15 at 17 19 06

- Set to your sandbox URL 
Screenshot 2023-08-15 at 17 20 20

- And open your sandbox to non-proxied requests with
```
sudo global-http enable
```

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/memberships.php
chrismccluskey: Thanks, I'll refactor this. I think v1 used the `/delete` syntax, but you're right that v2 uses the `DELETE` method instead.

I think the tests were working because this block of code isn't actually hit in this flow - we can only delete products from calypso at the moment, which means `IS_WPCOM` will be true. This code was included so that we can proxy to WPCOM in the future if we add delete functionality to the gutenberg blocks and panels. As such, I don't have a manual test for this block yet.

chrismccluskey:
@millerf Thanks for the info about global http. Previously my tests were done using a local build of jetpack, so I think my local docker pulled the IP from my host machine's hosts file. I made some assumptions about all of this working with Jurassic Ninja. I'll test this way and modify the test plan for this PR accordingly.

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/memberships.php
chrismccluskey: This `delete_product` endpoint is registered around line 155 in this file:

```php
			array(
					'methods'             => WP_REST_Server::DELETABLE,
					'callback'            => array( $this, 'delete_product' ),
					'permission_callback' => array( $this, 'get_status_permission_check' ),
				),
```

millerf: > we can only delete products from calypso at the moment, which means IS_WPCOM will be true.

Yes, I figured that out later...

>  This code was included so that we can proxy to WPCOM in the future if we add delete functionality to the gutenberg blocks and panels

Yes, makes sense. I think we actually talked about it last week.

> This `delete_product` endpoint is registered around line 155 in this file:

Sorry. For some reason I totally missed it...

millerf (changes requested):
It is working fine now. This is overall some great work! And a big leap in learning Jetpack/Calypso overall structure.

I believe there are still a couple of parameters missing when editing the product. 

See the following video:

https://github.com/Automattic/jetpack/assets/790558/b7804dff-ce57-4907-bb94-a6abab577d38

Same issue happens when adding a product. 

https://github.com/Automattic/jetpack/assets/790558/19afe0f8-2156-4496-9eec-95d01b42503b

This is totally understandable as adding the product from Jetpack (in the payments block side panel) does not provide such options, and was therefore not  implemented originally...
 
Screenshot 2023-08-16 at 12 20 43



This is the object passed when creating
```
{
    "currency": "USD",
    "price": "45",
    "title": "Title",
    "interval": "1 year",
    "buyer_can_change_amount": true,
    "multiple_per_user": true,
    "welcome_email_content": "some message",
    "subscribe_as_site_subscriber": true,
    "type": "donation",
    "is_editable": true
}
```

And editing (for reference):
```
{
    "ID": 14,
    "currency": "USD",
    "price": "45.00",
    "title": "Title",
    "interval": "1 year",
    "buyer_can_change_amount": true,
    "multiple_per_user": true,
    "welcome_email_content": "",
    "subscribe_as_site_subscriber": true,
    "type": "donation",
    "is_editable": true
}
```

millerf:
@chrismccluskey Upon merging that, it will take one day for it to be pushed on WPCOM. But you can also create a patch and have it merged on WPCOM beforehand (which I would prefer).

chrismccluskey:
@millerf thanks for testing and for providing the payloads! I've added the missing properties. I've also created D119139-code which includes these changes.

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/memberships.php
```diff
@@ -353,12 +356,15 @@ public function update_product( \WP_REST_Request $request ) {
...
-			'type'                    => $type,
+			'title'                        => $request['title'],
+			'price'                        => $request['price'],
+			'currency'                     => $request['currency'],
+			'buyer_can_change_amount'      => $buyer_can_change_amount,
+			'interval'                     => $request['interval'],
+			'type'                         => $type,
+			'welcome_email_content'        => $request['welcome_email_content'],
+			'subscribe_as_site_subscriber' => $request['subscribe_as_site_subscriber'],
+			'multiple_per_user'            => $request['multiple_per_user'],
```
millerf, line 367: It looks perfect. 
As a side note, I usually don't like duplicated code, and I would have refactored some of that. I know some prefer explicit code to prevent complexity.

What is your point-of-view about that? (no need to change anything here, this is perfectly fine)

millerf (approved):
LGTM and works perfectly

PR #32432, opened 2023-08-11
Add block.json metadata file to beta blocks

### LINKED ISSUE #32408: Blocks: ensure visibility of Jetpack blocks in w.org profile
### Impacted plugin

Jetpack

### What

As a site owner, I would like to be able to see all of the blocks brought to my site via Jetpack on the WordPress.org page.

### How

This needs to be done via adding block.json files to Jetpack to match the blocks that we are including in our bespoke bundling process. While splitting this up to use the block.json file is blocked a bit by the performance hit and the current Gutenberg work for async loading, we can add the block.json files now and begin to gain value.
```[tasklist]
### Tasks
- [ ] https://github.com/Automattic/jetpack/pull/32187
- [ ] https://github.com/Automattic/jetpack/pull/32431
- [ ] https://github.com/Automattic/jetpack/pull/32432
- [ ] https://github.com/Automattic/jetpack/pull/32433
- [ ] https://github.com/Automattic/jetpack/pull/32497
- [ ] https://github.com/Automattic/jetpack/pull/32508
- [ ] https://github.com/Automattic/jetpack/pull/32536
```

*****

Primary issue: #17099
Internal reference: pedMtX-RS-p2

monsieur-z:
See #32408

## Proposed changes:

Add `block.json` metadata files for Jetpack beta blocks.

`v6-video-frame-poster`, `ai-assistant-form-support`, and `videopress/video-chapters` are Jetpack extensions but not blocks.

### Other information:

## Jetpack product discussion

pedMtX-RS-p2

Content is extracted from the metadata found in various files in the block folder.

No.

## Testing instructions:

Review the files content.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-chat/block.json
+++ b/projects/plugins/jetpack/extensions/blocks/ai-chat/block.json
@@ -1,0 +1,21 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/block.json",
+	"apiVersion": 3,
+	"name": "jetpack/ai-chat",
+	"title": "AI Chat (Experimental)",
+	"description": "Provides summarized chat across a site’s content, powered by AI magic.",
+	"keywords": [ "AI", "GPT", "Chat", "Search" ],
+	"version": "12.5.0",
+	"textdomain": "jetpack",
+	"category": "text",
+	"icon": "superhero",
+	"supports": {
+		"align": true,
+		"alignWide": true,
+		"customClassName": true,
+		"className": true,
+		"html": false,
+		"multiple": false,
+		"reusable": false
+	}
+}
--- a/projects/plugins/jetpack/extensions/blocks/ai-chat/block.json
+++ b/projects/plugins/jetpack/extensions/blocks/ai-chat/block.json
@@ -17,5 +17,6 @@
 		"html": false,
 		"multiple": false,
 		"reusable": false
-	}
+	},
+	"editorScript": "file:../editor-beta.js"
 }
--- a/projects/plugins/jetpack/extensions/blocks/amazon/block.json
+++ b/projects/plugins/jetpack/extensions/blocks/amazon/block.json
@@ -1,0 +1,17 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/block.json",
+	"apiVersion": 3,
+	"name": "jetpack/amazon",
+	"title": "Amazon",
+	"description": "Promote Amazon products and earn a commission from sales.",
+	"keywords": [ "amazon", "affiliate" ],
+	"version": "12.5.0",
+	"textdomain": "jetpack",
+	"category": "earn",
+	"icon": "amazon",
+	"supports": {
+		"align": true,
+		"alignWide": false,
+		"html": false
+	}
+}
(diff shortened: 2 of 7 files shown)


projects/plugins/jetpack/extensions/blocks/ai-chat/block.json
```diff
@@ -0,0 +1,21 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/block.json",
+	"apiVersion": 3,
```
jeherve, line 3: Since we support the current version of WP (6.3) minus 1, so currently WP 6.2, should we keep this to version 2, to be compatible with the previous version of WordPress?

projects/plugins/jetpack/extensions/blocks/amazon/block.json
```diff
@@ -0,0 +1,17 @@
...
+	"$schema": "https://schemas.wp.org/trunk/block.json",
+	"apiVersion": 3,
+	"name": "jetpack/amazon",
+	"title": "Amazon",
+	"description": "Promote Amazon products and earn a commission from sales.",
+	"keywords": [ "amazon", "affiliate" ],
+	"version": "12.5.0",
+	"textdomain": "jetpack",
+	"category": "earn",
+	"icon": "amazon",
```
jeherve, line 11: Will it work to provide a custom string that does not match a dashicon?

monsieur-z: It surprised me, but it is [a valid icon](https://developer.wordpress.org/resource/dashicons/#amazon).

projects/plugins/jetpack/extensions/blocks/ai-chat/block.json
monsieur-z: We can definitely do that to stay on the safer side. From [this document](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-api-versions/) it seems it might not be a concern, but I'm not sure I'm getting all of it.

jeherve: I suppose we can give it a try as v3 for now, and see what happens :)

jeherve (dismissed):

jeherve (approved):

simison:
@monsieur-z FYI we keep working on Paywall and Subscription blocks, so before you make these active, just double check everything is up-to-date for description, keywords etc! :-) 

I synced latest changes in https://github.com/Automattic/jetpack/pull/32551#discussion_r1298062225 and will try remember do so going forward too.

Question: how should we go about a custom SVG icon for the Paywall block? I see in `block.json` it's now empty. **Edit:** just saw the [icons PR](https://github.com/Automattic/jetpack/pull/32568). 👍

monsieur-z:
Thanks for handling this @simison! Having content in two places is not ideal, but we needed to add the metadata files for the block descriptions to appear in wp.org, which was a pressing issue.

Eventually, the goal is to make `block.json` files the go-to way to register blocks since it's the recommended way to do. We'll add to our roadmap having a single source of truth.

simison:
Yep! Looking forward to more standards way of loading blocks. :-)

Meanwhile you could import the `block.json` into JS, and read values from there? I'm just not sure what implications would be for translations.

```js
import { name, description, icon } from './block.json';

export const settings = {
	name,
	description,
	icon,
	// ...
}
```

PR #32436, opened 2023-08-11
Add Atomic Deploy E2E Tests to the CI workflow for updating atomic staging sites

dpasque:
Fixes #

## Proposed changes:
This updates the `update-jetpack-staging-sites.yml` GH workflow to include a step that triggers a E2E tests for the Atomic staging sites that have been updated. It makes a remote request that triggers a TeamCity build. 

BEFORE MERGING:

- [x] D118763-code -- the corresponding diff needs to be landed on the backend to receive the request.
- [x] I haven't actually added the secrets yet -- was going to wait until before merge to do that!
- [x] Does this need a CHANGELOG entry? From the docs I read here, it didn't seem like it because it's not part of a project, but I wanted to double check :) 
- [x] Get TeamCity run green (fixing tests, etc.)
- [x] Figure out why the request signature isn't working in the remote workflow execution.

### Other information:

- [ ] Have you written new tests for your changes, if applicable? N/A
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)? N/A

Nope!

## Testing instructions:
I'm actually not entirely sure how we want to test this, because it's part of a workflow run that's intended to be run at very discrete points, and I don't wait to go updating all those staging sites when people aren't expecting! 😅  Any suggestions?


kraftbj:
> Does this need a CHANGELOG entry? From the docs I read here, it didn't seem like it because it's not part of a project, but I wanted to double check :)

You're good. Anything inside the `projects` directory will, but the things outside aren't distributed/versioned.

samiff (commented):
I took a look at D118763-code and think you should be able to merge that at this point. Once that gets deployed, and the new env variables are added to the Jetpack repo, you should be able to manually dispatch the [Update Jetpack Staging Test Sites](https://github.com/Automattic/jetpack/actions/workflows/update-jetpack-staging-sites.yml) workflow to test this PR:

Markup on 2023-08-11 at 13:55:17

dpasque:
So, running the script locally with all the same values it's triggering correctly, but running the workflow on the GitHub server, it's rejecting the request signature. 🤔 Something must be getting borked in the secret value hydrating in. I'll pick this back up next week and see if I can find a secure way to debug what's going wrong there.

dpasque:
Also, because this is run every time there's a new Beta build available, let's definitely wait to merge this until the TeamCity run is green haha... 😅

.github/files/jetpack-staging-sites/run-atomic-deploy-e2e-tests.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+# This script is ran as part of the `UpdateJetpackStaging` TeamCity build.
+# It triggers an E2E test run against pre-defined atomic sites that have been updated with the newest Jetpack build.
+
+if [ -z "$SIGNATURE_KEY" ] || [ -z "$TRIGGER_URL" ]; then
```
anomiex: Let's prefer Bash syntax since we're using Bash.
```suggestion
if [[ -z "$SIGNATURE_KEY" || -z "$TRIGGER_URL" ]]; then
```

.github/files/jetpack-staging-sites/run-atomic-deploy-e2e-tests.sh
```diff
@@ -0,0 +1,20 @@
...
+  echo "Missing required ENV variables: both 'SIGNATURE_KEY' and 'TRIGGER_URL' are required. Aborting."
+  exit 1
+fi
+
+echo "Starting atomic deploy E2E tests..."
+
+REQUEST_SIGNATURE=$(echo -n | openssl dgst -sha256 -hmac "$SIGNATURE_KEY")
+curl --fail-with-body -s -X POST -H "X-Jetpack-Atomic-Deploy-E2E-Signature: $REQUEST_SIGNATURE" "$TRIGGER_URL"
```
anomiex: Looks like the command you're using here produces a value for `REQUEST_SIGNATURE` along the lines of "SHA2-256(stdin)= b613679a0814d9ec772f95d778c35fc5ff1697c493715653c6c712144292c5ad", which of course doesn't work due to that "SHA2-256(stdin)=" prefix.

Something like this should work
```suggestion
REQUEST_SIGNATURE=$(echo -n | openssl mac -digest sha256 -macopt "key:$SIGNATURE_KEY" HMAC)
curl --fail-with-body -s -X POST -H "X-Jetpack-Atomic-Deploy-E2E-Signature: ${REQUEST_SIGNATURE,,}" "$TRIGGER_URL"
```

.github/files/jetpack-staging-sites/run-atomic-deploy-e2e-tests.sh
```diff
@@ -0,0 +1,20 @@
...
+
+if [ $? -eq 0 ]; then
```
anomiex:
```suggestion
if [[ $? -eq 0 ]]; then
```

Or, for that matter, you could just do like
```
if curl ...; then
```

.github/files/jetpack-staging-sites/run-atomic-deploy-e2e-tests.sh
dpasque: Ahhh, thank you so much for catching that! 🙂  That's really weird, because I ran this in Bash on my local system and wasn't seeing the prefix attached in that value. 🤔  But that would for sure explain why that wasn't working here!

dpasque:
The different hashing command was the ticket (thank you so much @anomiex!) and it looks like the trigger is working now: 
https://github.com/Automattic/jetpack/actions/runs/5893484233

dpasque:
We're also pretty close to the Atomic tests being green -- we're just waiting on a few more PR fix merges! 👍 Once those are running green, I'll make an update here.

.github/files/jetpack-staging-sites/run-atomic-deploy-e2e-tests.sh
anomiex: Different version of openssl maybe? I have 3.0.10, and the Actions runner probably has [3.0.2](https://packages.ubuntu.com/jammy/openssl).

dpasque: Ahh yeah, that's probably it, good point!

dpasque:
Huzzah! The TeamCity build is now running green, so I think it should be safe to merge this trigger. Does it seem ready to go? cc @anomiex / @samiff

anomiex (approved):
Let's go for it!

Don't forget to update PCYsg-xsv-p2#github-secrets with the info about the new secrets.

PR #32437, opened 2023-08-11
 CRM: Automations Contact transitional status condition

gogdzl:
## Proposed changes:

* This PR adds the contact transitional status condition and respective tests.
   * Operators:
      * from_to

### Other information:

## Jetpack product discussion
https://github.com/Automattic/zero-bs-crm/issues/3190

No.

## Testing instructions:

### DIFF
--- a/projects/plugins/crm/src/automation/commons/conditions/contacts/class-contact-transitional-status.php
+++ b/projects/plugins/crm/src/automation/commons/conditions/contacts/class-contact-transitional-status.php
@@ -83,6 +83,12 @@ class Contact_Transitional_Status extends Base_Condition {
 		$status_was = $this->get_attributes()['previous_status_was'];
 		$status_is  = $this->get_attributes()['new_status_is'];
 
+		if ( ! $this->is_valid_operator( $operator ) ) {
+			$this->condition_met = false;
+			$this->logger->log( 'Invalid operator: ' . $operator );
+			throw new Automation_Exception( 'Invalid operator: ' . $operator );
+		}
+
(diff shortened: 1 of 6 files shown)


projects/plugins/crm/src/automation/commons/conditions/contacts/class-contact-transitional-status.php
```diff
@@ -0,0 +1,182 @@
...
+				$this->condition_met = ( $data['old_status_value'] === $status_was ) && ( $data['contact']['data']['status'] === $status_is );
+				$this->logger->log( 'Condition met?: ' . ( $this->condition_met ? 'true' : 'false' ) );
+
+				return;
+		}
+
+		$this->condition_met = false;
+		$this->logger->log( 'Invalid operator: ' . $operator );
+
+		throw new Automation_Exception( 'Invalid operator: ' . $operator );
```
robertf4: This could be a check at the start of the function like the contact status transitional data check and use `$valid_operators` to explicitly check that the operator is valid, and then have a default case in the `operator_switch` which logs something saying that we have a valid but unimplemented operator.

robertf4 (commented):
Looks good, just had one minor comment.

projects/plugins/crm/src/automation/commons/conditions/contacts/class-contact-transitional-status.php
gogdzl: Thanks for the suggestion! Added in 594db33.

robertf4 (approved):
This looks good to me!

PR #32438, opened 2023-08-11
Make AI assistant input receive focus

CGastrell:
This PR is a wee change to set focus on the assistant input on render.

## Proposed changes:
The change is a small one, just adding a useEffect on the input ref to imperatively set focus on it.

While simple in concept, if you're engaging the editor with your keyboard, it is a bit counter intuitive to have to reach the mouse only to focus on the input and start typing what you need from the assistant.

Before:
no-focus

After:
with-focus

### Other information:

No

## Testing instructions:
Checkout and `jetpack build plugins/jetpack`. Insert a form block (which, in particular was stealing focus from the variation picker) or any other block and enable the assistant.
Focus should be immediately in the assistant input.

### DIFF
--- a/projects/js-packages/ai-client/changelog/update-ai-assistant-input-focus
+++ b/projects/js-packages/ai-client/changelog/update-ai-assistant-input-focus
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Make the AI Assistant input to receive focus on render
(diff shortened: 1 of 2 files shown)


renatoagds:
Just pushing back what was discussed in Slack: p1691782497118009-slack-C054LN8RNVA

I would say that we shouldn't do it, since it's not a good direction for accessibility. We already have a [lint rule](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/f0d2ddb65f21278ad29be43fb167a1092287b4b1/docs/rules/no-autofocus.md) to not use `autoFocus` in HTML. But I would like to have a opinions from our Designers.

cc/ @ederrengifo and @madebynoam

CGastrell:
Closing in favor of #32447 as the implementation was moved to the plugin extension

PR #32443, opened 2023-08-11
CRM Automations: Add workflow table

robertf4:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3210

## Proposed changes:

This PR adds a table component to display many workflows in a table.

### Other information:

## Jetpack product discussion

## Testing instructions:

1. There is no functionality yet, so primarily review the code.
2. Put `add_filter( 'jetpack_crm_feature_flag_automations', '__return_true' );` in an mu-plugin.
3. Visit `/wp-admin/admin.php?page=jpcrm-automations#/automations/test-workflow-table`. Verify that the visuals match the figma designs.


gogdzl (changes requested):
This PR's table:
image

Figma:
image

Should the headers match 100% the Figma file?
Also, the horizontal spacing looks different.
[edit] Horizontal spacing from the checkboxes to the first collumn.

gogdzl:
Some more issues:

- In the Figma file the header has `studio-gray-60` color.
- There is no sorting indicator in the column headers (I even think that they should be clickable)

projects/plugins/crm/src/js/components/automations-admin/components/checkbox/index.tsx
```diff
@@ -1,25 +1,21 @@
-import classnames from 'classnames';
 import styles from './styles.module.scss';
 
 type CheckboxProps = {
 	id: string;
 	checked?: boolean;
-	decorative?: boolean;
```
robertf4, line 7: Removing this as I misunderstood the design 😅

robertf4:
@gogdzl I implemented the sorting and addressed the other issues, and this one is ready for another look.

robertf4:
Note that sorting by status won't work until the store is implemented because the `active` prop won't actually be updated until then.

gogdzl:
This PR's table:
image


Figma:
image

Looks good!
Small nitpick would be to match the fonts of the titles, but if that is too complicated it already looks good to me!

gogdzl (approved):

PR #32447, opened 2023-08-13
AI Extension: make forms ai assistant input receive focus on first mount

CGastrell:
This PR is a wee change to set focus on the assistant input on render. It's a re-do from #32438 since it moved from the JS package to the plugin to address the specific case of forms assistant.

## Proposed changes:
The change is a small one, just adding a useEffect on the input ref to imperatively set focus on it.

While simple in concept, if you're engaging the editor with your keyboard, it is a bit counter intuitive to have to reach the mouse only to focus on the input and start typing what you need from the assistant.

For a11y reasons the focus is set only on first appearance, toggling the assistant off/on won't trigger the focus again.

Before:
no-focus

After:
with-focus

### Other information:

No

## Testing instructions:
Checkout and `jetpack build plugins/jetpack`. Insert a form block (which, in particular was stealing focus from the variation picker).
Focus should be immediately in the assistant input. Toggle the assistant on/off, the focus should not be set on the input anymore.

### DIFF
--- a/projects/plugins/jetpack/changelog/update-ai-assistant-input-focus
+++ b/projects/plugins/jetpack/changelog/update-ai-assistant-input-focus
@@ -1,4 +1,4 @@
 Significance: minor
-Type: added
+Type: other
 
-Make the AI Assistant input to receive focus on render
+Make the AI Assistant input to receive focus on first render for forms variations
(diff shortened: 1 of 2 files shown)


retrofox (approved):
LGTM. It works as expected.
We need to keep polishing the initial render of the bar, keeping in mind it opens as default when creating the block instance.
🚀

PR #32453, opened 2023-08-14
Super Cache: Fix deprecation notices

### LINKED ISSUE #32336: [Super Cache] More PHP 8 warnings
We have [new reports of PHP 8 warnings](https://wordpress.org/support/topic/php-8-1-php-deprecated-strstr/#post-16929870). These include:

```
[28-Jul-2023 11:49:59 UTC] PHP Deprecated: substr(): Passing null to parameter #1 ($string) of type string is deprecated in /home/wp-content/plugins/wp-super-cache/wp-cache-phase2.php on line 227

[28-Jul-2023 11:49:59 UTC] PHP Deprecated: strtolower(): Passing null to parameter #1 ($string) of type string is deprecated in /home/wp-content/plugins/wp-super-cache/wp-cache-phase2.php on line 808
```

These should be quick fixes to squeeze in, in advance of the 1.10 release.

haqadn:
Fixes #32336

## Proposed changes:

* Ensure `$wp_cache_request_uri` is always set to a string.

### Other information:

## Testing instructions:

None


donnchawp (changes requested):
I confirmed that `$_SERVER['REQUEST_URI']` is NULL in a script run from the command line, so that is likely where the null value is coming from.

I saw in `get_wp_cache_key` in wp-cache-phase2.php (around line 40) that we fixed the same problem by checking for null. I think we can probably remove that check now with this fix.

donnchawp (approved):
Looks great. I checked where $wp_cache_request_uri is used and it's good.

donnchawp:
deb8585fed7941dab2e30d76bd218be2c5eed10a

A-Printer:
@donnchawp The removal of the `if ( null === $url )` check causes a Deprecated notice to be thrown on the frontend when logged as an admin:

Capture d’écran 2023-08-18 à 13 26 52

Also, the Deprecated notice causes the REST API check to fail on the Site Health page:

Capture d’écran 2023-08-18 à 11 56 20

donnchawp:
Fixed in https://github.com/Automattic/jetpack/pull/32629

Issue #31638, opened 2023-06-29
WordPress 6.3 compatibility


jeherve:
WordPress 6.3 is set to release on August 8, 2023.

- Planning: [make.wordpress.org/core/6-3](https://make.wordpress.org/core/6-3/)

Let's use this issue to keep track of anything that may come up until then.

- [x] Update our "tested up to" version to 6.3 #31910
- [x] Update the version of WordPress required for Jetpack from WP 6.1 to WP 6.2:  ~**On hold for now, waiting for 12.6**~ #32762 
- [x] Remove 6.1 backward-compatible checks ~**On hold for now, waiting for 12.6**~ #32772

In addition to the above, we can follow the dev-notes that will drop here about the release:
[make.wordpress.org/core/tag/dev-notes+6-3](https://make.wordpress.org/core/tag/dev-notes+6-3)

- [x] [Layout updates](https://make.wordpress.org/core/2023/07/14/layout-updates-in-the-editor-for-wordpress-6-3/): changes in CSS specificity
- [x] [WP_Query used internally in get_pages()](https://make.wordpress.org/core/2023/07/14/wp_query-used-internally-in-get_pages/)
- [ ] [Improved Caching for Database Queries in WP_User_Query](https://make.wordpress.org/core/2023/07/14/improved-caching-for-database-queries-in-wp_user_query/)
- [x] [Image performance enhancements in WordPress 6.3](https://make.wordpress.org/core/2023/07/13/image-performance-enhancements-in-wordpress-6-3/): ensure that those enhancements do not conflict with our Lazy Images implementation.

Below, let's keep track of things we can start implementing once WordPress 6.3 is the minimum required version, in #32865.

- #32852
- #32853
- #32854
- #32863
- #32864

-------

For reference: prior art for WordPress 6.2 compat: #27795

jeherve:
> [Improved Caching for Database Queries in WP_User_Query](https://make.wordpress.org/core/2023/07/14/improved-caching-for-database-queries-in-wp_user_query/)

@Automattic/jetpack-kernel Do you think this is something that may impact us in sync?

sergeymitr:
Thanks for the ping @jeherve, there might be an issue here.

[...]

PR #32454, opened 2023-08-14
IS: add conditional to use new method introduced in WP 6.3

jeherve:
## Proposed changes:

`WP_Scripts::print_inline_script` was deprecated in WordPress 6.3, in favor of `WP_Scripts::get_inline_script_data`. Let's start using that to avoid deprecation notices, when we're running WP 6.3 or above.

See https://core.trac.wordpress.org/ticket/12009

### Other information:

## Jetpack product discussion

Primary issue: #31638

No

## Testing instructions:

Start with a site with more than 10 posts.

* Enable a classic theme that bundles support for Infinite Scroll, such as "Superhero".
* Go to Jetpack > Settings and enable Infinite Scroll (e.g. set it to "Load more posts as the reader scrolls down").
* Load the home page and scroll down
    * You should see more posts appearing as you scroll down the page.

### DIFF
--- a/projects/plugins/jetpack/changelog/update-infinite-scripts
+++ b/projects/plugins/jetpack/changelog/update-infinite-scripts
@@ -1,0 +1,4 @@
+Significance: patch
+Type: compat
+
+Infinite Scroll: avoid PHP notices when using the latest version of WordPress, 6.3.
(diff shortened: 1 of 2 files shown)


samiff (approved):
Worked for me 👍

PR #32464, opened 2023-08-14
Blocks: use UI dark color to ensure text is readable

### LINKED ISSUE #5036: Varia: Markdown + half-adopted editor styles = invisible markdown editing
## Steps to replicate
- Use a site that has a Varia theme
- Set the background color to something dark, and the text to something light.
- Start a new post or page
- Type /markdown to start a new Markdown block
- Add a little sample markdown, like this:

`# Hello World!

Programmed to work and not to feel
Not even sure that this is real
- Louie Zong`

So far I've confirmed this in Hever and Varia itself, and confirmed Zoologist does not suffer from this issue.

## Result
While editing the Markdown, the block editor shows an input that takes on the foreground (text) color, but forces a white background. So if I'm using a dark color scheme, I can't see much:

Screen Shot 2021-11-15 at 2 55 21 PM

## Expected
Either the Markdown block should use its own style, or it should accept the both the background and foreground color for the site. Maybe the background color for input in Varia is a bit too opinionated, since other themes seem fine.

* User report: Currently just a Twitter DM; [pb here](2ab6f-pb/)
* Follow-up ticket: I've asked the user to create a post in the forums so we can follow up there, but if necessary, we can use the link above to reply.

### LINKED ISSUE #19285: White background and white text when editing markdown block

1. Create new post
2. Add markdown block
3. Attempt to edit text

#### What I expected

The block to respect my [website's theme](https://constellations.scot/).

#### What happened instead

The block when editing text turns white, which combined with a white font makes it impossible to see anything. When the block does not have focus it respects my theme - and when posted the post respects my theme. It is only when in edit mode and the block has focus that this issue occurs.

image

jeherve:
Fixes #19285
Fixes https://github.com/Automattic/themes/issues/5036

## Proposed changes:

- This fixes an issue with the Markdown block when using a theme with dark background colors and white text.
- It also updates the Repeat Visitor block to stop using non-available colors, and use colors from the Gutenberg shared styles instead.

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

1. Go to Appearance > Themes and switch to a block theme like Twenty Twenty Three.
2. Go to Appearance > Site Editor > Styles
3. Switch to a style with a dark background and white text. You can edit a style to define those colors too.
4. Go to Posts > Add New
5. Add a Markdown block. You should be able to type in some text, and you should see your text clearly (dark color on white background).
6. Add a Repeat Visitor block; the notice should be visible to you (dark text on a white background).


samiff (approved):
Tested well 👍

PR #32472, opened 2023-08-15
 CRM: Automations Invoice field contains condition

gogdzl:
## Proposed changes:

* This PR adds the invoice field contains condition and respective tests.
   * Operators:
      * contains
      * does_not_contain

This PR also fixes https://github.com/Automattic/zero-bs-crm/issues/3279 by making the check of operator validity generic.

### Other information:

## Jetpack product discussion
https://github.com/Automattic/zero-bs-crm/issues/3201

No.

## Testing instructions:


projects/plugins/crm/src/automation/commons/conditions/invoices/class-invoice-field-contains.php
```diff
@@ -0,0 +1,187 @@
...
+			case 'does_not_contain':
+				$this->condition_met = ( strpos( $data['data'][ $field ], $value ) === false );
+				$this->logger->log( 'Condition met?: ' . ( $this->condition_met ? 'true' : 'false' ) );
+
+				return;
+		}
+		$this->condition_met = false;
+		$this->logger->log( 'Invalid operator: ' . $operator );
+
+		throw new Automation_Exception( 'Invalid operator: ' . $operator );
```
robertf4: This could be a check at the beginning of the function using `$valid_operators`, and there could be a default switch case which throws an "unimplemented operator exception".

projects/plugins/crm/src/automation/commons/conditions/invoices/class-invoice-field-contains.php
```diff
@@ -0,0 +1,187 @@
...
+	/**
+	 * Checks if the invoice has at least the necessary keys to detect if a field
+	 * contains some value.
+	 *
+	 * @since $$next-version$$
+	 *
+	 * @param array $invoice_data The invoice data.
+	 * @return bool True if the data is valid to detect if a field contains some value, false otherwise
+	 */
+	private function is_valid_invoice_field_contains_data( array $invoice_data ): bool {
```
robertf4, line 97: I noticed that this pattern is present in these classes. I'm wondering if there could be a `$valid_fields` property which is abstract in the `Base_Condition` class and then there could be a generic implementation to do this functionality which would be called something like `valid_fields_have_data()` that uses that `$valid_fields` property to check that all such fields are set. Then each implementation of `Base_Condition` wouldn't have to repeat this implementation.

projects/plugins/crm/src/automation/commons/conditions/invoices/class-invoice-field-contains.php
gogdzl: Opened a new issue to make this generic: https://github.com/Automattic/zero-bs-crm/issues/3279

gogdzl: Added to this PR's description that I'm also fixing this issue.

projects/plugins/crm/src/automation/commons/conditions/invoices/class-invoice-field-contains.php
gogdzl: We will have to open an issue and make this generic way of checking outside of this PR, because it will involve some decisions about the multidimensional nature of $invoice_data.

projects/plugins/crm/src/automation/class-base-condition.php
```diff
@@ -76,4 +93,23 @@ public function condition_met(): bool {
...
+	 * @param string $operator The operator.
+	 * @return void
+	 *
+	 * @throws Automation_Exception If the operator is invalid for this condition.
+	 */
+	protected function check_for_valid_operator( string $operator ): void {
+		if ( ! in_array( $operator, $this->valid_operators, true ) ) {
+			$this->condition_met = false;
+			$this->logger->log( 'Invalid operator: ' . $operator );
+			throw new Automation_Exception( 'Invalid operator: ' . $operator );
```
cleacos: Good improvement in adding this generic method here. Maybe we should add a new exception code: `CONDITION_INVALID_OPERATOR`

projects/plugins/crm/src/automation/commons/conditions/invoices/class-invoice-field-contains.php
```diff
@@ -0,0 +1,165 @@
...
+
+	/**
+	 * Get the category of the invoice field contains condition.
+	 *
+	 * @since $$next-version$$
+	 *
+	 * @return string The category 'jpcrm/invoice_condition'.
+	 */
+	public static function get_category(): string {
+		return 'invoice_condition';
```
cleacos:
```suggestion
		return 'invoice';
```

projects/plugins/crm/src/automation/commons/conditions/invoices/class-invoice-field-contains.php
```diff
@@ -0,0 +1,165 @@
...
+	public static function get_type(): string {
+		return 'condition';
+	}
+
+	/**
+	 * Get the category of the invoice field contains condition.
+	 *
+	 * @since $$next-version$$
+	 *
+	 * @return string The category 'jpcrm/invoice_condition'.
```
cleacos:
```suggestion
	 * @return string The category 'jpcrm/invoice'.
```

projects/plugins/crm/src/automation/commons/conditions/invoices/class-invoice-field-contains.php
cleacos: We are using high-level categories like: `contact`, `invoice`, `company`, ...

Maybe an improvement should be to allow multiple categories... returning an array.

cleacos (approved):
Test passed. I added some commits adding the code exception for the condition operator.

PR #32473, opened 2023-08-15
CRM: Automations Quote condition for status changed

gogdzl:
## Proposed changes:

* This PR adds the quote condition 'Status Changed' and respective tests.
* This PR also introduces a new function `jpcrm_get_quote_statuses` that returns quote statuses strings.

### Other information:

## Jetpack product discussion
https://github.com/Automattic/zero-bs-crm/issues/3192

No.

## Testing instructions:


projects/plugins/crm/src/automation/commons/conditions/quotes/class-quote-status-changed.php
```diff
@@ -0,0 +1,188 @@
...
+			case 'is_not':
+				$this->condition_met = ( $data['data'][ $field ] !== $value );
+				$this->logger->log( 'Condition met?: ' . ( $this->condition_met ? 'true' : 'false' ) );
+
+				return;
+		}
+		$this->condition_met = false;
+		$this->logger->log( 'Invalid operator: ' . $operator );
+
+		throw new Automation_Exception( 'Invalid operator: ' . $operator );
```
robertf4: I have the same suggestion here as the previous PR I reviewed: check this at the beginning of the function using `$valid_operators` and then have a default case in the switch that throws an "unimplemented operator exception".

cleacos (changes requested):
The quote model doesn't have the `status` property. Determining the virtual status of a quote requires performing some checks with the accepted field.

gogdzl:
> The quote model doesn't have the status property. Determining the virtual status of a quote requires performing some checks with the accepted field.

Thanks a lot for pointing that out! Sometimes quotes do have a status field, but just in some cases ( e.g. sorting ), and they are indeed calculated in the code.

I added a fix for that and everything should be corrected.

projects/plugins/crm/src/automation/commons/conditions/quotes/class-quote-status-changed.php
```diff
@@ -0,0 +1,170 @@
...
+
+	/**
+	 * Get the category of the quote status changed condition.
+	 *
+	 * @since $$next-version$$
+	 *
+	 * @return string The category 'jpcrm/quote_condition'.
+	 */
+	public static function get_category(): string {
+		return 'quote';
```
cleacos:
```suggestion
		return __( 'quote', 'zero-bs-crm' );
```

cleacos (dismissed):
Cool. The code looks good, and the tests passed. I committed a small change (category translation)

LGTM!

robertf4 (approved):
I approve this PR, it is the best!

PR #32391, opened 2023-08-10
e2e tests: updates to the recommendation assistant tests

jeherve:
Fixes #32467

## Proposed changes:

Making some changes to the tests so they start passing again.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

Internal reference: p1691598965621699-slack-C03QNBQKG73

## Does this pull request change what data or activity we track or use?

No

## Testing instructions:

Do the tests pass?

PR #32475, opened 2023-08-15
Social: Add SIG toggle for Jetpack Settings

gmjuhasz:
This change adds the ability to be able to toggle SIG and select default templates from the Jetpack Settings page. This is needed so people with only the main Jetpack plugin can access this setting too.

CleanShot 2023-08-15 at 12 45 44 png

## Proposed changes:

* Restructured the SIG components a bit
* Registering the Social store is now happening in `publicize-components`
* I added 2 new components for opening the template picker with a button, and a generic toggle for SIG. We can use these on the Social admin page too in a next change
* I added these components into _inc under the main Publicize toggle.

### Other information:

## Jetpack product discussion

1205142973434174-as-1203857426851211

No.
## Testing instructions:

* Go to `wp-admin/admin.php?page=jetpack#/sharing` on your site, and compare the looks with the image above
* You should see the SIG toggle and the template picker only if
  * You have the publicize module enabled with the main toggle
  * You have the Advanced plan - the only way currently to have the feature
* Test the functionality
  * Toggling the state you work, and stay the same on refresh
  * Changing templates should stay the same on refresh
  * You can open the Social admin page too, change it there, then refresh the Jetpack settings page to see if the changes went throug.

* A generic sanity testing would be nice too, as the store structure is changed a bit.
  * Create a new post and publicize it.
  * Play with the settings in the editor, you should see no errors.


projects/plugins/social/src/js/index.js
```diff
@@ -1,11 +1,7 @@
 import { ThemeProvider } from '@automattic/jetpack-components';
-import { SOCIAL_STORE_ID, SOCIAL_STORE_CONFIG } from '@automattic/jetpack-publicize-components';
-import { createReduxStore, register } from '@wordpress/data';
 import * as WPElement from '@wordpress/element';
 import React from 'react';
 import AdminPage from './components/admin-page';
-const store = createReduxStore( SOCIAL_STORE_ID, SOCIAL_STORE_CONFIG );
-register( store );
```
gmjuhasz, line 8: Now this gets initialised with using the publicize-components package

projects/plugins/jetpack/changelog/add-sig-toggle-jetpack-settings
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: major
+
+Added SIG toggle for Jetpack Settings
```
jeherve: SIG may not mean much to most folks. How about something like:

```suggestion
Dashboard: add Social Image Generator options to Jetpack Social settings.
```

projects/plugins/jetpack/changelog/add-sig-toggle-jetpack-settings
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: major
```
jeherve: I don't know that I would classify this as a major enhancement. I think enhancement would be enough.

```suggestion
Type: enhancement
```

projects/plugins/jetpack/_inc/lib/admin-pages/class-jetpack-redux-state-helper.php
```diff
@@ -241,6 +241,28 @@ function_exists( 'wp_theme_has_theme_json' ) ? wp_theme_has_theme_json() : WP_Th
...
 			/** This filter is documented in plugins/jetpack/modules/subscriptions/subscribe-module/class-jetpack-subscribe-module.php */
 			'isSubscriptionModalEnabled'  => apply_filters( 'jetpack_subscriptions_modal_enabled', false ),
+			'socialInitialState'          => self::get_publicize_initial_state(),
+		);
+	}
+
+	/**
+	 * Gets the initial state for the Publicize module.
+	 *
+	 * @return mixed|null
```
jeherve: Should we change this here?

```suggestion
	 * @return array|null
```

jeherve (changes requested):
This tests well and looks good, nice work! I only left some minor nitpicks.

projects/plugins/jetpack/_inc/lib/admin-pages/class-jetpack-redux-state-helper.php
gmjuhasz: Nice catch thanks!

jeherve (approved):

jeherve:
I see now that the Social e2e tests are failing. It may be good to make sure it's not caused by this PR?

gmjuhasz:
> I see now that the Social e2e tests are failing. It may be good to make sure it's not caused by this PR?

Yeah, that was some flaky tests I think, now it's only the Jetpack post-connection test is failing due to timeout, which is also flaky

jeherve:
> now it's only the Jetpack post-connection test is failing due to timeout, which is also flaky

I'm working on this in #32391, hopefully that will help!

PR #32477, opened 2023-08-15
e2e config: add a new test site

jeherve:
## Proposed changes:

In this PR, we create a new VIP e2e test suite, targetting an existing VIP test site that always runs the latest version of `trunk`.

### Other information:

## Jetpack product discussion

* pdWQjU-rN-p2

* No

## Testing instructions:

* Is CI happy?


adimoldovan (approved):
Looks good!

adimoldovan:
Hi @jeherve, I noticed the run against the new site [didn't work](https://github.com/Automattic/jetpack/actions/runs/5878263819/job/15940006837#step:11:18). I believe the problem is the `TEST_SITE` env variable not being set and the local Docker env was being targeted instead. This should have been set to the name of the config entry for the site, I believe `vip` in this case. [Here's](https://github.com/Automattic/jetpack/blob/fa9330974af9b7d026e9e7d3e234754cc41a488b/.github/workflows/e2e-tests.yml#L159) where it's being set for atomic.

jeherve:
@adimoldovan Thank you! I just opened #32509 to try to add that.

PR #32478, opened 2023-08-15
Forms block: allow transforming to a subscribe block

simison:
## Proposed changes:

* Add a simple transform from Forms block to a Subscription block

Screenshot 2023-08-15 at 15 29 06

Screenshot 2023-08-15 at 15 29 20

### Other information:

## Testing instructions:

* Build Forms package locally
* Add Forms block
* Try converting it to subscribe block

### DIFF
--- a/projects/packages/forms/changelog/update-form-subscribe-block
+++ b/projects/packages/forms/changelog/update-form-subscribe-block
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Forms block - allow transforming to a subscribe block.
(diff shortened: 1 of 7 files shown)


simison:
FYI also considered transforming into the "button under input" version because it's closer to how forms looks originally, but honestly I think we should be more opinionated about it and not do that. The single-input design (vs multi input) works much better with button on the same line, so that's what we should transform to.

image

jeherve (approved):
That's quite a difference between a full-fledged form and a subscribe form, but maybe it can be helpful for some. 👍

PR #32479, opened 2023-08-15
Boost: Fix not deferring concatenated stylesheets when using Critical CSS

### LINKED ISSUE #32449: Boost: Concatenated CSS not deferring w/ Critical CSS
### Impacted plugin

Boost

### Quick summary

When Critical CSS is enabled, stylesheets on the page are deferred by setting their `media` attribute to `not all` until they have loaded, at which point they are enabled using JavaScript.

However, it looks like concatenated CSS is not being deferred, as the `<link>` tag used has `media="all"` on it in the page source.

e.g.:
```
<link data-handles='wp-block-navigation,jetpack_css' rel='stylesheet' id='all-css-6e4a05cbae8f522e796180c2ea011fed' href='https://example.com/_jb_static/??/wp-includes/blocks/navigation/style.min.css,/wp-content/plugins/jetpack/css/jetpack.css?m=1691982771&cb=1' type='text/css' media='all' />
```

This will undermine the speed benefit of Critical CSS on some sites.

### Steps to reproduce

1. Create a JN test site with Boost installed
2. Turn on Critical CSS and Concatenate CSS
3. Look at the page source of your home page. Make sure Critical CSS is present.
4. Look for a concatenated CSS link tag in your source, like the example above.
5. You will note that it is `media="all"`, which will cause it to be a render-blocker, despite having Critical CSS on the page.

Concatenated CSS files should be deferred when Critical CSS is enabled.

### What actually happened

Concatenated CSS files were not deferred.

### Impact

Some (< 50%)

### Available workarounds?

Yes, easy to implement

_No response_

### Logs or notes

_No response_

dilirity:
Fixes https://github.com/Automattic/jetpack/issues/32449

Some background - this got removed in https://github.com/Automattic/jetpack/pull/31058 and that caused concatenated stylesheets to no longer be deferred. This PR adds it back, but changes the handles from an array, to a comma separated string, to prevent plugins using the filter from complaining.

Not sure if this is the best thing to do, but it was the cleanest.

Maybe we shouldn't use the filter, but call https://github.com/Automattic/jetpack/blob/3104e377ae2d45ad2d4ffadfd839cee47aa871df/projects/plugins/boost/app/lib/critical-css/Display_Critical_CSS.php#L34 if Critical/Cloud CSS is enabled, as the concatenated stylesheet shouldn't be modified by anything other than Boost?

## Proposed changes:

* Add a filter to allow Critical/Cloud CSS logic to hook onto and prepare the concatenated stylesheet tag for deferring.

### Other information:

## Jetpack product discussion

n/a

no

## Testing instructions:

* Follow the steps listed in the "Steps to reproduce" section of https://github.com/Automattic/jetpack/issues/32449

### DIFF
--- a/projects/plugins/boost/app/lib/minify/Concatenate_CSS.php
+++ b/projects/plugins/boost/app/lib/minify/Concatenate_CSS.php
@@ -194,6 +194,10 @@ class Concatenate_CSS extends WP_Styles {
 
 				$style_tag = apply_filters( 'page_optimize_style_loader_tag', $style_tag, $handles, $href, $media );
 
+				// Allow manipulation of the stylesheet tag.
+				// For example - making it deferred when using with Critical CSS.
+				$style_tag = apply_filters( 'style_loader_tag', $style_tag, implode( ',', $handles ), $href, $media );
+
 				// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
 				echo $style_tag . "\n";
 
--- a/projects/plugins/boost/changelog/fix-not-deferring-concatenated-stylesheets
+++ b/projects/plugins/boost/changelog/fix-not-deferring-concatenated-stylesheets
@@ -1,4 +1,4 @@
-Significance: minor
+Significance: patch
 Type: fixed
 
 Fixed not being able to defer concatenated stylesheets, causing them to be render-blocking when used with Critical CSS.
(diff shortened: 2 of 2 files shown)


pyronaur (approved):
👍

haqadn:
d448e958fd

PR #32480, opened 2023-08-15
Subscribe Modal: Don't pre register the template part

lezama:
We haven't found a stable way of registering template parts from a plugin, let's not register the template until we have a solid solution.

## Proposed changes:

* Removes the registration of the modal template part into template parts list.
* Still allows users to define a custom `jetpack-subscribe-modal`.

### Other information:

NO

## Testing instructions:

* Turn on subscribe modal
* Check that you get a default modal
* Go to the site editor and register a template part with the `jetpack-subscribe-modal` name
* Check that you get a modal with custom content
* Extra: It should also work if your theme defines a `jetpack-subscribe-modal` template part.

### DIFF
--- a/projects/plugins/jetpack/changelog/update-subscribe-modal-template-part-loading
+++ b/projects/plugins/jetpack/changelog/update-subscribe-modal-template-part-loading
@@ -1,0 +1,4 @@
+Significance: minor
+Type: other
+
+Subscribe modal: don't use template parts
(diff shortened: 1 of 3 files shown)


edanzer:
To confirm: with this solution, a potential downside is that the customer would not be able to edit the modal?

lezama:
To confirm: with this solution, a potential downside is that the customer would not be able to edit the modal?

yes, but I am currently working on a workaround, stay tuned!

lezama:
@edanzer pushed a workaround. Users should still be able to define a `jetpack-subscribe-modal` by manually adding it with this specific name.

edanzer (approved):
This tests as expected on both Jetpack self-hosted and WordPress.com simple sites.

I pushed a small fix for the modal not closing when someone adds a custom template that removes the close button.

PR #32482, opened 2023-08-15
Mailchimp block: allow transforming to a Subscribe block

simison:
## Proposed changes:

* Allow transforming Mailchimp block to subscribe block

Screenshot 2023-08-15 at 16 07 41

### Other information:

## Testing instructions:

* Insert Mailchimp block and transform it to subscribe block


jeherve (approved):
It works, but we lose any customizations we may have added to the Mailchimp block (button colors, button text, ...). Is that okay?

I'll approve for now since I do not think this is a blocker.

simison:
Yeah, I felt like carrying customizations over was good follow-up for another day. :-)  I'd probably start from the button label.

Issue #29971, opened 2023-04-06
[Super Cache] Deprecation notice on PHP 8.1


dilirity:
### Impacted plugin

Super Cache

### Quick summary

A user reported a deprecation warning when running Super Cache on PHP 8.1 - https://wordpress.org/support/topic/compatibility-with-php-8-1-7/

### Steps to reproduce

TBD

### A clear and concise description of what you expected to happen.

There should be no notice.

### What actually happened

There is a notice.

### Impact

Some (< 50%)

### Available workarounds?

No but the platform is still usable

mcaskill:
I've encountered this issue on my end.

I've patched `wp_cron_preload_cache()` in my project with the following:

```diff
 	if ( $page_on_front != 0 && ( $post_id == $page_on_front || $post_id == $page_for_posts ) )
 		continue;
 	$url = get_permalink( $post_id );
-
+	if ( !is_string( $url ) ) {
+		wp_cache_debug( "wp_cron_preload_cache: skipped $post_id. Expected a URL, received: " . gettype( $url ) );
+		continue;
+	}
 	if ( wp_cache_is_rejected( $url ) ) {
 		wp_cache_debug( "wp_cron_preload_cache: skipped $url per rejected strings setting" );
 		continue;
```

The `get_permalink()` function is documented as returning either `string` or `false` but with its various filters, could return anything.

I hope this helps.

Unfortunately, I don't have time to create a pull request.

bobmatyas:
Another user shared their solution on:

https://wordpress.org/support/topic/compatibility-with-php-8-1-7/#post-16926772

haqadn:
Although related, this doesn't seem to be a duplicate of #32336

donnchawp:
This is fixed in trunk, so the next release will have it. I'll leave this open because @mcaskill reported a different issue I'll fix.

donnchawp:
The issue @mcaskill reported has been fixed in #32484

PR #32484, opened 2023-08-15
Super cache - fix 2 null/false warnings

donnchawp:
As reported by @mcaskill in a comment on #29971 the get_permalink() function can return false which we didn't handle in the preload function. They provided a small patch to check which I added to this PR.

It was observed by @dilirity during testing that there was a NULL warning in trunk:
> PHP Deprecated:  str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in /usr/local/src/jetpack-monorepo/projects/plugins/super-cache/wp-cache-phase2.php on line 54

I noticed this only happens in wp-admin and so, this PR fixes this by disabling caching in "the backend". wp-cache-phase1.php is not loaded in wp-admin, where $wp_cache_request_uri is defined, which caused the problem. This also has the side effect of reducing the amount of extra logging done in the debug log in areas of a site that aren't cached any way.

## Proposed changes:
* Check that the return value from get_permalink() is a string, and handle it if not.
* Disable the caching system if the current URL is in "the backend", which is wp-admin, wp-login.php and other areas.

### Other information:

## Jetpack product discussion
p1692102560916519-slack-C016BBAFHHS

No

## Testing instructions:
I'm not sure how to replicate the permalink issue with preloading. It might be caused by deleting a post while preloading, when the preload system has the post_id in the "to check" array in memory.
The issue found by Peter can be replicated by using trunk, going to wp-admin and looking at the debug log.
Apply this patch and the `str_replace` warning disappears because it has not been run.

### DIFF
--- a/projects/plugins/super-cache/changelog/super-cache-fix-null-warnings
+++ b/projects/plugins/super-cache/changelog/super-cache-fix-null-warnings
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+WP Super Cache: fix null/false warning in PHP8.1
--- a/projects/plugins/super-cache/wp-cache-phase2.php
+++ b/projects/plugins/super-cache/wp-cache-phase2.php
@@ -393,6 +393,11 @@ function wp_cache_postload() {
 		return true;
 	}
 
+	if ( wpsc_is_backend() ) {
+		wp_cache_debug( 'wp_cache_postload: backend detected, not running' );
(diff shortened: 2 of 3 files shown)


haqadn (approved):
👍🏼 Looks good!

donnchawp:
ca8acbe76de50231cf7e06e5d63bd139f2f2fad9

PR #32495, opened 2023-08-15
AI Assistant: Support backend prompt building when composing forms

### LINKED ISSUE #32613: AI Extension: Move Form prompt to the backend
Let's start the process of moving prompts to the backend with the Form, that has only one possibility and is easier to transfer and test.

lhkowalski:
Fixes #32613.

Depends on D119078-code.

## Proposed changes:

* Change messages type to turn the `content` field optional and to add the `context` property
* Start sending the `context` on form completion requests

### Other information:

## Testing instructions:

* Apply D119078-code to your sandbox and sandbox the `public-api` domain
* Go to the block editor and open the browser development console
* Insert a Jetpack Form block
* Ask for some form on the AI Extension (for example, a "burguer ordering form")
* Notice the prompt structure on the request debug logs:
   * Before this change, the full prompt for the form creation would be present on the request
   * After this change, we are adding a `context` property with the necessary context to build the prompt on the server side; the original prompt is still being sent so we do a swift transition

Screenshot 2023-08-21 at 16 43 32

* Make sure the form is generated correctly
* Make sure followup requests are handled properly as well


retrofox (commented):
Can we keep sending the content? And handle this case in the endpoint? So the switching process will be smooth.

dhasilva (approved):
LGTM. Well done!

dhasilva:
[DO NOT MERGE] - The diff needs to be landed first. Once that is done we can merge this one

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
```diff
@@ -251,6 +252,9 @@ Strictly follow those rules:
 - When the user provides instructions, translate them into appropriate Gutenberg blocks and Jetpack form structure.
 - Avoid sensitive or controversial topics and ensure your responses are grammatically correct and coherent.
 - If you cannot generate a meaningful response to a user's request, reply only with "__JETPACK_AI_ERROR__". This term should only be used in this context, it is used to generate user facing errors.`,
+			context: {
+				type: 'form-ai-extension-system-prompt',
```
retrofox, line 256: We should use a TS type for this and another context type. Please don't take it as a blocker. Let's do it in a follow-up.

lhkowalski: Sure! The next step involves moving more complex prompts, so I am planning to type everything since the context property will be around a lot of places. Thank you for the suggestion!

PR #32497, opened 2023-08-15
Add production blocks metadata

andrii-lysenko:
See https://github.com/Automattic/jetpack/issues/32408

## Proposed changes:

This PR adds block json metadata files to production blocks in order to make them visible through w.org.

### Other information:

## Jetpack product discussion

pedMtX-RS-p2

No

## Testing instructions:

Review the files content.


projects/plugins/jetpack/extensions/blocks/payments-intro/block.json
```diff
@@ -0,0 +1,21 @@
+{
```
jeherve: I don't know if we need this one, since that's just a parent block for the Payments button, Donations, and Paid content blocks?

projects/plugins/jetpack/extensions/blocks/send-a-message/block.json
```diff
@@ -0,0 +1,17 @@
+{
```
jeherve, line 1: I don't know what we should do with this one. Could we just surface the WhatsApp block instead?

projects/plugins/jetpack/extensions/blocks/social-previews/block.json
```diff
@@ -0,0 +1,12 @@
+{
```
jeherve: This is more of a plugin than a block, I think we could leave it out.

jeherve (commented):
I left a few comments below.

Overall, it seems there is a mix of spaces and tabs in the different .json files. Should we stick to just tabs?

projects/plugins/jetpack/extensions/blocks/send-a-message/block.json
andrii-lysenko: Hmm, What do you mean by surfacing?

andrii-lysenko: You mean exposing only data from `whatsapp-button`?

kraftbj (approved):

projects/plugins/jetpack/extensions/blocks/send-a-message/block.json
jeherve: > You mean exposing only data from whatsapp-button?

Right, this. 👍

projects/plugins/jetpack/extensions/blocks/payments-intro/block.json
jsnajdr: `jetpack/payments-intro` is a block like any other: it's registered with `register_block_type`, can be inserted from the inserter, and is serialized as `<!-- wp:jetpack/payments-intro /-->`. What is special about is that it's meant to be a temporary placeholder, to be quickly replaced by one of the "real" payment blocks. But that's just semantics, formally it's a block.

projects/plugins/jetpack/extensions/blocks/social-previews/block.json
jsnajdr: This is right, `social-previews` registers only a plugin (adding stuff to sidebar and the pre-publish panel), there's no block at all. But removing `payments-intro/block.json` was a mistake.

projects/plugins/jetpack/extensions/blocks/payments-intro/block.json
jeherve: Right, but I'm not sure it's one that we must publicize on W.org, given that we already publicize the "real" payment blocks. It may create more confusion than anything else.

jsnajdr: Having a `block.json` file should be useful and preferred for all blocks. For example, they are then properly visible in the `/block-types` REST endpoint. If we don't want to publicize them on W.org, maybe there is a `block.json` flag for that? @gziolo will know.

gziolo: You don't need `block.json` to expose the block in the REST endpoint. All you need is to register it with `register_block_type` so it's available in the block type registry.

I don't know precisely how W.org scans the code of the plugin when looking for block types. I bet though, that when you alias the function with something else, it won't list it.

jsnajdr: > All you need is to register it with `register_block_type` so it's available in the block type registry.

Jetpack currently doesn't do server-side block registration with the `block.json`. The subset of fields that it passes to `register_block_type` is a very small subset, and it's duplicated. So, the info available in the REST endpoint will be very limited. It would be a nice follow-up to improve this.

> I don't know precisely how W.org scans the code of the plugin when looking for block types.

That would be nice to know, especially if there is any "opt-out from public registry listing" flag. It would be very useful for cases like this one.

PR #32498, opened 2023-08-15
Jetpack CRM: Use WP_UnitTestCase instead of WorDBless

kallehauge:
Jetpack CRM's codebase is a bit messy, and when trying to use WorDBless we sometimes end up fighting with different globally defined logic and assumptions that wasn't made with tests in mind.
We, therefore, want to start using [WP_UnitTestCase](https://make.wordpress.org/core/handbook/testing/automated-testing/phpunit/) instead like WordPress.org.

(Additionally: it's only a bonus that we will be able to do actual integration tests towards a "real" site to get away from a lot of the mocking we're doing to ensure our fragile legacy code doesn't break anything).

Fixes https://github.com/Automattic/zero-bs-crm/issues/3282

## Proposed changes:

* Update the test bootstrap logic [to follow the docs example](https://github.com/Automattic/jetpack/blob/trunk/docs/examples/bootstrap.php).
* Remove old `DBModeCheck` logic since we no longer have support for DAL 1 or 2 (our old data layer models).
    * The reason we remove it in this PR is an obscure race condition that happens when spinning up the test WordPress site.
* Add new parameter to `zeroBSCRM_database_reset` so we can bypass the permission check.
* Add a new Jetpack CLI command to run Jetpack CRM tests inside Docker (`jetpack docker phpunit-crm`)
* Update existing tests to use one of our new base case classes instead of `WorDBless\BaseTestCase`:
    * `JPCRM_Base_Test_Case`: ensures we do not let changed to the global `$zbs` logic bleed into other tests. 
    * `JPCRM_Base_Integration_Test_Case`: this will reset Jetpack CRM tables after each test so previous test data doesn't bleed into other tests.

### Other information:

## Jetpack product discussion

pbhBOz-3Uj-p2

No

## Testing instructions:

If you're using the Docker environment from Jetpack Monorepo, then it should be as easy as running one command:

* Spin up the Jetpack Monorepo Docker environment
* Run: `jetpack docker phpunit-crm`


projects/plugins/crm/composer.json
```diff
@@ -10,8 +10,8 @@
 		"codeception/module-db": "^1.0",
 		"codeception/module-filesystem": "^1.0",
 		"codeception/util-universalframework": "^1.0",
-		"automattic/wordbless": "0.4.2",
-		"yoast/phpunit-polyfills": "1.0.4"
+		"yoast/phpunit-polyfills": "1.0.4",
+		"symfony/polyfill-php81": "^1.27"
```
kallehauge: We needed to add `symfony/polyfill-php81` as a dev dependency to be able to use `WP_UnitTestCase`.

The reason is that `enum_exists` did not exist in the test environment which made mocking fail.

projects/plugins/crm/includes/ZeroBSCRM.Core.php
```diff
@@ -491,15 +491,6 @@ public static function instance() {
 		return self::$_instance;
 	}
 
-	/**
-	 * Cloning is forbidden.
-	 *
-	 * @since 2.1
-	 */
-	public function __clone() {
```
kallehauge, line 499: We had to remove this to be able to clone the class in our base test case (this is to restore the global state of `$zbs` after each test to avoid mocking bleeding into other tests.

projects/plugins/crm/composer.json
anomiex: What is trying to use `enum_exists` unguarded? That seems like a bug in whatever that is, so this would be the wrong solution.

projects/plugins/crm/tests/php/bootstrap.php
```diff
@@ -6,19 +6,61 @@
...
+	$_tests_dir = getenv( 'WP_DEVELOP_DIR' );
+	if ( file_exists( "$_tests_dir/tests/phpunit/" ) ) {
+		$_tests_dir .= '/tests/phpunit/';
+	}
+} elseif ( false !== getenv( 'WP_TESTS_DIR' ) ) {
+	// WordPress core environment variable.
+	$_tests_dir = getenv( 'WP_TESTS_DIR' );
+} elseif ( file_exists( dirname( dirname( $_plugin_root ) ) . '/tests/phpunit/includes/bootstrap.php' ) ) {
+	// Installed inside wordpress-develop.
+	$_tests_dir = dirname( dirname( $_plugin_root ) ) . '/tests/phpunit/includes/bootstrap.php';
```
anomiex: This bit is why your tests are failing to find things in CI. The corresponding bit in Jetpack's boostrap is
https://github.com/Automattic/jetpack/blob/89f43a22a4624fc6e3d56c6ccf0d061b7f3893a3/projects/plugins/jetpack/tests/php/bootstrap.php#L30-L32
I _think_ with what you have here you probably need two more levels of `dirname()`.

projects/plugins/crm/composer.json
kallehauge: It didn't really leave a trail, so I've had to go by assumptions, but the only places `enum_exists` is being used are `doctrine/instantiator` and `myclabs/deep-copy`. Both of them are dependencies of phpunit

anomiex: `yoast/phpunit-polyfills` should have chosen an appropriate version of phpunit for the version of PHP installed. Did that not happen?

projects/plugins/crm/tests/php/bootstrap.php
kallehauge: Your hunch was correct! It was that specific path that wasn't correct 🎉 

I'll create a separate PR to fix the path in [the example bootstrap file](https://github.com/Automattic/jetpack/blob/trunk/docs/examples/bootstrap.php) when I'm back from AFK in a couple of weeks

projects/plugins/crm/composer.json
kallehauge: It didn't appear to do so, no. At least locally but I don't see why it would be different in GitHub

anomiex: Let's give it a try anyway, I think it really will work without it in GitHub.

kallehauge: Incoming!

kallehauge: Just sharing what I got locally for context:

```
There was 1 error:

1) Automattic\Jetpack\CRM\Tests\REST_Contacts_Controller_Test::test_get_item_unknown_fatal_error
Error: Call to undefined function enum_exists()

/usr/local/src/jetpack-monorepo/projects/plugins/crm/tests/php/rest-api/v4/class-rest-contacts-controller-test.php:85
```

kallehauge: It seems you were correct in your assumption @anomiex 👍 
Now I just have to figure out why my own Jetpack Monorepo environment is complaining 😱

gogdzl (approved):
It works!
Ship it!

projects/plugins/crm/composer.json
kallehauge: I update and recreated my Docker container and the issue went away; but I can't say if there's an underlying issue since what I've done is go from PHP v8.0.28 to v8.2.7 and `enum_exists` is natively supported in v8.1+

Guess: some fringe version check that failed with 8.0.28 in `yoast/phpunit-polyfills`?

That being said: GitHub is happy and others on my team can successfully build, so I'm calling it a resolved ✅

PR #32502, opened 2023-08-16
Boost: Fix C.CSS Generation Failure

haqadn:
Fixes pc9hqz-1XT-p2

## Proposed changes:
* Do not require a nonce verification to determine generating state
* Always pass generation meta with critical CSS state

### Other information:

## Jetpack product discussion
pc9hqz-1XT-p2

No

## Testing instructions:
Test if both local critical CSS and cloud CSS generating correctly.


projects/plugins/boost/app/modules/optimizations/critical-css/Generator.php
```diff
@@ -23,32 +23,17 @@ public function __construct() {
...
+		if ( isset( $_GET[ self::GENERATE_QUERY_ACTION ] ) ) {
+			return true;
+		}
 
-		// Die if the nonce is invalid.
-		if ( $is_generating && ! Nonce::verify( $generate_nonce, self::GENERATE_QUERY_ACTION ) ) {
-			die();
+		// If there is an HTTP header indicating that we are generating, i.e. boost-cloud
+		if ( isset( $_SERVER['HTTP_X_GENERATE_CRITICAL_CSS'] ) ) {
+			return true;
```
pyronaur: Note caching: if the page is cached and there is no GET param, for the boost cloud request, this is going to return the cached version of the page, which might have this variable set to `false`, even if the header is set. I haven't walked the call tree to understand what the effects are of `is_generating_critical_css`, but providing a boolean based on header for a generic public function like this can cause problems down the line, even if there are none now.

pyronaur (dismissed):
Unsure about caching behavior here

projects/plugins/boost/app/modules/optimizations/critical-css/Generator.php
haqadn: I assumed this was for hydra. But, now that I double-checked it, apparently, hydra doesn't actually send this header. So, I am guessing we can remove this check and add the same query param-based check in hydra. 

If `is_generating_critical_css()` is true, a meta tag is added to the page. The generator can look for this meta to ensure that it has reached the correct page and not some WAF or 404. Currently, there is nothing like this for hydra, and it would be quite useful.

Did I miss any usage of this header? You introduced it a while ago.

projects/plugins/boost/app/data-sync/Critical_CSS_State_Entry.php
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	public function get( $fallback_value = false ) {
+		// WordPress looks at argument count to figure out if a fallback value was used.
+		// Only provide the fallback value if it's not the default ( false ).
+		if ( $fallback_value !== false ) {
+			$value = get_option( $this->option_key, $fallback_value );
+		} else {
+			$value = get_option( $this->option_key );
+		}
```
dilirity: Small thing :)

The [default value](https://developer.wordpress.org/reference/functions/get_option/#parameters) of [`get_option`](https://developer.wordpress.org/reference/functions/get_option/) is `false` so you can just pass in the `$fallback_value` variable.

projects/plugins/boost/app/modules/optimizations/critical-css/Generator.php
```diff
@@ -23,32 +23,12 @@ public function __construct() {
...
-			$generate_nonce = sanitize_key(
-				$_GET[ self::GENERATE_QUERY_ACTION ]
-			);
-		} elseif ( ! empty( $_SERVER['HTTP_X_GENERATE_CRITICAL_CSS'] ) ) {
-			$generate_nonce = sanitize_key(
-				$_SERVER['HTTP_X_GENERATE_CRITICAL_CSS']
-			);
-		}
-
 		// If GET parameter or header set, we are trying to generate.
```
dilirity: This comment is now outdated. Header is no longer checked.

projects/plugins/boost/app/modules/optimizations/critical-css/Generator.php
```diff
@@ -23,32 +23,12 @@ public function __construct() {
 	 * phpcs:disable WordPress.Security.NonceVerification.Recommended
 	 */
 	public static function is_generating_critical_css() {
```
dilirity, line 24: Since we're only checking the `$_GET` index, why not just:

```php
return isset( $_GET[ self::GENERATE_QUERY_ACTION ] );
```

projects/plugins/boost/app/modules/optimizations/critical-css/Generator.php
```diff
@@ -23,32 +23,12 @@ public function __construct() {
 	 * phpcs:disable WordPress.Security.NonceVerification.Recommended
 	 */
```
dilirity, line 23: Method description is also outdated.

dilirity (changes requested):
A few small things code-wise. I couldn't test if it works well, as JN is down.

dilirity (approved):
:shipit:

haqadn:
fb6ca5a719ec

PR #32504, opened 2023-08-16
repo-gardening: Update docs for new `add_labels` feature

anomiex:
## Proposed changes:

The inline comment and added doc didn't correctly represent the structure of the input (although the added example did).

The example can use yaml's multi-line string syntax to be represented in a clearer manner.

May as well take the opportunity to be a little more explicit in the doc as to how paths are matched too.

### Other information:

## Jetpack product discussion

None

No

## Testing instructions:

* Proofread the changed docs.


dereksmart (approved):
Thanks!

PR #32508, opened 2023-08-16
Add block.json metadata file to production blocks

monsieur-z:
See https://github.com/Automattic/jetpack/issues/32408

## Proposed changes:
Add block.json metadata files for Jetpack production blocks. This work is continued in https://github.com/Automattic/jetpack/pull/32497.

### Other information:

## Jetpack product discussion

pedMtX-RS-p2

Content is extracted from the metadata found in various files in the block folder.

No.

## Testing instructions:

Review the files content.


projects/plugins/jetpack/extensions/blocks/cookie-consent/block.json
```diff
@@ -0,0 +1,69 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/block.json",
+	"apiVersion": 3,
+	"name": "jetpack/cookie-consent",
+	"title": "Cookie Consent",
+	"description": "Displays a customizable cookie consent banner. To display this block on all pages of your site, please add it inside a Template Part that is present on all your templates, like a Header or a Footer.",
```
jeherve, line 6: Not for this PR, but this makes me realize that we do not use consistent phrasing in our block descriptions. Sometimes we use third-person like here, while in other descriptions we typically use second person. Maybe in a follow-up PR we can update all blocks to use a consistent messaging.

projects/plugins/jetpack/extensions/blocks/mailchimp/block.json
```diff
@@ -0,0 +1,59 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/block.json",
+	"apiVersion": 3,
+	"name": "jetpack/mailchimp",
+	"title": "Mailchimp",
+	"description": "A form enabling readers to join a Mailchimp audience.",
```
jeherve, line 6: Still on the topic of phrasing, here I think we could say "Add a form" to be consistent with other blocks. Again though, that's probably something we should do in a follow-up PR.

jeherve (approved):
This looks good to me. I left some comments, but that really should be for a follow-up PR imo.

PR #32529, opened 2023-08-17
Payments Button: Remove donation toggle from the block settings

chrisfromthelc:
Fixes https://github.com/Automattic/gold/issues/125
Context: p1691608946001779/1691528039.212859-slack-C01B6KEJ5GE

## Proposed changes:

* This patch removes the toggle for "Mark this payment plan as a donation" from the Payment Button block. 

### Other information:

## Jetpack product discussion

n/a

n/a

## Testing instructions:

1. Switch to this branch.
2. Ensure that your test site is connected to Jetpack/WPCOM.
3. Add a new payment button to a page/post.
4. In the payment button sidebar settings, it should no longer show the option to create a plan as a donation.
Screen Shot 2023-08-26 at 3 15 05 PM


projects/plugins/jetpack/extensions/shared/components/product-management-controls/inspector-control.js
```diff
@@ -88,9 +85,9 @@ export default function ProductManagementInspectorControl() {
 		if ( isCustomTitle ) {
 			return;
 		}
-		setTitle( getTitleByProps( isMarkedAsDonation, interval ) );
+		setTitle( getTitleByProps( false, interval ) );
```
millerf:
```
getTitleByProps( false, interval )
```
I believe we could remove this first parameter from the method as it will always be false...

millerf (changes requested):

projects/plugins/jetpack/extensions/shared/components/product-management-controls/inspector-control.js
chrisfromthelc: I've removed this.

millerf (approved):
LGTM

Issue #31508, opened 2023-06-22
[Sync] Enhancement: Support HPOS orders for WooCommerce in sync package


vedanshujain:
### Impacted plugin

Jetpack Sync

### What

Add support to sync WooCommerce orders when HPOS is enabled

### How

WooCommerce's HPOS feature will soon be generally available and will be enabled by default. Currently, Jetpack's sync package only supports the post tables, which will stop working on the sites which have HPOS enabled but are not backing the data to posts (for WooCommerce orders data).

To support HPOS we would need to:

### Potentially add a new module for orders on various WooCommerce order lifecycle actions 

In [class-posts.php](https://github.com/Automattic/jetpack/blob/trunk/projects/packages/sync/src/modules/class-posts.php) we are [listening on various posts lifecycle events](https://github.com/Automattic/jetpack/blob/trunk/projects/packages/sync/src/modules/class-posts.php#L135-L156), we would need to add a similar listener for orders and listen to order lifecycle events as well. Perhaps, this should be done in a new module connecting it to the [JP Sync's listener](https://github.com/Automattic/jetpack/blob/bca061f2bc37d0384a8c093096896ba3d53ecfdc/projects/packages/sync/src/class-listener.php#L85).

Depending upon what makes most sense, this can be done either:
1. Using the posts events itself, but sending order data in post data compatible format.
2. Create new events and send order data with WC_Order structure.

### Create corresponding HPOS tables on the backup sites 

If earlier we chose to create new events for orders, in WP.COM infra, we would need to create HPOS replica tables which can handle the new events.

jeherve:
cc @Automattic/jetpack-kernel

zinigor:
This is now shipped with #32530 and related diffs.

PR #32530, opened 2023-08-17
Add HPOS events to JP sync

vedanshujain:
Partially addresses #31508

## Proposed changes:

This PR adds a new module called `WooCommerce_HPOS_Order` which gets hooked when HPOS is in use and adds data to the sync queue for orders that are modified. Note that data excludes any order metadata, and PII as [per the allowlist](https://github.com/Automattic/jetpack/blob/trunk/projects/packages/sync/src/modules/class-woocommerce.php#L486). The event for full sync is also registered/hooked and will send all pending orders (based on batch size).

### Other information:

## Jetpack product discussion

See link https://wp.me/p9dueE-7KK

## Testing instructions:

Note that I don't have access to a wp.com sandbox or the sync dev environment, so the testing steps are limited to making sure that order data is being added to the queue. We will also we relying very heavily on WP shell to simulate conditions that allow us to test.

#### Pre requisite
1. If you are using a local environment, add the following snippets to force JP to think that it's connected and loads the sync module:
```php
\Automattic\Jetpack\Status\Cache::set( 'is_offline_mode', false ); // lie to the cache to prevent it from checking actual connection.

define( 'JETPACK_BLOG_TOKEN', ';,test' ); // Dummy blog token to make JP think that it's connected.

// Set a false option, again to make JP think that it's connected.
add_filter(
	'jetpack_options',
	function ( $value, $key ) {
		if ( $key === 'id' ) {
			return 1;
		}
		return $value;
	},
	10, 2
);
```
2. Set HPOS to authoritative, by following [the get started guide](https://woocommerce.com/document/high-performance-order-storage/#section-2).
3. Create at least couple of orders by going to WooCommerce > Orders > Add new so that we have something to sync.
#### Testing the full sync

1. We will use the wp shell to simulate a full sync, and open a shell environment by running `wp shell`
2. Run the following command and test along:
```php
//Register our module.
$listener = \Automattic\Jetpack\Sync\Listener::get_instance(); // Gets the listener, but more importantly will hook our new module.
$module = end(\Automattic\Jetpack\Sync\Modules::get_modules() ); // should be WooCommerce_HPOS_Order module
$module->init_before_send(); // This is usually called by the Sender, but we hook it explicitly so that our hooks are registered.

// Let's get the Full_Sync object
$full_sync = new \Automattic\Jetpack\Sync\Modules\Full_Sync();
$full_sync->reset_data(); // empty any existing so that we can test the queue
$full_sync->start();

$status = $full_sync->get_enqueue_status()['woocommerce_hpos_orders']; // Test that this has a count of pending orders/chunk size.

// Let's take a peek at the queue to make sure our order data is in fact queued. We would need to reset our queue though because there would be a lot of noise from other full sync modules.

$queue = new \Automattic\Jetpack\Sync\Queue( 'full_sync' );
$queue->reset();

$module->enqueue_full_sync_actions( $status, 100, false ); // Call full sync enqueue, but only for our module.

$queue->size(); // Should be amount of orders, or 100 whichever is lower.
$queue->peek( 100 ); // Test that full sync orders are queued.

$orders_object_chunk = $queue->peek(1); // Test that data is populated as expected.
```

#### Testing live sync

1. Again, we will use the wp shell to help with our testing, open a shell environment by running `wp shell`.
2. Run the following commands:
```php
// Let's clear the current sync queue.
$queue = new \Automattic\Jetpack\Sync\Queue( 'sync' );
$queue->reset();

$order = wc_create_order(); // Let's create an order.
// Test that order was queued under the action `woocommerce_after_order_object_save`
$queue->peek(100);
```


samiff (commented):
I tested from the perspective of seeing the new sync actions queue up.

- [Enabled](https://woocommerce.com/document/high-performance-order-storage/#section-2) Woo's HPOS in my local Jetpack dev environment.
- Verified new orders were being added to the `wc_orders` table and not `wp_posts`.
  - I see a wp_post still created with post_type `shop_order_placehold`, not entirely sure what this is used for.
- Created new orders in wp-admin and observed new sync listener `woocommerce_after_order_object_save` queueing up (I error logged from the sync queue `insert_item()` method).
  - I do notice that I see `woocommerce_after_order_object_save` queue up multiple times that look mostly identical (different timestamp), maybe this is normal?

vedanshujain:
> I see a wp_post still created with post_type shop_order_placehold, not entirely sure what this is used for.

This is expected, whenever an order is created, an equivalent post is also added with this post type and SAME id. This allows switching between two table sources if needed, without having to update external systems with new IDs.

> I do notice that I see woocommerce_after_order_object_save queue up multiple times that look mostly identical

Yeah this is also expected on save, is there a way to check and remove an existing item if it's already enqueued?

zinigor:
I don't think we should be worried about duplicating Sync events, the order gets updated on WordPress.com side with a statement that updates on duplicate entries:

```
"INSERT INTO wp_102812540_woocommerce_order_itemmeta (meta_id,order_item_id,meta_key,meta_value)
VALUES ('4042','450','_product_id','8893')
ON DUPLICATE KEY UPDATE order_item_id = '450',meta_key = '_product_id',meta_value = '8893'"
```

projects/packages/sync/src/modules/class-woocommerce-hpos-orders.php
```diff
@@ -0,0 +1,295 @@
...
+	/**
+	 * Hooks sync listners on order modify events.
+	 *
+	 * @access public
+	 *
+	 * @param callable $callable Action handler callable.
+	 */
+	public function init_listeners( $callable ) {
+		add_action( 'woocommerce_after_order_object_save', $callable );
+		add_action( 'woocommerce_delete_order', $callable );
```
samiff, line 93: `woocommerce_delete_order` doesn't seem to fire when I'm deleting from `/wp-admin/admin.php?page=wc-orders`. `woocommerce_trash_order` does seem to fire. I see there was some talk about this previously: https://github.com/woocommerce/woocommerce/issues/32543

@vedanshujain Can you confirm which deletion hooks are needed here, thanks!

vedanshujain: good catch, added a fix to address this here https://github.com/Automattic/jetpack/pull/32530/commits/fd7798f5f8b86b0e6a67e2dbdfbb56f159dd97bc.

samiff: Thanks, when deleting an order, I now see a `woocommerce_after_order_object_save` sync event followed by the `woocommerce_trash_order` one. Which I think is fine, but maybe @zinigor could confirm.

One other question I had was in regards to the order `status`. For non-HPOS orders I take it that is represented by the actual `post_status` such as `wc-pending`. Do we need to send along such a status for storage into the HPOS table @vedanshujain ?

PS: I was able to see the `id` now coming across 👍

bisko (approved):

PR #32531, opened 2023-08-17
Flickr embeds: avoid errors when a request to Flickr fails

jeherve:
## Proposed changes:

Let's add an additional safeguard before we try using items from a response we get from Flickr.

See p1692257171912309-slack-CBG1CP4EN

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

* Go to Jetpack > Settings and enable Shortcode embeds.
* Embed a Flickr video as explained here: https://wordpress.com/support/images/flickr-photos/
* It should be displayed nicely on the frontend.

### DIFF
--- a/projects/plugins/jetpack/modules/shortcodes/flickr.php
+++ b/projects/plugins/jetpack/modules/shortcodes/flickr.php
@@ -217,6 +217,11 @@ function flickr_shortcode_video_markup( $atts, $id, $video_param ) {
 
 		$embed_page = wp_remote_get( $embed_url );
 
+		// Bail if the request returns an error.
+		if ( is_wp_error( $embed_page ) ) {
+			return '';
+		}
+
 		// Get the video url from embed html markup.
 
 		preg_match( '/video.+src=\"([^\"]+)\"/', $embed_page['body'], $matches );
--- a/projects/plugins/jetpack/modules/shortcodes/flickr.php
+++ b/projects/plugins/jetpack/modules/shortcodes/flickr.php
@@ -218,7 +218,7 @@ function flickr_shortcode_video_markup( $atts, $id, $video_param ) {
 		$embed_page = wp_remote_get( $embed_url );
 
 		// Bail if the request returns an error.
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/modules/shortcodes/flickr.php
```diff
@@ -217,6 +217,11 @@ function flickr_shortcode_video_markup( $atts, $id, $video_param ) {
 
 		$embed_page = wp_remote_get( $embed_url );
 
+		// Bail if the request returns an error.
+		if ( is_wp_error( $embed_page ) ) {
```
fgiannar: Nice! I wonder if we want to explicitly check for array here (minor / not blocking)
```suggestion
		if ( ! is_array( $embed_page ) ) {
```

fgiannar (dismissed):
Thanks, Jeremy! Left a small inline suggestion (not a blocker though).

projects/plugins/jetpack/modules/shortcodes/flickr.php
jeherve: That works for me. 👍

fgiannar (approved):
LGTM 🚢

jeherve:
To close the loop on this, Flickr video embeds are broken on their end as well. I contacted them to report the issue. Here is my email below:

------

Hi there,

I've worked with Flickr's oEmbed endpoint for a few years, and it seems that recent changes caused some issues with the data returned in video embeds.

We can take the example of [this video](https://www.flickr.com/photos/lucyobe/2671401960/). Once can query the oEmbed endpoint for this video via [this query](https://www.flickr.com/services/oembed/?url=https://www.flickr.com/photos/lucyobe/2671401960/). The query returned the expected data, with an iFrame including the video content, at this URL:

https://embedr.flickr.com/photos/2671401960

Unfortunately, the source of the video isn't correct at this embed URL, and thus the video cannot be played.

Would it be possible to update the embed pages so they use the right video URLs again?

Thank you.

Best,

Jeremy Herve

PR #32532, opened 2023-08-17
Launchpad: Update the URL of 'Enable subscribers modal' task

mashikag:
Markup on 2023-08-17 at 09:57:17

## Proposed changes:

* Update the URL of the "Enable subscribers modal" task

### Other information:

## Jetpack product discussion

- p7DVsv-ioX-p2#comment-47038
- p1692248242076719-slack-C02TCEHP3HA

No, it does not. 

## Testing instructions:

* Apply the PR to your sandbox.
* Ensure that you have a test site where the current theme is `lettre`, or the `site_intent` blog option is set to `newsletter` and the site has a block theme applied 
  * 'site_intent` can be modified with the aid of `wpsh` as such:
  ```
    switch_to_blog( $test_blog_id);
    update_option( 'site_intent', 'newsletter' );
  ```
* Sandbox `public-api.wordpress.com`
* Do an API test call to `/wpcom/v2/sites/$your-test-site-slug/launchpad?checklist_slug=intent-free-newsletter`
   Markup on 2023-08-17 at 12:20:49


simison:
Oh, thanks for catching this. I thought I changed this from Calypso but didn't realize it's also in Jetpack 🤦

simison (dismissed):
I haven't tested but the URL is correct. https://wordpress.com/settings/newsletter/

Should you gate the change behind the filter?

mashikag:
Thanks @simison ! 🙇 

> Should you gate the change behind the filter?

TBH, I would not gate it behind the filter. The new Newsletter Settings page is already accessible in all environments.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -354,7 +354,7 @@ function wpcom_launchpad_get_task_definitions() {
 				if ( ( new Automattic\Jetpack\Status\Host() )->is_atomic_platform() ) {
 					return admin_url( 'admin.php?page=jetpack#/discussion' );
```
daledupreez: @mashikag, will we also need the URL to be updated for Atomic sites? If so, what's the best way to detect whether we should use the new URL or not?

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -341,7 +341,7 @@ function wpcom_launchpad_get_task_definitions() {
 			},
 			'is_complete_callback' => 'wpcom_is_task_option_completed',
 			'get_calypso_path'     => function ( $task, $default, $data ) {
-				return '/settings/reading/' . $data['site_slug_encoded'] . '#newsletter-settings';
+				return '/settings/newsletter/' . $data['site_slug_encoded'];
```
daledupreez, line 353: Seeing these two changes next two each other highlights a difference: should this function also be redirecting to a Jetpack URL when we're on Atomic? (We can handle this in a follow-up fix.)

daledupreez (dismissed):
LGTM! I have some questions that we can handle in follow-ups as needed.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
simison: Why do we redirect to Jetpack settings pages on Atomic? Those pages are for example missing most of the Newsletter settings, and are designed for self hosted Jetpack plugin customers, not .com customers, so it doesn't make sense.

Let's keep things consistent link all .com customers to .com settings pages.

mashikag: As per @simison's point, I will update the URL to be the same for all users, incl. ~Jetpack self-hosted~ Atomic sites.

simison: >  will update the URL to be the same for all users, incl. Jetpack self-hosted sites.


@mashikag Jetpack self hosted sites don't get this launchpad UI at all.

mashikag: @simison, thanks for the correction. I meant to say "incl. Atomic sites". 🙇

simison (approved):

daledupreez (commented):
LGTM!

daledupreez (approved):
That was meant to be an approval! 🙃

PR #32536, opened 2023-08-17
Blocks: add "Beta" suffix to Beta blocks' description

jeherve:
Follow-up from #32432

## Proposed changes:

As discussed during the team chat, this can be a quick and dirty way to mark our Beta blocks as "Beta" for now.

We could do the same for Experimental blocks, but it just so happens that our Experimental blocks are already marked as experimental in their titles today.

### Other information:

## Jetpack product discussion

* Primary issue: #32408 

* No

## Testing instructions:

Not much to test here. Make sure all the Beta blocks are covered here.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-chat/block.json
+++ b/projects/plugins/jetpack/extensions/blocks/ai-chat/block.json
@@ -2,7 +2,7 @@
 	"$schema": "https://schemas.wp.org/trunk/block.json",
 	"apiVersion": 3,
 	"name": "jetpack/ai-chat",
-	"title": "AI Chat (Experimental)",
+	"title": "AI Chat (Beta)",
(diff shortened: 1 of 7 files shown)


projects/plugins/jetpack/extensions/blocks/ai-chat/block.json
```diff
@@ -2,7 +2,7 @@
 	"$schema": "https://schemas.wp.org/trunk/block.json",
 	"apiVersion": 3,
 	"name": "jetpack/ai-chat",
-	"title": "AI Chat (Experimental)",
+	"title": "AI Chat (Beta)",
```
jeherve, line 5: I'm making that change to clearly mark it as Beta (vs. the other AI blocks currently marked as experimental).

monsieur-z (approved):

PR #32541, opened 2023-08-17
VideoPress: fix false values being removed from shortcodes

nunyvega:
Fixes  1822-gh-Automattic/GreenHouse

## Proposed changes:

* Use 0 as the false value for attributes set as "attribute=false" in the shortcode so the value is not empty.

Video attributes when setting controls as false, before this change:
```
(
    [w] => 1280
    [h] => 720
    [at] => 0
    [loop] =>
    [autoplay] => 1
    [cover] => 1
    [muted] =>
    [controls] =>
    [playsinline] =>
    [useaveragecolor] =>
    [preloadcontent] => metadata
)
```
After this change:
```
(
    [w] => 1280
    [h] => 720
    [at] => 0
    [loop] =>
    [autoplay] => 1
    [cover] => 1
    [muted] =>
    [controls] => 0
    [playsinline] =>
    [useaveragecolor] =>
    [preloadcontent] => metadata
)
```
### Other information:

No
## Testing instructions:

* Create a JN site using Jetpack VideoPress
* Upload one video and get its shortcode
* Add the video to a post using the shortcode and set controls=false
* Confirm that controls are not visible

### DIFF
--- a/projects/packages/videopress/changelog/fix-videopress-shortcode-false-values
+++ b/projects/packages/videopress/changelog/fix-videopress-shortcode-false-values
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fix false values not working on shortcodes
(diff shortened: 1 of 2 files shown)


retrofox (approved):
LGTM. The fix worked as expected. But I wonder why it doesn't work now.
I'm pretty sure we addressed these part of the implementation :-/

cc @dhasilva since, if I'm not wrong, you worked on this.

Accepted, but let's double-check.

PR #32543, opened 2023-08-17
AI Extension: Add default required custom text value in prompt for forms

nunyvega:
Fixes AI setting long and non default values to the required text due to not knowing the default one. 

## Proposed changes:

* Add a default value (same default value used by the form) to our prompt. 

### Other information:

## Testing instructions:

* Go to editor, add a form block
* Ask for a form and verify that required values and text are added correctly
* Ask the AI to iterate/translate/ modify the form and verify that the required field is also translated.

### DIFF
--- a/projects/plugins/jetpack/changelog/remove-required-custom-text-from-prompt
+++ b/projects/plugins/jetpack/changelog/remove-required-custom-text-from-prompt
@@ -1,0 +1,5 @@
+Significance: patch
+Type: other
+Comment: modifying prompt of unused feature
+
+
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/changelog/remove-required-custom-text-from-prompt
```diff
@@ -0,0 +1,5 @@
+Significance: patch
+Type: other
+Comment: modifying prompt of unused feature
```
dhasilva: The feature is used now

dhasilva (approved):
LGTM

PR #32551, opened 2023-08-17
Paywall Block: Improved UI

lezama:
Implements https://www.figma.com/file/qQxuiiWjgNV0AiYg6070eo/Paywall-Block?type=design&node-id=114-6927&mode=design&t=eXNSggsXhUEvjg2z-0

## Proposed changes:

**Block Panel**

Screenshot 2023-08-17 at 11 41 48

**Doc Panel**

Screenshot 2023-08-17 at 11 25 17

### Other information:

NO

## Testing instructions:

* Add a paywall block, Jump back and forward between the block panel and the document panel, play with the settings and enjoy 🥂

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/paywall/edit.js
+++ b/projects/plugins/jetpack/extensions/blocks/paywall/edit.js
@@ -1,25 +1,34 @@
 import './editor.scss';
 import { JetpackEditorPanelLogo } from '@automattic/jetpack-shared-extension-utils';
 import { InspectorControls } from '@wordpress/block-editor';
-import { PanelBody } from '@wordpress/components';
+import { Flex, FlexBlock, PanelBody, PanelRow, Spinner } from '@wordpress/components';
 import { useEntityProp } from '@wordpress/core-data';
 import { useSelect } from '@wordpress/data';
-import { store as editorStore } from '@wordpress/editor';
+import { PostVisibilityCheck, store as editorStore } from '@wordpress/editor';
 import { __ } from '@wordpress/i18n';
 import { arrowDown, Icon } from '@wordpress/icons';
-import { accessOptions } from '../../shared/memberships/constants';
+import {
+	accessOptions,
+	META_NAME_FOR_POST_LEVEL_ACCESS_SETTINGS,
+} from '../../shared/memberships/constants';
 import { useAccessLevel } from '../../shared/memberships/edit';
-import { NewsletterAccessDocumentSettings } from '../../shared/memberships/settings';
+import { NewsletterAccessRadioButtons } from '../../shared/memberships/settings';
+import {
+	getShowMisconfigurationWarning,
+	MisconfigurationWarning,
+} from '../../shared/memberships/utils';
 
 function PaywallEdit( { className } ) {
 	const postType = useSelect( select => select( editorStore ).getCurrentPostType(), [] );
 	const accessLevel = useAccessLevel( postType );
 	const [ , setPostMeta ] = useEntityProp( 'postType', postType, 'meta' );
 
+	if ( ! accessLevel || accessLevel === accessOptions.everybody.key ) {
+		setPostMeta( { [ META_NAME_FOR_POST_LEVEL_ACCESS_SETTINGS ]: accessOptions.subscribers.key } );
+	}
+
 	const getText = key => {
 		switch ( key ) {
-			case accessOptions.everybody.key:
-				return __( 'Change visibility to enable paywall', 'jetpack' );
 			case accessOptions.subscribers.key:
 				return __( 'Subscriber-only content below', 'jetpack' );
 			case accessOptions.paid_subscribers.key:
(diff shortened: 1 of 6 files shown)


projects/plugins/jetpack/extensions/blocks/paywall/edit.js
```diff
@@ -46,18 +55,82 @@ function PaywallEdit( { className } ) {
...
+							{ canEdit && (
+								<div className="editor-post-visibility">
+									<NewsletterAccessRadioButtons
+										isEditorPanel={ true }
+										onChange={ setPostMeta }
+										accessLevel={ _accessLevel }
+										stripeConnectUrl={ stripeConnectUrl }
+										hasNewsletterPlans={ hasNewsletterPlans }
+										paywallBlockSettings={ true }
+									/>
```
edanzer, line 122: Questions: 
   - If I'm reading this right, we're re-using the NewsletterAccessRadioButtons component within the Paywall block?
   - Does it make sense to show options for 'Everyone' and 'All Subscribers' for a Paywall block? I would think the point of the block is to put the content below behind a paywall for paid subscribers?

projects/plugins/jetpack/extensions/shared/memberships/settings.js
```diff
@@ -165,52 +177,61 @@ function NewsletterAccessSetupNudge( { stripeConnectUrl, isStripeConnected, hasN
...
 
-function NewsletterAccessRadioButtons( {
+export function NewsletterAccessRadioButtons( {
 	onChange,
 	accessLevel,
 	hasNewsletterPlans,
 	stripeConnectUrl,
 	isEditorPanel = false,
+	hasPaywallBlock = false,
+	paywallBlockSettings = false,
```
edanzer, line 187: This is a taste/preference, but for vars/props that are booleans, I think it can be easier to read code if they're named such that it's clear they're booleans. Like for this one, something like isPaywallBlock.

edanzer: I think this boolean is actually isPaywallBlockSettings - so a conditional that says whether this instance of the NewsletterAccessRadioButtons is for Newsletter visibility or for Paywall settings. So 'is...' rather than 'show...'

projects/plugins/jetpack/extensions/shared/memberships/settings.js
```diff
@@ -165,52 +177,61 @@ function NewsletterAccessSetupNudge( { stripeConnectUrl, isStripeConnected, hasN
...
-					{ key === accessLevel && key !== accessOptions.everybody.key && (
-						<p className="editor-post-visibility__notice">
-							<NewsletterNotice accessLevel={ accessLevel } />
-						</p>
-					) }
-				</div>
-			) ) }
+			{ Object.keys( accessOptions ).map(
+				key =>
+					( ! paywallBlockSettings || key !== accessOptions.everybody.key ) && (
```
edanzer, line 197: Seems this should work but wondering if it makes sense for the `! paywallBlockSettings` check to be outside of the map() function? Seems like if paywallBlockSettings === false, we don't render anything from this map function.

edanzer: Answering my own question: no because we want to use it combination with properties for each accessOption. It's how we don't show everyone only when  loading paywall block.

projects/plugins/jetpack/extensions/shared/memberships/settings.js
```diff
@@ -165,52 +177,61 @@ function NewsletterAccessSetupNudge( { stripeConnectUrl, isStripeConnected, hasN
...
+								checked={ key === accessLevel }
+								className="editor-post-visibility__radio"
+								id={ `editor-post-${ key }-${ instanceId }` }
+								name={ `editor-newsletter-access__setting-${ instanceId }` }
+								aria-describedby={ `editor-post-${ key }-${ instanceId }-description` }
+								disabled={
+									( key === accessOptions.paid_subscribers.key &&
+										( ! isStripeConnected || ! hasNewsletterPlans ) ) ||
+									( key === accessOptions.everybody.key && hasPaywallBlock )
+								}
```
edanzer, line 211: > key === accessOptions.everybody.key && hasPaywallBlock

So for this code, if a Paywall Block exists on the post, we disallow showing the post to everyone. I think I probably missed background convo on this, but I can imagine many situations where a user might want to show the content above the paywall to everyone. Doing so allows you to maximize visibility of the post, and thus possible sales of paid subscriptions.

edanzer (dismissed):
Some complex logic and lots of code I haven't interacted with before, so it took me a bit to fully follow what's going on. But I've got it now and I think this all makes sense. 

Basically we're using the same component for both the Newsletter Visibility Panel and the Paywall Block Settings. And we need to conditionally show/hide/disable options depending the settings in each and how they interact with each other. 

Looking at the Figma and the the code, this seems to be testing as I'd expect. 

Added a few questions/comments, but none are blockers to merging.

projects/plugins/jetpack/extensions/blocks/paywall/index.js
```diff
@@ -27,7 +27,7 @@ export const title = __( 'Paywall', 'jetpack' );
 export const settings = {
 	title,
 	description: __(
-		'Add a paywall block to make part of your content exclusive to paid subscribers.',
+		'Content before this block will be shown in the excerpt on your archives page.',
```
simison, line 30: I'll update this also in [`block.json`](https://github.com/Automattic/jetpack/edit/af6bbd095b61806f8dcd806ce92d6c8acf7eee76/projects/plugins/jetpack/extensions/blocks/paywall/block.json) that we now have since https://github.com/Automattic/jetpack/pull/32432. I don't think it's used yet since blocks are being migrated.


Edit: updated in https://github.com/Automattic/jetpack/pull/32551/commits/b80b32832f611c7af400f580a7412d3f7f023443

simison (approved):
Tested well, too.

PR #32570, opened 2023-08-17
Sitemaps: Add WooCommerce product post type to sitemaps if WooCommerce is present

### LINKED ISSUE #32572: WooCommerce & Jetpack Sitemaps: Add support for WooCommerce products post type
### What

Add support for WooCommerce Products to Jetpack Sitemaps

### Why

WooCommerce utilizes the products post type which Jetpack Sitemaps does not include support for by default. 

It is possible to add this support via a code snippet. Others utilize a dedicated SEO plugin for their sitemaps instead.

> **How to add more Post Types to the sitemap**
> The default sitemap only includes Posts, Pages, and the Custom Post Types included in Jetpack. If you’d like to add a new Custom Post Type to the default sitemap, you can use the jetpack_sitemap_post_types filter to specify your own array of post types to include.

* https://jetpack.com/support/sitemaps/

Given WordPress.com's goals, it appears that we might want to foster some collaboration between WooCommerce and Jetpack to help ensure products are added to Jetpack Sitemaps.

### How

_No response_

oskosk:
Add compatibility with WooCommerce types when generating sitemaps. Non-visible products are skipped when generating the sitemap.

Fixes #32572

<table>
<tr>
<td>
 <td>
<tr>
 <td>Before
 <td>After
</table>

## Proposed changes:

* Adds a filter to `jetpack_sitemap_post_types` adding the `product` post type to be mapped.
* * Adds a filter to `jetpack_sitemap_skip_post` skipping `product` if it's not visible

### Other information:

## Jetpack product discussion

p1692300608136739-slack-C03Q87XT1QF

No

## Testing instructions:
* Install WooCommerce, and Jetpack running this branch (e.g. [Jurassic Ninja + Jetpack (this branch) + WooCommerce]( https://jurassic.ninja/create?jetpack-beta&branches.jetpack=add/sitemap-compatibility-with-woocommerce&wp-debug-log&woocommerce) )
* Connect Jetpack
* Set up your Woo Store.
* Add A product named **Shirt**. Publish it
* Add a second product named **T-shirt**, but this time set its catalog visibility to hidden before publishing.
* Enable Jetpack Settings -> Traffic -> Sitemaps
* Visit [your-site]/sitemap.xml
* Click the first sitemap.
* Confirm the Shirt product you added is listed there.
* Confirm the T-Shirt product you added is **not** listed there.
  image

## Test on WooExpress site

* Install [Jetpack beta plugin](https://github.com/Automattic/jetpack-beta/archive/refs/heads/trunk.zip)
* Make sure your site is published and/or set to be reachable by engines.
* Visit Jetpack -> Jetpack Beta
* Click Jetpack
* In the field **Feature Branch**, type `woocommerce` and from the dropdown select `add / sitemap compatibility with woocommerce
`
  image
* Add A product named **Shirt**. Publish it
* Add a second product named **T-shirt**, but this time set its catalog visibility to hidden before publishing.
* Enable Sitemaps under Tools -> Marketing -> Traffic (At the bottom)
  image
* Visit [your-site]/sitemap.xml
* Click the first sitemap.
* Confirm the product you added is listed there.
* Confirm the T-Shirt product you added is **not** listed there.
  image

### DIFF
--- a/projects/plugins/jetpack/3rd-party/woocommerce.php
+++ b/projects/plugins/jetpack/3rd-party/woocommerce.php
@@ -57,7 +57,9 @@ function jetpack_woocommerce_add_to_sitemap( $post_types ) {
 function jetpack_woocommerce_skip_hidden_products_in_sitemap( $skip, $post ) {
 	if ( $post !== null && $post->post_type === 'product' ) {
 		$product = wc_get_product( $post->ID );
-		$skip    = ! $product->is_visible();
+		if ( $product ) {
+			$skip = ! $product->is_visible();
+		}
 	}
 	return $skip;
 }
--- a/projects/plugins/jetpack/3rd-party/woocommerce.php
+++ b/projects/plugins/jetpack/3rd-party/woocommerce.php
@@ -23,12 +23,28 @@ function jetpack_woocommerce_integration() {
 	add_action( 'woocommerce_share', 'jetpack_woocommerce_social_share_icons', 10 );
 
 	/**
+	 * Add product post type to Jetpack sitemap present.
+	 */
+	add_filter( 'jetpack_sitemap_post_types', 'jetpack_woocommerce_add_to_sitemap' );
+
+	/**
 	 * Wrap in function exists check since this requires WooCommerce 3.3+.
 	 */
 	if ( function_exists( 'wc_get_default_products_per_row' ) ) {
 		add_filter( 'infinite_scroll_render_callbacks', 'jetpack_woocommerce_infinite_scroll_render_callback', 10 );
 		add_action( 'wp_enqueue_scripts', 'jetpack_woocommerce_infinite_scroll_style', 10 );
 	}
+}
+
+/**
+ * Add product post type to sitemap if Woocommerce is present.
+ *
+ * @param array $post_types Array of post types included in sitemap.
+ */
+function jetpack_woocommerce_add_to_sitemap( $post_types ) {
+	$post_types[] = 'product';
+
+	return $post_types;
 }
 
 /**
--- a/projects/plugins/jetpack/3rd-party/woocommerce.php
+++ b/projects/plugins/jetpack/3rd-party/woocommerce.php
@@ -23,9 +23,10 @@ function jetpack_woocommerce_integration() {
 	add_action( 'woocommerce_share', 'jetpack_woocommerce_social_share_icons', 10 );
 
 	/**
-	 * Add product post type to Jetpack sitemap present.
+	 * Add product post type to Jetpack sitemap while skipping hidden products.
 	 */
 	add_filter( 'jetpack_sitemap_post_types', 'jetpack_woocommerce_add_to_sitemap' );
+	add_filter( 'jetpack_sitemap_skip_post', 'jetpack_woocommerce_skip_hidden_products_in_sitemap', 10, 2 );
 
 	/**
 	 * Wrap in function exists check since this requires WooCommerce 3.3+.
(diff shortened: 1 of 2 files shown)


kraftbj:
Some related conversations about if including all products—including "Hidden" ones—is the best UX:
pdhcfv-5i-p2#comment-113
https://github.com/woocommerce/woocommerce/issues/27954

I think, from an owner's POV, if a product was marked "hidden", I would not expect it to be indexed into search engines. It is confusing with the "search results" item there, as that's meant for the internal site search, but nevertheless.

beaulebens:
Good call-out @kraftbj. Given the context, it would be best to leave these hidden products out of the sitemap IMHO. It won't actually impact their visibility if someone has the direct link (expected functionality), but will avoid encouraging additional traffic to them via search engines (which seems to be the desired intent).

Unfortunately to remove them, it looks like we'd have to introduce an additional clause/join against a taxonomy (`product_visibility`)... which is [filterable](https://github.com/woocommerce/woocommerce/blob/278366def9b7f29bbdcccc2c6d6fd42144d25db8/plugins/woocommerce/includes/wc-product-functions.php#L835).

oskosk:
@beaulebens @kraftbj in c7c7e130563542a44159c96bd90c0043f69a9936 I pushed a commit to filter out those products that are not visible

image

projects/plugins/jetpack/3rd-party/woocommerce.php
```diff
@@ -31,6 +37,31 @@ function jetpack_woocommerce_integration() {
...
+/**
+ * Skip hidden products when generating the sitemap.
+ *
+ * @param bool    $skip Whether to skip the post.
+ * @param WP_Post $post The post object.
+ */
+function jetpack_woocommerce_skip_hidden_products_in_sitemap( $skip, $post ) {
+	if ( $post !== null && $post->post_type === 'product' ) {
+		$product = wc_get_product( $post->ID );
+		$skip    = ! $product->is_visible();
```
kraftbj: I'm waiting for the sitemap generation to happen. Should we be defensive here? `wc_get_product` can return `null` or `false`, in addition to a `WC_Product` object.

kraftbj:
```suggestion
		$skip    = ! (method_exists( $product, 'is_visible') && $product->is_visible() );
```
or check if it is a WC_Product, etc?

kraftbj (dismissed):
Confirmed this PR adds products to the sitemap and does not include "hidden" ones. Can merge now and add the defense later or can re-approve later.

projects/plugins/jetpack/3rd-party/woocommerce.php
oskosk: Thanks. Pushed 4a443f0bd0cb8affa43aa8a0f010daf03359ffe5 addressing the observation

kraftbj (approved):

oskosk:
@kraftbj I pushed a commit checking for the return value to be truthy. I'm not sure about checking the class too. We're already deliberately hooking to WoCommerce by checkin `class_exists( 'WooCommerce' )`,  checking the post type to be 'product', and relying on the return value of a Woo specific `wc_get_product()`.

kraftbj:
> @kraftbj I pushed a commit checking for the return value to be truthy. I'm not sure about checking the class too. We're already deliberately hooking to WoCommerce by checkin `class_exists( 'WooCommerce' )`, checking the post type to be 'product', and relying on the return value of a Woo specific `wc_get_product()`.

Makes sense. I do tend to get more defensive than is beneficial. The truth check should be sufficient. Thanks! Approved and set to automerge once CI finishes.

PR #32571, opened 2023-08-17
Refactor/make memberships class more dry

chrismccluskey:
Related to 106-gh-automattic/gold
Contains changes from https://github.com/Automattic/jetpack/pull/32652

## Proposed changes:

* Refactors the class which handles responses for the membership products endpoints to minimize code duplication, making the class more DRY ("Don't Repeat Yourself").

### Other information:

## Testing instructions:

Test to ensure there isn't a regression with this patch.

1. Sandbox the public API.
2. Download the patched version of Jetpack to your sandbox with `bin/jetpack-downloader test jetpack refactor/make-memberships-class-more-dry`.
3. Visit https://wordpress.com/earn/payment-plans in a browser and choose a site that is setup with Earn payments.
4. Click to add a new payment plan. The plan should add successfully.
5. Click to edit the created payment plan, make some adjustments and click save. The plan should save successfully and your edits should persist if you refresh the page.
6. Click to delete the created payment plan. The plan should delete, and if you refresh the page the plan should not re-appear in the list of payment plans.
7. Return to your Jetpack site and create a new post, this time with a payment button. Click the icon to select the payment plan and you should see a list of payment plans. Click add new payment plan and fill out the form, clicking save when done. The plan should exist when you return to the calypso live page and view your payment plans.


millerf (commented):
It also look like there are some argument missing from the endpoint (subscribe_as_site_subscriber, etc...)
Maybe we could get those parameters from a helper function as well (as they are shared between creation and update)?

chrismccluskey:
> It also look like there are some argument missing from the endpoint (subscribe_as_site_subscriber, etc...) Maybe we could get those parameters from a helper function as well (as they are shared between creation and update)?

That's actually what `get_payload_for_product` does. It's called at the top of `create_product` and `update_product` and it is used to build the payload, which is actually identical between the create and update callbacks. The only difference between the two is that update has an `product_id`, but that comes from the url parameters and not from query args.

millerf (approved):
LGTM. Maybe you should rebase and check tests are passing before merging?

PR #32578, opened 2023-08-18
Paywall block: avoid undefined variable warning

simison:
@lezama can you double check the logic here, I'm not entirely sure I followed it right.

We could add some comments between lines to make it easier to follow what happens.

## Proposed changes:

Fixes:
```
PHP Warning:  Undefined variable $paywalled_content in /usr/local/src/jetpack-monorepo/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php on line 774
```
### Other information:

## Testing instructions:

* Post which is free _and_ paid as a free _and_ paid user (so 4 test cases)
* Same as above but with paywall block in the post
* Check both emails and the blog

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -760,19 +760,17 @@ function add_paywall( $the_content ) {
 		return $the_content;
 	}
 
+	$post_access_level = Jetpack_Memberships::get_post_access_level();
+	if ( jetpack_is_frontend() ) {
+		$paywalled_content = get_paywall_blocks( $post_access_level );
+	} else {
+		// emails
+		$paywalled_content = get_paywall_simple();
+	}
+
 	// Partially free content with paywall
 	if ( has_block( $block_name ) ) {
-		$post_access_level = Jetpack_Memberships::get_post_access_level();
-		if ( jetpack_is_frontend() ) {
-			$paywalled_content = get_paywall_blocks( $post_access_level );
-		} else {
-			// emails
-			$paywalled_content = get_paywall_simple();
-		}
-		$paywalled_content = strstr( $the_content, '<!-- wp:' . $block_name . ' /-->', true ) . $paywalled_content;
-	} else {
-		// Fully paywalled content
-		$paywalled_content = get_paywall_blocks( $post_access_level );
+		return strstr( $the_content, '<!-- wp:' . $block_name . ' /-->', true ) . $paywalled_content;
 	}
 
 	return $paywalled_content;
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -760,15 +760,17 @@ function add_paywall( $the_content ) {
...
-			$paywalled_content = get_paywall_blocks( $post_access_level );
-		} else {
-			// emails
-			$paywalled_content = get_paywall_simple();
-		}
-		$paywalled_content = strstr( $the_content, '<!-- wp:' . $block_name . ' /-->', true ) . $paywalled_content;
+		return strstr( $the_content, '<!-- wp:' . $block_name . ' /-->', true ) . $paywalled_content;
 	}
 
 	return $paywalled_content;
```
edanzer, line 776: This PR does seem to resolve the error for me. 

I have a question a bit beyond just resolving the error. We seem to return $paywall_content from this function whether the paywall block exists on the page or not? That was the case both before and after the changes in this PR. I would think that if there's no paywall block on the page, we should return the_content? ie:

```
if ( Jetpack_Memberships::user_can_view_post() ||  ! has_block( jetpack/paywall )) {
		return $the_content;
} else {
  if ( jetpack_is_frontend() ) {
		$paywalled_content = get_paywall_blocks( $post_access_level );
	} else {
		// emails
		$paywalled_content = get_paywall_simple();
	}
  return strstr( $the_content, '<!-- wp:' . $block_name . ' /-->', true ) . $paywalled_content;
}
```

lezama: we return the_content earlier, here:

https://github.com/Automattic/jetpack/pull/32578/files#diff-9164d71f603706b3c6cc58c71ba3f2fbb12b1dba772adca76793f60c24ddbca7R760

edanzer: Yup, based on Jetpack_Memberships::user_can_view_post(). 

But there's a separate check for has_block( jetpack/paywall ). So question is: what happens here if you get past the first check (ie, Jetpack_Memberships::user_can_view_post() is false), and then the post does not have a paywall block? 

In that situation, we would currently return either `get_paywall_blocks( $post_access_level )` or `get_paywall_simple()`.

lezama: yes, that's what we want, means that the full content of the post is replaced by the paywall

edanzer: To clarify after chatting about this: The `get_paywall_blocks( $post_access_level )` and `get_paywall_simple()` blocks are not only used to get content based on the paywall block (which I assumed based on method names) but also for broader newsletter post visibility. When `has_block( jetpack/paywall )` is false, this code will block the full content of the post based on newsletter visibility settings. When `has_block( jetpack/paywall )` is true, it'll show the content above the paywall, then block the remainder of the content.

edanzer (approved):

PR #32579, opened 2023-08-18
Google Fonts: Register Google Fonts to theme json default to keep them in the font selection list

arthur791004:
Fixes https://github.com/Automattic/dotcom-forge/issues/3162, https://github.com/Automattic/wp-calypso/issues/78605

## Proposed changes:

* [Gutenberg adds the missing fonts to the theme json data](https://github.com/WordPress/gutenberg/blob/57d7861eca9a526d896adf717f64d6483bb6cc6f/lib/class-wp-theme-json-resolver-gutenberg.php#L283). However, when it tries to merge the global styles (selected style variation) into the theme json data, the [mergeBaseAndUserConfigs](https://github.com/WordPress/gutenberg/blob/eac3f039b0bd094af35291987390ea1c5cc473d9/packages/edit-site/src/components/global-styles/global-styles-provider.js#L24) will override the old array (no merging) so that the font families from the theme json data will be overridden if the style variation has its own font families set.
* The above behavior seems to be correct as the font family list should be the same as the style variation. However, on Jetpack (or WP.com), we want to display the google fonts that are not included in the theme or its style variation. So, we'd prefer to inject the google fonts into the core data so people can still pick the google fonts even if the current style variation has its font definition.

| Before | After |
| - | - |
| image | image |

### Other information:

## Testing instructions:

**WP.com**

* Apply this patch to your sandbox
* Go to Theme Showcase
* Activate the Pendant theme with any style variation (non-default one)
* Go to the Site Editor
* Select Styles > Typography > Text > Font
* Ensure you're able to see all Google Fonts in the list instead of just 3

**JN site**

* Create a [Jurassic Ninja sandbox](https://jurassic.ninja/create/?jetpack-beta&branch=feat/register-google-fonts-to-theme-json-default) against this PR.
* Go to Theme Showcase
* Install and activate the Pendant theme via https://public-api.wordpress.com/rest/v1/themes/download/pendant.zip
* Go to the Site Editor
* Select Styles > Typography > Text > Font
* Ensure you're able to see all Google Fonts in the list instead of just 3
* Install the Gutenberg plugin and repeat the above steps
* Ensure everything still works well

### DIFF
--- a/projects/plugins/jetpack/modules/google-fonts.php
+++ b/projects/plugins/jetpack/modules/google-fonts.php
@@ -169,4 +169,6 @@ function jetpack_register_google_fonts_to_theme_json_default( $theme_json ) {
 	return new WP_Theme_JSON_Data_Gutenberg( $raw_data, 'default' );
 }
 
-add_filter( 'wp_theme_json_data_default', 'jetpack_register_google_fonts_to_theme_json_default' );
+if ( class_exists( 'WP_Fonts_Utils' ) ) {
+	add_filter( 'wp_theme_json_data_default', 'jetpack_register_google_fonts_to_theme_json_default' );
+}
--- a/projects/plugins/jetpack/modules/google-fonts.php
+++ b/projects/plugins/jetpack/modules/google-fonts.php
@@ -140,3 +140,33 @@ add_filter( 'pre_render_block', '\Automattic\Jetpack\Fonts\Introspectors\Blocks:
 // re-registration (at priority 22) of the core blocks it de-registers (at the default priority 10),
 // otherwise Gutenberg caches an incorrect state.
 add_action( 'init', '\Automattic\Jetpack\Fonts\Introspectors\Global_Styles::enqueue_global_styles_fonts', 22 );
+
+/**
+ * Register google fonts to the default theme json data
+ *
+ * @param WP_Theme_JSON_Data_Gutenberg $theme_json The theme json data of core.
+ * @return WP_Theme_JSON_Data_Gutenberg The theme json data with registered google fonts.
+ */
+function jetpack_register_google_fonts_to_theme_json_default( $theme_json ) {
+	$google_font_families = apply_filters( 'jetpack_google_fonts_list', JETPACK_GOOGLE_FONTS_LIST );
+
+	$raw_data = $theme_json->get_data();
+	if ( empty( $raw_data['settings']['typography']['fontFamilies']['default'] ) ) {
+		$raw_data['settings']['typography']['fontFamilies']['default'] = array();
+	}
+
+	$default_font_families = array_map(
+		array( 'WP_Fonts_Utils', 'get_font_family_from_variation' ),
+		$raw_data['settings']['typography']['fontFamilies']['default']
+	);
+
+	$to_add = array_diff( $google_font_families, $default_font_families );
+	foreach ( $to_add as $font_family ) {
+		$font_family_handle = WP_Fonts_Utils::convert_font_family_into_handle( $font_family );
+		$raw_data['settings']['typography']['fontFamilies']['default'][] = wp_fonts()->to_theme_json( $font_family_handle );
+	}
+
+	return new WP_Theme_JSON_Data_Gutenberg( $raw_data, 'default' );
+}
+
+add_filter( 'wp_theme_json_data_default', 'jetpack_register_google_fonts_to_theme_json_default' );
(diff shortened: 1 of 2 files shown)


taipeicoder (dismissed):
Tested and works great 👍 
Screenshot 2023-08-22 at 2 35 04 PM

anomiex:
Trying this on a Jurassic Ninja site without the Gutenberg plugin installed, it fatals as soon as the Google Fonts module is enabled.
```
Fatal error: Uncaught TypeError: array_map(): Argument #1 ($callback) must be a valid callback, class "WP_Fonts_Utils" not found in .../public/wp-content/plugins/jetpack-dev/modules/google-fonts.php:160
```

arthur791004:
Ooops...nice catch! I added the check to hook the action only when the `WP_Fonts_Utils` class exists. Thank you!

projects/plugins/jetpack/modules/google-fonts.php
```diff
@@ -140,3 +140,35 @@ function jetpack_add_google_fonts_provider() {
...
+		$raw_data['settings']['typography']['fontFamilies']['default']
+	);
+
+	$to_add = array_diff( $google_font_families, $default_font_families );
+	foreach ( $to_add as $font_family ) {
+		$font_family_handle = WP_Fonts_Utils::convert_font_family_into_handle( $font_family );
+		$raw_data['settings']['typography']['fontFamilies']['default'][] = wp_fonts()->to_theme_json( $font_family_handle );
+	}
+
+	return new WP_Theme_JSON_Data_Gutenberg( $raw_data, 'default' );
```
anomiex: I wonder whether this class will also cause trouble in the future: if `WP_Fonts_Utils` makes it into core I'd guess core will still use `WP_Theme_JSON_Data` rather than `WP_Theme_JSON_Data_Gutenberg`.

anomiex (commented):
On a JN site now it works without Gutenberg. With Gutenberg installed, it dies with the following errors
```
Deprecated: Function wp_register_webfont_provider is deprecated since version GB 15.1! Use wp_register_font_provider instead. in .../wp-includes/functions.php on line 5453

Warning: Array to string conversion in .../wp-content/plugins/gutenberg/lib/experimental/fonts-api/class-wp-fonts-resolver.php on line 212

Warning: Array to string conversion in .../wp-content/plugins/gutenberg/lib/experimental/fonts-api/class-wp-fonts-resolver.php on line 212

Warning: Undefined array key "theme" in .../wp-content/plugins/gutenberg/lib/experimental/fonts-api/class-wp-fonts-resolver.php on line 207

Fatal error: Uncaught TypeError: array_merge(): Argument #1 must be of type array, null given in .../wp-content/plugins/gutenberg/lib/experimental/fonts-api/class-wp-fonts-resolver.php:208
Stack trace:
#0 .../wp-content/plugins/gutenberg/lib/experimental/fonts-api/class-wp-fonts-resolver.php(208): array_merge(NULL, Array)
#1 .../wp-content/plugins/gutenberg/lib/experimental/fonts-api/class-wp-fonts-resolver.php(117): WP_Fonts_Resolver::get_settings()
#2 .../wp-includes/class-wp-hook.php(310): WP_Fonts_Resolver::register_fonts_from_theme_json('')
#3 .../wp-includes/class-wp-hook.php(334): WP_Hook->apply_filters(NULL, Array)
#4 .../wp-includes/plugin.php(517): WP_Hook->do_action(Array)
#5 .../wp-settings.php(632): do_action('init')
#6 .../wp-config.php(103): require_once('/srv/users/user...')
#7 .../wp-load.php(50): require_once('/srv/users/user...')
#8 .../wp-admin/admin.php(34): require_once('/srv/users/user...')
#9 {main} thrown in .../wp-content/plugins/gutenberg/lib/experimental/fonts-api/class-wp-fonts-resolver.php on line 208

There has been a critical error on this website. Please check your site admin email inbox for instructions.
```
What seems to be happening is that it winds up with the same list of fonts in both 'default' and 'theme' the first time it reaches https://github.com/WordPress/gutenberg/blob/v16.5.0/lib/experimental/fonts-api/class-wp-fonts-resolver.php#L212, so the `array_unique` keeps only 'default' and on the next iteration it blows up.

arthur791004:
> the array_unique keeps only 'default' and on the next iteration it blows up.

I don't have any idea how to avoid the above behavior since the missing fonts are injected after the `wp_theme_json_data_theme` filter is triggered...

However, I found the `register_fonts_from_theme_json` function runs only once when the `init` hook of the WP is called. Hence, I made the new hook be injected when the `wp_loaded` hook is triggered so we can avoid the array_unique removing the value of the `theme`. Do you think this way is acceptable?

anomiex:
> Do you think this way is acceptable?

I have no idea. :shrug: There would seem to be a risk that, since it's only called once, then necessary data won't get cached. But I don't know enough about Gutenberg to know if that's really a concern or not.

arthur791004:
Thanks for your review! Now I assume this issue should be resolved by Gutenberg so closing this one in favor of https://github.com/WordPress/gutenberg/issues/54008

PR #32586, opened 2023-08-18
tools: Improve install-monorepo.sh script

anomiex:
## Proposed changes:

The big change here is adding some better error handling around the attempt to `jetpack cli link`. If `pnpm setup` hasn't been done successfully, `pnpm link --global` will silently fail and so `jetpack cli link` will too.

Beyond that, this also pulls in our "chalk-lite" library and uses those functions to format the output a little nicer.

### Other information:

## Jetpack product discussion

p1691445603197599-slack-CDLH4C1UZ

No

## Testing instructions:

* Ideally, try this on a fresh Mac that doesn't have pnpm or anything installed yet.


kraftbj:
In testing it on a fresh Mac, it installed `pnpm`, but `pnpm` wasn't yet available. There's a notice to run `source ~/.zshrc` to enable pnpm. A fresh terminal window does work, so that makes sense.

tools/install-monorepo.sh
```diff
@@ -18,102 +19,100 @@ if [[ $1 ]]; then
...
 	brew install jq
 fi
 
-echo "Checking if pnpm is installed..."
+info "Checking if pnpm is installed..."
 if ! command -v pnpm &>/dev/null; then
-	echo "Installing pnpm"
-	curl -f https://get.pnpm.io/v6.16.js | node - add --global pnpm
+	info "Installing pnpm"
+	curl -fsSL https://get.pnpm.io/install.sh | sh -
```
kraftbj:
```suggestion
	curl -fsSL https://get.pnpm.io/install.sh | sh -
	source ~/.zshrc
```

Adding this here resolved the issue for me.

kraftbj (changes requested):
One tweak to have the script sing.

tools/install-monorepo.sh
anomiex: This will not work for people using bash instead of zsh. Or dash or fish or ksh.

For that matter it might not always work for people running zsh, if their `~/.zshrc` includes stuff that's not compatible with the bash that the current script is running with.

... So many different ways to install JS crap, and they all suck.

gravityrail: Would this work?
```
if [[ $0 == "zsh" ]]; then
  source ~/.zshrc
fi
...
```

gravityrail: It wouldn't solve all your problems, but personally this PR is an improvement for me and I'd like to get it merged :)

anomiex: No. But c766a4f might.

kraftbj (approved):

PR #32590, opened 2023-08-18
 AI Extension: do not send request when AI assistant input is empty

### LINKED ISSUE #32589: AI Extension: pressing enter key in the Assistant input triggers an error in the console
Image

Let's block the input request when it's empty

retrofox:
This PR checks if the AI Assistant bar input is empty when the user presses the enter key. If so, it won't call the send() function, avoiding showing an error in the dev console./

Fixes https://github.com/Automattic/jetpack/issues/32589

## Proposed changes:

* AI Extension: do not send request when AI assistant input is empty

### Other information:

## Testing instructions:

* Go to the block editor
* Create a Jetpack Form block instance
* **Before**: Confirm that typing the enter key in the input when it's empty triggers an error in the dev-console
image

* **After**: dev console does not show any error.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/jetpack-contact-form/components/ai-assistant-bar/index.tsx
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/jetpack-contact-form/components/ai-assistant-bar/index.tsx
@@ -100,6 +100,11 @@ export default function AiAssistantBar( {
 	const { removeNotice } = useDispatch( noticesStore );
 
 	const onSend = useCallback( () => {
+		// Do not send the request if the input value is empty.
+		if ( ! inputValue?.length ) {
+			return;
+		}
+
 		// Remove previous error notice.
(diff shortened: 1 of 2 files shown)


lhkowalski (approved):
Working as expected.

Before the change, I can see the console error:

Screenshot 2023-08-18 at 15 00 44

After the change, nothing happens and the error does not appear in the console.

projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/jetpack-contact-form/components/ai-assistant-bar/index.tsx
```diff
@@ -100,6 +100,11 @@ export default function AiAssistantBar( {
 	const { removeNotice } = useDispatch( noticesStore );
 
 	const onSend = useCallback( () => {
+		// Do not send the request if the input value is empty.
+		if ( ! inputValue?.length ) {
```
CGastrell, line 104: What about a min length value? Like... 7 (the length of "contact" word). That should avoid getting nonsense prompts like single letters or numbers, or people trying to greet the AI and getting a form in response.

PR #32591, opened 2023-08-18
Launchpad: Use task ID to validate status update request

valterlorran:
## Proposed changes:

* With the changes introduced on https://github.com/Automattic/jetpack/pull/31471 we should be able to update a task with it's task ID and not the ID map prop. However, the request to update the task status used the method `get_task_key` to validate whether the incoming task was valid. This would make it impossible to update tasks by their ID if it contains the id_map prop.
* Instead of using the id_map to check if the incoming task is valid, we check by the ID.

### Other information:

## Testing instructions:

* Apply this patch to your Sandbox
* Create a new site
* Navigate to the Developer Console
* Make a PUT request to `WP REST API` -> `wpcom/v2` -> `/sites/:siteId/launchpad`, with the following body:
```
checklist_statuses = { "connect_social_media": true }
```
* The request should be completed without errors

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/changelog/Use the task id to validate the task update request
+++ b/projects/packages/jetpack-mu-wpcom/changelog/Use the task id to validate the task update request
@@ -1,0 +1,4 @@
+Significance: patch
+Type: added
+
+Use the task id to validate the task update request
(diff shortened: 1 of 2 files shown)


andres-blanco (changes requested):
I've tested successfully on console, but WordPress.com Tests are Failing D119375-code. 
This should be fixed before merging.

andres-blanco (approved):
Looking good! Also the tests from WPCOM are passing.

image

PR #32607, opened 2023-08-21
Subscriptions: Remove newsletter check for subscribe modal

edanzer:
## Proposed changes:

We currently check for site_intent === newsletter or theme === lettre before loading any code for the subscribe modal. This PR removes this check. Note that: 
   - The modal will still show only WordPress.com sites, not self hosted sites.
   - The modal will not show on the frontend until/unless a user enables it. The option to enable the modal is off by default.
   - We still need to remove a check in Calypso to show the 'Enable subscribe modal' option as well (https://github.com/Automattic/wp-calypso/pull/80703)

### Other information:

## Jetpack product discussion

None.

No.

## Testing instructions:

We want to test modal can now be enabled on any WordPress.com simple site (not just those with newsletter site_intent or Lettre theme). To check this, you'll need to remove the extra check in Calypso as well. 

1) Setup
   - Test on any WordPress.com simple site that does _not_ have newsletter / Lettre. 
   - Run the bin command provided by github below to load this branch in your sandbox. Sandbox public API and the domain of your test site. 
   - Checkout and start the trunk branch of calypso. Temporarily commend out this isNewsletterSite check to ensure the option will show in calypso with PR https://github.com/Automattic/wp-calypso/pull/80703
   - The modal should load.
   - Test also using non-FSE theme; the theme should still load.

2) Test
   - Access your site's dashboard via calypso localhost, and go to Settings > Reading > Newsletter. Confirm the 'Enable subscribe modal' option is visible.
   - Enable the modal and save. 
   - In a non-logged in browser (won't see the modal if logged in), go to any single blog post on the front end, scroll, pause, and confirm the modal shows. 

3) As a safety check, load this branch in your Jetpack dev environment, and confirm that the option to show the modal still does not show, and also that the modal does not load on any frontend single post.

### DIFF
--- a/projects/plugins/jetpack/changelog/update-remove-subscribe-modal-check
+++ b/projects/plugins/jetpack/changelog/update-remove-subscribe-modal-check
@@ -1,0 +1,4 @@
+Significance: minor
+Type: enhancement
+
+Subscriptions: Remove newsletter check for subscribe modal.
--- a/projects/plugins/jetpack/changelog/update-remove-subscribe-modal-check
+++ b/projects/plugins/jetpack/changelog/update-remove-subscribe-modal-check
@@ -1,4 +1,4 @@
 Significance: minor
 Type: enhancement
 
-Subscriptions: Remove newsletter check for subscribe modal.
+Subscriptions: Remove newsletter flow+theme and FSE theme check for the subscribe modal. Leave WP.com check.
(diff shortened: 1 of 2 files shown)


simison (dismissed):
👍  WP.com check remains so this isn't launched at Jetpack yet.

simison:
I removed the FSE check as well since the modal itself works fine on FSE themes, here's Mayland for example:

image

The ability to edit the modal isn't possible in non-FSE theme though, but that gating is handled in Calypso where we just don't show the edit link in that case.

simison (approved):

simison (changes requested):
On further testing, there's an issue I'm invesigating and testing with https://cicerosletters.wordpress.com/2023/08/15/new-cicero/

```
Fatal error: Uncaught Error: Call to undefined method Jetpack_Memberships::get_connected_account_id()
```

Edit, solved in https://github.com/Automattic/jetpack/pull/32620

simison (approved):

lezama (approved):

PR #32610, opened 2023-08-21
CRM: Automations expose attributes to frontend

gogdzl:
This PR addresses the need to expose attributes to the frontend.
 As it stands, we only have internal IDs/keys for our attributes (such as operators, values, fields, etc.). This means that to communicate with the front-end, we require a translation layer to bridge our base model's internal architecture and the frontend interface. To address this, this PR introduces a new class named Attribute_Definition, designed to encapsulate the rendering details for each of our steps.

*NOTE:*
- Conditions requiring a field list are annotated with TODO. Currently, we lack a comprehensive list of potential fields for our objects (e.g., contacts, invoices, quotes, and transactions) that can be utilized in automations. Addressing this is beyond the scope of this PR.
- @cleacos recommended renaming `Attribute_Definition` to `Step_Attribute`. During implementation, I found this occasionally confusing since we already possess a property named `$attribute` within our steps. This property (`$attribute`) remains unchanged and continues to store the selected attributes. Meanwhile, `Attribute_Definitions` represent the viable selections available.

## Proposed changes:
* Introduces `Attribute_Definition` class.
* Adds a title to our steps.

### Other information:

## Jetpack product discussion
pbhBOz-3UQ-p2
https://github.com/Automattic/zero-bs-crm/issues/3285

No.

## Testing instructions:
* Review the implementation of the Attribute_Definitions class.
* Ensure that all frontend requirements for rendering conditions are addressed in this PR.
* Thoroughly verify the modifications made in our tests.

### DIFF
--- a/projects/plugins/crm/src/automation/class-attribute-definition.php
+++ b/projects/plugins/crm/src/automation/class-attribute-definition.php
@@ -50,14 +50,6 @@ class Attribute_Definition {
 	const TEXT = 'text';
 
 	/**
-	* Represents a text input.
-	*
-	* @since $$next-version$$
-	* @var string
-	*/
-	const HIDDEN = 'hidden';
-
-	/**
 	* Represents a date input.
 	*
 	* @since $$next-version$$
--- a/projects/plugins/crm/src/automation/class-attribute-definition.php
+++ b/projects/plugins/crm/src/automation/class-attribute-definition.php
@@ -1,0 +1,261 @@
+<?php
+/**
+ * Attribute Definition
+ *
+ * @package automattic/jetpack-crm
+ * @since $$next-version$$
+ */
+
+namespace Automattic\Jetpack\CRM\Automation;
+
+/**
+ * Step Attribute.
+ *
+ * The Step Attribute represents
+ *
+ * @since $$next-version$$
+ */
+class Attribute_Definition {
+
+	/**
+	* Represents a dropdown selection input.
+	*
+	* @since $$next-version$$
+	* @var string
+	*/
+	const SELECT = 'select';
+
+	/**
+	* Represents a checkbox input.
+	*
+	* @since $$next-version$$
+	* @var string
+	*/
+	const CHECKBOX = 'checkbox';
+
+	/**
+	* Represents a textarea input.
+	*
+	* @since $$next-version$$
+	* @var string
+	*/
+	const TEXTAREA = 'textarea';
+
+	/**
+	* Represents a text input.
+	*
+	* @since $$next-version$$
+	* @var string
+	*/
+	const TEXT = 'text';
+
+	/**
+	* Represents a text input.
+	*
+	* @since $$next-version$$
+	* @var string
+	*/
+	const HIDDEN = 'hidden';
+
+	/**
+	* Represents a date input.
+	*
+	* @since $$next-version$$
+	* @var string
+	*/
+	const DATE = 'date';
+
+	/**
+	* Represents a date and time input.
+	*
+	* @since $$next-version$$
+	* @var string
+	*/
+	const DATETIME = 'datetime';
+
+	/**
+	* Represents a numerical input.
+	*
+	* @since $$next-version$$
+	* @var string
+	*/
+	const NUMBER = 'number';
+
+	/**
+	* Represents a password input.
+	*
+	* @since $$next-version$$
+	* @var string
+	*/
+	const PASSWORD = 'password';
+
+	/**
+	 * The slug (key) that identifies this attribute.
+	 *
+	 * @since $$next-version$$
+	 * @var string
+	 */
+	protected $slug;
+
+	/**
+	 * The title (label) for this attribute.
+	 *
+	 * @since $$next-version$$
+	 * @var string
+	 */
+	protected $title;
+
+	/**
+	 * The description for this attribute.
+	 *
+	 * @since $$next-version$$
+	 * @var string
+	 */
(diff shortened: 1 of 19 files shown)


projects/plugins/crm/src/automation/class-attribute-definition.php
```diff
@@ -0,0 +1,261 @@
...
+	*/
+	const SELECT = 'select';
+
+	/**
+	* Represents a checkbox input.
+	*
+	* @since $$next-version$$
+	* @var string
+	*/
+	const CHECKBOX = 'checkbox';
```
robertf4, line 34: Should these constants be a different class? Something like an enum of input types?

projects/plugins/crm/src/automation/commons/conditions/contacts/class-contact-field-changed.php
```diff
@@ -18,26 +19,37 @@
...
+
+		// TODO: Fetch automation fields from our DAL.
+		$contact_fields = array(
+			'id'       => __( 'ID', 'zero-bs-crm' ),
+			'fname'    => __( 'First Name', 'zero-bs-crm' ),
+			'lname'    => __( 'Last Name', 'zero-bs-crm' ),
+			'fullname' => __( 'Full Name', 'zero-bs-crm' ),
+		);
+
+		$this->valid_operators = array(
```
robertf4, line 39: I wonder if operators such as these could be declared and reused from somewhere so that the copy for them doesn't have to be repeated in each condition.

gogdzl: I had the same feeling. I will try to improve this!

cleacos: It doesn't seem trivial to improve it in a simple way. I think here is fine to "duplicate" it on each condition. In the end, they are their own operators from each condition implementation... and only a few lines.

Maybe we could have commons operators list in Base_Condition. It's not worth trying to optimize it now. Let's do it in another PR later.

projects/plugins/crm/src/automation/commons/conditions/invoices/class-invoice-field-contains.php
```diff
@@ -18,27 +19,36 @@
...
 	 */
-	private $valid_attributes = array(
-		'field',
-		'operator',
-		'value',
-	);
+	public function __construct( array $step_data ) {
+		parent::__construct( $step_data );
+
+		// TODO: Fetch automation fields from our DAL.
```
cleacos, line 31: For this to-do task, we could use the `objectModel`, which has all the information and label translation done. We have to add a method to retrieve the model, `get_object_model()`.

projects/plugins/crm/src/automation/class-attribute-definition.php
cleacos: Do you refer to the PHP enumeration? This was added from PHP 8, and we are compatible until PHP 7.2.

projects/plugins/crm/src/automation/commons/conditions/invoices/class-invoice-field-contains.php
gogdzl: This can be done in a later PR, agree?

projects/plugins/crm/src/automation/class-attribute-definition.php
```diff
@@ -0,0 +1,261 @@
...
+	*/
+	const TEXT = 'text';
+
+	/**
+	* Represents a text input.
+	*
+	* @since $$next-version$$
+	* @var string
+	*/
+	const HIDDEN = 'hidden';
```
cleacos: For our use case, step attributes, the hidden should not be an input for the end user.

projects/plugins/crm/src/automation/class-attribute-definition.php
robertf4: Ah, in that case disregard this 😄

projects/plugins/crm/src/automation/class-attribute-definition.php
gogdzl: Removed in 99ecac9.

projects/plugins/crm/src/automation/class-attribute-definition.php
cleacos: 👍

projects/plugins/crm/src/automation/class-base-step.php
```diff
@@ -112,7 +150,20 @@ abstract public static function get_slug(): string;
...
+	/**
+	 * Set the title of the step.
+	 *
+	 * @since $$next-version$$
+	 *
+	 * @param string $title Set title of the step.
+	 */
+	public static function set_title( string $title ) {
+		self::$title = $title;
+	}
```
cleacos: I thought you would revert it, and leave the original abstract static get_title method.

cleacos (changes requested):
One test is failing:

>  Test if an exception is being correctly thrown for wrong operators for transitional status.
>  
>  Failed asserting that exception of type "Automattic\Jetpack\CRM\Automation\Automation_Exception" is thrown.

If we implement `get_title` in Base_Step, the rest of the step implementation will overwrite it. If we can leave as it was, as static method/value, since all the step parameters are designed using only getters and not setters.

cleacos (approved):
Well, the code looks good. Tests passed.

LGTM!

PR #32615, opened 2023-08-21
Newsletters: Access panel with Paywall Block 

lezama:
Screenshot 2023-08-22 at 20 49 21

## Proposed changes:

* Rename Newsletter Visibility panel to Access.
* When the Paywall block is present, display the Paywall Icon, block name and description in the Access panel.
* Add link to select/focus the Paywall block.
* Fix "div inside p"  js warning.
* Remove notices in favor of counts

### Other information:

## Testing instructions:

* Add and Remove a Paywall Block
* Click on the Paywall block icon/name inside the Access 
* Try the different subscriber options


projects/plugins/jetpack/extensions/shared/memberships/settings.js
```diff
@@ -255,12 +165,13 @@ export function NewsletterAccessDocumentSettings( { accessLevel, setPostMeta } )
 				isLoading: isApiStateLoading(),
 				stripeConnectUrl: getConnectUrl(),
 				hasNewsletterPlans: getNewsletterProducts()?.length !== 0,
-				hasPaywallBlock: getBlocks().some( block => block.name === 'jetpack/paywall' ),
+				foundPaywallBlock: getBlocks().find( block => block.name === paywallBlockName ),
```
simison, line 167: Nit; I did like the previous `hasPaywallBlock` as a const name little better, but either works. :-) Maybe even `postHasPaywallBlock` to be extra clear?

projects/plugins/jetpack/extensions/shared/memberships/settings.js
```diff
@@ -219,16 +128,17 @@ export function NewsletterAccessRadioButtons( {
...
 							>
-								{ accessOptions[ key ].label }
+								{ accessOptions[ key ].label }{ ' ' }
+								{ key !== accessOptions.everybody.key && (
+									<>
+										{ ' ' }
+										{ sprintf(
+											'(%1$s)',
+											getReachForAccessLevelKey( key, emailSubscribers, paidSubscribers )
+										) }
```
simison: This whole section with multiple empty strings, `sprintf` and whatnot felt complicated to grok at glance but I don't have great alternative. 🤔  I had something like this in mind — separating the presentation from the rest of noise:

```js
const accessLabel = accessOptions[ key ].label;
const reach = key !== accessOptions.everybody.key ? getReachForAccessLevelKey( key, emailSubscribers, paidSubscribers ) : false;
```
```jsx
<label>
    { accessLabel }
    { subscriberReach && ` (${ subscriberReach })` }
</label>
```

simison:
Not a blocker but should be improved soon. — on mobile:

- "Click" term feels off
- The tap area is close to radio button which can cause accidental taps.
- Most importantly, the sidebar stays open so it looks like nothing happens since I can't see the editor.

Screenshot 2023-08-23 at 15 12 53

simison (dismissed):
When selecting "paid subscribers",  pre-publish panel still indicates it goes out to "all subscribers". 

Should we also indicate the count here? (good follow up if so)

image
Screenshot 2023-08-23 at 15 16 16

simison (dismissed):
Small nits: 

- spacing+widow here that would be easy to fix
    image

- Mobile issues where the sidebar staying open is biggest problem. https://github.com/Automattic/jetpack/pull/32615#issuecomment-1689854859

- About some code readability I noted in PR

Otherwise this is now testing well, none of the above are blockers for merging.

Also needs a rebase.

simison (approved):

PR #32623, opened 2023-08-22
CRM: Automations, use the original trigger data during the workflow execution

cleacos:
Fixes: https://github.com/Automattic/zero-bs-crm/issues/3284

## Proposed changes:

This PR updates the Automation Engine, when executing workflows, we have to keep the original data through the workflow execution and transform it before each step execution and pass the data type transformed, if any.

I couldn't add a test because of some dependencies and issues with the Data_Type/Transformer code (it needs to be updated because some data format issues).

### Other information:

## Jetpack product discussion
N/A

N/A

## Testing instructions:

This change doesn't affect the current behavior, so everything should work as before.

- Check the code.
- Run the tests and check that everything passes.

### DIFF
--- a/projects/plugins/crm/src/automation/class-automation-engine.php
+++ b/projects/plugins/crm/src/automation/class-automation-engine.php
@@ -399,9 +399,7 @@ class Automation_Engine {
 
 		while ( $step_data ) {
 			try {
-				$step_slug = $step_data['slug'];
-
-				$step_class = $step_data['class_name'] ?? $this->get_step_class( $step_slug );
+				$step_class = $step_data['class_name'] ?? $this->get_step_class( $step_data['slug'] );
 
 				if ( ! class_exists( $step_class ) ) {
 					throw new Automation_Exception(
@@ -413,6 +411,8 @@ class Automation_Engine {
 
 				/** @var Step $step */
 				$step = new $step_class( $step_data );
+
+				$step_slug = $step->get_slug();
 
 				if ( isset( $step_data['attributes'] ) && is_array( $step_data['attributes'] ) ) {
 					$step->set_attributes( $step_data['attributes'] );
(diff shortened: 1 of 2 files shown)


projects/plugins/crm/src/automation/class-automation-engine.php
```diff
@@ -379,23 +379,27 @@ public function init_workflows() {
...
 		$this->get_logger()->log( sprintf( 'Executing workflow: %s', $workflow->name ) );
 
 		$step_data = $workflow->get_initial_step();
+
 		// Convert the trigger data into a data type instance.
-		// This might change while running steps if we have to transform the data
-		// to e.g. allow invoice triggers to work together with contact actions.
-		$data_type = $this->get_data_type_instance( $trigger::get_data_type(), $trigger_data );
+		try {
+			$trigger_data_type = $this->get_data_type_instance( $trigger::get_data_type(), $trigger_data );
```
cleacos, line 392: Here we get the original data type from the trigger and we will use it during the workflow execution.

projects/plugins/crm/src/automation/class-automation-engine.php
```diff
@@ -408,13 +412,15 @@ public function execute_workflow( Automation_Workflow $workflow, Trigger $trigge
...
+				$step_slug = $step->get_slug();
+
 				if ( isset( $step_data['attributes'] ) && is_array( $step_data['attributes'] ) ) {
 					$step->set_attributes( $step_data['attributes'] );
 				}
 
 				$this->get_logger()->log( '[' . $step->get_slug() . '] Executing step. Type: ' . $step->get_data_type() );
 
-				$data_type = $this->maybe_transform_data_type( $data_type, $step::get_data_type() );
+				$data_type = $this->maybe_transform_data_type( $trigger_data_type, $step::get_data_type() );
```
cleacos, line 423: Get the data type before the Step execution.

robertf4 (approved):
Looks good to me!

PR #32627, opened 2023-08-22
Boost: Prevent SEO plugins from breaking Critical CSS generation

### LINKED ISSUE #32596: Boost: Register URL parameters used by Boost
Yoast SEO removes query parameters via:
**Yoast SEO > Settings > Advanced > Crawl Optimization > Advanced: URL cleanup**

This is affecting critical CSS generation by Boost since they rely on `jb-generate-critical-css`. [According to Yoast](https://yoast.com/help/advanced-crawl-settings-permalink-cleanup/#h-what-s-a-registered-query-parameter), the plugin developers should register query parameters for them to not get removed. I am assuming they mean [`query_vars`](https://developer.wordpress.org/reference/hooks/query_vars/) filter.

So here is what needs to be done:
- Register `jb-generate-critical-css` using the `query_vars` filter.
- See if it still gets removed, in that case, find out which registration they mean.
- See if there are more query parameters like this and register them.

dilirity:
Fixes https://github.com/Automattic/jetpack/issues/32596

## Proposed changes:

* Add compatibility files for `Yoast SEO` and `All in One SEO` plugins.
* Compatibility prevents plugins from stripping `jb-generate-critical-css` from URLs, by using their filters.

### Other information:

## Jetpack product discussion

n/a

no

### Test that Critical CSS generation does not work

* Setup latest stable Boost (free);
* Add and activate Yoast SEO;
* Go to **Yoast SEO > Settings > Advanced > Crawl Optimization > Advanced: URL cleanup** and enable `Remove unregistered URL parameters`;
   CleanShot 2023-08-22 at 17 16 27
* Go to the Boost dashboard and regenerate Critical CSS;
* Generation should fail (you could also check the network tab, and there should be 301 redirects).

You can repeat the above, but instead of Yoast SEO, you can install All in One SEO, and enable the `Crawl Cleanup` from **All in on SEO > Search Appearance > Advanced** (make sure `Remove Query Args` is set to `Yes`).

CleanShot 2023-08-22 at 17 17 21

### Test that Critical CSS generation works

* Setup Boost (free) via Jetpack Beta;
* Checkout the `add/boost/seo-plugins-compatibility` branch;
* Follow the steps from the previous section, to enable the "query cleanup" feature of either Yoast SEO or All in One SEO;
* Go to the Boost dashboard and regenerate Critical CSS;
* Generation should work without a problem.

### DIFF
--- a/projects/plugins/boost/changelog/add-seo-plugins-compatibility
+++ b/projects/plugins/boost/changelog/add-seo-plugins-compatibility
@@ -1,4 +1,4 @@
 Significance: patch
 Type: fixed
 
-Adds the Critical CSS generation token (jb-generate-critical-css), to the allowed query args lists of Yoast SEO and All in One SEO. This prevents them from removing it and breaking Critical CSS generation.
+Critical CSS: Improved compatibility with Yoast SEO and All in One SEO to ensure smooth Critical CSS generation.
--- a/projects/plugins/boost/changelog/add-seo-plugins-compatibility
+++ b/projects/plugins/boost/changelog/add-seo-plugins-compatibility
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Adds the Critical CSS generation token (jb-generate-critical-css), to the allowed query args lists of Yoast SEO and All in One SEO. This prevents them from removing it and breaking Critical CSS generation.
--- a/projects/plugins/boost/compatibility/yoast.php
+++ b/projects/plugins/boost/compatibility/yoast.php
@@ -9,4 +9,4 @@ namespace Automattic\Jetpack_Boost\Compatibility\Yoast;
 
 // Add the Critical CSS generation query arg to Yoast's allowed query args list.
 // This prevents Yoast from removing the query arg, which breaks Critical CSS generation.
-add_filter( 'Yoast\WP\SEO\allowlist_permalink_vars', array( '\Automattic\Jetpack_Boost\Modules\Optimizations\Critical_CSS\Critical_CSS', 'add_critical_css_query_arg_to_list' ) );
+add_filter( 'Yoast\WP\SEO\allowlist_permalink_vars', array( '\Automattic\Jetpack_Boost\Modules\Optimizations\Critical_CSS\Generator', 'add_generate_query_action_to_list' ) );
--- a/projects/plugins/boost/compatibility/yoast.php
+++ b/projects/plugins/boost/compatibility/yoast.php
@@ -7,10 +7,6 @@
 
 namespace Automattic\Jetpack_Boost\Compatibility\Yoast;
 
-function prevent_c_css_query_arg_removal( $default_allowed_extravars ) {
-	$default_allowed_extravars[] = 'jb-generate-critical-css';
-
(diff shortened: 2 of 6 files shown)


projects/plugins/boost/changelog/add-seo-plugins-compatibility
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Adds the Critical CSS generation token (jb-generate-critical-css), to the allowed query args lists of Yoast SEO and All in One SEO. This prevents them from removing it and breaking Critical CSS generation.
```
haqadn: As this is going to the changelog, consider something like:
> Critical CSS: Improved compatibility with Yoast SEO and All in One SEO to ensure smooth Critical CSS generation.

projects/plugins/boost/compatibility/yoast.php
```diff
@@ -0,0 +1,12 @@
...
+ * Yoast SEO compatibility for Boost
+ *
+ * @package automattic/jetpack-boost
+ */
+
+namespace Automattic\Jetpack_Boost\Compatibility\Yoast;
+
+// Add the Critical CSS generation query arg to Yoast's allowed query args list.
+// This prevents Yoast from removing the query arg, which breaks Critical CSS generation.
+add_filter( 'Yoast\WP\SEO\allowlist_permalink_vars', array( '\Automattic\Jetpack_Boost\Modules\Optimizations\Critical_CSS\Critical_CSS', 'add_critical_css_query_arg_to_list' ) );
```
haqadn: I assume `query_vars` did not work, and Yoast relies on their own filter.

However, I think we should still look into using the filter alongside this one. This is because other less popular plugins may still rely on `query_vars` as it is a WP standard.

haqadn (commented):
I have tested this, and it works. However, I have some nitpicks in the form of comments.

projects/plugins/boost/compatibility/yoast.php
dilirity: `query_vars` isn't used to make this work in the context of both plugins - they both rely on custom logic and their own filters. While reading the ticket description, initially I also thought that we should use the filter. However, after several re-reads of the documentation, I figured out that with the word "registered", they just mean registered in the context of the plugin - via UI or code, not by query_args.

I also checked out the code for any references of `query_vars` in both plugins and they don't use it at all in the context of query arg removal.

projects/plugins/boost/changelog/add-seo-plugins-compatibility
dilirity: Thanks! I've updated the changelog.

projects/plugins/boost/compatibility/yoast.php
dilirity: After reading on query vars, I felt it was unnecessary to add it (since we're not working with WP_Query or using it for pretty permalinks).

However since WordPress is used in ways we can't predict, I've added the token to the WP query vars list (only if the get param is present). This also affects the filters of the plugins, as they use the updated method.

haqadn (approved):
👍🏼

PR #32527, opened 2023-08-16
AI Assistant: Specify input background color

dhasilva:
Fixes #32525

## Proposed changes:

* Just specifies the background color explicitly, as some themes may try to overwrite it.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* Add an AI Assistant block or extension
* Check that the background color remains white
* Install and activate the Livro theme
* Check that the background is white

| Before | After
| -|-
|  | 

The placeholder centralization issue is unrelated. Seems to be a problem with Gutenberg.

PR #32628, opened 2023-08-22
AI Extension: Specify input background color on form

dhasilva:
Follow-up to #32527

## Proposed changes:

* Applies the same fix from the AI Assistant block to the package component used by the Form extension

### Other information:

## Testing instructions:

* Use the Livro theme
* Add a Form
* Check that the AI input field has a white background

| Before | After
-|-
|  |


nunyvega (approved):
Works well!

PR #32629, opened 2023-08-22
Super Cache: fix PHP warning about NULL used in str_replace

donnchawp:
If you disabled caching for logged in users then a warning about about using NULL showed in the PHP error log:

> PHP Deprecated: str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in .../plugins/wp-super-cache/wp-cache-phase2.php on line 54

In this PR the code that defines $wp_cache_request_uri is moved up in wp-cache-phase1.php so it always runs. It wouldn't run if caching was disabled for logged in users due to the way checks were done.
Later on when that variable was used, functions received the value "NULL" and in the latest versions of PHP8 that caused a warning to show.

The linter complained the value was unsanitized but this code executes before WordPress is entirely loaded. It is eventually unslashed and sanitized in wp_cache_postload().

Related thread: https://wordpress.org/support/topic/php-deprecated-str_replace-passing-null/

## Proposed changes:
* Move code that defines $wp_cache_request_uri up so it runs all the time.
* Sanitize that variable in wp_cache_postload()

### Other information:

No

## Testing instructions:
* Using a host with PHP8.2* apply this patch.
* Disable caching for logged in users.
* Visit your site and you shouldn't see that warning notice.

### DIFF
--- a/projects/plugins/super-cache/changelog/fix-super_cache_wp_cache_request_uri_defined
+++ b/projects/plugins/super-cache/changelog/fix-super_cache_wp_cache_request_uri_defined
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Caching: make sure $wp_cache_request_uri is defined to avoid warnings about "NULL" parameters.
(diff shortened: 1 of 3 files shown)


haqadn (approved):
Tested and it works nice 👍🏼

PR #32499, opened 2023-08-16
Connection: adjust the way we render initial state

sergeymitr:
Fixes #32499

## Proposed changes:
Currently, trying to render connection initial state multiple times on the same page will lead to `null` returned instead of the valid JavaScript code. That `null` will be passed to `wp_add_inline_script()` and subsequently to `stripos()`, which does not support non-string values for the `$haystack` argument.

This PR changes the way we render connection initial state to prevent re-render attempts, and simplifies it for consuming plugins/packages.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion
p1690838332688659-slack-CDLH4C1UZ
1191179647901802-as-1205228449251052

## Does this pull request change what data or activity we track or use?
No.

## Testing instructions:
Open the browser console and confirm the global variable `JP_CONNECTION_INITIAL_STATE` exists:

1. Jetpack Dashboard
2. Edit a post
3. Jetpack Social
4. Activate Jetpack Social (free is enough), disable Jetpack, and go to the post editor
5. Jetpack Search
6. Jetpack Starter Plugin
7. “Move to WordPress.com”
8. VideoPress
9. My Jetpack
10. Jetpack Protect
11. Disconnect and deactivate Jetpack, activate it again and see the huge “Set up Jetpack” modal.

PR #32630, opened 2023-08-22
Ensure Initial_State::render() always returns a string

markjaquith:
## Proposed changes:

Updates `Initial_State::render()` to return an empty string instead of `null` when is it called a second time.

This is necessary because passing `null` to `wp_add_inline_script()` will result in a PHP deprecation notice, as WordPress blindly expects a string to be passed in.

This also brings the function into compliance with its PHPDoc return type, which is `string`.

## Testing instructions:

No special instructions. Just run unit tests and verify no PHP Deprecated notices are thrown.

No.

### DIFF
--- a/projects/packages/connection/changelog/php-81-null-inline-script-deprecation
+++ b/projects/packages/connection/changelog/php-81-null-inline-script-deprecation
@@ -1,0 +1,3 @@
+Significance: patch
+Type: other
+Comment: Ensure Initial_State::render() always returns a string
--- a/projects/packages/connection/src/class-initial-state.php
+++ b/projects/packages/connection/src/class-initial-state.php
@@ -53,7 +53,7 @@ class Initial_State {
 	 */
 	public static function render() {
 		if ( self::$rendered ) {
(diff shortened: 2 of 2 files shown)


oskosk:
Thanks @markjaquith. We'll check it out

sergeymitr:
Hi @markjaquith, thanks for working on this.
We merged #32499 last week that fixes the deprecation notice.

We introduced [new method](https://github.com/Automattic/jetpack/blob/cbb30bc15d0524156a28536668c16a59295207ba/projects/packages/connection/src/class-initial-state.php#L69) `render_script()` which is now the preferred way for rendering Connection initial state as it fixes the problem you found and also simplifies things a bit.

We usually try to avoid return type changes to preserve backward compatibility, e.g. if there's `null === $initial_state` running somewhere.

> This also brings the function into compliance with its PHPDoc return type, which is string.

Good catch, we'll need to get the PHPDoc updated.

I'm going to close the PR, but I appreciate you dedicating the time to fix this.

PR #32633, opened 2023-08-22
Docker: add user switching

kraftbj:
Adds John Blackbourn's User Switching plugin as a default plugin for our Docker instance. He's the author of Query Monitor and WP Control, which we already use, an active Core contributor, and former Automattician.

Additionally, automatically generate an user at each of the default WordPress roles (editor, author, subscriber, contributor). 

This combination will allow for easier testing of different user roles (or logged out, a feature of User Switching) while developing locally.

### Other information:

## Jetpack product discussion

Inspired by pdWQjU-tq-p2

No.

## Testing instructions:

* Assuming you have the Jetpack monorepo setup as usual and Docker installed.
* jetpack docker up -d # if docker is not yet up.
* jetpack docker uninstall # this will delete your existing Docker site.
* jetpack docker install # this will run the install.sh script with the changes.
* jetpack docker user list
* See five users: wordpress (the existing default admin), author, contributor, editor, subscriber.
* Verify that "Switch to" is an option when hovering over their line item on the wp-admin user list.

### DIFF
--- a/tools/docker/bin/install.sh
+++ b/tools/docker/bin/install.sh
@@ -22,15 +22,15 @@ wp --allow-root core install \
 wp --allow-root option update blog_public 0
 
 if [ "$COMPOSE_PROJECT_NAME" == "jetpack_dev" ] ; then
-		# Add editor, author, contributor, and subscriber users
-  	roles=("author" "editor" "subscriber" "contributor")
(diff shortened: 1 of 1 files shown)


anomiex (approved):
Works. Let's do it.

PR #32635, opened 2023-08-22
AI Extension: Enhance blocks parsing flickering with child blocks

### LINKED ISSUE #32516: AI Extension: Asking for required form fields is too inconsistent
Asking for specific fields to be required or optional is inconsistent. Prompt tweaks are necessary to address that specifically.

ref pedxs5-i3-p2#comment-123

dhasilva:
Fixes #32516
Test with D119992-code

## Proposed changes:

* Changes the way the blocks are verified to avoid flickering and multiple history entries with invalid child blocks

### Other information:

## Testing instructions:

* Test with D119992-code
* Check that fields with children, like multiple choice inputs, do not flicker when updated and only one history change is added when a new child is inserted


retrofox:
In your description, you mention **_and lack of wp: in the block name (e.g. `)_.** I don't see where this PR updates the block name.

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
```diff
@@ -270,15 +273,55 @@ Strong requirements:
...
 	- \`Phone Number Field\`: <!-- wp:jetpack/field-telephone {"label":FIELD_LABEL,"required":IS_REQUIRED,"requiredText":REQUIRED_TEXT,"placeholder":PLACEHOLDER_TEXT} /-->
 	- \`URL Field\`: <!-- wp:jetpack/field-url {"label":FIELD_LABEL,"required":IS_REQUIRED,"requiredText":REQUIRED_TEXT,"placeholder":PLACEHOLDER_TEXT} /-->
-	- \`Multiple Choice (Checkbox)\`: <!-- wp:jetpack/field-checkbox-multiple {"label":FIELD_LABEL,"required":IS_REQUIRED,"requiredText":REQUIRED_TEXT, "options": [OPTION_ONE, OPTION_TWO, OPTION_THREE]} /-->
-	- \`Single Choice (Radio)\`: <!-- wp:jetpack/field-radio {"label":FIELD_LABEL,"required":IS_REQUIRED,"requiredText":REQUIRED_TEXT, "options": [OPTION_ONE, OPTION_TWO, OPTION_THREE]} /-->
+	- \`Multiple Choice (Checkbox)\`: <!-- wp:jetpack/field-checkbox-multiple {"label":FIELD_LABEL,"required":IS_REQUIRED,"requiredText":REQUIRED_TEXT, "options": []} --><!-- wp:jetpack/field-option-checkbox {"label":"OPTION_ONE"} /--><!-- wp:jetpack/field-option-checkbox {"label":"OPTION_TWO"} /--><!-- /wp:jetpack/field-checkbox-multiple -->
```
retrofox: Why do you replace the `options` attribute with children blocks here?

retrofox:
I've asked, "can you add a select with types of pizza?"

image

Is it something that these changes should address?

lhkowalski:
After we moved the form prompt building to the backend, I think this change may take place there. We also discussed about the need to expand the list of messages, adding new messages to the list like you did. I am working on a follow up to the backend patch to allow this.

dhasilva:
> In your description, you mention and lack of wp: in the block name (e.g. `). I don't see where this PR updates the block name.

Not in the block name, but in the rules. They were lacking the `wp:` in the response sometimes, which made the block invalid and removed. That is why _some_ of the fields are removed.

dhasilva:
Moving the changes to the backend.

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
retrofox: Did we migrate the prompt to the server side with or without these changes?

retrofox: Ignore me... checking the patch

retrofox (approved):
The PR handles the flickering issues when the app inserts the block, but let's keep in mind it doesn't solve the root problem. It seems to hide it by adding a new process.

image

Anyway, I think it's a nice improvement that will impact improve the UX :shipit:

PR #32639, opened 2023-08-23
Newsletter Categories: Add feature flag helper

### LINKED ISSUE #80916: Newsletter Categories: Create helper function that checks feature flag & blog sticker
### Details
Create a helper function that checks both the feature flag & blog sticker & that returns true if either one of them is set to true

### Checklist

- [ ] Helper is present
- [ ] Helper returns true when feature flag is set
- [ ] Helper returns true when blog sticker is applied

### Related
https://github.com/Automattic/wp-calypso/issues/80910

TimBroddin:
Closes https://github.com/Automattic/wp-calypso/issues/80916

## Proposed changes:
This PR adds a helper function that checks if either a blog sticker is present (on Simple/Atomic), or a feature flag is set.

To enable this, we can add a filter called jetpack_newsletter_categories in jetpack-mu-wpcom.

We'll also need something similar for inside Gutenberg (client-side).

### Other information:

No

## Testing instructions:
- Build the Jetpack plugin & test on Docker
- Add a Newsletter Sign up block
- Verify that there's no regressions & the block is rendered correctly on the frontend

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -1070,18 +1070,3 @@ function get_paywall_simple() {
 <!-- /wp:columns -->
 ';
 }
-
-/**
- * Determine if the newsletter categories are enabled for the site.
- *
- * This function checks whether the blog has the 'newsletter_categories' sticker enabled.
- * If the sticker is enabled, the function will return true.
- * If not, it applies the 'jetpack_newsletter_categories' filter and returns the result (default is false).
- */
-function newsletter_categories_enabled_for_site() {
-	if ( function_exists( 'has_blog_sticker' ) && has_blog_sticker( 'newsletter_categories' ) ) {
-		return true;
-	}
-
-	return apply_filters( 'jetpack_newsletter_categories', false );
-}
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -1079,7 +1079,7 @@ function get_paywall_simple() {
  * If not, it applies the 'jetpack_newsletter_categories' filter and returns the result (default is false).
  */
 function newsletter_categories_enabled_for_site() {
-	if ( function_exists( 'has_blog_sticker' ) && has_block_sticker( 'newsletter_categories' ) ) {
+	if ( function_exists( 'has_blog_sticker' ) && has_blog_sticker( 'newsletter_categories' ) ) {
 		return true;
 	}
 
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -1070,3 +1070,18 @@ function get_paywall_simple() {
 <!-- /wp:columns -->
 ';
 }
+
+/**
+ * Determine if the newsletter categories are enabled for the site.
+ *
+ * This function checks whether the blog has the 'newsletter_categories' sticker enabled.
+ * If the sticker is enabled, the function will return true.
+ * If not, it applies the 'jetpack_newsletter_categories' filter and returns the result (default is false).
+ */
+function newsletter_categories_enabled_for_site() {
+	if ( function_exists( 'has_blog_sticker' ) && has_block_sticker( 'newsletter_categories' ) ) {
+		return true;
+	}
+
+	return apply_filters( 'jetpack_newsletter_categories', false );
+}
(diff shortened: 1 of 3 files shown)


mashikag (dismissed):
Looks good to me. 👍

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -904,3 +904,18 @@ function get_paywall_simple() {
...
+
+/**
+ * Determine if the newsletter categories are enabled for the site.
+ *
+ * This function checks whether the blog has the 'newsletter_categories' sticker enabled.
+ * If the sticker is enabled, the function will return true.
+ * If not, it applies the 'jetpack_newsletter_categories' filter and returns the result (default is false).
+ */
+function newsletter_categories_enabled_for_site() {
+	if ( function_exists( 'has_blog_sticker' ) && has_block_sticker( 'newsletter_categories' ) ) {
```
yansern:
```suggestion
	if ( function_exists( 'has_blog_sticker' ) && has_blog_sticker( 'newsletter_categories' ) ) {
```
Might be a typo here.

yansern (dismissed):

TimBroddin:
Stupid typo 🫠 Thanks for the review @yansern

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -904,3 +904,18 @@ function get_paywall_simple() {
...
+ * This function checks whether the blog has the 'newsletter_categories' sticker enabled.
+ * If the sticker is enabled, the function will return true.
+ * If not, it applies the 'jetpack_newsletter_categories' filter and returns the result (default is false).
+ */
+function newsletter_categories_enabled_for_site() {
+	if ( function_exists( 'has_blog_sticker' ) && has_blog_sticker( 'newsletter_categories' ) ) {
+		return true;
+	}
+
+	return apply_filters( 'jetpack_newsletter_categories', false );
```
jeherve: I would recommend adding a docblock for that filter, so our Codex is automatically updated with this new hook.

```suggestion
	/**
	 * Force enable or disable Jetpack's Newsletter Categories feature.
	 *
	 * @module subscriptions
	 * @since $$next-version$$
	 *
	 * @param bool $jetpack_newsletter_categories Should Jetpack's Newsletter Categories be enabled. Default to false.
	 */
	return apply_filters( 'jetpack_newsletter_categories', false );
```

jeherve: Maybe that hook should be a bit more verbose, so folks know what it exactly does (enable the feature)?

jeherve (changes requested):
I left a few comments below.

Since this is going to be used for more than the block, I wondered if this wouldn't be better placed in `modules/subscriptions.php` instead?

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
yansern: > Maybe that hook should be a bit more verbose, so folks know what it exactly does (enable the feature)?

Perhaps `jetpack_newsletter_categories_enabled`?

Looking at the [Codex](https://developer.jetpack.com/) all forms of `disabled/enabled/disable/etc.` suffixing/prefixing are used so there's no set convention.

yansern:
To follow-up:
- [ ] Move the `newsletter_categories_enabled_for_site` function to `modules/subscriptions.php`
- [ ] Add docblock.
- [ ] Rename to `jetpack_newsletter_categories_enabled`.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
jeherve: > Perhaps jetpack_newsletter_categories_enabled?

I like it 👍

TimBroddin:
@yansern I think this name can be confusing because a site owner can also enable/disable newsletter categories, and this setting controls it from our side. Would it be better to name this `jetpack_newsletter_categories_feature_flag_enabled`?

yansern:
> @yansern I think this name can be confusing because a site owner can also enable/disable newsletter categories, and this setting controls it from our side. Would it be better to name this `jetpack_newsletter_categories_feature_flag_enabled`?

Maybe we can omit the `newsletter_categories_enabled_for_site` helper function altogether and rely on just applying the filter to get the value back, e.g.

```
$is_enabled = apply_filters( 'jetpack_newsletter_categories_enabled', false );
```

I can see there are 3 things that want to hook into the `jetpack_newsletter_categories_enabled` filter, and we can [order them by priority](https://developer.wordpress.org/reference/functions/add_filter/#parameters).
- check site options (below default priority = 9, execute first)
- ~check feature flag (priority = 8, execute after the previous one)~ (i don't think we need this)
- check blog sticker (priority = 7, execute last)

TimBroddin:
@yansern I updated it with your suggestions. Could you take a look?

TimBroddin:
Abandoning this one, as discussed with @yansern here: p1694079644799039-slack-C02TCEHP3HA

PR #32641, opened 2023-08-23
Fix a Global Styles UI performance issue

BogdanUngureanu:
Instead of loading the Jetpack blocks CSS for the Site Editor using the admin_init hook, we'll switch to the `enqueue_block_editor_assets` hook. In this manner, the CSS script is loaded in the iframe, instead of the main site-editor.php page.

By doing so, the UI for Global Styles is more responsive.

Fixes https://github.com/Automattic/dotcom-forge/issues/3565 

## Proposed changes:

* Use `enqueue_block_editor_assets` hook instead of `admin_init`

### Other information:

## Testing instructions:

* Use Jetpack beta on an Atomic site
* Make sure that your theme is TT3 (for easier testing)
* Go to the site editor and open the Global Styles UI.
* Go to Layout and play with the control. Notice that the feedback is way faster than before.
* To better notice the change, switch to the slider button (the right icon) and slide the margins/padding. The input value will update faster.

**Check for regressions:**

Make sure that the styles for payment blocks are still working fine (e.g. donations block).

### DIFF
--- a/projects/plugins/jetpack/class.jetpack.php
+++ b/projects/plugins/jetpack/class.jetpack.php
@@ -691,6 +691,11 @@ class Jetpack {
 		require_once JETPACK__PLUGIN_DIR . 'class.jetpack-gutenberg.php';
 		add_action( 'plugins_loaded', array( 'Jetpack_Gutenberg', 'load_independent_blocks' ) );
 		add_action( 'plugins_loaded', array( 'Jetpack_Gutenberg', 'load_block_editor_extensions' ), 9 );
+		/**
+		 * We've switched from enqueue_block_editor_assets to enqueue_block_assets because the assets with the former are loaded on the main site-editor.php.
+		 *
+		 * With the latter, the assets are now loaded in the SE iframe; the implementation is now faster because Gutenberg doesn't need to inject the assets in the iframe on client-side.
+		 */
 		add_action( 'enqueue_block_assets', array( 'Jetpack_Gutenberg', 'enqueue_block_editor_assets' ) );
 		add_filter( 'render_block', array( 'Jetpack_Gutenberg', 'display_deprecated_block_message' ), 10, 2 );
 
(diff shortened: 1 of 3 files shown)


projects/plugins/jetpack/class.jetpack.php
```diff
@@ -691,7 +691,7 @@ private function __construct() {
 		require_once JETPACK__PLUGIN_DIR . 'class.jetpack-gutenberg.php';
 		add_action( 'plugins_loaded', array( 'Jetpack_Gutenberg', 'load_independent_blocks' ) );
 		add_action( 'plugins_loaded', array( 'Jetpack_Gutenberg', 'load_block_editor_extensions' ), 9 );
-		add_action( 'enqueue_block_editor_assets', array( 'Jetpack_Gutenberg', 'enqueue_block_editor_assets' ) );
+		add_action( 'enqueue_block_assets', array( 'Jetpack_Gutenberg', 'enqueue_block_editor_assets' ) );
```
mmtr, line 699: Jetpack needs to also support the previous version of WordPress (6.2) where this hook doesn't enqueue styles in the editor, so we still need `enqueue_block_editor_assets` for some scenarios.

BogdanUngureanu: Both hooks were introduced in WP 5.0.0. I guess we just need to check if it breaks in older versions of WP.

mmtr: Yes, the `enqueue_block_assets` hook has been available for a while, but I think only as of WP 6.3 enqueues styles to the iframed site editor. Or did I misunderstand what changed in WP 6.3.?

mmtr: NVM, manually tested on WP 6.2 and this seems to work!

mmtr (approved):
Thanks @BogdanUngureanu! Great job investigating the issue and elegant fix 💥

jeherve:
I'm catching up with this but I was wondering if this may be what caused the issue reported here:
p1693004089973979-slack-C010KDAPG49

What is the impact of this change when the editor needs to be enqueued on the frontend, like on p2? Could it stop blocks from appearing in the frontend editor?

mmtr:
@jeherve Yeah, nice catch. We completely forgot that in some contexts (e.g. P2) the editor can be loaded on the frontend, so we'd need to add a workaround for that.

PR #32647, opened 2023-08-23
Newsletters: Fix Non WPCOM subscribers paywalled emails

lezama:
## Proposed changes:

* Fix email only (non wpcom users) subscribers receiving a paywalled email when post has a paywall block but access is set to "All subscribers".

### Other information:

No

## Testing instructions:

* See instructions at D119666-code


edanzer (approved):
Tested this alongside the associated phab diff, and it is testing as expected. Looks like it may still need a rebase, and one other test is failing. But assuming those will be resolved, I've approved.

PR #32650, opened 2023-08-23
VideoPress: ensure private videos rendered by the divi module are playable

jgcaruso:
Fixes # https://github.com/Automattic/greenhouse/issues/1798

## Proposed changes:

* enqueue the token bridge in the render method of the VideoPress Divi to support private videos
* update the regex used to parse videopress urls to be consistent between the admin and front-end renderers

### Other information:

nope

## Testing instructions:

* On a test site with Divi installed, enable both Jetpack and Jetpack VideoPress plugins
* In the Divi page builder, add a VideoPress block/module and embed a video that is a private video (or make it private after embedding)
* the video should load in the editor
* publish the page
* the video should load on the published (non editor) page
* in an unauthenticated browser, you should see the "This video is private" message
* on a front-end (non-editor) page that does not include any videopress videos, the `token-bridge.js` script should not get loaded
* RETURN to trunk
* reload the page in both browsers, you will eventually get the "Post message timed out" error


pgk (approved):
Worked well!

jeherve (approved):
This tests well for me. 🚢

PR #32659, opened 2023-08-24
Paywall block: Add toolbar

lezama:
## Proposed changes:

Screenshot 2023-08-24 at 08 48 35

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Does this pull request change what data or activity we track or use?

No

## Testing instructions:

* Make sure it works 😄

PR #32663, opened 2023-08-24
Paywall Block: Disable Paid Subscribers button if no plans

lezama:
Context p1692749551691859-slack-C052XEUUBL4

## Proposed changes:

image

### Other information:

NO

### DIFF
--- a/projects/plugins/jetpack/changelog/update-paywall-toolbar-2
+++ b/projects/plugins/jetpack/changelog/update-paywall-toolbar-2
@@ -1,0 +1,4 @@
+Significance: minor
+Type: enhancement
+
+Paywall block: Disable Paid subscribers button if plans are not set up.
--- a/projects/plugins/jetpack/extensions/blocks/paywall/edit.js
+++ b/projects/plugins/jetpack/extensions/blocks/paywall/edit.js
@@ -20,6 +20,15 @@ function PaywallEdit( { className } ) {
 	const accessLevel = useAccessLevel( postType );
 	const [ , setPostMeta ] = useEntityProp( 'postType', postType, 'meta' );
 
+	const { stripeConnectUrl, hasNewsletterPlans } = useSelect( select => {
+		const { getNewsletterProducts, getConnectUrl } = select( 'jetpack/membership-products' );
+		return {
+			stripeConnectUrl: getConnectUrl(),
+			hasNewsletterPlans: getNewsletterProducts()?.length !== 0,
+		};
+	} );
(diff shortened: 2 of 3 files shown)


simison (approved):
Tests well; checked Stripe connected and disconnected.

Checked posts panel, block panel, and block toolbar.

I think this could be improved by adding the copy (without the link) from sidebar as a tooltip into the disabled button, but that's fine also in follow-up.

Screenshot 2023-08-24 at 17 19 01

crisbusquets:
Hi!

The gray used when `Paid subscribed` is not available isn't accessible. In the Figma file I used `#949494` (Gray600), which is also not accessible... `#757575` (Gray700)  is, but I'm unsure if it's too dark to convey `disabled`.

Screenshot 2023-08-24 at 18 42 29

What do you think?

simison:
> The gray used when Paid subscribed is not available isn't accessible. In the Figma file I used #949494 (Gray600), which is also not accessible... #757575 (Gray700) is, but I'm unsure if it's too dark to convey disabled.

@crisbusquets I think we're using regular toolbar UI components with disabled state in Gutenberg, so this should get fixed there. I wouldn't want us to override those styles as it tends to break eventually.

@jasmussen is there anything we could help with in core? E.g. just bump the grey variable somewhere. I [noticed some spacing issues](https://github.com/Automattic/jetpack/pull/32659#pullrequestreview-1593419229) with these components, too.

jasmussen:
Gray 600 has 3:1 contrast which is appropriate for UI, and Gray 700 has 4.5:1 contrast appropriate for text. 

Disabled buttons technically [do not need to meet any contrast ratio at all](https://www.w3.org/TR/WCAG21/#contrast-minimum):

> Text or images of text that are part of an inactive [user interface component](https://www.w3.org/TR/WCAG21/#dfn-user-interface-components), that are [pure decoration](https://www.w3.org/TR/WCAG21/#dfn-pure-decoration), that are not visible to anyone, or that are part of a picture that contains significant other visual content, have no contrast requirement.

To that end, the componentry and colors are correct, and the question is more one of: should the "Paid subscribers" button actually be disabled or not. That depends on the behavior — are these meant to be tabs, like the HTML block? 

tabs

Or is the button meant to be disabled until you purchase the feature that enables you to use the tabs? If yes, then another option is to keep it enabled, but invoke a modal dialog or confirm to inform you what you need to do to enable _the feature_. In other words, the fact that the _feature_ is disabled, might not equate well to the _button_ being disabled.

simison:
Thanks, very helpful advice. 🙇‍♂️

PR #32666, opened 2023-08-24
Newsletter categories: Render non-interactive newsletter category pills in the editor 

### LINKED ISSUE #80904: Newsletter Categories: Render non-interactive newsletter category pills in the editor
### Details

Render non-interactive newsletter category pills in the editor above the Subscribe block

### Checklist

- [ ] Render the categories, only if the site has the newsletter categories setting enabled

### Related

_No response_

TimBroddin:
Closes https://github.com/Automattic/wp-calypso/issues/80904

## Proposed changes:
- This adds the newsletter categories as pills in the editor.

CleanShot 2023-08-24 at 16 23 51@2x

### Other information:

## Testing instructions:
For completeness, please test this on WPCOM & Jetpack self hosted.

1. Checkout this PR and run the command below for WPCOM or build it and run it though JT
2. Mark some categories as newsletter categories: `POST /wpcom/v2/sites/<siteid>/newsletter-categories/<categoryid>` 
3. Create a new post/page & add the subscribe block
4. The newsletter categories should be above the input field


projects/plugins/jetpack/extensions/blocks/subscriptions/view.scss
```diff
@@ -1,3 +1,5 @@
+
+
```
phcp: Could you remove those extra lines?

projects/plugins/jetpack/extensions/blocks/subscriptions/hooks/use-newsletter-categories.js
```diff
@@ -0,0 +1,33 @@
...
+				setError( true );
+			} finally {
+				setLoading( false );
+			}
+		};
+
+		fetchData();
+	}, [] );
+
+	return { data, enabled: data.length, error, loading };
```
phcp: Just confirm, `enabled: data.length` means that the API should reflect the user choice on the Newsletter Settings page (to enable/disable newsletter categories), right?

phcp: I'm thinking of a scenario where the user enabled newsletter categories, selected a few of them, and saved; then the user decided to disable newsletter categories. In the current implementation, the selected categories are still there, so the API would return categories.

TimBroddin: You're right. In that case we need to extend the API endpoint to expose this setting.

TimBroddin: I added the extra field in D119783-code & updated the code of this PR to use it.

projects/plugins/jetpack/extensions/blocks/subscriptions/hooks/use-newsletter-categories.js
```diff
@@ -0,0 +1,35 @@
+import apiFetch from '@wordpress/api-fetch';
+import { select } from '@wordpress/data';
+import { useEffect, useState } from '@wordpress/element';
+import { CONNECTION_STORE_ID } from '../../../../../../js-packages/connection';
```
jeherve: Here it may be best to import from the published package instead:

```suggestion
import { CONNECTION_STORE_ID } from '@automattic/jetpack-connection';
import apiFetch from '@wordpress/api-fetch';
import { select } from '@wordpress/data';
import { useEffect, useState } from '@wordpress/element';
```

Better yet, since all we need here is the blog ID, we should be able to get it from the data passed in PHP, with `window?.Jetpack_Editor_Initial_State?.wpcomBlogId`

projects/plugins/jetpack/extensions/blocks/subscriptions/hooks/use-newsletter-categories.js
```diff
@@ -0,0 +1,35 @@
...
+	const [ error, setError ] = useState( false );
+	const [ loading, setLoading ] = useState( true );
+
+	useEffect( () => {
+		const blogId = select( CONNECTION_STORE_ID ).getBlogId();
+
+		const fetchData = async () => {
+			try {
+				const newsLetterCategories = await apiFetch( {
+					url: `https://public-api.wordpress.com/wpcom/v2/sites/${ blogId }/newsletter-categories`,
```
jeherve: We've typically not queried `public-api` directly in our blocks. Instead we query a local endpoint that in turn queries WordPress.com (or makes a direct request when we're on WordPress.com).  The local endpoints live in `projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/`. 

There is some documentation about it here:

https://github.com/Automattic/jetpack/blob/695768a21b1341684a4da54fa4abc39093cd7ed8/projects/plugins/jetpack/extensions/README.md#L262-L271

projects/plugins/jetpack/extensions/blocks/subscriptions/view.scss
```diff
@@ -96,3 +96,32 @@
...
+	.wp-block-jetpack-subscriptions__newsletter-category {
+		padding: 6px 10px;
+		font-family: "SF Pro Text", -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif;
+		font-size: 14px;
+		font-weight: 500;
+		line-height: 20px;
+		color: var(--studio-gray-50);;
+		border: 1px solid var(--studio-gray-5);
+		border-radius: var(--border-radius);
```
jeherve: I wonder if we could use WordPress colors here, for more consistency?

Here are some examples:
https://github.com/Automattic/jetpack/blob/3a3c4adc1c0913bac224ddb832713bd6101fe1ab/projects/plugins/jetpack/extensions/blocks/donations/view.scss#L28

projects/plugins/jetpack/extensions/blocks/subscriptions/hooks/use-newsletter-categories.js
TimBroddin: Thanks for the pointers once again @jeherve .

I've made the needed API changes: https://github.com/Automattic/jetpack/pull/32666/commits/596bcef21f4fc3f89fc44768f728e72f86cede54

I do have a question regarding defusion: When I test with the jetpack-downloader, I can see the endpoint file on my WP.COM Simple site. However, even after deleting the existing endpoint, I cannot get it to execute it using the WP.com console. Do files inside `wpcom-endpoints` not get loaded on Simple Sites?

It's not a problem for now, since everything works. Just interested.

jeherve: > Do files inside wpcom-endpoints not get loaded on Simple Sites?

Yeah, it's a bit confusing. You'd need to do something like D110282-code (check the last file being added, or other files in that directory for some more examples).

TimBroddin: Thank you! It's much cleared now. I'll create a separate patch for WPCOM, removing the endpoint, after the Jetpack release happens on WPCOM.

jeherve (changes requested):
Noting that you'll need to look at the failing js tests in `extensions/blocks/subscriptions/test/edit.js` before we can merge this:
https://github.com/Automattic/jetpack/actions/runs/6010883027/job/16303218143?pr=32666

TimBroddin:
> Noting that you'll need to look at the failing js tests in `extensions/blocks/subscriptions/test/edit.js` before we can merge this: https://github.com/Automattic/jetpack/actions/runs/6010883027/job/16303218143?pr=32666

Fixed it! And somehow managed to bring in all the labels while doing a rebase 🫠

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-newsletter-categories-list.php
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	/**
+	 * Gets the site's newsletter categories
+	 *
+	 * @return array|WP_Error list of newsletter categories
+	 */
+	public function get_newsletter_categories() {
+
+		if ( defined( 'IS_WPCOM' ) && IS_WPCOM && function_exists( 'require_lib' ) ) {
```
jeherve: Since you're doing a `( new Host() )->is_wpcom_simple()` check above, maybe we have an opportunity to create a new method in this class to check for wpcom?

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-newsletter-categories-list.php
```diff
@@ -0,0 +1,91 @@
...
+	public function register_routes() {
+		$options = array(
+			'show_in_index'       => true,
+			'methods'             => 'GET',
+			// if this is not a wpcom site, we need to proxy the request to wpcom
+			'callback'            => ( ( new Host() )->is_wpcom_simple() ) ? array(
+				$this,
+				'get_newsletter_categories',
+			) : array( $this, 'proxy_request_to_wpcom_as_user' ),
+			'permission_callback' => '__return_true',
```
jeherve: Should we restrict access to folks who can manage options on the site?

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-newsletter-categories-list.php
```diff
@@ -0,0 +1,91 @@
+<?php
+/**
+ * REST API endpoint for the Jetpack Blogroll block.
+ *
+ * @package automattic/jetpack
+ * @since 12.2
```
jeherve:
```suggestion
 * @since $$next-version$$
```

projects/plugins/jetpack/extensions/blocks/subscriptions/view.scss
```diff
@@ -96,3 +98,32 @@
...
+	margin-top: 8px;
+	margin-bottom: 24px;
+	display: flex;
+	justify-content: center;
+	flex-wrap: wrap;
+	gap: 8px;
+
+	.wp-block-jetpack-subscriptions__newsletter-category {
+		padding: 6px 10px;
+		font-family: "SF Pro Text", -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif;
```
jeherve: We may be able to use more variables here, like 

```suggestion
		font-family: $default-font;
```

projects/plugins/jetpack/extensions/blocks/subscriptions/view.scss
```diff
@@ -96,3 +98,32 @@
...
+		font-size: 14px;
+		font-weight: 500;
+		line-height: 20px;
+		color: $gray-600;
+		border: 1px solid #dcdcde;
```
jeherve: Could this be a `$gray-200`?

```suggestion
		border: 1px solid $gray-200;
```

projects/plugins/jetpack/extensions/blocks/subscriptions/view.scss
```diff
@@ -96,3 +98,32 @@
...
+	margin-bottom: 24px;
+	display: flex;
+	justify-content: center;
+	flex-wrap: wrap;
+	gap: 8px;
+
+	.wp-block-jetpack-subscriptions__newsletter-category {
+		padding: 6px 10px;
+		font-family: "SF Pro Text", -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif;
+		font-size: 14px;
```
jeherve: Could we use `$default-font-size` here?

projects/plugins/jetpack/extensions/blocks/subscriptions/edit.js
```diff
@@ -179,6 +183,10 @@ export function SubscriptionEdit( props ) {
 		padding: getPaddingStyleValue( padding ),
 	};
 
+	const cssVars = {
+		'--border-radius': borderRadius ? borderRadius + 'px' : DEFAULT_BORDER_RADIUS_VALUE + 'px',
```
jeherve: I wonder if we would need to be more specific with the variable name here, to avoid potential mixups later?

jeherve (changes requested):

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-newsletter-categories-list.php
TimBroddin: I guess I can just leave out this check, since this callback will only be invoked on WPCOM.

phcp (approved):
Good job @TimBroddin! Tested and it works as expected.

TimBroddin:
Hey @jeherve 

The changes you asked for were made, could you rereview? 🙏

jeherve (approved):
Looks good and tests well for me. 👍

PR #32671, opened 2023-08-24
Social: Use Auto Conversion logic on the UI

gmjuhasz:
This change updates the logic to disable certain connections if the selected media is invalid. Now if auto conversion is enabled, the connections wont get disabled if the conversion is possible

## Proposed changes:

* Added the new logic for handling the conversion
* Updated the notices accordingly

### Other information:

## Jetpack product discussion

https://github.com/orgs/Automattic/projects/733/views/2?pane=issue&itemId=36022365

No.
## Testing instructions:

* Check out this patch while also having the `jetpack-social-image-auto-convert` blog sticker on your site
  * Turn off the auto-conversion in the Social admin page. All media validation should work as before
    * Connections should get disabled if the media is invalid for them
  * Turn on the auto-conversion
    * Selecting a convertible media that is invalid for a connection, should not disable the connection and the notice should appear stating that it will be converted
    * Selecting a non-convertible media (videos, or images that are more than 55Mb-s) should result the same errors as with auto-conversion disabled.
*


projects/js-packages/publicize-components/src/hooks/use-media-restrictions/index.js
```diff
@@ -137,10 +162,14 @@ const useMediaRestrictions = (
...
 					return errs;
 			  }, {} );
+
 		if ( JSON.stringify( newErrors ) !== JSON.stringify( errors.current ) ) {
 			errors.current = newErrors;
 		}
-		return errors.current;
+		return {
+			validationErrors: errors.current,
+			isConvertible: isMediaConvertible( media.metaData ),
```
pablinos, line 171: As we discussed this morning, I think `isConvertible` should be a combination of `isMediaConvertible` and whether there is a FILE_TYPE_ERROR. That also means that whether an image is convertible could vary based on the connection type

projects/js-packages/publicize-components/src/components/media-section/index.js
```diff
@@ -2,7 +2,7 @@ import { ThemeProvider, getRedirectUrl } from '@automattic/jetpack-components';
 import { Disabled, ExternalLink, Notice, BaseControl } from '@wordpress/components';
 import { useCallback } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
-import { Fragment } from 'react';
+import React, { Fragment } from 'react';
```
pablinos: Fragment should be imported from `@wordpress/element` and to avoid importing React we can use [this pattern](https://github.com/WordPress/gutenberg/blob/5dbe0da79971af4aa38641630a32f7814d50c159/packages/element/src/serialize.js#L540) of importing it for the types/docs

projects/js-packages/publicize-components/src/hooks/use-dismiss-notice/index.js
```diff
@@ -7,12 +7,14 @@ import { useCallback, useMemo, useState } from 'react';
  * @property {string} instagram - The name of the Instagram notice.
  * @property {string} advancedUpgradeEditor - The name of the advanced upgrade nudge in the editor.
  * @property {string} advancedUpgradeAdmin - The name of the advanced upgrade nudge in the admin page.
+ * @property {string} autoConversion - The name of the auto conversion notice.
```
pablinos, line 10: I've noticed we're importing from 'react' on line 3 of this file too. We should update that to `@wordpress/element`

pablinos (commented):
The code looks reasonable. I'm have trouble getting a suitable invalid image, to get the notice about not being able to convert it. I used a long video and got the message that it was invalid for Instagram, which I think tests the same logic.

projects/js-packages/publicize-components/src/hooks/use-media-restrictions/index.js
gmjuhasz: Updated it!

projects/js-packages/publicize-components/src/components/media-section/index.js
gmjuhasz: Oh yeah I always forget this, thanks!

projects/js-packages/publicize-components/src/hooks/use-dismiss-notice/index.js
gmjuhasz: Updated!

spsiddarthan (commented):
This is working well so far, question is when do we want to merge it.

projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -149,14 +156,44 @@ export default function PublicizeForm( {
 		</>
 	);
 
+	const renderValidationNotice = () => {
```
pablinos, line 159: Not for now, but I think there's enough here that we should think about making a notices component to handle these, or maybe creating separate components for them.

projects/js-packages/publicize-components/src/hooks/use-media-restrictions/index.js
```diff
@@ -137,10 +162,15 @@ const useMediaRestrictions = (
...
 			  }, {} );
+
 		if ( JSON.stringify( newErrors ) !== JSON.stringify( errors.current ) ) {
 			errors.current = newErrors;
 		}
-		return errors.current;
+		return {
+			validationErrors: errors.current,
+			isConvertible:
+				isMediaConvertible( media.metaData ) && ! ( FILE_TYPE_ERROR in errors.current ), // Currently we cannot convert file-types.
```
pablinos: It looks like we could remove the file type error here, because we'll launch with file type conversions working.

I'm not sure this is working as expected either as I uploaded a WebP and got the conversion notice for Instagram.

pablinos (commented):
[The logic is getting a bit wild](https://github.com/Automattic/jetpack/pull/32671/files#diff-60b54a297ef92f40c4b903f181a44bc7989e0ebcab1039e841232ec4a5c581a6R255), but when we can convert the image, we're still applying the `is-disabled` class on the connection which fades it out.

CleanShot 2023-09-01 at 11 48 06

pablinos (approved):
This looks good to me. Let's test it with the backend changes.

PR #32676, opened 2023-08-24
Add/user content link track and redirection to original URL

### LINKED ISSUE #80864: Post Link Tracking: track click information and redirect the user to the final URL
When the user clicks one of the tracking links in the email, some logic will be triggered that does two things:

a) We store the link tracking info in the `email_link_stats` table.
b) We redirect the user to the original URL.

This code will live in Jetpack (an iframe) and WPCOM.

JuanLucha:
Closes https://github.com/Automattic/wp-calypso/issues/80864

## Proposed changes:
When the user clicks one of the tracking links in their emails (like "new post email", for example), some logic will be triggered that does two things:

a) We store the link tracking info in the email_link_stats table.
b) We redirect the user to the original URL. The "original URL" was the URL that the author of the content assigned to the link. That URL was processed before sending the email, converting it to a "tracking URL" instead.

### Other information:

## Jetpack product discussion
pdDOJh-2i1-p2

It will allow us to track links in the body of the posts and comments, that we couldn't track until now. More info on this here: pdDOJh-1X2-p2

## Testing instructions:
- TBD


projects/plugins/jetpack/jetpack-user-content-tracking-redirects.php
```diff
@@ -0,0 +1,77 @@
...
+		// Grab the original URL from the database. The table has this structure:
+		global $wpdb;
+		$url_info = $wpdb->get_row(
+			$wpdb->prepare( 'SELECT url, email_link_mapping_id FROM email_link_mapping WHERE uuid = %s', $uuid )
+		);
+
+		// Store in the click database the information about the click. This is the structure:
+		$wpdb->query(
+			$wpdb->prepare(
+				'INSERT INTO email_link_stats (subscriber_id, user_id, post_id, blog_id, email_link_mapping_id) VALUES (%d, %d, %d, %d, %d) ON DUPLICATE KEY UPDATE total_clicks = total_clicks + 1',
```
yansern, line 42: Would these database calls work on Atomic or self-hosted Jetpack sites?

jeherve (commented):
If I understand correctly, this will be used to save data in a db on WordPress.com when someone accesses a page on the frontend of a WordPress.com Simple, WoA, or self-hosted site using the Jetpack plugin, when that page includes specific GET parameters?

It looks like the information saved is limited to subscriber_id, user_id, post_id, blog_id, email_link_mapping_id, all info we get from the GET parameters?

Since this information will only be available on sites that are connected to WordPress.com, I think we should only run such code when a site is connected to WordPress.com.

In fact, instead of a new file at the root of the Jetpack plugin, should this be part of the Connection package and the existing `Tracking` class, with the same safeguards (having agreed to ToS, having accepted tracking, ...)?
This may also allow us to use the existing mechanisms in place to send the data over to WordPress.com when the code does not run in the WordPress.com environment.

JuanLucha:
> If I understand correctly, this will be used to save data in a db on WordPress.com when someone accesses a page on the frontend of a WordPress.com Simple, WoA, or self-hosted site using the Jetpack plugin, when that page includes specific GET parameters?
> 
> It looks like the information saved is limited to subscriber_id, user_id, post_id, blog_id, email_link_mapping_id, all info we get from the GET parameters?
> 
> Since this information will only be available on sites that are connected to WordPress.com, I think we should only run such code when a site is connected to WordPress.com.
> 
> In fact, instead of a new file at the root of the Jetpack plugin, should this be part of the Connection package and the existing `Tracking` class, with the same safeguards (having agreed to ToS, having accepted tracking, ...)? This may also allow us to use the existing mechanisms in place to send the data over to WordPress.com when the code does not run in the WordPress.com environment.

Hi, Jeremy!
Good point! I'll check the Connection package and how it handles the tracking.

PR #32680, opened 2023-08-24
VideoPress: Check if module is active before initializing

### LINKED ISSUE #32235: VideoPress Block: CSS file is loaded on all pages, even when the block isn't used
### Steps to reproduce

CGastrell:
VideoPress initializer takes place even without the module being active, leading to load unneeded resources.

Fixes #32235

## Proposed changes:
This PR adds an extra check to see if the module is active before continuing with the registration. When the theme is NOT a block theme, the enqueuing process is deferred to `wp_enqueue_scripts` hook, where the content of the post can be parsed and checked for a VP block and, only then, enqueue the needed assets.

### Other information:

No

## Testing instructions:
Create a site and connect it to Jetpack. Don't enable VideoPress yet. Visit the site's homepage (or any page), check on the sources to verify `jetpack_vendor/automattic/jetpack-videopress/build/block-editor/blocks/video/view.css` is NOT being loaded.

Change theme to a classic theme, like 2020. Visit a page/post on frontend where NO video blocks are present. See that the mentioned file above is not being loaded. Then visit a page WITH a video block to see the file is being loaded properly.

### DIFF
--- a/projects/packages/videopress/src/class-initializer.php
+++ b/projects/packages/videopress/src/class-initializer.php
@@ -356,7 +356,7 @@ class Initializer {
 			add_action(
 				'wp_enqueue_scripts',
 				function () use ( $videopress_video_metadata_file ) {
-					if ( ! has_block( 'videopress/video', get_the_content() ) && ! has_block( 'jetpack-videopress/video', get_the_content() ) ) {
+					if ( ! has_block( 'videopress/video', get_the_content() ) ) {
 						return;
 					}
 
--- a/projects/packages/videopress/src/class-initializer.php
+++ b/projects/packages/videopress/src/class-initializer.php
@@ -356,16 +356,16 @@ class Initializer {
 			add_action(
 				'wp_enqueue_scripts',
 				function () use ( $videopress_video_metadata_file ) {
-					if ( ! has_block( 'videopress/video', get_the_content() ) ) {
+					$post_content = get_the_content();
+					// TODO: we need to also parse the site content for FSE cases (block/shortcode could be on footer, header, etc)
+					if ( ! has_block( 'videopress/video', $post_content ) && ! has_shortcode( $post_content, 'videopress' ) ) {
 						return;
 					}
-
 					self::enqueue_block_assets( $videopress_video_metadata_file );
 				}
 			);
 			return;
 		}
-
 		self::enqueue_block_assets( $videopress_video_metadata_file );
 	}
 
--- a/projects/packages/videopress/src/class-initializer.php
+++ b/projects/packages/videopress/src/class-initializer.php
@@ -353,7 +353,7 @@ class Initializer {
 
 		// for non block themes, we defer the enqueuing to the frontend, so we're able to tell if we need the assets
 		// TODO: this is draft code, shall we go this path I'll clean up and split code to its own method
-		if ( ! $is_block_theme ) {
+		if ( ! $is_block_theme && ! is_admin() ) {
 			add_action(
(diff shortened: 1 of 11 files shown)


retrofox (commented):
If my memory serves me right, we agreed to show the VideoPress video block when the module is inactive or, even more, when the site is disconnected.

When this happens, we provide a prompt that directs the user to the appropriate page to resolve the problem with the site or module.
This behavior aims to make the block visibility as much as possible for the users.

If we want to keep it (I think we do), we'll have to tackle the loading CSS issue from another perspective.

Does it make sense for you?

CGastrell:
> If my memory serves me right, we agreed to show the VideoPress video block when the module is inactive or, even more, when the site is disconnected.
> 
> When this happens, we provide a prompt that directs the user to the appropriate page to resolve the problem with the site or module. This behavior aims to make the block visibility as much as possible for the users.
> 
> If we want to keep it (I think we do), we'll have to tackle the loading CSS issue from another perspective.
> 
> Does it make sense for you?

Yes, seems like the way to go if we want better discoverability. I'll ask on the issue as this seems theme specific.

projects/packages/videopress/src/class-initializer.php
```diff
@@ -348,6 +348,33 @@ public static function register_videopress_video_block() {
...
 
+		// check current theme
+		$is_block_theme = wp_get_theme()->is_block_theme();
+
+		// for non block themes frontend, we defer the enqueuing to the frontend, so we're able to tell if we need the assets
+		if ( ! $is_block_theme && ! is_admin() ) {
+			add_action(
+				'wp_enqueue_scripts',
+				function () use ( $videopress_video_metadata_file ) {
+					if ( ! has_block( 'videopress/video', get_the_content() ) && ! has_block( 'jetpack-videopress/video', get_the_content() ) ) {
```
retrofox: If I'm correct, this condition only works in the post-content context. What happens if there is a VideoBlock video block instance in the footer added by a widget?

The second condition caught my attention because I've never seen the `jetpack-videopress` namespace before. Where is it being used?

CGastrell: Good questions!

First one, most likely you're correct. Any ideas how to run the same parsing/check throughout all template parts besides the post content?

Second one, my bad, I did a search and misunderstood the output, I thought there was a jetpack-videopress/video block, I'll remove it

retrofox: It's like [has_block()](https://developer.wordpress.org/reference/functions/has_block/) accepts a string as a secondary parameter. Can we get the site content instead of the post one? That seems to be kinda tricky. I Never did it, honestly.

CGastrell: > It's like [has_block()](https://developer.wordpress.org/reference/functions/has_block/) accepts a string as a secondary parameter. Can we get the site content instead of the post one? That seems to be kinda tricky. I Never did it, honestly.

I just remembered this issue only happens with  _classic_ themes, block themes render just fine and don't load the CSS unnecessarily. Moreover, block themes can use a VP block on templates and that also works properly.

Just for research purposes, I've been asking about some filter or global function to return the site (although we're talking about the page render, right?), just like `get_the_content()` does for posts, and it seems there is no such equivalent. I've been advised to maybe move to a more atomic filter (like [render_block](https://github.com/WordPress/gutenberg/issues/38120)) and load any assets there. I don't think we need this right now, the block is working fine in that regard.

dhasilva (approved):
LGTM!

| Before | After
|-|-
|   |

PR #32683, opened 2023-08-25
Earn/tier selector

n3f:
Fixes https://github.com/Automattic/gold/issues/119. (Adds tier selector under the paid subscribers access).

## Proposed changes:

* Create a TierSelector component and add to post paid newsletter access selector.

### Other information:

## Testing instructions:

* Apply the diff to WPCOM (e.g. `bin/jetpack-downloader test jetpack earn/tier-selector`) (this is required by an API change used on public-api.wordpress.com, alternately you could use with jurassic ninja, but make sure you push this change to sandbox and you global-http enable to properly sandbox your site).
* Get a test site working on your sandbox.
* Connect stripe and create a couple plans (they need to be monthly paid newsletter plans)(This PR should work for cases with 2 or more "tiers" defined.
* Enable paid newsletters
* Create a post and set the access to "Paid subscribers only". You should then see your tiers defined and be able to select them.
* Set the tier and save the post.
* Ensure reloading the post selects the same tier.

### DIFF
--- a/projects/plugins/jetpack/extensions/shared/memberships/settings.js
+++ b/projects/plugins/jetpack/extensions/shared/memberships/settings.js
@@ -98,13 +98,15 @@ function TierSelector( { onChange } ) {
 
 	// Find the current tier meta
 	const postType = useSelect( select => select( editorStore ).getCurrentPostType(), [] );
+	// Destructure the tierId from the meta (set tierId using the META_NAME_FOR_POST_TIER_ID_SETTINGS constant)
 	const [ { [ META_NAME_FOR_POST_TIER_ID_SETTINGS ]: tierId } ] = useEntityProp(
 		'postType',
 		postType,
 		'meta'
 	);
 
-	// Tiers don't apply if less than 2 products
+	// Tiers don't apply if less than 2 products (this is called here because
+	// the hooks have to run before any early returns)
 	if ( products.length < 2 ) {
 		return;
 	}
@@ -198,8 +200,7 @@ export function NewsletterAccessRadioButtons( {
 						{ key === accessOptions.paid_subscribers.key &&
 							key === accessLevel &&
 							isStripeConnected &&
-							hasNewsletterPlans &&
-							! postHasPaywallBlock && <TierSelector onChange={ onChange }></TierSelector> }
+							hasNewsletterPlans && <TierSelector onChange={ onChange }></TierSelector> }
 					</div>
 				);
 			} ) }
--- a/projects/plugins/jetpack/sal/class.json-api-metadata.php
+++ b/projects/plugins/jetpack/sal/class.json-api-metadata.php
@@ -4,6 +4,9 @@
  *
  * @package automattic/jetpack
  */
+
+use const Automattic\Jetpack\Extensions\Subscriptions\META_NAME_FOR_POST_TIER_ID_SETTINGS;
+
 /**
  * Base class for WPCOM_JSON_API_Metadata
  */
@@ -57,7 +60,7 @@ class WPCOM_JSON_API_Metadata {
 		// display the correct newsletter access in Calypso.
 		$whitelist = array(
 			'_jetpack_newsletter_access',
-			'_jetpack_newsletter_tier_id',
+			META_NAME_FOR_POST_TIER_ID_SETTINGS,
 		);
 
 		if ( in_array( $key, $whitelist, true ) ) {


projects/plugins/jetpack/extensions/shared/memberships/settings.js
```diff
@@ -85,6 +85,46 @@ function NewsletterAccessSetupNudge( { stripeConnectUrl, isStripeConnected, hasN
...
 }
 
+function TierSelector() {
+	// TODO: filter on currency?
+	const products = useSelect( select => select( membershipProductsStore ).getProducts() )
+		.filter(
+			product =>
+				product.subscribe_as_site_subscriber &&
+				product.interval === '1 month' &&
+				product.buyer_can_change_amount === false
```
millerf:
```
product.buyer_can_change_amount === false
```

I understand the idea about that, but let's not worry about that for now..
Meaning if subscribers can change the price of their subscription, it is fine. What counts is the price set by the content creator.

projects/plugins/jetpack/extensions/shared/memberships/settings.js
```diff
@@ -85,6 +85,46 @@ function NewsletterAccessSetupNudge( { stripeConnectUrl, isStripeConnected, hasN
...
+		)
+		.sort( product => Number( product.price ) )
```
millerf: We will need to think about how to handle multiple currencies here... But let's not worry about it for now.

projects/plugins/jetpack/extensions/shared/memberships/settings.js
```diff
@@ -138,6 +178,18 @@ export function NewsletterAccessRadioButtons( {
 							{ accessLabel }
 							{ reach }
 						</label>
+						{ /*
+						 * This adds a tier selector when:
+						 * - the paid_subscribers option is selected
+						 * - stripe is selected
```
millerf:
```suggestion
						 * - stripe is connected
```
Small typo

projects/plugins/jetpack/extensions/shared/memberships/settings.js
```diff
@@ -138,6 +178,18 @@ export function NewsletterAccessRadioButtons( {
...
+						 * - there are newsletter plans
```
millerf:
```suggestion
						 * - there are newsletter plans
```
More than 1 newsletter plan. Otherwise it works as it is working right now.

projects/plugins/jetpack/extensions/shared/memberships/settings.js
```diff
@@ -138,6 +178,18 @@ export function NewsletterAccessRadioButtons( {
...
+						 * - this isn't a paywall block
```
millerf:
```suggestion
						 * - this isn't a paywall block
```
It will work if it has paywall as well.
It will be a strong incentive for user to upgrade to the next tier for example. (CC @simison for validation)

projects/plugins/jetpack/extensions/shared/memberships/settings.js
n3f: thanks!

projects/plugins/jetpack/extensions/shared/memberships/settings.js
n3f: OK, but this needs to be looked at more in depth because I believe if a user thinks they can set a paywall block they might think they could have different sections for the different subscribers but that won't work with our implementation because the tier is a post level setting. (i.e. when it changes, the entire post tier is set and not just a small section.)

I'll keep the check for now, so respond to this if we should remove it going forward.

projects/plugins/jetpack/extensions/shared/memberships/settings.js
```diff
@@ -138,6 +188,19 @@ export function NewsletterAccessRadioButtons( {
...
+							// - stripe is connected
+							// - there are newsletter plans (the component will
+							//   check for more than 1 plan)
+							// - this isn't a paywall block
+						 }
+						{ key === accessOptions.paid_subscribers.key &&
+							key === accessLevel &&
+							isStripeConnected &&
+							hasNewsletterPlans &&
+							! postHasPaywallBlock && <TierSelector onChange={ onChange }></TierSelector> }
```
n3f:
```suggestion
							&& <TierSelector onChange={ onChange }></TierSelector> }
```

If we want to use the same selector in the paywall block too, let's remove this line.

projects/plugins/jetpack/sal/class.json-api-metadata.php
```diff
@@ -53,8 +53,14 @@ public static function is_internal_only( $key ) {
 			return false;
 		}
 
-		// We want to always return the `_jetpack_newsletter_access` key to display the correct newsletter access in Calypso.
-		if ( $key === '_jetpack_newsletter_access' ) {
+		// We want to always return the `_jetpack_newsletter_access` key to
+		// display the correct newsletter access in Calypso.
+		$whitelist = array(
+			'_jetpack_newsletter_access',
```
millerf, line 62: Do you think we can use META_NAME_FOR_POST_TIER_ID_SETTINGS and META_NAME_FOR_POST_LEVEL_ACCESS_SETTINGS? Just so we don't miss them in the future? I am not sure about the env context...

projects/plugins/jetpack/extensions/shared/memberships/settings.js
```diff
@@ -85,6 +90,50 @@ function NewsletterAccessSetupNudge( { stripeConnectUrl, isStripeConnected, hasN
...
 
+function TierSelector( { onChange } ) {
+	// TODO: figure out how to handle different currencies
+	const products = useSelect( select => select( membershipProductsStore ).getProducts() )
+		.filter( product => product.subscribe_as_site_subscriber && product.interval === '1 month' )
+		.sort( ( p1, p2 ) => Number( p2.price ) - Number( p1.price ) );
+
+	// Find the current tier meta
+	const postType = useSelect( select => select( editorStore ).getCurrentPostType(), [] );
+	const [ { [ META_NAME_FOR_POST_TIER_ID_SETTINGS ]: tierId } ] = useEntityProp(
```
millerf, line 102:
```
[ { [ META_NAME_FOR_POST_TIER_ID_SETTINGS ]: tierId } ]
```
I have ever seen such a notation 😅

projects/plugins/jetpack/extensions/shared/memberships/settings.js
lezama: >  different sections for the different subscribers

I think "Premium content" block should be used for this use case not the Paywall

millerf: > I'll keep the check for now, so respond to this if we should remove it going forward.

Yes, we need to remove  that check. 
The paywall will:
-  only appear to people that don't have access (not in the selected tier)
- the newsletter will be sent to everyone

Basically it will work exactly as it is now... But the access will be given to less people...

projects/plugins/jetpack/extensions/shared/memberships/settings.js
n3f: Added some clarifying comments...

projects/plugins/jetpack/sal/class.json-api-metadata.php
n3f: I'm not sure, I thought there must have been a reason, but I don't see a separate commit in the repo that tried it. I'm a little leery of making the change to both constants.  Should we just try with `META_NAME_FOR_POST_TIER_ID_SETTINGS`? (I'm checking in that change)

PR #32685, opened 2023-08-25
Boost: Improve ISA error UI

dilirity:
Fixes https://github.com/Automattic/boost-cloud/issues/352

## Proposed changes:

* Fix boost core util package not checking indexes returned from Boost Cloud when there's an error;
* Update error UI spacing;
* Add suggestion for the error UI.

### Before

CleanShot 2023-08-25 at 12 33 53

### After

CleanShot 2023-08-25 at 12 33 25

### Other information:

## Jetpack product discussion

n/a

no

### The easiest way to test this is with a modified Boost Developer plugin

- Setup this version of Boost (premium);
- Request a report to get the results UI to show up;
- In the Boost Developer source code, replace https://github.com/Automattic/boost-developer/blob/trunk/Shield_Client.php#L15-L49 with
   <details>
   <summary>the following code:</summary>

   ```php
	public function post( $endpoint, $payload ) {
		$this->set_default_headers( $endpoint, $payload );

		$args = array(
			'body'      => $payload,
			'headers'   => $this->headers,
			'timeout'   => 30,
			'sslverify' => false,
			'method'    => 'POST',
		);

		$response = wp_remote_post( $endpoint, $args );

		$response = array(
			'body'     => '{"statusCode":429,"error":"Too Many Requests"}',
			'response' => array(
				'code'    => 429,
				'message' => 'Too Many Requests',
			),
		);
		$code     = wp_remote_retrieve_response_code( $response );

		if ( is_wp_error( $response ) ) {
			return $response;
		}

		$json = json_decode( wp_remote_retrieve_body( $response ) );

		if ( 200 !== $code ) {
			$default_message = sprintf(
			/* translators: %d is a numeric HTTP error code */
				__( 'HTTP %d while communicating with WordPress.com', 'jetpack-boost' ),
				$code
			);

			$err_code = empty( $json->statusCode ) ? 'http_error' : $json->statusCode;
			$message  = empty( $json->error ) ? $default_message : $json->error;

			return new \WP_Error( $err_code, $message );
		}

		return $json;
	}
   ```
   </details>
- Request a report again and you should see the updated version of the error UI.

To see the "before" version, simply use the latest (or trunk) version of boost, but leave the replaced lines in Boost Developer (as that simulates the 429 error).


projects/plugins/boost/app/assets/src/js/utils/get-isa-error-suggestion.ts
```diff
@@ -0,0 +1,11 @@
+import { __ } from '@wordpress/i18n';
+
+export default function getIsaErrorSuggestion( code ) {
+	let suggestion = '';
+
+	if ( 429 === code ) {
+		suggestion = __( 'Minimum time between requests is 1 hour.', 'jetpack-boost' );
```
haqadn: We may change the threshold in boost-cloud from time to time. So, it would be nice if the message was coming from the API instead.

haqadn (approved):

PR #28913, opened 2023-02-13
E2E tests: better scoping

adimoldovan:
## Proposed changes:

Update e2e tests scoping to use jetpack cli dependencies command and run tests whenever any dependency of the specified target plugin changes.

- moved the target definition from `package.json` to `e2e-matrix.js` file for more control. You can now define a group of tests to run multiple times for different conditions.
- for monorepo changes there is one small group to run, just to validate the infrastructure didn't break
- for changes to files that have no tests assigned, test report and slack notifications jobs no longer run

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion
pd5faL-pt-p2#comment-308

## Does this pull request change what data or activity we track or use?
no

## Testing instructions:
* [see test run](https://github.com/Automattic/jetpack/actions/runs/4162842667/jobs/7202505998) for changes to monorepo
* [see test run](https://github.com/Automattic/jetpack/actions/runs/4162815533) for changes to files without tests assigned
* [see test run](https://github.com/Automattic/jetpack/actions/runs/4162613454/jobs/7202023879) for changes to only boost plugin

PR #32686, opened 2023-08-25
Jetpack Boost: Split and Parallelize Jetpack Boost Tests

pyronaur:
## Proposed changes:
- Split Boost tests into multiple directories (base, critical-css, modules)
- Run Jetpack Boost tests in parallel
- Updated file paths in test files to reflect new directory structure

### Other information:
N/A

## Jetpack product discussion
N/A

N/A

## Testing instructions:
- Run all Jetpack Boost tests


pyronaur:
@jeherve @anomiex 👋  I'll appreciate a thorough shaking on GitHub workflow files, my knowledge in GitHub workflows is limited.

I'll post a P2 post tomorrow with more details, but to summarize what I'm trying to achieve here:

We're seeing local tunnel timeouts quite often when running tests and the whole E2E test suite for boost takes up to 20 minutes sometimes, so if an E2E test fails, we have two choices - either yolo or run the test again, and possibly again and again.

So with this - I'm doing a couple things:

* Splitting up Jetpack Boost tests into multiple directories so that they can run in parallel
* Caching the build directory based on the last commit and project "group" so that multiple tests can share the same build (build once, then run multiple tests in parallel)
* If a test fails - we can now re-run only (hopefully) 1 out of 3 tests and it won't need to rebuild, so re-runs due to tunnel failures now take ~3-4 minutes instead of 10-20.


I'll appreciate any suggestions you might have. I think a lot more could be done here, perhaps even caching the whole testing environment in a docker image and only a couple of build files to optimize the build times even more, but for now - at least this fixes the immediate pain point for us.

.github/workflows/e2e-tests.yml
```diff
@@ -44,171 +44,220 @@ jobs:
...
     needs: create-test-matrix
-    timeout-minutes: 60
+    timeout-minutes: 30
+    concurrency:
+      group: e2e-build-${{ github.event_name }}-${{ github.sha }}-${{ github.event.action }}-${{ matrix.buildGroup }}
+      cancel-in-progress: true
     strategy:
       fail-fast: false
       matrix:
         include: ${{ fromJson( needs.create-test-matrix.outputs.matrix ) }}
```
anomiex: This doesn't work very well. You wind up spawning three different "E2E: Build jetpack-boost" jobs, then hoping that the `concurrency` above will kill two of them without wasting too many resources. Better to only spawn the jobs you need, which probably will mean you'll need multiple outputs from `create-test-matrix`.

.github/workflows/e2e-tests.yml
```diff
@@ -44,171 +44,220 @@ jobs:
...
-          echo "::group::Prepare build artefacts"
-          tar --xz -xf "$BUILD_DIR/jetpack-build/build.tar.xz" -C "$BUILD_DIR"
-          echo "::endgroup::"
-        fi
+      - name: Restore ${{ matrix.buildGroup }} build cache
+        id: jetpack-build-cache
+        uses: actions/cache/restore@v3
+        with:
+          path: .
+          key: ${{ matrix.buildGroup }}-${{ github.sha }}
```
anomiex, line 83: A cache for every sha may run into https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy, causing more useful caches to be ejected in favor of these.

OTOH, we may have enough cache churn that it doesn't really matter. I'm not sure. Looks like GitHub's own codeql action does the same thing.

.github/files/e2e-tests/e2e-matrix.js
```diff
@@ -8,62 +8,87 @@ const projects = [
 		testArgs: [ 'specs/connection', '--retries=1' ],
 		targets: [ 'plugins/jetpack' ],
 		suite: '',
+		buildGroup: 'jetpack-connection',
```
anomiex: Do we really need separate build groups for jetpack-connection, jetpack-pre-connection, jetpack-post-connection, jetpack-sync, and jetpack-blocks? Or can they all share one build like you're doing for jetpack-boost?

.github/workflows/e2e-tests.yml
```diff
@@ -44,171 +44,220 @@ jobs:
...
-        if [[ "${SUITE}" != atomic* && "${SUITE}" != vip ]]; then
-          echo "::group::Start docker environment"
-          pnpm run env:up
-          echo "::endgroup::"
+      - name: Save ${{ matrix.buildGroup }} build cache
+        if: steps.jetpack-build-cache.outputs.cache-hit != 'true'
+        id: jetpack-build-cache-save
+        uses: actions/cache/save@v3
+        with:
+          path: .
```
anomiex: Can we avoid caching the entire monorepo along with the built files? `actions/checkout` on line 111 below will already be checking it out again anyway.

Even better if we can also avoid caching all of `vendor/` and `node_modules/` everywhere too without too much loss in speed. The package managers' caches should already be cached in other keys, but they'd still need to be re-installed which would take some amount of time.

.github/workflows/e2e-tests.yml
```diff
@@ -44,171 +44,220 @@ jobs:
...
-      run: |
-        if [ "$GITHUB_EVENT_NAME" == workflow_run ]; then
-          echo "::group::Prepare build artefacts"
-          tar --xz -xf "$BUILD_DIR/jetpack-build/build.tar.xz" -C "$BUILD_DIR"
-          echo "::endgroup::"
-        fi
+      - name: Restore ${{ matrix.buildGroup }} build cache
+        id: jetpack-build-cache
+        uses: actions/cache/restore@v3
+        with:
```
anomiex, line 78: Use the `lookup-only` option mentioned at https://github.com/actions/cache/blob/main/restore/README.md?

.github/workflows/e2e-tests.yml
```diff
@@ -44,171 +44,220 @@ jobs:
...
+        id: jetpack-build-cache-save
+        uses: actions/cache/save@v3
+        with:
+          path: .
+          key: ${{ steps.jetpack-build-cache.outputs.cache-primary-key }}
+
+  e2e-tests:
+    name: "${{ matrix.project }} e2e tests"
+    runs-on: ubuntu-latest
+    if: ${{ always() }} &&  github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name
```
anomiex: Why use `always()` here? You probably don't want it to try to run if the build failed.

.github/workflows/e2e-tests.yml
```diff
@@ -44,171 +44,220 @@ jobs:
           MATRIX="$(node .github/files/e2e-tests/e2e-matrix.js)"
           echo "$MATRIX"
           echo "matrix=$MATRIX" >> "$GITHUB_OUTPUT"
-
-  e2e-tests:
-    name: "${{ matrix.project }} e2e tests"
+  build-projects:
```
anomiex, line 64: Just a little formatting
```suggestion

  build-projects:
```

.github/workflows/e2e-tests.yml
```diff
@@ -44,171 +44,220 @@ jobs:
...
+        id: jetpack-build-cache
+        uses: actions/cache/restore@v3
+        with:
+          path: .
+          key: ${{ matrix.buildGroup }}-${{ github.sha }}
+      - name: Checkout code
+        if: steps.jetpack-build-cache.outputs.cache-hit != 'true'
+        uses: actions/checkout@v3
 
-        if [ "$GITHUB_EVENT_NAME" == workflow_run ] || [ "$GITHUB_EVENT_NAME" == repository_dispatch ]; then
```
anomiex, line 106: You seem to have lost this condition on running the build. If either of these are true, the build is supposed to be fetched from elsewhere instead of being built here.

.github/workflows/e2e-tests.yml
```diff
@@ -44,171 +44,220 @@ jobs:
...
+
+      - name: Build projects
+        if: steps.jetpack-build-cache.outputs.cache-hit != 'true'
+        env:
+          COMPOSER_ROOT_VERSION: "dev-trunk"
+          BUILD_DIR: ./build-output
+          PROJECT_PATH: ${{ matrix.path }}
+          SUITE: ${{ matrix.suite }}
+        run: |
           if [[ "${SUITE}" != atomic* ]]; then
```
anomiex: There doesn't seem to be a whole lot of point in doing all the above steps if the build is going to be skipped because `"${SUITE}" != atomic*`

.github/workflows/e2e-tests.yml
```diff
@@ -44,171 +44,220 @@ jobs:
...
-    timeout-minutes: 60
+    timeout-minutes: 30
+    concurrency:
+      group: e2e-build-${{ github.event_name }}-${{ github.sha }}-${{ github.event.action }}-${{ matrix.buildGroup }}
+      cancel-in-progress: true
     strategy:
       fail-fast: false
       matrix:
         include: ${{ fromJson( needs.create-test-matrix.outputs.matrix ) }}
-    if: github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name
```
anomiex, line 57: You seem to have lost this condition on running the build. If this check doesn't pass, there's no point in building since the tests won't run anyway.

.github/workflows/e2e-tests.yml
```diff
@@ -44,171 +44,220 @@ jobs:
...
+          if [ "$GITHUB_EVENT_NAME" == workflow_run ] || [ "$GITHUB_EVENT_NAME" == repository_dispatch ]; then
+            echo "::group::Prepare build artefacts"
+            tar --xz -xf "$BUILD_DIR/jetpack-build/build.tar.xz" -C "$BUILD_DIR"
+            echo "::endgroup::"
+
+            echo "::group::Update volume mapping"
+            .github/files/e2e-tests/map-plugins-for-e2e-env.sh
+            echo "::endgroup::"
+          else
+            echo "Skipping build preparation; using pre-built artifacts."
```
anomiex: Re some of the above comments, if we _don't_ hit this line then `build-projects` having built was a waste of time.

It might even be worth checking for that and failing to force fixing of the wasted time.

jeherve:
@adimoldovan If you get the chance and have some bandwidth, do you think you could take a quick look at this?

Thank you!

.github/workflows/e2e-tests.yml
```diff
@@ -44,171 +44,220 @@ jobs:
...
+    needs: [ create-test-matrix, build-projects ]
+    timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        include: ${{ fromJson( needs.create-test-matrix.outputs.matrix ) }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Restore $${{ matrix.buildGroup }} build cache
+        id: jetpack-build-cache
```
anomiex, line 152: Should probably be an `if` here to only bother with this if we're in one of the cases where the build was needed.

.github/workflows/e2e-tests.yml
```diff
@@ -44,171 +44,220 @@ jobs:
...
+        with:
+          name: test-output-${{ matrix.project }}
+          path: ${{ matrix.path }}/output
 
   test-report:
     name: "Test report"
     runs-on: ubuntu-latest
-    if: ${{ success() || failure() }}
-    needs: [e2e-tests]
+    if: ${{ always() }}
```
anomiex: Using `success() || failure()` rather than `always()` here was intentional, see https://github.com/Automattic/jetpack/pull/28913#discussion_r1105079565

.github/workflows/e2e-tests.yml
```diff
@@ -240,12 +289,11 @@ jobs:
           repository: automattic/jetpack-e2e-reports
           event-type: ${{ steps.report-context.outputs.event_type }}
           client-payload: ${{ steps.report-context.outputs.client_payload }}
-
   slack-notification:
     name: "Slack notification"
     runs-on: ubuntu-latest
-    if: ${{ success() || failure() }}
-    needs: [e2e-tests]
+    if: ${{ always() }}
```
anomiex: Here too.

.github/workflows/e2e-tests.yml
pyronaur: I agree - this is a crude way of doing it. But it does at least put out a fire. From the tests I've done so far - all the concurrent jobs are immediately shut down, so not much resource waste here - just annoying two canceled jobs.

I'll appreciate help here if you've got time, alternatively, if we can tolerate this approach for now, I can find time to clean this up later.

.github/workflows/e2e-tests.yml
pyronaur: More useful caches are probably stored as artifacts, I can't speak for churn. Any suggestions?

.github/files/e2e-tests/e2e-matrix.js
pyronaur: I don't know - that's why I wanted you and @jeherve to have a look. I know that boost isn't going to have a cache conflict with itself. How might tests between jetpack-connections react, I don't know, so I kept it safe for now. Feel free to make edits ( or I can make them tomorrow ).

.github/workflows/e2e-tests.yml
pyronaur: I'm not sure about that. Build structure for all the various packages and plugins may differ, so isolating a specific pattern might not be that easy. 

Are you concerned with the cache size here?

.github/workflows/e2e-tests.yml
pyronaur: 👍

.github/workflows/e2e-tests.yml
pyronaur: I probably don't, but "skips" are considered failures - because of the concurrency statement above - the e2e-tests never start after the build is done without `always()`

.github/workflows/e2e-tests.yml
pyronaur: 👍  Will fix.

.github/workflows/e2e-tests.yml
pyronaur: 🤔  Good point

.github/workflows/e2e-tests.yml
pyronaur: 👍  I'll take a closer look

.github/workflows/e2e-tests.yml
pyronaur: They seemed intentional (although might be worth linking to that comment via a comment), but I assumed that this might run into the same problem as I mentioned earlier - the job not even trying to start if the previous job in the chain was cancelled.

.github/workflows/e2e-tests.yml
anomiex: Should be about as simple as adding this to the "Create test plan" step above
```
BUILD_MATRIX=$(jq -c '[ .[] | { path: .path, buildGroup: .buildGroup, suite: .suite } ] | unique' <<<"$MATRIX")
echo "$BUILD_MATRIX"
echo "build-matrix=$BUILD_MATRIX" >> "$GITHUB_OUTPUT"
```
then adding an appropriate output to the job near line 20 and using that output here.

.github/workflows/e2e-tests.yml
anomiex: Let's go with this for now.

.github/workflows/e2e-tests.yml
anomiex: May as well not bloat it if we don't have to, we only get 10GB total for all caches before GH starts evicting old ones.

The `path` here can be a list of newline-separated paths, it doesn't have to be just one. So you might `find . -type f > /tmp/before.txt` before the build, `find . -type f > /tmp/after.txt` after, then do like
```
PATHS=$(diff /tmp/before.txt /tmp/after.txt | sed -n 's/^> //p')
source .github/files/gh-funcs.sh
gh_set_output paths "$PATHS"
```
to dump the list of added files into an output.

Then here you'd do `path: ${{ steps.some-id.outputs.paths }}` to read that list in.

.github/workflows/e2e-tests.yml
anomiex: As written you won't ever have any skips I think. Although you probably will once you fix some of the other comments. OTOH, you will have cancellations, which I recommended you fix above (line 58).

If https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions is accurate, skips should fall under the default `success()` as they're not failure or cancel. But if they don't, you'd probably want something along the lines of `always() && ( needs.e2e-tests.result == 'success' || needs.e2e-tests.result == 'skipped' )` to still only run on success or skip.

.github/workflows/e2e-tests.yml
anomiex: > but I assumed that this might run into the same problem as I mentioned earlier - the job not even trying to start if the previous job in the chain was cancelled.

That appears to be intentional too. The following comment after the one I linked shows some tests with cancelled runs which did not run these jobs and that was described as "validated".

Another good reason to fix the bogus "cancelled" you introduced above (see line 58).

.github/workflows/e2e-tests.yml
anomiex: Turns out that didn't work for multiple reasons: you can't specify too long of a list, and also the path passed to actions/cache/restore needs to match the one passed to actions/cache/save for some reason.

I managed to come up with something else though, and it brought the caches for Boost and Jetpack down from 280M and 310M to 8.7M and 44M. Seems worth the effort.

haqadn (approved):
I am giving it a ✅ because it looks like the changes worked, the reviewers were addressed, and I couldn't find any obvious issues. However, I do not have good knowledge of the e2e-tests.yml configuration. So, if in doubt, please get another ✅ 

image

anomiex:
Looks like this broke the `repository_dispatch` runs, e.g. https://github.com/Automattic/jetpack/actions/runs/6024492530 (although [some of them seem to have been broken in different ways before](https://github.com/Automattic/jetpack/actions/runs/6010764087)).

I'll look at making a followup PR.

PR #32690, opened 2023-08-25
Boost: Show Performance History upgrade notice

haqadn:
Fixes Automattic/boost-cloud#362

## Proposed changes:
* Added a new Popover react component in js-packages/components
* Updated storybook config to easily develop react components in boost by allowing scss import from within boost
* Show upgrade notice for free users in Boost.

### Other information:

## Jetpack product discussion
None

None

## Testing instructions:

<details>

image
<summary>Test to see if you see the upgrade prompt in performance history panel</summary>

</details>

<details>

image

<summary>Go to projects/js-packages/storybook and run `pnpm run storybook:dev` and see if Boost components are listed.</summary>

</details>


projects/js-packages/components/components/boost-score-graph/uplot-line-chart.tsx
```diff
@@ -95,8 +95,8 @@ export default function UplotLineChart( { data, range, periods }: UplotChartProp
 	const uplot = useRef< uPlot | null >( null );
 	const uplotContainer = useRef( null );
 
-	const lastDesktopScore = data[ 1 ][ data[ 1 ].length - 1 ];
-	const lastMobileScore = data[ 2 ][ data[ 2 ].length - 1 ];
+	const lastDesktopScore = typeof data[ 1 ] !== 'undefined' ? data[ 1 ][ data[ 1 ].length - 1 ] : 0;
+	const lastMobileScore = typeof data[ 2 ] !== 'undefined' ? data[ 2 ][ data[ 2 ].length - 1 ] : 0;
```
pyronaur: Might be worth describing exactly what `uPlot.AlignedData` is.

This `typeof` check doesn't inspire confidence. If it can be undefined, then may be it can be other types too?

projects/js-packages/components/components/popover/style.scss
```diff
@@ -0,0 +1,21 @@
+.jp-popover {
+	width: 600px;
```
pyronaur, line 2: This might be problematic for smaller screens?

pyronaur (changes requested):
Looks good, but I left 2 minor comments of uncertainties I had during the review.

projects/js-packages/components/components/popover/style.scss
haqadn: There is a max-width below it which would contain it for mobile devices. However, I have improved the styles after that.

projects/js-packages/components/components/boost-score-graph/uplot-line-chart.tsx
haqadn: I have updated the docblock to add an example for the `uPlot.AlignedData` structure. I explained the undefined-ness in a comment as well.

projects/js-packages/components/components/boost-score-graph/uplot-line-chart.tsx
```diff
@@ -86,7 +86,7 @@ function getColor( score: number, opacity = 'FF' ) {
  * UplotLineChart component.
  *
  * @param {object} props - The props object for the UplotLineChart component.
- * @param {uPlot.AlignedData} props.data - The data for the uPlot chart.
+ * @param {uPlot.AlignedData} props.data - Formatted data for the uPlot chart. In the format [ [...timestamps], [...desktopScores], [...mobileScores] ].
```
pyronaur: Can we create a typescript type for this? Something like:

```
type uPlotGraphPoint = [ number[], number[], number[] ];
```

and then instead of accepting any UplodChartProp - we accept exactly what this function expects.

```ts
UplotLineChart( data: Array<uPlotGraphPoint | undefined> )
```

I'm still not entirely sure what the expected data format is there - so the above typescript is just illustrative of what I think this should look like.

projects/js-packages/components/components/popover/style.scss
pyronaur: 👍

pyronaur (changes requested):

projects/js-packages/components/components/boost-score-graph/use-boost-score-transform.ts
```diff
@@ -28,6 +28,6 @@ export function useBoostScoreTransform(
 			);
 		}
 
-		return [ timestamps, ...valueArray ];
+		return [ timestamps, ...valueArray ] as ScoreGraphAlignedData;
```
pyronaur: timestamps and value array should be defined as `number[]` and then this could be inferred without `as`

projects/js-packages/components/components/boost-score-graph/index.tsx
```diff
@@ -22,15 +23,21 @@ export interface BoostScoreGraphProps {
...
 	endDate: number;
 	title?: string;
-	isLoading?: boolean;
+	isPlaceholder?: boolean;
 }
 
+export type ScoreGraphAlignedData = [
+	number[], // timestamps
+	number[], // desktop_overall_score
+	number[] // mobile_overall_score
```
pyronaur: Seems like we want this to be an object with keys? If there's an external dependency that requires an array with exactly 3 number entries, we can then do `Object.values()` when interacting with that component.

pyronaur (changes requested):

projects/js-packages/components/components/boost-score-graph/use-boost-score-transform.ts
haqadn: The `valueArray` would always have two child arrays because of `keysToExtract`. Which typsecript wasn't able to detect and that's why I used `as`. Anyway, I have refactored it to explicitly add each array element.

projects/js-packages/components/components/boost-score-graph/index.tsx
haqadn: This value itself is prepared for the third-party library. So, I don't think we should make it an object and then pass it through `Object.values` again.

pyronaur (dismissed):

pyronaur (dismissed):

pyronaur (approved):
:shipit:

PR #32692, opened 2023-08-25
Remove logic that prevents site admins being tracked and add store_admin property to WooCommerce analytics events

### LINKED ISSUE #10427: Add admin views/actions to the woocommerce-analytics module
**Description:** the woocommerce-analytics module implementation excludes store admin actions in the store front end. This approach makes sense to avoid having noisy data coming from test actions. However, it can be quite useful to understand when and how merchants are interacting with the store front end. The best solution to this problem is to start tracking admin views/action and add a property to identify them, allowing us to segment the data.
This will need to be implemented in the Jetpack repo. More info about prior art here: pb0Spc-ou-p2#comment-701

**Acceptance criteria:**

- [ ] Change the logic to ensure admin views/actions are tracked when the store admin interacts with all the pages included in the woocommerce-analytics module (product, cart, checkout, order confirmation)
- [ ] Add a `store_admin` [true|false] property to identify when the event is fired by a store admin

Ideally, we should include this change in the next Jetpack release.

opr:
Fixes https://github.com/woocommerce/woocommerce-blocks/issues/10427

## Proposed changes:

* Change `should_track_store` function so it no longer checks whether the user is an administrator, admins will now be tracked.
* Add a `store_admin` property when sending events, true if the user is an admin, false otherwise.

### Other information:

## Jetpack product discussion

A previous discussion about this tracking can be found here pb0Spc-ou-p2#comment-701 however it is very old.

We also have pca54o-6dR-p2 where the effort is listed as a cycle goal (point 3). 

Yes, we now track store admins, as well as regular store users. We also track whether the user is logged in as a store admin on every WooCommerce analytics event.

## Testing instructions:

* Ensure site is connected, WooCommerce is installed, products, payment methods, and shipping methods are available. (Cash on Delivery and Free shipping will be fine).
* Ensure WooCommerce analytics is running.
* As an admin user: add an item to your cart and go to the checkout.
* Check out and then visit Tracks and find your event. (I spoofed my user agent so I could find the event easily)
* Check the event for the `store_admin` property, which should be `1`
* Repeat as a logged _out_ (e.g. guest) user, the event should be logged, and should have the `store_admin` property but it should be `0`
* Repeat as a logged in, but _not_ admin user, (e.g. a customer), the event should be logged, and should have the `store_admin` property but it should be `0`

### DIFF
--- a/projects/plugins/jetpack/changelog/add-store-admin-property
+++ b/projects/plugins/jetpack/changelog/add-store-admin-property
@@ -1,0 +1,4 @@
+Significance: minor
+Type: other
+
+Track store admin actions in WooCommerce analytics. Add a store_admin property to all WooCommerce analytics events.
--- a/projects/plugins/jetpack/modules/woocommerce-analytics/class-jetpack-woocommerce-analytics.php
+++ b/projects/plugins/jetpack/modules/woocommerce-analytics/class-jetpack-woocommerce-analytics.php
@@ -50,10 +50,6 @@ class Jetpack_WooCommerce_Analytics {
 		if ( is_admin() ) {
 			return false;
 		}
-		// Don't track site admins.
-		if ( is_user_logged_in() && in_array( 'administrator', wp_get_current_user()->roles, true ) ) {
-			return false;
-		}
 		// Make sure Jetpack is installed and connected.
 		if ( ! Jetpack::is_connection_ready() ) {
 			return false;
--- a/projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-universal.php
+++ b/projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-universal.php
@@ -90,6 +90,7 @@ class Jetpack_WooCommerce_Analytics_Universal {
 			'ui'          => $this->get_user_id(),
 			'url'         => home_url(),
 			'woo_version' => WC()->version,
+			'store_admin' => in_array( 'administrator', wp_get_current_user()->roles, true ),
 		);
 		$cart_checkout_info = self::get_cart_checkout_info();
 		return array_merge( $site_info, $cart_checkout_info );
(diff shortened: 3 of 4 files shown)


jeherve (dismissed):
This looks fine by me, but I'll let someone more familiar with the project do a final approval. 👍

jenhooks:
Hi @opr! Regarding the privacy update:

> Yes, we now track store admins, as well as regular store users. We also track whether the user is logged in as a store admin on every WooCommerce analytics event.

I'm assuming [this is the page it will go on](https://jetpack.com/support/privacy/woocommerce-analytics/#privacy) - can you confirm? Also, can you make a recommendation for the verbiage to use? I assume the "Site Owners / Users" sections under both the "Activity Tracked" and "Data Synced" areas will need to be updated.

opr:
> Hi @opr! Regarding the privacy update:
> 
> > Yes, we now track store admins, as well as regular store users. We also track whether the user is logged in as a store admin on every WooCommerce analytics event.
> 
> I'm assuming [this is the page it will go on](https://jetpack.com/support/privacy/woocommerce-analytics/#privacy) - can you confirm? Also, can you make a recommendation for the verbiage to use? I assume the "Site Owners / Users" sections under both the "Activity Tracked" and "Data Synced" areas will need to be updated.

Yep that page seems right!

I think the change we'd need to add would be under `Activity Tracked -> Site Owners / Users` and we would repeat the information from the `Site Visitors` section.

> We also track when site owners / users view products, add and remove products from their cart, view checkout, and complete orders. We also track device used. 
>
> Events include metadata such as page URL; product name, product ID, quantity, and product category; unique user ID (randomly generated identifier); WooCommerce version; blocks used on cart & checkout pages.

I skipped the stuff about the account creation since it is already mentioned in the paragraph above.

I don't _think_ we have to update data synced section though, but not 100% sure.

senadir (dismissed):
Confirmed this is working fine! If there's no other blocks, I will merge this after rebasing it.

senadir (approved):
Reapproving after rebase

Issue #32602, opened 2023-08-21
Blocks: use a single source of truth for blocks metadata


monsieur-z:
Following https://github.com/Automattic/jetpack/issues/32408, a `block.json` metadata file was created for each block so that wp.org could list them on [the Jetpack plugin page](https://wordpress.org/plugins/jetpack/). However, the block registration methods in both the back and front ends don't consider these files yet: they read metadata from an object that's exported in a JS file. Metadata is duplicated.

The recommended method for block registration is to specify a path to a `block.json` file. We should try that and get rid of data that's not present in that file.
```[tasklist]
### Tasks
- [ ] https://github.com/Automattic/jetpack/pull/32697
- [x] Switch all Beta blocks to be registered via their block.json file
- [ ] https://github.com/Automattic/jetpack/pull/32698
- [ ] https://github.com/Automattic/jetpack/pull/32815
- [ ] https://github.com/Automattic/jetpack/pull/33073
- [ ] https://github.com/Automattic/jetpack/pull/33076
- [ ] https://github.com/Automattic/jetpack/issues/33174
- [ ] https://github.com/Automattic/jetpack/issues/32603
- [ ] https://github.com/Automattic/jetpack/issues/32604
- [ ] https://github.com/Automattic/jetpack/issues/33173
- [x] Switch all Experimental blocks to be registered via their block.json file
```

PR #32697, opened 2023-08-25
Enable block registration by specifying the path to block.json

monsieur-z:
See https://github.com/Automattic/jetpack/issues/32602

## Proposed changes:

Allows the registration of blocks by passing the path to the `block.json` metadata file to `Blocks::jetpack_register_block`.

In more details:
- In PHP
  - Update `Blocks::jetpack_register_block` and `Jetpack_Gutenberg::load_assets_as_required` to accept a path as first argument
  - Add helper methods
  - Add unit tests
- In JS
  - Make sure the `beta` mention is only rendered once
  - Make `registerJetpackBlock` accept both prefixed and unprefixed names

### Other information:

## Jetpack product discussion

pedMtX-RS-p2

No.

## Testing instructions:

Regression testing:
- Check that Jetpack blocks are still working as expected

Functional testing:
- Test in coordination with https://github.com/Automattic/jetpack/pull/32698

### DIFF
--- a/projects/packages/blocks/src/class-blocks.php
+++ b/projects/packages/blocks/src/class-blocks.php
@@ -135,7 +135,7 @@ class Blocks {
 		if ( file_exists( $filename ) ) {
 			try {
 				$metadata = wp_json_file_decode( $filename, array( 'associative' => true ) );
-			} catch ( Exception $e ) {
+			} catch ( \Exception $e ) {
 				$metadata = array();
 			}
 		}
--- a/projects/packages/blocks/src/class-blocks.php
+++ b/projects/packages/blocks/src/class-blocks.php
@@ -27,7 +27,7 @@ class Blocks {
 	 *
 	 * @since 1.1.0
 	 *
-	 * @param string $slug Slug of the block.
+	 * @param string $slug Slug of the block or absolute path to the directory containing the block.json file.
 	 * @param array  $args {
 	 *     Arguments that are passed into register_block_type.
 	 *     See register_block_type for full list of arguments.
@@ -40,6 +40,22 @@ class Blocks {
 	 * @return WP_Block_Type|false The registered block type on success, or false on failure.
 	 */
 	public static function jetpack_register_block( $slug, $args = array() ) {
+		$block_type = null;
+
+		// If the path to block.json is passed, find the slug in the file then create a block type
+		// object to register the block.
+		// Note: passing the path directly to register_block_type seems to loose the interactivity of
+		// the block once in the editor once it's out of focus.
+		if ( '/' === substr( $slug, 0, 1 ) ) {
+			$metadata = self::get_block_metadata_from_file( $slug );
+			$name     = self::get_block_name_from_metadata( $metadata );
+
+			if ( ! empty( $name ) ) {
+				$slug       = $name;
+				$block_type = new \WP_Block_Type( $slug, array_merge( $metadata, $args ) );
+			}
+		}
+
 		if ( 0 !== strpos( $slug, 'jetpack/' ) && ! strpos( $slug, '/' ) ) {
 			_doing_it_wrong( 'jetpack_register_block', 'Prefix the block with jetpack/ ', 'Jetpack 9.0.0' );
 			$slug = 'jetpack/' . $slug;
@@ -99,7 +115,53 @@ class Blocks {
 			}
 		}
 
-		return register_block_type( $slug, $args );
+		return register_block_type( isset( $block_type ) ? $block_type : $slug, $args );
+	}
+
+	/**
+	 * Read block metadata from a block.json file.
+	 *
+	 * @param string $filename The path to the block.json file or its directory.
+	 *
+	 * @return array The block metadata.
+	 */
+	public static function get_block_metadata_from_file( $filename ) {
+		$metadata = array();
+		$needle   = '/block.json';
+		$filename = $needle === substr( $filename, -strlen( $needle ) ) ? $filename : $filename . $needle;
+
+		if ( file_exists( $filename ) ) {
+			try {
+				// phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents
+				$metadata = json_decode( file_get_contents( $filename ), true );
+			} catch ( Exception $e ) {
+				$metadata = array();
+			}
+		}
+
+		return $metadata;
+	}
+
+	/**
+	 * Get the block name from the its metadata.
+	 *
+	 * @param array $metadata The block metadata.
+	 *
+	 * @return string The block name.
+	 */
+	public static function get_block_name_from_metadata( $metadata ) {
+		return ! isset( $metadata['name'] ) || empty( $metadata['name'] ) ? '' : $metadata['name'];
+	}
+
+	/**
+	 * Get the block feature name (i.e. the name without the `jetpack` prefix) from its metadata.
+	 *
+	 * @param array $metadata The block metadata.
+	 *
+	 * @return string The block feature name.
+	 */
+	public static function get_block_feature_from_metadata( $metadata ) {
+		return str_replace( 'jetpack/', '', self::get_block_name_from_metadata( $metadata ) );
 	}
 
 	/**
--- a/projects/packages/blocks/src/class-blocks.php
+++ b/projects/packages/blocks/src/class-blocks.php
@@ -40,7 +40,7 @@ class Blocks {
 	 * @return WP_Block_Type|false The registered block type on success, or false on failure.
 	 */
 	public static function jetpack_register_block( $slug, $args = array() ) {
-		$block_type = null;
+		$block_type = $slug;
 
 		// If the path to block.json is passed, find the slug in the file then create a block type
 		// object to register the block.
@@ -115,7 +115,7 @@ class Blocks {
 			}
 		}
 
-		return register_block_type( isset( $block_type ) ? $block_type : $slug, $args );
+		return register_block_type( $block_type, $args );
 	}
 
 	/**
--- a/projects/packages/blocks/src/class-blocks.php
+++ b/projects/packages/blocks/src/class-blocks.php
@@ -40,7 +40,7 @@ class Blocks {
 	 * @return WP_Block_Type|false The registered block type on success, or false on failure.
 	 */
 	public static function jetpack_register_block( $slug, $args = array() ) {
-		if ( 0 !== strpos( $slug, 'jetpack/' ) && ! strpos( $slug, '/' ) ) {
+		if ( 0 !== strpos( $slug, 'jetpack/' ) && 0 !== strpos( $slug, '/' ) ) {
 			_doing_it_wrong( 'jetpack_register_block', 'Prefix the block with jetpack/ ', 'Jetpack 9.0.0' );
 			$slug = 'jetpack/' . $slug;
 		}
--- a/projects/packages/blocks/src/class-blocks.php
+++ b/projects/packages/blocks/src/class-blocks.php
@@ -40,7 +40,9 @@ class Blocks {
 	 * @return WP_Block_Type|false The registered block type on success, or false on failure.
 	 */
 	public static function jetpack_register_block( $slug, $args = array() ) {
-		if ( 0 !== strpos( $slug, 'jetpack/' ) && 0 !== strpos( $slug, '/' ) ) {
+		// Slug doesn't start with `jetpack/`, isn't an absolute path, or doesn't contain a slash
+		// (synonym of a namespace) at all.
+		if ( 0 !== strpos( $slug, 'jetpack/' ) && 0 !== strpos( $slug, '/' ) && ! strpos( $slug, '/' ) ) {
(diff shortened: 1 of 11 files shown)


projects/plugins/jetpack/extensions/shared/register-jetpack-block.js
```diff
@@ -16,28 +18,32 @@ import { addFilter } from '@wordpress/hooks';
  * @returns {object|boolean} Either false if the block is not available, or the results of `registerBlockType`
  */
 export default function registerJetpackBlock( name, settings, childBlocks = [], prefix = true ) {
```
monsieur-z, line 22: In `block.json`, `name` must be prefixed with `jetpack/`. These changes account for that.

projects/plugins/jetpack/extensions/editor.js
```diff
@@ -78,10 +78,16 @@ function setBetaBlockTitle( settings, name ) {
...
 
+	const { title, keywords } = settings;
+	const titleSuffix = '(beta)';
+	const betaKeyword = 'beta';
+
 	return {
 		...settings,
-		title: `${ settings.title } (beta)`,
-		kewords: [ ...settings.keywords, 'beta' ],
+		title: title.toLowerCase().endsWith( titleSuffix )
```
monsieur-z, line 87: Some blocks have a `name` suffixed by `(beta)` in their `block.json` file. This makes sure the 'beta' mention doesn't appear more than once.

projects/packages/blocks/src/class-blocks.php
```diff
@@ -99,7 +115,53 @@ function () use ( $feature_name, $method_name ) {
 			}
 		}
 
-		return register_block_type( $slug, $args );
+		return register_block_type( isset( $block_type ) ? $block_type : $slug, $args );
```
andrii-lysenko: Just a small nitpick, maybe we could avoid using the ternary operator inside the function call and instead declare some new variable using it there.

projects/packages/blocks/src/class-blocks.php
```diff
@@ -40,6 +40,22 @@ class Blocks {
...
 	 */
 	public static function jetpack_register_block( $slug, $args = array() ) {
+		$block_type = null;
+
+		// If the path to block.json is passed, find the slug in the file then create a block type
+		// object to register the block.
+		// Note: passing the path directly to register_block_type seems to loose the interactivity of
+		// the block once in the editor once it's out of focus.
+		if ( '/' === substr( $slug, 0, 1 ) ) {
+			$metadata = self::get_block_metadata_from_file( $slug );
```
andrii-lysenko, line 57: Does this happen on build or in runtime? If it's a runtime is it slowing down the registration?

andrii-lysenko (commented):
The changes looks reasonable and good aside of some questions. However, I'm not having as much experience so suggest for an additional pair of eyes on this one. Maybe @jeherve or @kraftbj could take a look?

projects/packages/blocks/src/class-blocks.php
monsieur-z: At runtime. I assume it would be fast since passing the path to `block.json` when calling [register_block_type](https://developer.wordpress.org/reference/functions/register_block_type/) is the recommended way to register blocks.

projects/packages/blocks/src/class-blocks.php
```diff
@@ -99,7 +115,53 @@ function () use ( $feature_name, $method_name ) {
...
+	}
+
+	/**
+	 * Read block metadata from a block.json file.
+	 *
+	 * @param string $filename The path to the block.json file or its directory.
+	 *
+	 * @return array The block metadata.
+	 */
+	public static function get_block_metadata_from_file( $filename ) {
```
jeherve, line 130: Do you think you could add tests for the new methods you're introducing, in `projects/packages/blocks/tests/php/test-blocks.php`?

projects/packages/blocks/src/class-blocks.php
```diff
@@ -99,7 +115,53 @@ function () use ( $feature_name, $method_name ) {
...
+		$metadata = array();
+		$needle   = '/block.json';
+		$filename = $needle === substr( $filename, -strlen( $needle ) ) ? $filename : $filename . $needle;
+
+		if ( file_exists( $filename ) ) {
+			try {
+				// phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents
+				$metadata = json_decode( file_get_contents( $filename ), true );
```
jeherve: Could we rely on `wp_json_file_decode` here?

projects/plugins/jetpack/extensions/editor.js
jeherve: 👍 Maybe we can remove the (beta) suffix from the block.json file entries now, and then land this so we don't need the extra logic?

projects/plugins/jetpack/extensions/shared/register-jetpack-block.js
jeherve: Maybe we could extend the function's docblock to explain what `name` can be / should be?

projects/plugins/jetpack/extensions/editor.js
monsieur-z: The beta mention in the `block.json` name and description is necessary if we want to display it in [the Jetpack plugin wp.org page](https://wordpress.org/plugins/jetpack/). wp.org only parses `block.json` to generate the blocks listing.

projects/packages/blocks/src/class-blocks.php
monsieur-z: Sure thing. I'm just curious why: is it a WordPress or A8C best practice?

projects/packages/blocks/src/class-blocks.php
monsieur-z: Wasn't aware of that function, thanks!

projects/packages/blocks/src/class-blocks.php
kraftbj: I'm not sure if it is a formal "best practice", but I can see where it is harder to parse while reading what is being returned with the operator in there.

projects/plugins/jetpack/extensions/editor.js
jeherve: Ah right, I forgot about that!

kraftbj:
Unit test failures

>  Unexpected incorrect usage notice for jetpack_register_block.
  Prefix the block with jetpack/  (This message was added in version Jetpack 9.0.0.)
  Unexpected incorrect usage notice for WP_Block_Type_Registry::register.
  Block type names must contain a namespace prefix. Example: my-plugin/my-custom-block-type (This message was added in version 5.0.0.)
  Failed asserting that an array is empty.

projects/packages/blocks/src/class-blocks.php
```diff
@@ -40,11 +40,29 @@ class Blocks {
 	 * @return WP_Block_Type|false The registered block type on success, or false on failure.
 	 */
 	public static function jetpack_register_block( $slug, $args = array() ) {
-		if ( 0 !== strpos( $slug, 'jetpack/' ) && ! strpos( $slug, '/' ) ) {
+		// Slug doesn't start with `jetpack/`, isn't an absolute path, or doesn't contain a slash
+		// (synonym of a namespace) at all.
+		if ( 0 !== strpos( $slug, 'jetpack/' ) && 0 !== strpos( $slug, '/' ) && ! strpos( $slug, '/' ) ) {
```
monsieur-z, line 45: I prefer to be explicit about the conditions here.

projects/packages/blocks/tests/php/test-blocks.php
```diff
@@ -330,4 +330,99 @@ public function test_jetpack_register_block_with_existing_editor_style() {
...
+
+		$result = Blocks::get_block_name_from_metadata( array( 'name' => $name ) );
+
+		$this->assertEquals( $result, $name );
+	}
+
+	/**
+	 * Test reading the feature name of a block from its metadata.
+	 *
+	 * @since 1.4.23
```
jeherve: This will be automatically replaced by the correct version number when we next release a new version of the Blocks package (which may be 1.5.0 or something else).

```suggestion
	 * @since $$next-version$$
```

For reference:
https://github.com/Automattic/jetpack/blob/3a3c4adc1c0913bac224ddb832713bd6101fe1ab/projects/packages/README.md#L107-L116

projects/packages/blocks/tests/php/test-blocks.php
```diff
@@ -330,4 +330,99 @@ public function test_jetpack_register_block_with_existing_editor_style() {
...
+
+		// phpcs:ignore MediaWiki.PHPUnit.SpecificAssertions.assertIsArray -- assertIsArray not supported by all PHP versions we support.
+		$this->assertTrue( is_array( $result ) );
+		$this->assertEmpty( $result );
+	}
+
+	/**
+	 * Test reading the name of a block from its metadata.
+	 *
+	 * @since 1.4.23
```
jeherve:
```suggestion
	 * @since $$next-version$$
```

projects/packages/blocks/tests/php/test-blocks.php
```diff
@@ -330,4 +330,99 @@ public function test_jetpack_register_block_with_existing_editor_style() {
...
+
+		// phpcs:ignore MediaWiki.PHPUnit.SpecificAssertions.assertIsArray -- assertIsArray not supported by all PHP versions we support.
+		$this->assertTrue( is_array( $result ) );
+		$this->assertNotEmpty( $result );
+	}
+
+	/**
+	 * Test reading metadata from a file that doesn't exist.
+	 *
+	 * @since 1.4.23
```
jeherve:
```suggestion
	 * @since $$next-version$$
```

projects/packages/blocks/tests/php/test-blocks.php
```diff
@@ -330,4 +330,99 @@ public function test_jetpack_register_block_with_existing_editor_style() {
...
+
+		// phpcs:ignore MediaWiki.PHPUnit.SpecificAssertions.assertIsArray -- assertIsArray not supported by all PHP versions we support.
+		$this->assertTrue( is_array( $result ) );
+		$this->assertNotEmpty( $result );
+	}
+
+	/**
+	 * Test reading metadata from a block.json file by specifying its folder.
+	 *
+	 * @since 1.4.23
```
jeherve:
```suggestion
	 * @since $$next-version$$
```

projects/packages/blocks/tests/php/test-blocks.php
```diff
@@ -330,4 +330,99 @@ public function test_jetpack_register_block_with_existing_editor_style() {
...
+	public function test_jetpack_register_block_from_metadata_file() {
+		$result = Blocks::jetpack_register_block( __DIR__ . '/fixtures/block.json' );
+
+		$this->assertInstanceOf( 'WP_Block_Type', $result );
+	}
+
+	/**
+	 * Test reading metadata from a block.json file by specifying its path.
+	 *
+	 * @since 1.4.23
```
jeherve:
```suggestion
	 * @since $$next-version$$
```

projects/packages/blocks/tests/php/test-blocks.php
```diff
@@ -330,4 +330,99 @@ public function test_jetpack_register_block_with_existing_editor_style() {
 			remove_filter( 'jetpack_is_standalone_block', '__return_false' );
 		}
 	}
+
+	/**
+	 * Test registering a block by specifying the path to its metadata file.
+	 *
+	 * @since 1.4.23
```
jeherve:
```suggestion
	 * @since $$next-version$$
```

jeherve (changes requested):

jeherve (dismissed):
This looks good to me now, it should be good to go imo!

projects/packages/blocks/src/class-blocks.php
```diff
@@ -99,7 +117,52 @@ function () use ( $feature_name, $method_name ) {
...
+	 */
+	public static function get_block_metadata_from_file( $filename ) {
+		$metadata = array();
+		$needle   = '/block.json';
+		$filename = $needle === substr( $filename, -strlen( $needle ) ) ? $filename : realpath( $filename . $needle );
+
+		if ( file_exists( $filename ) ) {
+			try {
+				$metadata = wp_json_file_decode( $filename, array( 'associative' => true ) );
+			} catch ( Exception $e ) {
```
jeherve: I'm not sure if this is really necessary, but wanted to suggest it anyway:

```suggestion
			} catch ( \Exception $e ) {
```

jeherve (approved):

projects/packages/blocks/src/class-blocks.php
mreishus: Core ended up moving away from parsing individual block.json files: https://make.wordpress.org/core/2022/10/07/improved-php-performance-for-core-blocks-registration/

I am proposing a 3rd party block registration system here: https://core.trac.wordpress.org/ticket/62002

PR #32698, opened 2023-08-25
Refactor: register business hours block with block.json path 

monsieur-z:
See https://github.com/Automattic/jetpack/issues/32602

## Proposed changes:

Registers the _Business Hours_ block by specifying the path to its `block.json` file. This was made possible by https://github.com/Automattic/jetpack/pull/32697. The goal is to avoid the duplication of block metadata and use `block.json` as the source of truth.

In more details:
- Server side
  - Add a helper method to get the path to a block's metadata file (in the build folder)
  - Add tests
  - Register the block by specifying the path to `block.json`
- Client side
  - Remove `editorScript` from `block.json` file (automatically added during the build)
  - Move metadata from `index.js` to `block.json`
  - Delete `index.js`
  - Create a function that reads the icon from the metadata and generates a React component, in `shared-extension-utils`
  - Create a helper to register blocks by passing the metadata object
  - Register the block by passing the metadata
  - Write i18n-friendly CSS
  - Update block tests

### Other information:

## Jetpack product discussion

pedMtX-RS-p2

No.

## Testing instructions:

Test that the _Business Hours_ block behaves as in production.

Screenshot 2023-09-13 at 3 50 35 PM

_Note: blocks metadata don't seem to be localized at the moment in production._


projects/plugins/jetpack/extensions/blocks/business-hours/block.json
```diff
@@ -84,5 +84,64 @@
 			]
 		}
 	},
+	"example": {
```
monsieur-z, line 87: From `index.js`.

projects/plugins/jetpack/extensions/blocks/business-hours/business-hours.php
```diff
@@ -22,23 +28,9 @@
  */
 function register_block() {
 	Blocks::jetpack_register_block(
-		BLOCK_NAME,
+		get_block_path(),
 		array(
 			'render_callback' => __NAMESPACE__ . '\render',
-			'supports'        => array(
```
monsieur-z, line 28: Data already present in `block.json`.

projects/plugins/jetpack/extensions/blocks/business-hours/style.scss
```diff
@@ -8,7 +8,7 @@
 
 	dt {
 		font-weight: bold;
-		margin-right: 0.5em;
+		margin-inline-end: 0.5em;
```
monsieur-z, line 11: Styles are still registered and enqueued manually in PHP. Eventually, they'll be determined from `block.json` and we won't be able to specify `view.css` or `view.rtl.css` at runtime. Let's write code that supports internationalization from the get-go.

projects/packages/blocks/src/class-blocks.php
```diff
@@ -350,4 +350,20 @@ public static function is_standalone_block() {
...
+	 * Returns the path to the directory (in the build folder) containing the block.json metadata
+	 * file of a block.
+	 *
+	 * @since 1.4.23
+	 *
+	 * @param string $block_src_dir The path to the folder containing the block source code.
+	 *
+	 * @return string
+	 */
+	public static function get_path_to_block_metadata( $block_src_dir ) {
```
monsieur-z: I tried to add a test for that function, but I'm having trouble dealing with constants. Any insight would be appreciated.

monsieur-z: Using `Constants::set_constant( 'JETPACK__PLUGIN_FILE', '/a/b/c/index.php' );` at the beginning of my test doesn't work. I end up with the following error: `Error: Undefined constant "Automattic\Jetpack\JETPACK__PLUGIN_FILE"`.

projects/plugins/jetpack/extensions/blocks/business-hours/block.json
```diff
@@ -84,5 +84,63 @@
 			]
 		}
 	},
-	"editorScript": "file:../editor.js"
```
monsieur-z, line 87: Now generated automatically by the build step.

projects/plugins/jetpack/extensions/shared/get-block-icon-from-metadata.js
```diff
@@ -0,0 +1,32 @@
+/* eslint-disable jsdoc/no-undefined-types */
```
monsieur-z: JSDoc doesn't like `React.Component` as a type.

monsieur-z:
I did rebase the branch with `add/enable-block-registration-by-path`, but PHP tests are still failing here. Running `jetpack docker phpunit` locally from `projects/plugins/jetpack` doesn't show any failing test: I believe this may resolve itself once the base branch is merged.

projects/plugins/jetpack/extensions/blocks/business-hours/style.scss
jeherve: Good catch!

projects/plugins/jetpack/extensions/shared/get-block-icon-from-metadata.js
jeherve, line 1: What do you think about moving this to the shared-extension-utils package, so we can use it outside of the Jetpack plugin as well?

projects/plugins/jetpack/extensions/shared/register-jetpack-block.js
```diff
@@ -57,3 +58,24 @@ export default function registerJetpackBlock( name, settings, childBlocks = [],
...
+/**
+ * Wrapper around registerJetpackBlock to register a block by specifying its metadata.
+ *
+ * @param {object }metadata - Metadata of the block (content of block.json)
+ * @param {object} settings - See registerJetpackBlock.
+ * @param {object} childBlocks - See registerJetpackBlock.
+ * @param {boolean} prefix - See registerJetpackBlock.
+ * @returns {object|boolean} Either false if the block is not available, or the results of `registerBlockType`
+ */
+export function registerJetpackBlockFromMetadata( metadata, settings, childBlocks, prefix ) {
```
jeherve, line 71: It's interesting, I would have expected a modified `registerJetpackBlock`, ready to handle different type of data passed to it, but I suppose a new function makes sense as well. 👍 It will probably be easier to understand that way!

projects/packages/blocks/src/class-blocks.php
jeherve: Right, this is a good indicator that we'd need to rework this, I think. `JETPACK__PLUGIN_FILE` will not work when folks want to use the method outside of Jetpack.
Maybe we could do something similar to this?
https://github.com/Automattic/jetpack/blob/3ff19e36551e19df74e26e87f9691846dfdd0be8/projects/packages/assets/src/class-assets.php#L136-L157

projects/packages/blocks/src/class-blocks.php
```diff
@@ -352,4 +352,20 @@ public static function is_standalone_block() {
 		 */
 		return apply_filters( 'jetpack_is_standalone_block', $is_standalone_block );
 	}
+
+	/**
+	 * Returns the path to the directory (in the build folder) containing the block.json metadata
+	 * file of a block.
+	 *
+	 * @since 1.4.23
```
jeherve:
```suggestion
	 * @since $$next-version$$
```

projects/js-packages/shared-extension-utils/src/get-icon-color.js
```diff
@@ -0,0 +1,25 @@
+import colorStudio from '@automattic/color-studio';
```
monsieur-z: Duplication of `/projects/plugins/jetpack/extensions/shared/block-icons.js`. I'll remove that file once all blocks are using `getBlockIconComponent`.

projects/packages/blocks/composer.json
```diff
@@ -3,11 +3,13 @@
...
 	"license": "GPL-2.0-or-later",
-	"require": {},
+	"require": {
+		"automattic/jetpack-constants": "@dev"
+	},
 	"require-dev": {
 		"automattic/wordbless": "dev-master",
 		"brain/monkey": "2.6.1",
-		"yoast/phpunit-polyfills": "1.1.0",
+		"yoast/phpunit-polyfills": "1.0.4",
```
monsieur-z: Composer required a valid version.

projects/plugins/jetpack/extensions/blocks/business-hours/business-hours.php
```diff
@@ -12,33 +12,16 @@
...
-
 /**
  * Registers the block for use in Gutenberg
  * This is done via an action so that we can disable
  * registration if we need to.
  */
 function register_block() {
 	Blocks::jetpack_register_block(
-		BLOCK_NAME,
+		__DIR__,
```
monsieur-z, line 22: I had to replace `Blocks::get_path_to_block_metadata( __DIR__ )` (the path to the `block.json` file in the build folder) with `__DIR__` (the path to the `block.json` file in the source folder). The former implementation worked locally but made the test jobs in the CI pipeline fail: files in the build folder don't seem to exist at the time the tests are run. Please let me know if you have any insights about this.

The `block.json` files in the build folder include paths to scripts and stylesheets. Currently, block assets are registered and enqueued manually, so it doesn't matter which `block.json` we use for registration. But eventually, we'll need to use the one in the build folder.

projects/packages/blocks/src/class-blocks.php
```diff
@@ -49,13 +50,16 @@ public static function jetpack_register_block( $slug, $args = array() ) {
...
-		// If the path to block.json is passed, find the slug in the file then create a block type
-		// object to register the block.
+		// If a path is passed, find the slug in the file then create a block type object to register
+		// the block.
 		// Note: passing the path directly to register_block_type seems to loose the interactivity of
 		// the block once in the editor once it's out of focus.
 		if ( '/' === substr( $slug, 0, 1 ) ) {
-			$metadata = self::get_block_metadata_from_file( $slug );
-			$name     = self::get_block_name_from_metadata( $metadata );
+			// Test if build directory exists. Fall back to reading block.json from source directory.
```
monsieur-z: The build folder for blocks (`_inc/block`) is not available when running the tests from the CI jobs. I'm still looking into why this is the case. In the meantime, we fall back to reading the `block.json` file from the block source directory so that tests pass. Hopefully, this is a temporary solution. Since block assets are currently registered and enqueued manually, reading from the `block.json` file in the build or source folder doesn't matter.

This logic is implemented here so that calling `jetpack_register_block` to register a block stays as straightforward as possible.

projects/plugins/jetpack/class.jetpack-gutenberg.php
```diff
@@ -478,9 +478,13 @@ public static function load_assets_as_required( $type, $script_dependencies = ar
 			return;
 		}
 
-		// Retrieve the feature from block.json if its path is passed.
+		// Retrieve the feature from block.json if a path is passed.
 		if ( '/' === substr( $type, 0, 1 ) ) {
-			$feature = Blocks::get_block_feature_from_metadata( Blocks::get_block_metadata_from_file( $type ) );
+			// Test if build directory exists. Fall back to reading block.json from source directory.
```
monsieur-z: Same comment as for `jetpack_register_block`.

monsieur-z:
Since the last review, I:
- Moved the icon logic to the `shared-extension-utils` package.
- Updated the implementation of `get_path_to_block_metadata` by using `Jetpack_Constants`
- Added tests
- Passed the path to the block source directory to `jetpack_register_block`. As noted in the inline comment, that function takes care of reading the `block.json` file in the build folder, and falls back to the source folder.

projects/js-packages/shared-extension-utils/package.json
```diff
@@ -17,6 +17,7 @@
 		"test": "jest"
 	},
 	"dependencies": {
+		"@automattic/color-studio": "2.5.0",
```
jeherve: It's a bit annoying that we have to bring that whole dependency just for a single color. Part of me would want to just hardcode the color like we do in other places like here:
https://github.com/Automattic/jetpack/blob/3a3c4adc1c0913bac224ddb832713bd6101fe1ab/projects/js-packages/components/components/jetpack-logo/index.tsx#L7

What do you think?

projects/packages/blocks/composer.json
jeherve: This is odd. [1.1.0 is a valid version](https://github.com/Yoast/PHPUnit-Polyfills/releases/tag/1.1.0), and it's what we use all over the monorepo today.

projects/js-packages/shared-extension-utils/package.json
monsieur-z: I'm okay with that.

monsieur-z:
I updated the PR with the following:
- Removed the color studio dependency from the `shared-extension-utils` package and hardcoded the needed color instead.
- Forced version `1.1.0` of `yoast/phpunit-polyfills` after the fact, which didn't seem to cause any problem in the build pipeline.
- Moved the logic to handle a fallback `block.json` directory to `get_path_to_block_metadata` and updated the rest of the code accordingly.

projects/packages/blocks/tests/php/test-blocks.php
```diff
@@ -339,7 +340,7 @@ public function test_jetpack_register_block_with_existing_editor_style() {
 	 * @covers Automattic\Jetpack\Blocks::get_block_metadata_from_file
 	 */
 	public function test_jetpack_register_block_from_metadata_file() {
-		$result = Blocks::jetpack_register_block( __DIR__ . '/fixtures/block.json' );
+		$result = Blocks::jetpack_register_block( __DIR__ . '/fixtures/test-block/block.json' );
```
monsieur-z, line 343: Updated the file structure in `fixtures` for clarity and to facilitate `test_get_path_to_block_metadata`.

fgiannar (commented):
Thanks for working on this, Kevin!

I tested this as follows: 

1. Added a business hours block on latest stable
2. Made a few changes and switched to the current branch
3. Confirmed the block is there and my changes persisted
4. Added a new business hours block and confirmed it behaves as expected
5. Confirmed no related errors in console

The only thing I found confusing (though doesn't look related to your changes) is that when editing the block the hours seem to be in my timezone but in view mode they appear in UTC
Screenshot 2023-09-14 at 10 32 08

Screenshot 2023-09-14 at 10 32 18

jeherve (approved):
Let's give this a try! 🚢

PR #32710, opened 2023-08-28
[Newsletter] Introduction to tiers for paid-subscribers

millerf:
Fixes https://github.com/Automattic/gold/issues/118
Fixes https://github.com/Automattic/gold/issues/119
Fixes https://github.com/Automattic/gold/issues/140

Needs those related PRs:
 - ~WPCOM: D120952-code~ (merged)
 - ~WPCOM: D121198-code~ (merged)
- ~Calypso: https://github.com/Automattic/wp-calypso/pull/81014~ [OLD]
- Calypso: https://github.com/Automattic/wp-calypso/pull/82728

## Proposed changes:

* Adds tiers to Newsletters
* This branch merges all those PRs into a feature branch:
  - https://github.com/Automattic/jetpack/pull/32683
  - https://github.com/Automattic/jetpack/pull/32658
  - https://github.com/Automattic/jetpack/pull/32681
  
### Other information:

No.

### Unit tests

- start docker `jetpack docker up -d`
- `jetpack docker phpunit -- -c tests/php.multisite.xml --testsuite=subscriptions --filter=test_newsletter_tiers`

### Live tests

This needs to be tested on both WPCOM simple, and JN/Atomic/self-hosted sites:

### on Jurassic Ninja or self-hosted: make sure to 
 - make sure  to enable the newsletter options in the jetpack settings.
 - make sure to set `JETPACK__SANDBOX_DOMAIN` to your sandbox address (see `wp-admin/options-general.php?page=companion_settings` in JN) and enable global http in your sandbox.

### Common tests for WPCOM Simple sites and self-hosted
- ~Apply D119740-code on your sandbox~ (merged)
- Patch this PR on your sandbox
```
bin/jetpack-downloader test jetpack add/newsletter-tiers-feature
```

- Run this https://github.com/Automattic/wp-calypso/pull/81014 via `yarn start` or [the "live" link](https://calypso.live/?image=registry.a8c.com/calypso/app:build-78948) in https://github.com/Automattic/wp-calypso/pull/81014
- Use `define( 'USE_STORE_SANDBOX', true );`
- Sandbox `public-api.wordpress.com`, `subscribe.wordpress.com` and the test site (if testing on a simple site).
- on Jurrassic-Ninja , link the instance to call your sandbox 

On a Simple site or a JN site:
- From the Earn dashboard, connect Stripe if needed
- Create a 3 tier plans on a site (Gold, SIlver and Bronze for example) via Calypso
Screenshot 2023-08-28 at 16 58 13

- Create a new post with a Silver tier (example this post: https://bestmemberships.wordpress.com/2023/08/28/silver-tier-content/) 
Screenshot 2023-08-28 at 17 29 53

- Open a new incognito and go to the test page
- You should NOT be able to see the post
Screenshot 2023-08-28 at 17 32 56

- Open a new incognito and Subscribe to a a Silver tier
Screenshot 2023-08-28 at 17 52 07

- You should be able to see the post
Screenshot 2023-08-28 at 18 51 08

- Open a new incognito and Subscribe to a a Bronze tier with a new user (use `EMAIL+%SOME_RANDOM_NB%@YOUR_EMAIL.COM`
- You should be NOT able to see the post
- Open a new incognito and Subscribe to a Gold tier with a new user (use `EMAIL+%SOME_RANDOM_NB%@YOUR_EMAIL.COM`
- You SHOULD be able to see the post

#### Use cases that cannot be tested
It looks like emails are sent from a global async job and therefore is very hard to test (might run into weird issues). So this will need to be tested once it is on production... 
- Send a gated post for Gold -> you should receive an email only on your "Gold" email
- Send a gated post for Bronze -> you should receive the email on your "Gold" and "Silver" email
- Send a gated post for Silver -> you should receive the email  on all the emails

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
+++ b/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
@@ -217,7 +217,7 @@ abstract class Token_Subscription_Service implements Subscription_Service {
 				'fields'         => 'ids',
 				'post_type'      => \Jetpack_Memberships::$post_type_plan,
 				'meta_key'       => 'jetpack_memberships_tier',
-				'meta_value'     => $tier_product_id,
+				'meta_value'     => $tier_id,
 			)
 		);
 
@@ -232,7 +232,8 @@ abstract class Token_Subscription_Service implements Subscription_Service {
 		}
 
 		foreach ( $user_abbreviated_subscriptions as $subscription_plan_id => $details ) {
-			$end = is_int( $details->end_date ) ? $details->end_date : strtotime( $details->end_date );
+			$details = (array) $details;
+			$end = is_int( $details['end_date'] ) ? $details['end_date'] : strtotime( $details['end_date'] );
 			if ( $end < time() ) {
 				// subscription not active anymore
 				continue;
--- a/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
+++ b/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
@@ -219,7 +219,7 @@ abstract class Token_Subscription_Service implements Subscription_Service {
 				)
 			);
 
-			if ( empty( $subscription_post_id ) === 0 ) {
+			if ( empty( $subscription_post_id ) ) {
 				// No post linked to this plan
 				continue;
 			}
--- a/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
+++ b/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
@@ -176,10 +176,15 @@ abstract class Token_Subscription_Service implements Subscription_Service {
 
 		// We now need the tier price and currency, and the same for the annual price (if available)
 		$tier_meta         = get_post_meta( $tier_id );
-		$tier_price        = $tier_meta['jetpack_memberships_price'][0];
-		$tier_currency     = $tier_meta['jetpack_memberships_currency'][0];
-		$tier_product_id   = $tier_meta['jetpack_memberships_product_id'][0];
+		$tier_price        = isset( $tier_meta['jetpack_memberships_price'] ) ? $tier_meta['jetpack_memberships_price'][0] : null;
+		$tier_currency     = isset( $tier_meta['jetpack_memberships_currency'] ) ? $tier_meta['jetpack_memberships_currency'][0] : null;
+		$tier_product_id   = isset( $tier_meta['jetpack_memberships_product_id'] ) ? $tier_meta['jetpack_memberships_product_id'][0] : null;
 		$annual_tier_price = $tier_price * 12;
+
+		if ( $tier_price === null || $tier_currency === null || $tier_product_id === null ) {
+			// There is an issue with the meta
+			return false;
+		}
 
 		// At this point we know the post is
 		$linked_post_tier = get_posts(
@@ -231,9 +236,14 @@ abstract class Token_Subscription_Service implements Subscription_Service {
 			}
 
 			$metas                 = get_post_meta( $subscription_post_id );
-			$subscription_price    = $metas['jetpack_memberships_price'][0];
-			$subscription_currency = $metas['jetpack_memberships_currency'][0];
-			$subscription_interval = $metas['jetpack_memberships_interval'][0];
+			$subscription_price    = isset( $metas['jetpack_memberships_price'] ) ? $metas['jetpack_memberships_price'][0] : null;
+			$subscription_currency = isset( $metas['jetpack_memberships_currency'] ) ? $metas['jetpack_memberships_currency'][0] : null;
+			$subscription_interval = isset( $metas['jetpack_memberships_interval'] ) ? $metas['jetpack_memberships_interval'][0] : null;
+
+			if ( $subscription_price === null || $subscription_currency === null || $subscription_interval === null ) {
+				// There is an issue with the meta
+				continue;
+			}
 
 			if ( $tier_currency !== $subscription_currency ) {
 				// For now, we don't count if there are different currency (not sure how to convert price in a pure JP env)
--- a/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
+++ b/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
@@ -9,6 +9,7 @@
 namespace Automattic\Jetpack\Extensions\Premium_Content\Subscription_Service;
 
 use Automattic\Jetpack\Extensions\Premium_Content\JWT;
+use const Automattic\Jetpack\Extensions\Subscriptions\META_NAME_FOR_POST_TIER_ID_SETTINGS;
 
 /**
  * Class Token_Subscription_Service
@@ -103,9 +104,12 @@ abstract class Token_Subscription_Service implements Subscription_Service {
 			);
 			$subscriptions      = (array) $payload['subscriptions'];
 			$is_paid_subscriber = $this->validate_subscriptions( $valid_plan_ids, $subscriptions );
+		} else {
+			// Token not valid. We bail even unless the post can be accessed publicly.
+			return $this->user_has_access( $access_level, false, false, get_the_ID(), array() );
 		}
 
-		$has_access = $this->user_has_access( $access_level, $is_blog_subscriber, $is_paid_subscriber, get_the_ID() );
+		$has_access = $this->user_has_access( $access_level, $is_blog_subscriber, $is_paid_subscriber, get_the_ID(), $subscriptions );
 		// phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
 		$current_user = $old_user;
 		return $has_access;
@@ -118,10 +122,11 @@ abstract class Token_Subscription_Service implements Subscription_Service {
 	 * @param bool   $is_blog_subscriber Is user a subscriber of the blog.
 	 * @param bool   $is_paid_subscriber Is user a paid subscriber of the blog.
 	 * @param int    $post_id Post ID.
+	 * @param array  $user_abbreviated_subscriptions User subscription abbreviated.
 	 *
 	 * @return bool Whether the user has access to the content.
 	 */
-	protected function user_has_access( $access_level, $is_blog_subscriber, $is_paid_subscriber, $post_id ) {
+	protected function user_has_access( $access_level, $is_blog_subscriber, $is_paid_subscriber, $post_id, $user_abbreviated_subscriptions ) {
 		$has_access = false;
 		if ( is_user_logged_in() && current_user_can( 'edit_post', $post_id ) ) {
 			// Admin has access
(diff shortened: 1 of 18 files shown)


pottedmeat:
Testing on Calypso for a Jetpack Ninja site:

- I created Gold and Silver plans.
- I updated the Gold plan by tripling the priced to 3/36
- It created a new monthly plan with the yearly price.

image

chrisfromthelc (commented):
> using the price order to gate or allow access.

The above quote is from the original issue. 

I ran into a problem with this. I made all tiers $1 when testing, and this seems to break things and not reliably gate content.

There could be cases where a user might want to run a promotion where the price of one tier might match the price or be lower than the price of a lower tier. This would obviously break the intended gating of the content. The logic of the gating needs to be unaware of the pricing.

n3f:
I can verify that the testing instructions are working for me as described with one exception: If you are using a new site (e.g. jurassic ninja) you need to also enable the newsletter options in the jetpack settings.

Also, does this PR remove the option to only have a one-time subscription to a newsletter? Will this functionality come back?

I agree with @chrisfromthelc 's comment too.  Another wrinkle might occur if the admin changes the tier pricing structure such that (this is a contrived example) Bronze costs more than Gold. In that case the Gold subscribers would stop seeing the Bronze posts... (ownerships are matched to the plans they have).

And what happens if an admin scraps the current tier structure and creates a new one?

I can tell you because I just tried.  The existing subscribers no longer have access to old posts (all those they would have "paid" for.)  New subscribers are able to see **_all_** old posts though...

millerf:
@chrisfromthelc 

> There could be cases where a user might want to run a promotion where the price of one tier might match the price or be lower than the price of a lower tier. This would obviously break the intended gating of the content.

Thanks for the feedback Chris. I understand the idea. Maybe we can think about promotions in the future and introduce a "promotion" price in the tier.
 However for the time being (and for Team51 use-case), we will stick to using prices as a way to order tiers.
 
 @n3f 
 > Another wrinkle might occur if the admin changes the tier pricing structure such that (this is a contrived example) Bronze costs more than Gold.
 
 I understand the concern, but as explained, I don't think we will support such use-case in the first implementation. Good to keep in mind, though.
 
 > I ran into a problem with this. I made all tiers $1 when testing, and this seems to break things and not reliably gate content.

Interesting. I need to look into that, because I should have handled this use case: https://github.com/Automattic/jetpack/pull/32710/files#diff-a2de8fe49437410f2cef32a4b1cb8a63cbe0727afd32c4e8312fcc73c913ad8bR229 (if the tier subscribed has the same price as the post tier price, it should not be gated). I'll create a unit-test for this...

> And what happens if an admin scraps the current tier structure and creates a new one?

Interesting use-case...

> The existing subscribers no longer have access to old posts (all those they would have "paid" for.) New subscribers are able to see all old posts though...

🤔 This check:  https://github.com/Automattic/jetpack/pull/32710/files#diff-a2de8fe49437410f2cef32a4b1cb8a63cbe0727afd32c4e8312fcc73c913ad8bR158 should handle that use case and give access to both, as the post tier is not in plans anymore... Could it be that the Jetpakc plan sync was not yet done when you checked?

millerf:
> I can verify that the testing instructions are working for me as described with one exception: If you are using a new site (e.g. jurassic ninja) you need to also enable the newsletter options in the jetpack settings.

Nice catch!


> Also, does this PR remove the option to only have a one-time subscription to a newsletter? Will this functionality come back?

I don't think it is planed... But that is another discussion we can have with @simison ... Maybe adding a third tab with one-tim e plans...

chrismccluskey:
I tested this and it works as described! Looks nice!

n3f:
> And what happens if an admin scraps the current tier structure and creates a new one? _(Previous subscribers can no longer view that old content)_

@millerf , I rechecked this today... And it doesn't appear to be a caching issue.

I created [a post](https://decisive-wasp.jurassic.ninja/2023/08/31/silver-tier-post/) with `silver` access. (After subscribing and then deleting that plan) A subscribed user (nef+mfauax96sp-silver@a8c.com) gets prompted to "resubscribe". I tried with a logged-in version of this user as we typically show that email/PIN login for subscribed users to get them logged in and it's not being displayed... (I appended `token=` with the result of `new \A8C\FSE\Earn\API\V1\Authentication\Auth_Endpoint( 223001755, 0, 'nef+mfauax96sp-silver@a8c.com' )->get_auth_token_without_redirect();`. BTW, the endpoint code is supplying a subscription to the JWT.)

However, this might not be an issue with this PR as the regular subscribe block does the same thing too, even though the user is obviously a subscriber.

n3f:
Also, [this text](https://github.com/Automattic/jetpack/blob/add/newsletter-tiers-feature/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php#L1039) might need to be changed if it's a tiers email...

millerf:
> Also, [this text](https://github.com/Automattic/jetpack/blob/add/newsletter-tiers-feature/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php#L1039) might need to be changed if it's a tiers email...

I am not understanding it here. What would you put? Tiers are a sub-category of paid-newsletters, so I believe this is still relevant...

millerf:
> created [a post](https://decisive-wasp.jurassic.ninja/2023/08/31/silver-tier-post/) with silver access. (After subscribing and then deleting that plan) A subscribed user ([nef+mfauax96sp-silver@a8c.com](mailto:nef+mfauax96sp-silver@a8c.com)) gets prompted to "resubscribe".

I now gets the issue. The issue is not with the tiers, but with paid-subscribers, as we check if the user is subscribed [to one of the "valid plans"](https://github.com/Automattic/jetpack/blob/0ec125e669f558e4e49e9267e40310b03866e552/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php#L233).

So that is more like an issue on the  paid-subscribers level, and a tricky one.

n3f:
> > Also, [this text](https://github.com/Automattic/jetpack/blob/add/newsletter-tiers-feature/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php#L1039) might need to be changed if it's a tiers email...
> 
> I am not understanding it here. What would you put? Tiers are a sub-category of paid-newsletters, so I believe this is still relevant...

The text "You are a free subscriber" might not apply.  This could be a bronze/silver subscriber on a gold tier post. it should probably just remove that sentence. (e.g. "Upgrade your subscription to get access to the rest of this post and other paid-subscriber only content.")

millerf:
> The text "You are a free subscriber" might not apply. This could be a bronze/silver subscriber on a gold tier post. it should probably just remove that sentence. (e.g. "Upgrade your subscription to get access to the rest of this post and other paid-subscriber only content.")

Oh I see. You're right... I'll create a separate task for that.

millerf:
@n3f Let's get this one out?

n3f (dismissed):
LGTM with a couple caveats we're handling.

n3f (dismissed):
still 👍

projects/plugins/jetpack/sal/class.json-api-metadata.php
```diff
@@ -4,6 +4,9 @@
  *
  * @package automattic/jetpack
  */
+
+use const Automattic\Jetpack\Extensions\Subscriptions\META_NAME_FOR_POST_TIER_ID_SETTINGS;
```
jeherve: I realize that's not the first time we use such a pattern, but I think this is the first time we do this outside the block directory. I worry this may cause issue since in some edge-case scenarios, one can filter out some blocks, or the whole blocks:
https://github.com/Automattic/jetpack/blob/cae8e177d313ec448325387d564d5292bd0e62ab/projects/plugins/jetpack/class.jetpack-gutenberg.php#L771-L778

Should we support those cases where this will not be available?

projects/plugins/jetpack/extensions/shared/memberships/settings.js
```diff
@@ -85,6 +90,52 @@ function NewsletterAccessSetupNudge( { stripeConnectUrl, isStripeConnected, hasN
...
+						value={ product.id }
+						onChange={ event => {
+							const obj = {};
+							obj[ META_NAME_FOR_POST_TIER_ID_SETTINGS ] = event?.target?.value;
+							return onChange && onChange( obj );
+						} }
+						id={ `editor-post-tier-${ product.id }` }
+					/>
+					<label htmlFor={ `editor-post-tier-${ product.id }` }>
+						{ product.title } subscribers
```
jeherve: We'll probably want to make this translatable.

projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
```diff
@@ -127,13 +129,114 @@ protected function user_has_access( $access_level, $is_blog_subscriber, $is_paid
...
+				array(
+					'posts_per_page' => 1,
+					'fields'         => 'ids',
+					'post_type'      => \Jetpack_Memberships::$post_type_plan,
+					'meta_key'       => 'jetpack_memberships_product_id',
+					'meta_value'     => $subscription_plan_id,
+				)
+			);
+
+			if ( empty( $subscription_post_id ) === 0 ) {
```
jeherve: We can probably simplify that check?

```suggestion
			if ( empty( $subscription_post_id ) ) {
```

projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
```diff
@@ -127,13 +129,114 @@ protected function user_has_access( $access_level, $is_blog_subscriber, $is_paid
...
+			// If the tier is not in the plans, we bail
+			return false;
+		}
+
+		// We now need the tier price and currency, and the same for the annual price (if available)
+		$tier_meta         = get_post_meta( $tier_id );
+		$tier_price        = $tier_meta['jetpack_memberships_price'][0];
+		$tier_currency     = $tier_meta['jetpack_memberships_currency'][0];
+		$tier_product_id   = $tier_meta['jetpack_memberships_product_id'][0];
+		$annual_tier_price = $tier_price * 12;
```
jeherve: Do we need to do any kind of additional checks and casting to avoid any potential errors here and below? I'm guessing the answer depends on the data that may be saved and available in post meta.

jeherve (commented):
Left a few questions below.

projects/plugins/jetpack/extensions/shared/memberships/settings.scss
```diff
@@ -75,3 +75,12 @@
 	}
 
 }
+
+.editor-post-tiers {
```
simison: A bit generic top level selector — would be good to prefix with `jetpack-`

projects/plugins/jetpack/extensions/shared/memberships/settings.js
```diff
@@ -85,6 +90,52 @@ function NewsletterAccessSetupNudge( { stripeConnectUrl, isStripeConnected, hasN
...
+	// the hooks have to run before any early returns)
+	if ( products.length < 2 ) {
+		return;
+	}
+
+	return (
+		<div className="editor-post-tiers">
+			{ products.map( product => (
+				<div key={ product.id }>
+					<input
```
simison: Would be good to use the core components for basic form elements if possible. See [`RadioControl`](https://wordpress.github.io/gutenberg/?path=/docs/components-radiocontrol--docs). That'll give us better visual consistency. 

Fine to do in a follow-up PR after merging this.

projects/plugins/jetpack/sal/class.json-api-metadata.php
millerf: Should we just include the constant file there? Can it actually be done?

millerf: Fixed

projects/plugins/jetpack/extensions/shared/memberships/settings.js
millerf: Done

projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
millerf: Fixed

projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
millerf: Fixed

projects/plugins/jetpack/extensions/shared/memberships/settings.scss
millerf: Done

projects/plugins/jetpack/extensions/shared/memberships/settings.js
millerf: We'll do it in this one. @n3f Can you do this, please?

n3f: Fix here: https://github.com/Automattic/jetpack/pull/32928/files

millerf:
> I ran into a problem with this. I made all tiers $1 when testing, and this seems to break things and not reliably gate content.

@chrisfromthelc 

I tried creating 3 1$ plans on https://relieved-llama.jurassic.ninja/ (the post with silver tier is https://relieved-llama.jurassic.ninja/2023/09/08/test-silver-tier-with-same-price/).

And everything seems to work fine...

projects/plugins/jetpack/extensions/shared/memberships/settings.js
```diff
@@ -85,6 +98,49 @@ function NewsletterAccessSetupNudge( { stripeConnectUrl, isStripeConnected, hasN
...
+
+	return (
+		<div className="jetpack-editor-post-tiers">
+			<RadioControl
+				label={ __( 'Choose Newsletter Tier', 'jetpack' ) }
+				hideLabelFromVision={ true }
+				selected={ Number( tierId ) }
+				options={ products.map( product => {
+					/* Translators: %s is the tier label created by site user */
+					const label = sprintf( __( '%s subscribers', 'jetpack' ), product.title );
```
jeherve: You'll want to move the translator comment right above the translatable, otherwise the linting tools are not going to be happy:

```suggestion
					const label = sprintf(
						/* Translators: %s is the tier label created by site user */
						__( '%s subscribers', 'jetpack' ),
						product.title
					);
```

millerf: @n3f

n3f (approved):
A lot of good work here.  Glad to see it ready!

PR #32712, opened 2023-08-28
Social: Add the change setting logic to the auto-conversion feature

gmjuhasz:
My previous change added the notice for the auto-conversion, and disabled the connections accordingly. This change adds the functionality to the `Change setting` button, which opens the Jetpack Social admin page, or the Jetpack settings sharing section for Jetpack in a new tab, then refreshes the state on focus

## Proposed changes:

* 

### Other information:

## Testing instructions:

* Turn off Social and have Jetpack enabled.
  * Go to the Jetpack settings page and turn on the auto conversion setting
  * Open up the editor and create a new post
  * Select a media file that is convertible, but not valid for some connections - for example a 10Mb jpg image
  * You should see the notice that it will be converted. If you dismissed already, run `jetpack docker wp option delete jetpack_social_dismissed_notices` to bring it back
  * On the notice click change settings button. It should open up Jetpack settings on the sharing screen.
  * Turn off the auto conversion.
  * Go back to the editor, the page should reflect the changes without needing to refresh

* Do the same with Jetpack Social enabled only. The only difference is that the button should direct you to the social admin page

CleanShot 2023-09-01 at 18 03 14 png


projects/js-packages/publicize-components/src/hooks/use-page-has-focus/index.js
```diff
@@ -0,0 +1,32 @@
...
+
+const hasFocus = () => document?.hasFocus() ?? false;
+
+/**
+ * Hook that returns whether the page has focus or not.
+ *
+ * @param {boolean} runOnRender - Whether to run the callback on render.
+ * @returns {boolean} Whether the page has focus or not.
+ */
+const usePageHasFocus = ( runOnRender = false ) => {
```
pablinos: I'm not sure this hook is working efficiently. A request happens on focus and then the connection test results  request is made, which results in a rerender and two more settings requests are made.

Could we use the library and debouncing that we do for the connection test results? [That uses the react-page-visibility library](https://github.com/Automattic/jetpack/blob/565c2117912e1fc41dc39401798d01f29d50e9b8/projects/js-packages/publicize-components/src/components/settings-button/index.js#L23) which is a component but I see it [does have a hook](https://www.npmjs.com/package/react-page-visibility)

projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -123,6 +125,18 @@ export default function PublicizeForm( {
...
+	const pageHasFocus = usePageHasFocus();
+	const refreshOptions = useDispatch( SOCIAL_STORE_ID ).refreshAutoConversionSettings;
+
+	const isAutoConversionEnabled = useSelect(
+		select => select( SOCIAL_STORE_ID ).isAutoConversionEnabled(),
+		[]
+	);
+
+	if ( pageHasFocus ) {
+		refreshOptions();
```
pablinos: I think you had an idea about only checking the option if we need to. Could we do this only if we're going to render the notice with the 'change setting' link? Then if someone turns it off we'll only update once, and if someone dismisses the notice, we won't update at all.

pablinos (commented):
It's working for me, but I think we need to change the approach with the focus hook. It's triggering a lot of API requests and we could probably reduce the circumstances that we check the setting.

projects/js-packages/publicize-components/src/hooks/use-page-has-focus/index.js
gmjuhasz: Sounds good, changed, now we are using that.

projects/js-packages/publicize-components/src/components/form/index.js
gmjuhasz: Yeah good point, I missed adding that. Added.!

projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -193,7 +221,10 @@ export default function PublicizeForm( {
 			shouldUploadAttachedMedia,
 		}
 	);
-	const shouldAutoConvert = isAutoConversionEnabled && isConvertible;
+	const shouldAutoConvert =
+		( isAutoConversionEnabled === undefined
+			? isAutoConversionEnabledInit
+			: isAutoConversionEnabled ) && isConvertible;
```
gmjuhasz: Since we are kind of using initial state, then the store itself too, we have to have this logic, to use the initial state first, and the store after.

pablinos: Can we initialise the store from the initial state?

projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -125,15 +134,20 @@ export default function PublicizeForm( {
 		[ dismissNotice, NOTICES ]
 	);
 
-	const pageHasFocus = usePageHasFocus();
+	const shouldAutoRefresh = useRef( false );
+	const enableAutoRefresh = useCallback( () => {
+		shouldAutoRefresh.current = true;
+	}, [ shouldAutoRefresh ] );
+
+	const pageHasFocus = usePageVisibility();
```
pablinos: I think we might want to set `shouldAutoRefresh` to `false` again after we have fetched the new settings, because otherwise we call `refreshOptions` on every re-render. We could maybe do it with `useEffect` and have `pageHasFocus` as a dependency.

What do you think about putting all this into a `useAutoConversionSettingUpdate` (I realise the naming could be better) hook?

projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -429,6 +446,7 @@ export default function PublicizeForm( {
 												{ __( 'Got it', 'jetpack' ) }
 											</Button>,
 											<Button
+												onClick={ enableAutoRefresh }
```
pablinos: I was wondering about using `shouldShowNotice` somewhere to enable it rather than needing to click on the button, so if we're showing this notice and we lose focus, when we regain focus we'll dispatch the action to refresh once.

projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -125,15 +134,20 @@ export default function PublicizeForm( {
...
+	const pageHasFocus = usePageVisibility();
 	const refreshOptions = useDispatch( SOCIAL_STORE_ID ).refreshAutoConversionSettings;
 
 	const isAutoConversionEnabled = useSelect(
 		select => select( SOCIAL_STORE_ID ).isAutoConversionEnabled(),
 		[]
 	);
 
-	if ( pageHasFocus ) {
+	if ( pageHasFocus && shouldAutoRefresh.current ) {
```
pablinos: We ended up debouncing the requests for connection-test-results in case someone is flicking back and forth between tabs. Maybe we should do the same here?

I'm not sure why we debounced rather than throttled, but we might want to use throttle instead so an initial call is made quickly but we don't make any further calls for a few seconds.

pablinos (commented):
I know this is actually part of the previous change, but I think we might have to do something with the sizing/loyout of the buttons because when it has focus it looks like this:

image

projects/js-packages/publicize-components/src/components/form/styles.module.scss
```diff
@@ -19,3 +19,9 @@
 	display: inline-block;
 	margin-top: 6px;
 }
+
+.change-settings-button {
+	padding: 0;
+	height: auto;
+	margin-left: 15px;
```
manzoorwanijk: RTL aware CSS

```suggestion
	margin-inline-start: 15px;
```

pablinos (approved):
This looks good. It tests well for me and only makes a single request

PR #32714, opened 2023-08-28
CRM Automations UI: Add store and workflows reducer

robertf4:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3181
Fixes https://github.com/Automattic/zero-bs-crm/issues/3205

## Proposed changes:

This PR accomplishes two tasks:
1. Creates a Redux store for our React apps.
2. Adds a `workflows` reducer to that store for use by the Automations UI.

The store uses `@wordpress/data` and is found at `/src/state/store`. This folder sets the pattern that future projects can use to add new reducers as needed, and the store can then be provided to multiple views so that state can be shared between them.

The workflows reducer will keep the primary state for the Automations UI. I have created an initial set of actions which will support our MVP, although more actions will need to be created for future expanded versions of the Automations module.

### Other information:

## Testing instructions:

This code is not called anywhere yet, so just review the code and make sure the tests are comprehensive, correct, and pass. Run them with `pnpm test` from the root crm directory.


projects/plugins/crm/src/js/state/action-types.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * All action types that can be used in the store should be defined here in order to ensure
+ * they are unique. Keep this file in alphabetical order.
+ */
+export type ACTIVATE_WORKFLOW = 'ACTIVATE_WORKFLOW';
+export type DEACTIVATE_WORKFLOW = 'DEACTIVATE_WORKFLOW';
+export type HYDRATE_WORKFLOWS = 'HYDRATE_WORKFLOWS';
+export type SET_ATTRIBUTE = 'SET_ATTRIBUTE';
```
gogdzl, line 8: ❤️

projects/plugins/crm/src/js/state/automations-admin/reducer.ts
```diff
@@ -0,0 +1,60 @@
...
+					...newState,
+					[ workflow.id ]: getIdentifiedWorkflow( workflow ),
+				} ),
+				{}
+			);
+		case 'ACTIVATE_WORKFLOW':
+			if ( ! action.id || ! state[ action.id ] ) {
+				return state;
+			}
+			return { ...state, ...{ [ action.id ]: { ...state[ action.id ], active: true } } };
```
gogdzl: It could actually be simplified to this:
`return { ...state, [ action.id ]: { ...state[ action.id ], active: true }, };`

projects/plugins/crm/src/js/state/automations-admin/reducer.ts
```diff
@@ -0,0 +1,60 @@
...
+		case 'DEACTIVATE_WORKFLOW':
+			if ( ! action.id || ! state[ action.id ] ) {
+				return state;
+			}
+			return { ...state, ...{ [ action.id ]: { ...state[ action.id ], active: false } } };
```
gogdzl: It could actually be simplified to this:
`return { ...state, [ action.id ]: { ...state[ action.id ], active: false }, };`

gogdzl (changes requested):
I've made two comments, aside from this PR looks good to me.

gogdzl (approved):
Tests look good to me and they all pass:

image

Good job!

PR #32715, opened 2023-08-28
Ensure correct reporting for Cart/Checkout shortcode/blocks properties in WooCommerce analytics events

opr:
## Proposed changes:

* If WooCommerce Blocks is version 10.6.0 or above, it will be using the Cart and Checkout templates to power those WooCommerce pages. In this case, the incorrect values may be reported for the following properties on `woocommerceanalytics_%` events:
  * `cart_page_contains_cart_block`
  * `cart_page_contains_cart_shortcode`
  * `checkout_page_contains_cart_block`
  * `checkout_page_contains_cart_shortcode`
* The changes introduced in this PR are:
* Get the version of WC blocks, if it is greater than 10.6 load the templates from the WC Blocks `BlockTemplatesController` - load the Cart and Checkout templates and check what they contain. This is the same logic as the existing checks, however the string (page content vs. template content) being searched is different.

### Other information:

## Jetpack product discussion

A previous discussion about this tracking can be found here pb0Spc-ou-p2#comment-701 however it is very old.

We also have pca54o-6dR-p2 where the effort is listed as a cycle goal (point 3).

No, it just ensures the data being tracked is accurate, but the data is not changed.

## Testing instructions:

- Ensure site is connected, WooCommerce is installed, products, payment methods, and shipping methods are available. (Cash on Delivery and Free shipping will be fine).
- Install WooCommerce 7.8.0 and a blocks theme e.g. Twenty Twenty-Three.
- Ensure WooCommerce analytics is running.
- Go to Pages -> Cart. Change its content to contain the shortcode block, and have it display the WooCommerce cart. (`[woocommerce_cart]`).
- Do the same for Pages -> Checkout and enter `[woocommerce_checkout]` into the shortcode block.
- Add an item to your cart and go to the checkout.
- Check out and then visit Tracks and find your event. (I spoofed my user agent so I could find the event easily)
- Check the event for the following properties:
  - `cart_page_contains_cart_block`: `false` (or 0)
  - `cart_page_contains_cart_shortcode`: `true` (or 1)
  - `checkout_page_contains_checkout_block`: `false` (or 0)
  - `checkout_page_contains_checkout_shortcode`: true (or 1)
- Go back to the pages from the earlier steps, remove the shortcodes and replace them with the Cart and Checkout blocks.
- Add an item to your cart and check out again, the new values should be:
  - `cart_page_contains_cart_block`: true (or 1)
  - `cart_page_contains_cart_shortcode`: false (or 0)
  - `checkout_page_contains_checkout_block`: true (or 1)
  - `checkout_page_contains_checkout_shortcode`: false (or 0)
- Feel free to change the setup so one page has the shortcode and one page doesn't and mix it up.
- Update to WooCommerce 8.0
- Repeat the steps above, however when you go to Pages -> Checkout and Pages -> Cart you should notice that it opens the site editor instead of the usual post editor.
- Repeat with a classic theme, e.g Storefront


projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-universal.php
```diff
@@ -625,6 +625,48 @@ public static function get_cart_checkout_info() {
...
+					$cart_template     = get_block_template( $cart_template_id );
+					$checkout_template = get_block_template( $checkout_template_id );
+				}
+
+				// Update the info transient with data we got from the templates, if the site isn't using WC Blocks we
+				// won't be doing this so no concern about overwriting.
+				$info = array(
+					'cart_page_contains_cart_block'     => str_contains( $cart_template->content, '<!-- wp:woocommerce/cart' ),
+					'cart_page_contains_cart_shortcode' => str_contains( $cart_template->content, '[woocommerce_cart]' ),
+					'checkout_page_contains_cart_block' => str_contains( $checkout_template->content, '<!-- wp:woocommerce/checkout' ),
```
jeherve: Hm, this should not trigger an error since WordPress includes a polyfill. @anomiex Do you know why this is happening?

projects/plugins/jetpack/to-test.md
```diff
@@ -5,6 +5,32 @@
 - **At any point during your testing, remember to [check your browser's JavaScript console](https://wordpress.org/support/article/using-your-browser-to-diagnose-javascript-errors/#step-3-diagnosis) and see if there are any errors reported by Jetpack there.**
 - Use the "Debug Bar" or "Query Monitor" WordPress plugins to help make PHP notices and warnings more noticeable and report anything of note you see.
 
+### WooCommerce Analytics
```
jeherve, line 8: Thanks for including those testing instructions! 👍 I've also added the "Needs testing" label, so the release manager knows to expect this.

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-universal.php
anomiex: It's a little complicated.

For the main phpcs run that uses PHPCompatibilityWP, there's a disable in that ruleset there so no warning. But since PHPCompatibility hasn't had a release since 2019, we also do a run using their "develop" branch to check for PHP 8 things. That run doesn't have any exclusions for WordPress polyfills, so the warning gets generated.

Complicating the matter further is that some of our packages are intended to be usable outside of a WordPress environment, so for them it would be good to still have warnings like this. Right now we only accidentally get that for them thanks to the "develop" branch run. Except for the ones manually overridden in `.github/files/phpcompatibility-dev-phpcs.xml`.

At some point I should really clean all that up. For now, adding to `.github/files/phpcompatibility-dev-phpcs.xml` is probably the most sensible thing.

opr: > For now, adding to .github/files/phpcompatibility-dev-phpcs.xml is probably the most sensible thing.

Is this something I should add in this PR? Happy to do so just need to confirm it's the right step.

anomiex: Yes, you can do that.

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-universal.php
```diff
@@ -602,7 +602,16 @@ public static function get_cart_checkout_info() {
 		$transient_name = 'jetpack_woocommerce_analytics_cart_checkout_info_cache';
 
 		$info = get_transient( $transient_name );
-		if ( false === $info ) {
+
+		// Return cached data early to prevent additional processing, the transient lasts for 1 day.
+		if ( false && false !== $info ) {
```
ralucaStan: The parser will never go further than the 1st false in this condition. I think you might have a typo.

opr: Yes, leftover from debugging. Thanks for spotting I will update.

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-universal.php
```diff
@@ -625,10 +634,49 @@ public static function get_cart_checkout_info() {
...
+			$checkout_template = get_block_template( $checkout_template_id );
+		}
+
+		// Update the info transient with data we got from the templates, if the site isn't using WC Blocks we
+		// won't be doing this so no concern about overwriting.
+		$info = array(
+			'cart_page_contains_cart_block'             => str_contains( $cart_template->content, '<!-- wp:woocommerce/cart' ),
+			'cart_page_contains_cart_shortcode'         => str_contains( $cart_template->content, '[woocommerce_cart]' ),
+			'checkout_page_contains_checkout_block'     => str_contains( $checkout_template->content, '<!-- wp:woocommerce/checkout' ),
+			'checkout_page_contains_checkout_shortcode' => str_contains( $checkout_template->content, '[woocommerce_checkout]' ),
```
senadir: This should return a 1/0 instead of true/false, returning a boolean causes a no value for false instead of 0.

```suggestion
			'cart_page_contains_cart_block'             => str_contains( $cart_template->content, '<!-- wp:woocommerce/cart' ) ? 1 : 0,
			'cart_page_contains_cart_shortcode'         => str_contains( $cart_template->content, '[woocommerce_cart]' ) ? 1 : 0,
			'checkout_page_contains_checkout_block'     => str_contains( $checkout_template->content, '<!-- wp:woocommerce/checkout' ) ? 1 : 0,
			'checkout_page_contains_checkout_shortcode' => str_contains( $checkout_template->content, '[woocommerce_checkout]' ) ? 1 : 0,
```

senadir (dismissed):
This is testing fine with me, left a single comment, if you're fine with it Thomas, we can merge.

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-universal.php
senadir: This matches the behavior of `post_contains_text`

opr: Yes please go ahead!

senadir (approved):
Reapproving after rebase.

PR #32716, opened 2023-08-28
CRM: Automation - Update the data type to avoid data inconsistencies using entities

cleacos:
Fixes: https://github.com/Automattic/zero-bs-crm/issues/3291

## Proposed changes:

This is a proof of concept/proposal to use entities as a way to manage well-known data, unified in one place. The current DAL implementation handles many data formats when we are going to add or update data, or when we retrieve the data from the database, we get different structures. 

For example, for the contact entity:

- `tidy_contact`, used by the Data_Type_Contact, expects data from the database, with the typical prefix `zbsc_`, like this `$contact->zbsc_status`. 
- When a contact is added/updated the expected format is different, is like:
   - `$contact = [ 'id' => -1, 'owner' => -1, 'data' => array( 'status' => 'lead') );`
   - And the old data has the same format returned from `tidy_contact` method (`get_contact` use it to tidy up)
- `tidy_contact` expects data from the database, so we cannot use it for managing data that is not from the database.

So we are managing different data structures. This proposal PR (draft) introduces the first concept of entity, it just adds a Contact class that holds the Contact information along with two methods that transform the data. Probably those methods will be the same for the rest of the entities so we could extract them to a `trait` file or parent class, or even into a Factory layer to make the transformation. I'd keep it as simple as we can, and then in the future, we could move to better entities and DAL system.

That is the idea of this change. It doesn't require too much time because the entities only have the properties and the mapping, which is super easy to implement after this PR, it will require one day more to adapt the rest of the entities and remove the Data_Type if we don't need it. The Contact entity in this PR has more fields that we really need for now, but I left it here to see how it'd look for a future full implementation of the entity, the right way. If the simple implementation is okay, I'll remove those methods and fields not needed.

In this first approach I didn't remove Data_Type classes, but probably using entities we don't need this middle layer.

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/projects/plugins/crm/changelog/fix-crm-3291-data-inconsistencies-transformers-unify-data
+++ b/projects/plugins/crm/changelog/fix-crm-3291-data-inconsistencies-transformers-unify-data
@@ -1,0 +1,5 @@
+Significance: patch
+Type: changed
+Comment: Update the Data_Type system for using entities for unifying data with a common entity
+
+
--- a/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Contacts.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Contacts.php
@@ -3022,12 +3022,12 @@ class zbsDAL_contacts extends zbsDAL_ObjectLayer {
 									'prev_contact' => $previous_contact_obj,
                                     ));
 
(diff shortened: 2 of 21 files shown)


cleacos:
Closing this PR. The issue was closed, and the work on this PR will be moved to another PRs.

PR #32720, opened 2023-08-28
Link styles and scripts to block.json automatically

monsieur-z:
## Proposed changes:

`block.json` metadata files require at least an `editorScript` field to be valid. This field was manually added to each `block.json` file, but generating it during the build makes more sense. This will prevent forgetting to update the value when moving a block from one mode to another (e.g., beta to production).

This PR adds the `editorScript`, `editorView`, `viewScript`, or `style` fields to `block.json` files at build time.

### Other information:

## Jetpack product discussion

pedMtX-RS-p2

No.

## Testing instructions:

- Run `cd projects/plugins/jetpack`
- Rebuild the blocks by running `rm -rf _inc/blocks && pnpm build-extensions`
- Choose a couple of blocks, and notice that the values for `editorScript`, `editorView`, `viewScript`, or `style` in `block.json` are correct.

_Note: blocks are currently not registered through their `block.json` file. This PR shouldn't impact the behaviour of blocks._

### DIFF
--- a/projects/plugins/jetpack/changelog/update-block-assets-build
+++ b/projects/plugins/jetpack/changelog/update-block-assets-build
@@ -1,0 +1,4 @@
+Significance: minor
+Type: enhancement
+
+Link styles and scripts to block.json automatically
--- a/projects/plugins/jetpack/tools/webpack.config.extensions.js
+++ b/projects/plugins/jetpack/tools/webpack.config.extensions.js
@@ -199,6 +199,48 @@ module.exports = [
 						to: '[path][name][ext]',
 						context: path.join( __dirname, '../extensions/blocks' ),
 						noErrorOnMissing: true,
+						// Automatically link scripts and styles
+						transform( content, absoluteFrom ) {
+							let metadata = {};
+
+							try {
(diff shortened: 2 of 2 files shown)


jeherve (approved):
This looks good and tests well for me, nice work.

It had me wondering about how RTL stylesheets are handled in WordPress blocks, but I suppose that's something that may be handled in core itself, not in each plugin's `block.json`.

monsieur-z:
> This looks good and tests well for me, nice work.
> 
> It had me wondering about how RTL stylesheets are handled in WordPress blocks, but I suppose that's something that may be handled in core itself, not in each plugin's `block.json`.

A `rtl.css` version is generated by our build, but I don't know if it's Jetpack or core that actually enqueues that version when it's needed. `block.json` files don't seem to allow specific stylesheets for internationalization.

My take on this is to write RTL-compatible CSS when possible, like https://github.com/Automattic/jetpack/pull/32698/files#diff-6aa20748b5497ded217b7f39a5ec124fa873246c41fca5a3c2559c4c7c4d4012

PR #32724, opened 2023-08-29
Subscriptions: Tweak subscribe modal text/spacing

edanzer:
## Proposed changes:

- Adds a a bit more vertical spacing between some elements in subscribe modal. 
- Updates the paragraph text to be shorter. There were several considerations here. 
   - First, the current text suggests that a user needs to subscribe to gain access to content. This is not the case with the subscribe modal, which does not block content. So I've updated the wording to be more reflective of that fact. 
   - Second, the current text was frequently wrapping just 1-2 words as orphans on the second line. I shorten the text to avoid line wrapping. I considered adjusting left/right padding, but (a) the impact varies by theme and (b) there's no way in the site editor to have different padding on a paragraph between desktop and mobile. Any padding that was enough for desktop was too much for mobile. And we didn't want to use custom css, which would be hard to edit.
   
**Before (on Lettre)**
lettre-before

**After (Lettre)**
lettre

**After (Lettre - Mobile)**
lettre-mobile

**After (Twenty Twenty-Three)**

**After (Twenty Twenty-Two)**

**After (Astra)**
astra

### Other information:

No.

## Testing instructions:

1) Checkout this branch in your local jetpack dev environment.
2) Add add_filter( 'jetpack_subscriptions_modal_enabled', '__return_true', 20 ); to tools/dockers/mu-plugins/0-sandbox.php.
3) Go to Jetpack > Settings > Newsletter and enable the Enable subscriber popup setting.
4) In a non-logged-in browser or private tab, open a single post on your test site, and scroll. The modal should open. confirm the new text looks like above and spacing looks reasonable.

### DIFF
--- a/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
+++ b/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
@@ -132,7 +132,7 @@ class Jetpack_Subscribe_Modal {
 		// translators: %s is the name of the site.
 		$discover_more_from = sprintf( __( 'Discover more from %s', 'jetpack' ), get_bloginfo( 'name' ) );
 		$continue_reading   = __( 'Continue Reading', 'jetpack' );
-		$subscribe_text     = __( 'Subscribe now to keep reading and gain access to the full archive.', 'jetpack' );
+		$subscribe_text     = __( 'Subscribe now to keep reading and get access to the full archive.', 'jetpack' );
 
 		return <<<HTML
 	<!-- wp:group {"style":{"spacing":{"padding":{"top":"var:preset|spacing|70","bottom":"var:preset|spacing|70","left":"var:preset|spacing|70","right":"var:preset|spacing|70"},"margin":{"top":"0","bottom":"0"}},"border":{"color":"#dddddd","width":"1px"}},"layout":{"type":"constrained","contentSize":"450px"}} -->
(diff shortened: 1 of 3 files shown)


projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.css
```diff
@@ -34,3 +34,13 @@
...
+
+/*
+ * These text-wrap properties still have limited browser
+ * support, but based on feedback still adding them for when
+ * they are supported.
+ */
+.jetpack-subscribe-modal__modal-content p {
+	text-wrap: balance;
+	text-wrap: pretty;
+}
```
edanzer, line 46: These text-wrap properties are experimental and have limited browser support. They will cause no harm when support is lacking, but could improve the style of the modal when support exists. Balance has more support than pretty, which is why I added both.

Slack convo: p1693271532532469/1693269433.261429-slack-C052XEUUBL4

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -132,14 +132,14 @@ public function get_subscribe_template_content() {
 		// translators: %s is the name of the site.
 		$discover_more_from = sprintf( __( 'Discover more from %s', 'jetpack' ), get_bloginfo( 'name' ) );
 		$continue_reading   = __( 'Continue Reading', 'jetpack' );
-		$subscribe_text     = __( 'Subscribe to the newsletter to keep reading and get access to the full archive.', 'jetpack' );
```
lezama, line 135: A note that we are losing the "get access to the full archive" incentive.

lezama (dismissed):

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
edanzer: I decided to kind-of revert the text change. I just cut some text to achieve my goal of keeping it to one line, rather than changing the text in a fundamental way. 

Looks like this now (in English with default Lettre theme):
modal-upate

For the record, I still think we should change this. The current text is still misleading - it is currently more what you'd put on a paywall or on a post that is subscribers-only.

But that's a separate conversation.

lezama (approved):

PR #32727, opened 2023-08-29
Boost: Update critical CSS URL parameter

haqadn:
## Proposed changes:

Update critical CSS generation parameter from `"1"` to the current timestamp to avoid cached redirects. Even after #32627 is merged, users may still find their critical CSS to be broken because of cached 301 redirects. This PR works around that.

### Other information:

## Testing instructions:

* Generate critical CSS manually and make sure it works
* Use an old version of Boost with one of the SEO plugins to reproduce the issue fixed by #32627, upgrade boost to this PR and make sure the critical CSS generation goes smoothly.


pyronaur (approved):
LGTM

PR #32738, opened 2023-08-29
CRM: Allow admin users to accept quotes

cleacos:
Fixes https://github.com/Automattic/zero-bs-crm/issues/2862

## Proposed changes:

This PR fixes this issue that doesn't allow admin users to accept quotes if it were necessary.

### Other information:

## Jetpack product discussion
N/A

N/A

## Testing instructions:
- Check the code
- Create a quote for an admin user
- Try to accept it by clicking on `Share the Link or [preview]` below the quote, after save it, it will send you to the Client Portal and you will see the Quote, click on accept:
   - In trunk, it will return an error.
   - In this branch, the quote will be accepted.

### DIFF
--- a/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
@@ -1104,8 +1104,7 @@ function ZeroBSCRM_accept_quote() {
 
 		// validate that this has been posted by the contact associated with the quote
 		if ( ! $uinfo->ID
-			|| zeroBS_getCustomerIDWithEmail( $uinfo->user_email ) !== zeroBSCRM_quote_getContactAssigned( $quoteID )
-			|| zeroBSCRM_permsQuotes()
+			|| zeroBS_getCustomerIDWithEmail( $uinfo->user_email ) !== zeroBSCRM_quote_getContactAssigned( $quoteID ) // phpcs:ignore
 		) {
 			zeroBSCRM_sendJSONError( array( 'access' => 1 ), 403 );
 		}
--- a/projects/plugins/crm/changelog/fix-crm-2862-allow-to-accept-quotes-by-admin
+++ b/projects/plugins/crm/changelog/fix-crm-2862-allow-to-accept-quotes-by-admin
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Quotes could be accepted by admin users
--- a/projects/plugins/crm/changelog/fix-crm-2862-allow-to-accept-quotes-by-admin
+++ b/projects/plugins/crm/changelog/fix-crm-2862-allow-to-accept-quotes-by-admin
@@ -1,4 +1,4 @@
 Significance: patch
 Type: fixed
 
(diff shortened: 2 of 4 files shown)


gogdzl:
I'm still getting this message:

image

But the quote gets accepted (both Admin quotes showed the message above, but got accepted):

image

cleacos:
@gogdzl I tried to replicate it, but I always get the expected response:

image

gogdzl:
Fixed the problem. Debugging with different admins I could find the error:

telegram-cloud-photo-size-1-5177181169924549604-y

It was caused by a notice being fired during the ajax call.

Fixed it in https://github.com/Automattic/jetpack/pull/32738/commits/92c99a667084ec906f38a2517e138dee8af8e34a


Not it runs with no errors:

telegram-cloud-photo-size-1-5177181169924549605-y

projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
```diff
@@ -1104,8 +1104,7 @@ function ZeroBSCRM_accept_quote() {
 
 		// validate that this has been posted by the contact associated with the quote
 		if ( ! $uinfo->ID
-			|| zeroBS_getCustomerIDWithEmail( $uinfo->user_email ) !== zeroBSCRM_quote_getContactAssigned( $quoteID )
-			|| zeroBSCRM_permsQuotes()
+			|| zeroBS_getCustomerIDWithEmail( $uinfo->user_email ) !== zeroBSCRM_quote_getContactAssigned( $quoteID ) // phpcs:ignore
```
tbradsha, line 1107: I believe technically this second condition could result in `false !== false`?

I don't see any real security risk, but imagine I'm a WP subscriber with no CRM access. I'd be logged in (so have a user ID set), but I have a no email in CRM, so if a quote is not assigned to a contact, in theory I can accept it.

tbradsha:
Rebased.

cleacos:
> Fixed the problem. Debugging with different admins I could find the error:

Okay, I see, it seems it was another issue. Thanks for fixing it.

projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
cleacos: Really you cannot share a quote until you have a contact associated with it.

cleacos: But we can add an extra condition (paranoid mode on):

`if ( ! $uinfo->ID || zeroBSCRM_quote_getContactAssigned( $quoteID ) === false || zeroBS_getCustomerIDWithEmail( $uinfo->user_email ) !== zeroBSCRM_quote_getContactAssigned( $quoteID )`

tbradsha (approved):

projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
tbradsha: I'm probably overly paranoid. :^)

PR #32741, opened 2023-08-29
AI Chat block: First pass at chatting with site functionality

dereksmart:
## Proposed changes:

This updates the AI Chat block to: 
- Chat with site(bot)
- Basic panel setting
- Registers endpoint for asking about the site
- Registers endpoint for sending feedback about the response

The feature relies on Jetpack Search and a site connection. 

The AI Chat block is in experimental mode and is not displayed to any users currently without a constant set. The wpcom endpoints are limited to a12s. 

This PR does not account for wpcom sites. 

## Jetpack product discussion

p1HpG7-oxA-p2

No

## Testing instructions:

It should work out of the box with
- A site owned by an a11n (or sandboxed connection for any site)
- Jetpack Search purchased (even the free tier)
- `define( 'JETPACK_BLOCKS_VARIATION', 'beta' );`

Try chatting with the site in the editor. 
Try chatting with the site in the front end. 
Try to modify the "ask" button. 
Try to send feedback. 

You can also test some filterable customizations, such as the answer prompt: 
```
add_filter( 'jetpack_ai_chat_answer_prompt', function() {
    return 'Talk like a cowboy';
} );
```


artpi:
So what we want for this UX is to copy Team51 work and reuse the same block they are using and converge.

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-ai.php
```diff
@@ -117,6 +119,137 @@ public function register_basic_routes() {
 				),
 			)
 		);
+
+		register_rest_route(
```
artpi, line 116: one thing with proxying this endpoint is that AFAIK we support streaming on WPCOM end and I don't think every server running JP does, so this can break streaming

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-ai.php
```diff
@@ -117,6 +119,137 @@ public function register_basic_routes() {
...
+				'query'         => $question,
+				/**
+				 * Filter for an answer prompt override.
+				 * Example: "Talk like a cowboy."
+				 *
+				 * @param string $prompt_override The prompt override string.
+				 *
+				 * @since $$next-version$$
+				 */
+				'answer_prompt' => apply_filters( 'jetpack_ai_chat_answer_prompt', false ),
```
artpi, line 215: Are we calling this "AI Chat" or "AI Search" ?
AI Search seems a little more honest, because this is one-shot AND technically built on top of Jetpack Search

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-ai.php
```diff
@@ -117,6 +119,137 @@ public function register_basic_routes() {
...
+	public function request_chat_with_site( $request ) {
+		$question = $request->get_param( 'query' );
+		$blog_id  = \Jetpack_Options::get_option( 'id' );
+		$response = Client::wpcom_json_api_request_as_blog(
+			sprintf( '/sites/%d/jetpack-search/ai/search', $blog_id ) . '?force=wpcom',
+			2,
+			array(
+				'method'  => 'GET',
+				'headers' => array( 'content-type' => 'application/json' ),
+				'timeout' => 30,
```
artpi: probs too low

projects/plugins/jetpack/changelog/update-ai-chat-block
```diff
@@ -0,0 +1,4 @@
+Significance: minor
```
artpi, line 1: MAJOR!!

artpi (commented):
Left some random feedback :)

projects/plugins/jetpack/changelog/update-ai-chat-block
dereksmart: image

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-ai.php
dereksmart: Yeah, should probably max this out. I had to raise it from the default 10s when testing connected to sandbox.

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-ai.php
dereksmart: If all goes according to plan, eventually it won't be dependent on search

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-ai.php
dereksmart: Interesting. What are the server requirements for streaming to work?

artpi: The server has to support Server-Side Events

dereksmart:
:robot: AI summarized this PR :robot:
### What and why?
This PR updates the AI Chat block with new functionalities which include the ability to chat with the site(bot), adjust basic panel settings, register endpoint for asking about the site and register endpoint for sending feedback about the response. The feature relies on Jetpack Search and a site connection. 

The enhancement is still considered experimental and is not currently displayed to any users unless a constant is set. Please note that this PR does not apply to WordPress.com sites and the wpcom endpoints are limited to a12s. 

Related Jetpack product discussion can be found [here](p1HpG7-oxA-p2).

### Testing Steps
- Test the PR with a site owned by an a11n (or sandboxed connection for any site), Jetpack Search purchased (even the free tier), and `define( 'JETPACK_BLOCKS_VARIATION', 'beta' );`
- Try chatting with the site in the editor and in the front end.
- Try to modify the "ask" button and send feedback. 
- You can also test some filterable customizations, such as the answer prompt by adding the filter: 
```add_filter( 'jetpack_ai_chat_answer_prompt', function() { return 'Talk like a cowboy'; } );```

Please note that this PR does not change what data or activity we track or use.

<details>
    <summary>Evaluation of original PR description</summary>
    The original PR description contains all the crucial details about the changes the PR introduces and provides clear testing steps. However, the descriptions could be made more concise and to the point. Regularly updating the PR with the latest changes and status also helps keep everyone in the loop. 
</details>

---

This comment will be updated when relevant updates are made to the PR.
Please give :thumbsup: or :thumbsdown: to indicate if it is helpful.

artpi:
Wow, this is close!

I tested it on WPCOM with Jetpack downloader and so far 2 issues stood out:
Most importantly, the JS seems to be using wrong endpoint

Zrzut ekranu 2023-09-8 o 17 27 25

We need to fix that :D

On the Gutenberg side, I never deployed actual React to the frontend of the site, so I am not familiar yet with how those blocks work but we probably don't want it to work inside Gutenberg? Or do we? I'm kind of on the fence.


Zrzut ekranu 2023-09-8 o 17 27 51

We probably want more customization options, but this is all cosmetics :)

dereksmart:
> We need to fix that

For posterity of the thread (already talked in Slack) it was expected in this PR that WP API rest endpoints aren't registered or working in wpcom. https://github.com/Automattic/jetpack/pull/32741/files#diff-38c22c02a5a6b50d822b14683ebbb1d3bcdd8cd313cb0aec8130811d5f4ab06bR93

artpi:
Works in Jetpack!
Zrzut ekranu 2023-09-8 o 18 22 08

artpi (approved):
debug.log is empty, this is good to go IMO

PR #32744, opened 2023-08-29
Admin page: rely on initial state passed from PHP for theme data

jeherve:
Follow-up to #32722

## Proposed changes:

This is a bit of refactoring that came to my mind as I was looking at #32722. @edanzer Let me know what you think about it, this is just an idea :)
Since we already query for information about the theme when we first load the dashboard, maybe we can use that directly.

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

Follow the instructions in #32722.


edanzer (approved):
@jeherve - Thanks for the follow up! Code changes here look straightforward and good to me. And this test as expected.

PR #32748, opened 2023-08-29
WIP: Refactor/membership currencies from wpcom

chrisfromthelc:
Fixes #

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/projects/plugins/jetpack/extensions/store/membership-products/actions.js
+++ b/projects/plugins/jetpack/extensions/store/membership-products/actions.js
@@ -22,6 +22,11 @@ export const setConnectedAccountDefaultCurrency = connectedAccountDefaultCurrenc
 	connectedAccountDefaultCurrency,
 } );
 
+export const setConnectedAccountMinimumCurrency = connectedAccountMinimumCurrency => ( {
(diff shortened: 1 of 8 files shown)


projects/plugins/jetpack/extensions/store/membership-products/actions.js
```diff
@@ -49,8 +54,9 @@ export const saveProduct =
 			return;
 		}
 
+		const currencyMinimums = registry.select( STORE_NAME ).getConnectedAccountMinimumCurrency();
```
millerf, line 57: Yes, good.

chrisfromthelc:
Closing this PR since we won't be able to accomplish what was planned currently.

millerf:
We closed this PR as We need the list of currencies on the front-end for [this check](https://github.com/Automattic/jetpack/blob/abc4ef1e2722ca695c86dcb5b844f85f4323f999/projects/plugins/jetpack/extensions/blocks/donations/view.js#L84). And it would not be possible without an extra call to the backend which could end up extremely expensive in term of resources...

PR #32750, opened 2023-08-29
AI Extension: introduce Create with voice jetpack block

### LINKED ISSUE #32731: Create with Voice: introduce jetpack/create-with-voice block
Attributes:

* `src`: source audio file

retrofox:
This PR creates the first implementation of the `Create with voice` block:

image

image

image

image

### TS
Additionally, it moves TS files to the root of `extensions/` folder to increase its supports covering blocks (AI Assistant and this new one), plugins (AI Extension), etc...
Currently, we are duplicating files unnecessarily. 

Fixes #

## Proposed changes:

* AI Extension: add create-with-voice beta extension

### Other information:

## Testing instructions:

* Go to the block editor
* Create a Create with voice block instance
* Confirm the block is available only when `beta` extension is active.
* Take a look at the TS.  You can take a look how your IDE detect the globals, for instance:

image


projects/plugins/jetpack/extensions/blocks/create-with-voice/create-with-voice.php
```diff
@@ -0,0 +1,50 @@
+<?php
+/**
+ * "Create with voice" Block.
+ *
+ * @since 12.2
```
lhkowalski:
```suggestion
 * @since 12.5
```

lhkowalski (changes requested):
It's working nicely:

- The beta feature control is happening correctly (it's a beta block)
- I can insert the block on the editor:

Screenshot 2023-08-30 at 12 56 02

- I can see the placeholder:

Screenshot 2023-08-30 at 12 56 13

- BUT I can't select it to be able to remove it. The only way to remove it is by editing the code on the code editor. Is there a simple way to handle this? If not, we can merge as is and iterate later. If yes, we could start allowing it now :)

lhkowalski (approved):
Let's fix the "undeletable" bug on a follow-up. 🚀

PR #32567, opened 2023-08-17
Add permanent dismissal of launchpad save modal

daledupreez:
Fixes https://github.com/Automattic/wp-calypso/issues/76880

## Proposed changes:

* This PR updates the launchpad modal to be permanently dismissable by three distinct pieces:
   1.  Implement `wpcom_launchpad_is_fse_next_steps_modal_hidden()` and `wpcom_launchpad_set_fse_next_steps_modal_hidden()` functions for the launchpad, which store the flag as the `hide_fse_next_steps_modal` sub-option of `wpcom_launchpad_config`.
   2. Add `hide_fse_next_steps_modal` as a modifiable option to the `POST` launchpad API.
   3. Update the launchpad save modal to hide the modal when `wpcom_launchpad_is_fse_next_steps_modal_hidden()` returns true, and to correctly update the option via the REST API above

### Other information:

- [x] Have you written new tests for your changes, if applicable?
- [x] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [x] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

This is a WordPress.com-specific feature.

## Testing instructions:

* Apply both parts of this PR to your WordPress.com development sandbox as per the instructions in [this comment](https://github.com/Automattic/jetpack/pull/32567#issuecomment-1682981605)
* Create a new WordPress.com site with the Write intent and get to the fullscreen launchpad for your site
* Ensure that both `public-api.wordpress.com` and your site's `*.wordpress.com` subdomain are both pointing at your development sandbox
* From the fullscreen launchpad, click on the task to write a post
* Write some content and publish the post
* You should see the "Great progress" prompt appear (possibly after another prompt)
* Select the "Don't show again" checkbox
* Click on the "Back to Edit" button
* Refresh the page
* Open the developer tools, and open the Console tab
* Ensure you're running in the context of `post-new.php` or `post.php`
* Type `window.Jetpack_LaunchpadSaveModal`, and verify that the `hideFSENextStepsModal` value is present and set to `true`
* Go back to the full-screen launchpad
* Click on the task to create a new post
* Edit and then publish the new post
* Verify that you don't see the "Great progress" modal - this means the update went through from the "Back to Edit" button and took effect correctly on loading the post editor
* On the back end, delete the `wpcom_launchpad_config` option for this blog via `switch_to_blog( $your_blog_id ); delete_option( 'wpcom_launchpad_config' );`
[...]

PR #32751, opened 2023-08-29
Reduce api calls for launchpad modal

daledupreez:
Related to https://github.com/Automattic/jetpack/pull/32567

## Proposed changes:

* This PR adds some additional logic to the launchpad post publish modal to avoid making API calls when the setting hasn't actually changed. There were two situations where we were making API calls that we shouldn't:
   1. If the _Don't show this again_ checkbox was not checked, but the user clicked on a button or closed the modal, we'd send a redundant API call
   2. In some cases, we close the modal in favour of showing another modal, but then re-open our modal after that modal has been closed, which can also trigger an unnecessary API call. That modal juggling is handled via the following `useEffect()` code block: https://github.com/Automattic/jetpack/blob/abce9f2b0fda88238e65a2a2aca14c7b44758195/projects/plugins/jetpack/extensions/plugins/launchpad-save-modal/index.js#L128-L159

### Other information:

## Jetpack product discussion

This changes a WordPress.com specific modal, so no discussion was involved.

No changes to tracking or data.

## Testing instructions:

* Ensure that you have an unlaunched WordPress.com site that is still showing the fullscreen modal -- if you don't have one, you can create a new site with the "Write & Publish" intent via https://wordpress.com/start - make sure you pause when you get to the fullscreen launchpad
* Apply this change to your WordPress.com development sandbox
* Ensure that `public-api.wordpress.com` _and_ the `*.wordpress.com` subdomain for your site are both pointing at your development sandbox
* Click on a task to publish or write a new post, or navigate directly to `https://wordpress.com/post/:siteSlug`
* Open your browser's development tools, and ensure that they retain data across navigation events
* Write some content for the post
* Publish the post
* Verify that you see the "Great progress" modal (you may see another "Post published!" modal first)
* Verify that there isn't a `POST` call to the `launchpad` REST API
* Click on the "Back to edit" button
* Verify that there still isn't a `POST` call to the `launchpad` REST API
* Navigate to `https://wordpress.com/post/:siteSlug`
* Write some content in the post, and publish the post
* Verify that you see the "Great progress" modal (again, it may be preceded by the "Post published!" modal)
* Check the _Don't show this again_ checkbox
* Click on the "Next steps" button
* Verify that there is a `POST` API call to the `launchpad` REST API

The basic idea here is that we want to limit calls to those that actually change the value.

### DIFF
--- a/projects/plugins/jetpack/changelog/reduce-api-calls-for-launchpad-modal
+++ b/projects/plugins/jetpack/changelog/reduce-api-calls-for-launchpad-modal
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
+Launchpad modal: reduce API calls for modal dismiss flag
--- a/projects/plugins/jetpack/extensions/plugins/launchpad-save-modal/index.js
+++ b/projects/plugins/jetpack/extensions/plugins/launchpad-save-modal/index.js
@@ -69,13 +69,16 @@ export const settings = {
 			siteIntentOption,
 		} = window?.Jetpack_LaunchpadSaveModal || {};
 
+		const hideFSENextStepsModalBool = !! hideFSENextStepsModal;
+
 		const [ isModalOpen, setIsModalOpen ] = useState( false );
-		const [ dontShowAgain, setDontShowAgain ] = useState( !! hideFSENextStepsModal );
-		const [ isChecked, setIsChecked ] = useState( !! hideFSENextStepsModal );
+		const [ dontShowAgain, setDontShowAgain ] = useState( hideFSENextStepsModalBool );
+		const [ isChecked, setIsChecked ] = useState( hideFSENextStepsModalBool );
 
 		const isInsideSiteEditor = document.getElementById( 'site-editor' ) !== null;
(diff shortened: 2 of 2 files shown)


markbiek:
I'm having difficulties with this, though I suspect it's because I'm not setting up the test environment properly.

I ran `bin/jetpack-downloader test jetpack try/reduce-api-calls-for-launchpad-modal` which has my sandbox looking like this:

CleanShot 2023-08-29 at 16 16 13@2x

I'm assuming there's a build process that's happening to compile that modified JS.

I've published my first post and I see this modal
CleanShot 2023-08-29 at 16 20 00@2x

but I never see the "Great progress" modal.

I do see some `No route was found matching the URL and request method` errors in the console but I can't tell which route they're failing on so that may be unrelated.

daledupreez:
@markbiek, I've just re-tested with a newly created site with the Write & Publish intent, and I am seeing the correct behaviour when I click on the "X" to dismiss the "Your post is published" modal. (I am not permanently dismissing that modal.) Are you correctly sandboxing your new `site.wordpress.com` domain? And is the launchpad option set to `full` for the site?

markbiek (approved):
Working as expected!

PR #32765, opened 2023-08-30
Paywall Block: Improve toolbar

lezama:
#### Before

Screenshot 2023-08-30 at 12 31 12

#### After

Screenshot 2023-08-31 at 13 14 44

Screenshot 2023-08-31 at 13 14 52

### Other information:

## Testing instructions:

* Add a Paywall block and play with the toolbar.

### DIFF
--- a/projects/plugins/jetpack/changelog/update-paywall-block-toolbar
+++ b/projects/plugins/jetpack/changelog/update-paywall-block-toolbar
@@ -1,0 +1,4 @@
+Significance: minor
+Type: enhancement
+
+Paywall Block: Improved toolbar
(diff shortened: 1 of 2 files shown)


edanzer (approved):
Tested this fairly exhaustively on both Jetpack self hosted and on WordPress.com. I tried:
  - Selecting the paid option from the toolbar with no stripe connection and no plan (get modal)
  - Selecting the paid option from the toolbar with stripe but no plan (get modal)
  - Selecting the paid option from the toolbar with stripe and plan (works)

One small thing is I think we can improve the styling of the modal - at the least make it less wide maybe. But fine to handle that as a follow up.

PR #32767, opened 2023-08-30
AI Client: introduce useMediaRecording() hook

### LINKED ISSUE #32661: Create with voice: research approach/external libraries to record audio
We could consider using an external library or maybe create our implementation based on [MediaDevices](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices) and [MediaRecorder](https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder).

[RecordRTC](https://github.com/muaz-khan/RecordRTC) seems to be interesting.

retrofox:
This PR introduces a custom hook to handle media recording from the client.

Part of #32661

## Proposed changes:

* AI Client: introduce useMediaRecording() hook

## Testing instructions:

The hook is not being used in any place. A follow-up PR is coming on.
There, we can test the implementation deeply. Inspecting the code visually, and taking a look at the [readme file](https://github.com/Automattic/jetpack/blob/797f455b3913de58c15e3d7ecd114cc1a86d5742/projects/js-packages/ai-client/src/hooks/use-media-recording/Readme.md) would be enough for now.

### DIFF
--- a/projects/js-packages/ai-client/src/hooks/use-media-recording/Readme.md
+++ b/projects/js-packages/ai-client/src/hooks/use-media-recording/Readme.md
@@ -1,0 +1,48 @@
+# `useMediaRecording` Custom React Hook
+
+## Description
+
+`useMediaRecording` is a custom React hook for handling media recording functionalities in a React application. It provides an easy way to start, pause, resume, and stop media recording, as well as to track the current recording state.
+
+Based on [MediaRecorder](https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder) API
+
+## API
+
+The hook returns an object with the following properties and methods:
+
+- `start: () => void`: Start the media recording
+- `pause: () => void`: Pause the current media recording
+- `resume: () => void`: Resume a paused recording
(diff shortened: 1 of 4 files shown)


projects/js-packages/ai-client/src/hooks/use-media-recording/Readme.md
```diff
@@ -0,0 +1,48 @@
...
+import useMediaRecording from './useMediaRecording';
+
+const MediaRecorderComponent = () => {
+	const { start, pause, resume, stop, state } = useMediaRecording();
+
+	return (
+		<div>
+			<h1>Media Recorder</h1>
+			<p>Current State: { state }</p>
+			<button onClick={ start } disabled={ state === 'recording' }>
```
dhasilva, line 33:
```suggestion
			<button onClick={ start } disabled={ state !== 'inactive' }>
```
If it is either recording or paused the `Start` should be disabled, right? Or else we could start again after pausing.

dhasilva (approved):
LGTM! Just left a minor comment about the example.

PR #32768, opened 2023-08-30
Remove unused variable

millerf:
Remove a var that was always false...

## Proposed changes:

* `shouldUpgrade` was always false here...


chrisfromthelc (approved):
LGTM

PR #32762, opened 2023-08-30
General: update WordPress requirements to WP 6.2

jeherve:
See #31638

## Proposed changes:

Now that WordPress 6.3 and folks have had some time to update to the most recent version of WordPress, let's start requiring WP 6.2 for everyone wanting to use our different plugins.

This also reverts #32366.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

* #31638

## Does this pull request change what data or activity we track or use?

* No

## Testing instructions:

* Is CI happy?

PR #32772, opened 2023-08-30
General: remove WP 6.1 backwards compatibility checks

jeherve:
## Proposed changes:

This removes backwards compatibility checks for WordPress 6.1, since our plugins will now require WP 6.2 (see #32762).

### Other information:

## Jetpack product discussion

* #31638

* No

## Testing instructions:

1. Is CI happy?
2. Check the different apps impacted by this change.


projects/packages/backup/src/js/index.js
```diff
@@ -11,24 +11,19 @@ register( store );
...
+function renderDash() {
 	const container = document.getElementById( 'jetpack-backup-root' );
 
 	if ( null === container ) {
 		return;
 	}
 
-	// @todo: Remove fallback when we drop support for WP 6.1
-	const component = (
+	render(
```
anomiex: Looks like you got all these backwards. It's the `if` case (using `createRoot().render()`) that needs to be kept, not the `else` case.

See https://github.com/Automattic/jetpack/actions/runs/6029259977/job/16358366809?pr=32772#step:9:695 for the problem this causes.

anomiex:
Looks like `projects/plugins/jetpack/tests/php/_inc/lib/class-tweetstorm-requests-transport-override.php` still has some references to old Requests library classes (`Requests_Transport`, and `Requests_Response` and `Requests_Exception` in a docblock).

Other stuff left from the comment in the other PR:

* https://github.com/Automattic/jetpack/blob/a97e707dbc159f0774bd832ab379dd3e97c946f4/projects/packages/waf/tests/php/integration/bootstrap.php#L22
* Probably https://github.com/Automattic/jetpack/blob/a97e707dbc159f0774bd832ab379dd3e97c946f4/tools/e2e-commons/pages/wp-admin/block-editor.js#L23
* https://github.com/Automattic/jetpack/blob/a97e707dbc159f0774bd832ab379dd3e97c946f4/projects/packages/videopress/src/client/block-editor/blocks/video/edit.tsx#L586-L591

Also these comments seem likely to need a version bump, it looks like it didn't make it into 6.2 (or 6.3).
* https://github.com/Automattic/jetpack/blob/a97e707dbc159f0774bd832ab379dd3e97c946f4/projects/packages/google-fonts-provider/src/class-utils.php#L48
* https://github.com/Automattic/jetpack/blob/a97e707dbc159f0774bd832ab379dd3e97c946f4/projects/plugins/jetpack/modules/google-fonts.php#L116
* https://github.com/Automattic/jetpack/blob/a97e707dbc159f0774bd832ab379dd3e97c946f4/projects/plugins/jetpack/modules/google-fonts.php#L127

anomiex:
Did a search for `function_exists` being called on `wp_`-prefixed functions. Not entirely sure these can all be removed, since often enough WordPress doesn't load files in some entry points (and sometimes wpcom lacks functions loaded even if WP itself would have them).

I wonder if this can be cleaned up, looks like `wp_get_global_styles()` was added in WordPress 5.9.
https://github.com/Automattic/jetpack/blob/a97e707dbc159f0774bd832ab379dd3e97c946f4/projects/packages/google-fonts-provider/src/introspectors/class-global-styles.php#L43-L48

Same for `wp_is_block_theme()`.
https://github.com/Automattic/jetpack/blob/a97e707dbc159f0774bd832ab379dd3e97c946f4/projects/packages/search/src/initializers/class-initializer.php#L156
https://github.com/Automattic/jetpack/blob/a97e707dbc159f0774bd832ab379dd3e97c946f4/projects/plugins/jetpack/class.jetpack-admin.php#L135
https://github.com/Automattic/jetpack/blob/a97e707dbc159f0774bd832ab379dd3e97c946f4/projects/plugins/jetpack/modules/masterbar/admin-menu/class-jetpack-admin-menu.php#L212
https://github.com/Automattic/jetpack/blob/a97e707dbc159f0774bd832ab379dd3e97c946f4/projects/plugins/jetpack/modules/masterbar/admin-menu/class-wpcom-admin-menu.php#L326

This too, unless the package wants to continue supporting WP older than 5.5.0
https://github.com/Automattic/jetpack/blob/a97e707dbc159f0774bd832ab379dd3e97c946f4/projects/packages/status/src/class-status.php#L172-L174

Here's one back to WP 4.9:
https://github.com/Automattic/jetpack/blob/a97e707dbc159f0774bd832ab379dd3e97c946f4/projects/plugins/beta/src/class-hooks.php#L443

Both of these were added in 6.1:
https://github.com/Automattic/jetpack/blob/a97e707dbc159f0774bd832ab379dd3e97c946f4/projects/plugins/boost/app/lib/class-storage-post-type.php#L208-L209

One from 5.2:
https://github.com/Automattic/jetpack/blob/a97e707dbc159f0774bd832ab379dd3e97c946f4/projects/plugins/jetpack/modules/masterbar/masterbar/class-masterbar.php#L396

5.5:
https://github.com/Automattic/jetpack/blob/a97e707dbc159f0774bd832ab379dd3e97c946f4/projects/plugins/jetpack/modules/shortcodes/recipe.php#L639
https://github.com/Automattic/jetpack/blob/a97e707dbc159f0774bd832ab379dd3e97c946f4/projects/plugins/jetpack/tests/php/modules/shortcodes/test-class.recipe.php#L187 (and three more later in that file)

5.8: https://github.com/Automattic/jetpack/blob/a97e707dbc159f0774bd832ab379dd3e97c946f4/projects/plugins/jetpack/modules/widget-visibility/widget-conditions.php#L69

projects/plugins/jetpack/_inc/lib/admin-pages/class-jetpack-redux-state-helper.php
```diff
@@ -205,10 +205,8 @@ public static function get_initial_state() {
 				'support'      => array(
 					'infinite-scroll' => current_theme_supports( 'infinite-scroll' ) || in_array( $current_theme->get_stylesheet(), $inf_scr_support_themes, true ),
 					'widgets'         => current_theme_supports( 'widgets' ),
-					'webfonts'        => (
-						// @todo Remove conditional once we drop support for WordPress 6.1
-						function_exists( 'wp_theme_has_theme_json' ) ? wp_theme_has_theme_json() : WP_Theme_JSON_Resolver::theme_has_support()
-					) && function_exists( 'wp_register_webfont_provider' ) && function_exists( 'wp_register_webfonts' ),
+					'webfonts'        => wp_theme_has_theme_json()
+						&& ( function_exists( 'wp_register_webfont_provider' ) || function_exists( 'wp_register_webfonts' ) ),
```
jeherve, line 209: `wp_register_webfonts` and `wp_register_webfont_provider` are not yet in Core.

jeherve:
I took care of the comments you left above, thanks for the feedback!

> Did a search for function_exists being called on wp_-prefixed functions. Not entirely sure these can all be removed, since often enough WordPress doesn't load files in some entry points (and sometimes wpcom lacks functions loaded even if WP itself would have them).

I like that idea, that's a clever way to find things that can maybe be removed. I took a look and removed those that seemed safe to remove. 👍

anomiex (dismissed):
Let's go for it.

jeherve:
This will need another review after bumping package versions. Thank you!

anomiex (approved):

PR #32774, opened 2023-08-30
Added WooCommerce orders table to checksum logic.

zinigor:
Depends on #32530 

Adds a definition of the new WooCommerce HPOS table to the Jetpack Sync checksum functionality. This will make sure the histograms returned by the Sync endpoints would allow checksums to function properly.

## Proposed changes:
* Adds a new WooCommerce HPOS table definition.

### Other information:

## Jetpack product discussion

p9dueE-7MN-p2

No

## Testing instructions:
* TBA


projects/packages/sync/src/replicastore/class-table-checksum.php
```diff
@@ -304,6 +304,14 @@ protected function get_default_tables() {
 				'table_join_field'          => 'order_item_id',
 				'is_table_enabled_callback' => array( $this, 'enable_woocommerce_tables' ),
 			),
+			'wc_orders'                  => array(
+				'table'                     => "{$wpdb->prefix}wc_orders",
+				'range_field'               => 'id',
+				'key_fields'                => array( 'id', 'parent_order_id' ),
+				'checksum_text_fields'      => array( 'type', 'billing_email', 'payment_method_title', 'transaction_id', 'ip_address', 'customer_note' ),
```
mdbitz: Checksums should be lightweight, are all these columns critical? Without looking into all the HPOS data I would lean towards ip_address and customer_note being not included.

projects/packages/sync/src/replicastore/class-table-checksum.php
```diff
@@ -304,6 +304,30 @@ protected function get_default_tables() {
 				'table_join_field'          => 'order_item_id',
 				'is_table_enabled_callback' => array( $this, 'enable_woocommerce_tables' ),
 			),
+			'wc_orders'                  => array(
+				'table'                     => "{$wpdb->prefix}wc_orders",
+				'range_field'               => 'id',
+				'key_fields'                => array( 'id', 'parent_order_id' ),
```
bisko: This should be fine to be just `id`. The orders should be unique by `id` if I recall correctly. No need to make a compound key if the entries are going to be 1:1 between the Jetpack site and WPCOM. Especially for non-meta tables.

bisko (approved):

PR #32776, opened 2023-08-30
CRM: Automations conditions for contact tags

gogdzl:
## Proposed changes:

* This PR adds conditions for contact tags:
  * has_tag
  * tag_added
  * tag_removed

### Other information:

## Jetpack product discussion
https://github.com/Automattic/zero-bs-crm/issues/3236
https://github.com/Automattic/zero-bs-crm/issues/3237
https://github.com/Automattic/zero-bs-crm/issues/3247

No.

## Testing instructions:

### DIFF
--- a/projects/plugins/crm/src/automation/commons/conditions/contacts/class-contact-tag.php
+++ b/projects/plugins/crm/src/automation/commons/conditions/contacts/class-contact-tag.php
@@ -1,0 +1,209 @@
+<?php
+/**
+ * Jetpack CRM Automation Contact_Tag condition.
+ *
+ * @package automattic/jetpack-crm
+ */
+
+namespace Automattic\Jetpack\CRM\Automation\Conditions;
+
+use Automattic\Jetpack\CRM\Automation\Attribute_Definition;
+use Automattic\Jetpack\CRM\Automation\Automation_Exception;
+use Automattic\Jetpack\CRM\Automation\Base_Condition;
+use Automattic\Jetpack\CRM\Automation\Data_Types\Data_Type_Contact;
+
+/**
+ * Contact_Tag condition class.
+ *
+ * @since $$next-version$$
+ */
+class Contact_Tag extends Base_Condition {
+
+	/**
+	 * Contact_Tag constructor.
+	 *
+	 * @since $$next-version$$
+	 *
+	 * @param array $step_data The step data.
+	 */
+	public function __construct( array $step_data ) {
+		parent::__construct( $step_data );
+
+		$this->valid_operators = array(
+			'tag_added'   => __( 'Tag was added', 'zero-bs-crm' ),
+			'tag_removed' => __( 'Tag was removed', 'zero-bs-crm' ),
+			'has_tag'     => __( 'Has tag', 'zero-bs-crm' ),
+		);
+
+		$this->set_attribute_definitions(
+			array(
+				new Attribute_Definition( 'operator', __( 'Operator', 'zero-bs-crm' ), __( 'Determines how the field is compared to the specified value.', 'zero-bs-crm' ), Attribute_Definition::SELECT, $this->valid_operators ),
+				new Attribute_Definition( 'tag', __( 'Tag', 'zero-bs-crm' ), __( 'Contact Tag to compare with.', 'zero-bs-crm' ), Attribute_Definition::TEXT ),
+			)
+		);
+	}
+
+	/**
+	 * Checks if a given tag name exists in the tags array.
+	 *
+	 * @param array  $data     The data array containing the 'tags' key.
+	 * @param string $tag_name The name of the tag to check for.
+	 *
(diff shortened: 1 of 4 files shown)


projects/plugins/crm/src/automation/commons/conditions/contacts/class-contact-tag.php
```diff
@@ -0,0 +1,209 @@
...
+	 *
+	 * @param mixed  $data Data passed from the trigger.
+	 * @param ?mixed $previous_data (Optional) The data before being changed.
+	 * @return void
+	 *
+	 * @throws Automation_Exception If an invalid operator is encountered.
+	 */
+	public function execute( $data, $previous_data = null ) { // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
+		if ( $previous_data === null ) {
+			$this->logger->log( 'Invalid previous contact tag data' );
```
cleacos: I'm not sure if this is okay for all the cases, for instance, `has_tag` condition should work with new contacts. For example, when a new Contact is created, we don't have previous data, it will be null, but we can check `has_tag` condition.

On the other hand, for `tag_added` or `tag_removed` we need information about the previous state.

cleacos: In summary, that check would make sense only if the condition is for a tag added or removed.

projects/plugins/crm/src/automation/commons/conditions/contacts/class-contact-tag.php
```diff
@@ -0,0 +1,209 @@
...
+
+	/**
+	 * Get the category of the contact tag condition.
+	 *
+	 * @since $$next-version$$
+	 *
+	 * @return string The category 'contact'.
+	 */
+	public static function get_category(): string {
+		return 'contact';
```
cleacos: For now, we get the category label, translated, and capitalized. Until we implement the Category class.
```suggestion
		return __( 'Contact', 'zero-bs-crm' );
```

cleacos (changes requested):
- I checked the code. It seems good. 
- I left a comment about one check performed in the code we have to update. 
- Tests run OK.

projects/plugins/crm/tests/php/automation/contacts/class-contact-condition-test.php
```diff
@@ -148,4 +162,112 @@ public function test_transitional_status() {
...
+		$previous_contact = $contact_data;
+
+		// Testing when the condition has been not been met because the contact does not have said tag.
+		$contact_tag_condition->execute( $contact_data, $previous_contact );
+		$this->assertFalse( $contact_tag_condition->condition_met() );
+
+		// Testing when the condition has been met.
+		$contact_data['tags'][] = array(
+			'id'          => 3,
+			'objtype'     => 1,
```
tbradsha: Do we have to use magic numbers here and below?

tbradsha:
Is there a reason the condition is contact-specific? Basically any object can have tags, so to me it makes sense to have a generic class that handles tag-related conditions...

projects/plugins/crm/src/automation/commons/conditions/contacts/class-contact-tag.php
gogdzl: Addressed in https://github.com/Automattic/jetpack/pull/32776

projects/plugins/crm/src/automation/commons/conditions/contacts/class-contact-tag.php
gogdzl: Addressed in 236c3ae

projects/plugins/crm/tests/php/automation/contacts/class-contact-condition-test.php
gogdzl: Removed in 446e3fe.

cleacos:
Basically, the step will handle a specific object type or class. For example, a Contact_Tag condition will handle the Contact type and check for the tag. To generalize it and avoid code duplication, we can use traits to manage tags, but each Step will process a well-known data type or class for them.

projects/plugins/crm/src/automation/commons/conditions/contacts/class-contact-tag.php
cleacos: I suppose it is this commit: https://github.com/Automattic/jetpack/pull/32776/commits/7f5b2250b2edf0d9edf6719c9a14cc43f07f69b2

cleacos: 👍  Nice extraction!

cleacos (approved):
The code looks good. Tests passed!

LGTM!

cleacos:
I'm going to merge it... I need the code in trunk.

PR #32780, opened 2023-08-31
Boost: Add Image CDN quality control

haqadn:
Fixes Automattic/boost-cloud#364

## Proposed changes:
* Added a quality slider for premium users.
* Registered a new data-sync entry to check premium features from front-end.

### Other information:

## Jetpack product discussion
None

None

## Testing instructions:
* Apply the filter `add_filter( 'jetpack_boost_has_feature_image-cdn-quality', '__return_true' );` to enable the feature
* Check Boost settings to see if the quality slider is visible, and that changing the slider changes the image quality.
* While the slider is not at 100, remove the filter to disable the feature.
* Check the image quality to make sure they are at 100%.

image


pyronaur (changes requested):
This looks really good - nice to see data-sync doing what it was designed to do, making this change really straight forward 👍

There's one thing that's missing from UX perspective - "I changed the value, I don't like the result, but I don't remember what was the original value.". We probably need a reset button there somewhere.

projects/plugins/boost/app/lib/Premium_Features.php
```diff
@@ -9,26 +9,39 @@ class Premium_Features {
...
+		}
 
-		if ( ! is_array( $features ) ) {
-			$features = Boost_API::get( 'features' );
-			if ( ! is_array( $features ) ) {
-				$features = array();
+		$features = array();
+		// Prepare a list of features after applying jetpack_boost_has_feature_* filter for each feature.
+		foreach ( $all_features as $feature ) {
+			if ( apply_filters( "jetpack_boost_has_feature_$feature", in_array( $feature, $available_features, true ) ) ) {
```
pyronaur:
```suggestion
			if ( apply_filters( "jetpack_boost_has_feature_{$feature}", in_array( $feature, $available_features, true ) ) ) {
```

projects/plugins/boost/app/modules/optimizations/image-cdn/class-image-cdn.php
```diff
@@ -18,4 +21,19 @@ public static function get_slug() {
...
+	 * Add quality arg to existing photon args.
+	 *
+	 * @param $args array - Existing photon args.
+	 *
+	 * @return mixed
+	 */
+	public function add_quality_args( $args ) {
+		if ( Premium_Features::has_feature( Premium_Features::IMAGE_CDN_QUALITY ) ) {
+			$args['quality'] = jetpack_boost_ds_get( 'image_cdn_quality' );
+		}
```
pyronaur: Instead of performing this check on every image, I think this should probably be done at the `setup` level.

projects/plugins/boost/wp-js-data-sync.php
```diff
@@ -162,5 +166,8 @@ function jetpack_boost_initialize_datasync() {
 $css_excludes_entry = new Minify_Excludes_State_Entry( 'minify_css_excludes' );
 jetpack_boost_register_option( 'minify_js_excludes', Schema::as_array( Schema::as_string() )->fallback( Minify_JS::$default_excludes ), $js_excludes_entry );
 jetpack_boost_register_option( 'minify_css_excludes', Schema::as_array( Schema::as_string() )->fallback( Minify_CSS::$default_excludes ), $css_excludes_entry );
+jetpack_boost_register_option( 'image_cdn_quality', ( new Schema( new Type_Constrained_Number( 1, 100 ) ) )->fallback( 100 ) );
```
pyronaur: I think the fallback should be whatever the default value for photon quality slider is right now

pyronaur (changes requested):
A few more things

projects/plugins/boost/wp-js-data-sync.php
```diff
@@ -162,5 +165,47 @@ function jetpack_boost_initialize_datasync() {
...
 jetpack_boost_register_option( 'minify_js_excludes', Schema::as_array( Schema::as_string() )->fallback( Minify_JS::$default_excludes ), $js_excludes_entry );
 jetpack_boost_register_option( 'minify_css_excludes', Schema::as_array( Schema::as_string() )->fallback( Minify_CSS::$default_excludes ), $css_excludes_entry );
+jetpack_boost_register_option(
+	'image_cdn_quality',
+	Schema::as_assoc_array(
+		array(
+			'jpg'  => Schema::as_assoc_array(
+				array(
+					'quality'  => Schema::as_number(),
+					'lossless' => Schema::as_boolean()->fallback( false ),
```
pyronaur: If the lossless fallback is false, there should also probably be a number fallback to the original quality?

projects/plugins/boost/wp-js-data-sync.php
```diff
@@ -162,5 +165,47 @@ function jetpack_boost_initialize_datasync() {
...
+			'png'  => array(
+				'quality'  => 80,
+				'lossless' => false,
+			),
+			'webp' => array(
+				'quality'  => 80,
+				'lossless' => false,
+			),
+		)
+	)
```
pyronaur, line 206: Then this could be avoided? Or is this a Schema::assoc_array limitation?

pyronaur (dismissed):
Looks good, but left a schema comment - have a look if it's relevant.

pyronaur (dismissed):

pyronaur (approved):

PR #32789, opened 2023-08-31
Fix: Exclude Pantheon domains starting with live- from known staging

bindlegirl:
Pantheon live sites automatically get a domain like this: live-site-name.pantheonsite.io. Our checks for staging sites don't recognize this and place live sites in staging mode. 
## Proposed changes:

* Edited regex so domains starting with live- will be ignored when performing staging check.

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/projects/packages/status/changelog/fix-pantheon-staging-domain
+++ b/projects/packages/status/changelog/fix-pantheon-staging-domain
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Exclude domains starting with live from known Pantheon staging domains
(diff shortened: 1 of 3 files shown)


jeherve (dismissed):
Do you think you could add a few valid and invalid examples to `get_is_staging_site_known_hosting_providers_data` in the tests, so we cover this change and ensure this works?

Thanks!

bindlegirl:
I added all standard Pantheon subdomains as listed [here](https://docs.pantheon.io/guides/domains/platform-domains) to tests. Thanks for the suggestion @jeherve

jeherve (approved):

PR #32790, opened 2023-08-31
Launchpad: Fixes the task path property

valterlorran:
## Proposed changes:

* Some tasks used the `get_task_url` key to define the path that Calypso interprets, but this key was not being used. The correct key is `get_calypso_path`.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* Sandbox the public API and use the command below to apply this diff to your sandbox
```
bin/jetpack-downloader test jetpack-mu-wpcom-plugin update/use-get-calypso-path-prop
```
* Create a new account and a new site with the "build" intent by selecting 'Promote myself or business' on the goals page.
* Go through the flow until you land on the Customer Home.
* On your sandbox, update the `is_visible_callback` callback value to `__return_true` in the `domain_claim` task definition:
```
> wp-content/mu-plugins/jetpack-mu-wpcom-plugin/production/vendor/automattic/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php

'domain_claim'                    => array(
	'get_title'            => function () {
		return __( 'Claim your free one-year domain', 'jetpack-mu-wpcom' );
	},
	'is_complete_callback' => 'wpcom_launchpad_is_domain_claim_completed',
	'is_visible_callback'  => '__return_true',
	'get_calypso_path'     => function ( $task, $default, $data ) {
		return '/domains/add/' . $data['site_slug_encoded'];
	},
),
```
* Open the Network tab on the dev console, and look for the `/wpcom/v2/sites/:siteSlug/launchpad?_locale=user&checklist_slug=intent-build&_envelope=1` URL.
* Verify that you can see the `calypso_path` on the `verify_email` and `claim_domain` tasks
Screen Shot 2023-08-31 at 10 23 49

For the `domain_upsell` & `site_launched` task, since we are not using it on the Customer Home launchpad, I decided to remove the path definition.

PR #32796, opened 2023-08-31
Launchpad: Adapt the pre launch tasks to work on the Customer Home

valterlorran:
Related to https://github.com/Automattic/wp-calypso/issues/81186

## Proposed changes:

* This PR adapts the pre-launch tasks for the Write intent to be compatible with the Customer Home launchpad. I added the `get_calypso_path` to several tasks that are in the write intent task list.

### Other information:

## Testing instructions:

* Apply this diff in your sandbox with the following command
```
bin/jetpack-downloader test jetpack-mu-wpcom-plugin update/pre-launch-write-customer-home
```
* Use the Calypso Live link on the following PR: https://github.com/Automattic/wp-calypso/pull/80661
* Create a new site and select the `Write & Publish` option on the Goal step
* Go through the setup flow until you reach the pre-launch launchpad
* Skip the pre-launch launchpad
* You should land on the Customer Home
* Click on the tasks to ensure the click event works.

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -38,6 +38,9 @@ function wpcom_launchpad_get_task_definitions() {
 			},
 			'is_complete_callback' => '__return_true',
 			'is_disabled_callback' => 'wpcom_launchpad_is_design_step_enabled',
+			'get_calypso_path'     => function ( $task, $default, $data ) {
+				return '/setup/update-design/designSetup?siteSlug=' . $data['site_slug_encoded'];
+			},
 		),
 		'domain_claim'                    => array(
 			'get_title'            => function () {
@@ -82,6 +85,9 @@ function wpcom_launchpad_get_task_definitions() {
 			'subtitle'             => 'wpcom_launchpad_get_plan_selected_subtitle',
 			'is_complete_callback' => '__return_true',
 			'badge_text_callback'  => 'wpcom_launchpad_get_plan_selected_badge_text',
+			'get_calypso_path'     => function ( $task, $default, $data ) {
+				return '/plans/' . $data['site_slug_encoded'];
+			},
 		),
 		'setup_general'                   => array(
 			'get_title'            => function () {
--- a/projects/packages/jetpack-mu-wpcom/changelog/update-pre-launch-write-customer-home
+++ b/projects/packages/jetpack-mu-wpcom/changelog/update-pre-launch-write-customer-home
@@ -1,0 +1,4 @@
+Significance: patch
+Type: changed
+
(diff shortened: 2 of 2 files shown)


projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -99,7 +105,7 @@ function wpcom_launchpad_get_task_definitions() {
 			},
 			'isLaunchTask'          => true,
 			'add_listener_callback' => 'wpcom_launchpad_add_site_launch_listener',
-			'get_task_url'          => function ( $task, $default, $data ) {
+			'get_calypso_path'      => function ( $task, $default, $data ) {
```
daledupreez: I just tested and we _do not_ want this to be returned, as the current client-side logic in https://github.com/Automattic/wp-calypso/pull/80661 will redirect to this URL instead of showing the in-page site launch modal. You're also removing this key in https://github.com/Automattic/jetpack/pull/32790, which is the right behaviour, so I think you can drop this change altogether.

```suggestion
			'get_task_url'          => function ( $task, $default, $data ) {
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -38,6 +38,9 @@ function wpcom_launchpad_get_task_definitions() {
 			},
 			'is_complete_callback' => '__return_true',
 			'is_disabled_callback' => 'wpcom_launchpad_is_design_step_enabled',
+			'get_calypso_path'     => function ( $task, $default, $data ) {
+				return '/setup/update-design/designSetup?siteSlug=' . $data['site_slug_encoded'];
+			},
```
daledupreez, line 43: I just worked through this flow and there is an incompatibility between the flow implementation and the skip logic we want to introduce. As a result of the following logic, we will always push the user back to `/setup/free/launchpad`, when we should be redirecting them to `/home/:siteSlug` when the `launchpad_screen` option is set to `skipped`.
https://github.com/Automattic/wp-calypso/blob/trunk/client/landing/stepper/declarative-flow/update-design.ts#L72-L74

daledupreez (changes requested):
There are two important items we need to resolve before we can land this change.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
daledupreez: Noting that I am addressing this issue in https://github.com/Automattic/wp-calypso/pull/81317

daledupreez (approved):
Thanks for cleaning this up! The changes look good to me, so I am going to go ahead and get this merged so we don't miss the weekly alpha release.

PR #32800, opened 2023-08-31
Paywall Block: Email confirmation pending state

lezama:
#### Before

Screenshot 2023-08-31 at 19 04 44

#### After

Screenshot 2023-08-31 at 18 49 30

### Other information:

## Testing instructions:

* Try subscribing with an email that's not associated with a wpcom user

### DIFF
--- a/projects/plugins/jetpack/changelog/update-paywall-subscription-pending
+++ b/projects/plugins/jetpack/changelog/update-paywall-subscription-pending
@@ -1,0 +1,4 @@
+Significance: minor
+Type: enhancement
+
+Paywall: Better content for when waiting for email validation.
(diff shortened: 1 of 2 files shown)


edanzer (approved):
Changes look straightforward and this tests as expected. 

I might tweak the text, but it's entirely a matter of taste, and I wouldn't be surprised if others end up having thoughts. We can deploy and change later depending on feedback. If I were going to tweak it myself I might consider: 
 - Use Verify vs Validate in the new header
 - Add a period at the end of the sentence
 - Replace 'You can now...' with 'Please' 
 
With those changes you'd get: 

verify

I'd love to shorten the paragraph a bit more so it drops to one line by default. It may also be worth adding that text-wrap: pretty/balance CSS that we discussed on some other recent PRs CSS even though it has limited browser support. 

Again none of that feedback is a blocker, and any could be handled in a follow up. It's already a solid improvement.

lezama:
Merging so I don't need to wait for all the tests and everything to pass again 😆 

But yes, let's work on the copy! I will push a PR with your suggestions for discussion. Thanks!

PR #32801, opened 2023-08-31
Paywall Block: Update Settings panel

lezama:
## Proposed changes:

* Use the same logic that display a dialog when trying to select paid subscribers without all the requirements.
* Use core's [RadioControl](https://wordpress.github.io/gutenberg/?path=/docs/components-radiocontrol--docs)
 instead of a custom implementation of the radio buttons.

### Other information:

## Testing instructions:

* On a site without paid plans or stripe connected play with the settings from the panel and the toolbar


edanzer (approved):
Tests as expected and simplifies the code.

PR #32807, opened 2023-09-01
Setup blogroll attributes and context

heavyweight:
This PR wires the attributes and context that we need to the blogroll block
We have the settings defined here: p9Jlb4-8nF-p2#block-settings

Fixes #
https://github.com/Automattic/wp-calypso/issues/80909

## Proposed changes:
* Add the blogroll context
* Add the blogroll item attributes

### Other information:

## Testing instructions:

* Checkout this branch
* Build jetpack
* Apply this patch [print_attributes.patch](https://github.com/Automattic/jetpack/files/12498972/print_attributes.patch)
* Edit a post and add the blogroll block
* Check the console and the attributes should be printed
* When you change the blogroll settings the changes should also be printed

Screenshot 2023-09-01 at 16 48 40


heavyweight:
Need to update the unit tests 🫢

agrullon95 (approved):
@heavyweight - I'm assuming you removed the console log to print the toggled setting?

I tested on simple and atomic. I was able to see and update the setting toggles for the blogroll block.

PR #32808, opened 2023-09-01
Add a Paid Newsletter Recommendation card

mikestottuk:
### Changes proposed in this Pull Request:
This PR adds a Recommendation to set up a Paid Newsletter. It triggers in the following conditions:-

- A site has the Newsletter module active (subscriptions)
- A site has over 100 subscribers but less than 100k subscribers
- It triggers as an action hooked to the fetch of the Subscriber stats (so needs the operator to visit the stats page).

### Jetpack product discussion

See: pbNhbs-8a3-p2

### Does this pull request change what data or activity we track or use?
No

### Testing instructions:
A pickle of a PR to test. It took me a while to get my test site to show as having more than 100 subscribers. I ended up having to manually add them from `https://wordpress.com/subscribers/{site}` importing via CSV they didn't seem to process on my ngrok site. 

- Get over 100 subscribers on your site 
- Clear the Jetpack Dev options 
- Visit the Dashboard there should be no recommendations shown
- Go to Stats
- Click the Subscribers tab - it should show you have over 100 subscribers
- Visit the Dashboard again - the red bubble should show up next to Recommendations
- Click the Recommendations tab - the prompt to set up a Paid Newsletter will show
- This then takes you through the flow of setting up a new plan. See the product discussion for a diagram of that flow.

It's worth nothing that on code review, the `total` in the current stats endpoint doesn't total the subscribers, it needs to pass `?type=all` to get around this I've summed the `total_wpcom` and `total_email` data to give the total which is the same with or without `?type=all` being passed. 

See: p1694436980423499/1694436172.515229-slack-C02TCEHP3HA for discussion.

### Here's how it should look

https://github.com/Automattic/jetpack/assets/4077604/6297e40a-89a9-42ec-992b-60e2726b117a


projects/plugins/jetpack/_inc/client/recommendations/feature-utils.js
```diff
@@ -597,6 +604,41 @@ export const getStepContent = ( state, stepSlug ) => {
...
+					'jetpack'
+				),
+				descriptionList: [
+					__( 'Paid newsletter subscription.', 'jetpack' ),
+					__( 'Publish premium content.', 'jetpack' ),
+					__( 'Earn money through your Newsletter.', 'jetpack' ),
+				],
+				ctaText: __( 'Learn More', 'jetpack' ),
+				ctaLink: getRedirectUrl( 'jetpack-support-paid-newsletter' ),
+				illustration: 'assistant-newsletter',
```
jboland88: The success step is an interesting idea! 
It's a bit outside the pattern for what the assistant does now, we don't typically provide these kinds of confirmation screens as recommendation steps. 

I wonder if there's a way to provide some stronger confirmation on the Calypso side instead of adding an extra step in this flow?

projects/plugins/jetpack/_inc/client/state/recommendations/reducer.js
```diff
@@ -338,6 +338,8 @@ const stepToNextStepByPath = {
 		monitor: 'related-posts',
 		'related-posts': 'newsletter',
 		newsletter: 'site-accelerator',
+		'paid-newsletter': 'summary',
+		'paid-newsletter-success': 'summary',
```
jboland88: How do users reach the success step?
I don't see this mentioned in the test plan?

projects/plugins/jetpack/_inc/lib/class-jetpack-recommendations.php
```diff
@@ -145,6 +147,38 @@ public static function init_conditional_recommendation_actions() {
 
 		// Monitor for changes in plugins that have auto-updates enabled
 		add_action( 'update_site_option_auto_update_plugins', array( static::class, 'plugin_auto_update_settings_changed' ), 10, 3 );
+
+		// Subscriber count check. This runs when stats are fetched.
+		add_action( 'jetpack_followers_fetched', array( static::class, 'recommend_paid_newsletter' ), 10, 2 );
```
jboland88: Nitpick, I think the last argument here should be a `1` since there's only one argument passed to the called function.

projects/packages/stats/src/class-wpcom-stats.php
```diff
@@ -231,8 +231,12 @@ public function get_views_by_country( $args = array() ) {
 	public function get_followers( $args = array() ) {
 
 		$this->resource = 'followers';
+		$followers      = $this->fetch_stats( $args );
 
-		return $this->fetch_stats( $args );
+		// run an action which allows us to fire a recommendation.
+		do_action( 'jetpack_followers_fetched', $followers );
```
jboland88, line 237: I wonder if there's another place that we can trigger an action to evaluate follower counts.
Where do we allow new subscribers, could we trigger an action each time someone subscribes to the blog?

This works OK, but it seems a bit out of place in this utility getter function.

projects/plugins/jetpack/_inc/lib/class-jetpack-recommendations.php
```diff
@@ -145,6 +147,38 @@ public static function init_conditional_recommendation_actions() {
...
+		// return if the subscriptions module is not active
+		if ( ! Jetpack::is_module_active( 'subscriptions' ) ) {
+			return;
+		}
+
+		// if followers array does not have total key, return.
+		if ( array_key_exists( 'total', $followers ) ) {
+			// the endpoint used in class-wpocom-stats does not pass type=all as a parameter. Summing the parts gives the total for this purpose
+			$total_subscribers = $followers['total_email'] + $followers['total_wpcom'];
+			if ( $total_subscribers >= 100 && $total_subscribers < 100000 ) {
```
jboland88, line 176: For future tests, I think we can just temporarily edit this line to change the requirements to enable the recommendation.

jboland88 (changes requested):
I was not sure how the success step fits in with the rest of the changes?

I'm curious if we can trigger an action when a new subscriber signs up for the blog? (I was digging around for this logic but did not have time to get to the bottom of it)

projects/plugins/jetpack/_inc/client/recommendations/feature-utils.js
mikestottuk: We have cards for welcome to paid products, which is where I got the idea from e.g.

`wp-admin/admin.php?page=jetpack#/recommendations/welcome-complete`
`welcome-security`

> I wonder if there's a way to provide some stronger confirmation on the Calypso side instead of adding an extra step in this flow?

Yes, could show a post plan setup modal on Calypso - it just leaves people in place. Let me strip it out of this PR and keep this one focussed on the recommendation stage.

projects/packages/stats/src/class-wpcom-stats.php
mikestottuk: I had a good look through here: https://developer.jetpack.com/hooks/

> This works OK, but it seems a bit out of place in this utility getter function.

Maybe we need to have a webhook which WP.com fires to notify sites when they hit milestones.

mikestottuk: > This works OK, but it seems a bit out of place in this utility getter function.

Other options:-

- Keep it as it is, and it triggers when the subscriber counts are fetched.
- Do a timed cron based check where we call the WP.com API, would rely on the cronjob being triggered.
- Do something else, there's Achievements, which can email a notification & show a badge when milestones are hit:  fbhepr%2Skers%2Sjcpbz%2Sjc%2Qpbagrag%2Szh%2Qcyhtvaf%2Snpuvrirzragf.cuc%3Se%3Qn7p82158%23658-og

projects/plugins/jetpack/_inc/lib/class-jetpack-recommendations.php
mikestottuk: Yeah, set it to 101 :)

projects/plugins/jetpack/_inc/client/state/recommendations/reducer.js
mikestottuk: See above, I'll break it into two PRs

projects/packages/stats/src/class-wpcom-stats.php
jeherve, line 237: When adding a new hook, you'll want to add a docblock so our parser can automatically add the hook to our [Jetpack Codex](https://developer.jetpack.com/).

Here is an example:
https://github.com/Automattic/jetpack/blob/5dcfe76bb621fecb87673f81b8324a68c9740628/projects/packages/stats/src/class-main.php#L221

jeherve (commented):
Quickly noting that since you're updating the Recommentations wizard, you would need to also update the matching e2e test in `specs/post-connection/recommendations.test.js`.

projects/plugins/jetpack/_inc/lib/class-jetpack-recommendations.php
```diff
@@ -145,6 +147,38 @@ public static function init_conditional_recommendation_actions() {
...
+				self::enable_conditional_recommendation( self::PAID_NEWSLETTER_RECOMMENDATION );
+			}
+		} else {
+			return;
```
jboland88, line 180: Nitpick, this return is not needed since there's no subsequent logic.

jboland88 (commented):
This tested well, the only thing that I noticed is that there seems to be an image missing. I think there needs to be an `assistant-newsletter.webp` image. I noticed this because it did not load for me:

Screenshot 2023-09-14 at 4 13 10 PM

I think adding the doc block per @jeherve 's comment would be good. Then running the version fixer script and this should be good to go.

projects/packages/stats/src/class-wpcom-stats.php
mikestottuk: @jboland88 - for when I pick this back up. What are your thoughts here?

jboland88: @mikestottuk - IMO the way that you have it now is fine. There may be a better solution out there, however, I think the other solutions are likely going to be more complex to implement.

If this recommendation does well and we want to improve how it is triggered, we could look at iterating in a later PR.

jboland88:
@mikestottuk - this went stale, I'm going to go ahead and close it

PR #32814, opened 2023-09-01
AI Assistant: Add language reminder to action prompts

### LINKED ISSUE #31820: AI Assistant: English content is generated after asking for some modification on non-Engllish content
ref p1688718666140519-slack-C054LN8RNVA

> The AI Assistant UI is in French as I would expect
I sent a prompt in French and it worked fine (generated content was in French)
I then worked on the generated content with the AI content editing tool and noticed that sometimes (not always though) summarize, simplify, change tone, makes the language of the text switch to English.

dhasilva:
Fixes #31820

## Proposed changes:

* Adds a rule to the text of the toolbar options prompts.

After many tests, adding it to the rules array had little to no effect, but adding it directly to the message seems to work very well. Maybe we'll need to rethink where rules are defined.

### Other information:

## Testing instructions:

* Add an AI Assistant block
* Ask for some content in a language other than English
* Iterate over the content in its language with the toolbar options
* Check that the content is not translated to English at any point
* Translate the content to English
* Check that the next iterations return English as normal
* Check for any regressions


retrofox (commented):
After a few tests, it works pretty well, considering the AI response feels imperceptible sometimes.
About the prompt, are we duplicating the same rule?

```
Rewrite your last answer with an informal tone. Ensure that your response is in the same language as your last answer. Do not switch to any language other than the language of your last answer in your response.
```

` Ensure that your response is in the same language as your last answer.` and `Do not switch to any language other than the language of your last answer in your response` seem to try to achieve the same translation goal.

nunyvega:
Almost all of the options worked pretty fine except for some problems that I had a couple times with “Change tone”.

It worked for me when tweaking `languageReminder` to:
```. Ensure that your response is in the same language as the post content. Do not switch to any language other than the language of ${ subject } in your response`;```

dhasilva:
> About the prompt, are we duplicating the same rule?

I tried with each phrase separately, but it seemed to me that having both worked much better than having only one.

dhasilva:
> It worked for me when tweaking `languageReminder` to: ...

I changed that, but withg a check for title generation.

retrofox (approved):
LGTM! It keeps the same language after the following requests. Bellow, I asked to change the tone, and the suggestion preserved the language (Spanish)


Screenshot 2023-09-13 at 16 23 25

image

PR #32824, opened 2023-09-04
Featured images fallback: add srcset, width/height, and loading attributes

sgomes:
In an effort to move away from the `devicepx` script towards native functionality, this PR adds a `srcset` to qualifying featured images obtained via `jetpack_featured_images_fallback_get_image` (which is used as a filter on `post_thumbnail_html`). The browser will automatically pick the most appropriate image from the options in the `srcset`, depending on the screen pixel ratio and the current zoom value.

Qualifying images also get `width` and `height` to avoid layout shifts.

Furthermore, all images get the `loading` attribute explicitly defined for an ancillary performance improvement: `lazy` on non-singular pages where several featured images will most likely be loaded, and `eager` on singular pages where the featured image will likely be at the top and is thus a strong candidate for the Largest Contentful Paint element.

Qualifying images are defined as those for which we know the source width and height at serving time, which typically means they're attachment images.

**Important note regarding wpcom:** as far as I can tell, this change will have no effect on wpcom, since a different codepath is used there instead, in a nearly-identical file. I will create a companion diff to this PR in wpcom once these changes are accepted.

## Proposed changes:
* Add support for width-based `srcset`s to `Jetpack_PostImages::generate_cropped_srcset`
* For qualifying images, add `srcset` to the image returned by `jetpack_featured_images_fallback_get_image`
* For qualifying images, add `width` and `height` to the image returned by `jetpack_featured_images_fallback_get_image`
* Add `loading=lazy` by default to all images returned by `jetpack_featured_images_fallback_get_image` on non-singular pages, and `loading=eager` by default to all images returned by `jetpack_featured_images_fallback_get_image` on singular pages

### Other information:

## Jetpack product discussion
None.

No.

## Testing instructions:

Start by creating at least two blog posts, with one image each:
- One blog post with a large attachment image, i.e., an image uploaded to the site
- One blog post with a large external image, i.e., an image provided from an external URL

For themes that use cropped thumbnails:
- Apply a theme with cropped thumbnails, such as Gazette
- Visit a blog post with an external image
  - Inspect the featured image (in the case of Gazette, it's a banner at the top)
  - Ensure the featured image is correctly resized and cropped to the theme's desired aspect ratio
  - Ensure the featured image has the `loading` attribute set to `eager`
- Visit a blog post with an attachment image
  - Inspect the featured image (in the case of Gazette, it's a banner at the top)
  - Ensure the featured image is correctly resized and cropped to the theme's desired aspect ratio
  - Ensure the featured image has the `loading` attribute set to `eager`
  - Ensure the featured image has a `srcset` attribute with multiple sources defined
  - Ensure the featured image has defined `width` and `height` attributes
- Visit a directory page, e.g. `/2023/`
  - Inspect the featured images (in the case of Gazette, they're only the post thumbnails defined as `` tags with the `wp-post-image` class; some other images are style backgrounds instead and are not relevant here)
  - Ensure they have the `loading` attribute set to `lazy`

For themes that use uncropped thumbnails:
- Apply a theme with uncropped thumbnails, such as Baskerville 2
- Visit a blog post with an external image
  - Inspect the featured image (in the case of Baskerville 2, it's the large image at the top)
  - Ensure the featured image is correctly resized and uncropped, maintaining the original aspect ratio
  - Ensure the featured image has the `loading` attribute set to `eager`
- Visit a blog post with an attachment image
  - Inspect the featured image (in the case of Baskerville 2, it's the large image at the top)
  - Ensure the featured image is correctly resized and uncropped, maintaining the original aspect ratio
  - Ensure the featured image has the `loading` attribute set to `eager`
  - Ensure the featured image has a `srcset` attribute with multiple sources defined
  - Ensure the featured image has defined `width` and `height` attributes
- Visit a directory page, e.g. `/2023/`
  - Inspect the featured images (in the case of Baskerville 2, they're the post thumbnails, and they're inside `div`s with the `featured-media` class)
  - Ensure they have the `loading` attribute set to `lazy`


sgomes:
**Important note regarding wpcom:** as far as I can tell, this change will have no effect on wpcom, since a different codepath is used there instead, in a nearly-identical file. I will create a companion diff to this PR in wpcom once these changes are accepted.

jeherve:
> Visit a blog post with an external image
> Inspect the featured image (in the case of Gazette, it's a banner at the top)

Do you have tips on how to do this? In my tests external images do not seem to get handled like internal images do, they do not appear as featured images. e.g. in [this post](https://jeherve5.wpcomstaging.com/2023/09/05/blog-post-with-a-large-external-image/).

sgomes:
> > Visit a blog post with an external image
> > Inspect the featured image (in the case of Gazette, it's a banner at the top)
> 
> Do you have tips on how to do this? In my tests external images do not seem to get handled like internal images do, they do not appear as featured images. e.g. in [this post](https://jeherve5.wpcomstaging.com/2023/09/05/blog-post-with-a-large-external-image/).

Thank you for taking a look, @jeherve! In my case, I used an image hosted in wpcom (`<blogname>.files.wordpress.com`), which may be going through a different codepath than an image hosted elsewhere?

jeherve:
> I used an image hosted in wpcom (`<blogname>.files.wordpress.com`), which may be going through a different codepath than an image hosted elsewhere?

That doesn't seem to change anything [in my test](https://jeherve5.wpcomstaging.com/2023/09/05/blog-post-with-a-large-external-image/). Looking at how we fetch featured image fallbacks, I wouldn't expect external images to be picked up really:
https://github.com/Automattic/jetpack/blob/3a3c4adc1c0913bac224ddb832713bd6101fe1ab/projects/plugins/jetpack/modules/theme-tools/content-options/featured-images-fallback.php#L41-L48

sgomes:
> Looking at how we fetch featured image fallbacks, I wouldn't expect external images to be picked up really

Thank you for taking another look, @jeherve! 

Sorry for the late reply, but it took me a while to understand what was happening. After some debugging it looks like this is because I was testing in wpcom, which allows for unsized external images. Vanilla Jetpack can too, but not by default; you need to add a filter to match wpcom behaviour. I've made that easy to test at the bottom, but read on for the details if you're interested.

To elaborate, I can say that in my case, the image is coming from HTML. `Jetpack_PostImages::get_images` defaults to `'from_html'  => true`, so even though `from_html` is not being passed as an explicit option in the code you mentioned, it's still `true` and thus Jetpack ends up looking for images in the HTML.

When I tried setting up an ephemeral site, I saw the same behaviour that you did; I couldn't get external images to be picked up as the featured image when inserting them with an image block. Images added in a classic block work, but that's only because the classic block adds their `width`/`height` to the HTML.

The difference in behaviour between wpcom and the ephemeral site I'm testing turned out to be `jetpack_postimages_ignore_minimum_dimensions`; if it's filtered to return `true`, it will pick up any images, whether or not they have any dimensions defined. And that filter is indeed set in wpcom. So I guess the only way to test with unsized external images is to do:

```php
add_filter( 'jetpack_postimages_ignore_minimum_dimensions', '__return_true' );
```

To make that a little easier, I set up a new branch (`test/featured-images-fallback-add-srcset-with-filter`) that adds the filter as an extra commit on top of the ones in this PR, and that you can use with the Jetpack Beta plugin to test the behaviour of sites that allow for unsized images. I have verified that the branch works as intended on my ephemeral site, so you should be able to use it for testing as well.

Thanks again!

jeherve (approved):
This seems to test well for me now, thank you!

sgomes:
Thank you so much for the review, @jeherve! 🙇 And sorry again for all the false starts, glad we got there in the end 🙂

PR #32826, opened 2023-09-04
JITM - show a link to settings after a module is activated

mikestottuk:
Allow a JITM that activates a module to be redirected to an admin page if activated rather than simply fade out.

## Proposed changes:
* This adds a new flow to redirect to settings if a JITM is clicked which has an activate module as the CTA. Previously this would just fade out after 2 seconds and leave people where they were.
* If nothing is passed (e.g. we don't have a line like `->settings_redirect( 'jetpack#/newsletter' )`) the JITM will fade away as previously.
* Review alongside D120595-code

### Other information:

## Jetpack product discussion
See here: pbNhbs-88F-p2#comment-17900

No

## Testing instructions:

* See D120595-code for detailed testing instructions.


projects/packages/jitm/src/js/jetpack-jitm.js
```diff
@@ -208,6 +209,11 @@ jQuery( document ).ready( function ( $ ) {
 				$( '#jitm-banner__activate a' ).text( window.jitm_config.activated_module_text );
 				$( '#jitm-banner__activate a' ).attr( 'disabled', true );
 
+				// Redirect to module settings page if a redirect is set in the JITM envelope.
+				if ( $activate_button.data( 'redirect' ) ) {
+					window.location.href = window.jitm_config.jetpack_admin_url + $activate_button.data( 'redirect' );
+				}
+
```
jboland88: Something about this doesn't sit right.

From a UX perspective, it seems unexpected that clicking an "Activate" button would redirect the user IMO.

I also think redirecting with `window.location.href` by pulling from the `window` object and the DOM might come with some risks. It could leave the door open to redirect somewhere else if another script from a different plugin decided to do something unfriendly and manipulate the window and DOM. 

cc @IanRamosC for some additional thoughts/ second opinion on this redirect method.

Would it work to swap the activate button for a "settings" button after the activation is complete?

jboland88 (commented):
I am not sure about this solution - It works, but it feels like a hack. I'm going to ask for a second opinion from the team.

projects/packages/jitm/src/js/jetpack-jitm.js
mikestottuk: > Would it work to swap the activate button for a "settings" button after the activation is complete?

Interesting - the main issue was the fade away without a clear next step. Swapping it to stay in place and have Activate => Settings would work. Then they can dismiss it with the "X" or click settings to go to where the re-direct would be. 

Let me modify on that basis.

projects/packages/jitm/src/js/jetpack-jitm.js
```diff
@@ -185,7 +186,7 @@ jQuery( document ).ready( function ( $ ) {
 			var $activate_button = $( this );
 
 			// Do not allow any requests if the button is disabled.
-			if ( $activate_button.attr( 'disabled' ) ) {
+			if ( $activate_button.attr( 'disabled' ) || $activate_button.attr( 'link' ) ) {
```
jboland88: Curious here, wouldn't this disable clicks on the updated settings button?

projects/packages/jitm/src/js/jetpack-jitm.js
```diff
@@ -206,8 +207,16 @@ jQuery( document ).ready( function ( $ ) {
 				},
 			} ).done( function () {
 				$( '#jitm-banner__activate a' ).text( window.jitm_config.activated_module_text );
-				$( '#jitm-banner__activate a' ).attr( 'disabled', true );
+				
+				// Present a link to module settings page if a settings link is set in the JITM envelope.
+				if ( $activate_button.data( 'settings_link' ) ) {
+					$( '#jitm-banner__activate a' ).text( window.jitm_config.settings_module_text );
+					$( '#jitm-banner__activate a' ).attr( 'href', window.jitm_config.jetpack_admin_url + $activate_button.data( 'redirect' ) );
```
jboland88: Nice improvement over the redirect 👍  

What would you think about passing the full settings URL for the activated module in the JITM response and adding a settings button in the JITM template? 
Then, when you get to this step, you could just hide the activation button and show the settings button.

This allows us to avoid splitting up the URL into two parts, one that comes from the plugin and one that comes from the JITM response and having to re-assemble here. It will likely be less fragile to have the entire settings URL coming from a single source.

jboland88 (changes requested):
This is improved, I had one comment about having the entire settings url coming from the JITM response (similar to how we specify the CTA URL) instead of part of the URL coming from the plugin and then needing to assemble the URL in code.

projects/packages/jitm/src/js/jetpack-jitm.js
mikestottuk: Yeah was trying to test this bit and whether it sent me to the link, clicking settings without it started to try and activate the module again.

projects/packages/jitm/src/js/jetpack-jitm.js
mikestottuk: Yeah, I think I'll do this - have another button drawn if it has the data in the envelope and then show hide it with JS

projects/packages/jitm/src/js/jetpack-jitm.js
IanRamosC: >cc @IanRamosC for some additional thoughts/ second opinion on this redirect method.

Just saw the comment now (that this seems to be resolved already), but I wanted to leave a comment for posterity. I agree with you. It's risky to redirect that way, especially if the redirect URL comes from the `window` object or anything else we don't have control over unless we validate it.

projects/packages/jitm/src/js/jetpack-jitm.js
```diff
@@ -72,13 +72,26 @@ jQuery( document ).ready( function ( $ ) {
...
 					envelope.activate_module +
+					'" data-settings_link="' + envelope.settings_link +
 					'" type="button" class="jitm-button is-compact is-primary jptracks" data-jptracks-name="nudge_click" data-jptracks-prop="jitm-' +
 					envelope.id +
 					'-activate_module">' +
 					window.jitm_config.activate_module_text +
 					'</a>';
 				html += '</div>';
 			}
+			if ( envelope.settings_link){
```
jboland88: If this is (in theory) always going to be used as part of the module activation routine, would it make sense to nest this logic in the check for `if ( envelope.active_module )` ?

projects/packages/jitm/src/js/jetpack-jitm.js
```diff
@@ -205,8 +218,12 @@ jQuery( document ).ready( function ( $ ) {
 					$( '#jitm-banner__activate a' ).attr( 'disabled', true );
 				},
 			} ).done( function () {
-				$( '#jitm-banner__activate a' ).text( window.jitm_config.activated_module_text );
-				$( '#jitm-banner__activate a' ).attr( 'disabled', true );
```
jboland88: I think we'd want to keep these lines in the case that the activation button did not have a settings button to show. Could these lines just be moved to after the new if statement you added?

jboland88 (changes requested):
I think this is looking close - just a couple minor points of feedback.

jboland88 (dismissed):
It looks like you might need to run the version update script. Otherwise, this LGTM!

robertsreberski (dismissed):
Looks like dependencies are updated correctly 👍

projects/packages/jitm/src/js/jetpack-jitm.js
```diff
@@ -72,14 +72,27 @@ jQuery( document ).ready( function ( $ ) {
 				html +=
 					'<a href="#" data-module="' +
 					envelope.activate_module +
-					'" type="button" class="jitm-button is-compact is-primary" data-jptracks-name="nudge_click" data-jptracks-prop="jitm-' +
+					'" data-settings_link="' + envelope.module_settings_link +
+					'" type="button" class="jitm-button is-compact is-primary jptracks" data-jptracks-name="nudge_click" data-jptracks-prop="jitm-' +
```
jboland88: We actually don't need the `jptracks` class on buttons with the `jitm-button` class anymore - these buttons are tracked from within this file now.

projects/packages/jitm/src/js/jetpack-jitm.js
```diff
@@ -72,14 +72,27 @@ jQuery( document ).ready( function ( $ ) {
...
 					'">' +
 					window.jitm_config.activate_module_text +
 					'</a>';
 				html += '</div>';
+				if ( envelope.module_settings_link){
+					html += '<div class="jitm-banner__action" id="jitm-banner__settings" style="display:none;">';
+					html +=
+						'<a href="' +
+						envelope.module_settings_link +
+						'" type="button" class="jitm-button is-compact is-primary jptracks" data-jptracks-name="nudge_click" data-jptracks-prop="jitm-' +
```
jboland88: Same here, could you remove `jptracks` to avoid duplicate tracking?

jboland88 (approved):
Re-approving after version management updates

PR #32834, opened 2023-09-04
beta: Chain PHP error handlers

anomiex:
## Proposed changes:

We set an error handler to log certain messages to Jetpack. If there was another error handler registered before us, chain to that handler instead of calling PHP's own default handler.

### Other information:

## Jetpack product discussion

None

No

## Testing instructions:

* Call `set_error_handler()` in wp-config.php, and make sure your handler still gets called when Jetpack Beta Tester is active.

### DIFF
--- a/projects/plugins/beta/src/class-hooks.php
+++ b/projects/plugins/beta/src/class-hooks.php
@@ -26,6 +26,13 @@ class Hooks {
 	protected static $instance = null;
 
 	/**
+	 * Previous error handler.
+	 *
+	 * @var callable
+	 */
+	protected static $prev_error_handler = null;
+
+	/**
 	 * Main Instance
 	 */
 	public static function instance() {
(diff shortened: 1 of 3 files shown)


projects/plugins/beta/src/class-hooks.php
```diff
@@ -595,7 +609,11 @@ public static function custom_error_handler( $errno, $errstr, $errfile, $errline
 			}
 		}
 
-		// Returning false makes the error go through the standard error handler as well.
-		return false;
+		if ( self::$prev_error_handler ) {
+			return call_user_func( self::$prev_error_handler, $errno, $errstr, $errfile, $errline );
+		} else {
```
ice9js, line 614: Nit: I guess we could drop the else in this case and just have `return false;`. Just a personal preference in the spirit of keeping things simple, it's up to you ;)

ice9js (approved):
Looks good to me 👍

projects/plugins/beta/src/class-hooks.php
anomiex: Doesn't matter to me. But looks like the auto-merge already auto-merged.

PR #32838, opened 2023-09-05
Import: import attachment improvement

ouikhuan:
Fixes Automattic/dotcom-forge#3640

## Proposed changes:

* Remove the `post_date_gmt` check when looking for an existing attachment.
* Disable big image handling when uploading attachment via unified importer to match the behavior as simple site.
* Checking `-scaled` image filename to prevent duplication, and for backward compatibility

### Other information:

## Jetpack product discussion

No

No

## Testing instructions:

* Spin up a JN site and apply this PR or test it locally. Your site needs to be connected to WordPress.com.
* Navigate to `https://wordpress.com/import/:site_slug` and select WordPress importer.
* Get the `substack-small.xml` sample file from D120640-code and performs an import.
* Review the results in media and perform another import using the same file.
* Go to media and make sure the attachments aren't created twice, and there's no images with `-scaled` append to the filename. `9713ae89-a4d3-4c2d-abe5-ade9ed291462_3929x1953.jpeg` will be the one you're looking for.

### DIFF
--- a/projects/packages/import/changelog/update-import-attachment-improvement
+++ b/projects/packages/import/changelog/update-import-attachment-improvement
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Handles scaled images for attachments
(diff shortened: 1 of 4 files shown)


bogiii (approved):
Works as described. All good! 🚢

PR #32849, opened 2023-09-05
Paywall: Add login link to simple sites

lezama:
## Proposed changes:

Add a login link on simple sites:

Screenshot 2023-09-05 at 14 33 55

### Other information:

## Testing instructions:

* On a simple site, while logged out, try to access a post you are already subscribed to.
* After login in you should be taken back to the post and get access.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -1004,7 +1004,7 @@ function get_paywall_blocks_subscribe_pending() {
 <!-- wp:paragraph {"align":"center","style":{"typography":{"fontSize":"14px"},"spacing":{"margin":{"top":"10px","bottom":"10px"}}}} -->
 <p class="has-text-align-center" style="margin-top:10px;margin-bottom:10px;font-size:14px">' . $subscribe_text . '</p>
 <!-- /wp:paragraph -->
-
+</div>
 <!-- /wp:group -->
 ';
 }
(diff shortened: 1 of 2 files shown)


edanzer (approved):
Code changes are surprisingly simple and straightforward for what's being added. The way the link is added looks really nice. 

And this tests as expected. I tested on a simple site and confirm the log in link does how if logged out. I also tested on self-hosted jetpack, and confirmed the log in link does not show.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -956,6 +957,18 @@ function get_paywall_blocks( $newsletter_access_level ) {
...
 
+	$access_question = $is_paid_post
+		? __( 'Already a paid subscriber?', 'jetpack' )
+		: __( 'Already a subscriber?', 'jetpack' );
+
+	$sign_in = '';
+	if ( ! is_user_logged_in() && ( new Host() )->is_wpcom_simple() ) {
+		$sign_in_link = add_query_arg( 'redirect_to', rawurlencode( get_post_permalink() ), 'https://wordpress.com/log-in' );
+		$sign_in      = '<!-- wp:paragraph {"align":"center", "typography":{"fontSize":"14px","fontWeight":"400";"color":"#646970"}} -->
+<p class="has-text-align-center" style="font-size:14px;font-weight:400;color:#646970">' . esc_html( $access_question ) . ' <a href="' . $sign_in_link . '" class="jetpack-subscriber-paywall-login" style="font-size:14px;font-weight:400;color:#646970">' . esc_html__( 'Log in', 'jetpack' ) . '</a></p>
```
simison, line 968: I think it's not worth having fixed colour and instead let theme's primary text colour take over.

That way we won't have problems with themes with different colours.

PR #32850, opened 2023-09-05
Add/allow premium to use subcribe as site subscriber

millerf:
After some discussion we realised that preventing users to use "newsletter" plans for Premium content and payment button is actually not a proper behaviour. 

This PR reverts https://github.com/Automattic/jetpack/pull/31200

 
## Proposed changes:

* Reverting https://github.com/Automattic/jetpack/pull/31200

### Other information:

## Testing instructions:

* Go to '..'
*


simison:
@millerf what do you think if we'd add some indicator after each plan that they're "newsletter" or "donations" plans?

Doesn't have to be in this PR and I think it's a good follow-up next week when Cris is back.

jeherve (approved):

PR #32867, opened 2023-09-05
Update/hide subscribe modal on post access

edanzer:
## Proposed changes:

Hide the subscribe modal if the entire post is set to be visible only to subscribers or paid subscribers. In such cases, the post content itself already has a subscribe block, so showing the modal is redundant.

### Other information:

## Jetpack product discussion

None. 

No.

## Testing instructions:

1) Setup: In your local Jetpack test environment, publish a post and set Access to subscribers only or paid subscribers only. Do not add paywall block to the post. You'll also want to be sure the subscribe modal is enabled at Jetpack > Settings > Newsletters. 
2) In a browser where you are not logged in, open that post, scroll, and confirm the modal does not load. 
   - Extra check: If you previously dismissed the modal, there may be a cookie in your browser that prevents it from showing. To avoid that, you may want to check storage/cookies in your browser dev tools, search for 'dismiss' and if a cookie for dismissing the modal shows delete it.  
3) Test that he modal still does load on a post that is accessible to everyone.

### DIFF
--- a/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
+++ b/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
@@ -194,7 +194,7 @@ HTML;
 		global $post;
 		$access_level              = get_post_meta( $post->ID, META_NAME_FOR_POST_LEVEL_ACCESS_SETTINGS, true );
 		$is_accessible_by_everyone = Token_Subscription_Service::POST_ACCESS_LEVEL_EVERYBODY === $access_level;
-		if ( has_block( 'jetpack/paywall' ) || ! $is_accessible_by_everyone ) {
+		if ( ! $is_accessible_by_everyone ) {
 			return false;
 		}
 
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -189,7 +191,10 @@ public function should_user_see_modal() {
 		}
 
 		// Don't show if post is for subscribers only or has paywall block
-		if ( has_block( 'jetpack/paywall' ) ) {
+		global $post;
+		$access_level              = get_post_meta( $post->ID, META_NAME_FOR_POST_LEVEL_ACCESS_SETTINGS, true );
+		$is_accessible_by_everyone = Token_Subscription_Service::POST_ACCESS_LEVEL_EVERYBODY === $access_level;
+		if ( has_block( 'jetpack/paywall' ) || ! $is_accessible_by_everyone ) {
```
lezama: I think we could just check `! $is_accessible_by_everyone`

lezama (dismissed):

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
edanzer: @lezama - To confirm (since you've worked with the access code more), does $access_level include or account for if there is a paywall block on the page? I assume if that's the case, it means we update this post meta if someone adds a paywall block to the post.

lezama: we add the post meta whenever someone restricts the access to the post, with or without paywall block

lezama (approved):

PR #32869, opened 2023-09-05
Add some subscription meta if exists.

n3f:
Add required meta for https://github.com/Automattic/gold/issues/143

## Testing Instructions

* Look at the subscription form in developer tools and verify that the `post_id` and `tier_id` hidden inputs are set correctly.


n3f:
@millerf, maybe we can just add this to #32710?

millerf (approved):
LGTM!

PR #32871, opened 2023-09-06
Launchpad: Reset the screen status when completing the Newsletter task list

valterlorran:
Related to:
* https://github.com/Automattic/wp-calypso/pull/81433
* D120759-code

## Proposed changes:

* The behavior for the newsletter is slightly different from the Build and Write intent. The Newsletter is completed when the user first publishes a post, so we can't leave the `launchpad_screen` as skipped after completing the first publish task. We need to reset the `launchpad_screen` to prevent users from falling into the condition added on D120759-code.

### Other information:

## Testing instructions:

* To test this PR, please follow the instructions on: https://github.com/Automattic/wp-calypso/pull/81433

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -805,6 +805,13 @@ function wpcom_launchpad_track_publish_first_post_task() {
 	// Since we share the same callback for generic first post and newsletter-specific, we mark both.
 	wpcom_launchpad_mark_launchpad_task_complete_if_active( 'first_post_published' );
 	wpcom_launchpad_mark_launchpad_task_complete_if_active( 'first_post_published_newsletter' );
+
+	$site_intent      = get_option( 'site_intent', false );
+	$launchpad_screen = get_option( 'launchpad_screen', false );
+	if ( 'newsletter' === $site_intent && 'skipped' === $launchpad_screen ) {
+		// First publishing a post should make the pre-launch disappear in case it's a Newsletter.
+		update_option( 'launchpad_screen', 'full' );
+	}
 }
 
 /**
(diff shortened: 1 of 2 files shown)


valterlorran:
On second thought, we can add an extra check on D120759-code to verify if the "first_post_publish_newsletter" is completed, instead of adding this condition here. 🤔

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -805,6 +805,13 @@ function wpcom_launchpad_track_publish_first_post_task() {
 	// Since we share the same callback for generic first post and newsletter-specific, we mark both.
 	wpcom_launchpad_mark_launchpad_task_complete_if_active( 'first_post_published' );
 	wpcom_launchpad_mark_launchpad_task_complete_if_active( 'first_post_published_newsletter' );
+
+	$site_intent      = get_option( 'site_intent', false );
+	$launchpad_screen = get_option( 'launchpad_screen', false );
+	if ( 'newsletter' === $site_intent && 'skipped' === $launchpad_screen ) {
+		// First publishing a post should make the pre-launch disappear in case it's a Newsletter.
+		update_option( 'launchpad_screen', 'full' );
```
daledupreez, line 813: I think we need to update the option's value to `'off'`, as we want the fullscreen launchpad to be disabled, not re-enabled. If the value gets set back to 'full', we may end up having redirects from various locations back to the fullscreen launchpad.
```suggestion
		update_option( 'launchpad_screen', 'off' );
```

daledupreez: I think your suggestion in [this comment](https://github.com/Automattic/jetpack/pull/32871#issuecomment-1707642088) is the right approach here -- I think that keeping the value as "skipped" and then switching which set of tasks we show in Customer Home is going to be the right approach.

daledupreez: I've gone ahead and made the relevant changes in D120759-code - I think we can close this PR.

valterlorran: Thanks for tackling it!

PR #32886, opened 2023-09-06
AI Client: add and expose reset() from useAiSuggestions() hook

retrofox:
This PR adds and exposes the `reset()` helper function to/from the useAiSuggestions() react custom hook. It's used when the consumer needs to reset the current state of the requesting process.

Fixes #

## Proposed changes:

* AI Client: add and expose reset() from useAiSuggestions() hook

### Other information:

## Testing instructions:

* Since the `reset()` is not used yet (follow-up PR coming on), a visual review and checking the compilation is oaky will be enough for now


lhkowalski (approved):
The code builds correctly:

Screenshot 2023-09-06 at 15 33 22

PR #32887, opened 2023-09-06
Paywall: Update login link

lezama:
## Proposed changes:

* Use "link" login instead.
* Make it available on paid posts only.

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/projects/plugins/jetpack/changelog/update-paywall-login-link
+++ b/projects/plugins/jetpack/changelog/update-paywall-login-link
@@ -1,0 +1,4 @@
+Significance: minor
+Type: other
+
+fix link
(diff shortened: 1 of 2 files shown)


simison (approved):
Tested on simple site — works as expected, I got returned to the post after clicking magic login link and I was logged in. 🎉 

We could improve copy all around but that we can follow-up with.

BTW I also tested this for Jetpack site and the link didn't show up as expected. I removed the WP.com check and the login flow worked pretty well for Jetpack sites too! Only the redirecting didn't work — I ended up at WP.com instead. So if we fix the redirect, we could upen this up for Jetpack sites as well.

image

image

image

image

lezama:
>  BTW I also tested this for Jetpack site and the link didn't show up as expected. I removed the WP.com check and the login flow worked pretty well for Jetpack sites too! Only the redirecting didn't work — I ended up at WP.com instead. So if we fix the redirect, we could upen this up for Jetpack sites as well.

But it doesn't get you rid of the paywall I think

PR #32894, opened 2023-09-06
jetpack: Restore some WP Requests library v1 back-compat code

anomiex:
## Proposed changes:

It seems WordPress.com Simple skipped that part of their upgrade to WordPress 6.2, so despite their saying they're on 6.3.1 now the new Requests library classes are missing there.

### Other information:

## Jetpack product discussion

p1694008147802019/1693971459.257519-slack-C4N88L95W

No

## Testing instructions:

* Load this on a Simple sandbox and hit these code paths, see that they work now.

### DIFF
--- a/projects/plugins/jetpack/_inc/lib/class-jetpack-tweetstorm-helper.php
+++ b/projects/plugins/jetpack/_inc/lib/class-jetpack-tweetstorm-helper.php
@@ -1651,17 +1651,25 @@ class Jetpack_Tweetstorm_Helper {
 
 		$requests = array_filter( $requests );
 
-		$hooks = new \WpOrg\Requests\Hooks();
+		// @todo Remove this check when wpcom finally finishes updating to WP 6.2 (apparently they only mostly did so as of yet).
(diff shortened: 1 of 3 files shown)


jeherve (commented):
This looks good to me. I think we could add a link to the p2 post in some of those inline comments, so it's easier for one to find out if WordPress.com simple has been updated. Not a blocker though.

anomiex:
> I think we could add a link to the p2 post in some of those inline comments, so it's easier for one to find out if WordPress.com simple has been updated. Not a blocker though.

That being p94tqY-t5-p2#comment-1414?

Personally, if it's me checking these `@todo` comments in the future then I think I'd be more likely to search the code on wpcom to see if the classes were added yet than to try to track down a P2 about it happening. Too easy for someone to forget to comment on exactly that thread to rely on it.

The P2 link could be useful if we wanted to ping if it still hasn't happened yet at some point in the future, but :shrug: seems easier to just ask the relevant people in Slack or in a new "systems request" post.

ice9js (approved):

PR #32901, opened 2023-09-06
CRM: Resolves #2410 - clean up rewrite rule logic

tbradsha:
Resolves Automattic/zero-bs-crm#2410 - clean up rewrite rule logic

## Proposed changes:

This tackles a long-standing issue where calls to the API would return a 404 until one flushes WP rewrite rules (e.g. goes to WP Settings → Permalinks).

I added an optional parameter to `jpcrm_install_core_extension()` and `jpcrm_uninstall_core_extension()` that takes a `bool` value depending on whether one needs to refresh the rewrite rules or not.

The actual flush runs on the `init` hook, the same as with Client Portal. Down the road we might consider removing `jpcrm_client_portal_flush_rewrite_rules_if_needed()` and replacing `jpcrm_client_portal_set_rewrite_rules_changed()` with `jpcrm_flag_for_flush_rewrite()`, though the latter has implications for Client Portal Pro.

### Other information:

## Testing instructions:

1. Disable API module if enabled: `/wp-admin/admin.php?page=zerobscrm-modules`
2. Save permalinks (or just visit the page): `/wp-admin/options-permalink.php`
3. Enable API module: `/wp-admin/admin.php?page=zerobscrm-modules`
4. Go to `/zbs_api/`

In `trunk`, the page will show a 404.

In the `fix/crm/2410/clean_up_rewrite_rules` branch, the page will show a JSON (error) response as expected.

### DIFF
--- a/projects/plugins/crm/changelog/fix-crm-2410-clean_up_rewrite_rules
+++ b/projects/plugins/crm/changelog/fix-crm-2410-clean_up_rewrite_rules
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+API: rewrite rules are now flushed after enabling module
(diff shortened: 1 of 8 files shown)


gogdzl (approved):
Great job! I like how this adds some improvement while fixing a long standing issue!
LGTM!

PR #3302, opened 2016-01-19
Display Posts Widget: Don't show empty URL error for new instances of the widget

bisko:
Empty URL error was shown for new instances, which were not yet saved. This means that the error was visible from the moment the widget was dragged from the Available Widgets area to the Widget Area.

Now the error checks if the widget has been already saved, before showing the error.

PR #32903, opened 2023-09-06
CRM: Resolves #3302 - old branding leakage

tbradsha:
Resolves Automattic/zero-bs-crm#3302 - old branding leakage

## Proposed changes:

While looking into another issue, I discovered we still had a reference to `ZBSCRM` in the built-in forms. This PR:

* Updates the strings from `ZBSCRM Form` to `Jetpack CRM Form`
* Adds translation to said strings
* Applies coding standards to the relevant files

Note that the three files that are changed could be better abstracted at some point (they're all effectively the same code with one difference) but that's out of scope here.

### Other information:

## Testing instructions:

1. Create a form: `/wp-admin/admin.php?page=zbs-add-edit&action=edit&zbstype=form`
2. Visit the form's iframe URL, e.g. `/crmforms/content/?fid=1`

In `trunk`, the document title (generally shown in the browser tab) will say `ZBSCRM Form`.

In the `fix/crm/3302-old_brand_leakage` branch, the document title says `Jetpack CRM Form`.


gogdzl (approved):
Code looks good, tested all three different form styles, using shortcodes and iframes. It works!

PR #3303, opened 2016-01-19
Display Posts Widget: Properly handle situations where Jetpack is not in `jetpack` plugin directory or `jetpack.php` is under a different name.

bisko:
The Display Posts Widget didn't properly handle situations where `jetpack` and `jetpack.php` could be under different names. This lead to the cron not activating/deactivating when doing such to the Jetpack plugin.

Now these cases are handled by properly detecting Jetpack plugin file name and directory.

PR #30644, opened 2023-05-11
CRM: Resolves 2874 - handle missing transaction status

tbradsha:
Resolves Automattic/zero-bs-crm#2874 - handle missing transaction status

## Proposed changes:

If a transaction was synced with an unofficial status, the transaction status dropdown when editing will show as if the initial status is selected (generally "Succeeded").

This PR adds the unofficial status to the dropdown so that it accurately reflects the current transaction status.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

1. Create a transaction with an unofficial status (e.g. use DevTools to edit the select options).
2. View the transaction editor.

In `trunk` the status will appear to be "Succeeded".

In `fix/crm/2874-handle_missing_transaction_status` the status will be the unofficial status you used.

Issue #1497, opened 2015-01-14
Add more Infinite Scroll jQuery events


ste93cry:
Following request @georgestephanis in issue #1488, I'm opening this ticket to request the `post-load` event in the Infinite Scroll jQuery plugin to be namespaced. It would be useful to add also some more extension points, for example a `pre-load` event or something that allows the editing of the `query_args` from client side before sending the AJAX request. As event data it would be useful to have access to the `Scroller` instance, to edit it's internal variables if needed (like the page number).

georgestephanis:
Anyone know if we have third party integrations to the existing event already that namespacing would break?

jeherve:
We've advertised `post-load` on the support page since the beginning, so there might indeed be quite a few people using it in personal projects out there:
http://jetpack.me/support/infinite-scroll/#js-events

ste93cry:
For fallback support, you could mark as deprecated that event and trigger it twice, one time as before and another time under the appropriate namespace that new events should use too. Then, in a major new version you remove the old trigger. This means that the plugin won't break immediatly the compatibility and users will have time to upgrade their scripts in order to use the new feature

georgestephanis:
I really don't like the idea of breaking backward compatibility without actual cause (not just 'best practices' but the current pattern actually breaking things sufficiently to justify breaking other things while fixing it)

ste93cry:
This is why I suggested to trigger the event twice, before the deprecated event is removed in a major new version.

PR #32909, opened 2023-09-06
CRM: Resolves #3303 - handle invoice status dropdown mismatches

tbradsha:
Resolves Automattic/zero-bs-crm#3303 - handle invoice status dropdown mismatches

## Proposed changes:

If the invoice status in the database doesn't match an available status, the edit page dropdown defaults to the first item on the list, making it appear to be selected.

Quotes don't have real statuses, and contact statuses are not translated. This was already solved for transactions (#30644), but we need to the same for invoices.

This PR does the following:

* If the status is an unexpected one, it appends the status as a dropdown option and selects it.
* WooSync was translating some statuses before saving, and now it should properly save the status in English.
* Along with the above, the invoice object now has a `status_label` key that can be used to output the translated status (e.g. listviews, edit page, client portal, and contact/company profiles).

The larger issue involves the fact that we store labels instead of keys in the database (#1497), which hopefully will be addressed at a future date. This PR begins to lay the groundwork for invoice status keys, though the "key" stored in the database is just the English name for the status at this stage.

Note also that this won't fix any existing errant saves in the database. We haven't had any complaints on this to date, but if it comes up after this is released, the easiest fix is to do a full Woo resync.

### Other information:

## Testing instructions:

1. Change the WP language to Spanish.
2. Create a Woo order and sync it to the CRM.

In `trunk`:
* 👍 It will show as expected in Spanish in the listview and Client Portal.
* 👍 It will show as expected in Spanish in the assigned contact or company profile.
* 👎  The status will save in the database in Spanish.
* 👎  Listview quickfilters don't match statuses.
* 👎 When editing the invoice, the status dropdown will look like it's set to the first status in the dropdown (`Borrador`), whereas it just wasn't able to find a match, so the dropdown defaults to the first item.
* 👎 If you set the invoice to a non-existent status via the database, the above point applies; the dropdown seems to have the first item selected.
* 👎 If you change the WP language back to English, the status shows in Spanish for the first two points.

In the `fix/crm/3303-fix_invoice_status_dropdown_mismatches` branch:
* 👍 It will show as expected in Spanish in the listview and Client Portal.
* 👍 It will show as expected in Spanish in the assigned contact or company profile.
* 👍 The status will save in the database in English.
* 👍  Listview quickfilters match statuses.
* 👍 When editing the invoice, the status dropdown will properly select the current status.
* 👍 If you set the invoice to a non-existent status via the database, the dropdown shows said status as selected.
* 👍 If you change the WP language back to English, the status shows in English everywhere.

### DIFF
--- a/projects/plugins/crm/modules/woo-sync/includes/class-woo-sync.php
+++ b/projects/plugins/crm/modules/woo-sync/includes/class-woo-sync.php
@@ -265,11 +265,11 @@ class Woo_Sync {
 			);
 
 			if ( in_array( $order_status, $paid_statuses, true ) ) {
-				$status = __( 'Paid', 'zero-bs-crm' );
+				$status = 'Paid';
 			} elseif ( $order_status === 'checkout-draft' ) {
-				$status = __( 'Draft', 'zero-bs-crm' );
+				$status = 'Draft';
 			} else {
-				$status = __( 'Unpaid', 'zero-bs-crm' );
+				$status = 'Unpaid';
 			}
 		} elseif ( $obj_type_id === ZBS_TYPE_TRANSACTION ) {
 			// note that transaction statuses aren't translated, as they're user-configurable
--- a/projects/plugins/crm/admin/company/view.page.php
+++ b/projects/plugins/crm/admin/company/view.page.php
@@ -588,7 +588,7 @@ item"><?php esc_html_e( 'Tasks', 'zero-bs-crm' ); ?></div><?php } ?>
 
 										$invoice_val = $invoice['total'];
 
-										$invoice_status = $invoice['status'];
+										$invoice_status = $invoice['status_label'];
 
 										echo '<tr>';
 										echo '<td><a href="' . esc_url( $invoice_url ) . '">' . esc_html( $id_ref_str ) . '</a></td>';
--- a/projects/plugins/crm/admin/contact/view.page.php
+++ b/projects/plugins/crm/admin/contact/view.page.php
@@ -876,7 +876,7 @@ item"><?php esc_html_e( 'Tasks', 'zero-bs-crm' ); ?></div><?php } ?>
 
 									$invoice_val = $invoice['total'];
 
-									$invoice_status = $invoice['status'];
+									$invoice_status = $invoice['status_label'];
 
 									echo '<tr>';
 									echo '<td><a href="' . esc_url( $invoice_url ) . '">' . esc_html( $id_ref_str ) . '</a></td>';
--- a/projects/plugins/crm/changelog/fix-crm-3303-fix_object_status_dropdown_mismatches
+++ b/projects/plugins/crm/changelog/fix-crm-3303-fix_object_status_dropdown_mismatches
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Invoices: Handle status translations consistently
--- a/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
@@ -5427,9 +5427,9 @@ function zeroBSCRM_AJAX_sendInvoiceEmail_v3( $email = '', $invoiceID = -1, $atta
 
 			// once the invoice is sent it will mark it as unpaid (automatically)
 			// (if is draft)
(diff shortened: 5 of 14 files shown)


projects/plugins/crm/modules/woo-sync/includes/class-woo-sync.php
```diff
@@ -265,11 +265,11 @@ public function get_default_status_for_order_obj( $obj_type_id, $order_status )
...
 			if ( in_array( $order_status, $paid_statuses, true ) ) {
-				$status = __( 'Paid', 'zero-bs-crm' );
+				$status = 'Paid';
 			} elseif ( $order_status === 'checkout-draft' ) {
-				$status = __( 'Draft', 'zero-bs-crm' );
+				$status = 'Draft';
 			} else {
-				$status = __( 'Unpaid', 'zero-bs-crm' );
+				$status = 'Unpaid';
 			}
```
gogdzl, line 273: Don't we need to add some kind of migration to be backwards compatible?

gogdzl: We have this problem throughout the whole codebase.
E.g.:
```
includes/ZeroBSCRM.DAL3.Helpers.php

		// retrieves statuses from object :)
		function zeroBSCRM_getInvoicesStatuses(){
		    
		    // for DAL3+ these are hard typed, probably need to sit in the obj:
		    return array(
		    	__( 'Draft', 'zero-bs-crm' ),
		    	__( 'Unpaid', 'zero-bs-crm' ),
		    	__( 'Paid', 'zero-bs-crm' ),
		    	__( 'Overdue', 'zero-bs-crm' ),
		    	__( 'Deleted', 'zero-bs-crm' )
		    );
		}
```

tbradsha: > Don't we need to add some kind of migration to be backwards-compatible?

I mentioned this briefly above; to my knowledge, only WooSync had this bug. A migration gets messy because we have know way of knowing if:

1. The user has changed languages since, or
2. The value in the database is not something they manually added.

Since there's an easy workaround (Woo resync), I think avoiding a migration is best until we actually move to some real key-based status setup.

tbradsha: > We have this problem throughout the whole codebase.

Indeed, my goal here was primarily to address the specific Woo situation, and to make output more i18n-friendly. But I'll poke around and see if there are other cases we can clean things up.

tbradsha: Fixed code block that was meant to set an invoice to "Unpaid" if it was "Draft" and emailed to a user: 56380a3f3d

Because `'Draft' !== __( 'Draft', 'zero-bs-crm' )` if the status has translations, the condition was never met previously.

tbradsha: Fixed invoice bulk action status change which hasn't worked previously on non-English installs: ab85f34521

tbradsha: Fixed invoice bulk action status change which hasn't worked previously on non-English installs: ab85f34521

tbradsha: WooSync status mapping was previously using translated status, but now correctly uses the English status with translated output: 6ebf0c04f4

tbradsha:
@gogdzl I've widened the scope of this PR to cover all invoice statuses I could find, and I think they're all properly handled (along with some bug fixes that have existed since who knows when).

tbradsha:
Rebased, as this has been sitting some time...

gogdzl (approved):
It works and it is a solid move towards fixing our statuses issues with translation.

Great job!

cleacos:
A minor issue with this change. This happens on the WooSync order mapping in sites translated to other languages:

Before the change:

image

After the change:

image

As you can see, any previous mapping will be reset to the default. So if the user had a different mapping (low probability), they will lose it.

tbradsha:
Thanks, @cleacos! I have notes from our call as well and will look into this today to see if it's something we can/should address.

tbradsha:
Handled in #33254.

PR #32910, opened 2023-09-06
AI Assistant: Use backend system prompt

### LINKED ISSUE #28569: Ai Assistant: Move prompt engineering to the backend
Currently, the AI Paragraph block is putting together the prompt based on title, content and categories on the frontend, in JavaScript.

I think there is value in sending payload with:
- post_id
- title
- content
- category ids (no need to fetch names, they already are on the backend)

and have the endpoint (on WPCOM end) put together the prompt.

That way, we could:
- Experiment with different prompt variations without having to release jetpack
- experiment with different models (babbage / ada)

I think this can be achieved by:
1. Keeping everything exactly as it is, just add title, content and categories to the payload
2. Introduce a filter in the Jetpack_Ai_helper class that can override prompt based on the above
3. Hook into this filter on WPCOM to tweak the prompt if needed

lhkowalski:
Part of #28569.

## Proposed changes:

* Start using backend system prompts
* This change uses a `role=jetpack-ai` message to request a generated system prompt from the backend
* The message being sent will be transformed to a system message and a relevant content message, if applicable
* The format of the message is:

```
{
   role: 'jetpack-ai',
   context: {
      type: 'ai-assistant-initial-prompt',
      for: 'the prompt type to which the system prompt will be used, since the type can modify the system prompt'
      content?: 'relevant content for the request, when applicable',
      custom_system_prompt?: 'the play ground system prompt; when set, will override the backend version'
   }
}
```

* The function that generates the message is replacing the current mechanism of creating the prompt but I did not remove the code yet, since we will need a whole clean-up to remove all the dead flows
* This change focus on the system prompts so it's more simple to test and check just this change instead of changing the whole app; follow-up PRs will start enabling backend prompts for the user actions as well
* The code contains some logic to generate user prompt messages as well, but it's not enabled yet and we will have the opportunity to test it on the next PRs

Some lateral changes made:

* Removed the "fake user-assistant" history that was used to set the content when using the block extension, since now the initial message will take care of creating the content history on the backend
* Removed the `contentType: generated` set when transforming extended blocks to the assistant block, since the `generatedContent` value is a source more valid given the new way of handling the relevant content on the backend

### Other information:

## Testing instructions:

* ~~Test in pair with D121131-code~~ backend change got deployed already
* ~~Make sure your test site is pointing to the `public-api`in your sandbox, and your sandbox is running the code above~~
* Go to the block editor and open the browser console
* Enable debug messages by running the following command; refresh the window after it:

```
localStorage.setItem( 'debug', '*' );
```

* Insert an AI Assistant block and ask for some completion (example: a haiku about code)
* Notice the `jetpack-ai-client:ask-question` log message
   * the message will follow the new format
   * if the post has some content, confirm the correct relevant content is present on the message `context.content` property
   * confirm that the `context.for` property contains the right target action

Screenshot 2023-09-08 at 18 54 14

* Confirm there is no `role=system` message
* Confirm the request works as expected:

Screenshot 2023-09-08 at 18 56 33

* Confirm the same happens for all the other actions (simplify, change tone, translate, etc)
* Make sure we don't have regressions, and that you can keep iterating over the content and the `ai-assistant-initial-message` is always present:

Screenshot 2023-09-08 at 18 58 06

* Use the logstash logs to confirm that the prompt being executed on the OpenAI API is correct
* Do the same logtash log check for actions that would modify the system prompt, like the "Simplify" that would add three more rules to it:

Screenshot 2023-09-08 at 18 51 42

Screenshot 2023-09-08 at 19 01 27

Screenshot 2023-09-08 at 19 02 24

* Confirm that you can set a custom system prompt on the client playground and that custom prompt will take precedence over the backend version:

Screenshot 2023-09-08 at 19 03 58

Screenshot 2023-09-08 at 19 04 45

Screenshot 2023-09-08 at 19 05 19


CGastrell (approved):
Everything worked as described, both on logs and resulting responses. I'm approving, but I think being so new to the code as I am, a second look should be in order. At your consideration 👍

dhasilva (approved):
LGTM

lhkowalski:
Let's keep this as do not merge for now, in favor of #33018.

lhkowalski:
Closing in favor of #33018.

PR #32912, opened 2023-09-06
Add meta to subscription form GET parameters.

n3f:
Required by https://github.com/Automattic/gold/issues/143.

## Proposed changes:
* Adds the post metadata added in #32869 to the iframe GET parameters.

## Testing instructions:

* On clicking "subscribe" you can check the developer tools iframe loaded that it has the post_id and tier_id GET params.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/view.js
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/view.js
@@ -19,6 +19,12 @@ domReady( function () {
 		form.payments_attached = true;
 		form.addEventListener( 'submit', function ( event ) {
 			const email = form.querySelector( 'input[type=email]' ).value;
+			const email_clause = email ? `&email=${ encodeURIComponent( email ) }` : '';
+			const post_id = form.querySelector( 'input[name=post_id]' )?.value;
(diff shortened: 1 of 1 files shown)


n3f (commented):
self approval, this was pretty small and I think @millerf wants to merge tomorrow.

projects/plugins/jetpack/extensions/blocks/subscriptions/view.js
```diff
@@ -19,6 +19,12 @@ domReady( function () {
 		form.payments_attached = true;
 		form.addEventListener( 'submit', function ( event ) {
 			const email = form.querySelector( 'input[type=email]' ).value;
+			const email_clause = email ? `&email=${ encodeURIComponent( email ) }` : '';
+			const post_id = form.querySelector( 'input[name=post_id]' )?.value;
+			const post_id_clause = post_id ? `&post_id=${ post_id }` : '';
+			const tier_id = form.querySelector( 'input[name=tier_id]' )?.value;
+			const tier_id_clause = tier_id ? `&tier_id=${ tier_id }` : '';
```
millerf, line 26: If we have the post_id, let's extract the tier from there in the backend, instead of passing it around...

PR #32922, opened 2023-09-07
Social: Disable `Share as a social post` if there is no media

gmjuhasz:
In this change I added a simple logic that disables the share as a social post checkbox if there is no media on the post. If there is a featured image, an attached media, or SIG enabled, it won't be disabled anymore.

## Proposed changes:

* Added a new prop to the component to be able to disable it.
* Piped it back up to the form where we have all the data about media.

### Other information:

## Jetpack product discussion

https://github.com/orgs/Automattic/projects/733/views/2?pane=issue&itemId=37830733

No.
## Testing instructions:

* Open up the editor. On a new post with SIG off, 'Share as a social post' checkbox should be disabled, and the text below should describe why is that. Do any of the following to turn off the disabling:
  * Enable SIG
  * Add a featured image
  * Add any attached media

CleanShot 2023-09-07 at 14 26 15 png


projects/js-packages/publicize-components/src/components/social-post-control/index.js
```diff
@@ -1,16 +1,17 @@
...
 import styles from './styles.module.scss';
 
 /**
  * The component that handles the social post checkbox.
  *
+ * @param {boolean} disabled -  Indicates whether the checkbox is disabled or not.
  * @returns {object} The SocialPostControl Component.
  */
-export default function SocialPostControl() {
+export default function SocialPostControl( { disabled = false } ) {
```
manzoorwanijk, line 24: We don't necessarily need a default value here, do we?

manzoorwanijk (dismissed):
Looks good

projects/js-packages/publicize-components/src/components/social-post-control/index.js
gmjuhasz: Yeah true, I thought it's easier to read like this, but I can remove that, what do you think?

manzoorwanijk: No, nothing big. It's fine either way.

manzoorwanijk (approved):

PR #32927, opened 2023-09-07
AI Excerpt: do not disable Generate button when there is a request error

retrofox:
When there is an error in the suggestion response, the Generate button should not be disabled. This PR checks the request status and don't disable the button when it's `error`.

Fixes #

## Proposed changes:

* AI Excerpt: do not disable Generate button when there is a request error

### Other information:

## Testing instructions:

* Go to the block editor
* Ensure your post has content
* Open the block sidebar
* Identify the AI Excerpt panel
* Try to generate bad conditions to generate a response error. For instance, disable the network 

Screenshot 2023-09-07 at 17 35 41

* Request an excerpt suggestion
* Take a look at the jetpack-ai-query request. Confirm content has the current content of the post

Before (trunk) | After (this PR)
------|------
Screenshot 2023-09-07 at 17 36 59 | Screenshot 2023-09-07 at 17 36 08

### DIFF
--- a/projects/plugins/jetpack/changelog/update-ai-excerpt-do-not-disable-when-error
+++ b/projects/plugins/jetpack/changelog/update-ai-excerpt-do-not-disable-when-error
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
+AI Excerpt: do not disable Genertate button when there is a request error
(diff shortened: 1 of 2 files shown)


nunyvega (approved):
Works well!

Issue #32914, opened 2023-09-07
Docker Feature Request: Easy ability to change php versions for unit tests


kraftbj:
### Impacted plugin

None / Other

### What

Easily switch versions of php for the Docker container

### How

Right now, if I wanted to switch php versions in Docker, I believe the best path is `jetpack docker sh` and use `apt install` to install a new version of php and required libraries.

It would be nice if there was a way to do something like `jetpack docker php 8.2` or `jetpack docker config --php=8.2` or whatever to switch between versions of PHP.

In addition to simply installing the basic `php8.2` deb package, it would need to install the libraries that WordPress requires (e.g. `php8.2-xml php8.2-mbstring php8.2-mysql`).

ice9js:
Hmm... I think a `--php=*` option is probably the way to go. Just going by docker's best practices, I believe it's better to have a separate image for each PHP version instead of dynamically modifying an existing container. It also makes it easier to keep track of all the other little bits you mentioned.

Just thought I'd note it down for when we come back to it.

kraftbj:
That makes sense. Do you think it'd make sense to see if possible to pull in existing php containers and separate the WordPress application from the php layer?

Of course, if an existing container doesn't work for us, we can grow our own.

ice9js:
In general yes, though looking at it right now, looks like we currently just have one image with everything on it. There is nothing inherently wrong with that, but I think it'd be good start versioning it, that way you can always pull what you need.

Official WordPress images would be a good example, with tags like `6.3.1-php8.0-apache`, but you can see how that quickly that can become an issue itself. I mean, let's just take that and throw node with it's own version into the mix. Since ours is specifically meant for development, maybe we can get away with being less explicit and 'remembering' which image has what versions installed.

Another solution would be to have all PHP versions installed simultaneously which I think you can do. And we could maybe dynamically assign `php` to the required one.

anomiex:
> Do you think it'd make sense to see if possible to pull in existing php containers and separate the WordPress application from the php layer?

I'm not sure what this means. But it's unlikely there are any existing base images that have PHP, Node, and pnpm all installed (and all at versions we want).

Building additional images isn't particularly hard, just time-consuming. Particularly the ARM-compatible images for Mac users.

> Another solution would be to have all PHP versions installed simultaneously which I think you can do. And we could maybe dynamically assign php to the required one.

[...]

PR #32929, opened 2023-09-07
docker: Add script for selecting different PHP versions

anomiex:
## Proposed changes:

`jetpack docker select-php` (or `select-php` inside the container) is now avalable to install and run a different version of PHP inside the container.

This is something of a proof of concept. See the addition to `tools/docker/README.md` for some important caveats.

### Other information:

## Jetpack product discussion

#32914

No

## Testing instructions:

* Temporarily change https://github.com/Automattic/jetpack/blob/2fe018f5e7fb3da1fd8c808b3fcdf8654a95ca32/tools/docker/docker-compose.yml#L37 to refer to `ghcr.io/automattic/jetpack-wordpress-dev:pr-32929` and restart your docker env (`jetpack docker stop && jetpack docker up -d`).

### DIFF
--- a/tools/cli/commands/docker.js
+++ b/tools/cli/commands/docker.js
@@ -328,6 +328,8 @@ const buildExecCmd = argv => {
 		opts.push( 'bash' );
 	} else if ( cmd === 'db' ) {
 		opts.push( 'mysql', '--defaults-group-suffix=docker' );
+	} else if ( cmd === 'select-php' ) {
+		opts.push( 'select-php', argv.version );
 	} else if ( cmd === 'phpunit' ) {
 		// @todo: Make this scale.
 		console.warn( chalk.yellow( 'This currently only run tests for the Jetpack plugin.' ) );
@@ -622,6 +624,18 @@ export function dockerDefine( yargs ) {
 					command: 'sh',
 					description: 'Access shell on WordPress container',
 					builder: yargExec => defaultOpts( yargExec ),
+					handler: argv => execDockerCmdHandler( argv ),
+				} )
+				.command( {
+					command: 'select-php <version>',
+					description:
+						'Select the version of PHP for use inside the container. See documentation for important notes!',
+					builder: yargCmd => {
+						yargCmd.positional( 'version', {
+							describe: 'The version to select, or "default".',
+							type: 'string',
+						} );
+					},
 					handler: argv => execDockerCmdHandler( argv ),
 				} )
 				.command( {
--- a/tools/docker/README.md
+++ b/tools/docker/README.md
@@ -212,6 +212,25 @@ wp> get_bloginfo( 'name' );
 
 Note that each `wp shell` session counts as a single request, causing unexpected situations with WP cache. You might want to run [`wp_cache_flush()`](https://developer.wordpress.org/reference/functions/wp_cache_flush/) between requests you expect to get cached by WordPress.
 
+### Changing PHP versions
+
+You can select different versions of PHP. For example, to use PHP 8.0 inside the container:
+
+```sh
+jetpack docker select-php 8.0
+```
+
+If you're already inside the container after using `jetpack docker sh`, you can use `select-php 8.0` there too.
+
+Note some caveats:
+
+* Running `jetpack docker down` or otherwise recreating the containers will reset to the default PHP version.
+* If you're wanting the new version of PHP to be used to serve web requests, you'll need to `jetpack docker stop && jetpack docker up -d` or the like.
+* You may need to update Composer packages from inside the container before you can successfully run `phpunit` or the like, in order to install a version compatible with the new version of PHP.
+  * For example, `jetpack docker exec -- composer -d /usr/local/src/jetpack-monorepo/projects/plugins/jetpack update`
+  * Be careful not to commit any resulting changes to `composer.lock` files!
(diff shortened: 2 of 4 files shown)


tools/docker/README.md
```diff
@@ -212,6 +212,25 @@ wp> get_bloginfo( 'name' );
...
+jetpack docker select-php 8.0
+```
+
+If you're already inside the container after using `jetpack docker sh`, you can use `select-php 8.0` there too.
+
+Note some caveats:
+
+* Running `jetpack docker down` or otherwise recreating the containers will reset to the default PHP version.
+* If you're wanting the new version of PHP to be used to serve web requests, you'll need to `jetpack docker stop && jetpack docker up -d` or the like.
+* You may need to update Composer packages from inside the container before you can successfully run `phpunit` or the like, in order to install a version compatible with the new version of PHP.
```
ice9js, line 229: I have a couple of questions here:

- Do we expect anything else other than PHPUnit or its dependencies to cause issues?
- If so, it seems to me these are potentially compatibility issues that need addressing.
- For phpunit itself, maybe it would be interesting to consider using PHAR releases instead of composer. That way we can also have multiple versions installed if need be.

tools/cli/commands/docker.js
```diff
@@ -328,6 +328,8 @@ const buildExecCmd = argv => {
 		opts.push( 'bash' );
 	} else if ( cmd === 'db' ) {
 		opts.push( 'mysql', '--defaults-group-suffix=docker' );
+	} else if ( cmd === 'select-php' ) {
```
ice9js, line 331: While switching the default version on the entire PHP container could be useful when some manual debugging is required, I'm thinking the most common use case for this will be to run unit tests.  
Maybe we could also expand the `jetpack docker phpunit` command with a `--php=` option like so:

```
jetpack docker phpunit /some/path/to/some/tests --php=7.1
```

What I find particularly appealing about this is the given version would only affect the single command, so you don't have to worry about switching back to what you had before.

tools/cli/commands/docker.js
```diff
@@ -624,6 +626,18 @@ export function dockerDefine( yargs ) {
 					builder: yargExec => defaultOpts( yargExec ),
 					handler: argv => execDockerCmdHandler( argv ),
 				} )
+				.command( {
+					command: 'select-php <version>',
+					description:
+						'Select the version of PHP for use inside the container. See documentation for important notes!',
+					builder: yargCmd => {
+						yargCmd.positional( 'version', {
+							describe: 'The version to select, or "default".',
```
ice9js, line 635: Hmm, can we maybe put a list of supported version numbers in there? I personally find little things like that quite helpful, even if it's just to check the expected format. It's a small list anyway.

Adding some basic validation on that would be cool too.

ice9js (commented):
Thanks Brad! Personally I'm not a huge fans of scripts or tools that cause you to make a mess and don't clean up after themselves, so that's where I'm coming from with my comments. I'm hoping maybe some of these adjustments will help with that.

Let me know what you think.

tools/docker/README.md
anomiex: > Do we expect anything else other than PHPUnit or its dependencies to cause issues?

Some of changelogger's Symfony deps probably will too. At least some PHPCS sniffs may crash with particularly old PHP as well.

> If so, it seems to me these are potentially compatibility issues that need addressing.

There's nothing we can do about upstream projects deciding to drop support for older versions of PHP more aggressively than we do.

> For phpunit itself, maybe it would be interesting to consider using PHAR releases instead of composer. That way we can also have multiple versions installed if need be.

That would make it harder for everything outside of the docker container to run phpunit, as then they'd have to download the appropriate phar instead of letting Composer handle it.

We already have things set up so Composer can choose the correct version of phpunit to install based on the PHP version. The problem is that Composer has to be run with that PHP version to do that, and that `composer.lock` can only lock one version (and checking that into the repo is [usually recommended](https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control)).

tools/cli/commands/docker.js
anomiex: Unfortunately it's not that simple considering the need for re-running composer to get a compatible version of phpunit (and maybe other deps, like if we keep the use of `johnkary/phpunit-speedtrap`), and in some cases upgrading or downgrading composer to get a compatible version of that.

We could certainly try something that would only work for `jetpack docker phpunit` if we want. It'd still need a script like up to line 60 of the one in this PR, and we might avoid screwing up the monorepo `vendor/` by installing PHPUnit and whatever other deps out-of-tree (with a composer.json making use of `.config.platform.php` to get the right versions) and running from there instead.

tools/cli/commands/docker.js
anomiex: Not here without manually hard-coding them. It depends on which versions are available from https://deb.sury.org/.

I didn't bother with validation here since the exec'ed command itself does what would be the same validation.

tools/cli/commands/docker.js
anomiex: > We could certainly try something that would only work for `jetpack docker phpunit` if we want.

Proof of concept for that is in #32979.

anomiex:
Since this probably breaks too much other stuff, I think it's unlikely we'll go ahead with this. https://github.com/Automattic/jetpack/pull/32979 happened for the more limited case of `jetpack docker phpunit --php`.

Issue #32401, opened 2023-08-10
Block discoverability: allow a block to be inserted even when its associated module is disabled


jeherve:
Jetpack currently loads about 70 blocks and plugins in the editor. A small portion of those blocks aren't immediately available though. If the feature is tied to a Jetpack module (as available in `projects/plugins/jetpack/modules`), we've typically not loaded the block until the associated module was activated.

We've typically done that by placing the code that was registering the block in files that are only loaded when the module is active, or by adding a `is_module_active` check`. Here is an example:
https://github.com/Automattic/jetpack/blob/4acb09fdd8653e5c0c0b2f414bb9257d13cb43a0/projects/plugins/jetpack/extensions/blocks/wordads/wordads.php#L33-L59

While this make sense, it's also not ideal for site owners:

- Someone may have heard / read about Jetpack offering a specific feature, but may not find it in their block editor because the module isn't active yet.
- Someone may be looking for a specific block, and looking for a plugin that will help them even though they wouldn't have to look for a new plugin if they knew the Jetpack plugin that's already installed on their site offers that functionality.

Ideally, we would offer all the blocks available in Jetpack, **with a prompt to activate the feature when necessary.** This is something we've already implemented with the Subscriptions block:

<img width="715" alt="image" src="https://github.com/Automattic/jetpack/assets/426388/3951a57c-56fa-460f-abdd-c7d09a1b383b">
<img width="296" alt="image" src="https://github.com/Automattic/jetpack/assets/426388/ab0fc71c-203d-410b-9f7a-5e5c0dd7d365">

Let's aim to offer a similar experience for the other features that rely on modules, listed below.

This would also address the problem of obscure error messages when a block is used on a site where the module was disabled: #11903 

### Implementation 

In #29044, we've introduced the `useModuleStatus` hook in `@automattic/jetpack-shared-extension-utils`, that allows us to turn a module on from the block editor. This can be used for the modules below.

While for some of those modules the change should be fairly straightforward, for others we'll need to watch out for any meta data that may be created on module activation.

### Block List

```[tasklist]
### Tasks
- [x] **Subscriptions** (block & plugin) #29044
- [x] **Tiled Gallery** (block) #13471
- [x] **Related Posts** (block) #11861
- [x] **Publicize** (plugin) #12367
- [x] **Contact Form** (block) #10903
- [ ] https://github.com/Automattic/jetpack/issues/33161
- [ ] https://github.com/Automattic/jetpack/issues/33162
[...]

PR #32930, opened 2023-09-07
Related Posts Block: make the block available even when the module is disabled

### LINKED ISSUE #11861: Related Posts Block: Available in Block Picker even if Disabled
#### Steps to reproduce
1. On a WP.com test site, start a new post in the block editor, and add a 'Related Posts' block.
2. Publish and view the post to verify that the block is present on the frontend.
3. At `https://wordpress.com/settings/traffic/<yourSite>`, disable the Related Posts toggle
4. Reload the published posts, and verify that the Related Posts block is no longer visible
5. Edit the post again in Calypso. 

#### What I expected

The block to be disabled (showing an error that it's unavailable), and to be missing from the block picker

#### What happened instead

The block is still there (displaying example content), and is still shown in the block picker

#### Context

In Jetpack, this works as expected.

I'm guessing something is wrong with block registration.

/cc @tyxla 

#### Primary issue

#32401

jeherve:
Fixes #11861

## Proposed changes:

This PR changes the behavior of the Related Posts block when the feature is not available.

### Current behavior

**On WordPress.com Simple**

On simple, you're able to disable the Related Posts feature via the `enabled` option of the `relatedposts` Jetpack option. This setting is available under Settings > Reading:

`https://yourdotcomsimplesite.com/wp-admin/options-reading.php`

When this happens, the block is still available in the editor but will not work properly, as outlined in #11861.

**On self-hosted sites**

On self-hosted sites (WoA and Jetpack), you can disable the Related Posts module, under Jetpack > Settings > Traffic

In this scenario, the block isn't registered in the editor.

### Future behavior

On both environments, the Related Posts block will remain registered and available. When the feature or the module will be disabled:

- We do not display anything on the frontend.
- In the block and site editors, we display a placeholder with a button inviting you to activate the feature. The placeholder contents and the support links should be a bit different whether you're on WordPress.com simple or self-hosted.

Screenshot 2023-09-15 at 17 44 14

- When the feature is enabled (either the module or the option), we display the block contents just like before.

### Other information:

## Jetpack product discussion

* See #32401

* This PR introduces new Tracks events when the placeholder is displayed, and when one enables the feature via the placeholder button.

## Testing instructions:

**This needs to be tested on simple and on self-hosted.** To test this, it's best if you use a site with a lot of existing content, so related posts can be displayed.

### On WordPress.com Simple

> **Note**
> You will need to apply this diff, as well as D121657-code to bring in the new endpoint.

1. Start a new post in the block editor, and add a 'Related Posts' block.
2. Publish and view the post to verify that the block is present on the frontend.
3. At https://wordpress.com/settings/traffic/<yourSite>, disable the Related Posts toggle
4. Reload the published posts, and verify that the Related Posts block is no longer visible
5. Edit the post again ; the block should be there, but with a placeholder content.
6. Click on the button to enable related posts back.
7. Verify that the related posts appear again on the frontend.

### On self-hosted

1. Go to Jetpack > Settings > Traffic
2. Disable the Related Posts module
3. Go to Posts > Add New
4. Add some content and save your draft.
5. Add a Related Posts Block
6. See the placeholder content
7. Enable the module from the button.
8. Publish and verify that you can see the related posts on the frontend.


projects/plugins/jetpack/extensions/blocks/related-posts/test/edit.js
```diff
@@ -54,29 +55,59 @@ const posts = [
...
+	__esModule: true,
+	...jest.requireActual( '@automattic/jetpack-shared-extension-utils' ),
+	useModuleStatus: jest.fn().mockReturnValue( {
+		isModuleActive: true,
+		isLoadingModules: false,
+		isChangingStatus: false,
+		changeStatus: jest.fn(),
+	} ),
+} ) );
+jest.mock( '@wordpress/element', () => ( {
```
ice9js: Hmm, it seems mocking modules only affects the current file so this doesn't work and the tests fail. If you delete this and replace it with the following code, all is well: 

```
import { select } from '@wordpress/data';
import { store as editorStore } from '@wordpress/editor';

...

beforeEach( () => {
        ...
        
	jest.spyOn( select( editorStore ), 'getCurrentPost' ).mockReturnValue( {
		[ 'jetpack-related-posts' ]: posts
	} );
} );
```

ice9js (dismissed):

projects/plugins/jetpack/extensions/blocks/related-posts/test/edit.js
anomiex: That replacement does seem more straightforward than what's going on here.

But I think the failure is not due to anything about "only affects the current file", it's because `useSelect` comes from `@wordpress/data` rather than `@wordpress/element`. I think the wording you're looking at from https://jestjs.io/docs/jest-object#jestmockmodulename-factory-options is referring to different _test files_, not to files required by the _current_ test.

I see [extensions/blocks/gathering-tweetstorms/test/editor.js](https://github.com/Automattic/jetpack/blob/74ba77bc25f6cd5879c8996e6c82b46e0bc39c2d/projects/plugins/jetpack/extensions/blocks/gathering-tweetstorms/test/editor.js) successfully mocks `useSelect` by doing it in a different way, for example.

ice9js: I tried changing `@wordpress/element` to `@wordpress/data` first but that wouldn't run at all. I guess if you mocked `useState` directly, maybe that'd work, but I didn't investigate further.

I'd argue mocking the entire `useSelect` is not a great idea as it completely bypasses any logic inside the callback itself (which in this case, there is), potentially leading to uncaught errors, more maintenance or the tests becoming unstable.

jeherve: 0625b7cbaab516dbdeae1ee1d9b47dac5a24fe9f did it. Thank you both!

anomiex:
Seems to work in self-hosted.

On Simple, when adding the block with the feature enabled I get a flash of "You'll need to activate the Related Posts feature" before it switches over to displaying the block. I think your new `useRelatedPostsStatus` needs to differentiate "loading status" from "changing status" like `useModuleStatus` does, so you can use the "loading status" indicator at https://github.com/Automattic/jetpack/blob/ad77b19b12159b7ce5bb8dc9a4d49dfd5de60864/projects/plugins/jetpack/extensions/blocks/related-posts/edit.js#L182 along with that one from `useModuleStatus`.

jeherve:
> On Simple, when adding the block with the feature enabled I get a flash of "You'll need to activate the Related Posts feature" before it switches over to displaying the block.

ah that's a good point, thank you! I just committed 4fda28dfd45e2b301b600ae434dd4edba2532ff5 to help with that.

anomiex (approved):
Seems to work.

sajmes:
On several simple sites, I am seeing the prompt to activate the related posts block in the editor, even when the setting in yoursimplesite.com/wp-admin/options-reading.php for related posts is enabled.

Clicking the activate option in the editor does nothing.

A user reported this across three sites, and I can replicate on a test site. All previously added related posts blocks are no longer working on their site.

I'm not sure if this is related to this, but it seems possible @jeherve 

Ticket: 7012332-zd-a8c

jeherve:
@sajmes Thanks for letting me know. Could you try on your test sites and on the customer's site again? It should work now.

sajmes:
@jeherve  - I checked this and it appears that for existing related post blocks to work, the block needs to be activated from any post on the site. On one of the user's sites, I activated this and the block is displaying related posts on their existing posts. 

On their other site, I did not click the activate option on the related posts block, and the site is still not displaying related posts on posts where that block was previously added.

Can you replicate that? 

Is this something we can automatically activate for existing blocks if so?

jeherve:
As discussed in Slack, I'm thinking that this may happen for all the folks who had added Related Posts blocks to their site without fully activating the feature in Settings > Reading.

To fix this issue, one would have to do the following:

1. Go to Settings > Reading
2. If the Related Posts are set to be displayed, set them to be hidden, and then save changes.
3. When the settings reload, mark the related posts to be displayed again.
4. Save your changes.

That should hopefully fix things.

renata-franco:
Is this issue related to this change?

CleanShot 2023-09-25 at 12 47 51

jeherve:
@renata-franco No, this isn't related. You'll see this grayed out block whenever the block has already been inserted elsewhere on the page, and when the block has a restriction of one block per page.

renata-franco:
> @renata-franco No, this isn't related. You'll see this grayed out block whenever the block has already been inserted elsewhere on the page, and when the block has a restriction of one block per page.

Thank you, Jeremy! Fixed.

PR #32706, opened 2023-08-28
Load the block editor assets in the iframe in WP-Admin

BogdanUngureanu:
Load the assets with enqueue_block_assets hook in WP-Admin and enqueue_block_editor_assets on the front-end.
## Proposed changes:

* Use two hooks to load the assets based on the page.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* Use Jetpack Beta and switch to this branch.
* Go to Post Editor and Site Editor and insert a donations block
* Notice that the block renders correctly on the page.

PR #32932, opened 2023-09-07
Connection: use JS to check if initial state is already rendered

sergeymitr:
We recently modified the Connection package initialization to only render initial state once using a PHP property:
https://github.com/Automattic/jetpack/pull/32499

However, [recently introduced changes](https://github.com/Automattic/jetpack/pull/32706/files#diff-028fe0b6a09c6ed2b0a61f21d8db4d21cf8ca35c8904f94f78503619455fe31fR700) to block rendering highlighted a problem with this approach.
PHP doesn't care about client side, and will not allow re-rendering even if it's actually needed due to the browser using separate environments (e.g. iframe and main window).

This PR will use JavaScript for that purpose. It's less optimal because initial state will occasionally be included into the page multiple times, but there's not much data for that to be an issue.

## Proposed changes:
- Use JavaScript to check if Connection initial state has been rendered instead of PHP variable.

### Other information:

## Jetpack product discussion
p1693495835507849-slack-CDLH4C1UZ

No.

## Testing instructions:
While testing, you might see the `Store "jetpack-connection" is already registered.` error reported in the browser console.
Please ignore that, we'll address that separately.

### Making sure we fixed the problem
1. Create new JN site and checkout Jetpack Bleeding Edge (not this branch).
2. Connect Jetpack.
3. Open the browser console and go to "Appearance -> Editor". Confirm you see the "Initial state is missing" error reported there.
4. Checkout the branch and reload the "Appearance -> Editor" page.
5. Confirm you don't see the error.
6. Use browser console to confirm that global variable `JP_CONNECTION_INITIAL_STATE` is defined and contains valid connection data.
7. Repeat the steps above on a WoA site.

### Making sure we didn't break anything
Open the browser console and confirm the global variable `JP_CONNECTION_INITIAL_STATE` exists on some of these pages (no need to check them all):

1. Jetpack Dashboard
8. Edit a post
9. Jetpack Social
10. Activate Jetpack Social (free is enough), disable Jetpack, and go to the post editor
11. Jetpack Search
12. Jetpack Starter Plugin
13. “Move to WordPress.com”
14. VideoPress
15. My Jetpack
16. Jetpack Protect
17. Disconnect and deactivate Jetpack, activate it again and see the huge “Set up Jetpack” modal.


jeherve (commented):
Things seem to work, but I still get a js error in the block editor; shouldn't this avoid that?

Screenshot 2023-09-08 at 09 59 55

sergeymitr:
Hi @jeherve, thanks for taking a look!

That reflects a fundamental problem with blocks, because they have their own compiled JS, and include their own copy of Connection JS package.
The error means that new store wasn't created, so the existing one gets reused, which is just what we need. It's not supposed to cause any issues.

I have an idea on how we could overcome this, but it would be a bigger project.
And I don't feel comfortable experimenting with smaller workarounds right before the release.

oskosk (approved):
Tests as described

PR #32938, opened 2023-09-08
Social: Limit auto-conversion accepted types

gmjuhasz:
This small change adds the list of Photon accepted mime types, so we only convert images if the uploaded image is from this list.

### Other information:

## Jetpack product discussion

https://github.com/orgs/Automattic/projects/733/views/2?pane=issue&itemId=38038597

No.
## Testing instructions:

* Download a test .gif image, for example [this](https://giphy.com/gifs/xThuWu82QD3pj4wvEQ) one
* On trunk upload is as an attached media. You should see the notice that it will be converted. If you do not see it, you might have dismissed the notice before. Bring it back with `jetpack docker wp option delete jetpack_social_dismissed_notices`
* Now check out this change, and rebuild the plugin. WIth the .gif selected it should state that the media is invalid.


manzoorwanijk (approved):
Works well.

PR #32940, opened 2023-09-08
Jetpack Sync: Drop Sync custom queue table when Jetpack is disconnected

fgiannar:
Drops Sync custom queue table when Jetpack is disconnected

## Proposed changes:

* `Sync\Main`: Add `jetpack_site_disconnected` hook, that will ensure Sync cleanup (drop custom table, reset Sync data etc) when Jetpack is disconnected. This needs to also happen on shutdown to prevent fatals.

### Other information:

## Jetpack product discussion

p9dueE-7e6-p2#comment-10025

No

## Testing instructions:
Pre-requisites: A JP connected site with the Jetpack plugin active

- Go to Jetpack Debugger and enable the Sync custom queue table in the Sync settings section
- In cli run: `wp db tables --all-tables` and confirm `wp_jetpack_sync_queue` table is present
- Deactivate Jetpack
- In cli run: `wp db tables --all-tables` and confirm `wp_jetpack_sync_queue` table is NOT present
- Make sure no fatals occurred

**Repeat the above steps with at least 2 Jetpack plugins present.**

Note: There's an unhandled edge case when 2 Jetpack plugins are present (eg one on this branch - Protect, another on latest stable - Jetpack). In this case, if we first deactivate Protect and then Jetpack, the table will not be dropped. See linked Jetpack product discussion for details.

### DIFF
--- a/projects/packages/sync/changelog/update-sync-drop-table-on-disconnection
+++ b/projects/packages/sync/changelog/update-sync-drop-table-on-disconnection
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Jetpack Sync: Drop Sync custom queue table when Jetpack is disconnected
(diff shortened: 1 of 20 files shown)


zinigor (approved):
This looks good to me, let's get it in before the weekly today so we test it in the beta?

PR #32943, opened 2023-09-08
AI Excerpt: re-enable generate button when options change

retrofox:
This PR re-enables the Generate button when the user changes the AI Excerpt options. For now, it means when it changes the _Number of words_ through the range control

Fixes #

## Proposed changes:

* AI Excerpt: re-enable generate button when options change

### Other information:

## Testing instructions:

* Go to the block editor
* Open the block sidebar
* Look at the AI Excerpt panel
* Request an excerpt suggestion
* Confirm the `Generate` button gets disabled from the requests until the response is onde.
* Confirm that, after change the number or words, the Generate button gets enabled again

https://github.com/Automattic/jetpack/assets/77539/85118264-612a-4494-859b-ac148f5bb616

### DIFF
--- a/projects/plugins/jetpack/changelog/update-ai-escerpt-re-enable-when-options-change
+++ b/projects/plugins/jetpack/changelog/update-ai-escerpt-re-enable-when-options-change
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
+AI Excerpt: re enable generate button when options change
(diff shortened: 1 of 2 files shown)


lhkowalski (dismissed):
It's working as described.

After the generation, the button gets disabled:

Screenshot 2023-09-08 at 16 28 12

If I change the number of words, the button is enabled again:

Screenshot 2023-09-08 at 16 28 25

---

It feel a bit weird to have a `reenable` state. I would prefer to have a state for the enabled status of the button (`isGenerateButtonEnabled` for example), so you could set it to `false` when the requestingState demands it to be inactive, and set it back to `true` when something happen that should activate it. It would be simpler to understand.

What do you think?

I am leaving the approval since it's better to have it working on a sub-optimal way now than not having this feature at all :)

retrofox:
> It feel a bit weird to have a `reenable` state. I would prefer to have a state for the enabled status of the button (`isGenerateButtonEnabled` for example), so you could set it to `false` when the requestingState demands it to be inactive, and set it back to `true` when something happen that should activate it. It would be simpler to understand.
> 
> What do you think?

Using `isGenerateButtonEnabled` sounds good to me.

> I am leaving the approval since it's better to have it working on a sub-optimal way now than not having this feature at all :)

👍

nunyvega (approved):
works well!

PR #32907, opened 2023-09-06
Jetpack CRM Hack Week Emails

robertf4:
Fixes #

## Proposed changes:

*

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* Go to '..'
*

PR #32944, opened 2023-09-08
CRM: Add crm webpack alias

robertf4:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3305

## Proposed changes:

In tsconfig.json there is a "crm/*" path which allows the reference of JS code by direct paths instead of relative paths. e.g. "crm/components/*" instead of "../../*" for component files.

This allowed the TS compiler to understand these paths, but Webpack was not correctly understanding these paths. This PR adds a Webpack alias so that it will understand these paths.

### Other information:

No

## Testing instructions:

Build the JavaScript with `pnpm build` and make sure everything compiles. The actual error only happened at runtime and there is no code in trunk that will currently exhibit this, so there is no functionality to test. This was found and fixed in https://github.com/Automattic/jetpack/pull/32907 so you may run that if you really want to test it.

### DIFF
--- a/projects/plugins/crm/webpack.config.js
+++ b/projects/plugins/crm/webpack.config.js
@@ -120,6 +120,9 @@ const crmWebpackConfig = {
 	},
 	resolve: {
 		...jetpackWebpackConfig.resolve,
+		alias: {
+			"crm": path.resolve(__dirname, "src/js/")	
+		}
 	},
 	node: false,
 	plugins: [
--- a/projects/plugins/crm/webpack.config.js
+++ b/projects/plugins/crm/webpack.config.js
@@ -122,7 +122,7 @@ const crmWebpackConfig = {
 		...jetpackWebpackConfig.resolve,
 		alias: {
 			...jetpackWebpackConfig.resolve.alias,
-			"crm": path.resolve(__dirname, "src/js/")	
+			crm: path.resolve( __dirname, 'src/js/' ),
 		}
 	},
 	node: false,
--- a/projects/plugins/crm/webpack.config.js
+++ b/projects/plugins/crm/webpack.config.js
@@ -121,6 +121,7 @@ const crmWebpackConfig = {
 	resolve: {
 		...jetpackWebpackConfig.resolve,
 		alias: {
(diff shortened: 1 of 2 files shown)


projects/plugins/crm/webpack.config.js
```diff
@@ -120,6 +120,9 @@ const crmWebpackConfig = {
 	},
 	resolve: {
 		...jetpackWebpackConfig.resolve,
+		alias: {
+			"crm": path.resolve(__dirname, "src/js/")	
+		}
```
zinigor, line 126: We're potentially overwriting Jetpack Webpack Config's `alias` property this way, can we instead extend it using something like this: https://github.com/Automattic/jetpack/blob/0bc1ca706c403d13b91ed5fbd9831cb6a6c219d0/projects/packages/search/tools/webpack.instant.config.js#L47 ?

robertf4: Sure, I made the update.

projects/plugins/crm/webpack.config.js
```diff
@@ -120,6 +120,10 @@ const crmWebpackConfig = {
 	},
 	resolve: {
 		...jetpackWebpackConfig.resolve,
+		alias: {
+			...jetpackWebpackConfig.resolve.alias,
+			"crm": path.resolve(__dirname, "src/js/")
```
anomiex: It's unfortunate that crm is globally ignoring eslint. The suggestion it would make here seems useful.
```suggestion
			crm: path.resolve( __dirname, 'src/js/' ),
```

anomiex (commented):
Seems like this will do what's intended. Personally I can't say I'm too fond of having things be confusing about "crm" versus "[crm](https://www.npmjs.com/package/crm)", but if this is what your team wants then I won't try to stop you. I'll leave final approval for your team.

Left a suggestion inline, but feel free to ignore it if you want.

tbradsha (approved):
Could always go with `jpcrm` in `tsconfig.json` and as the alias, but I agree, it's unlikely to be an issue here.

PR #3306, opened 2016-01-19
Merge/slideshow jan2016

dereksmart:
Merge changes from wpcom

cc @lancewillett

PR #32951, opened 2023-09-08
CRM: Resolves #3306 - fix unit test deprecation notices

tbradsha:
CRM: Resolves Automattic/zero-bs-crm#3306 - fix unit test deprecation notices

## Proposed changes:

Due to the use of dynamic properties in some of our classes, we get several deprecation notices when running unit tests on PHP 8.2. This PR makes most of the properties static. The `CRM_Modules` class uses an exception ([`#[\AllowDynamicProperties]`](https://www.php.net/manual/en/class.allowdynamicproperties.php)), as the current setup adds arbitrary modules as properties and refactoring is out of scope.

### Other information:

## Testing instructions:

Run `composer run phpunit` in PHP 8.2

In `trunk`, one will see lots of deprecation messages.

In the `fix/crm/3306-fix_unit_test_deprecation_notices` branch, the deprecation notices are no more.


gogdzl (approved):
LGTM!

PR #32959, opened 2023-09-08
Remove/earn from wpteams

chrismccluskey:
Fixes 141-gh-automattic/gold

## Proposed changes:

* Removed the paywall feature for P2 blogs.

### Other information:

No.

## Testing instructions:

* First, let's recreate the issue. Do this by opening a new post on a P2 blog. The issue is that you can select the paywall block.
* To fix the issue and test this patch, download it to your sandbox and repeat the first step. You should no longer be able to select the paywall block.

### DIFF
--- a/projects/plugins/jetpack/changelog/remove-earn-from-wpteams
+++ b/projects/plugins/jetpack/changelog/remove-earn-from-wpteams
@@ -1,0 +1,4 @@
+Significance: patch
+Type: other
+
+Disable earn features for P2.
--- a/projects/plugins/jetpack/extensions/blocks/donations/donations.php
+++ b/projects/plugins/jetpack/extensions/blocks/donations/donations.php
@@ -21,6 +21,14 @@ const BLOCK_NAME   = 'jetpack/' . FEATURE_NAME;
  * registration if we need to.
  */
 function register_block() {
+	/*
+	 * Disable the feature on P2 blogs
+	 */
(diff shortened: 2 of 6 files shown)


jeherve:
I see we had done that for Subscriptions as well, in #31374. I wonder if it wouldn't be best to keep that logic separated, and have a compatibility file on the p2 side, where we decide where we want to unregister some blocks. This way, we would also be able to do that without making any changes in Jetpack. It would look something like this:

```php
/**
 * Get rid of blocks we do not want.
 *
 * @param array $extensions List of available extensions.
 */
add_filter(
	'jetpack_set_available_extensions',
	function ( $extensions ) {
		$extensions = array_filter(
			$extensions,
			function ( $extension ) {
				return $extension !== 'subscriptions' && $extension !== 'paywall';
			}
		);
		return $extensions;
	}
);
```

What do you think?

chrismccluskey:
> What do you think?

Hello @jeherve! I didn't see your message before I pushed 2bf4e81. I think that's a much cleaner approach. I'll refactor and use that approach instead. Thanks!

chrismccluskey:
It looks like this PR isn't needed.

millerf:
> It looks like this PR isn't needed.

~@chrismccluskey What do you mean?~

Just saw https://github.com/Automattic/gold/issues/141#issuecomment-1716705886 👍

PR #32963, opened 2023-09-10
Launchpad/navigator api endpoint

andres-blanco:
Fixes #

## Proposed changes:

This diff adds the /launchpad/navigator endpoint to fetch data for Launchpad Navigator

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* Apply this branch on your sandbox
* Make a GET request to `/sites/[YOUR_TEST_SITE]/launchpad/navigator`
* Verify the result is correct

PR #32964, opened 2023-09-10
Launchpad/update current checklist

andres-blanco:
Depends on https://github.com/Automattic/jetpack/pull/32963
## Proposed changes:

*

### Other information:

## Testing instructions:

You can test the null value behavior using this snippet in the API Console console:
```
fetch('//public-api.wordpress.com/wpcom/v2/sites/[YOUR_SITE]/launchpad/navigator', {
  method: 'POST',
  headers: {
    'Content-Type': 'application/json',
    // add other headers as required, like Authorization
  },
    credentials: 'include',
  body: JSON.stringify({
    active_checklist_slug: null
  })
})
.then(response => response.json())
.then(data => console.log(data))
.catch(error => console.log('Error:', error));
```

And hardcoding the `can_access` method to always return `true`

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
@@ -719,11 +719,11 @@ function wpcom_launchpad_set_current_active_checklist( $checklist_slug ) {
 		$wpcom_launchpad_config = array();
 	}
 
-	if ( isset( $wpcom_launchpad_config['current_checklist_slug'] ) && $checklist_slug === $wpcom_launchpad_config['current_checklist_slug'] ) {
+	if ( isset( $wpcom_launchpad_config['active_checklist_slug'] ) && $checklist_slug === $wpcom_launchpad_config['active_checklist_slug'] ) {
 		return true;
 	}
 
-	$wpcom_launchpad_config['current_checklist_slug'] = $checklist_slug;
+	$wpcom_launchpad_config['active_checklist_slug'] = $checklist_slug;
 
 	return update_option( 'wpcom_launchpad_config', $wpcom_launchpad_config );
 }
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
@@ -710,9 +710,11 @@ function wpcom_launchpad_get_active_checklist() {
 function wpcom_launchpad_set_current_active_checklist( $checklist_slug ) {
 	$wpcom_launchpad_config = get_option( 'wpcom_launchpad_config' );
 
-	$checklists = wpcom_launchpad_checklists()->get_all_task_lists();
-	if ( ! array_key_exists( $checklist_slug, $checklists ) ) {
-		return false;
+	if ( null !== $checklist_slug ) {
+		$checklists = wpcom_launchpad_checklists()->get_all_task_lists();
+		if ( ! array_key_exists( $checklist_slug, $checklists ) ) {
+			return false;
+		}
 	}
 
 	if ( ! $wpcom_launchpad_config || ! is_array( $wpcom_launchpad_config ) ) {
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
@@ -721,11 +721,17 @@ function wpcom_launchpad_set_current_active_checklist( $checklist_slug ) {
 		$wpcom_launchpad_config = array();
 	}
 
-	if ( isset( $wpcom_launchpad_config['active_checklist_slug'] ) && $checklist_slug === $wpcom_launchpad_config['active_checklist_slug'] ) {
-		return true;
+	if ( null === $checklist_slug ) {
+		if ( ! isset( $wpcom_launchpad_config['active_checklist_slug'] ) ) {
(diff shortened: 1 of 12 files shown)


projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
```diff
@@ -680,6 +680,33 @@ function wpcom_launchpad_set_fse_next_steps_modal_hidden( $should_hide ) {
...
+ * @param string $checklist_slug The slug of the launchpad task list to mark as active.
+ * @return bool Whether the option update succeeded.
+ */
+function wpcom_launchpad_set_current_active_checklist( $checklist_slug ) {
+	$wpcom_launchpad_config = get_option( 'wpcom_launchpad_config' );
+
+	$checklists = wpcom_launchpad_checklists()->get_all_task_lists();
+	if ( ! array_key_exists( $checklist_slug, $checklists ) ) {
+		return false;
+	}
```
daledupreez: I think we should allow `$checklist_slug` to be `null`, and only run this logic if it's not null.
```suggestion
	if ( null !== $checklist_slug ) {
		$checklists = wpcom_launchpad_checklists()->get_all_task_lists();
		if ( ! array_key_exists( $checklist_slug, $checklists ) ) {
			return false;
		}
	}
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
```diff
@@ -680,6 +680,33 @@ function wpcom_launchpad_set_fse_next_steps_modal_hidden( $should_hide ) {
...
+
+	if ( ! $wpcom_launchpad_config || ! is_array( $wpcom_launchpad_config ) ) {
+		$wpcom_launchpad_config = array();
+	}
+
+	if ( isset( $wpcom_launchpad_config['active_checklist_slug'] ) && $checklist_slug === $wpcom_launchpad_config['active_checklist_slug'] ) {
+		return true;
+	}
+
+	$wpcom_launchpad_config['active_checklist_slug'] = $checklist_slug;
```
daledupreez: We should handle `null` here as well.
```suggestion
	if ( null === $checklist_slug ) {
		if ( ! isset( $wpcom_launchpad_config['active_checklist_slug'] ) ) {
			return true;
		}
		unset( $wpcom_launchpad_config['active_checklist_slug'] );
	} else {
		if ( isset( $wpcom_launchpad_config['active_checklist_slug'] ) && $checklist_slug === $wpcom_launchpad_config['active_checklist_slug'] ) {
			return true;
		}
		$wpcom_launchpad_config['active_checklist_slug'] = $checklist_slug;
	}
```

projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad-navigator.php
```diff
@@ -37,10 +37,52 @@ public function register_routes() {
...
 					'permission_callback' => array( $this, 'can_access' ),
 				),
+				array(
+					'methods'             => WP_REST_Server::EDITABLE,
+					'callback'            => array( $this, 'update_navigator_options' ),
+					'permission_callback' => array( $this, 'can_access' ),
+					'args'                => array(
+						'active_checklist_slug' => array(
+							'description' => 'The slug of the checklist to set as active.',
+							'type'        => 'string',
```
daledupreez: As per other comments, this should allow null values.

projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad-navigator.php
```diff
@@ -37,10 +37,52 @@ public function register_routes() {
...
+
+	/**
+	 * Updates Launchpad navigator-related options and returns the result
+	 *
+	 * @param WP_REST_Request $request Request object.
+	 */
+	public function update_navigator_options( $request ) {
+		$input = $request->get_json_params();
+
+		if ( ! isset( $input['active_checklist_slug'] ) ) {
```
daledupreez: This should allow null, which `isset()` may skip.

daledupreez (commented):
I have some comments from code inspection.

projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad-navigator.php
```diff
@@ -37,10 +37,68 @@ public function register_routes() {
...
+	 * @param WP_REST_Request $request Request object.
+	 */
+	public function update_navigator_options( $request ) {
+		$updated = array();
+		$input   = $request->get_json_params();
+
+		foreach ( $input as $key => $value ) {
+			switch ( $key ) {
+				case 'active_checklist_slug':
+					$updated = wpcom_launchpad_set_current_active_checklist( $input['active_checklist_slug'] );
```
daledupreez: Should we not put this into a key/subvalue of `$updated`? I ask because the current code does two things:
1. We change the type of `$update` to be a boolean instead of an array
2. We make the `foreach` loop's behaviour a bit more complicated in cases where we have multiple `$key` values, as we could end up overwriting the type and any existing values when we move through the loop.

daledupreez (approved):
Changes look good to me!

PR #32974, opened 2023-09-11
Required Review Action: retrieve GitHub server URL from context

### LINKED ISSUE #32973: required-review action links to non-existing check on GitHub Enterprise
### Impacted plugin

None / Other

### Quick summary

The required-review GitHub Action works on GitHub Enterprise, however the check that is generated on the pull request has a broken "details" link, since it has the public GitHub URL hardcoded.

### Steps to reproduce

1. Use the action in GitHub Enterprise
2. Click the `details` link next to the check
3. See the 404 returned from github.com

The details link should use the server on which the action was ran as its base, e.g. https://mycompany.com

### What actually happened

The details link uses https://github.com as its base and returns a 404.

### Impact

Some (< 50%)

### Available workarounds?

_No response_

### Logs or notes

_No response_

ldaneliukas:
This is required in order to correctly support GitHub Enterprise usage, otherwise the check detail links are broken.

Fixes #32973

## Proposed changes:
* Retrieve GitHub server URL from the context instead of hardcoding it in as github.com

## Testing instructions:

* Utilize the required-review GitHub Action in any Actions workflow with a sample configuration
* Create a pull request
* Click "Details" next to the Require Review check in the CI section

Example:
```
---
name: Required Review test
on: pull_request

jobs:
  check:
    name: Required Reviews
    runs-on: ubuntu
    steps:
      - uses: Automattic/action-required-review@v3
        with:
          token: ${{ secrets.GITHUB_TOKEN }}
          requirements: |
            - name: All files
              paths:
                - '**'
              teams:
                - test-team
```

No

## Affected projects
GitHub Action for required review


jeherve (dismissed):
Thanks for the contribution!

Could you add the missing info mentioned in the comment above mine?

Thank you.

anomiex:
That's a good catch, thanks!

It  looks like we probably have a similar issue in one of our other actions. Would you mind having a look at that one too?

<details>

These could probably make use of [`$GITHUB_SERVER_URL`](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables):
* https://github.com/Automattic/jetpack/blob/cd5b4408bfdfe0db89ee9da134e4a1b0c723cf68/projects/github-actions/push-to-mirrors/push-to-mirrors.sh#L43
* https://github.com/Automattic/jetpack/blob/cd5b4408bfdfe0db89ee9da134e4a1b0c723cf68/projects/github-actions/push-to-mirrors/push-to-mirrors.sh#L72
* https://github.com/Automattic/jetpack/blob/cd5b4408bfdfe0db89ee9da134e4a1b0c723cf68/projects/github-actions/push-to-mirrors/push-to-mirrors.sh#L74
* https://github.com/Automattic/jetpack/blob/cd5b4408bfdfe0db89ee9da134e4a1b0c723cf68/projects/github-actions/push-to-mirrors/push-to-mirrors.sh#L107

I'm not sure whether the use of `@users.noreply.github.com` to make email addresses (e.g. [here](https://github.com/Automattic/jetpack/blob/cd5b4408bfdfe0db89ee9da134e4a1b0c723cf68/projects/github-actions/push-to-mirrors/push-to-mirrors.sh#L45C82-L45C82)) would want something similar, or what that something might be.

</details>

If you don't want to look, that's ok, just say so. I'll make my best guess in a separate PR and hope someone using GitHub Enterprise will eventually correct it if needed.

ldaneliukas:
> That's a good catch, thanks!
> 
> It looks like we probably have a similar issue in one of our other actions. Would you mind having a look at that one too?
> 
> If you don't want to look, that's ok, just say so. I'll make my best guess in a separate PR and hope someone using GitHub Enterprise will eventually correct it if needed.

I've updated the other areas where the use of `github.com` would impact GitHub Enterprise users, however I don't think the default no-reply email matters here. As I see it is customisable so if needed, users can use that to adjust it to their needs.

projects/github-actions/push-to-mirrors/push-to-mirrors.sh
```diff
@@ -40,7 +40,7 @@ fi
 
 if [[ -z "$COMMIT_MESSAGE" ]]; then
 	MONOREPO_COMMIT_MESSAGE=$(cd "${SOURCE_DIR:-.}" && git show -s --format=%B $GITHUB_SHA)
-	COMMIT_MESSAGE=$( printf "%s\n\nCommitted via a GitHub action: https://github.com/%s/actions/runs/%s\n" "$MONOREPO_COMMIT_MESSAGE" "$GITHUB_REPOSITORY" "$GITHUB_RUN_ID" )
+	COMMIT_MESSAGE=$( printf "%s\n\nCommitted via a GitHub action: %s/%s/actions/runs/%s\n" "$GITHUB_SERVER_URL" "$MONOREPO_COMMIT_MESSAGE" "$GITHUB_REPOSITORY" "$GITHUB_RUN_ID" )
```
anomiex: You inserted the new argument in the wrong place in the list.
```suggestion
	COMMIT_MESSAGE=$( printf "%s\n\nCommitted via a GitHub action: %s/%s/actions/runs/%s\n" "$MONOREPO_COMMIT_MESSAGE" "$GITHUB_SERVER_URL" "$GITHUB_REPOSITORY" "$GITHUB_RUN_ID" )
```

projects/github-actions/push-to-mirrors/CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.1.0] - 2023-09-12
+### Fixed
+- Replaced hardcoded values of github.com with variables to support GitHub Enterprise. [#32974]
+
```
anomiex: Don't update the CHANGLOG.md files directly.

The thing to do is to create a file in `projects/github-actions/push-to-mirrors/changelog/` with appropriate contents. There are tools to help with formatting that file correctly:

* If you've done `pnpm install`, then `pnpm jetpack changelog add` should guide you through it.
* Or, in the `projects/github-actions/push-to-mirrors/` directory, you can do `composer install` and then `composer exec changelogger add`.

projects/github-actions/required-review/CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.2.0] - 2023-09-12
+### Fixed
+- Replaced hardcoded values of github.com with variables to support GitHub Enterprise. [#32974]
+
```
anomiex: As above, don't do this, add a file in `projects/github-actions/required-review/changelog/` instead.

anomiex (changes requested):

projects/github-actions/push-to-mirrors/CHANGELOG.md
ldaneliukas: My bad, I've never actually used `jetpack` and don't have composer installed (since I do nothing with PHP) so updated it manually.

anomiex (commented):
One more thing, I think: we should now add `GITHUB_SERVER_URL` to the list at https://github.com/Automattic/jetpack/blob/77cb1e6004148dbb8f75183bcacbeae87abd8d89/projects/github-actions/push-to-mirrors/push-to-mirrors.sh#L7-L10

anomiex:
Also I see the CI check has flagged that a version number needs to be updated to "3.2.0-alpha" in https://github.com/Automattic/jetpack/blob/77cb1e6004148dbb8f75183bcacbeae87abd8d89/projects/github-actions/required-review/package.json#L3

anomiex (approved):
Let's go for it! Assuming CI is happy this time.

PR #32979, opened 2023-09-11
docker: Add a `--php` option to `jetpack docker phpunit`

### LINKED ISSUE #32914: Docker Feature Request: Easy ability to change php versions for unit tests

anomiex:
Fixes #32914

## Proposed changes:

To make it easier to run tests locally with a specified version of PHP.

This turns out to be a bit more complex than it sounds, as we need to be able to install the necessary version of PHP inside the container and we need to make sure the correct versions of Composer deps (particularly phpunit) are installed too. And then use those rather than whatever is in the monorepo install.

### Other information:

## Jetpack product discussion

https://github.com/Automattic/jetpack/issues/32914
https://github.com/Automattic/jetpack/pull/32929#discussion_r1321337776

No

## Testing instructions:

* Temporarily change https://github.com/Automattic/jetpack/blob/197ba02d06a1ab3a2833f17dd2e5d38ba8474ce6/tools/docker/docker-compose.yml#L37 to refer to `ghcr.io/automattic/jetpack-wordpress-dev:pr-32979` and restart your docker env (`jetpack docker stop && jetpack docker up -d`).
* Then try `jetpack docker phpunit --php=7.0`, or `jetpack docker phpunit-crm --php=7.2`


ice9js (commented):
Overall, I'd say this looks solid for a first pass and would already be a huge improvement. In terms of fine tuning it, I see two opportunities, but we can work on those in separate PRs:

- How about only displaying the headers for PHP/Composer install unless `-v`/verbose mode is set, to make things a tiny bit cleaner? I'm also ok if we don't do this, since you'd only have to go through it once per container.
- Can we rely on PHPUnit's PHAR releases instead of having it as a composer dependency? I mean, I wouldn't mind keeping the latest in composer and having PHARs for the older versions but my concern is that'd create a conflict with the autoloader. The goal here would be to eliminate the need to mess with dependencies/composer for unit testing altogether.

Let me know what you think and I think we can start working on getting it merged :)

anomiex:
> * How about only displaying the headers for PHP/Composer install unless `-v`/verbose mode is set, to make things a tiny bit cleaner? I'm also ok if we don't do this, since you'd only have to go through it once per container.

That could turn out to be confusing, as phpunit also has a `-v` option.

> * Can we rely on PHPUnit's PHAR releases instead of having it as a composer dependency? I mean, I wouldn't mind keeping the latest in composer and having PHARs for the older versions but my concern is that'd create a conflict with the autoloader. The goal here would be to eliminate the need to mess with dependencies/composer for unit testing altogether.

If PHPUnit were the only thing, probably. Although it would be more complex as we'd need to figure out the appropriate PHPUnit version to download from the PHP version instead of letting Composer handle it.

But for Jetpack we also need to get an appropriate version of `johnkary/phpunit-speedtrap`, and at some point for CRM we might need to get an appropriate version of the codeception packages they use, etc.

And we also need to not break when the lock file contains too-new versions of changelogger's symfony deps even though they're almost certainly not used during the tests.

ice9js (approved):
As we already talked about, this seems to be installing dependencies separately for each PHP version so while we can't skip `composer install`, at least it should mess up the project dependencies too much once you're done testing.

Let's get this out :)

PR #32986, opened 2023-09-11
AI Extension: Only show input automatically for empty forms

### LINKED ISSUE #32737: AI Extension: Only show input automatically for empty forms
Otherwise the input will be shown when an user adds a child block directly:

https://github.com/Automattic/jetpack/assets/77539/7f70b825-3d9a-481c-9e66-104069145b8c

dhasilva:
Fixes #32737

## Proposed changes:

* Checks if the block is a form block with no inner blocks to determine if the visibility flag should start as true

### Other information:

## Testing instructions:

* Go to the block editor
* Add a Form block
* Check that the input is displayed immediately
* Separately, add a checkbox (or any form child) directly through the block selector
* Check that the input is not displayed immediately

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-ai-assistant-input-empty-forms
+++ b/projects/plugins/jetpack/changelog/fix-ai-assistant-input-empty-forms
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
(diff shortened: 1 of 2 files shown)


nunyvega (approved):
looks good!

Issue #32950, opened 2023-09-08
AI Chat block: Finish feedback UX


dereksmart:
Right now there's a few things incomplete with the Feedback UX: 
- Response doesn't clear after submitting
- Allows submitting multiple times 
- It looks bad
- The text field is too small 
- Whatever else you can find.

Issue #32947, opened 2023-09-08
AI Chat block: Handle errors gracefully 


dereksmart:
There is basically no error handling in the AI Chat block.

Things should generally fail gracefully and with some design.

PR #32989, opened 2023-09-11
AI Chat: Error handling, separate components, and feedback UX

dereksmart:
Addresses: 
- #32947 
- #32950 

This PR started by adding some simple error handling, which led to a barrage of refactors and tangental UX improvements. The QuestionAnswer component started to get a bit unruly itself. The hope is it will make it easier to improve even more on the UX later. 

image

## Proposed changes:

The main ones are: 
- New components: Feedback, CopyButton, and DisplayError. 
- Adds some basic styling to each component. The feedback area looks more uniform now. 
- Error handling: Added some sanity checks in the endpoint responses, ensuring it sends errors with unexpected responses so the client will have an easier time. Also displays errors. 
- Feedback submission: Doesn't allow multiple submissions for the same question/response (cacheKey). 

## Jetpack product discussion

peGwbA-Be-p2

No. 

## Testing instructions:

Test the errors: 
- Manually introduce an error in the code. Either connect it to your sandbox and malform the response, or do so in the plugin code. 
- Do so for both feedback and asking a question. An error message should appear in red. 

Test the feedback: 
- Submit a question. 
- Leave some feedback about the response, it should submit and you should get a thanks message. 
- Submit the same exact question. It should feed you a cached response and the thank you message should still be there. 
- Submit a new question and the feedback form should be reset, allowing you to submit more. 

Test the copy button: 
- Still works


artpi (approved):
There are some small tweaks we could be doing, but this takes care of the most important issues and tests well!

Issue #32949, opened 2023-09-08
AI Chat block: Configurable prompt


dereksmart:
There should be a UX for configuring the prompt. Either as an override or otherwise more "comprehensive" way in wpcom. 

Right now the override is customizable, but only through a filter `jetpack_ai_chat_answer_prompt`. 

See also peGwbA-BB-p2#comment-1111

artpi:
D121589-code
https://github.com/Automattic/jetpack/pull/32994

dereksmart:
merged #32994

PR #32994, opened 2023-09-12
Chat with a blog: Prompt config

artpi:
Addresses: #32949 

This adds the ability to override a prompt in Jetpack "Chat with a blog" feature. This allows for tweaking the prompt to change the behavior of the chat.

Zrzut ekranu 2023-09-13 o 14 09 21
Zrzut ekranu 2023-09-13 o 14 09 00

NO

## Testing instructions:

1. Check out D121589-code on wpcom sandbox
2. Do `sudo global-http enable` on sandbox
3. Get a Jetpack site, connect to WPCOM
4. Purchase / enable Jetpack Search
6. Point Jetpack sync to your sandbox in `/wp-admin/options-general.php?page=companion_settings`
7. Set JETPACK_BLOCKS_VARIATION to beta
8. Publish a few posts
9. Insert the block on a page
10. **Add additional instructions in the PROMPT OVERRIDE sidebar field (under Advanced)**
11. Publish/save
12. Open the page on the frontend
13. Ask a question relating to your published posts
14. Get a WPCOM site, sandbox public-api AND site domain, apply D121589-code
15. bin/jetpack-downloader test jetpack chat-with-site/wpcom
16. Get Jetpack search product ( click upgrade jetpack search on /wp-admin/admin.php?page=jetpack-search )
17. Publish a few posts
9. Insert the block on a page
10. **Add additional instructions in the PROMPT OVERRIDE sidebar field (under Advanced)**
11. Publish/save
12. Ask a question relating to your published posts

### DIFF
--- a/projects/packages/search/changelog/prompt-config
+++ b/projects/packages/search/changelog/prompt-config
@@ -1,0 +1,4 @@
+Significance: patch
+Type: added
+
+
--- a/projects/packages/search/src/class-settings.php
+++ b/projects/packages/search/src/class-settings.php
@@ -37,6 +37,7 @@ class Settings {
 		// NOTE: This contains significant code overlap with class-jetpack-search-customize.
 		$setting_prefix = Options::OPTION_PREFIX;
 		$settings       = array(
+			array( $setting_prefix . 'ai_prompt_override', 'string', '' ),
 			array( $setting_prefix . 'color_theme', 'string', 'light' ),
 			array( $setting_prefix . 'result_format', 'string', 'minimal' ),
 			array( $setting_prefix . 'default_sort', 'string', 'relevance' ),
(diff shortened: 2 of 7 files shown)


dereksmart:
> IF we should enable it

IMO this functionality is going to be the root expectation for such a feature. I suggest that we add the setting place for now, and have a discussion on how we should approach it and dedicated task for improving the default prompt. For example, maybe it makes sense to have a pre-defined set of prompt overrides to use as a v1.

artpi:
> For example, maybe it makes sense to have a pre-defined set of prompt overrides to use as a v1.

We can also allow people to add additional instruction in this prompt, but the full override would only be possible through filter

dereksmart (approved):
Nice work! It works as expected. My site is talking like a pirate. 

I was thinking it might be nice to store this as post meta. That way folks could set different personalities on different pages. The data would be synced in the same way, with the allowlist in wpcom side just adding the meta field instead of the option. 

Anyway not a blocker. We'll do another pass for copy during the design phase.

PR #33013, opened 2023-09-12
AI Client: check media record ref before removing the listeners

retrofox:
Before removing the function listener, we need to ensure that the media recorder instance has been established. There are various reasons why the instance may be undefined, such as the user failing to provide recording permissions. Failing to make these adjustments will result in the application malfunctioning.

Fixes #

## Proposed changes:

*  AI Client: check media record ref of the useMediaRecording() hook before removing the listeners

### Other information:

## Testing instructions:

* Test with Jetpack plugin 
* Enable `beta` extensions
* Go to the block editor
* Create a Create With Voice block instance
* Start to record an audio
* Ensure the browser asks for permission to record (you can use a new incognito browser instance)
* Confirm you see the browser prompt asking you to record the voice

image

* Before accepting (or not) the permissions, try to delete the block

image

* Before (trunk), confirm the app breaks 

image

image

* After (this PR) confirm the block is removed without any issue.

image

### DIFF
--- a/projects/js-packages/ai-client/changelog/update-create-with-voice-removeing-block-bug
+++ b/projects/js-packages/ai-client/changelog/update-create-with-voice-removeing-block-bug
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+AI Client: check media record ref of the useMediaRecording() hook before to remove the listeners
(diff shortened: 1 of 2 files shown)


nunyvega (approved):
Fixes the issue 
🚢

PR #33015, opened 2023-09-12
AI Assistant: Remove upgrade button when user is not administrator

### LINKED ISSUE #31165: AI Assistant: Reconnect Jetpack button does not work for non-admin users
If the user is not connected to WordPress.com, we prompt they to reconnect:

Image

This button works fine for admin users, but it will fail when the user does not have acccess to the given Jetpack dashboard page:

Image

This is the case for `author` users, for example.

---

It's possible to connect users of any role to WordPress.com, we just need to send them to the rigth connection page.

For authors, for example, the button on the Jetpack dashboard led me to `wp-admin/admin.php?page=jetpack&from=unlinked-user-connect#/connect-user`.

We need to detect which role the user has and show the correct link to them, or then link to a place that will work for any user role.

dhasilva:
Fixes #31165

## Proposed changes:

* Disables the input when the user is not connected
* Change the upgrade nudge for users that are not administrators, removing the upgrade button

### Other information:

## Testing instructions:

* Create a non-admin user, like an Author
* Use the AI Assistant block to spend the 20 free requests
* As the non-admin user, check that the message does not include the upgrade button
* As the admin user, check that the button is displayed

| Before | After
| - | -
| 2023-09-12_18-13-44-before-reconnect | 2023-09-12_18-13-19-after-reconnect
| 2023-09-12_18-12-54-before-author | 2023-09-12_18-11-28-after-author

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-ai-assistant-upgrade-author
+++ b/projects/plugins/jetpack/changelog/fix-ai-assistant-upgrade-author
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
+AI Assistant: Remove upgrade button when user is not administrator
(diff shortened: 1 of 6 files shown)


CGastrell (approved):
Worked as described! :shipit:

PR #33019, opened 2023-09-12
Launchpad: remove checklist from navigator

andres-blanco:
Removes a checklist from the navigator checklist list

## Proposed changes:

*

### Other information:

## Testing instructions:

* Open API console
* Run a PUT/POST on `wpcom/v2` > `/sites/[SITE]/launchpad/navigator`
* Add a slug to the `remove_checklist_slug` body param
* Verify it removes it from the list
* Verify that the correct value is returned in the `new_active_checklist` response parameter.
image

This is what the response should look like:
image

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
@@ -731,32 +731,43 @@ function wpcom_launchpad_navigator_update_checklists( $new_checklists ) {
  * Removes a checklist from the list of checklists that are currently available for the navigator.
  *
  * @param string $checklist_slug The slug of the checklist to remove.
- * @return bool Whether the option update succeeded.
+ * @return array Array with two values: whether the option update succeeded, and the new active checklist slug.
  */
 function wpcom_launchpad_navigator_remove_checklist( $checklist_slug ) {
 	$wpcom_launchpad_config = get_option( 'wpcom_launchpad_config', array() );
 
 	if ( ! isset( $wpcom_launchpad_config['navigator_checklists'] ) ) {
-		return false;
+		return array(
+			'updated'              => false,
+			'new_active_checklist' => null,
+		);
 	}
+
+	$current_active_checklist = wpcom_launchpad_get_active_checklist();
 
 	$checklists = $wpcom_launchpad_config['navigator_checklists'];
 	// Find if $checklist_slug is in the checklists array. If it is, remove it.
 	$key = array_search( $checklist_slug, $checklists, true );
 	if ( $key === false ) {
-		return true;
+		return array(
+			'updated'              => false,
+			'new_active_checklist' => $current_active_checklist,
+		);
 	}
 
 	unset( $checklists[ $key ] );
 
-	$current_active_checklist = wpcom_launchpad_get_active_checklist();
+	$new_active_checklist = $current_active_checklist;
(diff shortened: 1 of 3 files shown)


projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
```diff
@@ -727,6 +727,30 @@ function wpcom_launchpad_navigator_update_checklists( $new_checklists ) {
...
+ *
+ * @param string $checklist_slug The slug of the checklist to remove.
+ * @return bool Whether the option update succeeded.
+ */
+function wpcom_launchpad_navigator_remove_checklist( $checklist_slug ) {
+	$wpcom_launchpad_config = get_option( 'wpcom_launchpad_config', array() );
+
+	if ( ! isset( $wpcom_launchpad_config['navigator_checklists'] ) ) {
+		return false;
+	}
```
daledupreez: Can we add some more defensive code for non-array data in the option?

Also, should we return `true` or `false` if we didn't have a valid list to start with?
```suggestion
	if ( ! $wpcom_launchpad_config || ! is_array( $wpcom_launchpad_config ) ) {
		return false;	
	}

	if ( ! isset( $wpcom_launchpad_config['navigator_checklists'] ) || ! is_array( $wpcom_launchpad_config['navigator_checklists'] ) ) {
		return false;
	}
```

projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad-navigator.php
```diff
@@ -47,6 +47,11 @@ public function register_routes() {
 							'type'              => array( 'null', 'string' ),
 							'validate_callback' => array( $this, 'validate_checklist_slug_param' ),
 						),
+						'remove_checklist_slug' => array(
+							'description' => 'The slug of the checklist to set as active.',
```
daledupreez: This comment is not accurate.
```suggestion
							'description' => 'The slug of the checklist to remove from the active list.',
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
```diff
@@ -727,6 +727,30 @@ function wpcom_launchpad_navigator_update_checklists( $new_checklists ) {
...
+
+	$checklists = $wpcom_launchpad_config['navigator_checklists'];
+	// Find if $checklist_slug is in the checklists array. If it is, remove it.
+	$key = array_search( $checklist_slug, $checklists, true );
+	if ( $key === false ) {
+		return true;
+	}
+
+	unset( $checklists[ $key ] );
+	return wpcom_launchpad_navigator_update_checklists( $checklists );
```
daledupreez: If this removed item is also the active checklist, we need to update the currently active checklist. As per discussion, we should use the most recently active checklist if it exists and is valid, or update it to null otherwise.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
andres-blanco: `$wpcom_launchpad_config` will always be an array since the default value in `get_option` is an array. So checking for `! isset( $wpcom_launchpad_config['navigator_checklists']` should cover both paths.

valterlorran (approved):
LGTM!

PR #33028, opened 2023-09-13
Blocks: remove extra registration of free plugins already available

jeherve:
## Proposed changes:

I'm removing things that seem to be a bit unnecessary for us: those plugins are free, and thus we do not need any conditional loading based on the plan in use on your site.

### Other information:

## Jetpack product discussion

References:

- D121535-code
- 141-gh-gold

* No

## Testing instructions:

You'll want to test this on a free Jetpack site and a free WordPress.com simple site.

In both environments, you'll want to add the following code snippet, either to [a functionality plugin](https://jetpack.com/support/adding-code-snippets/) on a self-hosted site or to the `0-sandbox.php` file in your WordPress.com simple sandbox:

```php
add_filter( 'jetpack_set_available_extensions', function ( $extensions ) {
     l( $extensions ); // note that if you test this in an environment where l() is not available, you should not use it. use something like error_log( print_r()) instead.
     return $extensions;
}, 12 );
```

Then go to Posts > Add New, and ensure that the features are available in the editor.

Behind the scenes in your logs, you should see a difference between this patch and without it: the items we removed in this PR should not be duplicated in the extensions list anymore.


jeherve:
Closing this for now, we can resume this discussion later.

Issue #24290, opened 2022-05-09
Investigate future steps for Lazy Loading


thingalon:
Now that [Safari has native support for lazy loading images](https://webkit.org/blog/12445/new-webkit-features-in-safari-15-4/), we can consider whether our implementation of lazy-load can be scaled back / simplified.

[According to caniuse](https://caniuse.com/loading-lazy-attr), all of the major browsers support lazy loaded images natively.

There are a couple of paths ahead of us to consider:

1. We can test that the `loading=lazy` attribute works as well as our custom Lazy Loading method, and simplify the module to use that, or
2. We can add new features to our Lazy Loading module such as placeholder images and fixed image sizing.

As a first step, we need to explore the above two options. This issue involves:

* Investigate if there is a substantial behavior difference between our (Jetpack's) implementating of Lazy Loading, and the attribute-based implementations in major browsers,
* Test the idea of using Placeholder images in Jetpack's Lazy Loading implementation.
* Post a P2 about the above test so that we can decide on the path ahead.

samiff:
> Consider the impact of switching over.

I think core's implementation for native lazy images should work great for the majority of users, and the main reason at this point to keep around our JS based lazy loading would be to provide additional features - the most compelling being custom image placeholders (blurred image, dominant color, etc.).

jeherve:
> the most compelling being custom image placeholders (blurred image, dominant color, etc.).

That's a good point. If we were to keep our custom implementation around though, we would have to make those potential customizations more prominent, because right now most people wouldn't know how to make use of those.

thingalon:
Blurred placeholders are something that we've wanted to investigate for Boost for a while. It may be worth having that feature available for Boost, but not for Jetpack, to encourage more users to install Boost ;)

jeherve:
Closing in favor of #33048.

PR #33030, opened 2023-09-13
Lazy loading images: Update description on settings toggle to discourage enabling it

oskosk:
See #24290 

## Proposed changes:

This pull request aims to discourage the activation of the JavaScript-based image lazy loading feature from the Performance settings panel. Given that over 90% of modern browsers now support native lazy loading, this feature is no longer a necessary optimization.

The new description will be

_Most modern browsers already support lazy loading. With over 90% of current browsers offering native support, enabling this feature may be unnecessary._

This feature is interfering with the work being done on the Interactivity API

See https://github.com/WordPress/gutenberg/issues/54396

Related https://github.com/Automattic/wp-calypso/pull/81679
### Other information:

## Jetpack product discussion

pdWQjU-wV-p2

* No

## Testing instructions:

* Check the feature description.

### DIFF
--- a/projects/plugins/jetpack/_inc/client/performance/speed-up-site.jsx
+++ b/projects/plugins/jetpack/_inc/client/performance/speed-up-site.jsx
@@ -297,7 +297,7 @@ const SpeedUpSite = withModuleSettingsFormHelpers(
 						>
 							<p>
 								{ __(
-									'Lazy-loading images will improve your site’s speed and create a smoother viewing experience. Images will load as visitors scroll down the screen, instead of all at once.',
+									'Most modern browsers already support lazy loading. With over 90% of current browsers offering native support, enabling this feature may be unnecessary',
 									'jetpack'
 								) }
 							</p>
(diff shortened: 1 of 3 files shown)


jeherve (commented):
Should we also take that opportunity to update https://github.com/Automattic/jetpack/blob/3a3c4adc1c0913bac224ddb832713bd6101fe1ab/projects/plugins/jetpack/modules/lazy-images.php#L4 ?

kraftbj:
I'll add a commit to update the module description too.

kraftbj:
Related: https://github.com/Automattic/jetpack/pull/33039

jeherve (commented):
This looks good to me. I think I may personally prefer if we kept a description of what the module does before we actually suggest not to enable it ; after all, if all we show is why you should not use it, then why is it even in Jetpack to begin with?

I'll let others weigh in on that.

oskosk:
@jeherve I feel you, but explaining here becomes a matter of just narrating history. _Why it worked before, why it was needed._   So it's still a disservice, or confusing in the best case for non-savvy users. For those tech-savvy, the name of the feature is self-explanatory today in 2023. We also have a support page explaining the feature in detail. 

Let's merge this one and see how we evolve later, even if we choose to go for a better description.

oskosk:
Thanks for the commit @kraftbj

kraftbj (approved):

PR #33041, opened 2023-09-13
Jetpack Boost: Refactor MultiProgress component

pyronaur:
Part of the "Refactor Global State" project (pc9hqz-286-p2).

### Other information:

## Jetpack product discussion
N/A

N/A

## Testing instructions:
🦄

### DIFF
--- a/projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/store/isa-summary.ts
+++ b/projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/store/isa-summary.ts
@@ -18,11 +18,13 @@ export enum ISAStatus {
 	Stuck = 'error_stuck',
 }
 
-const zGroup = z.object( {
+const zSummaryGroup = z.object( {
 	issue_count: z.number(),
 	scanned_pages: z.number(),
 	total_pages: z.number(),
 } );
+
+type Summary_Group = z.infer< typeof zSummaryGroup >;
 
 const image_size_analysis_summary = jetpack_boost_ds.createAsyncStore(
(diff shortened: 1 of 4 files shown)


projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/store/isa-summary.ts
```diff
@@ -59,6 +61,29 @@ export const isaGroupLabels = {
 	other: __( 'Other', 'jetpack-boost' ),
 };
 
+export function isaGroupLabel( group: keyof typeof isaGroupLabels | string ) {
```
haqadn, line 64: TIL `keyof typeof`. Btw, isn't `keyof typeof isaGroupLabels | string` equivalent of `string`. I am curious if writing it this way gives any advantage like autocompletion?

haqadn (approved):

haqadn:
Looks much nicer and cleaner 👍🏼

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/store/isa-summary.ts
pyronaur: The purpose is three-fold: 

1. It communicates with the developer reading the code that I'm not just expecting any string, I'm expecting a isaGroupLabels key.
2. IDE Autocomplete is able to pick up on that
3. When a key is stored in a variable, that variable can be just a string not an union, so this avoids unnecessary `as "foo" | "bar"` that would likely be necessary to satisfy this function call.

PR #33038, opened 2023-09-13
reverts d5ca47d8de after finding out it was not working right with P2

CGastrell:
This PR reverts the change introduced in d5ca47d8de as we found out it was not working right on P2s

## Proposed changes:
This is a branch resulting from `git revert d5ca47d8de`

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion
p1694603935108069-slack-C03TA48NSUX

## Does this pull request change what data or activity we track or use?
No

## Testing instructions:
Apply and test using a classic theme. Visit the frontend to see the `jetpack_vendor/automattic/jetpack-videopress/build/block-editor/blocks/video/view.css` is being loaded even if no video block is present.

PR #33036, opened 2023-09-13
VideoPress: load VideoPress block on p2

nunyvega:
Fixes VideoPress block not loading in P2 front-end editor nor comments whenever there are no previous videos in the post or comments, introduced in [#32680](https://github.com/Automattic/jetpack/pull/32680)

## Proposed changes:

* Check if the page loa

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* Apply this PR and use bin/jetpack-downloader to apply it on your dotcom sandbox
* Test it and verify that assets load on P2 but are still skipped in the cases proposed by the original PR

PR #33042, opened 2023-09-13
VideoPress: improve asset enqueuing for P2 and Classic themes

### LINKED ISSUE #32235: VideoPress Block: CSS file is loaded on all pages, even when the block isn't used

CGastrell:
This PR is a re-add for a better VideoPress assets enqueuing mechanism.

The issue was solved on #32680 (d5ca47d8de53df832e67ac8b9d6bda3663c3e8df), but then reverted on #33038 (c9fa94de7886af75b65b8c75e642fb529144eb31) as we discovered an issue where P2s wouldn't load the block.

@nunyvega Added extra logic to the original change on #33036 and this PR is the result of that change.

Fixes #32235 

## Proposed changes:
The change adds an extra check to see if
- the theme is a block theme
- the site is a P2

When the theme is NOT a block theme and NOT a P2, the enqueuing process is deferred to `wp_enqueue_scripts` hook, where the content of the post can be parsed and checked for a VP block and, only then, enqueue the needed assets.

### Other information:

## Jetpack product discussion
p1694603935108069-slack-C03TA48NSUX

No

## Testing instructions:

- [ ] Create a site and connect it to Jetpack. Don't enable VideoPress yet. Visit the site's homepage (or any page), check on the sources to verify jetpack_vendor/automattic/jetpack-videopress/build/block-editor/blocks/video/view.css is NOT being loaded.

- [ ] Change theme to a classic theme, like 2020. Visit a page/post on frontend where NO video blocks are present. See that the mentioned file above is not being loaded. Then visit a page WITH a video block to see the file is being loaded properly.

- [ ] Apply the patch on .com and check on a P2 site. Load some site which holds no videoblock, start a new post (or new comment) and try and add a VideoPress block. Block should be available and working correctly.

### DIFF
--- a/projects/packages/videopress/changelog/change-vp-asset-enqueuing
+++ b/projects/packages/videopress/changelog/change-vp-asset-enqueuing
@@ -1,0 +1,4 @@
+Significance: minor
+Type: changed
+
+Add VideoPress assets enqueuing logic after the revert c9fa94de7886af75b65b8c75e642fb529144eb31 (reverted d5ca47d8de53df832e67ac8b9d6bda3663c3e8df). This time P2s should not be affected.
--- a/projects/packages/videopress/src/class-initializer.php
+++ b/projects/packages/videopress/src/class-initializer.php
@@ -348,6 +348,37 @@ class Initializer {
 			return;
 		}
 
+		// check current theme
+		$is_block_theme = wp_get_theme()->is_block_theme();
+
+		// Check if the site is a P2 site
+		$is_p2_site = function_exists( '\WPForTeams\is_wpforteams_site' ) && \WPForTeams\is_wpforteams_site( get_current_blog_id() );
+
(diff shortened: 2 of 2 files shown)


lhkowalski (approved):
Code looks good.

I tested on a Jetpack site. When using the 20-23 theme (a block theme), the enqueueing is handled correctly, both on your change and on trunk:

- no video block, no style file
- video block present, style file present

When using 20-20 (a classic theme), then your code fixes the enqueueing logic.

- If the page does not have a video, the style is not present:

Screenshot 2023-09-14 at 14 21 18

Screenshot 2023-09-14 at 14 26 46

- If the page has a video, the style is present:

Screenshot 2023-09-14 at 14 24 26

Screenshot 2023-09-14 at 14 24 34

---

I also tested on a P2 site. Given the logic in your code, the style will always be present, right? 

- I tested on a page without a video and the style was there
- I tested on a page with a video, and the style was there as well

PR #33047, opened 2023-09-13
Subscribe Block: Add optional login button

lezama:
## Proposed changes:

Screenshot 2023-09-13 at 15 05 10

See p1694613846242699-slack-C052XEUUBL4

* Adds "Switch Account" when user is already logged and the post is paid.
* Moves the login link to the subscribe block to simplify the paywall template logic.
* You should be able to show login buttons by adding the showLogin attribute to the jetpack/subscribe block

```
<!-- wp:jetpack/subscriptions {"borderRadius":50,"borderColor":"primary","className":"is-style-compact","showLogin": true} /-->
```

### Other information:

## Testing instructions:

* Check the paywall while logged in as a free subscriber on a paid post.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
+++ b/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
@@ -48,15 +48,9 @@ abstract class Token_Subscription_Service implements Subscription_Service {
 	 * How to obtain one of these (or what exactly it is) is
 	 * still a WIP (see api/auth branch)
 	 *
-	 * @inheritDoc
-	 *
-	 * @param array $valid_plan_ids List of valid plan IDs.
-	 * @param array $access_level Access level for content.
-	 *
-	 * @return bool Whether the user can view the content
+	 * @return array|false The token payload or false otherwise
 	 */
-	public function visitor_can_view_content( $valid_plan_ids, $access_level ) {
-
+	public function get_token_payload() {
 		// URL token always has a precedence, so it can overwrite the cookie when new data available.
 		$token = $this->token_from_request();
 		if ( $token ) {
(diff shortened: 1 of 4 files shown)


projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -686,7 +686,27 @@ function render_wpcom_subscribe_form( $data, $classes, $styles ) {
...
+
+		require_once JETPACK__PLUGIN_DIR . 'extensions/blocks/premium-content/_inc/subscription-service/include.php';
+		$subscription_service = \Automattic\Jetpack\Extensions\Premium_Content\subscription_service();
+
+		$show_login_sign_in_link = add_query_arg( 'redirect_to', rawurlencode( get_post_permalink() ), 'https://wordpress.com/log-in/link' );
+		if ( ! is_user_logged_in() && empty( $subscription_service->get_token_payload() ) ) {
+			$show_login_action_text = __( 'Log in', 'jetpack' );
+		} else {
+			$show_login_sign_in_link = add_query_arg( 'redirect_to', rawurlencode( $show_login_sign_in_link ), 'https://wordpress.com/log-in?action=logout' );
+			$show_login_action_text  = __( 'Switch accounts', 'jetpack' );
```
millerf, line 707: I don't understand the "switch account" use case...

It will happen when the user is loggued OR payload is not empty.... In both those cases shouldn't we just "not" display the login? Or a logout button?

projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
```diff
@@ -48,15 +48,9 @@ public function initialize() {
...
-	 *
-	 * @param array $valid_plan_ids List of valid plan IDs.
-	 * @param array $access_level Access level for content.
-	 *
-	 * @return bool Whether the user can view the content
+	 * @return array|false The token payload or false otherwise
 	 */
-	public function visitor_can_view_content( $valid_plan_ids, $access_level ) {
-
+	public function get_token_payload() {
```
millerf, line 53: Nice!

millerf: Maybe we could also change it [here](https://github.com/Automattic/jetpack/blob/3b773377bff3362955827f58910fe43d2c2441eb/projects/plugins/jetpack/extensions/blocks/premium-content/login-button/login-button.php#L50)

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
lezama: the "Switch account" links to logout, it's a matter of copy 😄, @crisbusquets is giving all this a design iteration (p1694681489008179/1694613846.242699-slack-C052XEUUBL4), let's see what she propose!

projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
lezama: nice, let's note it for a follow up PR.

millerf: Maybe it could be integrated in this one? It would also add another user of `get_token_payload()` and legitimate its creation even more ;)

kraftbj (changes requested):
Needs a trunk merge and has conflicts that need to be dealt with.

PR #33050, opened 2023-09-13
Update/blocks contact form discoverability

### LINKED ISSUE #10903: Gutenberg: Enable Contact Form block by default / Block Discoverability

1. Install Jetpack v6.8.1
2. Add `define( 'JETPACK_DEV_DEBUG', true );` to wp-config.php.
3. Add new post.
4. Open the block inserter and scroll to Jetpack. Note the Form block is not available.
5. Visit `[yourjetpack.blog]/wp-admin/admin.php?page=jetpack_modules`. Note that the Contact Form module is not enabled despite [documentation](https://jetpack.com/support/jetpack-blocks/form-block/) saying that it should be activated by default.
6.Enable the `Contact Form` module. At this point the Form block becomes available in the editor.

#### What I expected

I expected the Form block to be available in the Gutenberg block inserter and in slash commands by default.

#### What happened instead

Instead I had to search through every setting to look for a way to enable the `Form` block. When I did not find a setting, I eventually found the Debug link in the footer where I then followed the link to enable the Contact Form module, which in turn enabled the Form block.

**Screenshots**
Screenshot of the block inserter immediately after plugin activation, showing that the `Form` block is missing:
image

#### Primary issue

#32401

andrii-lysenko:
Fixes #10903

## Proposed changes:

This change will alow registering Contact Form Block even if the Contact Form module is disabled. It should show the placeholder allowing to enable the module in one click. This is aimed to make our blocks more discoverable.

### Other information:
There's an issue when the module is not active and we are showing placeholder, we would probably want to have AI Assistant disabled. In current status, `useModuleStatus` logic won't help us sync different parts of code. We would either want to have some control over the Assistant from within block it is working on (eg from Contact Form block we could say to to show AI assistant), or, create a proper `store`, where the status will be handled. @jeherve, @retrofox what do you think?

## Testing instructions:

* Spin up the branch locally and tunnel with JT to your sandbox
* Build `packages/forms` and `plugins/jetpack`
* Navigate to `/wp-admin/admin.php?page=jetpack_modules` and make sure that Contact Form module is disabled
* Create a new post `/wp-admin/post-new.php`
* In blocks menu make sure you see the `Form` block and are able to add it to the page. It will have loading state first

Screenshot at Sep 14 15-35-30

Screenshot at Sep 14 15-36-07

* Activate the module and check that the form works as expected.


projects/packages/forms/src/contact-form/class-contact-form-plugin.php
```diff
@@ -314,13 +275,6 @@ public static function disable_forms_view_script_concat( $do_concat, $handle ) {
 	 * Register the contact form block.
 	 */
 	private static function register_contact_form_blocks() {
```
jeherve, line 277: I think it may be worth adding a comment here to explain that this registers the child blocks only, and that the parent block is registered separately so it can be more readily available?

Or maybe we could bring everything that's block-related to the new block class, to keep things together?

projects/plugins/jetpack/extensions/blocks/contact-form/contact-form.php
```diff
@@ -0,0 +1,12 @@
+<?php
+/**
+ * Subscriptions Block.
```
jeherve:
```suggestion
 * Contact Form Block.
```

projects/plugins/jetpack/extensions/index.json
```diff
@@ -52,7 +52,8 @@
 		"videopress/video",
 		"ai-assistant-support",
 		"ai-assistant-form-support",
-		"paywall"
+		"paywall",
+		"contact-form"
```
jeherve: Personal pet-peeve of mine, and I know it hasn't been done for the past few blocks above, but I think those long lists are easier to parse when the items are loaded alphabetically. :)

projects/packages/forms/src/blocks/contact-form/class-contact-form-block.php
```diff
@@ -0,0 +1,90 @@
+<?php
+/**
+ * Subscriptions Block.
```
jeherve:
```suggestion
 * Contact Form Block.
```

projects/packages/forms/src/blocks/contact-form/class-contact-form-block.php
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+namespace Automattic\Jetpack\Extensions\Contact_Form;
+
+use Automattic\Jetpack\Assets;
+use Automattic\Jetpack\Blocks;
+use Automattic\Jetpack\Forms\ContactForm\Contact_Form;
+
+/**
+ * Contact Form block render callback.
```
jeherve, line 17: I think I would go into more details here, to explain the overall infrastructure of the block in this package:

- A parent block in its own class
- Child blocks loaded separately (maybe they should be part of this class too, in their own method?)
- How a plugin can load those different items on their end.

Basically, I think we would benefit from a how-to in here so one can easily load the contact form block in their own plugin in the future.

projects/packages/forms/src/blocks/contact-form/components/jetpack-contact-form-placeholder.js
```diff
@@ -0,0 +1,34 @@
...
+	const { tracks } = useAnalytics();
+
+	const enableModule = () => {
+		return changeStatus( true );
+	};
+
+	return (
+		<Placeholder
+			icon={ settings.icon }
+			instructions={ __(
```
jeherve, line 13: All over this package, the textdomain is `jetpack-forms`

jeherve (commented):
This is looking good! I left a few notes below.

monsieur-z (dismissed):
The PR tested as specified in the instructions.

The code changes look sensible to me.

monsieur-z:
I've just noticed that while we show the placeholder in the post editor, a short code is showing in the preview:
Screenshot 2023-09-15 at 2 59 56 PM
Screenshot 2023-09-15 at 2 59 49 PM

andrii-lysenko:
@monsieur-z Thanks for spotting that out. This should be fixed now

monsieur-z (approved):
Works as expected!

As a side note, the block could be registered with the recommended way (using a `block.json` file) but this is probably out of scope for this PR.

PR #33054, opened 2023-09-13
Replace Calypso progress bar with the one from VideoPress

anomiex:
## Proposed changes:

Calypso has a nice progress bar. But it's bundled in the `@automattic/components` package that has a _lot_ of very Calypso-specific dependencies that we don't need or want in Jetpack, some of which we're already using pnpmfile to skip installing. Also Calypso has not released a new version of since June 2022, which suggests that use outside of their monorepo is not a priority for them.

Meanwhile our own VideoPress package has a good start on a progress bar component. Let's move that into our `@automattic/jetpack-components` package and use it instead of Calypso's, allowing us to remove that dependency and the accompanying pnpmfile hacks.

I see Gutenberg also has an experimental progress bar component, but the visual appearance is very different (plus it's still experimental).

### Other information:

## Jetpack product discussion

p1655760691502999-slack-CBG1CP4EN among others.

No

## Testing instructions:

* Check that things still work, and that any differences look reasonable.
  * Bars in the VideoPress package/plugin should show no changes at all.
  * VideoPress card Jetpack's dashboard (when you have a non-unlimited plan and some videos uploaded) should be about the same, except the bar is now green instead of blue.
  * Same for the VideoPress option in Settings → Performance.
  * Backup plugin admin page storage usage (again with a limited-storage plan) should look the same. In particular, the different colors for different levels of fullness should show up.
  * The small progress bar in the upper left corner of the Recommendations cards should also look about the same.


anomiex:
Feel free to directly update this with necessary changes if you want, I'm not all that well-versed in the details of frontend stuff.

anomiex:
Hmm. Now that I did this, I see there's yet another progress bar in [`projects/plugins/protect/src/js/components/progress-bar/`](https://github.com/Automattic/jetpack/tree/2e9cc48512ab474f3699ab85047fb90d972ed39a/projects/plugins/protect/src/js/components/progress-bar), which seems to have an interface somewhat closer to the Calypso one...

oskosk:
Are there any visual changes that would result from this change?

anomiex:
The only one I noticed I already called out in the testing section: (emphasis added)

> * VideoPress card Jetpack's dashboard (when you have a non-unlimited plan and some videos uploaded) should be about the same, **except the bar is now green instead of blue**.
> * Same for the VideoPress option in Settings → Performance.

Which seems ok to me, the use of Calypso's default blue there rather than a Jetpack green was probably a mistake.

But I haven't checked whether corners have exactly the same radius or the like, so there might be a pixels' difference in something like that.

CGastrell (dismissed):
I only tested VideoPress and Jetpack Plugin changes, but those look good (both code and UI). I think it could come in handy to have someone from Backup team to take a look as I'm not too familiar with it.

CGastrell (approved):
Let's ship this, unless you want to ping someone from backup to really test there :shipit:

PR #33056, opened 2023-09-13
CRM: Fix address custom fields not showing up until field sorts are saved

gogdzl:
When addresses are using custom sorting (`wp-admin/admin.php?page=zerobscrm-plugin-settings&tab=fieldsorts`) and a new custom address field is saved, the new field does not show anywhere in the CRM but in the custom sorting page. After hitting save in the custom sorting page, even without modifying anything, everything works as expected.

This happens for two reasons: custom addresses use `addr_` as a prefix for custom fields, and the code is not adding new fields to the sorting array. This PR fixes these two problems.

## Proposed changes:
* Fix the bug!

### Other information:

## Jetpack product discussion
https://github.com/Automattic/zero-bs-crm/issues/3089

No.

## Testing instructions:
0. Save custom field sorts: `/wp-admin/admin.php?page=zerobscrm-plugin-settings&tab=fieldsorts`
1. Create a address custom field: `/wp-admin/admin.php?page=zerobscrm-plugin-settings&tab=customfields`
2. Go to the new/edit contact page: `/wp-admin/admin.php?page=zbs-add-edit&action=edit&zbstype=contact`

In `trunk` the address custom field is missing.
In `fix/crm/3089-fix-address-custom-fields` the address custom field is showing.


coder-karen (commented):
Testing following the testing instructions works well.

It looks a side-effect here though is that adding a new address custom field without ever visiting the field sorts page and clicking save - `/wp-admin/admin.php?page=zerobscrm-plugin-settings&tab=fieldsorts` - results in all other address fields being removed from the edit contact page, until field sorts are saved (vs the current behaviour which would just add the fields to the bottom of all address fields):

Screenshot 2023-09-14 at 15 36 54

tbradsha:
> It looks a side-effect here though is that adding a new address custom field without ever visiting the field sorts page and clicking save [...]

That sounds like what I reported here as the previous behaviour: 3089-gh-Automattic/zero-bs-crm#issuecomment-1718157455

But:

> vs the current behaviour which would just add the fields to the bottom of all address fields

Is this a regression or a pre-existing issue? Either way, @gogdzl, this should also be addressed.

gogdzl:
Thanks both for pointing this out.

The code was adding all new fields for both field sorting and field hiding, so when a new field was added all fields were identified as being new fields, thus being added to the field hiding list.

I've added a check to add the new fields only for sorting.

The code is not that readable because it has to keep track of new fields, old fields, and keep the previous field sort and field hide settings. And it has to do it with several different constraints:
 - `address` is called `addresses` and also `zbsAddressFields` depending on the context
 -  only address custom fields use `addr_` as a prefix, 
 - addresses have a `migrate` special case for grouping purposes, 
 - we have a `$GLOBALS` key for some fields, not all, 
 - we don't have the names of deleted fields at hand so we can just iterate through them)
 - and some more

Have this in mind while reviewing the code, and simplifications are welcome if you spot a place where we can improve.

I would be glad if both of you (@tbradsha  and @coder-karen) could test again, as testing this proved to be tricky.

coder-karen:
Tested again thoroughly now, all previous as well as new behaviours.

The issue I mentioned above is now fixed - custom (and all) address fields show up both with and without saving the field save sorts page first.

> That sounds like what I reported here as the previous behaviour: https://github.com/Automattic/zero-bs-crm/issues/3089#issuecomment-1718157455

From my testing just now (testing previous behaviour), as you mentioned saving field sorts first then adding and saving a new custom address field, followed later by another then removes all address custom fields from showing on the view and edit contact pages or client portals (but not all address fields). Saving field sorts again brings them back.
The new behaviour was that all address fields were being removed except for the custom fields when adding a custom address field but not visiting the field sorts page :) The latest commit has fixed that though.

projects/plugins/crm/admin/settings/custom-fields.page.php
```diff
@@ -301,8 +301,9 @@
 	 */
 	$sort_types_map = array(
 		'address'  => array(
-			'custom_key' => 'addresses',
-			'obj_key'    => 'zbsAddressFields',
+			'custom_key'   => 'addresses',
+			'obj_key'      => 'zbsAddressFields',
+			'force_prefix' => 'addr_',
```
coder-karen: I think possibly naming 'force_prefix' to 'force_addr_prefix' might be more self explanatory?
```suggestion
			'force_addr_prefix' => 'addr_',
```

Then also affects line 341.

coder-karen (commented):
In terms of the code and readability, I've left just one suggestion.

projects/plugins/crm/admin/settings/custom-fields.page.php
gogdzl: Thanks! Addressed in a029281.

tbradsha (approved):
It works! Ideally someday it'll get a rewrite so the logic can be less convoluted (not your fault), but not a priority at this point.

coder-karen (approved):

PR #33057, opened 2023-09-13
Classic Editor Markdown: support double-tilde strikethrough

### LINKED ISSUE #3752: Classic Editor: Add Support for Double-Tilde Strikethrough.
#### What I expected

With Markdown support enabled it appears that enclosing text in double-tilde (~~) in a Wordpress post or article does not result in ~~strike-through~~ text. There is a certain irony in reporting this bug as it works fine in GitHub's editor.
#### What happened instead

Instead of getting strike-through text the pair of tildes are rendered as text. Thus:

```
some text ~~I want a strike-through here~~ but I don't get it.
```

Create a Wordpress post with Markdown enabled and enclose a word in double-tilde like so: `~~word~~` and it won't be rendered as strike-through.
#### Additional Notes

I've grown accustom to GitHub flavored Markdown and would like to see the feature included in Jetpack. Including tilde strike-through support won't conflict with plain vanilla markdown or common Markdown extensions like [Markdown Extra](https://michelf.ca/projects/php-markdown/extra/).

Also, it's worth noting that several markdown editors support this usage including [dillinger.io](http://dillinger.io), and [stackedit.io](http://stackedit.io) so it seems support for double tilde has entered common usage.

kraftbj:
Fixes #3752

When we moved to the Markdown block for the Block Editor, we used Markdown-It, which natively supports ~~strikethrough~~ (`~~` for strikethrough). The Classic Editor used markdown-php which does not support that.

This adds GPL code to the Jetpack flavor of Markdown Extra to support the double-tilde sequence.

## Proposed changes:

* Adds support for double-tilde strikethrough

### Other information:

## Jetpack product discussion

None. Just pinged on an old issue.

No.

## Testing instructions:

* Jetpack + Connection
* `wp plugin install classic-editor --activate`
* `wp jetpack module activate markdown`
* `wp post create --post_title="Test" --post_content="~~strikethrough~~" --post_status=public`
* View the post on the front end. With patch, it will be a strikethrough. Without it, it won't.

Please note that the markdown is rendered on save, so you'll want to re-save any testing post between applying/removing the patch.

### DIFF
--- a/projects/plugins/jetpack/_inc/lib/markdown/gfm.php
+++ b/projects/plugins/jetpack/_inc/lib/markdown/gfm.php
@@ -4,6 +4,7 @@
  *
  * @author Evan Solomon
  * @author Matt Wiebe <wiebe@automattic.com>
+ * @author Brandon Kraft <kraft@automattic.com> -- Strikedown support converted from GPL code at https://github.com/annaesvensson/yellow-markdown/blob/main/markdown.php
  * @link https://github.com/evansolomon/wp-github-flavored-markdown-comments
  *
  * Add a few extras from GitHub's Markdown implementation. Must be used in a WordPress environment.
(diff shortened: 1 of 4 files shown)


kraftbj:
I should add more tests for things like `~~strikethrough~~` (the sequence surrounded by code backticks).

jeherve (approved):
Thanks for adding some tests to cover the changes. I wanted to test some more use-cases so I ended up refactoring the tests to use a data provider with more scenarios, I hope you don't mind. => 3cba9696984830bd8c9c5078a4ac71b1ab6ab9b0

This tests well and should be good to go now. 🚢

PR #33060, opened 2023-09-14
Blogroll: Append wpcom site to blogroll

agrullon95:
Fixes #

## Proposed changes:

Adds feature to append WPCOM site to blogroll.

### Other information:

## Testing instructions:

* Checkout branch
* Sync to simple site / atomic site.
* Add blogroll block and open blogroll item appender popup.
* Verify you can filter through the suggestions and append to the blogroll block.
* Verify you can search for a WPCOM site using its URL and append it to the blogroll block.

CleanShot 2023-09-14 at 00 36 23@2x


projects/plugins/jetpack/extensions/blocks/blogroll/components/blogroll-appender/index.js
```diff
@@ -26,22 +29,91 @@ export default function BlogrollAppender( { subscriptions, clientId } ) {
...
+				}
+			} else {
+				setResults( subscriptions );
+			}
+		}, 1000 );
+
+		return () => {
+			clearTimeout( cancellableSearch );
+		};
+	}, [ searchInput, subscriptions ] );
```
alshakero: These should be a hook.

projects/plugins/jetpack/extensions/blocks/blogroll/utils.js
```diff
@@ -20,3 +20,22 @@ export function createBlockFromSubscription( subscription ) {
...
 	} );
 }
+
+export function checkIfValidDomain( siteURL ) {
+	const regEx =
+		/(http(s)?:\/\/.)?(www\.)?[-a-zA-Z0-9@:%._+~#=]{2,256}\.[a-z]{2,64}\b([-a-zA-Z0-9@:%_+.~#?&//=]*)/g;
+
+	if ( ! siteURL ) {
+		return false;
+	}
```
alshakero: Why create the regex before checking for null?

projects/plugins/jetpack/extensions/blocks/blogroll/utils.js
```diff
@@ -20,3 +20,22 @@ export function createBlockFromSubscription( subscription ) {
 		description,
 	} );
 }
+
+export function checkIfValidDomain( siteURL ) {
+	const regEx =
+		/(http(s)?:\/\/.)?(www\.)?[-a-zA-Z0-9@:%._+~#=]{2,256}\.[a-z]{2,64}\b([-a-zA-Z0-9@:%_+.~#?&//=]*)/g;
```
alshakero: Bad regex it will match `https://ddd.com/ddd`. We want only `ddd.com`.

projects/plugins/jetpack/extensions/blocks/blogroll/utils.js
```diff
@@ -20,3 +20,22 @@ export function createBlockFromSubscription( subscription ) {
...
+	if ( ! siteURL ) {
+		return false;
+	}
+
+	let validUrl;
+
+	try {
+		validUrl = new URL( siteURL );
+	} catch ( e ) {
+		validUrl = siteURL.match( regEx );
```
alshakero: We should use the match, not the input.
If somebody gives http://google.com/hello, we only want the google.com.
```suggestion
		return new URL( siteURL ).host;
	} catch ( e ) {
		return siteURL.match ( regEx );
```

alshakero (changes requested):

projects/plugins/jetpack/extensions/blocks/blogroll/components/blogroll-appender-results/style.scss
```diff
@@ -12,6 +12,9 @@
 	> ul {
 		margin: 0;
 		max-height: 200px;
+		list-style: none;
+		padding-left: 0px !important;
+		overflow: scroll;
```
heavyweight: Screenshot 2023-09-18 at 13 02 29


```suggestion
		overflow-y: auto;
```

projects/plugins/jetpack/extensions/blocks/blogroll/components/blogroll-appender-search/style.scss
```diff
@@ -8,6 +8,15 @@
 		position: relative;
 	}
 
+	.components-text-control__input {
+		background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='-2 -2 24 24' width='24' height='24' aria-hidden='true' focusable='false'%3E%3Cpath d='M6.734 16.106l2.176-2.38-1.093-1.028-3.846 4.158 3.846 4.157 1.093-1.027-2.176-2.38h2.811c1.125 0 2.25.03 3.374 0 1.428-.001 3.362-.25 4.963-1.277 1.66-1.065 2.868-2.906 2.868-5.859 0-2.479-1.327-4.896-3.65-5.93-1.82-.813-3.044-.8-4.806-.788l-.567.002v1.5c.184 0 .368 0 .553-.002 1.82-.007 2.704-.014 4.21.657 1.854.827 2.76 2.657 2.76 4.561 0 2.472-.973 3.824-2.178 4.596-1.258.807-2.864 1.04-4.163 1.04h-.02c-1.115.03-2.229 0-3.344 0H6.734z'%3E%3C/path%3E%3C/svg%3E");
+		background-size: 32px;
```
heavyweight: The icon seems smaller in the figma files

```
    background-size: 24px;
    background-position-y: calc(100% - 8px);
```

projects/plugins/jetpack/extensions/blocks/blogroll/components/blogroll-appender/index.js
```diff
@@ -1,22 +1,20 @@
 import { Button, Popover } from '@wordpress/components';
 import { dispatch } from '@wordpress/data';
-import { useState, useEffect } from '@wordpress/element';
+import { useState } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import { plus } from '@wordpress/icons';
-import { debounce } from '../../../../shared/debounce';
```
heavyweight, line 6: We should find a way to bring back debouncing especially now when we are hitting endpoints with search

heavyweight (changes requested):

heavyweight:
We also need a changelog entry to fix the checks

projects/plugins/jetpack/extensions/blocks/blogroll/components/blogroll-appender/index.js
agrullon95: I already have logic in place to handle debouncing for the search input. 
I could not get this debounce function to work in the `use-get-site-details` useEffect, so I wrote my own logic for it.

projects/plugins/jetpack/extensions/blocks/blogroll/utils.js
agrullon95: @alshakero - I updated the logic to return the valid domain or null. It will either get the host name from the new URL constructor or attempt to get the host from the regex match.

projects/plugins/jetpack/extensions/blocks/blogroll/components/blogroll-appender/index.js
agrullon95: I moved the logic into the `use-get-site-details` hook

heavyweight (dismissed):
LGTM we will revisit the hook once we have more detailed designs

projects/plugins/jetpack/extensions/blocks/blogroll/edit.js
```diff
@@ -46,14 +46,19 @@ export function BlogRollEdit( { className, attributes, setAttributes, clientId }
 		} ),
 	} );
 
+	const renderBlogrollAppender = () => {
+		if ( subscriptions && subscriptions.length ) {
+			return () => <BlogrollAppender subscriptions={ subscriptions } clientId={ clientId } />;
+		}
+		return null;
+	};
```
alshakero: What if you have none? You won't be able to append?

projects/plugins/jetpack/extensions/blocks/blogroll/use-get-site-details.js
```diff
@@ -0,0 +1,120 @@
...
+	enableSiteSearch = false,
+} ) {
+	const [ isLoading, setIsLoading ] = useState( false );
+	const [ errorMessage, setErrorMessage ] = useState( null );
+	const [ siteDetails, setSiteDetails ] = useState( subscriptions ?? [] );
+	const abortControllerRef = useRef();
+	const fetchSiteDetails = useCallback( async () => {
+		const validDomain = getValidDomain( siteURL );
+
+		await fetch(
```
alshakero: Either `await` or `then`.

alshakero: It seems we're fetching even if we have the site details cached.

projects/plugins/jetpack/extensions/blocks/blogroll/utils.js
```diff
@@ -20,3 +20,28 @@ export function createBlockFromSubscription( subscription ) {
...
+	const pattern = new RegExp(
+		'^([a-zA-Z]+:\\/\\/)?' + // protocol
+			'((([a-z\\d]([a-z\\d-]*[a-z\\d])?)\\.)+[a-z]{2,})', // domain name
+		'i'
+	);
+
+	try {
+		return new URL( siteURL )?.host;
+	} catch ( e ) {
+		return siteURL.match( pattern ) ? siteURL.match( pattern )[ 2 ] : null;
```
alshakero: This runs the regex once:
```suggestion
		return siteURL.match( pattern )?.[ 2 ] ?? null;
```

projects/plugins/jetpack/extensions/blocks/blogroll/use-subscriptions.js
```diff
@@ -35,6 +36,11 @@ export default function useSubscriptions( { ignore_user_blogs } ) {
 						)
 					);
 				}
+				// Add placeholder image if site icon is missing
+				response = response.map( subscription => {
+					subscription.site_icon = getSiteIcon( subscription.site_icon );
+					return subscription;
+				} );
```
alshakero, line 43: I think this is better.
```suggestion
				response = response.map( subscription => {
				        // if the icon is an empty string, return `undefined` so the default attribute kicks in.
					subscription.site_icon = subscription.site_icon || undefined;
					return subscription;
				} );
```

alshakero (changes requested):
All tiny notes.

projects/plugins/jetpack/extensions/blocks/blogroll/use-subscriptions.js
agrullon95: Reverted this change. It won't display the place holder in the appender if we dont apply this logic:

```
subscription.site_icon = getSiteIcon( subscription.site_icon );
```

projects/plugins/jetpack/extensions/blocks/blogroll/use-get-site-details.js
agrullon95: It will only fetch once per siteURL. If you check the network tab, another request will not be made if you enter the same URL in the input field.

projects/plugins/jetpack/extensions/blocks/blogroll/edit.js
agrullon95: Im going to revert this and handle in separate PR

alshakero (approved):
Tested and all works.

PR #33063, opened 2023-09-14
Publicize: do not display Twitter in post-publish notice

jeherve:
## Proposed changes:

In the Classic Editor, we display a notice when you publish a post, indicating that your post has been sent to all connected services.

Since Twitter is no longer supported, we should not display it in that message.
Screenshot 2023-09-14 at 11 09 11

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

**You can only test this on an existing site, where you had connected Publicize to a Twitter account in the past.**

1. Go to Plugins > Add New, install and activate the Classic Editor plugin.
2. Go to Posts > Add New, write a new post.
3. Publish your post.
4. You should not see Twitter mentioned in the post-publish notice.


projects/packages/publicize/src/class-publicize-base.php
```diff
@@ -1427,6 +1427,11 @@ public function update_published_message( $messages ) {
 
 		$labels = array();
 		foreach ( $services as $service_name => $display_names ) {
+			// Twitter connections do not trigger Publicize anymore. Skip.
+			if ( 'Twitter' === $service_name ) {
```
kraftbj, line 1431: The capitalization caught my attention, but traced it back. It is `twitter` in the options table, but it is converted to ucfirst via https://github.com/Automattic/jetpack/blob/1cb88291aff63bf80754db0a8719759c4140c0fe/projects/packages/publicize/src/class-publicize-base.php#L347

kraftbj (approved):

PR #33064, opened 2023-09-14
AI Chat: Add nudges so that AI Chat is properly configured

### LINKED ISSUE #32955: AI Chat block: Upgrade/purchase prompt
Whether the feature is behind the search plan or has it's own billing, the block should use an upgrade prompt to notify them of the required thing that is needed to use. The upgrade prompt should not show on the front end. 

Similar UX of the following, but imagine "this block requires a _____ plan": 

image

artpi:
For the time being, this assumes that Jetpack search plan is ALL that you need to activate the feature. In the future, this probably should change. However, with delays in the Jetpack Metered Billing product, I am treating this as part of Jetpack Search for now.

Zrzut ekranu 2023-09-14 o 11 26 06

Fixes https://github.com/Automattic/jetpack/issues/32955

## Proposed changes:
* Pass in a new parameter in `window.Jetpack_Editor_Initial_State.search`
* Add a nudge to check status of Jetpack connection
* Add a nudge to check if Jetpack search is enabled

NO

## Testing instructions:

1. Get a Jetpack site
2. `define( 'JETPACK_BLOCKS_VARIATION', 'beta' );`
3. Publish a few posts
4. Insert the block on a page
5. **Observe the prompt to connect Jetpack**
6. Connect to WPCOM.
7.  **Observe the prompt to enable Jetpack search.**
10. Enable it
11. Observe all prompts gone.
12. publish the page from 4
13. Open the page on the frontend
14. Ask a question relating to your published posts
15. Get a WPCOM site, sandbox public-api AND site domain
16. `bin/jetpack-downloader test jetpack chat-with-site/wpcom`
4. Publish a few posts
5. Insert the block on a page
7.  **Observe the prompt to enable Jetpack search.**
10. Enable it
11. Observe all prompts gone.
12. Publish the page with block
13. Open the page on the frontend
14. Ask a question relating to your published posts

### DIFF
--- a/projects/plugins/jetpack/class.jetpack-gutenberg.php
+++ b/projects/plugins/jetpack/class.jetpack-gutenberg.php
@@ -756,7 +756,7 @@ class Jetpack_Gutenberg {
 		// We need this information in the ai-chat feature, because it cant be activated if instant search is not enabled.
 		if ( Jetpack::is_module_active( 'search' ) ) {
 			$initial_state['search'] = array(
-				'is_enabled' => (bool) get_option( 'instant_search_enabled' ),
+				'instant_search_enabled' => (bool) get_option( 'instant_search_enabled' ),
 			);
 		}
 
--- a/projects/plugins/jetpack/class.jetpack-gutenberg.php
+++ b/projects/plugins/jetpack/class.jetpack-gutenberg.php
@@ -12,7 +12,6 @@ use Automattic\Jetpack\Connection\Initial_State as Connection_Initial_State;
 use Automattic\Jetpack\Connection\Manager as Connection_Manager;
 use Automattic\Jetpack\Constants;
 use Automattic\Jetpack\Current_Plan as Jetpack_Plan;
-use Automattic\Jetpack\Search\Plan as Search_Plan;
 use Automattic\Jetpack\Status;
 use Automattic\Jetpack\Status\Host;
 
@@ -751,14 +750,6 @@ class Jetpack_Gutenberg {
 					'image'     => $auto_conversion_settings->is_enabled( 'image' ),
 				),
 				'jetpackSharingSettingsUrl'       => esc_url_raw( admin_url( 'admin.php?page=jetpack#/sharing' ) ),
-			);
-		}
-
-		// We need this information in the ai-chat feature, because it cant be activated if instant search is not enabled.
-		if ( Jetpack::is_module_active( 'search' ) ) {
-			$search_plan             = new Search_Plan();
-			$initial_state['search'] = array(
-				'instant_search_enabled' => (bool) $search_plan->supports_instant_search(),
 			);
 		}
 
--- a/projects/plugins/jetpack/class.jetpack-gutenberg.php
+++ b/projects/plugins/jetpack/class.jetpack-gutenberg.php
@@ -12,6 +12,7 @@ use Automattic\Jetpack\Connection\Initial_State as Connection_Initial_State;
 use Automattic\Jetpack\Connection\Manager as Connection_Manager;
 use Automattic\Jetpack\Constants;
 use Automattic\Jetpack\Current_Plan as Jetpack_Plan;
+use Automattic\Jetpack\Search\Plan as Search_Plan;
 use Automattic\Jetpack\Status;
 use Automattic\Jetpack\Status\Host;
 
@@ -755,8 +756,9 @@ class Jetpack_Gutenberg {
 
 		// We need this information in the ai-chat feature, because it cant be activated if instant search is not enabled.
 		if ( Jetpack::is_module_active( 'search' ) ) {
+			$search_plan             = new Search_Plan();
 			$initial_state['search'] = array(
-				'instant_search_enabled' => (bool) get_option( 'instant_search_enabled' ),
+				'instant_search_enabled' => (bool) $search_plan->supports_instant_search(),
 			);
 		}
 
--- a/projects/plugins/jetpack/class.jetpack-gutenberg.php
+++ b/projects/plugins/jetpack/class.jetpack-gutenberg.php
@@ -753,6 +753,13 @@ class Jetpack_Gutenberg {
 			);
 		}
 
+		// We need this information in the ai-chat feature, because it cant be activated if instant search is not enabled.
+		if ( Jetpack::is_module_active( 'search' ) ) {
+			$initial_state['search'] = array(
+				'is_enabled' => (bool) get_option( 'instant_search_enabled' ),
+			);
+		}
+
 		wp_localize_script(
 			'jetpack-blocks-editor',
 			'Jetpack_Editor_Initial_State',
--- a/projects/plugins/jetpack/extensions/blocks/ai-chat/components/nudge-connect/index.tsx
+++ b/projects/plugins/jetpack/extensions/blocks/ai-chat/components/nudge-connect/index.tsx
@@ -1,0 +1,39 @@
+/*
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+/*
+ * Internal dependencies
+ */
+import { Nudge } from '../../../../shared/components/upgrade-nudge';
+import useAutosaveAndRedirect from '../../../../shared/use-autosave-and-redirect';
+
+const ConnectPrompt = () => {
+	const checkoutUrl = `${ window?.Jetpack_Editor_Initial_State?.adminUrl }admin.php?page=my-jetpack#/connection`;
+	const { autosaveAndRedirect, isRedirecting } = useAutosaveAndRedirect( checkoutUrl );
+
+	if ( window?.Jetpack_Editor_Initial_State?.jetpack?.is_current_user_connected ) {
+		return null;
+	}
+
+	const goToCheckoutPage = event => {
+		autosaveAndRedirect( event );
+	};
+
+	return (
+		<Nudge
+			buttonText={ __( 'Reconnect Jetpack', 'jetpack' ) }
+			checkoutUrl={ checkoutUrl }
+			className={ 'jetpack-ai-connect-banner' }
+			description={ __( 'Your account is not connected to Jetpack at the moment.', 'jetpack' ) }
+			goToCheckoutPage={ goToCheckoutPage }
+			isRedirecting={ isRedirecting }
+			visible={ true }
+			align={ null }
+			title={ null }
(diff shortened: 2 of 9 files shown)


projects/plugins/jetpack/extensions/global.d.ts
```diff
@@ -71,6 +71,12 @@ declare global {
 			siteLocale: string;
 			'ai-assistant': SiteAIAssistantFeatureEndpointResponseProps;
 			screenBase?: string;
+			search?: {
+				instant_search_enabled: boolean;
+			};
+			jetpack: {
+				is_current_user_connected: boolean;
```
dereksmart: This is already available in the JP_CONNECTION_INITIAL_STATE object https://github.com/Automattic/jetpack/pull/33064/files#diff-5ed459628f52ee22070733dd3841879ef252227b0532a586d545267cacfae371R34

dereksmart: Can we repurpose these functions in the ai-assistant lib? https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/connection/index.ts

projects/plugins/jetpack/class.jetpack-gutenberg.php
```diff
@@ -752,6 +752,13 @@ public static function enqueue_block_editor_assets() {
 			);
 		}
 
+		// We need this information in the ai-chat feature, because it cant be activated if instant search is not enabled.
+		if ( Jetpack::is_module_active( 'search' ) ) {
```
dereksmart: Better to use the Search_Plan()->supports_search() package method which also works in wpcom. https://github.com/Automattic/jetpack/blob/273fb4c4e309c1abd2698921ce6ca9669e6dc2a6/projects/plugins/jetpack/_inc/lib/class-jetpack-ai-helper.php#L101

dereksmart (changes requested):
Couple of changes then good to go. 

curious why tsx? is that needed for using the initial state?

projects/plugins/jetpack/extensions/global.d.ts
artpi: > This is already available in the JP_CONNECTION_INITIAL_STATE object

Yes, and it is already available on `Jetpack_Editor_Initial_State` object as well.
I decided to use the Jetpack_Editor_Initial_State because it's consistently available on both wpcom and Jetpack.

I only added a type to a field that already exists.

> Can we repurpose these functions in the ai-assistant lib?

I was considering it, but it's not the same feature/package. I can imagine them not being available at the same time.
Ai Chat is not a part of jetpack ai, but jetpack search. It seemed easier to just read an existing option that is already exposed all across JP in Jetpack_Editor_Initial_State

projects/plugins/jetpack/class.jetpack-gutenberg.php
artpi: Done

artpi:
> curious why tsx? is that needed for using the initial state?

I was copying ai assistant :D

https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/jetpack/extensions/blocks/ai-assistant/components/connect-prompt/index.tsx

projects/plugins/jetpack/extensions/blocks/ai-chat/components/nudge-connect/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+/*
+ * Internal dependencies
+ */
+import { Nudge } from '../../../../shared/components/upgrade-nudge';
+import useAutosaveAndRedirect from '../../../../shared/use-autosave-and-redirect';
+
+const ConnectPrompt = () => {
```
jeherve: Since this isn't specific to the AI Chat feature, maybe it could be moved up?

projects/plugins/jetpack/class.jetpack-gutenberg.php
```diff
@@ -753,6 +754,14 @@ public static function enqueue_block_editor_assets() {
...
+			$search_plan             = new Search_Plan();
+			$initial_state['search'] = array(
+				'instant_search_enabled' => (bool) $search_plan->supports_instant_search(),
```
jeherve: If this is only needed for AI Chat, part of me wonders if it wouldn't be easier for our future selves to understand if we did something like this?
https://github.com/Automattic/jetpack/blob/abce9f2b0fda88238e65a2a2aca14c7b44758195/projects/plugins/jetpack/extensions/plugins/launchpad-save-modal/launchpad-save-modal.php#L42-L43

projects/plugins/jetpack/extensions/blocks/ai-chat/components/nudge-enable-search/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+	return (
+		<Nudge
+			buttonText={ __( 'Enable Jetpack Search', 'jetpack' ) }
+			checkoutUrl={ checkoutUrl }
+			className={ 'jetpack-ai-connect-banner' }
+			description={ __(
+				'You need to enable Jetpack Search so that Jetpack AI can index your site.',
+				'jetpack'
+			) }
+			goToCheckoutPage={ goToCheckoutPage }
```
jeherve: Can we rely on `useModuleStatus` from `@automattic/jetpack-shared-extension-utils` to activate the module straight from the editor?

projects/plugins/jetpack/class.jetpack-gutenberg.php
artpi: Good idea!

artpi: I don't like this global approach. I'm gonna use the `jetpack-search/plan` endpoint

dereksmart: What don't you like about it and why do you prefer making an additional request?

projects/plugins/jetpack/extensions/global.d.ts
dereksmart: > Ai Chat is not a part of jetpack ai, but jetpack search.

Technically both can be considered part of Jetpack (the plugin) right now. Since, for example, the block is not available in the Jetpack Search plugin until it's isolated into a consumable package.

projects/plugins/jetpack/class.jetpack-gutenberg.php
artpi: Because we either:

- Make an additional request for people who DO want to use that block
- Add a parameter to hundreds of millions of sites (everybody) who may or may not want to use that block

Collective waste is much bigger if we add parameter everywhere

dereksmart: A new request brings new considerations like: 
- Caching 
- Not calling it from the front-end
- Only calling it in context (when the block is present)
- Error handling 
- State of the block when waiting for response

artpi: Ok, I don't have strong opinions here :D

projects/plugins/jetpack/extensions/blocks/ai-chat/components/nudge-enable-search/index.tsx
artpi: That is something we probably want to look into later, as specifics to which plan this belongs are still ironed out. We want to show some information ASAP

projects/plugins/jetpack/extensions/blocks/ai-chat/components/nudge-connect/index.tsx
artpi: We can probably drop this requirement anyways in the future, as we probably don't want to register the block at all until JP is connected.
This is abundance of caution, and as such I am not ready to take on the responsibility of creating THE connection nudge

projects/plugins/jetpack/class.jetpack-gutenberg.php
dereksmart: Let's go with initial state approach and pass it in a new localized variable like Jer suggested. It will make it just a little more portable for us later too.

projects/plugins/jetpack/extensions/global.d.ts
dereksmart: Pushed a change that uses `Automattic\Jetpack\Search\Initial_State as Search_Initial_State` and passing it within it's own initial state object as suggested.

projects/plugins/jetpack/class.jetpack-gutenberg.php
dereksmart: Done in [8c0bd9e](https://github.com/Automattic/jetpack/pull/33064/commits/8c0bd9ea572c86e3839b07e2a1e6d03a43a8a0b7)

projects/plugins/jetpack/extensions/blocks/ai-chat/components/nudge-enable-search/index.tsx
dereksmart: Yeah we'll need to come back to this as we haven't yet settled on whether or which the plan will be required or not. If so, we'll want it trigger a plan purchase flow.

dereksmart (dismissed):
@artpi I pushed a change which passes all of the initial state data from the Search package. If you're ok with it, I'm ok with merging this 

Edit: In the interest of only sending the data that we need (Search Initial State adds a LOT of db calls), I modified the new initial state object to only include details about the status of search.

dereksmart (approved):

dereksmart (changes requested):
Actually this isn't working in wpcom. The Admin URL when trying to enable search is trying to take me to an iframe editor which isn't working.

dereksmart (dismissed):
Pushed another fix. Should be working in wpcom now. Please verify though, I was having pretty inconsistent results.

artpi:
> Please verify though, I was having pretty inconsistent results.

Ok, figured it out.
When you are doing calypsoify, then :

> Jetpack_Editor_Initial_State.adminUrl
> 'https://membershipsandboxed1.wordpress.com/wp-admin/?frame-nonce=b5b6f90811'

That frame-nonce needs to go.

dereksmart (approved):
I already spent too long trying to figure out of there was a "proper" approach to the frame-nonce. Nothing is documented. Let's do it!

PR #33065, opened 2023-09-14
Admin Menu: Display Newsletter settings submenu item for jetpack connected sites

mashikag:
## Proposed changes:

* Add "Newsletter" submenu option to "Settings" menu item in jetpack connected sites
* Remove the "Newsletter" and "Podcasting" feature flag checks. They are not required anymore. Both of the features are released to PROD.

### Other information:

## Jetpack product discussion

p1694683896468089-slack-C02TCEHP3HA

## Testing instructions:

* Spin up an instance of a non-WPCOM hosted site with the jetpack plugin installed (you may use jurassic.tube)
* Apply the PR code to your test site
* Jetpack-connect your test site with your WPCOM account.
* Access the calypso portal for the site (you can do so by searching for the site at https://wordpress.com/sites)
* Ensure that the "Newsletter" settings submenu item is visible

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-admin-menu-display-newsletter-and-podcast-settings
+++ b/projects/plugins/jetpack/changelog/fix-admin-menu-display-newsletter-and-podcast-settings
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
+Admin Menu: Display Newsletter settings submenu item for jetpack connected sites
--- a/projects/plugins/jetpack/modules/masterbar/admin-menu/class-admin-menu.php
+++ b/projects/plugins/jetpack/modules/masterbar/admin-menu/class-admin-menu.php
@@ -371,34 +371,8 @@ class Admin_Menu extends Base_Admin_Menu {
 
 		$this->update_submenus( 'options-general.php', $submenus_to_update );
 
-		if (
-			/**
-			 * Filter to enable the Newsletter Settings section in Calypso UI.
-			 *
-			 * @since 12.5
-			 * @module masterbar
-			 *
-			 * @param bool false Enable newsletter setting section? Default to false.
-			 */
-			apply_filters( 'calypso_use_newsletter_settings', false )
-		) {
-			add_submenu_page( 'options-general.php', esc_attr__( 'Newsletter', 'jetpack' ), __( 'Newsletter', 'jetpack' ), 'manage_options', 'https://wordpress.com/settings/newsletter/' . $this->domain, null, 7 );
-		}
-
-		if (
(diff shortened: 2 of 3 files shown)


jeherve (commented):
Should we take that opportunity to add a check for the module status in there? If the Subscriptions module is disabled, maybe we shouldn't display the menu item (or better yet, we should offer a toggle to activate the module in the Calypso screen I suppose).

mashikag:
> we should offer a toggle to activate the module in the Calypso screen I suppose

I believe you mean to display the toggle within the Newsletter settings page. Hence we should still display the submenu item even when the subscription module is off. This sounds good to me. 

@jeherve are you aware if there is a dedicated endpoint for activating jetpack modules?

What about the toggle when the module is already active? I assume we should still display it. Otherwise it would be a little weird to have the toggle go away once the module is activated. 

cc @yansern, @simison

jeherve:
> are you aware if there is a dedicated endpoint for activating jetpack modules?

Check `{ activateModule } from 'calypso/state/jetpack/modules/actions';`, that should help.

> What about the toggle when the module is already active? I assume we should still display it. Otherwise it would be a little weird to have the toggle go away once the module is activated.

You can check https://github.com/Automattic/wp-calypso/pull/74209 for an example of an implementation that could work, I think.

mashikag:
Thanks so much for the input on this one, @jeherve. It will save me some time when implementing the toggle in Calypso. Either way we should be good to progress with this PR, right? Simple and atomic sites already do have the submenu item, and so should Jetpack. I wrongfully assumed that the Jetpack's implementation of admin menu calls the parent's `add_options_menu` like the Atomic does when working on the new settings page.

jeherve:
> Either way we should be good to progress with this PR, right?

I think so, yes, although I think we should have that toggle in Calypso sooner rather than later. Since the Subscriptions module is not automatically enabled, folks may find themselves visiting that new (to them) Calypso screen and not being able to do anything there.

> so should Jetpack

I don't know if it's that useful for Jetpack customers. The settings in that page are already available in wp-admin for Jetpack site owners.

yansern:
> If the Subscriptions module is disabled, maybe we shouldn't display the menu item (or better yet, we should offer a toggle to activate the module in the Calypso screen I suppose).

Yes, I would prefer that a toggle that activate the subscriptions module from the Newsletter Settings page, using that `activateModule` toggle. This helps with discovery of the feature when users are glancing through Calypso.

yansern (approved):
Activation toggle followed-up in https://github.com/Automattic/wp-calypso/pull/81840.

PR #33066, opened 2023-09-14
Change priority of load_map_block_settings action

### LINKED ISSUE #32991: Map block: private sites show a form to enter map access token.
### Impacted plugin

Jetpack

### Quick summary

When the site is set to private **(not private post)**, Map blocks show a different input form stating an access token is required. This only happens on AT, and not on Simple private sites.

I searched internally and in closed issues but it does not appear the "why" is not documented anywhere.

### Steps to reproduce

1. set a site to private.
2. start a new post.
3. add the Map block.

Map block asks for the coordinates like with when the site is public.

### What actually happened

Map block requests a token.

### Impact

All

### Available workarounds?

Atomic

### Logs or notes

_No response_

TimBroddin:
Fixes #32991

## Proposed changes:
The map block settings were not being loaded because the action ran before the script was registered. This PR changes the priority of the action, so the script is registered for sure.

### Other information:

- [X] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:
1. Apply this PR
2. Test using the instructions in FG: PCYsg-Osp-p2 
3. Mapkit should load on both WPCOM & Atomic

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/class-jetpack-mu-wpcom.php
+++ b/projects/packages/jetpack-mu-wpcom/src/class-jetpack-mu-wpcom.php
@@ -45,7 +45,7 @@ class Jetpack_Mu_Wpcom {
 		add_action( 'admin_enqueue_scripts', array( __CLASS__, 'unbind_focusout_on_wp_admin_bar_menu_toggle' ) );
 
 		// Load the Map block settings.
-		add_action( 'enqueue_block_assets', array( __CLASS__, 'load_map_block_settings' ) );
+		add_action( 'enqueue_block_assets', array( __CLASS__, 'load_map_block_settings', 999 ) );
 
 		/**
 		 * Runs right after the Jetpack_Mu_Wpcom package is initialized.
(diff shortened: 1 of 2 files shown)


projects/packages/jetpack-mu-wpcom/src/class-jetpack-mu-wpcom.php
```diff
@@ -45,7 +45,7 @@ public static function init() {
 		add_action( 'admin_enqueue_scripts', array( __CLASS__, 'unbind_focusout_on_wp_admin_bar_menu_toggle' ) );
 
 		// Load the Map block settings.
-		add_action( 'enqueue_block_assets', array( __CLASS__, 'load_map_block_settings' ) );
+		add_action( 'enqueue_block_assets', array( __CLASS__, 'load_map_block_settings', 999 ) );
```
jeherve: I think you'd have to move the priority outside of the array here:

```suggestion
		add_action( 'enqueue_block_assets', array( __CLASS__, 'load_map_block_settings' ), 999 );
```

jeherve (changes requested):

projects/packages/jetpack-mu-wpcom/src/class-jetpack-mu-wpcom.php
TimBroddin: 🫠 My bad. I first (wrongly) made the PR on the wrong repo. I then used the Github editor to quickly make the fix here. I wanted to go too quickly. Thanks for catching this.

jeherve (approved):

PR #33074, opened 2023-09-14
Add post publish quick share buttons

manzoorwanijk:
Related to https://github.com/Automattic/jetpack-reach/issues/37

## Proposed changes:

* Add sharing buttons UI to be used in post-publish panel in another PR

### Other information:

## Jetpack product discussion
pdrWKz-1b4-p2

No

## Testing instructions:

* Add the buttons to some panel UI
* Check if they work alright

Screenshot 2023-09-18 at 5 12 38 PM


projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
```diff
@@ -0,0 +1,47 @@
...
+export const ShareButtons: React.FC = () => {
+	const networks = useAvailableNetworks();
+	const prepareUrl = usePrepareUrl();
+
+	const onClick = useCallback( ( event: React.MouseEvent< HTMLAnchorElement > ) => {
+		if ( event.target instanceof HTMLAnchorElement ) {
+			const { network } = event.target.dataset;
+
+			// TODO Add tracking here
+			network;
```
gmjuhasz: What does this do?

projects/js-packages/publicize-components/src/components/share-buttons/usePrepareUrl.ts
```diff
@@ -0,0 +1,37 @@
...
+	const getEditedPostAttribute = useSelect( select => {
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any
+		return ( select( 'core/editor' ) as any ).getEditedPostAttribute satisfies (
+			attribute: string
+		) => unknown;
+	}, [] );
+
+	return useCallback(
+		( urlWithPlaceholders: string ) => {
+			// TODO update this to use custom message
```
gmjuhasz: Should we retrieve the custom message in this change too, or will that be a separate task?

projects/js-packages/publicize-components/src/components/share-buttons/useAvailableNetworks.tsx
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Returns the list of available networks.
+ *
+ * @returns {Array<Network>} The list of available networks.
+ */
+export function useAvailableNetworks() {
+	return useMemo< Array< Network > >(
+		() => [
+			{
+				label: __( 'Twitter', 'jetpack' ),
```
gmjuhasz: Probably this is okay for now, I wonder when do we have rename all these to X or Twitter/X

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
```diff
@@ -0,0 +1,47 @@
...
+		}
+	}, [] );
+
+	return (
+		<div className="share-buttons">
```
gmjuhasz: In Jetpack repo we have some obfuscation with the classnames, which happen automatically if you use the classnames like this: [Import](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/social/src/js/components/instagram-notice/index.jsx#L11) and [usage](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/social/src/js/components/instagram-notice/index.jsx#L56).

I think we should stick to that when possible to maintain the same semantics across the repo. You probably have to add a tsconfig here too as we did not have TS components in the `publicize-components` so far, and the style as module won't be recognized, but using the base tsconfig should have the decralation, like in Social: https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/social/tsconfig.json

gmjuhasz (commented):
This looks good, had some comments on small semantics, but working as expected!

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
manzoorwanijk: Nothing, it can be used when adding tracking. It is just to avoid that ESLint error for unused variables.

projects/js-packages/publicize-components/src/components/share-buttons/usePrepareUrl.ts
manzoorwanijk: We can do that in a separate task.

projects/js-packages/publicize-components/src/components/share-buttons/useAvailableNetworks.tsx
manzoorwanijk: I wanted to be consistent with other parts of Jetpack for now.

pablinos (commented):
Could we use [the social icons](https://github.com/WordPress/gutenberg/tree/9fcfaac513297b0517cf33c696a59d2324dd8e1e/packages/block-library/src/social-link/icons) from core? I think it would be best if we don't have the network label too. It takes up a lot of space, and we can have the title of the link be the network for anyone that doesn't recognise the icon.

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
manzoorwanijk: > In Jetpack repo we have some obfuscation with the classnames, which happen automatically if you use the classnames like this: [Import](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/social/src/js/components/instagram-notice/index.jsx#L11) and [usage](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/social/src/js/components/instagram-notice/index.jsx#L56).

That is called CSS modules. I think we have a mixed usage in `publicize-components` - e.g. [here](https://github.com/Automattic/jetpack/blob/f21d6084b7c3d944664330057b96c71ed42d2d28/projects/js-packages/publicize-components/src/components/connection-icon/index.jsx#L30) and [here](https://github.com/Automattic/jetpack/blob/f21d6084b7c3d944664330057b96c71ed42d2d28/projects/js-packages/publicize-components/src/components/social-previews/modal.js#L35)

> I think we should stick to that when possible to maintain the same semantics across the repo. You probably have to add a tsconfig here too as we did not have TS components in the publicize-components so far, and the style as module won't be recognized, but using the base tsconfig should have the decralation, like in Social: https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/social/tsconfig.json
    
I don't think that is required and it should work by default. We already have TS components in the `publicize-components` added by you [here](https://github.com/Automattic/jetpack/blob/f21d6084b7c3d944664330057b96c71ed42d2d28/projects/js-packages/publicize-components/src/components/connection-icon/index.jsx).

manzoorwanijk:
> Could we use [the social icons](https://github.com/WordPress/gutenberg/tree/9fcfaac513297b0517cf33c696a59d2324dd8e1e/packages/block-library/src/social-link/icons) from core? I think it would be best if we don't have the network label too. It takes up a lot of space, and we can have the title of the link be the network for anyone that doesn't recognise the icon.


At some places in Jetpack, we control the fill color of the icons and the icons from Core do not support that.

EDIT: Also, Core does not expose those icons.

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
gmjuhasz: Yeah, I think those places should have used the css module too. What do you think @pablinos which is the preferred way? 

> I don't think that is required and it should work by default. We already have TS components in the publicize-components added by you [here](https://github.com/Automattic/jetpack/blob/f21d6084b7c3d944664330057b96c71ed42d2d28/projects/js-packages/publicize-components/src/components/connection-icon/index.jsx).

Oh, I thought we only added in Social so far, I was mistaken then.

projects/js-packages/publicize-components/src/components/share-buttons/useAvailableNetworks.tsx
```diff
@@ -0,0 +1,32 @@
...
+	networkName: SocialIconSlug;
+	url: string;
+};
+
+/**
+ * Returns the list of available networks.
+ *
+ * @returns {Array<Network>} The list of available networks.
+ */
+export function useAvailableNetworks() {
```
pablinos: What's the advantage to having this as a hook with a memoized return? Could this just be a static array?

manzoorwanijk: It's for i18n which will break on static arrays because the language is initiated way after the modules are loaded.

pablinos: I see. Could we pass it through `__` as it's rendered instead?

manzoorwanijk: Then i18n string extractor build plugins will not work.

manzoorwanijk: I have updated it to use a static array. Thanks

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
manzoorwanijk: Alright, I have switched to CSS module after enabling TS support with a `tsconfig.json` file. Thank.

Let us see if CI is happy about that.

manzoorwanijk: OK, CI seems to be happy now.

projects/js-packages/publicize-components/src/components/share-buttons/usePrepareUrl.ts
```diff
@@ -0,0 +1,37 @@
...
+			let text =
+				getEditedPostAttribute( 'meta' )?.jetpack_seo_html_title ||
+				getEditedPostAttribute( 'title' );
+			let url = getEditedPostAttribute( 'link' );
+
```
pablinos: Could we reorganise this so that the urls are only recalculated when these values change?

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
```diff
@@ -0,0 +1,43 @@
...
+import React from 'react';
+import { useCallback } from 'react';
+import { availableNetworks } from './available-networks';
+import styles from './styles.module.scss';
+import { usePrepareUrl } from './usePrepareUrl';
+
+export const ShareButtons: React.FC = () => {
+	const prepareUrl = usePrepareUrl();
+
+	const onClick = useCallback( ( event: React.MouseEvent< HTMLAnchorElement > ) => {
```
pablinos: Might we want to open the window here, so we can control its size and location?

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
```diff
@@ -0,0 +1,43 @@
...
+		if ( event.target instanceof HTMLAnchorElement ) {
```
pablinos: When would it not be an instance of HTMLAnchorElement?

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
pablinos: I don't think this handler relies on any component props. If so, perhaps it can be defined externally?

manzoorwanijk:
@pablinos @gmjuhasz can you please have another look at this?

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
manzoorwanijk: I wanted to keep it simple for now. Do you want me to implement that in this PR or it can be done later?

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
manzoorwanijk: That is for TypeScript to ensure that we can use `event.target`'s dataset in that if block. It can be used for tracking to see which network button was clicked.

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
manzoorwanijk: Aha, good point. Thanks.

manzoorwanijk: Done

projects/js-packages/publicize-components/src/components/share-buttons/usePrepareUrl.ts
manzoorwanijk: Done. Thanks

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
manzoorwanijk: I have changed it to open a window here. Thanks

projects/js-packages/publicize-components/src/components/share-buttons/usePrepareUrl.ts
manzoorwanijk: I have done that in this PR. Thanks.

gmjuhasz (approved):

PR #33077, opened 2023-09-14
Paywall: Add custom filter

lezama:
See: p1694638019596869/1694628745.600779-slack-C053A0D3WGK

## Proposed changes:

* Add a new filter `'jetpack_custom_paywall_blocks' so plugins can define their own default paywall.

### Other information:

## Testing instructions:

* On a plugin add a filter for `'jetpack_custom_paywall_blocks'` and return some custom content.
* Verify it replaces paywalls.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -1064,9 +1064,6 @@ function get_paywall_content( $post_access_level, $email_confirmation_pending = 
 	if ( $email_confirmation_pending ) {
 		return get_paywall_blocks_subscribe_pending();
 	}
-	if ( ! jetpack_is_frontend() ) { // emails
-		return get_paywall_simple();
-	}
 	if ( doing_filter( 'get_the_excerpt' ) ) {
 		return '';
 	}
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -1044,6 +1041,13 @@ function get_paywall_content( $post_access_level, $email_confirmation_pending =
  * @return string
  */
 function get_paywall_blocks( $newsletter_access_level ) {
+	$custom_paywall = apply_filters( 'jetpack_custom_paywall_blocks', false );
+	if ( ! empty( $custom_paywall ) ) {
+		return $custom_paywall;
```
simison, line 1082: Would it make sense to pass same conditionals as below, or even wrap the final block-content output into this filter?

simison: Thinking further, suppose they have access to same conditional checks (logged/paying/etc) when they hook into the filter anyway. :-)

How you implemented this might actually be cleanest at the end for now.

simison (approved):
My jetpack dev setup is a bit broken so can't test-run this.

However code-wise looks good.

I think you also need some PHP doc above the filter to tell since which version it's been there.

jeherve:
> I think you also need some PHP doc above the filter to tell since which version it's been there.

Yes. We do not add new hooks to the plugin without documenting them. Without a docblock, no page will be automatically generated in [the Jetpack codex](https://developer.jetpack.com/).

I would strongly recommend adding one. The docblock has to include 4 pieces of info:

- The description of the filter (if this filter is meant as an overwrite to our own render, I don't know if it should include the word "blocks"?)
- The module it belongs to, so it can be added to the right category in the Codex. `@module subscriptions`
- The version when it was added: `@since $$next-version$$` (see https://github.com/Automattic/jetpack/blob/3a3c4adc1c0913bac224ddb832713bd6101fe1ab/projects/packages/README.md#L109-L116 )
- The parameter (what should it be here? Right now it defaults to `false`, but I assume we'd expect third-parties to return a string? Maybe the default should be an empty string?)

Since we're talking about parameters, would `$is_paid_post` be useful to third-parties when they want to display info in their own rendering function? Maybe we could define it before that hook, and pass that variable to the hook so it's useful for third-parties too?

lezama:
@jeherve the filter here is intended for Team 51, they need it for a particular client.

The plan is to make the paywall editable via template parts in the site editor. I wouldn't document the filter as the best way to  modify the paywall.

lezama:
Should I still add the docblock?

jeherve:
I would still add the docblock. You can mention in the description that better ways to customize the paywall will come in future versions

PR #33081, opened 2023-09-14
Add Enhanced Ownership

obenland:
Fixes https://github.com/Automattic/dotcom-forge/issues/3655 and https://github.com/Automattic/dotcom-forge/issues/3657

## Proposed changes:

* Adds a feature that allows users to designate a legacy contact, including settings in General Settings in wp-admin and callbacks to return and update the values in the 1.4 settings API endpoint
* Introduced Locked Mode, including settings in General Settings in wp-admin and callbacks to return and update the values in the 1.4 settings API endpoint, removing the ability to add/edit/delete content, plugins, themes, and disables comments side wide.

### Other information:

Yes. We log changes to the contents of the legacy contact and whether locked mode was enabled/disabled.

## Testing instructions:

* Sandbox public-api and a test site on Simple.
* Add a 100-year plan in your Store Admin to your test site.
* Visit /wp-admin/options-general.php on your test site and scroll to the end of the page.
* Check the Lock Mode option and Save Changes.
* Populate the Legacy Contact input field and Save Changes.
* Check the blog RC to make sure audit log entries appear as expected.
* Apply https://github.com/Automattic/wp-calypso/pull/81650 and spin up your local Calypso instance.
* Navigate to General Settings and look for Enhanced Ownership settings.
* The values you set in wp-admin should show up there. They should also update when you change them from Calypso.

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/100-year-plan/locked-mode.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/100-year-plan/locked-mode.php
@@ -110,10 +110,6 @@ function wpcom_locked_mode_render() {
  * @return array
  */
 function wpcom_lm_remove_post_capabilities( $caps, $cap ) {
-	if ( ! wpcom_site_has_feature( WPCOM_Features::LOCKED_MODE ) ) {
-		return $caps;
-	}
-
 	// Return if option is not set.
 	if ( ! get_option( 'wpcom_locked_mode' ) ) {
 		return $caps;
--- a/projects/packages/jetpack-mu-wpcom/src/features/100-year-plan/locked-mode.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/100-year-plan/locked-mode.php
@@ -1,0 +1,200 @@
+<?php
+/**
+ * Locked Mode.
+ *
+ * @package automattic/jetpack-mu-wpcom
+ */
+
+/**
+ * Registers Locked Mode settings.
+ */
+function wpcom_lm_register_settings() {
+	if ( ! wpcom_site_has_feature( WPCOM_Features::LOCKED_MODE ) ) {
+		return;
+	}
+
+	register_setting(
+		'general',
+		'wpcom_locked_mode',
+		array(
+			'type'              => 'boolean',
+			'description'       => 'Whether the site is in locked mode.',
+			'show_in_rest'      => true,
+			'default'           => false,
+			'sanitize_callback' => function ( $value ) {
+				return (bool) $value;
+			},
+		)
+	);
+}
+add_action( 'admin_init', 'wpcom_lm_register_settings' );
+add_action( 'rest_api_init', 'wpcom_lm_register_settings' );
+
+/**
+ * Adds settings to the v1 API site settings endpoint.
+ *
+ * @param array $settings A single site's settings.
+ * @return array
+ */
+function wpcom_lm_get_options_v1_api( $settings ) {
+	if ( wpcom_site_has_feature( WPCOM_Features::LOCKED_MODE ) ) {
+		$settings['wpcom_locked_mode'] = (bool) get_option( 'wpcom_locked_mode' );
+	}
+
+	return $settings;
+}
+add_filter( 'site_settings_endpoint_get', 'wpcom_lm_get_options_v1_api' );
+
+/**
+ * Updates settings via public-api.wordpress.com.
+ *
+ * @param array $input             Associative array of site settings to be updated.
+ *                                 Cast and filtered based on documentation.
+ * @param array $unfiltered_input  Associative array of site settings to be updated.
+ *                                 Neither cast nor filtered. Contains raw input.
+ * @return array
+ */
+function wpcom_lm_update_options_v1_api( $input, $unfiltered_input ) {
+	if ( isset( $unfiltered_input['wpcom_locked_mode'] ) ) {
+		$input['wpcom_locked_mode'] = (bool) $unfiltered_input['wpcom_locked_mode'];
+	}
+
+	return $input;
+}
+add_filter( 'rest_api_update_site_settings', 'wpcom_lm_update_options_v1_api', 10, 2 );
+
+/**
+ * Registers the settings section and fields.
+ */
+function wpcom_lm_settings_page_init() {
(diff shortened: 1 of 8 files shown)


projects/packages/jetpack-mu-wpcom/src/features/100-year-plan/locked-mode.php
```diff
@@ -0,0 +1,196 @@
...
+/**
+ * Removes post and page creation/editing capabilities for locked mode sites.
+ *
+ * @param array  $caps The user's capabilities.
+ * @param string $cap  The capability name.
+ * @return array
+ */
+function wpcom_lm_remove_post_capabilities( $caps, $cap ) {
+	// Return if option is not set.
+	if ( ! get_option( 'wpcom_locked_mode' ) ) {
```
obenland: Not sure how smart it is to add an option check to every cap check. I wonder if we could move it to an init callback and only add this map_meta_cap filter when needed.

DustyReagan: That is a good idea. To be on the safe side.

projects/packages/jetpack-mu-wpcom/src/features/100-year-plan/locked-mode.php
```diff
@@ -0,0 +1,196 @@
...
+		case 'edit_plugins':
+		case 'delete_plugins':
+		case 'edit_files':
+		case 'upload_files':
+		case 'edit_comment':
+		case 'moderate_comments':
+		case 'manage_categories':
+		case 'manage_links':
+		case 'import':
+		case 'edit_posts':
```
DustyReagan: 'edit_posts' appears twice.

DustyReagan (dismissed):
This is working great for me! The options save seamlessly between Calypso and wp-admin. 🎉

However, the Audit Trail in MC appears to only record updates. It doesn't record the initial setting. You can replicate this by submitting '' to clear out the Legacy Contact, and then save a new value. That first value doesn't appear in the Audit Trail, but subsequent updates do.

I approved to unblock the PR, but it might be worth implementing your idea on reducing option checks first.

projects/packages/jetpack-mu-wpcom/src/features/100-year-plan/locked-mode.php
obenland: Thanks, fixed in https://github.com/Automattic/jetpack/pull/33081/commits/165fcf11985b90dffe421d375eb4393952ea58ec

candy02058912:
> However, the Audit Trail in MC appears to only record updates. It doesn't record the initial setting. You can replicate this by submitting '' to clear out the Legacy Contact, and then save a new value. That first value doesn't appear in the Audit Trail, but subsequent updates do.

I was able to replicate this ➕

candy02058912:
> Check the blog RC to make sure audit log entries appear as expected.

Are we expecting locking/unlocking to produce audit logs?

projects/packages/jetpack-mu-wpcom/src/features/100-year-plan/locked-mode.php
```diff
@@ -0,0 +1,200 @@
...
+/**
+ * Adds the action to limit capabilities for locked mode sites when the Locked Mode feature is available
+ * and the option enabled.
+ */
+function wpcom_lm_maybe_add_map_meta_cap_filter() {
+	if ( wpcom_site_has_feature( WPCOM_Features::LOCKED_MODE ) && get_option( 'wpcom_locked_mode' ) ) {
+		add_filter( 'map_meta_cap', 'wpcom_lm_remove_post_capabilities', 10, 2 );
+	}
+}
+add_action( 'plugins_loaded', 'wpcom_lm_maybe_add_map_meta_cap_filter' );
```
DustyReagan: I think `plugins_loaded` may have already fired when this code runs, so the capabilities aren't being limited. I put `error_log(did_action('plugins_loaded'));` before this line to check (it's returning `1`). But, `setup_theme` seems to work. Do you think `setup_theme` runs early enough to suffice?

*edit: `setup_theme` isn't early enough. I tried. 😅

obenland: Oh, I see, this file gets loaded on `plugins_loaded`. Weird, it worked for me in wp-admin for some reason?
If you add a priority of 11 as a third parameter, does that make it work?

DustyReagan: Adding a priority of 11 works! That's a bit mystifying to me, because with the code below, I can see `plugins_loaded` has already fired once. So, I'd assume the priority wouldn't matter at that point. But, I guess it fires again?

```
error_log( print_r( did_action( 'plugins_loaded' ) ) );
add_action( 'plugins_loaded', 'wpcom_lm_maybe_add_map_meta_cap_filter', 11 );
```

obenland: Yeah, `did_action` has no concept of priorities and just returns the amount of times `do_action()` was called. But yes, you can register callbacks at a later priority than the priority it currently works through.

obenland: Fixed in https://github.com/Automattic/jetpack/pull/33081/commits/6f6829233b5bc30aa44896787dffb727ac443722

obenland:
> Are we expecting locking/unlocking to produce audit logs?

Yes it should! If it doesn't (which sounds like is the case with the initial value of legacy contact, I'll have to take another look at those audit_log callbacks. But it's not something that needs changing in this PR.

DustyReagan (approved):
Setting the priority to 11 did the trick! LGTM!

jenhooks:
Howdy @obenland! Jetpack is included in the labels here, and I'm trying to figure out if this is something we need to add to our [Privacy Center](https://jetpack.com/support/privacy/). Please advise?

obenland:
I don't think so. The things we track are only available on Simple and WoA sites. 
The part that added the Jetpack label was that we had to expand the list of options we sync from a Jetpack site, but those should only have values in WP.com

PR #33082, opened 2023-09-14
Add tier logic for premium content

millerf:
See this discussion: p1694698889470569-slack-C01B6KEJ5GE
Part of https://github.com/Automattic/gold/issues/151
Prevent this use case to happen: p1694700121016389/1694698889.470569-slack-C01B6KEJ5GE
## Proposed changes:

*  If the Premium content is a "tier", use the tier logic (higher tier gives access), to prveent issue when upgrading (and therefore cancelling a lower subscription that could be used by the premium content.

### Other information:

## Testing instructions:

* Create 2 tiers Silver and Gold
* Create a Premium content block and use Silver plan~
* Subscribe to the Silver plan : you should have access to Premium content 
* Subscribe to the Gold plan : you should STILL have access to Premium content

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/access-check.php
+++ b/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/access-check.php
@@ -96,7 +96,22 @@ function current_visitor_can_access( $attributes, $block ) {
 
 	$paywall      = subscription_service();
 	$access_level = Token_Subscription_Service::POST_ACCESS_LEVEL_PAID_SUBSCRIBERS; // Only paid subscribers should be granted access to the premium content
-	$can_view     = $paywall->visitor_can_view_content( array( $selected_plan_id ), $access_level );
+	$tier_ids     = \Jetpack_Memberships::get_all_newsletter_plan_ids();
+	if ( in_array( $selected_plan_id, $tier_ids, true ) ) {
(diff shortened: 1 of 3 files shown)


projects/plugins/jetpack/extensions/blocks/premium-content/_inc/access-check.php
```diff
@@ -96,7 +96,22 @@ function current_visitor_can_access( $attributes, $block ) {
 
 	$paywall      = subscription_service();
 	$access_level = Token_Subscription_Service::POST_ACCESS_LEVEL_PAID_SUBSCRIBERS; // Only paid subscribers should be granted access to the premium content
-	$can_view     = $paywall->visitor_can_view_content( array( $selected_plan_id ), $access_level );
+	$tier_ids     = \Jetpack_Memberships::get_all_newsletter_plan_ids();
+	if ( in_array( $selected_plan_id, $tier_ids, true ) ) {
```
lezama, line 100: would it make sense to do this `if` inside `visitor_can_view_content`?  Could be reused in other places

millerf: `visitor_can_view_content` is also used in other places, but we might not want it (it would run as a double for Newsletter).

It might be a good part for a good refactor, but I don't feel like tackling that right now...

PR #33083, opened 2023-09-14
AI Client: add `model` param to request helpers

retrofox:
On this PR, we add a `model` option to the helper that performs AI requests. For instance, `askQuestion()`,  `suggestionEventSource()`, etc.

Fixes #

## Proposed changes:

* AI Client: add `model` param to request helpers

### Other information:

## Testing instructions:

For now, it's a matter of checking the dev-env compiles properly the new changes suggested here.
Stop and restart `jetpack build plugins/jetpack` should help to check that.
Also, a visual review would be nice.


CGastrell (approved):
Code and docs look good. Backend diff works fine. LGTM, just remember I'm the team rookie :shipit:

retrofox:
> LGTM, just remember I'm the team rookie :shipit:

Thanks. I appreciate the review, and almost all of us belong to that team :-)

These AI Client libraries help to perform the AI requests. Under the hood, the PR introduces the ability to set the model via the `model` parameter. The fundamental piece here is the `SuggestionEventSource` class. But also need to propagate the parameter from other auxiliary libraries such as `askQuestion`. 
These changes will be in evidence once we implement the UI part in the AI products, allowing users to switch between the Open AI models when requesting suggestions.

PR #33084, opened 2023-09-14
AI Assistant: Change conditions to show AI guidelines on input

### LINKED ISSUE #32773: AI Assistant: Fix disclaimer logic of the AIControl component
ref p1693399189852639-slack-C054LN8RNVA

The disclaimer is displayed before the content is generated and not after.  The disclaimer should be shown only after the content is generated.

dhasilva:
Fixes #32773

## Proposed changes:

* Changes the conditions to show the guidelines. For the AI Assistant, it is now shown whenever there is any content in the container. For the Form extension it is now shown when there is a request happening or it is already done.

### Other information:

## Testing instructions:

* Go to the block editor
* Add an AI Assistant block
* Check that the guideline message is not shown at first
* Make a request
* Check that the guideline message appears only when there is some content.
* Add a Form
* Check that the guideline message is not shown at first either
* Make a request
* Check that the guideline message appears

It is easier to spot the difference by first testing on `trunk`.


retrofox (commented):
LGTM, with a minor comment.

AI Assistant block instance works as expected:

image


But for the Jetpack Form AI Assistant, it shows the help also in the `requesting` state. There isn't content on this state, so I'd suggest considering adding this state to the `showGuideLine` condition.

retrofox (approved):

worldomonation:
I am testing Jetpack Version 12.6-6240849897-g0eae74ad, and noted the AI bias disclaimer still shows when using the Forms AI generation feature, similar to what @retrofox noted in https://github.com/Automattic/jetpack/pull/33084#pullrequestreview-1627325974.

The disclaimer does not show when the Forms block is initially inserted. See screenshot below.

image


Are there plans to follow up to apply consistent behavior for when the disclaim is expected to show?

PR #33085, opened 2023-09-14
Boost: Add a "Fix" button to the Image Size Analyzer records to fix image display problems

donnchawp:
This PR adds a "Fix" button to each record in the Image Size Analyzer report. When clicked it saves a description of the correct dimensions of the image, and will apply those fixes when the post is displayed.

Fixes https://github.com/Automattic/boost-cloud/issues/371

## Proposed changes:
* Changes the "Fix on Page" button in Image Stabilizer Reports to "Fix".
* Clicking that button will insert a postmeta record with the proper dimensions of the image. It's keyed by the MD5 of the image name or the attachment ID of the image.
* Displaying a post with a fixed image will cause the HTML img tag of any fixed images to be modified so it loads the correct image through Photon, using srcset and sizes attributes.

### Other information:

## Jetpack product discussion
pc9hqz-27e-p2

No

## Testing instructions:
* Apply patch.
* run `npm run dev` in projects/plugins/boost/
* Edit wp-config.php and add `define( 'ISA_FIX_BUTTON', true );`
* Run an Image Size Analysis of the site and view report.
* Click on any record to view the details of an image problem.
* Click on the "Fix" button.
* Go into PHPMyAdmin to check if a new postmeta record was created in the fixed post. It's called "jb_image_fixes".
* Reload the page and verify that the image is fixed and matches the width of the content area.

Note: wp-config.php is in tools/docker/wordpress/ if you're using a local install.


projects/plugins/boost/app/modules/image-size-analysis/Image_Size_Analysis_Fixer.php
```diff
@@ -0,0 +1,123 @@
...
+	public static function setup() {
+		add_filter( 'the_content', array( __CLASS__, 'fix_content' ), 99 );
+		add_filter( 'wp_calculate_image_srcset', array( __CLASS__, 'fix_image_attachments' ), 10, 5 );
+		add_filter( 'wp_calculate_image_sizes', array( __CLASS__, 'fix_image_sizes' ), 10, 5 );
+	}
+
+	public static function fix_url( $url ) {
+		$url = preg_replace( '/-\d+x\d+\.jpg/', '.jpg', $url ); // remove "-1024x768" from the end of the URL.
+		$url = preg_replace( '/\?.*/', '', $url ); // remove query string.
+		$url = str_replace( 'i0.wp.com/', '', $url ); // remove i0.wp.com/ from the beginning of the URL.
```
pyronaur: Can we leverage [`parse_url()` ](https://www.php.net/manual/en/function.parse-url.php) instead?

projects/plugins/boost/app/modules/image-size-analysis/Image_Size_Analysis_Fixer.php
```diff
@@ -0,0 +1,123 @@
...
+	// phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
+	public static function fix_image_attachments( $sources, $size_array, $image_src, $image_meta, $attachment_id ) {
+		global $post;
+
+		$fixes = self::get_fixes( $post->ID );
+
+		$image_width = 0;
+
+		// remove XxY dimension from $image_src as that's what's recorded by Image_Size_Analysis.
+		$image_src_key = self::fix_url( $image_src );
```
pyronaur: Maybe `fix_url` should be called something that describes the return of the function? I'm not entirely clear currently what it returns, but in case what's left of the cleaned URL is just the domain (for example) then I'd call this `get_url_domain()` or along those lines.

projects/plugins/boost/app/modules/image-size-analysis/Image_Size_Analysis_Fixer.php
```diff
@@ -0,0 +1,123 @@
...
+		$fixes[ $post_id ] = get_post_meta( $post_id, 'jb_image_fixes', true );
+		if ( ! $fixes[ $post_id ] ) {
+			$fixes[ $post_id ] = array();
+		}
+
+		return $fixes[ $post_id ];
+	}
+
+	// phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
+	public static function fix_image_attachments( $sources, $size_array, $image_src, $image_meta, $attachment_id ) {
```
pyronaur: Is `image_src` the `<img src=""` tag or the image URL? This can be confusing because there are functions like [`wp_get_attachment_image_src`](https://developer.wordpress.org/reference/functions/wp_get_attachment_image_src/) that produce HTML. 

If the function does expect HTML, maybe this should be called `image_html` or `image_html_src` ?

projects/plugins/boost/app/modules/image-size-analysis/Image_Size_Analysis_Fixer.php
```diff
@@ -0,0 +1,123 @@
...
+				'value'      => $image_width,
+			);
+		}
+
+		return $sources;
+	}
+
+	// phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
+	public static function fix_image_sizes( $sizes, $size, $image_src, $image_meta, $attachment_id ) {
+		global $post;
```
pyronaur: Is it possible to use `get_post()` here instead?

projects/plugins/boost/app/modules/image-size-analysis/Image_Size_Analysis_Fixer.php
```diff
@@ -0,0 +1,123 @@
...
+		$fixes         = self::get_fixes( $post->ID );
```
pyronaur: I might be wrong, but I think `$post` isn't guaranteed to be set here?

projects/plugins/boost/app/modules/image-size-analysis/Image_Size_Analysis_Fixer.php
```diff
@@ -0,0 +1,123 @@
...
+		$image_width   = 0;
+		$image_src_key = self::fix_url( $image_src );
+		$attachment_id = attachment_url_to_postid( esc_url( $image_src ) );
```
pyronaur: `esc_url()` is used for presentation, for interacting with database there's [`esc_url_raw()`](https://developer.wordpress.org/reference/functions/esc_url_raw)

projects/plugins/boost/app/modules/image-size-analysis/Image_Size_Analysis_Fixer.php
```diff
@@ -0,0 +1,123 @@
...
+
+		if ( $attachment_id && isset( $fixes[ $attachment_id ] ) ) {
+			$image_width = $fixes[ $attachment_id ]['image_width'];
```
pyronaur, line 140: Looks like `$fixes[ $attachment_id ]` can be an empty array, so `image_width` possibly doesn't exist? I think that would probably cause a warning/notice in PHP?

projects/plugins/boost/app/rest-api/endpoints/Image_Analysis_Fix.php
```diff
@@ -0,0 +1,66 @@
...
+		$image_url = Image_Size_Analysis_Fixer::fix_url( $request->get_param( 'image_url' ) );
+
+		$attachment_id = attachment_url_to_postid( esc_url( $image_url ) );
+		if ( $attachment_id ) {
+			$fixes[ $attachment_id ] = $params;
+		} else {
+			$fixes[ md5( $image_url ) ] = $params; // hot linked image, possibly from another site.
+		}
+
+		update_post_meta( $request->get_param( 'post_id' ), 'jb_image_fixes', $fixes );
```
pyronaur: Possibly not a real-world issue - but there's no validation or sanitization on `$params`.

projects/plugins/boost/app/rest-api/endpoints/Image_Analysis_Fix.php
```diff
@@ -0,0 +1,66 @@
...
+
+		// Send a success response.
+		return rest_ensure_response(
+			array(
+				'ok' => true,
+			)
+		);
```
pyronaur: We're sending `ok` even if `update_post_meta` failed here.

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/recommendations/row-types/ImageSizeRow.svelte
```diff
@@ -24,6 +28,22 @@
...
+	async function fixImageSize() {
+		const data = {
+			image_url: details.image.url,
+			image_width: details.image.dimensions.expected.width.toString(),
+			image_height: details.image.dimensions.expected.height.toString(),
+			post_id: post_id.toString(),
+			nonce: Jetpack_Boost.fixImageNonce,
+		};
+		return await api.post( '/image-size-analysis/fix', data );
+	}
```
pyronaur: I think we'll try to deprecate `api.post()` during the refactor in favor of using DataSync - perhaps we should use DataSync here so that we don't have to refactor this later?

pyronaur (dismissed):
👋  Great stuff! I've left comments/questions inline. I'll give this a spin and see how it works.

pyronaur:
Just tested - fixing works 🤩 - but the UI is unresponsive when clicking the "Fix" button. I  wonder if we should have the "Edit Page" appear in the expanded state and show the fix button inline and call it something like "Fix Automattically" or "Auto-fix" making image fixing convenient.

projects/plugins/boost/app/rest-api/endpoints/Image_Analysis_Fix.php
dilirity: Tagging along here - I would advise prefixing all meta records with an underscore. Underscored meta records are private and don't show in the Custom FIelds section, so at least we can prevent them from being edited via the UI 😅 
CleanShot 2023-09-27 at 13 13 31

donnchawp:
> Just tested - fixing works 🤩 - but the UI is unresponsive when clicking the "Fix" button. I wonder if we should have the "Edit Page" appear in the expanded state and show the fix button inline and call it something like "Fix Automattically" or "Auto-fix" making image fixing convenient.

I think you're right about swapping the position of the Edit Page and Fix buttons. As discussed elsewhere, I'm going to change the Fix button to Undo when it's clicked. Which will of course let someone undo the fix.

projects/plugins/boost/app/modules/image-size-analysis/Image_Size_Analysis_Fixer.php
donnchawp: I had called it clean_up_url previously, as it removes things like GET parameters, and "i0.wp.com/". I guess it's normalis(z)ing the URL.

projects/plugins/boost/app/modules/image-size-analysis/Image_Size_Analysis_Fixer.php
donnchawp: In the way we're working now, it should be, as we're calling it via filters in the loop. Once we start fixing images outside the loop, when we use an output buffer this will have to change, so I'll investigate.

projects/plugins/boost/app/modules/image-size-analysis/Image_Size_Analysis_Fixer.php
donnchawp: Good catch!

projects/plugins/boost/app/modules/image-size-analysis/Image_Size_Analysis_Fixer.php
pyronaur: Hmm, a phpDoc with examples would be nice to have there then :) - I don't think there's much difference between fix and normalize.

projects/plugins/boost/app/modules/image-size-analysis/Image_Size_Analysis_Fixer.php
pyronaur: If I've ever learnt anything with WordPress is that never assume anything 😛  - using `get_post` would probably solve that though.

projects/plugins/boost/app/rest-api/endpoints/Image_Analysis_Fix.php
pyronaur: @dilirity now that's a reaaaly nice catch 👍

projects/plugins/boost/app/modules/image-size-analysis/Image_Size_Analysis_Fixer.php
donnchawp: It's the URL. I'll rename it $image_url.

projects/plugins/boost/app/modules/image-size-analysis/Image_Size_Analysis_Fixer.php
donnchawp: Done!

projects/plugins/boost/app/modules/image-size-analysis/Image_Size_Analysis_Fixer.php
donnchawp: Added a phpDoc to the function.

projects/plugins/boost/app/modules/image-size-analysis/Image_Size_Analysis_Fixer.php
donnchawp: Done

projects/plugins/boost/app/rest-api/endpoints/Image_Analysis_Fix.php
donnchawp: Added validation and an "_" at the start of the postmeta field name. Thanks!

projects/plugins/boost/app/rest-api/endpoints/Image_Analysis_Fix.php
donnchawp: I've just addressed that. The button won't toggle to the other state if no changes were made or saved. There's no indication in the UI, apart from the button not changing.

projects/plugins/boost/app/modules/image-size-analysis/Image_Size_Analysis_Fixer.php
donnchawp: Done

projects/plugins/boost/app/modules/image-size-analysis/Image_Size_Analysis_Fixer.php
donnchawp: Using get_post() now.

projects/plugins/boost/app/admin/class-config.php
```diff
@@ -53,6 +59,8 @@ public function constants() {
 			'nonces'        => Nonce::get_generated_nonces(),
 		);
 
+		$constants['autoFix'] = defined( 'AUTOFIX' ) && AUTOFIX ? true : false;
```
dilirity: Before releasing this out into the wild - let's make `AUTOFIX` more unique, so there's no collision with anything else.

donnchawp: Good idea. Done!

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/recommendations/row-types/ImageSizeRow.svelte
donnchawp: We're going with api.post() just to get this merged, and then refactor it along with everything else afterwards.

haqadn (approved):
Tested the feature and it works. The only thing sticks out is the data-sync integration. However, DS may need some adjustments to accommodate this as well so we can address this later before releasing the feature.

Issue #26792, opened 2022-10-12
Non-plain-text block description is being deprecated in WordPress 6.2


anomiex:
### Impacted plugin

Jetpack

### Steps to Reproduce

1. Run Jetpack (or its extension tests) with a version of WordPress or the Gutenberg plugin that provides @wordpress/blocks 11.18.0 or later.
2. See deprecation warnings.

Affected blocks are [recurring-payments](https://github.com/Automattic/jetpack/blob/2c48c9f31f6284b8aef0a8b986f14b9ff19fdc62/projects/plugins/jetpack/extensions/blocks/recurring-payments/index.js#L37-L42), [subscriptions](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/jetpack/extensions/blocks/subscriptions/index.js#L25-L43), [simple-payments](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/jetpack/extensions/blocks/simple-payments/index.js#L32-L45), and [wordads](https://github.com/Automattic/jetpack/blob/2c48c9f31f6284b8aef0a8b986f14b9ff19fdc62/projects/plugins/jetpack/extensions/blocks/wordads/index.js#L24-L31).

### A clear and concise description of what you expected to happen.

No deprecation warnings.

### What actually happened

"Declaring non-string block descriptions is deprecated since version 6.2."

### Platform (Simple, Atomic, or both?)

Simple, Atomic, Self-hosted

### Reproducibility

Consistent

### Available workarounds?

No but the platform is still usable

kraftbj:
Asked on the Gutenberg PR what's the preferred way to pass a link into a description https://github.com/WordPress/gutenberg/pull/44455#issuecomment-1276461347

jeherve:
I've been looking into this, and I'm not sure what's the best approach we could take.

### Remove the existing custom description

This would be the nuclear option. We would lose a lot of data for each block; we do use custom descriptions with multiple paragraphs and external links in at least 9 blocks in Jetpack at the moment. See #29757 for implementation.

### Add a new panel to each block

We could manually add `InspectorControls` with a `PanelBody` to each one of the blocks where we have a custom description with a link, but this would separate our custom description from the default description.

We could extract this to Jetpack's `registerJetpackBlock`, by using the `editor.BlockEdit` filter to add a custom `InspectorControls` with a custom description to each block where a custom description was passed when registering the block.

[...]

PR #33086, opened 2023-09-14
Refactor production blocks registration 1

monsieur-z:

## Proposed changes:

Registers the following production blocks by specifying the path to their block.json file:  _AI Assistant_, _Blogging Prompt_, _Calendly_, _Contact Info_, _Cookie Consent_.

  - Remove `icon.js`
  - Fix JS tests

### Other information:

## Jetpack product discussion

pedMtX-RS-p2

No.

## Testing instructions:

- Spin up a test site

Note: I haven't been able to test the _Cookie Consent_ block.


projects/plugins/jetpack/extensions/blocks/ai-assistant/ai-assistant.php
```diff
@@ -28,7 +25,7 @@ function register_block() {
 		|| ! ( new Status() )->is_offline_mode()
 	) {
 		Blocks::jetpack_register_block(
-			BLOCK_NAME,
+			__DIR__,
```
andrii-lysenko, line 29: Not the biggest concern, but I find it a bit sensitive, as it is possible that at some point we will have a block that will be renamed, or the folder and we might have a mismatch

andrii-lysenko (commented):
The code looks good and tests well locally.

I have a small nitpick, but it's not mandatory so you could decide wether change it or not. 

The only thing - is that one test from [PHP tests: PHP 8.2 WP trunk](https://github.com/Automattic/jetpack/actions/runs/6483692177/job/17605875077?pr=33086#logs) is failing. Is it a concern?

projects/plugins/jetpack/extensions/blocks/ai-assistant/ai-assistant.php
monsieur-z: Having a matching folder name is already a requisite, I believe. We could make sure to document it and add a check in the build process if we feel that's a concern.

monsieur-z:
Thanks @andrii-lysenko! I commented inline.

> [The only thing - is that one test from [PHP tests: PHP 8.2 WP trunk](https://github.com/Automattic/jetpack/actions/runs/6483692177/job/17605875077?pr=33086#logs) is failing. Is it a concern?

No, tests for that version of PHP have been flaky lately (I believe this is due to some WordPress updates) and are not currently required. I'll check again after the rebase just to be sure, though.

andrii-lysenko (dismissed):
LGTM!

projects/plugins/jetpack/extensions/blocks/ai-assistant/editor.js
```diff
@@ -1,4 +1,39 @@
...
+ * Supports and extensions
+ */
+import './supports';
+import './extensions/ai-assistant';
+import './extensions/jetpack-contact-form';
+
+registerJetpackBlockFromMetadata( metadata, {
+	edit,
+	save: () => null,
+	// The `description` property in `block.json` doesn't support formatting or markup.
```
jeherve: I wonder if we should stop adding any kind of non-string in there, for all our blocks? That would allow us to close #26792 in the process. :)

monsieur-z: I originally removed all links since the block description component in the block inserter cannot be hovered, and thus, users could not interact with the links. But then I realized users could click on the links from the Jetpack sidebar and added them back.

But now we have a valid reason to keep it simple! Thanks for the heads-up @jeherve.

jeherve (commented):
I'm noticing an issue with style previews for the Calendly block: both styles look the same with this branch:

Screenshot 2023-10-23 at 19 29 13

jeherve (approved):
This tests well. I think this is ready to merge! 🥳 

I only have one remark, but I do not think it is blocker; it could be fixed in a follow-up PR.

It looks like we're missing chocolate chips on our cookie:

**Production**

Screenshot 2023-10-24 at 19 14 35

**This PR**

Screenshot 2023-10-24 at 19 14 50

PR #33095, opened 2023-09-14
Add jetpack endpoint for jetpack product data

CodeyGuyDylan:
## Proposed changes:

Add infrastructure to query and store product data into state. Right now it will not be used anywhere

### Other information:

## Jetpack product discussion

pbNhbs-80w-p2

We are pulling data from WPCOM to store Jetpack product data into state. Right now that data will not be used and this PR does not add any tracking

## Testing instructions:

This has to be tested in tandem with D121691-code

1. Checkout this branch locally
2. Go to `projects/packages/my-jetpack/_inc/components/my-jetpack-screen/index.jsx` and add the following import
```jsx
import useProductData from '../../hooks/use-product-data';
```
and add the hook to line 94
```
const { productData, fetchingProductData } = useProductData();
```
This will call the API so it can be viewed in the Network panel

3. Checkout this diff: D121691-code and follow the instructions on that diff to finish testing this PR

### DIFF
--- a/projects/packages/my-jetpack/_inc/state/resolvers.js
+++ b/projects/packages/my-jetpack/_inc/state/resolvers.js
@@ -13,6 +13,7 @@ import {
 	REST_API_SITE_PRODUCTS_ENDPOINT,
 	REST_API_CHAT_AVAILABILITY_ENDPOINT,
 	REST_API_CHAT_AUTHENTICATION_ENDPOINT,
+	REST_API_SITE_PRODUCT_DATA_ENDPOINT,
 	PRODUCTS_THAT_NEEDS_INITIAL_FETCH,
 } from './constants';
 import resolveProductStatsRequest from './stats-resolvers';
@@ -129,6 +130,19 @@ const myJetpackResolvers = {
 				dispatch.setAvailableLicensesIsFetching( false );
 			}
 		},
+
+	getProductData: () => {
+		async ( { dispatch } ) => {
+			dispatch.setProductDataIsFetching( true );
+
+			try {
+				dispatch.setProductData( await apiFetch( { path: REST_API_SITE_PRODUCT_DATA_ENDPOINT } ) );
+				dispatch.setProductDataIsFetching( false );
+			} catch ( error ) {
+				dispatch.setProductDataIsFetching( false );
+			}
+		};
+	},
 };
 
 const getProductStats = {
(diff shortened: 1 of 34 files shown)


projects/packages/my-jetpack/src/class-rest-product-data.php
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	/**
+	 * Gets the product data for all products
+	 *
+	 * @return array|WP_Error
+	 */
+	public static function get_all_product_data() {
+		$site_id        = \Jetpack_Options::get_option( 'id' );
+		$wpcom_endpoint = sprintf( 'sites/%d/jetpack-product-data/backup?locale=%2$s&force=wpcom', $site_id, get_user_locale() );
```
jboland88: This looks like the backup specific endpoint, is that intentional?

CodeyGuyDylan: oops, no it is not, let me remove that!

Doesn't matter too much right now since it's not really being used but yes that was accidental

projects/packages/my-jetpack/_inc/state/resolvers.js
```diff
@@ -129,6 +130,19 @@ const myJetpackResolvers = {
...
+
+	getProductData: () => {
+		return async ( { dispatch } ) => {
+			dispatch.setProductDataIsFetching( true );
+
+			try {
+				dispatch.setProductData( await apiFetch( { path: REST_API_SITE_PRODUCT_DATA_ENDPOINT } ) );
+				dispatch.setProductDataIsFetching( false );
+			} catch ( error ) {
+				dispatch.setProductDataIsFetching( false );
```
robertsreberski, line 142: You can use try / catch / finally. and in finally run `dispatch.setProductDataIsFetching( false );`. But it's very optional nitpick

robertsreberski (dismissed):
LGTM, Works great in tandem with wpcom diff.

projects/packages/my-jetpack/_inc/state/resolvers.js
CodeyGuyDylan: I think that seems unnecessary in this case personally. I do it used elsewhere in this file but for the most part it is not. I'll leave it off for now 😄

robertsreberski (approved):
LGTM after rebase

PR #3317, opened 2016-01-21
Tracks: prefix all functions with jetpack_

jeherve:
Fixes #3316

PR #33096, opened 2023-09-15
CRM: Resolves #3317 - prevent tag slug duplication and better fallback support

tbradsha:
Resolves Automattic/zero-bs-crm#3317 - prevent tag slug duplication and better fallback support

## Proposed changes:

This cleans up two fairly big issues:

1. Tags with different labels could be normalised to the same slug. We now make sure tags get a unique slug (incremented as needed).
2. The fallback tag schema (`tag-X`) capped at 1024, and looped through each iteration with a database call. We now make a single DB call to determine what X should be.

This does not address Automattic/zero-bs-crm#3316, which when resolved will contain a migration to resurface the duplicates.

### Other information:

## Testing instructions:

1. Go to any object's tag manager (e.g. Quotes): `/wp-admin/admin.php?page=tag-manager&tagtype=quote`
2. Add the following four tags: `cat` `Cat` `ПРОМО` `ПРОМО ПАКЕТ`

In `trunk`, you'll see the following slugs: `cat`, `cat`, `tag-1`, `tag-2`. Not easy to verify, but each subsequent `tag-X` will result in more database calls (e.g. `tag-8` will have 7 more database calls than `tag-1`).

In the `fix/crm/3317-better_generic_tag_fallback_handling` branch, you'll see the following slugs: `cat`, `cat-1`, `tag-1`, `tag-2`. Not easy to verify, but each `tag-X` will have the same amount of database calls.

### DIFF
--- a/projects/plugins/crm/includes/ZeroBSCRM.DAL3.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.DAL3.php
@@ -3746,7 +3746,7 @@ class zbsDAL {
 		$slug_exists = $this->tag_slug_exists( $obj_type_id, $slug );
 
 		// slug as provided doesn't exist, so use that
-		if ( ! $slug_exists ) {
+		if ( ! $slug_exists && $slug !== 'tag' ) {
 			return $slug;
 		}
 
--- a/projects/plugins/crm/includes/ZeroBSCRM.DAL3.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.DAL3.php
@@ -3753,7 +3753,7 @@ class zbsDAL {
 		$slug_base = $slug . '-';
 
 		// get last iteration of tag slug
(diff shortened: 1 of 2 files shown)


projects/plugins/crm/includes/ZeroBSCRM.DAL3.php
```diff
@@ -3708,62 +3713,59 @@ public function getTagObjStats($args=array()){
...
+		if ( ! $slug_exists && $slug !== 'tag' ) {
+			return $slug;
+		}
 
-                $i++;
+		$slug_base = $slug . '-';
 
-            }
+		// get last iteration of tag slug
+		$sql_query = 'SELECT CAST(TRIM(LEADING %s FROM zbstag_slug) AS SIGNED) AS next_slug_id FROM ' . $ZBSCRM_t['tags'] . ' WHERE zbstag_slug LIKE CONCAT(%s,"%") AND zbstag_objtype = %d ORDER BY next_slug_id DESC LIMIT 1000'; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
```
gogdzl: Why `DESC LIMIT 1000` and not `DESC LIMIT 1`?

gogdzl (changes requested):
Code looks good, and fixes this nasty problem that's been haunting us.
I just have one small question about the SQL.

projects/plugins/crm/includes/ZeroBSCRM.DAL3.php
tbradsha: Because I left it in a debug state...or I wanted to make sure you were paying attention? 🤪 Addressed in 79ac83957e, and I also changed the column alias.

gogdzl (approved):
LGTM!

PR #33098, opened 2023-09-15
launchpad: hide modal on first post for bloggers

roo2:
Draft PR because we still need to decide whether this is the right approach:

For wpcom simple sites, we've found an issue where the "First post congratulations" modal and the launchpad modal are both shown at the same time.

p1694530274144539-slack-C057AH42XQD

We had code in the launchpad modal which was supposed to hide it on the first publish of the first post. Unfortunately, the logic checks for `isPublishingPost` which is true on the fist render cycle, but then there is another render cycle where `isPublishingPost` is now false, but `isSavingPost` is still true. This caused the modal to be shown anyway.

With this PR, we will check the original state of `hasNeverPublishedPostOption` instead. So any saves on the first load of the editor for a blog (`site_intent === 'write'`) will not trigger the launchpad modal.

No

## Testing instructions:

Honestly, I haven't done jetpack development for a while, here is how I tested it.

`const { tracks } = useAnalytics();` [on line 88]( https://github.com/Automattic/jetpack/compare/fix/launchpad-hide-modal-on-first-post?expand=1#diff-13206839dd3315d528c316c3405fbf89039de38e648bc7b16ac786db91fbb37dL88) threw an error in my sandbox, so I had to comment it out for testing.
```
#build code: 
jetpack watch plugins/jetpack
#upload to sandbox
jetpack rsync jetpack <SANDBOX_USER>@<MY_SANDBOX>:~/public_html/wp-content/mu-plugins/jetpack-plugin/production
```

I then set `has_never_published_post` option to simulate a new site.
```
#in wpsh
update_blog_option(<SITE_ID>, 'has_never_published_post', true);
```

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-launchpad-hide-modal-on-first-post
+++ b/projects/plugins/jetpack/changelog/fix-launchpad-hide-modal-on-first-post
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
+hide launchpad modal on first post for bloggers
--- a/projects/plugins/jetpack/extensions/plugins/launchpad-save-modal/index.js
+++ b/projects/plugins/jetpack/extensions/plugins/launchpad-save-modal/index.js
@@ -39,25 +39,18 @@ const updateLaunchpadSaveModalBrowserConfig = config => {
 
 export const settings = {
 	render: function LaunchpadSaveModal() {
-		const {
-			isSavingSite,
-			isSavingPost,
-			isPublishingPost,
-			isCurrentPostPublished,
-			postLink,
-			postType,
-		} = useSelect( selector => ( {
-			isSavingSite: selector( editorStore ).isSavingNonPostEntityChanges(),
-			isSavingPost: selector( editorStore ).isSavingPost(),
-			isPublishingPost: selector( editorStore ).isPublishingPost(),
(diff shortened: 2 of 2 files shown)


lupus2k (approved):
I tested with a new and old site with `write` intent, and it works as expected. 🎉

daledupreez (commented):
Thanks for tackling this, @roo2!

Without testing, I _think_ we should be able to move forward with your proposed approach as we're handling the following:
* If the site doesn't have the `write` intent, we don't show the "You published your first post" modal, and your changed logic won't be triggered because it was already specific to the `write` intent
* If the site does have the `write` intent, we should show the "You published your first post" modal for the first publish, which should show a "Next steps" button/link when the user is still busy with the fullscreen launchpad (i.e. the `launchpad_screen` option is set to `full`). If they stay on the editor, we should get them back to the launchpad if they try to get back, or we should get them back to the launchpad via the "Next steps" button. If they have skipped the fullscreen launchpad, they only need to get back to Customer Home, which should not be too complicated.

I think we definitely want these modals to be more coordinated in future, but that's definitely a bigger project.

cc @Automattic/serenity for visibility

roo2:
ok, I hit merge, just to confirm so the first time the editor loads (for `site_intent==write` users), it will no longer show the launchpad modal on save.

PR #33099, opened 2023-09-15
skip git internals with rsync

roo2:
I noticed running rsync, it was uploading .git internal files to my sandbox, I added a simple line to skip this:
No

### Testing instructions 

I was using this on my wpcom sandbox.
```
jetpack build plugins/jetpack

jetpack rsync jetpack <SANDBOX_USER>@<SANDBOX_SERVER>:~/public_html/wp-content/mu-plugins/jetpack-plugin/production
```

With this change , `.git/**` blobs should no longer be uploaded, speeding up the rsync

### DIFF
--- a/tools/cli/commands/rsync.js
+++ b/tools/cli/commands/rsync.js
@@ -228,6 +228,9 @@ async function addVendorFilesToFilter( source, filters ) {
  */
 async function rsyncToDest( source, dest, pluginDestPath ) {
 	const filters = new Set();
+
+	// Exclude git internals
+	filters.add( '- .git' );
 	// To catch files required in dev builds.
 	await addVendorFilesToFilter( `${ source }/vendor/`, filters );
 	await buildFilterRules( source, '', filters );
--- a/tools/cli/commands/rsync.js
+++ b/tools/cli/commands/rsync.js
@@ -229,7 +229,7 @@ async function addVendorFilesToFilter( source, filters ) {
 async function rsyncToDest( source, dest, pluginDestPath ) {
 	const filters = new Set();
 
-	// Exclude git internals
+	// Exclude any `.git` dirs, mostly in case someone ran composer with --prefer-source (or composer fell back to that).
 	filters.add( '- .git' );
 	// To catch files required in dev builds.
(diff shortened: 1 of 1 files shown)


anomiex:
> I noticed running rsync, it was uploading .git internal files to my sandbox

I cannot reproduce this. There shouldn't be any `.git` dirs in the stuff that's synced.

Did you do something weird like run `composer install --prefer-source` or otherwise tell composer to use that option? Or did Composer have a network hiccup and fall back to that for some package? That'll bring in a lot more than just `.git` files though, like all the tests and other normally excluded files in the upstream repo.

dereksmart:
Hm, I also can't reproduce. I also tried to manually add a .git/test.txt in the root of projects/plugins/jetpack/ and it wasn't transferred. The change should be harmless, however it would be helpful to figure out how your situation is such

roo2:
hmm yea, for me it was `/.git/` folders in the `vendor` dependencies. I didn't run `composer install --prefer-source` by intention, but maybe that's what happened somehow 😅.

Do you think it's still worth adding this check just incase any one else gets into this state, or even deliberately uses `--prefer-source`? I don't think we'd ever want to include `.git` internals in the sync

tools/cli/commands/rsync.js
```diff
@@ -228,6 +228,9 @@ async function addVendorFilesToFilter( source, filters ) {
  */
 async function rsyncToDest( source, dest, pluginDestPath ) {
 	const filters = new Set();
+
+	// Exclude git internals
```
anomiex: Let's make this clearer about what we're doing.
```suggestion
	// Exclude any `.git` dirs, mostly in case someone ran composer with --prefer-source (or composer fell back to that).
```

anomiex (commented):
> Do you think it's still worth adding this check just incase any one else gets into this state, or even deliberately uses `--prefer-source`? I don't think we'd ever want to include `.git` internals in the sync

Shouldn't hurt anything. Left a suggestion inline for making the comment clearer.

anomiex (approved):
LGTM

PR #33101, opened 2023-09-15
Boost: Remove global state from Score component

dilirity:
## Proposed changes:

* Moves global store to parent component in favor of props.

### Other information:

## Jetpack product discussion

pc9hqz-286-p2

no

## Testing instructions:

* Make sure Score section is working as expected (toggling a feature on or off should refresh the scores).
CleanShot 2023-09-15 at 11 55 12


projects/plugins/boost/app/assets/src/js/pages/settings/Settings.svelte
```diff
@@ -15,14 +17,25 @@
...
 <ReRouter to="/getting-started" when={$shouldGetStarted}>
 	<div id="jb-dashboard" class="jb-dashboard jb-dashboard--main">
 		<Header />
 
 		<div class="jb-section jb-section--alt jb-section--scores">
-			<Score />
+			<Score
+				{activeModules}
+				criticalCssCreated={$criticalCssState.created}
+				criticalCssIsGenerating={$isGenerating}
```
pyronaur: During the refactor, I prefer to pull these up to JS scope so that the code that touches component has no `$` store connection - I think that's going to make refactoring easier down the line when we need to actually refactor those reactive values.

pyronaur (dismissed):

projects/plugins/boost/app/assets/src/js/pages/settings/Settings.svelte
dilirity: I've moved the stores to the Main component.

However, is there an alternative to passing props down the chain?

I kind of don't like doing it this way:
https://github.com/Automattic/jetpack/pull/33101/files#diff-cb1f7a8b9eab8e3b52de49d7a814a20ac5be9d6e2109ac5af2ec15c0e2d5277eR47-R49

and

https://github.com/Automattic/jetpack/pull/33101/files#diff-02098ca8bbf7a7f4d25f317f6e9360afa2a7d75284c249bbb53d0c72567cb7b6R30

pyronaur (approved):
Looks good 👍

PR #33106, opened 2023-09-15
Social Change/move auto conversion notice up

gmjuhasz:
This changes moves our auto-conversion notice near the other Instagram notice so it makes more sense for the featured image validation.

### Other information:

## Jetpack product discussion

https://github.com/orgs/Automattic/projects/733/views/2?pane=issue&itemId=38718952

No.
## Testing instructions:

* Select a media that is convertible and not valid

| Before  | After |
| CleanShot 2023-09-15 at 14 29 51 png | CleanShot 2023-09-15 at 14 28 17 png |

### DIFF
--- a/projects/js-packages/publicize-components/changelog/change-move-auto-conversion-notice-up
+++ b/projects/js-packages/publicize-components/changelog/change-move-auto-conversion-notice-up
@@ -1,0 +1,4 @@
+Significance: minor
+Type: changed
+
+Move auto-conversion notice near the Instagram one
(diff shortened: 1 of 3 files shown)


pablinos:
Changelogger is complaining, but otherwise this looks good.

CleanShot 2023-09-15 at 14 05 39

I wonder if we should change this to 'the image' or something else, because it doesn't make as much sense when it's the featured image. What do you think?

pablinos (approved):

PR #33108, opened 2023-09-15
Boost: Remove global state from Hero, Tabs and Table components

dilirity:
## Proposed changes:

* Moves global state to parent component in favor of props;

### Other information:

## Jetpack product discussion

pc9hqz-286-p2

no

## Testing instructions:

* Make sure the "Hero" section of the Image Size Analysis works as expected.
* Request a new report and open the detailed page - it should dynamically update as the report progresses.
image

* Make sure the "Tabs" section of the Image Size Analysis works as expected.
CleanShot 2023-09-15 at 17 49 17

* Make sure the "Table" section of the Image Size Analysis works as expected.
CleanShot 2023-09-15 at 17 48 24


projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/RecommendationsPage.svelte
```diff
@@ -37,12 +50,26 @@
...
 		<div class="jb-container">
-			<Hero {needsRefresh} {refresh} />
-			<Tabs />
+			<Hero
+				{needsRefresh}
+				{refresh}
+				{isImageCdnModuleActive}
+				{isaLastUpdated}
+				{hasActiveGroup}
+				totalIssueCount={$totalIssueCount}
```
pyronaur: We'll eventually get rid of global variables everywhere, might be a good idea to group this with the rest

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/RecommendationsPage.svelte
```diff
@@ -37,12 +50,26 @@
...
 
 		<div class="table-wrap">
-			<Table {needsRefresh} {refresh} />
+			<Table
+				{needsRefresh}
+				{refresh}
+				isaDataLoading={$isaDataLoading}
+				{activeGroup}
+				{images}
+				isaSummary={$isaSummary}
```
pyronaur: Same here

pyronaur (dismissed):
Looks good, but left suggestions

projects/plugins/boost/app/assets/src/js/modules/image-size-analysis/RecommendationsPage.svelte
dilirity: I'm not sure I understand what you mean by "group this with the rest" 😅

pyronaur: Oh, sorry, I meant that `totalIssueCount={$totalIssueCount}` could become `totalIssueCount={issueCount}` and instead define on top of the file that `$: issueCount = $totalIssueCount` or something along those lines - so that global store subscriptions aren't present in the template side of things - it's a minor thing, but I think it will help conversion to react later.

dilirity: Thanks! Updated in 6fbd59d.

pyronaur (approved):
:shipit:

PR #33110, opened 2023-09-15
Boost: Remove global state from Pagination component

dilirity:
## Proposed changes:

* Moves global state to parent component in favor of props;
* Update next/previous buttons to use a separate component.

### Other information:

## Jetpack product discussion

pc9hqz-286-p2

no

## Testing instructions:

* Make sure ISA Pagination is working as expected.
CleanShot 2023-09-18 at 16 56 59


pyronaur (dismissed):
Clicking left and right arrows no longer works because the connection to `current` reactivity is now lost.

pyronaur:
To fix that, I think `<button>` needs to become a `<Link>` instead and link to the relevant page.

dilirity:
@pyronaur great catch - thanks!

I've updated the buttons to use the Link component.

~~I've also updated their behavior - they'll hide when necessary, instead of always showing. I kind of felt it didn't make sense to show them and disable them, as they are very tiny.~~

pyronaur (dismissed):
Looks like this might work. Although you could use `slot` instead of `if right` check, removing a condition - would make the component a bit simpler imho.

But it's fine if you prefer to keep it this way.

dilirity:
Thanks for the suggestion @pyronaur! I've updated the pagination arrow component to use slots in 02584d3.

pyronaur (commented):
Looks good! :shipit:

pyronaur (approved):

PR #33121, opened 2023-09-15
CRM: Resolves #3316 - better slug generation

### LINKED ISSUE #3316: Fatal Error when Mojo Marketplace is Present
When the Mojo Marketplace plugin is present, activating Jetpack 3.9.0 triggers the following fatal error:

```
Fatal error: Cannot redeclare tracks_record_event_raw() (previously declared in
/public_html/wp-content/plugins/mojo-marketplace-wp-plugin/tests/jetpack-onboarding-tracks/lib/tracks/client.php:29)
in /public_html/wp-content/plugins/jetpack/_inc/lib/tracks/client.php on line 31
```

There was no conflict with Mojo Marketplace and Jetpack 3.8.2.

tbradsha:
Resolves Automattic/zero-bs-crm#3316 - better slug generation

## Proposed changes:

In #33096 we solved issues with duplicates and fallback slugs. This PR switches tag slug generation from `$zbs->DAL->makeSlug()` to WP's native `sanitize_key()` (initially I looked at both `sanitize_title()` and `sanitize_title_with_dashes()`, but neither limits to alphanumeric chars + dashes + underscores despite what their docs say).

It also adds a new migration that regenerates tag slugs. I wanted to simplify the logic, but was struggling to do so while covering all scenarios. Feel free to suggest more condensed code! This is the current logic:

```
1. Generate slug:
    * If blank, use `tag` base:
        * If old slug looks like /^tag-\d+$/ and it's not been used, use it.
        * Otherwise, generate a new `tag-N` slug iteration. It'll be unique, as we check the database and increment by one.
    * If not blank:
        * If it hasn't been used, use it.
        * If it has been used, but the old slug was iterated and has the expected base and is available, use old slug.
        * If the new and old slug are unavailable, iterate.
2. Store slug in database if it doesn't match its old slug.
3. Cache slug in `$used_slugs`, preventing reuse by other tags.
```

### Other information:

## Testing instructions:

Start in `trunk`. Add tags like with a mix of ASCII and non-ASCII chars (e.g. `Product: (ПРОМО ПАКЕТ)`). Also, add some tags with duplicate slugs via the database.

In `trunk`, you'll see duplicate slugs, and the mixed-char labels will have a slug of `tag-N` on most systems.

Then switch to the `fix/crm/3316-better_slug_generation` branch. The migration should run, fixing duplicates and giving the mixed-char tags a slug derived from the ASCII chars (not always pretty, but more unique than `tag-N`).

### DIFF
--- a/projects/plugins/crm/changelog/fix-crm-3316-better_slug_generation
+++ b/projects/plugins/crm/changelog/fix-crm-3316-better_slug_generation
@@ -1,0 +1,5 @@
+Significance: patch
+Type: fixed
+
+Tags: better slug generation
+Tags: added tag slug migration
--- a/projects/plugins/crm/includes/ZeroBSCRM.DAL3.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.DAL3.php
@@ -3199,7 +3199,7 @@ class zbsDAL {
             if (!isset($data['name']) || empty($data['name'])) return false;
 		if ( empty( $data['slug'] ) ) {
 
-			$potential_slug = $this->makeSlug( $data['name'] ); // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable
+			$potential_slug = sanitize_key( $data['name'] ); // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable
 
 			// catch empty slugs as per gh-462, chinese characters, for example
 			if ( empty( $potential_slug ) ) {
--- a/projects/plugins/crm/includes/ZeroBSCRM.DAL3.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.DAL3.php
@@ -3203,12 +3203,12 @@ class zbsDAL {
 
 			// catch empty slugs as per gh-462, chinese characters, for example
 			if ( empty( $potential_slug ) ) {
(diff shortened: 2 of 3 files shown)


gogdzl:
The migration produced some weird slugs.

From:
image

To:
image

Worked well for all my quote cat tags:

image

gogdzl (changes requested):
Code looks good, but I left a comment about some weird tags. I don't know if that is the expected behavior or not. (e.g. isn't expected 'Product: First Hoodie' to generate the tag 'product-first-hoodie'?)

tbradsha:
Rebased.

tbradsha:
> I left a comment about some weird tags. I don't know if that is the expected behavior or not. (e.g. isn't expected 'Product: First Hoodie' to generate the tag 'product-first-hoodie'?)

This is indeed expected behaviour when using WP's built-in `sanitize_key()`. While we could write our own wrapper or filter that modifies the function, that's somewhat the reason we had the problem in the first place, and I think it best to use the existing function as-is.

As mentioned in the original description, originally I considered `sanitize_title()` and `sanitize_title_with_dashes()`. However, despite the docs claiming they only do alphanumeric, dashes, and underscores, both convert the Cyrillic to `%xx` chars, which are definitely not as friendly to work with.

gogdzl (approved):
It works as advertised. Nice job, @tbradsha =)

PR #33126, opened 2023-09-15
phpcs: Use PHPCompatibilityWP for dev-develop run

anomiex:
## Proposed changes:

Because PHPCompatibility hasn't had a release since 2019, we have CI run with their dev-develop branch to catch issues relevant to PHP 8. But when doing so, we weren't respecting the rules that PHPCompatibilityWP disables because WordPress provides polyfills. We had a manual list of a few such functions, but better would be to actually load PHPCompatibilityWP. This PR does that.
    
However, an upside to the lack of checking for WordPress's polyfills was that we caught cases where code that is intended to run outside of WordPress (and so doesn't have those polyfills) tried to use these functions. Since we're losing that, let's make it happen properly by creating a "Jetpack-Compat-NoWP" ruleset that turns the relevant rules back on.

Then I went on to do some more DRY, introducing "Jetpack-NoWP" and "Jetpack-Tests" rulesets, and some cleanup to remove no-longer-needed ignore directives and cmpatibility code paths.

### Other information:

## Jetpack product discussion

None

No

## Testing instructions:

* Is CI happy?


anomiex:
Split into #33287 and #33288.

PR #33128, opened 2023-09-16
[CRM] Remove is_admin bootstrap condition for automations v2

kallehauge:
Remove the (at the time it was added) temporary `is_admin` condition while we figured out what to do to prevent logic leaking into the front-end.

This is bigger issue and will be looked into separately.

Fixes https://github.com/Automattic/zero-bs-crm/issues/3314

## Proposed changes:

* Remove `is_admin` condition around Automations bootstrap logic.

### Other information:

## Jetpack product discussion

pbhBOz-2P4-p2#comment-4943

No.

## Testing instructions:

If our automated integration tests still work, then everything should be good to go.

### DIFF
--- a/projects/plugins/crm/changelog/crm-update-remove-is-admin-bootstrap-condition
+++ b/projects/plugins/crm/changelog/crm-update-remove-is-admin-bootstrap-condition
@@ -1,0 +1,5 @@
+Significance: patch
+Type: changed
+Comment: Automations v2 is still in development, so this is an invisible change
+
+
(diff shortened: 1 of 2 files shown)


cleacos (approved):
The code looks good.

About `while we figured out what to do to prevent logic leaking into the front end.` We only have to load triggers (which add WP hooks), only from active workflows. That is minimal stuff to load, I don't think we should over-engineer it, trying to optimize a thing that is not a super-blocking code (read from DB active (indexed) workflows and add WP hooks).

LGTM!

PR #33129, opened 2023-09-17
CRM: Add automation workflows table

kallehauge:
This PR introduces a new automations workflow table.

It could be argued that the table should be named "automations" but that would IMO not be the correct name since a workflow is "a named datatype" in the overall automations systems and when someone wanted to look up a workflow, they would not look for the "automations" table.

An additional benefit is that it can co-exist next to the current automations table as well, so we're not stepping on any toes while hiding the new automation system behind a feature flag. 

Fixes https://github.com/Automattic/zero-bs-crm/issues/3198

**Update**

I've removed `zbs_team` as a column even though it was suggested by @cleacos.
The main reason being that I don't see us going to that level of customization any time soon where specific teams can set up their own workflows + "automations v1" didn't include it either.

Honestly, I contemplated not including the owner as well, but seeing who created the workflow is probably OK - but we're basically using it as "created by" instead of an "owner" at that point.

## Proposed changes:

* Add a new `workflows` table.
* Add a migration script that ensures the workflow table is added.

### Other information:

## Jetpack product discussion

* p1693381526547099/1693338179.961359-slack-CTXBP902X
* pbhBOz-2ES-p2
* pbhBOz-3Wf-p2

No.

## Testing instructions:

* Spin up a CRM installation using this patch
* Load any page (e.g. go to `/wp-admin`)
* Check that the new `*_zbs_workflows` table exists
* Check that the structure of the table matches what is described in the code


kallehauge:
I've removed `zbs_team` as a column even though it was suggested by @cleacos.
The main reason being that I don't see us going to that level of customization any time soon where specific teams can set up their own workflows + "automations v1" didn't include it either.

Honestly, I contemplated not including the owner as well, but seeing who created the workflow is probably OK - but we're basically using it as "created by" instead of an "owner" at that point.

gogdzl:
I'm getting the following error:

```
WordPress database error Invalid default value for 'created_at' for query CREATE TABLE IF NOT EXISTS wp_zbs_workflows(
	`id` INT NOT NULL AUTO_INCREMENT,
	`zbs_site` INT NOT NULL,
	`zbs_owner` INT NOT NULL,
	`name` VARCHAR(255) NOT NULL,
	`description` VARCHAR(255) NULL DEFAULT NULL,
	`category` VARCHAR(255) NOT NULL,
	`triggers` LONGTEXT NOT NULL,
	`initial_step` LONGTEXT NOT NULL,
	`active` TINYINT(1) NOT NULL DEFAULT 0,
	`version` INT(14) NOT NULL DEFAULT 1,
	`created_at` INT(14) NOT NULL DEFAULT CURRENT_TIMESTAMP,
	`updated_at` INT(14) NOT NULL DEFAULT CURRENT_TIMESTAMP,
	PRIMARY KEY (`id`),
	INDEX `name` (`name` ASC),
	INDEX `category` (`category` ASC),
	INDEX `created_at` (`created_at` ASC)
	) ENGINE = InnoDB
	DEFAULT CHARACTER SET = utf8
	COLLATE = utf8_general_ci 
```

gogdzl (dismissed):
I'm getting the following error:

```
WordPress database error Invalid default value for 'created_at' for query CREATE TABLE IF NOT EXISTS wp_zbs_workflows(
	`id` INT NOT NULL AUTO_INCREMENT,
	`zbs_site` INT NOT NULL,
	`zbs_owner` INT NOT NULL,
	`name` VARCHAR(255) NOT NULL,
	`description` VARCHAR(255) NULL DEFAULT NULL,
	`category` VARCHAR(255) NOT NULL,
	`triggers` LONGTEXT NOT NULL,
	`initial_step` LONGTEXT NOT NULL,
	`active` TINYINT(1) NOT NULL DEFAULT 0,
	`version` INT(14) NOT NULL DEFAULT 1,
	`created_at` INT(14) NOT NULL DEFAULT CURRENT_TIMESTAMP,
	`updated_at` INT(14) NOT NULL DEFAULT CURRENT_TIMESTAMP,
	PRIMARY KEY (`id`),
	INDEX `name` (`name` ASC),
	INDEX `category` (`category` ASC),
	INDEX `created_at` (`created_at` ASC)
	) ENGINE = InnoDB
	DEFAULT CHARACTER SET = utf8
	COLLATE = utf8_general_ci 
```

**MySQL version: 8.0.16 through PHP extension MySQLi**

projects/plugins/crm/includes/ZeroBSCRM.Database.php
```diff
@@ -879,6 +880,31 @@ function zeroBSCRM_createTables(){
...
+	`zbs_owner` INT NOT NULL,
+	`name` VARCHAR(255) NOT NULL,
+	`description` VARCHAR(255) NULL DEFAULT NULL,
+	`category` VARCHAR(255) NOT NULL,
+	`triggers` LONGTEXT NOT NULL,
+	`initial_step` LONGTEXT NOT NULL,
+	`active` TINYINT(1) NOT NULL DEFAULT 0,
+	`version` INT(14) NOT NULL DEFAULT 1,
+	`created_at` INT(14) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+	`updated_at` INT(14) NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
coder-karen: I wonder if this should be similar to the following?

```suggestion
	`created_at` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
	`updated_at` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
```

coder-karen (commented):
I get the same error as @gogdzl if I switch to the branch on an existing JN site, and if I attempt to spin up a new JN site from the branch directly it isn't possible ('Oh No! There was a problem launching the new WP. (The API responded in a weird way).').

projects/plugins/crm/includes/ZeroBSCRM.Database.php
cleacos: This^^. For using `CURRENT_TIMESTAMP` the column type should be `TIMESTAMP` OR `DATETIME`.

Also, for the `updated_at` column, after the `DEFAULT CURRENT_TIMESTAMP` clause we could add this `ON UPDATE CURRENT_TIMESTAMP`

cleacos: MySQL has this strange requirement: "one CURRENT_TIMESTAMP per table".

`Incorrect table definition; there can be only one TIMESTAMP column with CURRENT_TIMESTAMP in DEFAULT or ON UPDATE clause `

So let's do it programmatically. I'll fix it setting the default to 0.

tbradsha: Semantically, I'd prefer to leave `updated_at` as `NULL` until it's been updated (vs. just created).

projects/plugins/crm/includes/ZeroBSCRM.Database.php
```diff
@@ -879,6 +880,33 @@ function zeroBSCRM_createTables(){
...
+	// phpcs:disable WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
+	$sql = 'CREATE TABLE IF NOT EXISTS ' . $ZBSCRM_t['automation-workflows'] . '(
+	`id` INT NOT NULL AUTO_INCREMENT,
+	`zbs_site` INT NOT NULL,
+	`zbs_owner` INT NOT NULL,
+	`name` VARCHAR(255) NOT NULL,
+	`description` VARCHAR(600) NULL DEFAULT NULL,
+	`category` VARCHAR(255) NOT NULL,
+	`triggers` JSON NOT NULL,
+	`initial_step` VARCHAR(255) NOT NULL,
```
cleacos, line 893: The `initial_step` is the internal step ID (array key) inside of the workflow `steps`. It could be an integer or a string.

cleacos (dismissed):
I pushed some column updates. The table is created as expected.

cleacos (dismissed):
Conflict on migration file resolved.

coder-karen (dismissed):
Works well for me now, and the `zbs_workflows` looks good:

Screenshot 2023-09-20 at 12 28 25

projects/plugins/crm/includes/ZeroBSCRM.Database.php
coder-karen: Looks like it is now set to 0.

projects/plugins/crm/includes/ZeroBSCRM.Database.php
```diff
@@ -879,6 +880,33 @@ function zeroBSCRM_createTables(){
...
+	// Add table to store automation workflows.
+	// phpcs:disable WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
+	$sql = 'CREATE TABLE IF NOT EXISTS ' . $ZBSCRM_t['automation-workflows'] . '(
+	`id` INT NOT NULL AUTO_INCREMENT,
+	`zbs_site` INT NOT NULL,
+	`zbs_owner` INT NOT NULL,
+	`name` VARCHAR(255) NOT NULL,
+	`description` VARCHAR(600) NULL DEFAULT NULL,
+	`category` VARCHAR(255) NOT NULL,
+	`triggers` JSON NOT NULL,
```
kallehauge: **General context:**

The JSON data type isn't super common in the WordPress world. There tends to be a trend to serialize the data instead.

[WordPress.org server requirements](https://en-gb.wordpress.org/about/requirements/) mentions both MySQL and MariaDB:

> [MySQL](https://www.mysql.com/) version 5.7 or greater, OR [MariaDB](https://mariadb.org/) version 10.3 or greater.

**MySQL / MariaDB differences**

With the general trend to serialize arrays/objects combined with the lack of true support in MariaDB made me go with `LONGTEXT` instead of `JSON`: https://mariadb.com/kb/en/json-data-type/

projects/plugins/crm/includes/ZeroBSCRM.Database.php
```diff
@@ -879,6 +880,33 @@ function zeroBSCRM_createTables(){
...
+	`initial_step` VARCHAR(255) NOT NULL,
+	`steps` JSON NOT NULL,
+	`active` TINYINT(1) NOT NULL DEFAULT 0,
+	`version` INT(14) NOT NULL DEFAULT 1,
+	`created_at` TIMESTAMP NOT NULL DEFAULT 0,
```
kallehauge: MariaDB doesn't seem to have the same CURRENT_TIMESTAMP() limitation, so nice catch!
I do, however, feel like we should never allow" `created_at` to be null, so perhaps using it here makes sense?

cleacos: I preserved the `NOT NULL` in both, `created_at` and `updated_at. Both will be set programmatically on insert and `updated_at` on update. In PHP the int declaration, the default value is `0`, so... should we allow nullable values on both? (See Thomas comment)

projects/plugins/crm/includes/ZeroBSCRM.Migrations.php
```diff
@@ -38,6 +38,7 @@
 	'task_offset_fix', // removes task timezone offsets from database
 	'refresh_user_roles', // Refresh user roles
 	'regenerate_tag_slugs', // Regenerate tag slugs
+	'611', // 6.1.1 - Create "workflows" table.
```
tbradsha: Can we give this a descriptive name instead?

projects/plugins/crm/includes/ZeroBSCRM.Database.php
tbradsha: I strongly prefer not using 0 as a default for timestamps, as that's an actual timestamp. Why not do `NOW()` for `created_at` and `NULL` for `updated_at`?

Also, I'll note that the other tables use `created` and `lastupdated` if we want to be consistent.

projects/plugins/crm/includes/ZeroBSCRM.Migrations.php
```diff
@@ -1105,6 +1106,22 @@ function zeroBSCRM_migration_560() { // phpcs:ignore WordPress.NamingConventions
...
 
+/**
+ * Migration 6.1.1
+ *
+ * This migration will:
+ * - Make sure all tables are up-to-date. Practically speaking, then we're creating a new "workflows" table.
+ *
+ * @return void
+ */
+function zeroBSCRM_migration_611() { // phpcs:ignore WordPress.NamingConventions.ValidFunctionName.FunctionNameInvalid
```
tbradsha: See above; I'd prefer a descriptive name.

projects/plugins/crm/includes/ZeroBSCRM.Database.php
```diff
@@ -56,6 +56,7 @@
   $ZBSCRM_t['eventreminders']         = $wpdb->prefix . "zbs_event_reminders";
   $ZBSCRM_t['tax']                    = $wpdb->prefix . "zbs_tax_table";
   $ZBSCRM_t['security_log']           = $wpdb->prefix . "zbs_security_log";
+  $ZBSCRM_t['automation-workflows']  = $wpdb->prefix . 'zbs_workflows'; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase, Universal.WhiteSpace.PrecisionAlignment.Found
```
tbradsha, line 59: Let's avoid ignoring rules (`Universal.WhiteSpace.PrecisionAlignment.Found`) when possible...

projects/plugins/crm/includes/ZeroBSCRM.Database.php
cleacos: Okay let's do this:

`created_at`  => `NOT NULL DEFAULT NOW()`
`updated_at` => `DEFAULT NULL`

projects/plugins/crm/includes/ZeroBSCRM.Migrations.php
cleacos: I agree. A migration per context/issue is a better idea instead of per version:

`create_workflows_table`

projects/plugins/crm/includes/ZeroBSCRM.Migrations.php
cleacos: `zeroBSCRM_migration_ create_workflows_table()`

projects/plugins/crm/includes/ZeroBSCRM.Migrations.php
tbradsha: Per-context migrations allow for easy removal of obsolete migrations in the future, too.

projects/plugins/crm/includes/ZeroBSCRM.Database.php
cleacos: Yes, JSON type is a LONGTEXT for both, but with JSON we could perform some specific queries and also the internal validation they perform.

Using LONGTEXT will work anyway, for now, we will not perform any special query with the JSON data there. So fine with using `LONGTEXT`.

tbradsha (approved):

coder-karen:
It looks like this issue still isn't fully resolved, as testing new CRM PRs results in similar database errors (testing on Jurassic Ninja sites), only this time it is `Invalid default value for 'updated_at'`.
More in Slack: p1695366780951289-slack-CTXBP902X

PR #33133, opened 2023-09-18
Boost Score Graph: Improve new graph

haqadn:
## Proposed changes:
* Instead of showing empty space before the first datapoint, show graph lines coming to the first data point.
* Fix: don't show fresh start dialog if user has previously dismissed it.
* Data Sync: Load fallback value while merging an entry

### Other information:

## Jetpack product discussion
pc9hqz-28r-p2#comment-1720

No

## Testing instructions:
- Upgrade to boost premium
- Dismiss the prompt that says "Boost will now record performance"
- Refresh the page to make sure it doesn't appear again
- Ones there is initial data in the graph, make sure that the first datapoint has the graph lines continued from before the beginning of the graph.


projects/plugins/boost/app/assets/src/js/pages/settings/sections/History.svelte
```diff
@@ -74,6 +75,8 @@
 	<ReactComponent
 		this={PerformanceHistory}
 		onToggle={onToggleHistory}
+		isFreshStart={$dismissedAlerts.performance_history_fresh_start !== true}
+		onDismissFreshStart={() => ( $dismissedAlerts.performance_history_fresh_start = true )}
```
dilirity: Since we're removing global stores - can you bubble these up to the parent component?

You can see how I've done it here - https://github.com/Automattic/jetpack/pull/33278/files#diff-0c8a77ded084513471e3c3be7ce4faf020dc88955b75d71c02d29405b7e703ffL7-L8

haqadn: Updated https://github.com/Automattic/jetpack/pull/33133/commits/a92a7031f1ebcdeec564cf2d735b37256811fe6d

dilirity:
I just deployed a JN to test and the graph looks like this immediately after opening it:
CleanShot 2023-09-27 at 18 41 28

Leaving this here as I'll check again tomorrow when there's more data.

For now it appears to be working as expected.

projects/plugins/boost/app/data-sync/Mergeable_Array_Entry.php
```diff
@@ -0,0 +1,31 @@
...
+	public function __construct( $option_key ) {
+		$this->option_key = $option_key;
+	}
+
+	public function get( $fallback_value = false ) {
+		// WordPress looks at argument count to figure out if a fallback value was used.
+		// Only provide the fallback value if it's not the default ( false ).
+		if ( $fallback_value !== false ) {
+			return get_option( $this->option_key, $fallback_value );
+		}
```
dilirity, line 20: TIL about this.

Also - I wonder what made you add these lines of code. Any problems when passing `false` every time?

projects/plugins/boost/app/assets/src/js/stores/dismissed-alerts.ts
dilirity, line 1: In #33277 I added a Data Store for [dismissed_score_prompt](https://github.com/Automattic/jetpack/blob/ddcfc74819f0d76a8021e65f91e4445d95f67504/projects/plugins/boost/app/assets/src/js/stores/prompt.ts) for the score prompts. Wonder if it makes sense to create a follow up PR that moves them to this data store once this PR is merged.

What do you think?

projects/plugins/boost/app/assets/src/js/stores/dismissed-alerts.ts
```diff
@@ -0,0 +1,52 @@
...
+	dismissedAlertsSchema
+);
+
+type MapType = z.infer< typeof dismissedAlertsSchema >;
+
+/**
+ * This function creates an SyncedStoreCallback.
+ * It's wrapped in a function so that pendingValues can be scoped to this function.
+ */
+function createDismissedAlertSyncAction() {
```
dilirity, line 18: I'm still learning my way around DS. Could you explain why this is necessary?

As I understand, we only have the one alert for the History graph.

dilirity (changes requested):
This looks great!

I left some questions and am waiting for the JN to gather data so I can check the graph tomorrow.

Also - the changelogger isn't happy again 😞

dilirity:
This is what the graph looks today.

I know there's only data for 27 and 28th, but shouldn't it start from the 26th (just like it did for the 27th)?

CleanShot 2023-09-28 at 10 54 31

projects/plugins/boost/app/data-sync/Mergeable_Array_Entry.php
haqadn: I copied this part of the code from the existing get method from [`Data_Sync_Option`](https://github.com/Automattic/jetpack/blob/a92a7031f1ebcdeec564cf2d735b37256811fe6d/projects/packages/wp-js-data-sync/src/class-data-sync-option.php) class, which is part of data sync.

So the actual author is @pyronaur and he might be able to answer better. My understanding is that, depending on the presence of a provided default value, the `default_option_<option>` filter is applied. Now, if a default value wasn't provided, these lines of code will retain the default behaviour of WP.

projects/plugins/boost/app/assets/src/js/stores/dismissed-alerts.ts
haqadn: Yes!

projects/plugins/boost/app/assets/src/js/stores/dismissed-alerts.ts
haqadn: Since we are handling multiple dismissable alerts in multiple places, this would make them all manageable in the same place. I intend to remove different ways of dismissing alerts and make them consistent with DS. This will require a followup PR as you suggested.

haqadn:
> This is what the graph looks today.
> 
> I know there's only data for 27 and 28th, but shouldn't it start from the 26th (just like it did for the 27th)?
> 
> CleanShot 2023-09-28 at 10 54 31

Since you do not have real data on 26th. It makes sense to hide the fake data point. However, I see from your screenshot that it was showing the ticks for the 26th, which it shouldn't. I have fixed it in https://github.com/Automattic/jetpack/pull/33133/commits/a88fcd09582847686ae73ebdde451e671ae57a33.

projects/plugins/boost/app/assets/src/js/stores/dismissed-alerts.ts
dilirity: Neat. I've added the ticket for it (https://github.com/Automattic/jetpack/issues/33402).

dilirity:
> > This is what the graph looks today.
> > I know there's only data for 27 and 28th, but shouldn't it start from the 26th (just like it did for the 27th)?
> > CleanShot 2023-09-28 at 10 54 31
> 
> Since you do not have real data on 26th. It makes sense to hide the fake data point. However, I see from your screenshot that it was showing the ticks for the 26th, which it shouldn't. I have fixed it in [a88fcd0](https://github.com/Automattic/jetpack/commit/a88fcd09582847686ae73ebdde451e671ae57a33).

Great! Here's what the graph looks like after applying the latest update:

image

dilirity (dismissed):
:shipit:

dilirity:
This looks great! Screenshot from freshly deployed Jurassic Ninja:

CleanShot 2023-10-10 at 16 38 28

dilirity (approved):
:shipit:

PR #33118, opened 2023-09-15
Social Menu & Social Media Icons: Add X support

jeherve:
## Proposed changes:

Update the icon font, the social icons widget, the social menu, the social-logos package to include the new X icon.

Noting that this purposefully does not replace the existing Twitter icon. This will allow site owners to decide whether the icon they'd like to use. It should be less forceful that way, we will not update the existing Twitter icon that may be in use on many sites; site owners will do that themselves if they'd like. This is the approach that will most likely be taken for Core as well: https://github.com/WordPress/gutenberg/pull/54092

Related PR: https://github.com/Automattic/social-logos/pull/126/

image
image

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

* N/A

## Does this pull request change what data or activity we track or use?

* No

## Testing instructions:

* Go to `yoursite.com/wp-content/plugins/jetpack/_inc/social-logos/social-logos.html`
* Check the new logos
* Go to Jetpack > Settings and enable extra sidebar widgets.
* Install and activate the Dara theme on your site.
* Go to Plugins > Add New and install the Classic Widgets plugin.
* Go to Appearance > Customize > Menus
* Create a new menu with some links, e.g. `https://www.facebook.com/jetpackme` and `https://x.com/jetpack`
* Assign that menu to the Social Menu location
* Go to Appearance > Widgets menu
* Drag the Social Icons widget to one of the active widget areas, and add some links again.
* Check how things look like on the frontend, for both your social menu and the Social Icons Widget you created.
* Go to Jetpack > Settings > Social and enable sharing buttons. **Do not look for a X icon just yet, we are just checking to ensure there are no regressions with the existing icons**.
* Go to Settings > Sharing and add some sharing buttons to your site. Use the icon-only sharing type.
* The buttons should look good in the admin preview, as well as on the frontend when your changes are saved.

PR #33134, opened 2023-09-18
Sharing: Add X button

### LINKED ISSUE #32178: Sharing / Social: update Twitter logo
### What

Update the logo for Twitter to the new logo

Feature request from user: 6586135-zd-woothemes

### Why

Twitter has recently updated its logo: https://techcrunch.com/2023/07/24/twitter-has-officially-changed-its-logo-to-x/

### How

_No response_

---

_Just adding a couple places where this appears; feel free to update with more (-cuemarie)_:

- [ ] Social Icons block
- [ ] Sharing Buttons

### LINKED ISSUE #33315: Enhancement: Change Twitter share button to X.
### Impacted plugin

Social

### What

Need to change twitter function to X.

### How

_No response_

jeherve:
Fixes #32178
Fixes #33315

## Proposed changes:

This PR does 2 things:

1. It adds a new button for the X service.
2. It changes the default sharing buttons added on sites that enable the module for the first time from Twitter and Facebook to Facebook and X.

**Notes**

- It **does not** replace the existing Twitter button, for the reasons we discussed in #33118. This is less forceful on site owners.
- The new Share_X class extends the existing Twitter class so we can rely on that for now.
- At some point in the future, we may be able to merge the 2, once the Twitter brand is more deprecated than it is today.

### Other information:

## Jetpack product discussion

No discussion on this, but some previous related PRs:

- https://github.com/Automattic/social-logos/pull/126
- #33118

* No

## Testing instructions:

1. Go to Jetpack > Settings > Social and enable sharing buttons.
2. Go to Settings > Sharing and add a Twitter and an X sharing button.
3. They should both look good in wp-admin and on the frontend.
4. Clicking the button should work too.

> **Note**
> Matching Calypso PR: https://github.com/Automattic/wp-calypso/pull/81835


zinigor (dismissed):
I don't know if I'm testing this wrong, but the actual sharing doesn't work for me. When I click the X sharing button, a new window opens that just takes me to the same page. Can it be due to the fact that the URL parameter still says `twitter` instead of `X`?

jeherve:
Ah, that's my bad. I had only tested on a site where the old Twitter button was also active. This obviously won't always be the case, so we cannot just extend the Twitter class. I've just pushed 4a8fc54c8ea3682bd468869aa56ee115509e16ca
 to take care of this issue. It does duplicate a few things, but I think this is fine for now; at some point in the future we will eventually get rid of the Twitter class.

sdixon194 (approved):
Looks good to me and works! No official button is available yet right? Approving this PR so we can merge and test in the beta, and if there's an option for an official button later we can add it to a followup.

jeherve:
> No official button is available yet right? 

Correct. Maybe one will be added at some point in the future.

jeherve:
@sdixon194 Following-up to your remark about official buttons, I just opened #33474 to improve the button when using the Official style.

Thank you!

PR #33136, opened 2023-09-18
CRM: Update the workflow schema to use pointer steps

cleacos:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3312

## Proposed changes:

This PR updates the Workflow schema and the steps inside of it. Now, steps are linked through pointers, with the step id.

Example of Workflow:

```
array(
    'id'           => 'testing',
    'name'         => 'New Contact',
    'active'       => true,
    'version'      => 1,
    'category'     => 'Contact',
    'triggers'     => array(
        'jpcrm/invoice_created',
    ),
    'initial_step' => 'step_1',
    'steps'        => array(
        'step_1' => array(
            'slug'       => 'jpcrm/update_contact_status',
            'attributes' => array(
                'new_status' => 'Customer',
            ),
            'next_step'  => 'step_2',
        ),
        'step_2' => array(
            'slug'       => 'jpcrm/send_email',
            'attributes' => array(
                'to'      => 'admin@example.com',
                'subject' => 'New Customer',
                'body'    => 'A new customer has been created.',
            ),
            'next_step'  => null,
        ),
    ),
);
```
Now, instead of having the `initial_step` with all the step info linked (folder scheme), we have the new `steps` array with an index (step ID), and the `initial_step` has the step id (index). Also, `next_step` `next_step_true` `next_step_false` is the ID of the next step.

**Note**: We should work in our model with the Step instance rather than retrieving IDs. This will be handled in a following PR.

### Other information:

## Jetpack product discussion
pbhBOz-3Wf-p2

N/A

## Testing instructions:
- Check the code.
- Check that the tests continue passing.

### DIFF
--- a/projects/plugins/crm/src/automation/class-automation-engine.php
+++ b/projects/plugins/crm/src/automation/class-automation-engine.php
@@ -414,10 +414,6 @@ class Automation_Engine {
 
 				$step_slug = $step->get_slug();
 
-				if ( isset( $step_data['attributes'] ) && is_array( $step_data['attributes'] ) ) {
-					$step->set_attributes( $step_data['attributes'] );
-				}
-
 				$this->get_logger()->log( '[' . $step->get_slug() . '] Executing step. Type: ' . $step->get_data_type() );
 
 				$data_type = $this->maybe_transform_data_type( $trigger_data_type, $step::get_data_type() );
--- a/projects/plugins/crm/src/automation/class-automation-engine.php
+++ b/projects/plugins/crm/src/automation/class-automation-engine.php
@@ -293,8 +293,8 @@ class Automation_Engine {
 			);
 		}
 
-		$step_name                     = $class_name::get_slug();
-		$this->steps_map[ $step_name ] = $class_name;
+		$step_slug                     = $class_name::get_slug();
+		$this->steps_map[ $step_slug ] = $class_name;
 	}
 
 	/**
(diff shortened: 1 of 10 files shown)


projects/plugins/crm/src/automation/class-automation-engine.php
```diff
@@ -293,8 +293,8 @@ public function register_step( string $class_name ) {
 			);
 		}
 
-		$step_name                     = $class_name::get_slug();
-		$this->steps_map[ $step_name ] = $class_name;
+		$step_slug                     = $class_name::get_slug();
+		$this->steps_map[ $step_slug ] = $class_name;
```
gogdzl, line 297: Nice change!

gogdzl (approved):
The only part I find confusing is the fact that we allow strings and integers as valid step_ids.

I just would like some clarification on that matter. Otherwise, it works!

cleacos:
> The only part I find confusing is the fact that we allow strings and integers as valid step_ids.

I left it open because mainly, and for now, workflow steps are not going to be stored alone in DB, so that is just an index we have to set and it could be `1`, `2`,... or use another format like `s1`, `s2`,... or `step_1`, `step_2`. Better to read it in an array. So because steps are not as storable as a DB entity, we don't have the unique ID concept here is just an index inside of the workflow.

cleacos:
Merging it to trunk...

PR #33137, opened 2023-09-18
CRM: Automations - add the invoice entity class

coder-karen:
## Proposed changes:

* This PR adds the invoice entity class. See https://github.com/Automattic/zero-bs-crm/issues/3318
* The object properties include field map properties plus similar properties to https://github.com/Automattic/jetpack/pull/32716/files#diff-baa1aa6360d39c2a01d2f4b62762a1667e481345b3bb0771f6d86d4c48d0fed2 where relevant, minus associative
* The field map properties are generally retrieved [from here](https://github.com/Automattic/jetpack/blob/52aeb9a918509c6abd84ed0a6cab91880a779e1a/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Invoices.php#L37), as well as [this DB list](https://github.com/Automattic/jetpack/blob/e156ed85552fed62107978a7f41197640c7aa9b9/projects/plugins/crm/includes/ZeroBSCRM.Database.php#L678) (which includes property defaults generally).
* [`tidy_invoice`](https://github.com/Automattic/jetpack/blob/52aeb9a918509c6abd84ed0a6cab91880a779e1a/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Invoices.php#L2105), as with [`tidy_contact`](https://github.com/Automattic/jetpack/blob/e156ed85552fed62107978a7f41197640c7aa9b9/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Contacts.php#L3568) which was mentioned in this PR description - https://github.com/Automattic/jetpack/pull/32716 - includes additional fields so those aren't included.
* I've not included site or team properties as they weren't included for contacts - they are the only properties missing from the object model lists.

### Other information:

## Jetpack product discussion

https://github.com/Automattic/zero-bs-crm/issues/3318
p1694789026607679/1694779506.282519-slack-CTXBP902X
p1695023975304399-slack-CTXBP902X

No.

## Testing instructions:

* Un-defined as of yet. Currently it can be compared to the https://github.com/Automattic/jetpack/pull/32716/files#diff-baa1aa6360d39c2a01d2f4b62762a1667e481345b3bb0771f6d86d4c48d0fed2, with the constraints mentioned in the linked issue.


projects/plugins/crm/src/entities/class-invoice.php
```diff
@@ -0,0 +1,363 @@
...
+	 * Custom fields.
+	 *
+	 * @var array Custom fields.
+	 */
+	private $custom_fields = array();
+
+	/**
+	 * {@inheritDoc}
+	 */
+	protected function get_field_map(): array {
```
cleacos: This will be extracted to a mapper class.

projects/plugins/crm/src/entities/class-invoice.php
```diff
@@ -0,0 +1,363 @@
...
+	 * @var int
+	 */
+	public $created = -1;
+
+	/**
+	 * DB field name mapping. db_field => model_field.
+	 *
+	 * @var array
+	 */
+	private $field_map = array(
```
cleacos: This will be extracted to a Factory (Invoice_Factory) or Data Mapper (Invoice_Data_Mapper) class.

projects/plugins/crm/src/entities/class-invoice.php
coder-karen: Ok thanks, removed now.

projects/plugins/crm/src/entities/class-invoice.php
coder-karen: Got it, thanks.

projects/plugins/crm/src/entities/class-invoice.php
```diff
@@ -0,0 +1,310 @@
...
+ * @since $$next-version$$
+ */
+class Invoice {
+
+	/**
+	 * The DB ID of the object.
+	 *
+	 * @var int
+	 */
+	public $id = -1;
```
tbradsha, line 22: Probably not worth changing here since we're just matching DAL, but I really want to move away from -1 meaning "not set".

cleacos: Yes, in this case, we are mirroring the model from DAL. So we have to continue using -1.

projects/plugins/crm/src/entities/class-invoice.php
```diff
@@ -0,0 +1,310 @@
...
+	 * @var array
+	 */
+	public $files = array();
+
+	/**
+	 * The invoice status.
+	 *
+	 * @var string
+	 */
+	public $status = 'Draft';
```
cleacos: I know it was in the reference draft-PR with the Contact entity. But we won't use default values, leaving it empty. (I'll do the same on Contact entity into another PR)

```suggestion
	public $status = '';
```

cleacos (dismissed):
Let's move forward. The entity looks good. Only a small change, set the default status as empty (commit suggested) for now. Later, if we have to set the default status, we have to translate it using `__( 'Draft', 'zero-bs-crm' )` in the constructor. 

LGTM!

cleacos (approved):

tbradsha:
> Later, if we have to set the default status, we have to translate it

No, please don't! :^) The `status` is now a key, and the `status_label` (or its future equivalent here) is the translated version; see #32909.

cleacos:
>  The status is now a key, and the status_label (or its future equivalent here) is the translated version;

😮

PR #33149, opened 2023-09-18
Publicize: display the editor modal regardless of module status

### LINKED ISSUE #12367: Add ability to turn on publicize from within the Gutenberg Jetpack sidebar
We want to add the ability to enable Publicize right from within the Gutenberg Jetpack sidebar. This will raise visibility of Jetpack within Gutenberg and allow folks to enable a widely-used feature easily.

Design and details TBD.

#### Primary issue

#32401

jeherve:
Fixes #12367

## Proposed changes:

- Always register the Publicize editor extension.
- Track views of that placeholder and clicks to enable the module.

Screenshot 2023-09-18 at 16 41 19
Screenshot 2023-09-18 at 16 40 28

### Other information:

## Jetpack product discussion

* #32401 

* Yes, it starts tracking 2 things: views of the placeholder, and clicks to activate the module.

## Testing instructions:

* Start your test on a site where the Publicize module is not yet enabled.
* Open your browser's JavaScript console, and enable viewing tracks items: `localStorage.setItem( 'debug', 'dops:analytics*' );`
* Go to Posts > Add New
* Click on the "share this post" panel title.
    * You should see the Tracks event in your JavaScript console.
* Click on the button.
    * You should see the module getting enabled, and the event in your JS console.

You can also test on WordPress.com Simple; the module is always enabled there so the panel should never display the placeholder.


monsieur-z (commented):
I could be missing a configuration or something, but the _Share this post_ panel is not showing for me in the sidebar when I create a new post. I've tried with both a JN and my local site.

andrii-lysenko:
I have the same situation as @monsieur-z described above. Interesting that if I manually enable Sharing and then disable. Trying to add the post will show the correct behavior while the same session. However, after reloading the page social sharing is still not visible

jeherve:
Thank you both! I just pushed 10c393820e8ec699cf8bb45a94b254e85da44640 and ac4501d642d6b17266e9b68bec6033a1b810d6e9 to fix that (I didn't run into the issue on my local install because I have the Jetpack Social plugin activated there.

fgiannar (changes requested):
Thanks for working on this, Jeremy!

I tested on a JN site (created from the corresponding link in the PR):
- I didn't see any tracks events in my console after running: `localStorage.setItem( 'debug', 'dops:analytics*' );`
- After clicking the “Activate Jetpack Social” button the “Share this post” section **disappears**. It appears after reloading the page as follows:
screenshot_2023-09-19_at_12 25 48

jeherve:
> I didn't see any tracks events in my console after running: localStorage.setItem( 'debug', 'dops:analytics*' );

That does seem to work on my end. Do you see the Tracks events on other pages of the site, like in the Jetpack dashboard?

> After clicking the “Activate Jetpack Social” button the “Share this post” section disappears.

Good catch! The last 2 commits should fix that 👍

fgiannar (approved):
Tested once more (after the latest commits) and can confirm that the "Share this post" section doesn't disappear anymore but is loaded as follows:

Screenshot 2023-09-19 at 13 09 10

aka without the prompt to connect (which appears after reloading the page).
I believe the above can be addressed in a follow up PR.
-> See discussion in p1695118435230779-slack-C02JJ910CNL

Finally, I was also able to see the tracks events. FYI I had to enable verbose level of logging in Chrome's dev console to see those
Screenshot 2023-09-19 at 13 08 32

jeherve:
> I believe the above can be addressed in a follow up PR.

I logged an issue for this here:
https://github.com/Automattic/jetpack/issues/33172

PR #33152, opened 2023-09-18
Adds woo HPOS support to the google analytics module.

### LINKED ISSUE #31974: Google Analytics compatibility with WooCommerce HPOS
### Impacted plugin

Jetpack

### Quick summary

Both Google Analytics Legacy and Universal use `get/update_post_meta()` with WooCommerce orders. Both classes have similar code. The `postmeta` table is not used when Woo HPOS is authoritative and the posts sync is disabled. `$order->get_meta()` and `$order->update_meta_data()` should be used for compatibility with both datastores. 

https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/jetpack/modules/google-analytics/classes/wp-google-analytics-universal.php#L122:L189

Optional fixes:

```
$order          = new WC_Order( $order_id );
```

should be the loader function which will return false when the order doesn't exist

```
$order          = wc_get_order( $order_id );
```

`$order->get_total_shipping()` is deprecated, use `$order->get_shipping_total()` instead.

### Steps to reproduce

1. I found this using the following bash script:

```
for P in wpdb get_post metadata_exists post_exists _post_meta _metadata 'wp_(insert|update|delete|trash|untrash|transition)_post' '_post[a-zA-Z]*_cache' shop_order; do grep -nRE "$P" ./ | grep 'order' | grep -vE '\.js' ; done | sort -u
```

`*post_meta` functions would not be used

### What actually happened

They are

### Impact

Some (< 50%)

### Available workarounds?

Yes, easy to implement

Simple, Atomic, Self-hosted

### Logs or notes

This is primarily an issue when sync is disabled.

zinigor:
Fixes #31974 

## Proposed changes:
* Adds a parallel method for Google Analytics module to track purchases with WooCommerce HPOS-enabled sites.

### Other information:

No.

## Testing instructions:

* TBA
*


projects/plugins/jetpack/modules/google-analytics/classes/wp-google-analytics-universal.php
```diff
@@ -188,6 +195,76 @@ public function maybe_track_purchases( $command_array ) {
...
+	 * @see https://github.com/woocommerce/woocommerce/wiki/High-Performance-Order-Storage-Upgrade-Recipe-Book
+	 *
+	 * @since $$next-version$$
+	 * @param array $command_array Array of commands.
+	 * @return array `$command_array` with additional commands conditionally added.
+	 */
+	public function maybe_track_hpos_purchases( $command_array ) {
+		global $wp;
+
+		$order_id = isset( $wp->query_vars['order-received'] ) ? $wp->query_vars['order-received'] : 0;
```
rrennick, line 214: I'm not sure if you use them in JP but this could be a null coalesce.

rrennick (approved):
@zinigor This looks great from a HPOS perspective. I just had the one coding style suggestion.

projects/plugins/jetpack/modules/google-analytics/classes/wp-google-analytics-universal.php
zinigor: Yeah, unfortunately we're still at PHP 5.6 minimum, but someday :)

PR #33155, opened 2023-09-18
CRM: Resolves #3324 - implement tag entity

### LINKED ISSUE #3324: Sitemaps: Include links to the sitemaps
When activating Sitemaps or when clicking the module title via Jetpack->Settings, the module info should include the links to the sitemaps available (instead of passing users to the Support site and seeing the URLs there).

There doesn't seem to be a natural way to figure this out.

tbradsha:
CRM: Resolves Automattic/zero-bs-crm#3324 - implement tag entity

## Proposed changes:

This implements a basic Tag entity object.

Note that Tags are not considered objects in DAL, but instead are implemented at the DAL level. I consulted `ZeroBSCRM.DAL3.php` and the database table for reasonable values. I added the following properties:

* `id`: the ID in the table
* `obj_type_id`: the object type ID it belongs to
* `label`: the user-friendly tag label (referred to as "name" elsewhere, but I called it label here...can change back if we prefer)
* `slug`: the tag slug
* `created`: Unix timestamp of tag creation
* `lastupdated`: Unix timestamp of tag's last update

UTS fields are initially set to `null`, as `-1` is a valid UTS timestamp.

### Other information:

## Testing instructions:

N/A; implemented as best I could based on the inferred schema.

### DIFF
--- a/projects/plugins/crm/changelog/add-crm-3324-implement_tag_entity
+++ b/projects/plugins/crm/changelog/add-crm-3324-implement_tag_entity
@@ -1,0 +1,5 @@
+Significance: patch
+Type: added
+Comment: Created Tag class
+
+
--- a/projects/plugins/crm/src/entities/class-tag.php
+++ b/projects/plugins/crm/src/entities/class-tag.php
@@ -16,14 +16,14 @@ class Tag {
 	 *
 	 * @var int
 	 */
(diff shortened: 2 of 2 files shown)


cleacos (approved):
LGTM!

PR #3323, opened 2016-01-21
Add breadcrumb support for any hierarchical post type or taxonomy

eliorivero:
Previously breadcrumbs were only available for pages. This PR adds makes them available to any hierarchical post type.
It also introduces support for category views or any hierarchical taxonomy.

PR #33158, opened 2023-09-18
CRM: Resolves #3323 - implement task entity

tbradsha:
CRM: Resolves Automattic/zero-bs-crm#3323 - implement task entity

## Proposed changes:

This implements a basic Task entity object.

Note that the original issue mentioned Events, but I went with Tasks, as that's what we use throughout the UI, and also "events" could get confusing when dealing with the event manager system:

* `id`: the ID in the table
* `owner`: the WP user ID for the user that owns the task
* `title`: the task title
* `description`: the task description
* `start`: Unix timestamp of task start
* `end`: Unix timestamp of task end
* `is_completed`: Whether task has been completed or not
* `show_in_portal`: Whether task should show in portal
* `show_in_calendar`: Whether task should show in calendar
* `created`: Unix timestamp of task creation
* `lastupdated`: Unix timestamp of task last update

UTS fields are initially set to `null`, as `-1` is a valid UTS timestamp.

### Other information:

## Testing instructions:

N/A at the moment.

### DIFF
--- a/projects/plugins/crm/src/entities/class-task.php
+++ b/projects/plugins/crm/src/entities/class-task.php
@@ -9,6 +9,10 @@ namespace Automattic\Jetpack\CRM\Entities;
 
 /**
  * Task class.
+ *
+ * Note that DAL and the database currently use the legacy term `event` instead
+ * of `task`, but to match the UI and to prevent confusion with the Automations
+ * Event Manager, we'll be using `task` here.
  */
 class Task {
 	/**
--- a/projects/plugins/crm/src/entities/class-task.php
+++ b/projects/plugins/crm/src/entities/class-task.php
@@ -1,0 +1,90 @@
+<?php
+/**
+ * Task Entity.
+ *
+ * @package automattic/jetpack-crm
+ */
+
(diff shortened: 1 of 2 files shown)


cleacos:
In the backend, let's keep the same name we already have for it, `event`. The current model with the DAL, and the current Automation code use `event`. In the UI we could use any name that fits better for the end user. I prefer the name Task everywhere, but at least let's keep the same name in the backend.

projects/plugins/crm/src/entities/class-task.php
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Task Entity.
+ *
+ * @package automattic/jetpack-crm
+ */
+
+namespace Automattic\Jetpack\CRM\Entities;
+
+/**
+ * Task class.
```
cleacos, line 11: We should comment here about the relationship with the `events` on the DAL (events endpoint) and DB (events table) sides.

tbradsha: 3b36c25d3e

cleacos (approved):
LGTM!

PR #33160, opened 2023-09-19
Add Stats section that will be used in My Jetpack

IanRamosC:
## Proposed changes:

* Abstract the wrapper for product cards into its own component
* Moved "Status" to its own component - which is used to display the status (Active, Inactive, etc) of a specific product in a Card
* Add a new "Stats Section" component that looks like the highlight cards that are displayed on the Stats page. This component looks like the screenshot below

Card:
image

Stats Section:
image

### Other information:

## Testing instructions:

* Run this branch locally
* Run `jetpack build js-packages/storybook`
* Then `cd projects/js-packages/storybook && pnpm run storybook:dev`
* Visit the Storybook page and confirm that the Stats section matches the one proposed in the design


projects/packages/my-jetpack/_inc/components/card/style.module.scss
```diff
@@ -0,0 +1,53 @@
+.container {
+	--actions-size: 28px;
+	--status-size: 8px;
```
robertsreberski: Nitpick idea: I think we can define local variables as scss variables rather than css, so that they will be built during transpilation.

projects/packages/my-jetpack/_inc/components/stats-section/count-comparison-card.jsx
```diff
@@ -0,0 +1,95 @@
+import { numberFormat } from '@automattic/jetpack-components';
+import { Card } from '@wordpress/components';
+import { arrowDown, arrowUp, Icon } from '@wordpress/icons';
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+import React from 'react';
+import styles from './style.module.scss';
+
+const subtract = ( a, b ) => {
+	return a === null || b === null || b === undefined ? null : a - b;
```
robertsreberski: What is `a = undefined`? I could think of the following suggestion to make it clearer. it will also not allow for any `null` or `undefined`, but also `string` etc
```suggestion
	if ( typeof a !== 'number' || typeof b !== 'number' ) {
	    return null;
	}
	
	return a - b;
```

projects/packages/my-jetpack/_inc/components/stats-section/count-comparison-card.jsx
```diff
@@ -0,0 +1,95 @@
...
+};
+
+export const percentCalculator = ( part, whole ) => {
+	if ( part === null || whole === null || whole === undefined ) {
```
robertsreberski: same here

projects/packages/my-jetpack/_inc/components/stats-section/count-comparison-card.jsx
```diff
@@ -0,0 +1,95 @@
...
+					className={ styles[ 'stats-card-count-value' ] }
+					title={ Number.isFinite( count ) ? String( count ) : undefined }
+				>
+					{ numberFormat( count, shortenedNumberConfig ) }
+				</span>{ ' ' }
+				{ difference !== null ? (
+					<span
+						className={ classNames( styles[ 'stats-card-difference' ], {
+							[ styles[ 'stats-card-difference--positive' ] ]: difference < 0,
+							[ styles[ 'stats-card-difference--negative' ] ]: difference > 0,
```
robertsreberski, line 71: What if difference is 0?

projects/packages/my-jetpack/_inc/components/stats-section/count-comparison-card.jsx
```diff
@@ -0,0 +1,95 @@
...
+			{ icon && <div className={ styles[ 'stats-card-icon' ] }>{ icon }</div> }
+			{ heading && <div className={ styles[ 'stats-card-heading' ] }>{ heading }</div> }
+			<div className={ styles[ 'stats-card-count' ] }>
+				<span
+					className={ styles[ 'stats-card-count-value' ] }
+					title={ Number.isFinite( count ) ? String( count ) : undefined }
+				>
+					{ numberFormat( count, shortenedNumberConfig ) }
+				</span>{ ' ' }
+				{ difference !== null ? (
```
robertsreberski, line 67: Shouldn't we just do `!! difference` to also handle case that it's `0` or `undefined`?

robertsreberski (commented):
Looks solid already 💪 I just left some comments

projects/packages/my-jetpack/_inc/components/stats-section/count-comparison-card.jsx
IanRamosC: Yes, I just copied that over and didn't change.

projects/packages/my-jetpack/_inc/components/stats-section/count-comparison-card.jsx
IanRamosC: Then it doesn't have to add the class. It remains with the same style

projects/packages/my-jetpack/_inc/components/stats-section/count-comparison-card.jsx
IanRamosC: We want to display that section even if `difference` is 0, and the only 2 possibles returns from that function are a number or null.

projects/packages/my-jetpack/_inc/components/card/index.jsx
```diff
@@ -0,0 +1,32 @@
+import { Text } from '@automattic/jetpack-components';
```
CodeyGuyDylan, line 1: I think we should try to make all new files such as this and others TypeScript files unless there was some sort of configuration issue that made that more difficult 🤔  What do you think?

projects/packages/my-jetpack/_inc/components/product-card/status.jsx
```diff
@@ -0,0 +1,29 @@
...
+	const isPurchaseRequired =
+		status === PRODUCT_STATUSES.NEEDS_PURCHASE ||
+		status === PRODUCT_STATUSES.NEEDS_PURCHASE_OR_FREE;
+	const flagLabel = PRODUCT_STATUSES_LABELS[ status ];
+	const statusClassName = classNames( styles.status, {
+		[ styles.active ]: isActive,
+		[ styles.inactive ]: isInactive || isPurchaseRequired,
+		[ styles.error ]: isError,
+		[ styles[ 'is-fetching' ] ]: isFetching || isInstallingStandalone || isDeactivatingStandalone,
+	} );
```
CodeyGuyDylan: Feel free to ignore this comment, I can't help but feel this could be simplified a bit with maybe less variables? However I did sit here and think about what could be done better for a while and didn't come up with anything solid so I could be overthinking it

projects/packages/my-jetpack/_inc/components/stats-section/count-comparison-card.jsx
```diff
@@ -0,0 +1,99 @@
...
+		return null;
+	}
+
+	return a - b;
+};
+
+export const percentCalculator = ( part, whole ) => {
+	if ( typeof part !== 'number' || typeof whole !== 'number' ) {
+		return null;
+	}
```
CodeyGuyDylan, line 28: I see this being used twice in this file, could we move this to a `validateNumber` function or something like that? Or is there a function like it that already exists somewhere else in the plugin?

projects/packages/my-jetpack/_inc/components/stats-section/count-comparison-card.jsx
```diff
@@ -0,0 +1,99 @@
...
+	const shortenedNumberConfig = { maximumFractionDigits: 1, notation: 'compact' };
+
+	return (
+		<Card className={ styles[ 'stats-card' ] }>
+			{ icon && <div className={ styles[ 'stats-card-icon' ] }>{ icon }</div> }
+			{ heading && <div className={ styles[ 'stats-card-heading' ] }>{ heading }</div> }
+			<div className={ styles[ 'stats-card-count' ] }>
+				<span
+					className={ styles[ 'stats-card-count-value' ] }
+					title={ Number.isFinite( count ) ? String( count ) : undefined }
```
CodeyGuyDylan, line 63: Here I noticed if somehow the "count" is a negative number the card will still display it. I don't know if that's even possible with wherever we are getting that number from, but we could display a 0 instead of a negative number if somehow that does happen

projects/packages/my-jetpack/_inc/components/stats-section/count-comparison-card.jsx
```diff
@@ -0,0 +1,99 @@
...
+						</span>
+						<span className={ styles[ 'stats-card-difference-absolute-value' ] }>
+							{ numberFormat(
+								differenceMagnitude,
+								differenceMagnitude > 9999 ? shortenedNumberConfig : {}
+							) }
+						</span>
+						{ percentage !== null && (
+							<span className={ styles[ 'stats-card-difference-absolute-percentage' ] }>
+								{ ' ' }
```
CodeyGuyDylan: I see there is some margin between the number and the percentage in quotes, could we just increase that a bit instead of adding a space like this?

CodeyGuyDylan (dismissed):
I personally think the design that is currently implemented looks great, I did, however, notice a few discrepancies from the figma comps when switching between the two that mostly has to do with font sizes

Example: 
This is 36px (this PR)
image

This is 48px (comps)
image

The position and size of the label also seem a bit off from the comps. Like I said I think it still looks great I just wanted to make sure this was intentional

One more note about the design, the percentages don't look great on smaller screen sizes, and it looks like in the comps we are supposed to hide them for mobile sizes
image

projects/packages/my-jetpack/_inc/components/card/index.jsx
IanRamosC: This PR would be the one introducing TypeScript to My Jetpack and, while I did think about it when I was initially working on this task, I don't think it's in the scope of this task. It wouldn't be difficult to set it up, but I'd rather have that done in a different PR.

projects/packages/my-jetpack/_inc/components/stats-section/count-comparison-card.jsx
IanRamosC: Ooops, not intentional 👀

projects/packages/my-jetpack/_inc/components/stats-section/count-comparison-card.jsx
IanRamosC: This is a very personal opinion, but I don't usually like to abstract such things, considering that we would pretty much abstract it to a function that its only responsibility is to do exactly the same thing as the raw operation is doing: checking if a variable is a number.

Or you mean abstracting checking n parameters at once? Something like:
```js
const validateNumber = ( ...numbers ) => {
    return numbers.every( number => typeof number === 'number' );
}
```

Either way, that would add a tiny level of complexity that I personally don't think is necessary.

projects/packages/my-jetpack/_inc/components/stats-section/count-comparison-card.jsx
```diff
@@ -0,0 +1,98 @@
...
+		<Card className={ styles[ 'stats-card' ] }>
+			{ icon && <div className={ styles[ 'stats-card-icon' ] }>{ icon }</div> }
+			{ heading && <div className={ styles[ 'stats-card-heading' ] }>{ heading }</div> }
+			<div className={ styles[ 'stats-card-count' ] }>
+				<span
+					className={ styles[ 'stats-card-count-value' ] }
+					title={ Number.isFinite( count ) ? String( count ) : undefined }
+				>
+					{ numberFormat( count, shortenedNumberConfig ) }
+				</span>{ ' ' }
```
IanRamosC: There's an extra space here as well that will be removed in the PR that follows this one (for populating the fields with real data).

projects/packages/my-jetpack/_inc/components/card/style.module.scss
```diff
@@ -0,0 +1,50 @@
+.container {
+	padding: calc( var( --spacing-base ) * 3 ); // 24px
+	background: var( --jp-white );
+	border-radius: var( --jp-border-radius );
+	box-shadow: 0 0 40px rgba(0, 0, 0, 0.08);
+	height: 100%;
+	display: flex;
+	flex-direction: column;
+	box-shadow:
+		0 0 0 1px var( --jp-gray-10 ) inset,
```
jboland88, line 10: Nitpick: It may be something from storybook, but it looks like it's overriding this style with a 1.5px box shadow. Coming from an `is-secondary` class.

However, this is also how the buttons are currently styled in My Jetpack - the comps are different than the existing buttons we have on other cards.

Screenshot 2023-09-26 at 10 05 42 AM

projects/packages/my-jetpack/_inc/components/stats-section/count-comparison-card.jsx
```diff
@@ -0,0 +1,106 @@
...
+ * @param {number} props.previousCount     - Previous count.
+ * @param {React.ReactNode} props.icon     - Icon to display.
+ * @param {React.ReactNode} props.heading  - Card heading.
+ * @returns {object} CountComparisonCard React component.
+ */
+const CountComparisonCard = ( { count, previousCount, icon, heading } ) => {
+	const difference = subtract( count, previousCount );
+	const differenceMagnitude = Math.abs( difference );
+	const percentage = Number.isFinite( difference )
+		? percentCalculator( Math.abs( difference ), previousCount )
```
jboland88: nit: could differenceMagnitude be re-used here as well?

jboland88 (dismissed):
This LGTM! Nice work, thanks for doing some improvements by creating card and status components!

I had two tiny nits, no blockers.

projects/packages/my-jetpack/_inc/components/card/style.module.scss
IanRamosC: I can update the style of the button, but since this is going to be displayed on the same screen as the existing buttons, I'd rather leave it like this (at least for now).

jboland88 (commented):
Giving another ✅ after the version bump

jboland88 (approved):
✅ actually selecting "Approve" this time 🤣

PR #33175, opened 2023-09-19
Jetpack Boost: Refactor the connection globals

pyronaur:
Part of the "Refactor Global State" project (pc9hqz-286-p2).

### Other information:

## Jetpack product discussion
N/A

N/A

## Testing instructions:
* Try various connection flows to thoroughly test out the connection process and compare it to the current process. There should be no noticeable changes.


dilirity (dismissed):
There's a changelog missing.

projects/plugins/boost/app/admin/class-config.php
```diff
@@ -183,11 +182,13 @@ public static function set_getting_started( $value ) {
 	 */
 	public static function is_getting_started() {
 		// Aside from the boolean flag in the database, we also assume site already got started if they have premium features.
-		return \get_option( 'jb_get_started', false ) && ! Premium_Features::has_feature( Premium_Features::CLOUD_CSS ) && ! ( new Status() )->is_offline_mode();
```
dilirity, line 186: With the removal of `jb_get_started`, I'm thinking that we don't need logic tied to it at all. Should we add a ticket to clean this up in a separate PR?

projects/plugins/boost/app/assets/src/js/pages/getting-started/GettingStarted.svelte
```diff
@@ -1,100 +1,45 @@
...
-
+	async function initialize( plan: 'free' | 'premium' ) {
 		try {
 			// Make sure there is a Jetpack connection
-			await connection.initialize();
+			await initializeConnection();
 
 			// Record this selection. This must be done after the connection is initialized.
-			recordBoostEvent( 'premium_cta_from_getting_started_page_in_plugin', {} );
+			recordBoostEvent( `${ plan }_cta_from_getting_started_page_in_plugin`, {} );
```
dilirity, line 33: The other day I had to search for events in Boost. Having the full name of an event would be helpful in finding it.

Do you think expanding the comment to include the full names of both versions makes sense here?

dilirity (dismissed):
Great work on simplifying this!

I left a suggestion, but other than that - it's ready to merge.

:shipit:

projects/plugins/boost/app/assets/src/js/pages/getting-started/GettingStarted.svelte
pyronaur: Fixed

dilirity (approved):

PR #33179, opened 2023-09-19
Jetpack: Fix undefined variable warnings in featured images fallback

anomiex:
## Proposed changes:

PR #32824 introduced some new code, and unfortunately some code paths do not define variables that are needed later on. The later code handles the resulting `null`s properly, but we should still avoid the warnings.

### Other information:

## Jetpack product discussion

p1695140087285159/1695132622.239659-slack-C01U2KGS2PQ

No

## Testing instructions:

* Run the tests from #32824, checking the log for warnings?

### DIFF
--- a/projects/plugins/jetpack/modules/theme-tools/content-options/featured-images-fallback.php
+++ b/projects/plugins/jetpack/modules/theme-tools/content-options/featured-images-fallback.php
@@ -66,7 +66,7 @@ function jetpack_featured_images_fallback_get_image( $html, $post_id, $post_thum
 			$width  = isset( $image['width'] ) ? intval( $image['width'] ) : null;
 			$height = isset( $image['height'] ) ? intval( $image['height'] ) : null;
 
-			if ( $image['src_width'] && $image['src_height'] && $image['width'] && $image['height'] ) {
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/modules/theme-tools/content-options/featured-images-fallback.php
```diff
@@ -63,9 +63,10 @@ function jetpack_featured_images_fallback_get_image( $html, $post_id, $post_thum
 
 			$image_sizes = '';
 
+			$width  = isset( $image['width'] ) ? intval( $image['width'] ) : null;
+			$height = isset( $image['height'] ) ? intval( $image['height'] ) : null;
+
 			if ( $image['src_width'] && $image['src_height'] && $image['width'] && $image['height'] ) {
```
zinigor: We also have an
```
Undefined array key "src_width"
```
on this line, should we add an `isset( $image['src_width'] )` here?

anomiex: Probably more like
```suggestion
			if ( ! empty( $image['src_width'] ) && ! empty( $image['src_height'] ) && ! empty( $image['width'] ) && ! empty( $image['height'] ) ) {
```

zinigor (approved):
Thanks for the quick fix!

sgomes:
Thank you for the fix, @anomiex, and I apologise for the oversight 😞

PR #33180, opened 2023-09-19
AI Chat: Show guideline message 

dereksmart:
## Proposed changes:

This change exports the GuidelineMessage component from the jetpack-ai-client package and uses it in the AI Chat block. 

The guideline message shows only in the editor and when the block is focused/selected: 

image

## Jetpack product discussion

N/a

No
## Testing instructions:

- Make sure it only shows in the editor context
- It should only show when the block is selected
- The links should work


artpi:
Shouldn't we be showing this message on the frontend, where people actually ask questions?

dereksmart:
Not sure tbh. It links to a8c privacy policy. A front end default
disclaimer should be customizable or toggle-able. Lets merge this and
discuss during beta

artpi (approved):
Honestly I would rather have that disclaimer in the sidebar, but I guess that way is consistent with the AI Assistant.

PR #33181, opened 2023-09-19
Subscriptions Block: Do not optout from any newsletter categories when none selected

mashikag:
Fixes an issue where a user who did not select any newsletter categories, when subscribing to a site using the block, was in turn subscribing to the site with all the categories opted out from. Instead, the expectation is that the user should not be opted out from any newsletter categories in such case, in another words they should be subscribed to all categories.

## Proposed changes:

* When subscribing, verify if no newsletter category pills have been selected. If none are chosen, then omit including any newsletter category exclusions in the subscription request.

## Jetpack product discussion

p1695132146012329-slack-C02TCEHP3HA

No, it does not.

## Testing instructions:

* The PR branch can be tested on either a jetpack-connected site, simple or atomic
* The test site should have the "newsletter categories" blog RC sticker enabled
* Go to `/settings/newsletter/$site-slug`
* Enable newsletter categories on the site, and create some if there is none
  
* On the test site publish a page, or a post, with the Subscribe block
* Open an incognito window in your browser, have the network tab opened on the side, go to the page/post that you created in the previous step. 
* Enter some email address, leave all the newsletter category pills unselected and hit the **Subscribe**" button
* Verify that the network call to the subscribe endpoint did not have any `excluded_newsletter_categories` appended to its URL.

* Similarly verify that the exclusions are still being passed when some newsletter categories are selected

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-subscriptions-block-no-category-excl-when-none-selected
+++ b/projects/plugins/jetpack/changelog/fix-subscriptions-block-no-category-excl-when-none-selected
@@ -1,0 +1,4 @@
+Significance: patch
+Type: other
+
+Subscriptions Block: Do not append newsletter category exclusions to subscribe url, when none categories were checked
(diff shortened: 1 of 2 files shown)


phcp (approved):
Good job! Changes are working as expected.

PR #33182, opened 2023-09-19
Publicize: Reinstate the connect an account link

### LINKED ISSUE #33172: Publicize: add link to manage connections to initial editor plugin description
Right now, sites that have just enabled the publicize feature will see this:

Screenshot 2023-09-19 at 12 13 34

It would be nice if there was a link to the connections page on the original description. The Instagram notice will not always be there I assume. If it's not there, the plugin isn't too useful:

Screenshot 2023-09-19 at 12 14 41

*******

This was brought up in https://github.com/Automattic/jetpack/pull/33149#pullrequestreview-1632918779
Related discussion: p1695118435230779-slack-C02JJ910CNL

pablinos:
This changes ensures there's a link to connect an account at all times in the 'Share this post' panel in the sidebar.

Fixes #33172

## Proposed changes:

With no connections, we now show the 'Connect an account' link again.

image

This also refactors the connection refresh logic, so it's not tied to the PublicizeSettingsButton.

### Other information:

No

## Testing instructions:
* On a fresh site, enable auto-sharing or install Jetpack Social
* Open the editor on a new post
* Check that you are shown the Instagram notice - dismiss it.
* It should be replaced by a 'Connect an account' link
* Click it and connect an account
* Return to the page and check the UI updates after 2 seconds


projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -372,6 +376,14 @@ export default function PublicizeForm( {
 							? renderInstagramNotice()
 							: renderValidationNotice() ) }
 				</>
+			) : (
+				! shouldShowInstagramNotice && (
+					<PanelRow>
+						<ExternalLink href={ connectionsAdminUrl } target="_blank">
```
manzoorwanijk: I think `ExternalLink` already has `target="_blank"`
```suggestion
						<ExternalLink href={ connectionsAdminUrl }>
```

projects/js-packages/publicize-components/src/hooks/use-refresh-connections/index.js
```diff
@@ -0,0 +1,30 @@
...
+			debouncedRefresh.cancel();
+		}
+
+		if ( pageHasFocus && shouldAutoRefresh.current ) {
+			debouncedRefresh();
+			shouldAutoRefresh.current = false;
+		}
+	};
+
+	return { refresh };
```
manzoorwanijk: May be we can simply return the callback wrapped in `useCallback` instead of an object?

projects/js-packages/publicize-components/src/components/form/index.js
pablinos: Ah good point! I just restored the old code here.

projects/js-packages/publicize-components/src/hooks/use-refresh-connections/index.js
pablinos: I've made it so the function is returned directly. I don't think `useCallback` would help here as the function isn't used as a prop and so wouldn't trigger a render.

manzoorwanijk (approved):
Looks great now.

PR #33186, opened 2023-09-19
AI Extension: Show AI Form extension with connection nudge for disconnected users

### LINKED ISSUE #32557: AI Extension: Extension is not visible to disconnected users
The form extension is not visible when the user is not connected to a WordPress.com account. This can happen easily on a site with multiple users.

The extension will not work without a connection, but we can show it and leverage it's visibility to prompt the user to reconnect, the same way we do for the Assistant:

Image

dhasilva:
Fixes #32557

## Proposed changes:

* Removes two checks disabling the extension, both for connection (one for an error that happens when the user is not connected)
* Moves the connection check to the extension by showing a nudge, exactly as the AI Assistant

### Other information:

## Testing instructions:

* Connect your test site to Jetpack as admin
* Create a non-admin user, like an Author
* Log in as the non-admin user
* Go to the block editor without connecting the user to Jetpack
* Add a form
* Check that the extension is visible
* Check that there is a nudge with the message "Your account is not connected to Jetpack at the moment."
* Click on the connection button and connect the user
* Go back to the block editor
* Check that the form is working as normal

| Before | After
|-|-
| |

### DIFF
--- a/projects/plugins/jetpack/changelog/update-ai-form-disconnected
+++ b/projects/plugins/jetpack/changelog/update-ai-form-disconnected
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
+AI Extension: Show AI Form extension with connection nudge for disconnected users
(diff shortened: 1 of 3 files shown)


nunyvega (approved):
Works well!

🚢

PR #33187, opened 2023-09-19
AI Chat: Fix feedback styles and icon SVGs

dereksmart:
## Proposed changes:

This PR fixes the style of the feedback section: 
- Uses theme styles for button colors 
- Matches styles for the components used in the block and elsewhere in WP

This PR also adds a few SVG's for use in the shared icons. This means that icons will now render on the front end of the site when not logged in. The alternate to this was to load all dashicons styles on the front end, which did not seem ideal. 

The SVGs were copied from seth's site, so they will match the styling there.  

image

## Jetpack product discussion

N/A

No

## Testing instructions:

- Ask a question 
- Once answered, you should see the feedback buttons in the same color as your theme. 
- Click the copy button, make sure it looks ok
- Click one of the thumbs. The input field should display nicely. 

To achieve maximum testing, do those steps while logged out incognito or something.


dereksmart:
All of those version bumps are because of 

image

artpi (approved):
Nice!

PR #33194, opened 2023-09-19
CRM: Resolves #3327 - respect notify param in `create_event` API endpoint

tbradsha:
CRM: Resolves Automattic/zero-bs-crm#3327 - respect notify param in `create_event` API call

## Proposed changes:

The `create_event` API endpoint expects a `notify` param per the docs:
https://automattic.github.io/jetpack-crm-api-docs/#create-event

It should have created a 24-hr notification if one passes `24` as the value. However, there were a few things preventing this. The PR does the following:

* In `zeroBS_addUpdateEvent()` we were adding the `reminder` key one level too shallow (`$args['reminders']` → `$args['data']['reminders']`).
* Deletion of the reminder requires the key to exist with an empty value. As such, I initialise the key with an empty array.
* There was groundwork laid in the backend to allow for multiple notifications, but we only allow one notification via the editor, and while the endpoint had some code that was partially built out, the vars it was using were never actually set or used. I just chopped this for now and stuck with one notification.
* Updates `event` to `task` (see Automattic/zero-bs-crm#3330)

Note: while I could've done a lot more cleanup here, I tried to stay to an extremely limited scope, as the current API system isn't a priority. For example, if an owner of value `-1` is passed to DAL, it sets it to the current WP user. However, when using the API there is no current WP user, so it sets the owner to `0`. This affects all objects, but isn't a big deal...usually. I fixed a PHP notice it was causing.

### Other information:

## Testing instructions:

```
Pass the following to the `create_event` API endpoint:

{"title":"some task title","customer":-1,"notes":"this is the description","to":"2018-02-01 03:30","from":"2018-02-01 03:30","notify":24,"complete":0,"owner":-1,"event_id":""}

In `trunk`, the task will be created, but:
* You'll get a PHP notice when visiting the task in the CRM.
* The reminder is not added to the database or ticked in the CRM.
* The response gives a notify value of `-1`.

In the `fix/crm/3327-respect_notify_param_in_event_api` branch, the above is handled correctly:
* The PHP notice is no more.
* The reminder is created in the database and ticked in the CRM.
* The response gives a notify value of `24`.

As a bonus, pass the same payload with a `notify` value of `-1` and the new `event_id` value. The reminder will be removed from the database.

### DIFF
--- a/projects/plugins/crm/api/create_event.php
+++ b/projects/plugins/crm/api/create_event.php
@@ -10,112 +10,102 @@
  * Date: 04/06/2019
  */
 
-	// V3.0 version of API
-
-/*
-======================================================
-	Breaking Checks ( stops direct access )
-	====================================================== */
 if ( ! defined( 'ZEROBSCRM_PATH' ) ) {
 	exit;
 }
-/*
-======================================================
-	/ Breaking Checks
-	====================================================== */
 
-		$json_params    = file_get_contents( 'php://input' );
-		$potentialEvent = json_decode( $json_params, true );
+$json_params    = file_get_contents( 'php://input' );
+$potential_task = json_decode( $json_params, true );
 
-		// define
-		$eventFields = array();
-		$eventID     = -1;
-		$event_reminders = array();
+// define
+$task_fields    = array();
+$task_id        = -1;
+$task_reminders = array();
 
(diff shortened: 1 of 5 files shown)


coder-karen (approved):
This looks great, works well (also the 'bonus')!

PR #33196, opened 2023-09-19
Status: Do not check WP.com for protect multisite option

kraftbj:
The Protect check on multisite site not needed on WordPress.com due to the unique protection setup in place there. We can specifically abort out early to avoid an extra query.

## Proposed changes:

* Updates check to end early if on WP.com.

### Other information:

## Jetpack product discussion

n/a

no

## Testing instructions:

* Load `sun` on wp.com simple and see the extra query.
* Apply patch and see no query.

### DIFF
--- a/projects/packages/status/changelog/update-no-wpcom-module-check
+++ b/projects/packages/status/changelog/update-no-wpcom-module-check
@@ -1,0 +1,4 @@
+Significance: patch
+Type: changed
+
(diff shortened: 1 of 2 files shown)


dereksmart (approved):

chrisbliss18 (approved):
Ran through the testing instructions and can confirm that it reduced the queries from 2 to 1.

PR #33200, opened 2023-09-19
Launchpad: Add earn checklist

edanzer:
**UPDATE: I am changing this PR to add a generic Earn checklist, and may add an earn-newsletter checklist later**

## Proposed changes:
* Adds 'earn' checklist to Launchpad. We'll use this to show a Launchpad checklist on Earn pages.

### Other information:

## Jetpack product discussion

None.

No.

## Testing instructions:

1) This can be tested via the testing instructions in its companion frontend PR here: 
https://github.com/Automattic/wp-calypso/pull/81930

2) To test in isolation: 
   - Run  to load this branch in your sandbox. 
   - Sandbox public-api and the domain of your test site. 
   - Go to `https://developer.wordpress.com/docs/api/console/`, select `WP REST API` and `wpcom/v2`, and enter the endpoint: `/sites/YOURDOMAIN/launchpad?checklist_slug=earn`
   - Confirm you receive back a valid checklist with the stripe_connected and paid_offer_created tasks.

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -486,6 +486,32 @@ function wpcom_launchpad_get_task_definitions() {
 				return '/page/' . $data['site_slug_encoded'];
 			},
 		),
+
+		// Earn tasks
+		'stripe_connected'                => array(
+			'get_title'           => function () {
+				return __( 'Connect a Stripe account', 'jetpack-mu-wpcom' );
+			},
+			'is_visible_callback' => '__return_true',
+			'get_calypso_path'    => function ( $task, $default, $data ) {
+				if ( function_exists( 'get_memberships_connected_account_redirect' ) ) {
+					return get_memberships_connected_account_redirect(
+						get_current_user_id(),
+						get_current_blog_id()
+					);
+				}
+				return '/earn/payments/' . $data['site_slug_encoded'];
+			},
+		),
+		'paid_offer_created'              => array(
+			'get_title'           => function () {
+				return __( 'Set up an offer', 'jetpack-mu-wpcom' );
+			},
+			'is_visible_callback' => '__return_true',
+			'get_calypso_path'    => function ( $task, $default, $data ) {
+				return '/earn/payments-plans/' . $data['site_slug_encoded'];
+			},
+		),
 	);
 
 	$extended_task_definitions = apply_filters( 'wpcom_launchpad_extended_task_definitions', array() );
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -886,25 +886,47 @@ function wpcom_launchpad_get_newsletter_subscriber_count() {
 /**
  * Determines if Stripe has been connected.
  *
- * @return bool Whether or not Stripe account is connected.
+ * @return bool Whether Stripe account is connected.
  */
-function wpcom_is_stripe_connected() {
-	require_lib( 'memberships' );
-	$blog_id  = get_current_blog_id();
-	$settings = (array) get_memberships_settings_for_site( $blog_id );
-	return boolval( $settings['connected_account_id'] );
+function wpcom_launchpad_is_stripe_connected() {
+	$membership_settings = wpcom_launchpad_get_membership_settings();
+	if ( ! $membership_settings ) {
+		return false;
+	}
+	return isset( $membership_settings['connected_account_id'] ) && $membership_settings['connected_account_id'] !== '';
 }
 
 /**
- * Determines if paid membership plans exists
+ * Determines if any paid membership plan exists.
  *
- * @return bool Whether or not paid membership plans exist.
+ * @return bool Whether paid plan exists.
  */
-function wpcom_has_paid_membership_plans() {
+function wpcom_launchpad_has_paid_membership_plans() {
+	$membership_settings = wpcom_launchpad_get_membership_settings();
+	if ( ! $membership_settings ) {
+		return false;
+	}
+	return isset( $membership_settings['products'] ) && is_array( $membership_settings['products'] ) && ( count( $membership_settings['products'] ) > 0 );
+}
+
+/**
+ * Get membership settings.
+ *
+ * @return array|null Membership settings or null.
+ */
+function wpcom_launchpad_get_membership_settings() {
+	$is_atomic = defined( 'IS_ATOMIC' ) && IS_ATOMIC;
+
+	// Memberships lib is only available on Simple sites.
+	// A follow up will fetch membership settings for Atomic.
+	if ( $is_atomic ) {
+		return null;
+	}
+
 	require_lib( 'memberships' );
 	$blog_id  = get_current_blog_id();
 	$settings = (array) get_memberships_settings_for_site( $blog_id );
-	return count( $settings['products'] ) > 0;
+	return $settings;
 }
 
 /**
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -891,7 +891,7 @@ function wpcom_launchpad_get_newsletter_subscriber_count() {
 function wpcom_is_stripe_connected() {
 	require_lib( 'memberships' );
 	$blog_id  = get_current_blog_id();
-	$settings = get_memberships_settings_for_site( $blog_id );
+	$settings = (array) get_memberships_settings_for_site( $blog_id );
 	return boolval( $settings['connected_account_id'] );
 }
 
@@ -903,7 +903,7 @@ function wpcom_is_stripe_connected() {
 function wpcom_has_paid_membership_plans() {
 	require_lib( 'memberships' );
 	$blog_id  = get_current_blog_id();
-	$settings = get_memberships_settings_for_site( $blog_id );
+	$settings = (array) get_memberships_settings_for_site( $blog_id );
 	return count( $settings['products'] ) > 0;
 }
 
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -490,7 +490,7 @@ function wpcom_launchpad_get_task_definitions() {
 		// Earn tasks
 		'stripe_connected'                => array(
 			'get_title'           => function () {
-				return __( 'Connect a Stripe account', 'jetpack-mu-wpcom' );
+				return __( 'Connect a Stripe account to collect payments', 'jetpack-mu-wpcom' );
 			},
 			'is_visible_callback' => '__return_true',
 			'get_calypso_path'    => function ( $task, $default, $data ) {
@@ -505,7 +505,7 @@ function wpcom_launchpad_get_task_definitions() {
 		),
 		'paid_offer_created'              => array(
 			'get_title'           => function () {
-				return __( 'Set up an offer', 'jetpack-mu-wpcom' );
+				return __( 'Set up an offer for your supporters', 'jetpack-mu-wpcom' );
 			},
 			'is_visible_callback' => '__return_true',
 			'get_calypso_path'    => function ( $task, $default, $data ) {
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -493,7 +493,7 @@ function wpcom_launchpad_get_task_definitions() {
 				return __( 'Connect a Stripe account to collect payments', 'jetpack-mu-wpcom' );
 			},
 			'is_visible_callback'  => '__return_true',
-			'is_complete_callback' => 'wpcom_is_stripe_connected',
+			'is_complete_callback' => 'wpcom_launchpad_is_stripe_connected',
 			'get_calypso_path'     => function ( $task, $default, $data ) {
 				if ( function_exists( 'get_memberships_connected_account_redirect' ) ) {
 					return get_memberships_connected_account_redirect(
@@ -508,7 +508,7 @@ function wpcom_launchpad_get_task_definitions() {
 			'get_title'            => function () {
 				return __( 'Set up an offer for your supporters', 'jetpack-mu-wpcom' );
 			},
-			'is_complete_callback' => 'wpcom_has_paid_membership_plans',
+			'is_complete_callback' => 'wpcom_launchpad_has_paid_membership_plans',
 			'is_visible_callback'  => '__return_true',
 			'get_calypso_path'     => function ( $task, $default, $data ) {
 				return '/earn/payments-plans/' . $data['site_slug_encoded'];
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -489,11 +489,12 @@ function wpcom_launchpad_get_task_definitions() {
 
 		// Earn tasks
 		'stripe_connected'                => array(
-			'get_title'           => function () {
+			'get_title'            => function () {
 				return __( 'Connect a Stripe account to collect payments', 'jetpack-mu-wpcom' );
 			},
-			'is_visible_callback' => '__return_true',
-			'get_calypso_path'    => function ( $task, $default, $data ) {
+			'is_visible_callback'  => '__return_true',
+			'is_complete_callback' => 'wpcom_is_stripe_connected',
+			'get_calypso_path'     => function ( $task, $default, $data ) {
 				if ( function_exists( 'get_memberships_connected_account_redirect' ) ) {
 					return get_memberships_connected_account_redirect(
 						get_current_user_id(),
@@ -504,11 +505,12 @@ function wpcom_launchpad_get_task_definitions() {
 			},
 		),
 		'paid_offer_created'              => array(
-			'get_title'           => function () {
+			'get_title'            => function () {
 				return __( 'Set up an offer for your supporters', 'jetpack-mu-wpcom' );
 			},
-			'is_visible_callback' => '__return_true',
-			'get_calypso_path'    => function ( $task, $default, $data ) {
+			'is_complete_callback' => 'wpcom_has_paid_membership_plans',
+			'is_visible_callback'  => '__return_true',
+			'get_calypso_path'     => function ( $task, $default, $data ) {
 				return '/earn/payments-plans/' . $data['site_slug_encoded'];
 			},
 		),
@@ -879,6 +881,30 @@ function wpcom_launchpad_get_newsletter_subscriber_count() {
 	}
 
 	return (int) $total_subscribers;
+}
+
+/**
+ * Determines if Stripe has been connected.
+ *
+ * @return bool Whether or not Stripe account is connected.
+ */
+function wpcom_is_stripe_connected() {
+	require_lib( 'memberships' );
+	$blog_id  = get_current_blog_id();
+	$settings = get_memberships_settings_for_site( $blog_id );
+	return boolval( $settings['connected_account_id'] );
+}
+
+/**
+ * Determines if paid membership plans exists
+ *
+ * @return bool Whether or not paid membership plans exist.
+ */
+function wpcom_has_paid_membership_plans() {
+	require_lib( 'memberships' );
+	$blog_id  = get_current_blog_id();
+	$settings = get_memberships_settings_for_site( $blog_id );
+	return count( $settings['products'] ) > 0;
 }
 
 /**
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
@@ -191,6 +191,14 @@ function wpcom_launchpad_get_task_list_definitions() {
 			),
 			'is_enabled_callback' => 'wpcom_launchpad_is_paid_newsletter_enabled',
 		),
+		'earn-newsletter'        => array(
+			'title'               => 'Newsletter',
+			'task_ids'            => array(
+				'set_up_payments',
+				'newsletter_plan_created',
+			),
+			'is_enabled_callback' => 'wpcom_launchpad_is_newsletter_or_write_intent',
+		),
 	);
 
 	$extended_task_list_definitions = apply_filters( 'wpcom_launchpad_extended_task_list_definitions', array() );
@@ -913,6 +921,20 @@ function wpcom_launchpad_is_paid_newsletter_enabled() {
 	}
 
 	return wpcom_launchpad_has_goal_paid_subscribers() && apply_filters( 'wpcom_launchpad_intent_paid_newsletter_enabled', false );
+}
+
+/**
+ * Checks if site has newsletter or write as site intent.
+ *
+ * @return bool True if the task list is enabled, false otherwise.
+ */
+function wpcom_launchpad_is_newsletter_or_write_intent() {
+	$intent = get_option( 'site_intent', false );
+	if ( 'newsletter' !== $intent || 'write' !== $intent ) {
+		return true;
+	}
+
+	return false;
 }
 
 // Unhook our old mu-plugin - this current file is being loaded on 0 priority for `plugins_loaded`.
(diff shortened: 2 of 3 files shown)


valterlorran:
Hi @edanzer, would you mind adding a standalone testing instruction here? We can test these changes by requesting `/sites/:siteSlug/launchpad?checklist_slug=earn-newsletter` through the developer console.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
```diff
@@ -907,6 +915,20 @@ function wpcom_launchpad_is_paid_newsletter_enabled() {
...
 }
 
+/**
+ * Checks if site has newsletter or write as site intent.
+ *
+ * @return bool True if the task list is enabled, false otherwise.
+ */
+function wpcom_launchpad_is_newsletter_or_write_intent() {
+	$intent = get_option( 'site_intent', false );
+	if ( 'newsletter' !== $intent || 'write' !== $intent ) {
```
daledupreez: This condition will always be true. I think we may need `&&` here instead of `||`.
```suggestion
	if ( 'newsletter' !== $intent && 'write' !== $intent ) {
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
```diff
@@ -907,6 +915,20 @@ function wpcom_launchpad_is_paid_newsletter_enabled() {
 	return wpcom_launchpad_has_goal_paid_subscribers() && apply_filters( 'wpcom_launchpad_intent_paid_newsletter_enabled', false );
 }
 
+/**
+ * Checks if site has newsletter or write as site intent.
+ *
+ * @return bool True if the task list is enabled, false otherwise.
+ */
+function wpcom_launchpad_is_newsletter_or_write_intent() {
```
daledupreez: Putting aside the logic issue with the condition in the function, I think we may need to revisit the naming for this function. As far as I can tell, the function should return true when the intent is _not_ `write` or `newsletter`. Can we update the naming so the naming aligns with what we're checking.
```suggestion
function wpcom_launchpad_is_not_newsletter_or_write_intent() {
```

edanzer: @daledupreez The intent is to return true if intent is either write or newsletter (ie, this checklist should be enabled for sites with write/newsletter intent). I do need to adjust though.

```
function wpcom_launchpad_is_newsletter_or_write_intent() {
	$intent = get_option( 'site_intent', false );
	if ( 'newsletter' === $intent || 'write' === $intent ) {
		return true;
	}

	return false;
}
```

I'll update. One odd thing I need to explore is why this has been working as intended when tested alongside my frontend PR.

edanzer: I'm testing this via dev console know. No matter whether I update that condition or not, I get back the same expected checklist, which is why it works. I guess the callback just changes the value of he is_enabled property on the response.

edanzer:
@daledupreez @valterlorran -

1) I've added specific instructions for testing via the dev console.

2) I've fixed/updated the conditional for the is_enabled callback. Pretty obvious logic issue. I was surprised when I saw it that it was working at tall. But I think I misunderstood the purpose of the is_enabled callback. Regardless of whether that callback returns true or false, you still get back a checklist with the right tasks. So the frontend PR was working. It's just a matter of whether the is_enabled property on the response root is true or false. Not sure what downstream effect that has, but I was not making use of that property myself. 

Anyways, I believe this PR should be working as expected now. Thanks for the reviews.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
```diff
@@ -907,6 +915,20 @@ function wpcom_launchpad_is_paid_newsletter_enabled() {
...
+	$intent = get_option( 'site_intent', false );
+	if ( 'newsletter' === $intent || 'write' === $intent ) {
```
valterlorran: Can we restrict this a little more? I tested with a Free Newsletter and a site with the Write intent, and both returned `enabled = true`, but the checklist was empty. So maybe it shouldn't be enabled when the checklist is empty.

Screen Shot 2023-09-28 at 17 13 24

edanzer: > but the checklist was empty

Hmm. Not able to duplicate this. I wouldn't think an empty checklist would be related to the is_enabled callback? If you're seeing an empty checklist, that would seem like it would reflect some other other issue wouldn't it? 

To put it otherwise, if I send a request to `/sites/YOURDOMAIN/launchpad?checklist_slug=earn-newsletter`, given that I'm providing a specific checklist slug, shouldn't I **always** get back the `earn-newsletter` checklist, which is defined to have two items? 

From playing with it, I'd think the is_enabled callback only changes whether the is_enabled property on the request is true or false.

edanzer: To confirm, you'll get 0 checklist items if the site_goals don't include paid_subscribers because the task definitions for set_up_payments and create_paid_newsletter both include `'is_visible_callback' => 'wpcom_launchpad_has_goal_paid_subscribers',`

I need to think how to handle that. I need to be able to load those tasks even if a site doesn't have that goal. 

Options: 
   - Just set a site_goal of paid_newsletters when someone visits the earn page (since they've expressed monetization intent)
   - Create different tasks that don't depend on that is_visible callback.

edanzer: FYI - I recall now that is_visible callback was originally set because we're using the same task definitions for free and paid newsletters and didn't want to show them on paid newsletters. But that intention doesn't travel well at this point, later, when we want to encourage those with free newsletters to monetize. 

It's also a bit odd / non-ideal because I've created a checklist specifically to fetch those two items, but I can't do so without updating the site_goals.

edanzer: I've made several changes based on this and also broader discussion about Launchpad on Earn:
   - I've created two new tasks for for stripe_connected and paid_offer_created that don't have the limitation of needing to have a paid_newsletter goal. 
   - Both tasks above have __return_true as their is_visible callback. For now, I can't think of any reason we would not want those visible on the Earn launchpad.
   - I've updated the checklist slug to be 'earn' instead of 'earn-newsletter'. We had been planning to add a general launchpad to earn after doing it for newsletters. I'm reversing the order and doing the general launchpad first.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
```diff
@@ -183,6 +183,14 @@ function wpcom_launchpad_get_task_list_definitions() {
 			),
 			'is_enabled_callback' => 'wpcom_launchpad_is_paid_newsletter_enabled',
 		),
+		'earn'                   => array(
+			'title'               => 'Newsletter',
```
daledupreez: Minor note: this should probably be `Earn` or similar, as it's no longer specific to newsletters.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -456,6 +456,32 @@ function wpcom_launchpad_get_task_definitions() {
 				return '/page/' . $data['site_slug_encoded'];
 			},
 		),
+
+		// Earn tasks
+		'stripe_connected'                => array(
+			'get_title'           => function () {
+				return __( 'Connect a Stripe account', 'jetpack-mu-wpcom' );
```
daledupreez: Users may not know what `Stripe` is, or why they might need an account, so I think we may need additional context here. How about `Connect a Stripe account for payments` or `Connect a Stripe account to collect payments`?

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -456,6 +456,32 @@ function wpcom_launchpad_get_task_definitions() {
...
+						get_current_user_id(),
+						get_current_blog_id()
+					);
+				}
+				return '/earn/payments/' . $data['site_slug_encoded'];
+			},
+		),
+		'paid_offer_created'              => array(
+			'get_title'           => function () {
+				return __( 'Set up an offer', 'jetpack-mu-wpcom' );
```
daledupreez: I have no idea what an "offer" is or who it might be for. Assuming this is a "payment plan", can we use wording like `Set up a payment plan for [site] visitors`?

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -456,6 +456,32 @@ function wpcom_launchpad_get_task_definitions() {
...
+
+		// Earn tasks
+		'stripe_connected'                => array(
+			'get_title'           => function () {
+				return __( 'Connect a Stripe account', 'jetpack-mu-wpcom' );
+			},
+			'is_visible_callback' => '__return_true',
+			'get_calypso_path'    => function ( $task, $default, $data ) {
+				if ( function_exists( 'get_memberships_connected_account_redirect' ) ) {
+					return get_memberships_connected_account_redirect(
```
daledupreez, line 499: I followed the link that gets returned here and it looks like we're push users directly to Stripe to kick off the connection workflow. I think we need to make sure we're giving the user more context to make sure they have a good sense of what will happen next when they visit the site, as the change may feel really abrupt.

daledupreez (commented):
This is working as described, but I have some concerns about the UX we'll end up with. Could you include some links to the designs or other context that you're working from so I can share that feedback in the right place?

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
daledupreez: Noting that I've shared this feedback on the original designs at p9Jlb4-8NN-p2#comment-9257.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
daledupreez: Also shared on the original design: p9Jlb4-8NN-p2#comment-9257

edanzer:
UPDATE: Got it. Additional details on this here: p9Jlb4-8NN-p2#comment-9257

Hey @daledupreez - Here are Figma designs for the earn page: DqXCQr1dEWpF3P2dIEwiwd-fi-791_88014
   - You mentioned 'concerns about the ux' based on this PR... can you clarify what you may be concerned about? 
   - I'll ping and talk to Zap about a) if we might update the the task text for 'stripe' and 'offer' and b) what if anything we might do to provide some intermediate context between the task itself and Stripe.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
edanzer: Yup, updated to 'Earn'.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
edanzer: Updated this and payment text string based on comments here: 
p9Jlb4-8NN-p2#comment-9257

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
edanzer: Based on the P2 comments linked elsewhere, we're still going directly to stripe, but have updated the text strings to have greater clarity.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
edanzer: Updated this based on the separate P2 discussion.

edanzer:
For @daledupreez and others... I've updated the checklist title to 'Earn' and updated the task text strings for greater clarity based on the related conversation at p9Jlb4-8NN-p2#comment-9257. We're still be linking directly to Stripe, but sounds like that's fine given the additional text string clarity.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
edanzer: Just a note that the method for which this comment was added has been removed because I've updated this PR to be generic, and to show the earn launchpad for all sites, not just newsletter / blog / write.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
edanzer: Just a note that the method for which this comment was added has been removed because I've updated this PR to be generic, and to show the earn launchpad for all sites, not just newsletter / blog / write.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -855,6 +883,30 @@ function wpcom_launchpad_get_newsletter_subscriber_count() {
 	return (int) $total_subscribers;
 }
 
+/**
+ * Determines if Stripe has been connected.
+ *
+ * @return bool Whether or not Stripe account is connected.
+ */
+function wpcom_is_stripe_connected() {
```
lezama: should we add it to the `lib` instead? sounds useful

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
```diff
@@ -191,6 +191,14 @@ function wpcom_launchpad_get_task_list_definitions() {
 			),
 			'is_enabled_callback' => 'wpcom_launchpad_is_paid_newsletter_enabled',
 		),
+		'earn'                   => array(
+			'title'               => 'Earn',
+			'task_ids'            => array(
+				'stripe_connected',
+				'paid_offer_created',
+			),
+			'is_enabled_callback' => '__return_true',
```
lezama, line 200: sure this doesn't make it show up somewhere else?

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
daledupreez: Agreed -- this should be `wpcom_launchpad_is_stripe_connected()` if we're keeping it in the Launchpad code, or it should move somewhere else (where "somewhere else" could still be in jetpack-mu-wpcom).
```suggestion
function wpcom_launchpad_is_stripe_connected() {
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -855,6 +883,30 @@ function wpcom_launchpad_get_newsletter_subscriber_count() {
...
+	require_lib( 'memberships' );
```
daledupreez: Is this library defined on Atomic sites or in Jetpack? Assuming it's available via Jetpack, we also need to make sure that the relevant module(s) are enabled.
At a glance, this looks likely to break in that context.

Also, I really dislike `require_lib()` because I don't know what code it is actually looking for in any particular context.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -855,6 +883,30 @@ function wpcom_launchpad_get_newsletter_subscriber_count() {
...
+	return boolval( $settings['connected_account_id'] );
+}
+
+/**
+ * Determines if paid membership plans exists
+ *
+ * @return bool Whether or not paid membership plans exist.
+ */
+function wpcom_has_paid_membership_plans() {
+	require_lib( 'memberships' );
```
daledupreez, line 926: As above for the source of the library/code.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -855,6 +883,30 @@ function wpcom_launchpad_get_newsletter_subscriber_count() {
...
+/**
+ * Determines if Stripe has been connected.
+ *
+ * @return bool Whether or not Stripe account is connected.
+ */
+function wpcom_is_stripe_connected() {
+	require_lib( 'memberships' );
+	$blog_id  = get_current_blog_id();
+	$settings = (array) get_memberships_settings_for_site( $blog_id );
+	return boolval( $settings['connected_account_id'] );
```
daledupreez: Will this output a warning if `$settings['connected_account_id']` is undefined? Should we add an `isset()` wrapper to be super-safe?
```suggestion
	return isset( $settings['connected_account_id'] ) && boolval( $settings['connected_account_id'] );
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -855,6 +883,30 @@ function wpcom_launchpad_get_newsletter_subscriber_count() {
...
+/**
+ * Determines if paid membership plans exists
+ *
+ * @return bool Whether or not paid membership plans exist.
+ */
+function wpcom_has_paid_membership_plans() {
+	require_lib( 'memberships' );
+	$blog_id  = get_current_blog_id();
+	$settings = (array) get_memberships_settings_for_site( $blog_id );
+	return count( $settings['products'] ) > 0;
```
daledupreez: Are we guaranteed that `$settings['products']` will be defined and an array? I feel we should be defensive against the data being undefined, especially as `count()` will throw exceptions for null values in newer versions of PHP.
```suggestion
	return isset( $settings['products'] ) && is_array( $settings['products'] ) && ( count( $settings['products'] ) > 0 ); // we could also use `array() !== $settings['products']` for the final check if we want
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -855,6 +883,30 @@ function wpcom_launchpad_get_newsletter_subscriber_count() {
...
+	$settings = (array) get_memberships_settings_for_site( $blog_id );
+	return boolval( $settings['connected_account_id'] );
+}
+
+/**
+ * Determines if paid membership plans exists
+ *
+ * @return bool Whether or not paid membership plans exist.
+ */
+function wpcom_has_paid_membership_plans() {
```
daledupreez: As above RE naming, can we please keep the `wpcom_launchpad_` prefix, or move the code elsewhere?
```suggestion
function wpcom_launchpad_has_paid_membership_plans() {
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
daledupreez: I think this should be OK -- we don't have any reason a user shouldn't be able to load the task list.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
edanzer: Sure. Can handle as a follow up. Your comment made me think I should double check it doesn't already exists in the lib. As far as I can tell, it doesn't, which is surprising.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
edanzer: No, I don't think there's anyplace where a launchpad checklist is automatically displayed based on this property.

daledupreez (commented):
I've noted a few key items as comments.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
edanzer: Yes, I was just asking about this on slack. The require_lib() method will always exist in simple/atomic environments, but certain libraries may not. Someone thought memberships may not be present by default on atomic, but I'm not sure how that could be possible given how much other code depends on it. 

I don't have strong feelings one way or the other for require_lib(), but it is very widely used in our codebases, so I've assumed it is acceptable, or even preferred, for many. But I'm happy to adjust.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
edanzer: Sure! Happy to update.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
edanzer: Good call. Added the extra checks.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
edanzer: Good call. Added the extra checks.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
edanzer: Updated method names.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
edanzer: @daledupreez - I've updated this so we're only checking membership settings on simple for now. Checking the membership settings on atomic seems like it will require some extra work. 

I'm still using require_lib( 'memberships' ) as it still seems like the most straightforward and common way to pull in the needed methods. Let me know if you have an alternative approach in mind.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
edanzer: https://github.com/Automattic/jetpack/pull/33200#discussion_r1355760498

lezama (approved):

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
daledupreez: Just to clarify, my dislike of `require_lib()` stems from the fact that it effectively masks the underlying dependency for a casual reader, making the dependency somewhat implicit, unless the reader happens to recall exactly how `require_lib()` does lookups. In this particular case, I can't tell what `'memberships'` is, so I can't easily tell what code we're expecting to be present (and is not present, in some cases!).

As a proposed alternative, I always prefer an explicit `require_once`, as it is generally 100% explicit as to what you're trying to pull in. In this specific case, I would go for something like:
```php
	require_once WP_CONTENT_DIR . '/lib/memberships/0-load.php';
```
To me, it would have been way easier to detect a possible problem with the expectations for this more explicit `require_once` approach, and very clear as to how we'd detect whether the files are present. 😁

edanzer: @daledupreez - I do understand your point. I had to do a fair amount of research today to understand how require_lib() works, where it searches for libraries, how it loads them, and when it's safe to use. By comparison, a require statement is simple and direct.

I wasn't quite sure how to include the library and associated methods without it, but you're right that including that 0-load file should do it. That's basically what require_lib() does under the hood. I'd be happy to update in a follow up.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -855,6 +883,52 @@ function wpcom_launchpad_get_newsletter_subscriber_count() {
...
+/**
+ * Get membership settings.
+ *
+ * @return array|null Membership settings or null.
+ */
+function wpcom_launchpad_get_membership_settings() {
+	$is_atomic = defined( 'IS_ATOMIC' ) && IS_ATOMIC;
+
+	// Memberships lib is only available on Simple sites.
+	// A follow up will fetch membership settings for Atomic.
```
daledupreez, line 921: Minor note: the code _and_ the data are only present and directly accessible from WPCOM Simple.

It looks like we should be able to use code very similar to the code block below to fetch this data when we're in an Atomic context. That code is for the `wpcom/v2/memberships/status` endpoint, which will make a proxied API call to WordPress.com when we're not running on WPCOM Simple.
https://github.com/Automattic/jetpack/blob/cce10b2a0c92b53b79bda816e8fcd3899fdb30ad/projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/memberships.php#L369-L401

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -855,6 +883,52 @@ function wpcom_launchpad_get_newsletter_subscriber_count() {
...
+	}
+	return isset( $membership_settings['products'] ) && is_array( $membership_settings['products'] ) && ( count( $membership_settings['products'] ) > 0 );
+}
+
+/**
+ * Get membership settings.
+ *
+ * @return array|null Membership settings or null.
+ */
+function wpcom_launchpad_get_membership_settings() {
```
daledupreez, line 917: Neatly done in wrapping all this logic and ensuring the code won't explode outside of WPCOM Simple.

PR #33203, opened 2023-09-20
Google Fonts: Integrate with new font library

arthur791004:
Related to https://github.com/Automattic/wp-calypso/issues/81875

## Proposed changes:

* As suggested, we served [the collection of the Jetpack Google Fonts on wpcom](https://s0.wp.com/i/font-collections/jetpack-google-fonts.json). The collection is generated via the `generate_google_fonts` CLI tool on wpcom.
* Next, this PR gets the collection data from the remote and injects the fonts to the default global styles settings on the fly by the `wp_theme_json_data_default` hook so we can display Google Fonts in the dropdown of the Font Family.
* Why the default?
  * The theme global styles will be overridden after people pick any style variation. It leads to the font family from the theme global styles disappearing. Hence, it would be better to use the default global styles instead of the theme. See https://github.com/Automattic/dotcom-forge/issues/3162, https://github.com/Automattic/wp-calypso/issues/78605 for more context.
  * There is a [WIP PR](https://github.com/WordPress/gutenberg/pull/54888) to allow developers to register the default fonts and display them in the “Default Fonts” section. The benefits are the default fonts are activated by default and cannot be configurable. So, when this module is deactivated, the fonts provided by this module won't stick in the user's global styles settings. ~However, the default fonts are invisible on the block settings as [the block settings pick font families from one of the origins](https://github.com/WordPress/gutenberg/issues/55011#issuecomment-1748198200). But I assume it's acceptable for now since the user will only see the installed fonts there if they install a font. As a result, I think it makes to inject them into the default now~ (The issue is resolved 🎉)

image

### Other information:

## Testing instructions:

* Create a [Jurassic Ninja sandbox](https://jurassic.ninja/create/?jetpack-beta&branch=feat/google-fonts-with-new-font-library) against this PR.
* Install [the latest Gutenberg](https://github.com/WordPress/gutenberg/releases), e.g.: v16.7.0-rc.1
* Go to `/wp-admin/admin.php?page=jetpack_modules` to activate the “Google Fonts (Beta)” module
* Go to the Site Editor
* Select Styles > Typography > Text > Font
* Ensure you're able to see all Google Fonts in the list
* Disable the Gutenberg plugin
* Select Styles > Typography > Text > Font again
* Ensure you can still see all Google Fonts in the list


pbking:
At first shake I notice:

Enabling and dis-abling this module causes the fonts introduced by this module to stick around. 
Changing what fonts are in the list don't cause fonts added to be removed from the Global Styles list.
It did seem to remove them from other lists?

I didn't test a lot of scenarios though, just tinkered with it for a bit.

Ultimately this seems like a lot of logic in place for something the Font Library should be doing; I'm concerned that if this is they way Jetpack would augment the Font Library then the pipes to work with the Font Library aren't quite right yet.  I would rather try to do this more simply while the Font Library is relatively malleable. 

Ideally:

The module could provide a list of fonts to make available with a `.json` file.
These fonts point to remote locations for the assets (as outlined [here](https://github.com/WordPress/gutenberg/pull/54477/files?short_path=7b011a0#diff-7b011a05b51e36c6b10b8aa64a5f0cf6a5ae535e3810bbb79ad130768829cdcd))
to take advantage of publicly available CDNs OR 
(I think more useful) include local references to the fonts (just like themes do).
(just as with themes) changing the collection in the `.json` file changes what fonts are available to the user.
Fonts that the user has used in content or Global Styles settings are automatically activated.

arthur791004:
> Enabling and dis-abling this module causes the fonts introduced by this module to stick around.
Changing what fonts are in the list don't cause fonts added to be removed from the Global Styles list.
It did seem to remove them from other lists?

Yes, so I'd prefer to keep the Google Fonts invisible on the Font Library modal in the current iteration to avoid people controlling the fonts from the Jetpack. I'm not sure whether the Google Fonts module has to remove the added fonts from the Global Styles when disabling this module (and is it possible?).

> The module could provide a list of fonts to make available with a .json file.
These fonts point to remote locations for the assets (as outlined [here](https://github.com/WordPress/gutenberg/pull/54477/files?short_path=7b011a0#diff-7b011a05b51e36c6b10b8aa64a5f0cf6a5ae535e3810bbb79ad130768829cdcd))
to take advantage of publicly available CDNs OR
(I think more useful) include local references to the fonts (just like themes do).

Yep, I just noticed we have the collection of the JSON file but my concern is the `downloadFromUrl` might be changed by Google. Is it correct? Or is the asset URL available forever?

Maybe keeping the font assets in the Jetpack repo makes more sense but I'm unsure whether Jetpack allows us to add those font assets 🤔

arthur791004:
@pbking I've created a JSON file of the font collection [here](https://s0.wp.com/i/font-collections/jetpack-google-fonts.json) and now this module loads the data from the remote and injects the fonts we want to the default global styles settings!

projects/plugins/jetpack/modules/google-fonts/load.php
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * @package automattic/jetpack
+ */
+
+if ( class_exists( 'WP_Font_Library' ) && class_exists( 'WP_Font_Face' ) && class_exists( 'WP_Font_Collection' ) ) {
+	// WordPress 6.4 or above with the new Font Library.
+	require_once __DIR__ . '/current/load-google-fonts.php';
+} elseif ( class_exists( 'WP_Fonts' ) || class_exists( 'WP_Webfonts' ) ) {
+	// WordPress 6.3 compat.
+	require_once __DIR__ . '/wordpress-6.3/load-google-fonts.php';
```
pbking: Calling this "WordPress 6.3 compatible" isn't exactly correct.
It's more accurate to say that it's "Gutenberg Fonts API compatible" (since the API it leverages is available only in Gutenberg, that library MAY be available in other versions of WordPress, etc)

projects/plugins/jetpack/modules/google-fonts/load.php
```diff
@@ -0,0 +1,14 @@
+<?php
+/**
+ * Load the google fonts based on the current WordPress version.
+ *
+ * @package automattic/jetpack
+ */
+
+if ( class_exists( 'WP_Font_Library' ) && class_exists( 'WP_Font_Face' ) && class_exists( 'WP_Font_Collection' ) ) {
```
pbking: The Font Library won't be available in WordPress 6.4 since it's currently only available in Gutenberg.  It's planned to be available in WordPress 6.5.  However this technique IS available in situations other than WordPress 6.4 (or 6.5).

arthur791004: Are there any changes I have to make here? I made a change to only check the `WP_Font_Face` class as I assume we still rely on it to generate and print font styles. WDYT?

pbking: The class isn't actually used or depended on at all as the logic exists now, WP_Font_Collection WAS used but now that collection data is just loaded manually and the class isn't needed.  The data provided in that file that is loaded is everything needed for the fonts to work; no extra logic needed. That allows the logic to be font library/api agnostic.  Doing it like this means that it ONLY works in Gutenberg, wereas it COULD work WITHOUT Gutenberg.  Considering that it HASN'T worked without Gutenberg in the past I don't know if that matters though.

So I guess the answer is "If you want this to work WITHOUT gutenberg it CAN you'll just need to change that if() logic a bit."  "If you don't care for this to work without Gutenberg then it's fine and dandy as it is"

projects/plugins/jetpack/modules/google-fonts/wordpress-6.3/constants.php
```diff
@@ -0,0 +1,61 @@
+<?php
+/**
+ * Define the constants of the Jetpack Google Fonts module.
+ *
+ * @package automattic/jetpack
+ */
+
+/**
+ * Curated list of Google Fonts
+ * See https://wp.me/p9Jlb4-22P
```
jeherve: Since we're making changes, we may as well fix that. Let's avoid direct links like that:

```suggestion
 * See p9Jlb4-22P-p2
```

projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
```diff
@@ -0,0 +1,137 @@
+<?php
+/**
+ * Load the google fonts by the new Font Library. See https://wp.me/pNEWy-hhx.
```
jeherve: Let's switch to shorthand:

```suggestion
 * Load the google fonts by the new Font Library. See pNEWy-hhx-p2.
```

projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
```diff
@@ -0,0 +1,137 @@
...
+		return null;
+	}
+
+	$data = json_decode( wp_remote_retrieve_body( $response ), true );
+	if ( $data === null ) {
+		return null;
+	}
+
+	// Replace the google fonts api url if the custom one is provided.
+	$custom_google_fonts_api_url = \esc_url( apply_filters( 'jetpack_google_fonts_api_url', '' ) );
```
jeherve: Let's make our codex parser happy :)

```suggestion
	$custom_google_fonts_api_url = esc_url(
		/** This filter is documented in projects/packages/google-fonts-provider/src/class-google-fonts-provider.php */
		apply_filters( 'jetpack_google_fonts_api_url', '' )
	);
```

projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
```diff
@@ -0,0 +1,137 @@
...
+ * @return object[] The map of the the available Google Fonts.
+ */
+function jetpack_get_available_google_fonts_map( $google_fonts_data ) {
+	$jetpack_google_fonts_list  = array_map(
+		function ( $font_family ) {
+			return $font_family['name'];
+		},
+		$google_fonts_data['fontFamilies']
+	);
+	$google_font_list           = apply_filters( 'jetpack_google_fonts_list', $jetpack_google_fonts_list );
```
jeherve, line 62:
```suggestion
	/** This filter is documented in modules/google-fonts/wordpress-6.3/load-google-fonts.php */
	$google_font_list           = apply_filters( 'jetpack_google_fonts_list', $jetpack_google_fonts_list );
```

projects/plugins/jetpack/modules/google-fonts/load.php
arthur791004: Okay. I'll leave it as it is since it hasn't worked without Gutenberg in the past. I'd prefer to make it work without Gutenberg after WordPress 6.5 or above which supports the `WP_Font_Face` class.

jeherve (dismissed):
This tests well for me, and the changes make sense. I left a few nit-picks below, but nothing blocking (maybe the links, since we really try to stay away from private links in public repos in general?).

arthur791004:
> I left a few nit-picks below, but nothing blocking (maybe the links, since we really try to stay away from private links in public repos in general?).

Oh yes! We have to 👍

jeherve (approved):

PR #33206, opened 2023-09-20
Newsletter categories: Improve the handling of theme colors on both frontend & editor.

TimBroddin:
## Proposed changes:
- Improves the handling of theme colors on both frontend & editor.
- Fixes the wrapping in the editor

This defines 3 CSS variables:

- `--subscribe-block-accent-color`
- `--subscribe-block-text-color`
- `--subscribe-block-background-color`

Depending on the environment, these get filled in with values from the theme used.

### Other information:

## Testing instructions:
- Apply this PR
- Add a subscribe block to a site with newsletter categories enabled 
- Test with different themes

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/edit.js
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/edit.js
@@ -187,6 +187,9 @@ export function SubscriptionEdit( props ) {
 		'--subscribe-block-border-radius': borderRadius
 			? borderRadius + 'px'
 			: DEFAULT_BORDER_RADIUS_VALUE + 'px',
+		'--subscribe-block-text-color': 'var(--wp--preset--color--foreground)',
+		'--subscribe-block-background-color': 'var(--wp--preset--color--background)',
+		'--subscribe-block-accent-color': 'var(--wp--preset--color--foreground)',
 	};
 
 	const emailFieldStyles = {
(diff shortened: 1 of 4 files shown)


TimBroddin:
Don't review this yet, I still need to do some testing.

projects/plugins/jetpack/extensions/blocks/subscriptions/edit.js
```diff
@@ -187,6 +187,9 @@ export function SubscriptionEdit( props ) {
 		'--subscribe-block-border-radius': borderRadius
 			? borderRadius + 'px'
 			: DEFAULT_BORDER_RADIUS_VALUE + 'px',
+		'--subscribe-block-text-color': 'var(--wp--preset--color--foreground)',
+		'--subscribe-block-background-color': 'var(--wp--preset--color--background)',
+		'--subscribe-block-accent-color': 'var(--wp--preset--color--foreground)',
```
jeherve: This may be a stupid question, but will this work with classic themes as well?

TimBroddin: Not a stupid question at all 😃 

I assume it will fall back to the predefined defaults on Classic themes, but I need to conduct some more testing tomorrow. 

```
const DEFAULT_ACCENT_COLOR        = '#3c434a';
const DEFAULT_BACKGROUND_COLOR    = 'white';
const DEFAULT_TEXT_COLOR          = '#3c434a';
```
FYI: these newsletter categories will move to the `subscribe.wordpress.com` modal in i2 of this project, which will follow this iteration very soon. I'll ask @yansern if it makes sense to limit rendering the newsletter categories only to WPCOM. p9Jlb4-8ZT-p2

yansern: > I'll ask @yansern if it makes sense to limit rendering the newsletter categories only to WPCOM. p9Jlb4-8ZT-p2

Yes, limit the rendering of pills to just WPCOM (Simple & Atomic).

Jetpack's version will go straight to the modal version (i2) as seen in the P2 p9Jlb4-8ZT-p2.

TimBroddin:
I did some testing. I'd say it works most of the times, although for Livro Black the contrast is way too low.

| Twenty Sixteen (classic) | Lettre  |
|-|-|
| CleanShot 2023-09-21 at 10 02 31@2x | CleanShot 2023-09-21 at 09 41 00@2x |
| Livro Brown | Livro Black |
| CleanShot 2023-09-21 at 10 01 23@2x | CleanShot 2023-09-21 at 09 46 16@2x |
| Livro White | |
| CleanShot 2023-09-21 at 09 44 58@2x | | 

I attempted to use Jetpack Color to use a more readable contrast,

```php

	// make an attempt to make accent color readable on top of background color
	if ( $subscribe_block_background_color && $subscribe_block_accent_color ) {
		$bg_color = new \Jetpack_Color( $subscribe_block_background_color );
		$accent_color = new \Jetpack_Color( $subscribe_block_accent_color );
		$readable_accent_color = $accent_color->getReadableContrastingColor( $bg_color );
		if ( $readable_accent_color ) {
			$subscribe_block_accent_color = $readable_accent_color->toString();
		}
	}
```

But this fails, since `$global_styles` can contain CSS variables as a color value (such as `--wp--preset--color--background`), and I don't find a way to resolve these to their real value.

projects/plugins/jetpack/extensions/blocks/subscriptions/edit.js
TimBroddin: @jeherve as you predicted, this doesn't work in the editor for classic themes:

CleanShot 2023-09-21 at 10 19 10@2x

I'll append the default values to the styles here 👍

yansern (approved):
Tested a bunch of themes. Dark themes will have contrast issues. This is temporary as we're moving to subscribe modal (which will not inherit theme colors for now). 

Good to go.

Issue #33170, opened 2023-09-19
Lazy images: deprecate from Jetpack Boost


jeherve:
WordPress 6.4, scheduled to be released on November 7, 2023, will start conflicting with our Lazy Images functionality. Let's aim to have it removed from the Boost plugin by then.

```[tasklist]
### Tasks
- [ ] https://github.com/Automattic/jetpack/issues/33953
- [x] Remove the feature from Boost UI
```

*********

Primary issue: #33048 
Related issue for the Jetpack plugin: #33049

jeherve:
cc @haqadn Leaving that one to you :) Feel free to ping me for any coordination with the Jetpack plugin!

haqadn:
Thanks for creating the issue. @ederrengifo is helping us remove the feature while not leaving the users at a state of confusion.

jeherve:
Noting that in #33208, I'm poking at some of Boost' internals to address some of that confusion. I'll be happy to have your opinion on this!

ederrengifo:
Explored designs for the notifications to show and an idea for emailing `p1HpG7-oTQ`

ChrisBegley:
Hello, the Tiled Gallery block doesn't currently support native Lazy Loading: https://github.com/Automattic/jetpack/issues/33247

Without the Lazy Loading feature of Jetpack, there won't be any way to lazy load Tiled Galleries. That'll add a lot of "weight" to pages with galleries.

jeherve:
> Without the Lazy Loading feature of Jetpack, there won't be any way to lazy load Tiled Galleries. That'll add a lot of "weight" to pages with galleries.

That's correct. However, we must still proceed with deprecating Jetpack's Lazy Images implementation, as it conflicts with some recent changes in the Gutenberg plugin, that will come to WordPress in version 6.4, scheduled for next month.

#32598 already improved the performance of tiled galleries. Adding native lazy loading should help too, and we'll close #33247 when that's implemented.

haqadn:
Closing the issue as this is now complete

Issue #33049, opened 2023-09-13
Lazy Images: deprecate from the Jetpack plugin


kraftbj:
### Impacted plugin

Jetpack

### Quick summary

Jetpack-specific work for #33048 

Needed: updating the module, updating the admin dashboard, updating Calypso Blue

I'd like to see as much of the plugin-side work in for the October 10th release to provide buffer before WP 6.4's expected ship date of Nov 7 (which would also be the day of a Jetpack release).

I'm open to thoughts if we should just rip it out or discontinue with WP 6.4, etc. 

Internal reference: pdWQjU-wV-p2
```[tasklist]
### Tasks
- [ ] https://github.com/Automattic/jetpack/pull/33039
- [ ] https://github.com/Automattic/jetpack/pull/33030
- [ ] https://github.com/Automattic/jetpack/issues/33168
- [ ] https://github.com/Automattic/jetpack/issues/33166
- [ ] https://github.com/Automattic/jetpack/issues/33169
- [ ] https://github.com/Automattic/jetpack/issues/33167
- [ ] https://github.com/Automattic/jetpack/issues/33171
- [ ] https://github.com/Automattic/jetpack/pull/36155/
```

jeherve:
This can now be closed, since the feature is no longer in the Jetpack plugin as of https://github.com/Automattic/jetpack/pull/33782.

PR #33039, opened 2023-09-13
Lazy Images: disable when Gutenberg 16.6.0+ is active

kraftbj:
Fixes https://github.com/WordPress/gutenberg/issues/54396
Related https://github.com/Automattic/jetpack/issues/24290

## Proposed changes:

* Gutenberg's new Interactivity API and Lazy Images conflict, so aborting out when the instance is init.
* Lazy images is not needed with the vast majority of browsers having native support. Intent is to deprecation functionality completely.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

pdWQjU-wV-p2

## Does this pull request change what data or activity we track or use?

No.

## Testing instructions:

* The unit tests should be enough, but Jetpack + Lazy Images + Gutenberg 16.6
* Ensure that lazy images from Jetpack is not being used.

Issue #33169, opened 2023-09-19
Lazy Images: update Calypso module card to match wp-admin


jeherve:
We should update the UI in Calypso to match the changes that will be introduced in #33166. 

*******

Primary issue: #33049

PR #33208, opened 2023-09-20
Lazy Images: introduce deprecation notice in UI, and do not allow enabling it when disabled

### LINKED ISSUE #33166: Lazy Images: update module card to only allow disabling it
**Current experience**

image

#33030 takes a first step to encourage folks to use Core's Lazy loading functionality instead. 

In this task, we aim to do more:

1. Display a notice when the feature is disabled because it conflicts with a Gutenberg feature (see #33039).
2. Do not display the toggle when the feature is disabled.
3. Display a deprecation notice next to the toggle when the feature is enabled.

*******

Primary issue: #33049

jeherve:
Fixes #33166

## Proposed changes:

In this PR, we introduce:

- deprecation notices for the feature in the Jetpack plugin.
- logic to automatically disable the module when necessary.
- updated logic to detect whether the Lazy Images feature should be disabled. On top of the logic introduced in #33039, we now also check for future versions of WordPress (6.4) as well as sites running the development version of the Gutenberg plugin.

### Jetpack dashboard (Jetpack > Settings > Performance).

1. Always display a notice mentioning that the feature is no longer necessary, and explaining that it will be removed in November.
2. Do not allow enabling the module when it is disabled.
3. If you run a recent version of the Gutenberg plugin, customize the notice contents to explain that the feature is force-disabled now to avoid conflicts (see #33039).

Screenshot 2023-09-20 at 15 56 22
Screenshot 2023-09-20 at 15 56 08
Screenshot 2023-09-20 at 15 52 31

### Old Module list in the Jetpack plugin

Let's not make it available in the old module list **when it is disabled**. It does not need to be re-enabled.

### Jetpack Boost plugin

When the Lazy Images feature is flagged as conflicting with WordPress, we:

1. disable the feature
2. mark it as unavailable. It seems that right now, a feature that is marked as unavailable is just removed from the UI.

> **Warning**
> This means we'll need to make changes to the readme.txt file for the Boost plugin sooner rather than later. But I believe this will be tackled as part of #33170.

### Other information:

## Jetpack product discussion

- #33049
- pdWQjU-wV-p2

* Yes, it introduces a new tracked link when one wants more information about the feature deprecation.

## Testing instructions:

**With the Jetpack plugin**

* Start on a site that **does not** run this branch.
* Enable Lazy Images
* Switch to this branch
* Notice the new notice in Jetpack > Settings > Performance.
* Enable the Gutenberg plugin
* Notice that the notice contents change.
* Notice that you cannot toggle the Lazy Images feature anymore.
* Load any post on the frontend.
* Back to Jetpack > Settings > Performance, the feature is now inactive.
* You should not see the Lazy Images module as available.

**With the Jetpack Boost plugin**

* Now enable the Boost plugin
* You should not see the Lazy Images functionality in the plugin UI.

You can also start with a brand new site with just the Boost plugin:

* Enable Boost on a site that runs WordPress 6.3 without Gutenberg.
* You should see the Lazy Images feature in Jetpack > Boost
* Now add the Gutenberg plugin
* The Lazy Images feature should now be gone.


jeherve:
@kraftbj Before I open this up for reviews, I'd like to have your opinion on the overall UX of this change, as our temporary phase until the feature is fully removed.

Thank you!

Along this change, I would:

1. Add a JITM displayed to everyone with the module currently active. #33167
2. Update the support doc to explain the deprecation, since we're directing folks there (maybe I should direct folks to a specific anchor on the page, I thought about that, but maybe it's simpler to add a big notice at the top of the support doc) #33168
3. Update the Calypso UI to match this. #33169

kraftbj:
> maybe it's simpler to add a big notice at the top of the support doc)

Yeah, since this is a total deprecation and removal, no need to bury the notice in the text anywhere. Big and bold at the top is great.

Reviewing the UX itself now.

kraftbj:
Did not review the code so commenting, not reviewing.

Overall, I like this UX approach. It is pretty clean.

Things I noted:
* `wp jetpack module *` functions still list and let you activate lazy-images. I'm okay with this since it provides a way for folks to bring it back for whatever reason, but isn't something a regular person would stumble upon. Noting it for the sake of completeness.
* When using the Gutenberg force, the module itself is still active and displayed as such in the old modules list and the CLI. This isn't a big deal IMO, but would be nice to resolve.
* Bigger deal when using the Gutenberg force, since the module itself is still active, the Boost interface still shows Lazy Images as active. It won't load, of course, but that's confusing that the Jetpack regular interface and Boost interface show conflicting information.

Thus, my main suggestion would be:
* when using Gutenberg, let's on is_admin or within an admin hook, disable the module as well so we can have a consistent disabled experience for both states that would hit both plugins.

The lack of information in the Boost plugin is also a gap, but I know we're expecting HoG to handle the Boost-side of the deprecation.

jeherve:
@kraftbj Thanks for taking a look! I took a pass at addressing your feedback:

- We now deactivate the Jetpack module from the Lazy Images package if we're in a conflicting environment.
- In the Jetpack Boost plugin, if we're in a conflicting environment ~we deactivate the Jetpack module if necessary and~ (not anymore, see https://github.com/Automattic/jetpack/pull/33208#discussion_r1333910076) we mark the Boost feature as unavailable. Boost modules include a `is_available` that seems to be good for that purpose, although when a module is not available it is completely removed from the UI. I don't think it's ideal, but I don't think that's something we should change in this PR. I think I'll need a review from the Heart of Gold team on those changes, since I want to be sure it's really okay to hook those added behaviors to `is_available`.

Let me know what you think of the UX now.

projects/plugins/boost/app/modules/optimizations/lazy-images/class-lazy-images.php
```diff
@@ -16,6 +16,13 @@ public static function get_slug() {
 	}
 
 	public static function is_available() {
+		// In some scenarios, Lazy Images can conflict with other WordPress features.
+		// If we detect that, the feature should not be available.
+		if ( Jetpack_Lazy_Images::should_force_deactivate() ) {
+			Jetpack_Lazy_Images::disable();
```
haqadn, line 22: My only recommendation is to avoid disabling the module here. This function is for checking the availability

haqadn: I would recommend that you avoid disabling the module here. This function is for checking the availability and disabling the module would be a side effect. Besides, it would disable the module in the Jetpack plugin.

In practice, returning false from this function if force deactivation is needed should be enough for it to work.

haqadn (commented):
Thanks for taking the time to update Boost. We will work on improving the UX and messaging on the Boost side so it doesn't feel like the feature is missing for no reason. However, marking this as unavailable is still suitable to avoid a conflict.

projects/plugins/boost/app/modules/optimizations/lazy-images/class-lazy-images.php
jeherve: 👍 Removed that in 72fc3ec3779fe0478ca373629a02a481a598e4c0

projects/packages/lazy-images/tests/php/test_class.lazy-images.php
```diff
@@ -604,19 +604,58 @@ public function add_skip_lazy_class_to_attributes( $attributes ) {
...
+		$previous_version = $wp_version;
+
+		Constants::set_constant( 'IS_GUTENBERG_PLUGIN', $version_details['gutenberg'] );
+		Constants::set_constant( 'GUTENBERG_VERSION', $version_details['gutenberg_version'] );
+		$wp_version = $version_details['wp']; // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
+
+		$this->assertSame( $expected, Jetpack_Lazy_Images::should_force_deactivate() );
+
 		Constants::clear_constants();
+		$wp_version = $previous_version; // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
```
anomiex: FYI, if the test fails these lines won't be reached to reset things. If you want to fix that, you could use a try-finally or move the resetting logic to a `tear_down()` method.

projects/packages/lazy-images/tests/php/test_class.lazy-images.php
```diff
@@ -604,19 +604,58 @@ public function add_skip_lazy_class_to_attributes( $attributes ) {
...
+			),
+			'WordPress 6.4'    => array(
+				array(
+					'wp'                => '6.4.0',
+					'gutenberg'         => false,
+					'gutenberg_version' => false,
+				),
+				true,
+			),
+		);
```
anomiex, line 677: No test covering the `GUTENBERG_DEVELOPMENT_MODE` code path?

projects/plugins/jetpack/_inc/lib/admin-pages/class-jetpack-redux-state-helper.php
```diff
@@ -241,6 +241,49 @@ public static function get_initial_state() {
...
+
+	/**
+	 * Get information about the Gutenberg plugin and its Interactivity API support.
+	 *
+	 * @see https://make.wordpress.org/core/tag/interactivity-api/
+	 *
+	 * @return array
+	 */
+	private static function get_gutenberg_initial_state() {
+		$initial_state = array(
```
anomiex: This is only used once, it seems to me we may as well move it down to line 281. But feel free to ignore this suggestion if you disagree.

projects/plugins/jetpack/changelog/update-lazy-images-ui-deprecation
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: compat
+
+Lazy Images: prepare feature for its deprecation, coming in November . TYou will be able to rely on Lazy loading features provided by WordPress itself.
```
anomiex:
```suggestion
Lazy Images: prepare feature for its deprecation, coming in November. You will be able to rely on Lazy loading features provided by WordPress itself.
```

projects/plugins/jetpack/_inc/lib/admin-pages/class-jetpack-redux-state-helper.php
jeherve: 👍 52e06639eb7687885e90de51a1ca585ba9f01fbe

projects/packages/lazy-images/tests/php/test_class.lazy-images.php
jeherve: 👍 Done in 4ae55d67024739a1d10087b720b546378e1ae973

projects/packages/lazy-images/tests/php/test_class.lazy-images.php
jeherve: That's a good call. 9f62b74d690d452c7eee9f1cfba8f920c3395e99

anomiex:
Testing results:

> * Enable the Gutenberg plugin
> * Notice that the notice contents change.
> * Disable the Lazy Images feature

Can't, it's already showing as disabled.

> * Notice that you cannot enable it back.
> * Go to Jetpack > Settings > Modules
> * You should not see the Lazy Images module as available.

When I first try this, it shows the correct message and shows the slider as disabled but the option is not fully greyed out. The module is still present in the modules page and shows there as being *enabled*. After I reload a blog post, that apparently updates the database and now the option is fully greyed out and is no longer present on the modules page.

If I skip installing Gutenberg and just turn off the slider, it fully disables and greys out after the turning-off finishes.

> * You should see the Lazy Images feature in Jetpack > Boost

I note that this allows for turning the feature on when it's off, even though the corresponding slider in Jetpack's settings is disabled.

jeherve:
> Can't, it's already showing as disabled.

That's a good point. I need to update my testing instructions. Will do that now.

> When I first try this, it shows the correct message and shows the slider as disabled but the option is not fully greyed out. The module is still present in the modules page and shows there as being enabled. After I reload a blog post, that apparently updates the database and now the option is fully greyed out and is no longer present on the modules page.

Right, it will typically deactivate itself when Lazy Images is next loaded. I'm okay with this trade-off, personally.

> I note that this allows for turning the feature on when it's off, even though the corresponding slider in Jetpack's settings is disabled.

Yes, I think that's something that the folks in the Heart of Gold team will handle on their end, in #33170. I'll let them figure out the necessary UI updates that need to happen in the Boost plugin.

anomiex (approved):
Sounds good to me.

PR #33212, opened 2023-09-20
Fix video upload bug

CGastrell:
Fix a bug in the upload file queue iterator.

Fixes https://github.com/Automattic/jpop-issues/issues/8470

## Proposed changes:
This PR simply adds a `continue` statement on the file upload iteration when a video file fails the space/allowance check. The bug has been hidden since a filter takes place preventing the situation, but the filter is being fixed as it was wrongly messing with the returned value and then this bug became more evident and started to be triggered.

### Other information:

## Jetpack product discussion
p1695148738679889-slack-C9W0QF6KA
D122202-code

No

## Testing instructions:
The bug is not noticeable in regular use of the module, applying the PR should result in no changes uploading videos and free trial allowance.  This change is to land after and be tested against D122202-code

Checkout the diff on your sandbox. Sandbox public-api.wordpress.com. Follow instructions on the diff.

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-video-upload-bugfix
+++ b/projects/plugins/jetpack/changelog/fix-video-upload-bugfix
@@ -1,0 +1,4 @@
+Significance: minor
+Type: bugfix
+
+Skip video file addition to upload queue if it fails the space/allowance check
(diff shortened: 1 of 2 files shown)


dhasilva (approved):
LGTM! Tested with the diff following the diff instructions. Well done!

PR #33213, opened 2023-09-20
Sitebot button styling

artpi:
Fix a few issues with the AI chat button styling

NO

## Testing instructions:
Insert the block on the site that previously had styling inconsistencies


dereksmart (approved):
Love a simple fix

PR #33224, opened 2023-09-20
CRM Automations UI: Update Steps Data Model

robertf4:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3332

## Proposed changes:

The data model for Steps was updated so that the Workflow object has a map of Step objects, and the `next_step` property of the Step was updated to be a string instead of a Step object.

This PR implements those changes on the client.

### Other information:

## Jetpack product discussion

pbhBOz-3Wf-p2

No

## Testing instructions:

There is no way to functionally test this, so review the code and run the tests.

### DIFF
--- a/projects/plugins/crm/changelog/crm-update-automation-steps-data-model
+++ b/projects/plugins/crm/changelog/crm-update-automation-steps-data-model
@@ -1,0 +1,5 @@
+Significance: patch
+Type: changed
+Comment: Change data model for Automations Steps
+
+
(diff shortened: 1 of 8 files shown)


coder-karen (dismissed):
All tests well for me. I'm adding an approval but leaving in needs review as I may be missing a lot of the finer details here, so it would be good to get more eyes.

As an aside (not related to the changes here so not a blocker), in `src/js/state/automations-admin/test/reducer.js` there are two blocks of code that look like they could be rewritten - those blocks are line 15 and 30:
```
expect( Object.keys( state ).map( key => Number( key ) ) ).toEqual( [
					workflowOne.id,
					workflowTwo.id,
				] );
```
And it looks like they could be changed to:
```
				expect( { ...state } ).toEqual( {
					[ workflowOne.id ]: workflowOne,
					[ workflowTwo.id ]: workflowTwo,
				} );
```
Using the spread operator - tests well for me.

coder-karen (approved):

cleacos:
Adding @kallehauge as a reviewer for a second review.

kallehauge (approved):
LGTM 👍

PR #33226, opened 2023-09-21
Update: blocks sharing discoverability

### LINKED ISSUE #33162: Sharing: make plugin available in editor, even when the module is not active yet
See Primary issue: #32401

andrii-lysenko:
Fixes #33162

## Proposed changes:

- Always register the Sharing (Sharedaddy) editor extension.

### Other information:

* Yes: views of the placeholder (jetpack_editor_sharing_enable)
## Testing instructions:

* Spin this branch locally, for example, connect via JT
* Go to Posts > Add New
* Click on the "Likes and Sharing" panel title.
* Click on the button.
* Verify that the module is enabled and working as expected


projects/plugins/jetpack/extensions/plugins/sharing/sharing.php
```diff
@@ -0,0 +1,43 @@
...
+ * Register Sharing plugin.
+ *
+ * @return void
+ */
+function register_plugins() {
+	// Connection check.
+	if (
+		( new Host() )->is_wpcom_simple()
+		|| ( ( new Connection_Manager( 'jetpack' ) )->has_connected_owner() && ! ( new Status() )->is_offline_mode() )
+	) {
```
jeherve: I don't know if we need this here; the sharing buttons can be used without a connection to WordPress.com, in offline mode.

projects/plugins/jetpack/extensions/plugins/sharing/sharing.php
```diff
@@ -0,0 +1,43 @@
...
+add_action( 'jetpack_register_gutenberg_extensions', __NAMESPACE__ . '\register_plugins' );
+
+/**
+ * Allow display the Sharing UI in the editor.
+ */
+add_action(
+	'init',
+	function () {
+		if ( ! \Jetpack::is_module_active( 'sharedaddy' ) ) {
+			add_post_type_support( 'post', 'jetpack-sharing-buttons' );
```
jeherve: Should we support more post types here, like we do in the sharing module?
https://github.com/Automattic/jetpack/blob/0f1eace8d38a3ae9d8286f27df75b37133b29bdd/projects/plugins/jetpack/modules/sharedaddy/sharing.php#L753-L754

projects/plugins/jetpack/extensions/plugins/sharing/sharing.php
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @return void
+ */
+function register_plugins() {
+	// Connection check.
+	if (
+		( new Host() )->is_wpcom_simple()
+		|| ( ( new Connection_Manager( 'jetpack' ) )->has_connected_owner() && ! ( new Status() )->is_offline_mode() )
+	) {
+		// Register Publicize.
```
jeherve:
```suggestion
		// Register Sharing.
```

projects/plugins/jetpack/extensions/plugins/sharing/sharing.php
andrii-lysenko: Thanks!

projects/plugins/jetpack/extensions/plugins/sharing/sharing.php
andrii-lysenko: Oh, thanks, I missed that.

projects/plugins/jetpack/extensions/plugins/sharing/components/placeholder.scss
```diff
@@ -0,0 +1,3 @@
+.components-placeholder__learn-more {
```
jeherve: Just like I suggested here, should we switch our implementation a bit?
https://github.com/Automattic/jetpack/pull/33258#discussion_r1334181111

projects/plugins/jetpack/extensions/plugins/sharing/sharing.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+function register_plugins() {
+	// Register Sharing.
+	Jetpack_Gutenberg::set_extension_available( 'sharing' );
+}
+
+add_action( 'jetpack_register_gutenberg_extensions', __NAMESPACE__ . '\register_plugins' );
+
+/**
+ * Allow display the Sharing UI in the editor.
```
jeherve: Maybe it would be worth extending that comment a bit to explain why we need this in relation with `PostTypeSupportCheck`?

```suggestion
 * The Sharing panel is only displayed for post types that support sharing.
 * The sharing module declares support for sharing for all the public post types.
 * Let's do the same thing when the module isn't active yet.
```

projects/plugins/jetpack/extensions/plugins/sharing/components/placeholder.js
```diff
@@ -0,0 +1,41 @@
...
+			<Button
+				disabled={ isModuleActive || isLoading }
+				isBusy={ isLoading }
+				onClick={ enableSharingModule }
+				variant="secondary"
+			>
+				{ isLoading
+					? __( 'Activating Jetpack Sharing', 'jetpack' )
+					: __(
+							'Activate Jetpack Sharing',
```
jeherve: I wonder if we need the "Jetpack" prefix here? I think we market this feature as just "sharing"?

```suggestion
					? __( 'Activating Sharing', 'jetpack' )
					: __(
							'Activate Sharing',
```

jeherve (changes requested):

projects/plugins/jetpack/extensions/plugins/sharing/sharing.php
```diff
@@ -0,0 +1,40 @@
...
+
+add_action( 'jetpack_register_gutenberg_extensions', __NAMESPACE__ . '\register_plugins' );
+
+/**
+ * The Sharing panel is only displayed for post types that support sharing.
+ * The sharing module declares support for sharing for all the public post types.
+ * Let's do the same thing when the module isn't active yet.
+ */
+add_action(
+	'init',
```
jeherve: Should we / can we hook into `rest_api_init` here, like it's done here?
https://github.com/Automattic/jetpack/blob/0f1eace8d38a3ae9d8286f27df75b37133b29bdd/projects/plugins/jetpack/modules/sharedaddy/sharing.php#L777

jeherve (commented):
It tests well. I only have one tiny remark.

projects/plugins/jetpack/extensions/plugins/sharing/sharing.php
andrii-lysenko: Hmm, I've just thought of it. I'm not as experienced in this codebase yet. But, is that correct that when Sharing Module is disabled, this like you mentioned `add_action( 'rest_api_init', 'jetpack_post_sharing_register_rest_field' );` won't actually be run. Thus, until we enable the module, we actually don't have support for these? Maybe we shouldn't mark them as supported `add_post_type_support( $post_type, 'jetpack-sharing-buttons' );` by then? Or I'm getting it wrong?

andrii-lysenko: Oh, I think I see now why it was done in the first place. `<PostTypeSupportCheck supportKeys="{{module}}">` is basically providing a guard over rendering component. When we actually activate a module, that triggers component re-rendering , but, the module itself haven't activated everything, particularly, hasn't registered endpoints and marked `post type support`. I think ideally we would have to do `useModuleStatus` aware of actual module state and then we would be able to rely on it, and not registered `post type support` prior to Module activation itself

monsieur-z (commented):
This all looks good to me, but I've spotted a confusing behavior: after activating the feature, the `Show sharing buttons` is toggled off. However, when I hard refresh the page, it is toggled on. Is it that it should be on right after being activated, but the UI doesn't reflect that?

andrii-lysenko:
@monsieur-z Hmm, thanks for spotting this. So, there are definitely some store inconsistencies, which I also noticed within this draft: https://github.com/Automattic/jetpack/pull/33397. I'm not sure yet why this happened, maybe because of wordpress/data issues, or, there are also seem some problems that we load some packages twice, ending up creating several Stores. I'm wondering if we want to tackle it within this task, or it's not as big an issue for this particular task and we can address that as a separate task (I assume it might be quite complex). What do you think?
cc @jeherve

jeherve:
I think the problem is similar to the one you addressed when declaring sharing support for post types. Here, for the toggle to behave properly, this should be set:

https://github.com/Automattic/jetpack/blob/0f1eace8d38a3ae9d8286f27df75b37133b29bdd/projects/plugins/jetpack/modules/sharedaddy/sharing.php#L755-L766

So this can work:
https://github.com/Automattic/jetpack/blob/725a7c4c8a781cfbb70feadbebcd21df2506b841/projects/plugins/jetpack/extensions/plugins/sharing/components/show-sharing-control/index.js#L20-L23

andrii-lysenko:
@jeherve I see. Yes, this make sense, though still feel a bit tricky. I added these changes, and hope we would be able to address it in a different way later.

jeherve (approved):
This tests well for me. I added a bit of margin to improve the look of the 2 placeholders next to each other. It should be good to go now!

PR #33229, opened 2023-09-21
Add tracking to post publish share buttons

manzoorwanijk:
Related #33074

## Proposed changes:

* Add tracking to post publish sharing buttons

### Other information:

## Jetpack product discussion
pdrWKz-1b4-p2

## Testing instructions:

* On the post publish panel, confirm that you see the sharing buttons for Twitter and WhatsApp
* Open browser console 
* Run `localStorage.setItem('debug', 'dops:analytics');`
* Click on the sharing button
* Confirm that you see something like this in the console
```
dops:analytics - Super Props: {blog_id: 123456789}
dops:analytics Record event "jetpack_social_share_button_clicked" called with props {"network":"twitter","blog_id":123456789}
```
* Goto MC Live tracking
* Confirm that you see the event named `jetpack_social_share_button_clicked`

### DIFF
--- a/projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
+++ b/projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
@@ -1,10 +1,11 @@
 import { SocialServiceIcon } from '@automattic/jetpack-components';
 import { useAnalytics } from '@automattic/jetpack-shared-extension-utils';
 import { Button } from '@wordpress/components';
-import React, { useCallback } from 'react';
+import { useCallback } from '@wordpress/element';
 import { availableNetworks } from './available-networks';
 import styles from './styles.module.scss';
 import { usePrepareUrl } from './usePrepareUrl';
+import type React from 'react';
 
 export type ShareButtonsProps = {
 	buttonStyle?: 'icon' | 'text' | 'icon-text';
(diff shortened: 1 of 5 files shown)


projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
```diff
@@ -1,34 +1,11 @@
 import { SocialServiceIcon } from '@automattic/jetpack-components';
+import { useAnalytics } from '@automattic/jetpack-shared-extension-utils';
 import { Button } from '@wordpress/components';
-import React from 'react';
+import React, { useCallback } from 'react';
```
pablinos: This should come from `wordpress/elements` although I'm not sure that exposes the types. I think you can `import type` instead for that.

pablinos (commented):
Works well for me. Just a thought about the imports.

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
manzoorwanijk: We have imports from `react` everywhere and yes it doesn't import types. Since React is still loaded externally, using `import type` doesn't make any difference here.

pablinos: > We have imports from react everywhere

We do but we shouldn't. It's caused multiple versions of React to be imported in the past. Types are compile time dependencies, so importing specific types rather than `React` as a whole should give the bundler a helping hand. Thinking on this may have changed but we shouldn't be importing from React in the editor, even though it seems that's happening a fair amount!

manzoorwanijk: Updated. Thanks

manzoorwanijk: May be there should be ESLint rules to prevent non `type` imports from `react`.

pablinos (approved):
As discussed, let's iterate on thsi in #33261

PR #33145, opened 2023-09-18
Add tracking to WooCommerce my account page

senadir:
This PR does 2 things:

- It refactors universal class slightly and moves some of the shared logic to a trait.
- It adds tracking to my account page.

Closes https://github.com/woocommerce/woocommerce-blocks/issues/9730

## Proposed changes:

*

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [x] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Does this pull request change what data or activity we track or use?

This PR adds customer tracking to the my account page, it tracks what tabs they view and actions they do like adding payment methods, removing payment methods, updating addresses, and viewing orders. It doesn't track the actual content of address, payment method, or order, just the act of doing that.

### Regression testing:

- On a connected, live, WordPress account, check if existing tracks are being sent (either from the track page or using the Tracks Vigilante extension).
- View a product, add it to cart, change its quantity from cart, and remove it from cart. Make sure all of those events trigger fine.

### Tracking testing:

- On a connected, live, WordPress account, check if existing tracks are being sent (either from the track page or using the Tracks Vigilante extension).
- On the My Account page, test that:
- `woocommerceanalytics_my_account_tab_click` with prop `tab: logout` is triggered when you click log out.
- `woocommerceanalytics_my_account_page_view` with the prop `tab: $tabName` is being triggered on each top level tab you visit.
- `woocommerceanalytics_my_account_order_number_click` Is being triggered if you clicked an order number in the orders view.
- `woocommerceanalytics_my_account_order_action_click` is being triggered with prop `action: view` if you click the view button, and `action: pay` if you click the pay button, and `action: cancel` if you click the cancel button.
    - To see the pay and cancel buttons, change the order status to pending payment. 
- `woocommerceanalytics_my_account_address_click` with prop `address: billing | shipping` when you click on the button to edit an address.
- `woocommerceanalytics_my_account_address_save`  with prop `address: billing | shipping` when you save an address you edited.
- `woocommerceanalytics_my_account_details_save` When you click save on the Accounts Details page.
For those you will need a payment method installed like Stripe or WCPay:
- Check that:
- `woocommerceanalytics_my_account_payment_add` Is being triggered when you add a new payment method.
- `woocommerceanalytics_my_account_payment_save` Is being triggered when you save a payment method you're adding.
[...]

PR #33230, opened 2023-09-21
Fix static function call in WooCommerce Analytics

senadir:
This fixes the static call that get introduced by #33145 for a function that's no longer static.

## Testing instructions:

No testing needed, this should fix a warning about `Uncaught Error: Non-static method Jetpack_WooCommerce_Analytics_Universal::post_contains_text() cannot be called statically`

No

## Testing instructions

- On a connected, tracked website, run the following command on cli: `wp transient delete jetpack_woocommerce_analytics_cart_checkout_info_cache`
- Load the website on a live domain (jurassic or something connected) and add a product to cart.
- Make sure to error is thrown in the logs.

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-static-call-in-analytics
+++ b/projects/plugins/jetpack/changelog/fix-static-call-in-analytics
@@ -1,0 +1,5 @@
+Significance: patch
+Type: other
+Comment: Small typo fix
+
+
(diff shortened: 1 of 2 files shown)


zinigor (approved):
I was able to confirm that the PR fixes the problem, although I had to force the code to go into the affected logic branch. Otherwise the `if` condition makes it avoid the bug.

PR #33235, opened 2023-09-21
Subscriptions modal: enable for Jetpack sites

simison:
## Proposed changes:

* Enables Subscribe modal for Jetpack sites. Previously this was enabled just for .com sites.

### Other information:

## Testing instructions:

* Go to Settings -> Newsletter
* Enable modal
* Open a blog post as non logged in, non-subscribed user (i.e. incognito window would do)

### DIFF
--- a/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
+++ b/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
@@ -159,18 +159,6 @@ HTML;
 	}
 
 	/**
-	 * Returns true if we should load Newsletter content.
-	 *
-	 * @return bool
-	 */
-	public static function should_load_subscriber_modal() {
-		// Adding extra check/flag to load only on WP.com
-		// When ready for Jetpack release, remove this.
-		$is_wpcom = ( new Host() )->is_wpcom_platform();
-		return $is_wpcom;
-	}
-
-	/**
 	 * Returns true if a site visitor should see
 	 * the Subscribe Modal.
 	 *
@@ -228,14 +216,6 @@ HTML;
 	}
 }
 
-add_filter(
-	'jetpack_subscriptions_modal_enabled',
-	array(
-		'Jetpack_Subscribe_Modal',
-		'should_load_subscriber_modal',
-	)
-);
-
 /**
  * Filter for enabling or disabling the Jetpack Subscribe Modal
  * feature. We use this filter here and in several other places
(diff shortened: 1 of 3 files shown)


projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -244,9 +224,9 @@ public function has_subscription_cookie() {
  *
  * @since 12.4
  *
- * @param bool Defaults to false.
+ * @param bool Defaults to true.
  */
-if ( apply_filters( 'jetpack_subscriptions_modal_enabled', false ) ) {
+if ( apply_filters( 'jetpack_subscriptions_modal_enabled', true ) ) {
```
jeherve: Should we deprecate that filter in this PR, in a follow-up PR, or would you rather we keep it around for a bit longer?

simison: Any objections deprecating already in this PR?

What's the Jetpack way of deprecating a filter?

jeherve: I think we could remove the filter in this PR. 👍

Since it's unlikely that hook has been really used in the wild outside of Jetpack, I would just remove it, and add it to the list here just in case:
https://github.com/Automattic/jetpack/blob/d19049e08efdf0551d02825d552399b9d3b284d2/projects/plugins/jetpack/class.jetpack.php#L5848-L5851

simison: Moved the filter to deprecated hooks 👍  Thanks for the tip!

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -228,43 +215,11 @@ public function has_subscription_cookie() {
...
- * Filter for enabling or disabling the Jetpack Subscribe Modal
- * feature. We use this filter here and in several other places
- * to conditionally load options and functionality related to
- * this feature.
- *
- * @since 12.4
- *
- * @param bool Defaults to false.
- */
-if ( apply_filters( 'jetpack_subscriptions_modal_enabled', false ) ) {
```
lezama, line 249: We could keep the filter but default it to `true` in case site administrators want to disable/hide this feature?.

simison: Do we do that for other features? Feels like something we could add if there's demand, rather than leaving it in-case?

lezama (approved):

PR #33244, opened 2023-09-21
Add/one click sharing section

gmjuhasz:
Adds the previously added Quck-share panel to the editor too

## Proposed changes:

* Updates share-buttons a bit to match designs more
* Adds the new Dropdown to our panel

### Other information:

No.
## Testing instructions:

* Open up a new post, there should not be anything new.
* After publishing you should see the new Quick share Button if our panel is open
* Clicking the icon should open the Quick share dropdown
* Clicking on any of the icons there should work
* Clicking the learn more on the dropdown should open the help modal and close the dropdown.
* Test on mobile


projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
```diff
@@ -59,20 +59,24 @@ export function ShareButtons( { buttonStyle = 'icon', buttonVariant }: ShareButt
...
+							rel="noopener noreferrer"
+							onClick={ getOnClick( href ) }
+							data-network={ networkName }
+							className={ styles[ networkName ] }
+						>
+							{ 'text' === buttonStyle ? label : null }
+						</Button>
+						{ 'icon-text' === buttonStyle && (
+							<Text className={ styles.label }>Share on { label }</Text>
+						) }
```
manzoorwanijk: Why don't we make the whole button clickable here, instead of just the icon?

gmjuhasz: Good point we can make whole clickable.

gmjuhasz: @manzoorwanijk I just did that to match the designs
CleanShot 2023-09-27 at 12 17 30 png

manzoorwanijk: `Share on { label }` needs to be translated. I have done it in my commit.

projects/js-packages/publicize-components/src/components/share-buttons/styles.module.scss
```diff
@@ -32,3 +32,14 @@ $whatsapp: #2ab318;
 		color: var(--jp-black);
 	}
 }
+
+.container {
+	display: flex;
+}
+
+.label.label {
+	font-size: 15px;
```
pablinos: Looking at the designs, the text should be 13px to match the title of the panel. It is looking a bit big for me.

projects/js-packages/publicize-components/src/components/one-click-sharing-dropdown/index.js
```diff
@@ -0,0 +1,69 @@
...
+		/>
+	);
+};
+
+const OneClickSharingDropdown = ( { className: containerClass, onClickLearnMore } ) => {
+	const renderContent = useCallback(
+		( { onToggle } ) => {
+			return (
+				<div className="one-click-sharing-dropdown">
+					<ThemeProvider>
```
pablinos, line 21: Is this required? Could we rely on the editor extension, or similar, being wrapped in this component?

projects/js-packages/publicize-components/src/components/one-click-sharing-dropdown/index.js
```diff
@@ -0,0 +1,69 @@
...
+						<ShareButtons buttonStyle="icon-text" />
+						<hr className={ styles.divider } />
+						<Text className={ styles.description } variant="body-small">
+							{ __( 'Share with a Single Click!', 'jetpack' ) +
+								'✨ ' +
```
pablinos: I'm not sure about this. I know it's decoration, but we're breaking the text up. Could we add a placeholder for it (like we would with a link or component) and a translator note? I'm a bit concerned that it wouldn't make sense to be placed here in certain languages

projects/js-packages/publicize-components/src/components/panel/styles.module.scss
```diff
@@ -0,0 +1,10 @@
+.one-click-share-dropdown {
+	cursor: pointer;
+	position: absolute;
+	top: 13px;
+	right: 50px;
```
pablinos: This worried me a bit! But I've played around with zooming text etc, and stays in the right position. Is there anything else this could cause problems with? RTL languages for instance?

projects/js-packages/publicize-components/src/components/share-buttons/styles.module.scss
```diff
@@ -32,3 +32,14 @@ $whatsapp: #2ab318;
...
+	font-weight: 400;
+	line-height: 40px;
```
pablinos: Similarly, the icon size in the design is about 24px, and the line height of ends up being 36px, with margins/padding after setting this to 24px also.

Although fiddling about with that, I think it does make the icons quite small, but does feel better than the current size

image

projects/js-packages/publicize-components/src/components/one-click-sharing-dropdown/index.js
```diff
@@ -0,0 +1,69 @@
...
+import { useCallback } from 'react';
+import { ShareButtons } from '../share-buttons/share-buttons';
+import shareArrowIcon from './share-arrow-icon';
+import styles from './styles.module.scss';
+
+const renderToggle = ( { isOpen, onToggle } ) => {
+	return (
+		<Icon
+			className={ styles.icon }
+			size={ 18 }
```
pablinos: I think the SVG is 16px. Should this be 16 too?

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
```diff
@@ -59,29 +60,44 @@ export function ShareButtons( { buttonStyle = 'icon', buttonVariant }: ShareButt
...
-						target="_blank"
-						rel="noopener noreferrer"
-						onClick={ getOnClick( href, { network: networkName } ) }
-						data-network={ networkName }
-						className={ styles[ networkName ] }
-					>
-						{ 'icon' !== buttonStyle ? label : null }
-					</Button>
+					<div className={ styles.container } key={ networkName }>
+						<Button
```
pablinos, line 64: It would be great if we could make the whole row clickable, rather than just the button. What do you think about adding an :after element to the button that fills the div? So clicking anywhere on the div clicks the button.

There's probably better ways of doing it! I'm CSSing like it's 1999

pablinos (commented):
This is working well for me. A few thoughts, but it's pretty much there.

projects/js-packages/publicize-components/src/components/one-click-sharing-dropdown/index.js
manzoorwanijk: It is not part of the translated string. So I think it's better this way.

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
manzoorwanijk: Yes, I think we should make the whole row clickable but that will require quite a bit of changes given the way icon components are designed. It will need icon styling to be separate instead of the button component handling it for us.

manzoorwanijk: Done, please check now

projects/js-packages/publicize-components/src/components/one-click-sharing-dropdown/index.js
gmjuhasz: It is. We could move this into an outer component later, but without this the styles are not coming in

projects/js-packages/publicize-components/src/components/panel/styles.module.scss
gmjuhasz: I tried RTL with a chrome addon and looked okay for me.

projects/js-packages/publicize-components/src/components/one-click-sharing-dropdown/share-arrow-icon.tsx
```diff
@@ -0,0 +1,17 @@
+/**
+ * WordPress dependencies
+ */
+import { SVG, Path } from '@wordpress/components';
+
+const shareArrowIcon = (
```
manzoorwanijk: This can goto `js-packages/components/components/icons`

projects/js-packages/publicize-components/src/components/panel/styles.module.scss
manzoorwanijk: To make it [RTL aware](https://www.w3.org/TR/css-logical-1/#inset-properties), you can use

```suggestion
	position: absolute;
	inset-block-start: 13px;
	inset-inline-end: 50px;
```

projects/js-packages/publicize-components/src/components/one-click-sharing-dropdown/index.js
```diff
@@ -0,0 +1,69 @@
...
+							<Button
+								variant="link"
+								// eslint-disable-next-line react/jsx-no-bind
+								onClick={ () => {
+									onClickLearnMore();
+									// Close the dropdown
+									onToggle( false );
+								} }
+							>
+								{ __( 'Learn more..', 'jetpack' ) }
```
manzoorwanijk: 3 dots

```suggestion
								{ __( 'Learn more...', 'jetpack' ) }
```

projects/js-packages/publicize-components/src/components/one-click-sharing-dropdown/index.js
pablinos: Strange! I feel like we're using this in multiple places then and we'd be better off wrapping the whole sidebare in it. It must be used for the connection toggles for example. Something to look at.

projects/js-packages/publicize-components/src/components/one-click-sharing-dropdown/index.js
pablinos: It's not part of the translated string because we've broken it up. This has a unicode character with a trailing space, which make is like a piece of punctuation. I feel like it would be better to have one translated string with a placeholder for the problematic character, which we could replace and also provide a translator note, so that they know what it is.

projects/js-packages/publicize-components/src/components/panel/styles.module.scss
pablinos: Good idea! Is there enough browser support for this? I think It goes back far enough looking at [caniuse](https://caniuse.com/css-logical-props).

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
pablinos: Yep that works. Do you think we should close the popover when a link it clicked? At the moment we end up with an outline, as the element has focus. Or we could make the focus outline work a bit better

image

projects/js-packages/publicize-components/src/components/share-buttons/styles.module.scss
pablinos: Addressed in 94fc991b86

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
pablinos: I've noticed we haven't got a border on the Copy to clipboard button either. Is that intentional?

gmjuhasz: Fixed

projects/js-packages/publicize-components/src/components/one-click-sharing-dropdown/share-arrow-icon.tsx
gmjuhasz: Done

projects/js-packages/publicize-components/src/components/one-click-sharing-dropdown/index.js
gmjuhasz: Done

pablinos (approved):
This works well for me!

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
```diff
@@ -56,32 +57,57 @@ export function ShareButtons( { buttonStyle = 'icon', buttonVariant }: ShareButt
...
 				return (
-					<Button
-						key={ networkName }
-						icon={ icon }
-						variant={ buttonVariant }
-						aria-label={ label }
-						href={ href }
-						target="_blank"
-						rel="noopener noreferrer"
-						onClick={ getOnClick( href, { network: networkName } ) }
```
manzoorwanijk, line 70: This change of removing `, { network: networkName }` from the function call might have caused the tracking to stop working.

PR #33246, opened 2023-09-21
Locked Mode: Apply cap filter in API requests

obenland:
On API requests we've not switched to the correct site yet, when we look for feature and option on plugins_loaded.

## Proposed changes:

* Adds an action to run the cap filter determination once we switched to the correct site on API requests.

### Other information:

No

## Testing instructions:

* Apply this patch to your sandbox and sandbox public-api.
* Add a 100-year plan to a test site.
* Navigate to /settings/general and activate Locked Mode.
* Make sure the navigation bar updates and content menu items are removed.


DustyReagan (approved):
Works well! 👍

PR #3337, opened 2016-01-22
Merge/top posts jan2016

dereksmart:
Brings the file up to date.  

The wpcom dependent code is ugly, but necessary.  This way we're actually able to share the widget with dotcom instead of maintaining two separate classes over there.  

see phab diff `D886` and commit r130076-wpcom for more info

PR #33254, opened 2023-09-21
CRM: Resolves #3337 - add invoice status and mapping settings migration

tbradsha:
CRM: Resolves Automattic/zero-bs-crm#3337 - add invoice status and mapping settings migration

## Proposed changes:

In #32909 I made invoice statuses store in English, which fixed several inconsistencies. At the time, I opted out of doing a migration, as it was possible to do a full Woo resync as needed. However, @cleacos pointed out that the mapping settings could be lost as well, so this PR introduces a migration.

It makes changes if:
* the current locale is not `en_US`
* the translated text is not the same as English
* it finds a translated match in invoice mapping statuses and/or invoice statuses

As I was leery of changing WooSync settings, I also store a backup before making the change.

I also spotted a few other translation glitches:

* A new invoice would start with a status of `draft` (lowercase), so it didn't match in the dropdown. I changed this to `Draft`.
* In the left corner, there's some CSS based on invoice status that expects a class matching the lowercase name of the status. I added the conversion when outputting the class.
* As of the new PR, new invoices didn't have the new `status_label` in the default object, so would show undefined in the left corner. I corrected this.

### Other information:

## Testing instructions:

1. Switch WP to Spanish: `/wp-admin/options-general.php`
2. Add a few invoices: `/wp-admin/admin.php?page=zbs-add-edit&action=edit&zbstype=invoice`
3. Change and save mapping settings: `/wp-admin/admin.php?page=zerobscrm-plugin-settings&tab=woosync`
4. In the database, update the invoice statuses to `Pagado` or some other valid Spanish status.
5. In the database, update the mapping in the `zbscrm_dmz_ext_woosync` settings key to some valid Spanish status.

Note that steps 4 and 5 represent how WooSync would store things in JPCRM 6.1.

In `trunk`:
* Editing the invoice will show correctly, but only because of a fallback to handle unknown statuses. The status dropdown will show two of the same Spanish status, with the last one (fallback) selected.
* The invoice mappings saved in Spanish will not be selected.

In the `fix/crm/3337-invoice_status_migration` branch (you may need to manually run the `zeroBSCRM_migration_invoice_language_fixes` migration if you made the changes after the migration initially ran, as migrations only run automatically once):
* The invoice status is switched to English in the database, and the status select is working as expected.
* The invoice mappings are updated to English in the database, and the mapping selects are matched as expected.

### DIFF
--- a/projects/plugins/crm/changelog/fix-crm-3337-invoice_status_migration
+++ b/projects/plugins/crm/changelog/fix-crm-3337-invoice_status_migration
@@ -1,0 +1,5 @@
+Significance: patch
+Type: fixed
+Comment: Adds migration for invoice statuses and mapping settings
+
+
--- a/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Helpers.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Helpers.php
@@ -4743,7 +4743,8 @@ function zeroBSCRM_get_invoice_defaults( $obj_id = -1 ) {
 	}
 
 	$defaults = array(
-		'status'                    => 'draft',
+		'status'                    => 'Draft',
+		'status_label'              => __( 'Draft', 'zero-bs-crm' ),
 		'new_invoice'               => true,
 		'id'                        => $obj_id,
 		'invoice_items'             => array(),
--- a/projects/plugins/crm/includes/ZeroBSCRM.Migrations.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.Migrations.php
@@ -39,6 +39,7 @@ global $zeroBSCRM_migrations; $zeroBSCRM_migrations = array(
 	'refresh_user_roles', // Refresh user roles
 	'regenerate_tag_slugs', // Regenerate tag slugs
 	'create_workflows_table', // Create "workflows" table.
+	'invoice_language_fixes', // Store invoice statuses and mappings consistently
 	);
 
 global $zeroBSCRM_migrations_requirements; $zeroBSCRM_migrations_requirements = array(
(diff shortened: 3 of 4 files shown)


coder-karen (commented):
Following the testing instructions on a Jurassic Ninja site, after running the migration again I  can confirm the status and mappings in the database showed in English, and the status dropdown works (also no duplicated statuses).
I then switched to trunk afterwards, and the mappings seemed to all be removed. The status dropdown showed the duplicated statuses as explained.
I then went back to the branch again after adding mapping statuses back in, and these were then removed:
Screenshot 2023-09-25 at 09 08 36

The status dropdowns still showed the duplicated statuses but the correct one was selected.
Screenshot 2023-09-25 at 09 08 06

Is the expectation that moving back to trunk and then forward again to the branch (or as it will be back to an older CRM version then forward to a newer version) should result in correct mappings / no duplicate statuses?

tbradsha:
> Is the expectation that moving back to trunk

The migration system is (unfortunately) designed to be one-way, so the behaviour you're seeing when switching back to `trunk` or the 6.1 release is expected (statuses aren't mapped and the fallback invoice status is selected).

>  and then forward again to the branch

As long as you don't save anything or rerun the migration on the old version, moving back to the branch should show everything correctly. I'm not able to reproduce any issues on my end when going to `trunk` then back to the branch...are there other steps you took?

If you did save things on `trunk` or an old version, you'd need to rerun the migration upon switching to this branch.

> (or as it will be back to an older CRM version then forward to a newer version) should result in correct mappings / no duplicate statuses?

Technically `trunk` is an interim unreleased state of things, as we changed the way things match up in #32909. But same as above.

cleacos (approved):
It works as expected for me.

- Statuses are translated.
- The WooSync mapping shows the correct status in Spanish.
- The invoice list shows the correct status label.
- All the dropdown list shows the correct label (Spanish) and the name (English).
- The migration runs okay and it doesn't break anything.

LGTM!

coder-karen:
> The migration system is (unfortunately) designed to be one-way, so the behaviour you're seeing when switching back to trunk or the 6.1 release is expected (statuses aren't mapped and the fallback invoice status is selected).

Makes sense, thanks.

> As long as you don't save anything or rerun the migration on the old version

Yes that was what I did, so then your explanation would make sense here too.

🚢

PR #33255, opened 2023-09-21
Admin Menu: Make Home & Stats visible in Locked Mode

obenland:
For sites on 100-Year plans that have Locked Mode enabled, the Home and Stats menu item are currently removed because we disable the `edit_posts` capability.

On Simple sites, all available roles have `edit_posts` and `read` capabilities, so there is no difference in functionality in using `read` over `edit_posts` for the My Home and Stats menu items.

On WoA sites, the My Home menu item already uses the `read` capability. For stats we can use Jetpack's `view_stats` capability and filter the default options for the Stats menu to maintain access for all roles.

## Proposed changes:

* Update Stats menu item capability on WoA sites to `view_stats` to re-use Jetpack capability for that purpose.
* Add a default option callback to maintain access to Stats for all roles.
* Update My Home menu capability on Simple Sites to `read` to bring in line with Core.
* Update Stats menu item capability on Simple Sites to `read` so it doesn't get hidden when we remove the `edit_posts` capability.

### Other information:

No

## Testing instructions:

* Apply the diff to your sandbox and sandbox public-api.
* Apply the diff to a WoA site using Jetpack Beta.
* Test admin menu with simple and WoA sites with admin and contributor roles.
* Make sure My Home and Stats are always accessible.
* Upgrade a test site to the 100-year plan.
* In Settings General, enable Locked Mode.
* Make sure My Home and Stats are always accessible.


mmtr (dismissed):
> Changing Home from edit_posts to read will enable the Calypso My Home screeen for all users that currently have a Subscriber role. I'm not sure how big of a problem that is, but we should be at least aware of it.

I was concerned a little bit about this at first, but seems this is harmless since Subscribers don't have access to sites they are subscribed to from the "My Site(s)" menu.

> Changing Stats from edit_posts to view_stats will remove the Stats menu item for Editors, Authors, and Contributors

To clarify, this is only for Atomic sites, right? According to the code (and my tests) users of Simple sites will still have access to Stats as long as they have the `read` capability.

Which I think is the correct behavior, because Simple Sites don't have ([the ability to choose which roles can see stats](https://wordpress.com/support/stats/#authors)) so we need to ensure that Editors, Authors, and Contributors of Simple sites can continue accessing Stats.

obenland:
Yes, apologies, I've not had a chance to update the issue description. 

There are no subscriber-level roles on Simple, so every one has `edit_posts` AND `read`, so they can be used "interchangeably".

> Simple Sites don't have ([the ability to choose which roles can see stats](https://wordpress.com/support/stats/#authors))

It took me a while to figure that out on Friday. So yeah, with `read` there should be no changes on Simple. For WoA I think I'd like to at least add a filter to the default option value for Stats to not limit it to Admins only. Not sure if that's possible, but that'll be my next step on this.

projects/plugins/jetpack/modules/masterbar/admin-menu/class-atomic-admin-menu.php
```diff
@@ -444,4 +445,18 @@ public function wp_ajax_jitm_dismiss() {
 		}
 		wp_die();
 	}
+
+	/**
+	 * Extends the default value for which roles can access the Stats menu item.
+	 *
+	 * @see Automattic\Jetpack\Stats\Options::get_default_value()
```
mmtr: `get_default_value` does not exist in `Automattic\Jetpack\Stats\Options`. Maybe you meant `Automattic\Jetpack\Stats\Options::get_options()`?

See https://github.com/Automattic/jetpack/blob/f58c02baffc449b0bebd70421c8234e4740cea01/projects/packages/stats/src/class-options.php#L42

mmtr (commented):
Nice, this works as expected.

> Add a default option callback to maintain access to Stats for all roles.

This has the side effect of showing the Jetpack Stats section in Jetpack > Dashboard (`/wp-admin/admin.php?page=jetpack#/dashboard`) to contributors, authors, and editors, which was previously hidden to them:

Before | After
--- | ---
Screenshot 2023-09-26 at 13 52 17 | Screenshot 2023-09-26 at 13 53 02

I guess it sort of makes sense, since these users have access to the stats data anyway via  Calypso (`/stats/day/:site`), but I wonder if the Nav Unification logic is the best place to change the default roles since it's not immediately obvious that this also affects the Jetpack Dashboard.

projects/plugins/jetpack/modules/masterbar/admin-menu/class-atomic-admin-menu.php
obenland: It's `get_defaults()`! Gah, Copilot!

https://github.com/Automattic/jetpack/blob/f58c02baffc449b0bebd70421c8234e4740cea01/projects/packages/stats/src/class-options.php#L156

obenland:
> This has the side effect of showing the Jetpack Stats section in Jetpack > Dashboard

Nice catch! Agreed, I think of it more as fixing a bug than an unintended side effect. The bar graph in the sidebar now works, too!

obenland:
> I wonder if the Nav Unification logic is the best place to change the default roles since it's not immediately obvious that this also affects the Jetpack Dashboard.

I kept in the class because that's where we're using the capability. Would wpcomsh be a better place?

mmtr:
> Would wpcomsh be a better place?

I think so, yeah.

mmtr (approved):
Works great when paired with https://github.com/Automattic/wpcomsh/pull/1509

obenland:
@anomiex Is it okay to merge with these tests failing? Any idea how this diff would effect those?

obenland:
Hm, looks like they're failing on trunk, too

anomiex:
Yes, you should be ok. Details on the PHP test failure are in pdWQjU-zp-p2. The other one appears to be an issue with Actions rolling out an update from node 18.17.1 to 18.18.0 which apparently breaks things, I'm working on a workaround for that now.

PR #33258, opened 2023-09-21
Update/blocks seo tools discoverability

### LINKED ISSUE #33161: SEO Tools: make plugin available in editor, even when the module is not active yet
See Primary issue: #32401

andrii-lysenko:
Fixes #33161

## Proposed changes:

- Always register the Seo Tools editor extension.

### Other information:

* Yes: views of the placeholder (jetpack_editor_seo_enable)
## Testing instructions:

* Spin this branch locally, for example, connect via JT
* Go to Posts > Add New
* Click on the "SEO" panel title.
* Click on the button.
* Verify that the module is enabled and working as expected


projects/plugins/jetpack/extensions/blocks/seo/components/placeholder.js
```diff
@@ -0,0 +1,37 @@
...
+	const { tracks } = useAnalytics();
+
+	const enableSeoModule = () => {
+		tracks.recordEvent( 'jetpack_editor_seo_enable' );
+		return changeStatus( true );
+	};
+
+	return (
+		<>
+			<p>{ __( 'Activate Seo feature to optimize your site for search engines.', 'jetpack' ) }</p>
```
jeherve:
```suggestion
			<p>{ __( 'Activate the SEO feature and start optimizing your posts for search engines.', 'jetpack' ) }</p>
```

projects/plugins/jetpack/extensions/blocks/seo/components/placeholder.js
```diff
@@ -0,0 +1,37 @@
...
+			<Button
+				disabled={ isModuleActive || isLoading }
+				isBusy={ isLoading }
+				onClick={ enableSeoModule }
+				variant="secondary"
+			>
+				{ isLoading
+					? __( 'Activating Jetpack Seo', 'jetpack' )
```
jeherve:
```suggestion
					? __( 'Activating Jetpack SEO', 'jetpack' )
```

projects/plugins/jetpack/extensions/blocks/seo/components/placeholder.js
```diff
@@ -0,0 +1,37 @@
...
+					: __( 'Activate Jetpack Seo', 'jetpack', 0 ) }
```
jeherve:
```suggestion
					: __( 'Activate Jetpack SEO', 'jetpack', 0 ) }
```

projects/plugins/jetpack/extensions/blocks/seo/components/placeholder.js
```diff
@@ -0,0 +1,37 @@
...
+			</Button>
+
+			<div className="components-placeholder__learn-more">
+				<ExternalLink href={ getRedirectUrl( 'jetpack-support-seo-tools' ) }>
+					{ __( 'Learn more about Jetpack Seo.', 'jetpack' ) }
```
jeherve:
```suggestion
					{ __( 'Learn more about Jetpack SEO.', 'jetpack' ) }
```

projects/plugins/jetpack/extensions/blocks/seo/components/placeholder.scss
```diff
@@ -0,0 +1,3 @@
+.components-placeholder__learn-more {
```
jeherve: Should this be more specific to avoid targetting others containers with the same class in the editor? 

Or maybe we should go the opposite, and start using a common, Jetpack-unique class for all those placeholders, and then style it all from one place?

jeherve (changes requested):

projects/plugins/jetpack/extensions/blocks/seo/components/placeholder.scss
andrii-lysenko: Yes, good point. I think for now I will update the naming to be more specific. 
But after we finish up with all the blocks, I think we should create a generic component and reuse it.

projects/plugins/jetpack/extensions/blocks/seo/seo.php
```diff
@@ -0,0 +1,22 @@
...
+namespace Automattic\Jetpack\Extensions\Seo;
+
+use Jetpack_Gutenberg;
+
+/**
+ * Register SEO plugin.
+ *
+ * @return void
+ */
+function register_plugins() {
```
jeherve, line 20: The SEO Tools module requires a working connection to WordPress.com:
https://github.com/Automattic/jetpack/blob/3a3c4adc1c0913bac224ddb832713bd6101fe1ab/projects/plugins/jetpack/modules/seo-tools.php#L8-L9

With that in mind, I think we should add a check like here:
https://github.com/Automattic/jetpack/blob/13262b644b6d7f90e6da48e752fcaa0e512327d5/projects/plugins/jetpack/extensions/plugins/publicize/publicize.php#L31-L36

This would avoid issues when you're trying to access the SEO Panel when in offline module, for example.

jeherve (changes requested):

projects/plugins/jetpack/extensions/blocks/seo/seo.php
andrii-lysenko: Oh, I actually missed where to get this prerequisite data. I will know for the future. Thanks!

monsieur-z (dismissed):
The PR tested as instructed, and the code looks good.

A small detail: the SEO panel is open by default. Is that expected?
Screenshot 2023-09-28 at 8 47 27 AM

andrii-lysenko:
@monsieur-z this actually also confusing for me but this behavior was also present in the past. I think we would need to find the right person to ask and change that if needed in the future.

jeherve:
> the SEO panel is open by default. Is that expected?

That's a good point. I think we could update `<PanelBody />` props to only default to open when the module is active. I believe it should be open, but not when the feature is not active to not add any unnecessary noise.

andrii-lysenko:
@jeherve Thanks for idea, for some reason haven't thought of it at first. Updated as suggested

jeherve (approved):
This tests well for me. 🚢

PR #33259, opened 2023-09-21
SSO: add force 2fa functionality

kraftbj:
Brings https://github.com/automattic/jetpack-force-2fa into Jetpack itself behind a `jetpack_force_2fa` flag. Defaulted to off.

Defaults to forcing SSO + 2fa for admins when SSO is enabled and the filter is returned true. The capability (manage_options) can be modified with the `jetpack_force_2fa_cap` filter.

## Proposed changes:

* Adds functionality for forcing SSO+2fa for indicated user capabilities.

### Other information:

## Jetpack product discussion

n/a - Came up when talking about possibilities with VIP for a client.

No change.

## Testing instructions:

* Jetpack connnected + SSO
* Connect an account that does not have 2fa enabled to the Jetpack site (either cycle the connection or make a new admin user connected to a non-2fa WP.com account.
* Create a new user with subscriber or contributor role.
* Log out and log back into admin account with regular WP creds (not SSO) This should work.
* Enable flag via `add_filter( 'jetpack_force_2fa', '__return_true' );`
* Log out and log back in with regular WP creds. It should fail.
* Log in with WP.com SSO with an account that has 2fa enabled. It should work.
* Log out and login with the non-2fa WP.com account via SSO. It should fail.
* Add a filter to modify the cap, e.g. `add_filter( 'jetpack_force_2fa_cap, function() { return 'read' } );`
* Verify that the contributor forces SSO.

### DIFF
--- a/projects/plugins/jetpack/changelog/add-force-2fa
+++ b/projects/plugins/jetpack/changelog/add-force-2fa
@@ -1,0 +1,4 @@
+Significance: minor
+Type: enhancement
+
+SSO: offer ability to force a site to use Jetpack SSO with Two-Factor Authentication for certain roles.
--- a/projects/plugins/jetpack/modules/sso.php
+++ b/projects/plugins/jetpack/modules/sso.php
@@ -58,6 +58,28 @@ class Jetpack_SSO {
 		add_action( 'login_form_jetpack-sso', '__return_true' );
 
 		add_filter( 'wp_login_errors', array( $this, 'sso_reminder_logout_wpcom' ) );
+
+		/**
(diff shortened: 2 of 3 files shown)


kraftbj:
Going to mark this to review in the interest of iteration. I'd like to do some things with it:
1. some kind of failsafe way to login if the connection breaks
2. UI

but with it as it is, I think we can merge it and iterate in trunk.

zinigor (approved):
A quick glance doesn't show me anything wrong. I have rebased the PR onto trunk because otherwise the linter was complaining. Otherwise iterating in trunk sounds good!

PR #33261, opened 2023-09-22
Add "copy to clipboard" to share buttons

manzoorwanijk:
## Proposed changes:

* Add "Copy to clipboard" button to post publish share buttons.

### Other information:

## Jetpack product discussion
pdrWKz-1b4-p2

* On the post publish panel, confirm that you see the "Copy to clipboard" button along with sharing buttons for Twitter and WhatsApp
* Confirm that it works as expected

Screenshot 2023-09-22 at 11 34 59 AM


pablinos (dismissed):
The build is failing but this is working well for me.

projects/js-packages/publicize-components/src/components/share-buttons/copy-to-clipboard.tsx
```diff
@@ -0,0 +1,31 @@
...
+	buttonVariant,
+} ) => {
+	const prepareUrl = usePrepareUrl();
+	const { recordEvent } = useAnalytics();
+
+	const onCopy = useCallback( () => {
+		recordEvent( 'jetpack_social_share_button_clicked', { network: 'clipboard' } );
+	}, [ recordEvent ] );
+
+	const textToCopy = useCallback( () => prepareUrl( '{{text}}\n{{url}}', false ), [ prepareUrl ] );
```
pablinos: I'm not sure there's any advantage to separating this into another component. We have the analytics callback defined in ShareButtons, and it's already using `usePrepareUrl`.

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
```diff
@@ -1,16 +1,14 @@
...
 import styles from './styles.module.scss';
+import { ShareButtonProps } from './types';
 import { usePrepareUrl } from './usePrepareUrl';
 import type React from 'react';
 
-export type ShareButtonsProps = {
-	buttonStyle?: 'icon' | 'text' | 'icon-text';
-	buttonVariant?: React.ComponentProps< typeof Button >[ 'variant' ];
-};
+export type ShareButtonsProps = ShareButtonProps;
```
pablinos: I think if we bring the CopyToClipboard button in here then we can define the type here too

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
```diff
@@ -1,16 +1,14 @@
...
 
 export const ShareButtons: React.FC< ShareButtonsProps > = ( {
```
pablinos: Something for our discussion about style, but I think I prefer not using React.FC and declaring the type of the props

projects/js-packages/publicize-components/src/components/share-buttons/usePrepareUrl.ts
```diff
@@ -23,7 +23,7 @@ export function usePrepareUrl() {
 	}, [] );
 
 	return useCallback(
-		( urlWithPlaceholders: string ) => {
+		( urlWithPlaceholders: string, urlEncode = true ) => {
```
pablinos: I suppose this use case makes the name `usePrepareUrl` a bit inaccurate. Perhaps it should be `useShareButtonText`, and this argument could be something like `isUrl`.

This is probably more bikeshedding though!

pablinos (commented):
This is pretty close I reckon. It's working well for me, just some minor thoughts.

This could be done in a follow-up if necessary, but I wonder if we should have a slightly darker border on the button. @ilonagl, what do you think?

image

It's currently using one of the JP grey colours 
image

But I think the Figma version is slightly darker `#DCDCDE`. I'm not sure if we have that defined though.

ilonagl:
> But I think the Figma version is slightly darker #DCDCDE. I'm not sure if we have that defined though.

Figma shows that it's gray 5. I wonder if it's the same in the code?

projects/js-packages/publicize-components/src/components/share-buttons/copy-to-clipboard.tsx
manzoorwanijk: It was because that array of networks includes the list of social networks and all of them have a URL attached to them which will be opened in a window on click, but in case if Copy to Clipboard, it's not a social network and doesn't have a URL to open.

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
manzoorwanijk: I am not sure I understand what you mean here.

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
manzoorwanijk: I have made the changes, thanks.

projects/js-packages/publicize-components/src/components/share-buttons/usePrepareUrl.ts
manzoorwanijk: Good point. Thanks

manzoorwanijk:
Thank you @ilonagl @pablinos . I have used gray 5 there.

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
manzoorwanijk: Aha, got it. Thanks

Done ✅

pablinos (dismissed):
Works well. Thanks!

pablinos (approved):

PR #33267, opened 2023-09-22
CRM: Add the generic Entity Factory and the Contact Factory

cleacos:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3335

## Proposed changes:

This PR adds the Entity_Factory class with common methods to handle the data between the current DAL format (array) and the entities. It creates an entity instance from DAL data and also extracts the info from an instance to be given to the DAL system.

TL;DR: It transforms the data in both directions Entity <=> DAL.

**Notes**

- This is the first version of the system.
- I extracted the common methods to be able to be used by other CRM entity factories. 
- The `create` method could be extracted too, but we need to implement the generic validators
- With generic validators, and adding to each factory implementation the data to be validated could be a a good improvement. But for now, let's leave into each factory implementation how to create and validate the data.

**Tests**

I added some tests and set the test suite to the PHPUnit config file: `entities`

To run just the entities tests: `composer phpunit -- --testdox --testsuite entities`

### Other information:

## Jetpack product discussion
N/A

N/A

## Testing instructions:

- Check the code and get familiar with how the factories work.
- Run the tests suite. They should pass: `composer phpunit -- --testdox --testsuite entities`


coder-karen (approved):
Tests well, looks good 👍

PR #33269, opened 2023-09-22
Blogroll|: Fix hover/tab on appender list item

renancarvalho:
## Proposed changes:

* Improve accessibility on blogroll appender.

### Other information:

## Testing instructions:

- Checkout branch
- Go to the editor and add a blogroll block
- Open the appender, and you should be able to navigate on the suggestions using tab
- The suggestions should have a blue border when focused

image

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/blogroll/components/blogroll-appender-results/style.scss
+++ b/projects/plugins/jetpack/extensions/blocks/blogroll/components/blogroll-appender-results/style.scss
@@ -24,8 +24,8 @@
 				display: flex;
 				align-items: center;
 				align-self: stretch;
-				padding: 8px 4px;
 				gap: 12px;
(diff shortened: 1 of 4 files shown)


projects/plugins/jetpack/extensions/blocks/blogroll/components/blogroll-appender-results/style.scss
```diff
@@ -25,6 +25,11 @@
 				align-self: stretch;
 				padding: 8px 4px;
```
heavyweight: We can just apply this padding to the `li` and remove our focus rule.
We should respect the theme color schemes and not tie specific colors.

heavyweight (dismissed):
LGTM

heavyweight (approved):

PR #33270, opened 2023-09-22
CRM: Automations - adding factory classes for each entity

coder-karen:
## Proposed changes:

* This PR creates factory classes for each of the entities (minus the contact factory class, defined here -  https://github.com/Automattic/jetpack/pull/33267)
* These factories are needed to help 'hydrate' the entities, and to be able to use the entities within the Automation system
* The PR also includes entity tests, plus some sample data changes to match expected data, and edits to the task entity to match what is expected from the database.

### Other information:

## Jetpack product discussion

https://github.com/Automattic/zero-bs-crm/issues/3343
https://github.com/Automattic/zero-bs-crm/issues/3342
https://github.com/Automattic/zero-bs-crm/issues/3341
https://github.com/Automattic/zero-bs-crm/issues/3340
https://github.com/Automattic/zero-bs-crm/issues/3339

No.

## Testing instructions:

* Tests should pass: `jetpack docker phpunit-crm`, or `composer phpunit -- --testdox --testsuite entities` from the crm directory if not using Docker.


projects/plugins/crm/tests/php/entities/class-transaction-entity-test.php
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	/**
+	 * @testdox Test create transaction entity from input data and insert in DB via DAL.
+	 */
+	public function test_create_transaction_from_input_data_and_insert_into_DB() {
+
+		$transaction_data = $this->generate_transaction_data();
+
+		// Create the Transaction instance from the transaction data (tidy DAL format)
```
kallehauge, line 44: The factories are temporary code until we have repository classes for all entities, so we shouldn't put too much emphasis on them being percect, but inline comments should end with a `.` 😄

kallehauge (approved):
AFAIK then the code looks good and some manual testing confirms it to be true as well 🚀

PR #33271, opened 2023-09-22
AI Extension: Add keyboard shortcut for stop action on forms and update text

### LINKED ISSUE #31469: AI Assistant: add tooltips with keyboard shortcuts
Show keyboard shorctuts for SEND, ACCEPT and STOP

dhasilva:
Fixes #31469

## Proposed changes:

* Changes the `Send` text to `Send (ESC)` to make the shortcut visible
* Adds the shortcut to the AIControl component, making it work for Forms as well

### Other information:

## Jetpack product discussion

p1695400366663969-slack-C054LN8RNVA

## Testing instructions:

* Add an AI Assistant block and send a request
* Check that the `ESC` shortcut is visible
* Press ESC
* The suggestion should stop and the focus should be at the input
* Add a Form
* Check that the shortcut is visible and works as well

The component story has been updated as well and the `Stop` action can be seen on storybook by:
* Running  `( cd projects/js-packages/storybook && pnpm run storybook:dev --no-open )`
* Visiting http://localhost:50240/?path=/story/js-packages-ai-client-ai-control--default&args=state:requesting

| Before | After |
| - | - |
| 2023-09-22_14-42-45-before | 2023-09-22_14-41-26-after |


projects/js-packages/ai-client/src/components/ai-control/index.tsx
```diff
@@ -100,19 +115,34 @@ export function AIControl(
...
 		},
 		{
 			target: promptUserInputRef,
 		}
 	);
 
+	useKeyboardShortcut(
+		'esc',
+		e => {
+			e.stopImmediatePropagation();
```
retrofox: Why do we need to stop the event propagation here?

projects/js-packages/ai-client/src/components/ai-control/index.tsx
```diff
@@ -79,11 +79,26 @@ export function AIControl(
...
+		// Small delay to avoid focus crash
+		setTimeout( () => {
+			InputWrapperRef.current?.focus?.();
+		}, 100 );
+	};
+
+	const handleSend = useCallback( () => {
+		onSend?.( value );
+		// The input is disabled while the request is executing, so we need
+		// to focus on the wrapper for the stop keyboard shortcut to work
```
retrofox: So if the wrapper loses the focus, it won't be possible to cancel the suggestion request by pressing the `ESC` key, right? I mean for instance if the user clicks out of the block instance.

projects/js-packages/ai-client/src/components/ai-control/stories/index.stories.tsx
```diff
@@ -51,10 +51,9 @@ const Template = args => {
 };
 
 const DefaultArgs = {
-	loading: false,
 	isTransparent: false,
 	placeholder: '',
-	requestingState: 'init',
+	state: 'init',
```
retrofox, line 56: <3

projects/js-packages/ai-client/src/components/ai-control/index.tsx
```diff
@@ -79,11 +79,26 @@ export function AIControl(
...
+		focusOnWrapper();
+	}, [ onSend, value ] );
+
 	useKeyboardShortcut(
```
retrofox: The AI Assistant block [already handles the `ESC` keypress event](https://github.com/Automattic/jetpack/blob/8aa42e170d082fdb391a3af1dc667066b7403e9e/projects/plugins/jetpack/extensions/blocks/ai-assistant/edit.js#L395-L405). I think it should be consolidated in the same PR.

dhasilva: That event is for the whole block (red part), while the one added at the AIControl component only works on the green part:

projects/js-packages/ai-client/src/components/ai-control/index.tsx
dhasilva: To avoid double trigger in the AI Assistant because of this:
https://github.com/Automattic/jetpack/pull/33271#discussion_r1336206797

projects/js-packages/ai-client/src/components/ai-control/index.tsx
dhasilva: If the user clicks the AI Assistant wrapper, it still cancels the request, but clicking anywhere else outside of it the cancel shortcut won't work. I think it makes sense because we can have more than one assistants or forms requesting at the same time.

retrofox: yes, but it should be handled at block level. If the user selects the block it won't work I guess.

projects/js-packages/ai-client/src/components/ai-control/index.tsx
retrofox: so we are binding two functions for the same `ESC` keypress event?

retrofox (commented):
I have been examining the approach and conducting research on how we can resolve this issue.

In my opinion, it is not advisable to bind two functions to address the `ESC` keypress event in the same context (block representation in the editor canvas). Currently, the AI Assistant block binds a function to the main block wrapper element, while we are binding a new one to the AIControl component (child of the block wrapper).

Perhaps we could use the `<KeyboardShortcuts />` core component to handle the event at the block wrapper level.

However, this would not suffice when the user interacts from the block toolbar (requesting translation, changing tone, etc.). In such a scenario, the block is not selected/focused. To tackle this, we could either bind a new event to the block toolbar element or ensure the block is selected/focused once the app executes the request. I think we should choose the second option.

Also, while it is possible to register the `ESC` keypress function globally, I do not recommend this approach. The event should be handled at the block wrapper level, considering that multiple AI Assistant block instances could exist simultaneously.

projects/plugins/jetpack/extensions/blocks/ai-assistant/extensions/jetpack-contact-form/components/ai-assistant-bar/index.tsx
```diff
@@ -83,12 +85,15 @@ export default function AiAssistantBar( {
...
+		}, 100 );
+	};
+
 	const { requestSuggestion, requestingState, stopSuggestion, requestingError } = useAiContext( {
-		onDone: () => {
-			setTimeout( () => {
-				inputRef.current?.focus?.();
-			}, 10 );
-		},
+		onDone: focusOnPrompt,
```
retrofox, line 96: IMO, this action is too intrusive:

https://github.com/Automattic/jetpack/assets/77539/2d8bd1ce-d642-4bea-af0f-37fad64d5631

In the video above, I request a form. Then, I started to work on the excerpt. The app focuses on the AI Control textarea when the request is done.
Probably we should check if the block is selected before to focus on the textarea element.

dhasilva: Makes sense!

retrofox (approved):
LGTM :shipit:

PR #33277, opened 2023-09-25
Boost: Remove global state from PopOut component

dilirity:
## Proposed changes:

* Moves global store from PopOut component to parent component (Score);
* Updates "disable prompt" (score) logic to use Data Sync;
* Cleans up "disable prompt" code and global JS scope.

### Other information:

## Jetpack product discussion

pc9hqz-286-p2

no

## Testing instructions:

* Make sure the Score prompt works as expected.
CleanShot 2023-09-25 at 13 56 31

CleanShot 2023-09-25 at 13 57 56


dilirity:
A migration needs to be added for the old option - `jb_show_score_prompt`.

pyronaur (dismissed):
:shipit:

pyronaur (dismissed):
:shipit:

projects/plugins/boost/app/assets/src/js/pages/settings/sections/Score.svelte
```diff
@@ -112,6 +114,14 @@
 		$dismissedScorePromptStore = [ ...$dismissedScorePromptStore, id ];
 		dismissModal();
 	}
+
+	const panelStore = performanceHistoryPanelDS.store;
+	const onTogglePerformanceHistory = status => {
+		panelStore.set( status );
+	};
```
pyronaur: `$panelStore = status`

pyronaur (dismissed):

projects/plugins/boost/app/assets/src/js/pages/settings/sections/Score.svelte
pyronaur: Oooh, and - I got confused about the syntax here. I'm not sure if this is a rule - but we typically have named functions use the word function instead of being a const arrow function. So this probably is:

```ts
function togglePerformanceHistory( status: ... ) {
	$panelStore = status
}
```

Although - if that function is called directly from an even, I think the `status` is probably actually in `e.detail.status` ? (e.detail is how svelte passes event values)

dilirity: > I'm not sure if this is a rule - but we typically have named functions use the word function instead of being a const arrow function.

Gotcha. I've updated the syntax in ddcfc74.

> Although - if that function is called directly from an even, I think the status is probably actually in e.detail.status ? (e.detail is how svelte passes event values)

Thanks, but in this case it's react.

pyronaur (approved):
:doit:

PR #33278, opened 2023-09-25
Boost: Remove global state from History component

dilirity:
## Proposed changes:

* Moves global state to parent component (Score).

### Other information:

## Jetpack product discussion

pc9hqz-286-p2

no

## Testing instructions:

* Make sure the History component is togglable and works as expected.
CleanShot 2023-09-25 at 14 27 15

### DIFF
--- a/projects/plugins/boost/app/assets/src/js/pages/settings/sections/Score.svelte
+++ b/projects/plugins/boost/app/assets/src/js/pages/settings/sections/Score.svelte
@@ -9,6 +9,8 @@
 	import { scoreChangeModal, ScoreChangeMessage } from '../../../api/speed-scores';
 	import ErrorNotice from '../../../elements/ErrorNotice.svelte';
 	import ReactComponent from '../../../elements/ReactComponent.svelte';
+	import { performanceHistoryPanelDS } from '../../../stores/data-sync-client';
+	import { modulesState } from '../../../stores/modules';
 	import RefreshIcon from '../../../svg/refresh.svg';
 	import { recordBoostEvent } from '../../../utils/analytics';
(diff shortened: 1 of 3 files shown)


dilirity:
A migration needs to be added for the old option - `jb_show_score_prompt`.

projects/plugins/boost/app/assets/src/js/pages/settings/sections/Score.svelte
```diff
@@ -102,6 +104,14 @@
...
 		modalData = null;
 	}
+
+	const panelStore = performanceHistoryPanelDS.store;
+	const onTogglePerformanceHistory = status => {
+		panelStore.set( status );
+	};
+
+	$: performanceHistoryNeedsUpgrade = $modulesState.performance_history.available === false;
+	$: performanceHistoryIsOpen = $panelStore;
```
pyronaur, line 114: Should this be bubbled further up? Seems like this component now has more stores than it had before.

pyronaur (changes requested):

projects/plugins/boost/app/assets/src/js/pages/settings/sections/Score.svelte
dilirity: I've left that for the second pass.

dilirity: I'll refactor the Score component again, since both History and PopOut add things in there. I didn't want to bubble up further, as it gets messy with the other components that depend on Settings (which calls Score).

pyronaur (approved):
Ok, thanks for the clarification 👍

PR #33279, opened 2023-09-25
CRM: Add SQL transaction/rollback style to integration tests

cleacos:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3345

## Proposed changes:

This PR changes the way we reset the DB for each test. The right way to do it is with SQL transactions and then rollback it. 

Now we were resetting only the JPCRM database using the function zeroBSCRM_database_reset(); but in our test, we could add WP users and other DB items that are not part of JPCRM. We have to keep the testing DB clean and in the same state on each test, so we have to use the transaction way to roll back any change.

- START TRANSACTION
- Run test
- ROLLBACK

### Other information:

## Jetpack product discussion
N/A

N/A

## Testing instructions:
- Check the code.
- Tests should pass as before.

### DIFF
--- a/projects/plugins/crm/changelog/update-crm-3345-add-sql-transaction-to-integration-tests
+++ b/projects/plugins/crm/changelog/update-crm-3345-add-sql-transaction-to-integration-tests
@@ -1,0 +1,5 @@
+Significance: patch
+Type: changed
+Comment: Changed the method to ensure the DB transactions inside tests and to have a clean DB on each test instead of using the JPCRM reset database function.
+
+
--- a/projects/plugins/crm/tests/php/class-jpcrm-base-integration-test-case.php
+++ b/projects/plugins/crm/tests/php/class-jpcrm-base-integration-test-case.php
@@ -19,7 +19,26 @@ class JPCRM_Base_Integration_Test_Case extends JPCRM_Base_Test_Case {
 	public function tear_down(): void {
 		parent::tear_down();
 
-		zeroBSCRM_database_reset( false );
+		global $wpdb;
+
+		// Rollback the transaction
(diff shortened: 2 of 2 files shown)


coder-karen (approved):
This works for me - approving but leaving for additional reviews too.

kallehauge (commented):
@cleacos did this request come from a specific test failing?
I'm asking since all of this should already be handled when extending `WP_UnitTestCase`.

E.g. [committing](https://github.com/WordPress/wordpress-develop/blob/trunk/tests/phpunit/includes/abstract-testcase.php#L465-L468) is part of [`set_up_before_class`](https://github.com/WordPress/wordpress-develop/blob/trunk/tests/phpunit/includes/abstract-testcase.php#L78) and [`tear_down_after_class`](https://github.com/WordPress/wordpress-develop/blob/trunk/tests/phpunit/includes/abstract-testcase.php#L94)

(They should also rollback, etc)

cleacos:
> did this request come from a specific test failing?

Nope, I was checking the `JPCRM_Base_Integration_Test_Case` and saw the `zeroBSCRM_database_reset( false );` call, and since we can use WP users and other third-party DB elements (WC_Order) in our tests, the JPCRM database reset were not enough to clean up the database.

If `WP_UnitTestCase` does the job of creating and rolling back the transaction, I can update the PR and just remove the `zeroBSCRM_database_reset( false )` call because it's not necessary.

kallehauge:
> If WP_UnitTestCase does the job of creating and rolling back the transaction, I can update the PR and just remove the zeroBSCRM_database_reset( false ) call because it's not necessary.

If I remember correctly, then the tests didn't work as expected without the CRM reset (it "only" resets core WordPress things), but when I started looking through the code yesterday then it looks like it should work, so it requires some further tinkering to be sure

PR #33287, opened 2023-09-25
phpcs: Use PHPCompatibilityWP for dev-develop run, and DRY configs

anomiex:
## Proposed changes:

Because PHPCompatibility hasn't had a release since 2019, we have CI run with their dev-develop branch to catch issues relevant to PHP 8. But when doing so, we weren't respecting the rules that PHPCompatibilityWP disables because WordPress provides polyfills. We had a manual list of a few such functions, but better would be to actually load PHPCompatibilityWP. This PR does that.
    
However, an upside to the lack of checking for WordPress's polyfills was that we caught cases where code that is intended to run outside of WordPress (and so doesn't have those polyfills) tried to use these functions. Since we're losing that, let's make it happen properly by creating a "Jetpack-Compat-NoWP" ruleset that turns the relevant rules back on.

Then add some more standards to DRY some of our configs:

* We have several projects that are intended to run outside of a WordPress context, and these tend to disable the same set of WPCS rules. Add a "Jetpack-NoWP" ruleset that all of these can share and have them use it.
* We have several projects disabling the same set of rules for PHPUnit dirs, relaxing doc comment requirements (test names should generally be self-documenting) and WordPress function requirements. Add a "Jetpack-Tests" ruleset that all of these can share and have them use it.
    
  Also, let's include that ruleset by default for packages and plugins created by `jetpack generate`.

Cleanup of various `phpcs:ignore` and such that are no longer needed will be done in a followup.

### Other information:

## Jetpack product discussion

None I recall to link to. I've been saying I should do this for a while.

No

## Testing instructions:

* Is CI happy?


zinigor (dismissed):

anomiex:
Test failure is due to a WP core change. See p1695748336347839-slack-CBG1CP4EN

sdixon194 (approved):

PR #33288, opened 2023-09-25
phpcs: Clean up after #33287

anomiex:
## Proposed changes:

Changes in #33287 cleaned up our phpcs configs to better ignore rules that need ignoring where they need ignoring. That means we can get rid of a lot of `phpcs:ignore` and `phpcs:disable` directives that are no longer needed.

And while we're at it, let's remove some compatibility code for PHP before 5.6 and make use of WordPress's polyfills where we can.

### Other information:

## Jetpack product discussion

None

No

## Testing instructions:

* Is CI happy?


sdixon194 (dismissed):
Thanks for cleaning all this up!

sdixon194 (approved):
Thanks for cleaning all this up!

PR #33296, opened 2023-09-26
[CRM] Make Automations Workflow endpoints functional

kallehauge:
This PR will update the "Automation Workflows endpoints" that was previously added with dummy data, to being fully functional now. It aims to do so by:

* Complete the `[GET] /wp-json/jetpack-crm/v4/automation/workflows` endpoint
  * Params: `(bool) active`, `(string) category`, `(int) page`, `(int) per_page`, `(int) offset`.
* Complete the `[GET] /wp-json/jetpack-crm/v4/automation/workflows/{id}` endpoint
  * This only accepts the ID in the URL. 
* Complete the `[PUT] /wp-json/jetpack-crm/v4/automation/workflows/{id}` endpoint
  * Params: `(string) name`, `(string) description`, `(bool) active`, `(string) category`, `(string|int) initial_step`, `(object) steps`.
    * The "Steps" param is registered as an "object" because WP REST assumes "array" is a numeric array but we want to allow both strings and integers as the key. We also require a "slug" to be present as a simple verification that we're getting the correct format.

Fixes https://github.com/Automattic/zero-bs-crm/issues/3226

## Proposed changes:

* Rename controller from `REST_Automation_Controller` to `REST_Automation_Workflows_Controller` since, theoretically, if we add other endpoints than workflows, then it makes sense to separate them to improve readability and maintainability.
    * I've also renamed the `REST_Automation_Controller_Test` test class to `REST_Automation_Workflows_Controller_Test` to reflect the above change.
* Renamed methods inside the workflow controller to match default WordPress naming (`get_item`, `get_items`, etc).
* Added pagination to `Workflow_Repository::find_by`
* Changed the singular workflow paths to follow RESTful naming (`/jetpack-crm/v4/automation/workflow` => `/jetpack-crm/v4/automation/workflows/{id}`)
* Complete the workflow endpoints with dummy data:
    * Complete the `[GET] /wp-json/jetpack-crm/v4/automation/workflows` endpoint
    * Complete the `[GET] /wp-json/jetpack-crm/v4/automation/workflows/{id}` endpoint
    * Complete the `[PUT] /wp-json/jetpack-crm/v4/automation/workflows/{id}` endpoint
* Updated tests to use the workflow repository instead of relying on the dummy data that existed in the endpoint skeleton versions.

**Side-note:**

I've tried to keep a link between the old controller and test files, but I cannot for the life of me get it to work right now. So to save time while nearing the ETA, I've just chosen to accept it as is.

### Other information:

## Jetpack product discussion

* pbhBOz-3ss-p2
* pbhBOz-2ES-p2
* pbhBOz-3uX-p2

No.

### Automated tests

* `REST_Authentication_Test`: Updated to include our new endpoints.
* `REST_Automation_Workflows_Controller_Test`: Updated to use real integration tests + added new tests to verify additional expected behaviour like 404 response(s).
* `Workflow_Repository_Test`: Updated to test limit/offset arguments when fetching workflows in bulk.

### Manual testing

**Manual testing prep work**

```php
```

3. Make sure you have a workflow in your database. You can use the below snippet to add one.

```php
	[
		'name'         => 'New contact: Set status to "Customer"',
		'category'     => 'contact',
		'active'       => true,
		'triggers'     => [
		],
		'initial_step' => 1,
		'steps'        => [
			1 => [
				'slug'       => \Automattic\Jetpack\CRM\Automation\Actions\Update_Contact_Status::get_slug(),
				'attributes' => [
					'new_status' => 'Customer',
				],
				'next_step'  => null,
			],
		],
	]
);

$repo = new \Automattic\Jetpack\CRM\Automation\Workflow\Workflow_Repository();
$workflow = $repo->persist(  $workflow );
```

**GET All Workflows**

* Method: `GET`
* Path: `/wp-json/jetpack-crm/v4/automation/workflows`
* Payload: N/A

**GET Workflow**

* Method: `GET`
* Payload: N/A

**Update Workflow**

* Method: `PUT` (or `POST`, both works)
* Payload: `{ "name": "My updated name" }`


projects/plugins/crm/tests/php/rest-api/v4/class-rest-authentication-test.php
```diff
@@ -21,10 +21,22 @@ class REST_Authentication_Test extends REST_Base_Test_Case {
...
+				'/jetpack-crm/v4/automation/workflows',
+			),
+			'automation_workflows::get_item'    => array(
+				WP_REST_Server::READABLE,
+				'/jetpack-crm/v4/automation/workflows/123',
+			),
+			'automation_workflows::update_item' => array(
+				WP_REST_Server::CREATABLE,
+				'/jetpack-crm/v4/automation/workflows/123',
+			),
```
kallehauge, line 39: This dataprovider is a collection of endpoints that require the logged in cookie to access them.
By adding our new endpoints here, we automatically get some auth tests.

projects/plugins/crm/src/automation/workflow/class-workflow-repository.php
```diff
@@ -117,6 +114,15 @@ public function find_by( array $criteria ): array {
 			$query .= ' WHERE ' . implode( ' AND ', $where );
 		}
 
+		// Build pagination condition.
+		if ( isset( $criteria['page'] ) || isset( $criteria['per_page'] ) || isset( $criteria['offset'] ) ) {
```
cleacos: Better if we add the pagination/order functionality as params in the method:

`public function find_by( array $criteria, $order_by = 'id', int $limit = null, int $offset = 0 ): array {`

The reason is that we could have columns with those names: `page`, `per_page`, and `offset`.

I prefer the `limit` and `offset` terms to match with the SQL clauses.

By the way, `offset` and `page` would be the same.

cleacos: But it's true that in this case, we are limiting the criteria allowed keys. But if we are going to generalize it, maybe is better to set it as params from now.

kallehauge: We can definitely change the naming inside the repository itself.
The reason I went with that naming scheme is due to the REST API docs to follow best practice, but those can still be upheld in the endpoint controller instead: https://developer.wordpress.org/rest-api/using-the-rest-api/pagination/#pagination-parameters

I'll give it a look shortly

kallehauge: >By the way, offset and page would be the same.

Agreed. There's a comment that also explains it for the code itself:

https://github.com/Automattic/jetpack/pull/33296/files#diff-df77d7f24c7769b6989fff4eb24d14d497b61c055a2971875cd46af9d5df6d7cR306-R309

projects/plugins/crm/src/automation/workflow/class-workflow-repository.php
```diff
@@ -117,6 +120,22 @@ public function find_by( array $criteria ): array {
...
 		}
 
+		// Add limit/offset clause.
+		if ( $limit > 0 || $offset > 0 ) {
+			// It seems intuitive to provide "0" to mean "no limit", but technically it means that
+			// we do not want to return any results at all, so to help with the developer experience
+			// we convert "0" to a very high number instead.
+			if ( 0 === $limit ) {
+				$limit = 9999999999999;
+			}
```
cleacos: For my business case, `9999999999999` is not enough :^)

And if we use `PHP_INT_MAX`?

projects/plugins/crm/tests/php/rest-api/v4/class-rest-automation-workflows-controller-test.php
```diff
@@ -0,0 +1,307 @@
...
+		// actually retrieving the correct workflow and not just a false-positive response.
+		$this->assertSame( 'test_get_workflow_success', $response_data['name'] );
+	}
+
+	/**
+	 * GET Workflow: Test that we get a 404 if ID does not exist.
+	 *
+	 * @return void
+	 */
+	public function test_get_workflow_return_404_if_id_do_not_exist() {
```
cleacos, line 204: 👍

cleacos (dismissed):
The code looks good. Tests pass. The 3 manual tests work as expected, I can get all the workflows, get one workflow by the ID, and update the workflow.

**Bonus check:** I also tested to query workflows by filtering with other fields, and it works perfectly!!

- `/wp-json/jetpack-crm/v4/automation/workflows?active=0` => Not return anything, as it was expected (by my workflows in my DB)
- `/wp-json/jetpack-crm/v4/automation/workflows?active=1` => It returns the active workflows.
- `/wp-json/jetpack-crm/v4/automation/workflows?name=My updated name` => It returns the expected workflow.

LGTM!

cleacos (approved):

projects/plugins/crm/src/automation/workflow/class-workflow-repository.php
tbradsha: > :^)

@cleacos Imposter! But I agree, `PHP_INT_MAX` is better.

PR #33305, opened 2023-09-26
Boost: Remove global state from AdvancedCriticalCss component

dilirity:
## Proposed changes:

* Moves global store from AdvancedCriticalCss component to parent component (Settings).

### Other information:

## Jetpack product discussion

pc9hqz-286-p2

no

## Testing instructions:

* Make sure the **Critical Css Advanced** section works as expected and you can dismiss/undismiss issues.

To make the Critical CSS advanced recommendations section show up, you need to break a page in some way and regenerate Critical CSS.

The easiest way to do that is to add the following mu script. It will cause the homepage to return error 500 (don't forget to remove it after you finish testing 😅):
```php
<?php

add_action( 'template_redirect', function () {
	if ( is_home() ) {
		nonexistentfunction();
	}
} );

```

### Meta
CleanShot 2023-09-26 at 16 32 46

### Critical CSS advanced page
CleanShot 2023-09-26 at 16 32 26

### DIFF
--- a/projects/plugins/boost/app/assets/src/js/pages/settings/Settings.svelte
+++ b/projects/plugins/boost/app/assets/src/js/pages/settings/Settings.svelte
@@ -4,6 +4,7 @@
 	import Header from '../../sections/Header.svelte';
 	import config from '../../stores/config';
 	import { connection } from '../../stores/connection';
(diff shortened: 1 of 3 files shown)


donnchawp (approved):
Works as expected!

pyronaur (approved):

PR #33306, opened 2023-09-26
CRM: Automations - adding a clientwpuser trigger, data type and tests

coder-karen:
## Proposed changes:

* This PR adds a clientwpuser trigger, data type and tests.
* The trigger listens to the `register_user` hook, and then passes the user ID from the created user into the workflow.
* The test is an integration test that creates a workflow and creates a WP user, then checks that the workflow is executed using the new trigger and that the returned object is a WP User object.

### Other information:

## Jetpack product discussion

https://github.com/Automattic/zero-bs-crm/issues/3102

No.

## Testing instructions:

* Run the test - `jetpack docker phpunit-crm` or `composer phpunit -- --testdox --testsuite automation` from the CRM directory if not using Docker.


projects/plugins/crm/src/automation/class-automation-bootstrap.php
```diff
@@ -106,6 +106,7 @@ protected function register_triggers(): void {
 			\Automattic\Jetpack\CRM\Automation\Triggers\Quote_Updated::class,
 			\Automattic\Jetpack\CRM\Automation\Triggers\Transaction_Created::class,
 			\Automattic\Jetpack\CRM\Automation\Triggers\Transaction_Updated::class,
+			\Automattic\Jetpack\CRM\Automation\Triggers\ClientWPUser_Updated::class,
```
tbradsha: I suspect this was the intent:

```suggestion
			\Automattic\Jetpack\CRM\Automation\Triggers\ClientWPUser_Created::class,
```

tbradsha (dismissed):
There's one typo I've noted inline (`ClientWPUser_Updated` → `ClientWPUser_Created`). Tests pass and the code looks right, so other than that it should be good to go!

As a non-blocker note, is there a reason we've chosen to call it `ClientWPUser`? The "Client" part seems odd to me; while the WP user is typically used with the Client Portal in the CRM context, presumably the WP hook (and thus the trigger) we'd be using is more generic.

In fact, I'd suspect in the future we'd want a general `WP user created` trigger that runs a `create CRM contact` action, for example.

cleacos (approved):
The code looks good. Tests pass. Updated with some commits renaming the ClientWPUser_* by WP_User_*

LGTM!

tbradsha (approved):
Looks good to me!

PR #33310, opened 2023-09-26
Boost: Remove global state from Module component

dilirity:
I'm not entirely happy with the way I did it as now the `state` is passed to every module and there's also a `handleToggle` that handles storing the new state. Hopefully it can be improved.

## Proposed changes:

* Moves global state from Module component to parent component (Modules).

### Other information:

## Jetpack product discussion

pc9hqz-286-p2

no

## Testing instructions:

* Make sure toggling modules on and off works properly.


projects/plugins/boost/app/assets/src/js/pages/settings/elements/Module.svelte
```diff
@@ -1,32 +1,29 @@
...
-			const newState = $modulesState[ slug ].active;
-			if ( lastToggledState !== newState ) {
-				lastToggledState = newState;
-				dispatch( newState ? 'enabled' : 'disabled' );
-			}
+		state.active = ! isModuleActive;
+		if ( state.active ) {
+			dispatch( 'enabled' );
+		} else {
+			dispatch( 'disabled' );
```
pyronaur: It seems that this could be replaced with a reactive statement in Modules.svelte - it's already aware of the modules state, for example:

```js
$: if( $modulesState.critical_css.active && $modulesState.critical_css.available ) {
  resume();
}
```

projects/plugins/boost/app/assets/src/js/pages/settings/elements/Module.svelte
```diff
@@ -1,32 +1,29 @@
...
 		}
+
+		dispatch( 'toggle', {
+			active: state.active,
+			slug,
+		} );
```
pyronaur: If we remove the event above, then this could only fire "toggle" event that would be state independent - and let the state change be calculated in Modules.svelte too - removing the need for `state` prop.

pyronaur (dismissed):

projects/plugins/boost/app/assets/src/js/pages/settings/elements/Module.svelte
dilirity: Thank you for this <3

I've addressed this and your previous comment in dcb754d.

I didn't like passing the state. This feels cleaner now.

donnchawp (dismissed):
Code works for me. Modules toggled on and off and back again.

projects/plugins/boost/app/assets/src/js/pages/settings/sections/Modules.svelte
```diff
@@ -43,6 +43,18 @@
 
 	const suggestRegenerate = suggestRegenerateDS.store;
 
+	$: if ( $modulesState.critical_css.active && $modulesState.critical_css.available ) {
+		resume();
+	} else {
+		alreadyResumed = false;
+	}
```
pyronaur: This seems a bit awkward - ideally we should classify (wrap in a function?) the logic here - I haven't thought about this long and hard at the moment - but my gut reaction is "would this look awkward in React?". The goal is not only to move code around in the rafactor, but also to clean and package it up so that it's easier to port.

The awkward part is that in one, it's setting alreadyResumed to false, but in the other it's calling a function. This probably needs to invoke `resume` function with true/false and do the already resumed comparison there.

projects/plugins/boost/app/assets/src/js/pages/settings/sections/Modules.svelte
```diff
@@ -43,6 +43,18 @@
...
+
+	$: if ( $modulesState.cloud_css.active && $modulesState.cloud_css.available ) {
```
pyronaur: I think we should check for availability before active? That's a nitpick you can probably ignore :)

pyronaur (changes requested):

projects/plugins/boost/app/assets/src/js/pages/settings/sections/Modules.svelte
dilirity: Addressed in e08381f.

dilirity:
@pyronaur I decided to simplify the Module further by moving the availability check to the parent. Dunno if this makes sense to you as well. Thoughts?

Change done in f098504.

projects/plugins/boost/app/assets/src/js/pages/settings/elements/Module.svelte
```diff
@@ -1,74 +1,58 @@
...
 
 	export let toggle = true;
 	export let slug: string;
+	export let isActive: boolean;
 
 	const dispatch = createEventDispatcher();
 
-	$: isModuleActive = $modulesState[ slug ].active;
-	$: isModuleAvailable = $modulesState[ slug ].available;
+	$: isModuleActive = isActive;
```
pyronaur: isActive and isModuleActive is now a bit confused - both are used, both seem the same?

projects/plugins/boost/app/assets/src/js/pages/settings/sections/Modules.svelte
```diff
@@ -48,7 +48,19 @@
 
 	const suggestRegenerate = suggestRegenerateDS.store;
 
-	async function resume() {
+	$: if ( $modulesState.critical_css.available && $modulesState.critical_css.active ) {
```
pyronaur: I thought for this too long and now I'm questioning whether the available check should be here in the first place 🤔  - it should never be active if it's not available 🤔

projects/plugins/boost/app/assets/src/js/pages/settings/sections/Modules.svelte
```diff
@@ -48,7 +48,19 @@
...
+		resumeCriticalCssGeneration();
+	} else {
+		alreadyResumed = false;
```
pyronaur: This still has the same problem as before - it's set in varying levels of abstraction - meaning that here it's **unset** directly in a Svelte expression, and it's **set** a level deeper - in a function. This makes the `resumeCriticalCssGeneration` function virtually useless because it doesn't accept any arguments, it just takes in (and modifies) values outside its scope, relying on the parent scope.

pyronaur (changes requested):
I think that makes sense, as long as you don't need to read the value anywhere else within the components - but ā couple other things still need a good dusting 🙂

projects/plugins/boost/app/assets/src/js/pages/settings/elements/Module.svelte
dilirity: Wow - thanks. Not sure why I did that... I've simplified the check in 2701f1ac3c94a06c15ee0629d76538d70a07cb34.

projects/plugins/boost/app/assets/src/js/pages/settings/sections/Modules.svelte
dilirity: I had the same thought too, but I didn't check if that's the case for some reason.

Now that you mentioned it, I did check and you're right - a module can't be unavailable and active at the same time. It's can only be active if it's available.

Updated in a50757f.

projects/plugins/boost/app/assets/src/js/pages/settings/sections/Modules.svelte
dilirity: If Critical CSS was running and you refreshed the page, it should be started only once (that's by onMount). Currently, it gets started by the active check as well (which is needed in case you toggle off and on so it can resume).

I can't see a way around using an external variable to monitor whether the function was already called or not. I'm also finding it difficult to figure out how to do that just by passing arguments to the function, without having an external variable.

dilirity: I ended up removing the function altogether. I also removed the onMount call since the "is critical css active" check calls it anyway.

projects/plugins/boost/app/assets/src/js/pages/settings/sections/Modules.svelte
dilirity: Continued in the other review.

dilirity:
Closing this as it doesn't benefit the react migration much.

PR #33314, opened 2023-09-26
Subscriptions: Fix condition for showing modal

edanzer:
## Proposed changes:

* Fixes conditional for showing subscribe modal. We were checking post access level, but need to add a check in case the post meta for access level is not set at all. 

### Other information:

## Jetpack product discussion

None. 

No. 

## Testing instructions:

1) Use the bin command below to load this branch in your sandbox. Then sandbox the url of a WordPress simple test site. 
2) Login to your simple site, go to Settings > Newsletter, and enable the subscribe pop-up. 
3) In a separate browser where you are not logged in, open up a single post on the front end, scroll, and confirm the modal shows. 

Note the subscribe modal only shows....
   - on WordPress.com simple/atomic sites
   - on single blog posts
   - if logged out (cannot be logged in as an admin and see it)
   - if you are not already subscribed to the site
   - if post itself doesn't already require subscribing (ie, won't show if post is subscribers only or has paywall block)
   - if 'Enable subscriber pop-up' setting is turned on at Settings > Newsletters

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-subscribe-modal-not-loading
+++ b/projects/plugins/jetpack/changelog/fix-subscribe-modal-not-loading
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
+Subscriptions: Fix conditions for showing modal.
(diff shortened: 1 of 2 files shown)


lezama (approved):

simison (approved):
Simple! 👍

PR #33317, opened 2023-09-26
CRM: Update Automation core to use entities along the update of the Data_Type system

cleacos:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3334

## Proposed changes:

This PR updates the Automation core system to use entities through workflows and also it updates the Data_Type system to use entities too. Now, the workflow will use always a Data_Type. 

The Data_Type implementation encapsulates the data and provides some validation. The data inside of a Data_Type implementation could have any shape, it could be a CRM entity, and object, an array, a WC_Order, or a primitive value. The advantage of this approach is that the Step implementation will know what kind of data to accept and how to handle it.

### Other information:

## Jetpack product discussion
N/A

N/A

## Testing instructions:
- Check the code (roughly)
- Test should pass


projects/plugins/crm/src/automation/class-automation-bootstrap.php
```diff
@@ -34,51 +34,13 @@ final class Automation_Bootstrap {
 	public function init(): void {
 		$this->engine = Automation_Engine::instance();
 
-		$this->register_data_types();
```
cleacos, line 37: We don't need to register data_type along a slug. They are well-known elements (just use the class name).

cleacos:
Tests in my local env pass... 

I'm investigating why is failing here. Also, I'm doing a manual test.

projects/plugins/crm/src/automation/commons/actions/contacts/class-update-contact.php
```diff
@@ -73,33 +76,27 @@ public static function get_category(): ?string {
...
 	 *
 	 * @since $$next-version$$
 	 *
-	 * @param mixed  $data Data passed from the trigger.
-	 * @param ?mixed $previous_data (Optional) The data before being changed.
+	 * @param Data_Type $data Data passed from the trigger.
 	 */
-	public function execute( $data, $previous_data = null ) { // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
+	public function execute( Data_Type $data ) {
+		$this->validate( $data );
```
kallehauge, line 87: It's not blocking, so we shouldn't address it in this specific PR (we can do it in a follow-up if we feel like it's "critical"), but it would be awesome if we didn't have to do `$this->validate` in each step execute method.

**Suggestion:** We could move the logic to `Automation_Engine::execute_workflow` and check `validate()` before we run the `execute()` method.

**Why?**: It will make it more intuitive for 3. party developers to create steps.

projects/plugins/crm/src/automation/commons/conditions/contacts/class-contact-tag.php
```diff
@@ -65,80 +64,44 @@ private function has_tag_by_name( $data, $tag_name ) {
...
 		$operator = $this->get_attributes()['operator'];
 		$tag      = $this->get_attributes()['tag'];
 
 		$this->check_for_valid_operator( $operator );
 
-		if ( ! $this->check_for_valid_parameters( $operator, $data, $previous_data ) ) {
-			$this->condition_met = false;
-			return;
-		}
+		$previous_data = $data->get_previous_data();
```
kallehauge, line 88: We should add a guard clause to bail early if "previous data" isn't a `Contact` instance.
The current implementation could end in a fatal error when `$previous_data` is passed to `$this->has_tag_by_name()`

```suggestion
		$previous_data = $data->get_previous_data();
		
		if ( ! $previous_data instanceof Contact ) {
			$this->condition_met = false;

			return;
		}
```

kallehauge: Correction: we do not use `$previous_state` in `has_tag`, so we cannot add the check where I suggested, but we still need to check further down.

kallehauge (changes requested):
Really, really solid PR! I've left a few comments that I will be addressing myself.

kallehauge (approved):
I made a few changes to fix linting issues and the _very_ small amount of edge-cases I could find by initially going through it.

_(There are several of the "steps tests" that I would love to refactor into integration tests since they integrate into the system and could help catch incorrect mocking assumptions, but the PR achieves feature parity and it would be way out of scope)_

Very good job! 🚀 👍

projects/plugins/crm/src/automation/commons/actions/contacts/class-update-contact.php
cleacos: Yes, absolutely. I felt the same while I was adding it, but not energy to stop and think about how to do it and move it elsewhere. Moving it into `Automation_Engine::execute_workflow` sounds good.

projects/plugins/crm/src/automation/commons/conditions/contacts/class-contact-tag.php
cleacos: `previous_data` is validated in the `Data_Type_Base` constructor, calling to the `validate_data` method, that is implemented by each `Data_Type`.

cleacos:
> There are several of the "steps tests" that I would love to refactor into integration tests since they integrate into the system and could help catch incorrect mocking assumptions

Yes. The JPCRM DAL system is a world, and testing in a real scenario I discovered new issues, for instance, that the DAL sometimes doesn't retrieve the full data object, only the ID, so some validation fails because they try to find specific fields.

Also, there are some update cases where we don't send the previous state, for instance, in transactions. It's something we will have to correct/implement.

PR #33318, opened 2023-09-26
Fix erroneous path check in block registration

### LINKED ISSUE #33312: Blocks: Registration bug when path to block does not begin with /
### Impacted plugin

Jetpack

### Quick summary

The `Automattic\Jetpack\Blocks::jetpack_register_block()` function checks to see if the passed `$slug` is a path and has special logic to handle a passed path. The path detection is done by checking if the first character of `$slug` is `/`. This path check fails on systems where paths do not begin with `/`, such as Windows environments which have paths starting with a drive letter. For example: `C:\wamp64\www\user\wp-content\plugins\jetpack\extensions\blocks\business-hours`.

On a site with such paths, the blocks using the path registration technique will not load properly and create notices such as the following:

```
Notice: Function jetpack_register_block was called incorrectly

trigger_error( $message = 'Function WP_Block_Type_Registry::register was called incorrectly. Block type names must not contain uppercase characters.

WP_Block_Type_Registry-&gt;register( $name = 'jetpack/C:\wamp64\www\user\wp-content\plugins\jetpack\extensions\blocks\business-hours', $args = ['render_callback' =&gt; 'Automattic\Jetpack\Extensions\Business_Hours\render', 'editor_style' =&gt; 'jetpack-blocks-editor'] )
```

### Steps to reproduce

1. Install and activate the latest Jetpack on a site running on Windows.
2. Enable `WP_DEBUG` so that notices are generated.
3. Load a page on the site to trigger the notices.

### Reported instances on wp.org

* [Function jetpack_register_block; WP_Block_Type_Registry::register (5905) issue](https://wordpress.org/support/topic/function-jetpack_register_block-wp_block_type_registryregister-5905-issue/)
* [Function jetpack_register_block](https://wordpress.org/support/topic/function-jetpack_register_block-2/)

monsieur-z:
Fixes #33312

## Proposed changes:

In the `Blocks` and `Jetpack_Gutenberg` classes, the path checks don't take into account Windows paths. This PR fixes it.

### Other information:

## Jetpack product discussion

n/a

No.

### Prerequisites
Enable beta blocks in the Jetpack constants section if you're using a JN site, or by adding `define( 'JETPACK_BLOCKS_VARIATION', 'beta' );` to your `wp-config.php` otherwise.

### Regression Testing
Test that the following blocks are not broken: _Business Hours_, _AI Chat_, _Amazon_, _Blogroll_, _Google Docs_, _Google Sheets_, _Google Slides_, _Recipe_, _Create with Voice_

### Fix

To properly test this, you need a test site that runs on Windows and be able to see PHP logs. After connecting Jetpack, visit `Posts > Add New`. You should **not** see the warning `Function WP_Block_Type_Registry::register was called incorrectly. Block type names must not contain uppercase characters.`.

Alternatively, review the code.

### DIFF
--- a/projects/packages/blocks/changelog/fix-block-path-check
+++ b/projects/packages/blocks/changelog/fix-block-path-check
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+ Fix erroneous path check in Blocks class 
--- a/projects/packages/blocks/src/class-blocks.php
+++ b/projects/packages/blocks/src/class-blocks.php
@@ -43,7 +43,7 @@ class Blocks {
 	public static function jetpack_register_block( $slug, $args = array() ) {
 		// Slug doesn't start with `jetpack/`, isn't an absolute path, or doesn't contain a slash
 		// (synonym of a namespace) at all.
-		if ( 0 !== strpos( $slug, 'jetpack/' ) && 0 !== strpos( $slug, '/' ) && ! strpos( $slug, '/' ) ) {
+		if ( 0 !== strpos( $slug, 'jetpack/' ) && ! path_is_absolute( $slug ) && ! strpos( $slug, '/' ) ) {
 			_doing_it_wrong( 'jetpack_register_block', 'Prefix the block with jetpack/ ', 'Jetpack 9.0.0' );
 			$slug = 'jetpack/' . $slug;
 		}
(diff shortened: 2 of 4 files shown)


chrisbliss18 (approved):
I was able to test this on a WAMP site. Without this fix, I got the warnings as detailed in #33312. After applying these changes, the warnings no longer appeared and the affected blocks function as expected.

Approved!

Good work. Thanks for the quick fix.

sdixon194:
Cherry picked to release branch in c1e6ddd4471f7a13f4493c98b3059dd03eb2b0ef

PR #33319, opened 2023-09-26
AI Assistant: do not request AI feature data by default from the client

### LINKED ISSUE #33248: get_ai_assistance_feature severely degrades performance with uncached HTTP calls
### Impacted plugin

Jetpack

### Quick summary

When using the block editor, Jetpack makes 3 HTTP requests via `get_ai_assistance_feature` in the Jetpack AI helper class to fetch some data. These 3 HTTP requests are identical and take ~0.4-0.6s each, adding up to 1.5s slowdown on block editor screens during asset enquing.

Note that at no point had I attempted to use Jetpack AI features.

1. This should have been cached so only 1 HTTP request was needed
2. It shouldn't be necessary to do this for the Jetpack AI feature to work, nothing prevents this being done in a REST API endpoint so that it can be done at runtime in JS

### Steps to reproduce

Load a block editor screen with Jetpack activated.

I did not expect 3 slow uncached HTTP requests to happen in PHP to the WordPress.com API

### What actually happened

3 duplicate uncached HTTP requests were made for an AI feature I had not yet used.

### Impact

Most (> 50%)

### Available workarounds?

Self-hosted

### Logs or notes

_No response_

retrofox:
Fixes

## Proposed changes:

* 
AI Assistant: do not request AI feature data by default from the client

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-ai-feature/Readme.md
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-ai-feature/Readme.md
@@ -1,28 +1,41 @@
-# useAIFeature()
+# useAIFeature
 
-React custom hook that provides valuable data about AI requests for the site.
+The useAIFeature custom hook facilitates the retrieval of AI assistant feature properties, providing an interface for understanding the AI feature's availability, request limits, and any associated errors or upgrade requirements.
 
-```es6
+```jsx
 function UpgradePlan() {
 	const { hasFeature, count, refresh } = useAIFeature();
+
 	if ( ! hasFeature ) {
-		return null;
+		return (
+			<Button>Upgrade</Button>
+		);
 	}
 
 	return (
 		<div>
-			{ `You have made ${ count } requests so far.` }
-			<Button>Upgrade</Button>
-			<Button onClick={ refresh}>Refresh Data!</Button>
+			<p>
+				You have made ${ count } requests so far.
+			</p>
+			<Button onClick={ refresh }>Refresh data</Button>
 		</div>
 	);
 }
 ```
 
-# getAIFeature()
+### Options
 
-Async helper function that performes and returns relevant data about the AI Assistant feature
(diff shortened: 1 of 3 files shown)


projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-ai-feature/Readme.md
```diff
@@ -1,28 +1,41 @@
-# useAIFeature()
+# useAIFeature
 
-React custom hook that provides valuable data about AI requests for the site.
+The useAIFeature custom hook facilitates the retrieval of AI assistant feature properties, providing an interface for understanding the AI feature's availability, request limits, and any associated errors or upgrade requirements.
```
tomjn, line 3:
```suggestion
The `useAIFeature` custom hook facilitates the retrieval of AI assistant feature properties, providing an interface for understanding the AI feature's availability, request limits, and any associated errors or upgrade requirements.
```

tomjn:
@retrofox that the PR does nothing to address the issue, I suspect there's been a fundamental misunderstanding. There is nothing wrong with refreshing the data in the react hook.

The problem is not javascript triggered requests to the API, the problem is **PHP** triggered requests to the API. Jetpack makes a blocking request in PHP that takes ~0.4-0.5s on the server when enqueuing block assets. If the block asset enqueue fires multiple times it makes multiple unnecessary requests. As a result the entire block editor is delayed from loading by 1.5s.

CGastrell:
Looks good, didn't test as I'm AFK. But I'd like to be aware of what's the use case for this

tomjn:
@CGastrell when loading the block editor WP would make multiple HTTP requests to the Jetpack API to figure out things like how much balance was left and what percentage of the request quote was used, slowing things down a lot. There was a miscommunication though that gave the impression it was JS not PHP code at fault, and this PR was the result.

It might still be a useful change but I'm not in a place to say if that's the case, I guess whoevers the technical lead on these things in Jetpack will need to make the call

lhkowalski:
I will close this since it's not actually addressing the issue: prevent the initial state request to the features API. The original issue remains open: #33839.

PR #33325, opened 2023-09-27
[CRM] Add "Delete Workflow" endpoint

kallehauge:
Add an endpoint that can delete an automation workflow.

Fixes https://github.com/Automattic/zero-bs-crm/issues/3229

## Proposed changes:

* Add `[DELETE] /wp-json/jetpack-crm/v4/automation/workflows/{id}` endpoint.
* Add test to verify that the endpoint works.

### Other information:

## Jetpack product discussion

* pbhBOz-3ss-p2
* pbhBOz-2ES-p2
* pbhBOz-3uX-p2

No.

### Automated tests

* `REST_Authentication_Test`: Updated to include our new endpoint.
* `REST_Automation_Workflows_Controller_Test`: Updated with a new test to verify the endpoint works.

### Manual testing

**Manual testing prep work**

```php
```

**Delete Workflow**

* Method: `DELETE`
* Payload: N/A


projects/plugins/crm/src/rest-api/v4/class-rest-automation-workflows-controller.php
```diff
@@ -213,6 +218,44 @@ public function update_item( $request ) {
...
+			);
+		} catch ( Exception $e ) {
+			return new WP_Error(
+				'rest_unknown_error',
+				$e->getMessage(),
+				array( 'status' => 500 )
+			);
+		}
+
+		return rest_ensure_response( $this->prepare_item_for_response( $workflow, $request ) );
```
cleacos: if we return just an OK/200 instead of returning again the already deleted workflow?

cleacos (dismissed):
The code looks good. Tests pass. The manual testing works well, the workflow is deleted and any further read queries don't retrieve the workflow, as expected.

I left a minor comment, other than that, LGTM!

cleacos (approved):

PR #33354, opened 2023-09-27
Fix Jetpack Sync tests with newest WordPress Core revision mechanisms.

zinigor:
Fixes PHPUnit tests in WordPress trunk.

## Proposed changes:
* Adds a new filtering mechanism to the test event storage class.
* Uses the filtering mechanism to filter out revisions added by [the recent Core change](https://core.trac.wordpress.org/changeset/56714).

### Other information:

No.

## Testing instructions:

* Docker PHPUnit won't work to test this, so you'll have to set up the old fashioned local MySQL server and the WordPress development checkout with the test file.
* Run `WP_DEVELOP_DIR=~/workspace/wordpress-develop/tests/phpunit/ ./vendor/bin/phpunit --filter=WP_Test_Jetpack_Sync_Post`, make sure no tests fail.

### DIFF
--- a/projects/plugins/jetpack/tests/php/sync/server/class.jetpack-sync-server-eventstore.php
+++ b/projects/plugins/jetpack/tests/php/sync/server/class.jetpack-sync-server-eventstore.php
@@ -24,30 +24,48 @@ class Jetpack_Sync_Server_Eventstore {
 		);
 	}
 
-	public function get_all_events( $action_name = null, $blog_id = null ) {
+	/**
+	 * Returns all Sync events of a certain action type, of a certain blog, and filtered if necessary.
+	 *
+	 * @param String   $action_name Sync action slug, e.g. jetpack_sync_save_post.
+	 * @param Integer  $blog_id Blog ID filter - only return events for a given blog ID, defaults to current blog.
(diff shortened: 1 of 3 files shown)


projects/plugins/jetpack/tests/php/sync/server/class.jetpack-sync-server-eventstore.php
```diff
@@ -24,30 +24,48 @@ public function handle_remote_action( $action_name, $args, $user_id, $silent, $t
...
 					$events[] = $event;
 				}
 			}
+		} else {
+			$events = $this->events[ $blog_id ];
+		}
 
-			return $events;
+		if ( is_callable( $filter ) ) {
+			$events = array_filter( $events, $filter );
```
anomiex: Looks like tests are failing because this leaves holes in the array, which callers aren't expecting.
```suggestion
			$events = array_values( array_filter( $events, $filter ) );
```

zinigor: Whoops, broke more tests than fixed :)

anomiex (approved):
Looks sane to me, and CI is passing.

PR #33356, opened 2023-09-27
AI Assistant: Enable backend prompts for 10% of production sites

lhkowalski:
## Proposed changes:

* Keep backend prompts enabled for beta sites
* Also enable backend prompts for 10% of production sites (all sites where the `blog_id` finishes with `7`, a random choice)

This way we can have 10% of production sites using the new feature, and can scale to 50% and 100% when we are sure it's working as expected.

### Other information:

## Testing instructions:

* Test on a site with beta extensions enabled and confirm that it's still using the backend prompts:
   * To check if it's using backend prompt, notice the `jetpack-ai-client:ask-question Asking question:` log message
   * Backend prompts will have messages where `role = jetpack-ai`:

Screenshot 2023-09-27 at 15 24 36

* Now test on a site with production extensions enabled:
   * if the site ID finishes with a 7, you will see the backend prompts (`role = jetpack-ai`)
   * if the site ID does not finish with a 7, you will see the client prompts, with `role = system` and `role = user` messages:

Screenshot 2023-09-27 at 15 30 46

* To simplify your testing, you can switch the `7` on the code to the final number of your testing site ID, build the plugin again and test it; do it using the correct final number and a different final number, so you can confirm it switches between client and backend prompts


projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
```diff
@@ -43,10 +43,13 @@ export const PROMPT_TYPE_LIST = [
 
 export type PromptTypeProp = ( typeof PROMPT_TYPE_LIST )[ number ];
 
-// Support backend prompts as a beta extension.
+// Enable backend prompts for beta sites + 10% of production sites.
+const blogId = parseInt( window?.Jetpack_Editor_Initial_State?.wpcomBlogId );
 export const areBackendPromptsEnabled: boolean =
 	window?.Jetpack_Editor_Initial_State?.available_blocks[ 'ai-assistant-backend-prompts' ]
-		?.available || false;
+		?.available ||
```
CGastrell: shouldn't the first bool operation be `&&`? If not available (or not defined) should not allow for the evaluation to continue? 

Also, out of curiosity, why the last `|| false`?

lhkowalski: Just to close this, as we chatted in Slack:

- I need it to be an OR, since the user should use the backend prompts when it's a beta site as well as when the site ID was from the range specified; the evaluation will keep going until it gets to a truthy value or to the end of the expression
- The `|| false` was indeed not necessary :) removed on the last commit

Thanks for checking it!

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
```diff
@@ -43,10 +43,11 @@ export const PROMPT_TYPE_LIST = [
 
 export type PromptTypeProp = ( typeof PROMPT_TYPE_LIST )[ number ];
 
-// Support backend prompts as a beta extension.
+// Enable backend prompts for beta sites + 10% of production sites.
+const blogId = parseInt( window?.Jetpack_Editor_Initial_State?.wpcomBlogId );
 export const areBackendPromptsEnabled: boolean =
 	window?.Jetpack_Editor_Initial_State?.available_blocks[ 'ai-assistant-backend-prompts' ]
-		?.available || false;
+		?.available || blogId % 10 === 7;
```
retrofox, line 50: why seven? 😅

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
```diff
@@ -43,10 +43,11 @@ export const PROMPT_TYPE_LIST = [
 
 export type PromptTypeProp = ( typeof PROMPT_TYPE_LIST )[ number ];
 
-// Support backend prompts as a beta extension.
+// Enable backend prompts for beta sites + 10% of production sites.
+const blogId = parseInt( window?.Jetpack_Editor_Initial_State?.wpcomBlogId );
 export const areBackendPromptsEnabled: boolean =
 	window?.Jetpack_Editor_Initial_State?.available_blocks[ 'ai-assistant-backend-prompts' ]
```
retrofox: Should we check the `ai-assistant-backend-prompts` key?

```suggestion
	window?.Jetpack_Editor_Initial_State?.available_blocks?.[ 'ai-assistant-backend-prompts' ]
```

retrofox (dismissed):
LGTM! I suggest adding a comment that may be worth considering.

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
lhkowalski: It's a convenient number since my main testing site ID ends with 7 :) But could be any number, actually.

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/prompt/index.ts
lhkowalski: Indeed, I am going to commit your suggestion.

retrofox (approved):
LGTM

CGastrell (approved):
Worked great! :shipit:

PR #33363, opened 2023-09-27
Paywall: Make login work on custom domains

lezama:
Fixes  p1695650415430789-slack-C052XEUUBL4

## Proposed changes:

* Improve Paywall login link for custom domains.

### Other information:

## Testing instructions:

* Use a browser with blocked third party cookies
* Login to WPCOM
* Visit a "WPCOM simple site" on a custom domain where you are already a subscriber that has paywalls
* Visit a post with a paywall and click on "Already a Subscriber".
* Follow the flow – check that you get back to the post and that you have access to the paywalled content.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -1140,7 +1140,7 @@ function get_paywall_blocks( $newsletter_access_level ) {
 
 	$sign_in = '';
 	if ( ! is_user_logged_in() && ( new Host() )->is_wpcom_simple() ) {
-		$sign_in_link = 'aaaaaa' . wpcom_logmein_redirect_url( get_current_url() );
+		$sign_in_link = wpcom_logmein_redirect_url( get_current_url() );
 		$sign_in      = '<!-- wp:paragraph {"align":"center", "typography":{"fontSize":"14px","fontWeight":"400";"color":"#646970"}} -->
 <p class="has-text-align-center" style="font-size:14px;font-weight:400;color:#646970">' . esc_html( $access_question ) . ' <a href="' . $sign_in_link . '" class="jetpack-subscriber-paywall-login" style="font-size:14px;font-weight:400;color:#646970">' . esc_html__( 'Log in', 'jetpack' ) . '</a></p>
 <!-- /wp:paragraph -->';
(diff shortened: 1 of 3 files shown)


projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -1108,7 +1140,7 @@ function get_paywall_blocks( $newsletter_access_level ) {
 
 	$sign_in = '';
 	if ( ! is_user_logged_in() && ( new Host() )->is_wpcom_simple() ) {
-		$sign_in_link = add_query_arg( 'redirect_to', rawurlencode( get_post_permalink() ), 'https://wordpress.com/log-in/link' );
+		$sign_in_link = wpcom_logmein_redirect_url( get_current_url() );
```
simison, line 1143: Since the function isn't in Jetpack, worth adding a check if the function exists?

How does this work on Atomic sites?

lezama: > How does this work on Atomic sites?

it doesn't, it is limited to `is_wpcom_simple` right now.

simison (approved):
I couldn't get this work but I'm not entirely sure about testing steps — it's quite complex, so if this works for you that's ok.

Code on the other hand looks fine.

Note how it didn't log me in to the site, and unrelated to this how Calypso didn't have styles loaded at some point:

Screenshot 2023-09-29 at 16 46 03
Screenshot 2023-09-29 at 16 46 10
Screenshot 2023-09-29 at 16 46 15
Screenshot 2023-09-29 at 16 46 22
Screenshot 2023-09-29 at 16 46 36
Screenshot 2023-09-29 at 16 46 42

crisbusquets:
> I couldn't get this work but I'm not entirely sure about testing steps — it's quite complex, so if this works for you that's ok.
> 
> Code on the other hand looks fine.
> 
> Note how it didn't log me in to the site, and unrelated to this how Calypso didn't have styles loaded at some point:
> 
> Screenshot 2023-09-29 at 16 46 03 Screenshot 2023-09-29 at 16 46 10 Screenshot 2023-09-29 at 16 46 15 Screenshot 2023-09-29 at 16 46 22 Screenshot 2023-09-29 at 16 46 36 Screenshot 2023-09-29 at 16 46 42

Just to be sure: is the same flow as non-custom domains, right?
I'm reviewing it, and it would be cool if we only have to change it once (so we can also add metrics and all that).

simison:
FYI styling issues are now fixed in https://github.com/Automattic/wp-calypso/pull/82682

PR #33371, opened 2023-09-27
CRM Automations UI: Complete table integration

robertf4:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3211

## Proposed changes:

This PR hooks together the workflows table, bulk action, and edit modal component, as well as implements some Redux code which makes them work together.

image

image

### Other information:

## Jetpack product discussion

No

## Testing instructions:

1. Review the code.
2. Put add_filter( 'jetpack_crm_feature_flag_automations', '__return_true' ); in an mu-plugin.
3. Visit /wp-admin/admin.php?page=jpcrm-automations#//automations/manage. Compare visuals to what is found in the Figma. Test the checkboxes, bulk actions, activate/deactivate, and edit buttons.

### DIFF
--- a/projects/plugins/crm/src/js/components/automations-admin/components/workflows-home/index.tsx
+++ b/projects/plugins/crm/src/js/components/automations-admin/components/workflows-home/index.tsx
@@ -1,0 +1,37 @@
+import { useCallback } from 'react';
+import { useNavigate, useParams } from 'react-router-dom';
+import { BulkWorkflowActions } from '../bulk-workflow-actions';
+import { EditModal } from '../edit-modal';
+import { WorkflowTable } from '../workflow-table';
+import styles from './styles.module.scss';
+import type { AutomationsState } from 'crm/state/automations-admin/reducer';
+import type { Workflow } from 'crm/state/automations-admin/types';
+
+type WorkflowsHomeProps = {
+	workflows: AutomationsState[ 'workflows' ];
+};
+
+export const WorkflowsHome: React.FC< WorkflowsHomeProps > = ( { workflows } ) => {
+	const { id } = useParams< { id: string } >();
+
+	let workflow: Workflow | undefined;
+	if ( id ) {
+		const workflowId = Number( id );
+		workflow = workflows[ workflowId ];
+	}
+
+	const navigate = useNavigate();
+	const onEditModalClose = useCallback( () => {
+		navigate( '/automations/manage', { replace: true } );
+	}, [ navigate ] );
+
+	return (
+		<div className={ styles.container }>
+			<BulkWorkflowActions />
+			<WorkflowTable workflows={ Object.values( workflows ) } />
+			{ workflow && (
+				<EditModal isOpen={ !! workflow } onClose={ onEditModalClose } workflow={ workflow } />
+			) }
+		</div>
+	);
+};
--- a/projects/plugins/crm/src/js/components/automations-admin/components/workflows-home/index.tsx
+++ b/projects/plugins/crm/src/js/components/automations-admin/components/workflows-home/index.tsx
@@ -22,7 +22,7 @@ export const WorkflowsHome: React.FC< WorkflowsHomeProps > = ( { workflows } ) =
 
 	const navigate = useNavigate();
 	const onEditModalClose = useCallback( () => {
-		navigate( '/automations/manage', { replace: true } );
+		navigate( '/automations', { replace: true } );
 	}, [ navigate ] );
 
 	return (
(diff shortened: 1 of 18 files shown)


projects/plugins/crm/src/js/components/automations-admin/components/workflows-home/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+	return (
+		<div className={ styles.container }>
+			<BulkWorkflowActions />
+			<WorkflowTable workflows={ Object.values( workflows ) } />
+			{ workflow && (
+				<EditModal isOpen={ !! workflow } onClose={ onEditModalClose } workflow={ workflow } />
+			) }
+		</div>
+	);
```
coder-karen, line 36: I was wondering if it might be worth wrapping any of the components that might be more likely to have errors or be broken, in an error boundary? Only if that's something that could happen on occasion.

Something like:
```
<ErrorBoundary fallback={<p>Something holding text</p>}>
  <WorkflowTable workflows={ Object.values( workflows ) } />
</ErrorBoundary>
```
or whatever is more likely to cause issues? Along with creating the error boundary class etc. 

Maybe that's more an enhancement idea for the future (if it's relevant) since we're not making any of this visible just now?

robertf4: We should implement something like this but I considered it out of scope for this PR. The REST API calls could have errors (e.g. 4XX responses), which I would usually handle with something like
```
const { data, status, error } = useQuery( ... );

if ( error ) {
    return <ErrorDisplay error={ error }/>;
}

if ( status === 'loading' ) {
   return <LoadingIndicator />;
}

return <DataDisplay data={ data }/>
```

There are some generic catch all error handlers in Calypso and I think in the Jetpack plugin, but they're a part of an overall notification system which displays messages in popovers and we don't have that in CRM yet (at least not in the React part). The Figma design does include these so we will need to implement them at some point:

image

kallehauge: Just a heads up about the notifications that we have RNA components for it which hopefully makes it a bit smaller task 👏

projects/plugins/crm/src/js/state/automations-admin/actions.ts
```diff
@@ -10,7 +10,13 @@ export type AutomationsAction =
 	| HydrateWorkflowsAction
 	| ActivateWorkflowAction
 	| DeactivateWorkflowAction
-	| SetAttributeAction;
+	| SetAttributeAction
+	| SelectWorkflowAction
+	| DeselectWorkflowAction
+	| SelectAllWorkflowsAction
+	| DeselectAllWorkflowsAction
+	| ActivateAllSelectedWorkflowsAction
```
kallehauge: Is this meant to be `ActivateSelectedWorkflowsAction` ?

```suggestion
	| ActivateSelectedWorkflowsAction
```

projects/plugins/crm/src/js/state/automations-admin/actions.ts
```diff
@@ -10,7 +10,13 @@ export type AutomationsAction =
...
+	| DeactivateAllSelectedWorkflowsAction;
```
kallehauge: Is this meant to be `DeactivateSelectedWorkflowsAction`?

```suggestion
	| DeactivateSelectedWorkflowsAction;
```

kallehauge (changes requested):
It's not 100% functional for me.

1. The `Active` toggle doesn't work
2. When I open an edit modal, then the actual `attribute` value isn't selected
3. I cannot save a new value in the edit modal
4. We're missing "Trigger" and "Added" values in the table view

projects/plugins/crm/src/js/state/automations-admin/actions.ts
robertf4: Yes, that's correct.

coder-karen (commented):
I'm seeing a couple of console errors when viewing the Automations page at `/wp-admin/admin.php?page=jpcrm-automations#/automations/manage`, but they don't seem to be related to this PR. Since it's the UI and this is the last in progress UI PR I'll add them here anyway, maybe they're not relevant just now, or are easy addition / fixes?

The first was `Warning: Each child in a list should have a unique "key" prop. Check the render method of WorkflowTable. `
I think it's referring to line 19 of `automations-admin/components/workflow-table-header/index.tsx`, but then I think that those props are set just above in WorkflowTableHeaderProps - so it might be a false alarm?

The other was:
`Warning: validateDOMNesting(...): <tr> cannot appear as a child of <table>. Add a <tbody>, <thead> or <tfoot> to your code to match the DOM tree generated by the browser.`
If I wrap the `<tr>` in a `<tbody>` - line 50 in `automations-admin/components/workflow-table/index.tsx` - that warning disappears.

---

I also am probably missing something in terms of testing steps too - do I need to manually add a workflow?

kallehauge (approved):
Let's merge it 👍 

Some know missing aspects this merge assumes we'll fix in follow-ups:

* Update data types so they match API responses
* Save/load from DB
* More attribute configs
* There's a "Jetpack connection package" error in the console that is being triggered by  the automations-admin JS build

PR #33373, opened 2023-09-28
Internal release: Add task for email verification when custom domain present

mpkelly:
Fixes https://github.com/Automattic/dotcom-forge/issues/3948

## Proposed changes:

* Show a Launchpad task for unverified emails when the user purchases a domain

### Other information:

## Testing instructions:

* Patch your sandbox using `bin/jetpack-downloader test jetpack-mu-wpcom-plugin add/task_for_email_verification_when_custom_domain_present`
* Follow test plan at https://github.com/Automattic/wp-calypso/pull/82232


vykes-mac (commented):
I'm not sure if it's because the build has failed but the changes are not applied to the sandbox correctly.

Update: I got applied correctly second time I tried, It's working as is should, note D123297-code must also be applied for the `is_domain_email_unverified` function to work.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -130,6 +130,15 @@ function wpcom_launchpad_get_task_definitions() {
 				return '/me/account';
 			},
 		),
+		'verify_domain_email'             => array(
```
zaguiini, line 133: Having gone through the ICANN discussion, it doesn't seem like this definition is precise, right? Are we dispatching the ICANN verification e-mail when this is clicked?

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -130,6 +130,15 @@ function wpcom_launchpad_get_task_definitions() {
...
+			'get_title'           => function () {
+				return __( 'Verify email address', 'jetpack-mu-wpcom' );
+			},
+			'is_visible_callback' => 'wpcom_launchpad_is_domain_email_unverified',
+			'get_calypso_path'    => function () {
+				return '/me/account';
+			},
```
zaguiini: I'm thinking we should drop this and go with the ICANN verification logic: https://github.com/Automattic/wp-calypso/blob/49c6391c280b60c611e176315471e25a5b3e6f60/client/data/domains/use-domain-email-verfication.ts#L34-L38.

mpkelly: I'm not sure if we can send an email after clicking this. I think sending them to `/domains/manage/all/:domain/edit/:site` would be better.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
mpkelly: It will be precise once I update `is_domain_email_unverified` which I am doing now.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
mpkelly: I used `/domains/manage/:site` as it is easy to produce this URL and still contains all the relevant info, including allowing you to change the email address.

mpkelly:
I've fixed the tests but can't seem to get the GH task to update. Going to merge this anyway.

p-jackson (approved):
Tested with https://github.com/Automattic/wp-calypso/pull/82232 and working as expected

PR #33374, opened 2023-09-28
Newsletter categories: exposes value of wpcom_newsletter_categories_location to JavaScript

### LINKED ISSUE #82322: NC Subscribe Modal: Insert value of filter in JavaScript in `jetpack-mu-wpcom`
Make the changes necessary in `jetpack-mu-wpcom` to expose the value of the `wpcom_newsletter_categories_location` to the block editor. Similar to: 

https://github.com/Automattic/jetpack/blob/295c743a25612dbede6f4b768473cd75ebe19331/projects/packages/jetpack-mu-wpcom/src/class-jetpack-mu-wpcom.php#L122-L137

TimBroddin:
Fixes https://github.com/Automattic/wp-calypso/issues/82322

### Other information:

## Testing instructions:

- Follow the testing instructions below to apply to your sandbox
- Create a new post
- Inspect the `post-new.php` iframe in your devtools:

CleanShot 2023-09-28 at 11 30 02@2x

- Output the value of `Jetpack_Subscriptions` it should be `block`
- Insert these lines inside `0-sandbox.php` and refresh:

```
function wpcom_newsletter_categories_location() {
	return 'modal';
}

add_filter( 'wpcom_newsletter_categories_location', 'wpcom_newsletter_categories_location', 10 );
```

- `modal` should now be returned

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/class-jetpack-mu-wpcom.php
+++ b/projects/packages/jetpack-mu-wpcom/src/class-jetpack-mu-wpcom.php
@@ -48,6 +48,8 @@ class Jetpack_Mu_Wpcom {
 		// Load the Map block settings.
 		add_action( 'enqueue_block_assets', array( __CLASS__, 'load_map_block_settings' ), 999 );
 
+		// Load the Newsletter category settings.
+		add_action( 'enqueue_block_assets', array( __CLASS__, 'load_newsletter_categories_settings' ), 999 );
 		/**
 		 * Runs right after the Jetpack_Mu_Wpcom package is initialized.
 		 *
@@ -141,6 +143,26 @@ class Jetpack_Mu_Wpcom {
 
 		$map_provider = apply_filters( 'wpcom_map_block_map_provider', 'mapbox' );
 		wp_localize_script( 'jetpack-blocks-editor', 'Jetpack_Maps', array( 'provider' => $map_provider ) );
+	}
+
+	/**
+	 * Adds a global variable containing the map provider in a map_block_settings object to the window object.
+	 */
+	public static function load_newsletter_categories_settings() {
+		if (
+			! function_exists( 'get_current_screen' )
+			|| \get_current_screen() === null
+		) {
+			return;
+		}
+
+		// Return early if we are not in the block editor.
+		if ( ! wp_should_load_block_editor_scripts_and_styles() ) {
+			return;
+		}
+
+		$newsletter_categories_location = apply_filters( 'wpcom_newsletter_categories_location', 'block' );
+		wp_localize_script( 'jetpack-blocks-editor', 'Jetpack_Subscriptions', array( 'newsletter_categories_location' => $newsletter_categories_location ) );
 	}
 
 	/**
--- a/projects/packages/jetpack-mu-wpcom/src/class-jetpack-mu-wpcom.php
+++ b/projects/packages/jetpack-mu-wpcom/src/class-jetpack-mu-wpcom.php
@@ -146,7 +146,7 @@ class Jetpack_Mu_Wpcom {
 	}
 
 	/**
-	 * Adds a global variable containing the where the newsletter categories should be shown.
+	 * Adds a global variable containing where the newsletter categories should be shown.
 	 */
 	public static function load_newsletter_categories_settings() {
 		if (
--- a/projects/packages/jetpack-mu-wpcom/src/class-jetpack-mu-wpcom.php
+++ b/projects/packages/jetpack-mu-wpcom/src/class-jetpack-mu-wpcom.php
@@ -146,7 +146,7 @@ class Jetpack_Mu_Wpcom {
 	}
 
 	/**
-	 * Adds a global variable containing the map provider in a map_block_settings object to the window object.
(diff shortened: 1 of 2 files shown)


projects/packages/jetpack-mu-wpcom/src/class-jetpack-mu-wpcom.php
```diff
@@ -136,6 +138,26 @@ public static function load_map_block_settings() {
 		wp_localize_script( 'jetpack-blocks-editor', 'Jetpack_Maps', array( 'provider' => $map_provider ) );
 	}
 
+	/**
+	 * Adds a global variable containing the map provider in a map_block_settings object to the window object.
```
mashikag: I believe this comment is a copy/paste left over. 😁

TimBroddin: Shoot! Fixed it.

projects/packages/jetpack-mu-wpcom/src/class-jetpack-mu-wpcom.php
```diff
@@ -136,6 +138,26 @@ public static function load_map_block_settings() {
 		wp_localize_script( 'jetpack-blocks-editor', 'Jetpack_Maps', array( 'provider' => $map_provider ) );
 	}
 
+	/**
+	 * Adds a global variable containing the where the newsletter categories should be shown.
```
mashikag: I am sorry for being pesky about the comment, but I believe you missed a word: "containing the where the newsletter" 🙈

TimBroddin: I'm moving too fast :D Fixed it.
Thanks for your patience 🙏

mashikag:
BTW, I looked up the documentation on the `wp_localize_script`, and this is what it said ([here](https://developer.wordpress.org/reference/functions/wp_localize_script/#:~:text=Though%20localization%20is,to%20localize%20strings.)):

> Though localization is the primary use, it was often used to pass generic data from PHP to JavaScript, because it was originally the only official way to do that. [wp_add_inline_script()](https://developer.wordpress.org/reference/functions/wp_add_inline_script/) was [introduced in WordPress Version 4.5](https://make.wordpress.org/core/2016/03/08/enhanced-script-loader-in-wordpress-4-5/), and is now the best practice for that use case. `[wp_localize_script()](https://developer.wordpress.org/reference/functions/wp_localize_script/) ` should only be used when you actually want to localize strings.

I see the `wp_add_inline_script` is already used in the same file ([here](https://github.com/Automattic/jetpack/blob/39b239bdf7b1d5c059f0320a90fc35f8d6a6fbef/projects/packages/jetpack-mu-wpcom/src/class-jetpack-mu-wpcom.php#L164)). Is there a particular reason why you chose to use `wp_localize_script` instead? 🤔

mashikag (approved):
Either way I do not believe my above remark should be PR blocking. Other than that the code looks good, and works as expected. 👍

TimBroddin:
> BTW, I looked up the documentation on the `wp_localize_script`, and this is what it said ([here](https://developer.wordpress.org/reference/functions/wp_localize_script/#:~:text=Though%20localization%20is,to%20localize%20strings.)):
> 
> > Though localization is the primary use, it was often used to pass generic data from PHP to JavaScript, because it was originally the only official way to do that. [wp_add_inline_script()](https://developer.wordpress.org/reference/functions/wp_add_inline_script/) was [introduced in WordPress Version 4.5](https://make.wordpress.org/core/2016/03/08/enhanced-script-loader-in-wordpress-4-5/), and is now the best practice for that use case. `[wp_localize_script()](https://developer.wordpress.org/reference/functions/wp_localize_script/) ` should only be used when you actually want to localize strings.
> 
> I see the `wp_add_inline_script` is already used in the same file ([here](https://github.com/Automattic/jetpack/blob/39b239bdf7b1d5c059f0320a90fc35f8d6a6fbef/projects/packages/jetpack-mu-wpcom/src/class-jetpack-mu-wpcom.php#L164)). Is there a particular reason why you chose to use `wp_localize_script` instead? 🤔

`wp_localize_script` was recommended to me by @jeherve I think while working on the Maps block. It seems to be a clean way to expose values from PHP to JS, although I understand that we're probably not using it for what it was meant for 🙊

mashikag:
BTW, I run into an issue when testing, or at least I thought there was an issue. In Calypso I had "Classic View" selected for the posts page. As it comes out, this results in the post editor being loaded in a different manner (even when using the same **Add new post** submenu item). As a result there was no `new-post.php` iFrame present (or at least it is not referred to as such in the dev tools) but instead the values are found in the `top` iFrame. 👍

TimBroddin:
> BTW, I run into an issue when testing, or at least I thought there was an issue. In Calypso I had "Classic View" selected for the posts page. As it comes out, this results in the post editor being loaded in a different manner (even when using the same **Add new post** submenu item). As a result there was no `new-post.php` iFrame present (or at least it is not referred to as such in the dev tools) but instead the values are found in the `top` iFrame. 👍

Nice catch! Glad to hear that it also works correctly without an iframe.

PR #33376, opened 2023-09-28
Newsletter Categories: Add conditional rendering for newsletter categories based on the `wpcom_newsletter_categories_location` filter

### LINKED ISSUE #82323: NC Subscribe Modal: Render the form depending on the value of the `wpcom_newsletter_categories_location` filter in `Jetpack`
Read out the value of the `wpcom_newsletter_categories_location` (fallback to `block`) and use it to render the block:

- `block`: render the newsletter categories
- `modal`: don't render the newsletter categories

When the value is `modal` there's also no need to fetch the newsletter categories. 

This needs to happen both for the Gutenberg editor (JS), and the frontend (PHP).

Depends on #82322 for the JavaScript side.

TimBroddin:
Fixes https://github.com/Automattic/wp-calypso/issues/82323 

## Proposed changes:
- This PR disabled the rendering of the newsletter categories when the value of wpcom_newsletter_categories_location is something else than "block"

The changes are twofold:
1. When rendering the block on a site (with PHP), the filter value is checked in the part of the code that fetched the newsletter categories. If the filter value is something else than "block" (which is the default value when undefined), an empty array is returned. This empty array prevents the block from rendering inside `render_newsletter_categories`.
2. In the editor, the value of `window.Jetpack_Subscriptions?.newsletter_categories_location` is read inside the `useNewsletterCategories` hook. This property contains the value of the wpcom_newsletter_categories_location filter (see #33374). If this value is not set we set it to `block`, otherwise we use the defined value. The newsletter category data is only fetched when this value is set to block. Without data being fetched, the pills won't render.

After we've made the switch to show the categories in the Subscribe Modal, we'll do a follow-up PR to remove all traces of newsletter categories in the subscribe block.

### Other information:

No

## Testing instructions:
1. Add this PR to your sandbox, following the instructions below
2. Enable newsletter categories & select some categories
3. Add a subscribe block: newsletter categories should show in the editor & the site itself
4. Also apply this jetpack-mu-wpcom PR to your sandbox by following the testing instructions: #33374
5. Add this to `0-sandbox.php`:

```
function wpcom_newsletter_categories_location() {
	return 'modal';
}

add_filter( 'wpcom_newsletter_categories_location', 'wpcom_newsletter_categories_location', 10 );
```
6. Newsletter categories should be gone in both the editor & the site itself
7. Change to:

```
function wpcom_newsletter_categories_location() {
	return 'block';
}

add_filter( 'wpcom_newsletter_categories_location', 'wpcom_newsletter_categories_location', 10 );
```

8. The newsletter categories should show in the editor & the site itself

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -264,6 +264,12 @@ function fetch_newsletter_categories() {
  * @return array
  */
 function get_newsletter_categories() {
+	// opt out of rendering when we display newsletter catgeories in the subscribe modal
+	$should_render_newsletter_categories = apply_filters( 'wpcom_newsletter_categories_location', 'block' ) === 'block';
+	if ( ! $should_render_newsletter_categories ) {
+		return array();
+	}
+
 	$response = fetch_newsletter_categories();
 
 	if (
--- a/projects/plugins/jetpack/changelog/add-newsletter-categories-settings-check
+++ b/projects/plugins/jetpack/changelog/add-newsletter-categories-settings-check
@@ -1,0 +1,4 @@
+Significance: minor
+Type: other
+
+Add conditional rendering for newsletter categories based on the wpcom_newsletter_categories_location filter
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/hooks/use-newsletter-categories.js
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/hooks/use-newsletter-categories.js
@@ -22,7 +22,13 @@ const useNewsletterCategories = () => {
 			}
 		};
 
-		fetchData();
+		const newsletter_categories_location =
+			window.Jetpack_Subscriptions?.newsletter_categories_location ?? 'block';
+
+		// only fetch newsletter categories if they should be shown in the block (or the setting isn't defined)
+		if ( newsletter_categories_location === 'block' ) {
+			fetchData();
+		}
 	}, [] );
 
 	return { data, enabled, error, loading };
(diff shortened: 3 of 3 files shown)


phcp (approved):
Good job! Code looks good and works as expected.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -264,6 +264,12 @@ function fetch_newsletter_categories() {
  * @return array
  */
 function get_newsletter_categories() {
+	// opt out of rendering when we display newsletter catgeories in the subscribe modal
+	$should_render_newsletter_categories = apply_filters( 'wpcom_newsletter_categories_location', 'block' ) === 'block';
+	if ( ! $should_render_newsletter_categories ) {
+		return array();
+	}
```
jeherve, line 271: Could you add a docblock here, so our codex can be automatically updated to include this new hook?

I would also recommend using a `jetpack` prefix if you can, if this is distributed in Jetpack.

Something like this, maybe?

```suggestion
	/**
	 * Overwrite the default Newsletter categories location.
	 *
	 * @module subscriptions
	 *
	 * @since $$next-version$$
	 *
	 * @param string $location The default location for the Newsletter categories. default to 'block'.
	 */
	$newsletter_categories_location = apply_filters( 'jetpack_newsletter_categories_location', 'block' );

	// Opt out of rendering when we display newsletter catgeories in the subscribe modal.
	if ( 'block' !== $newsletter_categories_location ) {
		return array();
	}
```

jeherve (changes requested):

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
TimBroddin: Hey @jeherve . Sorry for getting back to this so late, I missed the notification.

This is only going to be here temporary. The newsletter categories will be moving from the block to the subscribe modal. The filter is just here to make the transition smooth (so we can flip the switch - and the categories should be gone everywhere). 

Therefor, I don't think this should be documented.

(Newsletter categories are only enabled on Simple & Atomic for now)

jeherve: Ok, that makes sense. If this ever comes up again, do not hesitate to note directly in the code when something is only there temporarily; this way we remember to take it off afterwards :)

TimBroddin:
Hey @jeherve 

Can this be merged?

jeherve (approved):

PR #33377, opened 2023-09-28
AI Excerpt: do not save the post before tp request the excerpt

retrofox:
When we started to work on this feature, the idea was to be able to collect the post content from the server. In order to guarantee to get the content updated, the app saves the post just before requesting the excerpt.
Now, since the post content is added in the request payload, we don't need to save the post.

Fixes #

## Proposed changes:

* AI Excerpt: do not save the post before tp request the excerpt

### Other information:

## Testing instructions:

* Go to the block editor
* Use the AI Excerpt feature
* It should work as expect 
* Confirm that the post is not saved before to request the excert 

Screenshot 2023-09-28 at 08 52 45

* Additionally, you could check the request send the post content

Screenshot 2023-09-28 at 08 53 36

### DIFF
--- a/projects/plugins/jetpack/changelog/update-ai-excerpt-do-not-save-post
+++ b/projects/plugins/jetpack/changelog/update-ai-excerpt-do-not-save-post
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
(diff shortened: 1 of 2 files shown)


CGastrell (approved):
Works as expected/described :shipit:

PR #33380, opened 2023-09-28
AI Excerpt: do not extend when the AI Assistant feature is not supported

retrofox:
The changes suggested on this PR don't extend the Excerpt panel when the site doesn't support the AI Assistant feature.

Fixes #

## Proposed changes:

* AI Excerpt: do not extend when the AI Assistant feature is not supported

### Other information:

## Testing instructions:

* Test with a site that supports the AI Assistant feature
* Go to the block editor
* Confirm the Exceprt panel is extended with the AI tools

image

* Now, test with a site that requires an upgrade, for instance
* Confirm the classic Excerpt panel is shown

image

### DIFF
--- a/projects/plugins/jetpack/changelog/update-ai-excerpt-do-not-extend-when-feature-is-not-supported
+++ b/projects/plugins/jetpack/changelog/update-ai-excerpt-do-not-extend-when-feature-is-not-supported
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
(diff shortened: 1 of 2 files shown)


CGastrell (approved):
It works, but as it's using the _initial state_, if you go over your quota the feature is already there:

image

I don't mind as this is a sort of exceptional case, but maybe we could take the _siteRequireUpgrade_ value from the state and update the inspector setting (either remove it or disable it and add the "Upgrade" button/teaser). Cool if we take this to another iteration.

:shipit:

PR #33383, opened 2023-09-28
Boost: Remove global state from from critical css related components

dilirity:
## Proposed changes:

* Remove state from CriticalCssStatus, CriticalCssShowStopperError, CriticalCssMeta and CloudCssMeta;
* State is now passed down from ancestors.

### Other information:

## Jetpack product discussion

pc9hqz-286-p2

no

## Testing instructions:

* Make sure that Critical/Cloud CSS generation works as expected and the UI behaves properly.
* You can cause the show stopper error to show up by causing the front-end to error 500.

### DIFF
--- a/projects/plugins/boost/app/assets/src/js/pages/settings/elements/CriticalCssShowStopperError.svelte
+++ b/projects/plugins/boost/app/assets/src/js/pages/settings/elements/CriticalCssShowStopperError.svelte
@@ -8,7 +8,6 @@
 	import { __ } from '@wordpress/i18n';
 	import ErrorNotice from '../../../elements/ErrorNotice.svelte';
 	import FoldingElement from '../../../elements/FoldingElement.svelte';
-	import { regenerateCriticalCss } from '../../../stores/critical-css-state';
 	import { CriticalCssState } from '../../../stores/critical-css-state-types';
 	import CriticalCssErrorDescription from './CriticalCssErrorDescription.svelte';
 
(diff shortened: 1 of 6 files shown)


projects/plugins/boost/app/assets/src/js/pages/settings/elements/CriticalCssShowStopperError.svelte
```diff
@@ -8,15 +8,18 @@
 	import { __ } from '@wordpress/i18n';
 	import ErrorNotice from '../../../elements/ErrorNotice.svelte';
 	import FoldingElement from '../../../elements/FoldingElement.svelte';
-	import { criticalCssState, regenerateCriticalCss } from '../../../stores/critical-css-state';
-	import { primaryErrorSet } from '../../../stores/critical-css-state-errors';
+	import { regenerateCriticalCss } from '../../../stores/critical-css-state';
```
haqadn: Do we want to pass `regenerateCriticalCSS` as a prop as well since this also interacts with the state?

dilirity: I've moved the function to the Modules component in b6ddc19.

Looking at the whole chain - there's something about passing down through 2 layers that I don't like.

However, I'm not sure how to improve it \: Any suggestions? Maybe event dispatching? But that would still have to bubble up...

haqadn (approved):
👍🏼 LGTM

PR #33385, opened 2023-09-28
ci: Try running phpunit tests on Windows

anomiex:
## Proposed changes:

We recently had a bug that only affected Windows servers. Since GH has Windows runners, let's see if that works.

Probably it'll fail due to lack of apt to install php and node, but may as well try it.

### Other information:

## Testing instructions:

* Is CI happy?


anomiex:
> Error: Container operations are only supported on Linux runners

So much for that. We need a mysql server to run WordPress for Jetpack's tests.

Issue #33335, opened 2023-09-27
Sitebot: Improve error handling


artpi:
From CFT: pdtkmj-1OZ-p2

> After a while, one tester started getting a “Your Offline error” when asking a new question and then when it started giving replies again, it would answer a previous question with the offline notice at the end of the page.

![Image](https://github.com/Automattic/jetpack/assets/3775068/ddceddd7-1072-4ec6-981d-d939fa5f6e75)

 
> I received Sorry, there was an error. The response is not a valid JSON response error for one of my questions, even though the bot returned a reply.

PR #33387, opened 2023-09-28
AI Chat: Better state handling, better error handling

artpi:
This PR fixes a few state inconsistencies found in Call for testing

## Proposed changes:
* Solves #33335 
* Clears out details of previous answer when new one is submitted
* All around state management
* Stop displaying the backend error message. I don't think we should be showing it on the frontend of the site, as the user cannot really do anything.

Zrzut ekranu 2023-09-28 o 19 12 55

NO

## Testing instructions:
* Check out everything, sandbox
* Break the endpoint
* See result,
* Submit different question, see error cleared
*

### DIFF
--- a/projects/plugins/jetpack/changelog/sitebot-erros-2
+++ b/projects/plugins/jetpack/changelog/sitebot-erros-2
@@ -1,0 +1,4 @@
+Significance: minor
+Type: enhancement
+
+Enhanced error presentation for ai chat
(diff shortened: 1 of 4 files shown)


dereksmart (approved):
Works well and I'm ok with a generic error message now, but note that one function of providing at least some information with the error can sometimes help Support diagnose issues. 

> I don't think we should be showing it on the frontend of the site, as the user cannot really do anything.

artpi:
> one function of providing at least some information with the error can sometimes help Support diagnose issues.

Yeah, I was thinking how that could work and the only useful use case would be when site owner is trying to debug issues and sees the error message on their site and contacts support.

Once we see what errors we actually have, we can see how can we surface them. For the time being, we'll log em.

PR #33389, opened 2023-09-28
[CRM] Add static workflow step data to API responses

kallehauge:
This PR will add missing static Step information to workflow responses.

The reason we need to expose this is to the UI can display all relevant information about steps.
The current information would limit the UI to only show attribute labels without any further context.

## Proposed changes:

* Add a unified way to get a full version of steps (including static information like title and attribute definition + dynamic values like attributes and "next step").
* Use the above mentioned unified way to fetch step data to expose static values via the API.
* Fix an edge-case bug where we forced "next step" to be an integer, but we allow strings. This could break future hardcoded workflows.
* Unify how we store "next step" amongst all steps (read: conditions + actions).
* Update some of the existing tests to include more up-to-date dummy data (e.g.: there's no reason for us to define "class_name" on steps anymore). This also made it easier for me to do direct comparisons between the original workflow before and after update/create.

### Other information:

## Jetpack product discussion

* pbhBOz-3Wf-p2

No.

### Automated tests

* `REST_Automation_Workflows_Controller_Test`: Updated so all the existing tests do not try to compare actual workflows with API responses that includes static information like step title and description.

### Manual testing

**Manual testing prep work**

```php
```

3. Make sure you have a workflow in your database. You can use the below snippet to add one.

```php
	[
		'name'         => 'New contact: Set status to "Customer"',
		'category'     => 'contact',
		'active'       => true,
		'triggers'     => [
		],
		'initial_step' => 1,
		'steps'        => [
			1 => [
				'slug'       => \Automattic\Jetpack\CRM\Automation\Actions\Update_Contact_Status::get_slug(),
				'attributes' => [
					'new_status' => 'Customer',
				],
				'next_step_true'  => null,
				'next_step_false' => null,
			],
		],
	]
);

$repo = new \Automattic\Jetpack\CRM\Automation\Workflow\Workflow_Repository();
$workflow = $repo->persist(  $workflow );
```

**GET All Workflows**

* Method: `GET`
* Path: `/wp-json/jetpack-crm/v4/automation/workflows`
* Payload: N/A

**GET Workflow**

* Method: `GET`
* Payload: N/A

## New Workflow Response Example

```json
{
    "id": "1",
    "zbs_site": -1,
    "zbs_owner": -1,
    "name": "New contact: Set status to \"Customer\"",
    "description": "This automation will change the status of a contact to \"Customer\" when they are created.",
    "category": "jpcrm/contact",
    "triggers": [
        "jpcrm/contact_created"
    ],
    "steps": {
        "1": {
            "title": "Update Contact Status Action",
            "description": "Action to update the contact status",
            "slug": "jpcrm/update_contact_status",
            "category": "Contact",
            "step_type": "action",
            "next_step_true": null,
            "next_step_false": null,
            "attributes": {
                "new_status": "Customer"
            },
            "attribute_definitions": {
                "new_status": {
                    "slug": "new_status",
                    "title": "New status",
                    "description": "The status that will be used for the contact.",
                    "type": "select",
                    "data": {
                        "Lead": "Lead",
                        "Customer": "Customer",
                        "Refused": "Refused",
                        "Blacklisted": "Blacklisted",
                        "Cancelled by Customer": "Cancelled by Customer",
                        "Cancelled by Us (Pre-Quote)": "Cancelled by Us (Pre-Quote)",
                        "Cancelled by Us (Post-Quote)": "Cancelled by Us (Post-Quote)"
                    }
                }
            }
        }
    },
    "initial_step": "1",
    "active": true,
    "version": 1,
    "created_at": 1695767766,
    "updated_at": 1695767766
}
```

## Previous/Old Workflow Response Example

```json
{
    "id": "1",
    "zbs_site": -1,
    "zbs_owner": -1,
    "name": "New contact: Set status to \"Customer\"",
    "description": "This automation will change the status of a contact to \"Customer\" when they are created.",
    "category": "contact",
    "triggers": [
        "jpcrm/contact_created"
    ],
    "steps": {
        "1": {
            "slug": "jpcrm/update_contact_status",
            "attributes": {
                "new_status": "Customer"
            },
            "next_step_true": null,
            "next_step_false": null,
            "attribute_definitions": {
                "new_status": {
                    "slug": "new_status",
                    "title": "New status",
                    "description": "The status that will be used for the contact.",
                    "type": "select",
                    "data": {
                        "Lead": "Lead",
                        "Customer": "Customer",
                        "Refused": "Refused",
                        "Blacklisted": "Blacklisted",
                        "Cancelled by Customer": "Cancelled by Customer",
                        "Cancelled by Us (Pre-Quote)": "Cancelled by Us (Pre-Quote)",
                        "Cancelled by Us (Post-Quote)": "Cancelled by Us (Post-Quote)"
                    }
                }
            }
        }
    },
    "initial_step": "1",
    "active": true,
    "version": 1,
    "created_at": 1695767341,
    "updated_at": 1695767341
}
```

### DIFF
--- a/projects/plugins/crm/src/automation/class-base-condition.php
+++ b/projects/plugins/crm/src/automation/class-base-condition.php
@@ -25,22 +25,6 @@ abstract class Base_Condition extends Base_Step implements Condition {
 	protected $logger;
 
 	/**
-	 * The next step if the condition is met.
-	 *
-	 * @since $$next-version$$
-	 * @var array|null
-	 */
-	protected $next_step_true = null;
-
-	/**
-	 * The next step if the condition is not met.
-	 *
-	 * @since $$next-version$$
-	 * @var array|null
-	 */
-	protected $next_step_false = null;
-
-	/**
 	 * If the condition is met or not.
 	 *
 	 * @since $$next-version$$
@@ -66,9 +50,7 @@ abstract class Base_Condition extends Base_Step implements Condition {
 	public function __construct( array $step_data ) {
 		parent::__construct( $step_data );
 
-		$this->next_step_true  = $step_data['next_step_true'] ?? null;
-		$this->next_step_false = $step_data['next_step_false'] ?? null;
-		$this->logger          = Automation_Logger::instance();
+		$this->logger = Automation_Logger::instance();
 	}
 
 	/**
@@ -76,9 +58,9 @@ abstract class Base_Condition extends Base_Step implements Condition {
 	 *
 	 * @since $$next-version$$
 	 *
-	 * @return int|null The next step data.
+	 * @return string|int|null The next step data.
 	 */
-	public function get_next_step_id(): ?int {
+	public function get_next_step_id() {
 		return ( $this->condition_met ? $this->next_step_true : $this->next_step_false );
 	}
 
--- a/projects/plugins/crm/src/automation/class-base-step.php
+++ b/projects/plugins/crm/src/automation/class-base-step.php
@@ -35,12 +35,20 @@ abstract class Base_Step implements Step {
 	protected $attribute_definitions = array();
 
 	/**
-	 * Next linked step.
+	 * The next step if the current one is successful.
 	 *
 	 * @since $$next-version$$
 	 * @var int|string|null
 	 */
-	protected $next_step_id = null;
+	protected $next_step_true = null;
+
+	/**
+	 * The next step if not successful.
+	 *
+	 * @since $$next-version$$
+	 * @var int|string|null
+	 */
+	protected $next_step_false = null;
 
 	/**
 	 * Base_Step constructor.
@@ -50,7 +58,9 @@ abstract class Base_Step implements Step {
 	 * @param array $step_data An array of data for the current step.
 	 */
 	public function __construct( array $step_data ) {
-		$this->attributes = $step_data['attributes'] ?? array();
+		$this->attributes      = $step_data['attributes'] ?? array();
+		$this->next_step_true  = $step_data['next_step_true'] ?? null;
+		$this->next_step_false = $step_data['next_step_false'] ?? null;
 	}
 
 	/**
@@ -98,25 +108,62 @@ abstract class Base_Step implements Step {
 	}
 
 	/**
-	 * Set the next step.
-	 *
-	 * @since $$next-version$$
-	 *
-	 * @param int|string|null $step_id The next linked step id.
-	 */
-	public function set_next_step( $step_id ) {
-		$this->next_step = $step_id;
-	}
-
-	/**
 	 * Get the next step.
+	 *
+	 * Unless anything else is defined, then we assume to only continue with the
+	 * next step if the current one is successful.
+	 * One example of this will be conditions where a certain criteria has not been met.
 	 *
 	 * @since $$next-version$$
 	 *
 	 * @return int|string|null The next linked step id.
 	 */
 	public function get_next_step_id() {
-		return $this->next_step_id;
+		return $this->get_next_step_true();
+	}
+
+	/**
+	 * Get the next step if the current one is successful.
+	 *
+	 * @since $$next-version$$
+	 *
+	 * @return int|string|null The next linked step id.
+	 */
+	public function get_next_step_true() {
+		return $this->next_step_true;
+	}
+
+	/**
+	 * Set the next step if the current one is successful.
+	 *
+	 * @since $$next-version$$
+	 *
+	 * @return int|string|null The next linked step id.
+	 */
+	public function set_next_step_true() {
+		return $this->next_step_true;
+	}
+
+	/**
+	 * Get the next step if the current one is falsy.
+	 *
+	 * @since $$next-version$$
+	 *
+	 * @return int|string|null The next linked step id.
+	 */
+	public function get_next_step_false() {
+		return $this->next_step_false;
+	}
+
+	/**
+	 * Set the next step if the current one is falsy.
+	 *
+	 * @since $$next-version$$
+	 *
+	 * @return int|string|null The next linked step id.
+	 */
+	public function set_next_step_false() {
+		return $this->next_step_false;
 	}
 
 	/**
@@ -192,4 +239,33 @@ abstract class Base_Step implements Step {
 	 * @return string|null The category of the step.
 	 */
 	abstract public static function get_category(): ?string;
+
+	/**
+	 * Get the step as an array.
+	 *
+	 * The main use-case to get the step as an array is to prepare
+	 * the items for an API response.
+	 *
+	 * @since $$next-version$$
+	 *
+	 * @return array The step as an array.
+	 */
+	public function to_array(): array {
+		$step = array(
+			'title'                 => static::get_title(),
+			'description'           => static::get_description(),
+			'slug'                  => static::get_slug(),
+			'category'              => static::get_category(),
+			'next_step_true'        => $this->get_next_step_true(),
+			'next_step_false'       => $this->get_next_step_false(),
+			'attributes'            => $this->get_attributes(),
+			'attribute_definitions' => array(),
+		);
+
+		foreach ( $this->get_attribute_definitions() as $attribute_definition ) {
+			$step['attribute_definitions'][ $attribute_definition->get_slug() ] = $attribute_definition->to_array();
+		}
+
+		return $step;
+	}
 }
--- a/projects/plugins/crm/src/automation/class-base-step.php
+++ b/projects/plugins/crm/src/automation/class-base-step.php
@@ -138,10 +138,11 @@ abstract class Base_Step implements Step {
 	 *
 	 * @since $$next-version$$
 	 *
-	 * @return int|string|null The next linked step id.
+	 * @param string|int|null $step_id The next linked step id.
+	 * @return void
 	 */
-	public function set_next_step_true() {
-		return $this->next_step_true;
+	public function set_next_step_true( $step_id ): void {
+		$this->next_step_true = $step_id;
 	}
 
 	/**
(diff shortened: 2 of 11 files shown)


projects/plugins/crm/tests/php/automation/class-automation-workflow-test.php
```diff
@@ -275,6 +276,7 @@ public function test_workflow_execution_with_condition_true() {
 		$automation = new Automation_Engine();
 		$automation->set_automation_logger( $logger );
 		$automation->register_trigger( Contact_Created_Trigger::class );
+		$automation->register_step( Contact_Condition::class );
```
kallehauge, line 279: Our hardcoded dummy workflow data had hardcoded the class name for the mock "Contact Condition" step, so we never had to register the step.
This was technically a false positive I had to correct after I updated the dummy data, but is otherwise unrelated to the PR itself.

projects/plugins/crm/tests/php/automation/class-automation-workflow-test.php
```diff
@@ -312,6 +314,7 @@ public function test_workflow_execution_with_condition_false() {
...
```
kallehauge, line 317: Our hardcoded dummy workflow data had hardcoded the class name for the mock "Contact Condition" step, so we never had to register the step.
This was technically a false positive I had to correct after I updated the dummy data, but is otherwise unrelated to the PR itself.

projects/plugins/crm/tests/php/rest-api/v4/class-rest-automation-workflows-controller-test.php
```diff
@@ -311,11 +311,16 @@ public function test_update_workflow_success() {
...
+					'next_step_true'  => 'updated_step_2',
+					'next_step_false' => null,
+					'attributes'      => array(),
 				),
 				'updated_step_2' => array(
-					'slug' => Contact_Condition::get_slug(),
+					'slug'            => Contact_Condition::get_slug(),
+					'next_step_true'  => null,
+					'next_step_false' => null,
+					'attributes'      => array(),
```
kallehauge, line 323: Updated dummy data so they more accurately represents what would be fetched by `Base_Step::to_array()` which makes direct comparisons in tests much easier.

projects/plugins/crm/tests/php/automation/tools/class-automation-faker.php
```diff
@@ -202,19 +203,21 @@ public function workflow_with_condition_action(): array {
...
-					'next_step_true'  => 1,
-					'next_step_false' => null,
 				),
 				// Step 1
 				1 => array(
-					'slug' => 'dummy_step',
+					'slug'            => Dummy_Step::get_slug(),
+					'next_step_true'  => null,
+					'next_step_false' => null,
+					'attributes'      => array(),
```
kallehauge, line 220: Updated dummy data so they more accurately represents what would be fetched by `Base_Step::to_array()` which makes direct comparisons in tests much easier.

robertf4 (dismissed):

projects/plugins/crm/src/automation/class-base-condition.php
```diff
@@ -66,19 +50,17 @@ abstract class Base_Condition extends Base_Step implements Condition {
...
+		$this->logger = Automation_Logger::instance();
 	}
 
 	/**
 	 * Get the next step.
 	 *
 	 * @since $$next-version$$
 	 *
-	 * @return int|null The next step data.
+	 * @return string|int|null The next step data.
```
tbradsha, line 61: Is there more info somewhere on why we allow both `string` and `int`?

projects/plugins/crm/src/automation/class-base-step.php
```diff
@@ -98,25 +108,62 @@ public function set_attribute_definitions( array $attribute_definitions ) {
...
 	 *
 	 * @since $$next-version$$
 	 *
-	 * @param int|string|null $step_id The next linked step id.
+	 * @return int|string|null The next linked step id.
 	 */
-	public function set_next_step( $step_id ) {
-		$this->next_step = $step_id;
+	public function get_next_step_id() {
+		return $this->get_next_step_true();
```
tbradsha, line 122: ❤️ I much prefer it hard-coded to `true` like this vs. using `next_step`.

projects/plugins/crm/src/automation/class-base-step.php
```diff
@@ -98,25 +108,62 @@ public function set_attribute_definitions( array $attribute_definitions ) {
...
+	/**
+	 * Set the next step if the current one is successful.
+	 *
+	 * @since $$next-version$$
+	 *
+	 * @return int|string|null The next linked step id.
+	 */
+	public function set_next_step_true() {
+		return $this->next_step_true;
+	}
```
tbradsha: This doesn't appear to set anything?

projects/plugins/crm/src/automation/class-base-step.php
```diff
@@ -98,25 +108,62 @@ public function set_attribute_definitions( array $attribute_definitions ) {
...
+
+	/**
+	 * Set the next step if the current one is falsy.
+	 *
+	 * @since $$next-version$$
+	 *
+	 * @return int|string|null The next linked step id.
+	 */
+	public function set_next_step_false() {
```
tbradsha: This doesn't appear to set anything?

projects/plugins/crm/src/automation/interface-step.php
```diff
@@ -36,13 +36,40 @@ public function execute( Data_Type $data );
...
+	public function get_next_step_true();
+
+	/**
+	 * Set the next step if the current one is successful.
+	 *
+	 * @since $$next-version$$
+	 *
+	 * @return int|string|null The next linked step id.
+	 */
+	public function set_next_step_true();
```
tbradsha: Do we need this?

projects/plugins/crm/src/automation/interface-step.php
```diff
@@ -36,13 +36,40 @@ public function execute( Data_Type $data );
...
+	public function get_next_step_false();
+
+	/**
+	 * Set the next step if the current one is falsy.
+	 *
+	 * @since $$next-version$$
+	 *
+	 * @return int|string|null The next linked step id.
+	 */
+	public function set_next_step_false();
```
tbradsha: Do we need this?

projects/plugins/crm/src/automation/class-base-step.php
kallehauge: Where's my "whistling emoji" when I need one - nice catch! Thank you!

_Context: They're there for the argument of always having matching setters when we have getters, but currently not used anywhere, so that's a nice catch._

projects/plugins/crm/src/automation/class-base-step.php
kallehauge: Where's my "whistling emoji" when I need one - nice catch! Thank you!

_Context: They're there for the argument of always having matching setters when we have getters, but currently not used anywhere, so that's a nice catch._

projects/plugins/crm/src/automation/class-base-step.php
tbradsha: Indeed, I'm fine for setters if they actually set things (just saw your last commit). :^)

projects/plugins/crm/src/automation/interface-step.php
kallehauge: Due to the way we've implemented the system, then yes.

Because we use `<something> instanceof Step` then we don't enforce folks to extend our abstract base class (`Base_step`), so 3. party developers could technically avoid the base class right now, so we have to force them to implemented the classes if they skip the base class.

I would personally prefer if we had consistently used `Base_Step` and `is_subclass_of( <something>, Base_Step::class )` so we could assume everything that exists in the abstract base class would be present, but we haven't

projects/plugins/crm/src/automation/interface-step.php
kallehauge: Due to the way we've implemented the system, then yes.

Because we use `<something> instanceof Step` then we don't enforce folks to extend our abstract base class (`Base_step`), so 3. party developers could technically avoid the base class right now, so we have to force them to implemented the classes if they skip the base class.

I would personally prefer if we had consistently used `Base_Step` and `is_subclass_of( <something>, Base_Step::class )` so we could assume everything that exists in the abstract base class would be present, but we haven't

projects/plugins/crm/src/automation/class-base-condition.php
kallehauge: It's indirectly documented in the example here: pbhBOz-3Wf-p2#comment-4808

There's been some previous discussions about it as well where the argument was used that strings would make it easier for us when we hardcode workflows

tbradsha: Thanks!

projects/plugins/crm/src/automation/class-base-step.php
kallehauge: My contextual remark was to say why it wasn't detected 😛 Normally our tests would crash if they were actually being used

projects/plugins/crm/src/automation/interface-step.php
tbradsha: > we use `<something> instanceof Step`

I'm not seeing that anywhere, to be honest, so I don't see the interface actually being used/needed (other than the fact it's currently referenced in the base class implementation).

To me it's essentially redundant vs. just implementing the base class). I agree that my preference is extending a parent class.

tbradsha (approved):

projects/plugins/crm/src/automation/interface-step.php
kallehauge: I guess it's not as bad as I thought/remembered. I know we have a few methods that type-hints return value as `method(): Step`

kallehauge: But I do find it unnecessary as well. I tend to complete move away from interfaces if I have an abstract class because you might as well implement what's in the interface as `abstract` and then add your generic logic as well

PR #33391, opened 2023-09-28
Jetpack AI: cache the AI assistant feature data for Jetpack sites

### LINKED ISSUE #33248: get_ai_assistance_feature severely degrades performance with uncached HTTP calls

retrofox:
@tomjn reported an issue where the app executes the exact requests multiple times in a brief interval. 
To prevent excessive requests to the dotcom side, this PR temporarily caches the first request for ten seconds, which reduces the initial rendering process time.

Fixes https://github.com/Automattic/jetpack/issues/33248

## Proposed changes:

* Jetpack AI: cache the AI assistant feature data for Jetpack sites

### Other information:

## Testing instructions:

* To test the cache system, you must add some logs in the code.

* Confirm the AI Assistant feature is propagated to the initial state.
  * Open the dev console of your dev client
  * Print the value of the following object

```js
Jetpack_Editor_Initial_State['ai-assistant']
```

* Confirm you see the data defined:
image

* Confirm the `ai-assistant-feature` endpoint response:
  * Open the Network tab of your dev client
  * Filter the requests by `ai`
  * Force the client to perform a request, closing and open the AI Jetpack panel, for instance
  * Confirm you see the response body

image


lhkowalski:
I tested it and can confirm it's working as intended. For the first load of the editor, we reduced the amount of calls to the HTTP endpoint to one:

Screenshot 2023-09-28 at 18 41 29

This will mitigate the issue and, as we chatted about on Slack, we can check if it's possible to not set an initial state at all for the AI Assistant as a follow up.

I will leave it open for now so we can collect insights from the other reviewers.

CGastrell (approved):
Looks good, prevents backend fetching effectively. Wondering if we might even increase those 10s to a full minute? Once I open the editor up until I have a fully functional UI, more than 10 seconds have elapsed (most of the time), and I don't see that the AI feature status might be changing (or be it a point of inflection) if checked within a minute.

:shipit:

retrofox:
> This will mitigate the issue and, as we chatted about on Slack, we can check if it's possible to not set an initial state at all for the AI Assistant as a follow up.

https://github.com/Automattic/jetpack/issues/33839

PR #33392, opened 2023-09-28
i18n-check-webpack-plugin: Provide a way to not break with webpack's mangleExports

anomiex:
## Proposed changes:

Webpack's optimization.mangleExports option, enabled by default in production mode, can on occasion mangle an export to the name of one of the i18n functions this module looks for.

To fix this, we provide a slightly modified copy of Webpack's internal `MangleExportsPlugin` that will never mangle to `__`, `_n`, `_x`, or `_nx`.

Then update our webpack-config package to disable `mangleExports` in favor of this new plugin.

### Other information:

## Jetpack product discussion

p1695921976230749-slack-CBG1CP4EN

No

## Testing instructions:

* Is CI happy?
* Check out c9c8b51924c21b35fa54dbd08fed694da3b68d9a, cherry-pick this, and see if `jetpack build -v --production plugins/crm` now works.

### DIFF
--- a/projects/js-packages/i18n-check-webpack-plugin/README.md
+++ b/projects/js-packages/i18n-check-webpack-plugin/README.md
@@ -32,6 +32,31 @@ Parameters recognized by the plugin are:
    - `babelOptions`: Supply options for Babel.
    - `functions`: Supply a custom list of i18n functions to recognize.
 
+## Export mangling
+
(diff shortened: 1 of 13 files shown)


zinigor (approved):
I don't know how exactly to test it, but it seems that the numbers of `__`,`_n`, and `_x` calls in built JS files with and without this PR have not changed, so it should be good.

PR #32815, opened 2023-09-01
Refactor beta blocks registration

monsieur-z:
See https://github.com/Automattic/jetpack/issues/32602

## Proposed changes:
Registers the beta blocks (_AI Chat_, _Amazon_, _Blogroll_, _Google Docs Embed_, _Recipe_, _Create with Voice_) by specifying the path to their block.json file. This was made possible by https://github.com/Automattic/jetpack/pull/32697. The goal is to avoid the duplication of block metadata and use block.json as the source of truth.

In more details:
- Update `registerJetpackBlockFromMetadata` to redeclare variations icons on the client.
- Add helper to get a block's name and feature name.
- For each block (when needed):
  - Write reading direction-agnostic CSS
  - Remove the `editorScript` property from `block.json` (automatically added during the build)
  - Convert icon from dashicon to SVG in the `icon` property of `block.json`
  - Remove `icon.js`
  - Import icon helpers from the `shared-extension-utils` package
  - Register the block client and server sides with `block.json`
  - Move data from `index.js` to `block.json` and `editor.js` and delete the file
  - Move data from `attributes.js` to `block.json` and delete the file
  - Fix JS tests

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

pedMtX-RS-p2

## Does this pull request change what data or activity we track or use?

No.

## Testing instructions:

- Enable beta blocks in the Jetpack constants section if you're using a JN site, or by adding `define( 'JETPACK_BLOCKS_VARIATION', 'beta' );` to your `wp-config.php` otherwise.
- Test that the blocks behave as in `trunk` (some newer ones, such as _Blogroll_, might not be functional yet). This PR shouldn't change anything from a user's standpoint.

A couple of notes:
- Blocks metadata don't seem to be localized in production.
- _Blogroll_ seems to be often updated  at the moment. The icons, for instance, don't currently match. I'll make sure to fix that in a subsequent PR.
- I've removed the links from the _Recipe_ description. It didn't lead to anything and I couldn't find a relevant page to link to.

Screenshot 2023-09-11 at 1 15 21 PM

PR #14528, opened 2020-01-30
Update BlockStylesSelector component and implement in Eventbrite

scruffian:
This makes some changes to the BlockStylesSelector component and implements it in the Eventbrite block.

Fixes #14527

#### Changes proposed in this Pull Request:

* The preview for the button Embed Type on the EventBrite block will now be a real block preview.
* The EventBrite block also now allows you to change the Embed Type from the BlockControls.
* The `useModal` attribute in the block is now deprecated and replaced with a `style` attribute.
* The component now also includes the inspector controls and creates a the block toolbar

#### Is this a new feature or does it add/remove features to an existing part of Jetpack?
* adding to existing features

#### Testing instructions:

* Add an Eventbrite block
* Add an event
* Check that you can still toggle between a button and an inline embed
Screenshot 2020-01-31 at 13 47 00

* Check that the Calendly, OpenTable and EventBrite blocks all have a Toolbar which allows you to change the embed style:
Screenshot 2020-01-31 at 14 16 42

* You also need to test the deprecation for old Eventbrite blocks using the old attributes.
* Create an Eventbrite block using master
* Set the Embed type to be "Modal"
* Save the post
* Check out this branch
* Reload the editor
* Check that the block is still in modal type
* Open the editor in code view and check you see `"style":"modal"` in the block attributes.

#### Proposed changelog entry for your changes:

* no changelog

PR #14383, opened 2020-01-17
Share the block style selector code

scruffian:
Fixes #14466 

#### Changes proposed in this Pull Request:

* OpenTable and Calendly both have a block style selector which does essentially the same thing. Let's extract that to a new shared component.

#### Testing instructions:

* Add a Calendly block and check that you can still change the styles in the sidebar
* Add an OpenTable block and check that you can still change the styles in the sidebar
Screenshot 2020-01-17 at 13 48 58
Screenshot 2020-01-17 at 13 48 51

#### Proposed changelog entry for your changes:

* no changelog

PR #33395, opened 2023-09-28
Gutenberg: Add Goodreads Block

### LINKED ISSUE #7920: Widgets: Add Sort Option to Goodreads Widget
When adding the Goodreads widget it displays the listed books in order of date added, and there is no way to change that. 

There should be options to sort the materials by date read, author, alphabetical by title, etc.

### LINKED ISSUE #74494: Add Goodreads as a Block
### What

We've tried to encourage our users to update to newer themes, but several legacy widgets, including Goodreads, do not work in block themes.

### Why

Currently, to get the old look Goodreads widget look with a newer theme, a user has to copy the GoodReads code, select just the "Static HTML" portion, and paste that into a block. And then they'd have to repeat this every time they update their Goodreads account.

### How

Can we create a Goodreads block so folks who enjoy the legacy widget are free to update their sites to newer themes?

I'd love to make it less complicated than the old version, too, and properly support readers as well as authors. [Here are the old steps.](https://wordpress.com/support/widgets/goodreads-widget/)

Aurorum:
Fixes Automattic/wp-calypso#74494
Fixes #7920 
## Proposed changes:

Adds a block for Goodreads so that users can update to newer themes - it's currently only available as a legacy widget. This block is similar to the existing widget in that it relies on the default Goodreads styles, but I've added a little bit more customisation which Goodreads offers that was never included in the legacy widget. This can be seen in the Settings panel, and it just controls which books display.

Goodreads also offers a Grid widget that was never used in Jetpack. I feel like Gutenberg is perfect at handling multiple styles, so I've added it with this block.

| Grid Widget                                      | Default Widget                                     |
| -------------------------------------------- | -------------------------------------------- |
| Screenshot 1 | Screenshot 2 |

I'm marking this as a draft because @supernovia mentioned in the original issue that it'd be good to support authors as well as readers. Unfortunately, I don't have an Authors Goodreads account, so I can't yet see how to implement those widgets. I'll do a bit more digging on that, but would be very grateful for any early feedback - it's been a while since I've tried my hand at making a block!

No changes. Regarding the above section, I don't believe that I'm able to test my code on WordPress.com, and I'm not familiar with e2e tests for Jetpack.

## Testing instructions:

- You should be able to create a Goodreads block using a profile URL - so for example: https://www.goodreads.com/user/show/1176283-matt-mullenweg
- Verify that it loads properly and that each of the settings in the sidebar works 
- Hopefully nothing is too out of place! 

Notably, you should also be able to test this with an Author URL. The Goodreads Widget directs users to documentation on how to acquire their user ID because it's different for authors; with this block, it's much simpler and only requires your profile URL. Example author profile: https://www.goodreads.com/author/show/1406384


projects/plugins/jetpack/extensions/blocks/goodreads/edit.js
```diff
@@ -0,0 +1,419 @@
...
+
+		if ( ! goodreadsId ) {
+			return;
+		}
+
+		const html = `
+		<style> [class^=gr_custom_container_] { border: 1px solid gray; border-radius: 10px; margin: auto; padding: 0 5px 10px 5px; background-color: #FFF; color: #000; width: 300px; }  [class^=gr_custom_header_] { border-bottom: 1px solid gray; width: 100%; padding: 10px 0; margin: auto; text-align: center; font-size: 120%; }  [class^=gr_custom_each_container_] { width: 100%; clear: both; margin: auto; overflow: auto; padding-bottom: 4px; border-bottom: 1px solid #aaa; }  [class^=gr_custom_each_container_] { width: 100%; clear: both; margin-bottom: 10px; overflow: auto; padding-bottom: 4px; border-bottom: 1px solid #aaa; }  [class^=gr_custom_book_container_] { overflow: hidden; height: 60px; float: left; margin-right: 6px; width: 39px; }  [class^=gr_custom_author_] { font-size: 10px; }  [class^=gr_custom_tags_] { font-size: 10px; color: gray; }  [class^=gr_custom_rating_] { float: right; }  [class^=gr_grid_book_container] { float: left; width: 98px; height: 160px; padding: 0px 0px; overflow: hidden; }  a { text-decoration: none; }  a:hover { text-decoration: underline; } img { max-width: 100%; }</style>
```
Aurorum: There has got to be a better way of doing this...but no idea what that would be!

Aurorum:
Although it would still be nice to add the Author widgets, I can't seem to figure out how to get access to them right now, so I'm marking this as ready for review! Nevertheless, I've added some changes so that it's much simpler for authors to use this block anyway.

Whereas the Goodreads widget requires people to read documentation about how to acquire their User ID in case they give us their Author ID, now you can just input your profile URL and the block will handle distinguishing between the two.

Aurorum:
@jeherve I’m looking at adding tests for this block and I’d appreciate some advice - what does the `fixtures` folder in each blocks folder actually do? eg. https://github.com/Automattic/jetpack/tree/trunk/projects/plugins/jetpack/extensions/blocks/eventbrite/test/fixtures

Is this necessary for the tests? Also, apologies if I’ve just completely missed it, but is there any documentation which you could point me to on this? Thanks!!

supernovia:
Awesome, thanks @Aurorum !

jeherve:
@Aurorum The fixtures represent mock data for each block. That data is used in the `runBlockFixtureTests` function here, basically ensuring that the block structure outcome you have in the fixtures matches what's actually outputted when you add a block:
https://github.com/Automattic/jetpack/blob/260c45af538dab84b7ccdbad52b869c4c0911efb/projects/plugins/jetpack/extensions/shared/test/block-fixtures.js#L13-L20

You can run `cd projects/plugins/jetpack && pnpm test-extensions` to test this locally.

Aurorum:
Thanks @jeherve - I think this block is good to go, except for tests which I don't have much experience with. 

I'm a little stuck on how to get the block to display when embedded. I've drawn this from the Calendly Block tests, but it's still set as embedding the profile for me: 

```js
	test( 'display block', async () => {
		const attributes = {
		      ...defaultAttributes,
		      goodreadsId: '1176283',
		      userInput: 'https://www.goodreads.com/user/show/1176283-matt-mullenweg',
		    };
		render( <GoodreadsEdit { ...{ ...defaultProps, attributes } } /> );

		let iframe;
		await waitFor( () => ( iframe = screen.getByTitle( 'Goodreads' ) ) );

		expect( iframe ).toBeInTheDocument();
	} );
```

Other than that, I think this is ready for review now.

projects/plugins/jetpack/extensions/blocks/goodreads/editor.js
```diff
@@ -0,0 +1,4 @@
+import registerJetpackBlock from '../../shared/register-jetpack-block';
+import { name, settings } from '.';
+
+registerJetpackBlock( name, settings );
```
jeherve: We've been working on moving away from this, and instead relying on a `block.json` to handle registration. You could check any of the other Beta blocks for some examples.
This was done in #32815.

projects/plugins/jetpack/extensions/blocks/goodreads/goodreads.php
```diff
@@ -0,0 +1,55 @@
+<?php
+/**
+ * Goodreads Block.
+ *
+ * @since 1.0.0 // Replace with the correct version number.
```
jeherve: You can use `$$next-version$$` as a placeholder here, and everywhere where you need to provide a version. It will be automatically replaced during the release process.

```suggestion
 * @since $$next-version$$
```

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-goodreads.php
```diff
@@ -0,0 +1,65 @@
...
+        add_action( 'rest_api_init', array( $this, 'register_routes' ) );
+    }
+
+    /**
+     * Register endpoint route.
+     */
+    public function register_routes() {
+        register_rest_route(
+            'wpcom/v2',
+            '/getGoodreadsUserId',
```
jeherve: I would recommend sticking to lowercase in what will be a URL. This route could be `/goodreads/user-id` for example (no need for the `get` keyword since we specify how the route can be called later).

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-goodreads.php
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * Get the user ID from the author ID.
+     *
+     * @param \WP_REST_Request $request request object.
+     *
+     * @return int Goodreads user ID (or 404 if not found).
+     */
+    public function get_goodreads_user_id( $request ) {
+        $profile_id = $request->get_param( 'id' );
+        $response   = wp_remote_get( 'https://www.goodreads.com/author/show/' . $profile_id );
```
jeherve: I think we can do a few things here:

1. We can build the URL separately, and escape it with `esc_url_raw` when we pass it to `wp_remote_get`.
2. We can validate the `id` param as part of the route registration, with `validate_callback`.

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-goodreads.php
jeherve, line 1: PHPCS prefers tabs over spaces, so you would need to update this file accordingly.

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-goodreads.php
```diff
@@ -0,0 +1,65 @@
...
+
+        if ( is_wp_error( $response ) ) {
+            return 404;
+        }
+
+        $body    = wp_remote_retrieve_body( $response );
+        $pattern = '/goodreads\.com\/user\/updates_rss\/(\d+)/';
```
jeherve: In this pattern, is it possible to define anchors for the beginning and end of line, and is it possible to be more specific with the domain? Basically we would want to avoid matching on `https://myevildomain.com/goodreads.com/user/updates_rss/12345`, for example.

projects/plugins/jetpack/extensions/blocks/goodreads/edit.js
```diff
@@ -0,0 +1,439 @@
...
+} from '@wordpress/components';
+import { Component } from '@wordpress/element';
+import { __, _x } from '@wordpress/i18n';
+import BlockStylesSelector from '../../shared/components/block-styles-selector';
+import testEmbedUrl from '../../shared/test-embed-url';
+import defaultExample from './default-example.png';
+import gridExample from './grid-example.png';
+import icon from './icon';
+
+export class GoodreadsEdit extends Component {
```
jeherve: In this repo we tend to prefer to implement all new components as [function components](https://react.dev/learn/your-first-component), using [hooks](https://react.dev/reference/react) to manage component state and lifecycle. Could you make the switch?

jeherve (changes requested):
Thanks for tackling this! I left a few remarks below:

********

The Grid view doesn't seem to display a grid in the editor, nor on the frontend:

Screenshot 2023-10-10 at 17 14 26

********

Those previews here are quite big, I think we would benefit from buttons instead, to reclaim that sidebar space:
image

********

> I'm a little stuck on how to get the block to display when embedded.

I see you're on the right track with those tests, apart from some formatting issues with the fixtures! Let's run those tests after the updates above, see how things go!

projects/plugins/jetpack/extensions/blocks/goodreads/edit.js
Aurorum: Sorry @jeherve, I've tried pretty much everything to make this switch but I keep getting loads of console errors - feels like every time I fix one, another appears 🙃 Would you have any advice/do you have any examples or documentation in this repo?

I based the block of the Eventbrite embed which is why it's like this currently! Do I need to rewrite the block entirely?

Aurorum: Ah, never mind - found some ideas from the Calendly block! Never tried it before, so let me know if there are any problems, but I've rewritten it now.

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-goodreads.php
Aurorum: I don't think so because the API reads the link from the middle of the Goodreads HTML (and for that reason, we also don't need to worry about matching). That being said, I've made those changes elsewhere for handling the user's inputted link.

Aurorum:
Thanks for the review @jeherve! I've addressed the points you've raised. 

> The Grid view doesn't seem to display a grid in the editor, nor on the frontend

Is the problem the fact that the images aren't aligned? I've taken the styles directly from Goodreads, and "Grid" is the terminology which they use for it, but we could tidy up the styling on our end if that's something which you think would be beneficial. Alternatively, we could rename it. 

I mainly see this style as being something useful for sidebars where you might want to showcase a range of books without taking up much space. In such cases, "Grid" is probably just about a fitting description. 

> Those previews here are quite big, I think we would benefit from buttons instead, to reclaim that sidebar space

Ah, I wanted to ask about that. I agree that the previews are far too big, but I still feel like the `BlockStylesSelector` component is appropriate here since it's also an embed. Instead, I wonder if the component is broken? I remember it used to look a lot more aesthetically pleasing. 

For example, on the Eventbrite block, it now looks like this. 

Screenshot 2023-10-11 at 21 27 26

In #14528, I can see that it used to look like this. Was changing this an intentional design choice or the result of a bug?

Screenshot 2023-10-11 at 21 30 01

I have a fondness to the second design, and I think that it'd make this block look much nicer too. It'd also be clearer which image applies to which style. I don't mind changing this to buttons though, and I definitely agree it'd be preferable to what it is now.

I just think that buttons would also add a layer of complexity for those trying to experiment with the two styles, and something like the old `BlockStylesSelector` design would be preferable and neat.

projects/plugins/jetpack/extensions/blocks/goodreads/index.js
jeherve, line 1: What's nice about the `block.json` implementation is that we do not need that file anymore. You can check our different Beta blocks for some examples.
You can entirely remove the file.

projects/plugins/jetpack/extensions/blocks/goodreads/goodreads.php
jeherve, line 1: Still on the topic of registering the block via `block.json`, you can check #32815 to see the few changes you would need to make to this file to full rely on `block.json`.

projects/plugins/jetpack/extensions/blocks/goodreads/block.json
```diff
@@ -0,0 +1,85 @@
...
+			"default": "default"
+		},
+		"userInput": {
+			"type": "string"
+		},
+		"widgetId": {
+			"type": "number"
+		}
+	},
+	"editorScript": "file:../editor.js"
```
jeherve: You can remove this. When we register the block via `block.json`, we do not have to also declare this since we still enqueue the JavaScript on the page via the existing big JavaScript bundle of Jetpack blocks.

projects/plugins/jetpack/extensions/blocks/goodreads/edit.js
```diff
@@ -0,0 +1,432 @@
+import React, { useState, useEffect, useRef } from 'react';
```
jeherve: You shouldn't have to import `React` at all here. `useState`, `useEffect`, `useRef` can be imported from a @wordpress package instead (`@wordpress/element`)

projects/plugins/jetpack/extensions/blocks/goodreads/edit.js
```diff
@@ -0,0 +1,432 @@
...
+
+const GoodreadsEdit = props => {
+	const [ userInput, setUserInput ] = useState( '' );
+	const [ displayPreview, setDisplayPreview ] = useState( false );
+	const [ isResolvingUrl, setIsResolvingUrl ] = useState( false );
+	const prevPropsRef = useRef( null );
+
+	useEffect( () => {
+		setUrl( props.attributes.userInput );
+	}, [ props.attributes.userInput ] );
```
jeherve: Do not hesitate to [deconstruct](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) `props` at the top of your function. It will make it easier to refer to the different constants afterwards. Here is an example:
https://github.com/Automattic/jetpack/blob/260c45af538dab84b7ccdbad52b869c4c0911efb/projects/plugins/jetpack/extensions/blocks/calendly/edit.js#L19-L20

projects/plugins/jetpack/extensions/blocks/goodreads/edit.js
jeherve, line 1: I think you could extract the logic that extracts data for the embed into your own [custom hook](https://react.dev/learn/reusing-logic-with-custom-hooks#extracting-your-own-custom-hook-from-a-component) in a new file. It would make the whole thing more readable and easy to understand.

You can find a number of examples in the repo, with file names that start with `use-`, if that's helpful!

jeherve (commented):
Nice work so far!

I ran the different GitHub checks we have for this PR, so you now have some additional feedback about your work. I also left a few notes below.

> Is the problem the fact that the images aren't aligned? I've taken the styles directly from Goodreads, and "Grid" is the terminology which they use for it, but we could tidy up the styling on our end if that's something which you think would be beneficial.

Yes, maybe we could benefit from aligning the images so it looks more like a grid?

> I agree that the previews are far too big, but I still feel like the BlockStylesSelector component is appropriate here since it's also an embed. Instead, I wonder if the component is broken? I remember it used to look a lot more aesthetically pleasing.

You're right, it does look broken. I believe it got messed up during a recent Gutenberg update.

To solve this issue for now, I wonder if there would be a way to ditch our custom `BlockStylesSelector`, and instead use Gutenberg's own `registerBlockStyle`:
https://developer.wordpress.org/block-editor/reference-guides/block-api/block-styles/

projects/plugins/jetpack/extensions/blocks/goodreads/index.js
Aurorum: Ah, nice - all done! Out of interest, do I not need to worry about the strings being translatable?

projects/plugins/jetpack/extensions/blocks/goodreads/edit.js
Aurorum: I kept getting a Lint error with this, but I've refactored this a little to make it more readable - let me know what you think!

Aurorum:
Thanks @jeherve!

> Yes, maybe we could benefit from aligning the images so it looks more like a grid?

Done. :)

> To solve this issue for now, I wonder if there would be a way to ditch our custom BlockStylesSelector, and instead use Gutenberg's own registerBlockStyle

This is more complicated, I'm afraid. From what I understand, the reason why `BlockStylesSelector` was introduced related to the fact that `registerBlockStyle` doesn't allow for changing attributes - it basically just adds another className to the div. That's fine for a lot of blocks, but not embeds like this one. 

@scruffian - am I understanding this correctly? Basing my thoughts on this comment, which is consistent with my testing: https://github.com/Automattic/jetpack/pull/14383#issuecomment-576693765

I think that a decision would need to be made about `BlockStylesSelector` because, ideally, there should probably be consistency between the Goodreads, Calendly and Eventbrite blocks. There might be some more options, but I think the choices broadly are:

- Adding the styles from WordPress/gutenberg#16873 (which were subsequently removed in favour of buttons) to Jetpack
- Refactoring `BlockStylesSelector` to use buttons and mimic Gutenberg's behaviour; I say 'mimic', since we can't just use Gutenberg because the embed blocks require changing the style attribute. 
- Adding a more simple dropdown Setting to allow switching between the styles like any other attribute

Aurorum:
Hey @jeherve, just looping back to this old one. If you have some time, I've heavily re-written this block. It now: 

- Places all the logic inside a hook so that it's easier to read
- Separates the controls from the rest of the block, also for readability
- Includes proper tests for the settings of the block (rather than just making sure anything renders)
- Removes the legacy widget 

I still can't register the two options as an official style because of the fact that it needs to be treated like an attribute. But I've resolved this by adding the icons to the Toolbar, like the Related Posts block does.

Screenshot 2023-12-04 at 13 15 33

Overall, I'm hoping that this works and looks better - both within the code and within the Editor. :) Thanks!!

projects/plugins/jetpack/extensions/blocks/goodreads/edit.js
Aurorum: Done!

scruffian:
> @scruffian - am I understanding this correctly? Basing my thoughts on this comment, which is consistent with my testing: https://github.com/Automattic/jetpack/pull/14383#issuecomment-576693765

That sounds right but it's been a long time since I worked on it!

Aurorum:
Tests should be fixed now

Aurorum:
Hey @jeherve - just looping back to this old one. I can see that the tests are passing except for the WordPress.com Tests. However, I don’t seem able to access the logs and see why they’re failing.

Are there any changes needed from my end?

jeherve:
@Aurorum I think we're getting there! Thanks for working on that! Everything works well for me, except for the number of items. Does this work for you? When I refresh the editor, the number resets back to 5.

Aurorum:
Whoops, that was just me registering the wrong type - not sure how I missed that! Should be fixed now. :) 

Thanks for taking another look!

jeherve (approved):
This looks good. Let's merge that first version!

jeherve:
Related diff: D138696-code

PR #33397, opened 2023-09-29
Add Jetpack Modules redux store to shared utils

andrii-lysenko:
Fixes https://github.com/Automattic/vulcan/issues/12

## Proposed changes:

In order to have a consistent state of Modules statuses across Monorepo I think it would be beneficial to have a Modules Redux store. This eventually could replace (or enhance) hook `useModuleStatus` and give more information. 

- Creating Redux store for general information on Jetpack modules inside `shared-extension-utils`
- Use selector and `dispatch` action to control the module state within Form package (and block)
- Update `withAiAssistantComponents` HOC to use `useIsPossibleToExtendJetpackFormBlock` hook to have reactive data

### Other information:

## Jetpack product discussion

* Epic: #32401
## Testing instructions:

- Clone this branch locally. You can run it with your Jurassic Tube setup.
- In `/wp-admin/admin.php?page=jetpack_modules` verify that Contact Form module is disabled
- Create new post `/wp-admin/post-new.php` and add new block

Screenshot at Oct 11 11-48-04

- Verify that no AI assistant present, when block is disabled
- Activate module. Verify that straight away the AI assistant is available

Screenshot at Oct 11 14-33-46

- Check any additional implications you could think of with the module or AI assistatnt


sergeymitr:
Hi @andrii-lysenko.
The Connection initial state issue was caused by the missing `window.JP_CONNECTION_INITIAL_STATE` variable.
I went ahead and [added](https://github.com/Automattic/jetpack/pull/33397/commits/84b9fcf1b9005edde36db387bb6d7432a49d700b) the "globals" initialization into the client tests config. Looks like it caused some conflicts with the JS-packages tests, I'll try to fix that.

andrii-lysenko:
@sergeymitr Thanks! Please let me know if there's something I need to do as well)

sergeymitr:
The Connection initial state tests seem to be working fine now.
There are a bunch of other errors though, feel free to ping me if you need help there.

andrii-lysenko:
@sergeymitr Thanks for all the help. Eventually, it seems that the problem was in my code, a bit unexpected one. Though it should be fixed now.

jeherve (commented):
I gave it a try, and it seems to test well on self-hosted. 👍

This doesn't seem to handle WordPress.com Simple just yet, where the contact form module is always active and that logic doesn't need to run.

**********

Is there something we could do to avoid this kind of errors?

image

jeherve:
Since I'm here; maybe we can take this opportunity to update the messaging a bit?

```diff
diff --git a/projects/packages/forms/src/blocks/contact-form/components/jetpack-contact-form-placeholder.js b/projects/packages/forms/src/blocks/contact-form/components/jetpack-contact-form-placeholder.js
index fd443ef92e..609f8ad9b7 100644
--- a/projects/packages/forms/src/blocks/contact-form/components/jetpack-contact-form-placeholder.js
+++ b/projects/packages/forms/src/blocks/contact-form/components/jetpack-contact-form-placeholder.js
@@ -11,7 +11,7 @@ export const ContactFormPlaceholder = ( { changeStatus, isLoading, isModuleActiv
 		<Placeholder
 			icon={ settings.icon }
 			instructions={ __(
-				"You'll need to activate the Contact Form plugin to use this block.",
+				'You’ll need to activate the Contact Form feature to use this block.',
 				'jetpack-forms'
 			) }
 			label={ settings.title }

```

andrii-lysenko:
>Is there something we could do to avoid this kind of errors?

@jeherve Yes, I also noticed this. It also seem to me that it's not something added in this PR, but it occurred to me on Trunk as well (for other redux stores). It seems that we load some packages twice and it's something we would definitely would need to fix. I just assume it would be a separate task.
It also can be noticed in `source` dev tab in browser

Screenshot at Oct 18 10-05-37

CGastrell:
Could use a .md with docs and examples/usage

andrii-lysenko:
@CGastrell Thanks for pointing this out. Added basic Readme. However, we might change it a bit in the future, providing a more accessible experience utilizing the hooks.

jeherve:
> This doesn't seem to handle WordPress.com Simple just yet, where the contact form module is always active and that logic doesn't need to run.

I'm afraid this is still an issue on WordPress.com Simple:

image

andrii-lysenko:
@jeherve oh, I haven't got this from the first time. Thanks, I'll take a look.

andrii-lysenko:
@jeherve Sorry it took so long, due to ongoing support rotation. I pushed I believe a fix for this situation. The only thing does Atomic also considered simple? I'm wiring my logic around `isSimpleSite()` which I hope should be sufficient.

projects/js-packages/shared-extension-utils/src/modules-state/actions.js
```diff
@@ -42,6 +43,10 @@ export function* updateJetpackModuleStatus( settings ) {
  * @returns {object} - an action object.
  */
 export function* fetchModules() {
+	// We don't fetch modules for Simple Site and aknowledge that all modules are active
+	if ( isSimpleSite() ) {
+		return true;
```
jeherve, line 48: Shouldn't it return an object instead, to match the expected return?

jeherve (commented):
> The only thing does Atomic also considered simple? I'm wiring my logic around isSimpleSite() which I hope should be sufficient.

`isSimpleSite()` should be enough indeed! That's also what we do in `useModuleStatus`:
https://github.com/Automattic/jetpack/blob/ef15d7e34a6ca8c5a1650f81052e4a78b2c47b55/projects/js-packages/shared-extension-utils/src/hooks/use-module-status/index.js#L47-L50

projects/js-packages/shared-extension-utils/src/modules-state/actions.js
andrii-lysenko: Oh, good catch. Though, it's error in JSDoc, it should `yield` object but returns `bool`.

jeherve (approved):
This tests well for me. 🚢 !

PR #33403, opened 2023-09-29
Blogroll: Hide subscribe button on atomic

### LINKED ISSUE #82346: Hide subscribe buttons on non WPCOM environments
We have lots of issues with nonce verification on non-wpcom env.
`pluggable.php` hosted on atomic sites generates nonce with different values. 
There is a difference between the session token and nonce tick.

Let's render the form and buttons only when running in wpcom for now.

We can check in PHP `if ( defined( 'IS_WPCOM' ) && IS_WPCOM )`

renancarvalho:
Fixes https://github.com/Automattic/wp-calypso/issues/82346

## Proposed changes:

* Hide the subscribe button on Atomic sites for now, this may get reverted in the future once we add atomic site support.

### Other information:

## Testing instructions:

- Pull and run this branch
- Open a simple site that uses blogroll and you should see the subscribe button available 
- Open an atomic site that uses blogroll and the buttons should not be visible.

Atomic:
image

Simple:
image

### DIFF
--- a/projects/plugins/jetpack/changelog/Hide subscribe buttons on atomic sites
+++ b/projects/plugins/jetpack/changelog/Hide subscribe buttons on atomic sites
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
+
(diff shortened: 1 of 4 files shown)


agrullon95 (approved):
Working well. The subscribe button appears on simple site and not on Atomic /jetpack site.

PR #33405, opened 2023-09-29
Add lint configs for VS Code

ice9js:
## Proposed changes:

This adds VS Code configuration for PHPCS and Prettier so their respective extensions use the correct executables inside the monorepo when active.

Extensions 

### Other information:

## Jetpack product discussion

pf4qpu-3P-p2

N/A

## Testing instructions:

You'll need to have the following extensions installed:

- [PHPCS](https://marketplace.visualstudio.com/items?itemName=shevaua.phpcs)
- [Prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode)

It might also be helpful to install the [ESLint extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) even though that one seems to work well out of the box.

- Open any PHP file inside the monorepo and make some invalid changes to the syntax. The errors should be highlighted and you should see the related explanation from PHPCS when hovering over them.
- Open any JS file from inside the monorepo and make invalid syntax changes too. If you have the ESLint extension, it'll highlight the errors.
- In any JS file, you should also be able to run the `Format Document` which will run prettier for just that file, or your selection.

Note: prettier should be using our `wp-prettier` fork installed in the monorepo and document should be formatted according to WP coding standards, not the prettier default.


.vscode/settings.json
```diff
@@ -19,5 +19,16 @@
 	"**/.cache/**",
 	"**/jetpack_vendor/**/jetpack_vendor/**",
 	"**/jetpack_vendor/**"
-	]
+	],
+	"phpcs.enable": true,
```
jeherve: I think this may only be useful for folks using the phpcs VS code extension, and not the `PHP_CodeSniffer` VS code extension ([reference](https://marketplace.visualstudio.com/items?itemName=obliviousharmony.vscode-php-codesniffer)).

.vscode/settings.json
```diff
@@ -19,5 +19,16 @@
...
+	"phpcs.executablePath": "vendor/bin/phpcs",
+	"phpcs.ignorePatterns": [
+		"**/.git/**",
+		"**/.svn/**",
+		"**/.hg/**",
+		"**/.cache/**",
+		"**/jetpack_vendor/**",
+		"**/vendor/**",
+	],
+	"prettier.prettierPath": "node_modules/.pnpm/wp-prettier@2.8.5/node_modules/wp-prettier/index.js"
```
jeherve: This may be risky, since that exact path may change as we update the package, and we'll have to remember to edit this line every time.

Looking at my own VS Code config, I do not seem to have `prettierPath` set, and it still works. I'm not sure if I did anything special to make things work :) I only have `configPath` set it seems.

ice9js: Hmm, so it will pick up `prettier` if you have it installed globally, which could be a problem in itself, especially since we want to be using our `wp-prettier` fork and not the original library.

anomiex: If we need a path, `node_modules/.bin/prettier` should work.

Unless of course vscode is assuming it's pointed at a JS file rather than an arbitrary executable. In that case maybe `tools/js-tools/node_modules/prettier/index.js`?

ice9js: I think there's an issue with the extension and that version of prettier, can't remember exactly, I'll need to track it down again to get the full context. But the gist is, I tried just pointing it at the module and it doesn't work, pointing it at `index.js` directly is a workaround but gets the job done.

The integration goes a little deeper than just executing a CLI command, hence why it's necessary.

anomiex: So, `tools/js-tools/node_modules/prettier/index.js` then?

sdixon194 (approved):
Perfect, this works for me!

PR #33408, opened 2023-09-29
AI Assistant: do not register plugins when not connected

jeherve:
## Proposed changes:

The AI Assistant editor plugins rely on a working connection to WordPress.com, so we should not see them when in offline mode for example.

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

* Start with a fresh JN site, running Bleeding edge.
* Go to Settings > Jetpack Constants
* Enable Offline mode
* Go to Posts > Add New
    * See the AI Assistant panel in the Jetpack plugin sidebar.
* Now switch to this branch
* Visit the post editor again
    * The AI Assistant panel should be gone.
* Go back to Settings > Jetpack Constants
* Disable Offline mode
* Connect your site to WordPress.com.
* Go to Posts > Add New
    * The AI Assistant panel should be there now.

**You will also want to make sure the feature remains available on WordPress.com Simple sites.**


monsieur-z (approved):
PR tested as instructed and AI panel is still visible on WPcom simple site. 🚢 

Not related to this PR (cc. @Automattic/jetpack-agora):
- The UI inside the panel is confusing as the disabled button doesn't look like a button.
- What are the conditions for the button to be enabled?


Screenshot 2023-10-10 at 9 54 46 AM

retrofox:
I have yet to check the PR, but registering the block even when the site is not connected is to increase its visibility. Under these conditions, the block should show a `Connect` button to help the user connect to the site. 

Is it possible that we could lose this behavior with these changes?

jeherve:
> Is it possible that we could lose this behavior with these changes?

@retrofox No, because the AI Assistant block is still registered when not connected. 
https://github.com/Automattic/jetpack/pull/33408/files#diff-88a75f3f147a74493e49bbc8071d4e37077fef15bb02b3f8471a8f0281535246R28

I only stopped registering the things that did not have a connect button when the site is not connected.

Do not hesitate to give the PR a try, and let me know if it behaves like you would expect!

dhasilva:
> What are the conditions for the button to be enabled?

There must be content, any content. It is only disabled when there is nothing in the post, IIRC.

retrofox:
I don't see the AI Assistant block when the site is disconnected:

image

I wonder if it's related to this PR, though.

PR #33285, opened 2023-09-25
Add .vscode/settings.json to Repo

sdixon194:
Adds a .vscode/settings.json file to the repo to add opinionated VS Code settings. 

## Proposed changes:

Adds a .vscode/settings.json file with an exclude list of intelephense so developers using the jetpack repo with VSCode will be able to properly use intelephense without causing performance problems. 

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

pf4qpu-3P-p2

## Does this pull request change what data or activity we track or use?

No.
## Testing instructions:

Not sure how to test this, but I suppose you could nuke your own VS Code Intelephense settings and checkout this PR and see if things continue to work.

PR #33412, opened 2023-09-29
Fix: protect wpcom urls from search and replace

bindlegirl:
IDC logic stores wpcom URLs to the database and uses them to display Safe Mode. Users sometimes use search & replace plugins to alter URLs on cloned sites and rewrite these values which causes domain change on wpcom side.

## Proposed changes:

* Using strrev() to reverse stored domains but keep them readable for debugging purposes before storing sync_error_idc option.
* Add a method to reverse options for comparison.
* Used `jetpack-options` filter to apply reversing and re-reversing
* Updated tests

### Other information:

## Jetpack product discussion

pf5801-3R-p2

## Testing instructions:

* You will need two JN sites with connected Jetpack.
* Use Jetpack beta to load this branch.
* Clone database from site A and import it to the site B. 
*  It should trigger Safe Mode. If not, try opening the Jetpack dashboard or My Jetpack.
* Ignore Safe Mode and install a search and replace plugin such as[ Better search replace](https://wordpress.org/plugins/better-search-replace/). Use the plugin to update all mentions of URL A with URL B in wp-options table.
* Safe Mode should still show up.
* Check audit logs and debug to make sure URLs didn't change.

## Alternative testing

* Create JN site, connect it and run this branch
* Run IDC tool from debugger tools
* Once you trigger Safe Mode, check the database for jetpack_sync_error_idc option. It should contain reversed wpcom values.


kraftbj:
I didn't test this, but wanted to say I think it's a clever solution. Simple and still possible to parse the value as a human.

.vscode/settings.json
anomiex, line 1: This probably shouldn't be being deleted. It was intentionally added in #33285.

bindlegirl: That's odd. That should be an untracked file. I added some custom stuff in there and had to remove it from cache but it shouldn't get commited as deleted. Not sure how to  undelete it now since it can't get commited.

bindlegirl: I think I undeleted it successfully now

anomiex: :shrug: @sdixon194 @zinigor Is there a way to have settings like pf4qpu-3P-p2 wants while also allowing people to set their own settings? Or should people set their settings somewhere else?

sergeymitr (approved):
Works great! 👍

.vscode/settings.json
sdixon194: Yeah the settings.json file in the root level .vscode is what VSCode will use. It will override the same settings in the user level settings.json (ctrl+cmd+p -> Open User Settings). You can specify any additional settings in your user settings, and as long as they're not the same as the ones specified in the root level workspace settings.json, should work fine.

Right now the only thing they should be opinionated about is the list of files to exclude from intelephense.

sdixon194: Ah I did some testing and if you have other workspace settings, it will overwrite them. 

This could be problematic, yeah.

bindlegirl: I solved my problem by using `git update-index --assume-unchanged .vscode/settings.json` 
Now I can add custom stuff to my settings file without pushing those changes to the repo.

I assume local settings will be overwritten when this file is updated but I don't expect this happening too often, so I'll just paste my changes back in when needed.

PR #33421, opened 2023-09-29
CRM: Resolves 3336, 3350, and 3351 - PHP 8.2 deprecation notices

tbradsha:
Resolves Automattic/zero-bs-crm#3336, Automattic/zero-bs-crm#3350, and Automattic/zero-bs-crm#3351 - PHP 8.2 deprecation notices

## Proposed changes:

This catches several deprecation notices that appear in PHP 8.2, of various types:

* Dynamic class properties: in most cases the fix was to declare them, and in a few cases they weren't used so they were removed
* Errant assignment of an array key/value to a `bool`: we now make this an empty array before assigning a key/value
* Errant passing of `null` to `stripslashes()`: we don't have details on how this was triggered (something in their settings), but I added a fallback to prevent it

---

I also updated Codeception's configuration to show errors in its output as relevant, and patched a deprecation in its `RunProcess.php` extension (we can't update, as Codeception 5 requires PHP 8). However, I believe this only affects PHP errors that are generated by the test code rather than by the pageloads it's testing, so unfortunately this doesn't help us too much.

It'd be nice if we found a way to get Codeception to output errors when deprecation notices popped up on our pages.

### Other information:

## Testing instructions:

Make sure you're on PHP 8.2.

In `trunk`, bounce around to each page (including new, edit, and view pages). You'll see deprecation notices throughout, even on non-CRM pages.

In `fix/crm/3336+3350+3351-PHP_deprecation_notices` the deprecation notices are all gone.

### DIFF
--- a/projects/plugins/crm/includes/ZeroBSCRM.GeneralFuncs.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.GeneralFuncs.php
@@ -157,9 +157,9 @@ function zeroBSCRM_stripSlashesFromArr( $value ) { // phpcs:ignore WordPress.Nam
 		return array_map( 'zeroBSCRM_stripSlashesFromArr', $value );
 	} elseif ( is_string( $value ) ) {
 		return stripslashes( $value );
-	} else {
-		return $value;
-	}
+	} 
+	return $value;
+
 }
 
    # from http://wordpress.stackexchange.com/questions/91900/how-to-force-a-404-on-wordpress
--- a/projects/plugins/crm/includes/ZeroBSCRM.GeneralFuncs.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.GeneralFuncs.php
@@ -141,15 +141,26 @@ function zerobscrm_doing_it_wrong( $function, $message, $version ) {
 
 	}
 
-
-	function zeroBSCRM_stripSlashesFromArr($value){
-	    $value = is_array($value) ?
-	                array_map('zeroBSCRM_stripSlashesFromArr', $value) :
-	                stripslashes($value);
-
-	    return $value;
+/**
+ * Recursively strips slashes from an array.
+ *
+ * Quite similar to zeroBSCRM_stripSlashes(), but that given
+ * its legacy status any changes to there and its returns may
+ * have negative implications
+ *
+ * @param array|string|null $value Some value to strip.
+ *
+ * @return array|string|null
+ */
+function zeroBSCRM_stripSlashesFromArr( $value ) { // phpcs:ignore WordPress.NamingConventions.ValidFunctionName.FunctionNameInvalid
+	if ( is_array( $value ) ) {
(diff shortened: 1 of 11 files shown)


projects/plugins/crm/includes/ZeroBSCRM.GeneralFuncs.php
```diff
@@ -141,15 +141,26 @@ function zeroBSCRM_stripSlashes($obj='',$return=true){
...
+ * @return array|string|null
+ */
+function zeroBSCRM_stripSlashesFromArr( $value ) { // phpcs:ignore WordPress.NamingConventions.ValidFunctionName.FunctionNameInvalid
+	if ( is_array( $value ) ) {
+		return array_map( 'zeroBSCRM_stripSlashesFromArr', $value );
+	} elseif ( is_string( $value ) ) {
+		return stripslashes( $value );
+	} else {
+		return $value;
 	}
```
coder-karen: Could this just be stripped down to the following:
```suggestion
	} 
	return $value;

```

coder-karen (commented):
Works well for me! All the deprecation notices I could replicate were removed, which was all but the stripslashes notices.

I added one suggestion.

I did then find a new notice, but have reported that one here: https://github.com/Automattic/zero-bs-crm/issues/3359 

I thought the issue may be related to this PR due to the removal of these two lines in `ZeroBSCRM.Edit.php`:
```
            'tag' => false,         // Discontinued v3.0 + zerobs_transactiontag
            'postType' => false,    // Discontinued v3.0 + //zerobs_transaction
```

However I can also replicate this in trunk.

It looks like adding `private $postType;` within the `ZeroBS__Metabox_LogsV2` class would fix that, like you've done elsewhere (eg. `ZeroBSCRM.MetaBoxes3.TagManager.php`), so likely worth adding to this PR?

cleacos (dismissed):
Other than Karen's comment... it worked for me.

I still see Client Portal Pro deprecation notices for the endpoints, but I recall that that issue is handled in another PR.

projects/plugins/crm/includes/ZeroBSCRM.GeneralFuncs.php
tbradsha: Yep, fine by me!

tbradsha:
Just needs a reapproval now.

@coder-karen:
> I did then find a new notice, but have reported that one here

Usually I'd just fix it in this PR, but since it already spans 3+ issues, I don't want to further bloat this one. I'll fix it separately. :^)

I accepted your code suggestion in da06d13, but it added a coding standards violation (fixed in 4f0624086c).

@cleacos:
> I still see Client Portal Pro deprecation notices for the endpoints, but I recall that that issue is handled in another PR.

Indeed, @gogdzl is working on that one: Automattic/zero-bs-crm#3357

cleacos (dismissed):
LGTM!

coder-karen (approved):

PR #33416, opened 2023-09-29
CRM Automations UI: Add TriggerInfo and use next_step_true

robertf4:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3354

## Proposed changes:

image

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

NtYgueI8oAgiWGHTbHpnjh-fi-1_29

## Does this pull request change what data or activity we track or use?

No

## Testing instructions:

1. Create a workflow with several different steps. For example you can run this on your site:
```
		$workflow = new \Automattic\Jetpack\CRM\Automation\Automation_Workflow(
			[
				'name'         => 'New contact: Set status to "Customer"',
				'description'  => 'This automation will change the status of a contact to "Customer" when they are created.',
				'category'     => 'contact',
				'active'       => true,
				'triggers'     => [
					\Automattic\Jetpack\CRM\Automation\Triggers\Contact_Created::get_slug(),
				],
				'initial_step' => 1,
				'steps'        => [
					1 => [
						'slug'       => \Automattic\Jetpack\CRM\Automation\Actions\Update_Contact_Status::get_slug(),
						'attributes' => [
							'new_status' => 'Customer',
						],
						'next_step_true'  => 2,
					],
					2 => [
						'slug'       => \Automattic\Jetpack\CRM\Automation\Actions\Update_Contact_Status::get_slug(),
						'attributes' => [
							'new_status' => 'Customer',
						],
						'next_step_true'  => 3,
					],
					3 => [
						'slug'       => \Automattic\Jetpack\CRM\Automation\Actions\Update_Contact_Status::get_slug(),
						'attributes' => [
							'new_status' => 'Customer',
						],
						'next_step_true'  => null,
					],
				],
			]
		);

		$repo = new \Automattic\Jetpack\CRM\Automation\Workflow\Workflow_Repository();
		$workflow = $repo->persist(  $workflow );
```
2. Visit `/wp-admin/admin.php?page=jpcrm-automations#/automations/` and click "Edit" on that workflow.
3. Verify that the trigger info displays at the top of the edit modal.
4. Verify that all steps display in the edit modal.

PR #33422, opened 2023-09-29
CRM Automations UI: Add workflow saving and reloading

robertf4:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3203

## Proposed changes:

This PR implements the saving and reloading of workflows.

### Other information:

## Jetpack product discussion

No

## Testing instructions:

Be sure you have test workflows on your site. If not then see this PR for instructions on making some: https://github.com/Automattic/jetpack/pull/33416.

1. Visit `/wp-admin/admin.php?page=jpcrm-automations#/automations`.
2. Toggle the active toggles on some of your workflows. Hard refresh the page to verify that the state is saved.
3. Select some workflows and use the bulk action to change their active state. Hard refresh the page to verify that the state is saved.
4. Click "Edit" for a workflow and change some attributes, then press "Save".  Hard refresh the page to verify that the state is saved.
5. Click "Edit" for a workflow, change some attributes, then press "Cancel". Now press "Edit" again and verify that your changes are gone. Hard refresh the page and click "Edit" again to verify that your changes were not saved.


projects/plugins/crm/src/js/components/automations-admin/components/attribute-config/index.tsx
```diff
@@ -61,6 +61,7 @@ const getEditValue = (
 							  } ) )
 							: []
 					}
+					value={ value }
```
robertf4, line 64: This just fixes a bug where on load the SelectControl would display the first option instead of the option selected in the store.

projects/plugins/crm/src/js/components/automations-admin/components/bulk-workflow-actions/index.tsx
```diff
@@ -1,26 +1,61 @@
 import { DropdownMenu, MenuItem } from '@wordpress/components';
-import { dispatch } from '@wordpress/data';
+import { useSelect } from '@wordpress/data';
+import { __ } from '@wordpress/i18n';
```
robertf4, line 3: There were a few non-internationalized strings in here, so I went ahead and corrected that.

projects/plugins/crm/src/rest-api/v4/class-rest-automation-workflows-controller.php
```diff
@@ -152,9 +152,22 @@ public function register_routes() {
 	 * @return WP_Error|WP_REST_Response
 	 */
 	public function get_items( $request ) {
+
+		// If $params['active'] is true then only active workflows should be returned.
```
robertf4: I changed this behavior because otherwise it is only possible to request either active workflows or inactive workflows, but not both together. The UI requires both together. Since the description of the `active` param is "Whether to return only active workflows", I thought this behavior seems to fit.

projects/plugins/crm/src/js/state/automations-admin/util.ts
```diff
@@ -1,26 +1,15 @@
...
- * Finds a step in a Workflow by ID.
+ * Gets a workflow which has been prepared for sending to the server.
  *
- * @param {Workflow} workflow - The workflow to search for the step
- * @param {string} stepId - The ID of the step to search for
- * @returns {object} An object containing the step, its previous step, and its next step if they exist, or undefined otherwise
+ * @param {Workflow} workflow - The workflow to prepare
+ * @returns {ServerPreparedWorkflow} The prepared workflow
  */
-export const findStep = ( workflow: Workflow, stepId: string ) => {
```
robertf4, line 10: I deleted this along with its corresponding tests as it isn't being used.

projects/plugins/crm/src/js/components/automations-admin/components/bulk-workflow-actions/index.tsx
```diff
@@ -1,26 +1,61 @@
...
-		onClose();
-	};
+	const onActivateClick = useCallback(
+		( onClose: () => void ) => () => {
+			for ( const workflowId of selectedWorkflows ) {
+				const workflow = workflows[ workflowId ];
+				mutateWorkflows(
+					{ ...workflow, active: true },
+					{
+						onSuccess: () => refetchWorkflows(),
```
robertf4, line 35: This looks like it would refetch workflows after each individual workflow in a bulk action is updated, but it doesn't due to `react-query` magic.

projects/plugins/crm/src/js/components/automations-admin/components/edit-modal/index.tsx
```diff
@@ -23,15 +30,17 @@ export const EditModal: React.FC< EditModalProps > = ( { isOpen, onClose, workfl
 		stepId = step.next_step_true;
 	}
 
+	const { mutate: mutateWorkflows } = useMutateAutomationWorkflows();
+
 	const onSave = useCallback( () => {
-		// TODO: save workflow here
+		mutateWorkflows( workflow );
```
robertf4, line 37: If I refetch here on success then the state doesn't change. If I refetch on failure then the user's changes are lost with no indication that it happened. Thus this save will fail silently until we implement error notifications.

projects/plugins/crm/src/rest-api/v4/class-rest-automation-workflows-controller.php
kallehauge: I made a slight modification to this approach.
This specific code has been reverted to what it was before, but the `active` API endpoint argument has been made optional (read: we've removed the `default` value), so unless anything else is specified, then it will return both active and inactive workflows

projects/plugins/crm/src/js/components/automations-admin/components/bulk-workflow-actions/index.tsx
kallehauge: giphy (32)

kallehauge (approved):

PR #32651, opened 2023-08-23
Subscriber block: present large subscriber numbers more elegantly

lezama:
Part of #32373

## Proposed changes:

Present large numbers more elegantly in subscriber block.

#### Before

Screenshot 2023-08-23 at 19 17 21

#### After

Screenshot 2023-08-23 at 19 16 46

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* Check that Subscribe widget and Subscribe blocks still work and look ok with the new format :)
* Sandbox sites with different amount of subscribers
* A site with millions: https://wordpress.com/blog/2009/11/25/blog-subscriptions/
* A site with thousands: https://jetpack.com/blog/
* A site with less than a thousand: <your test site>

PR #33430, opened 2023-10-01
Subscription block: drop unnecessary .0 from big subscriber counts

simison:
Follow-up to https://github.com/Automattic/jetpack/pull/32651

## Proposed changes:

* Drop unnecessary .0 from big subscriber counts 

image

### Other information:

## Testing instructions:

* Sandbox site with big subscription count, e.g. en.blog or mock the counter at `get_join_others_text()`
* See the subscriber count

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-subscriber-count-zero
+++ b/projects/plugins/jetpack/changelog/fix-subscriber-count-zero
@@ -1,0 +1,4 @@
+Significance: minor
+Type: enhancement
+
+Subscription block: drop unnecessary .0 from big subscriber counts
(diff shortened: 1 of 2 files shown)


kraftbj (approved):

kraftbj:
@simison I'll note that this would break for very large numbers, but I don't think that's a reason to stall the PR. :)

e.g. for a site with 1,000,000,000 subscribers.
```
wp> $subscriber_total = 1000000000
=> int(1000000000)
wp> number_format_i18n( $subscriber_total / 1000000, 1 )
=> string(7) "1,000.0"
wp> floatval(number_format_i18n( $subscriber_total / 1000000, 1 ))
=> float(1)
```

So, no sites can get a billion subscribers. Just not allowed :-P.

PR #33432, opened 2023-10-02
Jetpack AI Search: Release

artpi:
This releases Jetpack AI Search to production

## Proposed changes:
* Switch Jetpack AI Search from a beta block to production block

NO

## Testing instructions:
* Follow instructions from call for testing pdtkmj-1OZ-p2
* See that the block does not appear as beta anymore


dereksmart (approved):
Had to update the package versions. :shipit:

PR #33440, opened 2023-10-02
Jetpack Sync: Add support for supplying additional columns when doing a checksum.

bisko:
This PR fixes an issue that has existed since we shipped Checksum/Fix process. 

The process supports sending additional columns when doing a checksum, so we can fix data that has changed outside of the `post_modified_gmt` field. This happens when a customer updates the database in a way, that doesn't use the integrated WordPress APIs, e.g. direct `UPDATE wp_posts SET ...`.

The problem that we had is that while we received the columns that we wanted to add to the list of checksum columns, we never actually used that list. 

This PR updates that so now we can use the additional columns when we want to make a more comprehensive fix.

## Proposed changes:

* Actually use the supplied additional columns when making a Checksum/Fix for a Jetpack site.

## Jetpack product discussion

p9dueE-7eh-p2

No

## Testing instructions:

Testing instructions can be found in p9dueE-7UB-p2


sergeymitr (approved):
Can confirm that additional columns get passed into the SQL query and the resulting checksum reflects that.

PR #33445, opened 2023-10-03
Quick Share Buttons | Change Twitter icon to X

manzoorwanijk:
## Proposed changes:

* Add `XIcon` to `js-packages/components/components/icons`
* Replace Twitter with X in text strings in Quick Share buttons
* Fix the layout and styling of the Quick Share buttons

### Other information:

## Testing instructions:

* On the post publish panel, confirm that you see the <s>Twitter</s> X icon
* Confirm that it works as expected
* Open Jetpack sidebar
* Under "Share this post", click on the Quick Share icon
* Confirm that you see the <s>Twitter</s> X icon and the updated text.
* Click on "Learn more..." link
* Confirm the same for the modal

| Before | After |
|--------|--------|
| Screenshot 2023-10-03 at 10 44 13 AM | Screenshot 2023-10-03 at 10 42 52 AM |
| Screenshot 2023-10-03 at 10 44 54 AM | Screenshot 2023-10-03 at 10 43 24 AM |
| Screenshot 2023-10-03 at 10 44 39 AM | Screenshot 2023-10-03 at 10 43 06 AM | 

* Test the connections screen. You can change the new function to return `x` every time, then it should show the icons.

image


projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
```diff
@@ -52,19 +53,31 @@ export function ShareButtons( { buttonStyle = 'icon', buttonVariant }: ShareButt
...
+					label
+				);
+
 				return (
 					<div className={ styles.container } key={ networkName }>
 						<Button
 							icon={ icon }
 							variant={ buttonVariant }
-							aria-label={ label }
+							aria-label={ text }
```
pablinos, line 80: This is definitely better than just having the network name here, but could this cause problems with the Text component below and cause the label to be read twice? Is there a way to more closely associate the two components? Or perhaps use aria-hidden on the text?

pablinos (dismissed):
This is working ok. I had a question about the aria-label, but I don't think that's a blocker.

I've noticed that the hover state for the buttons looks the same as the new X icon. I don't know if we should do something about that? But if so it can be done in a follow-up

image

pablinos (approved):

projects/js-packages/publicize-components/src/components/share-buttons/share-buttons.tsx
manzoorwanijk: > aria-label takes precedence over the text included in the button

https://www.w3.org/WAI/WCAG22/Techniques/general/G208.html

sdixon194:
cherry picked to release branch in c37b685a972be56da1a99c2e288a680ba4c35d3d

PR #33451, opened 2023-10-03
External Media: update API access to authors and above

jeherve:
## Proposed changes:

Folks who cannot upload media do not need access to the feature, since they can only insert images using an existing URL.

### Other information:

## Jetpack product discussion

* p1696305792460289-slack-CRA4UEQQ3

* No

## Testing instructions:

In this case, you'll want to check that the feature continues to work for Authors and above. On a site that's connected to WordPress.com, 

1. Go to Posts > Add New
2. Add an image block
3. Click on the button to select an existing image
4. Try inserting an image from Pexels.
5. Try connecting the site to your Google Photos account.
6. Try disconnecting from Google if you are already connected.


sdixon194 (approved):
Works for me now 👍

PR #33458, opened 2023-10-04
Notifications: remove on block editor pages for Gutenberg compat

kraftbj:
Fixes an issue where the Notifications 'n' shortcut is bubbled up to the iframe when in the block editor, thus making it impossible to type `n`.

## Proposed changes:

* As a temporary bandaid until a real cause and resolution, remove the notifications bubble only on the block editor pages. The Site Editor does not an admin 
bar ever and the Post Editor has an option to show it, but WP defaults to not for new installs it appears.

### Other information:

## Testing instructions:

* Jetpack + Gutenberg 16.7
* Connect Jetpack and confirm that Notifications are active.
* Open the post editor and try to type "I do not like this bug"
* At the `n`, before the patch, it would not type. If the admin bar is displayed, the notifications panel would open.
* After the patch, there is no notifications item in the admin bar for only the post block editor page and you can type the full sample text.

### DIFF
--- a/projects/plugins/jetpack/modules/masterbar/masterbar/class-masterbar.php
+++ b/projects/plugins/jetpack/modules/masterbar/masterbar/class-masterbar.php
@@ -385,7 +385,7 @@ class Masterbar {
 		$this->add_reader_submenu( $wp_admin_bar );
 
 		// Right part.
-		if ( Jetpack::is_module_active( 'notes' ) && ! Jetpack_Notifications::is_block_editor() ) {
+		if ( Jetpack::is_module_active( 'notes' ) && ! \Jetpack_Notifications::is_block_editor() ) {
 			$this->add_notifications( $wp_admin_bar );
 		}
 
--- a/projects/plugins/jetpack/modules/masterbar/masterbar/class-masterbar.php
+++ b/projects/plugins/jetpack/modules/masterbar/masterbar/class-masterbar.php
@@ -385,7 +385,7 @@ class Masterbar {
 		$this->add_reader_submenu( $wp_admin_bar );
 
 		// Right part.
-		if ( Jetpack::is_module_active( 'notes' ) ) {
+		if ( Jetpack::is_module_active( 'notes' ) && ! Jetpack_Notifications::is_block_editor() ) {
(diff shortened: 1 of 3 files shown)


kraftbj:
Oh, why did I not just add a check for the block editor in the `init` hooked function? `get_current_screen` gets set up surprisingly late, so it is null even when firing the hook conditionally on `init` and `admin_init` (based on `is_admin()`). Instead of spending too much time to find an elegant DRY way, the helper function and short-circuiting the final execution vs hook setup feels like an okay compromise.

projects/plugins/jetpack/modules/masterbar/masterbar/class-masterbar.php
```diff
@@ -385,7 +385,7 @@ public function build_wpcom_masterbar( $wp_admin_bar ) {
 		$this->add_reader_submenu( $wp_admin_bar );
 
 		// Right part.
-		if ( Jetpack::is_module_active( 'notes' ) ) {
+		if ( Jetpack::is_module_active( 'notes' ) && ! Jetpack_Notifications::is_block_editor() ) {
```
ice9js: It should be `\Jetpack_Notifications` otherwise it's causing a fatal.

ice9js (approved):
Works for me 👍.

projects/plugins/jetpack/modules/masterbar/masterbar/class-masterbar.php
```diff
@@ -385,7 +385,7 @@ public function build_wpcom_masterbar( $wp_admin_bar ) {
 		$this->add_reader_submenu( $wp_admin_bar );
 
 		// Right part.
-		if ( Jetpack::is_module_active( 'notes' ) ) {
+		if ( Jetpack::is_module_active( 'notes' ) && ! \Jetpack_Notifications::is_block_editor() ) {
```
kraftbj, line 388: thanks for catching that @ice9js 🙇🏼

projects/plugins/jetpack/modules/masterbar/masterbar/class-masterbar.php
kraftbj: thanks for catching that!

PR #33456, opened 2023-10-03
Launchpad: Changed domain task visibility

nelsonec87:
Related to https://github.com/Automattic/wp-calypso/issues/80028

## Proposed changes:

* This PR changes the visibility of the domain Launchpad task.
* It makes the task always visible
* It makes the task disabled when the task is completed - has a custom domain.

## Does this pull request change what data or activity we track or use?

No

## Testing instructions:

* Apply this PR to your sandbox
* Create a new `build` site (`/start`)
* Follow the onboarding and launch the site
* Check that the `Customize your domain` is visible, enabled, and not completed;

image

* Buy a domain (remember to cancel the purchase afterward)
* Check that the task is now visible, marked as completed, and is disabled

image

PR #33462, opened 2023-10-04
Launchpad: Update copy for global styles in plan_selected task

taipeicoder:
Related to https://github.com/Automattic/wp-calypso/pull/82524

## Proposed changes:

* In wpcom, update the copy `custom` to `premium` styles in the context of global styles.

### Other information:

## Testing instructions:

* Apply this patch to your sandbox.
* Head to https://developer.wordpress.com/docs/api/console/ and select WP REST API → wpcom/v2.
* Call the endpoint `/sites/${ SITE_SLUG }/launchpad` with a site that has global styles gated (just select any theme with a non-default style variation).
* Ensure that the checklist item `plan_selected` the subtitle has changed to "Your site contains premium styles. Upgrade now to publish them and unlock tons of other features."


miksansegundo:
The copy change looks good. I'm not approving because all those files committed seem not required.

In the following screenshot I'm not using the PR with other copy changes, that's why the link still uses "custom".

Screenshot 2566-10-04 at 18 05 39

**Note:** It didn't work when testing it on my sandbox using `bin/jetpack-downloader` because the change is done in `sun` rather than `moon`. After editing the file manually in sandbox it worked fine.

miksansegundo (dismissed):
LGTM 👍 

The copy change looks good and all those files committed look good too.

In the following screenshot I'm not using the PR with other copy changes, that's why the link still uses "custom".

Screenshot 2566-10-04 at 18 05 39

**Note:** It didn't work when testing it on my sandbox using `bin/jetpack-downloader` because the change is done in `sun` rather than `moon`. After editing the file manually in sandbox it worked fine.

taipeicoder:
@jeherve I noticed that some PRs (like https://github.com/Automattic/jetpack/pull/33456) would bump the package versioning while others (like this PR, or https://github.com/Automattic/jetpack/pull/33374) won't. Is it because this PR will be bundled into the release version 4.14.0-alpha, hence it doesn't need version bumping?

jeherve:
@taipeicoder The version bump needs to match the one defined in all the changelog entries that were added since the last stable release. In this PR, the changelog entry defines a `minor` version bump:
https://github.com/Automattic/jetpack/pull/33462/files#diff-eef05f43fc068613b85deaf1382101a9cdb93c0d5a941aec4b0792245478145cR1

The current stable version of the package is `v4.13.0`:
https://packagist.org/packages/automattic/jetpack-mu-wpcom

A minor bump of that version would be `v4.14.0`, so for now, during development, the version in this repo should be `v4.14.0-alpha`.

- If you were to create a follow-up PR to this one, with `patch` defined in the changelog entry, no bump would need to occur since a patch bump (`v4.13.1-alpha`) will be negated by the minor bump already defined.
- If you were to create another follow-up PR with another `minor` bump in the changelog entry, the version wouldn't need to change again since the minor bump is already indicated.
- If you were to create another follow-up PR, this time defining `major` bump in the changelog entry, then you would be expected to bump the version to `v5.0.0-alpha`, to match the `major` bump requirement.

Does that make sense?

jeherve (dismissed):

jeherve (approved):

taipeicoder:
> @taipeicoder The version bump needs to match the one defined in all the changelog entries that were added since the last stable release. In this PR, the changelog entry defines a `minor` version bump: https://github.com/Automattic/jetpack/pull/33462/files#diff-eef05f43fc068613b85deaf1382101a9cdb93c0d5a941aec4b0792245478145cR1
> 
> The current stable version of the package is `v4.13.0`: https://packagist.org/packages/automattic/jetpack-mu-wpcom
> 
> A minor bump of that version would be `v4.14.0`, so for now, during development, the version in this repo should be `v4.14.0-alpha`.
> 
> * If you were to create a follow-up PR to this one, with `patch` defined in the changelog entry, no bump would need to occur since a patch bump (`v4.13.1-alpha`) will be negated by the minor bump already defined.
> * If you were to create another follow-up PR with another `minor` bump in the changelog entry, the version wouldn't need to change again since the minor bump is already indicated.
> * If you were to create another follow-up PR, this time defining `major` bump in the changelog entry, then you would be expected to bump the version to `v5.0.0-alpha`, to match the `major` bump requirement.
> 
> Does that make sense?

Thanks for the detailed explanation @jeherve!

PR #33476, opened 2023-10-04
CRM: Resolves #3356 - show custom transaction status filters

### LINKED ISSUE #3356: Stats + Carousel: Record individual image hits
On WordPress.com, page hits are recorded for each image viewed full-screen in the Carousel. In Jetpack, we do not collect that.

Enhancement: Collect these stats :)

Requested in 2509419-t

tbradsha:
CRM: Resolves Automattic/zero-bs-crm#3356 - show custom transaction status filters

## Proposed changes:

Transactions allow custom statuses, but the quickfilter dropdown in the listview was hard-coded. This PR modularises and lightly refactors the code that allowed custom contact and company statuses to show, and thus extends the ability to transaction statuses.

I also added a `jpcrm_add_custom_quickfilter` filter which allows one to add to the list of quickfilters in the dropdown (see testing instructions). There's not currently a direct application, so we can remove it as desired.

Note that the heavy use of globals and the existing data structures are all legacy, and are out of scope. Likewise, translation oddities and other hard-coded statuses (e.g. those in contacts and companies) are pre-existing and are unchanged by this PR. Finally, there were some linter rules about indents I ignored, as it errantly believes the code I edited should be indented.

### Other information:

## Testing instructions:

1. Add a custom transaction status: `/wp-admin/admin.php?page=zerobscrm-plugin-settings&tab=transactions`
2. Go to the transactions listview: `/wp-admin/admin.php?page=manage-transactions`
3. Check the listview filter dropdown and make sure it filters as expected.

In `trunk`, the listview filter dropdown does not show the new status.

In the `fix/crm/3356-show_custom_transaction_status_filters` branch, the listview filter dropdown shows the status, and it works as expected.

Bonus: test the below filter, which will add an item to the dropdown.

```
function jpcrm_add_custom_status_filter($quickfilters, $obj_type_id) {
	if ($obj_type_id === ZBS_TYPE_TRANSACTION) {
		$quickfilters['some_custom_filter'] = array( 'Some custom filter' );
	}

	return $quickfilters;
}
add_filter( 'jpcrm_add_custom_quickfilter', 'jpcrm_add_custom_status_filter', 10, 2 );
```

### DIFF
--- a/projects/plugins/crm/changelog/fix-crm-3356-show_custom_transaction_status_filters
+++ b/projects/plugins/crm/changelog/fix-crm-3356-show_custom_transaction_status_filters
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Transactions: filters now work for custom statuses
--- a/projects/plugins/crm/includes/ZeroBSCRM.List.Columns.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.List.Columns.php
@@ -649,52 +649,100 @@ $zeroBSCRM_filterbuttons_company['all'] = array( // phpcs:ignore WordPress.Namin
     //global $zbs;
     //$zeroBSCRM_columns_customer['current'] = $zbs->settings->get('customviews2');
 
+// phpcs:disable Generic.WhiteSpace.ScopeIndent.Incorrect, Generic.WhiteSpace.ScopeIndent.IncorrectExact
+/**
+ * Adds custom statuses as quickfilters
+ *
+ * This is a light refactor of some legacy code to make things a bit more modular,
+ * but obviously would benefit once we reduce our dependency on excessive globals.
+ */
+function jpcrm_add_custom_statuses_as_quickfilters() {
 
+	$obj_types_with_custom_statuses = array( ZBS_TYPE_CONTACT, ZBS_TYPE_COMPANY, ZBS_TYPE_TRANSACTION );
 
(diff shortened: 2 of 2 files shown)


coder-karen (commented):
Very nice!

> Bonus: test the below filter, which will add an item to the dropdown.

Just wondering about this - what could that filter be used for? If I add a custom filter via the snippet it won't show as a status option when adding a new transaction, so I'm not sure how it would be used 🤔

tbradsha:
> Just wondering about this - what could that filter be used for? If I add a custom filter via the snippet it won't show as a status option when adding a new transaction, so I'm not sure how it would be used 🤔

You're right, adding a status this way doesn't make sense. Quickfilters can be used for things other than statuses, so my example hook function wasn't a great example. 😅 

The filter isn't too useful right now; I mainly added it to allow for future extensibility should it be useful sometime down the road. It would have to be in conjunction with another hook in the DAL, where one would have some custom query logic, e.g. somewhere in here:

https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Contacts.php#L1535

But happy to take it out if you prefer!

coder-karen (approved):
> But happy to take it out if you prefer!

Perhaps it will have a use, though! I'm all for extensibility.

Generally though, I think this is great.

PR #33483, opened 2023-10-05
Blogroll: show errors and better loading state management

heavyweight:
This PR adds a spinner when the appender is loading subscriptions. The appender button is shown only after this request finishes.
To avoid showing two spinners this one is hidden when the recommendations are being loaded.
The PR also shows in the frontend possible errors from the requests we use.

Screenshot 2023-10-05 at 16 22 43

Fixes #
## Proposed changes:

*

### Other information:

## Testing instructions:

* In the editor create a new post and add the blogroll block
* When loading the recommendations there should be only one spinner visible
* Save the post and reload the page. When the subscriptions are loaded you should see a small spinner button before the button appender is shown.
* Provoke the requests we use to cause an error, or return error from the hooks and verify that the error messages are being shown in the block


agrullon95 (approved):
Verified the "blog appender" spinner appears then changed to the "+" icon when the request finished.

PR #33487, opened 2023-10-06
Internal release: add was_hosting_trial flag to indicate if site ever had a hosting free trial

mpkelly:
## Proposed changes:

* add a flag to site endpoint responses, `was_hosting_trial`. This flag is used internally by the Hosting Free Trial project. 

### Other information:

## Testing instructions:

* Test instructions added to internal issue.

### DIFF
--- a/projects/plugins/jetpack/changelog/fht-add_was_hosting_trial_flag
+++ b/projects/plugins/jetpack/changelog/fht-add_was_hosting_trial_flag
@@ -1,4 +1,4 @@
 Significance: minor
 Type: other
 
-527 - Sites API: add was_migration_trial flag to data returned for site details.[D[D[D[D[D
+527 - Sites API: add was_hosting_trial flag to data returned for site details.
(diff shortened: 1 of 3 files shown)


vykes-mac (commented):
@mpkelly when I use `v1.1` if the api i see the `was_hosting_trial` flag, however when I use v1.2 the flag is not present. The `was_migration_trial` and `was_ecommerce_trial` shows for both `v1.1` and `v1.2`, I'm assuming it should be the same for `was_hosting_trial`.

mpkelly:
> @mpkelly when I use `v1.1` if the api i see the `was_hosting_trial` flag, however when I use v1.2 the flag is not present. The `was_migration_trial` and `was_ecommerce_trial` shows for both `v1.1` and `v1.2`, I'm assuming it should be the same for `was_hosting_trial`.

@vykes-mac, thanks for finding that. I will fix it.

mpkelly:
I can see it in the 1.2 API. I was missing an entry in the `$site_format` so I added that. Can I get a re-review, @vykes-mac. 

Screenshot 2566-10-10 at 12 03 10

vykes-mac (approved):
This looks good @mpkelly

PR #33489, opened 2023-10-06
Launchpad: Adds the body prop to the tasks

valterlorran:
Fixes #

## Proposed changes:

* For my 20% time, I decided to test an idea we discussed in one of our calls with @daledupreez & @nuriapenya. The idea is to have a body with a description of the task and some other components like buttons, links, images, videos, etc.
* I added a fundamental data structure that allows us to render components on the front end. For now, I only made available a Text and Link component.

### Other information:

## Testing instructions:

* Apply this PR to your sandbox:
```shell
bin/jetpack-downloader test jetpack-mu-wpcom-plugin add/launchpad-task-body
```
* Edit your `0-sandbox.php` file and add the following code:
```php

function wpcom_launchpad_domain_customize_actions_filter( $actions, $task_id ) {
	if ( $task_id !== 'domain_customize' ) {
		return $actions;
	}
	return array(
		wpcom_launchpad_task_action_link(
			__( 'Learn more about', 'jetpack-mu-wpcom' ),
			'https://wordpress.com/support/domains/customize-your-domain/'
		),
		wpcom_launchpad_task_action_link(
			__( 'Help', 'jetpack-mu-wpcom' ),
			'https://wordpress.com/support/domains/customize-your-domain/'
		),
	);
}
add_filter( 'wpcom_launchpad_customize_task_actions', 'wpcom_launchpad_domain_customize_actions_filter', 10, 2 );

function wpcom_launchpad_domain_customize_content_filter( $content, $task_id ) {
	if ( $task_id !== 'domain_customize' ) {
		return $content;
	}
	return __( 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Pellentesque in tellus id eros scelerisque eleifend eget quis dui. Aenean eget sem non ex pulvinar commodo. Aliquam vel justo id ex rutrum faucibus id eu mi.', 'jetpack-mu-wpcom' );
}
add_filter( 'wpcom_launchpad_customize_task_content', 'wpcom_launchpad_domain_customize_content_filter', 10, 2 );

```
* Navigate to the developer console: https://developer.wordpress.com/docs/api/console/
* Make a `GET` request to `WP REST API -> wpcom/v2 -> /sites/:siteSlug/launchpad?_locale=user&checklist_slug=intent-build`
* Ensure you can see the `extended_content` property under the "Customize your domain" task.

Screen Shot 2023-10-30 at 11 19 05

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
@@ -371,7 +371,7 @@ class Launchpad_Task_Lists {
 		}
 
 		$extended_content = $this->load_extended_content( $task );
-		if ( $extended_content ) {
+		if ( is_array( $extended_content ) && $extended_content !== array() ) {
 			$built_task['extended_content'] = $extended_content;
 		}
 
@@ -457,28 +457,48 @@ class Launchpad_Task_Lists {
 	 * @return array|null The extended content for the task.
 	 */
 	private function load_extended_content( $task ) {
-		if ( ! isset( $task['extended_content'] ) ) {
+		if ( ! isset( $task['extended_content'] ) || ! is_array( $task['extended_content'] ) ) {
 			return null;
 		}
 
 		$extended_content_context = array();
 
-		foreach ( $task['extended_content'] as $context => $extended_content ) {
-			$loaded_extended_context = array();
+		foreach ( $task['extended_content'] as $extended_content_callback ) {
+			if ( isset( $extended_content_callback ) && is_callable( $extended_content_callback ) ) {
+				$extended_content_for_context = call_user_func( $extended_content_callback, $task['id'] );
 
-			$actions = $this->load_value_from_callback( $extended_content, 'actions', null, $task['id'] );
-			if ( $actions ) {
-				$loaded_extended_context['actions'] = $actions;
+				if (
+					is_array( $extended_content_for_context )
+					&& isset( $extended_content_for_context['contexts'] )
+					&& is_array( $extended_content_for_context['contexts'] )
+					// require either actions or content - we might want both, or only require content
+					&& ( isset( $extended_content_for_context['actions'] ) || isset( $extended_content_for_context['content'] ) )
+				) {
+					$current_contexts        = $extended_content_for_context['contexts'];
+					$current_content_updated = false;
+					$current_context_data    = array(
+						'actions' => array(),
+						'content' => array(),
+					);
+
+					if ( isset( $extended_content_for_context['actions'] ) && is_array( $extended_content_for_context['actions'] ) && array() !== $extended_content_for_context['actions'] ) {
+						$current_context_data['actions'] = $extended_content_for_context['actions'];
+						$current_content_updated         = true;
+					}
+
+					if ( isset( $extended_content_for_context['content'] ) && is_array( $extended_content_for_context['content'] ) && array() !== $extended_content_for_context['content'] ) {
+						$current_context_data['content'] = $extended_content_for_context['content'];
+						$current_content_updated         = true;
+					}
+
+					if ( $current_content_updated ) {
+						foreach ( $current_contexts as $current_context ) {
+							$extended_content_context[ $current_context ] = $current_context_data;
+						}
+					}
+				}
 			}
-
-			$content = $this->load_value_from_callback( $extended_content, 'content', null, $task['id'] );
-			if ( $actions ) {
-				$loaded_extended_context['content'] = $content;
-			}
-
-			$extended_content_context[ $context ] = $loaded_extended_context;
 		}
-
 		return $extended_content_context;
 	}
 
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
@@ -370,6 +370,11 @@ class Launchpad_Task_Lists {
 			}
 		}
 
+		$extended_content = $this->load_extended_content( $task );
+		if ( $extended_content ) {
+			$built_task['extended_content'] = $extended_content;
+		}
+
 		return $built_task;
 	}
 
@@ -443,6 +448,38 @@ class Launchpad_Task_Lists {
 			$task['subtitle'];
 		}
 		return '';
+	}
+
+	/**
+	 * Loads the extended content for a task.
+	 *
+	 * @param Task $task A task definition.
+	 * @return array|null The extended content for the task.
+	 */
+	private function load_extended_content( $task ) {
+		if ( ! isset( $task['extended_content'] ) ) {
+			return null;
+		}
+
+		$extended_content_context = array();
+
+		foreach ( $task['extended_content'] as $context => $extended_content ) {
+			$loaded_extended_context = array();
+
+			$actions = $this->load_value_from_callback( $extended_content, 'actions', null, $task['id'] );
+			if ( $actions ) {
+				$loaded_extended_context['actions'] = $actions;
+			}
+
+			$content = $this->load_value_from_callback( $extended_content, 'content', null, $task['id'] );
+			if ( $actions ) {
+				$loaded_extended_context['content'] = $content;
+			}
+
+			$extended_content_context[ $context ] = $loaded_extended_context;
+		}
+
+		return $extended_content_context;
 	}
 
 	/**
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -386,6 +386,9 @@ function wpcom_launchpad_get_task_definitions() {
 				// that are not in the Customer Home page. We should find a better way to handle this.
 				return '/domains/add/' . $data['site_slug_encoded'] . '?from=my-home';
 			},
+			'extended_content'     => array(
+				wpcom_launchpad_add_task_extended_context( array( 'launchpad-navigator' ) ),
+			),
 		),
 
 		'share_site'                      => array(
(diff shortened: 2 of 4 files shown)


sixhours:
This is working as described! Looks like it needs to be rebased. :)

Screenshot 2023-10-10 at 3 44 16 PM

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -349,6 +349,8 @@ private function build_task( $task ) {
 		$built_task['completed']    = $this->is_task_complete( $task );
 		$built_task['disabled']     = $this->is_task_disabled( $task );
 		$built_task['subtitle']     = $this->load_subtitle( $task );
+		$built_task['body']         = $this->load_value_from_callback( $task, 'body_callback' );
+		$built_task['body_context'] = isset( $task['body_context'] ) ? $task['body_context'] : array();
```
sixhours: What happens if we add a `body_callback` but no `body_context` -- does the callback get applied to all contexts by default? Do we need to do any kind of error checking around that?

valterlorran: We use a filter on the front end to check if the current context is in the array, which means that it shouldn't show in any context if the `body_context` is an empty array.  I removed the definition in case the `body_context` is empty either way to prevent sending more data to the front end unnecessarily.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -349,6 +349,8 @@ private function build_task( $task ) {
 		$built_task['completed']    = $this->is_task_complete( $task );
 		$built_task['disabled']     = $this->is_task_disabled( $task );
 		$built_task['subtitle']     = $this->load_subtitle( $task );
+		$built_task['content']      = $this->load_value_from_callback( $task, 'content_callback' );
+		$built_task['actions']      = $this->load_value_from_callback( $task, 'actions_callback' );
```
daledupreez: I have a concern around how we're modelling this that will cause complications down the line. As things stand we have a list of contexts and a single set of content and actions across all those contexts. From a longer-term perspective, would we ever want specific content or actions to be linked to particular contexts? Or will we always want the same content and actions across all contexts? I don't know the answer to this question, but we might want to set this up so the content is specifically linked to a context, which means we might want the JSON representation to look something more like the following:
```js
{
	/* standard task properties */
	"extended_content": {
		"context_example": {
			"content": { /* content */ },
			"actions": [ /* actions */ ],
			/* Maybe we need an additionalContexts array here? */
		},
		/* Other contexts with separate content/actions would have dedicated sections  */
	}
}
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -1829,3 +1832,50 @@ function wpcom_trigger_email_campaign() {
...
+}
+
+/**
+ * Adds the content to the domain customize task.
+ *
+ * @return string The content of the domain customize task.
+ */
+function wpcom_launchpad_domain_customize_content() {
+	return apply_filters( 'wpcom_launchpad_domain_customize_content', null );
+}
```
daledupreez: Might it be worth setting this up via a general-purpose function and filter? Is the thinking we'd use filters in the long term? Or just local functions?

If we're going to keep the filters in place, it may be a lot easier to have two general-purpose filters  that accept the task ID as an argument.
```suggestion
function wpcom_launchpad_customize_task_actions( $task_id ) {
	return apply_filters( 'wpcom_launchpad_customize_task_actions', null, $task_id );
}

/**
 * Adds custom content to tasks.
 *
 * @return string The content for the task.
 */
function wpcom_launchpad_customize_task_content( $task_id ) {
	return apply_filters( 'wpcom_launchpad_customize_content', null, $task_id );
}
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -386,6 +386,9 @@ function wpcom_launchpad_get_task_definitions() {
 				// that are not in the Customer Home page. We should find a better way to handle this.
 				return '/domains/add/' . $data['site_slug_encoded'] . '?from=my-home';
 			},
+			'body_context'         => array( 'launchpad-navigator' ),
+			'actions_callback'     => 'wpcom_launchpad_domain_customize_actions',
+			'content_callback'     => 'wpcom_launchpad_domain_customize_content',
```
daledupreez: It feels like it may be worth centralising this logic a bit so there is a single callback function to build the content, actions, and relevant contexts, along the lines of the long-term question I had above. Instead of having three separate pieces of logic, might it be simpler to have a single `extended_content_callback` that's responsible for returning the contexts, actions, and content?

daledupreez (commented):
I have some comments based on inspection and thinking about the data model we want.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -370,6 +370,11 @@ private function build_task( $task ) {
 			}
 		}
 
+		$extended_content = $this->load_extended_content( $task );
+		if ( $extended_content ) {
```
daledupreez: It may be worth validating that we have a non-empty array here, though I don't think that should be blocking.
```suggestion
		if ( is_array( $extended_content ) && $extended_content !== array() ) {
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -445,6 +450,38 @@ private function load_subtitle( $task ) {
...
 	}
 
+	/**
+	 * Loads the extended content for a task.
+	 *
+	 * @param Task $task A task definition.
+	 * @return array|null The extended content for the task.
+	 */
+	private function load_extended_content( $task ) {
+		if ( ! isset( $task['extended_content'] ) ) {
```
daledupreez: We should also check that `$task['extended_content']` is an array before we `foreach` over it.
```suggestion
		if ( ! isset( $task['extended_content'] ) || ! is_array( $task['extended_content'] ) ) {
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -445,6 +450,38 @@ private function load_subtitle( $task ) {
...
+			return null;
+		}
+
+		$extended_content_context = array();
+
+		foreach ( $task['extended_content'] as $context => $extended_content ) {
```
daledupreez: I _really_ dislike the variable name here, because we're using it as a callback and not as the content itself. I'd strongly prefer something like `$extended_content_callback`.
```suggestion
		foreach ( $task['extended_content'] as $context => $extended_content_callback ) {
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -1895,3 +1898,72 @@ function wpcom_launchpad_mark_theme_selected_complete( $new_theme, $old_theme )
...
+ * @param string $content_callback The callback to use for the task content.
+ * @return array The extended context.
+ */
+function wpcom_launchpad_add_task_extended_context( $contexts, $actions_callback = 'wpcom_launchpad_customize_task_actions', $content_callback = 'wpcom_launchpad_customize_task_content' ) {
+	return array(
+		'contexts' => $contexts,
+		'actions'  => $actions_callback,
+		'content'  => $content_callback,
+	);
+}
```
daledupreez, line 1945: The returned data structure here doesn't match what we're expecting on line 466, as that code is expecting the context to be used as the array key, and not be stored in a separate `contexts` key.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -445,6 +450,38 @@ private function load_subtitle( $task ) {
...
+
+			$content = $this->load_value_from_callback( $extended_content, 'content', null, $task['id'] );
+			if ( $actions ) {
+				$loaded_extended_context['content'] = $content;
+			}
+
+			$extended_content_context[ $context ] = $loaded_extended_context;
+		}
+
+		return $extended_content_context;
```
daledupreez: The structure of the loop here doesn't quite match the data structure for the returned data, and it feels like we're still unnecessarily splitting up the action and content from each other. It also feels like `$this->load_value_from_callback()` isn't ideal for this more complex pattern. I would much rather we tackled this via one callback, along with a loop structure more like the following:
```suggestion
		$extended_content = array();

		foreach ( $task['extended_content'] as $context_index => $extended_content_data ) {
			$content_by_context = array();

			if ( isset( $extended_content_callback ) && is_callable( $extended_content_callback ) ) {
				$extended_content_for_context = call_user_func( $extended_content_callback, $task['id'] );

				if (
					is_array( $extended_content_for_context )
					&& isset( $extended_content_for_context['contexts'] )
					&& is_array( $extended_content_for_context['contexts'] )
					// require either actions or content - we might want both, or only require content
					&& ( isset( $extended_content_for_context['actions'] ) || isset( $extended_content_for_context['content'] ) )
				) {
					$current_contexts = $extended_content_for_context['contexts'];
					$current_content_updated = false;
					$current_context_data = array(
						'actions' => array(),
						'content' => array(),
					);

					if ( isset( $extended_content_for_context['actions'] ) && is_array( $extended_content_for_context['actions'] ) && array() !== $extended_content_for_context['actions'] ) {
						$current_context_data['actions'] = $extended_content_for_context['actions'];
						$current_content_updated = true;
					}

					if ( isset( $extended_content_for_context['content'] ) && is_array( $extended_content_for_context['content'] ) && array() !== $extended_content_for_context['content'] ) {
						$current_context_data['content'] = $extended_content_for_context['content'];
						$current_content_updated = true;
					}

					if ( $current_content_updated ) {
						foreach ( $current_contexts as $current_context ) {
							$extended_content[ $current_context ] = $current_context_data;
						}
					}
				}
			}
		}

		return $extended_content_context;
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
valterlorran: Thanks for catching this. I knew I was missing one of my stashes! 🙄 
Screen Shot 2023-11-01 at 15 01 50

sixhours:
These are out of date, going to close for now. We can rework this if we decide to try it again.

PR #33490, opened 2023-10-06
Blogroll: front-end styling improvements

heavyweight:
This PR addresses two issues. We were using responsive breakpoints and not accounting for the fact that the block can be included in columns rows etc. This PR changes this targeting to be based on the container width.
Also on smaller devices, the subscribe submission form was not styled well. This is addressed here to have a column instead of a row direction.

Fixes #
https://github.com/Automattic/wp-calypso/issues/82665

## Proposed changes:

* Improve styling

### Other information:

## Testing instructions:

* Go to the admin and in a new post, add a 2 column block. In the first one add a paragraph and in the second one the blogroll.
* Add again the blogroll in the same post but outside of the column. We will use this one to compare the style targetting fixes.
* Save the post and visit the post url.
* The blogroll in the column, should have column direction, and the blogroll outside of the column should have row direction. Also when clicking subscribe, the form shown should show differently for the blogroll blocks


agrullon95 (dismissed):
Tested with 2 columns and without (desktop + mobile view).
Both blocks were displaying correctly.

agrullon95 (approved):

PR #32900, opened 2023-09-06
WPCOM Media: Add HEIC/HEIF image upload support

danielbachhuber:
See https://github.com/Automattic/wp-calypso/issues/55102
Originally https://github.com/Automattic/jetpack/pull/31719

## Proposed changes:

Ports the existing WordPress.com HEIC/HEIF support to the `jetpack-mu-wpcom` package so it can be used on both Simple and Atomic sites.

This PR only impacts the Media Library upload experience. It doesn't affect images uploaded via the Block Editor.

image

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

N/A

## Does this pull request change what data or activity we track or use?

No

## Testing instructions:

Here's an image you can test with if needed: https://danielbachhuber.com/wp-content/uploads/2023/07/IMG_3634.heif_.zip

### Simple

1. Create a new Premium Simple site and sandbox it.
2. Apply D120861-code to your sandbox.
3. Navigate to `/wp-admin/upload.php` on your Simple site.
4. Attempt to upload a HEIC image and verify you seen an error.
5.  Inside of your sandbox, run `bin/jetpack-downloader test jetpack-mu-wpcom-plugin improve/jetpack-wpcom-heif-support-v2`
6. Attempt to upload a HEIC image and verify it uploads as expected.

### Atomic

1. Create a new Business site, mark it as a WoA dev site, and take it Atomic.
2. Install the [Jetpack Beta Tester plugin](https://jetpack.com/download-jetpack-beta/) and add `define( 'JETPACK_MU_WPCOM_LOAD_VIA_BETA_PLUGIN', true );` to your `wp-config.php` file.
3. Navigate to `/wp-admin/admin.php?page=jetpack-beta&plugin=jetpack-mu-wpcom-plugin` to switch to the `improve/jetpack-wpcom-heif-support-v2` branch.
4. Navigate to `/wp-admin/upload.php` on your Atomic site.
5. Attempt to upload a HEIC image and verify it uploads as expected.

PR #33494, opened 2023-10-06
Allow Image CDN to Proxy HEIC / HEIF Images

xyu:
## Proposed changes:

The backend `Photon_OpenCV` extension actually does build in HEIC support using the following libraries:

```
libheif-decoder.cpp
libheif-decoder.h
libheif-encoder.cpp
libheif-encoder.h
```

What's more it's even allowed by `photon/index.php` because while the `$allowed_types` is currently set as:

```
49	$allowed_types = apply_filters( 'allowed_types', array(
50	        'gif',
51	        'jpg',
52	        'jpeg',
53	        'png',
54	) );
```

`libheif` returns `jpeg` as the type for HEIC images:

```
$ wp shell
wp> $i = new Photon_OpenCV();
=> object(Photon_OpenCV)#4849 (0) {
}
wp> $i->readimageblob( file_get_contents( 'img.heic' ) );
=> NULL
wp> $i->getimageformat()
=> string(4) "jpeg"
```

So all we need to do is to allow `heic` extensions to be passed through Photon. In fact we can test this right now and see that it works as expected:
- HEIC: https://xiao.blog/sample.heic
- JPEG: https://i0.wp.com/xiao.blog/sample.heic?w=600

### Other information:

## Jetpack product discussion

p9o2xV-35e-p2

Related: #32900 

No

## Testing instructions:

Check site with HEIC images load images from Photon instead.


projects/packages/image-cdn/CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.2.7] - 2023-10-06
+### Added
+- Added HEIC (`*.heic`) to list of images types allowed to be passed through Photon. [#33494]
```
jeherve: We typically do not edit the changelog directly, since when you work on your PR you do not yet know when the change will be released / what version of the package will include your changes. 

Instead, you can run `jetpack changelog add packages/image-cdn` and follow the prompts to add your changelog entry. It will generate a changelog entry file that you can commit and that will be merged alongside your PR. When we next release an update to the Image CDN package, we'll use the different changelog entry files from the different PRs that were merged since the last release to:

1. Decide what version should be released, based on the severity declared in the different changelog entry files.
1. Generate the changelog for that version.

Doing things that way will also fix the failed changelog GitHub Action on your PR.

projects/packages/image-cdn/src/class-image-cdn.php
```diff
@@ -33,7 +33,9 @@ final class Image_CDN {
 		'jpg',
 		'jpeg',
 		'png',
-		'webp', // Jetpack assumes Photon_OpenCV backend class is being used on the server. See link in docblock.
+		// Jetpack assumes Photon_OpenCV backend class is being used on the server. See link in docblock.
+		'webp', // Photon_OpenCV supports webp with libwebp-*, getimageformat() returns webp
+		'heic', // Photon_OpenCV supports webp with libheif-*, getimageformat() returns jpeg so does not match docblock
```
jeherve, line 38: Do you think you could add a test for this change, or edit the existing `get_test_image_cdn_validate_image_url_file_types_data_provider` test?

jeherve (changes requested):

projects/packages/image-cdn/CHANGELOG.md
xyu: Done in 4e4d30187b

projects/packages/image-cdn/src/class-image-cdn.php
xyu: Done in 49e965b619

jeherve (commented):
Should we extend this to also allow Jetpack's Search feature to support HEIC?
https://github.com/Automattic/jetpack/blob/3a3c4adc1c0913bac224ddb832713bd6101fe1ab/projects/packages/search/src/instant-search/lib/hooks/use-photon.js#L32

Here is how you would go about testing this:

1. Go to Jetpack > Settings > Performance and ensure that Instant Search is active.
2. Click to customize Instant Search and ensure that you are using the expanded mode, where thumbnails are displayed.
3. Search for a post that includes an HEIC image, using the Instant Search interface.
4. You should see the image appear in the search preview.

*********

Should we also extend the WordPress.com REST API to support HEIC?
https://github.com/Automattic/jetpack/blob/d59247968a6ced82775a75e41d0d44c890ec76d8/projects/plugins/jetpack/class.json-api-endpoints.php#L1598
https://github.com/Automattic/jetpack/blob/d59247968a6ced82775a75e41d0d44c890ec76d8/projects/plugins/jetpack/sal/class.json-api-post-base.php#L939
(and maybe in some other places?)
Looking at #32900, I'm guessing maybe we shouldn't, or should only limit it to WordPress.com sites?

xyu:
> Should we extend this to also allow Jetpack's Search feature to support HEIC?
> 
> https://github.com/Automattic/jetpack/blob/3a3c4adc1c0913bac224ddb832713bd6101fe1ab/projects/packages/search/src/instant-search/lib/hooks/use-photon.js#L32

This makes sense to me added with 2f67f34 

> Should we also extend the WordPress.com REST API to support HEIC?
> 
> https://github.com/Automattic/jetpack/blob/d59247968a6ced82775a75e41d0d44c890ec76d8/projects/plugins/jetpack/class.json-api-endpoints.php#L1598
> 
> https://github.com/Automattic/jetpack/blob/d59247968a6ced82775a75e41d0d44c890ec76d8/projects/plugins/jetpack/sal/class.json-api-post-base.php#L939
> 
> (and maybe in some other places?)
> Looking at #32900, I'm guessing maybe we shouldn't, or should only limit it to WordPress.com sites?

I'm not sure if we should add this to the API endpoints above, currently we kinda only have partial support for HEIC images in that we will _always_ convert HEIC images passed through Photon to JPEG which makes it compatible with all browsers. The downside is even for browsers that support HEIC like Safari we still convert the image which is not exactly what the client asked for. This probably is fine on the front end but I'm not sure if that would make it weird for a REST / programmatic endpoint. I.e. we advertise the URL to an HEIC image that's really a JPEG. What do you think @jeherve?

jeherve (approved):

inaikem:
Hi folks, checking in here for an old open ticket here: 7089554-zd-a8c

Can you please confirm if this fix has landed on AT sites yet? Thanks!

jeherve:
> Can you please confirm if this fix has landed on AT sites yet? Thanks!

It has, yes.

PR #33497, opened 2023-10-08
Add/keep deleted plans

millerf:
Fixes https://github.com/Automattic/gold/issues/149

This is only to keep sync in changes for D124684-code

And sync the meta needed.


pottedmeat (approved):
Code looks good and was tested.

PR #33508, opened 2023-10-09
phpcs-filter: Ignore `--basepath` option

anomiex:
## Proposed changes:

I had thought the use of this would be to point to the base of the monorepo. But it seems likely IDEs like VSCode are using it to point to various subdirectories of the monorepo instead.

Since, on re-reading the help docs for the parameter, that does seem like a valid use for the parameter, we shouldn't try to use it to determine the base of the monorepo after all.

### Other information:

## Jetpack product discussion

pf4qpu-3P-p2#comment-33

No

## Testing instructions:

* Run something like `vendor/bin/phpcs -p -s --basepath=projects/packages/assets/src/ projects/packages/assets/src/class-assets.php`


sergeymitr (approved):
Works as described 👍

PR #33519, opened 2023-10-09
AI Chat block: Fix button text wrapping in Firefox

### LINKED ISSUE #33516: AI Chat block: Button styling in Firefox
Styling for the button is off in Firefox at narrow viewports. 

image

p8oabR-1mw-p2#comment-7627

dereksmart:
Fixes #33516 

## Proposed changes:

`text-wrap` is [not supported](https://developer.mozilla.org/en-US/docs/Web/CSS/text-wrap) in Firefox, which led to the text overflowing in FF. 

Before:
image

After: 
image

No

## Testing instructions:

* Set a long string as the button text
* Look at the button on front end view in FF
* With this patch, it should not overflow
* Make sure it still looks good in other browsers.

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-ai-chat-button-wrapping
+++ b/projects/plugins/jetpack/changelog/fix-ai-chat-button-wrapping
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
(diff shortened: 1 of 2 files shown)


sdixon194 (approved):
Works for me!

PR #33522, opened 2023-10-10
Social: Add instagram aspect-ratio validation for images

gmjuhasz:
At the moment we only check aspect-ratio for videos. This should be fixed, this change adds the needed logic to correctly check the width and aspect-ratio requirements for Instagram

## Proposed changes:

*

### Other information:

No.
## Testing instructions:

* On trunk turn off auto-conversion on your site.
* Find an image that [does not meet the requirements](https://developers.facebook.com/docs/instagram-api/reference/ig-user/media/#image-specifications) for the aspect-ratio validation of Instagram
* Uploading that as attached media - if the filesize is correct - should not show errors.
* Check out this PR. Applying the same image should result in a warning.

### DIFF
--- a/projects/js-packages/publicize-components/src/components/media-picker/index.js
+++ b/projects/js-packages/publicize-components/src/components/media-picker/index.js
@@ -85,7 +85,11 @@ export default function MediaPicker( {
 								naturalHeight={ previewHeight || height }
 								isInline
 							>
-								<img src={ previewUrl } alt="" className={ styles[ 'preview-image' ] } />
+								<img
+									src={ previewUrl || sourceUrl }
+									alt=""
+									className={ styles[ 'preview-image' ] }
+								/>
 							</ResponsiveWrapper>
 						) }
 					</button>
--- a/projects/js-packages/publicize-components/src/components/media-picker/index.js
+++ b/projects/js-packages/publicize-components/src/components/media-picker/index.js
@@ -37,8 +37,11 @@ export default function MediaPicker( {
 	onChange,
 	wrapperClassName,
 } ) {
-	const { mediaData: { width, height, sourceUrl } = {}, metaData: { mime, length = null } = {} } =
-		mediaDetails;
+	const {
+		mediaData: { width, height, sourceUrl } = {},
+		metaData: { mime, length = null } = {},
+		previewData: { width: previewWidth, height: previewHeight, sourceUrl: previewUrl } = {},
+	} = mediaDetails;
 
 	const isImageLoading = ! sourceUrl || ! width || ! height || ! mime;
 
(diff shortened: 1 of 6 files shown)


projects/js-packages/publicize-components/src/hooks/use-media-restrictions/restrictions.js
```diff
@@ -126,6 +132,12 @@ export const RESTRICTIONS = {
 		allowedMediaTypes: [ 'image/jpg', 'image/jpeg', MP4, MOV, VIDEOPRESS ],
 		image: {
 			maxSize: 8,
+			minWidth: 320,
+			maxWidth: 1440,
+			aspectRatio: {
+				min: 4 / 5,
+				max: 1.91 / 1,
```
pablinos: Do you think we should just keep this as 1.91?

```suggestion
				max: 1.91,
```

projects/js-packages/publicize-components/src/hooks/use-media-restrictions/index.js
```diff
@@ -138,7 +158,7 @@ const getValidationError = ( metaData, mediaData, serviceName, shouldUploadAttac
 				mediaData?.height,
 				restrictions.video
 		  )
-		: getImageValidationError( sizeInMb, restrictions.image.maxSize );
+		: getImageValidationError( sizeInMb, mediaData?.width, mediaData?.height, restrictions.image );
```
pablinos: Can `mediaData` be null/undefined? If so, then we're not guarding against it in the `getImageValidationError` function and might cause problems with the ratio calculation.

image

Seems like it depends on what is null. Probably it's ok, but getting back NaN I could see might cause problems.

projects/js-packages/publicize-components/src/components/media-picker/index.js
```diff
@@ -75,15 +80,30 @@ export default function MediaPicker( {
...
 							></VideoPreview>
 						) : (
-							<ResponsiveWrapper naturalWidth={ width } naturalHeight={ height } isInline>
-								<img src={ sourceUrl } alt="" className={ styles[ 'preview-image' ] } />
+							<ResponsiveWrapper
+								naturalWidth={ previewWidth || width }
+								naturalHeight={ previewHeight || height }
+								isInline
+							>
+								<img src={ previewUrl } alt="" className={ styles[ 'preview-image' ] } />
```
pablinos: Should we also use `sourceUrl` here if we're also falling back to `width` and `height`?

```suggestion
								<img src={ previewUrl || sourceUrl } alt="" className={ styles[ 'preview-image' ] } />
```

projects/js-packages/publicize-components/src/components/media-picker/index.js
```diff
@@ -27,6 +27,7 @@ const clickHandler = open => e => {
  * @param {object} props.mediaDetails - The details of the media for preview
  * @param {Function} props.onChange - A callback that can be passed to parent for validation
  * @param {string} props.wrapperClassName - A class name to be added to the wrapper
+ * @param {object} props.allowedMediaTypes - Custom allowed media types
```
pablinos, line 30: Are we using this? Should this be passed in when auto conversion is not enabled?

pablinos (commented):
This is pretty close I think. It's working well for me, but maybe a couple of tweaks

projects/js-packages/publicize-components/src/components/media-picker/index.js
gmjuhasz: It's used here: fbhepr%2Skers%2Swrgcnpx%2Scebwrpgf%2Swf%2Qcnpxntrf%2Schoyvpvmr%2Qpbzcbaragf%2Sfep%2Spbzcbaragf%2Sfbpvny%2Qvzntr%2Qtrarengbe%2Scnary%2Szbqny.wf%3Se%3Qro7374p5%26zb%3Q898%26sv%3Q22%2382-og

It was not working before, this fixes it

manzoorwanijk (dismissed):
This looks good to me.

pablinos (approved):

PR #33283, opened 2023-09-25
Add Stats counts (highlights) logic to My Jetpack

IanRamosC:
## Proposed changes:

* Add the StatsSection component to My Jetpack page
* Add logic to only display the section if Stats module is active
* Connect Stats section to the store (and backend using the endpoints from Odyssey Stats)

image

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

pbNhbs-80w-p2

## Testing instructions:

NOTE: for now, the base branch is the one used in #33160. I'll do a `git rebase` as soon as the other PR is merge, and this one will be much cleaner.

* Point your site to this branch using Jetpack Beta or run this branch locally
* Visit `/wp-admin/admin.php?page=my-jetpack` with Stats active, and you'll see the new Stats section on this page
* Visit `/wp-admin/admin.php?page=jetpack_modules` and disable the Jetpack Stats module
* Go back to My Jetpack and confirm the section isn't there anymore, and there's no call to the API (`/jetpack/v4/stats-app/sites/{site_id}/stats/highlights?_locale=user`)
* With the module enabled, in the Stats section on My Jetpack, click on "See detailed stats". Confirm that it redirects to the Jetpack Stats page

PR #33523, opened 2023-10-10
My Jetpack: Use Jetpack Modules

pyronaur:
Fixes a fatal error introduced in #33283

Using Jetpack class directly works when Jetpack plugin is active, but without the plugin this throws fatals - for example in standalone plugins like Jetpack Boost.

### Other information:

## Jetpack product discussion
N/A

N/A

## Testing instructions:
1. Use trunk
2. Deactivate Jetpack
3. Visit my-jetpack, verify fatal error
4. Apply patch
5. Visit my-jetpack, verify no error


donnchawp (dismissed):
Works! No more fatal error.

donnchawp (approved):
Still works!

PR #33524, opened 2023-10-10
JITM: Fix jitms not being clickable

mikestottuk:
Fixes the bug encountered in testing that JITMs were not clickable due to tracking being moved and the default behaviour was prevented and buttons with external links didn't work. See video:-

https://github.com/Automattic/jetpack/assets/4077604/c80e8735-613a-47f2-bb15-6d4f27050a76

## Proposed changes:
* Trigger the button click event at the end of the tracking function.

## Jetpack product discussion
n/a

no

## Testing instructions:
You need a JITM to be visible, before the buttons were reacting to a click because the default click event was prevented so the JITM could be tracked. 

* Got to any page with a JITM
* A good example is the edit posts page (for Blaze)
* Notice the JITM buttons are not clickable to take you where they want you to go

Apply this patch
* JITMs should be clickable again

### DIFF
--- a/projects/packages/jitm/src/js/jetpack-jitm.js
+++ b/projects/packages/jitm/src/js/jetpack-jitm.js
@@ -259,7 +259,6 @@ jQuery( document ).ready( function ( $ ) {
 
 		// Handle tracking for JITM CTA buttons
 		$template.find( '.jitm-button' ).on( 'click', function ( e ) {
-			e.preventDefault();
 
 			var button = $( this );
 			var eventName = button.attr( 'data-jptracks-name' );
@@ -275,9 +274,7 @@ jQuery( document ).ready( function ( $ ) {
 			};
 
 			if ( window.jpTracksAJAX ) {
-				window.jpTracksAJAX.record_ajax_event(eventName, 'click', eventProp).always( function() {
-					window.location.href = button.attr('href');
-				} );
+				window.jpTracksAJAX.record_ajax_event(eventName, 'click', eventProp);
 			}
 		});
 	};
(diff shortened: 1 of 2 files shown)


projects/packages/jitm/src/js/jetpack-jitm.js
```diff
@@ -275,9 +275,11 @@ jQuery( document ).ready( function ( $ ) {
 			};
 
 			if ( window.jpTracksAJAX ) {
-				window.jpTracksAJAX.record_ajax_event(eventName, 'click', eventProp);
+				window.jpTracksAJAX.record_ajax_event(eventName, 'click', eventProp).always( function() {
+					window.location.href = button.attr('href');
+				} );
```
IanRamosC: The change kinda works, but I think we'll always want to redirect the user (not depending on `window.jpTracksAJAX`), so I'd move this out of this `if`

IanRamosC: It's also not considering cases where we have `target` set to `_blank`, and we'd eventually lose other attributes as well

IanRamosC (changes requested):
The code kinda works, but I left some comments about things that are blockers IMO.

I was thinking about the code again, and I think we can just remove `e.preventDefault();` since `onClick` runs before the actual redirect/page open (from the `href`), and we'd like to keep the link behavior after the click. I couldn't test it locally because my Jurassic Tube setup is borked.

IanRamosC (approved):
I can confirm it works as expected. :shipit:

PR #33527, opened 2023-10-10
Paywall block: remove "beta" from the name in block.json

simison:
I don't think the block.json is used anywhere yet, but Crew can confirm.

Anyway, the block was released a while back and isn't "beta" anymore. We just missed the block.json

## Proposed changes:

* Remove "beta" from the name in block.json

### Other information:

## Testing instructions:

* Nothing to test.

### DIFF
--- a/projects/plugins/jetpack/changelog/update-paywall-title-beta
+++ b/projects/plugins/jetpack/changelog/update-paywall-title-beta
@@ -1,0 +1,4 @@
+Significance: minor
+Type: other
+
+Paywall block: remove "beta" from the name in block.json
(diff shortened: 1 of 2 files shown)


lezama (approved):

monsieur-z (approved):

monsieur-z:
> I don't think the block.json is used anywhere yet, but Crew can confirm.

It's used to generate block information in the [WordPress.org Jetpack description](https://wordpress.org/plugins/jetpack/). It will eventually also be the source of truth to register the block.

PR #33537, opened 2023-10-10
Feat/google fonts with no library

pbking:
A possible alternative to #33203

This change registers fonts using the "Font Collection" technique experimented with in the new Font Library API [here](https://github.com/WordPress/gutenberg/pull/54888).  However, rather than depending on the Font Library to register the fonts collection as "default" fonts this module does that explicitly itself.

This was shown working in #33203, however it depended on Classes introduced in the Font Library API.  This required the "old" technique to remain (for when those classes were absent).

This change instead loads and handles the information itself, leveraging the same expected data format of Font Collections and inserting it (via filter) into the theme.json data.

The result is that the Fonts are available to the user to USE but not available to them to manage.  Which is the desired behavior of this module.

Another difference from #33203 is that the collection of fonts (that continue to be filterable via the `jetpack_google_fonts_list` filter) is derived from the font collection JSON data instead of a hard-coded collection.  In this way there is a single source-of-truth for the available fonts and no opportunity for those two collections to diverge.

NOTE: One remaining suggested change is to move the configured collection of fonts from the [hosted file](https://s0.wp.com/i/font-collections/jetpack-google-fonts.json) to instead include it in the module.

NOTE: I'm unsure how the `jetpack-google-fonts.json` file was generated.  It might be good to have a utility that generates that from a collection (like that present in the current version) using the Google Fonts API (but saving that as the `jetpack-google-fonts.json` file rather than realtime as the current version does.

NOTE: The configuration file is pointing to all `.ttf` assets.  However `.woff2` format is much smaller and universally available.  I believe that those assets should be considered instead.  I know that, be default, Google provides the `.ttf` assets.  However (in most cases) the `.woff2` assets are also available.


arthur791004:
>  This required the "old" technique to remain (for when those classes were absent).

I tried to inject the Google Fonts to the default global styles before but we found https://github.com/Automattic/jetpack/pull/32579#pullrequestreview-1595789454 that Gutenberg would try to avoid the duplication and it led to `the array_unique keeps only 'default' and on the next iteration it blows up.`. Hence, it seems to be a risk so I wanted to keep the “old” techniques. And there might be some edge cases we didn't notice 🫠

> Another difference from https://github.com/Automattic/jetpack/pull/33203 is that the collection of fonts (that continue to be filterable via the jetpack_google_fonts_list filter) is derived from the font collection JSON data instead of a hard-coded collection. In this way there is a single source-of-truth for the available fonts and no opportunity for those two collections to diverge

Sounds good! It makes sense to me to get the Google Fonts from the JSON data instead of a hard-coded collection 👍

> I'm unsure how the jetpack-google-fonts.json file was generated

p1696997604231419/1696951463.106089-slack-C057THCBY2X

> The configuration file is pointing to all .ttf assets. However .woff2 format is much smaller and universally available

Sounds like `.woff2` is good enough. I'm curious about why Core uses `.ttf` format 🤔

arthur791004:
I tested it with Gutenberg 16.5.0, the fonts might be registered twice. One was from the Jetpack Google Fonts module, and the other seemed to be from the `add_missing_fonts_to_theme_json` via the `wp_fonts`



Additionally, when switching to the [Pendant theme](https://public-api.wordpress.com/rest/v1/themes/download/pendant.zip), the error occurred. The reason seemed to be the above one as the `add_missing_fonts_to_theme_json` added the missing fonts to the theme data, and it led to the issue mentioned https://github.com/Automattic/jetpack/pull/32579#pullrequestreview-1595789454 😓



Last, when disabling Gutenberg, it's weird that the theme fonts wouldn't be loaded in the Editor



I'm not sure how many users still use the old version of Gutenberg, but it seems risky and I don't think it's worth getting rid of the old implementation. I also love to keep it clean if we can but I guess we can just remove the deprecated code in the future. WDYT?

pbking:
I don't believe it's necessary to continue to support versions of Gutenberg that old for this feature.  Especially since I believe it works now even without Gutenberg installed.  That said, I'm not responsible for maintaining this project and it's not my call, only my opinion.

> Additionally, when switching to the [Pendant theme](https://public-api.wordpress.com/rest/v1/themes/download/pendant.zip), the error occurred.

Also not able to reproduce an error with Pendant.  It has a troublesome history with fonts though.

> Last, when disabling Gutenberg, it's weird that the theme fonts wouldn't be loaded in the Editor

I didn't experience this (with recent versions of Gutenberg or without).  The font assets were always loaded, both with and without Gutenberg. Is this behavior only on older versions of Gutenberg?

I tested with and without the `FONT_LIBRARY_DISABLED` flag.  With and without Gutenberg activated.  With Pendant and Blockbase (which has a huge number of provided fonts that overlap Jetpack's) and a Blockbase child (Appleton). 

If it's just a matter of returning the "Fonts API" version to satisfy specifics of older versions of Gutenberg that's fine, though you may have the opportunity to do so before me.

Either way, this logic seems to demonstrate that the behavior of the module can continue as before during the transition to the Font Library.  Do you agree once the shim for older versions of Gutenberg are in place?

If you note other errors please be sure to note which version of WordPress, Gutenberg you are testing with.

arthur791004:
@pbking I tested with [the JN site against this PR](https://jurassic.ninja/create/?jetpack-beta&branch=feat/google-fonts-with-no-library) where the WP version is 6.3.2, installed [Gutenberg v16.5.0](https://github.com/WordPress/gutenberg/releases/tag/v16.5.0), and activated [the Pendant theme](https://public-api.wordpress.com/rest/v1/themes/download/pendant.zip).

Here is the demo:

https://github.com/Automattic/jetpack/assets/13596067/12fc9d49-b765-4155-a571-0d27cd4bc007

Additionally, I used the following command to test this PR on the simple site with the Gutenberg v16.8.1 (`FONT_LIBRARY_DISABLED=true`) and the active Pendant theme. 

```
bin/jetpack-downloader test jetpack feat/google-fonts-with-no-library
```

The issue still occurs but this time the page would keep loading to display the Editor correctly. So, the issue might exist with the latest Gutenberg and disabling the Font Library. However, I cannot figure out why you didn't encounter the same issue 😂

https://github.com/Automattic/jetpack/assets/13596067/a292e678-28d4-43ba-8871-aa2d11e7678f

<details>
  <summary>Click to see the screenshot and error log</summary>

image

```
[/plugins/gutenberg-core/v16.8.1/lib/experimental/fonts-api/class-wp-fonts-resolver.php:208 array_merge(), /plugins/gutenberg-core/v16.8.1/lib/experimental/fonts-api/class-wp-fonts-resolver.php:117 get_settings(), wp-includes/class-wp-hook.php:324 register_fonts_from_theme_json(), wp-includes/class-wp-hook.php:348 apply_filters(''), wp-includes/plugin.php:517 do_action(Array), wp-settings.php:653 do_action('init'), wp-config.php:7 require('wp-settings.php'), wp-load.php:50 require_once('wp-config.php'), wp-admin/admin.php:34 require_once('wp-load.php'), wp-admin/site-editor.php:12 require_once('wp-admin/admin.php')]
```
</details>

---

> I don't believe it's necessary to continue to support versions of Gutenberg that old for this feature. Especially since I believe it works now even without Gutenberg installed. That said, I'm not responsible for maintaining this project and it's not my call, only my opinion.

I think we have to avoid any breaking changes as some people might have the self-hosted site with the auto-updated Jetpack but stick the Gutenberg version to v16.5.0. Additionally, I don't think there is any harm in keeping the current implementation for the old Gutenberg or WP. Maybe @jeherve has more ideas about the support.

pbking:
I believe with these two commits ([1](https://github.com/Automattic/jetpack/pull/33203/commits/c4d9bd7979c9840665ee677ac622e8673b2975c5),[2](https://github.com/Automattic/jetpack/pull/33203/commits/33ffa44e304df200476fb49e8c53e95ddac8d3cd)) on your original branch that removes the dependency on the Font Library API.  That should be the change necessary to land this.  That change keeps the original functionality which seems to be needed for these edge cases.

I will close this.

arthur791004:
Nice, thanks for the review 👍

PR #33543, opened 2023-10-10
AI Assistant: Add "Create a form" toolbar action

### LINKED ISSUE #32521: AI Assistant: Add "Create a form..." to AI Assistant block
ref p8oabR-1jo-p2#comment-7512

The AI Assistant block does not work with Gutenberg code currently and so does not generate forms. We can inform the user about the Form Extension so they know that they can use the Form component directly for that.

dhasilva:
Fixes #32521

## Proposed changes:

* Adds a second `MenuGroup` with an option to create a form, that replaces the assistant with a new empty form

### Other information:

## Testing instructions:

* Go to the block editor
* Add an AI Assistant block
* Click on `Write with AI...` > `Create a form...`
* Check that the assistant is replaced with a form


retrofox (commented):
I have reservations about introducing an option to create a form within the block toolbar of an extended block instance, like a paragraph. The block toolbar is populated with tools directly related to the manipulation and formatting of the specific block instance.

Introducing a mechanism to create a disparate block type (e.g., a Jetpack Form) within the toolbar might not align with the established user expectations and design paradigms.
It could potentially lead to a cluttered interface and user confusion, as they may not associate the creation of forms with the context of editing a paragraph block.

CGastrell:
Code looks good and works fine, but I have to agree with @retrofox here: where does this end? How does the approach scale? Will we end up adding an entry for each _assistable block_ there? While Rob's comment seems like a direct request to add some sort of interaction between the assistant and the form block, it might be better to include/work on the prompt rather than adding an entry on the assistant's menu.

dhasilva:
Closing as the comments have good points, and p1697483720545659-slack-C054LN8RNVA

PR #33548, opened 2023-10-11
Search: allow only valid post types in Customberg

kangzj:
## Proposed changes:

If two or more post types are set and then these post types got removed (because of plugin removed etc), the excluded post type option would become disabled. The PR proposes to filter on the invalid post types, i.e. remove out-dated settings.

### Other information:

## Jetpack product discussion
panfyZ-2PW-p2

no

## Testing instructions:

* Build search and Jetpack locally `jetpack build packages/search && jetpack build plugins/jetpack`
* Update value of `jetpack_search_excluded_post_types` option to `aaa,bbb`
* Open `/wp-admin/admin.php?page=jetpack-search-configure`
* Ensure Excluded post types setting is available for changes
* Ensure you excluded post types stick after click saving/page refresh

image


bluefuton:
@kangzj nice! It looks like it needs a project versions bump.

projects/packages/search/src/customberg/hooks/use-search-options.js
```diff
@@ -49,10 +52,11 @@ export default function useSearchOptions() {
 	// Excluded Post Types is stored as a CSV string in site options. Convert into array of strings.
 	// Caveat: csv can be an empty string, which can produces [ '' ] if only csv.split is used.
```
bluefuton:
```suggestion
	// Caveat: csv can be an empty string, which can produce [ '' ] if only csv.split is used.
```

jsnmoon (approved):

PR #33560, opened 2023-10-11
Super cache: cancel the full preload scheduled job instead of the mini one

### LINKED ISSUE #33479: Super-Cache: the full preload job isn't unscheduled properly
When the preload is cancelled, it unschedules any jobs to preload. Unfortunately the "full" preload job isn't properly unscheduled.
In wpsc_cancel_preload() in wp-cache.php

ref: https://wordpress.org/support/topic/preload-error-v-1-10-0/

donnchawp:
When the preload is cancelled it wasn't cancelling the job that started the next round of preloading. This small change fixes that.

Fixes #33479

## Proposed changes:
* Fixed a typo calling wp_unschedule_event() so it unschedules $next_full_preload rather than $next_preload

### Other information:

no

## Testing instructions:
* Install Super Cache and start a preload.
* Clicking on "Cancel Cache Preload" and the preload won't be cancelled. The button will still show as a cancel button.
* Apply this patch and click the Cancel button and preloading will be cancelled.

### DIFF
--- a/projects/plugins/super-cache/changelog/fix-super-cache_cancel_full_preload
+++ b/projects/plugins/super-cache/changelog/fix-super-cache_cancel_full_preload
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Super Cache: cancel the full preload job correctly.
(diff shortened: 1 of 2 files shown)


pyronaur (approved):
👍

PR #8710, opened 2018-02-01
Previous link was incorrect

richardmuscat:
Fix incorrect blog post link

PR #33583, opened 2023-10-12
General: update Jetpack brand color to new reference

jeherve:
## Proposed changes:

This old color code isn't used anymore; let's switch to the new color instead.

### Other information:

## Jetpack product discussion

* No

* No

## Testing instructions:

* Not much to test here; check that the color code matches what we use elsewhere in the plugin


sergeymitr:
Hi @jeherve.
It looks like `#005b18` is mostly used as `green-70`, not the default green.
Also, I wasn't able to find an UI element to test the changes on, do you know if those lines are used anywhere?

jeherve:
> It looks like `#005b18` is mostly used as `green-70`, not the default green.

Good point, thank you! I fixed it in a new commit above.

> Also, I wasn't able to find an UI element to test the changes on, do you know if those lines are used anywhere?

I couldn't find one either, I'm not sure this is used anymore to be honest. I was mostly making the change to avoid confusion whenever someone may look for Jetpack green in the codebase.

monsieur-z (commented):
I'm nitpicking here, but Jetpack 50 (`#008710`) would be safer regarding accessibility. We've been using that color on jetpack.com since it provides better contrast with plain white.

anomiex:
> > Also, I wasn't able to find an UI element to test the changes on, do you know if those lines are used anywhere?
> 
> I couldn't find one either, I'm not sure this is used anymore to be honest. I was mostly making the change to avoid confusion whenever someone may look for Jetpack green in the codebase.

I managed to get one of these to appear by setting https://github.com/Automattic/jetpack/blob/06413693d854d56d5a341ff92b547bd8a59c37b1/projects/plugins/jetpack/class.jetpack.php#L396 to a non-empty string. There's at least one code path that does that more naturally, but I didn't see how to trigger it at a quick glance.

jeherve:
> I'm nitpicking here, but Jetpack 50 (`#008710`) would be safer regarding accessibility. We've been using that color on jetpack.com since it provides better contrast with plain white.

👍 Done in 0ed8bcddf7e49a966366a90bf87f23459c557564

monsieur-z (approved):
🚢 

I took the liberty of fixing a typo in `_main.scss`.

PR #27968, opened 2022-12-15
Fix PHP 8.2 issues when running Jetpack tests

anomiex:
#### Changes proposed in this Pull Request:

This includes Jetpack itself and a few things in packages that are caught by Jetpack's tests (but not the packages' own tests or they'd be in #27949 instead).

packages/connection:
* Declare fields in Manager class.

packages/search:
* Pass empty string instead of null to `add_submenu_page()`, internally the latter was being converted to the former anyway.

packages/sync:
* Declare field in Queue class.
* When loading action links HTML, prepend an XML encoding declaration instead of using `mb_convert_encoding()` with "HTML-ENTITIES".
  * Also add an appropriate character to the Jetpack test hitting this to make sure this works.

plugins/jetpack:
* Declare fields in lots of classes.
* Use local variables instead of properties that aren't used outside of the method they're set in.
* Add an `is_string` check before a `preg_replace` in `Grunion_Contact_Form_Plugin`
* Set `AllowDynamicProperties` on csstidy classes. This will make it easier if we ever try to bring in a newer upstream version.

#### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

#### Jetpack product discussion

https://github.com/Automattic/jetpack/issues/27927

#### Does this pull request change what data or activity we track or use?

No

#### Testing instructions:

* Is CI happy?
* Actually running this with PHP 8.2 would be a pain, since WordPress itself throws deprecation warnings for a bunch of tests.

PR #33585, opened 2023-10-12
Jetpack: Fix various test deprecation notices

anomiex:
## Proposed changes:

In #27981 we cleaned up tests that were failing due to core's new Requests library, but missed one that logs without failing since the use is in a data provider rather than a test. That in turn was masking several later tests that also depended on the back-compat layer.

Also in #27968 we fixed the normal test run deprecation warnings about undeclared properties, but missed a few that only run for multisite. Then we may as well clean up the legacy full sync tests too, even though those currently only run under PHP 7.0 for some reason.

### Other information:

## Jetpack product discussion

None

No

## Testing instructions:

* Is CI happy?
* Check the logs, are there any deprecation notices left in Jetpack's tests?


jeherve (approved):

jeherve:
Should be good to merge but it will need a final rebase I'm afraid.

PR #33598, opened 2023-10-13
Add endpoints whitelisting for media

therocket-gr:
Fixes #

## Proposed changes:
Added 2 new endpoint whitelistings

1. for the new media endpoint (so that we can fetch images from the site)
2. for the openverse media fetch

### Other information:

## Testing instructions:

You need to test locally by using the branch `origin/feature/new-image-sources` for widget to load the new media window
Screenshot 2023-10-13 at 17 16 31
Screenshot 2023-10-13 at 17 10 03

ensure that you can open the media and that the endpoints work


sbarbosa (dismissed):
Thanks @therocket-gr! The code looks good 👍
I tested the integration in my local and it worked correctly.

Screenshot 2023-10-13 at 15 02 16

jeherve (dismissed):

projects/packages/blaze/src/class-dashboard-rest-controller.php
```diff
@@ -329,6 +351,30 @@ public function get_dsp_blaze_posts( $req ) {
...
+	/**
+	 * Redirect GET requests to WordAds DSP Blaze media endpoint for the site.
+	 *
+	 * @param WP_REST_Request $req The request object.
+	 * @return array|WP_Error
+	 */
+	public function get_dsp_media( $req ) {
+		$site_id = $this->get_site_id();
+		if ( is_wp_error( $site_id ) ) {
+			return array();
```
timur987, line 363: I played a bit with the code and started thinking what if we, using the the Fail Fast approach, return `$site_id` if it's an instance of `WP_Error`?

I agree that sometimes we return an empty array but I do believe that it would be better to return an error instead.
For example the function `request_as_user` that's used by `get_dsp_generic`, we use here, actually returns an error is something went wrong with the request ([here](https://github.com/Automattic/jetpack/blob/trunk/projects/packages/blaze/src/class-dashboard-rest-controller.php#L587)).

Also, I checked that returning `WP_Error` instead of an empty array will produce `500` error but it is handled fine.

timur987 (dismissed):
Thanks, @therocket-gr!

The code looks great and works as advertised.

I left a comment in order to maybe improve the code, they're mostly minor though.

projects/packages/blaze/src/class-dashboard-rest-controller.php
therocket-gr: ignoring for now

timur987 (dismissed):
Thanks for resolving the conflicts!

timur987 (approved):
Thanks for resolving conflicts, @therocket-gr!

As before, changes look great.

PR #33602, opened 2023-10-13
Connection Toggle: Prevent onChange when the component is disabled

pablinos:
When the toggle was disabled, such as when the site had no shares remaining, or if the post didn't have an image and wanted to share to Instagram, the `onChange` handler was still being called when the toggle was clicked.

This change prevents that by checking the state of the `disabled` prop first before calling the event handler.

### Other information:

## Testing instructions:
* Create a test site with Social and create a connection that can be placed into a disabled state
* The easiest way to do this is to create an Instagram Business connection and then not set a featured image on a post, but you can also reduce the number of shares remaining to 0. Sandbox the API, but don't connect to your sandbox, so all connections look broken. Place a connection into a broken state etc.
* Create a post, and note that the connection is disabled.
* Without this change, clicking on the connection will still toggle its state. You might not see it enabled, but the appropriate post meta will be saved with the post, and the UI will update it's messaging.
* With this change that will not happen, and clicking on the connection will have no effect.

### Before
image

### After

image


manzoorwanijk (approved):
Works well

Issue #33174, opened 2023-09-19
Blocks: Switch all Production blocks to be registered via their block.json file


jeherve:
Now that:

- we have `block.json` files for all our blocks (see #32508)
- we can register a block via that metadata (see #32697)

Let's switch all our Production blocks to using those `block.json` files.

```[tasklist]
### Tasks
- [ ] #33086
- [ ] https://github.com/Automattic/jetpack/pull/33573
- [ ] https://github.com/Automattic/jetpack/pull/33604
- [ ] https://github.com/Automattic/jetpack/pull/33667
- [ ] https://github.com/Automattic/jetpack/pull/33682
- [ ] https://github.com/Automattic/jetpack/pull/33689
- [ ] https://github.com/Automattic/jetpack/pull/33694
```

*******

Primary issue: #32602

Related work for Beta and Experimental blocks:

- #32815 
- #33073

PR #33573, opened 2023-10-11
Refactor production blocks registration 2

monsieur-z:
- See https://github.com/Automattic/jetpack/issues/32602
- See https://github.com/Automattic/jetpack/issues/33174

## Proposed changes:

Registers the following production blocks by specifying the path to their block.json file:  _Donations_, _Eventbrite_, _Gif_, _Google Calendar_, _Image Compare_.

This was made possible by https://github.com/Automattic/jetpack/pull/32697. The goal is to avoid the duplication of block metadata and use block.json as the source of truth.

In more details, for each block (when needed):
  - Write reading direction-agnostic CSS
  - Remove the `editorScript` property from `block.json` (automatically added during the build)
  - Convert icon from dashicon to SVG in the `icon` property of `block.json`
  - Remove `icon.js`
  - Import icon helpers from the `shared-extension-utils` package
  - Register the block client and server sides with `block.json`
  - Move data from `index.js` to `block.json` and `editor.js` and delete the file
  - Move data from `attributes.js` to `block.json` and delete the file
  - Fix JS tests

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

pedMtX-RS-p2

## Does this pull request change what data or activity we track or use?

No.

## Testing instructions:

- Spin up a test site
- Create a new post and open the block inserter
- Test that the aforementioned blocks behave as in `trunk`. This PR shouldn't change anything from a user's standpoint.

Note: I haven't been able to test the _Cookie Consent_ block.

PR #33604, opened 2023-10-13
Refactor production blocks registration 3

monsieur-z:

## Proposed changes:

Registers the following production blocks by specifying the path to their block.json file:  _Mailchimp_, _Map_, _Markdown_, _OpenTable_, _Payment Buttons_.

  - Remove `icon.js`
  - Fix JS tests

### Other information:

## Jetpack product discussion

pedMtX-RS-p2

No.

## Testing instructions:

- Spin up a test site

Note: the Map block is currently broken on `trunk`. This PR fixes it: https://github.com/Automattic/jetpack/pull/33606.


jeherve (changes requested):
This one will need a rebase as well, I'm afraid.

jeherve (commented):
This tests well, but introduces the same issue with the Payments intro block, that I mentioned in the last PR (#33573)

image

jeherve (approved):

PR #33607, opened 2023-10-13
Update unpurchased backup card state my jetpack

CodeyGuyDylan:
## Proposed changes:

* Replace dropdown menu with the product logo
* Make sure all cards are the same height
* Pull in product data to Backup card
* Display site data to Backup card when card is in unpurchased state
* Set up infrastructure to handle the purchase state next

### Other information:

## Jetpack product discussion

P2: p1HpG7-oSY-p2
Figma: u7alAliUsO2ePGqbrxcGwC-fi-3813%3A7585

No

## Testing instructions:

1. Make sure to test this PR in conjunction with this diff: D123259-code
2. Get your local environment running or test this branch using the Jetpack beta plugin
3. Make sure your site has some pages/posts/images/etc. What can be displayed on the card are Pages, Posts, Comments, Images, Videos, and Audio files
4. Without a Backup plan, go to `/wp-admin/admin.php?page=my-jetpack`
5. Make sure the loading state looks alright for the site data on the card

NOTE: I wanted to do more than just display "Loading...", but it looks like that's what is normal around the plugin for now so we'll probably want to update all those at once if we do end up doing so

image

6. Once the card loads your site's data, make sure it looks like the Figma comps and that it scales well to mobile (should switch to showing 2 items on certain screen widths)

image

7. If you want to test out various values, you can manually edit the values on the diff you patched on your sandbox in the `wp-content/lib/jetpack-product-data/class.backups.php` file lines 128-144

8. Purchase VaultPress backup and go back to `/wp-admin/admin.php?page=my-jetpack`, make sure the card looks normal. We will be adding a purchased state for the data later in a separate PR

image


projects/js-packages/components/components/icons/index.tsx
```diff
@@ -213,6 +223,7 @@ export const ShareIcon: React.FC< BaseIconProps > = ( { size = 16, className, co
 };
 
 const jetpackIcons = {
+	'jetpack-ai': JetpackAiIcon,
```
elliottprogrammer: I recommend we should update the `README.md` file to include the new 'jetpack-ai' icon.
(`projects/js-packages/components/components/icons/README.md`).

projects/js-packages/components/components/icons/index.tsx
```diff
@@ -223,6 +234,7 @@ const jetpackIcons = {
 	search: SearchIcon,
 	social: SocialIcon,
 	star: StarIcon,
+	stats: StatsIcon,
```
elliottprogrammer: Same here for the 'stats' icon, I recommend we should add it to the `README.md` file.

elliottprogrammer (dismissed):
Tests well! It works as described. Code changes look good, I only think we should add the AI and Stats icons to the Icons README.md file, as stated in the code comments below.
Otherwise, LGTM! 👍

projects/js-packages/components/components/icons/index.tsx
CodeyGuyDylan: Thank you for this catch! I'll get those added 😄

projects/packages/my-jetpack/_inc/components/product-cards-section/backup-card/index.jsx
```diff
@@ -0,0 +1,138 @@
...
+		if ( siteData[ key ] === 0 ) {
+			return;
+		}
+
+		sortedData.push( [ key, siteData[ key ] ] );
+	} );
+
+	sortedData.sort( ( a, b ) => {
+		return a[ 1 ] < b[ 1 ] ? 1 : -1;
+	} );
```
jboland88: We should probably memoize this sort operation so that it does not happen on every render

projects/packages/my-jetpack/_inc/components/product-card/index.jsx
```diff
@@ -297,20 +139,6 @@ ProductCard.propTypes = {
...
-	menuItems: PropTypes.arrayOf(
-		PropTypes.shape( {
-			label: PropTypes.string,
-			icon: PropTypes.node,
-			onClick: PropTypes.func,
-		} )
-	),
-	onInstallStandalone: PropTypes.func,
-	onActivateStandalone: PropTypes.func,
-	onDeactivateStandalone: PropTypes.func,
```
jboland88: I notice by removing the menu that we are removing the action to install the standalone plugins from My Jetpack.

Just confirming that this is intended @eeeeeevon2023? Do we intend to provide another way for users to install the standalone plugins. If you don't have the main Jetpack plugin, there is an "install" option on each card, but if you do have the main Jetpack plugin, there's not a clear install path.

Since the standalone plugins offer the most updated UI and the latest features, I think we should still have a clear way to install them for each product from My Jetpack.

projects/packages/my-jetpack/src/products/class-backup.php
```diff
@@ -75,7 +75,11 @@ public static function get_title() {
 	 * @return string
 	 */
 	public static function get_description() {
-		return __( 'Save every change', 'jetpack-my-jetpack' );
+		if ( static::is_active() ) {
```
jboland88, line 78: I am still working on fully testing how this behaves - If this `is_active` call depends on the backup plugin being active, I wonder if we would get a scenario where the user has backups active but sees "Your site is not backed up" 🤔

jboland88 (commented):
I ran into a blocker with functional testing (reverting a different PR) that will require rebasing this PR 🙃 

I think the code looks OK, I had one suggestion to memoize and I am wanting to confirm that we are removing the controls to install standalone plugins from My Jetpack.

projects/packages/my-jetpack/_inc/components/product-card/index.jsx
eeeeeevon2023: >  I notice by removing the menu that we are removing the action to install the standalone plugins from My Jetpack.
Good catch. I noticed only today that the Figma file for the cards is out of date. The menu should not be removed. The rest of the tasks are correct. Let me rework this a bit tomorrow. Sorry-the covid brain fog has been really getting me lately. Give me tomorrow to sort this out!

projects/packages/my-jetpack/src/products/class-backup.php
jboland88: I was able to check this, the `Hybrid_Product` class checks if the standalone plugin is active OR if the main Jetpack plugin is active, so this check works fine.

projects/packages/my-jetpack/_inc/components/product-cards-section/index.jsx
```diff
@@ -21,21 +22,20 @@ const { showJetpackStatsCard = false } = window.myJetpackInitialState?.myJetpack
...
-		items.splice( 6, 0, StatsCard );
-	}
+	const items = {
+		backups: BackupCard,
+		scan: ScanAndProtectCard,
+		antispam: AntiSpamCard,
+		boost: BoostCard,
+		search: SearchCard,
+		videopress: VideopressCard,
+		stats: showJetpackStatsCard ? StatsCard : null,
```
jboland88, line 34: I am seeing the stats card at the top and a stats card in the product grid. I don't think that's caused by this PR, but seems like bug that we should clean up in another PR.

Screenshot 2023-10-19 at 4 19 46 PM

projects/packages/my-jetpack/_inc/components/product-cards-section/backup-card/index.jsx
```diff
@@ -0,0 +1,138 @@
...
+};
+
+// eslint-disable-next-line no-unused-vars
+const BackupCard = ( { admin, productData, fetchingProductData } ) => {
+	const { has_backup: hasBackups = false, site_data: siteData = {} } = productData || {};
+
+	return (
+		<ProductCard admin={ admin } slug="backup" showMenu isDataLoading={ fetchingProductData }>
+			{ /* todo: Add component for when users do have backups and an activity log */ }
+			{ hasBackups ? <></> : <NoBackupsValueSection siteData={ siteData } /> }
```
jboland88, line 135: I think because of the cache, I see the `NoBackupsValueSection` here just after completing the backup purchase. Would it make sense to rely on the product status here that the product card uses to show active/ inactive? That way we aren't introducing a different piece of data that determines if backups are active or inactive?

jboland88 (changes requested):
This tested well overall!
Let's memoize the sort logic and determine if we want to introduce a new data point for determining if backups are active.

projects/packages/my-jetpack/_inc/components/product-cards-section/index.jsx
CodeyGuyDylan: Added to the maintenance board

projects/packages/my-jetpack/_inc/components/product-cards-section/backup-card/index.jsx
CodeyGuyDylan: I removed the `new_backup` datapoints and went ahead and used the status, thank you for the suggestion!

projects/packages/my-jetpack/_inc/components/product-cards-section/backup-card/index.jsx
CodeyGuyDylan: Good catch

keoshi:
Jumping in for @eeeeeevon2023 and reiterating this point:

> The [ellipsis] menu should not be removed. The rest of the tasks are correct.

This means we shouldn't replace the current functionality where the ellipsis icon/menu is used to highlight the installation of the standalone plugin for each product.

image

Here's an updated look of the broader section as it relates to this PR, which can also be found in Figma — u7alAliUsO2ePGqbrxcGwC-fi-3930_3508 :

image

projects/packages/my-jetpack/_inc/components/product-card/index.jsx
```diff
@@ -253,17 +253,16 @@ const ProductCard = props => {
...
-				)
-			}
+			<Text variant="body-small" className={ styles.description }>
+				{ description }
+			</Text>
+
+			{ isDataLoading ? (
+				<span className={ styles.loading }>{ __( 'Loading…', 'jetpack-my-jetpack' ) }</span>
+			) : (
+				children
```
jboland88, line 263: Should we still retain the `isActive` check here? I think the only other card that has children currently is the VideoPress card.

I don't think this is a blocker, just wondering if this impacts any other cards.

jboland88 (approved):
I think this is ready to go once the backup diff is shipped 👍 

I had one question about how we are choosing what content to show on the cards - but I have a feeling that may get a bit more complex as we go along :)

projects/packages/my-jetpack/_inc/components/product-card/index.jsx
CodeyGuyDylan: As far as I can tell, this doesn't affect any cards since, like you said, they aren't using any children. The reasoning in the comment was that we don't want to show anything to the user unless it's active which is no longer true. 

Looking at the rest of the cards, it appears as though you are correct that only VideoPress uses children, and that card has its own check to not display them if VideoPress is not active so we are good there. I think we are good to remove this check

PR #33609, opened 2023-10-14
SEO Title field: Display char count, even when over suggested limit.

elliottprogrammer:
Jetpack SEO Tools:  Currently the SEO Title & SEO Description fields the show character count and suggested character limits, which is helpful. But when you have too many characters, it gives you a warning message without saying how many characters over the limit you are. So you have to keep deleting until you get under. It would be helpful if the current character count were shown, even after the suggested limit has been reached.

Related to: https://github.com/orgs/Automattic/projects/724/views/2?pane=issue&itemId=36521530

## Proposed changes:

* Updated the block to display the current char count, even after the suggested limit has been reached.

#### Screenshots

BEFORE | AFTER
--- | ---
Markup 2023-10-14 at 03 50 39 | Markup 2023-10-14 at 03 52 38

### Other information:

No
## Testing instructions:

- Do one of these:
    - Run this branch with a local Jetpack development environment ([Jetpack Docker environment](https://github.com/Automattic/jetpack/blob/trunk/tools/docker/README.md#readme)).
    - OR:
    - Spin up a jurassic.ninja or Atomic ephemeral site with Jetpack Beta plugin pointing to this branch (`update/counted-textarea-show-overcount`).
- Connect Jetpack, if not connected.
- Edit a post (any post).
- In the editor settings (right side column), at the top, click the Jetpack icon.
- Click to expand the "SEO" item.
- Click the "Activate Jetpack SEO" button (if not already activated).
- In the "SEO TITLE" field, start typing characters. Once you get past 70 characters, verify you see the addition of "([dd] used)", For example you should see, "It's recommended to use less than 70 characters in this field. (72 used)".  (See screenshots above)
- Do the same check for the "SEO DESCRIPTION" field --> after you get past 70 chars, verify you see the "([dd] used)"


projects/plugins/jetpack/extensions/plugins/seo/counted-textarea.js
```diff
@@ -12,8 +12,12 @@ export const CountedTextArea = ( { suggestedLimit, value, label, ...inputProps }
 			  )
 			: sprintf(
 					/* translators: Placeholder is a number of characters in a sentence. */
-					__( "It's recommended to use less than %d characters in this field.", 'jetpack' ),
-					suggestedLimit
+					__(
+						"It's recommended to use less than %1$d characters in this field. (%2$d used)",
```
jeherve: I would take that opportunity to move away from the double quotes and use a real apostrophe, maybe?

```suggestion
						'It’s recommended to use less than %1$d characters in this field. (%2$d used)',
```

jeherve (dismissed):
This looks good to me. I only have one nitpick, but approving in any case, I let you decide if it's worth changing.

jeherve (approved):

PR #33596, opened 2023-10-13
CRM: Fixing escaping in quote titles when rendered using template placeholders

coder-karen:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3151

## Proposed changes:

* This PR changes the quote sanitization / escaping from `	sanitize_text_field( $_POST['quote_fields']['zbscq_title'] );` to `sanitize_text_field( wp_unslash( $_POST['quote_fields']['zbscq_title'] ) );` in `ZeroBSCRM.AJAX.php`, in line with the sanitization and escaping that happens further down for `$quote_notes`.
* This prevents apostrophes added to new quote titles from being escaped and then the backslash being rendered in the editor, as well as preventing backslashes from being duplicated in the editor as well, when the quote title is rendered via a placeholder in a quote template.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

https://github.com/Automattic/zero-bs-crm/issues/3151

## Does this pull request change what data or activity we track or use?

No.

## Testing instructions:

* On trunk, within a quote template - `wp-admin/admin.php?page=manage-quote-templates` - add a quote title placeholder( `##QUOTE-TITLE##`
* Then, create a new quote - `/wp-admin/admin.php?page=zbs-add-edit&action=edit&zbstype=quote` - making sure to enter a quote title that includes an apostrophe
* Select the template from the dropdown, and the title should display with a backslash added (`Karen's Quote` becomes `Karen\'s Quote`)
* Try adding additional backslashes to the original title on a new quote, and when saving these should be duplicated as well.
* Check this branch out locally or using a Jurassic Ninja site, and repeat the same steps. The title should show as inputted (in terms of backslashes and apostrophes).

Issue #39, opened 2013-12-31
Uninstalling Jetpack warns you about Jetpack AND Tonesque


Ipstenu:
When you uninstall jetpack it tells you that you're also uninstalling Tonesque which ... is confusing. I'm tired of getting support tickets at DreamHost about that :)

PR #33614, opened 2023-10-16
CRM: Fixing HTML rendering issue in quote templates notes field 

coder-karen:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3366

## Proposed changes:

* This PR adds HTML encoding to the `zbsqt_notes` field when it is being 'tidied' (in the `tidy_quotetemplate` function) when retrived from the database, before being rendered. 
* Previously `strip_slashes` was being used, now a combination of `html_entity_decode` and `wp_kses` is used for more flexibility.

### Other information:

## Jetpack product discussion

https://github.com/Automattic/zero-bs-crm/issues/3366
https://github.com/Automattic/jetpack/pull/33596#pullrequestreview-1677168769

No.

## Testing instructions:

* On trunk, within a quote template - `wp-admin/admin.php?page=manage-quote-templates` - add a few words with an apostrophe, and other characters including forward slashes (for example `Karen's notes \\`).
* Click on 'Update Template', notice this becomes `Karen&#039;s notes \`. Update again, and it's now `Karen&amp;#039;s notes` .
* With this PR applied, the note field with those characters does not change.

### DIFF
--- a/projects/plugins/crm/changelog/fix-crm-quote-template-notes-escaping
+++ b/projects/plugins/crm/changelog/fix-crm-quote-template-notes-escaping
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Quote Templates: Fix issue with notes field rendering HTML entities in some cases
--- a/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.QuoteTemplates.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.QuoteTemplates.php
@@ -1056,7 +1056,7 @@ class zbsDAL_quotetemplates extends zbsDAL_ObjectLayer {
      * @return array quotetemplate (clean obj)
      */
     private function tidy_quotetemplate($obj=false,$withCustomFields=false){
-
+			global $zbs;
             $res = false;
 
             if (isset($obj->ID)){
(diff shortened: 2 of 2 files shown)


tbradsha (commented):
This works in this specific context, but I'm a bit nervous changing it directly in the DAL object, as generally escaping should be done as late as possible and changing here could affect things further down the pipeline. Any idea how the similar Quotes object handles this, as it doesn't seem to escape at this point?

tbradsha:
Also of note: Quotes and Quote Templates are currently storing the data in the database differently, so it'd be best to handle them the same in my opinion.

tbradsha:
Digging a bit deeper, quote data is processed in here (using `sanitize_textarea_field()`):
https://github.com/Automattic/jetpack/blob/d75eaf6ea14db5dd405478c4df6dd72b5b73ef7d/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Helpers.php#L2912

Which results in it not being processed here (using `zeroBSCRM_textProcess()`):
https://github.com/Automattic/jetpack/blob/d75eaf6ea14db5dd405478c4df6dd72b5b73ef7d/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Helpers.php#L3223

Perhaps the simplest option for now is to give the quote templates notes field a format of `textarea` and then add a case for `textarea` that properly uses `sanitize_textarea_field()`?

coder-karen:
Thanks for digging into this @tbradsha .

I've also been digging too, came to a similar point:


So both quotes and quote templates have a `save_data` function, and the value in the notes field is the same when being passed into that function, but different when being returned. This is happening because of some code in `zeroBS_buildObjArr` which is called in both `save_data` functions.

Within that function, we are unable to get a `$globFieldVarName` for quote templates, as there is no value set in `$fieldModelsByID` for them, so therefore we are just returning the starting array, or empty array if none set:

```
			if (
				!$zbs->isDAL3() && (empty($globFieldVarName) || $globFieldVarName == false || !isset($GLOBALS[ $globFieldVarName ]))
				) return $retArray;
```

Even if we did have a value, just below we also exclude quote templates from sanitization anyway, as you've shown in your shared snippet.

> Perhaps the simplest option for now is to give the quote templates notes field a format of textarea and then add a case for textarea that properly uses sanitize_textarea_field()?

I'll take a look at that, thanks!

tbradsha (approved):
Works as described, so approving!

PR #33631, opened 2023-10-17
wp-admin Site Default: default navigation to wp-admin when interface option is set to wp-admin

jeroenpf:
https://github.com/Automattic/dotcom-forge/issues/4155

See PT: pesfPa-lH-p2

Some users would like to be able to use the traditional wp-admin navigation. At the moment, this is possible for only certain pages and needs to be set per screen. There are also various screens added in Calypso that are not available in wp-admin but should still be available via the menu.

## Proposed changes:

This PR introduces a change where we read the `wpcom_admin_interface` option for a site. If this value exists, it can be either `wp-admin` or `calypso`. 

When the value is `wp-admin` some alterations have been made to keep the majority of menu items of the admin menu pointing to wp-admin pages (such as posts, pages, comments, media, etc.). At the same time, some items are still pointing to the calypso UI as is the case currently.

### Other information:

## Testing instructions:

- Go to or create a new Atomic site
- Install [Jetpack beta tester](https://jetpack.com/download-jetpack-beta/) plugin
- Via the Jetpack menu on your Atomic site, go to the Jetpack beta tester page, select Jetpack (manage)
- Search for the feature branch `add / atomic wpadmin menu` and activate that branch. 
- Your Atomic site should now be using that specific Jetpack plugin branch
- At this point, the menu should look like the regular Atomic site menu.
- Via `_cli` run `wp option update wpcom_admin_interface wp-admin` This will set the option that enables the default menu. 
- Check that the menu has changed and that the most menu items now point to wp-admin instead of Calypso.


wojtekn:
@jeroenpf I tested it, and it works as expected. Would it make sense to adjust menu item ordering to unify the experience? Or is one order WP core default and another one Calypso?

Screenshot 2023-10-19 at 12 06 01

Screenshot 2023-10-19 at 12 05 47

wojtekn (approved):

jeroenpf:
> @jeroenpf I tested it, and it works as expected. Would it make sense to adjust menu item ordering to unify the experience? Or is one order WP core default and another one Calypso?
> 
> Screenshot 2023-10-19 at 12 06 01
> 
> Screenshot 2023-10-19 at 12 05 47

Thanks for reviewing and for pointing that out. I think one will be the core default ordering as essentially we do not touch the order of the menu items when we enable the wp-admin interface setting. I will verify this before merging.

@wojtekn - I just verified that the order of the menu items for the wp-admin option follows the same order as an original wp install. I think we would want to keep the menu as similar as possible to the original wp-admin.

katinthehatsite (commented):
I think the PR does what it says it would do. After I set the option to `wp-admin` on WoA site, I could see the menu change to the following format:

Screenshot 2023-10-19 at 12 06 22 PM

I clicked through the following menu items and they all worked as expected:

```
* Posts/Pages
* Media
* Comments
* Users
* General Settings
* Writing Settings
* Reading Settings
* Discussion Settings
* Sharing Settings
```

All of these sent me to `wp-admin` which is what I am expecting. 

One thing that I noticed: 

* When I navigate to **Plugins -> Installed Plugins (looks good) -> Add New (gets me back to Calypso enhanced page)**. But if I go to **Installed Plugins -> Add new**, I am getting the expected `wp-admin` interface so this is sort of inconsistent. See below:

Screenshot 2023-10-19 at 12 13 29 PM

We can probably handle this in the follow-up issue when we check for these inconsistencies further. I think in the scope of this PR it is okay to ship as is.

katinthehatsite (approved):

jeroenpf:
> I think the PR does what it says it would do. After I set the option to `wp-admin` on WoA site, I could see the menu change to the following format:
> 
> Screenshot 2023-10-19 at 12 06 22 PM
> I clicked through the following menu items and they all worked as expected:
> 
> ```
> * Posts/Pages
> * Media
> * Comments
> * Users
> * General Settings
> * Writing Settings
> * Reading Settings
> * Discussion Settings
> * Sharing Settings
> ```
> 
> All of these sent me to `wp-admin` which is what I am expecting.
> 
> One thing that I noticed:
> 
> * When I navigate to **Plugins -> Installed Plugins (looks good) -> Add New (gets me back to Calypso enhanced page)**. But if I go to **Installed Plugins -> Add new**, I am getting the expected `wp-admin` interface so this is sort of inconsistent. See below:
> 
> Screenshot 2023-10-19 at 12 13 29 PM
> We can probably handle this in the follow-up issue when we check for these inconsistencies further. I think in the scope of this PR it is okay to ship as is.

Thanks! 

I think we need to look out for pages that we already served via wp-admin but that are linking to pages in Calypso. That is the same situation as you mentioned here. 

I will consider this fix done for now and I'll create a new issue for the plugins page.

jeherve:
It looks like this PR caused an issue with stats on WordPress.com Simple:
https://github.com/Automattic/wp-calypso/issues/83268

Would something like this be acceptable? #33702

PR #33635, opened 2023-10-17
Require that new projects have an initial change entry

anomiex:
## Proposed changes:

`jetpack build` chokes if there is no changelog entry at all, which can happen on creation of a new project if that project doesn't contain an initial change entry file.

The fix for this is threefold:

* Have `jetpack generate` include an initial change entry.
* Have `tools/check-changelogger-use.php` not count the addition of CHANGELOG.md as having a change entry.
* Improve the error produced by `jetpack build` to indicate that an initial change entry file may be needed.

This also fixes two small bugs in related code:

* `tools/check-changelogger-use.php` was trimming the first path character of uncommitted files in some cases.
* `jetpack changelog add` was counting `.gitkeep` as a change entry file.

### Other information:

## Jetpack product discussion

p1697549556880869/1697547904.858059-slack-CBG1CP4EN

No

## Testing instructions:

* Run `jetpack generate` and follow the prompts. Verify that the new project's `changelog/` dir contains a change entry.
* `git add` and `git commit` the new project, then run `tools/check-changelogger-use.php trunk HEAD`. It should complete successfully with no error output.
* Remove the change entry from the `changelog/` dir in your new project and commit. Run `tools/check-changelogger-use.php trunk HEAD` again. It should fail with an error about the lack of a change entry in your new project.
* Run `jetpack changelog add`. It should prompt to add a change entry to your new project.

### DIFF
--- a/tools/check-changelogger-use.php
+++ b/tools/check-changelogger-use.php
@@ -158,7 +158,7 @@ foreach ( glob( 'projects/*/*/composer.json' ) as $file ) {
 debug( 'Checking diff from %s...%s.', $base, $head );
 $pipes = null;
 $p     = proc_open(
-	sprintf( 'git -c core.quotepath=off diff --no-renames --name-only %s...%s', escapeshellarg( $base ), escapeshellarg( $head ) ),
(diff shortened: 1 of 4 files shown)


sdixon194 (approved):
Works perfectly for me.

PR #33641, opened 2023-10-17
Publicize: do not continuously watch over block changes

jeherve:
## Proposed changes:

Publicize to Twitter is no longer available, so we no longer need this feature loaded in the block editor. This should improve the block editor's performance.

### Other information:

## Jetpack product discussion

Related discussion: 8586-gh-jpop-issues

* No

## Testing instructions:

* Not much to test since the feature was removed from the interface a while ago.
* You should not see any JS errors in your browser console.

### DIFF
--- a/projects/plugins/jetpack/extensions/plugins/publicize/index.js
+++ b/projects/plugins/jetpack/extensions/plugins/publicize/index.js
@@ -8,7 +8,6 @@
  * displays the Publicize UI there.
  */
 import {
-	TwitterThreadListener,
 	PublicizePanel,
 	useSocialMediaConnections,
 	usePublicizeConfig,
@@ -64,8 +63,6 @@ const PublicizeSettings = () => {
 
 	return (
 		<PostTypeSupportCheck supportKeys="publicize">
-			<TwitterThreadListener />
-
 			<JetpackPluginSidebar>
 				<PublicizePanel enableTweetStorm={ true }>
 					<UpsellNotice />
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/plugins/publicize/index.js
```diff
@@ -64,8 +63,6 @@ const PublicizeSettings = () => {
 
 	return (
 		<PostTypeSupportCheck supportKeys="publicize">
-			<TwitterThreadListener />
-
 			<JetpackPluginSidebar>
 				<PublicizePanel enableTweetStorm={ true }>
```
pablinos, line 67: I think we could remove the `enableTweetStorm` prop too, but we could pick that up in another change.

pablinos (approved):
I was struggling to replicate the performance issues, but this should definitely help as it's unnecessary work.

projects/plugins/jetpack/extensions/plugins/publicize/index.js
jeherve: Yeah, there are many other things we could remove I think. Is removing traces of Twitter elements something that was on y'all's to-do list? I looked for a tracking issue, couldn't find any, but I assume that's something that came up when we deprecated Publicize to Twitter?

pablinos: Yes it came up, but I think there was a hope that we might be able to bring it back too, so we only removed the essential parts.

I've added a task to our backlog to tidy this up

jeherve: > I've added a task to our backlog to tidy this up

To close the loop on this, this is being removed in #33723.

PR #33645, opened 2023-10-17
CRM: Allow custom fields on transaction creation/update API endpoint

cleacos:
Fixes https://github.com/Automattic/zero-bs-crm/issues/3367

## Proposed changes:

This PR updates the transaction creation/update API endpoint to allow custom fields.

### Other information:

## Jetpack product discussion
N/A

N/A

## Testing instructions:

- In your JPCRM testing site, allow API connections. (Core Module)
- Create a custom field for transactions. (ie, `state`)
- Using an API tool like curl, Postman, or Insomnia, create a new transaction using custom fields.
   - `POST http://your-local-site.me/zbs_api/create_transaction/?api_key=jpcrm_pk_xxxx&api_secret=jpcrm_sk_xxxx`
```
{
	"orderid": "Transaction-1",
         "email": "me@example.com",
	"status": "Completed",
	"state": "Colorado",
	"total": 10000
}
```
- In trunk, the transaction will be created without the custom field
- In this branch, the transaction will be create with the custom field


projects/plugins/crm/api/create_transaction.php
```diff
@@ -27,19 +27,37 @@
 		$json_params = file_get_contents( 'php://input' );
 		$new_trans   = json_decode( $json_params, true );
 
-		// REQUIRED
-		$orderid = -1;
+// REQUIRED FIELDS
+$required_fields = array(
+	'orderid',
+	'email',
+);
```
tbradsha: I don't understand why we're doing this?

projects/plugins/crm/api/create_transaction.php
```diff
@@ -27,19 +27,37 @@
...
+			array(
+				'error'   => 400,
+				'message' => 'Missing required field: ' . $field,
+			)
+		);
+		exit();
+	}
+}
+
+$orderid = -1;
```
tbradsha, line 49: In the current implementation (which I don't think we should merge) we wouldn't need "default required" line.

projects/plugins/crm/api/create_transaction.php
```diff
@@ -27,19 +27,37 @@
...
+$email   = '';
```
tbradsha: In the current implementation (which I don't think we should merge) we wouldn't need these "default required" lines.

projects/plugins/crm/api/create_transaction.php
```diff
@@ -27,19 +27,37 @@
...
+$fname   = '';
+
 if ( isset( $new_trans['orderid'] ) ) {
 	$orderid = sanitize_text_field( $new_trans['orderid'] );
```
tbradsha, line 48: Pre-existing, but really should be cast to `int`.

projects/plugins/crm/api/create_transaction.php
tbradsha: I think you've misinterpreted the earlier comment. It's not required for the API call (user-side), but instead is required for the DAL call (server-side).

projects/plugins/crm/api/create_transaction.php
```diff
@@ -107,57 +125,33 @@
...
-			'tax_rate' => $rate,
+			'status'   => '',
+			'total'    => '',
+			'date'     => '',
+			'item'     => '',
+			'net'      => '',
+			'tax'      => '',
+			'fee'      => '',
+			'discount' => '',
+			'tax_rate' => '',
```
tbradsha: Rather than passing blank strings, just remove these...

projects/plugins/crm/api/create_transaction.php
```diff
@@ -107,57 +125,33 @@
...
 		);
 
+		foreach ( $new_trans as $field => $value ) {
+
+			if ( $field === 'orderid' || $field === 'email' || $field === 'fname' ) {
```
tbradsha: Why get the `orderid` and `fname` separately above rather than processing them here? I would suggest handling "required" blank fields after processing them.

projects/plugins/crm/api/create_transaction.php
```diff
@@ -107,57 +125,33 @@
...
+				continue;
+			}
+			if ( $field === 'date' ) {
+				$transaction_fields[ $field ] = gmdate( 'Y-m-d H:i:s', (int) $value );
+				continue;
+			}
+
+			$transaction_fields[ $field ] = sanitize_text_field( $value );
```
tbradsha: Given the purpose is to handle custom fields, this will not work unless it's a text-type field. Some expect numbers, arrays (or CSV, I can't remember), textarea...in particular, textarea is finicky here.

I believe the "correct" way to handle this would be to use `zeroBS_buildObjArr()` (which has its own problems, but should work in this scenario to sanitise and such).

projects/plugins/crm/api/create_transaction.php
cleacos: To ensure we get the required fields

projects/plugins/crm/api/create_transaction.php
cleacos: ?
It's required by the DAL, and here we add the check.

Note: Generally as a quick fix/improvement, made on the fly, I left the previous implementation and consideration, like the default values. In this case, I added the check at the API level ($email is required by the API to be able to get the contact id, required by the Transaction DAL)

projects/plugins/crm/api/create_transaction.php
cleacos: > Given the purpose is to handle custom fields, this will not work unless it's a text-type field. Some expect numbers, arrays (or CSV, I can't remember), textarea...in particular, textarea is finicky here.

Another potential issue for the future... For now, it manages well the most custom field types, all the basic ones: text, numbers, and also textareas. I don't see any issue here.

projects/plugins/crm/api/create_transaction.php
cleacos: I added two more fields as required ones: `status` and `total`.

projects/plugins/crm/api/create_transaction.php
cleacos: Removed

projects/plugins/crm/api/create_transaction.php
cleacos: - `fname` is required above, like `email`, to be able to create the contact if it does not exist. They are not part of the transaction object.
- `orderid` is not part of the transaction object. See `addOrUpdateTransaction`. We should use `ref`(I forgot it) instead of. (DALv3). Updated.

projects/plugins/crm/api/create_transaction.php
cleacos: I added `status` and `total` as required fields and removed `email`

projects/plugins/crm/api/create_transaction.php
cleacos: And removed `email` from the required ones. So `$email` and `$fname` will have a default value.

projects/plugins/crm/api/create_transaction.php
cleacos: Nope, the `orderid` is the external transaction reference, not the DB id. It could be a text.

tbradsha (approved):
Works okay. I'd prefer less default checks (pass-thru to DAL handles them), more robust datatype checks (e.g. this fails with textareas + newlines, which has bitten us many times in the past and continues to do so), and direct calls to DAL, but approving anyway.

Issue #31878, opened 2023-07-13
AI Assistant: Keep text formatting


dhasilva:
ref p58i-eRJ-p2#comment-58814

Right now we are losing text formatting and concatenating multiple blocks content, which makes them blend together. We need to find a way to keep different block's contents separated and keep the text formatting.

nunyvega:
@dhasilva are we good to close this or something else is missing?

nunyvega:
Adding a video to show how it's working now after your last PR:

https://github.com/Automattic/jetpack/assets/16329583/7f9b5ed6-e0f8-4624-92e4-0c6a4fbf9b01

dhasilva:
I think we may need some prompt tweaks to keep the formatting when making more significant changes. Some more tests are need before closing this issue.

nunyvega:
What would be an example of a formatting issue that needs fixing?

dhasilva:
Before:
![2023-07-18_16-31-05](https://github.com/Automattic/jetpack/assets/8486249/dd9123e8-8c39-45a6-aa4d-1a198f10da2e)

After grammar correction:
![2023-07-18_16-31-42](https://github.com/Automattic/jetpack/assets/8486249/00e47fc6-23a5-4755-a9ab-4124c42d8489)

Bold and italics are kept, but things like underline and strikethrough are not. And partial word formatting is lost as well.

dhasilva:
Closing as partially done. The underline style seems to be unsolvable in pure Markdown as there is no underline Markdown syntax. This is a spec issue, unfortunately.

PR #33647, opened 2023-10-17
AI Assistant: Add strikethrough Markdown syntax to Markdown generator

dhasilva:
Part of #31878
Intended to be used with D125660-code, but can be tested and used independently.

## Proposed changes:

* Adds the strikethrough Markdown syntax to Turndown. Turns out it was dropped when generating markdown, so the AI never received it in the first place.

Underline, though, is not possible to solve in a consistent manner because there is no Markdown syntax for underlined text. I was able to generate a custom syntax on Turndown, but the AI does not understand it and the Markdown parser would need an extension to be able to process it correctly.

### Other information:

## Testing instructions:

* Go to the block editor and create some text with strikethrough style
* Ask for some change by the AI Assistant
* Look for the network request and check that the content being sent is enclosed in `~~` characters

### DIFF
--- a/projects/plugins/jetpack/changelog/update-ai-assistant-style
+++ b/projects/plugins/jetpack/changelog/update-ai-assistant-style
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
+AI Assistant: Add strikethrough Markdown syntax to Markdown generator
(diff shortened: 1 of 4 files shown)


lhkowalski (approved):
Code looks good.

The strikethrough formatting is kept and translated to markdown before the request:

Screenshot 2023-10-17 at 17 58 51

I tested without and with the backend patch.

Without the backend patch, I got a bad result:

Screenshot 2023-10-17 at 17 59 15

Using the backend patch, I got the right result:

Screenshot 2023-10-17 at 18 01 54

---

Nice work using the library capabilities!

PR #33653, opened 2023-10-18
Subscribe modal: simplify URL construction

simison:
## Proposed changes:

* Simplify the way we construct URL for the iframe by using `URLSearchParams`, and move everything together for easier reading
* Check only for `email` before bailing out for missing email

### Other information:

## Testing instructions:

* Have a post with paywall, and subscribe. Modal should work as before.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/view.js
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/view.js
@@ -14,13 +14,13 @@ domReady( function () {
 	if ( ! form.payments_attached ) {
 		form.payments_attached = true;
 		form.addEventListener( 'submit', function ( event ) {
-			event.preventDefault();
-
 			const email = form.querySelector( 'input[type=email]' ).value;
 
 			if ( form.resubmitted || ! email ) {
 				return;
 			}
+
+			event.preventDefault();
 
 			const post_id = form.querySelector( 'input[name=post_id]' )?.value ?? '';
 			const tier_id = form.querySelector( 'input[name=tier_id]' )?.value ?? '';
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/view.js
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/view.js
@@ -14,33 +14,33 @@ domReady( function () {
 	if ( ! form.payments_attached ) {
 		form.payments_attached = true;
 		form.addEventListener( 'submit', function ( event ) {
+			event.preventDefault();
+
 			const email = form.querySelector( 'input[type=email]' ).value;
-			const email_clause = email ? `&email=${ encodeURIComponent( email ) }` : '';
-			const post_id = form.querySelector( 'input[name=post_id]' )?.value;
-			const post_id_clause = post_id ? `&post_id=${ post_id }` : '';
-			const tier_id = form.querySelector( 'input[name=tier_id]' )?.value;
-			const tier_id_clause = tier_id ? `&tier_id=${ tier_id }` : '';
 
 			if ( form.resubmitted || ! email ) {
 				return;
 			}
-			event.preventDefault();
 
-			const url =
-				'https://subscribe.wordpress.com/memberships/?' +
-				'blog=' +
-				form.dataset.blog +
-				'&plan=newsletter' +
-				'&source=jetpack_subscribe' +
-				'&post_access_level=' +
-				form.dataset.post_access_level +
-				'&display=alternate' +
-				post_id_clause +
-				tier_id_clause +
-				email_clause;
+			const post_id = form.querySelector( 'input[name=post_id]' )?.value ?? '';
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/subscriptions/view.js
```diff
@@ -14,33 +14,33 @@ domReady( function () {
...
+				source: 'jetpack_subscribe',
+				post_access_level: form.dataset.post_access_level,
+				display: 'alternate',
+				TB_iframe: true,
+			} );
+
+			const url = 'https://subscribe.wordpress.com/memberships/?' + params.toString();
 
 			window.scrollTo( 0, 0 );
-			tb_show( null, url + '&TB_iframe=true', null );
```
millerf, line 43: That is used by thickBox and probably should not be removed...

millerf: NVM. I see it was added to the iframe URL...
I would keep it outside so we know it is only for thickbox...

projects/plugins/jetpack/extensions/blocks/subscriptions/view.js
```diff
@@ -14,33 +14,33 @@ domReady( function () {
...
-			const email_clause = email ? `&email=${ encodeURIComponent( email ) }` : '';
-			const post_id = form.querySelector( 'input[name=post_id]' )?.value;
-			const post_id_clause = post_id ? `&post_id=${ post_id }` : '';
-			const tier_id = form.querySelector( 'input[name=tier_id]' )?.value;
-			const tier_id_clause = tier_id ? `&tier_id=${ tier_id }` : '';
 
 			if ( form.resubmitted || ! email ) {
 				return;
 			}
-			event.preventDefault();
```
millerf, line 27: I am not sure about what is the `form.resubmitted` use-case (and I could not find any data on it), but 
```
event.preventDefault();
```
should maybe stay after that check? So it can be handled natively?

projects/plugins/jetpack/extensions/blocks/subscriptions/view.js
simison: I can add a note on the line to make it clear; I find it easier to keep things together in one spot vs previously URL was split into 3 blocks of code. What do you think?

projects/plugins/jetpack/extensions/blocks/subscriptions/view.js
simison: Sure! I'll debug a bit and put it down if it affects.

projects/plugins/jetpack/extensions/blocks/subscriptions/view.js
```diff
@@ -14,33 +14,33 @@ domReady( function () {
...
-				post_id_clause +
-				tier_id_clause +
-				email_clause;
+			const post_id = form.querySelector( 'input[name=post_id]' )?.value ?? '';
+			const tier_id = form.querySelector( 'input[name=tier_id]' )?.value ?? '';
+
+			const params = new URLSearchParams( {
+				email: encodeURIComponent( email ),
+				post_id,
+				tier_id,
```
millerf, line 31: I checked on the back-end, and that works fine. I was skeptical if `null` and '' values were handled the same way. 
It seems like it: fbhepr%2Skers%2Sjcpbz%2Sjc%2Qpbagrag%2Szh%2Qcyhtvaf%2Sfhofpevor%2Qgb%2Qoybtf.cuc%3Se%3Qq4pqnn54%23216%2Q232-og

simison: Nice, thanks! Empty strings are little less prone to cause accidental problems e.g. with `empty()` checks rather than string-nulls.

millerf: >  e.g. with empty() checks rather than string-nulls.

Of course. But you never know how the backend was built until you check it...

projects/plugins/jetpack/extensions/blocks/subscriptions/view.js
millerf: Don't get me wrong. This refactor is way nicer.
But I'd personaly keep `url + '&TB_iframe=true'` here as it is very specific for thickbox... I prefer isolating concepts.

As we will end up removing thickbox soon, I can bet that  `TB_iframe=true` will be overlooked and kept in the URL if we add your change...

simison: sure 👍

projects/plugins/jetpack/extensions/blocks/subscriptions/view.js
simison: I didn't really find a way to test `resubmitted` quickly (would be curious to learn tho) so I just put it back where it was, not a biggie.

millerf (approved):
This looks way cleaner! :shipit:

PR #33656, opened 2023-10-18
[newsletters] retrieve subscriptions for self hosted

millerf:
Fixes https://github.com/Automattic/gold/issues/171
Fixes https://github.com/Automattic/gold/issues/164

Needs D125802-code

## Proposed changes:

* Allow mail-login so subscriptions are retrieved via email code

### Other information:

## Testing instructions:

* Launch a JN instance
* Connect JP
* Link JETPACK__SANDBOX_DOMAIN to your sandbox (don't forget `sudo global-http enable`)	
* Enable Earn (Stripe, add a plan)
* Patch D125802-code on your sandbox
* Sandbox subscribe.wordpress.com
* Create a post with paid-subscribers access
* Log using the "login" link
  
Screenshot 2023-10-19 at 11 57 24

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -1079,14 +1079,14 @@ function get_paywall_blocks( $newsletter_access_level ) {
 		$id           = '';
 		$sign_in_link = wpcom_logmein_redirect_url( get_current_url(), false, null, 'link' );
 		$button_text  = esc_html__( 'Log in', 'jetpack' );
-	} elseif ( class_exists( 'Automattic\Jetpack\Connection\Tokens\Jetpack_Token_Subscription_Service' ) &&
+	} elseif ( class_exists( 'Automattic\Jetpack\Extensions\Premium_Content\Subscription_Service\Jetpack_Token_Subscription_Service' ) &&
 		! $host->is_wpcom_simple() ) {
 		// We are on Jetpack and no cookie is set
 		$id              = 'jp_retrieve_subscriptions_link';
 		$sign_in_link    = '#'; // listening to "click" event in view.js
 		$access_question = __( 'I am already subscribed.', 'jetpack' );
 		if ( ! Jetpack_Token_Subscription_Service::has_token_from_cookie() ) {
-			$button_text = esc_html__( 'Retrieve subscriptions', 'jetpack' );
+			$button_text = esc_html__( 'Log in', 'jetpack' );
 		} else {
 			$button_text = esc_html__( 'Switch accounts', 'jetpack' );
 		}
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -1076,15 +1076,17 @@ function get_paywall_blocks( $newsletter_access_level ) {
 	$sign_in             = '';
 	$show_button_sign_in = false;
 	$host                = new Host();
-	if ( ! is_user_logged_in() && $host->is_wpcom_simple() ) {
+	if ( false && ( ! is_user_logged_in() && $host->is_wpcom_simple() ) ) {
+		$id                  = '';
 		$show_button_sign_in = true;
 		$sign_in_link        = wpcom_logmein_redirect_url( get_current_url(), false, null, 'link' );
 		$button_text         = esc_html__( 'Log in', 'jetpack' );
-	} elseif ( class_exists( 'Automattic\Jetpack\Connection\Tokens\Jetpack_Token_Subscription_Service' ) &&
-		! $host->is_wpcom_platform() ) {
+	} elseif ( true || ( class_exists( 'Automattic\Jetpack\Connection\Tokens\Jetpack_Token_Subscription_Service' ) &&
+		! $host->is_wpcom_platform() ) ) {
 		// We are on Jetpack and no cookie is set
 		$show_button_sign_in = true;
-		$sign_in_link        = '';
+		$id                  = 'retrieve_subscriptions_link';
+		$sign_in_link        = 'javascript:show_iframe_retrieve_subscriptions_from_email()';
 		$access_question     = __( 'I am already subscribed.', 'jetpack' );
 		if ( ! Jetpack_Token_Subscription_Service::has_token_from_cookie() ) {
 			$button_text = esc_html__( 'Retrieve subscriptions', 'jetpack' );
@@ -1095,7 +1097,7 @@ function get_paywall_blocks( $newsletter_access_level ) {
 
 	if ( $show_button_sign_in ) {
 		$sign_in = '<!-- wp:paragraph {"align":"center","style":{"typography":{"fontSize":"14px"}}} -->
-<p class="has-text-align-center" style="font-size:14px">' . esc_html( $access_question ) . ' <a href="' . $sign_in_link . '">' . $button_text . '</a></p>
+<p class="has-text-align-center" style="font-size:14px">' . esc_html( $access_question ) . ' <a id="' . $id . '" href="' . $sign_in_link . '">' . $button_text . '</a></p>
 <!-- /wp:paragraph -->';
 	}
 
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -1073,12 +1073,29 @@ function get_paywall_blocks( $newsletter_access_level ) {
 		? __( 'Already a paid subscriber?', 'jetpack' )
 		: __( 'Already a subscriber?', 'jetpack' );
 
-	$sign_in = '';
-	if ( ! is_user_logged_in() && ( new Host() )->is_wpcom_simple() ) {
-		$sign_in_link = wpcom_logmein_redirect_url( get_current_url(), false, null, 'link' );
+	$sign_in             = '';
+	$show_button_sign_in = false;
+	$host                = new Host();
+	if ( ! is_user_logged_in() && $host->is_wpcom_simple() ) {
+		$show_button_sign_in = true;
+		$sign_in_link        = wpcom_logmein_redirect_url( get_current_url(), false, null, 'link' );
+		$button_text         = esc_html__( 'Log in', 'jetpack' );
+	} elseif ( class_exists( 'Automattic\Jetpack\Connection\Tokens\Jetpack_Token_Subscription_Service' ) &&
+		! $host->is_wpcom_platform() ) {
+		// We are on Jetpack and no cookie is set
+		$show_button_sign_in = true;
+		$sign_in_link        = '';
+		$access_question     = __( 'I am already subscribed.', 'jetpack' );
+		if ( ! Jetpack_Token_Subscription_Service::has_token_from_cookie() ) {
+			$button_text = esc_html__( 'Retrieve subscriptions', 'jetpack' );
+		} else {
+			$button_text = esc_html__( 'Change user ', 'jetpack' );
+		}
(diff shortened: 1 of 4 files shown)


projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -1073,12 +1073,29 @@ function get_paywall_blocks( $newsletter_access_level ) {
...
+	if ( ! is_user_logged_in() && $host->is_wpcom_simple() ) {
+		$show_button_sign_in = true;
+		$sign_in_link        = wpcom_logmein_redirect_url( get_current_url(), false, null, 'link' );
+		$button_text         = esc_html__( 'Log in', 'jetpack' );
+	} elseif ( class_exists( 'Automattic\Jetpack\Connection\Tokens\Jetpack_Token_Subscription_Service' ) &&
+		! $host->is_wpcom_platform() ) {
+		// We are on Jetpack and no cookie is set
+		$show_button_sign_in = true;
+		$sign_in_link        = '';
+		$access_question     = __( 'I am already subscribed.', 'jetpack' );
```
simison: You can just update the `$access_question` to always be `Already a subscriber?`, see p9Jlb4-911-p2

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -1073,12 +1073,29 @@ function get_paywall_blocks( $newsletter_access_level ) {
...
+		if ( ! Jetpack_Token_Subscription_Service::has_token_from_cookie() ) {
+			$button_text = esc_html__( 'Retrieve subscriptions', 'jetpack' );
```
simison: We can keep the "log in" terminology here for clarity

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -1073,12 +1073,29 @@ function get_paywall_blocks( $newsletter_access_level ) {
...
+		} else {
+			$button_text = esc_html__( 'Change user ', 'jetpack' );
```
simison: Let's use "Switch accounts", p9Jlb4-911-p2

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -1073,12 +1073,29 @@ function get_paywall_blocks( $newsletter_access_level ) {
...
-		$sign_in_link = wpcom_logmein_redirect_url( get_current_url(), false, null, 'link' );
+	$sign_in             = '';
+	$show_button_sign_in = false;
+	$host                = new Host();
+	if ( ! is_user_logged_in() && $host->is_wpcom_simple() ) {
+		$show_button_sign_in = true;
+		$sign_in_link        = wpcom_logmein_redirect_url( get_current_url(), false, null, 'link' );
+		$button_text         = esc_html__( 'Log in', 'jetpack' );
+	} elseif ( class_exists( 'Automattic\Jetpack\Connection\Tokens\Jetpack_Token_Subscription_Service' ) &&
+		! $host->is_wpcom_platform() ) {
```
simison: Will this work also for Atomic site?

Neither of the cases otherwise cover Atomic sites.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -1073,12 +1073,31 @@ function get_paywall_blocks( $newsletter_access_level ) {
...
+	if ( ! is_user_logged_in() && $host->is_wpcom_simple() ) {
+		$id                  = '';
+		$show_button_sign_in = true;
+		$sign_in_link        = wpcom_logmein_redirect_url( get_current_url(), false, null, 'link' );
+		$button_text         = esc_html__( 'Log in', 'jetpack' );
+	} elseif ( class_exists( 'Automattic\Jetpack\Connection\Tokens\Jetpack_Token_Subscription_Service' ) &&
+		! $host->is_wpcom_platform() ) {
+		// We are on Jetpack and no cookie is set
+		$show_button_sign_in = true;
+		$id                  = 'retrieve_subscriptions_link';
```
simison: Prolly good to prefix with `jetpack_` just in case.

Any situation where we might have multiple of these links at the same view? `id` should be just once on the page.

projects/plugins/jetpack/extensions/blocks/subscriptions/view.js
```diff
@@ -6,7 +6,48 @@ import '../../shared/memberships.scss';
 import domReady from '@wordpress/dom-ready';
 import { handleIframeResult } from '../../../extensions/shared/memberships';
 
+// @ts-ignore
+function show_iframe_retrieve_subscriptions_from_email() {
+	const form = document.querySelector( '.wp-block-jetpack-subscriptions__container form' );
```
simison, line 11: Note that you might have multiple subscribe forms on the page simultaneously, and you might end up picking the email input from wrong one.

You could generate the whole subscribe block container an unique IDs across sessions, and rely on that.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -1073,12 +1073,31 @@ function get_paywall_blocks( $newsletter_access_level ) {
...
+		$sign_in_link        = 'javascript:show_iframe_retrieve_subscriptions_from_email()';
+		$access_question     = __( 'I am already subscribed.', 'jetpack' );
+		if ( ! Jetpack_Token_Subscription_Service::has_token_from_cookie() ) {
+			$button_text = esc_html__( 'Retrieve subscriptions', 'jetpack' );
+		} else {
+			$button_text = esc_html__( 'Change user ', 'jetpack' );
+		}
+	}
 
+	if ( $show_button_sign_in ) {
```
simison: You could check if `$sign_in_link` is empty or not, and avoid the separate var for the flag. Not a biggie either way.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
lezama: I like not using "log in", specially with this approach where we are not really login the user, in particular for Jetpack sites. "Get access" maybe?

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
lezama: @millerf in which case we don't show the button?

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
simison: I get where you both are coming with this (I talked about it with Fab in Malaga), but I don't think it really matters or that there would be confusion; it's right in the context of paywall, and they know they are subscriber, so I don't think we'll have issues. 

Let's not "optimize for minority with expense of clarity for majority". :-) There will always be very few admins of the site (who might expect to login to site admin) vs very many subscribers/readers. The few admins will learn quickly they can't use this to actually login.

"Sign in/log in" is very established wording too — and thus easy to translate in every language. Anything else will quickly get fuzzier quickly, especially across translations.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -1073,12 +1073,31 @@ function get_paywall_blocks( $newsletter_access_level ) {
...
+		$id                  = '';
+		$show_button_sign_in = true;
+		$sign_in_link        = wpcom_logmein_redirect_url( get_current_url(), false, null, 'link' );
+		$button_text         = esc_html__( 'Log in', 'jetpack' );
+	} elseif ( class_exists( 'Automattic\Jetpack\Connection\Tokens\Jetpack_Token_Subscription_Service' ) &&
+		! $host->is_wpcom_platform() ) {
+		// We are on Jetpack and no cookie is set
+		$show_button_sign_in = true;
+		$id                  = 'retrieve_subscriptions_link';
+		$sign_in_link        = 'javascript:show_iframe_retrieve_subscriptions_from_email()';
```
simison: Minor but probably best to just have `#` (or real fallback link) here and attach the click listener directly in JS after DOM finishes loading.

If the site is implementing [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP), they'll likely forbid JS in `href`. Ok to do in follow-up PR.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
simison: Just on Atomic currently in this PR, I think.

simison:
One of the follow-ups later on is to move this login link to be part of the subscribe block. That way the login link can be made visible under the block just with block setting, also outside the paywall.

lezama:
I am not getting the link on a Jetpack site (logged out)

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
millerf: I actually forgot to remove it, but this is how it is implemented (via DOM click event)

projects/plugins/jetpack/extensions/blocks/subscriptions/view.js
millerf: I agree. But I am using the same logic as the "payment" part. Not perfect, and needs improvement. Yes, like every bit of code on Earn...

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
millerf: > id should be just once on the page.

I know. But there seems to be a baked in that there is only one subscription form...

simison (commented):
- Testing on self hosted site with Firefox, incognito mode.
- Entering email of user who is a subscriber at the site.


https://github.com/Automattic/jetpack/assets/87168/0c19df80-f21c-420f-8e3a-a87009290263


Problems:
- When pressing "subscribe" button I get the code
- When pressing login, no code and it treats me as new subscriber.

image

millerf:
> * Testing on self hosted site with Firefox, incognito mode.

You forgot to patch your sandbox. Please follow the testing instructions...

lezama:
hmm, disregard my previous comment, I think I was testing something different

lezama:
I think if we change the button text to "log in" it would make the flow clearer for the user: 

https://github.com/Automattic/jetpack/assets/104869/3418cd6c-4d24-4337-b24f-8260af7e440d

After entering the email and hitting "log in" again I got this:

Screenshot 2023-10-19 at 09 53 01

(verified in another tab that I was logged out from wpcom)

chrismccluskey (approved):
LGTM. Login works just fine. Instructions did not mention to subscribe to site first, so I didn't - I was able to login, and even switch accounts, though I could not access the content because I was not a subscriber. I subscribed and opened a new private window. When I logged in as a subscriber, I was able to see the content.

millerf:
Got merged by mistake. Reverting.

projects/plugins/jetpack/extensions/blocks/subscriptions/view.js
simison: The line caused bugs when there were multiple subscribe blocks on the page, and we redirected to wrong blog posts after confirming subscription. Fixed in https://github.com/Automattic/jetpack/pull/33932

millerf: I took that into account and that was fixed in https://github.com/Automattic/jetpack/pull/33800/files#diff-364058ace64d8c69cdbbac7d7cc6afd5116da163ce99d8e2928b5a5194de4487R14, but you do not want that approach.

PR #33657, opened 2023-10-18
Jetpack Boost: Bare bones React DataSync Prototype

pyronaur:
## Proposed changes:
This is still WIP but I need to shift gears a bit before I can continue working on this. I'm going to refactor Image Size Analyzer part of the dashboard as I implement this further.

### Other information:

## Jetpack product discussion
N/A

N/A

## Testing instructions:
🌈


projects/js-packages/react-data-sync-client/src/DataSync.ts
```diff
@@ -0,0 +1,289 @@
+import { z } from 'zod';
```
pyronaur, line 1: ❗ This file is a carbon copy from `svelte-data-sync-client` - we'll remove svelte-data-sync-client eventually

projects/plugins/boost/app/assets/src/js/react-components/pages/settings/elements/MinifyMeta.tsx
```diff
@@ -0,0 +1,88 @@
...
+	inputLabel: string;
+	buttonText: string;
+	placeholder: string;
+	value: string[];
+}
+
+let nextIdIndex = 0;
+
+const MetaComponent = ( { inputLabel, buttonText, placeholder, type }: Props ) => {
+	const { data, mutate } = useDataSync( 'jetpack_boost_ds', type, minifyMetaTypes[ type ] );
```
haqadn: Can we call it something other than `type`? In this context where I don't know what the parameters of `useDataSync` are, `type` is a bit misleading. Maybe something like `option` or `optionKey`?

projects/plugins/boost/app/assets/src/js/react-components/pages/settings/elements/MinifyMeta.tsx
```diff
@@ -0,0 +1,88 @@
...
+						) }
+					</div>
+
+					<button
+						type="button"
+						className="components-button is-link"
+						onClick={ () => setIsEditing( true ) }
+					>
+						<img
+							src="/wp-content/plugins/jetpack-boost/app/assets/src/js/svg/pencil.svg"
```
haqadn: This wouldn't work if the wp is installed in a subdirectory or they have some weird setup for plugin directory.

projects/plugins/boost/app/assets/src/css/admin-style.scss
```diff
@@ -24,4 +24,7 @@
 @import 'components/support';
 @import 'components/snackbar';
 @import 'components/icon-tooltip';
-@import 'components/pricing-table.scss';
\ No newline at end of file
+@import 'components/pricing-table.scss';
+
+// React Migration CSS
+@import 'react/pages/settings/elements/MinifyMeta.scss'
```
haqadn: I wonder if we should set a proper directory structure for all react components. Something like storybooks would be nice where  all assets of a component lives in the same folder with it. This structure should also allow easy storybook integration.

But, this is probably out of scope for this PR.

projects/js-packages/react-data-sync-client/src/DataSyncHooks.ts
```diff
@@ -0,0 +1,48 @@
...
+		arg: Value;
+	};
+	const datasync = new DataSync( namespace, key, schema );
+	const { trigger, isMutating } = useSWRMutation< Value >(
+		key,
+		async ( _, data: MutationArguments ) => {
+			const value = data.arg; // schema.parse( data.arg );
+			const response = await datasync.SET( value );
+			// eslint-disable-next-line no-console
+			console.log( 'Response', response );
```
haqadn: Missed by accident?

haqadn (changes requested):
This looks very neat. One thing I am confused about, Are we handling the API calls within DS instead of letting SWR handle it? Does SWR not handle the API calls or are there advantages to keeping it within DS?

What would be SWRs role in case we handle the API calls in DS? Parsing and dispatching the requests?

projects/plugins/boost/app/assets/src/css/admin-style.scss
dilirity: I checked how other JP plugins are doing it. For example, Social's `connections` component - https://github.com/Automattic/jetpack/tree/e7067c4f42dac3539b1548f074088713f3384276/projects/plugins/social/src/js/components/connections

They have a flat components structure and the styles are in the same folder as each component. In this case though, when I created the MinifyMeta component, I blatantly copy-pasted the styles from the critical-css meta and it's now a mess. I tried rewriting the styles just now, but then I had to go and rewrite the critical-css markup so it can be its own thing and it got out of scope for this PR so I dropped it for now.

As for this PR - think we can leave it for now, but for future PRs we should definitely use the same structure as the rest of JP. We'll also need to rewrite some markup and styles to make them more independent (and less copy-pasted 😅). So your suggestion about storybook sounds like a great idea!

Also - I didn't know React scopes the styles it imports in a module (makes sense though). So we won't need to prefix anything with `jb` like we did for https://github.com/Automattic/jetpack/pull/33679.

Here's what the scoping looks like:
```js
<div className={ styles['critical-css__meta'] }>
```

gets transformed into this

CleanShot 2023-10-20 at 16 41 03

projects/plugins/boost/app/assets/src/js/react-components/pages/settings/elements/MinifyMeta.tsx
dilirity: Thanks for this - I also had a hard time understanding what it is and had to go digging 😅 .

Updated `type` to `optionKey`. I also updated related code in e0ad50932d330b20762937bada1ad91def4e9bc7.

pyronaur:
> Are we handling the API calls within DS instead of letting SWR handle it?

SWR itself doesn't perform network requests - instead it needs a `fetcher` which you can use to plug in your network request function, be that GraphQL, simple `fetch` or `axios` - or as in our case - a custom DataSync instance that handles the authentication with WordPress.

> Does SWR not handle the API calls or are there advantages to keeping it within DS?
> What would be SWRs role in case we handle the API calls in DS? Parsing and dispatching the requests?

The advantage is that we don't have to come up with a new paradigm of mutation and asynchronous state handling - SWR is already battle tested with 1.5m weekly downloads. We provide our callbacks and data, shape the data as necessary - SWR handles the state, mutation, caching, request de-duplication, etc.

projects/plugins/boost/app/assets/src/js/react-components/pages/settings/elements/MinifyMeta.tsx
pyronaur: Yep, this is temporary. Perhaps I should have marked this as a draft pr 😛

projects/js-packages/react-data-sync-client/src/DataSyncHooks.ts
pyronaur: Yesno - this PR isn't the complete DataSync - it's a prototype implementing a simple component. I need to go through a lot of iteration - I'll need this here for now. But, there's a code freeze issue that we need to talk about. I'll slack you.

projects/plugins/boost/app/assets/src/css/admin-style.scss
pyronaur: I created a `react` directory for styles that mimic the location of the component so that when we're ready to migrate to some CSS-in-JS build approach, we can just take the whole react directory and merge it with the actual react components directory and the styles will be right next to the components.

For now, especially for DataSync - I'm not concerned with the styles library or how they're imported in react, - I'm leveraging our global stylesheet instead.

projects/plugins/boost/app/assets/src/js/react-components/pages/settings/elements/MinifyMeta.tsx
dilirity: Fixed in efb8b71.

dilirity: Marking this as resolved as it's been fixed.

dilirity:
Once we sort out the build failing, we should be good to merge.

dilirity:
@haqadn I fixed the build and I think it's ready for merge.

projects/plugins/boost/app/assets/src/js/stores/minify.ts
```diff
@@ -1,15 +1,39 @@
...
+		'jetpack_boost_ds',
+		'config',
+		z.object( {
+			plugin_dir_url: z.string().url(),
+		} )
+	);
+	const { data } = useQuery();
+
+	return data;
+};
```
haqadn, line 39: I think this is meant to be in a separate file and is intended for more generic usage. But, we can address it in a different PR.

haqadn (approved):

projects/plugins/boost/app/assets/src/js/stores/minify.ts
dilirity: Yeah, I thought so too, but I wasn't sure where to put it as the whole structure is still a mystery.

PR #33663, opened 2023-10-18
Social: Add Nextdoor connection toggle

pablinos:
This makes the connections API call return Nextdoor connections and ensures that the icon has the correct colour.

## Proposed changes:

This adds Nextdoor to the 'known' connections when retrieving them via the API calls. It also adds the default colour to the icon for when it's displayed in the editor.

### Other information:

No

## Testing instructions:

* Create a Nextdoor connection on your site using D125526-code
* Load the editor and check that you can see the Nextdoor logo representing the connection

image


manzoorwanijk (commented):
For a brief moment, I see the connection active and then it is greyed out with the notice that the connection is broken.

Screenshot 2023-10-19 at 11 43 59 AM


Debugging shows the `brokenConnections` array to be this

Screenshot 2023-10-19 at 11 46 32 AM


I can confirm that the connection in Calypso is OK.

gmjuhasz:
> For a brief moment, I see the connection active and then it is greyed out with the notice that the connection is broken.

The editor tries the connection test endpoint after loading up, that's why it's healthy in the beginning and becomes broken afterwards when the request returns. As we cannot test the connection at the moment, we might have to hardcode to return true for nextdoor in the API too

pablinos:
> As we cannot test the connection at the moment, we might have to hardcode to return true for nextdoor in the API too

We do at the moment, don't we? As we have the `test_connection` method returning `true`. Was your dev site sandboxing the API @manzoorwanijk ? If the Nextdoor connection wasn't coming back in the API response, then it would be broken.

manzoorwanijk (approved):
Works well.

> Was your dev site sandboxing the API @manzoorwanijk ?

No, that was the problem. Thanks

PR #33665, opened 2023-10-18
tools: Fix edge case in changelogger-release.sh

anomiex:
## Proposed changes:

We had a situation when trying to release videopress, which depends on both sync and identity-crisis, where sync had no changes while identity-crisis did. The script released identity-crisis, bumping it from 0.10 to 0.11, but did not release sync, resulting in a problem because the existing sync package's `^0.10` dep could not be satisfied by the new 0.11 version.

The fix is conceptually simple enough: when a package has a major-ish version bump, any dependents also need to have a release.

This means that we need to change the script to process every package before any of its dependents (same order as when building) so we can flag those dependents before we consider them for release. Which makes it into a somewhat large refactoring, which includes running some safety checks before the real processing instead of during that processing.

### Other information:

## Jetpack product discussion

p1697638594374359/1697564420.947809-slack-CBG1CP4EN

No

## Testing instructions:

* Check out bf8a1ff4aacb245d68bd68533312ee65c1599a52, cherry-pick this, then run `tools/changelogger-release.sh -p -b -r 1.6-beta plugins/videopress`. `packages/sync` should be one of the packages being released.


ice9js (approved):
👍

PR #33667, opened 2023-10-18
Refactor production blocks registration 4

monsieur-z:

## Proposed changes:

Registers the following production blocks by specifying the path to their block.json file:  _Paywall_, _Pinterest_, _Podcast Player_, _Rating Star_, _Recurring Payments_.

  - Remove `icon.js`
  - Fix JS tests

### Other information:

## Jetpack product discussion

pedMtX-RS-p2

No.

## Testing instructions:

- Spin up a test site
- Test that the aforementioned blocks behave as in `trunk`. This PR shouldn't change anything from a user's standpoint. You'll need to activate subscriptions to test the _Paywall_ block.

Note: I couldn't test the _Recurring Payments_ block.


projects/plugins/jetpack/extensions/blocks/pinterest/pinterest.php
```diff
@@ -12,9 +12,7 @@
 use Automattic\Jetpack\Blocks;
 use WP_Error;
 
-const FEATURE_NAME = 'pinterest';
-const BLOCK_NAME   = 'jetpack/' . FEATURE_NAME;
-const URL_PATTERN  = '#^https?://(?:www\.)?(?:[a-z]{2}\.)?pinterest\.[a-z.]+/pin/(?P<pin_id>[^/]+)/?#i'; // Taken from AMP plugin, originally from Jetpack.
+const URL_PATTERN = '#^https?://(?:www\.)?(?:[a-z]{2}\.)?pinterest\.[a-z.]+/pin/(?P<pin_id>[^/]+)/?#i'; // Taken from AMP plugin, originally from Jetpack.
```
andrii-lysenko, line 15: Is it the same as `URL_REGEX` in `constants` ? Could it be reused?

andrii-lysenko: Oh, or it's the one below

projects/plugins/jetpack/extensions/blocks/payment-buttons/index.js
jeherve, line 1: Would this be a rebase issue with #33604? The Payments button is modified there.

projects/plugins/jetpack/extensions/blocks/podcast-player/edit.js
```diff
@@ -69,7 +69,7 @@ const PodcastPlayerEdit = ( {
 	replaceWithEmbedBlock,
 } ) => {
 	// Validated attributes.
-	const validatedAttributes = getValidatedAttributes( attributesValidation, attributes );
+	const validatedAttributes = getValidatedAttributes( metadata.attributes, attributes );
```
jeherve, line 72: This causes the block to break. You can try to add a podcast URL to the block and you will get the following errors:

image

projects/plugins/jetpack/extensions/blocks/rating-star/rating-star.php
jeherve, line 1: I believe this may be causing a notice:

```
Warning: Undefined array key "ratingStyle" in /wp-content/plugins/jetpack-dev/extensions/blocks/rating-star/rating-meta.php on line 123
```

projects/plugins/jetpack/extensions/blocks/recurring-payments/block.json
```diff
@@ -65,5 +65,5 @@
 			"type": "string"
 		}
 	},
-	"editorScript": "file:../editor.js"
+	"parent": [ "jetpack/payment-buttons" ]
```
jeherve, line 68: This block is currently broken for me:

image

andrii-lysenko: I also noticed this in console:
Screenshot at Oct 24 10-20-45

jeherve (changes requested):
Looks like the tests in `extensions/blocks/rating-star/test/validate.js` will need to be updated?

jeherve (approved):

PR #33670, opened 2023-10-18
AI Extension: Enable Form extension inside query loops

### LINKED ISSUE #32554: AI Extension: multiple buttons on the toolbar when using the block on a query loop block
When using the full site editor, if I place the form block on a location that is inside a query loop block, the assistant toolbar button will be repeated multiple times.

How to reproduce:

- test on a site with some published posts
- go to the full site editor
- add a Jetpack Form block on a location inside the frontpage query loop; this is a `twenty twenty-three` example location:

Screenshot 2023-08-17 at 12 32 19

- confirm the toolbar button is repeated multiple times:

Screenshot 2023-08-17 at 12 32 53

dhasilva:
Fixes #32554

## Proposed changes:

* Enables the Form extension again

It seems that the recent changes made by @retrofox have fixed the issue as a side-effect, so I'm only removing the query loop check and fixing some typescript issues in the same file.

### Other information:

## Testing instructions:

* Go to the site editor
* Add a Form inside a query loop
* Check that there is only one AI icon
* Check that the extension works with no regression both on the site editor and in a post

| Before | After
| -|-
|   |


lhkowalski (approved):
Code looks good.

Tested it before and after your change. Working as described.

_Before:_

Screenshot 2023-10-18 at 19 22 17

_After:_

Screenshot 2023-10-18 at 19 26 01

Issue #33359, opened 2023-09-27
AI Tiered Plans: Add request counter on sidebar


lhkowalski:
The tiered plans will be measured so we need to show the usage on the sidebar. Sidebar was chosen as a quick place to add the usage info.

To do this, we need to:
 
- Implement the component to show this information (see design files)
- Place the componet on the sidebar
- Expose monthly usage data as an API (or include it on the `
wpcom/v2/sites/SITE/jetpack-ai/ai-assistant-feature` endpoint)
- Connect the component with the real usage data (usage reset data TBD, between 1st of the month and purchase date)

---

Check design wrap up post for details: p1HpG7-oR3-p2

---

This is required for the MVP.

---

This task depends on:

- [x] Create the counter component based on the design
- [x] Provide the current period usage on some API endpoint https://github.com/Automattic/jetpack/issues/33566

---

First phase:

- [x] implement the usage indicator for the current plan structure

Second phase:

- [ ] implement the usage indicator for the multiple plans on the new tiered structure: https://github.com/Automattic/jetpack/issues/33887

lhkowalski:
The first phase of the work. We are now showing the usage on the current plan structure:

- an usage bar and a label with "X / Y free requests" on the free users
- a label with "X / ∞ requests" and an help text with "Unlimited requests for your site" on the paid users

lhkowalski:
I will close this since the second phase will be handled in the context of https://github.com/Automattic/jetpack/issues/33887.

PR #33671, opened 2023-10-18
AI Assistant: Add usage panel scaffolding

lhkowalski:
Part of #33359.

## Proposed changes:

* Add a placeholder panel to hold usage information in next iterations
* Divide the AI Assistant sidebar panel using sections, put the content feedback tool under "AI feedback on post" and the usage panel under "Usage" (accoding to Figma styles from p1HpG7-oMv-p2)

### Other information:

## Testing instructions:

* With beta extensions enabled, go to the block editor, open the Jetpack sidebar and look for the AI Assistant section; it should contain the new sections division and a placeholder text for the Usage Panel:

Screenshot 2023-10-18 at 18 33 50

* With the beta extensions disabled, check the sidebar again and confirm the "Usage" section is not present, but the "AI feedback on post" is still there and working as expected:

Screenshot 2023-10-18 at 18 35 03


projects/plugins/jetpack/extensions/plugins/ai-assistant-plugin/components/ai-assistant-plugin-sidebar/index.tsx
```diff
@@ -69,13 +75,28 @@ export default function AiAssistantPluginSidebar() {
...
+						</BaseControl>
 					</PanelRow>
 					{ requireUpgrade && (
 						<PanelRow>
 							<Upgrade onClick={ autosaveAndRedirect } type={ upgradeType } />
 						</PanelRow>
 					) }
+					{ isUsagePanelAvailable && (
+						<PanelRow>
+							<BaseControl
```
CGastrell, line 92: Are we sure we don't want the BaseControl inside UsagePanel? Seems to me the block could handle the entire component tree by itself.

Of course, this is no blocker and likely on your plans for later iterations :)

CGastrell (approved):
Left a comment for your consideration but nothing serious. :shipit:

projects/plugins/jetpack/extensions/plugins/ai-assistant-plugin/components/ai-assistant-plugin-sidebar/index.tsx
lhkowalski: I was going to add it inside the UsagePanel, but then I also needed to add a BaseControl around the Proofread component and was not thinking about changing it directly, so I added the BaseControl as a wrapper on both.

Thinking more about it, I think it's OK to add them on the component itself. I am starting to work on a follow-up now, so I am going to move the BaseControl inside the UsagePanel and the Proofread component.

Thank you for the comment 👍

PR #33674, opened 2023-10-19
WIP Media Inserter: Fix checking Google Photos connection when not needed

rcrdortiz:
Fixes p1697662881647649-slack-C04DZ8M0GHW

We have an issue with an excessive number of calls being made to the Google Photos endpoint which needs to be solved in order to bring back the number of calls made.

## Proposed changes:

* Changed the order in which we check the Google Photos connection so that we only check the connection when the media inserter is open.

No

### Other information:

## Testing instructions:

* Load this branch on an Atomic site.
* Edit or Create a new Post
* Open Chrome's Network tab or equivalent in your browser of choice.
* Filter by `media/list/google` 
* You should not see any requests until you open the media sidebar.

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-media-sidebar-checks-google-photos-connection-before-needed
+++ b/projects/plugins/jetpack/changelog/fix-media-sidebar-checks-google-photos-connection-before-needed
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
+The Google Photos media inserter only checks for the connection status when needed.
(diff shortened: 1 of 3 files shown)


BogdanUngureanu (approved):
Works as expected. The request is made only when the user opens the sidebar media. 

However, I also noticed that the http response code is 500. Is that expected?

Copons (approved):

rcrdortiz:
> Works as expected. The request is made only when the user opens the sidebar media.
> 
> However, I also noticed that the http response code is 500. Is that expected?

Yes, it's expected. To determine if the user is connected or not it tries calling the API (which fails if not connected)

PR #33677, opened 2023-10-19
Welcome message: Add support for the welcome email setting in wp-admin

### LINKED ISSUE #83081: Welcome email: Jetpack changes - wp-admin
Jetpack adds the textfields to wp-admin. Add a new field for the welcome message.

`modules/subscriptions.php`

TimBroddin:
Fixes https://github.com/Automattic/wp-calypso/issues/83081

## Proposed changes:
- Adds the field for the welcome message inside `wp-admin`
- Sets the default value for the default value we're using right now when sending the welcome email
- 
Reading Settings _ HelloWord _ WordPress

### Other information:

## Testing instructions:
1. Apply this PR to your JT site
2. Go to Settings > Reading and verify that the field is there and works.

If you need to reset the fields back to their default setting you can use this command:

```
docker exec -it jetpack_dev-wordpress-1 wp option delete subscription_options  --allow-root
```

### DIFF
--- a/projects/plugins/jetpack/modules/subscriptions.php
+++ b/projects/plugins/jetpack/modules/subscriptions.php
@@ -430,6 +430,14 @@ class Jetpack_Subscriptions {
 			'reading',
 			'email_settings'
 		);
+
+		add_settings_field(
+			'welcome',
+			__( 'Welcome email text', 'jetpack' ),
+			array( $this, 'setting_welcome' ),
+			'reading',
+			'email_settings'
+		);
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


projects/plugins/jetpack/modules/subscriptions.php
```diff
@@ -620,17 +628,44 @@ public function setting_comment_follow() {
...
 	/**
 	 * Get default settings for the Subscriptions module.
 	 */
 	public function get_default_settings() {
 		$site_url    = get_home_url();
 		$display_url = preg_replace( '(^https?://)', '', untrailingslashit( $site_url ) );
 
+		// TODO: Remove the locale check once the translation is ready.
+		$is_english_locale = str_starts_with( get_locale(), 'en' );
+		if ( $is_english_locale ) {
```
jeherve: I don't quite understand this. Since both strings are new and untranslated right now, what's the benefit in having this condition?

TimBroddin: I was wrongfully assuming that it would use the translation already present on WPCOM when deployed there. I'll fix this.

jeherve (approved):

PR #33679, opened 2023-10-19
Jetpack Boost: Extract ISA Styles

pyronaur:
## Proposed changes:
Extract CSS styles from components into dedicated files - this is a part of the React migration, we'll have to change our approach to styles either way, this is a step towards that.

### Other information:

## Jetpack product discussion
N/A

N/A

## Testing instructions:
🌈


dilirity (dismissed):
:shipit:

dilirity (approved):
After manually checking the components and their corresponding styles, I squashed some bugs that resulted after moving stuff around.

I think this is ready to ship.

@pyronaur I remember you mentioned something about moving stylesheets around to match how react is going to use them? I think for now we can leave this as is and slowly move them over to react as we go.

Since I'm not yet familiar with how react scopes stylesheets when importing them - would we need the jb prefix once we start using stylesheets inside components? Doesn't it do magic like svelte did to make styles unique to the component that uses it? Or is that just another way of loading stylesheets globally via the components themselves 😅

Issue #33655, opened 2023-10-18
Update deprecated inline style functions


jeherve:
Dev note: https://make.wordpress.org/core/2023/10/17/replacing-hard-coded-style-tags-with-wp_add_inline_style/

********

> the following functions have been deprecated and replaced with the new approach:
>
> `print_embed_styles()`
> `print_emoji_styles()`
> `wp_admin_bar_header()`
> `_admin_bar_bump_cb()`
> `the_block_template_skip_link()`

We do use `print_emoji_styles()` in the mu wpcom package:
https://github.com/Automattic/jetpack/blob/cbc1de32f2eec7a94b3ce3304f0d1f5a8f84b730/projects/packages/jetpack-mu-wpcom/src/features/coming-soon/coming-soon.php#L131

However, no change is necessary for now:

> In 6.4 this `print_emoji_styles()` function is now deprecated. Nevertheless, the above method for preventing emoji styles from being printed is retained, even though they are now being printed by `wp_enqueue_emoji_styles()`. This applies to the other deprecated functions as well, so no developer action is required.

*********

It seems we'll need to make changes in Boost and in Jetpack CRM though:

> For developers who are currently using the wp_print_styles() function, whether it’s in unit tests or within their own code, some adjustments may be necessary to ensure a smooth transition. You should follow the example set by Gutenberg. 

**In Boost:**

https://github.com/Automattic/jetpack/blob/3a3c4adc1c0913bac224ddb832713bd6101fe1ab/projects/plugins/boost/app/lib/critical-css/Admin_Bar_Compatibilty.php#L52

**In Jetpack CRM:**

https://github.com/Automattic/jetpack/blob/0f0602135373d2546974c47581db075601c4e2e1/projects/plugins/crm/admin/activation/before-you-go.php#L48
https://github.com/Automattic/jetpack/blob/ba835b99cc2ae09f2b8ce5676c0fa760fe99c46f/projects/plugins/crm/admin/activation/welcome-to-jpcrm.php#L68
https://github.com/Automattic/jetpack/blob/413617e798c5e783bfd44af61af907181d97c306/projects/plugins/crm/modules/woo-sync/admin/activation/welcome-to-woo-sync.php#L47
https://github.com/Automattic/jetpack/blob/9f999ab39a277e3dc140d26551de336fe114c306/projects/plugins/crm/public/forms/form-content.php#L36
[...]

PR #33680, opened 2023-10-19
Boost: Fix depreciation notice `wp_print_styles` WP 6.4

haqadn:
**⚠️ I am still trying to figure out if we really need this.**

Core seems to have found a weird way to deprecate `print_emoji_styles`. Even though `print_emoji_styles` itself isn't depricated, using it without this hack may still produce deprication notice. 

Related to  #33655

See WordPress/gutenberg#54806 and WordPress/gutenberg#54828

## Proposed changes:
* Temporarily remove action call to `print_emoji_styles` while using `wp_print_styles`

### Other information:

## Jetpack product discussion
N/A

No

## Testing instructions:
* Install WP 6.4 beta
* Generate critical CSS and ensure no deprication notice is present related to `print_emoji_styles`.


haqadn:
Not actually needed

PR #33682, opened 2023-10-19
Refactor production blocks registration 5

monsieur-z:

## Proposed changes:

Registers the following production blocks by specifying the path to their block.json file:  _Related Posts_, _Repeat Visitor_, _Slideshow_, _Story_.

  - Remove `icon.js`
  - Fix JS tests

### Other information:

## Jetpack product discussion

pedMtX-RS-p2

No.

## Testing instructions:

- Spin up a test site


projects/plugins/jetpack/tools/check-block-assets.php
```diff
@@ -43,6 +43,7 @@
 	'wp-html-entities',
 	'wp-i18n',
 	'wp-is-shallow-equal',
+	'wp-plugins',
```
jeherve: Did you add this in for the Story block? If we need to add this in here, maybe we could allow it just for the Story block, with a change to `$allowed`?

jeherve (commented):
This tests well. I only have a small question.

jeherve (commented):
I'll let someone else review that one (@andrii-lysenko maybe), because I'm running into an issue with the Pay with PayPal block, and I'm not sure if it's something to do with my setup. I am experiencing the same problem when on `trunk`, and only on self-hosted (a new JN site and an existing WoA site). It works well on wpcom simple.

The block seems to work well in the editor, but I do not see it on the frontend. And when refreshing the block editor, the data in the block disappears.

monsieur-z:
UPDATE: I moved all payment-related changes to a separate PR. This PR now effectively changes the following blocks: Related Posts, Repeat Visitor, Slideshow, Story.

jeherve (approved):

PR #33686, opened 2023-10-19
Only add WooCommerce tracking to core order actions

senadir:
This PR limits add the tracking initiator query var from being added to non woo-core actions, so it's only limited to view, pay, and cancel.

### Other information:

No

## Testing instructions:
Testing this PR might require some code to be used, or a longer route of installing a plugin.

Using code:
- In a live, Jetpack website, add this code:
```php
add_filter('woocommerce_my_account_my_orders_actions', function( $actions ) {
	$actions['test'] = [
			'url'  => $_REQUEST['_wp_http_referer'],
			'name' => __( 'Test', 'woocommerce' ),
	];
	return $actions;
});
```
- You will need WooCommerce installed, logged in, add a product to your Cart and Checkout.
- Got to the /my-account/orders page, you should see a couple of buttons, View, and Test, on each order.
- Make sure the view button link has a `?_wca_initiator=action` on it, and that the test button has nothing but the current page url.

Using a plugin
- Install [PDF invoices plugin.](https://wordpress.org/plugins/woocommerce-pdf-invoices-packing-slips/)
- You will need WooCommerce installed, add a product to your Cart and Checkout.
- Go to wp-admin > WooCommerce > Orders > The order you just created.
- Hit "PDF Invoice" on the Create PDF box.
- Got to the /my-account/orders page, you should see a couple of buttons, View, and Invoice, on each order.
- The View button link should have `?_wca_initiator=action` on it, the invoice link should work.

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-limit-tracking-to-core-buttons
+++ b/projects/plugins/jetpack/changelog/fix-limit-tracking-to-core-buttons
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
+Fixed a bug that prevent customers from downloading invoices from the my account page in WooCommerce
--- a/projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-my-account.php
+++ b/projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-my-account.php
@@ -232,6 +232,12 @@ class Jetpack_WooCommerce_Analytics_My_Account {
 			if ( ! isset( $action['url'] ) ) {
 				continue;
 			}
+
+			// Check if the action key is view, pay, or cancel.
+			if ( ! in_array( $key, array( 'view', 'pay', 'cancel' ), true ) ) {
+				continue;
+			}
+
 			$url                    = add_query_arg( array( '_wca_initiator' => 'action' ), $action['url'] );
 			$actions[ $key ]['url'] = $url;
 		}


anomiex (approved):
I had trouble reproducing the bug without this PR in order to verify that this PR actually fixed it. It turned out that that PDF Invoices plugin released a v3.7.0 yesterday to work around the bug from their end, I had to install v3.6.3 to reproduce it.

OTOH, the real cause of the bug that prompted this isn't really anything in here. It's that that PDF Invoices plugin [runs the URL through `esc_url()` early](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/be25d785958ed8fe7a73da5b0c400883608387a3/includes/class-wcpdf-endpoint.php#L142) ([twice!](https://github.com/wpovernight/woocommerce-pdf-invoices-packing-slips/blob/be25d785958ed8fe7a73da5b0c400883608387a3/includes/class-wcpdf-frontend.php#L65)), so when the hook here does `add_query_arg` it sees something like `/wp-admin/admin-ajax.php?action=generate_wpo_wcpdf&#038;document_type=invoice&#038;order_ids=16` and decides that everything after the `#` is a fragment. Fixing those to not escape early results in a working Invoice link.

senadir:
I think it's fine merging this regardless of the plugin status?

anomiex:
> I think it's fine merging this regardless of the plugin status?

Yes, sorry if I was unclear. My intention was to approve the PR while also noting the background in case that made a difference to you.

ice9js:
This change will be included in `12.7.1` (a49addbff6a220f3347e647c2eb3a2fc1699f185).

PR #33693, opened 2023-10-19
AI Extension: show "no content" notice when the extended block content is empty

retrofox:
This PR avoids requesting an AI suggestion from the AI extension when the block doesn't have content, showing a notice instead.

Fixes #

## Proposed changes:

* AI Extension: show "no content" notice when the extended block content is empty

### Other information:

## Testing instructions:

* Go to the block editor
* Add some spaces to a paragraph block instance
* Click on the AI Extension button of the block toolbar
* Click on one of the dropdown menu items
* Confirm that, since the block doesn't have content, the app shows a notice instead of performing a request

image

* Confirm that when the block has content, it works as expected

### DIFF
--- a/projects/plugins/jetpack/changelog/update-ai-extension-show-empty-info
+++ b/projects/plugins/jetpack/changelog/update-ai-extension-show-empty-info
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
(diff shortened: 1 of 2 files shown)


CGastrell (approved):
Works fine. I still think we should revisit this with a more elegant solution, but this should suffice to get back in track with the current situation

PR #33410, opened 2023-09-29
MyJetpack: add has_required_plan for VideoPress product class

CGastrell:
This PR adds the method override for VideoPress product class to perform the necessary purchase checks for the site

Fixes #33222 

## Proposed changes:
The issue described a circular flow where the purchase would go into the wrong upgrade path, but this all was coming from the wrong flag on the product details as `has_required_plan` was returning the parent's class default `true`

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Does this pull request change what data or activity we track or use?
No

## Testing instructions:
Start with a site with no VP subscription. Upload a video to exhaust the free quota.

Go to My Jetpack. See the VideoPress card now reads "Purchase" and shows as inactive. Click "Purchase" and follow the flow, purchase with credits. 

You should be taken back to My Jetpack screen to now show the VideoPress card active and with a "Manage" button as CTA.

PR #33697, opened 2023-10-19
Hotfix/revert 33410

CGastrell:
Revert #33410 as it can cause fatals

## Proposed changes:
Using Current_Plan falls into a use case where Jetpack is required to be installed (as it provides the lib for Current_Plan), causing fatals otherwise.

This _revert_ makes `has_required_plan` return `true`, as it is the default parent's class behavior (as it was before the change).

### Other information:

## Jetpack product discussion
p1697748366739339-slack-C02TQF5VAJD

No

## Testing instructions:

Without this patch:
- Install a standalone Jetpack plugin like Jetpack VaultPress Backups
- Make sure you have uninstalled and removed the main Jetpack plugin
- Navigate to My Jetpack
- You will get a fatal error

Apply patch and navigate to My Jetpack, fatal should be gone

### DIFF
--- a/projects/packages/my-jetpack/changelog/hotfix-revert-33410
+++ b/projects/packages/my-jetpack/changelog/hotfix-revert-33410
@@ -1,0 +1,4 @@
+Significance: patch
+Type: changed
+
+Make has_required_plan return true (as it was before #33410) as a way to revert the change.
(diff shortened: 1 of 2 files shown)


jboland88 (approved):
code LGTM, tested and this fixes the fatal error on My Jetpack

PR #33701, opened 2023-10-20
VideoPress: register block type with block metadata

### LINKED ISSUE #32235: VideoPress Block: CSS file is loaded on all pages, even when the block isn't used

CGastrell:
Use `register_block_type` to handle assets loading properly

Fixes #32235

## Proposed changes:
This PR removes old asset enqueueing code and uses the return value from `register_block_type` to get the script handle, which is then provided to `Connection_Initial_State::render_script` for proper initial state rendering.

The PR doesn't change any functional aspect, but tries to solve the mentioned issue after a failed attempt on #33042 (which ended being reverted).

### Other information:

No

### Test on .com:

Use the downloader script to apply on your sandbox. Point to your sandbox:
- [ ] a P2 site (might need to upgrade to P2+ to enable video uploading, ping me for a test-ready site)
- [ ] a simple site with a classic theme
- [ ] a simple site with a block theme

### Test on Jetpack and WoA sites
Use local or rsync, then setup:
- [ ] with classic theme
- [ ] with block theme
- [ ] with infinity scroll if available (search themes that support it)

### On all cases:

- [ ] publish multiple posts with and without videos
- [ ] use Settings -> Reading to list only 2 blog posts (so posts with videos are not on first render/load)
- [ ] see that posts/lists where no videos are published `view.css` is NOT loaded on sources under
  - [ ] wp-content/mu-plugins/jetpack-plugin/**/jetpack_vendor/automattic/jetpack-videopress/build/block-editor/blocks/video
  - [ ] wp-content/plugins/videopress/jetpack_vendor/automattic/jetpack-videopress/build/block-editor/blocks/video
  - [ ] wp-content/plugins/jetpack/jetpack_vendor/automattic/jetpack-videopress/build/block-editor/blocks/video
- [ ] see that scrolling into or loading a post with a video will show `view.css` loaded from one of those paths
- [ ] see that videos, both on lists and single posts, are loaded properly and can be played

### DIFF
--- a/projects/packages/videopress/changelog/fix-videopress-register-block-json
+++ b/projects/packages/videopress/changelog/fix-videopress-register-block-json
@@ -1,0 +1,4 @@
+Significance: minor
+Type: changed
+
+Use register_block_type and block.json to handle the block's scripts properly and fix assets being loaded when block is not present
--- a/projects/packages/videopress/composer.json
+++ b/projects/packages/videopress/composer.json
@@ -64,7 +64,7 @@
 			"link-template": "https://github.com/Automattic/jetpack-videopress/compare/v${old}...v${new}"
 		},
 		"branch-alias": {
-			"dev-trunk": "0.18.x-dev"
+			"dev-trunk": "0.19.x-dev"
 		},
 		"version-constants": {
 			"::PACKAGE_VERSION": "src/class-package-version.php"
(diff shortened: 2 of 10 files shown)


dhasilva (approved):
LGTM. `view.css` is not loaded.

PR #33695, opened 2023-10-19
Paywall: Cross platforms login link

lezama:
Alternative to #33627 and #33656

## Proposed changes:
* Similar to #33627 but uses an existing endpoint.
* Display the "log in" link in Jetpack sites
* See more details in D125480-code
* Test alongside with D126041-code to get the magic link flow instead of the user name login flow.

Screenshot 2023-10-19 at 17 41 55

Asks the user to log in 

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* See D125480-code

PR #33705, opened 2023-10-20
[newsletters] retrieve subscriptions for self hosted

millerf:
Reverts Automattic/jetpack#33703

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -1074,11 +1074,27 @@ function get_paywall_blocks( $newsletter_access_level ) {
 		: __( 'Already a subscriber?', 'jetpack' );
 
 	$sign_in = '';
-	if ( ! is_user_logged_in() && ( new Host() )->is_wpcom_simple() ) {
+	$host    = new Host();
+	if ( ! is_user_logged_in() && $host->is_wpcom_simple() ) {
+		$id           = '';
 		$sign_in_link = wpcom_logmein_redirect_url( get_current_url(), false, null, 'link' );
+		$button_text  = esc_html__( 'Log in', 'jetpack' );
+	} elseif ( class_exists( 'Automattic\Jetpack\Extensions\Premium_Content\Subscription_Service\Jetpack_Token_Subscription_Service' ) &&
+		! $host->is_wpcom_simple() ) {
+		// We are on Jetpack and no cookie is set
+		$id              = 'jp_retrieve_subscriptions_link';
+		$sign_in_link    = '#'; // listening to "click" event in view.js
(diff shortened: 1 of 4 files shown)


projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -1074,11 +1074,27 @@ function get_paywall_blocks( $newsletter_access_level ) {
...
+		$sign_in_link    = '#'; // listening to "click" event in view.js
+		$access_question = __( 'I am already subscribed.', 'jetpack' );
+		if ( ! Jetpack_Token_Subscription_Service::has_token_from_cookie() ) {
+			$button_text = esc_html__( 'Log in', 'jetpack' );
+		} else {
+			$button_text = esc_html__( 'Switch accounts', 'jetpack' );
+		}
+	}
 
+	if ( ! empty( $sign_in_link ) ) {
```
lezama, line 1095: let's use the approach in https://github.com/Automattic/jetpack/pull/33695 as a fallback, it should result in a better UX on all platforms.

`$sign_in_link = 'https://subscribe.wordpress.com/memberships/jwt?site_id=' . \Jetpack_Options::get_option( 'id' ) . '&redirect_url=' . get_current_url();`

millerf: What do you mean "as a fallback"?

lezama: when `! is_user_logged_in()` and `empty( $sign_in_link )`, or that would never happen?

lezama (approved):

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
millerf: Outside of WPCOM, `is_user_logged_in()` doesn't have any meaning, so 
```
! is_user_logged_in() && && $host->is_wpcom_simple()
``` 
is already covered and $sign_in_link is never empty,....

lezama: > is already covered and $sign_in_link is never empty,....

why do we check for `if ( ! empty( $sign_in_link ) ) {` then?

millerf: Because if you are logged in AND on WPCOM `$sign_in_link` is empty...

lezama: I think we should display `Switch accounts` in that case!

millerf: > I think we should display `Switch accounts` in that case!

So linking to logout? Yes it could be an idea...

PR #3369, opened 2016-02-02
Custom Post Types: Testimonials: add class if testimonial has a featured image

thomasguillot:

PR #33711, opened 2023-10-20
CRM: Resolves #3369 - allow WooSync contact to be assigned to existing company

tbradsha:
Resolves Automattic/zero-bs-crm#3369 - allow WooSync contact to be assigned to existing company

## Proposed changes:

If a Woo order was created with a company name that already existed in the CRM, the CRM contact would not be associated with the company. This is because the companies were created with the Woo customer info (e.g. email + address). A new customer using the same company name would have a different email and thus would not be able to find the existing company.

Given that this happens behind the scenes (via hooks and `cron`), there isn't a way to have the user decide on the fly.

If the company existed, we could handle this in three ways:
1. Continue to fail, as the email won't match and company name is a unique field. This is what we currently do, and it's not very intuitive, though the failure does show in WooSync logs.
2. Search for a company by name and use that ID. In edge cases, two customers might list the same company name for two distinct companies.
3. Add a setting to let the user decide between the two behaviours. This still requires option 2 to be implemented and adds a fair amount of complexity for a situation that hasn't previously been reported.

I opted for option 2 at this stage. With this PR I introduce the following changes:

* Companies can be searched by name via DAL (this actually existed, but there was a missing condition preventing this).
* Woo orders that have a company names field that matches an existing CRM company name will associate the CRM contact with the existing CRM company.
* Woo orders that have a company name that doesn't exist in the CRM will create a new CRM company without contact-specific fields.

### Other information:

## Testing instructions:

1. Add a new Woo order with the following:
    - surname: `鈴木`
    - company: `東京産業`
    - email: `suzuki@dev.null`
    - address: 123 Shady Ln
2. Verify CRM contact and CRM company are created and associated.

In `trunk`, the new company is created with the contact email address (as well as any contact address fields if provided).

In the `fix/crm/3369-assign_customer_to_existing_company` branch, the new company will only have the name field. All other fields will be blank.

3. Add a new Woo order with the following:
    - surname: `田中`
    - company: `東京産業`
    - email: `tanaka@dev.null`

In `trunk`, the CRM contact will not be associated with the existing CRM company.

In the `fix/crm/3369-assign_customer_to_existing_company` branch, the CRM contact is properly associated with the existing company.

### DIFF
--- a/projects/plugins/crm/changelog/fix-crm-3369-assign_customer_to_existing_company
+++ b/projects/plugins/crm/changelog/fix-crm-3369-assign_customer_to_existing_company
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+WooSync: contacts can now be assigned to existing companies
--- a/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Companies.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Companies.php
@@ -415,6 +415,8 @@ class zbsDAL_companies extends zbsDAL_ObjectLayer {
             (!empty($id) && $id > 0)
             ||
             (!empty($email))
+					||
+					( ! empty( $name ) )
             ||
             (!empty($externalSource) && !empty($externalSourceUID))
             ){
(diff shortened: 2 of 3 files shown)


coder-karen (approved):
This works well for me. I noticed I was having various issues when testing whilst not in incognito - even with the branch the user associated with the email address I was logged in with was being associated with the company even if none of the fields matched in the form, and then their details would change as a result. That - I assume - would make sense (that the logged in user is the user being used), though the changing of details wasn't expected. Not related to this PR anyway.

In incognito in trunk I could test the conditions as written - and in the branch the customer is linked as expected (and not in trunk).

PR #33221, opened 2023-09-20
CRM: Resolves #3330 - use "task" term instead of "event"

tbradsha:
CRM: Resolves Automattic/zero-bs-crm#3330 - use task instead of event

## Proposed changes:

This PR updates many references to use `task` instead of `event` wherever possible (just shy of 700 references).

**Note:** please ignore the one failing PHPCS check titled `Linting / PHP Code Sniffer (changes to excluded files only) (pull_request)`. We will skip this prior to merge; instructions to do so can be found in p1695240852475789-slack-CDLH4C1UZ.

Things that did NOT change:

* DAL (with the exception of constant updates) and DAL helper functions
* Database
* API endpoint names (i.e. `create_event` and `events`)
* Permission capabilities (e.g. `admin_zerobs_view_events`)
* Hard-coded low-level references to object type as 'event' or `zerobs_event` (e.g. in the URL 😢)
* Things that rely on dynamically-accessed globals (e.g. `$zeroBSCRM_columns_event`)
* Things that rely on dynamically-accessed JS vars (e.g. `zeroBSCRMJS_listView_event_bulkActionTitle_markcomplete`)
* Functions and event names in the old internal automator system

Of note, `ZBS_TYPE_TASK` now exists and has replaced `ZBS_TYPE_EVENT` throughout the codebase. The old constant still remains for backward-compatibility.

One can search for references with the following regex:

`\b(?!(pr|admin_zerobs_|addUpdate))\w*event(?!(s->|.prevent|Listener|ually))`

And exclude the following:

`js/lib/*, sass/semantic-ui/*, changelog/*, */LICENSE, licenses/*, *.txt, *.md, *.lock, *.dist, *.pot`

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

Make sure everything task-related runs as before. Tests should pass. The new Automation stuff should be properly renamed.

PR #33712, opened 2023-10-20
API - Fix Creating Event Endpoint

monagjr:
passed data has **"from"** and **"to"** keys not **"start"** and **"end"**, and hence without this update the Create Event API Endpoint keeps returning Error 100

## Testing instructions:

* POST /create_event
* Example POST data: 
```
'{"title":"My Event From the API","notify": 0,"customer":1,"to":"10/23/2023 00:00:00","from":"10/18/2023 00:00:00","complete":0}
```
* without proposed change,  it keeps returning  {"error":100}
* with proposed change, it creates the event as expected and returns expected data:
```
{"title":"My Event From the API","customer":1,"notes":"","to":"10\/23\/2023 00:00:00","from":"10\/18\/2023 00:00:00","notify":-1,"complete":-1,"owner":-1,"id":1}
```

NO.

### DIFF
--- a/projects/plugins/crm/includes/ZeroBSCRM.IntegrationFuncs.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.IntegrationFuncs.php
@@ -885,8 +885,8 @@ function zeroBS_integrations_addOrUpdateTask(
 		is_array( $data_array )
 		&& count( $data_array ) > 0
 		&& ! empty( $data_array['title'] )
-		&& ! empty( $data_array['start'] )
-		&& ! empty( $data_array['end'] )
+		&& ! empty( $data_array['from'] )
+		&& ! empty( $data_array['to'] )
 	) {
 			return zeroBS_addUpdateEvent( $task_id, $data_array, $task_reminders );
 	} else { // no source/id/fields
--- a/projects/plugins/crm/includes/ZeroBSCRM.IntegrationFuncs.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.IntegrationFuncs.php
@@ -885,8 +885,13 @@ function zeroBS_integrations_addOrUpdateTask(
 		is_array( $data_array )
 		&& count( $data_array ) > 0
 		&& ! empty( $data_array['title'] )
-		&& ! empty( $data_array['from'] )
-		&& ! empty( $data_array['to'] )
+		&& (
+			// old params
+			( ! empty( $data_array['to'] ) && ! empty( $data_array['from'] ) )
+			||
+			// new params
+			( ! empty( $data_array['start'] ) && ! empty( $data_array['end'] ) )
+		)
 	) {
 			return zeroBS_addUpdateEvent( $task_id, $data_array, $task_reminders );
 	} else { // no source/id/fields
--- a/projects/plugins/crm/changelog/patch-1
+++ b/projects/plugins/crm/changelog/patch-1
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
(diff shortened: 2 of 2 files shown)


tbradsha:
Introduced in #33221 when cleaning up legacy code and naming.

projects/plugins/crm/includes/ZeroBSCRM.IntegrationFuncs.php
```diff
@@ -885,8 +885,8 @@ function zeroBS_integrations_addOrUpdateTask(
 		is_array( $data_array )
 		&& count( $data_array ) > 0
 		&& ! empty( $data_array['title'] )
-		&& ! empty( $data_array['start'] )
-		&& ! empty( $data_array['end'] )
+		&& ! empty( $data_array['from'] )
+		&& ! empty( $data_array['to'] )
```
tbradsha: Great find! Due to other integrations, we'll need to support both versions. How does this look?

```suggestion
		&& (
			// old params
			( ! empty( $data_array['to'] ) && ! empty( $data_array['from'] ) )
			||
			// new params
			( ! empty( $data_array['start'] ) && ! empty( $data_array['end'] ) )
		)
```

monagjr: I don't think there is a need to check for both old and new params in this code (unless we do more code changes)

Reason: I can see the function enclosing this code (zeroBS_integrations_addOrUpdateTask) is only called once in the whole codebase from this line: https://github.com/Automattic/jetpack/blob/8832e428eb65643273e2e1c41116d6681ebdae64/projects/plugins/crm/api/create_event.php#L93

and the input ($task_fields) array is being populated only by "from" and "to" keys
https://github.com/Automattic/jetpack/blob/8832e428eb65643273e2e1c41116d6681ebdae64/projects/plugins/crm/api/create_event.php#L43
https://github.com/Automattic/jetpack/blob/8832e428eb65643273e2e1c41116d6681ebdae64/projects/plugins/crm/api/create_event.php#L47

tbradsha: Yep, the official API docs use `to` and `from`, but `start` and `end` are the "new" way to do it, matching the database. Down the road when we update the API, those will be the correct params.

At the moment, the only reason the `to` and `from` work is because of a legacy DAL translation function.

Given that we've supported the new way for quite some time, and that third-party integrations may already use this function, removing it could break existing code. Technically the function is not documented, so we can do what we want, but we generally try to not make breaking changes.

monagjr: Thanks for the extra details.

And I did not mean to remove the code / function 🙂  , I was just refering to the only usage for it.

Based on the where things are heading and the roadmap, I think we go with supporting both "old" and "new" params. 

I tested the shared code above https://github.com/Automattic/jetpack/pull/33712#discussion_r1367507504 and it seems to be working as expected 👍

tbradsha: > And I did not mean to remove the code / function

No problem! Really we're the ones that broke it (me specifically), so I'm grateful you spotted it and took the initiative to get it working!

tbradsha (approved):
Looks great. Thanks again!

monagjr:
when will it be released in the public version of the plugin?
Nov. 7th? or before?

tbradsha:
We don't currently have a fixed release schedule for Jetpack CRM, but it will be in the next version.

PR #33714, opened 2023-10-20
AI Assistant: Connect usage data to UsagePanel

lhkowalski:
Part of #33359.

## Proposed changes:

* Connect the usage bar on the usage panel to real usage data, so we display the current usage for the site
* Hide the `Requests will reset in X days.` message since the current plans do not reset; I kept the message there because it will be necessary for the tiered plans
* For paid plans, since there is not real limit, I am showing `X / ∞ requests` and fixing the usage percentage to 10%  so the usage bar shows some information

### Other information:

## Testing instructions:

* Test on a site with beta extensions enabled
* Go to the block editor and open the Jetpack sidebar; look for the "AI Assistant" section:

Screenshot 2023-10-19 at 18 05 48

---

* On the "AI Assistant" section, confirm you see the "Usage" panel with a message stating your usage + the usage bar:

Screenshot 2023-10-20 at 16 04 15

---

* To confirm the number is correct, open the WordPress.com Console (https://developer.wordpress.com/docs/api/console/) and do a request to `wpcom/v2` `/sites/YOUR-SITE/jetpack-ai/ai-assistant-feature`:

Screenshot 2023-10-20 at 16 09 32

---

* Test it on a site with the Jetpack AI plan and confirm you see the usage as a following (X / ∞):

Screenshot 2023-10-20 at 16 10 33

---

* Also test on a site without the plan and confirm you see the usage in another way (X / Y):

Screenshot 2023-10-20 at 16 10 55

---

* Notice that the `Requests will reset in X days.` is not visible on both cases


projects/plugins/jetpack/extensions/plugins/ai-assistant-plugin/components/usage-panel/index.tsx
```diff
@@ -6,26 +6,47 @@ import { __, sprintf } from '@wordpress/i18n';
...
-			<UsageBar usage={ 0.5 } />
+			<UsageBar usage={ usage } />
 
-			<p className="muted">
-				{
-					// translators: %1$d: number of days until the next usage count reset
-					sprintf( __( 'Requests will reset in %1$d days.', 'jetpack' ), 10 )
-				}
-			</p>
+			{ false && (
```
retrofox, line 42: A comment about this line here in the code would be nice

retrofox:
> For paid plans, since there is not real limit, I am showing X / ∞ requests and fixing the usage percentage to 10% so the usage bar shows some information



cc @ederrengifo @madebynoam in case you have some feedback here.

For instance, I don’t know, remove the infinite part and showing simply `n requests`?

retrofox (approved):
LGTM. I left a few comments that I don't consider blockers. Nice one1

ederrengifo:
The infinite (302 / infinite) seems good since communicates that users should be worried about any limit, but I think we shouldn't show the progress bar because it could send the wrong message. The rule here would be that only to show the progress bar when there is a limit in the current tier. cc @madebynoam

retrofox:
> The infinite (302 / infinite) seems good since communicates that users should be worried about any limit, but I think we shouldn't show the progress bar because it could send the wrong message. The rule here would be that only to show the progress bar when there is a limit in the current tier. cc @madebynoam

Could we consider doing something like this?

image

The requests bar looks nice imo.

PR #33717, opened 2023-10-20
AI Assistant: Enhance toolbar UX always showing disabled actions

### LINKED ISSUE #31917: AI Assistant: 🎨 Revisit never hiding the AI Assistant block menu items, instead just disabling them
We explored improvements for the block at two levels. First, we tried to identify the functions for each part of the block, this helps us to understand what users should expect from each part and open the opportunity to define where is expected to see changes over time:

Image

Then, we identify that when a block is created we can add consistency across instances. More specifically:
- When a new block is created, and there is no content, we show the fixed prompts disabled
- When content has been suggested but not accepted, we keep the "Write with AI..." menu available, but we populate it with prompts that work in this context.
- When a new block is created, and there is previous content, we show the same fixed prompts including the "improve" one, currently hidden.

Image

dhasilva:
Fixes #31917

## Proposed changes:

* Shows disabled actions instead of removing them
* Adds contextual suggestions for follow-up requests
* Adds a tsconfig file to fix typescript issues
* Moves some template to avoid this empty group:

### Other information:

## Testing instructions:

* Go to the block editor and add an AI Assistant block
* Check that the actions that need content are displayed, but disabled
* Ask for some content
* Check that the toolbar actions are enabled and there are contextual suggestions under "Write with AI..."
* Accept the content and add another AI Assistant block
* Check that the actions that require post content are enabled, with the exception of "Improve" that is meant for "chat" mode
* Check that there is no regression in the AI Assistant block and the extensions

https://github.com/Automattic/jetpack/assets/8486249/f2f6218e-0922-4e15-bff0-e2055d6168a7

### DIFF
--- a/projects/plugins/jetpack/changelog/update-ai-assistant-toolbar-ux
+++ b/projects/plugins/jetpack/changelog/update-ai-assistant-toolbar-ux
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
+AI Assistant: Enhance toolbar UX
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/components/i18n-dropdown-control/index.tsx
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/components/i18n-dropdown-control/index.tsx
@@ -1,7 +1,14 @@
 /*
  * External dependencies
  */
-import { MenuItem, MenuGroup, ToolbarDropdownMenu, DropdownMenu } from '@wordpress/components';
+import {
+	MenuItem,
+	MenuGroup,
+	ToolbarDropdownMenu,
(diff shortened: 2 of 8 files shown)


CGastrell (approved):
LGTM, and works well.

That said, I can't see where `contentIsLoaded` is coming from.

Also, any specifics to _regressions_? I've used the assistant and haven't seen any drawbacks so far, but maybe you have something in mind?

dhasilva:
`contentIsLoaded` is passed down as a prop that is `true` when the block's `content` attribute has something. Not the best name.

> Also, any specifics to regressions?

There is the case of asking for some new information, and cases for transforming existing content by using all toolbar options.

PR #33718, opened 2023-10-20
CRM: Resolves #3371 - remove buggy unused code

### LINKED ISSUE #3371: Comments: form uses wrong comment ID when user is logged in locally, logged out from wpcom
Steps to reproduce:
1. Under Settings > Discussion, check that box:
   > Users must be registered and logged in to comment
2. Log out of your site.
3. Log back in during a different account.
4. Log out of WordPress.com.
5. Open a post, and leave a comment.
6. The comment will be left under another post, using post ID `14`. That post ID is the same on 2 different blogs where I followed the steps above.

Here is how the comment form looks like in the inspector:
- When logged in to WordPress.com (looking good):

good
- When logged out:

Reported here:
https://wordpress.org/support/topic/comments-appear-under-wrong-post-jetpack2016

tbradsha:
Resolves Automattic/zero-bs-crm#3371 - remove buggy unused code

## Proposed changes:

If a user has some manually-translated statuses (e.g. via Loco Translate), it can cause issues. The root cause is that we have some legacy listview filter code that isn't used but that is also not properly escaped. This PR removes said code (the `zbsFilterButtons` JS var) and some additional unused code surrounding it.

### Other information:

## Testing instructions:

This is a bit hard to reproduce. In `trunk`, the easiest way is to edit `includes/ZeroBSCRM.List.php` and modify these lines:

https://github.com/Automattic/jetpack/blob/5cd649f44de4fe9bdcf7dc89d1e1d088d352e062/projects/plugins/crm/includes/ZeroBSCRM.List.php#L498-L500

To add the following:

```
                    <?php  $buttonCount = 0;

                        $currentFilterButtons['bad_filter'] = array( "'קצת אוהד 'לקוח'" );

                        #} Current cols
```

Then go to the contact listview and you'll see the JS error due to improper output escaping. Given this JS isn't even used, the `fix/crm/3371-remove_buggy_unused_code` branch simply removes it altogether.

### DIFF
--- a/projects/plugins/crm/changelog/fix-crm-3371-remove_buggy_unused_code
+++ b/projects/plugins/crm/changelog/fix-crm-3371-remove_buggy_unused_code
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Listview: remove legacy code
--- a/projects/plugins/crm/includes/ZeroBSCRM.List.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.List.php
@@ -493,15 +493,6 @@ class zeroBSCRM_list{
             // Vars for zbs list view drawer
             var zbsListViewParams = <?php echo wp_json_encode( $list_view_parameters ) ?>;
 
-            var zbsUnsortables = [<?php $c = 0; if (count($this->unsortables) > 0) foreach ($this->unsortables as $sortableStr) {
-
-                        if ($c > 0) echo ',';
-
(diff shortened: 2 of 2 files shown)


coder-karen (approved):
Yes that is a hard one to reproduce, I can reproduce the issue with the testing instructions you gave. I can also see none of the code appears to have been used anywhere too, so this looks good to me 👍

Issue #33360, opened 2023-09-27
AI Tiered Plans: Add upgrade button on the sidebar


lhkowalski:
The MVP design shows an upgrade button on the sidebar, under the usage indicator.

The button is always visible, even before the user reaches the limit.

We need to include the button and make it point to the right target. For Jetpack sites, the target is the upgrade interstitial. For Simple, the target is the checkout flow.

---

This is required for the MVP.

---

First phase:

- [x] implement the upgrade button for the current plan structure

Second phase:

- ~~implement the upgrade button for the multiple plans on the new tiered structure~~
   - I will mark this change as not necessary for closing the task right now since we have #33365 that will set the correct checkout URLs on the upgrade buttons

lhkowalski:
The first phase of this is done. Now we have an upgrade button available that will allow free users to upgrade to the paid plan. Now we need to follow-up on this to provide upgrade call-to-actions for the new plan structure.

lhkowalski:
The second phase of this will be completed in the scope of #33365.

PR #33720, opened 2023-10-20
AI Assistant: Add upgrade button to usage panel

lhkowalski:
Part of #33360.

## Proposed changes:

* Add an "Upgrade" button on the usage panel when the site does not have a plan
* Point the button to the checkout URL
* Only display the button when the user can purchase upgrades

The logic to show this button is different from the upgrade prompt that we are using on the Assistant itself: it will not wait for the user to reach the limit; it will always be visible while the user does not buy the upgrade.

The button style is the standard WordPress style, as per i6 designs: p1HpG7-oMv-p2.

Note: this is based on the current state of trunk, before merging #33714, so the usage data is not real.

### Other information:

## Testing instructions:

* Go to the block editor and open the Jetpack sidebar; look for the AI Assistant section and open it:

Screenshot 2023-10-19 at 18 05 48

---

* If the site already have a plan, you should not see the button
* If the site does not have a plan and the user is an admin or other user with `manage` powers, you should see the "Upgrade" button:

Screenshot 2023-10-20 at 19 07 22

---

* If the site does not have a plan and the user does not have `manage` powers, you should not see the button (test with an author or editor, for example)
* When the button is present, if you click it, it should lead to the right page of the upgrade flow:
   * When the site is a Jetpack site, it should redirect to the Jetpack AI interstitial
   * When the site is a Simple site, it should redirect to the checkout page

### DIFF
--- a/projects/plugins/jetpack/changelog/update-ai-assistant-add-upgrade-button-to-usage-panel
+++ b/projects/plugins/jetpack/changelog/update-ai-assistant-add-upgrade-button-to-usage-panel
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
+AI Assistant: add Upgrade button component on the UsagePanel.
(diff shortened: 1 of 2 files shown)


CGastrell (dismissed):
LGTM, works as described :shipit:

retrofox:
image

I suggest using the Upgrade prompt that we use in other sections.

CGastrell (approved):
Works great. We might want to make sure this is what we want after removing a subscription:
image

But we can work on it in a follow up

PR #33721, opened 2023-10-20
Publicize: fix tweet storm performance issue

ellatrix:
Fixes #

## Proposed changes:

* `TweetDivider` and its selectors through `withSelect` are currently being called for all blocks in a post on every keystroke, even if that post is not a tweet storm. We should avoid rendering the component entirely if a post is not a tweet storm.

This could be having a significant performance impact. All the green squares you see in the red box below are `withSelect` call form `TweetDivider`

Screenshot 2023-10-21 at 01 10 53

### Other information:

## Testing instructions:

* Go to '..'
*


jeherve (commented):
Noting that we stopped using that component last week, so this is effectively impossible to test, unless you were to revert #33641 in your test environment when testing.

We should be able to close this PR in favor of #33723 since we do not need any Twitter functionality anymore.

PR #33723, opened 2023-10-21
Social: Remove the editor tweetstorm components

pablinos:
## Proposed changes:

@ellatrix let us know that the TweetDivider component was causing performance issues in the browser, and addressed them in #33721.

In discussion about it, we decided to remove the components entirely as the feature can't be used while we don't have access to the twitter API.

This PR removes the components, ~we should follow up with removing the tweetstorm functionality from the `publicize` data store.~ Turns out that a lot of stuff was coupled together, so this also removes code from the store, as well as the the `gather-tweetstorm` editor extension, which could be another performance problem.

### Other information:

## Jetpack product discussion

p1697840484666559-slack-C02JJ910CNL

No

## Testing instructions:
* Ensure auto-sharing is enabled on a Jetpack site or use a WordPress.com site
* Load the editor
* Publish a test post
* Check the post is shared correctly
* Check for no related console errors in the browser


ellatrix (dismissed):
Awesome! Let's remember this is still in the git history and deleting it for now prevents copying the code to other features.

projects/js-packages/publicize-components/src/components/social-previews/twitter.js
```diff
@@ -14,14 +14,10 @@ function Twitter( { tweets } ) {
...
-	const { getTweetTemplate, getTweetStorm, getShareMessage, isTweetStorm } =
-		select( 'jetpack/publicize' );
+	const { getTweetTemplate, getShareMessage } = select( 'jetpack/publicize' );
 
-	let tweets = [];
-	if ( isTweetStorm() ) {
-		tweets = getTweetStorm();
-	} else {
-		tweets.push( {
+	const tweets = [
```
ellatrix: It's not good to return a new array on every select. This will cause the component to re-render whenever the store updates.

ellatrix (approved):
Let's fix the selector in a follow up

gmjuhasz (approved):
Works as described.

jeherve (approved):

PR #33724, opened 2023-10-21
Performance: VideoPress: avoid useEffect call for every block on typing

ellatrix:
Fixes #

## Proposed changes:

* We should return early before `useEffect` is called. Currently this effect is running for every single block on the page on every single keystroke.

### Other information:

### Test `<JetpackCoreVideoDeprecation />` block

The app filters all blocks and will extend only the `core/video` videopress  instances.

* Go to the block editor
* Create a few general block instances, like paragraphs
* Ensure the VideoPress video block is not supported for your site
  * Deactive the Jetpack module: `https://your-testing-site/wp-admin/admin.php?page=jetpack#/settings?term=video`
image
  * Deactivate the VideoPress standalone plugin 
 
image

* Open the React Dev tool. Search by `ConvertV6toV5Effect` string
* Confirm there isn't any `JetpackCoreVideoDeprecation` instance in the components three

* Create a `videopress/video` block instance
  * Switch the editor mode
  * Paste the following chunk of code:
 
```
<!-- wp:videopress/video { "title":"boats","description":"","caption":"","id":36,"guid":"z4N1oaIj","cacheHtml":"\u003ciframe title=\u0022VideoPress Video Player\u0022 aria-label='VideoPress Video Player' width='600' height='338' src='https://video.wordpress.com/embed/z4N1oaIj?cover=1\u0026amp;preloadContent=metadata\u0026amp;useAverageColor=1\u0026amp;hd=1' frameborder='0' allowfullscreen data-resize-to-parent=\u0022true\u0022 allow='clipboard-write'\u003e\u003c/iframe\u003e\u003cscript src='https://v0.wordpress.com/js/next/videopress-iframe.js?m=1674852142'\u003e\u003c/script\u003e","videoRatio":56.333333333333336,"privacySetting":2,"allowDownload":false,"rating":"G","isPrivate":false,"duration":14960} /-->
```

  * Return to visual mode
  * Confirm the editor identifies the new block like a `core/video` block instance
  * Confirm the block sidebar shows the transform info section

Screenshot 2023-10-21 at 12 39 04


retrofox (commented):
### convertV6toV5

I'm not able to reproduce a case to run this code.
In short, the VideoPress video block is always registered.

If you disable the Jetpack module and the VideoPress standalone plugin, the block is still available. But it shows the Connect banner:

image

The only way to unregister the VideoPress video block is to deactivate the Jetpack and the Standalone plugin.

CGastrell:
back and forth conversions work from/to v5/v6


https://github.com/Automattic/jetpack/assets/157240/cfad6f31-25b8-4541-ac02-94c1d9807f20

retrofox (approved):
* The `handleJetpackCoreVideoDeprecation` never runs—obsolete code. Let's remove it in a follow-up PR.
* Something similar happens with `convertV6toV5`.
The `if ( name !== 'core/missing' || attributes?.originalName !== 'videopress/video' )` condition never happens because the block is always registered even when the Jetpack module is deactivated. We should probably clean this code, too.

ice9js:
This change will be included in `12.7.1` (1a1423ab3aab1209107ec5b4897759fc07d0957b).

PR #33725, opened 2023-10-21
Performance: don't call useAnalytics for all paragraphs

ellatrix:
Fixes #

## Proposed changes:

* Currently `useAnalytics` is being called for every single block (including paragraph blocks) on every single keystroke. We must remove `useAnalytics` when it is not needed. `useAnalytics` is adding a bunch of select and effect hooks that are running needlessly.

### Other information:

## Testing instructions:

* Go to '..'
*


projects/plugins/jetpack/extensions/extended-blocks/paid-blocks/with-upgrade-banner.jsx
```diff
@@ -35,10 +35,10 @@ export default createHigherOrderComponent(
 			select => select( 'core/block-editor' ).hasSelectedInnerBlock( clientId, true ),
 			[]
 		);
-
+		const { hasParentBanner } = useContext( PaidBlockContext ) || {};
```
retrofox, line 38: Is it okay to call the hook conditionally?

ellatrix: It's not :) but you're already calling hook in this component conditionally anyway. Should be fine for now since blocks can't change type.

retrofox (approved):
LGTM. We've mentioned that calling hooks conditionally is not okay, which deserves a follow-up since it isn't introduced by this PR.

ice9js:
This change will be included in `12.7.1` (acbf1b5a1b9322f053fdd27b4fc0248578557d4b).

PR #33730, opened 2023-10-23
REST API settings endpoint: Fix google analytics option handling for Jetpack sites

fgiannar:
## Proposed changes:

* `WPCOM_JSON_API_Site_Settings_Endpoint`: Rely on whether this is a Jetpack site instead of the environment (WPCOM vs remote site) in order to determine whether the `wga` or `jetpack_wga` option should be used for fetching/storing Google Analytics tracking data.

### Other information:

## Jetpack product discussion

pf5801-5l-p2

No

## Testing instructions:
- Using the [Developer console](https://developer.wordpress.com/docs/api/console/) check the `sites/[SITE_ID]/settings` endpoint for both fetching and updating the `wga` setting
- Make sure to test on a Simple site too
- Bonus points for testing on a Jetpack site, forcing the endpoint to fetch the results from WPCOM (`?force=wpcom`)

### DIFF
--- a/projects/plugins/jetpack/json-endpoints/class.wpcom-json-api-site-settings-endpoint.php
+++ b/projects/plugins/jetpack/json-endpoints/class.wpcom-json-api-site-settings-endpoint.php
@@ -572,8 +572,20 @@ class WPCOM_JSON_API_Site_Settings_Endpoint extends WPCOM_JSON_API_Endpoint {
 	 * Get GA tracking code.
 	 */
 	protected function get_google_analytics() {
-		$option_name = defined( 'IS_WPCOM' ) && IS_WPCOM ? 'wga' : 'jetpack_wga';
+		$option_name = $this->get_google_analytics_option_name();
+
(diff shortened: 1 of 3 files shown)


projects/plugins/jetpack/json-endpoints/class.wpcom-json-api-site-settings-endpoint.php
```diff
@@ -715,6 +726,7 @@ public function update_settings() {
 					/** This action is documented in modules/widgets/social-media-icons.php */
 					do_action( 'jetpack_bump_stats_extras', 'google-analytics', $enabled_or_disabled );
 
+					$is_wpcom = defined( 'IS_WPCOM' ) && IS_WPCOM;
```
fgiannar, line 729: Not 100% sure if this only makes sense for Simple sites or we should modify the logic here too. For now, leaving as is

jeherve (approved):
This seems to test well for me. Thanks for adding a test!

PR #33733, opened 2023-10-23
JSPackages: make ContextualUpgradeTrigger icon click work

CGastrell:
ContextualUpgradeTrigger holds an arrow icon, but the icon doesn't have an `onClick` and blocks the click behavior of the component

https://github.com/Automattic/jetpack/assets/157240/829ee7cc-fb22-401a-8063-d2395e5224e4

## Proposed changes:
Set the icon to `pointer-events: none` so the click behavior falls back to the underlying component

### Other information:

No

## Testing instructions:
The component is used on some Jetpack products, like VideoPress, Search, Scan, Protect, Firewall.
The easiest way to see it is installing VideoPress plugin and upload a video. Then navigate to the dashboard `/wp-admin/admin.php?page=jetpack-videopress#/` and the upgrade banner (ContextualUpgradeTrigger) should show:
image

- [ ] Hover the arrow to see it behave as the rest of the component (since it doesn't receive the events)
- [ ] See that hovering still triggers the animation
- [ ] See that clicking on the arrow triggers the same action as the rest of the component

### DIFF
--- a/projects/js-packages/components/changelog/change-contextual-upgrade-arrow-behavior
+++ b/projects/js-packages/components/changelog/change-contextual-upgrade-arrow-behavior
@@ -1,0 +1,4 @@
+Significance: patch
+Type: changed
+
+Use pointer-events: none on arrow icon so its click behavior falls back to the container/underlying component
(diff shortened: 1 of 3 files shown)


retrofox (approved):
TIL a little bit more about [CSS pointer-events](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events#none) 


https://github.com/Automattic/jetpack/assets/77539/bcea07be-888f-4616-bea2-2847a6be8514


LGTM! 🚀

PR #33736, opened 2023-10-23
CRM: 3375 - Clean up usage of get_userdata wrapper functions

### LINKED ISSUE #3375: Activation request returns invalid HTTP
When activating one's jetpack installation the request to xmlrpc.php returns a response without `Content-Length` and without `Transfer-Encoding`. This causes varnish (tested with varnish version 4) to dismiss the response body.
According to RFC 2616 one of the two headers are mandatory.

Steps on how to reproduce:
1. Install wordpress with jetpack
2. Use netcat/telnet to simulate activation request.
3. Note that both `Content-Length` and `Transfer-Encoding` are missing.

Example request (removed activation code and replaced host header):

```
POST /xmlrpc.php?for=jetpack HTTP/1.0
Host: example.com
User-Agent: Jetpack by WordPress.com
Accept: */*
Content-Length: 198
Content-Type: application/x-www-form-urlencoded

<?xml version="1.0"?>
<methodCall>
<methodName>jetpack.verifyRegistration</methodName>
<params>
<param><value><string>******</string></value></param>
```

Here is an example response ( I removed the body for readability):
```
HTTP/1.1 200 OK
Date: Fri, 05 Feb 2016 12:48:37 GMT
Server: Apache
X-Powered-By: PHP/5.6.16
Connection: close
X-Clacks-Overhead: GNU Terry Pratchett
Vary: Accept-Encoding
Content-Type: text/xml; charset=UTF-8

<?xml version="1.0" encoding="UTF-8"?>
[...]
```

Please let me know if you need any more information or want me to test anything.

tbradsha:
Resolves Automattic/zero-bs-crm#3375 - clean up usage of get_userdata wrapper functions

## Proposed changes:

We have four functions that perform similar tasks, and for better readability this PR cleans them up.

* `zeroBS_getCustomerOwner`: a wrapper for `zeroBS_getOwner` that is not actually used, so removed
* `zeroBS_getWPUserSimple`: only called from `zeroBS_getOwnerObj`, so moved into parent function
* `zeroBS_getOwnerObj`: simplified logic and limited what `WP_User->data` fields are returned
* `zeroBS_getOwner`: no changes here, but in some cases it acts as a wrapper for `zeroBS_getOwnerObj`

I also did the following:
* removed some legacy pre-DAL3 code I ran across
* updated the `events` endpoint to use DAL calls

Note that paid extensions will be unaffected by the changes in this PR.

Additional context: p3btAN-2vJ-p2

### Other information:

## Testing instructions:

No functionality should have changed after merging this PR, but it should be easier to follow what the remaining functions do.


projects/plugins/crm/includes/ZeroBSCRM.DAL3.Helpers.php
```diff
@@ -1183,20 +1176,33 @@ function zeroBS_getOwner($objID=-1,$withDeets=true,$objType=-1,$knownOwnerID=-1)
 	return false;
 }
 
-function zeroBS_getOwnerObj($wpUserID=-1,$simpleUserData=true){
+// phpcs:ignore Squiz.Commenting.FunctionComment.Missing
```
gogdzl:
```suggestion
/**
 * Retrieves the owner object based on a given WP user ID.
 * 
 * This function gets the owner's data without revealing sensitive information
 * (e.g. `user_pass`).
 * 
 * @param int $wp_user_id The WordPress user ID. Default is -1.
 * 
 * @return array|bool Returns an associative array containing the 'ID' and 'OBJ' (user data object) if successful, false otherwise.
 */
```

gogdzl (dismissed):
It works as advertised, and it comes with a nice cleanup package!

I left a small suggestion to get rid of one phpcs:ignore, but not a blocker. Marking this as approved.

gogdzl (dismissed):

gogdzl (approved):

PR #3373, opened 2016-02-04
Open Graph: Correctly pull core's site icon

kraftbj:
In https://core.trac.wordpress.org/changeset/33605 (pre-4.3), the args for `get_site_icon_url` were rearranged to avoid the frequency of needing to pass null args. We did not account for this before shipping.

This PR:
* Rearranges the `get_site_icon_url()` call to use the correct args.
* Removes the `jetpack_site_icon` fallback as our minimum version is 4.3 so all sites will have access to Core's functionality.

Issue #3376, opened 2016-02-06
Add Social Tracking, after the initial post


captbrando:
The Social plugin from Mailchimp does (or used to do) this where once a post was pushed to social media, any replies or retweets/reshares were captured as comments inside of the post history. Love how I can capture discussions on my post in one place!

jeherve:
I'm afraid that's not something that's on our radar right now. If you'd like to add that feature to your site, I'd recommend a plugin like this one:
https://wordpress.org/plugins/gplus-comments/

You can use it alongside Jetpack's Comments module, and it will allow you to capture content that's posted to Facebook and Google+. 

The [Social plugin](https://wordpress.org/plugins/social/) does indeed do something similar, but was retired.

Thanks for the feedback!

PR #33740, opened 2023-10-23
CRM: Resolves #3373 and #3376 - address PHP 8.2 deprecation notices in the Client Portal

tbradsha:
Resolves Automattic/zero-bs-crm#3373 and Automattic/zero-bs-crm#3376 - address PHP 8.2 deprecation notices in the Client Portal

## Proposed changes:

Dynamic class properties are not allowed by default in the Client Portal. Since all endpoints use the `param_value` property, I added that as part of the `Client_Portal_Endpoint` class declaration.

After testing Client Portal Pro, I realised there are some other dynamic properties the extension selectively adds. There were three choices:

1. Declare those dynamic properties in the main class. I didn't like this option, because that introduces code only used by the extension.
2. Have an alternate class the endpoints use when the extension is enabled. I didn't like this option, as it relies on load order, adds extra complexity, and would likely duplicate code.
3. Allow dynamic properties with `#[\AllowDynamicProperties]`.

I went with the latter choice. That means the first commit wasn't needed in the end, but I still kept it so as to uphold best practices when possible.

### Other information:

## Testing instructions:

Go to each of the subpages in the Client Portal.

In `trunk`, most of the pages (i.e. `/details`, `/invoices`, `/invoices/1`, `/quotes`, `/quote/1`, `/transactions`) will generate a PHP deprecation notice that looks like this:
```
PHP Deprecated:  Creation of dynamic property Automattic\JetpackCRM\Some_Endpoint::$param_value is deprecated in /modules/portal/class-client-portal.php on line 503
```

In the `fix/crm/3373-deprecation_notices_in_Client_Portal` branch, the deprecation notice is no more.

### DIFF
--- a/projects/plugins/crm/modules/portal/class-client-portal-endpoint.php
+++ b/projects/plugins/crm/modules/portal/class-client-portal-endpoint.php
@@ -15,6 +15,7 @@ defined( 'ZEROBSCRM_PATH' ) || exit;
  * This class represents a single Client Portal endpoint (e.g. Quotes)
  * 
  */
+#[\AllowDynamicProperties]
 abstract class Client_Portal_Endpoint {
 	public $portal                       = null;
 	public $template_name                = null;
--- a/projects/plugins/crm/modules/portal/class-client-portal-endpoint.php
+++ b/projects/plugins/crm/modules/portal/class-client-portal-endpoint.php
@@ -29,6 +29,11 @@ abstract class Client_Portal_Endpoint {
 	public $template_args                = array();
 	public $template_path                = '';
 	public $default_template_path        = '';
+	/**
+	 * Option param value used by some endpoints
+	 * @var string
+	 */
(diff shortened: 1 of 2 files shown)


projects/plugins/crm/modules/portal/class-client-portal-endpoint.php
```diff
@@ -15,6 +15,7 @@
  * This class represents a single Client Portal endpoint (e.g. Quotes)
  * 
  */
+#[\AllowDynamicProperties]
```
coder-karen, line 18: Very nice!

coder-karen (approved):
This looks good and works well for me. 

I found even more deprecation notices while testing though, but even though they're in the Client Portal, they're unrelated. Added them to a new issue: https://github.com/Automattic/zero-bs-crm/issues/3377

PR #33748, opened 2023-10-23
Jetpack Sync: Add/sync woo remove order items

mavegaf:
Add the `woocommerce_remove_order_items` event to Jetpack Sync, added on WooCommerce on https://github.com/woocommerce/woocommerce/pull/37822 . This PR also needs changes on D126373-code to whitelist the event on WPCOM (already merged)

## Proposed changes:
* Add action listener `woocommerce_remove_order_items`
* Get the order items id and send them as a group using action `woocommerce_remove_order_item_ids`

### Other information:

## Testing instructions:
Test plan for Automatticians
* Create a test site with Payments configured. Probably JN is the simpler option (you will need to enable global-http)
** Include WooCommerce
** Include WooCommerce Payments
** Include WooCommerce Payments Dev Tools
** Include WooCommerce Payments JN Options
** Jetpack Beta with this branch (add/sync-woo-remove-order-items)
** Add a Backup plan
* ~~Apply D126373-code and configure your site to use your sandbox.~~ (already merged)
* As an admin, add some products to your website
* As a customer, add the products to your shopping cart and go to the checkout 
* Use a declined credit card to pay https://woocommerce.com/document/woopayments/testing-and-troubleshooting/testing/#declined-test-cards
* Verify you received the events `woocommerce_remove_order_item` with the order_item_id. You could do it by adding a log with the events before they are sent to backups (fbhepr%2Skers%2Sjcpbz%2Sjc%2Qpbagrag%2Szh%2Qcyhtvaf%2Swrgcnpx%2Sflap%2Spynff.wrgcnpx%2Qflap%2Qerjvaq.cuc%3Se%3Q29612p98%23230-og)
```
error_log( '-->send_events_to_backup:' . json_encode( $this->events ) );
```


fgiannar (commented):
@mavegaf Apologies for the late review here and many thanks for working on this!

I was wondering mostly from a performance POV, why do we need to "hack" our way here by listening for the `woocommerce_remove_order_items` action and converting it to multiple `woocommerce_remove_order_item` for each order item?
Would it be possible to send the `woocommerce_remove_order_items` action instead to WPCOM with all the item IDs and do the corresponding processing for each item there instead?

mavegaf:
> @mavegaf Apologies for the late review here and many thanks for working on this!
> 
> I was wondering mostly from a performance POV, why do we need to "hack" our way here by listening for the `woocommerce_remove_order_items` action and converting it to multiple `woocommerce_remove_order_item` for each order item? Would it be possible to send the `woocommerce_remove_order_items` action instead to WPCOM with all the item IDs and do the corresponding processing for each item there instead?

Hi. Thanks for the review.

Initially, I thought a solution by sending `woocommerce_remove_order_items` with an array of `order_item_ids` using a filter https://github.com/Automattic/jetpack/pull/33748/commits/a4893d7f4c929fc81b408da60243ec5452554214 (it needs a fix with the type)
However, when I integrated it, I realized that the current integration for events and backups was built expecting to receive the id of each event. Given that, I considered different alternatives:

1. The current one, transforming `woocommerce_remove_order_items` on `woocommerce_remove_order_item `
2. Pass the array of order_item_ids and handle the special case on WPCOM or VaultPress for the type `woocommerce_remove_order_items` 
3.  Add on WooCommerce monorepo a hook to send every [order_item_id after they were deleted. ](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/abstracts/abstract-wc-order.php#L824C23-L824C35)

I would be ok with the second solution to get a better performance, even if that means handling a special case. The current implementation has not been touched in a while, so a refactor shouldn't be necessary unless more cases like this appear.

However, I also consider the third option. One concern I have about the [woocommerce_remove_order_items](https://github.com/woocommerce/woocommerce/blob/f51995a470450679466a70c8e7308db19a0eb76f/plugins/woocommerce/includes/abstracts/abstract-wc-order.php#L822) action is at that point, we have not deleted any order_item yet; it is triggered before the deletion so it is just the intention, and in case of any unexpected problem, we could receive the event, but the order_items might not have been deleted.
There is another action triggered after the order_items were deleted, [woocommerce_removed_order_items](https://github.com/woocommerce/woocommerce/blob/f51995a470450679466a70c8e7308db19a0eb76f/plugins/woocommerce/includes/abstracts/abstract-wc-order.php#L843), but at that point, since the order_items were already removed, I don't have a way to recover the order_item_id which are the values I need.

So I think the third could be a good option considering we already handle other order_item actions individually like `added_order_item_meta`, `updated_order_item_meta`, `deleted_order_item_meta`, `woocommerce_new_order_item`, `woocommerce_update_order_item`, and this case will not have a higher number of order_items than the current that should be now in an order, and also it is something that only happens if the order creation failed (like in the payment failure). 

What solution do you consider more convenient? I can proceed with the second (send an array of order_item_ids and implement a special case), or explore the third option to add another action in the woocommerce plugin.

Thanks in advance for your comments


.

fgiannar:
Many thanks for sharing these approaches @mavegaf !

> One concern I have about the [woocommerce_remove_order_items](https://github.com/woocommerce/woocommerce/blob/f51995a470450679466a70c8e7308db19a0eb76f/plugins/woocommerce/includes/abstracts/abstract-wc-order.php#L822) action is at that point, we have not deleted any order_item yet; it is triggered before the deletion so it is just the intention, and in case of any unexpected problem, we could receive the event, but the order_items might not have been deleted.

I guess this concern applies to the current approach too, right? 

IMO the second approach would be the most performant since it would only trigger and send a single action to WPCOM including all the item IDs.

projects/packages/sync/src/modules/class-woocommerce.php
```diff
@@ -128,6 +148,8 @@ public function init_listeners( $callable ) {
 		add_action( 'woocommerce_new_order_item', $callable, 10, 4 );
 		add_action( 'woocommerce_update_order_item', $callable, 10, 4 );
 		add_action( 'woocommerce_delete_order_item', $callable, 10, 1 );
+		add_action( 'woocommerce_remove_order_items', array( $this, 'action_woocommerce_remove_order_items' ), 10, 2 );
```
zinigor: Should we move that `add_action` into the constructor with other filters?

zinigor (dismissed):
Overall the changes look good, but one of the `add_action` calls looks like it will better fit into another place.

mavegaf:
> Many thanks for sharing these approaches @mavegaf !
> 
> > One concern I have about the [woocommerce_remove_order_items](https://github.com/woocommerce/woocommerce/blob/f51995a470450679466a70c8e7308db19a0eb76f/plugins/woocommerce/includes/abstracts/abstract-wc-order.php#L822) action is at that point, we have not deleted any order_item yet; it is triggered before the deletion so it is just the intention, and in case of any unexpected problem, we could receive the event, but the order_items might not have been deleted.
> 
> I guess this concern applies to the current approach too, right?
> 
> IMO the second approach would be the most performant since it would only trigger and send a single action to WPCOM including all the item IDs.

Thanks for the review. I'll work next week on the changes in backend and will update this again to use an array of order_item ids.

Regarding the concern about the items that could have been not deleted yet, I think that is always possible with `woocommerce_remove_order_items` (It is probably unlikely the action fails to remove the order item ids and at the same time the order re-creation by a payment failure doesn't). 

I'll continue with the solution using this action (which is better than not receiving that information) , and maybe later, we could evaluate adding the removed order item ids to the action `woocommerce_removed_order_items` in the WooCoomerce project)



> Overall the changes look good, but one of the `add_action` calls looks like it will better fit into another place.

Thanks for the comment. I'll change how to get the order_item_ids with a filter in the constructor

mavegaf:
The changes are ready, but I'll move the PR to draft until the review of other changes and tests get priority

Initsogar (dismissed):
Tested and work as expected.

More details in: D126373-code

projects/packages/sync/src/modules/class-woocommerce.php
mavegaf: Moved to the constructor 👍

mavegaf:
The PR was moved from `draft` to `ready for review` again and tested by the backup team

fgiannar (dismissed):

jwebbdev (approved):

PR #33749, opened 2023-10-24
Boost: Add deprecation notice for lazy-loading

haqadn:
#33170 

## Proposed changes:
* Add a notice to lazy-loading mentioning that this feature is going away.
* Keep the module visible but disable the toggle if the feature is not available(force disabled).

### Other information:

## Jetpack product discussion
pc9hqz-2e6-p2

No

## Testing instructions:
<details>
<summary>See if you have the notice while the module is still working.</summary>

</details>
<details>
<summary>Install Gutenberg plugin 16.6.0+ or WP 6.4 and test if the module is force disabled</summary>

</details>

### DIFF
--- a/projects/plugins/boost/app/assets/src/js/elements/Notice.svelte
+++ b/projects/plugins/boost/app/assets/src/js/elements/Notice.svelte
@@ -15,7 +15,7 @@
 	};
 
 	export let level: 'info' | 'warning' | 'success' | 'error' = 'info';
-	export let title: string | null = null;
+	export let title: string;
 	export let message: string;
 	export let actions: ActionButton[] = [];
 	export let hideCloseButton = true;
@@ -31,7 +31,7 @@
 				key: index,
 				isLoading: !! action.isLoading,
 				disabled: !! action.disabled,
-				isExternalLink: !! action.isExternalLink || false,
+				isExternalLink: !! action.isExternalLink,
 				variant: action.variant || 'primary',
 			},
 			action.label
--- a/projects/plugins/boost/app/assets/src/js/elements/Notice.svelte
+++ b/projects/plugins/boost/app/assets/src/js/elements/Notice.svelte
@@ -10,12 +10,14 @@
 		onClick?: () => void;
 		isLoading?: boolean;
 		disabled?: boolean;
+		isExternalLink?: boolean;
+		variant?: 'primary' | 'secondary' | 'link' | 'tertiary';
 	};
 
 	export let level: 'info' | 'warning' | 'success' | 'error' = 'info';
-	export let title: string;
+	export let title: string | null = null;
 	export let message: string;
-	export let actions: ActionButton[] | undefined;
+	export let actions: ActionButton[] = [];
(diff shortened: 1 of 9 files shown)


haqadn:
@ederrengifo I improvised on the notice a little bit

projects/plugins/boost/app/assets/src/js/elements/Notice.svelte
```diff
@@ -10,12 +10,14 @@
 		onClick?: () => void;
 		isLoading?: boolean;
 		disabled?: boolean;
+		isExternalLink?: boolean;
+		variant?: 'primary' | 'secondary' | 'link' | 'tertiary';
 	};
 
 	export let level: 'info' | 'warning' | 'success' | 'error' = 'info';
-	export let title: string;
+	export let title: string | null = null;
```
pyronaur: If this is to make title optional, IIRC it should be `| undefined = undefined`

projects/plugins/boost/app/assets/src/js/elements/Notice.svelte
```diff
@@ -29,6 +31,8 @@
 				key: index,
 				isLoading: !! action.isLoading,
 				disabled: !! action.disabled,
+				isExternalLink: !! action.isExternalLink || false,
```
pyronaur: `!!` is already type-casting, I don't think you need false there. Although I'm not a fan of `!!` in general, but since it's used around there, I guess 🤷

projects/plugins/boost/app/assets/src/js/pages/settings/elements/Module.svelte
```diff
@@ -36,13 +37,14 @@
 	} );
 </script>
 
-{#if isModuleAvailable}
+{#if isModuleAvailable || ! hideIfUnavailable}
```
pyronaur: if not hide if unavailable 😕

projects/plugins/boost/app/assets/src/js/pages/settings/sections/Modules.svelte
```diff
@@ -202,7 +216,7 @@
 		</p>
 	</Module>
 
-	<Module slug="lazy_images">
+	<Module slug="lazy_images" hideIfUnavailable={false}>
```
pyronaur: Seems like we're adding an optional property to component that's only used in one component. Maybe it would be just simpler to wrap the content of the slot with a single `if` rather than changing the component?

pyronaur (changes requested):

projects/plugins/boost/app/assets/src/js/pages/settings/sections/Modules.svelte
haqadn: I was thinking in terms of using this for ISA as well. The latest design suggestion keeps the ISA module visible even in free. And, I think we'll need a toggle button to disable the fixes applied by ISA.

pyronaur (approved):
LGTM 🚢

PR #33761, opened 2023-10-24
Fix/login button disappears when logged in to wpcom

chrismccluskey:
Fixes 168-gh-automattic/gold

## Proposed changes:

* Require the is_user_logged_in check to be performed on wpcom platform.

### Other information:

- [x] Have you written new tests for your changes, if applicable?
- [x] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [x] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

### Setup
- Sandbox the public api and subscriptions
- Create a new JN site, configure jetpack constant and setup global http
- Connect stripe and Jetpack on the JN site
- Setup paid plans for the paid content block
- Create a post with a paid content block

### Replicate the issue
- Open private window, purchase sub via paid content block and pay for plan.
- Close private window, then open a new private window and return to post, ensure you cannot read the post.
- Open new tab in private window and login to wordpress.com with the subscriber account you used to pay for plan.
- Return to the browser tab with the post containing the paid content block and refresh, you should still not see the content.
- Click login, login button disappears.

### Test the patch
- Close any private windows you had open previously from replicating the issue
- Activate this branch in JN
- Apply D126524-code to your sandbox
- Run `bin/jetpack-downloader test jetpack fix/login-button-disappears-when-logged-in-to-wpcom` and apply on top of D126524, answering `n` to the first question asking if you want apply it to trunk.
- Open a new private window and return to post, ensure you cannot read the post.
- Open new tab in private window and login to wordpress.com with the subscriber account you used to pay for plan.
- Return to the browser tab with the post containing the paid content block and refresh, you should still not see the content.
- Click login, and you should be able to see the content.

PR #33763, opened 2023-10-25
Paywall: "Log in" Link – Improve the experience for WPCOM users

lezama:
## Proposed changes:

Improves the experience for WPCOM Users accessing subscriptions.

* Paywall "Log in" link – use the Magic Link flow instead of the "code flow" .
* User is not authenticated but logged in to wpcom – authenticate the user instead of going through the code flow. 
* User is not authenticated and not logged in to wpcom – send the user to the magic link flow instead of the code flow.

Next steps:
* Add site context to the login screen.
* Add site context to emails.
* User is not authenticated and not logged in to wpcom – Using a similar iframe as the "code flow" one, request a login link from the iframe without leaving the site to keep more context.
 
### Other information:

## Testing instructions:

Create a subscriber-only post in the following types of sites:

- Custom domain site
- Atomic site
- Jetpack site

Testing Scenarios:

- As a logged-in subscriber, test the link on all types of sites.
- As a logged-out subscriber, test the link on all types of sites.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -1069,32 +1069,36 @@ function get_paywall_blocks( $newsletter_access_level ) {
 		// translators: %s is the name of the site.
 		: esc_html__( 'Subscribe to get access to the rest of this post and other subscriber-only content.', 'jetpack' );
 
-	$access_question = $is_paid_post
-		? __( 'Already a paid subscriber?', 'jetpack' )
-		: __( 'Already a subscriber?', 'jetpack' );
+	$sign_in         = '';
+	$switch_accounts = '';
+	if ( is_user_auth() ) {
+		if ( ( new Host() )->is_wpcom_simple() ) {
+			$switch_accounts_link = wp_logout_url( get_current_url() );
+			$switch_accounts      = '<!-- wp:paragraph {"align":"center","style":{"typography":{"fontSize":"14px"}}} -->
+<p class="has-text-align-center" style="font-size:14px"><a href="' . $switch_accounts_link . '">' . __( 'Switch Accounts', 'jetpack' ) . '</a></p>
+<!-- /wp:paragraph -->';
 
-	$sign_in = '';
-	$host    = new Host();
-	if ( ! is_user_logged_in() && $host->is_wpcom_simple() ) {
-		$id           = '';
-		$sign_in_link = wpcom_logmein_redirect_url( get_current_url(), false, null, 'link' );
-		$button_text  = esc_html__( 'Log in', 'jetpack' );
-	} elseif ( class_exists( 'Automattic\Jetpack\Extensions\Premium_Content\Subscription_Service\Jetpack_Token_Subscription_Service' ) &&
-		! $host->is_wpcom_simple() ) {
-		// We are on Jetpack and no cookie is set
-		$id              = 'jp_retrieve_subscriptions_link';
-		$sign_in_link    = '#'; // listening to "click" event in view.js
-		$access_question = __( 'I am already subscribed.', 'jetpack' );
-		if ( ! Jetpack_Token_Subscription_Service::has_token_from_cookie() ) {
-			$button_text = esc_html__( 'Log in', 'jetpack' );
-		} else {
-			$button_text = esc_html__( 'Switch accounts', 'jetpack' );
 		}
-	}
+	} else {
+		if ( ( new Host() )->is_wpcom_simple() ) {
+			// custom domain
+			$sign_in_link = wpcom_logmein_redirect_url( get_current_url(), false, null, 'link' );
+		} else {
+			$sign_in_link = add_query_arg(
+				array(
+					'site_id'      => intval( \Jetpack_Options::get_option( 'id' ) ),
+					'redirect_url' => rawurlencode( get_current_url() ),
+					'v2'           => '',
+				),
+				'https://subscribe.wordpress.com/memberships/jwt'
+			);
+		}
+		$access_question = $is_paid_post
+			? __( 'Already a paid subscriber?', 'jetpack' )
+			: __( 'Already a subscriber?', 'jetpack' );
 
-	if ( ! empty( $sign_in_link ) ) {
 		$sign_in = '<!-- wp:paragraph {"align":"center","style":{"typography":{"fontSize":"14px"}}} -->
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -1122,6 +1115,23 @@ function get_paywall_blocks( $newsletter_access_level ) {
...
+ *
+ * @return boolean
+ */
+function is_user_auth() {
+	if ( is_user_logged_in() && ( new Host() )->is_wpcom_simple() ) {
+		return true;
+	}
+	if ( class_exists( 'Automattic\Jetpack\Extensions\Premium_Content\Subscription_Service\Jetpack_Token_Subscription_Service' ) ) {
+		if ( Jetpack_Token_Subscription_Service::has_token_from_cookie() ) {
+			return true;
```
millerf, line 1141: This assumption is false: the user can have the code and NOT be logged in to WPCOM.

For example: 
- When coming back from the checkout page, user has the subscription token but is NOT logged in to WPCOM.
- WPCOM session can expire or the user log out of WPCOM and still have the token on the hosted site

lezama: Yes, this method is not intended to check if the user is logged in to WPCOM.
It checks if it has valid credentials to check subscription status.

Returns `true` if we are on Simple and the user is logged in, or when we have a token/cookie.

We need it so we display "Switch accounts" whenever we know the users is authenticated(not logged in) and we know their membership status.

Method name and description could be better I guess, any suggestions?

simison: @millerf  What's the consiquence? 

If they're seeing the subscriber-only content at the site because of the token is present, but not paid content because they're not paid member, we suggest them to switch accounts based on this check?

simison: `has_access_to_subscriber_only_content` would be very clear? :-)

millerf: > we suggest them to switch accounts based on this check?

I agree this is not the best wording. The "Switch account" is here if they entered the wrong email for example when "getting access". Otherwise they have no way to remove the "wrong" token...

millerf: Use case is:
- User has a xx@gmail.com email address and has previously paid for the subscription and should have access
- User enters yx@gmail.com by mistake with the "Get access" (by error or because they use multiple emails)
- Then user does not have a way to get access with xx@gmail.com, unless we have a "switch account" flow...

millerf: > has_access_to_subscriber_only_content would be very clear? :-)

I kind of see what you are trying to do there, but:
- on WPCOM, you could be logged in WPCOM and not be a site subscriber
- on self-hosted, you could have a token and your subscription still pending (email not verified)

But I agree it means they have an account on WPCOM. So `has_wpcom_account()` would work...

Even after that, I don't see how does removing the "iframe flow" on self-hosted improves the user experience...

lezama: > on WPCOM, you could be logged in WPCOM and not be a site subscriber

so you get a switch account link.

> on self-hosted, you could have a token and your subscription still pending (email not verified)

you will get the message in the paywall. to verify the email.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -1073,30 +1073,23 @@ function get_paywall_blocks( $newsletter_access_level ) {
...
+	$button_text     = esc_html__( 'Log in', 'jetpack' );
+	$sign_in_link    = 'https://subscribe.wordpress.com/memberships/jwt?site_id=' . \Jetpack_Options::get_option( 'id' ) . '&redirect_url=' . get_current_url();
+	$access_question = __( 'I am already subscribed.', 'jetpack' );
+
+	if ( is_user_auth() ) {
+		$button_text = esc_html__( 'Switch accounts', 'jetpack' );
+		if ( ( new Host() )->is_wpcom_simple() ) {
+			// custom domain
+			$sign_in_link = wpcom_logmein_redirect_url( get_current_url(), false, null, 'link' );
 		} else {
```
millerf, line 1086: this "else" is not needed BTW: $sign_in_link is not modified from line 1077

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -1122,6 +1115,23 @@ function get_paywall_blocks( $newsletter_access_level ) {
...
+ * Returns true if user is auth for subscriptions check, otherwise returns false.
+ *
+ * @return boolean
+ */
+function is_user_auth() {
+	if ( is_user_logged_in() && ( new Host() )->is_wpcom_simple() ) {
+		return true;
+	}
+	if ( class_exists( 'Automattic\Jetpack\Extensions\Premium_Content\Subscription_Service\Jetpack_Token_Subscription_Service' ) ) {
+		if ( Jetpack_Token_Subscription_Service::has_token_from_cookie() ) {
```
simison, line 1140: Does `has_token_from_cookie` take into account token in the URL, https://github.com/Automattic/jetpack/pull/33761/files?

millerf: We're moving out of the token in URL. Source of truth is in the header. If it is in the URL, [it should be set early](https://github.com/Automattic/jetpack/blob/1fbf6b7ef3bbb8e3ccf98f7fcc3dfe8f9510f3cb/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php#L38) in the $_COOKIE.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
lezama: 👍 good catch, forgot to delete it when I moved it up

n3f (approved):
Worked for me.  I love how consistent this makes it across all the different flows.  (There were some small differences, which is something we need to stay aware of in case of regressions, but they were minor).

PR #32428, opened 2023-08-11
Introduced a PHPCS ignore statement to unblock PR merges.

zinigor:
This is not a fix, just a temporary stopgap to allow tests to pass on pull requests.

## Proposed changes:
* Introduces a phpcs:ignore statement on the usage of `utf8_encode` that's deprecated in PHP 8.2.

### Other information:
- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* Make sure tests pass.

PR #33777, opened 2023-10-25
CRM: Removing usage of deprecated function utf8_encode

coder-karen:
## Proposed changes:

* This PR removes the `utf8_encode` as it is now deprecated, and also removes the `zeroBSCRM_utf8ize` function in which it was located (it is not used by any extensions either). Instead, the two places in core CRM where the `zeroBSCRM_utf8ize` function was used (`zeroBSCRM_AJAX_mailDelivery_validateSMTP` and `zeroBSCRM_AJAX_mailDelivery_validateSMTPPorts`) now use `wp_json_encode()` with the `JSON_UNESCAPED_UNICODE` option` passed in.
* One suggestion was to use `mb_convert_encoding` to replace the original instance of `utf8_encode`, and I can confirm `mb_convert_encoding` does the same as `utf8_encode`, using example code provided here - https://www.php.net/manual/en/function.utf8-encode.php in testing.
* The other suggestion in the issue to remove unnecessary code also makes sense and is the approach used here.
* The output of `wp_json_encode( $res, JSON_UNESCAPED_UNICODE );` differs slightly to the original function, but from my perspective is a better output.
  * To clarify more, given a $res of `array(
		"debugs" => array("success:<h2>Debugging SMTP<\/h2>Some unexpected characters here: á"')
	);`, the previous output in the debug output would be `Debugging SMTP<\/h2>Some unexpected characters here: Ã¡` with original code, and
	`Debugging SMTP<\/h2>Some unexpected characters here: á` with new code (in `zeroBSCRM_AJAX_mailDelivery_validateSMTP`). The result is similar in `zeroBSCRM_AJAX_mailDelivery_validateSMTPPorts`, with the correct character being passed in. 

### Other information:

## Jetpack product discussion

https://github.com/Automattic/zero-bs-crm/issues/3277
https://github.com/Automattic/jetpack/pull/32428

No.

## Testing instructions:

* To test this out, you can try printing the following in your error log if not in either `zeroBSCRM_AJAX_mailDelivery_validateSMTP` or `zeroBSCRM_AJAX_mailDelivery_validateSMTPPorts`:
```
$res = array(
		"debugs" => array( "success:<h2>Debugging SMTP<\/h2>Some unexpected characters here: á" )
	);
```
Then on the branch:
```
error_log( wp_json_encode( $res, JSON_UNESCAPED_UNICODE ));
```
vs on trunk
```
error_log( json_encode( zeroBSCRM_utf8ize ($res ) ) );
```

* You can also test using an smtp value like á.smtp.com on both trunk and on the branch and visiting the network tab in the inspector console, clicking on the most recent admin-ajax.php call, and viewing the response (or error_log the response from `zeroBSCRM_AJAX_mailDelivery_validateSMTP` directly). For `zeroBSCRM_AJAX_mailDelivery_validateSMTPPorts` you can test by adding a $res into the code before the echo at the end of that function (use the $res above) - the result will display in the debug output.
* You can also test with a character encoding such as `\xEB` and the result should be the same.

### DIFF
--- a/projects/plugins/crm/admin/settings/mail-delivery.ajax.php
+++ b/projects/plugins/crm/admin/settings/mail-delivery.ajax.php
@@ -503,8 +503,7 @@ function zeroBSCRM_AJAX_mailDelivery_validateSMTP() {
 	}
 
 	header( 'Content-Type: application/json' );
-	// requires zeroBSCRM_utf8ize for proper debug passing
-	echo json_encode( zeroBSCRM_utf8ize( $res ) );
+	echo wp_json_encode( $res, JSON_UNESCAPED_UNICODE );
 	exit();
 }
 
@@ -574,8 +573,7 @@ function zeroBSCRM_AJAX_mailDelivery_validateSMTPPorts() {
 
 	$res['open'] = $okay;
 	header( 'Content-Type: application/json' );
-	// requires zeroBSCRM_utf8ize for proper debug passing
-	echo json_encode( zeroBSCRM_utf8ize( $res ) );
+	echo wp_json_encode( $res, JSON_UNESCAPED_UNICODE );
 	exit();
 }
 
--- a/projects/plugins/crm/changelog/remove-crm-utf8_encode-usage
+++ b/projects/plugins/crm/changelog/remove-crm-utf8_encode-usage
@@ -1,0 +1,4 @@
+Significance: patch
+Type: changed
+
+Mail delivery: Removed usage of deprecated function utf8_encode
--- a/projects/plugins/crm/includes/ZeroBSCRM.GeneralFuncs.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.GeneralFuncs.php
@@ -905,22 +905,6 @@ function zeroBSCRM_isJson( $str ) {
 		return $endArr;
 	}
 
-
-	// recursive utf8-ing 
-	// https://stackoverflow.com/questions/19361282/why-would-json-encode-return-an-empty-string
-	function zeroBSCRM_utf8ize($d) {
-	    if (is_array($d)) {
-	        foreach ($d as $k => $v) {
-	            $d[$k] = zeroBSCRM_utf8ize($v);
-	        }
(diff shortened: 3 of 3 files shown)


gogdzl (approved):
The original functions seems to double encode strings, that's why you got Ã¡ (`\u00c3\u00a1`) in your tests.

Your code looks good, and I agree with the changes, but I'm curious about what problem  `zeroBSCRM_utf8ize` was trying to solve.

coder-karen:
Thanks for reviewing @gogdzl !

One thought is that this function was possibly used elsewhere? Especially given the Stack Overflow link in the original function comment. Given the output in the functions that used it, I can't see how they could differ. And Googling the same issue (`json_encode` returning empty results), using `JSON_UNESCAPED_UNICODE` (though in this case with `json_encode`) is also given as a solution [here](https://medium.com/@onwuka_gideon/how-i-fixed-php-json-encode-returning-empty-result-e4ef97638ea1).

In an attempt to replicate the original issue I ran the following through the different functions: `array( "debugs" => array("") );`
And they all resulted in:
```
{"debugs":[""]}
```
Similarly `array( "debugs" => "" );` resulted in `{"debugs":""}` in all cases.
Perhaps if we knew what the original issue was we might be able to replicate, but I'm not too sure how we'd do that. I'm happy to wait for additional reviews too incase anyone has any thoughts on this too.

coder-karen:
Sending a ping to @tbradsha in case you have any additional historic context on this:

>  I'm curious about what problem zeroBSCRM_utf8ize was trying to solve.

Otherwise I can merge later today.

tbradsha (approved):
The review sent me down a rabbit hole. I don't have any further context as to the goal here, but thanks for going with the "chop code and use native functions" route over the patch route; going forward it's much more readable/clear what is happening.

If this function remained in its current state and was unwittingly used somewhere else where we store data, it could result in data integrity issues once we tweaked or replaced the original function. Because the function is no more, the worst that'd happen is bad debug output (and with this tweak, it's actually better debug output).

The end result looks good!

Issue #33048, opened 2023-09-13
Lazy Images: deprecate functionality


kraftbj:
### Impacted plugin

Jetpack, Boost

### Quick summary

The Lazy Images functionality of Jetpack and Boost is not compatible with new work being done in Gutenberg (shipped 16.6). This feature has overall been seeing reduced necessity due to the advancement of browser's native loading.

This is to track the overall deprecation of this feature.

### Logs or notes

Internal reference: pdWQjU-wV-p2

```[tasklist]
### Tasks
- [ ] https://github.com/Automattic/jetpack/pull/33039
- [ ] https://github.com/Automattic/jetpack/issues/33049
- [ ] https://github.com/Automattic/jetpack/issues/33170
- [x] Deprecate package
```

kraftbj:
Open question that I have:

Do we want to kill off the feature straightaway or have it stop working with WP 6.4, so existing clients can continue to use it for the moment until they update? I don't have a strong opinion and curious what others think.

jeherve:
@kraftbj I would have it stop working with Jetpack 12.7, that will come out around the time of WP 6.4.

I did have a question on my end as well, about the deprecation of the package:

Should we mark all the methods in the package as deprecated then deprecate the whole package, or should we leave it as it is and only stop using the package in our plugins? If the package remained usable, without any deprecation warnings, we could create a small plugin that only requires the package, for folks who would want to continue to use this package over Core's functionality.

cc @haqadn

kraftbj:
> @kraftbj I would have it stop working with Jetpack 12.7, that will come out around the time of WP 6.4.

Do you think we should aim for the October release to have the disable/unable to activate changes? But for sites actively using it without manual intervention, let it ride until removing it outright in November?

> Should we mark all the methods in the package as deprecated then deprecate the whole package

Personally, I would deprecate the package since we actively know it will break functionality in upcoming versions of WordPress without an intent on our end to resolve. Folks can still use a deprecated package/functions, but at least they should check to see why we did and if it matters to them. I wouldn't, say, delete them from Packagist or anything like that.

jeherve:
> Do you think we should aim for the October release to have the disable/unable to activate changes? But for sites actively using it without manual intervention, let it ride until removing it outright in November?

Yes, that's what I was planning. 👍

> Personally, I would deprecate the package since we actively know it will break functionality in upcoming versions of WordPress without an intent on our end to resolve.

👍

foleynotrose:
[...]

PR #19269, opened 2021-03-23
Search: Lazy load images

jsnmoon:
Fixes #19171.

#### Changes proposed in this Pull Request:
* Modularizes `photon` library invocation into a `usePhoton` hook.
* If `IntersectionObserver` is available, enable lazy loading of images. Otherwise, greedily load images as before.

#### Does this pull request change what data or activity we track or use?
No.

#### Testing instructions:
* Apply this change to your site with a Jetpack Search subscription.
* With either the Expanded or Product result format enabled, perform a search.
* Ensure that the result images are loaded lazily. You can confirm this via your browser's network console.
* Using a browser that doesn't support [`IntersectionObserver`](https://caniuse.com/intersectionobserver), repeat the testing instructions above. Ensure that the images (greedily) load as expected.

PR #33782, opened 2023-10-25
Lazy Images: remove from Jetpack plugin

### LINKED ISSUE #8281: Lazy Images: Improve reflow for images
Currently, we try and set the width and height attributes when possible for any image we lazy load. Ideally, this would help the browser to reserve the correct amount of space in the window to minimize reflows as we load the image later on.

But, many themes do something like the following for responsive layouts:

```css
img {
    max-width: 100%;
    height: auto;
}
```

Since the theme's CSS sets the `height` for the `img` tag, it overrides the height we set via the image attribute, which results in the image not reserving the correct amount of vertical space.

This doesn't break lazy images, but it can cause significant reflow in some cases, and in other cases, especially where a post has a ton of images, it may result in all of the images loading on `$( document ).ready`.

I think we can do better with the reflowing if we set the height and width via the `style` attribute on the image tag. Although, for the best experience, we may need to look at setting the aspect ratio as opposed to strict height and width values.

### LINKED ISSUE #8773: Lazy Images: Provide better placeholder options
Currently, we load a 1x1 transparent gif as a placeholder image when we initially load a page with lazy images. This seems to work functionally, but perhaps it'd be a better UX if we loaded placeholders similar to Calypso to show that something will be there and is being fetched?

### LINKED ISSUE #8879: Lazy Images: Breaks outdenting of big images in compatible themes

* Activate Lazy Images
* Activate any theme that outdents images via JavaScript (e.g. Independent-Publisher-2 or TwentySixteen)
* Watch posts with big images within the article (> 1100px width)

#### What I expected

Images should be "outdented", breaking out of the page width as the overall viewport is big enough to support this.

#### What happened instead
Images are not "outdented", stay within given article width.

### Additional Notes

The reason for this is that most themes only outdent once at the `load` or `ready` event and even then only look at the image width given by the DOM API.

I was able to fix this by creating a child theme of independent-publisher-2, adding a `lazyOutdent` method that was called by every lazy loaded image's `load` event handler (this also enabled me to add CSS transitions), but this might have the side effect of being called on any `load` and not just the one triggered by the lazy load - a JS hook might fix that.

Also using the `naturalWidth` (or - as a fallback - `offsetWidth`) DOM attribute was necessary to get the correct sizes of an image as supplied by `srcset` attributes to detect if an image will actually be big enough to be outdented after the browser selected the most appropriate one from the source set (`width` reports wrong values in that case).

---

**PS**: You might also want to look at this lazy loading technique which makes the use of `<noscript>` tags redundant by using an empty image for the `srcset` attribute, yet keeping the `src` intact for old browsers and making this solution better for sharing widgets. Also uses a data URL instead of an actual 1x1 pixel GIF to reduce HTTP overhead: https://github.com/ivopetkov/responsively-lazy

### LINKED ISSUE #8897: Lazy Load: Breaks Images in with Ultimate Shortcode Slider
Lazy Load causes broken images in a slider created with Ultimate Shortcodes. 

1. Install Ultimate Shortcodes plugin.
2. Create a post that includes a slider with images.
3. Load the post. Notice that no images appear.
4. Deactivate Lazy Load. The images appear.

### LINKED ISSUE #8905: Lazy load for background images
Just insert any image as a background for a section of the page. Any themes using image parallax will do this and pretty much any page created with WP Bakery (formerly Visual Composer) can easily add an image parallax to any row.

I see the preg_match_all function on Jetpack code is searching for
``

Problem is when you set an image as the background the code will look like this:

```
<section id=”section1″ data-overlay=”5″ data-vc-parallax-image=”https://image.url”> ::before
<div class=”background-image-holder” style=”transform: translate3d(0px, -1262.31px, 0px); background: url(“https://image.url”); opacity: 1;”>
```

Notice how the image is now loaded thru the "background:url" property. 
Jetpack won't act on those images.

#### What I expected
To have all images lazy loaded, including the ones set as layer backgrounds.

#### What happened instead
Jetpack doesn't find those images and therefore no lazy loading happens for them, which negatively impacts page performance, specially in Google Lighthouse. 

#### Why have this feature added
Most modern themes use image parallax effects, which is a really nice effect, and lazy loading those images is a must in terms of performance. 

Thanks!

### LINKED ISSUE #8917: Images Above Anchor Link Are Not Lazy Loaded
A user in the WordPress.org forum noticed that the Jetpack Lazy Load implementation does not lazy load images that are above a linked section within a page.

This implementation is different from an alternative plugin's approach, which the user recommended:

https://wordpress.org/plugins/simple-lazyload/

You can see an example of this alternative implementation by going here and scrolling up:

http://www.jangadafilmes.com.br/enin-2015/#dia13

This type of approach may be something that can be incorporated into future versions of the plugin.

### LINKED ISSUE #8996: Request: Provide init options to lazy load
It'll be really great to be able to customize some things via a JS options array and/or plugin settings:

1. The scrolling distance
2. Post-loading actions
3. Image selection/exclusion tweaks

### LINKED ISSUE #9043: Lazy Images: better support for content loaded via AJAX
> Whenever I use AJAX to load content from another page of my site, only the 1×1 placeholder comes in. How do I reinitialize JetPack’s lazy images to pop the proper image back in after Ajax?
>
> Just a basic jQuery load, loading HTML from one page into the current one.
-- https://wordpress.org/support/topic/using-ajax-only-loads-1x1-lazy-placeholder/

For now, here is a work around:
```js
function update_thumbs() {
    $('YOUR IMAGE SELECTOR').each(function() {
      var src = $(this).prop('src');
      if (src.indexOf('1x1.trans.gif') !== -1) {
        $(this).prop('src', $(this).data('orig-file'));
      }
    });
  }
```

### LINKED ISSUE #9509: Lazy Images: Images do not load with Maximage2 library

1. Activate the Soho theme on a site running Jetpack. See: https://wordpress.com/theme/soho
2. Make sure Lazy load images are activated.
3. Set up home page according to theme documentation, and enable slideshow: https://wordpress.com/theme/soho/setup
4. View home page.

#### What I expected

The slideshow should play, and show full width images. Like: 

screen shot 2018-05-08 at 11 55 44 am

#### What happened instead

Instead, you get a blank looking page, and the placeholder GIF is loading instead. Like: 

screen shot 2018-05-08 at 11 56 59 am
https://cloudup.com/cB52TGXiuB2

If you turn off Lazy load images, the slideshow works.

Some further testing reveals it seems to be caused by this library included in the theme: http://www.aaronvanderzwan.com/maximage/ You don't get any errors, but if you remove that, the slideshow plays. I wanted to report it here since the theme works fine without it, and the root cause might be in Lazy load images.

### LINKED ISSUE #9595: Lazy Images: Layouts using imagesLoaded.js and Masonry.js not working as expected
Themes using imageLoaded.js with masonry.js for layouts can end up with gaps and overlaps when Lazy Images are enabled. 

1. Activate a theme using Masonry and imagesLoaded JavaScript libraries for layout. From themes on WP.com, one good option is Baskerville 2.
2. Make sure Lazy load images are activated.
3. Make sure you have a number of posts with featured images.
4. Load the blog posts page.

#### What I expected

The posts would be displayed evenly spaced, like here: https://baskerville2demo.wordpress.com/

#### What happened instead

The posts are often further apart, or slightly overlapped. Resizing the browser window will correct the issue. 

On page load (note the gaps, and the overlap in the bottom-right corner):

After resizing the browser: 

Originally reported for the Opti theme here: 

* 2365-wpcom-premium-themes
* Follow-up ticket: 1143441-zen.

### LINKED ISSUE #9827: Lazy Loaded images hide featured images in no-JS mode.
1. Turn on Lazy Loading of images.
2. Visit the site where you can see at least one post with a featured image (I tested on Twenty Fifteen).
3. Disable JS.

#### What I expected
Featured images should be displayed as well as all other content images.

#### What happened instead
Featured images are replaced with a transparent image, and not loaded.

cc @ebinnion because this is a follow-up to #9780

### LINKED ISSUE #9896: Lazy Images Causes Full Size WooCommerce Product Image to Load When Selected
With Jetpack (6.3.2) and WooCommerce (3.4.3) installed, WooCommerce Lightbox [manually disabled](http://www.remicorson.com/disable-woocommerce-3-lightbox/), and Lazy Images active, selecting a WooCommerce Product image causes the full size image to load in the browser. After deactivating Lazy Images, selecting the image does nothing.

Why does Lazy Images cause this change in behavior? Is this intended? If so, should it be?

Reported [in the forums](https://wordpress.org/support/topic/infinite-scroll-lazy-load-3/).

### LINKED ISSUE #9929: Lazy Images module compatibility with "TwentyTwenty" and "Twenty20 Image Before-After" plugin

1. Enable Lazy Images module
2. Install either [TwentyTwenty](https://wordpress.org/plugins/twentytwenty/) or [Twenty20 Image Before-After](https://wordpress.org/plugins/twenty20/) plugins
3. Try those plugins

#### What I expected

I expected those plugins to show images correctly.

#### What happened instead

The don't load correctly.

### LINKED ISSUE #9974: Lazy Load: Header image gets temporarily expanded during load

Up until now, this has only has been reproduced on this site:

https://optstroy-prom.ru/

1. Enable Lazy Loading for Images here: https://optstroy-prom.ru/wp-admin/admin.php?page=jetpack#/settings
2. Refresh the site: https://optstroy-prom.ru/
3. You will notice the header gets temporarily expanded:
lazy-loading

#### What I expected

The site to load normally. As it does when Lazy Loading is deactivated.

#### What happened instead

The header got expanded while loading the site and then got back to normal.

Initially reported here: 5419074-hc

### LINKED ISSUE #10016: BuddyPress Group images can't be cropped when Lazy Images is enabled.

1. Activate BuddyPress and the User Groups feature.
2. Create a new Group. When you get to the photo, upload a photo from your computer.

#### What I expected

The photo should appear and be ready for cropping:

image
(Lazy Images disabled ^ )

#### What happened instead

The photo does not load:

image

1328476-zen

### LINKED ISSUE #10046: Lazy Images: Photon: WooCommerce: Item preview (product image) is low resolution when expanded
One of our hosting partners reported that the WooCommerce product image was low resolution after being expanded when using lazy images and photon.

This was reported to me via another a11n, so I don't have more details than that. But, we should investigate.

### LINKED ISSUE #10606: Lazy Load: Images not lazy loaded

1. Go to https://www.plastic2go.com.au/vented-plastic-fruit-vegetable-crates/
2. Check the top 2 content images' markup
3. Check loading of same images in developer tools network tab

#### What I expected

I expected the images to have lazy loading markup and to be loaded via the "jquery" initiator.

#### What happened instead

No lazy loading markup and "(index)" initiator. Two other images on the page have the correct markup and one of them is loaded via lazy loading (the other one is the same as the topmost image).

I thought this might be because the topmost image is included again on the page, so I tested a different image at the top of the page, and the other image was also incorrectly rendered.

### LINKED ISSUE #10607: Lazy Load not rendering base64 image
I have some image base64 encoded data:image/svg+xml;base64,PD94bW... when i active Lazy Load from jetpack it remove data: and image are not displayed.
I try also png tyoe and not work

### LINKED ISSUE #10707: CSP issues - remove inline scripts
#### Is your feature request related to a problem? Please describe.
Yes, I tried to enable [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) on my WordPress site. One of the things I wanted to block with CSP is inline scripts. To my surprise, I find out the JetPack add a lot of inline scripts. For example:
```
     <script>
	document.documentElement.classList.add(
		'jetpack-lazy-images-js-enabled'
	);
    </script>
```
I had to allow inline scripting in the policy so my site will keep working with the plugin.
#### Describe the solution you'd like
Use regular JS files instead of inline scripting.

#### Describe alternatives you've considered
Adding specific hashes for each one of the inline scripts (see it [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#Sources), down the list). This was not easy to go over all the scripts and might break on future releases.

### LINKED ISSUE #11054: Lazy Loading: Class exclusion images contain noscript tags
In 1655173-zen we have a user who has an embed plugin that includes image urls to copy and paste into other sites, and even though they are using the jetpack_lazy_images_blacklisted_classes filter it still is including the noscript tags included by lazy load (but excluding the lazy loading classes). Looking at code this appears to be because the should_skip_image_with_blacklisted_class method is called in the process_image_attributes method before we build the image, so it's just skipping over the lazy loading attributes versus skipping over the image entirely. We either need to add a second should_skip_image_with_blacklisted_class check inside the process_image_attributes process_image method or just move the entire check in there so that the noscript tags are not added.

https://github.com/Automattic/jetpack/blob/abf3988da482778e9d90ddb35b9ee2368a9d905d/modules/lazy-images/lazy-images.php#L182

### LINKED ISSUE #12344: Lazy Load: Prevents Images from showing in Elementor Pro Popups

1. Install Activate Elementor Pro. Make sure Jetpack Lazy Load images is going. 
2. Head to `Templates`>`Popups` and publish a pop up with an image
3. Take a look at your pop up on the live site. Images will be missing
4. Turn off Lazy Load. Your image will be back in the pop up upon refresh.

#### What I expected
I expected images to work on my site when lazy load was enabled

#### What happened instead
Lazy load prevents images from showing in an Elementor popup.

**Screenshots**
Turning Lazy Load on

lazy load

(Image link: https://cld.wthms.co/Lu8iuN )

Issue reported in 2023176-zen. 
For testing you can find our license for Elementor Pro in the Secret Store

### LINKED ISSUE #13518: Predominant color and low quality placeholders for lazy load
Lazy load is great. But it would be nice for user experience to see better placeholders (I'm aware of the filter, but it's not enough), I'd like a unique placeholder for each image.

Two techniques come to mind:

1. Predominant color placeholders: 
Shows a solid image of the predominant color, used by Google images, example:
google-images-placeholder

2. Low quality placeholder:
Consists on using a small pixelated version of the real image and apply a blurry CSS filter to it.
Details: https://jmperezperez.com/medium-image-progressive-loading-placeholder/

This is probably wishful thinking, but was worth asking for. Thanks!

### LINKED ISSUE #14161: Add lazy load for iframes (specifically shortcode embeds)
#### Is your feature request related to a problem? Please describe.

I have some pages with 20+ YouTube videos and the pages take quite some time to load on slow Internet connections.

#### Describe the solution you'd like

I noticed the Jetpack shortcode [youtube] uses an iframe to show the video, so a solution to my issue could be to use lazy load for iframes.

#### Describe alternatives you've considered

I saw some plugins do this, but I'd prefer to avoid adding more plugins to my WordPress installation.

### LINKED ISSUE #14191: Lazy Images: Disable when SiteGround's caching plugin is lazy loading
SiteGround's caching plugin:

https://wordpress.org/plugins/sg-cachepress/

... has a feature to lazy load images. If users have this installed along with Jetpack's lazy load feature, it can cause issues. Most of the time it's that images don't load properly.

I haven't investigated too deep, but I would think SG's solution integrates with their caching better than Jetpack's lazy loading, so we should consider disabling our feature when theirs is enabled.

When SG's lazy loading is enabled, `siteground_optimizer_lazyload_images` is set to `1` in `wp_options`. Maybe we can detect than and do this somewhere?

`add_filter( 'lazyload_is_enabled', false );`

### LINKED ISSUE #14281: WooCommerce Lazy Images Compat inline JS issue
When WooCommerce is active, Jetpack adds the following inline JS regardless of whether lazyload is enabled or not:

`jQuery( 'body' ).bind( 'wc_fragments_refreshed', function() {
			jQuery( 'body' ).trigger( 'jetpack-lazy-images-load' );
		} );`

It's inline script, so if jQuery is deferred (done normally for performance reasons), this causes a jQuery is not defined error. 

Source: https://github.com/Automattic/jetpack/blob/master/3rd-party/woocommerce.php#L105

I assume two things have to change:

- Check if Lazyload is even enabled.
- Ensure jQuery exists by executing on window.onload event. Or have it in a normal JS file with jQuery dependency rather than inline JS.

### LINKED ISSUE #15073: Lazy Images: Content is not processed if it contains an SSR Block that uses wp_get_attachment_image

I don't have an example in production, this is happening locally.

- Create a WordPress site using Jetpack Lazy Images and Woocommerce.
- Add a post with a [server side rendered](https://developer.wordpress.org/block-editor/packages/packages-server-side-render/) Woocommerce block.
  - One where it stores the IDs of the products rather than the rendered HTML.
  - Example: [Handpicked Product Block](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/master/assets/js/blocks/handpicked-products) 
- Add a core image block anywhere else in the post.

#### What I expected
I expected all images to be lazy loaded.

#### What happened instead
Instead, only the images in the server side rendered block are lazy loaded.

#### Trace

1. [Handpicked Products Block](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/master/assets/js/blocks/handpicked-products)
2. Uses Server Side Rendering, which is done in the [Abstract Product Grid class](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/src/BlockTypes/AbstractProductGrid.php#L319)
3. It calls [get_image](https://github.com/woocommerce/woocommerce/blob/master/includes/abstracts/abstract-wc-product.php#L1850) on each product, which uses `wp_get_attachment_image`
4. [process_image_attributes](https://github.com/Automattic/jetpack/blob/master/modules/lazy-images/lazy-images.php#L196) is applied to the image.
5. Later, [add_image_placeholders](https://github.com/Automattic/jetpack/blob/master/modules/lazy-images/lazy-images.php#L100) is applied to `the_content` but it bails because `data-lazy-src` is already present.

### LINKED ISSUE #16421: Jetpack Lazyload and Cumulative Layout Shift (CLS)
#### Is your feature request related to a problem? Please describe.

The recent google update has highlighted “Cumulative Layout Shift (CLS)” as an important feature to pay attention to.
Sadly, jetpack’s lazyload is damaging that metric, since when using it we get a shift in the content once the image is loaded.

#### Describe the solution you'd like
Several solutions would be great:
1. make lazy load wrap the img tag with <div style=”min-height:100px;”>  </div>
When the image height “min-height:100px;” is decided based on the height of the image.
This could be taken from the height parameter set in img (ideal first solution).
An even better fallback is to detect the image height and set it based on that (the best solution would offer to do it also for images that are hosted outside of the website).
2. to add a control parameter of "port height" that will indicate how close should the user be before the image will start loading. Currently, this value is 0. But if it were to be (say) 100px (or some other configurable method), then the images would be loaded before the user sees them (but after the rest of the page load, and only conditional on the user scrolling down).

These features would be great.

#### Describe alternatives you've considered
Turn off the plugin.

#### Additional context
https://web.dev/cls/

### LINKED ISSUE #17640: Lazy Images: images do not load after printing, in Safari
Follow-up from #17490 

> In Safari, the images did not finish reloading after I closed the print dialog box. I had to refresh the page, scroll down to load all the images, and then print. In Chrome, the images did load on their own without needing to refresh.

----

p8oabR-za-p2#comment-4693

### LINKED ISSUE #24002: Lazy Load: clarify that the feature settings are available in 2 different interfaces
### Quick summary

When Jetpack's Lazy load setting is enabled and the Jetpack Boost plugin is installed and activated, it by default enables the Lazy Load setting under the plugin.

And when I turn off Lazy load from the Jetpack Boost plugin, it turns off from the Jetpack Settings too. They both are interlinked.

Ref chat: 34778200-hc, Follow up: 4940855-zen

### Steps to reproduce

1. Have WordPress.com → Settings → Performance → Enable Lazy Loading for images enabled.
2. Install Jetpack Boost plugin and activate it.
3. The Lazy Load setting from the plugin is activated automatically.
4. Turn off Lazy load from Jetpack Settings, it turns off under the Jetpack Boost plugin too.
5. Turning off Lazy load under Jetpack Boost turns off Lazy load under Jetpack.

Didn't expect both the setting to be interdependent. Expected the setting to be stand-alone and work on its own.

### What actually happened

The settings are interdependent and cause confusion which one to use for the user. This also causes the images on the site to not load at some instance.

### Context

Customer report.

Internal reference and discussion as per potential next steps to avoid such confusion:
p1HpG7-fGP-p2

### Simple, Atomic or both?

Atomic

### Theme-specific issue?

_No response_

### Browser, operating system and other notes

_No response_

### Reproducibility

Consistent

### Severity

Most (> 50%)

### Available workarounds?

_No response_

### Workaround details

_No response_

### LINKED ISSUE #25659: Lazy Images: Woo product image not loading in Safari
There seems to be an issue with Jetpack lazy images + WooCommerce sometimes not displaying the product image correctly in **Safari**.

**Steps used to reproduce:**

* Create a JN site with WooCommerce and Jetpack. I tested with the 'Twenty Twenty' theme.
* Connect Jetpack and make sure lazy loading is enabled at `wp-admin/admin.php?page=jetpack#/performance`
* Activate WooCommerce and publish a test product with a product image included.
* View the published product page in Chrome/Firefox and refresh the page a few times, the product image should always display as expected.
* Next view the product page in Safari, the product image may not load. Note that this seems to be somewhat intermittent. Try refreshing the product page a few times to see.
* Disable Jetpack lazy images and try viewing again in Safari and the product image should always load as expected.

**Other notes:**

* When the image doesn't display, you can hover over the image and the zoomed portion displays okay.
* If you disable the hover zoom effect with the below snippet, the issue seems to go away even with JP lazy images enabled:

```php
function remove_image_zoom_support() {
    remove_theme_support( 'wc-product-gallery-zoom' );
}
add_action( 'wp', 'remove_image_zoom_support', 100 );
```

via: https://github.com/Automattic/jetpack/issues/23553#issuecomment-1215470801

### LINKED ISSUE #29585: Enhancement: Exclude featured images from Lazy Load in block themes
### Impacted plugin

Boost

### What

When image lazy loading is activated, the featured image in the Twenty Twenty-Three block theme will be lazy-loaded. This is not good for pagespeed.

### How

Wanting to exclude the featured image, displayed at the top of a single post in the Twenty Twenty-Three block theme from being lazy-loaded, I tried to give the featured image block the class `skip-lazy`. This does not have the desired effect. The class is set on the parent FIGURE element, not the IMG element itself and the image is still marked for lazy loading. Same goes for any regular image block, by the way...

To reproduce: install a clean WP 6.2-RC with some demo content with featured images, install Jetpack Boost (or Jetpack) and activate Lazy Image Loading. Open a post with a featured image in your browser, right-click featured image and choose Inspect. Notice the classes `attachment-post-thumbnail size-post-thumbnail wp-post-image jetpack-lazy-image jetpack-lazy-image--handled` indicating that it has been treated by Jetpack Lazy Image Loading.

To "fix" the issue in the Twenty Twenty-three theme, I added this little snippet as a custom plugin.

```
function my_customize_lazy_images( $blocked_classes ) {
    $blocked_classes[] = 'wp-post-image';
    return $blocked_classes;
}

add_filter( 'jetpack_lazy_images_blocked_classes', 'my_customize_lazy_images' );
```

This excludes all featured images from lazy-load. Better than before, but not ideal: also featured images inside a query loop will be excluded even if they are "below the fold".

So my request is actually two-fold:
1. Exclude featured images by default (ideally only the first 3 on each request, which would mean the top image on a single page or the top row in a query loop with three columns).
2. Exclude all child elements of an element with class `skip-lazy`, so that that class can be used on block level.

Thanks for considering :)

### LINKED ISSUE #31877: Lazy Images: Updating WooCommerce cart makes product images disappear
### Impacted plugin

Jetpack, Boost

### Quick summary

Editing the products in cart makes their thumbnails to no longer show after reload.

Tested with Twenty-Twenty Three and Storefront themes. In Chrome and Firefox.

https://github.com/Automattic/jetpack/assets/24195829/359d6dc3-12e4-4f19-9679-97dc39705b59

### Steps to reproduce

1. Create a new JN site with Jetpack and WooCommerce.
2. Connect Jetpack and enable **Enable Lazy Loading for images** at `/wp-admin/admin.php?page=jetpack#/performance`.
3. Import WooCommerce sample products.
4. Go to `/shop` add a few products to shopping cart.
5. Go to `/cart`. 
  a. Increasing and decreasing a product's quantity and clicking on `Update Cart` makes the product thumbnail disappear.
  b. Removing a product from cart by clicking on `x` makes the product thumbnail disappear.

Reloading the page should show the correct product thumbnails again. Disabling **Enable Lazy Loading for images** and retrying the above steps should make the product thumbnail display accordingly.

If a custom image is added to the cart section, i.e. an icon for increasing/decreasing the quantity (`woocommerce_before_quantity_input_field` hook), that image will also disappear.

Product thumbnails and other images in the cart section should not disappear when updating the cart.

### What actually happened

Removing a product or adding/subtracting its quantity makes the images in the cart section to not show anymore.

### Impact

Some (< 50%)

### Available workarounds?

Yes, easy to implement

_No response_

### Logs or notes

_No response_

### LINKED ISSUE #33171: Lazy Images: stop loading the package in the Jetpack plugin
Let's stop loading Lazy Images in Jetpack.

https://github.com/Automattic/jetpack/blob/fbb712b80810ce609127103cabda54bdd79d5a00/projects/plugins/jetpack/modules/lazy-images.php#L34

https://github.com/Automattic/jetpack/blob/8a9b0dd5ef83a9ed69d675f0d00e059f7bebbca9/projects/plugins/jetpack/composer.json#L38

******

Primary issue: #33049

jeherve:
Fixes #33171

## Proposed changes:

We deprecated the feature and issued notices to our users still using it in #33208. It is now time to remove it entirely.

- If the module is still enabled, we disable it.
- We do not display the feature in any of our UI elements anymore.
- We declare the file as deprecated, so we can safely remove it in a future version of the plugin.

### Other information:

Let's also take that opportunity to close related issues:

Fixes #10016
Fixes #9827
Fixes #24002
Fixes #9043
Fixes #11054
Fixes #8905
Fixes #17640
Fixes #15073
Fixes #16421
Fixes #8879
Fixes #9509
Fixes #9974
Fixes #9896
Fixes #8917
Fixes #10046
Fixes #8897
Fixes #9929
Fixes #9595
Fixes #10606
Fixes #12344
Fixes #10607
Fixes #25659
Fixes #29585
Fixes #14191
Fixes #13518
Fixes #10707
Fixes #8996
Fixes #8773
Fixes #8281
Fixes #14161
Fixes #31877
Fixes #14281

## Jetpack product discussion

* #33049

* Yes, it removes all Lazy Image links from the UI, and thus the attached tracking disappears with them.

## Testing instructions:

* You can start with a site running `trunk`, using the Lazy Images module: connect your site to WordPress.com, and in WP CLI, enable the module (you cannot enable it via the UI anymore): `wp jetpack module activate lazy-images`
* Switch to this branch.
    * You should not see any mentions of Lazy Images anymore.
    * No mention of the feature there either.
* Back to the CLI, run `wp jetpack module list`
    * No mention of the feature there either.

### DIFF
--- a/projects/plugins/jetpack/3rd-party/woocommerce.php
+++ b/projects/plugins/jetpack/3rd-party/woocommerce.php
@@ -134,29 +134,3 @@ function jetpack_woocommerce_infinite_scroll_style() {
 	}';
 	wp_add_inline_style( 'woocommerce-layout', $custom_css );
 }
-
-/**
- * Adds compat for WooCommerce and Lazy Loading.
- */
-function jetpack_woocommerce_lazy_images_compat() {
-	wp_add_inline_script(
-		'wc-cart-fragments',
-		"
-		jQuery( 'body' ).bind( 'wc_fragments_refreshed', function() {
-			var jetpackLazyImagesLoadEvent;
-			try {
-				jetpackLazyImagesLoadEvent = new Event( 'jetpack-lazy-images-load', {
-					bubbles: true,
-					cancelable: true
-				} );
-			} catch ( e ) {
-				jetpackLazyImagesLoadEvent = document.createEvent( 'Event' )
-				jetpackLazyImagesLoadEvent.initEvent( 'jetpack-lazy-images-load', true, true );
-			}
-			jQuery( 'body' ).get( 0 ).dispatchEvent( jetpackLazyImagesLoadEvent );
-		} );
-		"
-	);
-}
-
-add_action( 'wp_enqueue_scripts', 'jetpack_woocommerce_lazy_images_compat', 11 );
--- a/projects/plugins/jetpack/_inc/client/components/contextualized-connection/index.jsx
+++ b/projects/plugins/jetpack/_inc/client/components/contextualized-connection/index.jsx
@@ -109,12 +109,6 @@ const ContextualizedConnection = props => {
 								) }
 							</li>
 							<li>
-								{ createInterpolateElement(
-									__( '<strong>Free</strong> lazy image loading', 'jetpack' ),
-									{ strong: <strong /> }
-								) }
-							</li>
-							<li>
 								{ createInterpolateElement( __( '<strong>Free</strong> SEO tools', 'jetpack' ), {
 									strong: <strong />,
 								} ) }
--- a/projects/plugins/jetpack/_inc/client/performance/index.jsx
+++ b/projects/plugins/jetpack/_inc/client/performance/index.jsx
@@ -19,7 +19,7 @@ class Performance extends Component {
 			hasConnectedOwner: this.props.hasConnectedOwner,
 		};
 
-		const found = [ 'photon', 'videopress', 'lazy-images', 'photon-cdn', 'search' ].some(
+		const found = [ 'photon', 'videopress', 'photon-cdn', 'search' ].some(
 			this.props.isModuleFound
 		);
 
--- a/projects/plugins/jetpack/_inc/client/performance/speed-up-site.jsx
+++ b/projects/plugins/jetpack/_inc/client/performance/speed-up-site.jsx
@@ -1,11 +1,8 @@
 import { ToggleControl, getRedirectUrl } from '@automattic/jetpack-components';
-import { ExternalLink } from '@wordpress/components';
 import { __, sprintf } from '@wordpress/i18n';
-import classNames from 'classnames';
 import { FormFieldset } from 'components/forms';
 import { withModuleSettingsFormHelpers } from 'components/module-settings/with-module-settings-form-helpers';
 import { ModuleToggle } from 'components/module-toggle';
-import SimpleNotice from 'components/notice';
 import SettingsCard from 'components/settings-card';
 import SettingsGroup from 'components/settings-group';
 import analytics from 'lib/analytics';
@@ -148,50 +145,13 @@ const SpeedUpSite = withModuleSettingsFormHelpers(
 			}
 		};
 
-		trackDeprecatedLazyImagesLearnMore = () => {
-			analytics.tracks.recordJetpackClick( {
-				target: 'learn-more',
-				feature: 'lazy-images',
-				extra: 'deprecated-link',
-			} );
-		};
-
-		displayLazyImagesNotice = () => {
-			const disableStatusMessage = this.props.gutenbergInfo.hasInteractivityApi
-				? __( 'We’ve consequently disabled the Lazy Images feature on your site.', 'jetpack' )
-				: __( 'This feature will consequently be removed from Jetpack in November.', 'jetpack' );
-
-			return (
-				<SimpleNotice showDismiss={ false } status="is-info" className="jp-form-settings-notice">
-					{ __(
-						'Modern browsers now support lazy loading, and WordPress itself bundles lazy loading features for images and videos.',
-						'jetpack'
-					) }{ ' ' }
-					{ disableStatusMessage }{ ' ' }
-					{
-						<ExternalLink
-							href={ getRedirectUrl( 'jetpack-support-lazy-images' ) }
-							onClick={ this.trackDeprecatedLazyImagesLearnMore }
-						>
-							{ __( 'Learn more', 'jetpack' ) }
-						</ExternalLink>
-					}
-				</SimpleNotice>
-			);
-		};
-
 		render() {
 			const foundPhoton = this.props.isModuleFound( 'photon' );
 			const foundAssetCdn = this.props.isModuleFound( 'photon-cdn' );
-			const foundLazyImages = this.props.isModuleFound( 'lazy-images' );
 
-			if ( ! foundPhoton && ! foundLazyImages && ! foundAssetCdn ) {
+			if ( ! foundPhoton && ! foundAssetCdn ) {
 				return null;
 			}
-
-			const lazyImages = this.props.module( 'lazy-images' );
-			const isLazyimagesActive = this.props.getOptionValue( 'lazy-images' );
-			const isLazyimagesForceDisabled = this.props.gutenbergInfo.hasInteractivityApi;
 
 			// Check if any of the CDN options are on.
 			const siteAcceleratorStatus =
@@ -322,33 +282,6 @@ const SpeedUpSite = withModuleSettingsFormHelpers(
 									</ModuleToggle>
 								) }
 							</FormFieldset>
-						</SettingsGroup>
-					) }
-
-					{ foundLazyImages && (
-						<SettingsGroup
-							hasChild
-							module={ lazyImages }
-							className={ classNames( 'jp-form-settings-group--lazy_images', {
-								'jp-form-settings-group--lazy_images_disabled': ! isLazyimagesActive,
-							} ) }
-						>
-							{ this.displayLazyImagesNotice() }
-							<ModuleToggle
-								slug="lazy-images"
-								disabled={
-									this.props.isUnavailableInOfflineMode( 'lazy-images' ) ||
-									! isLazyimagesActive ||
-									isLazyimagesForceDisabled
-								}
-								activated={ isLazyimagesActive && ! isLazyimagesForceDisabled }
-								toggling={ this.props.isSavingAnyOption( 'lazy-images' ) }
-								toggleModule={ this.toggleModule }
-							>
-								<span className="jp-form-toggle-explanation">
-									{ __( 'Enable Lazy Loading for images', 'jetpack' ) }
-								</span>
-							</ModuleToggle>
 						</SettingsGroup>
 					) }
 				</SettingsCard>
--- a/projects/plugins/jetpack/_inc/client/performance/style.scss
+++ b/projects/plugins/jetpack/_inc/client/performance/style.scss
@@ -22,19 +22,3 @@
 		}
 	}
 }
-
-/* Lazy Images */
-.jp-form-settings-group--lazy_images {
-	.jp-form-settings-notice {
-		a.components-external-link {
-			color: $white;
-		}
-	}
-	&.jp-form-settings-group--lazy_images_disabled {
-		.jp-form-toggle-explanation {
-			color: $gray-5;
-		}
-	}
-}
-
-
--- a/projects/plugins/jetpack/_inc/client/state/modules/reducer.js
+++ b/projects/plugins/jetpack/_inc/client/state/modules/reducer.js
@@ -269,14 +269,13 @@ export function hasAnyOfTheseModules( state, modules = [] ) {
 /**
  * Check that the site has any of the performance features available.
  *
- * @param  {Object} state   Global state tree
+ * @param {object} state   - Global state tree
  *
- * @return {boolean}        True if at least one of the performance features is available
+ * @returns {boolean}        True if at least one of the performance features is available
  */
 export function hasAnyPerformanceFeature( state ) {
 	return hasAnyOfTheseModules( state, [
 		'carousel',
-		'lazy-images',
 		'photon',
 		'photon-cdn',
 		'search',
--- a/projects/plugins/jetpack/_inc/client/state/modules/test/selectors.js
+++ b/projects/plugins/jetpack/_inc/client/state/modules/test/selectors.js
@@ -166,7 +166,7 @@ describe( 'items selectors', () => {
 				jetpack: {
 					modules: {
 						items: {
-							'lazy-images': {},
+							photon: {},
 						},
 					},
 				},
--- a/projects/plugins/jetpack/changelog/rm-lazy-images-jetpack
+++ b/projects/plugins/jetpack/changelog/rm-lazy-images-jetpack
@@ -1,0 +1,4 @@
+Significance: patch
+Type: compat
+
+Lazy Images: remove the feature from the plugin. You can now rely on WordPress' own Lazy Image features on your site.
--- a/projects/plugins/jetpack/changelog/rm-lazy-images-jetpack#2
+++ b/projects/plugins/jetpack/changelog/rm-lazy-images-jetpack#2
@@ -1,0 +1,5 @@
+Significance: patch
+Type: other
+Comment: Updated composer.lock.
+
+
--- a/projects/plugins/jetpack/class.jetpack-modules-list-table.php
+++ b/projects/plugins/jetpack/class.jetpack-modules-list-table.php
@@ -116,8 +116,7 @@ class Jetpack_Modules_List_Table extends WP_List_Table {
 			<# var i = 0;
 			if ( data.items.length ) {
 			_.each( data.items, function( item, key, list ) {
-				if ( item === undefined ) return;
-				if ( 'lazy-images' == item.module && ! item.activated ) return; #>
+				if ( item === undefined ) return; #>
 				<tr class="jetpack-module <# if ( ++i % 2 ) { #> alternate<# } #><# if ( item.activated ) { #> active<# } #><# if ( ! item.available ) { #> unavailable<# } #>" id="{{{ item.module }}}">
 					<th scope="row" class="check-column">
 						<input type="checkbox" name="modules[]" value="{{{ item.module }}}" {{{ item.disabled }}} />
--- a/projects/plugins/jetpack/class.jetpack.php
+++ b/projects/plugins/jetpack/class.jetpack.php
@@ -300,11 +300,6 @@ class Jetpack {
 			'Rank Math'                            => 'seo-by-rank-math/rank-math.php',
 			'Slim SEO'                             => 'slim-seo/slim-seo.php',
 		),
-		'lazy-images'        => array(
-			'Lazy Load'              => 'lazy-load/lazy-load.php',
-			'BJ Lazy Load'           => 'bj-lazy-load/bj-lazy-load.php',
-			'Lazy Load by WP Rocket' => 'rocket-lazy-load/rocket-lazy-load.php',
-		),
 	);
 
 	/**
@@ -5976,11 +5971,6 @@ endif;
 			'jetpack_cache_plans'                          => array(
 				'replacement' => null,
 				'version'     => 'jetpack-6.1.0',
-			),
-
-			'jetpack_lazy_images_skip_image_with_atttributes' => array(
-				'replacement' => 'jetpack_lazy_images_skip_image_with_attributes',
-				'version'     => 'jetpack-6.5.0',
 			),
 			'jetpack_enable_site_verification'             => array(
 				'replacement' => null,
--- a/projects/plugins/jetpack/composer.json
+++ b/projects/plugins/jetpack/composer.json
@@ -34,7 +34,6 @@
 		"automattic/jetpack-import": "@dev",
 		"automattic/jetpack-ip": "@dev",
 		"automattic/jetpack-jitm": "@dev",
-		"automattic/jetpack-lazy-images": "@dev",
 		"automattic/jetpack-licensing": "@dev",
 		"automattic/jetpack-logo": "@dev",
 		"automattic/jetpack-my-jetpack": "@dev",
--- a/projects/plugins/jetpack/composer.lock
+++ b/projects/plugins/jetpack/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "efbfa6f913dfce5c8a13cf87a6e4cf99",
+    "content-hash": "d4543bd5fa21fe0b226972d98ea56739",
     "packages": [
         {
             "name": "automattic/jetpack-a8c-mc-stats",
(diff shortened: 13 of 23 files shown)


anomiex (approved):
Works as advertised.

I did a search looking for other references to lazy image loading. I suppose these things need to wait until it's removed from Boost too?

* Image CDN package support for our lazy-images feature:
  * https://github.com/Automattic/jetpack/blob/ce0c812108525b89c6e323f16d27227a68135ce4/projects/packages/image-cdn/src/class-image-cdn.php#L379-L391
  * https://github.com/Automattic/jetpack/blob/ce0c812108525b89c6e323f16d27227a68135ce4/projects/packages/image-cdn/src/class-image-cdn.php#L572-L581
  * https://github.com/Automattic/jetpack/blob/ce0c812108525b89c6e323f16d27227a68135ce4/projects/packages/image-cdn/src/js/image-cdn.js#L8-L14
* Adding "[abandoned](https://getcomposer.org/doc/04-schema.md#abandoned)" to `projects/packages/lazy-images/composer.json`, and an appropriate notice in the readme.
  * Then after publishing a final version we can delete it from the monorepo and archive https://github.com/Automattic/jetpack-lazy-images
* Boost-related reference in my-jetpack: https://github.com/Automattic/jetpack/blob/ce0c812108525b89c6e323f16d27227a68135ce4/projects/packages/my-jetpack/src/products/class-boost.php#L180-L196
* Boost-related reference at https://github.com/Automattic/jetpack/blob/ce0c812108525b89c6e323f16d27227a68135ce4/projects/plugins/jetpack/_inc/client/recommendations/feature-utils.js#L302

Also I wonder whether https://github.com/Automattic/jetpack/blob/rm/lazy-images-jetpack/projects/packages/search/src/instant-search/components/photon-image.jsx, which looks like it does its own lazy loading, should be changed to just use `loading="lazy"` too.

jeherve:
> I suppose these things need to wait until it's removed from Boost too?

Yes, once we do not use the Lazy Images package anywhere anymore, we'll be able to completely remove everything and properly deprecate the package. This is tracked in the primary issue: #33048. 

> Also I wonder whether [rm/lazy-images-jetpack/projects/packages/search/src/instant-search/components/photon-image.jsx](https://github.com/Automattic/jetpack/blob/rm/lazy-images-jetpack/projects/packages/search/src/instant-search/components/photon-image.jsx?rgh-link-date=2023-10-25T20%3A08%3A39Z), which looks like it does its own lazy loading, should be changed to just use loading="lazy" too.

That's a good point. @jsnmoon Looks like you worked on #19269 back in the day. Do you think we could remove this in favor of the loading attribute now?

jsnmoon:
> That's a good point. @jsnmoon Looks like you worked on #19269 back in the day. Do you think we could remove this in favor of the loading attribute now?

Yup, that sounds good!

PR #33783, opened 2023-10-25
docker: Use mpm_event and php-fpm instead of mpm_prefork and mod_php

anomiex:
## Proposed changes:

The main hope here is that using mpm_event rather than mpm_prefork will make our E2E tests run better.

But mod_php is not compatible with mpm_event. The modern way to do things is to use php-fpm instead.

And while we're at it, we may as well enable apache's http2 module too. Although actually making effective use of that may take more work (if anyone cares), since HTTP/2 likes to run with TLS but our dev environment doesn't do TLS.

### Other information:

## Jetpack product discussion

pc9hqz-2f3-p2#comment-1825

No

## Testing instructions:

* Temporarily change https://github.com/Automattic/jetpack/blob/9e6f6dc25b4fc2c522a3062e19826678c1e4fdb3/tools/docker/docker-compose.yml#L37 to refer to `ghcr.io/automattic/jetpack-wordpress-dev:pr-33783` and restart your docker env (`jetpack docker stop && jetpack docker up -d`). Then make sure everything still works.
* If you can run the E2E tests locally and reproduce the kinds of problems mentioned in pc9hqz-2f3-p2, see if this helps.

### DIFF
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -27,7 +27,7 @@ RUN --mount=type=cache,target=/var/lib/apt/lists/,sharing=private \
 	&& add-apt-repository ppa:ondrej/php \
 	&& curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg \
 	&& apt-get update \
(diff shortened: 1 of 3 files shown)


zinigor (approved):
Works well for me!

PR #33786, opened 2023-10-25
Paywall: Add Subscribers auth endpoint

lezama:
## Proposed changes:

* Adds an new endpoint `https://example.com/wp-json/jetpack/v4/subscribers/auth` that receives a subscriber token.
* The endpoint verifies the token is valid, stores it as cookie(`jp-premium-content-session`) and then redirects to the `redirect_url` parameter inside the token.

### Other information:

## Testing instructions:

* Apply D126644-code to a wpcom sandbox – It adds the `redirect_url` to the token.
* Sandbox subscribe.wordpress.com
* You will need an email that's already subscribed to your Jetpack site.
* Log in to WPCOM with that user.
* Go to a subscribers only post in your Jetpack site.
* Click "Log in" in the paywall.
* You should land in the post but this time without the paywall and no token in the url.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -1082,8 +1082,6 @@ function get_paywall_blocks( $newsletter_access_level ) {
 		if ( ( new Host() )->is_wpcom_simple() ) {
 			// custom domain
 			$sign_in_link = wpcom_logmein_redirect_url( get_current_url(), false, null, 'link' );
-		} else {
-			$sign_in_link = 'https://subscribe.wordpress.com/memberships/jwt?site_id=' . \Jetpack_Options::get_option( 'id' ) . '&redirect_url=' . get_current_url();
 		}
 	}
 
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -1069,13 +1069,11 @@ function get_paywall_blocks( $newsletter_access_level ) {
 		// translators: %s is the name of the site.
 		: esc_html__( 'Subscribe to get access to the rest of this post and other subscriber-only content.', 'jetpack' );
 
+	$button_text     = esc_html__( 'Log in', 'jetpack' );
+	$sign_in_link    = 'https://subscribe.wordpress.com/memberships/jwt?site_id=' . \Jetpack_Options::get_option( 'id' ) . '&redirect_url=' . get_current_url();
 	$access_question = $is_paid_post
 		? __( 'Already a paid subscriber?', 'jetpack' )
 		: __( 'Already a subscriber?', 'jetpack' );
-
-	$button_text     = esc_html__( 'Log in', 'jetpack' );
-	$sign_in_link    = 'https://subscribe.wordpress.com/memberships/jwt?site_id=' . \Jetpack_Options::get_option( 'id' ) . '&redirect_url=' . get_current_url();
-	$access_question = __( 'I am already subscribed.', 'jetpack' );
 
 	if ( is_user_auth() ) {
 		$button_text = esc_html__( 'Switch accounts', 'jetpack' );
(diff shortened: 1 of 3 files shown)


millerf:
I like the endpoint approach. Very lean.

A few comments:

- Not sure the endpoint should live in `class.core-rest-api-endpoints.php`. Maybe we could create `class.memberships-rest-api-endpoints.php` for it/them (I am planning to add a fewmore)?
- I don't get the 
```
if ( isset( $_GET['v2'] ) ) {
```
- On WPCOM simple sites without custom domains, we don't need the endpoint, we can set the cookie directly if headers are not sent
- I am not sure if this one is included (got lost on so many changes lately: but we should use D126644-code to also redirect the email subscription confirmation (this whole code needs so much cleanup)
- I would love this one as a standalone without the signin links changes. 
- Why do you add the redirection in the token?! It needs to be passed as a parameter. Also we could remove the domain to prevent weird redirection attacks
- Let's add a throttle protection on the endpoint
- There is so much room for refactor and clean (`get_and_set_token_from_request` can pretty much be removed everywhere else, the whole Auth endpoint is very convoluted)

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -1069,34 +1069,23 @@ function get_paywall_blocks( $newsletter_access_level ) {
 		// translators: %s is the name of the site.
 		: esc_html__( 'Subscribe to get access to the rest of this post and other subscriber-only content.', 'jetpack' );
 
+	$button_text     = esc_html__( 'Log in', 'jetpack' );
+	$sign_in_link    = 'https://subscribe.wordpress.com/memberships/jwt?site_id=' . \Jetpack_Options::get_option( 'id' ) . '&redirect_url=' . get_current_url() . '&v2';
```
millerf, line 1073: I see the v2 now

lezama: > I don't get the
if ( isset( $_GET['v2'] ) ) {

We need to distinguish if the jetpack site using `subscribe.wordpress.com/memberships/jwt` has the `wp-json/jetpack/v4/subscribers/auth` endpoint

lezama:
> Not sure the endpoint should live in class.core-rest-api-endpoints.php. Maybe we could create class.memberships-rest-api-endpoints.php for it/them (I am planning to add a fewmore)?

My first instinct was to create another file too but couldn't quickly find any other files registering `jetpack/v4/*` endpoints, seems like they are all centralized in this file 🤔  . @jeherve am I missing something?

> On WPCOM simple sites without custom domains, we don't need the endpoint, we can set the cookie directly if headers are not sent

Oh, do they still get a token? will change that in the WPCOM Phab

> I would love this one as a standalone without the signin links changes.

The parent branch is the one with the signin links changes, just for testing my use case, we can rebase against trunk or merge the other PR first

> Why do you add the redirection in the token?! It needs to be passed as a parameter. Also we could remove the domain to prevent weird redirection attacks

There's no way to trick the JWT, the check for a safe redirect is already done WPCOM side. This way we don't need to do the check again in Jetpack side.

> Let's add a throttle protection on the endpoint

Do we need it? There's no way for an attacker to build a valid token to a different redirect.

> I like the endpoint approach. Very lean.
> There is so much room for refactor and clean (get_and_set_token_from_request can pretty much be removed everywhere else, the whole Auth endpoint is very convoluted)

Yes! I would love to see that (not on this PR of course)

millerf:
> they are all centralized in this file 🤔 . @jeherve am I missing something?

I reused that code in  https://github.com/Automattic/jetpack/pull/33800, and I think the place I found is a bit better (see https://github.com/Automattic/jetpack/pull/33800/files#r1372991408)

> There's no way to trick the JWT, the check for a safe redirect is already done WPCOM side. This way we don't need to do the check again in Jetpack side.

🤔 I am not sure I follow. But it is definitely not a good thing to add it in the token...

> There's no way for an attacker to build a valid token to a different redirect.

I was more thinking of a direct attack on that endpoint with the same token?

lezama:
> 🤔 I am not sure I follow. But it is definitely not a good thing to add it in the token...

could you explain why?

> I was more thinking of a direct attack on that endpoint with the same token?

what would be the attack?

n3f (changes requested):
I'm not sure what happened but when I used a WPCOM simple site with a custom domain the login didn't work. That is the login "worked", but after the redirect from WordPress.com I was presented with the "Subscribe to continue reading" view.

I'm not sure if this is a 3rd party cookie problem or not (I was using a chrome incognito window).

It did work with a regular WPCOM simple site (using a wordpress.com domain).

lezama:
> I'm not sure what happened but when I used a WPCOM simple site with a custom domain the login didn't work. That is the login "worked", but after the redirect from WordPress.com I was presented with the "Subscribe to continue reading" view.

yes! I need to use `wpcom_logmein_redirect_url` but makes the auth test fail in wpcom, I need to figure out how to fix those.

lezama:
Merging https://github.com/Automattic/jetpack/pull/33815 that doesn't have the log in part

PR #33788, opened 2023-10-25
Prevent adding tier to payload when not provided

millerf:
Fixes Adding plan from paid-content block.

## Proposed changes:

* Fix a bug found by @chrismccluskey 
* Do not add "tier" to payload as `null` when not provided

### Other information:

## Testing instructions:

* On a Jn instance
* Set Jetpack
* Create a plan from a premium-content and make sure it gets added
* It would return a 500 before


chrismccluskey (approved):
LGTM and fixes the issue!

PR #33789, opened 2023-10-25
CRM: 3374 - allow events API endpoint to be filtered by owner

tbradsha:
Resolves Automattic/zero-bs-crm#3374 - allow events API endpoint to be filtered by owner

## Proposed changes:

In #33221, I inadvertently removed the ability to filter tasks returned by the `events` endpoint by owner. This PR restores that ability.

Note: prior to removal, the param used was `owned`. The [API docs](https://automattic.github.io/jetpack-crm-api-docs/#ownership-assignment) are ambiguous (may be `assign`, `owned`, or `owner`), and the Events endpoint example doesn't give the request example but uses `owner` in the response, so I allow for both `owned` and `owner`.

### Other information:

## Testing instructions:

1. Create a task assigned to user 1.
2. Create a task assigned to user 2.
3. Enable the API if not yet enabled.
4. Call the API `events` endpoint with no ownership param, with `owned`, and with `owner`.

In `trunk`, all tasks will be returned no matter the ownership param/value.

In the `fix/crm/3374-allow_events_API_filter_by_owner` branch, the following occurs:

* no param: all tasks
* `-1`: all tasks
* random string: all tasks
* WP ID: tasks assigned to WP user
* WP ID with no tasks assigned: no tasks

### DIFF
--- a/projects/plugins/crm/api/events.php
+++ b/projects/plugins/crm/api/events.php
@@ -28,7 +28,19 @@ jpcrm_api_check_http_method( array( 'GET' ) );
 // Process the pagination parameters from the query
 list( $page, $per_page ) = jpcrm_api_process_pagination();
 
-$owner = -1;
+/**
+ * Allow events to be filtered by owner. Docs are ambiguous about
+ * whether we should use `owned` or `owner`, so let's support both.
+ */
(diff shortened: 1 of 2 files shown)


gogdzl (approved):
LGTM!

PR #33800, opened 2023-10-26
Access modal mixed approach

millerf:
Works with D126720-code

Task list:
- [x] Do not use WPCOM page to generate the magic link 
- [x] Fix issue with multiple subscribe buttons
- [x] Replace "code" screen  with an "email" input (if not provided) 
- [x] Send a magic link in the background from WPCOM side (sent but needs redirect to WPCOM
- [ ] (WPCOM repo) Add an endpoint to redirect with token to the post/page from the redirect
- [ ] (Jetpack repo) Add an endpoint (or reuse the existing one) that set the new token in session/cookie on instance backend and closes the tab.
- [ ] (Jetpack ) Add an endpoint that accepts only calls from the front-end and accepts a "token" parameter and checks it against the backend one `$_COOKIE['token']`
- [ ] (Jetpack js) Create a JS polling routine that checks for token changes, and reload the page if token is changed.

## Proposed changes:

* See: p81Rsd-1vQ-p2#comment-1620
* Keep user in-site when getting subscriptions access
* streamline login process without opening new tab
* Can potentially be segmented for WPCOM users (using regular WPCOM email-less login)  vs self-hosted users 

## Testing instructions:

* Apply D126720-code to your sandbox
* For Simple sites:
  * Apply this PR on top (if working with Simple site)
* For Jetpack
  * Create a JN instance and link to your sandbox
  * Link Jetpack
* Setup Monetize and create tier plans
* Create a post with a paywall plan
* Click "Login"/"Get Access"
You should see something like this:

Uploading Screen Recording 2023-10-26 at 12.28.41.mov…

* Clicking on the email link open and close a new tab quickly, and reload the browsed page


projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/memberships.php
```diff
@@ -8,6 +8,14 @@
 
 use Automattic\Jetpack\Connection\Client;
 
+if ( ( defined( 'IS_WPCOM' ) && IS_WPCOM ) || Jetpack::is_connection_ready() ) {
+	// We load the necessary endpoints
+	require_once JETPACK__PLUGIN_DIR . '_inc/lib/class.memberships-rest-api-endpoints.php';
+
+	wpcom_rest_api_v2_load_plugin( 'WPCOM_REST_API_V2_Endpoint_Memberships' );
+	add_action( 'rest_api_init', array( 'Jetpack_Memberships_Json_Api_Endpoints', 'register_endpoints' ) );
```
millerf, line 16: Here @lezama ( being a couple of flags). I haven't been able to test it, though

lezama: for Jetpack only endpoints it seems it is better to register them where I am proposing in the other PR: https://github.com/Automattic/jetpack/pull/33786

See: p1698339772009019-slack-CBG1CP4EN

if we do it here we would register the endpoints for simple sites too and I think it's not needed?

projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
```diff
@@ -384,7 +384,8 @@ private function set_token_cookie( $token ) {
 		}
 
 		if ( ! empty( $token ) && false === headers_sent() ) {
-			setcookie( self::JWT_AUTH_TOKEN_COOKIE_NAME, $token, 0, '/', COOKIE_DOMAIN, is_ssl(), true ); // httponly -- used by visitor_can_view_content() within the PHP context.
+			// Set the token to expire in one week to prevent ever-lasting subscriptions (if cancelled) and to force user to reload
```
lezama, line 418: I think we should set the expire date at the token level instead of here, so if the token is too old we can require to log-in to get a new one.

lezama:
closing in favor of https://github.com/Automattic/jetpack/pull/33867

PR #33806, opened 2023-10-26
CRM: 3380 - bump minimum PHP to 7.4

tbradsha:
Resolves Automattic/zero-bs-crm#3380 - bump minimum PHP to 7.4

## Proposed changes:

See pbhBOz-3Yp-p2; we've decided to change the minimum PHP version to 7.4.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

Previous reference: #33003

PR #33810, opened 2023-10-26
CRM: Drop deps on old codeception versions

anomiex:
## Proposed changes:

Now that the minimum is PHP 7.4 (see #33806), we can drop some older versions from the dependencies.

### Other information:

## Jetpack product discussion

Follows up #33806

No

## Testing instructions:

* Is CI happy?


tbradsha (approved):
Confirmed dependency updates are valid, and tests run fine.

PR #33813, opened 2023-10-26
My Plan: update link to licensing management page

### LINKED ISSUE #33803: My Plan: Reposition actions at the top
### Problem

Reported in Slack by @simison: p1698302590416139/1698302544.355569-slack-C0D96691V

>“activate product” on the right-most extreme side when the CTA text is on the left side.
>Could both be right right or left aligned?

Image

This could even be more jarring when a paid product exists and an additional “View all purchases” button is displayed:

Image

### Proposal

- Get rid of the primary CTA “Activate a product”
- Use the copy “Activate it here” for that action instead.

Image

jeherve:
Fixes #33803

## Proposed changes:

Let's update the link to the licensing management page according to the discussion and screenshots in #33803:

> Get rid of the primary CTA “Activate a product”
> Use the copy “Activate it here” for that action instead.

> **Note**
> @keoshi I did not make any changes to the "Recommendations" link at the bottom of the header.

This also fixes some spaces <> tabs mixes in the stylesheet for that page.

**Before**

image

**After**

image

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

* Open your browser's JavaScript console.
* Enable debugging to see Tracks events by pasting the following: `localStorage.setItem( 'debug', 'dops:analytics*' );`
* Check the header of the page according to the screenshot above.
* Click on the "Activate it here" link.
    * The link should direct to the right page.
    * You should see a tracks event triggered in your JS console.

### DIFF
--- a/projects/plugins/jetpack/_inc/client/my-plan/style.scss
+++ b/projects/plugins/jetpack/_inc/client/my-plan/style.scss
@@ -3,63 +3,64 @@
 @import '../scss/typography';
 
 .jp-landing__plans {
-  	margin-top: 48px;
+	margin-top: 48px;
 	margin-bottom: 24px;
-  	box-shadow: 0 0 40px rgba(0, 0, 0, 0.08);
+	box-shadow: 0 0 40px rgba(0, 0, 0, 0.08);
 
 	.dops-button {
 		margin-right: 10px;
 	}
 
 	.dops-card.is-compact {
-	  margin-bottom: 0;
-	  border-bottom: 1px solid var(--jp-gray);
+		margin-bottom: 0;
+		border-bottom: 1px solid var(--jp-gray);
 
-	  	@include breakpoint( "<960px" ) {
-		border-bottom: 0;
-	  	}
+		@include breakpoint( "<960px" ) {
+			border-bottom: 0;
+		}
 	}
 
-  	.dops-card:first-child {
-	  border-radius: 4px 4px 0 0;
+	.dops-card:first-child {
+		border-radius: 4px 4px 0 0;
 	}
 
-  	.dops-card:last-child {
-	  border-radius: 0 0 4px 4px;
+	.dops-card:last-child {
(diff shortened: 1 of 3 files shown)


projects/plugins/jetpack/_inc/client/my-plan/style.scss
```diff
@@ -3,63 +3,64 @@
...
+		@include breakpoint( "<960px" ) {
+			border-bottom: 0;
+		}
 	}
 
-  	.my-plan-card__tag-line a {
-	  color: $black;
-	  text-decoration-line: underline;
+	.my-plan-card__tag-line a,
+	a.jp-landing__licensing-actions-link {
```
jeherve, line 42: This is the only real change in the file. The rest is just switching from spaces to tabs.

andrii-lysenko (commented):
This does seem reasonable to me and code looks good.

Maybe we could also iterate a bit on style, as it's a bit hard underline there. What do you think?
 
Screenshot at Oct 26 20-24-03

jeherve:
> Maybe we could also iterate a bit on style, as it's a bit hard underline there. What do you think?

I like it, personally, it makes it clear when you move hover a link. I didn't come up with those styles though, this is what we use elsewhere in the dashboard:

https://github.com/Automattic/jetpack/assets/426388/b3d3b620-0b05-4f45-a45b-afd27641879f

But I'm happy to update them if that's necessary! Maybe we can do that in a follow-up PR?

@keoshi what's your take on this?

keoshi:
> @keoshi what's your take on this?

It's a bit more opinionated than most link styles, but having the same color link as the text on the page warrants it IMO. It's also an intermittent state (on hover) so only a single element will have this styling on a page at any given moment.

If we feel this needs to be changed then we need to have a much broader discussion, since it affects all plugin pages and all plugins too.

keoshi (approved):
Change looks good to me! Thanks for doing this so quickly, Jeremy.

PR #33816, opened 2023-10-26
custom-css: PHP Rector updates (lessc)

mreishus:
## Proposed changes:

* Updated php rector issues for `custom-css/custom-css.php` and `custom-css/custom-css/preprocessors/lessc.inc.php` for our php8 migration rector repo ( see: 32481-pb/#js ) 

### Other information:

no

## Testing instructions:

* Copy to sandbox
```
scp ./projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php wpcom-sandbox:public_html/./wp-content/mu-plugins/jetpack-plugin/moon/modules/custom-css/custom-css/preprocessors/lessc.inc.php
scp ./projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php wpcom-sandbox:public_html/./wp-content/mu-plugins/jetpack-plugin/sun/modules/custom-css/custom-css/preprocessors/lessc.inc.php
scp ./projects/plugins/jetpack/modules/custom-css/custom-css.php wpcom-sandbox:public_html/./wp-content/mu-plugins/jetpack-plugin/sun/modules/custom-css/custom-css.php
scp ./projects/plugins/jetpack/modules/custom-css/custom-css.php wpcom-sandbox:public_html/./wp-content/mu-plugins/jetpack-plugin/moon/modules/custom-css/custom-css.php
```

(maybe the bot commands in the comment below are easier)

* have a simple site with premium plan or better
* in the customizer, go to the additional css panel. Add this css and change the dropdown to use the less preprocessor

```
// Variables
@primary-bg-color: #f4f4f4;
@secondary-bg-color: #4CAF50;
@text-color: #333;

// Mixin for text styling
.text-styling(@size, @color: @text-color) {
  font-size: @size;
  color: @color;
}

// Global styles to check if LESS is working
body {
  background-color: @primary-bg-color;
  .text-styling(56px);
}

h1 {
  .text-styling(36px, #ff0000);
}

// Nested Rules
.container {
  width: 100%;
  padding: 15px;
  background-color: @secondary-bg-color;
  .text-styling(18px, white);

  .box {
    width: 50%;
    margin: 10px auto;
    padding: 10px;
    background-color: white;
    border: 1px solid @secondary-bg-color;

    h2 {
      .text-styling(24px, @secondary-bg-color);
    }
  }
}

``` 

Save and the front-end of the site should now have a gray background with large text

### Note

There's one rector fix I didn't make:

The file treats `$importDir` strangle by initially assigning it to '' but always casting it to an array.. It seems safe to change to `[]`, but I wasn't 100% sure so I didn't.. should we?

### DIFF
--- a/projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
+++ b/projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
@@ -1696,7 +1696,7 @@ class lessc {
 	protected function op_color_number($op, $lft, $rgt) {
 		if ($rgt[0] == '%') $rgt[1] /= 100;
 
-		$lftCount = is_countable( $lft ) ? count( $lft ) : 0;
+		$lftCount = is_countable( $lft ) ? count( $lft ) : 1;
 		return $this->op_color_color($op, $lft,
 			array_fill(1, $lftCount - 1, $rgt[1]));
 	}
--- a/projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
+++ b/projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
@@ -1290,7 +1290,7 @@ class lessc {
 		} else {
 			if ($value[0] !== "list" || $value[1] != ",") $this->throwError("expecting list");
 			$values = $value[2];
-			$numValues = is_countable( $value ) ? count( $values ) : 0;
+			$numValues = is_countable( $values ) ? count( $values ) : 0;
 			if ($expectedArgs != $numValues) {
 				if ($name) {
 					$name = $name . ": ";
--- a/projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
+++ b/projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
@@ -2414,8 +2414,6 @@ class lessc_parser {
 		if (empty($this->buffer)) return false;
 		$s = $this->seek();
 
-		$block = null; // Initialize for wpcom's rector.
-
 		if ($this->whitespace()) {
 			return true;
 		}
@@ -2515,6 +2513,7 @@ class lessc_parser {
 			} catch (exception $e) {
 				$this->seek($s);
 				$this->throwError($e->getMessage());
+				$block = null; // Rector.
 			}
 
 			$hidden = false;
--- a/projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
+++ b/projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
@@ -3457,7 +3457,7 @@ class lessc_parser {
 	}
 
 	protected function genericList(&$out, $parseItem, $delim="", $flatten=true) {
-		$value = null;
+		$value = null; // Initialize output parameter to make wpcom's rector happy.
 		$s = $this->seek();
 		$items = array();
 		while ($this->$parseItem($value)) {
--- a/projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
+++ b/projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
@@ -2415,7 +2415,7 @@ class lessc_parser {
 		if (empty($this->buffer)) return false;
 		$s = $this->seek();
 
-		$block = null;
+		$block = null; // Initialize for wpcom's rector.
 
 		if ($this->whitespace()) {
 			return true;
@@ -3022,9 +3022,9 @@ class lessc_parser {
 	// arguments are separated by , unless a ; is in the list, then ; is the
 	// delimiter.
 	protected function argumentDef(&$args, &$isVararg) {
-		$value  = null;
-		$rhs    = null;
-		$newArg = null;
+		$value  = null; // Initialize output variable to make wpcom's rector happy.
+		$rhs    = null; // Initialize output variable to make wpcom's rector happy.
+		$newArg = null; // Initialize variable for rector.
 		$s = $this->seek();
 		if (!$this->literal('(')) return false;
 
--- a/projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
+++ b/projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
@@ -1304,7 +1304,6 @@ class lessc {
 	}
 
 	protected function toHSL($color) {
-		$H = null;
 		if ($color[0] == 'hsl') return $color;
 
 		$r = $color[1] / 255;
@@ -1325,7 +1324,7 @@ class lessc {
 
 			if ($r == $max) $H = ($g - $b)/($max - $min);
 			elseif ($g == $max) $H = 2.0 + ($b - $r)/($max - $min);
-			elseif ($b == $max) $H = 4.0 + ($r - $g)/($max - $min);
+			else $H = 4.0 + ($r - $g)/($max - $min);
 
 		}
 
--- a/projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
+++ b/projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
@@ -1,6 +1,6 @@
 <?php // phpcs:disable -- Copy-pasted library. Not reformatting to facilitate identification of actual changes.
 /**
- * lessphp v0.5.0
+ * lessphp v0.5.0 + php 8 compatibility changes by a8c (project is unmaintained)
  * https://leafo.net/lessphp
  *
  * LESS CSS compiler, adapted from http://lesscss.org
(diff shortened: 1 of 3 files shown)


projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
```diff
@@ -234,7 +234,7 @@ protected function compileMedia($media) {
 
 		$this->compileProps($media, $this->scope);
 
-		if (count($this->scope->lines) > 0) {
+		if ( is_countable( $this->scope->lines ) && count( $this->scope->lines ) > 0 ) {
```
anomiex, line 237: 🤷 The call in line 232 will always return an object with it set to an array, and the call on line 235 isn't going to change it. But I suppose we can't expect rector to be smart enough to see that.

projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
```diff
@@ -570,7 +570,8 @@ protected function patternMatch($block, $orderedArgs, $keywordArgs) {
 		} else {
 			$numMatched = $i + 1;
 			// greater than becuase default values always match
-			return $numMatched >= count($orderedArgs);
+			$orderedArgsCount = is_countable( $orderedArgs ) ? count( $orderedArgs ) : 0;
```
anomiex, line 573: Looks like the only code path getting here is via line 716, which will always pass an array. But, again, I guess we can't expect rector to understand that.

projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
```diff
@@ -1287,7 +1290,7 @@ public function assertArgs($value, $expectedArgs, $name="") {
 		} else {
 			if ($value[0] !== "list" || $value[1] != ",") $this->throwError("expecting list");
 			$values = $value[2];
-			$numValues = count($values);
+			$numValues = is_countable( $value ) ? count( $values ) : 0;
```
anomiex: You're testing the wrong variable here.
```suggestion
			$numValues = is_countable( $values ) ? count( $values ) : 0;
```

projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
```diff
@@ -1301,6 +1304,7 @@ public function assertArgs($value, $expectedArgs, $name="") {
 	}
 
 	protected function toHSL($color) {
+		$H = null;
```
anomiex: Really the thing to do would be to change the `elseif ($b == $max)` on line 1328 to just `else`. After doing `$max = max( $r, $g, $b )`, if the result is not equal to `$r` or `$g` then either it's equal to `$b`, or someone snuck a NaN in, or there's some really weird floating point error going on.

projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
```diff
@@ -1688,13 +1696,16 @@ protected function op_number_color($op, $lft, $rgt) {
 	protected function op_color_number($op, $lft, $rgt) {
 		if ($rgt[0] == '%') $rgt[1] /= 100;
 
+		$lftCount = is_countable( $lft ) ? count( $lft ) : 0;
```
anomiex: This seems like a case where more closely following the original semantics of `count()` would be better. Less likely to accidentally be passing -1 to `array_fill()`, which doesn't accept negative numbers.
```suggestion
		$lftCount = is_countable( $lft ) ? count( $lft ) : 1;
```
Although really I'd be surprised if this (or most of these really) ever is getting a non-array.

projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
```diff
@@ -2404,6 +2415,8 @@ protected function parseChunk() {
 		if (empty($this->buffer)) return false;
 		$s = $this->seek();
 
+		$block = null;
```
anomiex: Rector is probably not realizing that line 2518 will always throw. If we have to do anything, it would probably make more sense to put the dummy assignment after that line inside the catch block.

A comment there pointing out that the dummy assignment is just to make rector happy wouldn't be amiss either.

projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
```diff
@@ -3009,6 +3022,9 @@ protected function color(&$out) {
 	// arguments are separated by , unless a ; is in the list, then ; is the
 	// delimiter.
 	protected function argumentDef(&$args, &$isVararg) {
+		$value  = null;
```
anomiex: Both possibilities for `$method` in the call at line 3042 will be using `$value` as an output parameter, which is why the lack of assignment makes no difference.

If we want to be more semantically correct, we should probably move this to just before that line. OTOH, here is probably slightly faster if we have to do anything. Either way, I'd probably throw in a `// Initialize output parameter to make wpcom's rector happy` to make it clearer what's going on.

projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
```diff
@@ -3009,6 +3022,9 @@ protected function color(&$out) {
...
+		$rhs    = null;
```
anomiex: Same, with line 3047 this time.

projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
```diff
@@ -3009,6 +3022,9 @@ protected function color(&$out) {
...
+		$newArg = null;
```
anomiex: If this is doing anything at all, it would seem to be masking a logic error. Better would be to place it somewhere around line 3071, as if it's really possible to miss all the assignments down there then it's probably not correct to be reusing whatever happened to be set from an earlier pass through the loop.

projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
```diff
@@ -3441,6 +3457,7 @@ protected function literal($what, $eatWhitespace = null) {
 	}
 
 	protected function genericList(&$out, $parseItem, $delim="", $flatten=true) {
+		$value = null;
```
anomiex: Again, any `$parseItem` is using `$value` as an output parameter. But if we have to do anything, this is as good a place for the dummy assignment as any. Still, I might do like 🤷

```suggestion
		$value = null; // Initialize output parameter to make wpcom's rector happy.
```

projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
mreishus: Thanks!

projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
mreishus: $value and $rhs are output parameters that rector wants to be "initialized" before they're used on 3041 and 3046; 
$newArg could be initialized on each loop, but that might change the way the program works. I don't think there's anything wrong with $newArg in the way the program works, rector simply doesn't like it when path A doesn't set $newArg, path B sets $newArg, and both can land on an `if ( $newArg )`

projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
mreishus: I can't find a way to let it know without making a class for the $this->scope output

projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
```diff
@@ -2415,7 +2414,7 @@ protected function parseChunk() {
 		if (empty($this->buffer)) return false;
 		$s = $this->seek();
 
-		$block = null;
+		$block = null; // Initialize for wpcom's rector.
```
anomiex: Personally I'd still move this one down to just after line 2517. Or does rector find some other code path I'm not seeing?

projects/plugins/jetpack/modules/custom-css/custom-css/preprocessors/lessc.inc.php
anomiex: Yeah, the `$newArg` bit might change how it works, although if it does I'd be surprised it works correctly now. But I'll go along with "better to not risk it".

anomiex (dismissed):
Looks good enough to me. Haven't tested. One optional suggestion inline.

mreishus:
The "Pay with PayPal block in post editor" e2es fail no matter how many times I rebase, but I'm seeing them fail on other PRs too.

anomiex (approved):

PR #33818, opened 2023-10-26
Update to node 20

anomiex:
## Proposed changes:

Calypso has updated to node 20, which means it's time for us to do so as well.

Also, it has been two years since we switched from yarn to pnpm. Let's clean up the dummy `.engines.yarn` and remove the doc it references.

Also, since pnpm is the only thing (as far as I know) we use that checks package.json `.engines` and it's happy to check the monorepo root for everything, let's remove the engines from all the projects' package.jsons so we don't have to have such a giant PR in the future. That means we can also get rid of the linting that syncs all the projects' package.jsons and the special logic in the mirroring that strips it out of the mirrors.

### Other information:

## Jetpack product discussion

None specifically, but p4TIVU-aOV-p2 is relevant.
Announcement post: pdWQjU-Ab-p2

No

## Testing instructions:

* Does everything still work?


anomiex:
Submitted this to see if anything blows up in CI, but since Calypso reverted their update I'll leave this on hold for now.

anomiex:
Ugh, should have gone to bed an hour ago. Noting down what I've found so far to not lose context:

* Boost tests are failing early with a weird error, "Expected array buffer, or typed array to be returned for the "source" from the "transformSource" function but got null"
* This seems to have started with node 20.6.0, 20.5.1 works fine. I suspect https://href.li?https://github.com/nodejs/node/pull/47999
* Problem seems to be when you have a [`load` hook](https://nodejs.org/docs/latest-v20.x/api/module.html#loadurl-context-nextload) where you have esm -> import cjs that [gets a non-null source](https://github.com/microsoft/playwright/blob/4d8dd55b908f365cc5f683e7806ea7345d9de4d9/packages/playwright/src/transform/esmLoader.ts#L56-L64) -> require cjs that [tries to return a (default) null source](https://github.com/microsoft/playwright/blob/4d8dd55b908f365cc5f683e7806ea7345d9de4d9/packages/playwright/src/transform/esmLoader.ts#L53-L54), it blows up with that error.
* The simple fix of never returning a null source (see the "Here is an example hook that would opt-in to using the non-default behavior" bit [here](https://nodejs.org/docs/latest-v20.x/api/module.html#loadurl-context-nextload)) doesn't work either, eventually it hangs for some reason. Haven't worked out why yet.
* Never returning a non-null source for commonjs does work, at least for us. OTOH, nodejs's docs say that's going away at some point. 🤷
* Overall, not sure whether all the bugs are in nodejs or if playwright has some too. Playwright's bug may just be "don't try to use this broken feature yet".
* I wonder if we can work around all this by just converting more of the e2e config to esm (e.g. `tools/e2e-commons/config/playwright.config.default.cjs`)

anomiex:
> I wonder if we can work around all this by just converting more of the e2e config to esm (e.g. `tools/e2e-commons/config/playwright.config.default.cjs`)

Did that in #33846.

zinigor (dismissed):
I really like the cleanup, although the PR needs a conflict resolve again.

zinigor (dismissed):
Resolved the conflict, let's get it in!

anomiex:
The thing is that Calypso decided to wait on re-trying their update due to sickness and AFK (p4TIVU-aOV-p2#comment-11982), so they're still on v18. Are we ok with getting ahead of them?

anomiex:
Looks like their new PR is https://github.com/Automattic/wp-calypso/pull/83585

zinigor:
I'd say it's a good idea to wait for them. Their process is much more reliant on Node, and if they find regressions, we'll end up staying on Node 20 ahead of them longer than we expect, potentially causing workflow disruptions for people who work on both environments.

noahtallen:
While we ran out of time to ship this week due to some other priorities, we hope to have that PR shipped early next week. Other than the testing issues we ran into at the last minute, we aren't expecting any major issues hopefully!

anomiex:
Calypso updated, so we're good to go here.

zinigor (approved):
Poor labelling action has died, RIP.

PR #33823, opened 2023-10-26
Provide Jetpack Version for BlazePress App

timur987:
Main ticket is in Tumblr's Jira: `SELFSERVE-815`

Recently we introduced config for BlazePress App that provides initial data for the App in https://github.com/Automattic/jetpack/pull/30103.

This PR adds providing the current Jetpack's version to the same config.

## Proposed changes:

* Populate Jetpack's version in `'jetpack_version'` node.

### Other information:

## Testing instructions:

- Start Jetpack locally, use Jurassic Tube
- Open Tools > Advertising (if you don't see that menu item, re-establishing Jetpack connection might help)
- Open your browser's Dev Tools / Console
- Run `window.configData.initial_state.sites.items[window.configData.blog_id].options.jetpack_version` and ensure that you see your Jetpack plugin's version like in a sample below:

image


projects/packages/blaze/src/class-dashboard-config-data.php
```diff
@@ -39,6 +40,7 @@ public function get_data() {
 			'admin_page_base'          => $this->get_admin_path(),
 			'api_root'                 => esc_url_raw( rest_url() ),
 			'blog_id'                  => $blog_id,
+			'jetpack_version'          => Constants::get_constant( 'JETPACK__VERSION' ),
```
sbarbosa: If you move this option below, were we have all of the site information resides, you won't need to add the additional selector in https://github.com/Automattic/wp-calypso/pull/83590/files.

When jetpack Blaze loads, it loads the store using the `intial_state` prop. It will take the correct `jetpack_version` if you put it under `sites->items->[blog_id]`

It will make the Blaze loading easier, and we won't need to add another selector to Calypso.
What do you think?

timur987: Indeed, I didn't consider that option. Will do so, thanks @sbarbosa.

timur987: Moved in e4d9b952c6904c67eb7f113b9c9220801e4ffe30. Will update Calypso accordingly asap.

projects/packages/blaze/src/class-dashboard-config-data.php
```diff
@@ -8,6 +8,7 @@
 namespace Automattic\Jetpack\Blaze;
 
 use Automattic\Jetpack\Connection\Manager;
+use Automattic\Jetpack\Constants;
```
jeherve, line 11: Since we're using a new package here, could you add it as a dependency to the Blaze package?
https://github.com/Automattic/jetpack/blob/a771ddd7d937bdb7bd490d687d42aa939df922d6/projects/packages/blaze/composer.json#L8

Right now it's not an issue because the Jetpack plugin requires that `automattic/jetpack-constants` package already, but in the future the Blaze package may be used on sites that do not use that package, so we'll want to account for that.

After you've updated the list of dependencies, you'll want to run `tools/fixup-project-versions.sh` to ensure the lock files are updated too.

jeherve (changes requested):

projects/packages/blaze/src/class-dashboard-config-data.php
timur987: Thanks @jeherve! Hope I got you right, 38b199cfeb781f6c339b7c1b5fc605a99e8c5e0d contains your suggestion.

jeherve (approved):

PR #33827, opened 2023-10-26
Fix: wrong applied styles on Multiple checkbox in Contact From block

### LINKED ISSUE #31397: Form Block Multiple Choice checkbox - Font weight and color are not matching with the editor
### Quick summary

I am able to replicate it so far with the [Tazza theme](https://wordpress.com/theme/tazza). 

Form block's field appearance is not matching with the editor, for example, color and font weight: 

I have created a short video to demonstrate the issue: 

https://github.com/Automattic/wp-calypso/assets/54123436/206bebd5-1e69-4c22-a1be-a9cbd9f8a599

### Steps to reproduce

1. Install [Tazza theme](https://wordpress.com/theme/tazza) and activate it. 
2. Take a Form block. 
3. Add Multiple Choice ( checkbox ), and change the label colors to white or something else. 
Screen Shot on 2023-06-14 at 18:50:06
4. Check the frontend. 

The color and font-weight should match with the Editor and frontend. 

### What actually happened

The color and font-weight are not matching with the editor and frontend. 

### Impact

Some (< 50%)

### Available workarounds?

Yes, easy to implement

_No response_

### Logs or notes

Got a user here: #6396960-zen

And shared this custom CSS as a workaround: 

```
/* Adjust form multiple choice label color | ( 6396960-zen ) MT */
.contact-form label.checkbox-multiple span {
    color: #FFF !important;
    opacity: 1;
    font-size: 100%;
    font-weight: 400;
}
```

andrii-lysenko:
Fixes https://github.com/Automattic/jetpack/issues/31397

## Proposed changes:
This PR should fix wrongly applied styles. In the current state, we are applying `opacity` and other styles on all `span` elements, while we should only apply them on "(required)" line.

Note: Even though issue states about Tazza theme, issue persists on all themes.

### Other information:
This issue is a bit odd, as it seems that Styles used here are from `deprecated` Forms module, while the code used is from the new package. I don't think we should iterate on this within this Fix, but, it would be resolved when we fully remove old Forms module.
cc @jeherve 

## Testing instructions:

* Download the branch locally and run `jetpack build plugins/jetpack`
* Work with your JT or other jetpack setup environment
* Create a new post and add Jetpack Form to it
* Add a Multiple Choice (Checkbox) field
* In block editor change the Option Text color
* Save and check out the Post and verify that the styles the same as in Edit mode (see original issue for context on how it was before)
Screenshot at Oct 26 14-10-16


projects/packages/forms/src/contact-form/class-contact-form-field.php
```diff
@@ -454,7 +454,7 @@ class='grunion-field-label{$type_class}" . ( $this->is_error() ? ' form-error' :
 				. $extra_attrs_string
 				. '>'
 				. esc_html( $label )
-				. ( $required ? '<span>' . $required_field_text . '</span>' : '' )
+				. ( $required ? '<span class="required">' . $required_field_text . '</span>' : '' )
```
jeherve: I wonder if we should use a more specific class name to avoid potential conflicts with other plugins or themes that may apply styles to all `.required` containers. What do you think?

andrii-lysenko: Thanks for pointing this out. Updated. Though, I feel like removing old Forms module would be quite a challenge, as there's a lot stuff.

jeherve (approved):

PR #33832, opened 2023-10-27
Connection: add partner codes to connection URLs & deprecate Partner package

jeherve:
## Proposed changes:

This PR changes the way the Partner codes were added when connecting a site to a WordPress.com account, so this can work with standalone plugins.

This was previously only done in the Jetpack plugin, in 2 ways:

- The Partner package is initialized on plugins_loaded in the main Jetpack class.
- In Jetpack's `build_authorize_url` method we have a `jetpack_build_authorize_url` filter that is leveraged by the Partner package.

Standalone plugins, on the other hand, do not use `Jetpack::build_authorize_url()`. This method is obviously only available in the Jetpack plugin, and so standalone plugins do not have that `jetpack_build_authorize_url` filter.

We fix this discrepancy in this PR:

- We now include the Partner package functionality directly in the connection package.
- We initialize the Partner package in the connection package.
- We add the `jetpack_build_authorize_url` filter to the connection package's `get_authorization_url` method, so the partner codes can be added to all URLs.
- We stop initializing the Partner class in the Jetpack plugin, since it is now done in the connection package.

> ~**Note**~
> ~This does not add the Partner package as a new dependency for the Connection package, since the Partner package already declares the connection package as a dependency.~

> **Note**
> This PR is getting larger because we are deprecating the Partner package as part of this PR. It is no longer needed since the features it offered can be offered directly in the connection package.

### Other information:

## Jetpack product discussion

* pbtFPC-6ha-p2

* No

## Testing instructions:

There may be an easier way to test this, but here is how I went about it.

> **Warning**
> You will want to test the process in both the Jetpack plugin and a standalone plugin like Jetpack Protect

> **Warning**
> It would be useful to test with the Jetpack plugin running this branch, as well as another standalone plugin running a stable version, still using the Partner package.

* On a brand new site,
* Via WP CLI, set up an affiliate code and a subsiadiary code:
    * `wp option update jetpack_partner_subsidiary_id jeremy`
    * `wp option update jetpack_affiliate_code 123456`
* Install a Jetpack plugin (Jetpack or something else).
* Open your browser's network tools
* Filter for `jetpack.wordpress.com` requests
* Start the connection process
* Check the arguments added to the jetpack.wordpress.com request
    * It should include both an `aff` and a `subsidiaryId` argument, using the values defined above.


monsieur-z (dismissed):
I followed the instructions for testing both the Jetpack plugin and Boost and could see the query parameters. I appreciate how the code changes were documented 👍

jeherve:
The [test failure](https://github.com/Automattic/jetpack/actions/runs/6666229898/job/18117356986) outlines an interesting problem. It fails for the Inspect plugin, most likely because it doesn't include the My Jetpack package, so the Partner package is never required. I'm not quite sure how to fix that.

- My initial thought was to add the Partner package as a dependency of the Connection package, since we use `Partner` in the connection package. But since the Partner package already has the connection package as a dependency, it would create a circular dependency. Maybe that's okay?
- I could cheat and the Partner package as a dependency of the Inspect plugin, but that's not future-proof at all.

@anomiex Could I have your opinion about this?

anomiex:
> My initial thought was to add the Partner package as a dependency of the Connection package, since we use Partner in the connection package. But since the Partner package already has the connection package as a dependency, it would create a circular dependency. Maybe that's okay?

Offhand I don't know how Composer would handle it, but I know our CI will complain about a circular dependency. We'll want to avoid that.

> I could cheat and the Partner package as a dependency of the Inspect plugin, but that's not future-proof at all.

Yeah, that's cheating. We don't want that to be the long-term solution either, it'll just bite someone else who uses the Connection package in the future.

I see a few options that could work:

* Move the necessary code from Partner to Connection, since Connection needs it. If that makes Partner empty, that's fine, we've done that before.
* Make Connection "suggest" Partner, and guard any calls to it from Connection with appropriate checks (`is_callable`, `class_exists`, etc). Of course, that depends on whether working without Partner makes sense, and whether it might be confusing if e.g. Inspect's behavior differs based on whether some other plugin that provides Partner is also installed.

jeherve:
> I see a few options that could work:
> 
>     * Move the necessary code from Partner to Connection, since Connection needs it. If that makes Partner empty, that's fine, we've done that before.

That's something I was thinking about, but I wasn't sure everyone was going to be okay with that. 👍 
I've just done that in this PR.
I'll look at the test failures tomorrow, I have yet to understand what's causing the connection tests to fail now.

sergeymitr:
Hi @jeherve.
`Brain\Monkey` in `Partner_Test` did not count function mock calls correctly for some reason, which caused an error on the proper tear down attempt.
I adjusted the tests so they wouldn't need `Brain\Monkey` altogether, they seem to pass now.

sergeymitr:
Hi @jeherve.
Can you please fix the checks, and I'll give it a final look.

sergeymitr (dismissed):

sergeymitr (approved):

PR #33833, opened 2023-10-27
Boost: Update e2e tests to account for lazy loading removal

dilirity:
Jetpack removed lazy loading, so these tests have been trimmed down
because of that.
Since our module is disabled by default and can not be activated,
it shouldn't be activatable.

## Proposed changes:

* Fixes e2e tests related to recent lazy loading updates. Since the module is now visible in the UI but can not be activated, the tests now accommodate for that.

### Other information:

## Jetpack product discussion

n/a

no

## Testing instructions:

* Checkout this version of Boost;
* Run e2e tests - they should all pass (hopefully no tunnel issues).


projects/plugins/boost/tests/e2e/specs/base/jetpack-compatibility.test.js
```diff
@@ -1,94 +1,23 @@
...
 	test.beforeAll( async () => {
 		await prerequisitesBuilder().withActivePlugins( [ 'jetpack' ] ).build();
 	} );
 
-	test( 'The Jetpack lazy-image module activation reflects in Boost dashboard', async ( {
+	test( 'The Jetpack lazy-image module can not be activated in Boost dashboard', async ( {
 		page,
 	} ) => {
 		await prerequisitesBuilder( page ).withInactiveModules( [ 'lazy-images' ] ).build();
 		await activateModules( [ 'lazy-images' ] );
```
haqadn: I think lazy loading toggle tests should be removed for now until the module is fully removed.

haqadn (changes requested):
Thanks for looking into it. I think we should not have any lazy loading tests since currently it depend on too many variables like WP 6.4, Gutenberg 16.6.0. I think it will just add to the current flakyness of the tests.

projects/plugins/boost/tests/e2e/specs/base/jetpack-compatibility.test.js
dilirity: Good idea. I wasn't sure if I should leave it or remove it. I guess it will only hurt us having them here.

I've removed them in fd2e8fb and updated other tests depending on lazy loading in 5f3e766.

I also noticed out tests are outdated in general, so I opened a ticket for that https://github.com/Automattic/jetpack/issues/33877

haqadn (approved):
Thank you for taking care of this. Looks like we may need add more tests since we have added new features in Boost.

PR #33843, opened 2023-10-27
AI Assistant: expose `usage-period` in the ai-assistant-feature endpoint

### LINKED ISSUE #33844: AI Tiered Plans: Expose the next reset date on API (or how many days left before the reset)
Probably, we should collect all data into a `"current-period"` field.

Suggestion:

Image

retrofox:
It changes the response shape, collecting the usage=period data in the new `"usage-period"` field.

~requires D127135-code~ r279297-wpcom

Fixes #

## Proposed changes:

* AI Assistant: expose current period start in the ai-assistant-feature endpoint

### Other information:

## Testing instructions:

* Apply these changes in your sandbox
* Get sandboxed
* Hit the ai-assistant-feature endpoint
* Confirm you see the `"usage-period"` field in the response

Before | After
-------|-----
image|image

### DIFF
--- a/projects/plugins/jetpack/_inc/lib/class-jetpack-ai-helper.php
+++ b/projects/plugins/jetpack/_inc/lib/class-jetpack-ai-helper.php
@@ -391,7 +391,10 @@ class Jetpack_AI_Helper {
 			$usage_period_start = WPCOM\Jetpack_AI\Usage\Helper::get_usage_period_start_date( $blog_id );
 			$usage_period_start = isset( $usage_period_start ) ? $usage_period_start->format( 'Y-m-d H:i:s' ) : null;
 
-			$usage_period_requests_count = WPCOM\Jetpack_AI\Usage\Helper::get_usage_period_requests_count( $blog_id );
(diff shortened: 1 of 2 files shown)


CGastrell (dismissed):
Looks good, works fine.

lhkowalski:
We may need to expose the `get_next_period_start_date` value as well, since it will be necessary for the label stating when the request count will reset. The value being exposed on this PR is for the current period start, not the future one.

To expose the `get_next_period_start_date` from the `Period_Metered_Usage`, we may need to add a method to `WPCOM\Jetpack_AI\Usage\Helper` to handle the caching of it, the same we did for the current one.

With that in mind, maybe we could use `usage-period` instead of `current-period`, to accomodate `current-start`, `next-start` and `requests-count`.

lhkowalski (approved):
Code change looks good and works as expected.

---

_On site with a plan from September 26:_

Screenshot 2023-10-30 at 17 20 24

---

_On site without a plan:_

Screenshot 2023-10-30 at 17 20 58

PR #33852, opened 2023-10-30
Add/blogging-prompts: Add bloganuary field

roo2:
Part of pe7F0s-1iI-p2.

"Bloganuary" is a yearly promotion where bloggers are encouraged to write a blog post every day in January.

In order to integrate the writing prompts feature with the bloganuary promotion we were requested to add a new tag to posts created from blogging prompts. Id is of the form `bloganuary-yyyy-dd`.

This tag will be used so that bloggers can find other answers written as part of bloganuary.

## Proposed changes:

This change adds the ID to posts returned by the blogging-prompts v3 endpoint.

It would be possible to generate the ID on the front end, but I think that having it come from the back end would be a better design.

### Other information:

Note currently the endpoint is unused but it should be used by both calypso and the apps soon. pctCYC-KA-p2#comment-1085

No

## Testing instructions:
Apply the diff to calypso.

Call the API like so:

`https://public-api.wordpress.com/wpcom/v3/sites/<site_id>/blogging-prompts?after=--01-01&force_year=2024&order=desc`

It requires the `Authorization` and `Cookie` headers from a logged in claypso session

### DIFF
--- a/projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v3-endpoint-blogging-prompts.php
+++ b/projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v3-endpoint-blogging-prompts.php
@@ -306,6 +306,10 @@ class WPCOM_REST_API_V3_Endpoint_Blogging_Prompts extends WP_REST_Posts_Controll
 			$data['answered_link_text'] = __( 'View all responses', 'jetpack' );
 		}
 
+		if ( rest_is_field_included( 'bloganuary_id', $fields ) ) {
+			$data['bloganuary_id'] = 'bloganuary-' . $this->prepare_date_response( $prompt->post_date_gmt );
+		}
+
 		return $data;
 	}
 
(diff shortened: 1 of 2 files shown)


roo2:
note: I haven't been able to test the latest changes because of an issue with my sandbox

jeherve (commented):
This doesn't quite work for me, because `$post_date_gmt` is always in 2020:

Screenshot 2023-10-31 at 09 25 26

thomashorta:
> This doesn't quite work for me, because `$post_date_gmt` is always in 2020:
> 
> Screenshot 2023-10-31 at 09 25 26

Good point here, I believe that if the `force_year` parameter is passed we also need to force the year in the tag to be the passed year if we're building that tag dynamically.

roo2:
Thanks for the review @jeherve. I've updated the bloganuary id date field to also use `force_year` if it's available and also updated the [calypso PR](https://github.com/Automattic/wp-calypso/pull/83696) to pass that parameter along from the frontend.

jeherve (approved):

thomashorta (approved):
The changes look good to me now! Thanks! 🙇🏼

PR #33854, opened 2023-10-30
wp-admin Site Default: Fix/remove submenu from my sites

jeroenpf:
Fixes https://github.com/Automattic/dotcom-forge/issues/4405

When SSO is disabled, a submenu appears under the My Sites button. This submenu is not styled correctly and should not be there in the first place. 

## Proposed changes:

* A filter is used to determine if the standard menu should be loaded. This diff adds a filter to return true, preventing the submenu from being added.

### Other information:

## Testing instructions:

* Load this PR on a WoA site. You can do this by using the Jetpack Beta Tester or by syncing changes to a WoA dev blog.
* Enable the classic interface via the hosting configuration.
* Disable SSO via https://wordpress.com/settings/security - Disable the `Allow users to log in to this site using WordPress.com accounts` toggle.
* Go to the wp-admin interface of your WoA site and you should see the standard WordPress menu on the left (it will not have any of the Calypso links). 
* Move your mouse over the `My Sites` button in the admin bar and observe that, aside from hover feedback on the button itself, there is nothing strange happening.

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-remove-submenu-from-my-sites
+++ b/projects/plugins/jetpack/changelog/fix-remove-submenu-from-my-sites
@@ -1,0 +1,5 @@
+Significance: patch
+Type: bugfix
+Comment: It is a very trivial change
+
+
--- a/projects/plugins/jetpack/modules/masterbar/masterbar/class-masterbar.php
+++ b/projects/plugins/jetpack/modules/masterbar/masterbar/class-masterbar.php
@@ -394,6 +394,9 @@ class Masterbar {
 			$bar->remove_node( $node->id );
 		}
 
+		// This disables a submenu from being placed under the My Sites button.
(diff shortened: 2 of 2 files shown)


katinthehatsite:
I added some feedback to: https://github.com/Automattic/dotcom-forge/issues/4405 
I will wait for the response there before proceeding with the review

katinthehatsite (approved):
Works as advertised. 

Before:

Screenshot 2023-11-01 at 8 57 52 AM

After:

Screenshot 2023-11-01 at 8 54 45 AM

katinthehatsite:
@jeroenpf one thing - could you please test on self-hosted and simple to make sure nothing breaks? Thanks

jeroenpf:
@katinthehatsite thanks! I will double check this with a self-hosted and simple site although this piece of logic should only be triggered for Atomic sites and when the wpcom_admin_interface option is set to wp-admin.

PR #33390, opened 2023-09-28
Frontend email nag for custom domain 

vykes-mac:
Fixes https://github.com/Automattic/dotcom-forge/issues/3950

## Proposed changes:

* Show a frontend email nag for users with unverified email address of a custom domain
* 
image

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* Patch your sandbox using `bin/jetpack-downloader test jetpack-mu-wpcom-plugin add/domains-email-nag`
* Buy a domain with an account that has an unverified email address and map domain to a site
* Go to sites frontend while logged in and confirm you see a nag.
* Confirm clicking `Fix` takes you to /me/account

PR #33855, opened 2023-10-30
Internal release: record stat to track how often unverified domain email nag is shown

mpkelly:
Fixes n/a
Related https://github.com/Automattic/jetpack/pull/33390

## Proposed changes:

* Record a stat for how many times email nag added in https://github.com/Automattic/jetpack/pull/33390 is displayed

### Other information:

## Testing instructions:

* Apply the patch to your sandbox as per this comment https://github.com/Automattic/jetpack/pull/33855#issuecomment-1785044907
* Sandbox the wordpress.com subdomain for your site. 
* Follow the test plan here to get a site with an unverified email https://github.com/Automattic/wp-calypso/pull/83597
* Go to the site's front end and confirm a nah is shown. Check bump stat `wpcom_frontend_unverified_domain_email_nag_shown` and see that the stat has increased (there might be a delay)
* Is the stat name I used appropriate?

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
@@ -50,20 +50,6 @@ function get_domain_with_unverified_email() {
 }
 
 /**
- * Bump the MC stat for wpcom-domain-email-nag-shown.
- *
- * @param string $domain the domain to bump the stat for.
- */
-function bump_mc_stats( $domain ) {
-	$pixel_url = sprintf(
-		'http://pixel.wp.com/g.gif?v=wpcom-no-pv&x_%s=%s',
-		'wpcom-domain-email-nag-shown',
-		$domain
-	);
-	wp_remote_get( $pixel_url );
-}
-
-/**
  * Decides whether to render to the email nag.
  */
 function domain_email_nag() {
@@ -77,7 +63,11 @@ function domain_email_nag() {
 		return;
 	}
 
-	bump_mc_stats( $domain );
+	$pixel_url = sprintf(
+		'http://pixel.wp.com/g.gif?v=wpcom-no-pv&x_%s=%s',
+		'wpcom_frontend_unverified_domain_email_nag_shown',
+		$domain
+	);
 
 	wp_enqueue_style( 'wpcom-domain-email-nag-style', plugins_url( 'domain-nag.style.css', __FILE__ ), array(), Jetpack_Mu_Wpcom::PACKAGE_VERSION );
 
@@ -89,11 +79,12 @@ function domain_email_nag() {
 
 	?>
 	<div class="wp-domain-nag-sticky-message">
+		<img src="<?php echo esc_url( $pixel_url ); ?>" alt="" style="display:none;" />
 		<div class="wp-domain-nag-inner">
 			<p class="wp-domain-nag-text"><?php echo wp_kses( $notice, array( 'strong' => array() ) ); ?></p>
 			<a class="button" href="<?php echo esc_url( get_account_url() ); ?>"><?php esc_html_e( 'Fix', 'jetpack-mu-wpcom' ); ?></a>
 		</div>
-	</div>	
+	</div>
 	<?php
 }
 add_action( 'wp_footer', __NAMESPACE__ . '\domain_email_nag' );
--- a/projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
@@ -65,8 +65,8 @@ function domain_email_nag() {
 
 	$pixel_url = sprintf(
 		'http://pixel.wp.com/b.gif?v=wpcom-no-pv&x_%s=%s',
-		'wpcom_frontend_unverified_domain_email_nag_shown',
-		$domain
+		'wpcom_frontend_unverified_domain_email_nag',
+		'shown'
 	);
 
 	wp_enqueue_style( 'wpcom-domain-email-nag-style', plugins_url( 'domain-nag.style.css', __FILE__ ), array(), Jetpack_Mu_Wpcom::PACKAGE_VERSION );
--- a/projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
@@ -63,6 +63,8 @@ function domain_email_nag() {
 		return;
 	}
 
+	bump_stats_extras( 'wpcom-domain-email-nag-shown', $domain );
+
 	wp_enqueue_style( 'wpcom-domain-email-nag-style', plugins_url( 'domain-nag.style.css', __FILE__ ), array(), Jetpack_Mu_Wpcom::PACKAGE_VERSION );
 
 	$notice = sprintf(
@@ -77,7 +79,7 @@ function domain_email_nag() {
 			<p class="wp-domain-nag-text"><?php echo wp_kses( $notice, array( 'strong' => array() ) ); ?></p>
 			<a class="button" href="<?php echo esc_url( get_account_url() ); ?>"><?php esc_html_e( 'Fix', 'jetpack-mu-wpcom' ); ?></a>
 		</div>
-	</div>
+	</div>	
 	<?php
 }
 add_action( 'wp_footer', __NAMESPACE__ . '\domain_email_nag' );
--- a/projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
@@ -44,7 +44,7 @@ function get_domain() {
  */
 function is_custom_domain() {
 	$domain = get_domain();
-	return ! str_ends_with( $domain, 'wpcomstaging.com' ) && ! str_ends_with( $domain, 'wordpress.com' );
+	return ! str_ends_with( $domain, '.wpcomstaging.com' ) && ! str_ends_with( $domain, '.wordpress.com' );
 }
 
 /**
--- a/projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
@@ -57,32 +57,34 @@ function domain_email_nag() {
 
 	wp_enqueue_style( 'wpcom-domain-email-nag-style', plugins_url( 'domain-nag.style.css', __FILE__ ), array(), Jetpack_Mu_Wpcom::PACKAGE_VERSION );
 
-	$blog_id = Jetpack_Options::get_option( 'id' );
+	$blog_id = \Jetpack_Options::get_option( 'id' );
 
-	wc_enqueue_js(
-		sprintf(
-			"
-			const base = 'https://public-api.wordpress.com';
-			const path = '/v1/domains/%d/is-domain-email-unverified';
-			
-			fetch(base + path).then(function (result) {
-				if (result) {
-					result.json().then(function (body) {
-						if (body.unverified) {
-							const nag = document.querySelector('.wp-domain-nag-sticky-message');
-							if (nag) {
-								nag.style.display = 'block';
-								const statUrl =
-									'http://pixel.wp.com/b.gif?v=wpcom-no-pv&x_wpcom_frontend_unverified_domain_email_nag=shown';
-								fetch(statUrl);
-							}
+	if ( ! filter_var( $blog_id, FILTER_VALIDATE_INT ) ) {
+		return;
+	}
+
+	$script = sprintf(
+		"
+		const base = 'https://public-api.wordpress.com';
+		const path = '/v1/domains/%d/is-domain-email-unverified';
+		
+		fetch(base + path).then(function (result) {
+			if (result) {
+				result.json().then(function (body) {
+					if (body.unverified) {
+						const nag = document.querySelector('.wp-domain-nag-sticky-message');
+						if (nag) {
+							nag.style.display = 'block';
+							const statUrl =
+								'http://pixel.wp.com/b.gif?v=wpcom-no-pv&x_wpcom_frontend_unverified_domain_email_nag=shown';
+							fetch(statUrl);
 						}
-					});
-				}
-			});
-		",
-			$blog_id
-		)
+					}
+				});
+			}
+		});
+	",
+		$blog_id
 	);
 
 	$domain = get_domain();
@@ -100,6 +102,7 @@ function domain_email_nag() {
 			<a class="button" href="<?php echo esc_url( get_account_url() ); ?>"><?php esc_html_e( 'Fix', 'jetpack-mu-wpcom' ); ?></a>
 		</div>
 	</div>
+	<script><?php echo esc_js( $script ); ?></script>
 	<?php
 }
 add_action( 'wp_footer', __NAMESPACE__ . '\domain_email_nag' );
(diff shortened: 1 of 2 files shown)


projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
```diff
@@ -49,6 +49,20 @@ function get_domain_with_unverified_email() {
...
+ *
+ * @param string $domain the domain to bump the stat for.
+ */
+function bump_mc_stats( $domain ) {
+	$pixel_url = sprintf(
+		'http://pixel.wp.com/g.gif?v=wpcom-no-pv&x_%s=%s',
+		'wpcom-domain-email-nag-shown',
+		$domain
+	);
+	wp_remote_get( $pixel_url );
```
danielbachhuber: `wp_remote_get()` is blocking and will slow down the request. Could we display the pixel instead?

mpkelly: I will switch to rendering an image instead. Thanks!

projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
```diff
@@ -49,6 +49,20 @@ function get_domain_with_unverified_email() {
...
 
+/**
+ * Bump the MC stat for wpcom-domain-email-nag-shown.
+ *
+ * @param string $domain the domain to bump the stat for.
+ */
+function bump_mc_stats( $domain ) {
+	$pixel_url = sprintf(
+		'http://pixel.wp.com/g.gif?v=wpcom-no-pv&x_%s=%s',
+		'wpcom-domain-email-nag-shown',
```
p-jackson: I think we need to use pascal_case instead of kebab-case. Pretty sure invalid names get filtered out.
PCYsg-4S2-p2

projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
```diff
@@ -77,7 +93,7 @@ function domain_email_nag() {
 			<p class="wp-domain-nag-text"><?php echo wp_kses( $notice, array( 'strong' => array() ) ); ?></p>
 			<a class="button" href="<?php echo esc_url( get_account_url() ); ?>"><?php esc_html_e( 'Fix', 'jetpack-mu-wpcom' ); ?></a>
 		</div>
-	</div>
+	</div>
```
p-jackson: whitespace nitpick

projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
p-jackson: Maybe this?
`wpcom_frontend_unverified_domain_email_nag_shown`

I like saying frontend because it wasn't initially clear to me what context the nag was shown in.

p-jackson (commented):
Made some notes but I wasn't able to test this unfortunately.
I have an unverified email, and I can see the unverified domain email banner in the Calypso sidebar, but after sandboxing the frontend and applying this diff to my sandbox, I saw no banner 😕

mpkelly:
> * Sandbox the wordpress.com subdomain for your site.

@p-jackson, did you do this part? It would mean an entry in your `/etc/hosts` like `192.0.94.44 testname.wordpress.com`

projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
```diff
@@ -63,6 +63,12 @@ function domain_email_nag() {
 		return;
 	}
 
+	$pixel_url = sprintf(
+		'http://pixel.wp.com/g.gif?v=wpcom-no-pv&x_%s=%s',
```
p-jackson: I prefer using `b.gif` which does the same thing but renders a transparent pixel rather than a smiley face. I'm always paranoid that there'll be some error that causes the image to appear really big or something, and the smiley face would feel inappropriate then 😅 that's me being paranoid though.

I'm not seeing any of these stats appear in MC yet. I would have thought I'd see something from testing (even if I personally have only just managed to get the nag showing).

Maybe that's because we need to change the URL to something like. `&x_wpcom_frontend_unverified_domain_email_nag=shown`, with an equals in there. I don't see why the group and event name can't be the same. But maybe it can't and I just remembered that this way of doing things (separating out the verb into a separate name) is more common. So maybe it's required or something.

p-jackson (commented):
I was able to figure it out why I wasn't seeing the nag. It's because `class_exists( 'Domain_Management' )` returns `false` for me.

I can see why. `Domain_Management` is defined in `admin-plugins`, and admin plugins are only loaded when in the admin view. We're rendering the front end here and so I'd totally expect this class not to exist.

I tested the stat by requiring the `Domain_Management` code on my sandbox and I got the nag appearing. But I don't think it's good to do that in production. We keep front end rendering performant by not loading all the code, and who knows what spaghetti would get included if we started to require it.

I'll merge this once I see the event working in MC. But I think we'll need a follow up which finds a way to get icann info on the front end. Maybe it'll be ok as long as we only load code when we know it's the owner who's logged in 🤔

p-jackson (dismissed):
I discovered a way to test this change.

I've made some changes to use `b.gif` instead of `g.gif` and also anonymised the bump stat. I actually wonder whether using the domain as the event name might have caused issues with the URL not being encoded correctly.

projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
mpkelly: TIL. Thanks for this change.

mpkelly:
> I discovered a way to test this change.
> 

I am just testing it now and no longer see the banner. I can't see why, but I will take a closer look tomorrow. 

Edit. Been a long day. There is a discussion on Slack about removing the dependency on `Domain_Management` which I think should make this banner show consistently.

mpkelly:
I've updated this but haven't tested it yet. I will do that tomorrow (8th)

projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
```diff
@@ -32,47 +32,69 @@ function get_account_url() {
...
 		return;
 	}
 
-	$domain = get_domain_with_unverified_email();
+	wp_enqueue_style( 'wpcom-domain-email-nag-style', plugins_url( 'domain-nag.style.css', __FILE__ ), array(), Jetpack_Mu_Wpcom::PACKAGE_VERSION );
 
-	if ( ! $domain ) {
-		return;
-	}
+	$blog_id = Jetpack_Options::get_option( 'id' );
```
p-jackson: This one's causing an error on the front end
CleanShot 2023-11-09 at 20 10 55@2x

I bet it's a missing namespace.

projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
```diff
@@ -32,47 +32,69 @@ function get_account_url() {
...
-	foreach ( $domains as $domain ) {
-		if ( $domain['is_pending_icann_verification'] === true ) {
-			return $domain['domain'];
-		}
-	}
-	return false;
+/**
+ * Checks if the site is using a custom domain.
+ */
+function is_custom_domain() {
```
p-jackson, line 45: This `is_custom_domain` check means we'll only perform the check if the primary domain is custom. I found myself in a situation where my primary address was `*.wordpress.com` but I have a secondary domain which is custom. This check means I don't see the nag even though the API is supposed to be checking all my domains, not just the primary one.

I like the optimisation of trying to only enqueue the nag code when it's necessary, but I don't think we enough information here 🧐

projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
```diff
@@ -32,47 +32,69 @@ function get_account_url() {
...
+	$domain = get_domain();
+	return ! str_ends_with( $domain, 'wpcomstaging.com' ) && ! str_ends_with( $domain, 'wordpress.com' );
```
p-jackson:
```suggestion
	return ! str_ends_with( $domain, '.wpcomstaging.com' ) && ! str_ends_with( $domain, '.wordpress.com' );
```

We should have the extra dots in the substring check. Otherwise it'll return `false` if the user does something clever like register the `my-special-domain-wordpress.com` domain.

The security workshop has made me paranoid 😅

projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
```diff
@@ -32,47 +32,69 @@ function get_account_url() {
...
-	$domain = get_domain_with_unverified_email();
+	wp_enqueue_style( 'wpcom-domain-email-nag-style', plugins_url( 'domain-nag.style.css', __FILE__ ), array(), Jetpack_Mu_Wpcom::PACKAGE_VERSION );
 
-	if ( ! $domain ) {
-		return;
-	}
+	$blog_id = Jetpack_Options::get_option( 'id' );
 
-	wp_enqueue_style( 'wpcom-domain-email-nag-style', plugins_url( 'domain-nag.style.css', __FILE__ ), array(), Jetpack_Mu_Wpcom::PACKAGE_VERSION );
+	wc_enqueue_js(
```
p-jackson: `wc_enqueue_js` is defined in Woo so pretty sure we can't use it here.

`wc_enqueue_js` does its magic by subscribing to the `wp_footer` action and then echoing out the `<script>` tag. We can probably just do that ourselves.

projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
```diff
@@ -32,47 +32,69 @@ function get_account_url() {
...
+								const statUrl =
+									'http://pixel.wp.com/b.gif?v=wpcom-no-pv&x_wpcom_frontend_unverified_domain_email_nag=shown';
+								fetch(statUrl);
+							}
+						}
+					});
+				}
+			});
+		",
+			$blog_id
```
p-jackson: Seeing the variable injected directly into a JavaScript string made me feel nervous. Even though I'm not sure how a user could spoof it. Could we do something like this to be extra paranoid?

```
filter_var( $blog_id, FILTER_VALIDATE_INT )
```

There might also be a core WP function for this sort of thing.

projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
```diff
@@ -32,47 +32,69 @@ function get_account_url() {
...
-		return;
-	}
+	$blog_id = Jetpack_Options::get_option( 'id' );
 
-	wp_enqueue_style( 'wpcom-domain-email-nag-style', plugins_url( 'domain-nag.style.css', __FILE__ ), array(), Jetpack_Mu_Wpcom::PACKAGE_VERSION );
+	wc_enqueue_js(
+		sprintf(
+			"
+			const base = 'https://public-api.wordpress.com';
+			const path = '/v1/domains/%s/is-domain-email-unverified';
```
p-jackson: Similar to the above suggestion, maybe it's enough to do this?

```suggestion
			const path = '/v1/domains/%d/is-domain-email-unverified';
```

But I worry that it'll throw an error at runtime instead of fail silently.

p-jackson (commented):
Made some notes below. Also making a note that we should make sure to check this on atomic too before deploying. It's in the mu plugin so it runs on atomic too.

Another idea, probably for another diff, but I saw this UI in the launcher header (I'm not sure what we call it). Maybe this'd be a more attractive place for the nag to appear?

CleanShot 2023-11-09 at 19 58 42@2x

projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
mpkelly: Thanks for that. I will just use a script tag.

projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
mpkelly: Yeah, I would still prefer this as doing an API call every time the user is on the home feels too expensive to me, and I think most people will use their custom domain instead of ours.

projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
mpkelly: I still prefer this as doing an API call every time the user is on the home feels too expensive to me, and I think most people will use their custom domain instead of ours.

mpkelly:
It's been a bit of a fight but this is getting closer. I had to make some local modifications when testing because I am using a test store domain and these don't get included in the API response. I will need to buy a real domain to test this before merging. Still need to test on Atomic too. I also ran into a permission issue so need to debug the backend diff. 

image

projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
p-jackson: I guess if they're still using `.wordpress.com` as their primary address it's not as pressing an issue to verify their domain email. It won't break their site to the same extent.

projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
```diff
@@ -32,52 +32,80 @@ function get_account_url() {
...
+						}
+					}
+				});
+			}
+		});
+EOD;
+
+	$script = sprintf(
+		$script,
+		$blog_id
```
p-jackson, line 90: Just a heads up that this `$blog_id` value hasn't been run through the `filter_var` function.

projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
```diff
@@ -32,52 +32,80 @@ function get_account_url() {
...
 
 	?>
-	<div class="wp-domain-nag-sticky-message">
+	<div class="wp-domain-nag-sticky-message" style="display:none;">
 		<div class="wp-domain-nag-inner">
 			<p class="wp-domain-nag-text"><?php echo wp_kses( $notice, array( 'strong' => array() ) ); ?></p>
 			<a class="button" href="<?php echo esc_url( get_account_url() ); ?>"><?php esc_html_e( 'Fix', 'jetpack-mu-wpcom' ); ?></a>
 		</div>
 	</div>
+	<script><?php echo $script; // phpcs:ignore -- output generated on server ?></script>
```
p-jackson, line 108: I like this. Sticking a `<script>` right here is just simpler. I think we can add the `defer` attribute so the JS isn't render blocking and it behaves more like the script appearing in the footer.

projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
mpkelly: The way I did it was to return early if it is not a valid `int`. Should be ok like this?

mpkelly:
I'm getting a 403 now, so will need to send over a token with the request. Otherwise, I think this is ready.

projects/packages/jetpack-mu-wpcom/src/features/domain-email-nag/domain-email-nag.php
p-jackson: Oh I see your right. For some reason I remember `filter_var()` always returning a valid int even if you passed it garbage. But it actually returns `false`, so this is fine.

mpkelly:
This proved quite challenging and has performance implications for all sites with custom domains, so we figured it was not worth it. Closing.

PR #33856, opened 2023-10-30
Components: Fix IconTooltip Popover

haqadn:
## Proposed changes:
Seems like there was some changes in the Popover component and it started using portals. Which meant that it would embed it in a different position in the page breaking some of the styles we apply. This portals can be avoided by using `inline: true` argument in the popover. This PR does just that.

### Other information:

## Jetpack product discussion
pc9hqz-2ie-p2#comment-1829

No

## Testing instructions:
* Go to the boost in WP admin
* The pricing table should be visible
* Click on one of those info tooltip icons.
* The popovers should be properly styled.

image


dilirity:
:shipit:

dilirity (approved):
:shipit:

haqadn:
Cherry picked: a4915f0823a948c682

PR #33862, opened 2023-10-30
Jetpack Sync: Update Dedicated Sync init hook priority to 200

fgiannar:
Reverts the previous update of the Dedicated Sync `init` priority from 90 to 0 and sets it to 200 to ensure custom post types that are registered on init with a priority of `100` will still be synced.

It's important to note that this is a temporary solution till we update the logic of checking and syncing constants and callables before sending an action to WPCOM.

Our end goal in general is still to set the init hook priority to 0.

Please checkout linked discussions for more context.

### Other information:

## Jetpack product discussion

- p1697623116284589-slack-C02JQ08G0
- p9dueE-7Vs-p2

No

## Testing instructions:

- Add a custom post type on init with priority > `90`. You can use a plugin like Code Snippets or modify your `functions.php` file as follows:
```
function add_lala_cpt() {
	$args = array(
		'public' => true,
		'label'  => 'Lala',
	);
	register_post_type( 'lala', $args );
}

add_action( 'init', 'add_lala_cpt', 200 );
```

- Ensure the `post_types` callable is synced with the `lala` cpt

### DIFF
--- a/projects/packages/sync/changelog/fix-dedicated-sync-init-hook-priority
+++ b/projects/packages/sync/changelog/fix-dedicated-sync-init-hook-priority
@@ -1,0 +1,5 @@
+Significance: patch
+Type: fixed
+Comment: Update Dedicated Sync init hook priority to 200
+
+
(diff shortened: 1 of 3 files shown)


bisko (approved):
LGTM :shipit:

PR #33864, opened 2023-10-30
Repo Gardening: flag contact form changes in package as well

jeherve:
## Proposed changes:

The Contact Form feature used to live only as a module in the Jetpack plugin. Now that it lives in a separate package, we should still continue to automatically add the labels we were adding when the feature lived in the Jetpack plugin.

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

This cannot be tested in this repo, you will need to test in a fork. Here is an example:
https://github.com/jeherve/jetpack/pull/113


andrii-lysenko (approved):
Looks good and works as described 👍

PR #33865, opened 2023-10-30
Subscribe block: refactor block rendering into a single function

simison:
## Proposed changes:

* Block rendering was done in two separate functions for Jetpack and WP.com sites, although the rendering seemed identical at both with minor differences. Combining rendering into one.

### Other information:

## Testing instructions:

Test:

*  Subscribe block on WP.com simple
*  Subscribe block on Jetpack self hosted
* Subscribe widget

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -609,11 +609,12 @@ function get_post_access_level_for_current_post() {
  */
 function render_for_website( $data, $classes, $styles ) {
 	$blog_id            = \Jetpack_Options::get_option( 'id' );
-	$form_id            = 'subscribe-blog-' . $data['widget_id'];
+	$widget_id_suffix   = Jetpack_Subscriptions_Widget::$instance_count > 1 ? '-' . Jetpack_Subscriptions_Widget::$instance_count : '';
+	$form_id            = 'subscribe-blog' . $widget_id_suffix;
 	$form_url           = defined( 'SUBSCRIBE_BLOG_URL' ) ? SUBSCRIBE_BLOG_URL : '#';
 	$post_access_level  = get_post_access_level_for_current_post();
 	$post_id            = get_the_ID();
-	$subscribe_field_id = apply_filters( 'subscribe_field_id', 'subscribe-field-', $data['widget_id'] );
+	$subscribe_field_id = apply_filters( 'subscribe_field_id', 'subscribe-field' . $widget_id_suffix, $data['widget_id'] );
 	$tier_id            = get_post_meta( $post_id, META_NAME_FOR_POST_TIER_ID_SETTINGS, true );
 
 	ob_start();
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -580,11 +580,8 @@ function render_block( $attributes ) {
 	if ( ! jetpack_is_frontend() ) {
 		return render_for_email( $data, $styles );
 	}
-	if ( is_wpcom() ) {
-		return render_wpcom_subscribe_form( $data, $classes, $styles );
-	}
 
-	return render_jetpack_subscribe_form( $data, $classes, $styles );
+	return render_for_website( $data, $classes, $styles );
 }
 
 /**
@@ -602,7 +599,7 @@ function get_post_access_level_for_current_post() {
 }
 
 /**
- * Renders the WP.com version of the subscriptions block.
+ * Renders the subscriptions block at the site.
  *
  * @param array $data    Array containing block view data.
  * @param array $classes Array containing the classes for different block elements.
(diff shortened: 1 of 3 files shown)


lezama:
@Automattic/apex @Automattic/gold any concerns or particularities we should be careful here?

TimBroddin:
> @Automattic/apex @Automattic/gold any concerns or particularities we should be careful here?

No, we don't have any business here since moving the newsletter categories to the subscribe modal 😀

Thanks for cleaning this up @simison !

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -599,19 +596,22 @@ function get_post_access_level_for_current_post() {
...
  * @return string
  */
-function render_wpcom_subscribe_form( $data, $classes, $styles ) {
-	global $current_blog;
-
-	$form_id = 'subscribe-blog' . ( Jetpack_Subscriptions_Widget::$instance_count > 1 ? '-' . Jetpack_Subscriptions_Widget::$instance_count : '' );
-	$url     = defined( 'SUBSCRIBE_BLOG_URL' ) ? SUBSCRIBE_BLOG_URL : '';
+function render_for_website( $data, $classes, $styles ) {
+	$blog_id            = \Jetpack_Options::get_option( 'id' );
+	$form_id            = 'subscribe-blog-' . $data['widget_id'];
```
millerf: On WPCOM, form id was always "subscribe-blog". But I don't think it matters here.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -599,19 +596,22 @@ function get_post_access_level_for_current_post() {
...
+	$form_url           = defined( 'SUBSCRIBE_BLOG_URL' ) ? SUBSCRIBE_BLOG_URL : '#';
+	$post_access_level  = get_post_access_level_for_current_post();
+	$post_id            = get_the_ID();
+	$subscribe_field_id = apply_filters( 'subscribe_field_id', 'subscribe-field-', $data['widget_id'] );
```
millerf:
```suggestion
	$subscribe_field_id = apply_filters( 'subscribe_field_id', 'subscribe-field', $data['widget_id'] );
```
I think small typo here

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -776,22 +634,38 @@ function render_jetpack_subscribe_form( $data, $classes, $styles ) {
...
+								id="<?php echo esc_attr( $subscribe_field_id . '-label' ); ?>"
+								for="<?php echo esc_attr( $subscribe_field_id ); ?>"
 								class="screen-reader-text"
-								for="<?php echo esc_attr( $subscribe_field_id . '-' . $data['widget_id'] ); ?>">
+							>
 								<?php echo esc_html( $data['subscribe_placeholder'] ); ?>
 							</label>
-							<input type="email" name="email" required="required"
-								<?php if ( ! empty( $classes['email_field'] ) ) : ?>
-									class="<?php echo esc_attr( $classes['email_field'] ); ?> required"
```
millerf, line 789: Nice. Fixes the missing '"'

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -805,20 +679,16 @@ class="<?php echo esc_attr( $classes['email_field'] ); ?> required"
 							<input type="hidden" name="sub-type" value="<?php echo esc_attr( $data['source'] ); ?>"/>
 							<input type="hidden" name="redirect_fragment" value="<?php echo esc_attr( $form_id ); ?>"/>
 							<?php
+							wp_nonce_field( 'blogsub_subscribe_' . $blog_id, '_wpnonce', false );
```
millerf, line 686: That nonce was not present on the Jetpack form. Probably not an issue.

millerf: Also, it used to be surrounded with:
```
<?php
							if ( is_user_logged_in() ) {
								wp_nonce_field( 'blogsub_subscribe_' . get_current_blog_id(), '_wpnonce', false );
							}
							?>
```
```suggestion
							if ( ( new Host() )->is_wpcom() && is_user_logged_in() ) {
							       wp_nonce_field( 'blogsub_subscribe_' . $blog_id, '_wpnonce', false );
							}
```

millerf (commented):
Nice job 🙏

I checked all the forms depending on the environment: 32531-pb/

Difference between WPCOM without/with this patched
Screenshot 2023-10-31 at 10 58 46

Difference between Jetpack with/without the patch
Screenshot 2023-10-31 at 10 59 16

I don't see anything broken but I'll add some comments to make closer to production...

There is also something not proper with $subscribe_field_id. The logic before was a bit wonky, and I am not exactly sure how much it matters, but I think it is worth a small revisit...

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
simison: On Jetpack we had:

```
subscribe-blog
subscribe-blog-1
subscribe-blog-2
```

etc. The same for `$subscribe_field_id`.

I basically simplified and unified it a bit. It just tries to avoid multiple same IDs on the same page.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
simison: It was surrounded by logged-in check in Jetpack, but not on WP.com:

https://github.com/Automattic/jetpack/blob/bba5b43f5f58d491403009a7816b4626973f8b3d/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php#L695

https://github.com/Automattic/jetpack/blob/bba5b43f5f58d491403009a7816b4626973f8b3d/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php#L817-L821

I'm fine with either way, as long as it's the same and works. 😁

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
simison: I think it needs `. $data['widget_id']` added

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
millerf: > It was surrounded by logged-in check in Jetpack, but not on WP.com:

Oh I missed that. I guess keeping it on both is fine, then...

lezama: pushed [this commit](https://github.com/Automattic/jetpack/pull/33865/commits/31eba3884a6ffc1c893e7664afa5f2228e682e6d) so we always verify the nonce too

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
lezama: tried to clean up element ids with https://github.com/Automattic/jetpack/pull/33865/commits/dfbf7e3c537fc0ce786cffcce2044c8a22096e6c

simison (commented):
Haven't re-tested but reviewed Miguel's changes and they make sense.

lezama (approved):

PR #33871, opened 2023-10-30
Jetpack Search AI Block: Improve upgrade flow. 

dereksmart:
## Proposed changes:

There are now two different nudges that could show: 
1. Upgrade nudge
- Now shows and upgrade prompt if they are on the "free" plan. 
- Upgrade now takes them directly to the wordpress.com checkout cart. 
2. Activate Instant Search
- Shows if the site is paid plan, yet Instant Search is not active. 
- Instant search is the feature name used in the Search UI. 

image

image

### Other information:
A nice side effect of going directly to wordpress.com checkout flow is that it will preserve the redirect back to the editor after purchase. Also, there seemed to be no way to show a pricing page within wp-admin when the site is on a "free" plan. 

I looked into `useModuleStatus` https://github.com/Automattic/jetpack/pull/33064#discussion_r1328911440 as a way to enable the Instant Search setting directly from the editor. The issue is that it only activates the entire module, and the AI block depends on the Instant Search feature specifically. This means we'd need to either find or build a way to update the specific setting through API. For future reference, [this is how it's done](https://github.com/Automattic/jetpack/blob/3ead806e5862ecee00518b56fede626e1e08efa3/projects/packages/search/src/dashboard/components/module-control/index.jsx#L98C8-L98C27) in the Jetpack Search UI.  

## Jetpack product discussion

peGwbA-Be-p2#comment-1794

No

## Testing instructions:

- Start without Jetpack Search. 
- Add the block to the editor. You should see the upgrade prompt. 
- Click the upgrade prompt. It should take you to wordpress.com checkout. 
- Complete the purchase. 
- You should be redirected back to the the post editor with your changes saved. 

After purchasing: 
- Deactivate the "Instant Search" setting from the Jetpack Search admin page. 
- Go back to the editor with your block. 
- Test out the "Enable" nudge. It will redirect you back to the search settings to manually do it.


gibrown:
If we are allowing it to work on the free tier, what about showing a link to "Powered by Jetpack Search" on the front end when this is on the free plan and making it optional (but on by default) when the user is on a paid plan? This is what we do with the modal also.

PR #33875, opened 2023-10-31
Social: Fix initial state feature discrepancy

gmjuhasz:
Fixes an issue where feature toggles would appear disabled after turning on Publicize for the first time.

## Proposed changes:

* Added an option to ignore if Publicize is enabled, which we can use for populating the initial states
* Fixed a minor issue where feature toggles could be updated while Publicize was still turning on

### Other information:

## Jetpack product discussion

https://github.com/orgs/Automattic/projects/733/views/2?pane=issue&itemId=43051048

No.
## Testing instructions:

Without this patch:
* Create a JN site
* Connect Jetpack and select the free plan
* Go to settings->Sharing, and enable auto-sharing
* Notice that the auto-conversion setting is displayed as disabled
* Refreshing the page solves it
* On your JT site turning Publicize off in the admin page or Jetpack settings, refreshing and then turning it on causes the same issue

With this patch
* Doing the same, it should reflect the state for the first time too.

### DIFF
--- a/projects/packages/publicize/changelog/fix-social-initial-state-feature-discrepancy
+++ b/projects/packages/publicize/changelog/fix-social-initial-state-feature-discrepancy
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fixed an issue where Social initial states were messed up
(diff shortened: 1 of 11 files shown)


gmjuhasz:
Going with other approach

PR #33883, opened 2023-10-31
CLI Tools: update scaffolding, use block.json to register blocks.

### LINKED ISSUE #33173: Update jetpack CLI command to generate new blocks using block.json
Following #32697, we should encourage folks to use `block.json` to register new blocks in Jetpack. Updating the Jetpack CLI command would go a long way to do that:

https://github.com/Automattic/jetpack/blob/a6a9739405d4579bbb7b0dd9b63f495ee0e8ae49/projects/plugins/jetpack/class.jetpack-cli.php#L1919-L1948

*****

Primary issue: #32602

jeherve:
Fixes #33173

## Proposed changes:

Let's update our scaffolding to encourage others to use the new nomemclature introduced in #32602.

### Other information:

## Jetpack product discussion

* #32602

* No

## Testing instructions:

* On your local checkout, run `jetpack docker wp -- jetpack scaffold block "Amazing Rock" --slug="good-music" --description="Rock the best music on your site" --keywords="music, audio, media, four, five"`, then build with `jetpack build plugins/jetpack`, and observe the new block in the block editor.


monsieur-z (dismissed):
Thanks, these changes are welcome.

I'm having some trouble running scaffolding commands locally (even on `trunk`), so you may want to ask for additional testing. But everything seems good codewise.

projects/plugins/jetpack/wp-cli-templates/block-block-json.mustache
```diff
@@ -0,0 +1,18 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/block.json",
+	"apiVersion": 1,
```
anomiex, line 3: Should new blocks be defaulting to a later api version?

For version 2, it looks like we'll also have to have the default `edit` function use `useBlockProps` as documented at https://developer.wordpress.org/block-editor/reference-guides/block-api/block-edit-save/. And maybe something similar for the `save` function and `useBlockProps.save`.

Version 3 will probably have to wait until we're not using 'enqueue_block_editor_assets' to enqueue the monolithic editor.js though.

projects/plugins/jetpack/wp-cli-templates/block-block-json.mustache
```diff
@@ -0,0 +1,18 @@
...
+	"name": "jetpack/{{ slug }}",
+	"title": "{{ title }}",
+	"description": "{{ description }}",
+	"keywords": [{{keywords}}],
+	"version": "{{ nextVersion }}",
+	"textdomain": "jetpack",
+	"category": "grow",
+	"icon": "<svg viewBox='0 0 32 32' width='32' height='32' xmlns='http://www.w3.org/2000/svg'><path d='M9.33301 5.33325L10.4644 8.20188L13.333 9.33325L10.4644 10.4646L9.33301 13.3333L8.20164 10.4646L5.33301 9.33325L8.20164 8.20188L9.33301 5.33325Z'/><path d='M21.3333 5.33333L22.8418 9.15817L26.6667 10.6667L22.8418 12.1752L21.3333 16L19.8248 12.1752L16 10.6667L19.8248 9.15817L21.3333 5.33333Z'/><path d='M14.6667 13.3333L16.5523 18.1144L21.3333 20L16.5523 21.8856L14.6667 26.6667L12.781 21.8856L8 20L12.781 18.1144L14.6667 13.3333Z'/></svg>",
```
anomiex: Is it intentional that this is now using a "three differently-sized four-pointed stars" icon as a default icon rather than an "i-in-a-circle" icon?

projects/plugins/jetpack/wp-cli-templates/block-block-json.mustache
```diff
@@ -0,0 +1,18 @@
...
+	"supports": {
+		"align": true,
+		"alignWide": false,
+		"html": false
```
anomiex: Too bad we lose comments in this file.

I note the old version had different defaults here. Are the changes from that intentional?
```suggestion
		"align": false,
		"alignWide": true,
		"anchor": false,
		"customClassName": true,
		"className": true,
		"html": false,
		"multiple": true,
		"reusable": true
```

anomiex (commented):
When running the script, I note the following issues:

* If you specify `--keywords`, this produces
  ```
  "keywords": [Array],
  ```
  in the block.json which is not terribly helpful.
* It also doesn't properly encode the keywords, description, and such in the json, but the old version had that bug too.
* Another existing bug is that for some reason `--keywords` only accepts three keywords. If you try to do more, the output has all of them combined as the third.

projects/plugins/jetpack/wp-cli-templates/block-block-json.mustache
jeherve: @monsieur-z What's your take on this?

projects/plugins/jetpack/wp-cli-templates/block-block-json.mustache
jeherve: I changed it back in 88425e5a126a07ee53e5c9d25ece8f7ef1137663

jeherve:
@anomiex Thanks for the feedback. I addressed most of it above. I had a question about

> It also doesn't properly encode the keywords, description, and such in the json, but the old version had that bug too.

What type of encoding issue did you have in mind in this case? I tried `jetpack docker wp -- jetpack scaffold block "Block with accented characters éóőöàá" --slug="good-music" --description="Rock the best music of the 70’s" --keywords="mot clé, audio, media, four, five"` and didn't run into any issues, but I assume you mean something else?

anomiex:
> What type of encoding issue did you have in mind in this case? 

Try putting `"` (or `'` for the existing version) in the values. It doesn't encode them properly for JSON.

Newlines might be an issue too, if someone tries a multi-line description, but I didn't try that.

jeherve:
> > What type of encoding issue did you have in mind in this case?
> 
> Try putting `"` (or `'` for the existing version) in the values. It doesn't encode them properly for JSON.
> 
> Newlines might be an issue too, if someone tries a multi-line description, but I didn't try that.

Would something like d4c48c3a52a7ec121eaff71cbcbbc141170faf97 help?

anomiex:
Since we're dealing with json, it might be better to use actual `json_encode()` rather than `addcslashes()` which may handle things slightly differently.

Of course, that would mean that on the template side we'd need to not include the quotes since they're now in the source strings. Or else we'd need to do like `substr( 1, -1 )` to remove the quotes so the templates can still re-add them.

projects/plugins/jetpack/wp-cli-templates/block-block-json.mustache
monsieur-z: It makes sense to nudge people to use the latest version when possible. In this case (using 2 or 3), we'd probably need to edit the `save` and `edit` functions in the templates to include `useBlockProps();` (as in the link Brad copied) and make it visible it's required.

As for which version to use, I can't say I fully understand what [using version 3](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-api-versions/) involves. The iframing might be an issue for some, but it seems it wouldn't occur unless all blocks are registered with 3 anyway.

anomiex: What I've managed to gather is that we have to make sure the scripts will be enqueued inside the iframe. Which it seems may involve somehow using the 'enqueue_block_assets' hook rather than 'enqueue_block_editor_assets', or else doing the enqueuing via block.json.

https://developer.wordpress.org/block-editor/how-to-guides/enqueueing-assets-in-the-editor/#editor-content-scripts-and-styles (and [the following section](https://developer.wordpress.org/block-editor/how-to-guides/enqueueing-assets-in-the-editor/#block-scripts-and-styles)) might be relevant.

jeherve:
> Since we're dealing with json, it might be better to use actual `json_encode()` rather than `addcslashes()` which may handle things slightly differently.
> 
> Of course, that would mean that on the template side we'd need to not include the quotes since they're now in the source strings. Or else we'd need to do like `substr( 1, -1 )` to remove the quotes so the templates can still re-add them.

Is this what you had in mind? 1f39cf31150c6e3add89a5178c06cf4ec791a63e

It doesn't support new lines, but I personally think that's okay, I don't believe that is something that will be commonly passed when using the CLI command.

anomiex (approved):
Yes, that's what I had in mind. Also it supported newlines fine when I tested it.

projects/plugins/jetpack/wp-cli-templates/block-block-json.mustache
jeherve: 👍 Discussion can continue in #34120

PR #33886, opened 2023-10-31
Boost: Migrate svelte components to react

dilirity:
Migrate several components to React and added classname pnpm dependency.

## Proposed changes:

* ActivateLicense
* ImageCdnRecommendation
* Image Size Analysis - Hero
* RecommendationContext (tooltip)
* ScoreContext (tooltip)
* Image Size Analysis - OtherGroupContext (tooltip)
* Support

Tooltips throughout the UI now use the IconTooltip RNA component.

### Other information:

## Testing instructions:

* Check if the UI components work as expected.


dilirity:
> Speed Score feature › Should be able to hover info icon next to overall score and see the detailed overall score description popin 

Tests are broken. Info tooltip no longer works on hover - just on mouse click.

projects/plugins/boost/app/assets/src/js/react-components/image-size-analysis/hero/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+		hour12: true,
+	} );
+
+	const lastUpdated = formatter.format( isaLastUpdated );
+	const showLatestReport = hasActiveGroup && !! isaLastUpdated;
+
+	return (
+		<div
+			className={ classNames( styles.hero, {
+				[ styles[ 'fade-in' ] ]: fadeProp,
```
haqadn: I think the fade-in effect can be achieved purely with CSS using animation. So, I don't feel very comfortable involving the state in this. Can you try replacing it with [CSS animation](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_animations/Using_CSS_animations)?

haqadn: Here is an implementation just in case: https://github.com/animate-css/animate.css/blob/main/source/fading_entrances/fadeIn.css

projects/plugins/boost/app/assets/src/js/react-components/recommendation-context/index.tsx
```diff
@@ -0,0 +1,18 @@
+import { IconTooltip } from '@automattic/jetpack-components';
+import { ImageCdnRecommendation } from '../image-cdn-recommendation';
+import styles from './styles.module.scss';
+
+export const RecommendationContext = () => {
```
haqadn: Considering it is a simple usage of the IconTolltip, I think we can skip putting it in a separate component and place the IconTolltip directly in `Hero`.

projects/plugins/boost/app/assets/src/js/react-components/image-size-analysis/other-group-context/index.tsx
```diff
@@ -0,0 +1,29 @@
+import { IconTooltip } from '@automattic/jetpack-components';
+import { __ } from '@wordpress/i18n';
+import styles from './styles.module.scss';
+
+const OtherGroupContext = () => {
```
haqadn: I think we do not need this component either, just like `RecommendationContext`. But, I think you are using it as a stop-gap until `MultiProgress` is refactored into a react component, in which case, it is ok.

projects/plugins/boost/app/assets/src/js/react-components/score-context/index.tsx
```diff
@@ -0,0 +1,57 @@
+import { IconTooltip } from '@automattic/jetpack-components';
+import { __ } from '@wordpress/i18n';
+import styles from './styles.module.scss';
+
+const ScoreContext = () => {
```
haqadn: Same as `RecommendationContext` and `OtherGroupContext`.

haqadn (changes requested):
This covers a lot and it looks like it's working well. My recommendation is to organize the files in the proposed structure in the p2. And, I think non-reusable components like Tooltips can live in their parent component folder.

I have noticed a small issue in styles, the icon placement is wrong:

This PR:
image

Trunk:
image

haqadn:
I think this also breaks an e2e test since `IconTooltip` from RNA needs a click instead of a hover.

projects/plugins/boost/app/assets/src/js/react-components/score-context/index.tsx
dilirity: This will be removed when refactoring MultiProgress, as it holds styles. I've added a todo in MultiProgress about this.

projects/plugins/boost/app/assets/src/js/react-components/image-size-analysis/other-group-context/index.tsx
dilirity: This will be moved when refactoring MultiProgress, as it holds styles. I've added a todo in MultiProgress about this.

projects/plugins/boost/app/assets/src/js/react-components/score-context/index.tsx
dilirity: I've also included a todo for this one as well.

projects/plugins/boost/app/assets/src/js/react-components/recommendation-context/index.tsx
dilirity: This makes sense - removed!

projects/plugins/boost/app/assets/src/js/react-components/image-size-analysis/hero/index.tsx
dilirity: Oh wow! I tried replicating the svelte way via JS and didn't think to use CSS. Thanks!

haqadn (changes requested):
A couple of suggestions related to organization:

1. Instead of  `react-components/image-size-analysis/hero` we can have `react-components/sections/isa-hero`, which conforms to the flat structure.
2. `react-components/components/score-context` seems only useful for the future "score" section. In which case, it should live under something like `react-components/sections/score/context-tooltip.tsx`.

 
My thoughts regarding sections and components are: 
sections = Single use parts of UI
components = Generic UI elements which can be used by multiple sections
pages = What we mount as a route in navigation

What do you think?

dilirity:
@haqadn I've addressed the suggestions - thank you for those. I'll leave my comments about the structure in the P2, so we have the conversation in one place.

haqadn (approved):
:shipit:

PR #33890, opened 2023-10-31
[RNMobile] Fix broken block registration

SiobhyB:
Fixes a regression in the Jetpack app that followed the changes in https://github.com/Automattic/jetpack/pull/32697. With this fix, it's possible to add Jetpack blocks within the apps again.  

Gutenberg Mobile PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/6342

## Proposed changes:

* The `registerJetpackBlockFromMetadata` function has been taken from [the web's register-jetpack-block.js file](https://github.com/Automattic/jetpack/blob/f043df7738f5c7d28b4a2f6affd9e8812e314fdd/projects/plugins/jetpack/extensions/shared/register-jetpack-block.js#L80-L98) add [added to register-jetpack-block.native.js](https://github.com/Automattic/jetpack/blob/f043df7738f5c7d28b4a2f6affd9e8812e314fdd/projects/plugins/jetpack/extensions/shared/register-jetpack-block.native.js) for use in the apps.
* Some tweaks have been made to ensure the function works as expected with the apps, including the use of a custom function (`SvgXml` from [the react-native-svg library](https://github.com/software-mansion/react-native-svg)) for compatibility with React Native.

### Other information:

## Jetpack product discussion

N/A

No, it does not.

## Testing instructions:

* With these changes checked out, navigate to the editor in the iOS or Android app.
* Select + to insert a new block, verify that the following Jetpack blocks are visible: 
  * Contact Info
  * Story
  * Tiled Gallery (dev-environments only)
* Verify that it's possible to add each block and that any edits are saved as expected.

| Before  | After |
|  |  |


anomiex:
Build is failing because `react-native-svg` does not resolve in the non-native build. You may need to add a `.native.js` copy of the file or something.

projects/js-packages/shared-extension-utils/src/get-block-icon-from-metadata.native.js
```diff
@@ -0,0 +1,33 @@
...
+ * This prevents us from duplicating the markup in various places.
+ *
+ * Note: using an `img` tag and passing the SVG markup as a data URI doesn't allow us to
+ * dynamically set the icon color later on.
+ *
+ * @param {object} metadata - Block.json content
+ * @returns {JSX.Element} Icon component
+ */
+export function getBlockIconComponent( metadata ) {
+	return <SvgXml xml={ metadata.icon } />;
```
anomiex: What happens if the block has no `metadata.icon`, or has it set to a non-svg like instagram-gallery currently does?
https://github.com/Automattic/jetpack/blob/6960749e172aba929224b545fcb6461ed2423117/projects/plugins/jetpack/extensions/blocks/instagram-gallery/block.json#L11

anomiex:
You'll need to address the failing Linter test (running `tools/fixup-project-versions.sh` should make the necessary changes). All the E2E test failures look like known issues, you can ignore them.

I don't know enough about the apps to test or review that part of it myself though.

projects/js-packages/shared-extension-utils/src/get-block-icon-from-metadata.native.js
SiobhyB: @anomiex, good point! @fluiddot has suggested another approach via Slack that I'm planning to follow instead of this. I'll work on this and set as ready for review again when it's ready. (Just juggling a few tasks today, sorry for not being super responsive!)

SiobhyB: Noting that, to address this, in https://github.com/WordPress/gutenberg/pull/55742, a change has been made directly to [`components/block-icon/index.native.js`](https://github.com/WordPress/gutenberg/pull/55742/files#diff-ba3bb0b5d33390dd479b7b3eff6a656890a89fbcf11c7fd84e073968447a0bef) so that the parsing is handled on the Gutenberg side.

projects/js-packages/shared-extension-utils/src/get-block-icon-from-metadata.js
```diff
@@ -42,15 +42,19 @@ export function getBlockIconComponent( metadata ) {
...
+ * This function also sets the color of the icon.
  *
  * @param {object} metadata - Block.json content
  * @returns {object} Icon property for client registration
  */
 export function getBlockIconProp( metadata ) {
+	const isNative = Platform.isNative;
 	return {
-		src: getBlockIconComponent( metadata ),
+		src: isNative ? metadata.icon : getBlockIconComponent( metadata ),
```
fluiddot: Alternatively, we could create the native version of this file:

```
import getIconColor from './get-icon-color';

/**
 * Provides needed icon props to render the block icon, including the color.
 *
 * @param {object} metadata - Block.json content
 * @returns {object} Icon property for client registration
 */
export function getBlockIconProp( metadata ) {
	return {
		src: metadata.icon,
		foreground: getIconColor(),
	};
}
```

This option might give us more flexibility in the future, plus reduce friction in case there are changes in the web version that affect the native one. WDYT?

projects/plugins/jetpack/extensions/shared/register-jetpack-block.native.js
```diff
@@ -14,43 +19,87 @@ function requiresPlan( unavailableReason, details ) {
...
+	childBlocks = [],
+	prefix = true
+) {
+	const name = typeof nameOrMetadata === 'string' ? nameOrMetadata : nameOrMetadata.name;
+	const isNamePrefixed = name.startsWith( JETPACK_PREFIX );
+	const rawName = isNamePrefixed ? name.slice( JETPACK_PREFIX.length ) : name;
+
+	const { available, details, unavailableReason } = getJetpackExtensionAvailability( rawName );
 
 	const requiredPlan = requiresPlan( unavailableReason, details );
```
fluiddot, line 42: **Discrepancy between native and web version:** In the web version, we are using `requiresPaidPlan` instead of `requiresPlan`.

https://github.com/Automattic/jetpack/blob/eebad7d58da476499ccc090d3b78c02547918c33/projects/plugins/jetpack/extensions/shared/register-jetpack-block.js#L35

AFAIK we are not filtering yet blocks in the native version depending on the plan, so probably this it not an actual issue now.

projects/plugins/jetpack/extensions/shared/register-jetpack-block.native.js
```diff
@@ -14,43 +19,87 @@ function requiresPlan( unavailableReason, details ) {
...
 		}
 		return false;
 	}
 
-	const result = registerBlockType( `jetpack/${ name }`, {
+	const prefixedName = jpPrefix + rawName;
+	const result = registerBlockType( nameOrMetadata === 'object' ? nameOrMetadata : prefixedName, {
 		...settings,
-		title: betaExtensions.includes( name ) ? `${ settings.title } (beta)` : settings.title,
+		title: betaExtensions.includes( prefixedName ) ? `${ settings.title } (beta)` : settings.title,
```
fluiddot, line 58: **Discrepancy between native and web version:** In the web version, the beta suffix in the title is added using WP hooks. Eventually, we adopt this approach to remove this line.

https://github.com/Automattic/jetpack/blob/eebad7d58da476499ccc090d3b78c02547918c33/projects/plugins/jetpack/extensions/shared/register-jetpack-block.js#L49-L52

https://github.com/Automattic/jetpack/blob/3178c7f15114d90cc91f31332eef9cf060735f1d/projects/plugins/jetpack/extensions/editor.js#L81-L106

projects/plugins/jetpack/extensions/shared/register-jetpack-block.native.js
```diff
@@ -14,43 +19,87 @@ function requiresPlan( unavailableReason, details ) {
...
+ * @param {object} settings - See registerJetpackBlock.
+ * @param {object} childBlocks - See registerJetpackBlock.
+ * @param {boolean} prefix - See registerJetpackBlock.
+ * @returns {object|boolean} Either false if the block is not available, or the results of `registerBlockType`
+ */
+export function registerJetpackBlockFromMetadata( metadata, settings, childBlocks, prefix ) {
+	const mergedSettings = {
+		...settings,
+		icon: getBlockIconProp( metadata ),
+		title: metadata.title,
```
fluiddot, line 90: **Discrepancy between native and web version:** I noticed that the web version is not adding `title` here. Following https://github.com/Automattic/jetpack/pull/33890/commits/f043df7738f5c7d28b4a2f6affd9e8812e314fdd, I assume we need it in the native version. Is this accurate?

https://github.com/Automattic/jetpack/blob/eebad7d58da476499ccc090d3b78c02547918c33/projects/plugins/jetpack/extensions/shared/register-jetpack-block.js#L81-L85

fluiddot (commented):
@SiobhyB I encountered an error when adding a Story block:

Screenshot 2023-11-03 at 15 50 41

Seems we'd need to support SVG XML string sources directly in the [`Icon` component](https://github.com/WordPress/gutenberg/blob/a3802ba85a63cdbaa96a5b5636e22624ca2828bc/packages/components/src/icon/index.tsx), as the block icon might be used in more places apart from the `BlockIcon` component:

https://github.com/Automattic/jetpack/blob/eebad7d58da476499ccc090d3b78c02547918c33/projects/plugins/jetpack/extensions/blocks/story/edit.native.js#L25

https://github.com/Automattic/jetpack/blob/eebad7d58da476499ccc090d3b78c02547918c33/projects/plugins/jetpack/extensions/blocks/story/edit.native.js#L175-L176

projects/js-packages/shared-extension-utils/src/get-block-icon-from-metadata.js
SiobhyB: I don't feel overly strongly, but my thought process behind this approach was that it may help avoid the impact of changes on native being overlooked. I'm a bit worried that the native files were overlooked when the changes to block registration happened, so thought that trying to keep native logic in the main files when possible could help mitigate that in the future.

I'm open to separating this logic out, however.

fluiddot: Yep, good point. Definitely, each approach has its benefits and drawbacks. In terms of making more prominent the potential impact on the native version when introducing changes, I'd rather flag them via unit/E2E tests. Since we don't have them set in the Jetpack repository, sharing code in files might be our best option. That said, I agree with continuing with your approach 👍 .

projects/plugins/jetpack/extensions/shared/register-jetpack-block.native.js
SiobhyB: That's a good spot, I see that there's a related `requiresPlan` function in the native file, but don't believe it's currently doing much:

https://github.com/Automattic/jetpack/blob/eebad7d58da476499ccc090d3b78c02547918c33/projects/plugins/jetpack/extensions/shared/register-jetpack-block.native.js#L12-L18

As we're not changing either of these lines with this PR, I'm hesitant to tweak them, as I don't want to inadvertently expand the scope of this fix. I've opened a separate GitHub issue, though, to track the issue of filtering blocks based on plan: https://github.com/wordpress-mobile/gutenberg-mobile/issues/6353

Let me know if you think that addresses the problem you've raised here!

projects/plugins/jetpack/extensions/shared/register-jetpack-block.native.js
SiobhyB: @fluiddot, yes, I noticed that the title is only passed as part of the `metadata` object for native, and not with the `settings` object. I haven't been able to pinpoint the exact reason for the difference, however. 🤔

projects/plugins/jetpack/extensions/shared/register-jetpack-block.native.js
SiobhyB: Thanks Carlos, I note you've mentioned this could be a future improvement. Do you think there's anything we need to change here in the context of this PR? Or should we track this somewhere as a future improvement?

SiobhyB:
@fluiddot, thank you for the thorough review here!

> Seems we'd need to support SVG XML string sources directly in the [Icon component](https://github.com/WordPress/gutenberg/blob/a3802ba85a63cdbaa96a5b5636e22624ca2828bc/packages/components/src/icon/index.tsx), as the block icon might be used in more places apart from the BlockIcon component:

I have a sense we should stick to parsing the icon in [get-block-icon-from-metadata.native.js](https://github.com/Automattic/jetpack/pull/33890/files#diff-ae9a90c8ac89c759fb202a7e41fdf6f8c741f9e81b3e3186c37660b79133c859) in the Jetpack codebase. The reason behind my thinking is that there could be a number of places where the icon could be passed, and the equivalent web components haven't been updated to account for this in Gutenberg. Do you think that seems sound? I've made some changes that appear to work for our use case, and also seem tweaks in https://github.com/Automattic/jetpack/pull/33890/commits/1a127671742c8a8729d6180702bf046340223611 to account for other potential issues mentioned in this review. Looking forward to hearing what you think.

projects/js-packages/shared-extension-utils/src/get-block-icon-from-metadata.native.js
SiobhyB: Following the discovery that the icon can be passed to components other than `<BlockIcon/>`, I reverted the aforementioned change, so we're back to parsing the icon in `get-block-icon-from-metadata.native.js`. 

To account for cases where there may not be a block icon, or it may not be an svg string, I made some changes in https://github.com/Automattic/jetpack/pull/33890/commits/1a127671742c8a8729d6180702bf046340223611. I believe these changes bring us in alignment to how the logic works in the web file.

I used the `instagram` example to test that this works locally:

Screenshot_20231106_172034

projects/js-packages/shared-extension-utils/src/get-block-icon-from-metadata.native.js
```diff
@@ -0,0 +1,37 @@
+/* eslint-disable jsdoc/no-undefined-types */
+import { SvgXml } from 'react-native-svg';
```
fluiddot: @SiobhyB My only concern about importing `react-native-svg` is that it's not a dependency in the Jetpack repository. In the future, if we run unit tests for the native version in this repository, this might be an issue as this package won't be exposed and tests might fail.

One option we could explore is to export `SvgXml` as part of the primitives of SVG and the Gutenberg components package.

* https://github.com/WordPress/gutenberg/blob/trunk/packages/primitives/src/svg/index.native.js

This way we could import `SvgXml` here via the `@wordpress/primitives`. Strictly speaking, `SvgXml` is not a primitive of SVG, but we could consider it as this due to be the only solution for processing a SVG coming from string in the native version.

projects/plugins/jetpack/extensions/shared/register-jetpack-block.native.js
fluiddot: > As we're not changing either of these lines with this PR, I'm hesitant to tweak them, as I don't want to inadvertently expand the scope of this fix. I've opened a separate GitHub issue, though, to track the issue of filtering blocks based on plan: https://github.com/wordpress-mobile/gutenberg-mobile/issues/6353
> 
> Let me know if you think that addresses the problem you've raised here!

Oh, I didn't mean to imply that we should fix it here, I just wanted to share the discrepancies I found when comparing both versions in case there was something we'd need to follow up with. Tracking this in a GitHub issue sounds good, thanks 🙇 !

projects/plugins/jetpack/extensions/shared/register-jetpack-block.native.js
fluiddot: > [...] Do you think there's anything we need to change here in the context of this PR? Or should we track this somewhere as a future improvement?

My comment was merely a heads-up that this changed in the web version.  So I think there's nothing to apply in this PR related to this. As you pointed out, we could consider tracking it in a new GitHub issue. I'll create one as a follow-up.

fluiddot: **To close the loop:** I created https://github.com/wordpress-mobile/gutenberg-mobile/issues/6357 as a follow-up

projects/plugins/jetpack/extensions/shared/register-jetpack-block.native.js
fluiddot: That's interesting. I presume the blocks' title might be processed differently in the native version so we need to explicitly include it here. I don't foresee this as an important divergence between versions, so we can keep it as-is for now. In the future, if we want to unify the Jetpack blocks registration logic, we could consider revisiting this part.

projects/js-packages/shared-extension-utils/src/get-block-icon-from-metadata.js
SiobhyB: For posterity, sharing that I went ahead to create a separate `get-block-icon-from-metadata.native.js` file, as we started to make more native-specific changes that I thought could have caused confusion if kept in the same file.

projects/js-packages/shared-extension-utils/src/get-block-icon-from-metadata.native.js
SiobhyB: > My only concern about importing react-native-svg is that it's not a dependency in the Jetpack repository. In the future, if we run unit tests for the native version in this repository, this might be an issue as this package won't be exposed and tests might fail.

@fluiddot, thank you for sharing, I hadn't considered that concern, and can see how it could be viewed as a blocker now.

I like your suggestion of exporting `SvgXml` via `@wordpress/primitives`, as I do think the function could be considered relevant to that package. I've gone ahead to implement that in this PR at https://github.com/Automattic/jetpack/pull/33890/commits/647795fb67a69ed5eb93b5b497c28f26be6804dc and in Gutenberg at https://github.com/WordPress/gutenberg/pull/55742/commits/49f94c4acc20135c111c529583f00871d166f84e. Looking forward to hearing what you think!

projects/js-packages/shared-extension-utils/src/get-block-icon-from-metadata.native.js
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Generate an icon as a React component from the SVG markup defined in a block.json metadata file.
+ * This prevents us from duplicating the markup in various places.
+ *
+ * Note: using an `img` tag and passing the SVG markup as a data URI doesn't allow us to
+ * dynamically set the icon color later on.
+ *
+ * @param {object} metadata - Block.json content
+ * @returns {JSX.Element} Icon component
```
fluiddot: Nitpick: Since the function can return an empty string, we'd need to reflect it in the JSDoc.
```suggestion
 * @returns {JSX.Element|string} Icon component
```

fluiddot (dismissed):
LGTM 🎊 ! Thanks @SiobhyB for addressing this issue 🙇 !

I noticed that one of the E2E tests failed (`Jetpack blocks e2e tests`), the failure seems unrelated to these changes so I went ahead and restart it.

projects/js-packages/shared-extension-utils/src/get-block-icon-from-metadata.native.js
SiobhyB: Thank you, I've committed this change, it just looks like you'll need to re-approve the PR following the commit. 🙇‍♀️

fluiddot (approved):
LGTM 🎊 !

PR #33891, opened 2023-10-31
AI Aissistant: introduce tiers definition

CGastrell:
Add tiers definition for consumption

## Proposed changes:
This PR adds the base files for tiers definition for later consumption on upgrade and checkout processes.

Tiers are defined as 

- TIER_AI_FREE
- TIER_AI_MONTHLY_100
- TIER_AI_MONTHLY_200
- TIER_AI_MONTHLY_500
- TIER_AI_UNLIMITED

and carry basic information:
- slug
- limit
- readableLimit (optional, for UI usage when limit is something needing translation)

A simple `getNextTierByUsage` is also available, receiving a usage and returning the next tier that would fit the needs of the user.

NOTE: we could solve all things internally by using `useAIFeature.requestsCount` instead of receiving params, still a PoC

### Other information:

## Jetpack product discussion
https://github.com/orgs/Automattic/projects/667/views/14

No

## Testing instructions:
There is no noticeable change, this PR is just introductory for the base files needed and should mostly work as a PoC.
TBD

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/tiers/index.ts
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/tiers/index.ts
@@ -7,32 +7,38 @@ export type AiAssistantTier = {
 	slug: string;
 	limit: number;
 	readableLimit?: string;
+	value: number;
 };
 
 const TIER_AI_FREE: AiAssistantTier = {
 	slug: 'ai-assistant-tier-free',
 	limit: 20,
+	value: 0,
 };
 
 const TIER_AI_MONTHLY_100: AiAssistantTier = {
 	slug: 'ai-assistant-tier-100',
 	limit: 100,
+	value: 100,
 };
 
 const TIER_AI_MONTHLY_200: AiAssistantTier = {
 	slug: 'ai-assistant-tier-200',
 	limit: 200,
+	value: 200,
 };
 
 const TIER_AI_MONTHLY_500: AiAssistantTier = {
 	slug: 'ai-assistant-tier-500',
 	limit: 500,
+	value: 500,
 };
 
 const TIER_AI_UNLIMITED: AiAssistantTier = {
 	slug: 'ai-assistant-tier-unlimited',
 	limit: Infinity,
 	readableLimit: __( 'unlimited', 'jetpack' ),
+	value: 1,
 };
 
 const TIERS = [
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/tiers/index.ts
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/tiers/index.ts
@@ -1,0 +1,55 @@
+/**
+ * External dependencies
+ */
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/tiers/index.ts
```diff
@@ -0,0 +1,55 @@
...
+];
+
+/**
+ * Function getNextTierByUsage
+ * Given a current usage, returns the next TIER whose limit is higher than the current usage.
+ *
+ * @param {number} usage - The current usage.
+ * @returns {AiAssistantTier} The next AiAssistantTier whose limit is higher than the current usage.
+ */
+export function getNextTierByUsage( usage: number ): AiAssistantTier {
```
lhkowalski, line 59: Maybe instead of usage, we should use the current plan the site has, like a `getNextTierByCurrentTier`. The site may have the 200 plan and used 73 requests, but the upgrade path should lead them to the 500 plan in this case.

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/tiers/index.ts
lhkowalski, line 1: I think this definition is important. Like you said in the PR, we could rely on the `current-tier.count` value and go from it, but it's smart when we have some functions abstracting away the details.

projects/plugins/jetpack/extensions/blocks/ai-assistant/lib/tiers/index.ts
CGastrell: What would be the use case for that? If we're showing an upgrade nudge it's because the feature is requiring an upgrade.

Agree that the current tier should be available, and we should keep adding methods there to manage all scenarios/needs

CGastrell:
@lhkowalski overall the idea is that this support files land as a way to work on the UI with the tiered plans. We might remove them in the future (shall we get them from backend) or keep adding functions (getBySlug, getNext, getPrev...etc). Also, once the real plans are in place, we'll adjust these to match them

retrofox (commented):
I'm not convinced about redefining the client. Ideally, it should be defined and provided from a single place on the server side. Duplicate these definitions will add extra complexity, making the development and maintenance process harder IMO.

Ideally, from my PoV, we should collect these data from the server, and simply propagate them among all pieces of the app.

CGastrell:
> I'm not convinced about redefining the client. Ideally, it should be defined and provided from a single place on the server side. Duplicate these definitions will add extra complexity, making the development and maintenance process harder IMO.
> 
> Ideally, from my PoV, we should collect these data from the server, and simply propagate them among all pieces of the app.

Sounds as good to me. I can move this to the backend and make it be sent with either the AI data or some other endpoint. But however we do it, we'll need it available on the block, the interstitials and maybe some other places. Do you think I should add it to `Jetpack_Editor_Initial_State['ai-assistant']`?

CGastrell:
Closed as we're moving to D127450-code

PR #33892, opened 2023-10-31
Poc/tiered upgrade nudge

CGastrell:
PoC for tiered upgrades.

Follow up for #33891 

## Proposed changes:
This PR uses the tiers to build the corresponding upgrade nudge.
We're still lacking some _current plan_ information and a way to deal with a migration from regular plans to tiered. Some of this information should/could be stored and retrieved only from backend, which we might end up doing once the backend implementation is ready.
image

### Other information:

No

## Testing instructions:
Use a free plan and reach the requests limit. See the upgrade nudge on the block, the button should now read some "Upgrade to X requests". Reaching each tier limit should suggest the next tier based on usage. (100, 200, 500, unlimited)

### DIFF
--- a/projects/plugins/jetpack/changelog/add-ai-lib-tiers
+++ b/projects/plugins/jetpack/changelog/add-ai-lib-tiers
@@ -1,0 +1,4 @@
+Significance: minor
+Type: other
+
+Add AI tiers definition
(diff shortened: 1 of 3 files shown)


lhkowalski:
I saw the implementation and was going to say that we need to be able to support the current upgrade flow (from free to unlimited) while we don't pull the trigger and release the new plans, but then started to think a way of doing it and it's not possible using the `ai-assistant-feature` payload data. I think we need a beta feature flag for the tiered plan implementation, the same way we did for the Usage Panel implementation. This way we can implement everything behind this feature and then change it when it's ready.

PR #33897, opened 2023-10-31
Subscribe Block: Introduce "Subscribed" state

lezama:
Fixes #

## Proposed changes:

Improve the user interaction with the Subscribe block, provide feedback when they are subscribed.

Screenshot 2023-10-31 at 15 55 59

Before:

Screenshot 2023-11-01 at 16 13 42

### Other information:

No

## Testing instructions:

* Check that the Subscribe block reflects the "Subscribed" state when a user is subscribed.
* Check the Subscribe state under different styles and variations of the block.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -553,29 +553,29 @@ function render_block( $attributes ) {
 	$is_paid_subscriber       = get_attribute( $attributes, 'isPaidSubscriber', false );
 
 	$data = array(
-		'widget_id'              => Jetpack_Subscriptions_Widget::$instance_count,
-		'subscribe_email'        => $subscribe_email,
+		'widget_id'                     => Jetpack_Subscriptions_Widget::$instance_count,
+		'subscribe_email'               => $subscribe_email,
 
-		'wrapper_attributes'     => get_block_wrapper_attributes(
+		'wrapper_attributes'            => get_block_wrapper_attributes(
 			array(
 				'class' => $classes['block_wrapper'],
 			)
 		),
-		'subscribe_placeholder'  => get_attribute( $attributes, 'subscribePlaceholder', esc_html__( 'Type your email…', 'jetpack' ) ),
-		'submit_button_text'     => get_attribute( $attributes, 'submitButtonText', $is_paid_subscriber ? esc_html__( 'Upgrade', 'jetpack' ) : esc_html__( 'Subscribe', 'jetpack' ) ),
-		'subscribed_button_text' => get_attribute( $attributes, 'subscribedButtonText', esc_html__( 'Subscribed', 'jetpack' ) ),
-		'success_message'        => get_attribute(
+		'subscribe_placeholder'         => get_attribute( $attributes, 'subscribePlaceholder', esc_html__( 'Type your email…', 'jetpack' ) ),
+		'submit_button_text'            => get_attribute( $attributes, 'submitButtonText', $is_paid_subscriber ? esc_html__( 'Upgrade', 'jetpack' ) : esc_html__( 'Subscribe', 'jetpack' ) ),
+		'submit_button_text_subscribed' => get_attribute( $attributes, 'submitButtonTextSubscribed', esc_html__( 'Subscribed', 'jetpack' ) ),
+		'success_message'               => get_attribute(
 			$attributes,
 			'successMessage',
 			esc_html__( "Success! An email was just sent to confirm your subscription. Please find the email now and click 'Confirm Follow' to start subscribing.", 'jetpack' )
 		),
-		'show_subscribers_total' => (bool) get_attribute( $attributes, 'showSubscribersTotal' ),
-		'subscribers_total'      => get_subscriber_count( $include_social_followers ),
-		'referer'                => esc_url_raw(
+		'show_subscribers_total'        => (bool) get_attribute( $attributes, 'showSubscribersTotal' ),
+		'subscribers_total'             => get_subscriber_count( $include_social_followers ),
+		'referer'                       => esc_url_raw(
 			( is_ssl() ? 'https' : 'http' ) . '://' . ( isset( $_SERVER['HTTP_HOST'] ) ? wp_unslash( $_SERVER['HTTP_HOST'] ) : '' ) .
 			( isset( $_SERVER['REQUEST_URI'] ) ? wp_unslash( $_SERVER['REQUEST_URI'] ) : '' )
 		),
-		'source'                 => 'subscribe-block',
+		'source'                        => 'subscribe-block',
 	);
 
 	if ( ! jetpack_is_frontend() ) {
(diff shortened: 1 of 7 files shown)


projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -563,6 +563,7 @@ function render_block( $attributes ) {
 		),
 		'subscribe_placeholder'  => get_attribute( $attributes, 'subscribePlaceholder', esc_html__( 'Type your email…', 'jetpack' ) ),
 		'submit_button_text'     => get_attribute( $attributes, 'submitButtonText', $is_paid_subscriber ? esc_html__( 'Upgrade', 'jetpack' ) : esc_html__( 'Subscribe', 'jetpack' ) ),
+		'subscribed_button_text' => get_attribute( $attributes, 'subscribedButtonText', esc_html__( '✓ Subscribed', 'jetpack' ) ),
```
simison: Thinking if this should be `'✓ ' . esc_html__( 'Subscribed', 'jetpack' )` ?

lezama: if we are going to allow editing this in the future it could be good to add the flexibility of allowing to remove the ✓

simison: I agree — we should include the ✓ in string they can edit, but for translation purposes I'm not sure if it makes a difference to send it for translations? Seems just something translators could get wrong.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -723,7 +757,8 @@ class="<?php echo esc_attr( $classes['submit_button'] ); ?>"
 			</div>
 		</div>
 	</div>
-	<?php
+		<?php
+	}
```
simison: I think this end should be higher up before `show_subscribers_total` check, so that we keep showing that extra line if they've enabled it in the block settings.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -624,7 +625,40 @@ function render_for_website( $data, $classes, $styles ) {
 			'success_message' => $data['success_message'],
 		)
 	);
-	?>
+
+	if ( Jetpack_Memberships::user_can_view_post() ) { ?>
+		<div <?php echo wp_kses_data( $data['wrapper_attributes'] ); ?>>
+		<div class="jetpack_subscription_widget">
+			<div class="wp-block-jetpack-subscriptions__container">
```
simison: Can everything up to here be outside the if so that we don't need to duplicate it? It's shared across both sides of the if/else statement.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -624,7 +625,40 @@ function render_for_website( $data, $classes, $styles ) {
...
+				<div class="wp-block-jetpack-subscriptions__form-elements">
+					<p id="subscribe-submit"
```
simison: What do we need the `id` for?

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
dlind1: Including the checkmark in the string would allow changing it in different languages and also adjust its placement for RTL, but also introduces the danger of it being omitted/corrupted in some way. Since the checkmark is at the beginning of the word, it'd work well with RTL, and it probably shouldn't be changed to another symbol.

Given that I'd suggest extracting the checkmark outside of the string, as an exception to our "avoid string concatenation" rule 🙂

simison: So let's do this?
```php
$subscribed = esc_html__( 'Subscribed', 'jetpack' );

is_rtl() ? $subscribed . ' ✓' : '✓ ' . $subscribed
```
? :-) 

It's a pain to find the checkmark from keyboards/somewhere if you don't know what to do, and they might copy the wrong one because there are a few.

dlind1: Just `$subscribed = '✓ ' . esc_html__( 'Subscribed', 'jetpack' );` should work and the checkmark would be displayed correctly for RTL as long as `direction: rtl` is set.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
lezama: I think the text says something like "Join other x subscribers", that's why I left it behind

simison: 👍 Fine for this PR but let's find a way to bring it in, with copy changes perhaps?

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
lezama: ok! Done https://github.com/Automattic/jetpack/pull/33897/commits/1be309ea9fb3e3408c315c78b6ed3a1b2e803934

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
lezama: 😅 it's used on [styles here](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/jetpack/extensions/blocks/subscriptions/view.scss#L77), we definitely shouldn't be using id for this. Let's change that in a different PR?

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
lezama: https://github.com/Automattic/jetpack/pull/33897/commits/a8be5d5d8bd15a6952ad83a6045cbc184eb90bbf

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
simison: Yep sounds good!

simison (dismissed):
Works on subscriber-only post:

image

Doesn't work on homepage's block:

image

Paid post paywall shows subscribe button — next we should update this to "Become paid subscriber" button without input:

image

Was surprised the block in the footer didn't show "subsceribed",  tho.

simison:
I think we could let submit the form from the "subscribed" button at least on paid sites — that way they see the paid options:

image

→

image

millerf (dismissed):
LGTM.

Could we just keep the same approach as before and use:

```
/**
	 * Get the token payload .
	 *
	 * @return false|array.
	 */
	public function get_token_payload() {
		$token = $this->get_and_set_token_from_request();
		return $this->decode_token( $token );
	}
	```
Your code will still work as-is, but it seems cleaner...

lezama:
> I think we could let submit the form from the "subscribed" button at least on paid sites — that way they see the paid options:

✅ took me some time but it does it now

simison (approved):
This looks fantastic and works smoothly now! Rebased locally before testing.

Tested homepage, free post, footer, paid post, subscribers-only post.

Something to iterate on is showing "Become paid subscriber" instead of "Subscribed" especially in paid paywalls:

Screenshot 2023-11-03 at 13 53 26

Also "verify your email" in paid posts when I subscribed to become free subscriber needs attention:

Screenshot 2023-11-03 at 13 53 35

PR #33900, opened 2023-10-31
Subscriptions: Add param for magic login

edanzer:
## Proposed changes:

* Depends on this diff being deployed first: D127286-code
* Sets `redirect_to` param when calling wpcom_logmein_redirect_url(). This will result in the blog_id being added to the url for the magic login screen. For Jetpack sites, we'll just use the site_id param. 

### Other information:

## Jetpack product discussion

p9Jlb4-9mo-p2

No

## Testing instructions:

**Test on simple site:**
1) Load this phab diff in your sandbox
2) While keeping phab diff in place, run the bin command below to load this branch in your sandbox
3) Sandbox a simple test site
4) On your simple site, publish a post with a paywall block that is for subscribers only. 
5) In an incognito or logged out browser, open the paywalled post on the frontend. Click the link for already a subscriber / login.
6) You will be redirected to the magic login screen. Confirm that the url has the blog_id param set.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -955,7 +955,7 @@ function get_paywall_blocks( $newsletter_access_level ) {
 	} else {
 		if ( ( new Host() )->is_wpcom_simple() ) {
 			// custom domain
-			$sign_in_link = wpcom_logmein_redirect_url( get_current_url(), false, null, 'link', get_current_blog_id() );
+			$sign_in_link = wpcom_logmein_redirect_url( get_current_url(), false, get_current_url(), 'link' );
 		} else {
 			$sign_in_link = add_query_arg(
 				array(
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -955,7 +955,7 @@ function get_paywall_blocks( $newsletter_access_level ) {
 	} else {
 		if ( ( new Host() )->is_wpcom_simple() ) {
 			// custom domain
-			$sign_in_link = wpcom_logmein_redirect_url( get_current_url(), false, null, 'link' );
+			$sign_in_link = wpcom_logmein_redirect_url( get_current_url(), false, get_current_url(), 'link', get_current_blog_id() );
```
lezama: we don't need to pass the second `get_current_url()` anymore,right?

lezama (dismissed):

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
edanzer: Correct. Will remove shortly.

lezama (approved):

PR #33894, opened 2023-10-31
Jetpack AI: Use usage helper to set current tier value

lhkowalski:
Part of #33792.

## Proposed changes:

* Use the Jetpack AI Usage Helper to set the `current-tier.value` on the the `ai-assistant-feature` response, so it's ready when the new plans come alive

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* We need to check if the response is still the same as before
* Apply this change to your sandbox and sandbox the `public-api` endpoint
* Go to the WordPress.com Console and do a request to the `wpcom/v2` `/sites/$wpcom_site/jetpack-ai/ai-assistant-feature` endpoint:
* If the site used for the test has the paid plan, you should see `current-tier.value = 1`:

Screenshot 2023-10-31 at 15 22 11

* If the site used for the test does not have the paid plan, you should see `current-tier.value = 0`:

Screenshot 2023-10-31 at 15 23 04

PR #33901, opened 2023-10-31
AI Tiered Plans: pick and expose tier plan data

retrofox:
Follow up of https://github.com/Automattic/jetpack/pull/33894

This PR registers, collect and expose tier plan data from the useAiFeature() custom hook.

Fixes #

## Proposed changes:

* AI Tiered Plans: pick and expose tied plan data

### Other information:

## Testing instructions:

* Go to the block editor
* Open your dev tool of your favorite client (firefox, chrome, etc)
* Select the `Source` tab
* Browse the `use-ai-fearture` folder

Screenshot 2023-10-31 at 18 54 16

* Select the `index.ts` file
* Add a breakpoint just after returning the hook data (click on the L93):

Screenshot 2023-10-31 at 19 02 25

* In the block editor, open the Jetpack AI panel
* The app should stop running
* Mouse over the `data` variable
* Confirm you see the `currentTier` and the `usagePeriod` there

Screenshot 2023-10-31 at 19 05 22

### DIFF
--- a/projects/plugins/jetpack/changelog/update-tiered-plan-update-use-ai-feature
+++ b/projects/plugins/jetpack/changelog/update-tiered-plan-update-use-ai-feature
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
+AI Tiered Plans: pick and expose tied plan data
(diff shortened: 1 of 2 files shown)


CGastrell (approved):
LGTM. One question though: while on free tier, there is no data in usage-period, that means this is not the same prop we use to keep track of 20 free requests?

retrofox:
> LGTM. One question though: while on free tier, there is no data in usage-period, that means this is not the same prop we use to keep track of 20 free requests?

Yes, at least for now we can say it isn't the same data prop. We can use the regular `requestsCount` prop and eventually update the logic in case the data provided by the endpoint changes.

Thanks for the review

PR #33902, opened 2023-10-31
Add/app promo meta data to admin

enejb:
This PR attemts to add a Jetpack App promot to the admin pages that shows up in safari. 

## Proposed changes:
* It does this using the https://developer.apple.com/documentation/webkit/promoting_apps_with_smart_app_banners tech.

### Other information:

## Testing instructions:
TBD

### DIFF
--- a/projects/plugins/jetpack/_inc/lib/admin-pages/class.jetpack-admin-page.php
+++ b/projects/plugins/jetpack/_inc/lib/admin-pages/class.jetpack-admin-page.php
@@ -123,7 +123,7 @@ abstract class Jetpack_Admin_Page {
 		if ( current_user_can( 'jetpack_connect' ) && $connectable ) {
 			$this->add_connection_banner_actions();
 		}
-		add_action( 'wp_head', array( $this, 'add_ios_smart_banner' ) );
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/_inc/lib/admin-pages/class.jetpack-admin-page.php
```diff
@@ -202,6 +203,12 @@ public function admin_styles() {
 		wp_style_add_data( 'jetpack-admin', 'rtl', 'replace' );
 		wp_style_add_data( 'jetpack-admin', 'suffix', $min );
 	}
+	/**
+	 * Adds iOS Smart banner to let folks know about the Jetpack app on iOS.
+	 */
+	public function add_ios_smart_banner() {
```
jeherve, line 209: This feels like the wrong place to add that method, since the meta tag is added to every page in wp-admin.

What do you think about creating a new file loaded in `projects/plugins/jetpack/3rd-party/3rd-party.php` where we would add this, and potentially other things in the future?

I would also recommend adding a check to ensure this only gets added once, because in my tests the meta tag often got added 3 times.

jeherve (changes requested):

PR #33908, opened 2023-11-01
launchpad-task-definitions: Fix PHP 8.x fatal

xknown:
Add an isset check to `wpcom_launchpad_mark_customize_welcome_message_complete_on_add` to avoid PHP 8.x fatals (`TypeError : Cannot access offset of type string on string `) when an invalid `subscription_options` value is stored on the DB.

Fixes #

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*


jeherve:
This was fixed in #33923 and is no longer necessary.

PR #33910, opened 2023-11-01
CRM: 3362 - use modern string manipulation functions

tbradsha:
Resolves Automattic/zero-bs-crm#3362 - use modern string manipulation functions

## Proposed changes:

PHP 8 has some convenient string manipulation functions that are particularly useful when you just need a `bool`:

* `str_starts_with`
* `str_contains`
* `str_ends_with`

WP has had a polyfill for said functions as of v5.9, and now that we support a minimum of WP 6.0, we can use said functions for better code readability. This PR swaps out all instances I could find that could benefit from the above functions.

I also went ahead and did the same in JS codebase, with `substr` and `indexOf` being swapped out for `startsWith`, `includes`, and `endsWith` where it made sense.

Nota bene:
* We frequently used `strpos() > -1`. Thankfully in this scenario if `strpos()` evaluates to `false`, the resulting `false > -1` also evaluated to `false`. I'm not quite not sure the full reason (that operator results in a strict comparison, but `false < -1` evaluates to `true`), but in the end it worked as intended even if a bit sketchy. These instances were replaced with `str_contains()`.
* There are several instances where `strpos()` (no operator) and `strpos() > 0` are used. I suspect in some instances `>=0` was intended, but that would involve careful testing to verify. The swap equivalent would be `str_contains() && ! str_starts_with()`, but I didn't see a big benefit to changing those and thus left them alone for now.
* In some instances PHPCS forced some bigger changes than I was planning. Sorry, reviewer.

I noticed `class-client-portal-router.php` was missing visibility keywords on its methods, so I went ahead and cleaned up that entire file. I also removed some chunks of commented code I ran across while working on this PR.

### Other information:

## Testing instructions:

This is a beast of a PR, and ultimately everything should function as before I've done my best to cluster like changes in each commit, and I recommend reviewing commit-by-commit.

The most likely mistake is probably a missed `!` somewhere.


coder-karen (approved):
I went through each commit, couldn't find any typos. Initially was confused as I saw there are still some `substr` references, but realized the issue is about using it in the right place which is what you've done.

Also spun up a fresh test site with dummy data as well as Woo to test things out, everything looks good.

PR #33929, opened 2023-11-02
Search: use classic search for ios<16 devices

### LINKED ISSUE #33495: Bug: Jetpack Search results show no matches for browsers running on iOS version 14 or 15
### Impacted plugin

Search

### Quick summary

While attempting to use Jetpack Search on a site using an older iOS device (_specifically running iOS versions 14 or 15_), Jetpack Search does not return any results. 

I've tested this for both Safari and Chrome.

**Impact: For anyone running an older iOS device, it will appear as though Jetpack Search is not working.** 

### Steps to reproduce

Using BrowserStack or a physical device running iOS versions 14 or 15, 
Visit the following sites:
- https://jetpack.com/?s=creative

During my testing, I could replicate the issue across both Safari and Chrome in BrowserStack. 

BrowserStack - iPhone 15 (iOS 17) {screenshot below}

As expected Jetpack Search returns results, as the keyword appears in multiple places. 

Jetpack iOS 17 - iPhone 15

### What actually happened

BrowserStack - iPhone 13 (iOS 15) {screenshot below}

On older devices (iOS 14 or 15), Jetpack Search does not return any results. 

Jetpack iOS 15 - iPhone 13

### Impact

All

### Available workarounds?

Simple, Atomic, Self-hosted

### Logs or notes

This bug does not appear to affect iOS versions > 16. 

Could not test iOS versions less than 14.

kangzj:
Fixes #33495 

## Proposed changes:

Enable Classic Search if browser is iOS and version < 16. **A follow up task will be needed to enable the compatibility patch on WPCOM.**

### Other information:

## Jetpack product discussion
n/a

no

## Testing instructions:

* Prepare local / JN site with Instant Search enabled
* Open `/?s=test`
* Ensure search overlay pops up
* Choose a device with iOS < 16 on browserstack <- look at the secret store for account
* Refresh the page
* Ensure classic search returned results too
* Ensure all tests pass

image


bluefuton:
Seems like a reasonable solution given the anticipated increase in bundle size if these older iOS versions are supported (9kb to 25kb): https://github.com/Automattic/jetpack/issues/33495#issuecomment-1760561023

Rather than including our own browser regex to target older iOS, is there any sense in using `browserslist` to generate a regex of browsers not supported by the current configuration? e.g. https://github.com/browserslist/browserslist-useragent-regexp

kangzj:
> Rather than including our own browser regex to target older iOS, is there any sense in using browserslist to generate a regex of browsers not supported by the current configuration? e.g. https://github.com/browserslist/browserslist-useragent-regexp

@bluefuton That's an excellent point! I wanted to target all the unsupported browsers, but didn't mange to find a good way. Looking now!

projects/packages/search/compatibility/unsupported-browsers.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+function enable_classic_search_for_unsupported_browsers( $clasic_search_enabled ) {
+	if ( $clasic_search_enabled ) {
+		return true;
+	}
+	$user_agent = isset( $_SERVER['HTTP_USER_AGENT'] ) ? filter_var( wp_unslash( $_SERVER['HTTP_USER_AGENT'] ) ) : '';
+	// Generated by `browserslist-useragent-regexp`
+	// TODO: make the value re-generated on each build.
+	// !! The regex doesn't really work!!
+	return ! preg_match( '/Edge?\/1{2}[6-8](\.0|)(\.\d+|)|Firefox\/1{2}[78]\.0(\.\d+|)|Chrom(ium|e)\/(109\.0|1{2}[5-8]\.0)(\.\d+|)|Maci.+ Version\/(16\.6|17\.0)([,.]\d+|)( \(\w+\)|)( Mobile\/\w+|) Safari\/|Chrome.+OPR\/10[23]\.0\.\d+|(CPU[ +]OS|iPhone[ +]OS|CPU[ +]iPhone|CPU IPhone OS|CPU iPad OS)[ +]+(15[._][67]|16[._]6|17[._]0)([._]\d+|)|Opera Mini|Android:?[ /-]1{2}7(\.0|)(\.\d+|)|Android.+Chrom(ium|e)\/1{2}7\.0(\.\d+|)|SamsungBrowser\/2{2}\.0/i', $user_agent );
```
kangzj: @bluefuton  If the regex works well then this would be an ideal solution, but it doesn't 🤔

kangzj: I tested with latest Safari and Chrome - they both failed the regex 😢

projects/packages/search/compatibility/unsupported-browsers.php
```diff
@@ -0,0 +1,25 @@
...
+namespace Automattic\Jetpack\Search\Compatibility;
+
+add_filter( 'jetpack_search_classic_search_enabled', __NAMESPACE__ . '\enable_classic_search_for_unsupported_browsers', 10, 1 );
+
+/**
+ * Enable Classic Search when the frontend is not working.
+ *
+ * @param boolean $clasic_search_enabled Whether Classic Search is enabled.
+ */
+function enable_classic_search_for_unsupported_browsers( $clasic_search_enabled ) {
```
bluefuton:
```suggestion
function enable_classic_search_for_unsupported_browsers( $classic_search_enabled ) {
```

projects/packages/search/compatibility/unsupported-browsers.php
```diff
@@ -0,0 +1,25 @@
...
+	if ( $clasic_search_enabled ) {
```
bluefuton:
```suggestion
	if ( $classic_search_enabled ) {
```

projects/packages/search/src/initializers/class-initializer.php
```diff
@@ -112,11 +113,21 @@ protected static function init_search( $blog_id ) {
...
+		$success                   = false;
+		$is_instant_search_enabled = ( new Module_Control() )->is_instant_search_enabled();
+		if ( $is_instant_search_enabled ) {
 			// Enable Instant search experience.
 			$success = static::init_instant_search( $blog_id );
-		} else {
+		}
+
+		/**
+		 * Filter whther classic search should be enabled. By this stage, search module would be enabled already.
```
bluefuton:
```suggestion
		 * Filter whether classic search should be enabled. By this stage, search module would be enabled already.
```

projects/packages/search/src/initializers/class-initializer.php
```diff
@@ -112,11 +113,21 @@ protected static function init_search( $blog_id ) {
...
+		 *
+		 * @since Jetpack $$next_version$$
+		 * @param boolen initial value whether classic search is enabled.
```
bluefuton:
```suggestion
		 * @param boolean initial value whether classic search is enabled.
```

projects/packages/search/src/initializers/class-initializer.php
```diff
@@ -112,11 +113,21 @@ protected static function init_search( $blog_id ) {
...
+		 * @param boolen filtered result whether classic search is enabled.
```
bluefuton:
```suggestion
		 * @param boolean filtered result whether classic search is enabled.
```

projects/packages/search/compatibility/unsupported-browsers.php
```diff
@@ -0,0 +1,25 @@
...
+ * Compatibility for iOS <= 15.x
+ *
+ * @package automattic/jetpack-search
+ */
+namespace Automattic\Jetpack\Search\Compatibility;
+
+add_filter( 'jetpack_search_classic_search_enabled', __NAMESPACE__ . '\enable_classic_search_for_unsupported_browsers', 10, 1 );
+
+/**
+ * Enable Classic Search when the frontend is not working.
```
bluefuton:
```suggestion
 * Enable Classic Search for browsers that we don't support for Instant Search.
```

kangzj: a0f1913c93

a8ck3n (commented):
This doesn't work for me.

I set up a test JN site with Instant Search enabled. After forcing a sync and updating the index, things worked correctly on Desktop Safari but failed on an iPhone 13 with iOS 15.4 (via BrowserStack). It would only show the, "No posts were found" message.

Disabling Instant Search and testing again did work on iOS 15.4 and it listed a bunch of search results. Enabling Instant Search and testing again still resulted in no posts being found.

projects/packages/search/compatibility/unsupported-browsers.php
```diff
@@ -0,0 +1,39 @@
...
+	$version = str_replace( '_', '.', $matches[2] );
+	return $version;
+}
+
+/**
+ * Force enable Classic Search for browsers for iOS versions < 16.
+ *
+ * @param boolean $classic_search_enabled whether Classic Search is enabled.
+ */
+function enable_classic_search_for_unsupported_browsers( $classic_search_enabled ) {
```
a8ck3n, line 33: I think it might make sense to rename this to better reflect what the code is actually doing. This code doesn't "enable classic search" per say. It only tests if the version is less than 16. Maybe either `is_iOS_less_than_16` or `should_enable_classic_search` or something like that.

a8ck3n (approved):
I revisited this on the same test site after the weekend and now it's working like a charm.

Left a note but no blockers. :shipit:

kangzj:
> A follow up task will be needed to enable the compatibility patch on WPCOM.

I checked WPCOM, the file will be automatically loaded, so no follow up needed. The changes are not deployed to WPCOM yet tho.

`wpcom/wp-content/mu-plugins/site-search/class-wpcom-initializer.php?r=80d9798b#48`

projects/packages/search/compatibility/unsupported-browsers.php
kangzj: That's a good point - sorry I only saw this after merging the PR. It probably is not worth a PR itself, and I'll see what I can do in the future. Thanks @a8ck3n

PR #33931, opened 2023-11-02
Nextdoor: Fix the embed resizing

pablinos:
Nextdoor are now sending us the height of the embed using postMessage, so we need to react to that.

### Other information:

No.
## Testing instructions:

This patch needs to be tested on Jetpack and WPCOM sites too.

For Jetpack sites:

1. Add `define( 'JETPACK_BLOCKS_VARIATION', 'beta' );` to your site so your see the beta blocks.
2. Create a new post, and try adding a Nextdoor block with pasting the link into the body, or searching for the block then pasting the URL to the post
3. The block should appear with the right height after loading in the editor. Create at least 2 blocks to test the unique ids, then publish the post.
4. Check the published post, the blocks should be there with the right size too
5. Repeat steps 1-4 with and without the Gutenberg plugin activated, to test if it works with nested iframes.

For WPCOM sites:

1. Use `bin/jetpack-downloader` following the instructions below to test the changes on your wpcom site.
2. After sandboxing the site follow steps 2-4 above and make sure everything works fine

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/nextdoor/edit.js
+++ b/projects/plugins/jetpack/extensions/blocks/nextdoor/edit.js
@@ -1,6 +1,6 @@
 import { getBlockIconComponent } from '@automattic/jetpack-shared-extension-utils';
 import { getBlockDefaultClassName } from '@wordpress/blocks';
-import { Button, Placeholder, withNotices } from '@wordpress/components';
+import { SandBox, Button, Placeholder, withNotices } from '@wordpress/components';
 import { useState } from '@wordpress/element';
 import { __, _x } from '@wordpress/i18n';
 import './editor.scss';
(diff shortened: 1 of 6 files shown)


projects/plugins/jetpack/extensions/blocks/nextdoor/edit.js
```diff
@@ -93,23 +93,45 @@ export function NextdoorEdit( {
...
+				if ( event.source !== iframe.contentWindow ) {
+					return;
+				}
+				iframe.setAttribute( 'height', event.data.height );
+				var clientBoundingRect = document.body.getBoundingClientRect();
+				window.parent.postMessage( { action: 'resize', width: clientBoundingRect.width, height: clientBoundingRect.height }, '*' );
+			} );
+		</script>
+		<iframe id="nextdoor-embed" width="100%" height="200" frameBorder="0" src="${ iframeUrl }" title="Nextdoor" />
+		`;
```
pablinos: This is a bit of a mess, but the resize events weren't firing properly in the sandbox so I've ended up having to mimic the the resize functionality in [the SandBox component](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/sandbox/index.tsx#L24-L35).

Overall I think using a SandBox element is the way to go, as then we're in control of the iframe context, and can receive the messages.

pablinos (commented):
@gmjuhasz Would you mind picking this one up? It's working, but might need some tidying up, dependent on time.

The SandBox component is doing some pretty interesting things with adding script tags and serialising JS functions to strings which we could do here too, but it's probably safer to include an inline script like I'm doing now.

The alternative might be to have an external script that can be passed to the SandBox component to load. That way it could maybe make use of the util functions, and make things like the origin checks more DRY.

projects/plugins/jetpack/extensions/blocks/nextdoor/edit.js
```diff
@@ -93,23 +93,45 @@ export function NextdoorEdit( {
...
+				var iframe = document.getElementById( 'nextdoor-embed' );
+				if ( event.source !== iframe.contentWindow ) {
+					return;
+				}
+				iframe.setAttribute( 'height', event.data.height );
+				var clientBoundingRect = document.body.getBoundingClientRect();
+				window.parent.postMessage( { action: 'resize', width: clientBoundingRect.width, height: clientBoundingRect.height }, '*' );
+			} );
+		</script>
+		<iframe id="nextdoor-embed" width="100%" height="200" frameBorder="0" src="${ iframeUrl }" title="Nextdoor" />
```
manzoorwanijk: What if there are more than one nextdoor embeds on the page? `id` attribute will need to be different.

gmjuhasz: Fixed in https://github.com/Automattic/jetpack/pull/33931/commits/ccf50e7755f19436b44b2a5b4980a7c9407c7cfe

projects/plugins/jetpack/extensions/blocks/nextdoor/edit.js
```diff
@@ -93,23 +94,49 @@ export function NextdoorEdit( {
...
+				var iframe = document.getElementById( 'nextdoor-embed-${ clientId }' );
+				if ( event.source !== iframe.contentWindow ) {
+					return;
+				}
+				iframe.setAttribute( 'height', event.data.height );
+				var clientBoundingRect = document.body.getBoundingClientRect();
+				window.parent.postMessage( { action: 'resize', width: clientBoundingRect.width, height: clientBoundingRect.height }, '*' );
+			} );
+		</script>
+		<iframe id="nextdoor-embed-${ clientId }" width="100%" height="200" frameBorder="0" src="${ iframeUrl }" title="Nextdoor" />
```
manzoorwanijk: Should the `title` attribute be translated?

manzoorwanijk (dismissed):
This works well for me.

projects/plugins/jetpack/extensions/blocks/nextdoor/edit.js
gmjuhasz: Translated here https://github.com/Automattic/jetpack/pull/33931/commits/264f273f74e070b81bd98ec630a33aac66527cb9

spsiddarthan (approved):
Tests well for me as well!

PR #33932, opened 2023-11-02
Fixed subscribe block redirect issue on non-post pages

TimBroddin:
## Proposed changes:
The subscribe block was incorrectly redirecting to the post from `get_the_ID()` when used on the homepage and archive pages.

This PR also contains a small fix to bind the showing of the subscribe modal to all subscribe forms on the page.

### Other information:

No

## Testing instructions:
- Apply this PR
- Change the design of your blog: 
   - Add a subscribe block to the footer
   - Show post content inside the post loop for the homepage
- Create a post with a subscribe block 
- Inspect the subscribe blocks using developer tools:
   - When on the home page the footer subscribe block shouldn't contain a hidden `post_id` field
   - The subscribe block inside the post loop should contain a hidden `post_id` field containing the post_id of the post being shown
   - Both subscribe blocks on the post detail (single) page should contain a hidden `post_id` field containing the post_id of the post being shown

You can also try subscribing using the different blocks, the one in the footer on the home page should redirect to `Reader > Manage Subscriptions` while the other ones should redirect to the post.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -609,12 +609,19 @@ function get_post_access_level_for_current_post() {
  * @return string
  */
 function render_for_website( $data, $classes, $styles ) {
-	$blog_id            = \Jetpack_Options::get_option( 'id' );
-	$widget_id_suffix   = Jetpack_Subscriptions_Widget::$instance_count > 1 ? '-' . Jetpack_Subscriptions_Widget::$instance_count : '';
-	$form_id            = 'subscribe-blog' . $widget_id_suffix;
-	$form_url           = defined( 'SUBSCRIBE_BLOG_URL' ) ? SUBSCRIBE_BLOG_URL : '#';
-	$post_access_level  = get_post_access_level_for_current_post();
-	$post_id            = get_the_ID();
+	$blog_id           = \Jetpack_Options::get_option( 'id' );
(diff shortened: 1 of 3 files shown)


TimBroddin:
Rebasing this later today.

simison:
Rebased since I'm testing anyway. :-)

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -609,12 +609,19 @@ function get_post_access_level_for_current_post() {
...
+	$form_id           = 'subscribe-blog' . $widget_id_suffix;
+	$form_url          = defined( 'SUBSCRIBE_BLOG_URL' ) ? SUBSCRIBE_BLOG_URL : '#';
+	$post_access_level = get_post_access_level_for_current_post();
+
+	$post_id = null;
+	if ( in_the_loop() ) {
+		$post_id = get_the_ID();
+	} elseif ( is_singular( 'post' ) ) {
+		$post_id = get_queried_object_id();
+	}
```
simison, line 623: Something like this would also catch pages, and always have a fallback for the homepage? I'd like to avoid sending to subscription management in any situation.

```php
if ( in_the_loop() ) {
	$post_id = get_the_ID();
} elseif ( is_singular( 'post' ) || is_page() ) {
	$post_id = get_queried_object_id();
} else {
	$post_id = get_option( 'page_on_front' );
}
```

simison (approved):
Let's get this in and iterate.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
simison: https://github.com/Automattic/jetpack/pull/34086

PR #33933, opened 2023-11-02
wp-admin Site Default: Add notice banner for wp-admin users

jeroenpf:
Fixes https://github.com/Automattic/dotcom-forge/issues/4335

> The settings pages in the Classic wp-admin view have fewer options than their counterparts in Calypso. We should notify users about this fact so they understand that additional settings are available. We should also let them switch to the other version easily.

## Proposed changes:

* For a selected number of settings pages, we add a notice element via the `admin_notices` action hook.
* The notice will inform the user more settings are available via the default view.
* The notice will have a link that directs the user to the relevant Calypso page
* The notice is added to the: General, Writing, Reading and Discussion settings pages.

### Other information:

## Testing instructions:

* Load this PR onto a WoA site - you can use the Jetpack Beta Tester plugin or rsync the plugin to a WoA dev blog
* Via wpcalypso.wordpress.com, change the admin interface style setting to Classic
* Go to one of the applicable settings pages ( General, Writing, Reading, Discussion ) and observe that there is a notice informing the user of the Calypso view.
* Ensure that other wp-admin pages do not display this notice.
* Ensure that the links in the notice bring you to the appropriate page in the Calypso UI.

### DIFF
--- a/projects/plugins/jetpack/modules/masterbar/admin-menu/class-atomic-admin-menu.php
+++ b/projects/plugins/jetpack/modules/masterbar/admin-menu/class-atomic-admin-menu.php
@@ -501,7 +501,7 @@ class Atomic_Admin_Menu extends Admin_Menu {
 		// Close over the $switch_url variable.
 		$admin_notices = function() use ( $switch_url ) {
 			// translators: %s is a link to the Calypso settings page.
-			$notice = __( 'You are currently using the Classic view, which doesn\'t offer the same set of features as the Default view. To Access additional settings and features, <a href="%s">switch to the Default view</a>. ', 'jetpack' );
+			$notice = __( 'You are currently using the Classic view, which doesn\'t offer the same set of features as the Default view. To access additional settings and features, <a href="%s">switch to the Default view</a>. ', 'jetpack' );
 			?>
 			<div class="notice notice-warning">
 				<p><?php echo wp_kses( sprintf( $notice, esc_url( $switch_url ) ), array( 'a' => array( 'href' => array() ) ) ); ?></p>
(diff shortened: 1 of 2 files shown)


javierarce:
LGTM! Here's a screenshot for reference:


Screenshot 2023-11-02 at 12 04 12

projects/plugins/jetpack/modules/masterbar/admin-menu/class-atomic-admin-menu.php
```diff
@@ -463,4 +468,47 @@ public function wp_ajax_jitm_dismiss() {
...
+
+		// Show the notice for the following screens and map them to the Calypso page.
+		$screen_map = array(
+			'options-general'    => 'general',
+			'options-writing'    => 'writing',
+			'options-reading'    => 'reading',
+			'options-discussion' => 'discussion',
+		);
+
+		$mapped_screen = isset( $screen_map[ get_current_screen()->id ] )
```
danielbachhuber: `get_current_screen()` can return `null`: https://developer.wordpress.org/reference/functions/get_current_screen/#return

Need to make sure we have a valid screen object too.

jeroenpf: I added a check to ensure it is a WP_Screen

wojtekn (approved):
@jeroenpf code looks correct and works as expected. I re-run tests as they failed due to a MailChimp error which looked like a one-time issue not related to the changes.

jeroenpf:
Thanks @wojtekn - the mailchimp is indeed not related to this change and is happening on other branches too. Seems like the test depends on something external that is timing out.

PR #33938, opened 2023-11-02
Make Subscribe block default texts editable

### LINKED ISSUE #26377: Subscriber block: Allow editing of placeholder from the editor
Follow-up to #18072

The frontend respects the block attributes, but the block placeholder is not editable from the Editor right now

image

image

Updating by hand from the Code Editor view works, although it warns of a validation. Clicking Attempt Block Recovery works in the end. Seems to be a validation error related to the order of the attributes and the shortcode being rendered, or something like that

monsieur-z:
Fixes https://github.com/Automattic/jetpack/issues/26377

## Proposed changes:

The _Subscribe_ block consists of a text input and a button. This PR makes the input placeholder and button label editable.

| Before | After |
|-|-|
|||
|||

### Other information:

## Jetpack product discussion

n/a

No.

### Self-hosted site
- Visit `/wp-admin/post-new.php`
- Add the `Subscribe` block to the page, and activate subscriptions if they're not already
- Make sure the block is selected
- In the _Block_ panel of the sidebar, expand the _Settings_ section
- Notice you can update the input placeholder and submit label (they should be updated in real-time in the editor)
- Refresh the page and check the updated texts have been saved
- Preview the post and notice that you can see the custom texts
- Optionally, beginning with a version of the plugin that doesn't include these changes, add the `Subscribe` block to the page, update the plugin to use this branch, and reload the page. Notice that upgrading the plugin didn't cause any issue in the existing page.

### Simple site
- Download the branch
- Build the plugin: `jetpack build plugins/jetpack`
- Sync the build with your sandbox site: `jetpack rsync jetpack`
- Visit `<your site>/wp-admin/post-new.php`
- Repeat the above steps

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/controls.js
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/controls.js
@@ -264,7 +264,7 @@ export default function SubscriptionControls( {
 				<TextareaControl
 					value={ subscribePlaceholder }
 					label={ __( 'Input placeholder text', 'jetpack' ) }
-					help={ __( 'Edit the placeholder text of email address input.', 'jetpack' ) }
+					help={ __( 'Edit the placeholder text of the email address input.', 'jetpack' ) }
 					onChange={ placeholder => setAttributes( { subscribePlaceholder: placeholder } ) }
 				/>
 				<TextareaControl
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/controls.js
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/controls.js
@@ -51,6 +51,8 @@ export default function SubscriptionControls( {
 	subscriberCount,
 	textColor,
 	buttonWidth,
+	subscribePlaceholder,
+	submitButtonText,
 	successMessage,
 } ) {
 	const { isPublicizeEnabled } = usePublicizeConfig();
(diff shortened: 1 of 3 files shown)


projects/plugins/jetpack/extensions/blocks/subscriptions/controls.js
```diff
@@ -258,6 +260,19 @@ export default function SubscriptionControls( {
...
+					label={ __( 'Input placeholder text', 'jetpack' ) }
+					help={ __( 'Edit the placeholder text of the email address input.', 'jetpack' ) }
+					onChange={ placeholder => setAttributes( { subscribePlaceholder: placeholder } ) }
+				/>
+				<TextareaControl
+					value={ submitButtonText }
+					label={ __( 'Submit button label', 'jetpack' ) }
+					help={ __( 'Edit the label of the button a user clicks to subscribe.', 'jetpack' ) }
+					onChange={ text => setAttributes( { submitButtonText: text } ) }
+				/>
```
jeherve, line 275: Do we need this, given that one can click on the button to edit its text?

jeherve (commented):
Looks good, just a small question from me.

projects/plugins/jetpack/extensions/blocks/subscriptions/controls.js
monsieur-z: I'd say yes, since the submit button has no affordance that it can be edited. With most editable blocks, the label or the placeholder "transparency" somewhat help understand what can be edited.

jeherve: ah, I suppose that makes sense.

It looks like the sidebar controls over less flexibility with line breaks though?


https://github.com/Automattic/jetpack/assets/426388/8fef78fc-728a-4e83-be07-01966aab8fd4

monsieur-z: I don't necessarily see this as an issue, especially since a multi-line button would visually clash with the single-line input in this specific case.

But I don't mind looking into it. Maybe in a separate PR so that we can merge this one?

jeherve (approved):

PR #33941, opened 2023-11-02
Refactor payment-related blocks registration

monsieur-z:

## Proposed changes:

This PR is a follow up of https://github.com/Automattic/jetpack/pull/33842, which was reverted in https://github.com/Automattic/jetpack/pull/33895. The differences between the 2 PRs are the following:
- The simple payments block API version was set to 1 (this was done in https://github.com/Automattic/jetpack/pull/33889).
- The default value of the `currency` attribute for the same block was set to `USD` (replacing an invalid value).

See p1698765346173339-slack-C052XEUUBL4 for more details.

_Below's the description of the original PR._

Registers the following production blocks by specifying the path to their block.json file:  _Payments Intro_, _Paywall_, _Premium Content_, _Recurring Payment_, _Simple Payment_, _Subscriptions_.

  - Remove `icon.js`
  - Fix JS tests

### Other information:

## Jetpack product discussion

pedMtX-RS-p2

No.

## Testing instructions:

- Spin up a test site
- Optionally, test these blocks on a simple site


anomiex:
LGTM, the issue with the payment block is fixed now in some manual tests. Remind me to approve this after 12.8 is cut, it's too close right now for me to feel comfortable trying this again now.

anomiex:
Ok, 12.8 is out. This needs a trunk merge now, then we can give it a try.

anomiex (dismissed):

jeherve (approved):
This tests well for me. 🚢

projects/plugins/jetpack/extensions/blocks/paywall/editor.js
```diff
@@ -1,8 +1,12 @@
 import { unregisterBlockType } from '@wordpress/blocks';
 import { subscribe, select } from '@wordpress/data';
 import { store as editorStore } from '@wordpress/editor';
-import registerJetpackBlock from '../../shared/register-jetpack-block';
-import { name, settings } from '.';
```
fluiddot, line 5: 👋 @monsieur-z, I noticed that `name` is being used in line 21:
https://github.com/Automattic/jetpack/blob/30f477439a16f9720b63523b64a314846e02483a/projects/plugins/jetpack/extensions/blocks/paywall/editor.js#L21

This produces a crash in the Jetpack app. Could we revert this change and keep the `name` property?
Thanks 🙇 !

fluiddot: I went ahead and created a PR to fix this: https://github.com/Automattic/jetpack/pull/34274.
Let me know if you could take a quick look, thanks 🙇  !

PR #33942, opened 2023-11-02
AI Assistant: Add loading state to usage panel

### LINKED ISSUE #33936: AI Tiered Plans: Add loading state to components that use initial state data
ref p1698873875426409/1698850671.294749-slack-C054LN8RNVA

For performance reasons, the initial data will be fetched at the frontend editor level when necessary, avoiding computing values at page request time. Because of this the components that required this data will need a UI loading state.

dhasilva:
Fixes #33936

## Proposed changes:

* Exposes `loading` on `useAIFeature`
* Adds a `LoadingPlaceholder` component when loading on `UsagePanel`
* Uses `ThemeProvider` to load Jetpack CSS, necessary for the `LoadingPlaceholder` component

### Other information:

## Testing instructions:

* Go to the block editor and open the Jetpack sidebar
* Open the AI Assistant panel
* Check that the placeholder is displayed while loading

Now we are using initial data, but the initial data will be removed. Loading will happen only once after the store is set up.

https://github.com/Automattic/jetpack/assets/8486249/28d056a2-855f-4fe8-814b-89732ed23597

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-ai-feature/index.ts
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-ai-feature/index.ts
@@ -56,44 +56,40 @@ export async function getAIFeatures(): Promise< AIFeatureProps > {
 		path: '/wpcom/v2/jetpack-ai/ai-assistant-feature',
 	} );
 
-	try {
-		return {
-			hasFeature: !! response[ 'has-feature' ],
-			isOverLimit: !! response[ 'is-over-limit' ],
-			requestsCount: response[ 'requests-count' ],
-			requestsLimit: response[ 'requests-limit' ],
-			requireUpgrade: !! response[ 'site-require-upgrade' ],
-			errorMessage: response[ 'error-message' ],
-			errorCode: response[ 'error-code' ],
-			upgradeType: response[ 'upgrade-type' ],
-			usagePeriod: {
-				currentStart: response[ 'usage-period' ]?.[ 'current-start' ],
-				nextStart: response[ 'usage-period' ]?.[ 'next-start' ],
-				requestsCount: response[ 'usage-period' ]?.[ 'requests-count' ] || 0,
-			},
-			currentTier: {
-				value: response[ 'current-tier' ]?.value || 1,
-			},
-		};
-	} catch ( error ) {
-		console.error( error ); // eslint-disable-line no-console
-	}
+	return {
+		hasFeature: !! response[ 'has-feature' ],
+		isOverLimit: !! response[ 'is-over-limit' ],
+		requestsCount: response[ 'requests-count' ],
+		requestsLimit: response[ 'requests-limit' ],
+		requireUpgrade: !! response[ 'site-require-upgrade' ],
+		errorMessage: response[ 'error-message' ],
+		errorCode: response[ 'error-code' ],
+		upgradeType: response[ 'upgrade-type' ],
+		usagePeriod: {
+			currentStart: response[ 'usage-period' ]?.[ 'current-start' ],
+			nextStart: response[ 'usage-period' ]?.[ 'next-start' ],
+			requestsCount: response[ 'usage-period' ]?.[ 'requests-count' ] || 0,
+		},
+		currentTier: {
+			value: response[ 'current-tier' ]?.value || 1,
+		},
+	};
 }
 
 export default function useAIFeature() {
 	const [ data, setData ] = useState< AIFeatureProps >( AI_Assistant_Initial_State );
 	const [ loading, setLoading ] = useState< boolean >( false );
-	const [ failed, setFailed ] = useState< boolean >( false );
+	const [ error, setError ] = useState< Error >( null );
 
 	const loadFeatures = async () => {
 		setLoading( true );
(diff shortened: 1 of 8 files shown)


retrofox (commented):
I woulnd't introduce the ThemeProvider in the block editor context.
If I'm not wrong, ThemeProvider provides styles defined to be used in the (I don’t know how to explain it actually) "wp-admin" context. Consuming these styles could generate a disparity with core styles.

We must be cautious about how and where this component is consumed since it defines (and redefines) a large number of styles when, as is the case, only a couple are needed.
Also, Ideally, It should be used at the top of the app, as much as possible.

I understand the idea of ensuring compatibility between different contexts, but we should avoid using it at least for now.

dhasilva:
> Consuming these styles could generate a disparity with core styles

The styles are mostly prefixed.

And the LoadingPlaceholder component uses at least one such variable. In a general sense, it is easy to add a component and miss the fact that it uses some Jetpack variable that is not defined, and given that this is a Jetpack feature, I see no problem with adding Jetpack styles. As for style redefinition it is scoped to the panel, so even if it happens, it would not break anything outside of it.

I'll try do duplicate it manually so it works without the provider.

> Also, Ideally, It should be used at the top of the app, as much as possible.

Things like portals make top-level use not work. Even in this case, wrapping the Sidebar with it does not work.

projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-ai-feature/index.ts
```diff
@@ -82,13 +82,32 @@ export async function getAIFeatures(): Promise< AIFeatureProps > {
...
+	const [ loading, setLoading ] = useState< boolean >( false );
+	const [ failed, setFailed ] = useState< boolean >( false );
+
+	const loadFeatures = async () => {
+		setLoading( true );
+		setFailed( false );
+
+		try {
+			const aiFeatures = await getAIFeatures();
+			setData( aiFeatures );
```
retrofox, line 90: should set `failed` to `false` here?

retrofox: Also, does it make sense to store the error instead of the boolean state? It can be provided by the `getAIFeatures()` helper.

CGastrell:
Works nice. I'd second @retrofox if we don't exclusively need ThemeProvider, don't use it. What is it providing in this context?

The only nit: I'd like it if we could assign some css transition to the height or min/max-height of the container so it doesn't snap when the load is done. TBH, I'm not sure this is possible though

projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-ai-feature/index.ts
dhasilva: > should set failed to false here?

`error` is set to `null` when the loading starts, as that is a new request, so no need to set it on success.

dhasilva:
>  I'd like it if we could assign some css transition to the height or min/max-height of the container

Not possible with CSS alone in this case as the height is `auto`, set by the height of the children (the content replaces the loader). We could do it with JS but I think it is overkill for this case as it happens quickly and only once.

retrofox:
> > Consuming these styles could generate a disparity with core styles
> 
> The styles are mostly prefixed.
> 
> And the LoadingPlaceholder component uses at least one such variable. In a general sense, it is easy to add a component and miss the fact that it uses some Jetpack variable that is not defined, and given that this is a Jetpack feature, I see no problem with adding Jetpack styles. 

Adding Jetpack styles in this context could generate inconsistencies in terms of the UI. The styles of the app components app should not be defined by the Jetpack plugin. There are a few comments about a similar topic here p1HpG7-oMv-p2#comment-67063

Also, every time you create a `<ThemeProvider />` instance, it defines a bunch of CSS variables, which is not ideal for performance:

Screenshot 2023-11-03 at 09 50 05

It's a Theme provider, meaning it should be instantiated once in the app. Having multiple instances reflects that, in some way, the component is not used properly.

AFAICS, `<LoadingPlaceholder />` is the only component that consumes the styles provided by the `<ThemeProvider />` component. I tend to think that maybe that implementation isn't right either.

> > Also, Ideally, It should be used at the top of the app, as much as possible.
> 
> Things like portals make top-level use not work. Even in this case, wrapping the Sidebar with it does not work.

Exactly, that's another reason to think that maybe the `<ThemeProvider/>` usage doesn't scale very well here either. To put us on the same page, initially the `<ThemeProvider />` was designed to be used in other contexts, specifically for MyJetpack product.

dhasilva:
I see, there are discussions about the style consistency I was not aware of.
The ThemeProvider was removed.

I have to add, though, that we would need to test the performance hit of adding CSS variables, as it seems negligible to me.

retrofox:
> I have to add, though, that we would need to test the performance hit of adding CSS variables, as it seems negligible to me.

Yes, that's correct. Fair enough. In any case, my argument is not based exclusively on performance but on use that I consider inappropriate.

CGastrell:
Works nicely after the update. Good work!

CGastrell (approved):
LGTM! Thanks! :shipit:

PR #33945, opened 2023-11-03
wp-admin Site Default: respect user's view setting in Classic style

jeroenpf:
Fixes https://github.com/Automattic/dotcom-forge/issues/4455

The user can change the default view via the View menu. When the admin interface style has been set to wp-admin (Classic) all pages will link to wp-admin instead of Calypso. You can still use the 'view' selector but this does not persist beyond the current pageview.

Below, is a screenshot of the View menu that allows you to set which style to use on a per-page level.
CleanShot 2566-11-03 at 14 30 33@2x

## Proposed changes:

* We check if we need to load wp-admin based on the `wpcom_admin_interface` setting **after** we check for the preferred style selected by the user on the page level.

### Other information:

## Testing instructions:

* Using a WoA site, apply this PR. You can do this via the Jetpack Beta Tester plugin OR by rsyncing the Jetpack plugin to a WoA Dev blog.
* Via the hosting configuration page, select the Classic admin interface style.
* Go to a wp-admin page on your WoA site.
* Check that all links point to wp-admin except the ones that should link to Calypso.
* Change the view option of any wp-admin page to Default.
* Ensure you are correctly redirected to the Calypso version of the page.
* Ensure that the link in the menu also points to the Calypso version of the page, while all other links still point to wp-admin.

For this test, it is important that you have not set any view preference yet, otherwise, the results might be confusing. You can remove the `user option` `jetpack_admin_menu_preferred_views` (this is where we store the preferences for which view to use per page).

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-respect-user-view-setting
+++ b/projects/plugins/jetpack/changelog/fix-respect-user-view-setting
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
+Respect the user's view setting when wpcom_admin_interface is wp-admin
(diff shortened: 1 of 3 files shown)


wojtekn (approved):
The fix looks clear and works as expected. I can now use a default (Calypso) interface for some pages like Posts and Settings -> Writing, and wp-admin for the rest, and my choice is being preserved while I navigate through my site.

PR #33947, opened 2023-11-03
Prevent PHP warnings by adding `isset` and `is_array` checks to inner block loop in WooCommerce Analytics module

opr:
## Proposed changes:

* Hardens the code to prevent PHP warnings when accessing undefined array members.

### Other information:

## Jetpack product discussion

p1698931821469179-slack-C011ENB20Q1

No

## Testing instructions:

- Ensure the WooCommerce analytics module is enabled
- Set up your site so the WooCommerce Blocks `Cart` block is enabled on the Cart page, and the `Checkout` block is enabled on the Checkout page.
- Go to the Cart page and then proceed to the Checkout page.
- Ensure no PHP warnings are shown.

### DIFF
--- a/projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-universal.php
+++ b/projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-universal.php
@@ -470,7 +470,7 @@ class Jetpack_WooCommerce_Analytics_Universal {
 	 */
 	private function get_inner_blocks( $inner_blocks ) {
 		$block_names = array();
-		if ( isset( $inner_blocks['blockName'] ) && $inner_blocks['blockName'] ) {
+		if ( ! empty( $inner_blocks['blockName'] ) ) {
 			$block_names[] = $inner_blocks['blockName'];
 		}
 		if ( isset( $inner_blocks['innerBlocks'] ) && is_array( $inner_blocks['innerBlocks'] ) ) {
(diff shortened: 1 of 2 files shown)


PanosSynetos (dismissed):
Thanks for this @opr - Manually tested on a Woo Express site and it solves the warnings.

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-universal.php
```diff
@@ -470,10 +470,10 @@ public function remove_from_cart_via_quantity() {
 	 */
 	private function get_inner_blocks( $inner_blocks ) {
 		$block_names = array();
-		if ( $inner_blocks['blockName'] ) {
+		if ( isset( $inner_blocks['blockName'] ) && $inner_blocks['blockName'] ) {
```
jeherve: What do you think about doing something like this instead?

```suggestion
		if ( ! empty( $inner_blocks['blockName'] ) ) {
```

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-universal.php
```diff
@@ -520,11 +523,11 @@ function ( $block ) {
 		foreach ( $other_blocks as $block ) {
 
 			// This check is necessary because sometimes this is null when using templates.
-			if ( $block['blockName'] ) {
+			if ( isset( $block['blockName'] ) && $block['blockName'] ) {
```
jeherve:
```suggestion
			if ( ! empty( $block['blockName'] ) ) {
```

jeherve (dismissed):
I left some comments, but they are not blockers. 👍

opr:
@jeherve thanks for the suggestions, I applied them. I didn't realise `empty` didn't show warnings for non existent variables! Nice improvements. @PanosSynetos and your reviews got marked as stale after applying the new changes so if you could re-approve that would be ace.

jeherve (approved):

ice9js:
Cherry-picked to 12.8 in f1bf06f92e49767a75323abac43dc5ab0b9d23ed.

PR #33948, opened 2023-11-03
Subscriber page: add Launchpad tasks & task list

### LINKED ISSUE #83846: Launchpad: create tasklist
Create a tasklist named `subscribers` inside jetpack-mu-wpcom: src/features/launchpad/launchpad.php containing the tasks defined in the other tasks.

Depends on #83847 #83848 #83849

### LINKED ISSUE #83847: Launchpad: import subscribers task
Create a task for import subscribers in jetpack-mu-wpcom in `src/features/launchpad/launchpad-task-definitions.php` . This task should have an id_map pointing add the existing subscribers_added task, like so:

```
		'subscribers_imported'               => array(
			'get_title'            => function () {
				return __( 'Import existing subscribers', 'jetpack-mu-wpcom' );
			},
			'id_map'            => 'subscribers_added',
			'is_complete_callback' => 'wpcom_launchpad_is_task_option_completed',
			'is_visible_callback'  => 'wpcom_launchpad_has_goal_import_subscribers',
			'get_calypso_path'     => function ( $task, $default, $data ) {
				return '/subscribers/' . $data['site_slug_encoded'] . '#add-subscribers';
			},
		),
```

### LINKED ISSUE #83848: Launchpad: share site task
This task already exists as 

```
'share_site'                      => array(
			'get_title'            => function () {
				return __( 'Share your site', 'jetpack-mu-wpcom' );
			},
			'is_complete_callback' => 'wpcom_launchpad_is_task_option_completed',
		),
```

It can probably be reused.

### LINKED ISSUE #83849: Launchpad: add subscriber block task
Create the add subscriber block task inside jetpack-mu-wpcom: `projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php`

TimBroddin:
Fixes https://github.com/Automattic/wp-calypso/issues/83846
Fixes https://github.com/Automattic/wp-calypso/issues/83847
Fixes https://github.com/Automattic/wp-calypso/issues/83848
Fixes https://github.com/Automattic/wp-calypso/issues/83849

## Proposed changes:
- This adds the needed tasks for the Launchpad for Subscriber Page.
- It also adds a check that triggers the `subscriber_block_added` task as complete, when a template part is saved with a jetpack:subscribe block in it.
- It will hide the Add subscriber block task when not using an FSE theme

This PR does not:
- Open a modal when clicking Import subscribers
- Complete the Import subscribers task when clicking it

### Other information:

## Jetpack product discussion
pdDOJh-2Ee-p2

No
## Testing instructions:

- Apply this PR to your sandbox by following the instructions below
- Open Calypso in your IDE and add a launchpad component somewhere to `client/my-sites/subscribers/components/subscriber-list-container/subscriber-list-container.tsx`. Make sure to replace the siteSlug to that of the site you're testing with.

```js
import { Launchpad } from '@automattic/launchpad'
---
<Launchpad
siteSlug={ "broddinsandbox.wordpress.com" }
checklistSlug={ "subscribers" }
launchpadContext='my-app-context'

/>
```

CleanShot 2023-11-03 at 13 20 07@2x

Check that:
- Clicking `Import subscribers` adds `#add-subscribers` to the page url
- Clicking `Share your site` opens a share box, clicking copy marks the task as complete
- Clicking `Add a subscriber block` takes you to the site editor, adding a subscriber block to a template or template part marks this task as complete.
- Choosing a non-block editor theme (such as Twenty Twenty Five) hides the `Add a subscriber block` task

Also test on Jetpack/Atomic using the instructions below.

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/changelog/add-subscriber-launchpad-tasks
+++ b/projects/packages/jetpack-mu-wpcom/changelog/add-subscriber-launchpad-tasks
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Added Subscriber page Launchpad tasks
--- a/projects/packages/jetpack-mu-wpcom/changelog/add-subscriber-launchpad-tasks
+++ b/projects/packages/jetpack-mu-wpcom/changelog/add-subscriber-launchpad-tasks
@@ -1,4 +1,4 @@
 Significance: minor
 Type: added
 
-Added Subscriber page Launchpad tasks
+Added Subscribers page Launchpad tasks
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -1797,7 +1797,7 @@ add_action( 'post_updated', 'wpcom_launchpad_edit_page_check', 10, 3 );
  * @return bool True if we should show the task, false otherwise.
  */
 function wpcom_launchpad_is_add_subscribe_block_visible() {
-	return Blocks::is_fse_theme();
+	return is_callable( array( '\Automattic\Jetpack\Blocks', 'is_fse_theme' ) ) && \Automattic\Jetpack\Blocks::is_fse_theme();
 }
 
 /**
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -5,6 +5,8 @@
  * @package automattic/jetpack-mu-wpcom
  */
 
+use Automattic\Jetpack\Blocks;
+
 /**
  * Get the task definitions for the Launchpad.
  *
@@ -568,7 +570,7 @@ function wpcom_launchpad_get_task_definitions() {
 				return __( 'Add the Subscribe Block to your theme', 'jetpack-mu-wpcom' );
 			},
 			'is_complete_callback' => 'wpcom_launchpad_is_task_option_completed',
-			'is_visible_callback'  => '__return_true',
+			'is_visible_callback'  => 'wpcom_launchpad_is_add_subscribe_block_visible',
 			'get_calypso_path'     => function ( $task, $default, $data ) {
 				return '/site-editor/' . $data['site_slug_encoded'];
 			},
@@ -1788,6 +1790,15 @@ function wpcom_launchpad_edit_page_check( $post_id, $post ) {
 	wpcom_mark_launchpad_task_complete( 'edit_page' );
 }
 add_action( 'post_updated', 'wpcom_launchpad_edit_page_check', 10, 3 );
+
+/**
+ * Determine `add_subscribe_block` task visibility. The task is visible if using a FSE theme.
+ *
+ * @return bool True if we should show the task, false otherwise.
+ */
+function wpcom_launchpad_is_add_subscribe_block_visible() {
+	return Blocks::is_fse_theme();
+}
 
 /**
  * When a template or template part is saved, check if the subscribe block is in the content.
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -5,8 +5,6 @@
  * @package automattic/jetpack-mu-wpcom
  */
 
-use Automattic\Jetpack\Blocks;
-
 /**
  * Get the task definitions for the Launchpad.
  *
@@ -567,7 +565,7 @@ function wpcom_launchpad_get_task_definitions() {
 		),
 		'add_subscribe_block'             => array(
 			'get_title'            => function () {
-				return __( 'Add the Subscribe Block to your theme', 'jetpack-mu-wpcom' );
+				return __( 'Add the Subscribe Block to your site', 'jetpack-mu-wpcom' );
 			},
 			'is_complete_callback' => 'wpcom_launchpad_is_task_option_completed',
 			'is_visible_callback'  => 'wpcom_launchpad_is_add_subscribe_block_visible',
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -552,6 +552,27 @@ function wpcom_launchpad_get_task_definitions() {
 				return '/site-monitoring/' . $data['site_slug_encoded'];
 			},
 		),
+		'import_subscribers'              => array(
+			'get_title'            => function () {
+				return __( 'Import existing subscribers', 'jetpack-mu-wpcom' );
+			},
+			'id_map'               => 'subscribers_added',
+			'is_complete_callback' => 'wpcom_launchpad_is_task_option_completed',
+			'is_visible_callback'  => '__return_true',
+			'get_calypso_path'     => function ( $task, $default, $data ) {
+				return '/subscribers/' . $data['site_slug_encoded'] . '#add-subscribers';
+			},
+		),
+		'add_subscribe_block'             => array(
+			'get_title'            => function () {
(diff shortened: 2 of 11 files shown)


projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -552,6 +554,27 @@ function wpcom_launchpad_get_task_definitions() {
...
+			'id_map'               => 'subscribers_added',
+			'is_complete_callback' => 'wpcom_launchpad_is_task_option_completed',
+			'is_visible_callback'  => '__return_true',
+			'get_calypso_path'     => function ( $task, $default, $data ) {
+				return '/subscribers/' . $data['site_slug_encoded'] . '#add-subscribers';
+			},
+		),
+		'add_subscribe_block'             => array(
+			'get_title'            => function () {
+				return __( 'Add the Subscribe Block to your theme', 'jetpack-mu-wpcom' );
```
daledupreez: Nit: should this not be "Add the Subscribe Block to your _site_"? I wouldn't expect the block to be added to the theme itself.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -1768,6 +1791,44 @@ function wpcom_launchpad_edit_page_check( $post_id, $post ) {
 }
 add_action( 'post_updated', 'wpcom_launchpad_edit_page_check', 10, 3 );
 
+/**
+ * Determine `add_subscribe_block` task visibility. The task is visible if using a FSE theme.
+ *
+ * @return bool True if we should show the task, false otherwise.
+ */
+function wpcom_launchpad_is_add_subscribe_block_visible() {
+	return Blocks::is_fse_theme();
```
daledupreez: Nit: Can you please use the fully-qualified class name here? It's really hard to see what code you're calling from future diffs where the `use` statement may not be visible.

Also, is this function guaranteed to be defined/callable? It may be worth adding some defensive code to make sure the class exists first.
```suggestion
	return is_callable( array( '\Automattic\Jetpack\Blocks', 'is_fse_theme' ) ) && \Automattic\Jetpack\Blocks::is_fse_theme();
```

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
```diff
@@ -211,6 +211,14 @@ function wpcom_launchpad_get_task_list_definitions() {
 			),
 			'is_enabled_callback' => 'wpcom_launchpad_is_hosting_flow_enabled',
 		),
+		'subscribers'            => array(
+			'title'    => 'Subscribers',
+			'task_ids' => array(
+				'import_subscribers',
+				'share_site',
+				'add_subscribe_block',
```
daledupreez: Shouldn't this be the other way around? I would imagine we want the subscribe block to be added to the site before we ask users to share the site. I know our tasks aren't blocking, but we should still try to sequence them in a way that's likely to produce better outcomes.
```suggestion
				'add_subscribe_block',
				'share_site',
```

daledupreez (commented):
I have some minor suggestions, but nothing blocking, and things are testing well. I'll let other folks on Apex approve, however, as it will be much better if you're all more acquainted with the feature and testing it.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
TimBroddin: I agree. What do you think @crisbusquets ? This is about the order in which we display the tasks.

TimBroddin: I've commited the change. We can always revert this.

projects/packages/jetpack-mu-wpcom/changelog/add-subscriber-launchpad-tasks
```diff
@@ -0,0 +1,4 @@
+Significance: minor
+Type: added
+
+Added Subscriber page Launchpad tasks
```
ivan-ottinger: This is a minor nitpick:

```suggestion
Added Subscribers page Launchpad tasks
```

ivan-ottinger (commented):
Great work, Tim! 💯 

Everything is working as expected:
- all the tasks load (apart from _Add the Subscribe Block to your site_ when the site is non-FSE - as expected)
- it is possible to interact with the tasks as described in this PR
- I did not observe any errors / warnings, nor regressions (neither on the Calypso side, neither on the WPCOM side)

So far I tested on a Simple site only. Now will be testing on WoA.

---

There's one thing though I noticed that might be worth discussing:

When the user clicks on the _Add the Subscribe Block to your site_, they are simply directed to the editor without any guidance:

Markup on 2023-11-07 at 15:53:28

I believe most users will get lost here and won't know what to do next.

An ultimate solution I can imagine right now is to guide the user how they can add the Subscribe block with on-screen instructions - once they hit the editor (but this is something we won't have bandwidth to build or explore at the moment), so instead we could consider directing the user to the docs instead (e.g. https://wordpress.com/support/wordpress-editor/blocks/subscribe-block/)?

But maybe this was discussed somewhere already while I was on AFK. 🤔

TimBroddin:
> Great work, Tim! 💯
> 
> Everything is working as expected:
> 
> * all the tasks load (apart from _Add the Subscribe Block to your site_ when the site is non-FSE - as expected)
> * it is possible to interact with the tasks as described in this PR
> * I did not observe any errors / warnings, nor regressions (neither on the Calypso side, neither on the WPCOM side)
> 
> So far I tested on a Simple site only. Now will be testing on WoA.
> 
> There's one thing though I noticed that might be worth discussing:
> 
> When the user clicks on the _Add the Subscribe Block to your site_, they are simply directed to the editor without any guidance:
> 
> Markup on 2023-11-07 at 15:53:28
> 
> I believe most users will get lost here and won't know what to do next.
> 
> An ultimate solution I can imagine right now is to guide the user how they can add the Subscribe block with on-screen instructions - once they hit the editor (but this is something we won't have bandwidth to build or explore at the moment), so instead we could consider directing the user to the docs instead (e.g. https://wordpress.com/support/wordpress-editor/blocks/subscribe-block/)?
> 
> But maybe this was discussed somewhere already while I was on AFK. 🤔

Thanks for testing @ivan-ottinger !

Great suggestion about linking to the docs! We can override the action that will happen in Calypso itself (see [here](https://github.com/Automattic/wp-calypso/blob/7aeafddf7f86eded3b67fdc70ef441ad47510ca9/packages/launchpad/src/setup-actions.ts#L68-L68) how the share_site action is wired up), so this decision shouldn't be a blocker for this PR. I think like we should open the docs in a new tab when doing so.

@crisbusquets do you also think that linking to the docs on how to add the subscribe block is the better approach here? As Ivan said, it will confuse users when they're just dropped inside the site editor without any explanation.

(I've researched whether we can automatically add a subscribe block to the template, but this feels very hacky. We don't know how the result will look like if we just drop a subscribe block in the footer for example.)

ivan-ottinger:
> [...] I think like we should open the docs in a new tab when doing so.

Yes, that sounds good to me - that way the user wouldn't lose the page they are on.

ivan-ottinger:
Works great on WoA as well. 🙂👍🏼

ivan-ottinger (approved):

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -1768,6 +1789,44 @@ function wpcom_launchpad_edit_page_check( $post_id, $post ) {
 }
 add_action( 'post_updated', 'wpcom_launchpad_edit_page_check', 10, 3 );
 
+/**
+ * Determine `add_subscribe_block` task visibility. The task is visible if using a FSE theme.
+ *
+ * @return bool True if we should show the task, false otherwise.
+ */
+function wpcom_launchpad_is_add_subscribe_block_visible() {
+	return is_callable( array( '\Automattic\Jetpack\Blocks', 'is_fse_theme' ) ) && \Automattic\Jetpack\Blocks::is_fse_theme();
```
simison, line 1798: For classic themes the task could still nudge to add the subscribe widget in the Customizer?

PR #33949, opened 2023-11-03
wp-admin Site Default: Ensure settings point to wp-admin when wpcom_admin_interface is 'wp-admin'

jeroenpf:
Fixes https://github.com/Automattic/dotcom-forge/issues/4455

There was a bug where the general settings page and users page had a special exception to always point to the default view. This is desired behavior unless the new `wpcom_admin_interface` option is set to wp-admin. The logic to force links to go to wp-admin in response to the `wpcom_admin_interface` option was moved down to a later point after the exception was applied. 

See: https://github.com/Automattic/jetpack/pull/33945 where this bug was introduced

## Proposed changes:

* Only apply the settings and user page exception when the `wpcom_admin_interface` option is not `wp-admin`

### Other information:

## Testing instructions:

**Atomic site:**

- Using a WoA site, apply this PR. You can do this via the Jetpack Beta Tester plugin OR by rsyncing the Jetpack plugin to a WoA Dev blog.
- Via the hosting configuration page, select the Classic admin interface style.
- Go to a wp-admin page on your WoA site.
- Check that all links point to wp-admin except the ones that should link to Calypso.
- Change the view option of any wp-admin page to Default.
- Ensure you are correctly redirected to the Calypso version of the page.
- Ensure that the link in the menu also points to the Calypso version of the page, while all other links still point to wp-admin.
- Check that the settings page AND users page both point to wp-admin
- Check that you can change the view setting of the settings page and users page to default and that the menu items change accordingly. Also, verify that you can change back to the classic view again.
- Change the admin interface style back to default and check that the links now point to Calypso again and that the view settings on the per-page level are correctly applied. E.g. try to toggle between Default and Classic to verify that per-page level view settings work correctly. 

**Simple site**

- Load this PR on your sandbox `bin/jetpack-downloader test jetpack fix/settings-and-user-menu-incorrect-links`
- Ensure that the menu items are all pointing to Calypso, except for the pages that should always point to wp-admin regardless.

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-settings-and-user-menu-incorrect-links
+++ b/projects/plugins/jetpack/changelog/fix-settings-and-user-menu-incorrect-links
@@ -1,0 +1,5 @@
+Significance: patch
+Type: bugfix
+Comment: Its a small bugfix
+
+
--- a/projects/plugins/jetpack/modules/masterbar/admin-menu/class-admin-menu.php
+++ b/projects/plugins/jetpack/modules/masterbar/admin-menu/class-admin-menu.php
@@ -57,12 +57,13 @@ class Admin_Menu extends Base_Admin_Menu {
 	 * @return string
 	 */
 	public function get_preferred_view( $screen, $fallback_global_preference = true ) {
+		$force_default_view = in_array( $screen, array( 'users.php', 'options-general.php' ), true );
+		$use_wp_admin       = $this->use_wp_admin_interface( $screen );
+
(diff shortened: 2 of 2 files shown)


danielbachhuber (approved):
Thanks for tracking this down!

This now works as expected on my Atomic site:
1. With "Classic style" selected, I can switch to "Default" and my view selection persists as expected.
2. With "Default style" selected, I can switch to "Classic" and my view selection persists as expected.
3. When I switch between "Classic style" and "Default style", my original custom view selections still apply.

On my Simple site, the view selector works as expected.

PR #33951, opened 2023-11-03
wp-admin Site Default: Always links 'My Sites' to the Sites page

danielbachhuber:
Fixes https://github.com/Automattic/dotcom-forge/issues/4457

## Proposed changes:

Always links 'My Sites' to the Sites page when 'wp-admin' is the selected admin interface:

CleanShot 2023-11-03 at 05 18 43@2x

Because 'My Sites' always links to the Sites page, I'm also removing the 'Browse sites' link:

CleanShot 2023-11-03 at 05 21 49@2x

Also fixes a leftover issue from https://github.com/Automattic/jetpack/pull/33854 where it wasn't actually possible to click "My Sites".

### Other information:

## Jetpack product discussion

N/A

No.

## Testing instructions:

Atomic Setup:
1. Create a new Business site.
2. Register the site in WoA Developer Blog Manager.
3. Take the site Atomic.
4. Install the Jetpack Beta tester plugin.

With "Default style" selected, verify the WordPress.com masterbar, "Browse sites", and the Site card all appear as expected:

CleanShot 2023-11-03 at 05 30 25@2x

With "Classic style" selected, verify the core toolbar appears as expected, and "Browse sites" no longer appears:

CleanShot 2023-11-03 at 06 33 04@2x

Clicking on "My Sites" should take you to the Sites page.

Lastly, verify the WordPress.com masterbar, "Browse sites", and the Site card all appear as expected on a Simple site.

### DIFF
--- a/projects/plugins/jetpack/changelog/improve-wp-admin-toolbar-mods
+++ b/projects/plugins/jetpack/changelog/improve-wp-admin-toolbar-mods
@@ -1,0 +1,4 @@
+Significance: minor
+Type: enhancement
+
+Always links 'My Sites' to the Sites page when 'wp-admin' is the admin interface.
--- a/projects/plugins/jetpack/modules/masterbar/admin-menu/admin-menu.js
+++ b/projects/plugins/jetpack/modules/masterbar/admin-menu/admin-menu.js
@@ -11,6 +11,11 @@
 			return;
 		}
 
+		var adminbarBlogMenuList = adminbar.querySelector( '#wp-admin-bar-blog > .ab-sub-wrapper ul' );
+		if ( ! adminbarBlogMenuList ) {
+			return;
+		}
+
 		function setAriaExpanded( value ) {
 			var anchors = adminbar.querySelectorAll( '#wp-admin-bar-blog a' );
 			for ( var i = 0; i < anchors.length; i++ ) {
--- a/projects/plugins/jetpack/modules/masterbar/admin-menu/class-atomic-admin-menu.php
+++ b/projects/plugins/jetpack/modules/masterbar/admin-menu/class-atomic-admin-menu.php
@@ -179,6 +179,11 @@ class Atomic_Admin_Menu extends Admin_Menu {
 			return;
 		}
 
+		// Unnecessary because "My Sites" always links to the Sites page.
+		if ( 'wp-admin' === get_option( 'wpcom_admin_interface' ) ) {
+			return;
+		}
+
 		// Add the menu item.
 		add_menu_page( __( 'site-switcher', 'jetpack' ), __( 'Browse sites', 'jetpack' ), 'read', 'https://wordpress.com/sites', null, 'dashicons-arrow-left-alt2', 0 );
 		add_filter( 'add_menu_classes', array( $this, 'set_browse_sites_link_class' ) );
@@ -220,6 +225,11 @@ class Atomic_Admin_Menu extends Admin_Menu {
 	 * Adds site card component.
 	 */
 	public function add_site_card_menu() {
+		// Unnecessary because core toolbar links to the site.
+		if ( 'wp-admin' === get_option( 'wpcom_admin_interface' ) ) {
+			return;
+		}
+
 		$default        = plugins_url( 'globe-icon.svg', __FILE__ );
 		$icon           = get_site_icon_url( 32, $default );
 		$blog_name      = get_option( 'blogname' ) !== '' ? get_option( 'blogname' ) : $this->domain;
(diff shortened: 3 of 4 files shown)


jeroenpf:
I've tested the change according to the testing steps provided and it works on both Atomic and Simple sites.

Could you explain a bit more why we are removing the site card? This is a place where we place nudges via JITM so I thought it might still offer value in that sense. I am unsure what 'core home affordance' is.

Another thing I just noticed now is that the My Sites menu item is highlighted:
CleanShot 2566-11-03 at 20 16 25@2x
But you are not actually on the My Sites page. Is this what we want?

Finally, a little nit pick: i feel that the Dashboard menu item is too close to the admin bar. When looking at the original menu, there is some space there. With the browse sites and site card removed, maybe there should be some margin added:

CleanShot 2566-11-03 at 20 18 34@2x

jeroenpf (dismissed):
Since this works as intended, i am approving this. I did leave some comments that we can take into consideration.

danielbachhuber:
> Could you explain a bit more why we are removing the site card? This is a place where we place nudges via JITM so I thought it might still offer value in that sense. I am unsure what 'core home affordance' is.

@jeroenpf Here's some context on the Site card: pdtkmj-1Vq-p2#comment-3533

Good point re: JITM, though. That's probably an argument in favor of keeping it. I'll go ahead and restore it, which probably resolves the styling issue.

Core home affordance:

CleanShot 2023-11-03 at 06 24 50@2x

danielbachhuber:
> I've tested the change according to the testing steps provided and it works on both Atomic and Simple sites.

@jeroenpf Don't _just_ follow the testing instructions, though — also figure out where the testing instructions are wrong 😁

jeroenpf:
> > I've tested the change according to the testing steps provided and it works on both Atomic and Simple sites.
> 
> @jeroenpf Don't _just_ follow the testing instructions, though — also figure out where the testing instructions are wrong 😁

Sure, for context here is what i did additionally:

- Since there were some JS changes, i also verified that the mobile menu is still working properly and checked the mobile style in general.
- as mentioned, looked at the style 
- Reviewed the code itself, not just try the changes
- Check that when going to a Calypso page, the site card and browse sites still appear and that the My Sites link is unaffected (which should be the case since we do not load the customized admin bar in Calypso).

danielbachhuber:
@jeroenpf Thanks!

wojtekn (approved):
> Always links 'My Sites' to the Sites page when 'wp-admin' is the selected admin interface:
> Because 'My Sites' always links to the Sites page, I'm also removing the 'Browse sites' link:

It works well for me: when I click "My sites", I navigate to the Sites page. Then, when I click on the site name, I'm back in the wp-admin dashboard.

I know it's out of the scope of this project, but it would be cool to remove the "Switch site" on Calypso views and link the My Sites to Sites page for consistency. (However, the current site switcher in the sidebar preserves context, so it's a bigger change).

ilyasfoo:
Hi, @danielbachhuber 👋 

Sorry for the random ping, we're experiencing a regression on the upsell menu ([issue](https://github.com/Automattic/wc-calypso-bridge/issues/1370)). While debugging, I found out that removing [this check](https://github.com/Automattic/jetpack/pull/33951/files#diff-7b09ba18eab949b6e4eb2bbf6a4f66005bcf90f9cf7e1f8c9406e50badbbbd95R14-R17) restores it. I'm not sure if it's something that my site is missing, but I think the site I've created is pretty generic so I believe there could be a bug somewhere.

To reproduce with the free domain nudge:

1. Create a free site https://wordpress.com/setup/free and skip the domain step
2. Upgrade to business plan
3. Transfer the site to Atomic by installing any plugin
4. Go to the calypso dashboard, i.e: https://wordpress.com/home/SITE_URL
5. Observe the `Free domain available` nudge is shown
6. Go to the wp-admin dashboard, i.e: https://SITE_URL/wp-admin/
7. Observe the `Free domain available` nudge does not appear

image

After removing the code via browser override:

image

danielbachhuber:
@ilyasfoo Sorry about that! 🙃 And great debugging 😊

I was overzealous in my early return. I think https://github.com/Automattic/jetpack/pull/34167 should fix it. Do you have a good way of testing that PR on a Woo Express trial site?

ilyasfoo:
Awesome, that's super quick! Thanks @danielbachhuber!

To test, you can set up a WoA free trial site using this guide: https://wp.me/peapX7-1D4. With that, you should have enough control to upload a custom Jetpack build, I assume. The nudge should be shown by default in all Woo Express free trial sites, so no further instructions to reproduce it there.

PR #33952, opened 2023-11-03
Boost: Implement `BackButton` in React

haqadn:
The back button interacts with navigation. However, the current usage can be simply handled with a `window.location.hash` change.

## Proposed changes:
* Import SVG files in React components
* Convert BackButton to React component

### Other information:

## Jetpack product discussion
No

No

## Testing instructions:
* Test the back button in boost UI


projects/plugins/boost/app/assets/src/js/utils/navigate.ts
```diff
@@ -0,0 +1,3 @@
+export const navigate = ( to: string ) => {
+	window.location.hash = to;
+};
```
pyronaur, line 3: Don't abstract 1 line of code if it's not reused at the moment, especially if you need to create yet another utils file to achieve this.

pyronaur (dismissed):

dilirity (approved):
:shipit: 

I'm so happy because of this - great work!

projects/plugins/boost/app/assets/src/js/utils/navigate.ts
haqadn: It is there as a stop gap. I want to be able to replace the logic inside it with react-navigator.

PR #33959, opened 2023-11-03
Jetpack Custom-CSS: Do not `@import` from the filesystem

anomiex:
## Proposed changes:

The SCSS and LESS libraries we're using by default can attempt to process scss or less files from the filesystem for `@import` directives. There shouldn't be any need for this, so to be safe let's turn it off.

### Other information:

## Jetpack product discussion

p1698952414557559-slack-CBG1CP4EN

No

## Testing instructions:

* Enable "Custom CSS", and install an older theme like TwentyTwentyOne.
* Go to the "Additional CSS" in customize.php.
* Create a file on your server, outside the web root, with some css that will do something obvious (e.g. `* { background-color: red !important; }`). Don't name it with ".css" though, call it something like "foo.xyz" (or "foo.scss" or "foo.less" maybe).
* Select SCSS, and create some custom CSS that starts with `@import "../relative/path/to/foo.xyz";`. The styles from foo.xyz should have effect with Jetpack trunk but not with this PR.
* Select LESS, and create some custom CSS that starts with `@import "/absolute/path/to/foo.xyz";`. The styles from foo.xyz should have effect with Jetpack trunk but not with this PR.


jeherve (dismissed):

jeherve:
This tests well for me, but will need a rebase before we can merge I'm afraid.

jeherve (approved):

PR #33622, opened 2023-10-16
Subscribers: Allow admins to see subscribe modal

edanzer:
## Proposed changes:

* Allow site administrators to see the subscriber modal/popup. We had hidden this because admins obviously do not want to subscribe. But based on feedback from Happiness Engineers, it has been confusing for site managers when they can't see/demo the modal on the front end.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

None.

## Does this pull request change what data or activity we track or use?

No.

## Testing instructions:

**Test on Jetpack self hosted:**
* You'll need to test on a jurassic connected site. 
* Go to Jetpack > Settings > Newsletters and enable the subscriber modal. 
* While logged in as an admin, go to a frontend post without a paywall or limited access, scroll, and confirm the modal loads. 

**Test on Simple sites:**
* Run the bin command below to load this branch in your sandbox. Be sure to sandbox public-api and the domain of your test site. 
* Go to Settings > Newsletters and enable the subscriber modal. 
* While logged in as an admin, go to a frontend post without a paywall or limited access, scroll, and confirm the modal loads.

PR #33967, opened 2023-11-06
Subscribe modal: don't show when previewing post or theme

simison:
Feedback via slack p1699147934734569-slack-CTG7Q97G8 shared that when previewing blog posts, they might see subscribe banner.

Screenshot 2023-11-06 at 11 16 29

Previously we never showed modal for the logged in admins even if the conditions otherwise were met, but we had lots of feedback that it was hard for people to test & review the modal. We started showing modal to admins, too in https://github.com/Automattic/jetpack/pull/33622. Now we just need to ensure the modal doesn't show in places where they don't expect, because they're reviewing something elese (like the theme or post).

Adding quite a few of the get-params that we use at .com for different preview iframes, just to cover any current unexpected situations as well any future ones.

## Proposed changes:

* Check for preview get params before deciding to show the modal

### Other information:

## Testing instructions:

* As logged in admin, enable subscribe modal
* Have a free blog post
* Ensure you haven't dismissed banner by removing cookie `jetpack_subscribe_modal_dismissed`
* Load the post and see the banner
* Add `?preview` (or other getters added in the PR) to the URL and see modal no more.

### DIFF
--- a/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
+++ b/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
@@ -169,6 +169,12 @@ HTML;
 			return false;
 		}
 
+		// Don't show when previewing blog posts or site's theme
+		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		if ( isset( $_GET['preview'] ) || isset( $_GET['theme_preview'] ) || isset( $_GET['customize_preview'] ) || isset( $_GET['hide_banners'] ) ) {
+			return false;
+		}
+
 		// Don't show if one of subscribe query params is set.
 		// They are set when user submits the subscribe form.
 		// The nonce is checked elsewhere before redirect back to this page with query params.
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -169,6 +169,12 @@ public function should_user_see_modal() {
 			return false;
 		}
 
+		// Don't show when previewing blog posts or site's theme
+		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		if ( isset( $_GET['preview'] ) || isset( $_GET['theme_preview'] ) || isset( $_GET['customize_preview'] ) || isset( $_GET['hide_banners'] ) ) {
```
lezama, line 174: @oskosk we could extract this to some shared place. What would be the best file?

lezama (approved):

projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
simison: Good follow-up 👍

janmckell:
There has been some feedback about how this appears, after the fix:
[https://wordpress.com/forums/topic/newsletter-pop-up-in-previews/?view=all#post-4001272](https://wordpress.com/forums/topic/newsletter-pop-up-in-previews/?view=all#post-4001272
)

simison:
@janmckell should be fixed with https://github.com/Automattic/jetpack/pull/33973

PR #33970, opened 2023-11-06
Super Cache: Fix null parameter warning when enabling/disabling caching

donnchawp:
This small fix will fix a PHP warning that the parameter being passed to is_dir() is NULL in the enable/disable caching functions.

## Proposed changes:
* In wp_super_cache_enable() and wp_super_cache_disable() check if $supercachedir is defined and set it if not.

### Other information:

## Jetpack product discussion
ref: https://wordpress.org/support/topic/deprecated-is_dir/

no

## Testing instructions:
* Tail your error_log
* If you have PHP8.2 installed, then you'll see a warning about is_dir() when caching is enabled or disabled from the settings page.
* If you don't have PHP8.2 installed, then error_log $supercachedir in wp_super_cache_enable() and wp_super_cache_disable() and you'll see it's blank when you toggle caching.
* Apply PR and error_log $supercachedir after the changes and it will be set to the path of your super cache directory.

### DIFF
--- a/projects/plugins/super-cache/wp-cache.php
+++ b/projects/plugins/super-cache/wp-cache.php
@@ -1859,6 +1859,10 @@ function wp_super_cache_enable() {
 
 	$super_cache_enabled = true;
 
+	if ( ! $supercachedir ) {
+		$supercachedir = get_supercache_dir();
+	}
+
 	if ( is_dir( $supercachedir . '.disabled' ) ) {
 		if ( is_dir( $supercachedir ) ) {
 			prune_super_cache( $supercachedir . '.disabled', true );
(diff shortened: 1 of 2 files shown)


projects/plugins/super-cache/wp-cache.php
dilirity, line 1: Seeing `get_supercache_dir` referenced here, and earlier usage in https://github.com/Automattic/jetpack/blob/fix/null_parameter_supercachedir/projects/plugins/super-cache/wp-cache.php#L946 - does it need to be wrapped in `function_exists` here as well?

dilirity (commented):
Haven't tested, just something I noticed after skimming through.

projects/plugins/super-cache/wp-cache.php
donnchawp: No need for it. I should remove that function_exists check because the function is in the phase2.php file and that's always loaded by wp-cache.php now.

dilirity (approved):
Works great! :shipit: 

No more notices.

I did see a notice on production super cache, that doesn't appear on this branch (while on PHP 8.2):

```
[07-Nov-2023 09:43:10 UTC] PHP Deprecated:  str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in /var/www/html/wp-content/plugins/wp-super-cache/wp-cache-phase2.php on line 54
```

PR #33972, opened 2023-11-06
Fix accept header check and add new wpsc_accept_list filter for sites to modify that list

### LINKED ISSUE #33968: WP Super Cache: do not serve cached content if client doesn't accept text/html
If the page /author/username is loaded by a browser, then that page is cached. If a Fediverse client subsequently tries to access it, the cached HTML page will be served. If there is no cached page, then JSON code is served.

When serving a cached page, the plugin should make sure that the client can accept text/html.

Ref: https://wordpress.org/support/topic/wp-supercache-seems-to-ignore-non-default-accept-header/

In the future, we should cache 'application/activity+json' requests:
* Create a "sanitized content-type" for the type of request. 'text/html' vs 'application/activity+json' to start with.
* Add content-type to the cache key
* Modify the "incomplete page" check, so it only operates on HTML pages.

donnchawp:
As reported [here](https://wordpress.org/support/topic/wp-supercache-seems-to-ignore-non-default-accept-header/), the author page sometimes shows HTML to Fediverse clients using ActivityPub.

If the /author/username page is cached already, then it is delivered to the client.
This PR fixes a check of the Accept header, so it recognises those clients correctly. The plugin already had code that checks the Accept header is "text/html" and refuses to serve a cached file if it's not. That check was getting an incorrect Accept header so the cached file was served, regardless.

It also adds a filter, wpsc_accept_list, so the list of Accept headers can be modified by the site owner.

Fixes #33968

## Proposed changes:
* In wpsc_get_accept_header() fix the strpos() that checks the Accept header, so it detects if the header is found in position 0 of the list item.
* In that same function add a filter, wpsc_accept_list, around the list of non text/html Accept headers so the site owner can modify it.

### Other information:

No

## Testing instructions:
1. Install this plugin and activate caching.
2. Go to /author/username on your site using a private/incognito browser to prime the cache. Confirm that the page is cached by examining the bottom of the source.
3. At a terminal, use this to test with an ActivityPub Accept header. Replace URL with your own test site: `curl -v -sS -H 'Accept: application/activity+json' https://example.com/author/username/`
4. HTML should be shown if the page is cached already.
5. Apply this PR.
6. Rerun the `curl` command, and it should show a JSON array with author information.
7. Reload the browser window and confirm again that the page is cached.

### DIFF
--- a/projects/plugins/super-cache/wp-cache-phase2.php
+++ b/projects/plugins/super-cache/wp-cache-phase2.php
@@ -510,6 +510,7 @@ function wpsc_get_accept_header() {
 
 	if ( $accept === 'N/A' ) {
 		$json_list = apply_filters( 'wpsc_accept_list', array( 'application/json', 'application/activity+json', 'application/ld+json' ) );
+		$json_list = array_map( 'strtolower', $json_list );
 		// phpcs:ignore WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
 		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.MissingUnslash -- $accept is checked and set below.
 		$accept = isset( $_SERVER['HTTP_ACCEPT'] ) ? strtolower( filter_var( $_SERVER['HTTP_ACCEPT'] ) ) : '';
--- a/projects/plugins/super-cache/wp-cache-phase2.php
+++ b/projects/plugins/super-cache/wp-cache-phase2.php
@@ -509,13 +509,13 @@ function wpsc_get_accept_header() {
 	static $accept = 'N/A';
 
 	if ( $accept === 'N/A' ) {
-		$json_list = apply_filters( 'wpsc_accept_list', array( 'application/json', 'application/activity+json', 'application/ld+json' ) );
-		$json_list = array_map( 'strtolower', $json_list );
+		$accept_headers = apply_filters( 'wpsc_accept_headers', array( 'application/json', 'application/activity+json', 'application/ld+json' ) );
+		$accept_headers = array_map( 'strtolower', $accept_headers );
 		// phpcs:ignore WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
 		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.MissingUnslash -- $accept is checked and set below.
 		$accept = isset( $_SERVER['HTTP_ACCEPT'] ) ? strtolower( filter_var( $_SERVER['HTTP_ACCEPT'] ) ) : '';
 
-		foreach ( $json_list as $header ) {
+		foreach ( $accept_headers as $header ) {
 			if ( strpos( $accept, $header ) !== false ) {
 				$accept = 'application/json';
 			}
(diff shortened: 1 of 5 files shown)


projects/plugins/super-cache/wp-cache-phase2.php
```diff
@@ -509,13 +509,13 @@ function wpsc_get_accept_header() {
...
 	if ( $accept === 'N/A' ) {
-		$json_list = array( 'application/json', 'application/activity+json', 'application/ld+json' );
+		$json_list = apply_filters( 'wpsc_accept_list', array( 'application/json', 'application/activity+json', 'application/ld+json' ) );
 		// phpcs:ignore WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
 		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.MissingUnslash -- $accept is checked and set below.
 		$accept = isset( $_SERVER['HTTP_ACCEPT'] ) ? strtolower( filter_var( $_SERVER['HTTP_ACCEPT'] ) ) : '';
 
 		foreach ( $json_list as $header ) {
-			if ( strpos( $accept, $header ) ) {
+			if ( strpos( $accept, $header ) !== false ) {
```
dilirity, line 519: `$accept` is lowercased, but since there's now a filter for `$json_list`, I think `$header` should be lowercased as well, or simply use `stripos`.

projects/plugins/super-cache/wp-cache-phase2.php
```diff
@@ -509,13 +509,13 @@ function wpsc_get_accept_header() {
 	static $accept = 'N/A';
 
 	if ( $accept === 'N/A' ) {
-		$json_list = array( 'application/json', 'application/activity+json', 'application/ld+json' );
+		$json_list = apply_filters( 'wpsc_accept_list', array( 'application/json', 'application/activity+json', 'application/ld+json' ) );
```
dilirity: Based on the filter name - `$json_list` is `accept list`. Or `$accepted_headers`.

dilirity (changes requested):
A few small improvements 🎈

projects/plugins/super-cache/wp-cache-phase2.php
donnchawp: Thanks, good catch!

projects/plugins/super-cache/wp-cache-phase2.php
donnchawp: Thanks. Did that, and renamed the filter too to "wpsc_accept_headers" to reflect what the filter is actually working on.

dilirity (approved):
:shipit:

alecmuffett:
EDIT: IGNORE THIS, IT TURNS OUT THAT THE QUOTE MARKS ARE UNICODE AND SOMEHOW THAT PASSED THRU THE COMMANDLINE TO CURL AND BROKE THE ACCEPT HEADER.  APOLOGIES.

- I performed:
`curl -v -sS -H ‘Accept: application/activity+json’ https://alecmuffett.com/article/author/alecm`
…and I received HTML.  Is there a deeper issue?

donnchawp:
@alecmuffett - phew, I'm glad that bug is fixed. :)

PR #33961, opened 2023-11-03
Subscribe Block: Improves "Subscribed" state in WPCOM

lezama:
Fixes #

## Proposed changes:

* Improves subscribe button behavior introduced in https://github.com/Automattic/jetpack/pull/33897
* See p1699026721433949-slack-C05PV073SG3

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* On a simple site check the subscribe button while logged in / logged out.

PR #33973, opened 2023-11-06
Subscribe Block: Always show email input if not subscribed

### LINKED ISSUE #33964: Subscribe Block does not show Email input field for logged-in users on wpcom Simple sites
### Impacted plugin

Jetpack

### Quick summary

When logged in to a WordPress.com account (or SU'd as a user), the Subscribe block doesn't show an email input field, only the submit button. This only occurs on simple sites, but appears to happen in all themes, both classic and whatever-we're-calling-FSE-now. The button still works, and will subscribe your current account to the blog, but without the input field it throws off the layout and _looks_ broken to the user who is comparing it to what they see in the editor.

### Steps to reproduce

1. on a WordPress.com Simple (IE not-Atomic) site, place the Subscribe block anywhere on the site:
2. Screenshot 2023-11-04 at 2 25 07 PM
3. using the same logged-in browser session go look at the Subscribe block, and you will only see the button, no input field
4. Screenshot 2023-11-04 at 2 25 22 PM

The block should look on the front-end of the site, the same way it does in the editor

### What actually happened

it looks bad and wrong because the input field is missing

### Impact

Most (> 50%)

### Available workarounds?

Simple

### Logs or notes

Some of my teammates reported still seeing the input field in when logged-in with browsers other than Chrome, but for me it is still missing in every browser I try.

It seems _possible_ to me that this is an expected behavior, but I could not find any documentation to that affect. If it is indeed the expected behavior, I would propose that it shouldn't be, because it completely changes the size and shape of the element which can have very ugly consequences on the page layout.

lezama:
Fixes #33964

## Proposed changes:

* Always Display the email input if the user is not subscribed, even when the user is logged in to WPCOM or has the subscription token on Jetpack sites.

### Other information:

## Jetpack product discussion

Follow up to #33961

## Testing instructions:

* Try to reproduce #33964.
* TLDR: If you are not a subscriber you should always see the email input and a "Subscribe" button.
* Confirm that when subscribing you get a "✓Subscribed" button without email input.
* Please test both in Jetpack and in WPCOM.

### DIFF
--- a/projects/plugins/jetpack/changelog/update-subscribe-block-always-show-input
+++ b/projects/plugins/jetpack/changelog/update-subscribe-block-always-show-input
@@ -1,0 +1,4 @@
+Significance: minor
+Type: other
+
+rollback change
--- a/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
+++ b/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
@@ -145,10 +145,7 @@ abstract class Token_Subscription_Service implements Subscription_Service {
 	/**
 	 * Retrieves the email of the currently authenticated subscriber.
 	 *
-	 * This function checks if the current user has an active subscription. If the user is subscribed,
-	 * their email is returned. Otherwise, it returns an empty string to indicate no active subscription.
-	 *
-	 * @return string The email address of the subscribed user or an empty string if not subscribed.
+	 * @return string The email address of the current user.
 	 */
 	public function get_subscriber_email() {
 		$email = $this->get_token_property( 'blog_subscriber' );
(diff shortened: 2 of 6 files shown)


simison (changes requested):
Good now in terms of what shows up in different states.

Tested subscriber only and paid posts. Tested as logged out, and logged in, and subscribed states.

Tested paywall and tested subscribe block outside paywall.

Only thing that needs fixing is visual problem within subscriber-only paywall when I'm logged out:

Screenshot 2023-11-07 at 12 01 30

Otherwise all good:

Screenshot 2023-11-07 at 11 59 31
Screenshot 2023-11-07 at 11 59 27
Screenshot 2023-11-07 at 11 59 00
Screenshot 2023-11-07 at 11 58 19

Important follow up will be to show "Become paid subscriber" at least for those paid paywalls, or the whole site when paid plans are set up.

I did test that the modal works to become paid, tho:

image

simison (approved):
Tests well now with Firefox

PR #33974, opened 2023-11-06
wp-admin Site Default: Add back the Calypso stats menu item when admin interface is Classic

jeroenpf:
Fixes https://github.com/Automattic/dotcom-forge/issues/4456

The stats menu item is missing from the admin menu when the site is WoA and the admin interface style is set to wp-admin. This was by design, but we decided it was best to add it back.

## Proposed changes:

* Restore the top-level Stats menu item

Note that this essentially is a revert of a change made recently as part of the wp-admin Site default project where we decided to hide the menu option. The logic that was put in place to hide this menu item has been removed.

### Other information:

## Testing instructions:

* Go to a WoA blog and apply this PR. 
* Ensure that you see the Stats menu item at the top level of the admin menu.
* Check that the stats menu option appears for both the Classic and Default admin interface styles by switching back and forth between them.
* Also check that the menu still works properly for simple sites.
* Finally, check that all the links in the menu point to the expected pages, depending on which admin interface style is selected. Try toggling the per-page view option (classic, default) to make sure that works as intended as well.
* There is no need to test this with SSO turned off. No navigation customizations are applied if SSO is disabled.


danielbachhuber (approved):
Works well 👍 

I verified the Stats menu appears as expected with "Classic style" selected:

CleanShot 2023-11-06 at 13 15 34@2x

PR #33979, opened 2023-11-06
Repo Gardening: also gather p2 comments as support references

jeherve:
## Proposed changes:

Until now, when collecting support references (and consequently when judging whether an issue should be escalated or not), we looked at the number of tickets that had been posted on the issue.
However, in some cases user feedback may come as links to p2 comments instead of support tickets. Let's count those as support references as well.

### Other information:

## Jetpack product discussion

Related conversation: p1698810312902799-slack-CQD1HH4MA

* No

## Testing instructions:

This cannot be tested on this repo before merged, but you can test it in a forked repo. Here is an example:
https://github.com/jeherve/jetpack/issues/114#issuecomment-1796001066

### DIFF
--- a/projects/github-actions/repo-gardening/src/tasks/gather-support-references/index.js
+++ b/projects/github-actions/repo-gardening/src/tasks/gather-support-references/index.js
@@ -88,7 +88,7 @@ async function getIssueReferences( octokit, owner, repo, number, issueComments )
 
 		// xxx-zen is the preferred format for tickets.
 		// xxx-zd, as well as its uppercase version, is considered an alternate version.
-		const wrongId = supportId.match( /([0-9]*)-zd/i );
+		const wrongId = supportId.match( /^([0-9]*)-zd$/i );
 		if ( wrongId ) {
 			const correctedId = `${ wrongId[ 1 ] }-zen`;
 			correctedSupportIds.add( correctedId );
--- a/projects/github-actions/repo-gardening/src/tasks/gather-support-references/index.js
+++ b/projects/github-actions/repo-gardening/src/tasks/gather-support-references/index.js
@@ -50,7 +50,7 @@ async function getListComment( issueComments ) {
  */
 async function getIssueReferences( octokit, owner, repo, number, issueComments ) {
 	const ticketReferences = [];
-	const referencesRegexP = /([0-9]*-(?:zen|zd)|[a-zA-Z0-9-]+-p2#comment-[0-9]*)/gim;
+	const referencesRegexP = /[0-9]*-(?:zen|zd)|[a-zA-Z0-9-]+-p2#comment-[0-9]*/gim;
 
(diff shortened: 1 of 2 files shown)


projects/github-actions/repo-gardening/src/tasks/gather-support-references/index.js
```diff
@@ -44,7 +50,7 @@ async function getListComment( issueComments ) {
  */
 async function getIssueReferences( octokit, owner, repo, number, issueComments ) {
 	const ticketReferences = [];
-	const referencesRegexP = /[0-9]*-(?:zen|zd)/gim;
+	const referencesRegexP = /([0-9]*-(?:zen|zd)|[a-zA-Z0-9-]+-p2#comment-[0-9]*)/gim;
```
anomiex: Is it intentional that we're only doing P2 comments, and not references to entire P2 posts?

Also, I don't think the extra parens are needed, this should work
```suggestion
	const referencesRegexP = /[0-9]*-(?:zen|zd)|[a-zA-Z0-9-]+-p2#comment-[0-9]*/gim;
```

projects/github-actions/repo-gardening/src/tasks/gather-support-references/index.js
```diff
@@ -74,6 +80,12 @@ async function getIssueReferences( octokit, owner, repo, number, issueComments )
...
 
+		// Do not correct p2 comment references.
+		if ( supportId.includes( '-p2#comment-' ) ) {
+			correctedSupportIds.add( supportId );
+			return;
+		}
+
 		// xxx-zen is the preferred format for tickets.
 		// xxx-zd, as well as its uppercase version, is considered an alternate version.
 		const wrongId = supportId.match( /([0-9]*)-zd/i );
```
anomiex: We should probably anchor this pattern
```suggestion
		const wrongId = supportId.match( /^([0-9]*)-zd$/i );
```
After doing that, the check being added above shouldn't be needed since there's no way this could match.

projects/github-actions/repo-gardening/src/tasks/gather-support-references/index.js
jeherve: > Is it intentional that we're only doing P2 comments, and not references to entire P2 posts?

Yes, I chose to do that since we often use references to p2 posts in issues but they're not always support references. Comments on p2 posts, when we refer to them in issues, can almost always be considered as support references though, in my experience.

anomiex (approved):
Code looks good to me. Since you already have it set up, would you mind pulling the latest changes into your fork and running the test again to verify before we merge?

jeherve:
> Since you already have it set up, would you mind pulling the latest changes into your fork and running the test again to verify before we merge?

👍 It still works: https://github.com/jeherve/jetpack/issues/114#issuecomment-1796001066

PR #33980, opened 2023-11-06
get_author: wpcom: Add a new wpcom_pre_get_active_blog_for_user filter

mreishus:
## Proposed changes:

* Inside the WPCOM only section of `get_author()`, add a new filter called `wpcom_pre_get_active_blog_for_user` which can optionally bypass the call to `get_active_blog_for_user()`.
  * WPCOM has some special users with over 50,000 blogs that this can help out with. 

### Other information:

No

## Testing instructions:

* Unit tests + Manual inspection. 
  * If JP is not running on WPCOM, this patch should have no change in behavior. 
  * If JP is running on WPCOM but there is no filter attached, there should be no change in behavior.

```
scp projects/plugins/jetpack/class.json-api-endpoints.php wpcom-sandbox:./public_html/wp-content/mu-plugins/jetpack-plugin/sun/class.json-api-endpoints.php
scp projects/plugins/jetpack/class.json-api-endpoints.php wpcom-sandbox:./public_html/wp-content/mu-plugins/jetpack-plugin/moon/class.json-api-endpoints.php
```
API Console: GET 1.2 `/read/following?locale=en&meta=site%2Cfeed&number=40&order=DESC`

Should continue to see author information

### DIFF
--- a/projects/plugins/jetpack/class.json-api-endpoints.php
+++ b/projects/plugins/jetpack/class.json-api-endpoints.php
@@ -1455,7 +1455,9 @@ abstract class WPCOM_JSON_API_Endpoint {
 				if ( false === $active_blog ) {
 					$active_blog = get_active_blog_for_user( $id );
 				}
-				$site_id = $active_blog->blog_id;
+				if ( ! empty( $active_blog ) ) {
+					$site_id = $active_blog->blog_id;
+				}
 				if ( $site_id > -1 ) {
 					$site_visible = (
 						-1 !== (int) $active_blog->public ||
--- a/projects/plugins/jetpack/class.json-api-endpoints.php
+++ b/projects/plugins/jetpack/class.json-api-endpoints.php
@@ -1451,6 +1451,16 @@ abstract class WPCOM_JSON_API_Endpoint {
 				$nice       = $user->user_nicename;
 			}
 			if ( defined( 'IS_WPCOM' ) && IS_WPCOM && ! $is_jetpack ) {
+				/**
+				 * Allow customizing the blog ID returned with the author in WordPress.com REST API queries.
+				 *
+				 * @since $$next-version$$
+				 *
+				 * @module json-api
+				 *
+				 * @param bool|int $active_blog  Blog ID, or false by default.
+				 * @param int      $id           User ID.
+				 */
 				$active_blog = apply_filters( 'wpcom_pre_get_active_blog_for_user', false, $id );
 				if ( false === $active_blog ) {
 					$active_blog = get_active_blog_for_user( $id );
--- a/projects/plugins/jetpack/class.json-api-endpoints.php
+++ b/projects/plugins/jetpack/class.json-api-endpoints.php
@@ -1451,6 +1451,8 @@ abstract class WPCOM_JSON_API_Endpoint {
 				$nice       = $user->user_nicename;
 			}
 			if ( defined( 'IS_WPCOM' ) && IS_WPCOM && ! $is_jetpack ) {
+				$site_id = -1;
+
 				/**
 				 * Allow customizing the blog ID returned with the author in WordPress.com REST API queries.
 				 *
@@ -1461,8 +1463,6 @@ abstract class WPCOM_JSON_API_Endpoint {
 				 * @param bool|int $active_blog  Blog ID, or false by default.
 				 * @param int      $id           User ID.
 				 */
-				$site_id = -1;
-
 				$active_blog = apply_filters( 'wpcom_pre_get_active_blog_for_user', false, $id );
 				if ( false === $active_blog ) {
 					$active_blog = get_active_blog_for_user( $id );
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/class.json-api-endpoints.php
```diff
@@ -1451,8 +1451,11 @@ public function get_author( $author, $show_email_and_ip = false ) {
 				$nice       = $user->user_nicename;
 			}
 			if ( defined( 'IS_WPCOM' ) && IS_WPCOM && ! $is_jetpack ) {
-				$active_blog = get_active_blog_for_user( $id );
-				$site_id     = $active_blog->blog_id;
+				$active_blog = apply_filters( 'wpcom_pre_get_active_blog_for_user', false, $id );
+				if ( false === $active_blog ) {
+					$active_blog = get_active_blog_for_user( $id );
+				}
+				$site_id = $active_blog->blog_id;
```
andres-blanco: I was looking at `get_active_blog_for_user` code: fbhepr%2Skers%2Sjcpbz%2Sjc%2Qvapyhqrf%2Szf%2Qshapgvbaf.cuc%3Se%3Qrp171832%26zb%3Q927%26sv%3Q45%2348-og and it looks like it can return a `void` value if user has no blogs. Maybe it should go into another PR, but we should definitively check here to see if `$active_blog` is a `WP_Site` object.

mreishus: Added check here b5615314159f98dda12785dbef4efe2dc2d3b055

projects/plugins/jetpack/class.json-api-endpoints.php
```diff
@@ -1451,8 +1451,13 @@ public function get_author( $author, $show_email_and_ip = false ) {
...
 			if ( defined( 'IS_WPCOM' ) && IS_WPCOM && ! $is_jetpack ) {
-				$active_blog = get_active_blog_for_user( $id );
-				$site_id     = $active_blog->blog_id;
+				$active_blog = apply_filters( 'wpcom_pre_get_active_blog_for_user', false, $id );
+				if ( false === $active_blog ) {
+					$active_blog = get_active_blog_for_user( $id );
+				}
+				if ( ! empty( $active_blog ) ) {
+					$site_id = $active_blog->blog_id;
+				}
```
daledupreez: I think we also need to initialize `$site_id` to make sure it's not undefined, especially given that we're now setting it inside the `if` block.
```suggestion
				$site_id = -1;

				$active_blog = apply_filters( 'wpcom_pre_get_active_blog_for_user', false, $id );
				if ( false === $active_blog ) {
					$active_blog = get_active_blog_for_user( $id );
				}
				if ( ! empty( $active_blog ) ) {
					$site_id = $active_blog->blog_id;
				}
```

daledupreez (commented):
One small note from me about a code path that leaves `$site_id` undefined.

projects/plugins/jetpack/class.json-api-endpoints.php
```diff
@@ -1451,8 +1451,13 @@ public function get_author( $author, $show_email_and_ip = false ) {
 				$nice       = $user->user_nicename;
 			}
 			if ( defined( 'IS_WPCOM' ) && IS_WPCOM && ! $is_jetpack ) {
-				$active_blog = get_active_blog_for_user( $id );
-				$site_id     = $active_blog->blog_id;
+				$active_blog = apply_filters( 'wpcom_pre_get_active_blog_for_user', false, $id );
```
jeherve: Could you add a docblock, to make our parser happy?

```suggestion
				/**
				 * Allow customizing the blog ID returned with the author in WordPress.com REST API queries.
				 *
				 * @since $$next-version$$
				 *
				 * @module json-api
				 *
				 * @param bool|int $active_blog  Blog ID, or false by default.
				 * @param int      $id           User ID.
				 */
				$active_blog = apply_filters( 'wpcom_pre_get_active_blog_for_user', false, $id );
```

Maybe we could also update the filter name to make it clearer where the filter could be used?

projects/plugins/jetpack/class.json-api-endpoints.php
andres-blanco: Nice thanks! Dale had another comment on this check so I am closing this as resolved

projects/plugins/jetpack/class.json-api-endpoints.php
mreishus: Sounds good. I mistakenly thought the init on line 1394 applies, but it's actually within a different branch.

projects/plugins/jetpack/class.json-api-endpoints.php
mreishus: Done. Do you have any suggestions for the filter name?

jeherve: Maybe we could mention something about this being used in the API, and author-related? Something like `wpcom_api_pre_get_active_blog_author`? What do you think?

andres-blanco (dismissed):
LGTM

projects/plugins/jetpack/class.json-api-endpoints.php
mreishus: Sounds good to me!

mreishus:
Rebased + added [37f1e80](https://github.com/Automattic/jetpack/pull/33980/commits/37f1e80a8644b7321b1f08074ca50ab41f7d2536) (rename)

jeherve (approved):

Issue #33358, opened 2023-09-27
AI Tiered Plans: Update upgrade interstitial to show the next tier available


lhkowalski:
On the MVP, the upgrade interstitial will show the details for the next plan available for purchase. This was called Linear Flow.

- When the user has a free plan, the interstitial will show the 100 requests upgrade
- When the user has the 100 requests plan, the interstitial will show the 200 requests upgrade
- When the user has the 200 requests plan, the interstitial will show the 500 requests upgrade
- When the user has the 500 requests plan, the interstitial will show the "Contact Us" call to action so it can be sent to a sales pipeline [target of the button to be defined]

The plan being purchased is the same (it's a plan created as a Upgradeable Quantity plan), so the tiers are set using a quantity value (200 requests or 200 units or 2 units, TBD) that will then reflect on the price being charged on the checkout (similar to DIFM extra pages).

---

Check the design wrap up post: p1HpG7-oR3-p2

---

This is required for the MVP.

dhasilva:
- [x] Base component and logic [#33981]
- [x] TOS removal [#34076]
- [x] Linking the cta buttons to the right tier checkout page [p1699643698953239/1699642722.303119-slack-C054LN8RNVA] [#34177]
- [x] Show prices on different usage tiers [#34196]
- [x] Linking the "Contact us" button to the correct action [pbNhbs-8Iv-p2#comment-18831] [#34240]
- [x] Update the AI card to show "Upgrade" instead of "Manage" when the user has a limited tier [#34215]

PR #33981, opened 2023-11-06
AI Assistant: Change image and text for interstitial based on tiers

dhasilva:
Part of https://github.com/Automattic/jetpack/issues/33358

## Proposed changes:

* Adds a JetpackAIInterstitialMoreRequests component for the custom last upgrade
* Sets the features and description according to the next tier
* Changes the interstitial image

### Other information:

#### On a site without the AI plan, go to My Jetpack
* Go to the purchase page, and check that the only change is the image
* Check that the purchase happens as normal and the `Manage` button is still inactive
* Check that there are no regressions on other interstitial pages

#### On local dev environment
* Sandbox your public API and change the `$current_licensed_quantity` to a known tier
* Enable the tiers on your 0-sandbox file
* Manually go to `?page=my-jetpack#/add-jetpack-ai`
* Check that the unlimited tier is shown regardless of the licensed quantity
* Change the return value of `get_next_usage_tier` to the commented code
* Check that the next tier is show

Not included: The contact button link and the checkout URLs for the tiers


projects/packages/my-jetpack/src/products/class-jetpack-ai.php
```diff
@@ -74,6 +74,29 @@ public static function get_title() {
...
+			return 1;
+		}
+
+		// If the current tier is 1 or 500, there is no next tier.
+		if ( 1 === $current_tier || 500 === $current_tier ) {
+			return null;
+		}
+
+		// phpcs:ignore Squiz.PHP.CommentedOutCode.Found
+		// return $info['next-tier']['value'];
```
lhkowalski, line 96: I think we are close to make this line work even without the tiers in place 🤞

projects/packages/my-jetpack/_inc/components/product-interstitial/index.jsx
```diff
@@ -268,11 +265,26 @@ export function ExtrasInterstitial() {
...
 export function JetpackAIInterstitial() {
+	const slug = 'jetpack-ai';
+	const { detail } = useProduct( slug );
+	const { onClickGoBack } = useGoBack( { slug } );
+
+	const currentTier = detail?.[ 'ai-assistant-feature' ]?.[ 'current-tier' ]?.value;
+	const hasNextTier = ! [ 1, 500 ].includes( currentTier );
+
+	if ( ! hasNextTier ) {
+		return <JetpackAIInterstitialMoreRequests onClickGoBack={ onClickGoBack } />;
```
lhkowalski, line 276: In this case, are we going to show the `JetpackAIInterstitialMoreRequests` when the user has the unlimited plan?

lhkowalski:
Code looks very good. Going to test it now.

lhkowalski (approved):
I tested it and it's working as expected.

When the tiers are disabled, on a free site, I get the current interstitial:

Screenshot 2023-11-10 at 11 55 04

---

On my test site, with the line commented, I always see the same interstitial no matter the tier I have on the site:

Screenshot 2023-11-10 at 12 22 41

The only different that I noted is the absense of the price on the interstitial.

---

Still on my testing site, now with the line uncommented, I can see the progression of the interstitials:

Screenshot 2023-11-10 at 12 27 41

Screenshot 2023-11-10 at 12 27 50

Screenshot 2023-11-10 at 12 28 13

Screenshot 2023-11-10 at 12 28 38

Screenshot 2023-11-10 at 12 29 03

---

With the unlimited plan active and the line uncommented, I see the same interstitial that we show for the 500 plan, asking for contact. I think the user will never land in this page on this scenario, so it's ok for me.

---

Very nice work overal, that looked like a hard problem to solve and you found a pretty neat way of doing it. Also, well done on moving the "go back" feature to a custom hook and documenting it 💯

projects/packages/my-jetpack/_inc/components/product-interstitial/index.jsx
dhasilva: As it is, yes, but the page is not linked to once a user has a plan, so we need to think about how to migrate users.
Which reminds me that the `Manage` button should be upgraded as well.

PR #33984, opened 2023-11-07
Connection: Add "source" as a query param to the Jetpack connect URL

yashwin:
Related to https://github.com/Automattic/jetpack-manage/issues/34

## Proposed Changes

This PR adds "source" as a query param to the Jetpack connect URL.

### Other information:

No

## Testing instructions:

* Go to Jetpack Live Branch with Jetpack
* Append the URL with `wp-admin/admin.php?page=jetpack&connect_url_redirect=true&jetpack_connect_login_redirect=true&source=jetpack-manage`
* Verify the link you are redirected(/jetpack/connect/authorize) to has source=jetpack-manage in it.

### DIFF
--- a/projects/packages/connection/src/class-manager.php
+++ b/projects/packages/connection/src/class-manager.php
@@ -1916,8 +1916,7 @@ class Manager {
 				'site_created'          => $this->get_assumed_site_creation_date(),
 				'allow_site_connection' => ! $this->has_connected_owner(),
 				'calypso_env'           => ( new Host() )->get_calypso_env(),
-				// phpcs:disable WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
-				'source'                => isset( $_GET['source'] ) ? wp_unslash( $_GET['source'] ) : '',
+				'source'                => ( new Host() )->get_source_query(),
 			)
 		);
 
--- a/projects/packages/connection/src/class-manager.php
+++ b/projects/packages/connection/src/class-manager.php
@@ -1917,7 +1917,7 @@ class Manager {
 				'allow_site_connection' => ! $this->has_connected_owner(),
 				'calypso_env'           => ( new Host() )->get_calypso_env(),
 				// phpcs:disable WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
-				'source'                => isset( $_GET['source'] ) ? wp_unslash( $_GET['source'] ) : null,
+				'source'                => isset( $_GET['source'] ) ? wp_unslash( $_GET['source'] ) : '',
 			)
 		);
 
--- a/projects/packages/connection/src/class-manager.php
+++ b/projects/packages/connection/src/class-manager.php
@@ -1916,6 +1916,8 @@ class Manager {
 				'site_created'          => $this->get_assumed_site_creation_date(),
 				'allow_site_connection' => ! $this->has_connected_owner(),
 				'calypso_env'           => ( new Host() )->get_calypso_env(),
+				// phpcs:disable WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+				'source'                => isset( $_GET['source'] ) ? wp_unslash( $_GET['source'] ) : null,
 			)
 		);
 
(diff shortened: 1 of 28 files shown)


jkguidaven (dismissed):
Tested this with https://github.com/Automattic/wp-calypso/pull/83963 and I can confirm the source query param is now included. 👍 LGTM

projects/packages/connection/src/class-manager.php
```diff
@@ -1916,6 +1916,8 @@ public function get_authorization_url( $user = null, $redirect = null ) {
 				'site_created'          => $this->get_assumed_site_creation_date(),
 				'allow_site_connection' => ! $this->has_connected_owner(),
 				'calypso_env'           => ( new Host() )->get_calypso_env(),
+				// phpcs:disable WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+				'source'                => isset( $_GET['source'] ) ? wp_unslash( $_GET['source'] ) : '',
```
jeherve: For some other parameters, like the calypso env above, we rely on an allowlist of different options instead of allowing just about any string. Would it make sense here as well?

jeherve (commented):
This seems to make sense. I was wondering if it would be an option to rely on the existing `get_activation_source` instead of a query string, but I suppose that's a bit different here.

cc @sergeymitr for visibility.

projects/packages/connection/src/class-manager.php
yashwin: Thanks for the review, @jeherve!

> For some other parameters, like the calypso env above, we rely on an allowlist of different options instead of allowing just about any string. Would it make sense here as well?

Agree. I have updated it now: https://github.com/Automattic/jetpack/pull/33984/commits/56a6afcaf672c776ba5e79d1b5dead552fde268d.

projects/packages/status/src/class-host.php
```diff
@@ -119,4 +119,19 @@ public function get_calypso_env() {
 
 		return '';
 	}
+
+	/**
+	 * Return source query param value from the URL if exists in the allowed sources list.
+	 *
+	 * @return string "source" query param value
+	 */
+	public function get_source_query() {
```
jeherve, line 128: This  looks good to me. What do you think about adding a test for it in `projects/packages/status/tests/php/test-host.php`, to ensure we continue having proper coverage of the different methods in that class?

Thank you!

jeherve (changes requested):

projects/packages/status/src/class-host.php
yashwin: Thank you! I'm not familiar with tests in PHP. I'll figure it out and get back to this.

jeherve: Something like this should do the trick:

```diff
diff --git a/projects/packages/status/tests/php/test-host.php b/projects/packages/status/tests/php/test-host.php
index 9d29142ce1..6986241afc 100644
--- a/projects/packages/status/tests/php/test-host.php
+++ b/projects/packages/status/tests/php/test-host.php
@@ -132,6 +132,21 @@ class Test_Host extends TestCase {
 		$this->assertEquals( $env, $this->host_obj->get_calypso_env() );
 	}
 
+	/**
+	 * Test adding a source parameter to the Calypso URL.
+	 *
+	 * @covers Automattic\Jetpack\Status\Host::get_source_query
+	 * @dataProvider get_source_query_params
+	 *
+	 * @param string $source Source parameter.
+	 * @param string $expected Expected query string.
+	 */
+	public function test_get_source_query( $source, $expected ) {
+		$_GET['source'] = $source;
+		$this->assertEquals( $expected, $this->host_obj->get_source_query( $source ) );
+		unset( $_GET['source'] );
+	}
+
 	/**
 	 * Data provider for `test_get_calypso_env()` test method.
 	 *
@@ -145,4 +160,17 @@ class Test_Host extends TestCase {
 			'default'     => array( '' ),
 		);
 	}
+
+	/**
+	 * Data provider for `test_get_source_query()` test method.
+	 *
+	 * @return array
+	 */
+	public function get_source_query_params() {
+		return array(
+			'empty'   => array( '', '' ),
+			'valid'   => array( 'jetpack-manage', 'jetpack-manage' ),
+			'invalid' => array( 'invalid-param', '' ),
+		);
+	}
 }

```

Let me know what you think!

jeherve (changes requested):

jeherve (approved):

projects/packages/status/src/class-host.php
yashwin: Thank you so much, @jeherve! 🙇

PR #33987, opened 2023-11-07
Boost: Update header component

haqadn:
## Proposed changes:
* Convert the Header component to React

### Other information:

## Jetpack product discussion
pc9hqz-2jy-p2

No

## Testing instructions:
* Browse through different pages in Boost dashboard and make sure the header component is working as expected.


projects/plugins/boost/app/assets/src/js/pages/benefits/BenefitsInterstitial.svelte
```diff
@@ -27,18 +28,12 @@
 	if ( ! ( 'yearly' in pricing ) ) {
 		goToCheckout();
 	}
+
+	const activateLicense = React.createElement( ActivateLicense );
 </script>
 
 <div id="jb-dashboard" class="jb-dashboard">
-	<div class="jb-dashboard-header jb-benefits-header">
-		<div class="jb-container jb-container--fixed">
```
dilirity, line 34: Removing these classes caused a change in the header (notice the spacing on the left of the back arrow on current):

trunk:
CleanShot 2023-11-13 at 17 43 46

current:
CleanShot 2023-11-13 at 17 44 15

However, on the current version, the header is the same across all Boost pages, so it makes sense to not have custom styles for it. Just in case, pointing it out.

projects/plugins/boost/app/assets/src/js/react-components/sections/header/index.tsx
```diff
@@ -0,0 +1,55 @@
+import classNames from 'classnames';
+import React from 'react';
+import { navigate } from '../../../utils/navigate';
+import BackButton from '../../components/back-button';
+import ChevronRight from '../../svg/chevron-right';
+import Logo from '../../svg/logo';
+import styles from './styles.module.scss';
+
+interface HeaderProps {
```
dilirity: I'm still fuzzy on the details, but I feel we should be consistent when declaring types. In the back-button, we use `type`, and here `interface`.

projects/plugins/boost/app/assets/src/js/react-components/sections/header/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+							</div>
+							<div className={ classNames( styles.subpage ) }>{ subPageTitle }</div>
+						</>
+					) }
+				</div>
+
+				{ children }
+			</div>
+
+			{ subPageTitle !== '' && (
```
dilirity, line 46: Just an observation - maybe a separate prop for this can be used? That way, we can show the back button without having a sub title.

dilirity (changes requested):
I've noticed a few things.

I also merged it with trunk so we have the up-to-date version in here.

projects/plugins/boost/app/assets/src/js/pages/benefits/BenefitsInterstitial.svelte
haqadn: The original design didn't have the header in the upgrade page. It was later added to include the license activation link. So, I don't mind changing the width and make it consistent, same as you.

However, I notice from your screenshot that the background color was missing from the header. This was due to CSS class name mistake which I have fixed now in https://github.com/Automattic/jetpack/pull/33987/commits/b51a64b5b0b763f86bd90018e362ec7d506dc038.

projects/plugins/boost/app/assets/src/js/react-components/sections/header/index.tsx
haqadn: Updated https://github.com/Automattic/jetpack/pull/33987/commits/efb87f2e847d447cc16c1abb63f6100426b42c8c

projects/plugins/boost/app/assets/src/js/react-components/sections/header/index.tsx
haqadn: I think we should add the prop when we need to.

projects/plugins/boost/app/assets/src/js/pages/benefits/BenefitsInterstitial.svelte
dilirity: Oh wow - nice catch!

dilirity (approved):
:shipit:

PR #33995, opened 2023-11-07
Subscription Block: Enhance button text and styles.

lezama:
Fixes
## Proposed changes:

* Improve button text based on the user's subscription status. Subscribe - Upgrade subscription - Subscribed.

### Other information:

## Jetpack product discussion

peKye1-sP-p2

## Testing instructions:

* Test the block everywhere 😅 – different browsers – non sub – subscriber – paid subscriber – subs only post – paid only posts

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -707,7 +707,7 @@ function render_for_website( $data, $classes, $styles ) {
 								<?php endif; ?>
 								name="jetpack_subscriptions_widget"
 							>
-								<?php echo esc_html( $button_text ); ?>
+								<?php echo esc_html( sanitize_submit_text( $button_text ) ); ?>
 							</button>
 						</p>
 					</div>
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -740,7 +737,7 @@ class="<?php echo esc_attr( $classes['submit_button'] ); ?>"
 function render_for_email( $data, $styles ) {
 	$submit_button_wrapper_style = ! empty( $styles['submit_button_wrapper'] ) ? 'style="' . esc_attr( $styles['submit_button_wrapper'] ) . '"' : '';
 	$button_text                 = wp_kses(
-		html_entity_decode( $data['submit_button_text'], ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401 ),
+		html_entity_decode( get_submit_button_text( $data ), ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401 ),
```
simison: Looks like `sanitize_submit_text` already handles `html_entity_decode` and `wp_kses`?

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -912,6 +909,37 @@ function get_current_url() {
...
+/**
+ * Get the submit button text based on the subscription status.
+ *
+ * @param array $data Array containing block view data.
+ *
+ * @return string
+ */
+function get_submit_button_text( $data ) {
+	if ( ! Jetpack_Memberships::is_current_user_subscribed() ) {
+		return sanitize_submit_text( $data['submit_button_text'] );
```
simison: I think it would be fine to keep the sanitization closer to output.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
lezama: oh yes

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
lezama: good call

simison (dismissed):
"Subscribe/Subscribed" works well, and the the paid subscriptions was problematic for other reasons so couldn't test that now (p1699378863370269-slack-C01B6KEJ5GE)

edanzer (approved):
Testing as expected and code changes look mostly straightforward. 
   - For non-subscribed users, we get the normal subscribe block input field plus 'Subscribe' button.
   - For subscribed users, the input field in the subscribe block goes away, and button text is 'Subscribed'
   - If a user is a free subscriber, and tries to access content that is only for paid subscribe, we get the paywall block but with an 'Upgrade subscription' button. 
   
For the third case above, there is a small issue with the button alignment. Looks like below and is being addressed separately in https://github.com/Automattic/jetpack/pull/34003.

paywall upgrade

One other note for posterity - I wondered why the sanitization was needed as I thought we were providing the button text. But users can edit the button text in the site editor, so we need to account for user-provided values.

PR #33997, opened 2023-11-07
Remove/unneeded condition from product data api

CodeyGuyDylan:
## Proposed changes:

The backend of the product data is being stripped a bit since we are not pulling live activity log data from that endpoint. The last_finished_backup_time datapoint is being removed, and this is not used in any previous version of the plugin since that change has not yet been released in a Jetpack version, so we should be good to remove it

### Other information:

- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a P2 generated comment below with a script to run)?

## Jetpack product discussion

P2: p1HpG7-pn9-p2

No

## Testing instructions:

1. Checkout this branch
2. Get your local docker environement up and running
3. Point your local environment to your sandbox ( I have to enable global-http to get this working sometimes)
4. Patch D127975-code to your sandbox
5. Without VaultPress backup being installed, go to My Jetpack
6. You should see the unpurchased state of the card with a summary of what can be backed up on the site (if you had vaultpress backup and just cancelled it, I had some caching issues with it still returning as "active" on the backend, wasn't able to track down where that cache is set yet)
image
7. Purchase a VaultPress backup subscription and wait for there to be a backup. Add an asset like an image or video and wait for a realtime backup to be created.
8. Head back to My Jetpack and you should see the purchased state of the card
image

### DIFF
--- a/projects/packages/my-jetpack/src/class-rest-product-data.php
+++ b/projects/packages/my-jetpack/src/class-rest-product-data.php
@@ -64,7 +64,6 @@ class REST_Product_Data {
 
 		// If site has backups and the realtime backup capability, add latest undo event
 		if (
-			$body->backups->last_finished_backup_time &&
 			in_array( 'backup-realtime', $capabilities->data['capabilities'], true )
 		) {
 			$body->backups->last_undoable_event = self::get_site_backup_undo_event();
--- a/projects/packages/my-jetpack/src/class-rest-product-data.php
+++ b/projects/packages/my-jetpack/src/class-rest-product-data.php
@@ -62,15 +62,15 @@ class REST_Product_Data {
 
 		$capabilities = self::get_backup_capabilities();
 
+		if ( is_wp_error( $response ) || empty( $response['body'] ) || 200 !== $response_code ) {
+			return new WP_Error( 'site_products_data_fetch_failed', 'Site products data fetch failed', array( 'status' => $response_code ? $response_code : 400 ) );
+		}
+
 		// If site has backups and the realtime backup capability, add latest undo event
 		if (
 			in_array( 'backup-realtime', $capabilities->data['capabilities'], true )
 		) {
 			$body->backups->last_undoable_event = self::get_site_backup_undo_event();
-		}
-
-		if ( is_wp_error( $response ) || empty( $response['body'] ) || 200 !== $response_code ) {
-			return new WP_Error( 'site_products_data_fetch_failed', 'Site products data fetch failed', array( 'status' => $response_code ? $response_code : 400 ) );
 		}
 
 		return rest_ensure_response( $body, 200 );
@@ -112,24 +112,13 @@ class REST_Product_Data {
 	 * This will fetch the last rewindable event from the Activity Log and
 	 * the last rewind_id prior to that.
 	 *
-	 * @param int $page - numbered page of activity logs to fetch.
-	 *
 	 * @return array|WP_Error|null
 	 */
-	public static function get_site_backup_undo_event( $page = null ) {
-		if ( ! $page ) {
-			$page = 1;
-		}
-
-		// Cap out at checking 10 pages of activity log
-		if ( $page >= 10 ) {
-			return null;
-		}
-
+	public static function get_site_backup_undo_event() {
 		$blog_id = \Jetpack_Options::get_option( 'id' );
 
 		$response = Client::wpcom_json_api_request_as_user(
-			'/sites/' . $blog_id . '/activity?force=wpcom&page=' . $page,
+			'/sites/' . $blog_id . '/activity/rewindable?force=wpcom',
 			'v2',
 			array(),
 			null,
@@ -143,11 +132,6 @@ class REST_Product_Data {
 		$body = json_decode( $response['body'], true );
 
 		if ( ! isset( $body['current'] ) ) {
-			return null;
-		}
-
-		// If page has no item, we have reached the end of the activity log.
-		if ( ! isset( $body['current']['orderedItems'] ) || count( $body['current']['orderedItems'] ) === 0 ) {
 			return null;
 		}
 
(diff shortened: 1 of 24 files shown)


projects/packages/my-jetpack/src/class-rest-product-data.php
```diff
@@ -64,7 +64,6 @@ public static function get_all_product_data() {
 
 		// If site has backups and the realtime backup capability, add latest undo event
```
jboland88, line 69: Unrelated to this change, should the error check for the API response come before this additional populating of the response body?

jboland88 (dismissed):
I had one question unrelated to this changeset, this PR itself looks good.

projects/packages/my-jetpack/src/class-rest-product-data.php
```diff
@@ -64,7 +64,6 @@ public static function get_all_product_data() {
...
 		if (
-			$body->backups->last_finished_backup_time &&
 			in_array( 'backup-realtime', $capabilities->data['capabilities'], true )
 		) {
 			$body->backups->last_undoable_event = self::get_site_backup_undo_event();
```
jboland88, line 73: Another thought here, if we know the site has a backup capability, do we need to call the first API endpoint to `jetpack-prodcut-data`? Could we just look up the last rewindable event?

jboland88 (commented):
I am late to the party, but I think we may want to revisit some of the API layer here. 

The recursive call to `get_site_backup_undo_event` seems like it could take a while to run. We should be able to collect rewindable events directly from the `activity/rewindable` endpoint, and/ or, we could also increase the number of events in the return body, maybe get the first 100 items so we don't need to send repeat API requests.

Also, would we be able to allow the backup card to initiate the query for the latest rewindable event? Since we are moving away from a centralized data approach, it would be great to have each card be responsible for collecting its own data going forward.

projects/packages/my-jetpack/src/class-rest-product-data.php
CodeyGuyDylan: I had thought about this, but we are still set up here for the "products-data" endpoint to be the main API call for all the cards, not just backup. I know we want to move away from that, but I didn't know how much refactoring we wanted to do right now 🤔 

I think if we were to move that conditional and not call the API if it is true, we'd want to go with a larger refactor right now because the way things are set up right now, this would prevent any other product data from being queried. I know we aren't pulling any other product data right now but it could be confusing for the next person if we leave things as they are since it looks like that is what it is for. I hope I am explaining my thoughts clearly enough

Do you think this should be a larger refactor to move away from the product data API endpoint right now?

projects/packages/my-jetpack/src/class-rest-product-data.php
CodeyGuyDylan: That's a good call, I think we could move that above

CodeyGuyDylan:
> I am late to the party, but I think we may want to revisit some of the API layer here.
> 
> The recursive call to `get_site_backup_undo_event` seems like it could take a while to run. We should be able to collect rewindable events directly from the `activity/rewindable` endpoint, and/ or, we could also increase the number of events in the return body, maybe get the first 100 items so we don't need to send repeat API requests.

I was unaware of the `activity/rewindable` endpoint, I will definitely look into calling this endpoint instead. I had copied this function from the backup plugin so I had assumed this was the best way to go

> Also, would we be able to allow the backup card to initiate the query for the latest rewindable event? Since we are moving away from a centralized data approach, it would be great to have each card be responsible for collecting its own data going forward.

Speaking to my earlier thoughts about the refactor, I think this might be involved in a larger refactor to move away from the product-data endpoint. Do you think we should handle this now?

projects/packages/my-jetpack/src/class-rest-product-data.php
jboland88: To add some notes from our conversation:

I think we can leave this as-is in the current diff, but we will revisit what would be involved in refactoring this and what would make sense to tackle before the next release

elliottprogrammer:
Tested per the instructions.

- Without the Backup plugin being activated, I was able to see the unpurchased state of the card with a summary of what can be backed up on the site:
Screenshot on 2023-11-15 at 13-25-40

- After purchasing a VaultPress backup subscription, waiting for a backup, then adding some posts, images, comments, etc. Then waited a couple hours.. (I see 2 full backups have been completed). Then going back to My Jetpack, I do see the purchased state of the card, as the testing instructions describe.
However, I do not get any UNDO button or a description of the last event/action that took place (as the testing instructions image implies).  Is this expected? Ok? (Here's what I get: See image:)
Markup 2023-11-15 at 16 06 33

~~I'm running this on a local environment using Docker so therefore I don't think there is any way to enter backup credentials when using a local docker dev environment. And since there are not credentials, restores cannot executed without credentials, so maybe that's why it does not show any rewindable events or the Undo button? Because technically we cannot rewind without credentials anyway? 🤷‍♂️~~
Anyway I'm not completely sure if I'm getting the desired result after testing.. Any thoughts @CodeyGuyDylan?

**EDIT:** After a brief Slack discussion with @CodeyGuyDylan, he informed me that the site's last activity/action along with the "Undo" button in the Backup card should become visible when the plugin can detect/read any "live" _real-time_ backup events. However  _real-time_ backup events aren't triggered when using the STORE_SANDBOX, which _I was_ using STORE_SANDBOX, so that's why the last backup event and the Undo button were not showing up for me during my testing.

So now I've tested after purchasing a Backup subscription while **Not** using STORE_SANDBOX, and as a result the Backup card displays the last site activity/action detected and the "Undo" button properly, as expected. (See image)
Markup 2023-11-15 at 17 52 26

Works good, as described. The code appears to be all good.
LGTM! 👍

elliottprogrammer (dismissed):

elliottprogrammer (approved):

PR #23716, opened 2022-03-30
class.jetpack.php: Move useful methods to Status package for other plugins to use. 

dereksmart:
props @gravityrail 

Breaks out the refactors done in class.jetpack.php in #23587, as it's helpful/required for standalone plugins being built off of existing Jetpack monolith features. 

The following classes have been introduced to the status package:  
- `Automattic\Jetpack\CookieState` 
- `Automattic\Jetpack\Errors`
- `Automattic\Jetpack\Files`
- `Automattic\Jetpack\Modules`
- `Automattic\Jetpack\Paths`

Which means the following Jetpack methods no longer rely on Jetpack-only code, and neither will other plugins: 
- `Jetpack::update_active_modules()`
- `Jetpack::get_available_modules()`
- `Jetpack::get_module_slug()`
- `Jetpack::get_module_path()`
- `Jetpack::get_module()`
- `Jetpack::get_active_modules()`
- `Jetpack::is_module_active()`
- `Jetpack::is_module()`
- `Jetpack::catch_errors()`
- `Jetpack::activate_module()`
- `Jetpack::deactivate_module()`
- `Jetpack::admin_url()`
- `Jetpack::state()`
- `Jetpack::should_set_cookie()`
- `Jetpack::is_onboarding()`
- `Jetpack::glob_php()`

#### Changes proposed in this Pull Request:

- Unlocks lot of useful-to-other-plugin code from being held hostage in class.jetpack.php. 
- Refactors class.jetpack.php to use the code.
- Refactors the Search plugin to use the package code.

#### Jetpack product discussion

N/A

#### Does this pull request change what data or activity we track or use?

No.

#### Testing instructions:

- Interact with Jetapck modules in every way you can think of, and from every place you can think of.

**I would love to get this merged as early in the release cycle as possible in order to get maximum passive testing exposure.**

PR #33999, opened 2023-11-07
lazy-images: Hack to fix tests

anomiex:
## Proposed changes:

Looks like #23716 back April 2022 in made packages/status depend on the `Jetpack_Options` class without adding an actual dependency. And nothing caught it because any tests hitting those code paths happened to provide the dep themselves, until now that WorDBless brings in WP 6.4 which makes the lazy-images tests start trying to call the "disable" function that was added as part of the deprecation process for that package.

Unfortunately `Jetpack_Options` was merged into packages/connection not long after that happened, and Connection depends on Status so we can't make Status depend on Connection.

While we sort that out, let's hack the lazy-image tests to not try to disable the module like they didn't before so the failing tests don't block everything else.

### Other information:

## Testing instructions:

* Is CI happy?


sdixon194 (approved):
Approving for now so we can unblock failing tests.

PR #34004, opened 2023-11-08
VS Code: set PHP version level to 7.2

kraftbj:
In setting up VS Code, I updated the PHP version to 7.2, which edits the settings.json for the repo. Should we set this as part of the default settings file sent over?

## Proposed changes:

* Set VSC to PHP 7.2

### Other information:

## Jetpack product discussion

n/a

No.

## Testing instructions:

* Open in VS Code and see if the bottom-right indicates PHP 5.6 when opening a PHP file.


kraftbj:
Updated to PHP 7.0.

.vscode/settings.json
```diff
@@ -34,5 +34,6 @@
 		"**/jetpack_vendor/**",
 		"**/vendor/**",
 	],
-	"prettier.prettierPath": "tools/js-tools/node_modules/prettier/index.cjs"
-}
\ No newline at end of file
+	"prettier.prettierPath": "tools/js-tools/node_modules/prettier/index.cjs",
+	"php.version": "7.0"
```
kraftbj:
```suggestion
	"php.version": "7.2"
```

jeherve (commented):
I'm not exactly sure what this does, and how it's used. Given that we rely on PHP 8.2 for other tooling in the monorepo, maybe that's what we should use here too?

kraftbj:
@jeherve The intent with this is for the PHP version the code supports; not dev tooling.

jeherve (approved):

PR #34006, opened 2023-11-08
Consolidate excluded files VS Code settings

ice9js:
## Proposed changes:

This adds most of the common excludes from our VSC settings into `files.exclude`. Making this global should help as some plugins inherit from that, and of course it will help with VSC directly.

That said, not all plugins do in fact inherit it, so it's still necessary/good practice to define it for every plugin.

### Other information:

No.

## Testing instructions:

- Open the monorepo from VSC.
- You shouldn't observe any obvious slowdowns and the same files should still be ignored for both Intelephense and PHPCodeSniffer.


sdixon194 (approved):
Seems okay in my testing 👍

kraftbj (approved):

tbradsha:
An alternative to consider would be including a `settings.json.default`.

I only noticed this change today when looking for the `node_modules` folder in VS Code. Noting that I'd much rather avoid forcing IDE settings; particularly when it deals with hiding files. It's rather frustrating that I now can't have my own workspace settings.

manzoorwanijk:
I agree with not forcing this and that, too, when the configuration is not complete.

We could do it in the same way as we did in Calpyso, making it opt-in https://github.com/Automattic/wp-calypso/pull/67798

PR #34012, opened 2023-11-08
Update gravatar profile shortcode

unDemian:
Fixes p3btAN-2wb-p2 (GHE Gravatar 106081 issue)

## Proposed changes:

* Limit usage of gravatar profile shortcode to users in current blog when used with `who=<user_id>` attribute.
* Retrieving Gravatar user profile has been update to request the data from Gravatar similar to the [Widget](https://github.com/Automattic/jetpack/blob/455da7c1400b07696f566513da104fc5006320a0/projects/plugins/jetpack/modules/widgets/gravatar-profile.php#L443)
* Gravatar url was based on username but that will not be correct for non wpcom scenarios, it will now be based on the hashed email address

### Other information:

No

## Testing instructions:

Both on Dotcom or Dotorg simple/multisite
* Make sure regular `[gravatar_profile who="email"]` works as expected
* `[gravatar_profile who="user-id"]` works only for users of that blog

### DIFF
--- a/projects/plugins/jetpack/modules/shortcodes/gravatar.php
+++ b/projects/plugins/jetpack/modules/shortcodes/gravatar.php
@@ -79,7 +79,12 @@ function jetpack_gravatar_profile_shortcode( $atts ) {
 
 	// Can specify username, user ID, or email address.
 	if ( is_numeric( $atts['who'] ) ) {
-		$user = get_user_by( 'id', (int) $atts['who'] );
+		$user_id = (int) $atts['who'];
+		if ( ! is_user_member_of_blog( $user_id, get_current_blog_id() ) ) {
+			// Bail if the user id is not a member of this site.
+			return false;
+		}
+		$user = get_user_by( 'id', $user_id );
 	} elseif ( is_email( $atts['who'] ) ) {
 		$user = get_user_by( 'email', sanitize_email( $atts['who'] ) );
 	} elseif ( is_string( $atts['who'] ) ) {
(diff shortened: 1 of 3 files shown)


projects/plugins/jetpack/modules/shortcodes/gravatar.php
```diff
@@ -93,16 +98,60 @@ function jetpack_gravatar_profile_shortcode( $atts ) {
...
+	if ( empty( $profile ) ) {
+		$profile_url = sprintf(
+			'https://secure.gravatar.com/%s.json',
+			$hashed_email
+		);
+
+		$expire        = 300;
+		$response      = wp_remote_get(
+			esc_url_raw( $profile_url ),
+			array( 'User-Agent' => 'WordPress.com Gravatar Profile Widget' )
```
jeherve: Since this is also in use in the Jetpack plugin, maybe we could say "Jetpack" instead of "WordPress.com", for folks examining their logs and trying to understand why a self-hosted site would make "WordPress.com" requests to their site?

projects/plugins/jetpack/modules/shortcodes/gravatar.php
```diff
@@ -93,16 +98,60 @@ function jetpack_gravatar_profile_shortcode( $atts ) {
...
+				$profile = $profile['entry'][0];
+			} else {
+				// Something strange happened.  Cache for 5 minutes.
+				$profile = array();
+			}
+		} else {
+			// Cache for 15 minutes.
+			$expire  = 900;
+			$profile = array();
+		}
```
jeherve: I wonder if this could be simplified so whenever we have an error, we cache for 5 minutes? Something like this?

```diff
diff --git a/projects/plugins/jetpack/modules/shortcodes/gravatar.php b/projects/plugins/jetpack/modules/shortcodes/gravatar.php
index b56c90e396..a3949e0885 100644
--- a/projects/plugins/jetpack/modules/shortcodes/gravatar.php
+++ b/projects/plugins/jetpack/modules/shortcodes/gravatar.php
@@ -108,34 +108,32 @@ function jetpack_gravatar_profile_shortcode( $atts ) {
 			$hashed_email
 		);
 
-		$expire        = 300;
 		$response      = wp_remote_get(
 			esc_url_raw( $profile_url ),
 			array( 'User-Agent' => 'WordPress.com Gravatar Profile Widget' )
 		);
 		$response_code = wp_remote_retrieve_response_code( $response );
+		$expire        = 300; // Cache any errors for 5 minutes.
+		$profile       = array();
 
 		if ( $response_code === 200 ) {
-			$profile = wp_remote_retrieve_body( $response );
-			$profile = json_decode( $profile, true );
-
-			if ( is_array( $profile ) && ! empty( $profile['entry'] ) && is_array( $profile['entry'] ) ) {
-				// Cache for 15 minutes.
+			$profile = json_decode( wp_remote_retrieve_body( $response ), true );
+
+			if (
+				is_array( $profile )
+				&& ! empty( $profile['entry'] )
+				&& is_array( $profile['entry'] )
+			) {
+				// Sucessfully fetched the profile. Cache for 15 minutes.
 				$expire  = 900;
 				$profile = $profile['entry'][0];
-			} else {
-				// Something strange happened.  Cache for 5 minutes.
-				$profile = array();
 			}
-		} else {
-			// Cache for 15 minutes.
-			$expire  = 900;
-			$profile = array();
 		}
 
 		set_transient( $cache_key, $profile, $expire );
 	}
 
+	// Fetching the profile returned an error. Bail.
 	if ( empty( $profile ) ) {
 		return false;
 	}
```

projects/plugins/jetpack/modules/shortcodes/gravatar.php
```diff
@@ -93,16 +98,60 @@ function jetpack_gravatar_profile_shortcode( $atts ) {
 		return false;
 	}
 
+	$hashed_email = hash( 'sha256', strtolower( trim( $user->user_email ) ) );
+	$cache_key    = 'grofile-' . $hashed_email;
```
jeherve: Do you think we could use a `jetpack_` prefix transient name here, so folks looking at transients on their site immediately know where it is set? `grofile` may be a bit obscure for folks not familiar with the shortcode.

projects/plugins/jetpack/modules/shortcodes/gravatar.php
```diff
@@ -93,16 +98,60 @@ function jetpack_gravatar_profile_shortcode( $atts ) {
...
+	$gravatar_url     = 'https://gravatar.com/' . $hashed_email;
+	$user_location    = ! empty( $profile['currentLocation'] ) ? $profile['currentLocation'] : '';
+	$display_name     = ! empty( $profile['displayName'] ) ? $profile['displayName'] : '';
+	$user_description = ! empty( $profile['aboutMe'] ) ? $profile['aboutMe'] : '';
 
 	if ( defined( 'IS_WPCOM' ) && IS_WPCOM ) {
-		$avatar_url    = wpcom_get_avatar_url( $user->ID, 96 );
-		$avatar_url    = $avatar_url[0];
-		$user_location = get_user_attribute( $user->ID, 'location' );
+		$gravatar_url = 'https://gravatar.com/' . $user->user_login;
```
jeherve, line 148: Is there a point in having 2 different scenarios for `$gravatar_url`? Given that the <gravatar.com/hash> redirects, maybe we could always use the <gravatar.com/hash> format, to simplify things?

projects/plugins/jetpack/tests/php/modules/shortcodes/test-class.gravatar.php
jeherve, line 1: Thanks for adding some tests! ❤️

jeherve (commented):
This tests well for me. I only have a few minor questions / remarks below.

projects/plugins/jetpack/modules/shortcodes/gravatar.php
unDemian: I left it to avoid a redirect if possible, but I can update it if feels cleaner.

projects/plugins/jetpack/modules/shortcodes/gravatar.php
unDemian: Yeap, makes sense.

projects/plugins/jetpack/modules/shortcodes/gravatar.php
unDemian: Sure thing, updated.

projects/plugins/jetpack/modules/shortcodes/gravatar.php
unDemian: Makes sense, updated.

unDemian:
Thanks for taking the time, feedback addressed.

projects/plugins/jetpack/modules/shortcodes/gravatar.php
jeherve: That makes sense. It doesn't really make things that much more complex, so I'm not insisting on this. 👍

jeherve (dismissed):

jeherve (approved):

PR #34016, opened 2023-11-08
e2e tests: remove e2e tests for blocks.

jeherve:
## Proposed changes:

e2e block tests have been failing in this repo for the past few months; we do not monitor or maintain them anymore, and other similar tests have been implemented in Calypso. Those can now be removed from the Jetpack repo.

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

* You should not see any block tests running here anymore.


kraftbj (approved):
Could you link to the Calypso tests? I assume https://github.com/Automattic/wp-calypso/tree/trunk/test/e2e/specs/blocks but are there others you had in mind?

anomiex:
While I like getting rid of broken unmaintained tests, I want to note p58i-fJV-p2#comment-60088 as a counterpoint.

kraftbj:
> While I like getting rid of broken unmaintained tests, I want to note p58i-fJV-p2#comment-60088 as a counterpoint.

Nod. I think this test would need to be modified/changed in any event. We can bring it back if/when we have the bandwidth to do it. I suppose we could also ask ourselves if it makes sense for unit tests to only consume individual bites, e2e tests only work within flows fully within our control, and we have separate integration tests that actually tie everything together.

In other words, counterpoint noted. I approved it so it could be merged, but also open to the team making a case for reworking this test now vs later.

jeherve:
> Could you link to the Calypso tests? I assume [Automattic/wp-calypso@`trunk`/test/e2e/specs/blocks](https://github.com/Automattic/wp-calypso/tree/trunk/test/e2e/specs/blocks?rgh-link-date=2023-11-08T18%3A27%3A17Z) but are there others you had in mind?

Those are the ones I had in mind indeed.

> I approved it so it could be merged, but also open to the team making a case for reworking this test now vs later.

I was not planning on reworking those tests at the moment. @anomiex What do you think? Should we merge this, or would you prefer I close this PR and leave things as they are for now, until someone has the bandwidth to refactor those tests? I'm fine either way (although I would personally be happy to get rid of the noise for now).

anomiex:
I'm ok with merging this to get rid of the noise. I just wanted to make sure the other comment was seen.

I had tried looking at the MailChimp test a day or two before you posted this PR. After I logged in as the test user and used the "Reconnect" button to reconnect to MailChimp, it worked for a short time before the connection broke again. If we decide to follow up instead, we could either try to figure out where in wpcom it keeps disconnecting or we could have the test use the credentials it already has to click that "Reconnect" button when it shows up.

PR #34019, opened 2023-11-08
AI Assistant: introduce action to increase requests counter

retrofox:
In order to make the app optimistic when computing the AI requests count, this PR introduces a new action to deal with it.
It isn't used yet.

Fixes #

## Proposed changes:

* AI Assistant: introduce action to increase requests counter

### Other information:

## Testing instructions:

* Go to the block editor
* Open the dev console
* Open the Redux tool
* Filter the `wordpress-com/plans` store
* Populate the store by getting the AI Assistant feature data

https://github.com/Automattic/jetpack/assets/77539/a40fa1f7-a2bb-4217-ad54-266683ea9915

### DIFF
--- a/projects/plugins/jetpack/extensions/store/wordpress-com/reducer.ts
+++ b/projects/plugins/jetpack/extensions/store/wordpress-com/reducer.ts
@@ -2,6 +2,7 @@
  * Types
  */
 import {
+	ACTION_INCREASE_AI_ASSISTANT_REQUESTS_COUNT,
 	ACTION_REQUEST_AI_ASSISTANT_FEATURE,
 	ACTION_SET_PLANS,
 	ACTION_STORE_AI_ASSISTANT_FEATURE,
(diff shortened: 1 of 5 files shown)


projects/plugins/jetpack/extensions/store/wordpress-com/reducer.ts
```diff
@@ -0,0 +1,93 @@
...
+							...state.features.aiAssistant._meta,
+							isRequesting: false,
+						},
+					},
+				},
+			};
+		}
+
+		case ACTION_INCREASE_AI_ASSISTANT_REQUESTS_COUNT: {
+			return {
```
CGastrell, line 79: Shouldn't this also increase the `features.aiAssistant.usagePeriod.requestsCount` number? Or maybe we have some useless redundancy there?

I guess `aiAssistant.requestsCount` was the original one, used for free and all plans, and now we seem to be moving into this specific usage object. For now, I'd say we can update both in the same reducer?

CGastrell: Either that or decide which one is it and get rid of the other, but I feel this should be done on a separate PR

retrofox: I think we should increase both. Nice feedback

CGastrell (approved):
LGTM! Works as described!

projects/plugins/jetpack/extensions/store/wordpress-com/reducer.ts
lhkowalski: Did we ended up not increment the `features.aiAssistant.usagePeriod.requestsCount` number? I think it will be important on usage meter for the tiers.

PR #34023, opened 2023-11-08
Subscribe Block: Fix regression in the editor

### LINKED ISSUE #34043: Subscribe Block not showing correctly in the editor
### Quick summary

I have two reports of the Subscribe block not displaying correctly in the editor compared to the live site; essentially, the button always shows as being on a new line when it should be on the side, which is reflected correctly in the live version of the site.

### Steps to reproduce

1. Add a `Subscribe` block to any page/post
2. Make sure that the settings are set so that the button *does not* appear on a new line

The button should appear to the side of the text entry box.

### What actually happened

The button in the editor shows as being on a new line regardless.

Editor view: 

Live view: 

### Impact

Most (> 50%)

### Available workarounds?

Yes, easy to implement

Simple, Atomic

### Logs or notes

Workaround for now is to let the user know that this is only affecting the editor and that on the live version of the site it will display correctly.

lezama:
Fixes https://github.com/Automattic/jetpack/issues/34043

Needs a follow up investigation to make sure changing display settings works well.

## Proposed changes:

After:
Screenshot 2023-11-08 at 16 10 42

Before:
Screenshot 2023-11-08 at 15 53 12

### Other information:

## Testing instructions:

* Add a subscribe block in the editor

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-subscribe-block-editor
+++ b/projects/plugins/jetpack/changelog/fix-subscribe-block-editor
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
+Subscribe Block: Fix display in the editor.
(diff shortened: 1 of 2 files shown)


edanzer (approved):
Tests as expected and fixes the alignment issue in the editor. 

For clarity from @lezama - this PR is adding an outer .is-not-subscriber div that is missing in the editor but present on the frontend. Once added, needed extra styles kick in.

PR #34028, opened 2023-11-08
AI Assistant: add next-tier prop

CGastrell:
Add `next-tier` prop to the AI assistant feature endpoint response

Fixes #

## Proposed changes:
This PR adds `next-tier` prop as part of the response from the AI assistant feature. It is provided with the same data shape as `current-tier` since they both come from the same collection. 

The PR also adds the prop to the frontend hooks and store, but it doesn't enforce the data shape yet.

### Other information:

No

## Testing instructions:

Checkout and go to the editor, use the console to force the AI assistant feature fetch, see that the returned value includes `nextTier` as `null`
image

Apply the patch on your sandbox, enable the filter by adding this line:
```php
add_filter( 'jetpack_ai_tier_plans_enabled', '__return_true' );
```

then look at the ai-assistant-feature response:
image

image

NOTE: without the filter enabled, you'll get default value `null` on `next-tier`

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-ai-feature/index.ts
+++ b/projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-ai-feature/index.ts
@@ -25,6 +25,7 @@ export const AI_Assistant_Initial_State = {
 	currentTier: {
 		value: aiAssistantFeature?.[ 'current-tier' ]?.value || 1,
 	},
+	nextTier: aiAssistantFeature?.[ 'next-tier' ] || null,
 };
 
 export default function useAiFeature() {
--- a/projects/plugins/jetpack/extensions/store/wordpress-com/reducer.ts
+++ b/projects/plugins/jetpack/extensions/store/wordpress-com/reducer.ts
@@ -32,6 +32,7 @@ const INITIAL_STATE: PlanStateProps = {
 			_meta: {
 				isRequesting: false,
 			},
+			nextTier: null,
 		},
 	},
 };
--- a/projects/plugins/jetpack/extensions/store/wordpress-com/types.ts
+++ b/projects/plugins/jetpack/extensions/store/wordpress-com/types.ts
@@ -31,11 +31,11 @@ export type AiFeatureProps = {
 		nextStart: string;
 		requestsCount: number;
 	};
+	nextTier: null;
 };
 
 export type AiFeatureStateProps = AiFeatureProps & {
 	_meta?: {
 		isRequesting: boolean;
 	};
-	nextTier: null;
 };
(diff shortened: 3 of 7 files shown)


projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-ai-feature/index.ts
```diff
@@ -25,6 +25,7 @@ export const AI_Assistant_Initial_State = {
 	currentTier: {
 		value: aiAssistantFeature?.[ 'current-tier' ]?.value || 1,
 	},
+	nextTier: aiAssistantFeature?.[ 'next-tier' ] || null,
```
retrofox, line 28: I think we should get rid of the `null` part.

```suggestion
	nextTier: aiAssistantFeature?.[ 'next-tier' ],
```

projects/plugins/jetpack/extensions/store/wordpress-com/reducer.ts
```diff
@@ -32,6 +32,7 @@ const INITIAL_STATE: PlanStateProps = {
 			_meta: {
 				isRequesting: false,
 			},
+			nextTier: null,
```
retrofox, line 35: Should define a default value here?

```suggestion
			nextTier: {
				slug: '',
				value: 20,
				limit: 20,
				redeableLimit: '',
			},
```

projects/plugins/jetpack/extensions/store/wordpress-com/types.ts
```diff
@@ -31,6 +31,7 @@ export type AiFeatureProps = {
 		nextStart: string;
 		requestsCount: number;
 	};
+	nextTier: null;
```
retrofox, line 34: I think we can define the type, too:

```suggestion
		nextTier: {
		slug: string;
		value: TierValueProp;
		limit: number;
		redeableLimit: string;
	};
```

projects/plugins/jetpack/extensions/types.ts
```diff
@@ -29,4 +29,5 @@ export type SiteAIAssistantFeatureEndpointResponseProps = {
 		limit: number;
 		value: TierValueProp;
 	} >;
+	'next-tier'?: null;
```
retrofox, line 32: Same here?

```suggestion
	'next-tier'?: {
		slug: string;
		value: TierValueProp;
		limit: number;
		'redeable-limit': string;
	};
```

retrofox (commented):
I wasn't able to get the either the `current-tier` or `next-tier` objects:

Screenshot 2023-11-08 at 18 11 57

retrofox (approved):

projects/plugins/jetpack/extensions/store/wordpress-com/reducer.ts
CGastrell: I was still wondering if we should. Null seems like a good option. But in any case, that was for a follow up as I'm going to be shaping both current and next tier props

PR #34036, opened 2023-11-09
Super Cache: start the next cycle of dev with 1.11.1-alpha

donnchawp:
Starting the next dev cycle.

## Proposed changes:
version numbers updated

No

## Testing instructions:
None required

### DIFF
--- a/projects/plugins/super-cache/changelog/init-release-cycle#2
+++ b/projects/plugins/super-cache/changelog/init-release-cycle#2
@@ -1,5 +1,5 @@
 Significance: patch
 Type: changed
-Comment: Init 1.12.0-alpha
+Comment: Init 1.11.1-alpha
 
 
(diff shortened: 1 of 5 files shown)


projects/plugins/super-cache/changelog/init-release-cycle#2
```diff
@@ -0,0 +1,5 @@
+Significance: patch
+Type: changed
+Comment: Init 1.12.0-alpha
```
dilirity: I always get confused with versioning. Isn't it suppose to be 1.11.1-alpha?

dilirity (dismissed):
:shipit:

projects/plugins/super-cache/changelog/init-release-cycle#2
haqadn: I think so too. Based on the autoloader suffix and package.json

donnchawp: You're right. I changed it elsewhere but I forgot that one.

donnchawp: Fixed it!

dilirity (approved):
:shipit:

PR #34039, opened 2023-11-09
Fix/subscribe modal showing to subscribers

lezama:
Fixes #

## Proposed changes:

* Use new Jetpack_Memberships::is_current_user_subscribed in order to decide if we show the modal or not. The new method handles subscribers for Jetpack and WPCOM better.

### Other information:

## Testing instructions:

* On a Jetpack site make sure the subscribe modal doesn't show up if you are a subscriber

### DIFF
--- a/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
+++ b/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
@@ -197,26 +197,10 @@ HTML;
 
 		// Don't show if user is subscribed to blog.
 		require_once __DIR__ . '/../views.php';
-		if ( $this->has_subscription_cookie() || Jetpack_Subscriptions_Widget::is_current_user_subscribed() ) {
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/modules/subscriptions/subscribe-modal/class-jetpack-subscribe-modal.php
```diff
@@ -197,27 +197,11 @@ public function should_user_see_modal() {
 
 		// Don't show if user is subscribed to blog.
 		require_once __DIR__ . '/../views.php';
-		if ( $this->has_subscription_cookie() || Jetpack_Subscriptions_Widget::is_current_user_subscribed() ) {
+		if ( Jetpack_Memberships::is_current_user_subscribed() ) {
```
simison, line 200: Do we need to ensure the class is loaded, similar to Subscription block paywall?

https://github.com/Automattic/jetpack/blob/8c3dabed53c440f1061cb2fcabc66daf10e384ca/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php#L792

simison (approved):
Confirmed this fixes issues.

PR #34041, opened 2023-11-09
Add/bloganuary add tags to bloganuary posts

roo2:
Adds bloganuary tags as specified in pe7F0s-1iI-p2. Builds on https://github.com/Automattic/jetpack/pull/33852

When a prompt is created in January it now gets a `bloganuary_id` such as `bloganuary-2024-01` for the first prompt of January.

This PR adds tags `bloganuary` and `bloganuary-2024-01` to newly created posts created in January.

## Proposed changes:

I had to extend the blogging-prompts v3 api slightly so that the single GET for a prompt can take a `force_year` parameter. I then passed in `force_year` to the current year when adding tags. Otherwise the dates would be the original date which would be in 2020. I think it would have been possible to design this in a more robust way, but at this point I just want to get it all connected and working.
Screen Shot 2023-11-09 at 11 46 10 pm

### Other information:

- [ x ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion
 pe7F0s-1iI-p2

No

## Testing instructions:
Apply this diff to my wpcom sandbox.
Open the editor for a prompt in January e.g. 
`/post/$site_slug?answer_prompt=1812`
Edit the post.
Expand the post settings "tags" section as shown in screenshot above.
You should see the blogging prompt tags and also two bloganuary tags.
This should only happen for blogging prompts in January

It can also be tested in calypso along with https://github.com/Automattic/wp-calypso/pull/83696

### DIFF
--- a/projects/plugins/jetpack/_inc/blogging-prompts.php
+++ b/projects/plugins/jetpack/_inc/blogging-prompts.php
@@ -48,7 +48,7 @@ function jetpack_setup_blogging_prompt_response( $post_id ) {
 	if ( $prompt ) {
 		update_post_meta( $post_id, '_jetpack_blogging_prompt_key', $prompt_id );
 		wp_add_post_tags( $post_id, array( 'dailyprompt', "dailyprompt-$prompt_id" ) );
-		if ( ! empty( $prompt['bloganuary_id'] ) ) {
+		if ( array_key_exists( 'bloganuary_id', $prompt ) ) {
 			wp_add_post_tags( $post_id, array( 'bloganuary', $prompt['bloganuary_id'] ) );
 		}
 	}
--- a/projects/plugins/jetpack/_inc/blogging-prompts.php
+++ b/projects/plugins/jetpack/_inc/blogging-prompts.php
@@ -48,6 +48,9 @@ function jetpack_setup_blogging_prompt_response( $post_id ) {
 	if ( $prompt ) {
 		update_post_meta( $post_id, '_jetpack_blogging_prompt_key', $prompt_id );
 		wp_add_post_tags( $post_id, array( 'dailyprompt', "dailyprompt-$prompt_id" ) );
+		if ( ! empty( $prompt['bloganuary_id'] ) ) {
+			wp_add_post_tags( $post_id, array( 'bloganuary', $prompt['bloganuary_id'] ) );
+		}
 	}
 }
 
@@ -118,6 +121,8 @@ function jetpack_get_blogging_prompt_by_id( $prompt_id ) {
 
 	$request = new WP_REST_Request( 'GET', $route );
 	$request->set_param( '_locale', $locale );
+	$request->set_param( 'force_year', gmdate( 'Y' ) );
+
 	$response = rest_do_request( $request );
 
 	if ( $response->is_error() || WP_Http::OK !== $response->get_status() ) {
--- a/projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v3-endpoint-blogging-prompts.php
+++ b/projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v3-endpoint-blogging-prompts.php
@@ -128,6 +128,10 @@ class WPCOM_REST_API_V3_Endpoint_Blogging_Prompts extends WP_REST_Posts_Controll
 			return $this->proxy_request_to_wpcom( $request, $request->get_param( 'id' ) );
 		}
 
+		if ( $request->get_param( 'force_year' ) ) {
+			$this->force_year = $request->get_param( 'force_year' );
+		}
+
 		switch_to_blog( self::TEMPLATE_BLOG_ID );
 		$item = parent::get_item( $request );
 		restore_current_blog();
--- a/projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v3-endpoint-blogging-prompts.php
+++ b/projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v3-endpoint-blogging-prompts.php
@@ -537,11 +537,11 @@ class WPCOM_REST_API_V3_Endpoint_Blogging_Prompts extends WP_REST_Posts_Controll
 		}
 
 		$response_status = wp_remote_retrieve_response_code( $response );
-		$response_body   = json_decode( wp_remote_retrieve_body( $response ) );
+		$response_body   = json_decode( wp_remote_retrieve_body( $response ), true );
 
 		if ( $response_status >= 400 ) {
-			$code    = isset( $response_body->code ) ? $response_body->code : 'unknown_error';
-			$message = isset( $response_body->message ) ? $response_body->message : __( 'An unknown error occurred.', 'jetpack' );
+			$code    = isset( $response_body['code'] ) ? $response_body['code'] : 'unknown_error';
+			$message = isset( $response_body['message'] ) ? $response_body['message'] : __( 'An unknown error occurred.', 'jetpack' );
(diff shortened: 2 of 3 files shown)


projects/plugins/jetpack/_inc/blogging-prompts.php
```diff
@@ -48,6 +48,9 @@ function jetpack_setup_blogging_prompt_response( $post_id ) {
 	if ( $prompt ) {
 		update_post_meta( $post_id, '_jetpack_blogging_prompt_key', $prompt_id );
 		wp_add_post_tags( $post_id, array( 'dailyprompt', "dailyprompt-$prompt_id" ) );
+		if ( ! empty( $prompt['bloganuary_id'] ) ) {
```
jeherve: This causes a fatal:

```
Fatal error: Uncaught Error: Cannot use object of type stdClass as array in wp-content/plugins/jetpack-dev/_inc/blogging-prompts.php:51
```

jeherve (changes requested):

projects/plugins/jetpack/_inc/blogging-prompts.php
roo2: hmmm for me it is saying that $prompt is an array, I logged `gettype($prompt)` and it returned `array`, and the check seemes to work, so I think there must be different cases perhaps? I tested uploading to my wpcom sandbox with `https://wordpress.com/post/bloggerforeru.wordpress.com?answer_prompt=1812` for a bloganuary prompt and `https://wordpress.com/post/bloggerforeru.wordpress.com?answer_prompt=1850` for a normal prompt. Do you remember how you were testing when you got the fatal? 🙏

I updated the code to use `array_key_exists` which will just return false if it is passed a stdClass. It worked (on my machine & test site) but I'd be interested if the code adds the tag for you or always returns false?

Thank you for testing this!

.

projects/plugins/jetpack/_inc/blogging-prompts.php
```diff
@@ -48,6 +48,9 @@ function jetpack_setup_blogging_prompt_response( $post_id ) {
 	if ( $prompt ) {
 		update_post_meta( $post_id, '_jetpack_blogging_prompt_key', $prompt_id );
 		wp_add_post_tags( $post_id, array( 'dailyprompt', "dailyprompt-$prompt_id" ) );
+		if ( array_key_exists( 'bloganuary_id', $prompt ) ) {
```
eoigal, line 51: `jetpack_get_blogging_prompt_by_id` returns a stdClass object from what I can tell. I think this should be something like;

```suggestion
		if ( property_exists( $prompt, 'bloganuary_id' ) ) {
```

projects/plugins/jetpack/_inc/blogging-prompts.php
```diff
@@ -48,6 +48,9 @@ function jetpack_setup_blogging_prompt_response( $post_id ) {
...
+			wp_add_post_tags( $post_id, array( 'bloganuary', $prompt['bloganuary_id'] ) );
```
eoigal, line 52:
```suggestion
			wp_add_post_tags( $post_id, array( 'bloganuary', $prompt->bloganuary_id ) );
```

eoigal (commented):
This worked for me when testing. There was some confusion for me as the function `jetpack_get_blogging_prompt_by_id` going by its phpdoc function description but it looks like its `data` can be mixed and in this case, we are returning an array. 

Should we update the function to be clearer on that?

projects/plugins/jetpack/_inc/blogging-prompts.php
jeherve: > Do you remember how you were testing when you got the fatal?

I'm testing on a self-hosted site. In all my tests, I get an object back:

image

image

This fixes things on self-hosted, but from what you're saying it seems you're getting a different response on WordPress.com simple. I'll test there as well.
Edit: It is indeed an array on WordPress.com Simple! That's probably something we'd need to fix, ideally we would have the same response in both environments.

```diff
diff --git a/projects/plugins/jetpack/_inc/blogging-prompts.php b/projects/plugins/jetpack/_inc/blogging-prompts.php
index f00541e6ed..88ec1604f0 100644
--- a/projects/plugins/jetpack/_inc/blogging-prompts.php
+++ b/projects/plugins/jetpack/_inc/blogging-prompts.php
@@ -48,8 +48,8 @@ function jetpack_setup_blogging_prompt_response( $post_id ) {
 	if ( $prompt ) {
 		update_post_meta( $post_id, '_jetpack_blogging_prompt_key', $prompt_id );
 		wp_add_post_tags( $post_id, array( 'dailyprompt', "dailyprompt-$prompt_id" ) );
-		if ( array_key_exists( 'bloganuary_id', $prompt ) ) {
-			wp_add_post_tags( $post_id, array( 'bloganuary', $prompt['bloganuary_id'] ) );
+		if ( property_exists( $prompt, 'bloganuary_id' ) ) {
+			wp_add_post_tags( $post_id, array( 'bloganuary', $prompt->bloganuary_id ) );
 		}
 	}
 }

```

roo2: hmmm yes, good to test on both and bad to have different behavior on both! 😂 I'll look into whether I can standardize the `jetpack_get_blogging_prompt_by_id` function

roo2: Ahh I think I have found it, in the implementation of proxy_request_to_wpcom, we parse the response with `json_decode`, which converts it to a stdClass object, I'll add a second parameter to turn it into an array and test that tomorrow, then we would be consistent. Thanks for helping me get set up with the jurassic.ninja site, it looks like as the code is now, I'm getting a fatal error from `property_exists`, but I'll update it tomorrow :). 

https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v3-endpoint-blogging-prompts.php#L536

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v3-endpoint-blogging-prompts.php
```diff
@@ -533,11 +537,11 @@ public function proxy_request_to_wpcom( $request, $path = '' ) {
 		}
 
 		$response_status = wp_remote_retrieve_response_code( $response );
-		$response_body   = json_decode( wp_remote_retrieve_body( $response ) );
+		$response_body   = json_decode( wp_remote_retrieve_body( $response ), true );
```
roo2, line 540: this was the fix so that an array type is returned on both wpcom and self hosted

roo2:
I made the change so that it now returns associative array objects for both wpcom and self hosted. I retested self hosted and it now adds the bloganuary tag as expected. It looks like the only other place we use blogging prompts in self-hosted is for the API call that supports the prompt block. But API calls just parse stdClass and associative arrays into the same kind of JSON object so it doesn't matter either way :).

I noticed that in the [proxy_request_to_wpcom_as_user function](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/trait-wpcom-rest-api-proxy-request-trait.php#L48) used by email previews and newsletters we are making the same mistake which would result in stdClass being returned in stead of array objects. It appears that all of those endpoints are only called externally from JavaScript, so it would not be an issue at this point. But I think I will create another PR after this to standardise that code just in-case it's expanded or copied from later.

roo2:
@jeherve I've added https://github.com/Automattic/jetpack/pull/34417 to make the same refactoring for other/future endpoints. There were only three others effected and it appears none of them are even called from PHP :)

jeherve (approved):
This tests well for me now. 🚢

PR #34042, opened 2023-11-09
Boost: Migrate benefits interstitial component to react

dilirity:
## Proposed changes:

* Migrate Benefits Interstitial to react.
* Rename `BenefitsInterstitial` to `Upgrade`, as it's easier to understand to non-native speakers (and it also matches the `#upgrade` route).

### Other information:

## Jetpack product discussion

pc9hqz-28M-p2

no

## Testing instructions:

* Make sure the upgrade page works as expected.
* Clicking the `Upgrade Jetpack Boost` should record an event and redirect to the upgrade page.


projects/plugins/boost/app/assets/src/js/react-components/pages/upgrade/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+	return (
+		<div id="jb-dashboard" className="jb-dashboard">
+			<div className={ classNames( 'jb-dashboard-header', styles.benefits__header ) }>
+				<div
+					className={ classNames(
+						'jb-container jb-container--fixed',
+						styles[ 'container--fixed' ]
+					) }
+				>
+					<div className={ styles.header__logo }>
```
pyronaur: I think we don't need BEM anymore here

projects/plugins/boost/app/assets/src/js/react-components/pages/upgrade/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+								</li>
+								<li>{ __( 'Dedicated email support', 'jetpack-boost' ) }</li>
+							</ul>
+						</div>
+
+						<div className="jb-card__cta px-2 my-4">
+							{ 'yearly' in pricing && (
+								<PricingCard
+									title={ __( 'Jetpack Boost', 'jetpack-boost' ) }
+									icon={ `${ Jetpack_Boost.site.assetPath }../static/images/forward.svg` }
```
pyronaur: `${path}../static/..` 🤔 

Maybe `${ Jetpack_Boost.site.staticPath }/...`

projects/plugins/boost/app/assets/src/js/react-components/pages/upgrade/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+									priceBefore={ pricing.yearly.priceBefore / 12 }
+									priceAfter={ pricing.yearly.priceAfter / 12 }
```
pyronaur: Look-up historical conversations with pricing - I think we have pricing monthly config somewhere.

IIRC this might not work, for example if monthly pricing is 12$ then * 12 it's 144$, but maybe the promo says yearly pricing is 120$ then `120 / 12 = 10$`. All that is to say - we shouldn't `/ 12 ` to get pricing.

pyronaur (changes requested):
Looks good, but needs a rebase and look at the pricing history

projects/plugins/boost/app/assets/src/js/react-components/pages/upgrade/index.tsx
dilirity: Added new property to `Jetpack_Boost` for static path and fixed the path.

projects/plugins/boost/app/assets/src/js/react-components/pages/upgrade/index.tsx
dilirity: It was like that before the migration - https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/boost/app/assets/src/js/pages/upgrade/Upgrade.svelte#L77-L78. We haven't seen issues there, so I'll leave it for now.

pyronaur (approved):
Looks good! 🚢

PR #34045, opened 2023-11-09
Boost: Migrate Getting Started page to react

dilirity:
## Proposed changes:

* Migrate the Getting Started page (#getting-started) to react.

### Other information:

## Jetpack product discussion

pc9hqz-28M-p2

no

## Testing instructions:

* Make sure the Getting Started page is working as expected.


projects/plugins/boost/app/assets/src/js/pages/getting-started/getting-started.module.scss
```diff
@@ -0,0 +1,9 @@
+.pricing-table {
+	isolation: isolate;
+
+	h2 {
+		font-size: 36px !important;
+		font-weight: 700 !important;
+		line-height: 40px !important;
+	}
```
pyronaur: Why is this so `!important`

projects/plugins/boost/app/assets/src/js/pages/getting-started/getting-started.tsx
```diff
@@ -0,0 +1,87 @@
...
+	domain,
+} ) => {
+	const [ selectedPlan, setSelectedPlan ] = useState< 'free' | 'premium' | false >( false );
+	const [ snackbarMessage, setSnackbarMessage ] = useState< string >( '' );
+
+	if ( false !== selectedPlan ) {
+		initialize( selectedPlan );
+	}
+
+	async function initialize( plan: 'free' | 'premium' ) {
```
pyronaur: `initialize()` is called before it's defined

projects/plugins/boost/app/assets/src/js/pages/getting-started/getting-started.tsx
```diff
@@ -0,0 +1,87 @@
...
+			// Otherwise go to dashboard home.
+			// @todo - fix when react routing
+			// navigate( '/', { replace: true } );
+			navigate( '/' );
+		} catch ( e ) {
+			// Display the error in a snackbar message
+			setSnackbarMessage( e.message || 'Unknown error occurred during the plan selection.' );
+		} finally {
+			setSelectedPlan( false );
+		}
```
pyronaur: This is almost an infinite loop, or at least without this `finally` it would be. I think this probably needs to be wrapped in a react hook and the initialize function can be lifted out of the component.

pyronaur (dismissed):
I think this needs a bit of react hook magic

projects/plugins/boost/app/assets/src/js/pages/getting-started/getting-started.module.scss
dilirity: I guess I found a way without it - thanks 😅

projects/plugins/boost/app/assets/src/js/pages/getting-started/getting-started.tsx
dilirity: I'm not sure what moving the function outside the component would look like because of the state setter functions.

Also - how would it be an infinite loop without the `finally`? Because the setter would always get called?

dilirity: I attempted to resolve this in 36ec873f63149e8a2879a0ee5eff65515348a3b2.

projects/plugins/boost/app/assets/src/js/pages/getting-started/getting-started.tsx
dilirity: Should be resolved with 36ec873f63149e8a2879a0ee5eff65515348a3b2.

pyronaur (approved):
🎉  Looks good!

PR #34046, opened 2023-11-09
AI Assistant: Add tier data shape and types

CGastrell:
Add ts types for tiers

Fixes #

## Proposed changes:
This PR adds the definitions for tier types, including the limit prop and the optional readableLimit

### Other information:

No

## Testing instructions:
The change should introduce no errors. Keep the console open and use any flow to trigger AI usage:
- open usage panel
- use AI assistant block
- use Form block and ask assistant to build
- use suggestions from AI assistant on paragraphs

### DIFF
--- a/projects/plugins/jetpack/extensions/store/wordpress-com/types.ts
+++ b/projects/plugins/jetpack/extensions/store/wordpress-com/types.ts
@@ -8,15 +8,15 @@ export type UpgradeTypeProp = 'vip' | 'default';
 
 export type TierUnlimitedProps = {
 	slug: 'ai-assistant-tier-unlimited';
-	limit: 9223372036854776000n;
+	limit: 922337203685477600;
 	value: 1;
-	readableLimit: 'Unlimited';
+	readableLimit: string;
 };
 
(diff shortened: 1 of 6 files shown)


projects/plugins/jetpack/extensions/store/wordpress-com/types.ts
```diff
@@ -59,20 +74,13 @@ export type AiFeatureProps = {
 	errorMessage: string;
 	errorCode: string;
 	upgradeType: UpgradeTypeProp;
-	currentTier: {
-		value: TierValueProp;
-	};
+	currentTier: TierFreeProps | Tier100Props | Tier200Props | Tier500Props | TierUnlimitedProps;
```
retrofox: Maybe we could create a `TierProp` type, and reuse it:

CGastrell: Yeah, I think I did that in the reverse order (enum and then realized a TierProp would come in handy)

retrofox (approved):
Thanks for addressing these issues. LGTM!

PR #34025, opened 2023-11-08
AI Client: ensure dispatching unclear error prompt once per request/response

retrofox:
This PR fixes an issue that happens when detecting the `Unclear prompt error`.
The app tries to catch the error searching by the `JETPACK_AI_ERROR` string in the streaming connection.
Sometimes, and depending on the streaming data, the app dispatches the error twice (or maybe more?) because:

* The method is called twice
* The fullMessage prop contains the `JETPACK_AI_ERROR`

To fix this issue, I've introduced a flag in order to avoid more than one dispatch per request

Fixes #

## Proposed changes:

* AI Client: ensure dispatching unclear error prompt once per request/response

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* Go to the block editor
* Use the debug() lines
* Create an AI Assistant block instance
* Introduce an unclear prompt
* Request
* Confirm that, always, the app dispatched the event once per request, looking at the debug line:

Screenshot 2023-11-08 at 14 37 09

PR #34051, opened 2023-11-09
AI Client: do not dispatch `done` event when JETPACK_AI_ERROR

retrofox:
Follow-up of https://github.com/Automattic/jetpack/pull/34025

In the same way that a double `error` event was dispatched, fixed in https://github.com/Automattic/jetpack/pull/34025, this PR fixes the issue that happens sometimes when the prompt is unclear but in the `done` event. Visually, the editor shows the undesired `JETPACK_AI_ERROR`:

Screenshot 2023-11-09 at 12 25 36

Fixes #

## Proposed changes:

* AI Client: do not dispatch `done` event when JETPACK_AI_ERROR

### Other information:

## Testing instructions:

* Go to the block editor
* Set an unclear prompt message
* Send the request

**BEFORE (trunk)**
* Confirm that, sometimes, the app renders the `JETPACK_AI_ERROR`:
image

**AFTER (this PR)**
* Confirm that the app never renders the error message

### DIFF
--- a/projects/js-packages/ai-client/changelog/update-ai-client-prevent-unclear-error-message-on-done
+++ b/projects/js-packages/ai-client/changelog/update-ai-client-prevent-unclear-error-message-on-done
@@ -1,0 +1,4 @@
+Significance: patch
+Type: changed
+
+AI Client: do not dispatch `done` event when JETPACK_AI_ERROR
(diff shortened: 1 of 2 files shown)


dhasilva (approved):
Can't guarantee it _never_ happens, but it did not happen on the times I tested and the code makes sense.

PR #34053, opened 2023-11-09
Fix some wpcom rector complaints

anomiex:
## Proposed changes:

In the daily deploy in D128439-code, it complained about calling static methods with `$this->` in device-detection, and a place where `end()` needs to be passed a variable to avoid a PHP notice.

### Other information:

## Testing instructions:

* Stuff still works?


zinigor (approved):
Sanity check passed, looks good to me, but haven't tested.

PR #34054, opened 2023-11-09
WIP - Use register_settings for Social options

pablinos:
## Proposed changes:
This is on experiment to see if we can use register_setting in order to utilise the core functions and rest endpoints.

### Other information:

No

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/projects/packages/publicize/src/class-publicize-setup.php
+++ b/projects/packages/publicize/src/class-publicize-setup.php
@@ -43,7 +43,7 @@ class Publicize_Setup {
 		add_action( 'rest_api_init', array( new REST_Controller(), 'register_rest_routes' ) );
 		add_action( 'current_screen', array( static::class, 'init_sharing_limits' ) );
 		// add_action( 'rest_api_init', array( new Auto_Conversion\REST_Settings_Controller(), 'register_routes' ) );
-		add_action( 'rest_api_init', array( new Jetpack_Social_Settings\REST_Settings_Controller(), 'register_routes' ) );
+		add_action( 'rest_api_init', array( new Jetpack_Social_Settings\Settings(), 'register_settings' ) );
 
 		( new Social_Image_Generator\Setup() )->init();
 	}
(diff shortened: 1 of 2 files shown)


pablinos:
image


Using this we can call the core endpoint to get the settings in one go. The code to update the settings is untested, but in principle it should work.

I don't understand the current format of the SIG settings. There's a defaults array, but all the settings are the defaults, so that doesn't seem to make sense to me. I've flattened it for now, but that would mean we would need to migrate the SIG settings too

Similarly, I think we should simplify the auto conversion settings. I don't know when we're going to do videos, and we're future proofing in a way that just adds maintenance. I think just having a boolean makes things simpler.

gmjuhasz:
I tested this, but it gives me unknown route:
CleanShot 2023-11-10 at 13 30 05 png

On the https://developer.wordpress.com/docs/api/console/ it was working with `wp/w2/sites/<siteid>/settings`.

Also I know it's WIP, but so far I see some errors if I try to load my site because of some format issues with the initial state. I'm trying to find a place for this action where it could be registered on simple sites too, then we could just use the same w2 API to get and update the settings for WPCOM site

pablinos:
> I tested this, but it gives me unknown route:

I think that's because the route is wrong. It should be v2 not w2

> On the https://developer.wordpress.com/docs/api/console/ it was working with wp/w2/sites/<siteid>/settings.

Yes, on simple sites the route changes to having the /sites/<site id>/ part of the path. We cope with that by passing the URL in the initial state for the other endpoints. We might be able to make it more generic by passing an API base URL or something similar.

> Also I know it's WIP, but so far I see some errors if I try to load my site because of some format issues with the initial state.

Yeah it's definitely not in a working state, I was trying to prove concept with using register_setting and the core endpoints, which I think is possible. My main concern is register the settings based on the feature being available, and whether the feature checks might end up being too expensive to be run during `rest_api_init`. If it hits the database, I think that could be a problem.

pablinos:
@gmjuhasz I think this has been superseded by work we've done since. Is that right?

gmjuhasz:
> @gmjuhasz I think this has been superseded by work we've done since. Is that right?

Yes, looking at Pulicize setup, we already do this.

fbhepr%2Skers%2Swrgcnpx%2Scebwrpgf%2Scnpxntrf%2Schoyvpvmr%2Sfep%2Spynff%2Qchoyvpvmr%2Qfrghc.cuc%3Se%3Q24s18726-og

PR #34056, opened 2023-11-09
Jetpack: Fix open ai libs require paths

CGastrell:
Fix paths used to require libs and helpers

## Proposed changes:
This PR addresses a bug that never triggered. The paths used to require the classes and libs is wrong, but the condition under which those files should have been required was never met.

### Other information:

## Jetpack product discussion
p1699553597616979-slack-C054LN8RNVA

No

## Testing instructions:
There is no testing process that can cover the case, we'll just use the sandbox to try and mimic what was done before and what is done now to require the libs:

```bash
wpsh

# previous to this patch, this is the line that tried to load one of the classes
require_once WP_PLUGIN_DIR . '/openai/openai-request-count.php';
# see that it should trigger an error

# new path
require_once WP_CONTENT_DIR . '/lib/openai/openai-request-count.php';
# should output nothing

# try to run a method from the newly required file
print_r( OpenAI_Request_Count::get_count( 211199336 ) ); # use whatever blog ID here
# should print the requests count for the given blog ID
```

### DIFF
--- a/projects/plugins/jetpack/_inc/lib/class-jetpack-ai-helper.php
+++ b/projects/plugins/jetpack/_inc/lib/class-jetpack-ai-helper.php
@@ -348,8 +348,8 @@ class Jetpack_AI_Helper {
 			}
 
 			if ( ! class_exists( 'OpenAI_Limit_Usage' ) ) {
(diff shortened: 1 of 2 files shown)


retrofox (approved):
LGTM!

PR #34058, opened 2023-11-09
Changelogger: Add New Prompt for User Change

sdixon194:
## Proposed changes:

Per some recent discussions about how to properly format changelogs, this PR adds an additional prompt to changelogger to check if the change is user facing. If it's not, it automatically adds the `other` type so it won't appear in the readme.txt during release.

### Other information:

## Jetpack product discussion

pf4qpu-9d-p2

no
## Testing instructions:

You can use `jetpack changelog add <plugin>` to test and make sure the new prompt appears. If you answer Yes, it should then ask you for the type of change after. If you answer no, it should skip it and automatically give it the `other` type.


projects/packages/changelogger/src/AddCommand.php
```diff
@@ -242,11 +242,24 @@ protected function execute( InputInterface $input, OutputInterface $output ) {
 					$type = strtolower( $type );
 				}
 				if ( $isInteractive ) {
-					$question = new ChoiceQuestion( 'Type of change.', $types, $type );
-					$type     = $this->getHelper( 'question' )->ask( $input, $output, $question );
-					if ( null === $type ) { // non-interactive.
-						$output->writeln( 'Got EOF when attempting to query user, aborting.', OutputInterface::VERBOSITY_VERBOSE ); // @codeCoverageIgnore
-						return 1;
+					$question     = new ChoiceQuestion(
+						'Is this a change users would want to know about?',
```
jeherve: What if we added "self-hosted" to kind of exclude WoA and simple? I guess that's very Automattic-specific though, so not really appropriate for this package.

```suggestion
						'Is this a change self-hosted users would want to know about?',
```

What about "site admins"? Would it be more appropriate?


```suggestion
						'Is this a change site admins would want to know about?',
```

sdixon194: I like site admins, I'll plug that in for now.

sdixon194:
Reworked things per the thread here p1699566741898429-slack-C05Q5HSS013

sdixon194:
Since we moved to enquirer, I'm closing this in favor of https://github.com/Automattic/jetpack/pull/34523

PR #34062, opened 2023-11-10
 Media: Prevent fatals in HEIC/HEIF Support

jeroenpf:
Fixes https://github.com/Automattic/dotcom-forge/issues/4095

It is possible that the filename is empty when triggering the `jetpack_wpcom_transparently_convert_heif_upload_to_jpg` function via the `wp_handle_upload_prefilter` and  `wp_handle_sideload_prefilter` hooks. This function will detect a HEIC/HEIF image and try to convert it into a jpg image. 

We have received fatal errors in cases where the filename is empty or does not exist:
`PHP Fatal error:  Uncaught ValueError: Path cannot be empty in /.../jetpack-mu-wpcom/src/features/media/heif-support.php:32`

It is likely caused by the interaction with other plugins and I was not able to reproduce the problem without adding a little plugin that deliberately empties the filename.

## Proposed changes:

* Ensure that we detect empty filenames or files that do not exist and exit early - this will prevent fatal errors
* Log errors and information to logstash if we are in a WPCOM context

### Other information:

## Testing instructions:

I am unsure as to how the errors happened and was not able to reproduce them. This test will focus on ensuring that the conversion still works properly and that we are logging information in the context of WPCOM

**Simple**

* Apply this patch to a simple site by following the instructions in the jetpack-mu-wpcom docs: PCYsg-Osp-p2
* Go to the media uploader and upload a HEIC/HEIF file
* Ensure that the file is uploaded successfully as a JPG
* In logstash, you should see an info entry for your upload in the log2logstash index. It should show you the filename and the magic bytes that appear at the beginning of the image file that indicate the type of HEIC/HEIF. You can use `feature:heif2jpg` to find the log entry.
* You can try uploading a corrupted HEIC image. Doing this should trigger an error that will be logged in Logstash. You can use `feature:heif2jpg` to find the log entry.

Download a corrupted HEIC file that I provided in the original issue.

**Atomic**

Logging is disabled on Atomic sites.

* Apply this patch to your Atomic site by following the instructions in the jetpack-mu-wpcom docs: PCYsg-Osp-p2
* Go to the media uploader and upload a HEIC/HEIF file
* Ensure that the file is uploaded successfully as a JPG

**Simulate empty file on Atomic**
Because the fatal error is caused by receiving an empty filename, we want to verify that this no longer triggers a fatal error. Add in `wp-content/plugins` a file name `empty-file-uploader.php`with the following code:

```
<?php
/**
 * Plugin Name: Empty Filename Uploader
 * Description: A plugin to simulate an empty filename during upload for testing purposes.
 * Version: 1.0
 * Author: Test
 */

// Hook into the filters for handling file uploads and side loads.
add_filter('wp_handle_upload_prefilter', 'empty_filename_prefilter', 1);
add_filter('wp_handle_sideload_prefilter', 'empty_filename_prefilter', 1);

// Function to modify the filename to be empty.
function empty_filename_prefilter($file) {
    $file['tmp_name'] = '';
    return $file;
}
```

Activate the plugin and try to upload a valid HEIC file. The easiest way to observe that the fatal error is no longer happening is by inspecting the network tab in the browser and looking for the `async-upload.php` request. This request should return a JSON error: `Specified file failed upload test.`. Doing the same without this PR applied should cause a fatal error ( an error page will be returned AND an error `PHP Fatal error:  Uncaught ValueError: Path cannot be empty` will be displayed in the site monitoring php log)

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/media/heif-support.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/media/heif-support.php
@@ -17,14 +17,12 @@ function jetpack_wpcom_maybe_convert_heif_to_jpg( $filename ) {
 	}
 
 	if ( empty( $filename ) ) {
-
 		jetpack_wpcom_maybe_log_heif_to_jpg(
 			array(
 				'message'  => 'file path is empty',
 				'severity' => 'error',
 			)
 		);
-
 		return false;
 	}
 
@@ -35,7 +33,6 @@ function jetpack_wpcom_maybe_convert_heif_to_jpg( $filename ) {
 				'severity' => 'error',
 			)
 		);
-
 		return false;
 	}
 
--- a/projects/packages/jetpack-mu-wpcom/src/features/media/heif-support.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/media/heif-support.php
@@ -15,6 +15,19 @@ function jetpack_wpcom_maybe_convert_heif_to_jpg( $filename ) {
 	if ( ! class_exists( 'Photon_OpenCV' ) ) {
 		return false;
 	}
+
+	if ( empty( $filename ) || ! file_exists( $filename ) ) {
+
+		jetpack_wpcom_maybe_log_heif_to_jpg(
+			array(
+				'message'  => sprintf( 'file does not exist: %s', $filename ),
+				'severity' => 'error',
+			)
+		);
+
+		return false;
+	}
+
 	$valid_magic_bytes = array(
 		'ftypheic',
 		'ftypheix',
@@ -34,6 +47,14 @@ function jetpack_wpcom_maybe_convert_heif_to_jpg( $filename ) {
 		return false;
 	}
 
+	// Log that we found a HEIF image.
+	jetpack_wpcom_maybe_log_heif_to_jpg(
+		array(
+			'message'  => sprintf( 'Found HEIF image with magic bytes: %s', $magic_bytes ),
+			'severity' => 'info',
+		)
+	);
+
 	$img = new Photon_OpenCV();
 	try {
 		$img->readimage( $filename );
@@ -41,6 +62,13 @@ function jetpack_wpcom_maybe_convert_heif_to_jpg( $filename ) {
 		// Bad detection or malformed image
 		/** This action is documented in modules/widgets/social-media-icons.php */
 		do_action( 'jetpack_bump_stats_extras', 'heif2jpg', 'failed-to-read' );
+		jetpack_wpcom_maybe_log_heif_to_jpg(
+			array(
+				'message'    => sprintf( 'failed to read image: %s', $e->getMessage() ),
+				'error_code' => $e->getCode(),
+				'severity'   => 'error',
+			)
+		);
 		return false;
 	}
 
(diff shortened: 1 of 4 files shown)


projects/packages/jetpack-mu-wpcom/src/features/media/heif-support.php
```diff
@@ -15,6 +15,19 @@ function jetpack_wpcom_maybe_convert_heif_to_jpg( $filename ) {
 	if ( ! class_exists( 'Photon_OpenCV' ) ) {
 		return false;
 	}
+
+	if ( empty( $filename ) || ! file_exists( $filename ) ) {
+
+		jetpack_wpcom_maybe_log_heif_to_jpg(
+			array(
+				'message'  => sprintf( 'file does not exist: %s', $filename ),
```
danielbachhuber, line 32: Can we make it easier to distinguish between`empty( $filename )` and `! file_exists( $filename )`?

projects/packages/jetpack-mu-wpcom/src/features/media/heif-support.php
```diff
@@ -51,14 +79,35 @@ function jetpack_wpcom_maybe_convert_heif_to_jpg( $filename ) {
...
 }
 
+/**
+ * Log information about HEIF/HEIC conversions.
+ *
+ * @param array $params The parameters to log.
+ *
+ * @return void
+ */
+function jetpack_wpcom_maybe_log_heif_to_jpg( $params = array() ) {
```
danielbachhuber, line 97: It looks like wpcomsh has some code that logs to logstash (see `logstash_log_global_styles`). Can we take a similar approach here, and log in both contexts?

I asked in #dotcom-devs Slack

jeroenpf: Decided that logging on simple sites will be sufficient for now.

projects/packages/jetpack-mu-wpcom/src/features/media/heif-support.php
```diff
@@ -51,14 +87,35 @@ function jetpack_wpcom_maybe_convert_heif_to_jpg( $filename ) {
...
+	if ( ! defined( 'IS_WPCOM' ) || ! IS_WPCOM ) {
```
danielbachhuber: A better check might be `file_exists( WP_CONTENT_DIR . '/lib/log2logstash/log2logstash.php' );`, so we don't accidentally cause fatals if `WP_CONTENT_DIR . '/lib/log2logstash/log2logstash.php'` doesn't exist but `IS_WPCOM` is true.

danielbachhuber (dismissed):
This works as expected.

I applied the branch to my Simple site and verified:
1. The problem is written to Logstash when there's an error.
2. Valid heic files are still converted to jpeg as expected.

I couldn't get the branch applied to WoA site (it wouldn't show up in the Jetpack Beta tester) so I'd recommend doing a final smoke test there.

jeroenpf:
> This works as expected.
> 
> I applied the branch to my Simple site and verified:
> 
> 1. The problem is written to Logstash when there's an error.
> 2. Valid heic files are still converted to jpeg as expected.
> 
> I couldn't get the branch applied to WoA site (it wouldn't show up in the Jetpack Beta tester) so I'd recommend doing a final smoke test there.

It is strange you did not see the branch via the beta tester. This is not the jetpack plugin though, it is a `jetpack-mu-wpcom` plugin so maybe that's where it went wrong.

projects/packages/jetpack-mu-wpcom/src/features/media/heif-support.php
jeroenpf: Thanks for the suggestion, I've changed it.

danielbachhuber (approved):

PR #34072, opened 2023-11-10
Add support for script strategy

### LINKED ISSUE #32852: Assets: switch to registering scripts with `async` and `defer` attributes introduced in WordPress 6.3
> WordPress 6.3 introduces support for registering scripts with async and defer attributes as part of an enhancement to core’s existing Scripts API. 

Source: https://make.wordpress.org/core/2023/07/14/registering-scripts-with-async-and-defer-attributes-in-wordpress-6-3/

When Jetpack requires WordPress 6.3, we should consider updating our methods in Assets to take advantage of that.

*****

Primary issue: #32865

adamsilverstein:
Fixes https://github.com/Automattic/jetpack/issues/32852

## Proposed changes:

WordPress 6.3 introduces support for registering scripts with async and defer attributes as part of an enhancement to core’s existing Scripts API. Since Jetpack now requires WordPress 6.3, we can start relying on those script strategies in our plugins.

Source: https://make.wordpress.org/core/2023/07/14/registering-scripts-with-async-and-defer-attributes-in-wordpress-6-3/

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

* TBD.  This will require testing different features in the plugin.


projects/packages/assets/src/class-assets.php
```diff
@@ -103,6 +103,8 @@ public function script_add_async( $tag, $handle ) {
 	/**
 	 * A helper function that lets you enqueue scripts in an async fashion.
 	 *
+	 * @since n.e.x.t Leverages script strategy for WordPress 6.3 and greater.
```
anomiex:
```suggestion
	 * @since $$next-version$$ Leverages script strategy for WordPress 6.3 and greater.
```

projects/packages/assets/src/class-assets.php
```diff
@@ -307,6 +319,7 @@ public static function normalize_path( $path ) {
 	 * This wrapper handles all of that.
 	 *
 	 * @since 1.12.0
+	 * @since n.e.x.t Update the `in_footer` argument to accept an array to leverage script_strategy.
```
anomiex:
```suggestion
	 * @since $$next-version$$ Update the `in_footer` argument to accept an array to leverage script_strategy.
```

projects/packages/assets/src/class-assets.php
```diff
@@ -112,7 +114,17 @@ public function script_add_async( $tag, $handle ) {
 	 */
 	public static function enqueue_async_script( $handle, $min_path, $non_min_path, $deps = array(), $ver = false, $in_footer = true ) {
 		$assets_instance = self::instance();
-		$assets_instance->add_async_script( $handle );
+
+		// WordPress 6.3 introduces support for async/defer via the strategy attribute.
+		if ( version_compare( get_blog_info( 'version' ), '6.3', '>=' ) ) {
```
anomiex: Why `get_blog_info( 'version' )` rather than comparing with `$wp_version` like we do almost everywhere else?

Too bad there's no relevant public function or method we could feature-detect instead.

projects/packages/assets/src/class-assets.php
```diff
@@ -317,7 +330,7 @@ public static function normalize_path( $path ) {
 	 *  - `css_path`:         (string|null) `.css` to load. Default is to base it on `$path`.
 	 *  - `dependencies`:     (string[]) Additional script dependencies to queue.
 	 *  - `enqueue`:          (bool) Set true to enqueue the script immediately.
-	 *  - `in_footer`:        (bool) Set true to register script for the footer.
+	 *  - `in_footer`:        (bool|array) Pass an array of args to leverage script_strategy since WordPress 6.4. Set true to register script for the footer.
```
anomiex: Let's not do this. We're already taking an array of parameters, let's either use the existing 'async' parameter or deprecate that and create a new 'strategy' parameter that takes 'defer' or 'async' (or explicitly null/false/empty-string, which core seems to have missed documenting).

anomiex (dismissed):
Ok for a start. But at minimum we should also update the public `add_async_script()` function to do like you have on line 396 too.

We should probably also soft-deprecate that and `script_add_async()` (i.e. add `@deprecated` but not `_deprecated_function()`). Maybe even hard deprecate them (note that would also require delaying the hook registration until `script_add_async()` is called the first time).

We might even consider whether we want to release version 2.0 that drops compat with WordPress <6.3 instead of keeping around the back-compat code.

projects/packages/assets/src/class-assets.php
adamsilverstein: thanks for the tip!

projects/packages/assets/src/class-assets.php
adamsilverstein: > Why get_blog_info( 'version' ) rather than comparing with $wp_version like we do almost everywhere else?

I missed that was your approach in the codebase, either approach is valid. I will update here to match what you use.

projects/packages/assets/src/class-assets.php
adamsilverstein: >We're already taking an array of parameters, let's either use the existing 'async' parameter or deprecate that and create a new 'strategy' parameter that takes 'defer' or 'async'

I would suggest matching the core API signature, so accept a `strategy`. in_footer is currently still supported, although it may be deprecated in core at some point in the future.

> (or explicitly null/false/empty-string, which core seems to have missed documenting).

I believe because the intention is  this might be extended to support additional strategies (even by plugins), for example `worker`.

projects/packages/assets/src/class-assets.php
adamsilverstein: updated in caaec1424a

adamsilverstein:
> Ok for a start. But at minimum we should also update the public add_async_script() function to do like you have on line 396 too.

@anomiex - I'm not quite following this comment, where is this update needed (can you give a code link)?

projects/packages/assets/src/class-assets.php
adamsilverstein: I switched to using a new `strategy` option for the array in https://github.com/Automattic/jetpack/pull/34072/commits/e341de8c39a9bbc4d351bd6d8f7db6bf34aa06d6, leaving `in_footer` unchanged. I also extended support to include `async` in https://github.com/Automattic/jetpack/pull/34072/commits/733270b997e2d4a13fed55a0b5c3a1b2afb2b844.

I noticed the function `add_async_script` (and `script_add_async`) actually adds `defer` to the script tags, is that correct? and you don't use the `async` attribute at all, right? 

Would you be open to renaming these function(s)? As is, they are confusing - should you want to use `async` and `defer` they become problematic!

projects/packages/assets/src/class-assets.php
```diff
@@ -323,9 +338,12 @@ public static function normalize_path( $path ) {
 	 *  - `nonmin_path`:      (string) Non-minified script path.
 	 *  - `textdomain`:       (string) Text domain for the script. Required if the script depends on wp-i18n.
 	 *  - `version`:          (string) Override the version from the `asset_path` file.
+	 *  - `strategy`:         (string) Specify a script strategy to use, eg. `defer` or `async`. Default is `""`.
```
anomiex: The rest are in alphabetical order, let's keep that going.
```suggestion
	 *  - `strategy`:         (string) Specify a script strategy to use, eg. `defer` or `async`. Default is `""`.
	 *  - `textdomain`:       (string) Text domain for the script. Required if the script depends on wp-i18n.
	 *  - `version`:          (string) Override the version from the `asset_path` file.
```

projects/packages/assets/src/class-assets.php
anomiex: As I suggested [earlier](https://github.com/Automattic/jetpack/pull/34072#pullrequestreview-1727426868), we should probably deprecate those functions. Even if we renamed them we'd still need to keep the old confusing names around until the next major version bump, and there doesn't seem to be much point in having new correctly-named versions now that WordPress 6.3 includes the same functionality.

anomiex (commented):
> @anomiex - I'm not quite following this comment, where is this update needed (can you give a code link)?

https://github.com/Automattic/jetpack/blob/733270b997e2d4a13fed55a0b5c3a1b2afb2b844/projects/packages/assets/src/class-assets.php#L73-L81

projects/packages/assets/src/class-assets.php
```diff
@@ -377,9 +396,19 @@ function ( $d ) {
...
+			if ( version_compare( $wp_version, '6.3', '>=' ) ) {
+				wp_script_add_data( $handle, 'strategy', 'defer' );
+			} else {
+				// Pre 6.3 approach.
+				self::instance()->add_async_script( $handle ); // Adds `defer` to script tag.
+			}
 		}
+		if ( 'async' === $options['strategy'] && version_compare( $wp_version, '6.3', '>=' ) ) {
+			wp_script_add_data( $handle, 'strategy', 'async' );
+		}
```
anomiex: This seems kind of overdone, compared to something like this
```suggestion
		if ( version_compare( $wp_version, '6.3', '>=' ) ) {
			wp_register_script(
				$handle,
				$url,
				$options['dependencies'],
				$ver,
				array(
					'in_footer' => $options['in_footer'],
					'strategy'  => $options['strategy'],
				)
			);
		} else {
			// Pre 6.3 approach.
			wp_register_script( $handle, $url, $options['dependencies'], $ver, $options['in_footer'] );
			if ( $options['async'] || 'defer' === $options['strategy'] ) {
				static::add_async_script( $handle ); // Adds `defer` to script tag.
			}
		}
```

anomiex:
Also, it would be useful to add tests for the new "strategy" option in
https://github.com/Automattic/jetpack/blob/733270b997e2d4a13fed55a0b5c3a1b2afb2b844/projects/packages/assets/tests/php/test-assets.php#L380-L692

You'll probably need to update some of those tests anyway for where they look for the pre-6.3 way of doing things.

projects/packages/assets/src/class-assets.php
adamsilverstein: > As I suggested earlier, we should probably deprecate those functions. 

Ok, not sure how to do that in Jetpack, I'll see if I can find some examples.

projects/packages/assets/src/class-assets.php
adamsilverstein: I was trying to support the async attribue even in 6.3+, your version is cleaner for sure though so accepting that suggestion

adamsilverstein: well, that doesn't seem to work, I'll try applying manually.

image

adamsilverstein:
thanks for the feedback @anomiex - I'll let you know when I have updates or questions.

projects/packages/assets/src/class-assets.php
anomiex: Step one is to add an `@deprecated` tag in the phpdoc.

Step two would be to add a call to WordPress's `_deprecated_function()` function. But before doing this we'd want to clean up any existing uses within this repository. For `add_async_script` it looks like there's [just one outside this file](https://github.com/Automattic/jetpack/blob/733270b997e2d4a13fed55a0b5c3a1b2afb2b844/projects/packages/stats/src/class-tracking-pixel.php#L110). For `script_add_async` there's only this file and its tests, but you'd need to change this file to stop adding the filter unconditionally.

If you want to deprecate the 'async' option to this function too (which I'd recommend, it has the same naming confusion), you'd probably want to mention that deprecation in the `@since`, and mention that it's deprecated in the option's line too, and after checking that there are no uses you could use `_deprecated_argument()` to complain about it.

projects/packages/assets/src/class-assets.php
anomiex: Looks like GitHub doesn't like trying to avoid some of the limitations in its suggestions thing anymore. 😀 Manually would be the way to go then.

anomiex: Oh, I just noticed I left out a bit of logic in my suggestion. In the 6.3+ case, you'll probably want to do something like
```
if ( $options['strategy'] === '' && $options['async'] ) { // Handle deprecated 'async' option
    $options['strategy'] = 'defer';
}
```
before the call to `wp_register_script`.

adamsilverstein: Ah right, that was the bit I was trying to cover in my version :) Adjusting...

adamsilverstein: added in https://github.com/Automattic/jetpack/pull/34072/commits/90bc7726377deb7cf4624fb8315a4dba1d9ef0e4 and https://github.com/Automattic/jetpack/pull/34072/commits/f57f73fcb562321508a660c49e394acbab8f486a

projects/packages/assets/src/class-assets.php
adamsilverstein: thanks for the detailed instructions, I'll work on it!

adamsilverstein: I'm also deprecating `enqueue_async_script` and updating usage.

Regarding the tests for the deprecated functions, is it ok to leave those in place?

adamsilverstein: > As I suggested earlier, we should probably deprecate those functions.

This part should be complete.

adamsilverstein:
> You'll probably need to update some of those tests anyway for where they look for the pre-6.3 way of doing things.

@anomiex - Do you run tests against multiple WP versions? Should I leave the existing tests for pre 6.3, then add new tests for strategy for 6.3+?

projects/packages/assets/src/class-assets.php
adamsilverstein: I applied this already 👍🏼

projects/packages/assets/src/class-assets.php
anomiex: Yes, it's ok to leave the tests for the deprecated functions in place. They may need to be adjusted for any changes in the code though.

anomiex:
> Do you run tests against multiple WP versions? Should I leave the existing tests for pre 6.3, then add new tests for strategy for 6.3+?

**Short answer:** For this package's tests, there's no WP version installed at all. Any functions needed are mocked using [brain/monkey](https://packagist.org/packages/brain/monkey).

**Long answer:**

For our plugins we run the unit tests against the latest release (6.4), the trunk version, and the "previous" release (6.3). Although the plugin tests may or may not actually make use of the installed WordPress version. Our E2E tests I think only run against the current WP release. Then there are some other tests that live outside of this repository, some run against PRs and some run before deployment to WordPress.com.

But for the packages, like this one, the situation is more complicated. This package doesn't have any version of WordPress available at all; [brain/monkey](https://packagist.org/packages/brain/monkey) is used to mock functions as needed. Other of our packages use [automattic/wordbless](https://packagist.org/packages/automattic/wordbless) to bring in WordPress code to run against. And some of the more integration-y plugin tests may also exercise some code paths in the packages those plugins use.

projects/packages/assets/src/class-assets.php
```diff
@@ -64,18 +64,24 @@ public static function instance() {
...
 	 * A public method for adding the async script.
 	 *
+	 * @deprecated Since $$next-version$$, the `strategy` feature should be used instead.
+	 *
 	 * @param string $script_handle Script handle.
 	 */
 	public static function add_async_script( $script_handle ) {
+		_deprecated_function( __METHOD__, '$$next-version$$ ' );
 		$assets_instance                         = self::instance();
 		$assets_instance->defer_script_handles[] = $script_handle;
```
anomiex: You'll want to do something like this here so this keeps working in 6.3+.
```suggestion
		if ( version_compare( $wp_version, '6.3', '>=' ) ) {
			wp_script_add_data( $script_handle, 'strategy', 'defer' );
		} else {
			$assets_instance                         = self::instance();
			$assets_instance->defer_script_handles[] = $script_handle;
		}
```

projects/packages/assets/src/class-assets.php
```diff
@@ -111,6 +122,7 @@ public function script_add_async( $tag, $handle ) {
 	 * @param bool   $in_footer       Should the script be included in the footer.
 	 */
 	public static function enqueue_async_script( $handle, $min_path, $non_min_path, $deps = array(), $ver = false, $in_footer = true ) {
+		_deprecated_function( __METHOD__, '$$next-version$$ ' );
```
anomiex: Good call, I overlooked this one.

projects/packages/assets/src/class-assets.php
```diff
@@ -307,12 +319,13 @@ public static function normalize_path( $path ) {
 	 * This wrapper handles all of that.
 	 *
 	 * @since 1.12.0
+	 * @since $$next-version$$ Add a new `strategy` option to leverage >= 6.3. script strategy feature. The `async` option is deprecated.
```
anomiex: typo, and clarify what "6.3" refers to.
```suggestion
	 * @since $$next-version$$ Add a new `strategy` option to leverage WP >= 6.3 script strategy feature. The `async` option is deprecated.
```

projects/packages/assets/src/class-assets.php
```diff
@@ -307,12 +319,13 @@ public static function normalize_path( $path ) {
...
 	 * @param string $handle      Name of the script. Should be unique across both scripts and styles.
 	 * @param string $path        Minimized script path.
 	 * @param string $relative_to File that `$path` is relative to. Pass `__FILE__`.
 	 * @param array  $options     Additional options:
 	 *  - `asset_path`:       (string|null) `.asset.php` to load. Default is to base it on `$path`.
-	 *  - `async`:            (bool) Set true to register the script as async, like `Assets::enqueue_async_script()`
+	 *  - `async`:            (bool) Set true to register the script as async, like `Assets::enqueue_async_script()`. Deprecated since $$next-version$$ in favor of `strategy`.
```
anomiex: Even though we can't rename the option, we can at least make the docs more correct. 🙂
```suggestion
	 *  - `async`:            (bool) Set true to register the script as deferred, like `Assets::enqueue_async_script()`. Deprecated in favor of `strategy`.
```

EDIT: Also, it looks like our script that does the `$$next-version$$` token replacement won't replace it here. Maybe we should loosen it up, but for now the `@since` above is probably sufficient to document the version.

projects/packages/assets/src/class-assets.php
```diff
@@ -64,18 +64,24 @@ public static function instance() {
...
-		add_filter( 'script_loader_tag', array( $this, 'script_add_async' ), 10, 2 );
+		if( version_compare( $wp_version, '6.3', '<' ) ) {
+			add_filter( 'script_loader_tag', array( $this, 'script_add_async' ), 10, 2 );
+		}
 	}
 
 	/**
 	 * A public method for adding the async script.
 	 *
+	 * @deprecated Since $$next-version$$, the `strategy` feature should be used instead.
```
anomiex: Should we specifically mention to use "defer" with `strategy` to match the old behavior of this function? Since the name is misleading.

projects/packages/assets/src/class-assets.php
```diff
@@ -321,15 +334,26 @@ public static function normalize_path( $path ) {
...
+
 		if ( substr( $path, -3 ) !== '.js' ) {
 			throw new \InvalidArgumentException( '$path must end in ".js"' );
 		}
 
+		if ( $options['async'] ) {
+			_deprecated_argument(
+				__METHOD__,
+				'$$next-version$$',
+				'The `async` option is deprecated in favor of `strategy`'
```
anomiex: May as well be clear here too.
```suggestion
				'The `async` option is deprecated in favor of `strategy` set to "defer"'
```

projects/packages/assets/src/class-assets.php
```diff
@@ -321,15 +334,26 @@ public static function normalize_path( $path ) {
...
 	 * @throws \InvalidArgumentException If arguments are invalid.
 	 */
 	public static function register_script( $handle, $path, $relative_to, array $options = array() ) {
+		global $wp_version;
+
 		if ( substr( $path, -3 ) !== '.js' ) {
 			throw new \InvalidArgumentException( '$path must end in ".js"' );
 		}
 
+		if ( $options['async'] ) {
```
anomiex: Since you put this above the `$options += array(` bit below, this may result in a `PHP Warning:  Undefined array key "async"`.

projects/packages/assets/src/class-assets.php
```diff
@@ -343,6 +367,7 @@ public static function register_script( $handle, $path, $relative_to, array $opt
 			'media'            => 'all',
 			'minify'           => false,
 			'textdomain'       => null,
+			'strategy'         => '',
```
anomiex: Alphabetical order again.
```suggestion
			'strategy'         => '',
			'textdomain'       => null,
```

projects/packages/stats/src/class-tracking-pixel.php
```diff
@@ -107,7 +107,11 @@ public static function enqueue_stats_script() {
 		);
 
 		// Make sure the script loads asynchronously (add a defer attribute).
-		Assets::instance()->add_async_script( 'jetpack-stats' );
+		if ( version_compare( $wp_version, '6.3', '>=' ) ) {
```
anomiex: May as well just update the `wp_enqueue_script()` call above. We're in the process of dropping pre-6.3 code paths.

anomiex:
I'll go ahead and approve a test run here.

I note that one test that's going to fail will refer to the need for changelog entry files. https://github.com/Automattic/jetpack/blob/trunk/docs/monorepo.md#jetpack-changelogger has details on that.

projects/packages/assets/src/class-assets.php
```diff
@@ -64,18 +64,24 @@ public static function instance() {
...
 
 	/**
 	 * A public method for adding the async script.
 	 *
+	 * @deprecated Since $$next-version$$, the `strategy` feature should be used instead.
+	 *
 	 * @param string $script_handle Script handle.
 	 */
 	public static function add_async_script( $script_handle ) {
+		_deprecated_function( __METHOD__, '$$next-version$$ ' );
```
anomiex:
```suggestion
		_deprecated_function( __METHOD__, '$$next-version$$' );
```

projects/packages/assets/src/class-assets.php
```diff
@@ -84,10 +90,13 @@ public static function add_async_script( $script_handle ) {
 	 * Add an async attribute to scripts that can be loaded deferred.
 	 * https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script
 	 *
+	 * @deprecated Since $$next-version$$, the `strategy` feature should be used instead.
+	 *
 	 * @param string $tag    The <script> tag for the enqueued script.
 	 * @param string $handle The script's registered handle.
 	 */
 	public function script_add_async( $tag, $handle ) {
+		_deprecated_function( __METHOD__, '$$next-version$$ ' );
```
anomiex:
```suggestion
		_deprecated_function( __METHOD__, '$$next-version$$' );
```

projects/packages/assets/src/class-assets.php
anomiex:
```suggestion
		_deprecated_function( __METHOD__, '$$next-version$$' );
```

projects/packages/assets/src/class-assets.php
```diff
@@ -321,15 +334,26 @@ public static function normalize_path( $path ) {
...
 		if ( substr( $path, -3 ) !== '.js' ) {
 			throw new \InvalidArgumentException( '$path must end in ".js"' );
 		}
 
+		if ( $options['async'] ) {
+			_deprecated_argument(
+				__METHOD__,
+				'$$next-version$$',
+				'The `async` option is deprecated in favor of `strategy`'
+			);
```
anomiex: Here our `$$next-version$$` replacing script wants it all on one line. Or at least the `$$next-version$$` bit on the same line as `_deprecated_argument(`.
```suggestion
			_deprecated_argument( __METHOD__, '$$next-version$$', 'The `async` option is deprecated in favor of `strategy`' );
```
Hopefully phpcs doesn't want it otherwise! 😀

anomiex (commented):
The Build job failed because of some errors from the script that replaces the `$$next-version$$` tokens. It's somewhat conservative about where exactly it'll do the replacements.

projects/packages/assets/tests/php/test-assets.php
```diff
@@ -688,6 +688,36 @@ public static function provide_register_script() {
...
+						'single-file',
+						"$url_base/single-js-file.js?minify=false",
+						array(),
+						2883865438,
+						array(
+							'strategy' => 'defer',
+						),
+					),
+					array(),
+				),
```
anomiex: Having the `array()` here is what's causing the `'Variable of type integer' is not a valid function name.` error in the tests (e.g. [here](https://github.com/Automattic/jetpack/actions/runs/6894049950/job/18763774851?pr=34072#step:9:474)), it's saying that the code is expected to call `0()`. But since the number 0 is not a valid function name, :boom:.

Did you mean for that `array()` to be the value for `$extra` instead maybe?
```suggestion
				),
				array(),
```

projects/packages/assets/tests/php/test-assets.php
```diff
@@ -688,6 +688,36 @@ public static function provide_register_script() {
...
+						'single-file',
+						"$url_base/single-js-file.js?minify=false",
+						array(),
+						2883865438,
+						array(
+							'strategy' => 'async',
+						),
+					),
+					array(),
+				),
```
anomiex: Same deal as above. I'll make the suggestion here too in case that's what you intended.
```suggestion
				),
				array(),
```

projects/packages/assets/src/class-assets.php
jeherve: Should be fixed with d15a75d7f075ee6ddf2cb1ddd5f9b942304887ce

projects/packages/assets/src/class-assets.php
jeherve: Would c871128c71b554f7f5de979d071a2a795fa8051c be okay?

projects/packages/assets/changelog/add-script-strategy-support#2
anomiex, line 1: Huh? This *is* the Jetpack Assets Composer package.

jeherve: Oops. Meant to put this in the Stats package. Fixed in aaca0608a2e170b7917d6439d73b58fcf9c38605

jeherve:
I took a stab at updating this PR with the latest `trunk` and addressing the feedback above. Let's now see what the tests have to say. :)

projects/packages/assets/src/class-assets.php
anomiex: LGTM

projects/packages/assets/changelog/add-script-strategy-support
```diff
@@ -0,0 +1,4 @@
+Significance: minor
```
anomiex, line 1: Possibly we should make this major, since it's dropping support for WP before 6.3.

projects/packages/stats/src/class-tracking-pixel.php
```diff
@@ -132,7 +131,7 @@ public static function enqueue_stats_script() {
 		);
 
 		// Make sure the script loads asynchronously (add a defer attribute).
-		Assets::instance()->add_async_script( 'jetpack-stats' );
+		wp_script_add_data( 'jetpack-stats', 'strategy', 'defer' );
```
jeherve: This doesn't seem to work like I would expect:

- It doesn't add a `defer` attribute, just the `data-wp-strategy` attribute set to `defer`.
- Couldn't we update the `wp_enqueue_script` function just above and set it to defer there (although that doesn't seem to add the `defer` attribute either:

```php
wp_enqueue_script(
	'jetpack-stats',
	'https://stats.wp.com/e-' . gmdate( 'YW' ) . '.js',
	array(),
	null, // phpcs:ignore WordPress.WP.EnqueuedResourceParameters.MissingVersion -- The version is set in the URL.
	array(
		'in_footer' => true,
		'strategy'  => 'defer',
	)
);
```

jeherve (commented):
Most things seem to work now, except for the Stats script in my tests (see below).

projects/packages/stats/src/class-tracking-pixel.php
jeherve: > Couldn't we update the wp_enqueue_script function just above and set it to defer there (although that doesn't seem to add the defer attribute either)

Did that in 14dfa23e3154a7fba320cd1a35b739cce1ad0c1a

anomiex (dismissed):
Let's do it!

projects/packages/stats/src/class-tracking-pixel.php
anomiex: > It doesn't add a `defer` attribute, just the `data-wp-strategy` attribute set to `defer`.

Probably because [WordPress doesn't allow handles with inline "after" scripts to be deferred](https://github.com/WordPress/wordpress-develop/blob/6dd00b1ffac54c20c1c1c7721aeebbcd82d0e378/src/wp-includes/class-wp-scripts.php#L984-L987), which [that handle has](https://github.com/Automattic/jetpack/blob/10a2a12cca11cc4c5025fd11c43d2cfcccbc7d4f/projects/packages/stats/src/class-tracking-pixel.php#L150-L154).

jeherve:
> Let's do it!

I had to bump versions one more time. Could you approve this one again?

Thank you!

anomiex (approved):

PR #34073, opened 2023-11-10
RNMobile: Fix broken appearance of video blocks that were first created on the web

SiobhyB:
> **Note**
> The three failing E2E tests on this PR are also failing against other recently opened PRs. I don't believe they're related to this PR's changes.

Fixes a problem in which video blocks that were first created on the web appear broken when viewed in the app. 

Gutenberg Mobile: https://github.com/wordpress-mobile/gutenberg-mobile/pull/6371

## Proposed changes:

* Following the changes in https://github.com/WordPress/gutenberg/pull/49858, `<MediaPlaceholder>` is rendered whenever `src` is falsey. This creates a problem in the case of video blocks that are saved on the web, converted to VideoPress, and then opened in the app. The `src` is falsey in those cases, even though media is already associated with the block. 
* To prevent this problem, a default `src` value has been assigned to video blocks when it's detected that they've been converted to VideoPress.
* The default value is a string, in order for it to be truthy, but is `"undefined"` in an effort to still communicate that it's not set by default.
* See p9ugOq-2W8-p2 for a run-down of the wider issue associated with assigning a correct `src` value to the video in these cases.

### Other information:

## Jetpack product discussion

N/A.

No, it does not.

## Testing instructions:

* Follow the steps to begin a new post on a WordPress.com site. 
* Add the video block to the post and upload a video to it.
* Save the post.
* Navigate to the editor for the saved post on the web.
* Verify that the block is available and that you can access its settings.
* Note, that this fix restores the video block's previous behaviour, in which a placeholder is displayed in place of a thumbnail. Although this isn't ideal behaviour, it's better than a broken state and is something we can iterate on in the future.  

| Before  | After |
|  |  |


derekblank (approved):
I was able to test this successfully. Thanks for improving the Video block placeholder behavior! 🚀 

Also, I agree that the three E2E tests appear broken for other PRs as well, and are unrelated.

PR #34083, opened 2023-11-10
AI Assistant: compute if the site requires an upgrade depending on on the current tier too

retrofox:
This PR computes if the site requires an upgrade based also in the current tier

Fixes #

## Proposed changes:

* AI Assistant: compute if the site requires an upgrade depending on on the current tier too

### Other information:

## Testing instructions:

Run tests locally:

This PR adds tests to test these changes

* Take a look at the test data. Confirm they make sense.
* Then, run the tests locally

```
cd /projects/plugins/jetpack
```

```
pnpm run test-extensions extensions/store 
```

image


projects/plugins/jetpack/extensions/store/wordpress-com/reducer.ts
```diff
@@ -97,16 +97,35 @@ export default function reducer( state = INITIAL_STATE, action ) {
...
 
 		case ACTION_INCREASE_AI_ASSISTANT_REQUESTS_COUNT: {
-			// Increase request count;
 			const requestsCount = state.features.aiAssistant.requestsCount + action.count;
 
+			const currentTierValue = state.features.aiAssistant.currentTier?.value;
+
+			const isFreeTierPlan =
+				( typeof currentTierValue === 'undefined' && ! state.features.aiAssistant.hasFeature ) ||
+				currentTierValue === 1;
```
lhkowalski:
```suggestion
				currentTierValue === 0;
```

projects/plugins/jetpack/extensions/store/wordpress-com/reducer.ts
```diff
@@ -97,16 +97,35 @@ export default function reducer( state = INITIAL_STATE, action ) {
...
+
+			const requestsLimit =
+				isUnlimitedTierPlan || isFreeTierPlan // @todo: update once tier data is available
+					? state.features.aiAssistant.requestsLimit
+					: state.features.aiAssistant.currentTier?.limit;
+
+			const currentCount =
+				isUnlimitedTierPlan || isFreeTierPlan // @todo: update once tier data is available
+					? requestsCount
+					: state.features.aiAssistant.currentTier?.limit;
```
lhkowalski:
```suggestion
					: state.features.aiAssistant.usagePeriod?.requestsCount;
```

projects/plugins/jetpack/extensions/store/wordpress-com/reducer.ts
```diff
@@ -97,16 +97,35 @@ export default function reducer( state = INITIAL_STATE, action ) {
...
+
+			const isFreeTierPlan =
+				( typeof currentTierValue === 'undefined' && ! state.features.aiAssistant.hasFeature ) ||
+				currentTierValue === 1;
+
+			const isUnlimitedTierPlan =
+				( typeof currentTierValue === 'undefined' && state.features.aiAssistant.hasFeature ) ||
+				currentTierValue === 1;
+
+			const requestsLimit =
```
lhkowalski: I think here we should use:

- `state.features.aiAssistant.requestsLimit` for the free plan
- `Infinity` for the unlimited plan
- `state.features.aiAssistant.currentTier?.limit` for a plan with tier info

projects/plugins/jetpack/extensions/store/wordpress-com/reducer.ts
```diff
@@ -97,16 +97,35 @@ export default function reducer( state = INITIAL_STATE, action ) {
...
+
 			/**
 			 * Compute the AI Assistant Feature data optimistically,
 			 * based on the Jetpack_AI_Helper::get_ai_assistance_feature() helper.
 			 *
 			 * @see _inc/lib/class-jetpack-ai-helper.php
 			 */
-			const isOverLimit = requestsCount >= state.features.aiAssistant.requestsLimit;
+			const isOverLimit = currentCount >= requestsLimit;
 			const requireUpgrade = isOverLimit && ! state.features.aiAssistant.hasFeature;
```
lhkowalski: I know this line is already deployed, but it will fail when the user has a tier. When they have a tier, the `hasFeature` will be `true`, so the condition will always fail.

Maybe we shoud just check if it `isOverLimit`. When the site is over the limit, they require an upgrade.

lhkowalski (changes requested):
Left some comments, some can be addressed on a follow-up, but some probably should be solved on this PR.

projects/plugins/jetpack/extensions/store/wordpress-com/reducer.ts
retrofox: > When they have a tier, the hasFeature will be true

This is a mistake from my PoV. `hasFeature` should reflect whether the site supports the feature, no matter the reasons.

retrofox: 🙇‍♂️ Addressed https://github.com/Automattic/jetpack/pull/34083/commits/260106b6c55e711adb78971b0a956e452f0ebde7

CGastrell (approved):
Data makes sense and tests run fine, we can little by little be adding more complex tests there. Thanks! :shipit:

retrofox:
> Data makes sense and tests run fine, we can little by little be adding more complex tests there. Thanks! :shipit:

Correct. Definitely these kind of tests are ideal for the implementation

lhkowalski (approved):
Nice work!

PR #34084, opened 2023-11-11
Add/memberships email to checkout form

chrismccluskey:
Fixes 213-gh-automattic/gold

## Proposed changes:

* Adds a simple function to get the token from the cookie.
* Passes the token to the iframe to identify the user

### Other information:

## Testing instructions:

* Apply D128616-code to your sandbox.
* Download this branch to your sandbox on top of D128616-code (answer 'n' to switching to trunk then 'y' to patch over D128616-code).
* Visit a test site which has both a newsletter tier / subscription and a paid content block (preferably, a paid content block with a different subscription plan, higher than one of the tiers).
* Open a private window and purchase the lowest tier subscription as a new subscriber.
* In the same private window, open a new tab and visit the post with the paid content block.
* Click "Subscribe" for the paid content block. You should a modal showing you your saved cards.
Screenshot 2023-11-14 at 2 07 18 PM
* Click "New card".
* At the bottom of the form, your email address should be filled in automatically for the "Email" field.


projects/plugins/jetpack/extensions/shared/memberships.js
```diff
@@ -98,3 +102,19 @@ export const reloadPageWithPremiumContentQueryString = function (
 	}
 	document.location.href = newQueryString;
 };
+
+export const getTokenFromCookie = function () {
```
millerf: This already exist in `donations/view.js`.
Please remove it from there... and import this one to prevent duplication...

millerf (changes requested):

projects/plugins/jetpack/extensions/shared/memberships.js
chrismccluskey: We had a call about this today and it seems this code is being added to #33800 . I think either approach will work, and looking at your approach I think it might be a little more performant than mine. I'll refactor accordingly.

millerf (approved):

PR #34086, opened 2023-11-11
Subscribe block: return to correct page after subscribing even when not in a post

simison:
Follow up to https://github.com/Automattic/jetpack/pull/33932/files#r1384990526

Improves redirection to correct place which happens after confirming subscription.

Previously we easily sent users to subscription management, which isn't great if they started from the site:

image

## Proposed changes:

* Adds individual page IDs to subscription modal, and defaults to "page on front" id (static homepage or "blog" page) if none is found.

The ID will be used to pull paid status of the post object (which of course is available for blog posts only), but also to pull "redirect URL" which we send reader to after they confirm subscription. Otherwise they end up in Calypso subscription management. 

### Other information:

## Testing instructions:

* Create free and paid posts, test subscribing to those. After confirming subscription, you should land at the post.
* Subscribe from subscription block outside the posts loop, e.g. on homepage header/footer (Use Lettre theme for example) — land at homepage after confirming subscription.
* Subscribe from a page — land at the page after confirming subscription.

### DIFF
--- a/projects/plugins/jetpack/changelog/update-subscriptions-post-id
+++ b/projects/plugins/jetpack/changelog/update-subscriptions-post-id
@@ -1,0 +1,4 @@
+Significance: minor
+Type: enhancement
+
+Subscribe block: return to correct page even when not subscribing from a post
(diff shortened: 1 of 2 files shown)


TimBroddin (approved):
The code looks really good! Tested & approved.

lezama (approved):
nice and simple

PR #34092, opened 2023-11-13
Blogroll: Allow non-wpcom sites to be suggested

renancarvalho:
## Proposed changes:

Currently, Blogroll only allows `wpcom` sites to be suggested by site editors, this PR makes it possible to suggest non-wpcom sites, for instance `www.google.com`.

## Testing instructions:

* Apply and sync this branch to your local environment
* Open your simple site in the editor `wp-admin`
* Add the blogroll block and click on the `plus` button to add another item
* If you type a valid domain, you should be able to select it. Try `www.google.com` or any other valid domain.
* For these non-wpcom sites, we can't allow users to subscribe. Therefore, the subscribe button is invisible for the final users.

image

### Other information:


renancarvalho:
Moved https://github.com/Automattic/jetpack/pull/34112

PR #34098, opened 2023-11-13
Enhancement: add plugin-check workflow

### LINKED ISSUE #34097: Enhancement: Add Plugin Check workflow
### Impacted plugin

Jetpack

### What

Add the WordPress core plugin plugin check as a GitHub action for new PRs and releases

### Description
Plugin Check is a [plugin checking tool](https://github.com/WordPress/plugin-check) developed by the WordPress core team (see [original proposal](https://make.wordpress.org/plugins/2022/07/05/proposal-for-a-wordpress-plugin-checker/)). This tool is used as part of the plugin review process and will help ensure this plugin’s codebase remains up-to-date with the base required standards from the Plugin Review team. The tool includes both static and runtime checks and an ever-growing list of checks for various plugin development requirements and best practices.

### Details
The plugin check can be run automatically using a GitHub [action](https://github.com/swissspidy/wp-plugin-check-action) and the results of the Plugin Check analysis will be added in the form of annotations on new pull requests. Specific groups of checks can be enabled or disabled using the checks attribute. For an initial implementation, I suggest enabling all checks then disabling what you want to ignore. As the Plugin Check grows to include more tests, the list run in the automated action can be trimmed to whatever is preferred here.

adamsilverstein:
Fixes https://github.com/Automattic/jetpack/issues/34097

## Proposed changes:

* Enable the plugin-checker via a github workflow

### Other information:

## Testing instructions:
* Review the checks that ran for this PR
* Approve the workflow to run to see how it works: https://github.com/Automattic/jetpack/actions/runs/6855147597 
* Review the docs at https://github.com/swissspidy/wp-plugin-check-action to see what you might want to exclude

### DIFF
--- a/.github/workflows/plugin-check.yml
+++ b/.github/workflows/plugin-check.yml
@@ -6,16 +6,163 @@ on:
     - trunk
 
 jobs:
-  prepare:
-    name: "Check plugin"
-
+  check-backup-plugin:
+    name: "Check backup plugin"
     runs-on: ubuntu-latest
-
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
-
       - name: Run plugin check
         uses: swissspidy/wp-plugin-check-action@main
         with:
-            build-dir: ./projects/plugins
+          build-dir: ./projects/plugins/backup/
+  check-beta-plugin:
+    name: "Check beta plugin"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Run plugin check
+        uses: swissspidy/wp-plugin-check-action@main
+        with:
+          build-dir: ./projects/plugins/beta/src
+  check-boost-plugin:
+    name: "Check boost plugin"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Run plugin check
+        uses: swissspidy/wp-plugin-check-action@main
+        with:
+          build-dir: ./projects/plugins/boost/app
+  check-crm-plugin:
+    name: "Check crm plugin"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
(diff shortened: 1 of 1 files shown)


anomiex:
I think, before we go forward with reviewing the large amount of code this would wind up running within our monorepo, we should first discuss on https://github.com/Automattic/jetpack/issues/34097 what benefit this would bring to us in the first place.

.github/workflows/plugin-check.yml
```diff
@@ -0,0 +1,168 @@
...
+  check-videopress-plugin:
+    name: "Check videopress plugin"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Run plugin check
+        uses: swissspidy/wp-plugin-check-action@main
+        with:
+          build-dir: ./projects/plugins/videopress
```
anomiex, line 168: This could easily be DRYed using a [matrix](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix). Further, we wouldn't want to hard-code the list of plugins in the matrix either, we'd rather have a job look at the contents of the repo to prepare the matrix.

Also, each plugin is in `projects/plugins/X`. I don't know why you have `./projects/plugins/beta/src` and `./projects/plugins/boost/app` in there, as those are subdirectories of the plugins rather than the actual plugins.

Further, it seems that the the checker wants a _built_ version of the plugin. These directories contain the plugin sources. If it really needs a built version the best way would probably be to add a job to [our Post-Build workflow](https://github.com/Automattic/jetpack/blob/99b340807c502cd57b7629edf536dc214c463115/.github/workflows/post-build.yml), which runs after the build workflow and downloads the built artifacts to test against.

OTOH, it might be more efficient to create our own version of the action that can do multiple plugins within a single job. Particularly with how the Post-Build workflow has to report statuses back to the PR that might make more sense than a matrix.

PR #34101, opened 2023-11-13
Fix issue with newsletter plans not being found

chrismccluskey:
Fixes #

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-regression-with-post-access
+++ b/projects/plugins/jetpack/changelog/fix-regression-with-post-access
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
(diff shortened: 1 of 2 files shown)


n3f:
Discussion: p81Rsd-1u6-p2#comment-1630

millerf (changes requested):
I haven't looked at the issue properly, but this fix is error-prone: source of truth on cahed sites are on the external instance( see this P2 post: p81Rsd-1u6-p2 )
So this fix would create a lot of issue. LEt's close it and discuss about it on a call today.

chrismccluskey:
This PR should not be merged.

Issue #1258, opened 2014-11-04
CSS Concatenating: jetpack.css loads all styles, even if they've been manually dequeued


cferdinandi:
As of this weekend, I'm now seeing `jetpack.css` loading on the front-end, even though the only feature I'm using in Markdown. This file contains CSS for all Jetpack features.

I'm currently overriding this behavior using this snippet in `functions.php`, but this feels like it may be a bug:

``` php
add_filter( 'jetpack_implode_frontend_css', '__return_false' );
```

More info: https://www.twirlingumbrellas.com/wordpress/remove-deregister-jetpack-contact-form-styles/

jeherve:
Also reported here:
https://wordpress.org/support/topic/too-much-css-loading/

jeherve:
Related suggestion:

> Concatenated jetpack.css should only include CSS for active modules

#7656

jeherve:
This was brought up again in #11123.

cferdinandi:
Just wanted to comment that it's been six years and I'm _still_ seeing this issue. 😔

jeherve:
Reopening this, as it is still a valid enhancement.

adaminfinitum:
I came across this thread after a google search led me to [issue #15734](https://github.com/Automattic/jetpack/issues/15734), which is closed because it's basically a duplicate of this issue.

I'm glad this is still open because it is a valid enhancement, what I am working on now to address it is [adding a filter to disable the concatenation](https://developer.jetpack.com/hooks/jetpack_implode_frontend_css/) and then using the "Asset CleanUp" plugin to disable the files I don't need and letting my caching plugin concatenate everything else into as few files as possible.

Hopefully others will find this approach helpful.

jeherve:
This was also brought up in [this post](https://felix-arntz.me/blog/rebuilding-my-website-using-a-block-theme/):

> The only feature of Jetpack that I use in the frontend of my site is the Subscribe block that you can see in my site footer. There’s no way that simple text input and button needs 21.5 KB of CSS.

As we had discussed in #7656, we should try to find a way to not include everything in the concatenated file if it is not needed.

jeherve:
I'd like to come back to this issue and summarize the conversations above a bit:

[...]

PR #34110, opened 2023-11-14
Performance: only load concatenated CSS when needed by at least 2 modules

jeherve:
See #1258

## Proposed changes:

> **Note**
> This is only a partial improvement. We still need to address the larger questions around the implementation in today's world, as discussed here:
> https://github.com/Automattic/jetpack/issues/1258#issuecomment-1507814494

We currently enqueue a `jetpack.css` concatenated CSS file as soon as your site is connected to WordPress.com.
That `jetpack.css` file is built off a list of stylesheets from individual modules that are not enqueued when the concatenated file is enqueued.

In this commit, we add a new condition: for the file to be enqueued, there has to be at least 2 modules that need it. This will avoid loading the file when it wouldn't even be needed at all.
This requires a change in the implementation: we now check for active modules before we dequeue individual assets / enqueue jetpack.css. This is an additional call to get an option (see `( new Modules() )->get_active()`).
This also means that the new `$modules_with_concatenated_css` variable will need to be kept updated.

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

* Go to Jetpack > Settings > Modules, and deactivate all modules.
* In the site's `wp-config.php` file, ensure that `SCRIPT_DEBUG` is not set to true.
* Load the site's frontend
    * You should not see any `jetpack.css` file being loaded.
* Activate the Related Posts Module.
    * There should stil be no `jetpack.css` file, but you should now see a CSS file loaded for Related Posts.
* Activate the Likes module.
    * Now you should see a `jetpack.css` file, and no longer any CSS file loaded for Related Posts or Likes.


monsieur-z (approved):
The PR works as expected. The strategy and implementation look good too. :ship:

PR #34064, opened 2023-11-10
Fix post editor dirty state caused by publicize store

manzoorwanijk:
Opening the Jetpack sidebar on the post edit page results in dirty state of the post editor and thus preventing users from reloading or navigating away from the page.

Screenshot 2023-11-13 at 12 39 17 PM

## Proposed changes:

* Move the connections state and actions from `jetpack/publicize` to `jetpack-social-store` and update consumer components to use the new store.
* The source of truth for the connections will now be `jetpack-social-store` instead of post meta.
* Connections in post meta will only be synced to post meta when a user explicitly toggles a connection

**Note:**
After this PR, the store inside `projects/js-packages/publicize-components/src/store` technically contains the utility functions not the actual selectors or actions because they mostly talk to the `core/editor` store not the `jetpack/publicize` store. That means we can convert those selectors/actions to simple utility functions and get rid of `jetpack/publicize` store.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

- Run `jetpack build plugins/jetpack`
- Ensure that you have some social connections added for the site
- Goto post edit page for an existing post and open Jetpack sidebar
- Wait for the `connection-test-results` API request to finish
- Try to reload the page
- Confirm that their is no confirmation popup about unsaved changes
- Toggle a connection
- Confirm that the UI reflects the toggled connection state
- Try to reload the page
- Confirm that this time you see the popup
- Confirm that the social preview details like name and profile picture work correctly
- Create a new post and publish it
- Confirm that it is posted to the social media connections

PR #34111, opened 2023-11-14
Get rid of `jetpack/publicize` store

manzoorwanijk:
Following #34064, this PR cleans up the jetpack social store usage

## Proposed changes:

* Remove `jetpack/publicize` store
* Convert the store sections to re-usable utility functions
* Thus get rid of `lodash` and `refx` deps from `publicize-components` 

### Other information:

## Testing instructions:

* Go to post editor
* Confirm that all the Jetpack Social and SIG related settings work correctly


projects/js-packages/publicize-components/src/utils/types.ts
```diff
@@ -0,0 +1,17 @@
+export type SIGSettings = {
+	template?: string;
```
pablinos: Should the SIG settings also have the image to use?

projects/js-packages/publicize-components/src/utils/is-nextdoor-connection-supported.ts
```diff
@@ -0,0 +1,10 @@
+import { getJetpackData } from '@automattic/jetpack-shared-extension-utils';
+
+/**
+ * Checks if the Instagram connection is supported.
+ *
+ * @returns {boolean} Whether the Instagram connection is supported
+ */
+export function isNextdoorConnectionSupported() {
+	return !! getJetpackData()?.social?.isNextdoorConnectionSupported;
```
pablinos: Seeing these three like this makes me wonder if these should be combined in some way, there's quite a bit of boilerplate in here

projects/js-packages/publicize-components/src/components/social-previews/modal.js
```diff
@@ -72,7 +72,7 @@ export default withSelect( select => {
 	// Use the featured image by default, if it's available.
 	let image = featuredImageId ? getMediaSourceUrl( getMedia( featuredImageId ) ) : '';
 
-	const sigSettings = getImageGeneratorPostSettings();
+	const sigSettings = getImageGeneratorSettings();
```
pablinos: I think the use of Post in the name was to distinguish it from the blog/plugin settings. As long as we don't think it could be confusing then I think it's fine to remove.

pablinos (commented):
Looks good, but it's quite a big refactor. I think it would be nice to add some tests here too.

projects/js-packages/publicize-components/src/utils/types.ts
manzoorwanijk: Updated.

projects/js-packages/publicize-components/src/utils/is-nextdoor-connection-supported.ts
manzoorwanijk: Yeah, may be we should pass an array of available connections from the backend.

projects/js-packages/publicize-components/src/components/social-previews/modal.js
manzoorwanijk: Reverted to `getImageGeneratorPostSettings`

projects/js-packages/publicize-components/src/utils/is-nextdoor-connection-supported.ts
manzoorwanijk: Done ✅

manzoorwanijk:
@pablinos I have added the unit tests

projects/js-packages/publicize-components/src/components/connection/index.js
```diff
@@ -47,7 +46,7 @@ class PublicizeConnection extends Component {
 	 *
 	 * @returns {boolean} True if connection must be reauthenticated.
 	 */
-	connectionNeedsReauth = () => includes( this.props.mustReauthConnections, this.props.name );
+	connectionNeedsReauth = () => this.props.mustReauthConnections.includes( this.props.name );
```
pablinos, line 49: I'm guessing not, but could `mustReauthConnections` ever be null?

projects/js-packages/publicize-components/src/utils/get-share-message.js
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Gets the message that will be used hen sharing this post.
+ *
+ * @returns {string} The share message.
+ */
+export function getShareMessage() {
+	const { getEditedPostAttribute } = select( editorStore );
+	const meta = getEditedPostAttribute( 'meta' );
+	const message = `${ meta?.jetpack_publicize_message || '' }`;
```
pablinos, line 13: What benefit does this have over:

```suggestion
	const message = meta?.jetpack_publicize_message || '';
```

Is it that it ensures it's a string?

pablinos (approved):
Works well for me. I've got a couple of comments, but they're more just queries than anything else.

Thanks for adding the tests.

projects/js-packages/publicize-components/src/components/connection/index.js
manzoorwanijk: No, it will always be an array.

projects/js-packages/publicize-components/src/utils/get-share-message.js
manzoorwanijk: Yes, just to ensure it's a string to allow auto completion.

PR #34112, opened 2023-11-14
Blogroll: Allow non-wpcom sites to be suggested

renancarvalho:
## Proposed changes:

Currently, Blogroll only allows `wpcom` sites to be suggested by site editors, this PR makes it possible to suggest non-wpcom sites, for instance `www.google.com`.

## Testing instructions:

* Apply and sync this branch to your local environment
* Open your simple site in the editor `wp-admin`
* Add the blogroll block and click on the `plus` button to add another item
* If you type a valid domain, you should be able to select it. Try `www.google.com` or any other valid domain.
* For these non-wpcom sites, we can't allow users to subscribe. Therefore, the subscribe button is invisible for the final users.

image

### Other information:


projects/plugins/jetpack/extensions/blocks/blogroll/utils.js
```diff
@@ -8,14 +8,15 @@ export function createBlockFromRecommendation( attrs ) {
 }
 
 export function createBlockFromSubscription( subscription ) {
-	const { blog_id, site_icon, URL, name, description } = subscription;
+	const { blog_id, id, site_icon, URL, name, description, is_non_wpcom_site } = subscription;
 
 	return createBlockFromRecommendation( {
-		id: blog_id,
```
agrullon95, line 14: I'm seeing a 'Saving failed` error when I manually search and add a WPCOM site.
CleanShot 2023-11-16 at 15 06 32@2x

I no longer see the error if I change this line from `id: blog_id || id` to `id: blog_id`. Do we need the additional `id` value?

projects/plugins/jetpack/extensions/blocks/blogroll/use-get-site-details.js
```diff
@@ -12,12 +12,21 @@ export default function useGetSiteDetails( { siteURL, subscriptions, enabled = f
 
 	if ( wpcomSite ) {
 		results.unshift( {
-			id: wpcomSite?.ID,
+			blog_id: wpcomSite?.ID,
 			description: wpcomSite?.description,
 			URL: wpcomSite?.URL,
 			site_icon: getSiteIcon( wpcomSite?.logo?.url ),
 			name: wpcomSite?.name,
```
agrullon95: Not a blocker, but can we add `is_non_wpcom_site: false` to this for object structure consistency?

agrullon95 (dismissed):
I was able to manually add a wpcom site and non wpcom site.
Both appeared correctly on the frontend. The wpcom site had the subcribe button available and the non wpcom site did not have the subscribe button available.

I left comments related to `saving failed` error I ran into when adding and saving a WPCOM site.

heavyweight:
https://github.com/Automattic/jetpack/assets/7000684/6dcc70a3-37d3-4475-b134-4b496b478549

For some reason, I get an invalid site being added see video.
Debugging this..

heavyweight (changes requested):
the wrong replacement that happens is 
```
<!-- wp:jetpack/blogroll-item {"id":156306909,"name":"The View from Where I Sit","url":"https://theviewfromwheresit.blog","description":"Disability Advocacy and Empowerment"} /--></div>
<!-- /wp:jetpack/blogroll -->
```

projects/plugins/jetpack/extensions/blocks/blogroll/use-get-site-details.js
```diff
@@ -12,12 +12,21 @@ export default function useGetSiteDetails( { siteURL, subscriptions, enabled = f
...
 		} );
+	} else if ( validDomain && ! isLoading ) {
+		results.unshift( {
+			blog_id: validDomain,
```
heavyweight, line 26: `blog_id: validDomain` doesn't make sense to keep and track imo

projects/plugins/jetpack/extensions/blocks/blogroll/utils.js
renancarvalho: Thank you for reviewing, the issue was because sites coming from blogroll suggestions had their ID as a string and sites coming from search itself had them as integers. Apparently, it was not failing in the BE but showing this message in the FE. I pushed a change to threat all IDs as strings that fixed this issue.

projects/plugins/jetpack/extensions/blocks/blogroll/use-get-site-details.js
renancarvalho: This happens because our API is returning `Facebook` as a wpcom site, which obviously is not. I could also see it happening with `gmail.com`. I am digging this API to understand why it happens, but it is for sure nothing related to blogroll only. Likewise, I believe whoever uses `https://public-api.wordpress.com/rest/v1.1/sites` API is exposed to it.

renancarvalho: We need to have a unique ID somehow, empty IDs won't work. Since these `non_wpcom_sites` have no ID, we thought we could use their URL, which is unique and the only information we have about it, since the ID is a string, it works fine.

renancarvalho:
> Screen.Recording.2023-11-20.at.21.16.24.mov 
> For some reason, I get an invalid site being added see video. Debugging this..

This happens because our API is returning Facebook as a wpcom site, which obviously is not. I could also see it happening with gmail.com. I am digging this API to understand why it happens, but it is for sure nothing related to blogroll only. Likewise, I believe whoever uses https://public-api.wordpress.com/rest/v1.1/sites API is exposed to it.

heavyweight (dismissed):
Tested together on a call and looking good

heavyweight (approved):

PR #33969, opened 2023-11-06
Social: Fix store/initial-state out of sync issue

gmjuhasz:
In our store, a feature is available if the main Jetpack Social toggle is on. Right now if you have Jetpack Social off, and go to the settings page, turning Jetpack Social on will show SIG and auto conversion off even if they are on in the settings, as the initial state was loaded with Jetpack Social off

## Proposed changes:

* Refresh the store from the blog options when we turn Jetpack Social on

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

https://github.com/orgs/Automattic/projects/733/views/2?pane=issue&itemId=43051048
## Does this pull request change what data or activity we track or use?

No.
## Testing instructions:

* Open up your site and turn off Jetpack Social off on the admin page. Refresh the page and turn it on.
* After loading the features should be on if they were on before turning the feature off
* Without this change following these steps will show the features being off.

PR #34113, opened 2023-11-14
Social: Refactor feature option storing

gmjuhasz:
This change refactors how we store our options for SIG and auto conversion. We moved to use the register_setting function, this way we need no Rest controllers.
It's easier to review this following the commits.

FEATURE BRANCH OTHER PRs:
* Social: Remove deprecated feature files #34114
* Social: Refresh store on Publicize module toggle via 1 call #34142

## Proposed changes:

* f37dd1a70f2259ef236166a183b2eec1c1c1f2c5 - Added the new settings file and tests for it.
* d00d07e9d311f341e33b250b224c18a1f73cb359 - Updated the initial state, so it uses the new object. Also changed every usage where SIG needed the default template
* 012eb2ce2482f79ea11ff7a40c6ba08d38696a39 - This commit updates the UI parts to match the new data structure, and also updates the store so it calls the core wp/v2 endpoint to retrieve and update the settings.

### Other information:

## Jetpack product discussion

https://github.com/orgs/Automattic/projects/733/views/2?pane=issue&itemId=44049666

No.
## Testing instructions:

This probably needs a lot of testing to see that everything is working. You can run the tests on the backend with `composer run test-php` from the `publicize` folder.

To test the flow you can do the following:
* Have both Jetpack and Jetpack Social enabled. Open up The Jetpack Settings, and go to the sharing section. On another tab open up the Social admin page.
  * Toggle on/off SIG and the conversion in one of the pages, then refreshing the other page should reflect the changes. Try to change the template too.
  * Make sure that turning the main module off, refreshing, then turning it on shows the correct values.
* Have the conversion on, then create a new post. The default state of SIG in the editor should match the state of the toggle in the admin page. 

* Clear the dismissed notices with `jetpack docker wp option delete jetpack_social_dismissed_notices`
* Open up a new post with the conversion on, then upload an invalid image that can be converted. You should see the notice that the image will be converted. 
  * Click on change settings, then turn the conversion off
  * Switch back to the other tab, the page should reflect the changes without needing to refresh it


projects/js-packages/publicize-components/src/social-store/controls.js
```diff
@@ -92,23 +92,31 @@ export default {
...
+		return apiFetch( {
+			path: '/wp/v2/settings?_fields=jetpack_social_autoconvert_images',
+		} );
 	},
 	[ UPDATE_AUTO_CONVERSION_SETTINGS ]: function ( action ) {
 		return apiFetch( {
-			path: '/jetpack/v4/auto-conversion/settings',
+			path: '/wp/v2/settings',
 			method: 'POST',
-			data: action.settings,
```
manzoorwanijk, line 111: Why do we need multiple API calls? Can we make a single API call to get all the settings at once?

manzoorwanijk: We can add the settings fields separated by comma

`/wp/v2/settings?_fields=jetpack_social_image_generator_settings,jetpack_social_autoconvert_images`

pablinos: Yes, to fetch the settings we can just have one call.

gmjuhasz: Yeah, I'll do that.

gmjuhasz: This turned out to be a bigger change as I had to remove the old refreshes introduced in #33969, so I added the store functions in b6952a283d187e95d955adda49092e834a6cd2b2 and created another PR on this branch for the others #34142

manzoorwanijk (dismissed):
This works well now

manzoorwanijk (dismissed):

projects/packages/publicize/src/jetpack-social-settings/class-settings.php
```diff
@@ -0,0 +1,251 @@
...
+		);
+
+		// The feature cannot be enabled without Publicize.
+		if ( ! ( new Modules() )->is_active( 'publicize' ) ) {
+			$settings['autoConversionSettings']['enabled']       = false;
+			$settings['socialImageGeneratorSettings']['enabled'] = false;
+		}
+
+		if ( $with_available ) {
+			$settings['autoConversionSettings']['available']       = $this->is_auto_conversion_available();
```
pablinos, line 155: I was giving this a test, went to the Social admin page, and got this error:
```
Cannot access offset of type string on string in /usr/local/src/jetpack-monorepo/projects/packages/publicize/src/jetpack-social-settings/class-settings.php:144
```

My guess is that the options on my site are in a weird state, but we should maybe check for that. The return of `get_option` might not be an array.

pablinos (commented):
I've added a few thoughts. Probably the biggest thing we need to look at is updating things on the WPCOM side

projects/packages/publicize/src/jetpack-social-settings/class-settings.php
gmjuhasz: Added some safety checks, I tested it with clean state, then with the one someone would have coming from trunk

projects/packages/publicize/src/jetpack-social-settings/class-settings.php
```diff
@@ -0,0 +1,263 @@
...
+	 *
+	 * @return array
+	 */
+	public function get_settings( $with_available = false ) {
+		$this->migrate_old_option();
+
+		$settings = array(
+			'autoConversionSettings'       => get_option( self::OPTION_PREFIX . self::AUTOCONVERT_IMAGES, self::DEFAULT_AUTOCONVERT_IMAGES_SETTINGS ),
+			'socialImageGeneratorSettings' => get_option( self::OPTION_PREFIX . self::IMAGE_GENERATOR_SETTINGS, self::DEFAULT_IMAGE_GENERATOR_SETTINGS ),
+		);
```
pablinos, line 146: The problem I was having was that my auto conversion setting had got mangled in some way. And was set to just `1` rather than `[ 'enabled' => 1]`. I'm not sure we need to do anything much here, but we could check for it during migration.

projects/packages/publicize/src/jetpack-social-settings/class-settings.php
```diff
@@ -0,0 +1,263 @@
...
+	 * jetpack_social_image_generator_settings option. Now they are stored in the jetpack_social_settings
+	 * together with the auto conversion settings.
+	 *
+	 * TODO: Work out if this is possible on plugin upgrade
+	 *
+	 * @return void
+	 */
+	private function migrate_old_option() {
+		$auto_conversion_settings = get_option( 'jetpack_social_settings' );
+		if ( ! empty( $auto_conversion_settings['image'] ) ) {
```
pablinos: How about checking `empty( $auto_conversion_settings )` here? If we have any settings then we want to migrate them, and if someone has disabled the auto-conversion this will be `false` and we'll go back to the default which is `true`

projects/packages/publicize/src/jetpack-social-settings/class-settings.php
```diff
@@ -0,0 +1,263 @@
...
+			update_option( self::OPTION_PREFIX . self::AUTOCONVERT_IMAGES, array( 'enabled' => $auto_conversion_settings['image'] ) );
```
pablinos: Probably need to check the logic here, but then this would be something like:

```suggestion
			update_option( self::OPTION_PREFIX . self::AUTOCONVERT_IMAGES, array( 'enabled' => ! empty( $auto_conversion_settings['image'] ) ) );
```

projects/packages/publicize/src/jetpack-social-settings/class-settings.php
pablinos: What safety checks did you add? My problem was that `$settings['autoConversionSettings']` wasn't an array, which would probably not happen anywhere but my mangled dev site, but would it be worth validating the shape of the options that we get? Maybe that's overkill!

pablinos (commented):
Once I fixed my rogue setting problem it's working well. I'm not sure how much we want to safeguard against that, but there might be a couple of things worth adding.

gmjuhasz:
@pablinos 
>What safety checks did you add? My problem was that $settings['autoConversionSettings'] wasn't an array, which would probably not happen anywhere but my mangled dev site, but would it be worth validating the shape of the options that we get? Maybe that's overkill!

I think now I made sure that the object shape is correct, if it's not, the migration will run, and then it will be. That runs before getting here, so it should be fine I think.

projects/packages/publicize/src/jetpack-social-settings/class-settings.php
gmjuhasz: https://github.com/Automattic/jetpack/pull/34113/commits/0ba898a4fa9547ec6d74da7c16b70d0257a7e0e7

projects/packages/publicize/src/jetpack-social-settings/class-settings.php
gmjuhasz: https://github.com/Automattic/jetpack/pull/34113/commits/0ba898a4fa9547ec6d74da7c16b70d0257a7e0e7

projects/plugins/jetpack/tests/php/sync/test_class.jetpack-sync-options.php
```diff
@@ -219,6 +219,7 @@ public function test_sync_default_options() {
 			'jetpack_publicize_options'                    => array(),
 			'jetpack_connection_active_plugins'            => array( 'jetpack' ),
 			'jetpack_social_settings'                      => array( 'image' => true ),
```
pablinos, line 221: Do you think we should stop syncing this one?

pablinos (commented):
This looks good to me now. I wondered if we should stop syncing the other option, but otherwise, I think we might be able to finally get it merged!

projects/plugins/jetpack/tests/php/sync/test_class.jetpack-sync-options.php
gmjuhasz: I think we talked about leaving it for now on Friday, and let's remove it in the next one

projects/js-packages/publicize-components/src/social-store/actions/jetpack-social-settings.js
```diff
@@ -0,0 +1,39 @@
...
+		yield setSocialImageGeneratorSettings(
+			updatedSettings.jetpack_social_image_generator_settings
+		);
+		return true;
+	} catch ( e ) {
+		return false;
+	} finally {
+		yield setUpdatingAutoConversionSettingsDone();
+		yield setUpdatingSocialImageGeneratorSettingsDone();
+	}
```
manzoorwanijk, line 34: TIL that `finally` block is executed before the `return` statement inside `try/catch`.

manzoorwanijk (approved):
I tested it with and without Jetpack plugin and it works fine and the code also looks good.

projects/packages/publicize/src/jetpack-social-settings/class-settings.php
```diff
@@ -0,0 +1,263 @@
...
+					'schema' => array(
+						'type'       => 'object',
+						'properties' => array(
+							'enabled' => array(
+								'type' => 'boolean',
+							),
+						),
+					),
+				),
+				'type'         => 'boolean',
```
pablinos, line 101: I think something might be wrong here, this type should be `object`. Do you think that's how it ended up being the empty string? I think maybe something has been cast as a boolean and it's evaluated to false.

PR #34130, opened 2023-11-14
Paywall Block: Better access question

lezama:
See p9Jlb4-911-p2#comment-9659

## Proposed changes:

Screenshot 2023-11-14 at 19 01 29

For Simple Sites:

Subscriber-only paywall -> Already a subscriber? Log in.
Paid-subscriber only paywall -> Already a paid subscriber? Log in.
Higher tier paywall -> Already a higher-tier paid subscriber? Log in.

For Jetpack sites:

Subscriber-only paywall -> Already a subscriber?
Paid-subscriber only paywall -> Already a paid subscriber?
Higher tier paywall -> Already a higher-tier paid subscriber?

### Other information:

## Testing instructions:

* Test paywalls on Jetpack and Simple sites.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -987,11 +987,9 @@ function get_paywall_blocks( $newsletter_access_level ) {
 
 		}
 	} else {
-		$access_question = get_paywall_access_question( $newsletter_access_level );
 		if ( ( new Host() )->is_wpcom_simple() ) {
 			// custom domain
-			$sign_in_link     = wpcom_logmein_redirect_url( get_current_url(), false, null, 'link', get_current_blog_id() );
-			$access_question .= ' ' . __( 'Log in.', 'jetpack' );
+			$sign_in_link = wpcom_logmein_redirect_url( get_current_url(), false, null, 'link', get_current_blog_id() );
 		} else {
 			$sign_in_link = add_query_arg(
 				array(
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -1046,14 +1047,14 @@ function get_paywall_access_question( $post_access_level ) {
 			if ( $tier !== null ) {
 				return sprintf(
 				// translators:  Placeholder is the tier name
-					__( 'I am already a <i>%s</i> paid-subscriber', 'jetpack' ),
+					__( 'Already a <i>%s</i> paid subscriber?', 'jetpack' ),
```
simison: We talked with @crisbusquets to drop the tier name in the mid-sentence out, but didn't actually come up with alternative that I remember.

crisbusquets: This was the last solution we discussed:

Subscriber-only paywall -> Already a subscriber? Log in.
Paid-subscriber only paywall -> Already a paid subscriber? Log in.
Higher tier paywall -> Already a higher-tier paid subscriber? Log in.

The 'Log in' is the link. Or maybe everything could be clickable. What do you think?

simison: Nice, thanks! 👍 

Let's drop "Log in" as pre-caution to issues Fab and Miguel were raising, make sure that all sentences are question format (that makes it more actionable rather than without question mark like it's in production) and make the whole sentence clickable/underlined.

lezama: Perfect, thanks, applied your suggestions.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -1046,14 +1045,14 @@ function get_paywall_access_question( $post_access_level ) {
...
```
simison: Needs updating, see copy Cris shared below.

simison (dismissed):
Approving based on code review (untested). Just small copy update still and then we're good.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
lezama: done

simison (approved):

Issue #32865, opened 2023-09-05
WordPress 6.4 compatibility


jeherve:
WordPress 6.4 is set to release on November 7, 2023.

- Planning: [make.wordpress.org/core/6-4](https://make.wordpress.org/core/6-4/)
- PT: pf5801-82-p2 

Let's use this issue to keep track of anything that may come up until then.

```[tasklist]
- [ ] #33525
- [ ] https://github.com/Automattic/jetpack/issues/33643
- [ ] https://github.com/Automattic/jetpack/issues/33642
- [ ] https://github.com/Automattic/jetpack/issues/33654
- [ ] https://github.com/Automattic/jetpack/issues/33655
- [ ] https://github.com/Automattic/jetpack/pull/33709
- [ ] https://github.com/Automattic/jetpack/issues/33779
- [ ] https://github.com/Automattic/jetpack/issues/33874
```

And **in a follow-up plugin release** (most likely the December release, 12.9), to give folks time to update:

```[tasklist]
- [ ] https://github.com/Automattic/jetpack/pull/34127
- [ ] https://github.com/Automattic/jetpack/issues/34209
- [ ] #32852
- [ ] #32853
- [ ] #32854
- [ ] #32863
- [ ] #32864
- [ ] #31714
- [ ] #31716
- [x] Coordinate deactivation of the Lazy Images feature in Jetpack in favor of Core's (see #33048)
```

In addition to the above, we can follow the dev-notes that will drop here about the release:
[make.wordpress.org/core/tag/dev-notes+6-4](https://make.wordpress.org/core/tag/dev-notes+6-4)

-------

For reference: prior art for WordPress 6.3 compat: #31638

jeherve:
Marking this one as done; all tasks have now been handled.

PR #34135, opened 2023-11-15
Code Modernization: Replace strpos() with str_starts_with()

### LINKED ISSUE #32854: Code modernization: Replace usage of `strpos` with `str_starts_with`
Source: https://core.trac.wordpress.org/ticket/58012

Let's follow Core's lead once Jetpack requires WordPress 6.3.

Primary issue: #32865

jeherve:
Fixes #32854

## Proposed changes:

`str_starts_with()` was introduced in PHP 8.0 to perform a case-sensitive check indicating if the string to search in (haystack) begins with the given substring (needle).

WordPress core includes a polyfill for `str_starts_with()` on PHP < 8.0 as of WordPress 5.9.

This commit replaces usage of `strpos()` with `str_starts_with()`, making the code more readable and consistent, as well as improving performance in PHP 8+ environments.

While `strpos()` is slightly faster than the polyfill on PHP < 8.0, `str_starts_with()` is noticeably faster on PHP 8.0+, as it is optimized to avoid unnecessarily searching along the whole haystack if it does not find the needle.

> **Note**
> 
> This doesn't update uses of `strpos()` in a few places:
> - The autoloader since it can be used outside of WordPress.
> - The VaultPress plugin since it can still be used on older versions of WordPress, where the `str_starts_with()` shim did not exist.

Matching core change: https://core.trac.wordpress.org/ticket/58012

### Other information:

## Jetpack product discussion

* Primary issue: #32865

* No

## Testing instructions:

* Activate all touched plugins and watch out for fatals on the frontend and wp-admin.


kraftbj:
@jeherve A few of the tests need mocks for `str_starts_with` or something similar.

jeherve:
Good point. I just pushed 81955927e5d2c021fabcfb2702b575aae440e9b1 to try to fix that.

projects/packages/assets/src/class-assets.php
```diff
@@ -450,9 +450,9 @@ public static function wp_default_scripts_hook( $wp_scripts ) {
 		$content_dir = Jetpack_Constants::get_constant( 'WP_CONTENT_DIR' );
 		$abspath     = Jetpack_Constants::get_constant( 'ABSPATH' );
 
-		if ( strpos( $lang_dir, $content_dir ) === 0 ) {
+		if ( \str_starts_with( $lang_dir, $content_dir ) ) {
```
anomiex: I'm not sure that `wp-includes/compat.php` winds up getting loaded from the `wp-admin/load-scripts.php` entry point where this hook gets called from.

projects/packages/assets/tests/php/patchwork.json
```diff
@@ -1,3 +1,3 @@
 {
-	"redefinable-internals": [ "filemtime", "is_readable" ]
+	"redefinable-internals": [ "filemtime", "is_readable", "str_starts_with" ]
```
anomiex: No need to mock it, just polyfill it in bootstrap.php or something like

```
if ( ! function_exists( 'str_starts_with' ) ) {
    function str_starts_with( ...
}
```

projects/packages/assets/src/class-assets.php
anomiex: I note that `wp-admin/load-styles.php` has [a comment calling out that lack](https://github.com/WordPress/wordpress-develop/blob/0b8ca16ea3bd9722bd1a38f8ab68901506b1a0e7/src/wp-admin/load-styles.php#L76).

projects/packages/redirect/tests/php/patchwork.json
```diff
@@ -0,0 +1,3 @@
+{
+	"redefinable-internals": [ "str_starts_with" ]
```
anomiex: Same, no need to mock, just polyfill in bootstrap.php or something

projects/packages/assets/tests/php/patchwork.json
jeherve: 👍 4eaadb626f7a0b6e834a5534b80138beac8d4ae1

projects/packages/redirect/tests/php/patchwork.json
jeherve: 👍 4eaadb626f7a0b6e834a5534b80138beac8d4ae1

projects/packages/assets/src/class-assets.php
jeherve: That makes sense. c020b9f9e8ea93f414af7b966a86cb7ddd1715dc

projects/packages/assets/changelog/update-strpos-str-starts-with
```diff
@@ -0,0 +1,4 @@
+Significance: patch
+Type: changed
+
+Code Modernization: Replace usage of strpos() with str_starts_with().
```
anomiex:
```suggestion
Comment: Comment why we don't use str_starts_with() in Assets::wp_default_scripts_hook. No change to functionality.

```

anomiex (commented):
Needs a trunk merge, otherwise it looks ok to me

anomiex (approved):

PR #34139, opened 2023-11-15
Add accessible name to Contact Form dropdown

### LINKED ISSUE #30762: Jetpack Form Block: A11Y - ARIA input field doesn't have accessible text
### Impacted plugin

Jetpack

### Quick summary

There is an issue on the Jetpack Form Block regarding Accessibility. 

`A11Y - ARIA input field doesn't have accessible text`
Screenshot 2023-05-16 at 21 46 47

Here is a related issue: https://github.com/Automattic/jetpack/issues/16685 

### Steps to reproduce

1. Login to WP Admin
2. Add the Jetpack Form to a page 
3. Add a dropdown block 
4. Save and load the front end 
5. Inspect to see the missing A11y fields

_No response_

### What actually happened

_No response_

### Impact

Some (< 50%)

### Available workarounds?

_No response_

### Logs or notes

_No response_

monsieur-z:
Fixes https://github.com/Automattic/jetpack/issues/30762

## Proposed changes:

In the front-end of a site, the dropdown from _Contact Form_ doesn't have an accessible name. It makes it hard for screen reader users to know the purpose of the input.

Using a native `select` tag, the association with a `label` is done by adding the `for` attribute to the latter. However, Contact Form uses `jquery-ui-selectmenu`, which makes the `select` component invisible to assistive technologies and replaces it with a custom dropdown of which the root is a `span`. The association made with the `label` `for` attribute is lost.

The fastest solution to reestablish this association is to add the `aria-labelled-by` attribute to the aforementioned `span` and set the `label` `id` as its value.

Concretely, what's done in this PR is:
- Add an `id` to the `label` rendered just before the dropdown
- Reference that id on the `select` tag via a `data-` attribute, so that we can get it back later
- When initializing a jQuery UI select menu, retrieve that id
- Set that id to an `aria-labelled-by` attribute on the custom dropdown created by jQuery UI

_Note: the safest strategy to keep components accessible is usually to follow standards and use native solutions when available. In this case, we should consider getting rid of jQuery UI in favor of a native `select` and try to style that tag instead._

### Other information:

## Jetpack product discussion

n/a

No.

## Testing instructions:

- Create a new post: visit `/wp-admin/post-new.php`
- Add the _Contact Form_ block to the page
Screenshot 2023-11-15 at 10 20 30 AM

- Add a dropdown
Screenshot 2023-11-15 at 10 21 29 AM

- Open the page preview
- Inspect the dropdown and select the `span` jQuery UI created just after the `select` tag
- You should be able to find an _Accessibility_ panel (screenshots from Chrome)
Screenshot 2023-11-15 at 10 14 54 AM

- Notice the `span` has an accessible name (the _Name_ property in the _Computed properties_ panel below)
Screenshot 2023-11-15 at 10 15 26 AM

### DIFF
--- a/projects/packages/forms/src/contact-form/js/dropdown.js
+++ b/projects/packages/forms/src/contact-form/js/dropdown.js
@@ -13,14 +13,21 @@ jQuery( function ( $ ) {
 	} );
 
 	function initializeSelectMenu() {
-		$( '.contact-form .contact-form-dropdown' )
-			.selectmenu( {
-				classes: {
-					'ui-selectmenu-button': 'contact-form-dropdown__button',
-					'ui-selectmenu-menu': 'contact-form-dropdown__menu',
-				},
-			} )
-			.attr( 'aria-hidden', true )
-			.prop( 'tabindex', -1 );
+		$( '.contact-form .contact-form-dropdown' ).each( function () {
+			const $select = $( this );
+			const labelId = $select.data( 'contact-form-label-id' );
+
+			$select
+				.selectmenu( {
+					classes: {
+						'ui-selectmenu-button': 'contact-form-dropdown__button',
+						'ui-selectmenu-menu': 'contact-form-dropdown__menu',
+					},
+				} )
+				.attr( 'aria-hidden', true )
+				.prop( 'tabindex', -1 )
+				.selectmenu( 'widget' )
+				.attr( 'aria-labelledby', labelId );
+		} );
 	}
 } );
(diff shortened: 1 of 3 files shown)


projects/packages/forms/src/contact-form/js/dropdown.js
```diff
@@ -13,14 +13,21 @@ jQuery( function ( $ ) {
...
+			$select
+				.selectmenu( {
+					classes: {
+						'ui-selectmenu-button': 'contact-form-dropdown__button',
+						'ui-selectmenu-menu': 'contact-form-dropdown__menu',
+					},
+				} )
+				.attr( 'aria-hidden', true )
+				.prop( 'tabindex', -1 )
+				.selectmenu( 'widget' )
```
monsieur-z, line 29: Retrieve the root `span` tag.

andrii-lysenko (approved):
The code looks good. Tested on Self hosted and Simple, works as described 👍

PR #34142, opened 2023-11-15
Social: Refresh store on Publicize module toggle via 1 call

gmjuhasz:
This change updates how we refresh our options on the Publicize module change, and will use the common api fetch, so we can decrease the number of API calls.

## Proposed changes:

* Removed the separated refresh from the toggle components 07cf749fca37012a69c0c35dcf5aedbc7c7443dd
* Added a common refresh into the admin page and to Jetpack Dashboard. 61facfa7bbd9a7b492e9d295071314e28a22848a and 1902f5d6d4b1b808169a0acf0f70b4ed0f153bd5

### Other information:

## Jetpack product discussion

https://github.com/orgs/Automattic/projects/733/views/2?pane=issue&itemId=44049666

No.
## Testing instructions:

* Turn SIG and auto conversion on.
* Turn off the main Publicize module.
* Refresh the page
* Turn on the module, the child toggles should reflect their actual state
* There should be only 1 call to the API

### DIFF
--- a/projects/js-packages/publicize-components/src/components/refresh-jetpack-settings/index.js
+++ b/projects/js-packages/publicize-components/src/components/refresh-jetpack-settings/index.js
@@ -1,5 +1,4 @@
 import { useDispatch } from '@wordpress/data';
-import { useEffect } from '@wordpress/element';
 import { SOCIAL_STORE_ID } from '../../social-store';
 
 /**
@@ -13,11 +12,9 @@ import { SOCIAL_STORE_ID } from '../../social-store';
 export default function RefreshJetpackSocialSettingsWrapper( { shouldRefresh, children } ) {
 	const refreshOptions = useDispatch( SOCIAL_STORE_ID ).refreshJetpackSocialSettings;
 
-	useEffect( () => {
-		if ( shouldRefresh ) {
-			refreshOptions();
-		}
-	} );
+	if ( shouldRefresh ) {
+		refreshOptions();
+	}
 
 	return children;
 }
--- a/projects/js-packages/publicize-components/src/components/refresh-jetpack-settings/index.js
+++ b/projects/js-packages/publicize-components/src/components/refresh-jetpack-settings/index.js
@@ -1,0 +1,23 @@
+import { useDispatch } from '@wordpress/data';
+import { useEffect } from '@wordpress/element';
+import { SOCIAL_STORE_ID } from '../../social-store';
+
+/**
+ * HOC that refreshes all of the Jetpack Social settings in the store, to be used in class components.
+ *
+ * @param {object} props - The component props.
+ * @param {boolean} props.shouldRefresh - Whether or not to refresh the settings.
+ * @param {object} props.children - The children to render.
+ * @returns { object } The refreshJetpackSocialSettings function.
+ */
+export default function RefreshJetpackSocialSettingsWrapper( { shouldRefresh, children } ) {
+	const refreshOptions = useDispatch( SOCIAL_STORE_ID ).refreshJetpackSocialSettings;
+
+	useEffect( () => {
+		if ( shouldRefresh ) {
+			refreshOptions();
+		}
+	} );
+
+	return children;
+}
--- a/projects/plugins/jetpack/_inc/client/sharing/publicize.jsx
+++ b/projects/plugins/jetpack/_inc/client/sharing/publicize.jsx
@@ -1,4 +1,5 @@
 import { getRedirectUrl } from '@automattic/jetpack-components';
+import { RefreshJetpackSocialSettingsWrapper } from '@automattic/jetpack-publicize-components';
 import { createInterpolateElement } from '@wordpress/element';
 import { __, _x } from '@wordpress/i18n';
 import Card from 'components/card';
@@ -154,10 +155,14 @@ export const Publicize = withModuleSettingsFormHelpers(
 							>
 								{ __( 'Automatically share your posts to social networks', 'jetpack' ) }
 							</ModuleToggle>
-							{ shouldShowChildElements && hasAutoConversion && <AutoConversionSection /> }
-							{ shouldShowChildElements && hasSocialImageGenerator && (
-								<SocialImageGeneratorSection />
-							) }
+							<RefreshJetpackSocialSettingsWrapper
+								shouldRefresh={ ! isActive && this.props.isSavingAnyOption( 'publicize' ) }
+							>
+								{ shouldShowChildElements && hasAutoConversion && <AutoConversionSection /> }
+								{ shouldShowChildElements && hasSocialImageGenerator && (
+									<SocialImageGeneratorSection />
+								) }
+							</RefreshJetpackSocialSettingsWrapper>
 						</SettingsGroup>
 					) }
 
--- a/projects/plugins/social/src/js/components/admin-page/index.jsx
+++ b/projects/plugins/social/src/js/components/admin-page/index.jsx
@@ -30,16 +30,6 @@ const Admin = () => {
 
 	const refreshJetpackSocialSettings = useDispatch( SOCIAL_STORE_ID ).refreshJetpackSocialSettings;
 
-	useEffect( () => {
-		if (
-			isModuleEnabled &&
-			isUpdatingJetpackSettings &&
-			( isAutoConversionAvailable || isSocialImageGeneratorAvailable )
-		) {
-			refreshJetpackSocialSettings();
-		}
-	} );
-
 	const onUpgradeToggle = useCallback( () => setForceDisplayPricingPage( true ), [] );
 	const onPricingPageDismiss = useCallback( () => setForceDisplayPricingPage( false ), [] );
 
@@ -67,6 +57,18 @@ const Admin = () => {
 			isUpdatingJetpackSettings: store.isUpdatingJetpackSettings(),
 		};
 	} );
+
+	useEffect( () => {
+		if (
+			isModuleEnabled &&
+			isUpdatingJetpackSettings &&
+			( isAutoConversionAvailable || isSocialImageGeneratorAvailable )
+		) {
+			refreshJetpackSocialSettings();
+		}
+		// We want this to run only once, when the module is enabled.
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [ isModuleEnabled ] );
 
 	const moduleName = `Jetpack Social ${ pluginVersion }`;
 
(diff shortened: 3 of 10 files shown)


projects/plugins/social/src/js/components/admin-page/index.jsx
```diff
@@ -28,6 +28,18 @@ const Admin = () => {
 	const showConnectionCard = ! isRegistered || ! isUserConnected;
 	const [ forceDisplayPricingPage, setForceDisplayPricingPage ] = useState( false );
 
+	const refreshJetpackSocialSettings = useDispatch( SOCIAL_STORE_ID ).refreshJetpackSocialSettings;
+
+	useEffect( () => {
+		if (
+			isModuleEnabled &&
+			isUpdatingJetpackSettings &&
```
pablinos: Should this be:

```suggestion
			! isUpdatingJetpackSettings &&
```

projects/plugins/social/src/js/components/admin-page/index.jsx
```diff
@@ -28,6 +28,18 @@ const Admin = () => {
...
+			( isAutoConversionAvailable || isSocialImageGeneratorAvailable )
+		) {
+			refreshJetpackSocialSettings();
+		}
+	} );
```
pablinos: This hasn't got a dependency array, which will mean that it runs on every render. I think putting it a useEffect means that it runs after the render rather than before. Is that what's intended?

projects/plugins/social/src/js/components/admin-page/index.jsx
```diff
@@ -90,10 +102,10 @@ const Admin = () => {
 						<InstagramNotice onUpgrade={ onUpgradeToggle } />
 						<SocialModuleToggle />
 						{ ! isUpdatingJetpackSettings && isModuleEnabled && isAutoConversionAvailable && (
-							<AutoConversionToggle shouldRefresh />
+							<AutoConversionToggle />
```
pablinos: Should we still render it but disable it if the state is updating? It might mean that there's less page jank. I don't know if there's a busy/loading state that we can use with the toggles, but that would be great.

projects/js-packages/publicize-components/src/components/refresh-jetpack-settings/index.js
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * HOC that refreshes all of the Jetpack Social settings in the store, to be used in class components.
+ *
+ * @param {object} props - The component props.
+ * @param {boolean} props.shouldRefresh - Whether or not to refresh the settings.
+ * @param {object} props.children - The children to render.
+ * @returns { object } The refreshJetpackSocialSettings function.
+ */
+export default function RefreshJetpackSocialSettingsWrapper( { shouldRefresh, children } ) {
```
pablinos, line 12: I wonder if this file should be in a folder named `refresh-social-settings`

projects/plugins/jetpack/_inc/client/sharing/publicize.jsx
```diff
@@ -154,10 +155,14 @@ export const Publicize = withModuleSettingsFormHelpers(
 							>
 								{ __( 'Automatically share your posts to social networks', 'jetpack' ) }
 							</ModuleToggle>
-							{ shouldShowChildElements && hasAutoConversion && <AutoConversionSection /> }
-							{ shouldShowChildElements && hasSocialImageGenerator && (
-								<SocialImageGeneratorSection />
-							) }
+							<RefreshJetpackSocialSettingsWrapper
```
pablinos, line 158: Are we using this component here so we can encapsulate the `useEffect` because this is a class component?

That's probably fine, but it might be worth checking that we don't want to do something like handle it with a lifecycle method in here.

pablinos (commented):
Looks good. I need to give it a proper test, but I've left a few comments

projects/js-packages/publicize-components/src/components/refresh-jetpack-settings/index.js
```diff
@@ -0,0 +1,23 @@
...
+	const refreshOptions = useDispatch( SOCIAL_STORE_ID ).refreshJetpackSocialSettings;
+
+	useEffect( () => {
+		if ( shouldRefresh ) {
+			refreshOptions();
+		}
+	} );
```
manzoorwanijk: I don't think this needs `useEffect`. It can simply be `shouldRefresh && refreshOptions()`, which means this component may not even be required.

projects/plugins/jetpack/_inc/client/sharing/publicize.jsx
manzoorwanijk: `useEffect` is not even needed for this and thus a lifecycle method may be a good place for it.

projects/plugins/social/src/js/components/admin-page/index.jsx
manzoorwanijk: This `useEffect` may also be unnecessary.

projects/plugins/jetpack/_inc/client/sharing/publicize.jsx
gmjuhasz: Yeah, and also `useDispatch`. I added the useEffect to fix an issue earlier where the inner component was updated while the parent was changing, but it seems that's already fixed and it might not be needed. I think we still need the component to be able to dispatch the store function, or we need to wrap publicize.jsx into another HOC component

projects/js-packages/publicize-components/src/components/refresh-jetpack-settings/index.js
gmjuhasz: replied https://github.com/Automattic/jetpack/pull/34142#discussion_r1395348927

projects/plugins/social/src/js/components/admin-page/index.jsx
gmjuhasz: There is no page jank, the store has an isUpdating flag for each which disables the toggle while updating. Is that what you ask about?

projects/plugins/social/src/js/components/admin-page/index.jsx
gmjuhasz: I think it makes sense to update the store while the main module is turning on, this way it will be refreshed already when the toggles appear, reducing the page jank

projects/plugins/social/src/js/components/admin-page/index.jsx
gmjuhasz: https://github.com/Automattic/jetpack/pull/34142#discussion_r1395348927

projects/plugins/social/src/js/components/admin-page/index.jsx
pablinos: Not quite, I was thinking that there's an ` ! isUpdatingJetpackSettings` check, so if `isUpdatingJetpackSettings` we won't rending the toggle at all.

gmjuhasz: Oh I see what you mean. It's working this way on Jetpack Settings that's what I mimicked, but I can make that be there even with Publicize off if you think that's better.

projects/plugins/social/src/js/components/admin-page/index.jsx
```diff
@@ -62,8 +64,10 @@ const Admin = () => {
 		if (
 			isModuleEnabled &&
 			isUpdatingJetpackSettings &&
```
pablinos: I think we can remove this check now

projects/plugins/social/src/js/components/admin-page/index.jsx
```diff
@@ -62,8 +64,10 @@ const Admin = () => {
...
+			! hasRefreshedSettings.current &&
 			( isAutoConversionAvailable || isSocialImageGeneratorAvailable )
 		) {
+			hasRefreshedSettings.current = true;
 			refreshJetpackSocialSettings();
 		}
 		// We want this to run only once, when the module is enabled.
```
pablinos: It should be fine to include the other dependencies. `hasRefreshedSettings` should be stable as it's a ref, and the `is...Available` values shouldn't change either.

projects/plugins/social/src/js/components/admin-page/index.jsx
```diff
@@ -103,11 +107,11 @@ const Admin = () => {
 						{ shouldShowAdvancedPlanNudge && <AdvancedUpsellNotice /> }
 						<InstagramNotice onUpgrade={ onUpgradeToggle } />
 						<SocialModuleToggle />
-						{ ! isUpdatingJetpackSettings && isModuleEnabled && isAutoConversionAvailable && (
-							<AutoConversionToggle />
+						{ isModuleEnabled && isAutoConversionAvailable && (
+							<AutoConversionToggle disabled={ isUpdatingJetpackSettings } />
```
pablinos, line 113: `isUpdatingJetpackSettings` is whether the module value is being updated, I think? Have we got an equivalent for `isRefreshingSocialSettings`?

pablinos (commented):
Looks good. Couple of thoughts, as always! Sorry...

projects/plugins/social/src/js/components/admin-page/index.jsx
gmjuhasz: I think we cant as then it would refresh just when the toggle is on. That means if it's on, it will instantly refresh on each page refresh, while we only want it to refresh on off->on changes

projects/plugins/social/src/js/components/admin-page/index.jsx
gmjuhasz: Yeah, we have, and that one is already considered in the toggle itself. I added this here to avoid the scenario where the `isRefreshingSocialSettings` finishes faster than the module toggle update, and there is that weird state where the module toggle is still disabled and updating, but the children are not. This way it always waits for the main toggle to be not disabled.

projects/plugins/social/src/js/components/admin-page/index.jsx
pablinos: Oh, I thought we'd get around that by using the reference. It could be something like `hasEnabledModule` which would be set to `true` if it is enabled on page load, and so would prevent us from ever refreshing the settings. If the module is enabled but initially it wasn't, then we fetch the settings and set the reference to `true`

gmjuhasz: Done in https://github.com/Automattic/jetpack/pull/34142/commits/4b5f559992941d4383d6aa1c16e78841ca89d8ea

manzoorwanijk (approved):
It looks good now.

Issue #33824, opened 2023-10-26
AI Tiered Plans: revisit the feature endpoint to consider tiers when setting the flags


lhkowalski:
Example:

If the current tier is 200 and the current period usage is 201, the `site-requires-upgrade` should be true. Same for `has-feature` and all the other fields.

---

Some conversation about it: p1698344317896489-slack-C054LN8RNVA, at the end of the thread.

lhkowalski:
Below is a summary of the endpoint current properties, their meaning and some comments about changing it or not:

- `has-feature`
   - there is some discussion about the meaning of this field
   - currently it represents the value of `wpcom_site_has_feature( 'ai-assistant' )` to mark it it has the feature provided by some plan or any other mean
   - maybe it should represent if the feature is available on the site, considering limits of the free plan/tiers and all other factors that may block the feature; in this case, it would be the opposite, negated value of `is-over-limit`
- `is-over-limit`
   - originally represented if the site was over the free limit of requests
   - we are changing it to represent if the site is over the limit of requests that the current plan/tier provides to the site (free, unlimited, 100, 200, 500)
   - being addressed by #34151
- `requests-count`
   - represents the number of requests executed by the site since the beggining (the all time count)
   - could use a better name, like `allTimeRequestsCount`
- `requests-limit`
   - represents the number of free requests allowed to the site type
   - could use a better name, like `freeRequestsLimit`
- `usage-period`
   - info about the current usage period (start dates and request count)
   - no changes needed so far
- `site-require-upgrade`
   - represents if the site needs an upgrade to keep running requests
   - with tiered plans, the site may need an upgrade even when it already have a tier, to get a bigger one to allow more requests
   - we need to revisit it because currently it only asks for upgrades on free sites, and that is not the case anymore
   - a good value for it is the `is-over-limit` value; when the site is over limit, an upgrade is required
   - being addressed by #34170
- `upgrade-type`
   - represents the type of the upgrade message, refering to different site types
   - no changes needed so far
- `current-tier`
   - info about the current tier of the site
   - no changes needed so far
- `next-tier`
   - info about the next tier the site may/will upgrade to
   - no changes needed so far 
- `tier-plans`
   - the list of tiers enabled for the site
   - no changes needed so far

lhkowalski:
Updating, the last 3 fields that could deserve some changes are:

- `has-feature`:
[...]

PR #34151, opened 2023-11-15
Jetpack AI: Use the Usage Helper to set ai-assistant-feature endpoint fields

lhkowalski:
Part of #33824.

## Proposed changes:

* Use the `WPCOM\Jetpack_AI\Usage\Helper` to set:
   * `is-over-limit` (if the site is over the allowed limit, be it the free or the tiered one)
   * `requests-limit` (the free requests limit for the site type)
   * `requests-count` (the amount of requests since the beggining)
* Remove code that was checking for the OpenAI_* class presence, since they are not in use anymore

### Other information:

## Testing instructions:

* Let's test it comparing the values before and after the change
* Before applying this change to your sandbox, go to the WordPress.com console (https://developer.wordpress.com/docs/api/console/)
* Execute a request to the `wpcom/v2` `/sites/$wpcom_site/jetpack-ai/ai-assistant-feature` endpoint using your test site and note the results:

Screenshot 2023-11-15 at 17 45 35

* Apply this change to your sandbox and make sure you are sandboxing the `public-api`  
* Go to the WordPress.com console again
* Execute the same request as before, then compare the results:

Screenshot 2023-11-15 at 17 50 21

* The values for `requests-limit` and `requests-count` should be the same as before
* The value for `is-over-limit` may change because now a new rule is applied (the rule from D128826-code):
   * before, the rule was considering if the site was over the free limit of requests (a site with a paid plan could easily be over the 20 requests limit)
   * now, the rule considers if the site is over the limit of requests of the current plan it has (be it free or unlimited, and will also work for tiered plans)
   * confirm the value for `is-over-limit` makes sense by comparing the `requests-count` to the expected limit of the plan active on the site (free or unlimited)
* If possible, test for a Jetpack site and a Simple site

### DIFF
--- a/projects/plugins/jetpack/_inc/lib/class-jetpack-ai-helper.php
+++ b/projects/plugins/jetpack/_inc/lib/class-jetpack-ai-helper.php
@@ -343,31 +343,6 @@ class Jetpack_AI_Helper {
 
 		if ( defined( 'IS_WPCOM' ) && IS_WPCOM ) {
 			$has_ai_assistant_feature = \wpcom_site_has_feature( 'ai-assistant' );
-			if ( ! class_exists( 'OpenAI' ) ) {
-				\require_lib( 'openai' );
-			}
-
-			if ( ! class_exists( 'OpenAI_Limit_Usage' ) ) {
-				if ( is_readable( WP_CONTENT_DIR . '/lib/openai/openai-limit-usage.php' ) ) {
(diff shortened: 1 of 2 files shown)


CGastrell:
Tests look ok:
Without patch vs with patch. P2 site on top and simple site with AI subscription at the bottom
image

CGastrell (approved):
LGTM! Thanks :shipit:

PR #34161, opened 2023-11-16
Check for image data before calling imagecreatefromstring()

josephscott:
The current code can run into a few conditions where the end result is that imagecreatefromstring() is called on something that isn't actually an image.

I found two conditions where that happens. The first is where we end up with a NULL value at the end. The second is where a remote image is requested, but what you get back is something else ( like the HTML for an error page ).

This update adds checks to defend against both of those possible error conditions.


jeherve:
@josephscott Since this PR is open on a fork, it doesn't come with the utilities we have to run checks, and quickly test that PR on WordPress.com Simple. I've consequently opened a new PR with your changes and a few more things in #34164. Could you give it a look?

Thank you!

PR #34164, opened 2023-11-16
Toneque: Check for image before calling imagecreatefromstring()

jeherve:
## Proposed changes:

> **Note**
> This is a replay of #34161, on a branch instead of a fork.

The current code can run into a few conditions where the end result is that `imagecreatefromstring()` is called on something that isn't actually an image.

I found two conditions where that happens. The first is where we end up with a NULL value at the end. The second is where a remote image is requested, but what you get back is something else ( like the HTML for an error page ).

This update adds checks to defend against both of those possible error conditions.

It also moves away from a Jetpack-only function to a PHP function, `str_starts_with`, that has a polyfill in WP Core for folks still running an older version of PHP.

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

* TBD


jeherve (commented):
@josephscott Is there any chance you could check this PR against the site where you ran into the issue, or add some testing instructions to the PR for anyone wanting to reproduce this? I'm essentially curious as to when Tonesque ends up being called against something that is not an image. :)

Thank you!

josephscott:
@jeherve Thank you.  Yes, confirmed, I tried out this version of the patch and it solved the divide by zero issues I was seeing.

projects/plugins/jetpack/_inc/lib/tonesque.php
```diff
@@ -72,21 +72,24 @@ public static function imagecreatefromurl( $image_url ) {
...
 					}
 				}
 			}
 
 			if ( empty( $data ) ) {
 				$response = wp_safe_remote_get( $image_url );
-				if ( is_wp_error( $response ) ) {
+				if (
+					is_wp_error( $response )
+					|| ! str_starts_with( $response['headers']['content-type'], 'image/' )
```
anomiex: Hopefully there aren't too many misconfigured servers out there that serve images as `application/octet-stream` or the like.

anomiex (commented):
Looks ok. One comment, if we're fine with that possibility I'll approve.

projects/plugins/jetpack/_inc/lib/tonesque.php
josephscott: That would definitely be an issue.  Are you aware of any sites/services that do that?

anomiex: I don't know of any offhand, but the `$response['headers']['content-type']` will have whatever value the remote server decides to return for the content type, if it specifies a type at all.

Most servers will probably get it right, but some may not for some image types. For example, if someone's webserver is old enough it may not know about `image/webp` to call webp images that type.

josephscott: Do you have a preference on an alternative?  My goal is to avoid calling `imagecreatefromstring()` on something that isn't a function, and am happy to try out other approaches.

anomiex: Really calling `imagecreatefromstring()` (maybe with `@` and manual error handling) is probably the most straightforward way to check if you actually got an image.

Detecting an error page would be better done by looking at the response code. If the server returned a 2xx code, it's probably not an error page, and otherwise it probably is.

If you really want to check the type, you'd have to use something like `jetpack_mime_content_type()` except every function I see for doing something like that (including that one) wants to be passed a filename rather than the contents of the file.

josephscott: Hiding things behind `@` is generally discouraged.  I would be ok with a `2xx` response check, as that still does better than what we have now.

What about updating the conditional to be:

```php
            if ( empty( $data ) ) {
                $response = wp_safe_remote_get( $image_url );
                $response_code = (int) $response['response']['code'];
                if (
                    is_wp_error( $response )
                    || $response_code < 200
                    || $response_code > 299
                ) {
                    return false;
                }
                $data = wp_remote_retrieve_body( $response );
            }
```

/cc @jeherve

jeherve: Would something like this work for you?
493296901735d18c7e2d26c94af137c568401bf2

Edit: although you did say 2xx, so I suppose 0858e52e0835bf932b01649919a91e2c4def222c would be better, my bad.

anomiex: > Hiding things behind `@` is generally discouraged.

There's a difference between using `@` to ignore errors and using `@` to suppress PHP's default warnings in order to handle the problem in a better manner. But I'm fine with the response code check here, it's still an improvement.

anomiex (approved):

josephscott:
Thank you for getting this through, much appreciated!

PR #34165, opened 2023-11-16
Jetpack Sync: Gracefully handle exceptions when calculating checksums

fgiannar:
Jetpack Sync: Gracefully handle exceptions when calculating checksums

## Proposed changes:

* `Automattic\Jetpack\Sync\Replicastore\Table_Checksum_Users`: Adds missing `use Exception;`
* `Automattic\Jetpack\Sync\Replicastore`: Updates `checksum_histogram` method to gracefully handle exceptions by adding missing try/catch blocks and returning the corresponding `WP_Error`

**Bonus Performance Imrpovement:** I noticed that we are duplicating the same DB query (related to `get_range_edges`) when calculating buckets so I fixed that too.
**Bonus Performance Imrpovement 2:** No need to calculate buckets when retrieving the histogram in `checksum_histogram` when `$only_range_edges` is `true`. This means that we are only requesting the range edges and not the actual checksum histogram.

### Other information:

No

## Testing instructions:

- Using your sandbox you will need to trigger checksums/fetch range edges only while throwing Exceptions and confirm that you receive a corresponding WP_Error and the Exception was caught
- Using your sandbox ensure that when triggering checksums the buckets calculated match with/without the PR applies
- You will need to also test this on WPCOM too (see instructions below for testing on Simple sites)


darssen (dismissed):
Thanks a lot for working on this @fgiannar!

Changes look good! I especially like the find to not duplicate the same DB query when calculating buckets.

I tested changes in my sandbox for local environment and everything works as expected.

darssen (approved):

PR #34169, opened 2023-11-16
AI Assistant: default unlimited request limit

CGastrell:
Use same value as in backend for unlimited plan limit.

## Proposed changes:
This PR replaces the old value set by PHP_INT_MAX and uses a more JS safe value. It also removes the use of `Infinity` for the use of the new constant.

### Other information:

No

## Testing instructions:
The editor should just not break. Visual review, but also you can search for the values on the code. First run a build on the plugin (so we don't get false positives from older builds), either with build or watch. Then just `grep`:

```
// search for the old value, shouldn't be present anymore:
grep -r 922337203685477600 projects/plugins/jetpack/ --exclude-dir={node_modules,build,vendor,tests} --exclude={*css,*map,*md}

// search for the new value. This will give you some false positives on the `_inc` folder, ignore or just check that it's the same constant declared on different build contexts
grep -r 999999999 projects/plugins/jetpack/ --exclude-dir={node_modules,build,vendor,tests,test} --exclude={*css,*map,*md}

// same as above
grep -r UNLIMITED_PLAN_REQUESTS_LIMIT projects/plugins/jetpack/ --exclude-dir={node_modules,build,vendor,tests,test} --exclude={*css,*map,*md}
```


lhkowalski (approved):
Well done! Now both client and backend have the same limit.

PR #34173, opened 2023-11-16
Make Contact Form error fixing accessible

### LINKED ISSUE #14155: A11Y bug: Errors disappear in subscription & contactforms
Hello,
In jetpack subscription & contact form modules the error messages appear only for few seconds. The focus moves to the field that has the error & when user tabs out the error messages disappear. As a screen reader user I am unable to see the error messages using my arrow keys & this is a failure of WCAG SC 2.2.1 timing adjustable. 

The error messages must be available all time until the user successfully completes the form. Is there a hook or a temporary fix to achieve this until this is adopted into jetpack?

#### What happened instead

**Screenshots**

monsieur-z:
Fixes https://github.com/Automattic/jetpack/issues/14155 https://github.com/Automattic/jetpack/issues/34235

## Proposed changes:

In this PR, we make Contact forms more accessible by overwriting the native form validation. Browsers handle form validation differently, making accessible forms for all users difficult. Specifically, we:
- Overwrite the form native validation by adding the `novalidate` attribute on the `form`.
- Register and enqueue the custom validation script.
- Rely on the [Contraint Validation API](https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation#the_constraint_validation_api) to report fields validity.
- Show an error message above the submit button if one or more fields are invalid. This is the right place to display a message since the button is on the screen, and the user is focused on it. Scrolling or focusing on another place can be confusing. The error message is also read out loud by screen readers thanks to `aria-live=assertive`.
- For each invalid field, mark it as such and show a more specific message.

### Other information:

## Jetpack product discussion

n/a

No.

## Testing instructions:

- Create a new post
- Add a _Contact form_ block to the post
- View the preview
- Check that when one or more fields are invalid, an error message is displayed above the submit button and below each invalid field
- Fix the errors and make sure you can submit the form successfully
- Check that the layout is not broken for all three styles: _Default_, _Outlined_, and _Animated_
- Optionally, repeat these steps by adding various field types
- Optionally, test the above by adding a _Shortcode_ block instead
Screenshot 2023-11-21 at 10 48 48 AM

### Before

Notice that the focus is immediately switched to the first invalid field on submitting. When tabbing out of the field, the error message disappears. This requires a user to submit the form again to be able to see the error message.

https://github.com/Automattic/jetpack/assets/1620183/1a8831f2-615e-4062-9c1d-702cfc421d69

### After

On submitting the form, an error message is shown above the submit button but no scroll or focusing is happening. We let the user tab back and forth to investigate which fields need to be corrected. Error messages are visible at all time and announced by the screen reader.

https://github.com/Automattic/jetpack/assets/1620183/5eb6bf92-db6a-4fa6-8607-667fcebb2d50


projects/packages/forms/src/contact-form/css/grunion.css
```diff
@@ -172,7 +174,7 @@
 }
 
 .form-errors .form-error-message {
-	color: red;
+	color: var(--jetpack--contact-form--error-color);
```
monsieur-z, line 215: Error message displayed at the top of the form when client side validation is not operating.

projects/packages/forms/src/contact-form/class-contact-form-field.php
```diff
@@ -454,7 +454,7 @@ class='grunion-field-label{$type_class}" . ( $this->is_error() ? ' form-error' :
 				. $extra_attrs_string
 				. '>'
 				. esc_html( $label )
-				. ( $required ? '<span class="grunion-label-required">' . $required_field_text . '</span>' : '' )
+				. ( $required ? '<span class="grunion-label-required" aria-hidden="true">' . $required_field_text . '</span>' : '' )
```
monsieur-z, line 457: Screen readers say "required" for inputs with the `aria-required="true"` attribute. Here, we make the _Required_ label invisible to screen readers so that they don't read it twice.

projects/packages/forms/src/contact-form/class-contact-form.php
```diff
@@ -356,8 +357,7 @@ public static function parse( $attributes, $content ) {
 				$form_classes .= ' wp-block-jetpack-contact-form';
 			}
 
-			$r .= "<form action='" . esc_url( $url ) . "' method='post' class='" . esc_attr( $form_classes ) . "'>\n";
-			$r .= self::get_script_for_form();
```
monsieur-z, line 360: `get_script_for_form` used to disable the submit button once it was clicked. I suppose this was to prevent users from submitting the form multiple times, but it now prevents users from correcting the form.

monsieur-z:
Marking it back as `In Progress` as I've found some layout issues with the `Outlined` style.

projects/packages/forms/src/contact-form/class-contact-form-field.php
```diff
@@ -735,7 +739,14 @@ public function render_checkbox_multiple_field( $id, $label, $value, $class, $re
 				$checkbox_id    = "$id-$checkbox_value";
 
 				$field .= "<p class='contact-form-field'>";
-				$field .= "<input type='checkbox' id='" . esc_attr( $checkbox_id ) . "' name='" . esc_attr( $id ) . "[]' value='" . esc_attr( $checkbox_value ) . "' " . $class . checked( in_array( $option, (array) $value, true ), true, false ) . ' /> ';
+				$field .= "<input
```
monsieur-z, line 744: Reformatting for readability.

projects/packages/forms/src/contact-form/class-contact-form-field.php
```diff
@@ -679,11 +679,13 @@ public function render_radio_field( $id, $label, $value, $class, $required, $req
 	 * @return string HTML
 	 */
 	public function render_checkbox_field( $id, $label, $value, $class, $required, $required_field_text ) {
-		$field  = "<input id='" . esc_attr( $id ) . "' type='checkbox' name='" . esc_attr( $id ) . "' value='" . esc_attr__( 'Yes', 'jetpack-forms' ) . "' " . $class . checked( (bool) $value, true, false ) . ' ' . ( $required ? "required aria-required='true'" : '' ) . "/> \n";
+		$field  = "<div class='contact-form__checkbox-wrap'>";
```
monsieur-z, line 682: Group label and input in a wrapper so that the error message is not added in the same flexbox.

jeherve:
This may be a stupid question, I'm no accessibility expert, but right now VoiceOver doesn't tell what field is incorrect when I submit a form with incorrect data. It only tells me it's incorrect. Shouldn't it tell me what I missed?

monsieur-z:
> This may be a stupid question, I'm no accessibility expert, but right now VoiceOver doesn't tell what field is incorrect when I submit a form with incorrect data.

Not at all!

> It only tells me it's incorrect. Shouldn't it tell me what I missed?

That's the expected behavior (in case it interests someone, I've found this pattern in [this book](https://www.smashingmagazine.com/2016/10/inclusive-design-patterns/)).

Screen reader users know what fields are invalid and how to fix them by tabbing back in the form. Because invalid fields are identified with `aria-invalid` and because a description is linked with `aria-describedby`, each time an input gets focused, screen readers announce something like `Email, invalid, The email address is incorrect`. That's helpful since users don't have to remember what fields are invalid right after submitting the form.

That being said, we could also append the list of invalid fields to the error message read by the screen reader on submitting the form. However, I don't think that'd provide much value. I suspect this is not part of the pattern described in the book because, by the time the screen reader reaches that list, most users will have already tabbed back in the form.

andrii-lysenko (commented):
The code look reasonable to me. Tested it on Simple and Self hosted and it works as described. One small notice: could we clear up the error message on MouseHover or when start typing
Screenshot 2023-11-27 at 17 41 57

And maybe a sanity check from design team (probably just by reviewing video).

P.S. I actually think it's a great change, where we have the control over the validation. In future we could possibly add additional settings to the block, where user would be able to add specific validation rules from pre-defined (e.g. email check) to even some regex. What do you think?

monsieur-z:
> One small notice: could we clear up the error message on MouseHover or when start typing

Sure! It's already logged here: https://github.com/Automattic/jetpack/issues/34281

> P.S. I actually think it's a great change, where we have the control over the validation. In future we could possibly add additional settings to the block, where user would be able to add specific validation rules from pre-defined (e.g. email check) to even some regex. What do you think?

Yes, the native validation would probably limit us at some point anyway.

andrii-lysenko (approved):
@monsieur-z thanks for your replies. This looks good to me and ready to go!

eeeeeevon2023:
Screenshot 2023-11-28 at 11 12 48 AM

Screenshot 2023-11-28 at 11 16 42 AM


Let's get some vertical space between the message and the button. Use the same distance as above the message.

PR #34171, opened 2023-11-16
Site Favorites: Add 'is_user_favorite' property to sites on WordPress.com 

katinthehatsite:
Related to https://github.com/Automattic/dotcom-forge/issues/4539
Related to D129036-code

## Proposed changes:

This PR adds `is_user_favorite` property to sites to WordPress.com

### Other information:

- [ ] Have you written new tests for your changes, if applicable? N/A
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them? 
- [x] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

N/A

## Does this pull request change what data or activity we track or use?

N/A

## Testing instructions:

1. Apply this diff to your sandbox if not merged yet: D129036-code
2. Run `bin/jetpack-downloader test jetpack add/site-favorite-option-to-site` on your sandbox and apply it on top of the diff mentioned above
3. Navigate to https://developer.wordpress.com/docs/api/console/
4. Select `WP.COM API`, `v1.2.`, `GET`, `/me/sites`
5. Select one of the sites from the list and confirm that you can see the `is_user_favorite` property in the list
6. Navigate to https://developer.wordpress.com/docs/api/console/
7. Select `WP REST API`, `wpcom/v2`, `POST`, /sites/$site/site-favorite` (replace $site with the site ID of the site that you would like to make as favorite)
8. Select `WP.COM API`, `v1.2.`, `GET`, `/me/sites`
9. Confirm that the `is_user_favorite` property is set to `true` for the site that you used in step 7

PR #34170, opened 2023-11-16
Jetpack AI: fix site-require-upgrade field to be ready for tiered plans

lhkowalski:
Part of #33824.

## Proposed changes:

* Remove the check for the AI Assistant feature presence on the `site-require-upgrade` field, so even when the site has the feature, it may require an upgrade when it's over the usage limit

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* Let's test if the flag is working correctly for the current plan structure (free and unlimited paid)
* Apply this branch change to your sandbox and sandbox the `public-api` endpoint
* Go to the WordPress.com console to execute requests to `wpcom/v2` `/sites/$wpcom_site/jetpack-ai/ai-assistant-feature`
* Check the `site-require-upgrade` field on the response for some test scenarios:
   * Site with free plan and less than 20 requests -> `site-require-upgrade: false`
   * Site with free plan and more than 20 requests -> `site-require-upgrade: true`
   * Site with the paid plan -> `site-require-upgrade: false`
   * Internal P2 -> `site-require-upgrade: false`
   * Simple site with upgrade -> `site-require-upgrade: false`
* If you have a site with a tiered plan attached to it, test using it and confirm the field will be false when you are under the limit, and true when you are over the limit

PR #34176, opened 2023-11-16
AI Assistant: janitorial code rearrange and site ID retrieval

CGastrell:
Make for cleaner code and less prone to mishaps on blog ID usage.

## Proposed changes:
This PR carries no functional change, its sole purpose is to make the code more clear and less prone to misbehaves.
- use (and comment) specific calls to get the site ID on different contexts
- provide blog ID on methods that accept it as second param
- group code depending on exec context

### Other information:

## Jetpack product discussion
p1700146451177769-slack-C054LN8RNVA

No

## Testing instructions:
Test instructions are just to repeat the latest change tests for this file and expect the same output: #34170 & #34171 

* Let's test if the flag is working correctly for the current plan structure (free and unlimited paid)
* Apply this branch change to your sandbox and sandbox the `public-api` endpoint
* Go to the WordPress.com console to execute requests to `wpcom/v2` `/sites/$wpcom_site/jetpack-ai/ai-assistant-feature`
* Check the `site-require-upgrade` field on the response for some test scenarios:
   * Site with free plan and less than 20 requests -> `site-require-upgrade: false`
   * Site with free plan and more than 20 requests -> `site-require-upgrade: true`
   * Site with the paid plan -> `site-require-upgrade: false`
   * Internal P2 -> `site-require-upgrade: false`
   * Simple site with upgrade -> `site-require-upgrade: false`
* If you have a site with a tiered plan attached to it, test using it and confirm the field will be false when you are under the limit, and true when you are over the limit 

* Let's test it comparing the values before and after the change
* Before applying this change to your sandbox, go to the WordPress.com console (https://developer.wordpress.com/docs/api/console/)
* Execute a request to the `wpcom/v2` `/sites/$wpcom_site/jetpack-ai/ai-assistant-feature` endpoint using your test site and note the results:

Screenshot 2023-11-15 at 17 45 35

* Apply this change to your sandbox and make sure you are sandboxing the `public-api`  
* Go to the WordPress.com console again
* Execute the same request as before, then compare the results:

Screenshot 2023-11-15 at 17 50 21

* The values for `requests-limit` and `requests-count` should be the same as before
* The value for `is-over-limit` may change because now a new rule is applied (the rule from D128826-code):
   * before, the rule was considering if the site was over the free limit of requests (a site with a paid plan could easily be over the 20 requests limit)
   * now, the rule considers if the site is over the limit of requests of the current plan it has (be it free or unlimited, and will also work for tiered plans)
   * confirm the value for `is-over-limit` makes sense by comparing the `requests-count` to the expected limit of the plan active on the site (free or unlimited)
* If possible, test for a Jetpack site and a Simple site

### DIFF
--- a/projects/plugins/jetpack/_inc/lib/class-jetpack-ai-helper.php
+++ b/projects/plugins/jetpack/_inc/lib/class-jetpack-ai-helper.php
@@ -333,7 +333,9 @@ class Jetpack_AI_Helper {
 	 */
 	public static function get_ai_assistance_feature() {
 		if ( defined( 'IS_WPCOM' ) && IS_WPCOM ) {
-			$has_ai_assistant_feature = \wpcom_site_has_feature( 'ai-assistant' );
+			// On WPCOM, we can get the ID from the site.
+			$blog_id                  = get_current_blog_id();
+			$has_ai_assistant_feature = \wpcom_site_has_feature( 'ai-assistant', $blog_id );
 
 			if ( ! class_exists( 'WPCOM\Jetpack_AI\Usage\Helper' ) ) {
 				if ( is_readable( WP_CONTENT_DIR . '/lib/jetpack-ai/usage/helper.php' ) ) {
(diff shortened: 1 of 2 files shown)


lhkowalski (approved):
Tested the API for 5 different sites.

---

Site with free plan and less than 20 requests, before the change:

Screenshot 2023-11-16 at 16 58 32

Site with free plan and less than 20 requests, after the change:

Screenshot 2023-11-16 at 17 02 57

---

Site with free plan and more than 20 requests, before the change:

Screenshot 2023-11-16 at 16 59 26

Site with free plan and more than 20 requests, after the change:

Screenshot 2023-11-16 at 17 03 31

---

Site with the paid plan, before the change:

Screenshot 2023-11-16 at 17 00 01

Site with the paid plan, after the change (notice the `is-over-limit` is different in this case):

Screenshot 2023-11-16 at 17 03 57

---

Internal P2, before the change:

Screenshot 2023-11-16 at 17 00 55

Internal P2, after the change:

Screenshot 2023-11-16 at 17 08 37

---

Simple site with upgrade, before the change:

Screenshot 2023-11-16 at 17 01 43

Simple site with upgrade, after the change:

Screenshot 2023-11-16 at 17 04 46

---

I also tested the assistant on some sites, to make sure it's working as expected, and found no issues.

PR #34177, opened 2023-11-16
My Jetpack: Add direct checkout support for products with quantity-based plans

### LINKED ISSUE #33365: AI Tiered Plans: Ensure correct checkout URL for the upgrade buttons
The checkout will contain information about which tire the user is buying, so we need to make sure it's the correct value on every place we show the upgrade banner.

---

Questions to answer:

- [x] ~~How the DIFM product builds the URL?~~
- [x] Which parameters do we need to include?

---

The URL structure follows this pattern: `https://wordpress.com/checkout/jetpack/jetpack_ai_monthly:-q-100`

- It can be either `jetpack_ai_monthly` or `jetpack_ai_yearly`.
- `q` parameter can be either `100`, `200` or `500`

---

This is required for the MVP.

dhasilva:
Fixes #33365

## Proposed changes:

* Adds a `quantity` prop to the checkout workflow hook, adding the necessary `:-q-` part of the URL
* Adds a `directCheckout` prop to the interstitial so that products that have a required plan but can be further upgraded by usage tiers can go to the checkout page directly.

### Other information:

## Testing instructions:

Due to the transitioning state we are now, it is easier to test locally, as there is a hardcoded function return value that needs to be disabled.

To simulate the current tier, on your sandbox change the value of `$current_licensed_quantity` on `get_current_tier`, in the helper file (`lib/jetpack-ai/usage/helper.php`)

* On `class-jetpack-ai.php`, change the return value of `get_next_usage_tier` to enable the correct next tier
* Go to the interstitial page (admin.php?page=my-jetpack#/add-jetpack-ai)
* Change the current tier to 0, 100 and 200, refreshing the interstitial page
* Check that clicking the upgrade button links to the checkout page with the next tiers `:-q-` URL suffix (keep the network tab open, as that is visible in the first request after redirect)

* On a Jetpack site, check that the other products correctly link to their checkout pages without regressions.

### DIFF
--- a/projects/js-packages/connection/changelog/update-ai-assistant-intertitial-checkout-url
+++ b/projects/js-packages/connection/changelog/update-ai-assistant-intertitial-checkout-url
@@ -1,0 +1,4 @@
+Significance: patch
+Type: added
+
+Connection: Add optional quantity to product checkout workflow hook
--- a/projects/js-packages/connection/hooks/use-product-checkout-workflow/index.jsx
+++ b/projects/js-packages/connection/hooks/use-product-checkout-workflow/index.jsx
@@ -21,15 +21,16 @@ const defaultAdminUrl =
  * Custom hook that performs the needed steps
  * to concrete the checkout workflow.
  *
- * @param {object} props              - The props passed to the hook.
- * @param {string} props.productSlug  - The WordPress product slug.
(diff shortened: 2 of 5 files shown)


CGastrell (approved):
LGTM, works as described! :shipit:

PR #34179, opened 2023-11-17
Add: Jetpack App as a new upload media extension

enejb:
This PR adds a new modal that is found in the Select image drop down for a bunch of image/gallery blocks. 
See

Clicking the "Jetpack Mobile App" button you end up with 
Screenshot 2024-01-10 at 1 15 46 PM

Screenshot 2024-01-10 at 1 16 03 PM

And once you upload media via the Jetpack app. (or in a new tab) 
You can then see the newly added media you see something like 

Screenshot 2024-01-10 at 1 16 29 PM

The image then gets added once the photo is selected. 

Here is a video. 

https://github.com/Automattic/jetpack/assets/115071/ab3ab677-fb2c-466d-bc6f-231c2c06e7c0

## Testing instructions:
 Set `define( 'JETPACK_EXPERIMENTAL_BLOCKS', true ); `

Visit the editor. Notice the new Jetpack App Media modal when you try to add/replace image in a block. Sucha s image, gallery, tile gallery block. 

You can use the beta version for the apps running Jetpack App version 24.0 or later. (not yet released) 
Or you can open up a new tab and upload media under the same user. You should see the modal change and let you enter see the newly uploaded media. 

### Other information:

## Testing instructions:

* Go to '..'


projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-app-media.php
```diff
@@ -0,0 +1,176 @@
+<?php
+/**
+ * REST API endpoint for the External Media.
+ *
+ * @package automattic/jetpack
+ * @since 8.7.0
```
enejb: Todo: Update to the version that we need.

saygunnyc:
Hey @enejb, after reviewing the designs and testing, I wanted to share some updates and suggestions below.

- Let's use the Jetpack App icon and Jetpack Mobile App over the dropdown, so folks who are not familiar with Jetpack App Media Library can quickly understand this option is an upload from Jetpack Mobile App. It also then becomes more in line with how other options are presented. _– Name of app, app logo._

Screenshot 2024-01-09 at 4 19 05 PM

- Tweaked the original modal designs to feature 4px border-radius, shadow and, a close button on the modal like other modals. An example for this could be "Add page" and or "Choose a pattern" existing modals. Tweaked the iPhone image to get rid of the perspective, updated images from OpenVerse, and made some spacing, copy and font size adjustments. Removed the punctuation from titles.

Screenshot 2024-01-09 at 4 21 46 PM


- Used the same sized modal for once photos are uploaded. Since the action needed from the user at this point is to select which photos to be added. I tweaked the title and copy accordingly. Also tweaked the CTA copy to reflect an addition action, rather than selection because the user is already selecting images from uploaded ones within the modal. The next action and the main CTA is adding to post/page. It might be wise to re-use  and align to the existing Dotcom's media library's states for image selection on the modal. Right now we have 3 different types, the Jetpack App, Dotcom Media Library, this modal.

Screenshot 2024-01-09 at 4 26 21 PM


- Media Library selection
Screenshot 2024-01-09 at 4 27 25 PM



Here's the link to the Figma >  L43swkhWMdUsOFx5cSJ9op-fi-0%3A1
Hope this is helpful and let me know if there's anything else that needs attention.

zwarm:
@enejb - I tested the flow using Jetpack Android 24.0 and it works as expected from web to mobile device. Nicely done 👏 I'll leave the official review to folks who are more familiar with the web than I.

zwarm (dismissed):
@enejb All works as expected. See my previous comment. Thanks for wrangling this.

projects/plugins/jetpack/extensions/shared/get-blog-id.js
monsieur-z, line 1: Seems like this file's not needed.

projects/plugins/jetpack/images/app-image-upload.png
monsieur-z, line 1: Would it make sense to generate a JPEG with a white background to save some weight?

projects/plugins/jetpack/extensions/shared/external-media/sources/jetpack-app-media.js
```diff
@@ -0,0 +1,111 @@
...
+		getMedia( `/wpcom/v2/app-media?refresh=true&after=${ currentTime }`, false, false );
+	}, [ getMedia, currentTime ] );
+
+	const onCopy = useCallback(
+		items => {
+			insertMedia( items );
+		},
+		[ insertMedia ]
+	);
+	if ( media.length === 1 && ! multiple ) {
```
monsieur-z: Should this be wrapped in a `useEffect`?

projects/plugins/jetpack/extensions/shared/external-media/editor.scss
```diff
@@ -67,6 +67,110 @@ $grid-size: 8px;
...
+}
+
+.jetpack-external-media-wrapper__jetpack_app_media-title {
+	font-family: Recoleta, "Noto Serif", Georgia, "Times New Roman", Times, serif;
+	font-size: 24px;
+	font-weight: 400;
+	line-height: 40px;
+	letter-spacing: -0.32px;
+	margin: 0 0 14px 0;
+	color:#101517;
```
monsieur-z: Colors could probably be replaced by CSS variables equivalents.

projects/plugins/jetpack/extensions/shared/external-media/editor.scss
```diff
@@ -67,6 +67,110 @@ $grid-size: 8px;
...
+		opacity: 1;
+		transform: translateY(0);
+	}
+}
+
+.jetpack-external-media-wrapper__jetpack_app_media-title {
+	font-family: Recoleta, "Noto Serif", Georgia, "Times New Roman", Times, serif;
+	font-size: 24px;
+	font-weight: 400;
+	line-height: 40px;
```
monsieur-z: Unitless `line-height` is usually safer (40/24=1.67 in that instance). Some browsers could be quirky with this.

projects/plugins/jetpack/extensions/shared/external-media/editor.scss
```diff
@@ -67,6 +67,110 @@ $grid-size: 8px;
...
+		transform: translateY(-100%);
+	}
+	100% {
+		opacity: 1;
+		transform: translateY(0);
+	}
+}
+
+.jetpack-external-media-wrapper__jetpack_app_media-title {
+	font-family: Recoleta, "Noto Serif", Georgia, "Times New Roman", Times, serif;
```
monsieur-z, line 118: This might exist as a variable already.

projects/plugins/jetpack/class.jetpack-gutenberg.php
```diff
@@ -688,6 +688,7 @@ public static function enqueue_block_editor_assets() {
 			$is_current_user_connected = ( new Connection_Manager( 'jetpack' ) )->is_user_connected();
 		}
 
+		wp_enqueue_style( 'recoleta-font', '//s1.wp.com/i/fonts/recoleta/css/400.min.css', array(), Constants::get_constant( 'JETPACK__VERSION' ) );
```
monsieur-z: Recoleta doesn't seem to be used in Jetpack. I'd argue it's overkill to load a new font for a heading.

projects/plugins/jetpack/extensions/shared/external-media/editor.scss
```diff
@@ -67,6 +67,110 @@ $grid-size: 8px;
...
+	letter-spacing: -0.32px;
+	margin: 0 0 14px 0;
+	color:#101517;
+
+}
+.jetpack-external-media-wrapper__jetpack_app_media-description {
+	font-size: 14px;
+	font-weight: 400;
+	line-height: 20px;
+	letter-spacing: -0.15px;
```
monsieur-z: `letter-spacing` rules are probably not noticeable here.

projects/plugins/jetpack/images/app-image-upload.png
enejb: I tried this and even if we got some good image size reduction. The results were were not so great!

projects/plugins/jetpack/extensions/shared/get-blog-id.js
enejb: Fixed

projects/plugins/jetpack/extensions/shared/external-media/sources/jetpack-app-media.js
enejb: Great idea! Thanks

projects/plugins/jetpack/extensions/shared/external-media/editor.scss
enejb: Great suggestion!

projects/plugins/jetpack/extensions/shared/external-media/editor.scss
enejb: Updated to use variables.

projects/plugins/jetpack/extensions/shared/external-media/editor.scss
enejb: :+1:

projects/plugins/jetpack/class.jetpack-gutenberg.php
enejb: I also though about this as well but Recoleta is used on .com as the very common element font when we display banners about the Jetpack app. I will update the code so that we don't load it if we don't need it for now. 
and revisit this decision later.

projects/plugins/jetpack/class.jetpack-gutenberg.php
```diff
@@ -688,6 +688,9 @@ public static function enqueue_block_editor_assets() {
 			$is_current_user_connected = ( new Connection_Manager( 'jetpack' ) )->is_user_connected();
 		}
 
+		if ( self::blocks_variation() === 'experimental' && $is_current_user_connected ) {
+			wp_enqueue_style( 'recoleta-font', '//s1.wp.com/i/fonts/recoleta/css/400.min.css', array(), Constants::get_constant( 'JETPACK__VERSION' ) );
+		}
```
jeherve: Since the font isn't used anywhere else in the editor, could we do without it, and use the existing layout elements and styles that come with the block editor? In general, we attempt to remain as close to the core design elements in the block editor:

https://github.com/Automattic/jetpack/blob/45c2fd4126fe34ceb68800deda9aad764ac50ea9/projects/plugins/jetpack/extensions/README.md#L357

projects/plugins/jetpack/class.jetpack-gutenberg.php
```diff
@@ -706,9 +709,10 @@ public static function enqueue_block_editor_assets() {
 		}
 
 		$initial_state = array(
-			'available_blocks' => self::get_availability(),
-			'modules'          => $modules,
-			'jetpack'          => array(
+			'available_blocks'        => self::get_availability(),
+			'available_media_sources' => array( 'jetpack_app_media' => self::blocks_variation() === 'experimental' ),
```
jeherve: What do you think about only passing the value of `blocks_variation`  from PHP to JS, and introducing a new `getBlocksVariation` utility function in `extensions/shared`? That would keep things a bit more generic, and give us a tool we can use for other extensions in the future. Then, in the external media code, you'd rely on that function to conditionally add `jetpack_app_media` based off the blocks variation of your choice.

projects/plugins/jetpack/extensions/shared/external-media/sources/jetpack-app-media.js
```diff
@@ -0,0 +1,113 @@
...
+							size="100"
+							value={ `https://apps.wordpress.com/get/?campaign=qr-code-media#%2Fmedia%2F${ wpcomBlogId }` }
+						/>
+					</div>
+					<div className="jetpack-external-media-wrapper__jetpack_app_media-instructions">
+						<img
+							src={ imagePath + 'app-image-upload.png' }
+							srcSet={ `${ imagePath + 'app-image-upload.png' } 1x, ${
+								imagePath + 'app-image-upload-2x.png'
+							} 2x` }
```
jeherve: Nitpick, but since we're using template literals already, maybe we can get rid of `+` here and fully rely on template literals?

```suggestion
							src={ `${ imagePath }app-image-upload.png` }
							srcSet={ `${ imagePath }app-image-upload.png 1x, ${ imagePath }app-image-upload-2x.png 2x` }
```

This should also fix the 404 with the original image.

projects/plugins/jetpack/extensions/shared/external-media/sources/jetpack-app-media.js
```diff
@@ -0,0 +1,113 @@
...
+							alt="Screenshot of the Jetpack mobile app with the media upload highlighted."
```
jeherve: This should probably be translated. Another alternative would be to leave the alt tag empty, since this image is decorative.

projects/plugins/jetpack/extensions/shared/external-media/index.js
```diff
@@ -1,6 +1,7 @@
 import { isCurrentUserConnected } from '@automattic/jetpack-shared-extension-utils';
 import { useBlockEditContext } from '@wordpress/block-editor';
 import { addFilter } from '@wordpress/hooks';
+import { SOURCE_JETPACK_APP_MEDIA } from './constants'; // Move this import statement above the import statement for './media-button'
```
jeherve: I don't quite understand the comment here. Can it be removed?

```suggestion
import { SOURCE_JETPACK_APP_MEDIA } from './constants';
```

projects/plugins/jetpack/extensions/shared/external-media/index.js
```diff
@@ -10,7 +11,15 @@ function insertExternalMediaBlocks( settings, name ) {
 	if ( name !== 'core/image' ) {
 		return settings;
 	}
-
+	// Check if the Jetpack App Media source is available.
```
jeherve: Interestingly, before to start looking at the PR, I would have expected to find this logic in `projects/plugins/jetpack/extensions/shared/external-media/sources/index.js`. That's just me though. :)

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-app-media.php
```diff
@@ -0,0 +1,184 @@
+<?php
+/**
+ * REST API endpoint for the media uploaded by the Jetpack app.
+ *
+ * @package automattic/jetpack
+ * @since 13.0.0
```
jeherve:
```suggestion
 * @since $$next-version$$
```

This will be automatically replaced by the correct version on release.

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-app-media.php
```diff
@@ -0,0 +1,184 @@
...
+ */
+
+/**
+ * Media uploaded by the Jetpack app helper API.
+ *
+ * @since 13.0.0
```
jeherve:
```suggestion
 * @since $$next-version$$
```

projects/plugins/jetpack/extensions/shared/icons.js
```diff
@@ -237,6 +237,44 @@ export const JetpackLogo = ( { size = 24, border = 0, className, color = COLOR_J
 	);
 };
 
+export const JetpackMobileAppIcon = () => {
```
jeherve, line 240: My initial reaction with those is that it's 2 more icons we'll have to keep updated. I was wondering if we couldn't just use the existing regular Jetpack icon, but I suppose that wouldn't fit with the designs you had in mind.

projects/plugins/jetpack/extensions/shared/external-media/sources/jetpack-app-media.js
```diff
@@ -0,0 +1,113 @@
...
+import { JetpackAppIcon } from '../../icons';
+import useInterval from '../../use-interval';
+import MediaBrowser from '../media-browser';
+import { MediaSource } from '../media-service/types';
+import withMedia from './with-media';
+
+function JetpackAppMedia( props ) {
+	const { media, insertMedia, isCopying, multiple, getMedia } = props;
+
+	const wpcomBlogId = window?.Jetpack_Editor_Initial_State?.wpcomBlogId || 0;
```
jeherve, line 14: Should we bail entirely if we do not have a WordPress.com blog ID, since we will not be able to do anything with `0`?

jeherve (dismissed):
I gave this a try, and left a few notes below.

It's an interesting flow, not something I would have thought about, but I could see it being useful in my personal blogging! 

******

This doesn't work on WordPress.com simple yet. You'll need to add a new file to call in the new endpoint. You can check D121657-code for an example.

For now, this causes an error when you first call in the modal:

Screenshot 2024-01-12 at 12 34 36

It also makes it impossible to test the full flow with the upload on mobile.

*********

On the screenshot above, I also noticed 3 issues:

1. the image background isn't completely white, so it caused an interesting little red see-through on my small screen.
2. The modal includes a scrollbar on smaller screens. Maybe that's the best experience we can have, but maybe we could also try to resize the contents?
3. The original, non 2x image returned a 404 error. I left a suggestion below on how to reformat the image attributes, that should help with that.

**********

What do you think about making the feature available to experimental AND beta block variations? It would make testing a lot easier since a12s get the Beta blocks variation by default when proxied on WordPress.com Simple, and many a12s use Beta blocks in their development environment.

***********

The images, once uploaded, are not attached to the post where you want to insert them. This is problematic I think, especially on the long run since unattached images can often get lost when exporting / moving content to a different site. Would it be possible to attach the images to the right post in the Jetpack app?

projects/plugins/jetpack/extensions/shared/icons.js
enejb: Yeah the Jetpack icon is not the same as the design.

projects/plugins/jetpack/extensions/shared/external-media/sources/jetpack-app-media.js
enejb: Good call!

projects/plugins/jetpack/class.jetpack-gutenberg.php
enejb: True. This could work! Initially I though it would be good to be very specific as to who will see this but I think we can go ahead but I think limiting it to the block experiments should work just fine.

projects/plugins/jetpack/class.jetpack-gutenberg.php
enejb: @saygunnyc both @jeherve  and @monsieur-z make this point. About not using the Recoleta font. 
Should we try to update the modal to not use it? The modal would look something like this. 
Screenshot 2024-01-12 at 8 57 43 AM

enejb:
Thanks for the review @jeherve! 

>     The modal includes a scrollbar on smaller screens. Maybe that's the best experience we can have, but maybe we could also try to resize the contents?
I think vertical scroll bars are fine. the way the error notification is currently display is not great. I will work on fixing that. 


> What do you think about making the feature available to experimental AND beta block variations? It would make testing a lot easier since a12s get the Beta blocks variation by default when proxied on WordPress.com Simple, and many a12s use Beta blocks in their development environment.

I think that is a good idea. If we can get it into the hands of a11ns earlier the better. 

> The images, once uploaded, are not attached to the post where you want to insert them. This is problematic I think, especially on the long run since unattached images can often get lost when exporting / moving content to a different site. 
Would it be possible to attach the images to the right post in the Jetpack app?

Good point. I would be possible to do but this requires more mobile dev work. We will take this up if we see adoption of this feature. In the mean time I will add the post id to the QR code so that at least it is available early and we don't have to wait for a Jetpack release for this to be there.

projects/plugins/jetpack/extensions/shared/external-media/sources/jetpack-app-media.js
enejb: Good call!

projects/plugins/jetpack/extensions/shared/external-media/index.js
enejb: I wanted this logic to be run only once and and including it in the `external-media/sources/index.js` would make it harder. 🤔

projects/plugins/jetpack/extensions/shared/external-media/sources/jetpack-app-media.js
enejb: because of this `isCurrentUserConnected()` call in https://github.com/Automattic/jetpack/pull/34179/files/8bcaa103d32508623c2bbfb7c190fde1e054c45e#diff-7aaabf6ea20f69c89116cad6960da44b10dca2ebdb72a2560dcf4d76e55e88c6R29 we should always have this be available. I don't think we need to explicitly return early here. 

Since this would be a weird case where the button would show but not the UI.

projects/plugins/jetpack/class.jetpack-gutenberg.php
enejb: I chatted with @saygunnyc about this and ideally all the modals should be more aligned. Most of the Jetpack Mobile App Notices have this font on .com. 
I am not sure when we are going to revisit some of the other modals in Gutenberg but at this point lets ship this. 
Any reason not to ship it? Besides the extra bytes (34kb). Should we include a filter so folks can remove this feature in Jetpack if they are not using it? 
From testing in FF the font only gets downloaded once the font is used. (Modal is opened)  

Would love your thoughts on this @jeherve  and @monsieur-z !

jeherve: If this is meant to align the design with other existing notices on WordPress.com, do we need to add the font here? I expect it would already be added somewhere else on WordPress.com to handle the design of those other notices. Maybe we can rely on that, and fallback to a font family used on Core when that font is not available, to handle the design in core?

enejb: > If this is meant to align the design with other existing notices on WordPress.com, do we need to add the font here?

Yes. Other wise the font is not available in the browser. 

I don't really want to add more complexity to the code and have 2 different UIs just So we don't download the font in the browser for Jetpack users. The browser is already smart enough to figure out if it needs to download the font. At least for me the font only downloads when the modal is opened. See https://jaketrent.com/post/when-font-face-is-downloaded/ 

Or are there other concerns? 

> Aligning with core design elements. 

The UI is as much of a ty into to Jetpack app (marketing) as it is a UI element. 
For example in WordPress.com/discover we use a similar element like this. 
Screenshot 2024-01-18 at 11 02 36 AM

The font that tells you about why to get the app is the same.

jeherve: > Yes. Other wise the font is not available in the browser.

Right, that means that there is currently no other element in the block editor that uses that font. Adding the font now for this one modal will break consistency in the design.

I'm no designer though, so I'd appreciate if someone from a design team would chime in on that :)

zwarm (dismissed):
@enejb - I took the PR out for another spin and everything is working as I expect. 

Upon tapping the Jetpack App icon, the QR Code is displayed. Scanning it with my device launches the app. Following the app's instructions for media permission, I selected images for upload. Once the upload was complete, the post was updated with my chosen selections.

As before, I'll leave the code review for those more familiar with the specifics. 

Thanks

sdixon194 (approved):

Issue #34108, opened 2023-11-14
Sharing block: ensure that we do not load resources on the frontend unless needed


jeherve:
The Sharing Buttons block currently requires the `Sharing_Service` class:
https://github.com/Automattic/jetpack/blob/bb1b509ebab88c7f355d8ec8611b699b5e171c04/projects/plugins/jetpack/extensions/blocks/sharing-buttons/sharing-buttons.php#L15

As a result, the sharing buttons' CSS gets enqueued, even if you haven't set up any buttons yet.
https://github.com/jeherve/jetpack/blob/193e017dbfbfbdd2b78dd5f843357d1d04892c38/projects/plugins/jetpack/modules/sharedaddy/sharing-service.php#L907-L908

As we continue to iterate on the new block, we'll need to ensure we do not needlessly enqueue those styles.

********

Epic: #19694

andrii-lysenko:
Fixed in https://github.com/Automattic/jetpack/pull/34181

PR #34181, opened 2023-11-17
Enhance sharing buttons UI

### LINKED ISSUE #34116: Sharing block: enhance block view
Improve the display of the sharing buttons added via the block.

Here is a quick video about the current implementation, and potential directions:

https://github.com/Automattic/jetpack/assets/426388/b597b909-5833-4b32-87af-8df17f9f561f

If we were to go in the direction of the existing Core Social Icons block, we could have a parent block where we would define style (icon only, icon+text, text), and then child blocks for each one of the sharing buttons the site owner wishes the enable on their site.

********

image

Screenshot 2023-11-14 at 18 14 05
Screenshot 2023-11-14 at 18 13 47

*******

Epic: #19694

andrii-lysenko:
Fixes https://github.com/Automattic/jetpack/issues/34116 and https://github.com/Automattic/jetpack/issues/34108

## Proposed changes:

After discussion in a primary ticket we are moving forward with the implementation similar to Social Links in Gutenberg.
This PR adds new Sharing Button block alongside Sharing Buttons one that is used as a container. Also, added styles to match existing behaviour.

Note: `Official buttons` style is not yet available and we will add it on later stages. This is also still a dummy block with no logic connected and probably some code to be improved in the future.

https://github.com/Automattic/jetpack/assets/60262784/d8bf1503-acc9-40b8-9b2d-9a0c77634390

### Other information:

## Testing instructions:

- Make sure you have beta block enabled. In your environment (example Docker) wp-config.php you should have `define( 'JETPACK_BLOCKS_VARIATION', 'beta' );`
- Add `Sharing Buttons` block to a post and check it's behaviour.


projects/plugins/jetpack/extensions/blocks/sharing-button/sharing-button.php
```diff
@@ -0,0 +1,52 @@
...
+ */
+function render_block( $attr, $content, $block ) {
+	$post_id = $block->context['postId'];
+
+	$style_type  = $block->context['styleType'];
+	$style       = 'style-' . $style_type;
+	$data_shared = 'sharing-' . $attr['service'] . '-' . $post_id . $attr['service'];
+	$link_url    = get_permalink( $post_id ) . '?share=' . $attr['service'] . '&nb=1';
+
+	$content = str_replace( 'url_replaced_in_runtime', $link_url, $content );
```
andrii-lysenko, line 47: I think this is a temporary solution while block is still in `beta`.

monsieur-z (changes requested):
I noticed a couple of rough edges in the UI. The PR still has the `In Progress` label, so don't mind me if these are already known.

1. It'd be nice to have a preview of the block (definitely something that could be done later).
Screenshot 2023-11-24 at 8 59 14 AM

2. The label could be a bit clearer, e.g., _Click the plus icon to add a social network_.
Screenshot 2023-11-24 at 8 59 36 AM

3. The button style can be chosen in both the block's _Settings_ and _Styles_ panels. Should we leave the selection in _Styles_ only?
Screenshot 2023-11-24 at 9 00 45 AM
Screenshot 2023-11-24 at 9 00 48 AM

4. The rendering seems to be the same in the front-end, no matter the style.

5. When adding a button, the _Sharing button_ is proposed but has no icon? Should it even be there?
Screenshot 2023-11-24 at 9 01 34 AM

6. The button labels are rendered in lowercase and with no start padding in the front-end.
Screenshot 2023-11-24 at 9 02 26 AM

7. Is this block independent from the Sharing buttons settings?
Screenshot 2023-11-24 at 9 07 57 AM

andrii-lysenko:
@monsieur-z Thank you for a thorough review. Initially I also thought that we will want to wrap up the UX with this task, however, after discussion with Design, we enhanced a different approach here. So I wanted to roll out initial step, where we basically change the behavior from single block to parent-child. We would still need to do a fare amount of work with UI, including Ofiicial buttons, More button and some adjustmentes to the logic inside as well. So to keep things going, thought we could merge this as a step, but definitely not the final look. What do you think?

andrii-lysenko:
Though, most of them are pretty easy to fix, so I pushed it. As per the last point, yes, these two functionalities are currently disconnected and probably be like that till the very last step, not to provide issues to the users, while we are still working on the block.

projects/plugins/jetpack/extensions/blocks/sharing-buttons/editor.js
```diff
@@ -2,8 +2,31 @@ import { registerJetpackBlockFromMetadata } from '../../shared/register-jetpack-
 import metadata from './block.json';
 import edit from './edit';
 import save from './save';
+import './style.scss';
 
 registerJetpackBlockFromMetadata( metadata, {
 	edit,
 	save,
+	example: {
```
monsieur-z: I believe the example could be added to `block.json` instead, since it doesn't have dynamic values.

projects/plugins/jetpack/extensions/blocks/sharing-button/components/view-button.jsx
```diff
@@ -0,0 +1,24 @@
+import SocialIcon from 'social-logos';
+
+const ViewSocialButton = ( { service, url } ) => {
+	return (
+		<a
```
monsieur-z: From an accessibility perspective, we could make screen readers read something like `Share to Facebook` rather than `Facebook`. This would provide more context to screen reader users and avoid confusion if there are multiple link labelled `Facebook` on the page.

We could do that by adding an `aria-label` attribute on the `a` and, I believe, hiding the `span` with `aria-hidden="true"`.

projects/plugins/jetpack/extensions/blocks/sharing-button/style.scss
```diff
@@ -0,0 +1,82 @@
...
+.style-icon .jetpack-sharing-button__service-label {
+	display: none;
+}
+
+.jetpack-sharing-buttons__service-label.style-icon {
+	display: none;
+}
+
+.jetpack-sharing-button__service-label {
+	margin-left: 5px;
```
monsieur-z: It's probably safer to use `margin-inline-start` for internationalization reasons.

monsieur-z (dismissed):
Thanks for the additional context @andrii-lysenko. I left some comments but they can definitely be tackled later one.

andrii-lysenko:
@monsieur-z Thanks for highlighting these. I think it is worth having these added so I pushed an additional commit.

chrisbliss18 (approved):
I see that you made the changes that @monsieur-z requested. Everything looks good. Approved!

keoshi:
Commenting here after the PR merge but hope this feedback is useful. Love the suggestions that @monsieur-z left above, btw!

~1. Adding the block the first time asked me to activate the sharing feature. This is great, but it also added the old sharing buttons by default.~ Disregard this one, I was testing RC and not Bleeding Edge.

2. Agree with @monsieur-z that the placeholder copy could be `Click the plus icon to add a social network.`

image

3. The empty + unselected state is perfectly fine, but I wonder if we should try to do something like the social icons do. E.g.: add a visual flare to denote the block hasn't been touched yet.

image

4. The block options — moving icons, justification, alignment, orientation, styles — worked perfectly. Very neat that people will soon have much more flexibility to play around with these buttons.

image

5. The child blocks description is the same across the board but refers to the parent block. Perhaps we should replicate what the Social block is doing.
  - Suggested copy: `Display a button that allows people to share your content to a social media network.`

image

image


6. We could probably approximate the style of the new buttons with the older ones, so that nothing really changes in their site design. In particular icon size and font size.

image

7. "Official Buttons" didn't work for me; it shows the same styling as Icon & Text, but I'm guessing this hasn't been built out yet.

image

8. Finally, a sidenote: I'm seeing `BETA` tags in the child blocks as well. Would be great to get rid of these and just have one in the parent block so the editing interface shines through while we test the UI.

image

This is looking very, very promising @andrii-lysenko — awesome work so far! Thank you.

PR #34182, opened 2023-11-17
Jetpack Social| Fix scheduled posts UI

manzoorwanijk:
This is the base branch for fixing scheduled posts UI for Jetpack Social

## Proposed changes:

* #34183
* #34206 
* #34323
* #34392

### Other information:

## Testing instructions:

* This is a base branch
* Please test the individual brances


gmjuhasz (commented):
I found a minor issue. If you have 0 schedule shares, the bar looks wrong, it's not filling the space out. If I schedule at least 1 post it looks good.

CleanShot 2023-12-07 at 12 51 16 png

manzoorwanijk:
> I found a minor issue. If you have 0 schedule shares, the bar looks wrong, it's not filling the space out. If I schedule at least 1 post it looks good.

Thank you for catching that. I have pushed a fix

Screenshot 2023-12-07 at 5 55 17 PM

gmjuhasz (dismissed):
Thanks for fixing that. I tested the scenarios and all seemed to work for me 👍

projects/js-packages/components/components/record-meter-bar/style.scss
```diff
@@ -25,7 +25,7 @@
 			width: 1rem;
 			height: 1rem;
 			display: inline-block;
-			margin-inline-end: 0.4em;
+			margin-inline-end: 0.25em;
```
pablinos: I know this is only a minor change, but have tested it's ok for the other uses of this component?

projects/js-packages/publicize-components/src/components/form/enabled-connections-notice.tsx
```diff
@@ -0,0 +1,30 @@
...
+						'This post will be shared to %d network.',
+						'This post will be shared to %d networks.',
+						enabledConnections.length,
+						'jetpack'
+					),
+					enabledConnections.length
+				) }
+			</p>
+		</PanelRow>
+	) : null;
```
pablinos, line 31: It might be nice to add something for 0 connections. I'm not sure what though! "This post will not be shared unless you enable a connection" or "Enable a connection to share this post".

projects/js-packages/publicize-components/src/components/form/enabled-connections-notice.tsx
```diff
@@ -0,0 +1,30 @@
...
+import useSocialMediaConnections from '../../hooks/use-social-media-connections';
+import styles from './styles.module.scss';
+
+/**
+ * Displays enabled connections text.
+ *
+ * @returns {import('react').ReactElement} Enabled connections text.
+ */
+export function EnabledConnectionsNotice() {
+	const { enabledConnections } = useSocialMediaConnections();
```
pablinos, line 13: There can be lots of other reasons that connections are disabled. If they're broken, if we haven't got valid media etc. This won't take those into account, so it will end up saying that it's going to share to connections when it's not.

projects/js-packages/publicize-components/src/components/form/enabled-connections-notice.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+	return enabledConnections.length ? (
+		<PanelRow>
+			<p className={ styles[ 'enabled-connections-notice' ] }>
+				{ sprintf(
+					/* translators: %d: number of networks */
+					_n(
+						'This post will be shared to %d network.',
```
pablinos: Should we call them connections? They might have two for one network.

projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -92,26 +98,7 @@ export default function PublicizeForm( {
...
-				window.location.href = ev.target.href;
-			}
-			if ( target ) {
-				ev.preventDefault();
-				window.open( ev.target.href, target, 'noreferrer' );
-			}
-		},
-		[ autosave, isEditedPostDirty ]
-	);
+	const outOfConnections = showShareLimits && numberOfSharesRemaining <= enabledConnections.length;
```
pablinos, line 101: Similarly to my comment above, `enabledConnections` is not that accurate. Perhaps we should fix that in a follow-up as it's been this way for a while.

projects/js-packages/publicize-components/src/components/share-limits-bar/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+				label: __( 'used', 'jetpack' ),
+			},
+			scheduledCount > 0 && {
+				count: noticeType === 'default' ? scheduledCount : scheduledCount + usedCount,
+				backgroundColor: colorsForScheduled[ noticeType ],
+				label: noticeType === 'default' ? scheduledMessage : usedAndScheduledMessage,
+			},
+			{
+				count: remainingCount,
+				backgroundColor: 'var(--jp-gray-off)',
```
pablinos, line 55: Ah, I see why this isn't simple to have a different colour for the admin page and the editor. Could we make this point to another variable and then set it in the CSS?

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
```diff
@@ -0,0 +1,103 @@
...
+				message = messages.exceeded;
+			} else if ( totalSharesCount === shareLimit ) {
+				status = 'full';
+				noticeType = 'error';
+				message = messages.full;
+			}
+			// If they have used 80% of their limit, they are approaching it
+			else if ( totalSharesCount >= shareLimit * 0.8 ) {
+				status = 'approaching';
+				noticeType = 'warning';
```
pablinos, line 94: As I was talking about yesterday, now the bar is static, I think it might be worth making it go red when the share count is >= 90% of the limit. A bit like the warning light on a fuel gauge!

projects/js-packages/publicize-components/src/social-store/types.ts
```diff
@@ -0,0 +1,43 @@
...
+	sharesData: SharesData;
+	// on post editor
+	hasPaidPlan?: boolean;
+	// on Jetack Social admin page
+	jetpackSettings?: JetpackSettings;
+};
+
+declare global {
+	interface Window {
+		jetpackSocialInitialState?: SocialStoreState;
```
pablinos, line 41: I think these types are only like this in the admin page, or maybe the editor? How much of an issue is that?

pablinos (commented):
I've mainly just scanned through the code again, but a lot of it has been reviewed as part of other PRs.

We should check a few scenarios, mainly with and without a plan on both the Jetpack and Social plugins. We should also test this on WPCOM. We've been seeing some conflicts with different versions of plugins active on sites too, so we should probably check this in that scenario too.

projects/js-packages/components/components/record-meter-bar/style.scss
manzoorwanijk: Yes, I agree. https://github.com/Automattic/jetpack/pull/34392/commits/d080a694a18ca96dba6f3cb1df9730b223a6ef21 should not have been made here.

projects/js-packages/publicize-components/src/components/form/enabled-connections-notice.tsx
manzoorwanijk: I think we have this kind of message duplicated above the connections list.

| When a connection is ON | When no connection is ON |
|--------|--------|
| Screenshot 2023-12-11 at 11 17 26 AM | Screenshot 2023-12-11 at 11 17 34 AM |

Should we get rid of one or refine it?

projects/js-packages/publicize-components/src/components/form/enabled-connections-notice.tsx
manzoorwanijk: May be there should be a deterministic way to know that?

projects/js-packages/publicize-components/src/components/form/enabled-connections-notice.tsx
manzoorwanijk: I don't have a strong preference here.

projects/js-packages/publicize-components/src/components/share-limits-bar/index.tsx
manzoorwanijk: I am sorry but I didn't understand.

The color used in the designs is already defined as `--jp-gray-0` but that is also used as the background color for the admin page top section, thus rendering the bar in that section with `--jp-gray-0` makes it invisible.

projects/js-packages/publicize-components/src/social-store/types.ts
manzoorwanijk: That is why it's optional - `?:`

projects/js-packages/components/components/record-meter-bar/style.scss
manzoorwanijk: Restored.

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
manzoorwanijk: Done ✅ 

Screenshot 2023-12-11 at 11 33 14 AM

projects/js-packages/publicize-components/src/components/form/enabled-connections-notice.tsx
pablinos: Yeah perhaps we leave it how it is then, at least for now. I'm worried about jank as people disable/enable a connection. The majority of people only have one connection, so these messages will be disappearing/reappearing and changing above the connections, quite often. The one above the connections means the hit area for the toggle moves too. It's not a great UX. Let's sort it out in a follow-up

projects/js-packages/publicize-components/src/components/form/enabled-connections-notice.tsx
pablinos: There is, but the problem is, that it's not determined [in the right place](https://github.com/Automattic/jetpack/blob/10b15dde473e30fc22b17cf3f06442d0309cc466/projects/js-packages/publicize-components/src/components/form/index.js#L299-L306). This disabled logic needs to be in the store, or at least in that hook.

One of the problems is that 'enabled' is not the opposite of 'disabled' in this context. An 'enabled' connection is one that we will try to share to if the connection is not 'disabled', but there are a few reasons a connection could be disabled: out of shares, a broken connection and invalid media.

We need some better terms. The toggleable flag I think was probably created for this purpose, but isn't really being used.

Maybe we're trying to fix too much in this PR though, and we should follow up as soon as we can!

projects/js-packages/publicize-components/src/components/form/enabled-connections-notice.tsx
pablinos: I think we're using connections elsewhere, let's go with that.

projects/js-packages/publicize-components/src/components/share-limits-bar/index.tsx
pablinos: Yeah, so my thought was, how can we use a different colour on the admin page, or make it stand out in some other way. In the editor, with one scheduled post, the colour is pretty lost with this darker grey.

projects/js-packages/publicize-components/src/social-store/types.ts
pablinos: Sure, but that means we're defining the types, but only in one context. I suppose that's better than nothing!

projects/js-packages/publicize-components/src/components/form/enabled-connections-notice.tsx
manzoorwanijk: OK, I am going to create a task for it.

projects/js-packages/publicize-components/src/components/share-limits-bar/index.tsx
manzoorwanijk: Designers are the best people for that.

projects/js-packages/publicize-components/src/components/form/enabled-connections-notice.tsx
manzoorwanijk: Done.

projects/js-packages/publicize-components/src/components/form/enabled-connections-notice.tsx
manzoorwanijk: I have created a task for this on our board.

projects/js-packages/publicize-components/src/components/share-limits-bar/index.tsx
pablinos: Ok, then let's create another task for this so it doesn't get lost. Currently it doesn't follow the design, but there are reasons for that, and we should solve the problem.

manzoorwanijk: Done ✅

spsiddarthan (approved):
Tests well!

PR #34183, opened 2023-11-17
Jetpack Social | Warn the user that their scheduled posts won't get shared if they will get over limits.

manzoorwanijk:
Part of #34182, This PR fixes the messaging regarding scheduled posts for Jetpack Social to warn the user that their scheduled posts won't get shared if they will get over limits.

This PR is a result of a joint/paring session between me and. @gmjuhasz 

## Proposed changes:

* Add a visual meter bar on post edit page to display the current limit status
* Update the messaging around limits especially for scheduled posts

### Other information:

## Testing instructions:

* Boot up this PR
* Run `jetpack install plugins/social`
* Run `jetpack watch plugins/social`
* Goto post edit and open Jetpack Social sidebar
* Manipulate the selectors in `ackages/publicize-components/src/social-store/selectors/shares-data.ts`
* Test different scenarios to confirm the messaging works as expected

Jetpack Social admin page 
Screenshot 2023-11-17 at 5 00 12 PM

Different scenarios on post editor

Screenshot 2023-11-17 at 4 46 54 PM
Screenshot 2023-11-17 at 4 46 15 PM
Screenshot 2023-11-17 at 4 47 40 PM
Screenshot 2023-11-17 at 4 48 28 PM
Screenshot 2023-11-17 at 4 48 56 PM
Screenshot 2023-11-17 at 4 49 34 PM
Screenshot 2023-11-17 at 4 50 14 PM


projects/js-packages/publicize-components/src/components/share-limits-bar/index.tsx
```diff
@@ -0,0 +1,53 @@
+import { RecordMeterBar, Text } from '@automattic/jetpack-components';
+import { __ } from '@wordpress/i18n';
+import { useMemo } from 'react';
```
pablinos: As we're in the editor, I think this should be imported from `@wordpress/elements`

projects/js-packages/publicize-components/src/components/form/share-count-info.tsx
```diff
@@ -0,0 +1,36 @@
+import { ThemeProvider } from '@automattic/jetpack-components';
+import { useSelect } from '@wordpress/data';
+import { store as socialStore } from '../../social-store';
+import { ShareLimitsBar } from '../share-limits-bar';
+import styles from './styles.module.scss';
+
+export const ShareCountInfo: React.FC = () => {
```
pablinos, line 7: I feel like we need some linting rules around types. `React.FC` is not imported here. Should it be?

projects/js-packages/publicize-components/src/components/form/share-count-notice.tsx
```diff
@@ -0,0 +1,60 @@
...
+	if ( ! showShareLimits ) {
+		return null;
+	}
+
+	return (
+		<Notice type={ noticeType }>
+			{ message }
+			<br />
+			{ createInterpolateElement(
+				__( '<upgradeLink>Upgrade now</upgradeLink> to remove the limit.', 'jetpack' ),
```
pablinos: I think we currently say 'to share more' rather than remove the limit. They are essentially removing the limit, but at some point we're supposed be putting a limit on the basic plan.

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * @param {number} shareLimit - Share limit
+ *
+ * @returns {ReturnType<typeof getMessages>} Share limits messages
+ */
+export function getMessages( shareLimit: number ) {
+	return {
+		default: sprintf(
+			// translators: %1$d is the number of shares allowed in 30 days.
+			__( 'Share limit for 30 days: %1$d.', 'jetpack' ),
```
pablinos, line 22: I agree that we need to show the share limit somewhere, but this isn't quite right for me. Is there a way we could put a scale on the bar instead?

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
```diff
@@ -0,0 +1,101 @@
...
+		),
+		limitExceeded: __(
+			'You have exceeded your share limit. Your posts will not longer be shared.',
+			'jetpack'
+		),
+		scheduled: __(
+			'Your scheduled posts will not get shared after you reach the sharing limit.',
+			'jetpack'
+		),
+		approachingLimit: __( 'You are approaching your share limit.', 'jetpack' ),
```
pablinos: I'm trying to work out if it should be 'the share limit' or 'your share limit', maybe 'your' is better if we're going to have different limits on different plans.

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
```diff
@@ -0,0 +1,101 @@
...
+			// translators: %1$d is the number of shares allowed in 30 days.
+			__( 'Share limit for 30 days: %1$d.', 'jetpack' ),
+			shareLimit
+		),
+		limitExceeded: __(
+			'You have exceeded your share limit. Your posts will not longer be shared.',
+			'jetpack'
+		),
+		scheduled: __(
+			'Your scheduled posts will not get shared after you reach the sharing limit.',
```
pablinos: If this is for the situation that where they have posts that will not go out, then I think we should say that. 'You have scheduled posts that will not be shared because you will exceed the share limit' or something along those lines.

I've just thought that by the time we go to share them, we'll be looking at a different 30 day window, so it could be that they do have enough shares. @spsiddarthan is that right? If so, perhaps this softer message is ok, unless we can work out warning them with dates or something, but that feels far too complex. I feel like if something is scheduled to go out and it's not going to, then we need a big red shouty warning!

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
```diff
@@ -0,0 +1,101 @@
...
+	}, [] );
+
+	let noticeType: ShareLimits[ 'noticeType' ] = 'default';
+	const messages = getMessages( shareLimit );
+
+	let message = messages.default;
+
+	switch ( limitStatus ) {
+		case 'crossed':
+			noticeType = usedSharesCount >= shareLimit ? 'error' : 'warning';
```
pablinos: I don't think `limitStatus` could be `crossed` without them being over the share limit, but I'm not sure of the difference between `totalSharesCount` and `usedSharesCount`. If a connection has been toggled off and they're now below the limit, then `limitStatus` will be recalculated too, and we won't set this to 'warning'.

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
```diff
@@ -0,0 +1,101 @@
...
+			shareLimit: store.getShareLimit(),
+		};
+	}, [] );
+
+	let noticeType: ShareLimits[ 'noticeType' ] = 'default';
+	const messages = getMessages( shareLimit );
+
+	let message = messages.default;
+
+	switch ( limitStatus ) {
```
pablinos: Could this `switch` be added to the first `useSelect`? We get the status and then only use it for the switch statetment (as well as returning it of course).

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
```diff
@@ -0,0 +1,101 @@
...
+
+		// If they have used 80% of their limit, they are close
+		if ( totalSharesCount >= shareLimit * 0.8 ) {
+			return 'close';
+		}
+
+		return 'none';
+	}, [] );
+
+	const { scheduledShares, usedSharesCount, shareLimit } = useSelect( select => {
```
pablinos: This is also selecting things from the `socialStore`. It might be clearer this way, but perhaps we should combine the `useSelect`s?

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
```diff
@@ -0,0 +1,101 @@
...
+ */
+export function useShareLimits(): ShareLimits {
+	const limitStatus = useSelect( select => {
+		const store = select( socialStore );
+
+		const shareLimit = store.getShareLimit();
+		const totalSharesCount = store.getTotalSharesCount( { includeActiveConnections: true } );
+
+		if ( totalSharesCount > shareLimit ) {
+			return 'crossed';
```
pablinos: Should we be consistent with our naming, and call this `exceeded`? Or rename the uses of exceeded?

Similarly `close` could be `approaching`. I'm not sure what other term we're using for `full`.

projects/packages/publicize/src/class-publicize-base.php
```diff
@@ -1837,33 +1837,6 @@ public function get_supported_additional_connections() {
...
-	}
-
-	/**
-	 * Check if Nextdoor connection is enabled.
-	 *
-	 * @return bool
-	 */
-	public function has_nextdoor_connection_feature() {
-		return Current_Plan::supports( 'social-nextdoor-connection' );
-	}
```
pablinos: Is this part of this change or the larger store refactor? If it's the latter, perhaps we should break it out into a separate change

pablinos: I've also realised that we might have to be careful if the WPCOM side gets out of sync with the Jetpack plugin versions. We don't want to break the API

projects/js-packages/publicize-components/src/components/share-limits-bar/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+};
+
+export const ShareLimitsBar: React.FC< ShareLimitsBarProps > = ( {
+	currentCount,
+	maxCount,
+	scheduledCount,
+	activeConnectionsCount,
+	text,
+	className,
+} ) => {
```
pablinos: In fact, could we do:

```suggestion
export const ShareLimitsBar = ( {
	currentCount,
	maxCount,
	scheduledCount,
	activeConnectionsCount,
	text,
	className,
} : ShareLimitsBarProps ) => {
```
And have the return type inferred?

These are probably stylistic choices, which is why we could do with some lint rules.

pablinos (commented):
This is partly a problem because of the bug that allows me to have too many connections enabled, but I get the wrong message when `active + used > share limit`

image

And if I disable a connection then I also get the wrong message, because I haven't exceeded the limit yet

image

But, before we work on the messages too much, it might be best to wait, as @jeffgolenski is looking at tidying up the UI for us and bringing it close to the AI Assistant, which has a similar meter in the sidebar.

pablinos (changes requested):
Something about this change as it stands means that the share limit information is stitll being shown despite the site having a paid plan.

Hopefully it should be a straightforward fix.

CleanShot 2023-11-21 at 20 43 06

projects/packages/publicize/src/class-publicize-base.php
manzoorwanijk: I have reverted theses changes and I will create a new PR for it. Thanks

projects/js-packages/publicize-components/src/components/form/share-count-info.tsx
manzoorwanijk: `React` namespace is already in global scope when using [`react-jsx`](https://www.typescriptlang.org/tsconfig#jsx)

projects/js-packages/publicize-components/src/components/share-limits-bar/index.tsx
manzoorwanijk: Done. Thanks

projects/js-packages/publicize-components/src/components/form/share-count-notice.tsx
manzoorwanijk: Done. Thanks

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
manzoorwanijk: I don't have any strong opinions about it. It was what me and Geri came up with.

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
manzoorwanijk: I have updated the message for now.

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
manzoorwanijk: `totalSharesCount` includes both used and scheduled shares, while as `usedSharesCount` includes only the already shared count excluding the scheduled shares.

projects/js-packages/publicize-components/src/components/share-limits-bar/index.tsx
manzoorwanijk: Done. Thanks. Yeah, lint rule should help.

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
manzoorwanijk: > If a connection has been toggled off and they're now below the limit, then limitStatus will be recalculated too, and we won't set this to 'warning'.

If `usedSharesCount >= shareLimit`, that means they have already exceeded the limit without scheduled posts or connections.

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
manzoorwanijk: The intention was to break the logic down to keep it simple and clean.

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
manzoorwanijk: Again, the intention was to break the logic down to keep it simple and clean by avoiding too many branches of `if/else` at one place.

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
manzoorwanijk: I have done it. Thanks

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
manzoorwanijk: I have done it. Thanks

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
manzoorwanijk: Updated

manzoorwanijk:
> Something about this change as it stands means that the share limit information is stitll being shown despite the site having a paid plan.

This was actually due to store state initialization not working properly because we have the initial data scattered at different places in different shapes.

TIL 💡 that redux won't add the property to the state from the initial state https://github.com/Automattic/jetpack/pull/34183/commits/3968f51c4be81d832ef513c8ecc071c6f6dd1fb6 added for that property.

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
manzoorwanijk: > Is there a way we could put a scale on the bar instead?

May be we will need some design input on that

projects/packages/publicize/src/class-publicize-base.php
manzoorwanijk: I have created #34241 for this change.

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
pablinos: Yes, we can look at it with the designs.

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
pablinos: Got it!`usedSharesCount` is a subset of `totalSharesCount`. I think I was seeing it as the other way around.

projects/js-packages/publicize-components/src/components/form/index.js
```diff
@@ -72,7 +72,12 @@ export default function PublicizeForm( {
 	const hasInstagramConnection = connections.some(
 		connection => connection.service_name === 'instagram-business'
 	);
-
+	const { showShareLimits, numberOfSharesRemaining } = useSelect( select => {
+		return {
+			showShareLimits: select( socialStore ).showShareLimits(),
+			numberOfSharesRemaining: select( socialStore ).numberOfSharesRemaining(),
+		};
+	} );
```
pablinos, line 80: I've just noticed this could be:

```suggestion
	const { showShareLimits, numberOfSharesRemaining } = useSelect( select => ( {
			showShareLimits: select( socialStore ).showShareLimits(),
			numberOfSharesRemaining: select( socialStore ).numberOfSharesRemaining(),
	} ) );
```

Another stylistic change though. Take it or leave it. We can apply a lint rule if we decide we prefer one over the other.

projects/js-packages/publicize-components/src/components/form/share-count-info.tsx
pablinos: Cool. Do we need it at all though?

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
```diff
@@ -0,0 +1,120 @@
...
+			 */
+			// Case 1
+			if ( usedSharesCount >= shareLimit ) {
+				noticeType = 'error';
+			}
+			// Case 2a
+			else if ( scheduledShares > 0 ) {
+				noticeType = 'warning';
+				message = messages.scheduled;
+			}
```
pablinos, line 76: This is all pretty complex isn't it? I can see what the switch statement was doing now, but I'm not sure one is much clearer than the other.

I think we could only be greater than the share limit with scheduled shares, and we would want to show an error at that point I think. 

The noticeType is usually the same for each message. `error => exceeded, warning => scheduled` etc. Perhaps the noticeType should be part of the messages data structure?

Having said this, I think perhaps we should wait before making further changes here until we change the design, we may end up showing completely different notices.

pablinos (approved):
I think this is good to merge and we can update the design in a separate change. My only other comments are stylistic thoughts, which again we can do in a separate change if we want to.

projects/js-packages/publicize-components/src/components/form/share-count-info.tsx
manzoorwanijk: Hmm, I had initially added props to that component and then decided to use hooks instead and it remained there. It doesn't hurt though.

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
manzoorwanijk: Yeah, it's complicated

jeffgolenski:
> I think this is good to merge and we can update the design in a separate change. My only other comments are stylistic thoughts, which again we can do in a separate change if we want to.

Sounds good, thanks @pablinos & @manzoorwanijk ! Quick question about a specific use case I see above in a screenshot:

image

If a user has 0 active connections, then technically the scheduled shares can't happen because they have no networks to be sent to, right? Should we instead display a different notice that says "Reconnect your network connections to schedule your posts to be shared." or something of that sort?

manzoorwanijk:
> If a user has 0 active connections, then technically the scheduled shares can't happen because they have no networks to be sent to, right?

Sorry about the confusion @jeffgolenski. By "Active connections", we mean the connections that the user has enabled above there. May be we should name it as "Selected connections" or something similar to avoid the confusion?

PR #34186, opened 2023-11-17
Sync: silence the call to `gzinflate`

bor0:
## Proposed changes:

Silence the call to `gzinflate`:

- Can't always assume that the input is a valid compressed string
- There are [no functions](https://www.php.net/manual/en/function.gzinflate.php) that we can use to check if the data is valid
- WordPress core also uses `@gzinflate`.

Should fix a few warnings :)

### Other information:

No

## Testing instructions:
N/A


projects/packages/sync/src/class-json-deflate-array-codec.php
```diff
@@ -41,7 +41,7 @@ public function encode( $object ) {
 	 * @return array|mixed|object
 	 */
 	public function decode( $input ) {
-		return $this->json_unserialize( gzinflate( base64_decode( $input ) ) ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.obfuscation_base64_decode
+		return $this->json_unserialize( @gzinflate( base64_decode( $input ) ) ); // phpcs:ignore
```
anomiex: Let's not use generic `phpcs:ignore`. Better to be clear about what exactly we're ignoring.

Also, it seems like it might make more sense to store the result of `gzinflate` in a variable and check for `false` so we could skip trying to `json_unserialize` that.

bor0: Thank you! I agree. Addressed these in 436ac0e747.

fgiannar (dismissed):
LGTM 👍 

Apologies for the late review. Give me a ping to re-approve after resolving conflicts if needed

fgiannar (dismissed):

fgiannar (approved):
LGTM

PR #34189, opened 2023-11-17
Fix VideoPress and Memberships privacy setting override

pgk:
Fixes #

## Proposed changes:

* Only allow memberships to override privacy access if the post access level is set to an access level that requires subscription.

### Other information:

## Testing instructions:

* Create a post and add a private video.
* Do not set membership access. (i.e. the default which is 'everybody')
* Check the video post logged in to the site. It should work.
* Check the video post logged out. It should not work.


CGastrell:
LGTM, tests worked just fine. Waiting for a second look.

dhasilva (approved):
LGTM :rocket:

PR #34126, opened 2023-11-14
General: update our PHP version requirements to PHP 7.0

jeherve:
Fixes #31714

## Proposed changes:

Now that WordPress 6.4 is available, our plugins will be able to drop support for WordPress 6.2 and require WordPress 6.3 (to happen in a follow-up PR, #34127).

WordPress 6.3 introduced a new PHP version requirement: PHP 7.0. We can now match that requirement in all our tools.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

* #32865

## Does this pull request change what data or activity we track or use?

* No

## Testing instructions:

* Check if CI is happy
* Check that the changes to the codesniffer to remove one set of rules makes sense.

PR #34192, opened 2023-11-17
Declare PHP version requirement in all Composer packages

anomiex:
## Proposed changes:

This makes for a better experience for third parties using the packages, as they can see on Packagist which PHP versions the package is supposed to support.

Also this is a good excuse to make a major version bump for all the packages following #34126 which removed testing with PHP 5.6 and allows use of 7.0+ features going forward.

And since we're making systematic use of declared PHP versions now, let's lint for that and make use of it to skip phpunit tests where appropriate.

### Other information:

## Jetpack product discussion

None really, although https://github.com/Automattic/jetpack/pull/34126#discussion_r1394352324 is somewhat related

No

## Testing instructions:

* Is CI happy?
  * [x] Codesniffer tests skipped in PHP 7.0 with the new check for that. https://github.com/Automattic/jetpack/actions/runs/6907991966/job/18796345466?pr=34192#step:9:1639
  * [x] Codesniffer tests still ran in PHP 7.4 as expected. https://github.com/Automattic/jetpack/actions/runs/6907991966/job/18796346607?pr=34192#step:9:1781
* Try using `jetpack generate` to generate a package, it should include `.require.php` in composer.json.
* Try removing the `.require.php` from a package and see if `.github/files/lint-project-structure.sh` catches it with an appropriate message.

### DIFF
--- a/.github/files/lint-project-structure.sh
+++ b/.github/files/lint-project-structure.sh
@@ -7,6 +7,7 @@ cd $(dirname "${BASH_SOURCE[0]}")/../..
 BASE=$PWD
 . "$BASE/tools/includes/check-osx-bash-version.sh"
 . "$BASE/tools/includes/chalk-lite.sh"
+. "$BASE/.github/versions.sh"
 
 if [[ -n "$CI" ]]; then
 	function debug {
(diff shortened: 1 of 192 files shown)


jeherve (approved):
🚢

PR #34196, opened 2023-11-17
My Jetpack: Add Jetpack AI prices by tier to the interstitial page

dhasilva:
Part of #33358

## Proposed changes:

* Maps each Jetpack AI tier to its yearly price using the product's `price_tier_list`
* Shows the price in the interstitial component for already-activated products that are quantity-based 

### Other information:

## Testing instructions:

On JN:
* Apply this PR to a JN site and just check that it acts as normal, with no obvious regression in the checkout flow

Test locally to be able to disable the hard return of the unlimited plan:
* Look for the `get_next_usage_tier` on `class-jetpack-ai.php` and change its commented return value
* Sandbox the public API
* In your sandbox, change the `$current_licensed_quantity` value in the `get_current_tier` function of the Jetpack AI usage Helper class to mock different tiers 
* Go to the interstitial page (?page=my-jetpack#/add-jetpack-ai)
* Check that *a* price is displayed for all tiers. At this point it should be the same price
* On your sandbox's `0-sandbox.php` file, enable the store sandbox flag with `define( 'USE_STORE_SANDBOX', true );`
* Here cache will prevent the prices from loading correctly, so you may need to force a `return true` on `class-wpcom-products.php`' `is_cache_old` function.
* Reload the interstitial page
* Check that each tier shows a different price that matches the one on the prices spreadsheet (p1700251365617209/1700165215.316789-slack-C054LN8RNVA)

### DIFF
--- a/projects/packages/my-jetpack/_inc/components/product-detail-card/index.jsx
+++ b/projects/packages/my-jetpack/_inc/components/product-detail-card/index.jsx
@@ -120,8 +120,11 @@ const ProductDetailCard = ( {
 	 * Product needs purchase when:
 	 * - it's not free
 	 * - it does not have a required plan
+	 *
+	 * Or when:
+	 * - it's a quantity-based product
 	 */
-	const needsPurchase = ! isFree && ! hasRequiredPlan;
+	const needsPurchase = ( ! isFree && ! hasRequiredPlan ) || quantity != null;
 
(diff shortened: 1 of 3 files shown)


CGastrell (approved):
LGTM, we need to figure out how to connect to a _single source of truth_ kind of thing.

PR #34197, opened 2023-11-17
Update/refactor data request for my jetpack cards

jboland88:
## Proposed changes:

* This diff changes the way the backup card collects data to display. It removes a call to a remote WPCOM endpoint in favor of a local API call and makes an additional call more direct.
* Overall, the goal of this PR is to remove the dependency on the `sites/%d/jetpack-product-data` endpoint in favor of other existing remote endpoints or plugin-local endpoints.

No

## Testing instructions:
* Get a Jetpack testing site and check out this PR using the beta plugin
* Follow the test plan from https://github.com/Automattic/jetpack/pull/33997 and make sure that everything still works as described.
* Confirm that no errors show in the console while testing and that the automated tests are still passing

### DIFF
--- a/projects/packages/my-jetpack/_inc/components/product-cards-section/index.jsx
+++ b/projects/packages/my-jetpack/_inc/components/product-cards-section/index.jsx
@@ -1,6 +1,5 @@
 import { Container, Col } from '@automattic/jetpack-components';
 import React from 'react';
-import useProductData from '../../hooks/use-product-data';
 import AiCard from './ai-card';
 import AntiSpamCard from './anti-spam-card';
 import BackupCard from './backup-card';
@@ -22,8 +21,6 @@ const { showJetpackStatsCard = false } = window.myJetpackInitialState?.myJetpack
  * @returns {object} ProductCardsSection React component.
  */
 const ProductCardsSection = () => {
-	const { productData, fetchingProductData } = useProductData();
-
 	const items = {
 		backups: BackupCard,
 		scan: ScanAndProtectCard,
@@ -52,11 +49,7 @@ const ProductCardsSection = () => {
 
 				return (
 					<Col tagName="li" sm={ 4 } md={ 4 } lg={ 4 } key={ key }>
-						<Item
-							admin={ !! window?.myJetpackInitialState?.userIsAdmin }
-							productData={ productData[ key ] }
-							fetchingProductData={ fetchingProductData }
-						/>
+						<Item admin={ !! window?.myJetpackInitialState?.userIsAdmin } />
 					</Col>
 				);
 			} ) }
--- a/projects/packages/my-jetpack/_inc/state/actions.js
+++ b/projects/packages/my-jetpack/_inc/state/actions.js
@@ -21,10 +21,14 @@ const SET_CHAT_AVAILABILITY_IS_FETCHING = 'SET_CHAT_AVAILABILITY_IS_FETCHING';
 const SET_CHAT_AVAILABILITY = 'SET_CHAT_AVAILABILITY';
 const SET_CHAT_AUTHENTICATION_IS_FETCHING = 'SET_CHAT_AUTHENTICATION_IS_FETCHING';
 const SET_CHAT_AUTHENTICATION = 'SET_CHAT_AUTHENTICATION';
-const SET_PRODUCT_DATA_IS_FETCHING = 'SET_PRODUCT_DATA_IS_FETCHING';
-const SET_PRODUCT_DATA = 'SET_PRODUCT_DATA';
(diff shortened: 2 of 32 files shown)


projects/packages/my-jetpack/src/class-rest-product-data.php
```diff
@@ -47,67 +47,6 @@ public static function permissions_callback() {
...
-
-	/**
-	 * Get an array of backup/scan/anti-spam site capabilities
-	 *
-	 * @access public
-	 * @static
-	 *
-	 * @return WP_Error|\WP_REST_Response|null An array of capabilities
-	 */
-	public static function get_backup_capabilities() {
```
CodeyGuyDylan, line 87: This was originally used to see if the user's plan included realtime backups so that we didn't show anything funky if they had backup but not realtime. However after looking through this code, and remembering that we have a default state for if there are no undoable events, I don't think it is needed any longer. Nice cleanup

projects/packages/my-jetpack/src/class-rest-product-data.php
```diff
@@ -47,67 +47,6 @@ public static function permissions_callback() {
...
 	}
 
-	/**
-	 * Gets the product data for all products
-	 *
-	 * @return array|WP_Error
-	 */
-	public static function get_all_product_data() {
-		$site_id        = \Jetpack_Options::get_option( 'id' );
-		$wpcom_endpoint = sprintf( 'sites/%d/jetpack-product-data?locale=%2$s&force=wpcom', $site_id, get_user_locale() );
```
CodeyGuyDylan, line 57: I see you have a separate task to remove some more stuff on the backend. Will that be possible? With the old versions of the plugin calling this API endpoint still I think we will still need to return the backupable items at least, and that's essentially all it's doing right now

projects/packages/my-jetpack/_inc/state/actions.js
```diff
@@ -21,11 +21,15 @@ const SET_CHAT_AVAILABILITY_IS_FETCHING = 'SET_CHAT_AVAILABILITY_IS_FETCHING';
 const SET_CHAT_AVAILABILITY = 'SET_CHAT_AVAILABILITY';
```
CodeyGuyDylan, line 21: This is just a general comment and should not be addressed in this PR. With how many different types of data we are now querying, I think we might want to split up the actions/reducers/selectors/etc into different files so these files don't get so massive they're impossible to parse. Kinda like how it's currently done in the Jetpack plugin. Just a thought 😄

projects/packages/my-jetpack/_inc/components/product-cards-section/index.jsx
```diff
@@ -52,11 +49,7 @@ const ProductCardsSection = () => {
 
 				return (
 					<Col tagName="li" sm={ 4 } md={ 4 } lg={ 4 } key={ key }>
-						<Item
-							admin={ !! window?.myJetpackInitialState?.userIsAdmin }
-							productData={ productData[ key ] }
-							fetchingProductData={ fetchingProductData }
-						/>
+						<Item admin={ !! window?.myJetpackInitialState?.userIsAdmin } />
```
CodeyGuyDylan, line 52: It's nice that this refactor allowed us to pass less data through, I like this!

CodeyGuyDylan (dismissed):
I left some comments but didn't really request any changes. I like this approach because it minimizes the amount of data querying per request in a neat way.

I don't really have any issues with this approach, I think it is much better

Additionally I didn't encounter any issues when testing the card on My Jetpack

projects/packages/my-jetpack/_inc/components/product-cards-section/index.jsx
robertsreberski, line 52: I have the same remark as I left on @IanRamosC PR just now (https://github.com/Automattic/jetpack/pull/34384/files#discussion_r1410513881). Do you think we can use `useSelector` redux way of fetching data here?

`window.myJetpackInitialState` is sourced as a seed to initialize reducer, but we don't need to depend on the global window object after that. I think doing it React way is more encouraged, if there are no issues I don't see.

robertsreberski: Especially since you add and use selectors in other parts of the PR

robertsreberski (approved):
Rebase looks successful, code still works and looks good, and was previously approved by @CodeyGuyDylan. One small remark to be consistent in fetching state values - besides we can merge.

projects/packages/my-jetpack/_inc/components/product-cards-section/index.jsx
jboland88: I'm going to go ahead and merge this as-is for now. We are missing selectors/ reducers/ actions for any of the initial state object - I think this would be a nice improvement, but maybe good to refactor other similar instances at the same time.

PR #34202, opened 2023-11-20
Publicize: do not display module toggles when not connected

jeherve:
## Proposed changes:

When an admin is not connected to WordPress.com, let's not display the toggle to turn Publicize on or off. This way they would not inadvertently disable a feature that other authors on the site may be using for their own posts.

The invitation to connect your account to WordPress.com remains, so you can take action and toggle the module once you're connected.

**When not connected**

Screenshot 2023-11-20 at 11 24 01

**When connected**

Screenshot 2023-11-20 at 11 23 56

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

Related discussion: p1699929025605869-slack-CDLH4C1UZ

## Does this pull request change what data or activity we track or use?

* No

## Testing instructions:

* Start with a site that's connected to WordPress.com. 
* Go to Jetpack > Settings > Sharing
* Enable the Publicize feature.
* Go to Users > Add New and create a new admin.
* In a new browser, log in with that user.
* Go to Jetpack > Settings > Sharing.
   * You should only see an invitation to connect.

PR #34203, opened 2023-11-20
Publicize: allow all admins to toggle the feature on / off

jeherve:
This is an alternative to #34202

## Proposed changes:

*
All admins should be able to toggle features on and off, just like they can under Jetpack > Settings > Modules.

This commit also hides the upgrade link for folks that have not connected their account to WordPress.com yet, since they will not be able to purchase the upgrade without being connected.

**When user is not connected**

Screenshot 2023-11-20 at 11 33 47

**When user is connected**

Screenshot 2023-11-20 at 11 33 54

### Other information:

## Jetpack product discussion

Related discussion: p1699929025605869-slack-CDLH4C1UZ

* No

## Testing instructions:

* Start with a site that's connected to WordPress.com.Start with a site that's connected to WordPress.com.
* Go to Jetpack > Settings > Sharing
* Enable the Publicize feature.
* Go to Users > Add New and create a new admin.
* In a new browser, log in with that user.
* Go to Jetpack > Settings > Sharing. 
    * You should be able to turn the Publicize feature off, but you should not see the upgrade link.


pablinos:
> All admins should be able to toggle features on and off, just like they can under Jetpack > Settings > Modules.

Ah, if that's the case, then this is better for consistency. It feels like you could get into some slightly confusing states though. For example, if you turn the module on, but you haven't connected your account to WordPress.com, then you can't connect any social media accounts in Jetpack Cloud.

#34202 Might be safer even if it's less convenient. Perhaps I should check what happens with other modules that rely on the API.

jeherve:
> For example, if you turn the module on, but you haven't connected your account to WordPress.com, then you can't connect any social media accounts in Jetpack Cloud.

I don't know if that's an issue, given that we do not display the link to connect your social media accounts when your account isn't linked yet.

pablinos:
> I don't know if that's an issue, given that we do not display the link to connect your social media accounts when your account isn't linked yet.

Good point! Hopefully that's the only potential oddity. Let's go this route.

andrii-lysenko (approved):
The code looks reasonable and I haven't found any problems while testing on self-hosted. LGTM!

A quick question, we don't have this setting on Simple/Atomic. It's always enabled?

darssen (approved):
Nice work! Tested with trunk and the PR branch. Everything works as expected.

For **connected users,** the **functionality is the same** for both branches.

When logging in with a non-connected Admin, we now allow toggling the Auto-sharing function but don't show the Upgrade to Advance features link. See images below

Prior to the PR (Toggle disabled and update link present)
Selection_361


With the PR applied (Toggle enabled and update link not present)
Selection_360

jeherve:
> A quick question, we don't have this setting on Simple/Atomic. It's always enabled?

The upgrade isn't available on Simple/Atomic at the moment (see #32032).

simison:
@jeherve are you planning to do this for other modules as well?

I noticed similar line for Subscriptions module:

https://github.com/Automattic/jetpack/blob/12b00addaafb03d259c68c8162466f52ebc30455/projects/plugins/jetpack/_inc/client/newsletter/subscriptions-settings.jsx#L109-L111

jeherve:
> are you planning to do this for other modules as well?

Good point. #34358

PR #34207, opened 2023-11-20
Code Modernization: replace substr with str_starts/ends_with

### LINKED ISSUE #32864: Code modernization: Replace usage of substr with `str_starts_with` and `str_ends_with`
Source: https://core.trac.wordpress.org/ticket/58220

Let's follow Core's lead once Jetpack requires WordPress 6.3.

Primary issue: #32865

jeherve:
Fixes #32864

## Proposed changes:

`str_starts_with()` and `str_ends_with()` were introduced in PHP 8.0 to perform a case-sensitive check indicating if the string to search in (haystack) begins or ends with the given substring (needle).

WordPress core includes a polyfill for these functions on PHP < 8.0 as of WordPress 5.9.

This commit uses `str_starts_with()` and `str_ends_with()` where appropriate.

This aims to make the code more readable and consistent, as well as better aligned with modern development practices.

Note: this doesn't update uses in a few places, like the autoloader since it can be used outside of WordPress.

Matching core change: https://core.trac.wordpress.org/changeset/55990

### Other information:

## Jetpack product discussion

* Primary issue: #32865

* No

## Testing instructions:

* Activate all touched plugins and watch out for fatals on the frontend and wp-admin.


tools/check-changelogger-use.php
```diff
@@ -65,7 +65,7 @@ function usage() {
 			usage();
 			break;
 		default:
-			if ( substr( $argv[ $i ], 0, 1 ) !== '-' ) {
+			if ( ! str_starts_with( $argv[ $i ], '-' ) ) {
```
jeherve, line 68: @anomiex Could I have your opinion on this? Since we all run PHP 8.2 in our dev environments, should this trigger an error here?

anomiex: Yeah, this should be fine since we want 8.2 for the tooling.

To make phpcs happy we'll want to add `<rule ref="Jetpack-Compat-82" />` into `tools/.phpcs.dir.xml` and `tools/.phpcs.dir.phpcompatibility.xml` alongside what's already there. Probably the same to `.github/.phpcs.dir.xml` and `.github/.phpcs.dir.phpcompatibility.xml` too.

There's a possible exception in that some scripts might want to be run from the Tests or Linting jobs that install older PHP to test/lint with them, but we should be able to catch those when they make the Tests or Linting jobs fail.

jeherve: Giving this a try in 5c07039bd87bf29e6b29d3029d0c90c2991b201b

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
```diff
@@ -502,7 +502,7 @@ private function is_valid_admin_url_or_absolute_path( $input ) {
 		}
 
 		// Require that the string start with a slash, but not two slashes.
-		if ( '/' === substr( $input, 0, 1 ) && '/' !== substr( $input, 1, 1 ) ) {
+		if ( str_starts_with( $input, '/' ) && '/' !== substr( $input, 1, 1 ) ) {
```
anomiex: Maybe do this? I'm not sure.
```suggestion
		if ( str_starts_with( $input, '/' ) && ! str_starts_with( $input, '//' ) ) {
```

anomiex (dismissed):
One optional suggestion inline.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/class-launchpad-task-lists.php
jeherve: Yeah, why not? 👍

anomiex (approved):

PR #34127, opened 2023-11-14
General: update WP version requirements to WordPress 6.3.

jeherve:
## Proposed changes:

WordPress 6.4 is out, we can now require WP-1, i.e. WP 6.3.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

Primary issue: #32865

## Does this pull request change what data or activity we track or use?

* No

## Testing instructions:

* Is CI happy?

PR #31660, opened 2023-07-01
Theme Fields: load is_block_theme only for < WP 6.3

creativecoder:
## Proposed changes:

* Adds version check for loading is_block_theme field to wp/v2/themes API response, since this field is added by WP 6.3 and above

See related Core ticket: https://core.trac.wordpress.org/ticket/58123#ticket

**Note:** I'm AFK for quite a while starting next week, so feel free to take over this PR and merge it. Thanks!

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

See previous PR: https://github.com/Automattic/jetpack/pull/30144/files

## Does this pull request change what data or activity we track or use?

No.

## Testing instructions:

* Use REST API Tester from Jetpack debugger to check the response of `wp/v2/themes`. It should include `is_block_theme` from Jetpack if WP version is < 6.3, and from WP when the version is 6.3 or above.

PR #31579, opened 2023-06-26
Jetpack: Update test for WordPress core change

anomiex:
## Proposed changes:

WordPress core change https://core.trac.wordpress.org/changeset/56037 has changed the ordering of the `` parameters that our test was looking at. Add a check for a function added in that commit to determine which ordering to look for.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

p1687798900184079-slack-CDLH4C1UZ

## Does this pull request change what data or activity we track or use?

No

## Testing instructions:

* Is CI happy?

PR #34210, opened 2023-11-20
General: remove backwards-compatible code for versions < WP 6.3

### LINKED ISSUE #34209: General: remove 6.2 backward-compatible checks
Primary issue: #32865

jeherve:
Fixes #34209

## Proposed changes:

Remove code that is no longer necessary now that Jetpack requires WP 6.3+ (see #34127).

Related PRs:

- #31660 
- #32454 
- #31579

### Other information:

## Jetpack product discussion

Primary issue: #32865

* No

## Testing instructions:

* No functionality should change.


projects/plugins/jetpack/_inc/lib/core-api/wpcom-fields/theme-fields-is-block-theme.php
anomiex, line 1: I suppose we're only keeping this for the moment because of fbhepr%2Skers%2Sjcpbz%2Sjc%2Qpbagrag%2Serfg%2Qncv%2Qcyhtvaf%2Swrgcnpx%2Qsvryqf%2Sgurzr%2Qsvryqf%2Qvf%2Qoybpx%2Qgurzr.cuc%3Se%3Qq5psoq9r-og ? Or is there another reason?

If it's just that, maybe we should go ahead and delete that so we can delete the file here too.

jeherve: That's a good point. I thought about doing it the other way around out of habit, but why not removing it now; that does make things easier. Done in D129524-code (and removed from here in 2bacab152e3443c18ecd12fca6ac2a0527eb93da )

anomiex (approved):

PR #34211, opened 2023-11-20
CRM: Making sure invoice total count factors in credit notes and refunds correctly in invoice editor, preview and pdf view

coder-karen:
Issue: https://github.com/Automattic/zero-bs-crm/issues/2868

## Proposed changes:

* This PR primarily changes the logic in `ZeroBSCRM.InvoiceBuilder.php` and `ZeroBSCRM.DAL3.Obj.Invoices.php` in terms of how the invoice total is calculated and displayed when refunds and credit notes are factored in.
* This fixes the issue where refunds and credit notes were being incorrectly added to totals when previewing or viewing the pdf of an invoice, instead of being subtracted. However this 'fix' may itself need further scrutiny - it assumes that if a credit or refund is added (either minus or positive), it is meant to be subtracted from the invoice total.
* This PR still separates out Sales from Refunds / Credits due to the current behaviour of Sales being a little confusing (see below).
* Also, I did initially approach this issue by making sure that the preview and pdf pages matched the invoice edit page. What I've since realized is that fixing that would be a breaking change for people - if they had been inputting credits and refunds they would most likely have been entering them as negative amounts because the Client Portal and PDF are what matter most, as that is shared with customers. So the proposed fix is a catch-all for refunds and credits, and also allows further conversation about best practices.

### Existing quirks / related behaviour for comparison

When testing out totals I discovered some confusing behaviour (to me). This is likely expected, though without being able to grasp the logic behind it I'm noting some of the behaviour here, partly because anyone testing this PR may look at some of this behaviour and possibly be confused about expected vs new behaviour:

* When combined with WooCommerce and using WooSync and then setting WooSync settings to create invoices from orders, those invoices will have an associated transaction created that effectively pays off the invoice (assuming you've set transaction settings to include Pending statuses in the total value from the CRM transactions settings page). The transaction will be marked as a sale and have a positive value (matching the invoice total). 
* If you then apply a partial refund to that order within WooCommerce, that refund will generate a new transaction with a negative value sale. However that transaction is not automatically linked to an invoice. Linking it to the original invoice will only temporary work - each WooSync will unlink it. Linking it to the original invoice will however increase the invoice total.
* So, in effect, it appears that both positive and negative sales values look like they should pay off an invoice value given the current behaviour.
* I've shared more about this in Slack - p1700570444797999-slack-CTXBP902X - and now also in this issue - https://github.com/Automattic/zero-bs-crm/issues/3386. 
* For now however, I've approached this just by fixing what the seemingly current expected behaviour would be.

### Other information:

## Jetpack product discussion

Issue: https://github.com/Automattic/zero-bs-crm/issues/2868

No.

## Testing instructions:

* Using the [Jetpack Beta plugin](https://github.com/Automattic/jetpack-beta) on a test site with this branch applied or testing locally with the branch checked out, create a new invoice assigned to a contact, with a product and value greater than 0.
* Create a new transaction, set the transaction value to be less than the invoice value, and assign it to the contact and invoice just created. Test with all three types: Sale, Refund and Credit Note (when manually inputting, you can add either negative or positive for refund and credit - they should behave the same. With sales, the behaviour should match what it did before - positive value decreases the total amount, negative value increases it). See more in the existing quirks section for what may need to change in the future).
* On the invoice edit page, check the totals are correct. Then click Preview as well as downloading the pdf of the invoice - the totals should be correct (the total amount reduced for all three).


projects/plugins/crm/includes/ZeroBSCRM.InvoiceBuilder.php
```diff
@@ -1086,15 +1082,21 @@ function zeroBSCRM_invoicing_generateInvoiceHTML( $invoice_id = -1, $template =
...
 
-				// v3.0+ has + or - partials. Account for that:
-				if ( $partial['type_accounting'] === 'credit' ) {
-					// credit note, or refund
-					$balance = $balance + $partial['total'];
-				} else {
-					// assume debit
-					$balance = $balance - $partial['total'];
-				}
+					case __( 'Sale', 'zero-bs-crm' ):
```
coder-karen, line 1091: In these cases we're switching from checking the 'type_accounting', to checking for the 'type' (and the translated string too). It was noted here that isn't ideal - https://github.com/Automattic/zero-bs-crm/pull/2129/files#r875102950 - but that this is a bigger issue throughout. In this case, we're being consistent with the approach in `ZeroBSCRM.DAL3.Obj.Invoices.php`.

projects/plugins/crm/js/ZeroBSCRM.admin.invoicebuilder.js
```diff
@@ -1569,8 +1569,10 @@ function zbscrm_JS_calc_amount_due() {
 			// get
 			var v = jQuery( '.zbs-partial-value', ele ).text();
 
+			var label = jQuery( '.zlabel', ele ).text();
+
 			// detect +-
-			var multiplier = 1; // gets turned to -1 if negotive ()
+			var multiplier = 1; // gets turned to -1 if negative ()
 
 			// got -?
```
coder-karen, line 1577: It's unclear to me when the value amount would include a bracket, so leaving this for any unknown cases.

projects/plugins/crm/js/ZeroBSCRM.admin.invoicebuilder.js
```diff
@@ -1569,8 +1569,10 @@ function zbscrm_JS_calc_amount_due() {
...
 			if ( v.includes( '(' ) ) {
```
coder-karen, line 1578: It's unclear to me when the value amount would include a bracket, so leaving this for any unknown cases.

tbradsha: It's common in accounting for a negative value to be surrounded by parentheses instead of having the sign. I'm not sure if it's possible through normal flows to have them an invoice value (we block it client-side when manually adding an invoice, and our imports should have the proper sign), but apparently adding it in transactions then saving swaps things to a negative, so that's likely the intent.

coder-karen: Thanks for the explanation!

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Invoices.php
```diff
@@ -2753,52 +2753,57 @@ public function getOutstandingBalance($invoiceID=-1){
...
+										// these count as credits against invoice, but should be marked as negative (thus being deducted).
+										$transactions_total_value += $transaction_amount;
+										// These count as credits against invoice, and should be added.
+										// If inputted as negative, then add that amount.
+										if ( str_starts_with( $transaction_amount, '-' ) ) {
+											$transactions_total_value += $transaction_amount;
+										} else {
+											// If inputted as positive, then subtract that amount.
+											$transactions_total_value -= $transaction_amount;
+										}
```
cleacos: I don't understand this part at all...

if the invoice amount is 100 and the transaction is a Refund or a Credit Note with the amount of -30:

- `$invoiceTotalValue = 100.00`
- `$transaction_total_value = 0.0`
- `$transaction_amount = -30` initial amount
- `$transaction_total_value += $transaction_amount` => -30
- `$transaction_amount` is negative so `$transaction_total_value += $transaction_amount` => `$transaction_total_value = -60`
- Value to return: `$invoiceTotalValue + $transaction_total_value` = 40

If the transaction is positive, the `$transaction_total_value` will be `0` so the result to return is `100`.

coder-karen: Oh good catch. That extra line - `$transactions_total_value += $transaction_amount;` - was left in from the original code and should have been removed. I've taken it out now: 9267a249f0587ab46097ffbb7af52576d284c7ff

For me part of the challenge with that section of code is that I'm not sure how to actually test it so that's part of why I didn't catch that - but as the functionality here should match elsewhere it should be matching any changes elsewhere.

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.Invoices.php
```diff
@@ -2753,52 +2753,55 @@ public function getOutstandingBalance($invoiceID=-1){
...
+									case __( 'Refund', 'zero-bs-crm' ):
+									case __( 'Credit Note', 'zero-bs-crm' ):
+										// These count as credits against invoice, and should be added.
+										// If inputted as negative, then add that amount.
+										if ( str_starts_with( $transaction_amount, '-' ) ) {
+											$transactions_total_value += $transaction_amount;
+										} else {
+											// If inputted as positive, then subtract that amount.
+											$transactions_total_value -= $transaction_amount;
+										}
```
cleacos:
```suggestion

									$transactions_total_value -= 
abs( (float) $transaction_amount );

```

projects/plugins/crm/includes/ZeroBSCRM.InvoiceBuilder.php
```diff
@@ -621,23 +621,28 @@ function zeroBSCRM_invoicing_generateStatementHTML_v3( $contact_id = -1, $return
...
+								case __( 'Credit Note', 'zero-bs-crm' ):
+									// These count as credits against invoice, and should be added.
+									// If inputted as negative, then add that amount.
+									if ( str_starts_with( $partial['total'], '-' ) ) {
+										$balance = $balance + $partial['total'];
+									} else {
+										// If inputted as positive, then subtract that amount.
+										$balance = $balance - $partial['total'];
+									}
+									$payments += $partial['total'];
```
cleacos: Here we could do the same. Always add the absolute amount to the balance, ignoring the sign:

`$balance += abs( (float) $partial['total'] );`
`$payments -= abs( (float) $partial['total'] );`

( considering the balance as the pending amount to be paid )

projects/plugins/crm/includes/ZeroBSCRM.InvoiceBuilder.php
```diff
@@ -1086,14 +1091,26 @@ function zeroBSCRM_invoicing_generateInvoiceHTML( $invoice_id = -1, $template =
...
+					case __( 'Refund', 'zero-bs-crm' ):
+					case __( 'Credit Note', 'zero-bs-crm' ):
+						// These count as credits against invoice, and should be added.
+						// If inputted as negative, then add that amount.
+						if ( str_starts_with( $partial['total'], '-' ) ) {
+							$balance = $balance + $partial['total'];
+						} else {
+							// If inputted as positive, then subtract that amount.
+							$balance = $balance - $partial['total'];
+						}
```
cleacos: And I suppose that here the same, considering the balance as the pending amount to be paid:

`$balance += abs( (float) $partial['total'] );`

cleacos:
I had a call with Karen to discuss this PR... in the end, it needs some additional tests and it won't be ready for this release.

projects/plugins/crm/includes/ZeroBSCRM.InvoiceBuilder.php
coder-karen: Great suggestion. I have implemented this now, however with `$balance -=`, not `+=`, as it was adding the credit / refund amount to the invoice total. Perhaps I misunderstood what we talked about though - in this case with an invoice at £100, and a credit note of - or + £5, the total is now £95.

projects/plugins/crm/includes/ZeroBSCRM.InvoiceBuilder.php
coder-karen: See comment above.

coder-karen:
Adding an additional comment here as an update:

The current solution here is that in terms of transactions the behaviour of sales hasn't changed, but credits and refunds, when linked to an invoice, will always reduce the amount of invoice. This is both on the invoice editor itself, plus in preview and pdf.

cleacos:
Invoice: $40
Transaction: $10 (Sale)

I don't see the transaction amount deducted from the amount due in the invoice editor:

image

In the invoice preview, it is deducted as expected:

image

With transaction type `Credit` or `Refund` (either), the transaction amount in the invoice (Editor) is deducted:

image

projects/plugins/crm/js/ZeroBSCRM.admin.invoicebuilder.js
```diff
@@ -1582,10 +1582,13 @@ function zbscrm_JS_calc_amount_due() {
...
 			// if v is a negative number and also a Credit or Refund, add to amount_due, otherwise, deduct.
-			if ( v < 0 && ( label.includes( 'Refund' ) || label.includes( 'Credit' ) ) ) {
-				amount_due += v;
-			} else {
-				amount_due -= v;
+			// if ( v < 0 && ( label.includes( 'Refund' ) || label.includes( 'Credit' ) ) ) {
+			// 	amount_due += v;
+			// } else {
+			// 	amount_due -= v;
+			// }
```
cleacos: We have to remove this commented code.

coder-karen: Yes, thanks for testing! It looks like the logic wasn't fully completed, so this section should be (at quick glance):
```
if (  label.includes( 'Refund' ) || label.includes( 'Credit' ) ) {
amount_due -= Math.abs( v );
} else {
amount_due -= v;
}
```
Which would fix the sales amount issue above too.

cleacos:
Now, it works as expected.

I've found some issues with the WooCommerce order creating the linked invoice:

When you create a new order, JPCRM creates an invoice and a transaction. Even if the order is unpaid, the invoice says there is no pending amount (Amount due).
- When an order is refunded, totally or partially, the transaction refund amount appears in the invoice but with zero value. (Note: In theory, any refund should generate a new invoice, but in our case, we update the existing invoice, but that is another issue to discuss).
- The refund transaction is generated, marked as Refund status, and also as Refund type, and linked to the original invoice.

image

cleacos:
Even worse, when the WooSync cronjob executes the sync between orders and invoices/transaction, the refund transaction disappears from the invoice...

image

and the transaction is updated: Transaction Status from `Refunded` to `Completed` and the type, from `Refund` to `Sale`, and it removes the link with the invoice.

Another example:

image

cleacos (approved):
Approving this PR. It works as expected.

I left a comment about some issues I found related to the WooCommerce orders.

coder-karen:
Thank you @cleacos  for getting this over the line! I've added those points re WooSync to this issue: https://github.com/Automattic/zero-bs-crm/issues/3386

cleacos:
Thank you!

PR #34212, opened 2023-11-20
JPCRM. Adds a filter to update the WooSync data

ecairol:
JPCRM. Adds a filter to update the WooSync data before convering order into a CRM Object.

This filter would allow currency conversions during a WooSync process. Because the whole order_data object is being sent, it'd allow a external plugin to:
 - Add transaction extra metadata to keep track of the original amount along the exchange rate used (eg: EUR-USD)
 - Modify the total amount of the transaction: EUR imported values can be stored as USD in the transaction['total'] field

Fixes #

## Proposed changes:

* Pass the Order data thru a WP filter before converting it in CRM Objects

### Other information:

## Testing instructions:

* 
*

### DIFF
--- a/projects/plugins/crm/changelog/2023-11-20-17-49-45-010348
+++ b/projects/plugins/crm/changelog/2023-11-20-17-49-45-010348
@@ -1,0 +1,4 @@
+Significance: patch
+Type: added
+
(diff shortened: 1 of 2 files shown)


tbradsha (approved):
Fixed whitespace error and added a changelog. Otherwise looks good to go!

PR #34216, opened 2023-11-20
CLI: Refactor jetpack changelog command to pass directly to changelogger

sdixon194:
## Proposed changes:

This refactor lets us massage the changelog entries for the monorepo via the CLI, instead of just calling the changelogger in interactive mode. This will let us add any custom questions or data we want without interfering with the public package. 

This doesn't add the new prompt for if it's a user facing change, as discussed in https://github.com/Automattic/jetpack/pull/34058

It just lays the groundwork to do that.
### Other information:

## Jetpack product discussion

Internal ref: p1699566741898429-slack-C05Q5HSS013

No
## Testing instructions:

Make sure the `jetpack changelog add` command still works as expected:

- `jetpack changelog add <type>/<project> should add a changelog file to the specified project
- If you make a change to a project, just running `jetpack changelog add` should find the project and let you add a changelog file to it.
- If you make changes to multiple projects, you should have the option to add the same changelog file to all of them
- It should give the appropriate changelog types for projects with specified changelog types, like jetpack.


tools/cli/commands/changelog.js
```diff
@@ -261,27 +261,26 @@ async function changelogCommand( argv ) {
 }
 
 /**
- * Checks if any projects have special changelog configurations.
+ * Checks if any projects have special changelog type configurations.
  *
  * @param {Array} needChangelog - files that need a changelog.
  * @returns {Array} - array of projects with unique changelog configurations.
```
anomiex: Return type doc needs updating

tools/cli/commands/changelog.js
```diff
@@ -290,80 +289,102 @@ async function checkSpecialProjects( needChangelog ) {
...
+				argv,
+				proj,
+				projectChangeTypes[ proj ] ? projectChangeTypes[ proj ] : defaultTypes
+			);
+			console.log( chalk.green( `Running changelogger for ${ proj }!` ) );
+			argv = await formatAutoArgs( proj, argv, response );
+			await changelogArgs( argv );
+		}
+		return;
 	}
```
anomiex: We can avoid this code duplication by doing something like this above instead, before the `if ( promptConfirm.changelogConfirm || promptConfirm.sameChangelogFiles ) {` check.

```js
if ( promptConfirm.separateChangelogFiles ) {
    uniqueProjects.unshift( ...defaultProjects.splice( 0 ) );
}
```

Then, of course, that `if ( promptConfirm.changelogConfirm || promptConfirm.sameChangelogFiles ) {` check needs to change. Probably the most sensible thing would be to remove the check and always run the body, after changing `changelogAddPrompt` to return false where it would currently return `{ changelogConfirm: false }`.

tools/cli/commands/changelog.js
```diff
@@ -290,80 +289,102 @@ async function checkSpecialProjects( needChangelog ) {
...
+		if ( projectChangeTypes[ proj ] !== 'default' ) {
+			uniqueProjects.push( proj );
+		} else {
+			defaultProjects.push( proj );
+			projectChangeTypes[ proj ] = defaultTypes;
 		}
+	}
+
+	// Confirm what projects we're adding a changelog to, and how we want to add them.
+	const promptConfirm = await changelogAddPrompt( argv, defaultProjects, uniqueProjects );
```
anomiex: We need to avoid calling `changelogAddPrompt` if `argv.project` is set. Or `changelogAddPrompt` should just assume "yes" if `argv.project` is set, either way works.

tools/cli/commands/changelog.js
```diff
@@ -722,15 +730,21 @@ async function promptVersion( argv ) {
...
  */
-async function promptChangelog( argv, needChangelog ) {
+async function promptChangelog( argv, needChangelog, types ) {
 	const gitBranch = child_process
 		.spawnSync( 'git', [ 'branch', '--show-current' ] )
 		.stdout.toString()
 		.trim()
 		.replace( /\//g, '-' );
-	console.log( gitBranch );
+	const maxLength = Math.max( ...Object.keys( types ).map( key => key.length ) );
```
anomiex: `Array.reduce` doesn't get used often enough. 😀
```suggestion
	const maxLength = Object.keys( types ).reduce( ( a, v ) => v.length > a ? v.length : a, 0 );
```

tools/cli/commands/changelog.js
```diff
@@ -834,44 +823,56 @@ async function promptChangelog( argv, needChangelog ) {
  */
 async function changelogAddPrompt( argv, needChangelog, uniqueProjects ) {
```
anomiex: The `needChangelog` argument should be renamed to `defaultProjects` to better reflect what it is (and the doc comment above should be updated accordingly).

tools/cli/commands/changelog.js
```diff
@@ -834,44 +823,56 @@ async function promptChangelog( argv, needChangelog ) {
...
 	const totalProjects = [ ...needChangelog, ...uniqueProjects ];
+
+	// If only one project needs a changelog, confirm.
```
anomiex: There's some redundancy in the code below. You might want to restructure it something like this

```js
let prompts;

if ( totalProjects.length === 1 ) {
    prompts = { /* changelogConfirm: Add changelog for the one project? */ };
} else if ( defaultProjects.length <= 1 && uniqueProjects.length > 0 ) {
    prompts = { /* separateChangelogFiles: Found N projects needing changelogs. Create for each one? */ };
} else if ( defaultProjects.length > 0 && uniqueProjects.length === 0 ) {
    prompts = {
        /* sameChangelog: Found N projects. Same to all? */
        /* separateChangelogFiles: (else) For each individually? */
    };
} else {
    prompts = {
        /* sameChangelog: Found N projects that can use the same changelog and M that need individual ones. Same to the N? */
        /* separateChangelogFiles: (else) For each individually? */
    };
}

const response = await inquirer.prompt( prompts );

// Return the object only if some question was answered "yes", otherwise false.
return Object.values( response ).some( v => v ) ? response : false;
```

That also (in combination with my suggestion on line 388) should happen to fix some weird behavior in a case like `touch projects/packages/changelogger/foo.txt projects/plugins/jetpack/foo.txt && git add . && jetpack changelog add`
```
Checking if changelog files are needed. Just a sec...
? Found 1 projects that can accept the same changelog file.
  Found 1 project(s) that requires a separate changelog configuration. 
  Add same changelog file to 1 project(s)? No
? Create changelog for each project individually?
```
Just treat them all as unique.

tools/cli/commands/changelog.js
sdixon194: Nice catch! 48567a9057bfb7a292ca9664980130d9c11fc379

tools/cli/commands/changelog.js
sdixon194: Good thinking 👍 48567a9057bfb7a292ca9664980130d9c11fc379

tools/cli/commands/changelog.js
sdixon194: Oooh very fancy.

tools/cli/commands/changelog.js
sdixon194: Right now it still confirms the project we're releasing a changelog file to on line 828 - you think we should skip the confirmation altogether since someone is passing the project manually?

anomiex: Yes. That's the current behavior in trunk, and adding the extra confirmation seems silly.

sdixon194:
Okay, think this is good for another look 👍

tools/cli/commands/changelog.js
```diff
@@ -817,61 +806,64 @@ async function promptChangelog( argv, needChangelog, types ) {
...
+	const totalProjects = [ ...defaultProjects, ...uniqueProjects ];
+	let prompts;
 
-	// If only one project needs a changelog, confirm.
+	/* changelogConfirm: Add changelog for the one project? */
 	if ( totalProjects.length === 1 ) {
-		const response = await inquirer.prompt( {
+		prompts = {
 			type: 'confirm',
 			name: 'changelogConfirm',
```
anomiex: While my original suggestion still had "changelogConfirm" as the name here, it probably makes more sense overall to rename it like this:
```suggestion
			name: 'separateChangelogFiles',
```

That also fixes a bug where you didn't change line 866 to account for this now that the direct `return` is gone, without needing any change to line 866.

tools/cli/commands/changelog.js
```diff
@@ -817,61 +806,64 @@ async function promptChangelog( argv, needChangelog, types ) {
...
-		} );
-		return response;
-	}
-
-	// If multiple projects need a changelog that can be the same, confirm.
-	if ( uniqueProjects.length === 0 && needChangelog.length > 1 ) {
-		const response = await inquirer.prompt( [
+		};
+	} else if ( defaultProjects.length <= 1 && uniqueProjects.length > 0 ) {
+		/* separateChangelogFiles: Found N projects needing changelogs. Create for each one? */
```
anomiex: 😀 The comments from my pseudocode are kind of pointless as code comments, they just repeat what the prompt messages say (and in more stilted language).

A good comment might point out _why_ the first two only have one question (for "separateChangelogFiles") while the other two have two, namely that the "sameChangelog" mode only makes sense if defaultProjects,length > 1.

tools/cli/commands/changelog.js
```diff
@@ -817,61 +806,64 @@ async function promptChangelog( argv, needChangelog, types ) {
...
+		prompts = {
+			type: 'confirm',
+			name: 'separateChangelogFiles',
+			message: `Found ${ totalProjects.length } project(s) that need a changelog. Create changelog for each one?`,
+		};
+	} else if ( defaultProjects.length > 0 && uniqueProjects.length === 0 ) {
+		console.log( "Where I'm supposed to be" );
```
anomiex: Left this debugging message in 😀

tools/cli/commands/changelog.js
```diff
@@ -817,61 +806,64 @@ async function promptChangelog( argv, needChangelog, types ) {
...
+		/* sameChangelog: Found N projects. Same to all? */
+		prompts = [
 			{
 				type: 'confirm',
-				name: 'sameChangelogFiles',
-				message: `Found ${ needChangelog.length } project(s) that need a changelog. Create and add same changelog to all projects?`,
+				name: 'sameChangelog',
```
anomiex: Heh. I didn't really intend to rename this from "sameChangelogFiles" to "sameChangelog", I must have just left off the "Files". Either way works though.

tools/cli/commands/changelog.js
```diff
@@ -828,50 +806,65 @@ async function promptChangelog( argv, needChangelog ) {
...
+async function changelogAddPrompt( argv, defaultProjects, uniqueProjects ) {
+	const totalProjects = [ ...defaultProjects, ...uniqueProjects ];
+	let prompts;
+
+	/* changelogConfirm: Add changelog for the one project? */
+	if ( totalProjects.length === 1 ) {
+		prompts = {
+			type: 'confirm',
+			name: 'changelogConfirm',
+			message: `Add changelog for ${ totalProjects[ 0 ] }?`,
```
anomiex: Since we're here, we may as well copyedit the messages too.
```suggestion
			message: `Add a changelog for ${ totalProjects[ 0 ] }?`,
```

tools/cli/commands/changelog.js
```diff
@@ -828,50 +806,65 @@ async function promptChangelog( argv, needChangelog ) {
...
+		};
+	} else if ( defaultProjects.length <= 1 && uniqueProjects.length > 0 ) {
+		/* separateChangelogFiles: Found N projects needing changelogs. Create for each one? */
+		prompts = {
+			type: 'confirm',
+			name: 'separateChangelogFiles',
+			message: `Found ${ totalProjects.length } project(s) that need a changelog. Create changelog for each one?`,
```
anomiex:
```suggestion
			message: `Found ${ totalProjects.length } project(s) that need a changelog. Create a changelog for each one?`,
```

tools/cli/commands/changelog.js
```diff
@@ -828,50 +806,65 @@ async function promptChangelog( argv, needChangelog ) {
...
+	} else if ( defaultProjects.length > 0 && uniqueProjects.length === 0 ) {
+		console.log( "Where I'm supposed to be" );
+		/* sameChangelog: Found N projects. Same to all? */
+		prompts = [
 			{
 				type: 'confirm',
-				name: 'autoAdd',
-				message: `Found ${ needChangelog.length } project(s) that need a changelog. Create and add same changelog to all projects?`,
+				name: 'sameChangelog',
+				message: `Found ${ defaultProjects.length } project(s) that need a changelog. Create and add same changelog to all projects?`,
```
anomiex:
```suggestion
				message: `Found ${ defaultProjects.length } project(s) that need a changelog. Create and add the same changelog to all of them?`,
```

tools/cli/commands/changelog.js
```diff
@@ -828,50 +806,65 @@ async function promptChangelog( argv, needChangelog ) {
...
 			},
+			/* separateChangelogFiles: (else) For each individually? */
 			{
 				type: 'confirm',
-				name: 'autoPrompt',
+				name: 'separateChangelogFiles',
 				message: 'Create changelog for each project individually?',
```
anomiex:
```suggestion
				message: 'Create a changelog for each project individually?',
```

tools/cli/commands/changelog.js
```diff
@@ -828,50 +806,65 @@ async function promptChangelog( argv, needChangelog ) {
...
-		] );
-		return response;
+		];
+	} else {
+		/* sameChangelog: Found N projects that can use the same changelog and M that need individual ones. Same to the N? */
+		prompts = [
+			{
+				type: 'confirm',
+				name: 'sameChangelog',
+				message: `Found ${ defaultProjects.length } project(s) that can accept the same changelog and ${ uniqueProjects.length } projects that need individual ones. Create and add same changelog to ones we're able?`,
```
anomiex:
```suggestion
				message: `Found ${ defaultProjects.length } project(s) that can accept the same changelog and ${ uniqueProjects.length } project(s) that need individual ones. Create and add the same changelog to the ones we're able?`,
```

tools/cli/commands/changelog.js
```diff
@@ -828,50 +806,65 @@ async function promptChangelog( argv, needChangelog ) {
...
+			},
+			/* separateChangelogFiles: (else) For each individually? */
+			{
+				type: 'confirm',
+				name: 'separateChangelogFiles',
+				message: 'Create changelog for every project individually?',
```
anomiex:
```suggestion
				message: 'Create a changelog for each project individually?',
```

anomiex (commented):
One bug found, and a bunch of suggestions.

tools/cli/commands/changelog.js
sdixon194: lol nice catch

tools/cli/commands/changelog.js
sdixon194: Updated this for consistency

tools/cli/commands/changelog.js
sdixon194: Hah, I used these as a guidelines and meant to rework them. Removed them.

sdixon194:
I think this is good for hopefully one last look 👍

tools/cli/commands/changelog.js
```diff
@@ -290,81 +289,92 @@ async function checkSpecialProjects( needChangelog ) {
...
-				chalk.green( `Running auto changelogger for ${ needChangelog.length } project(s)!` )
-			);
-			const response = await promptChangelog( argv, needChangelog );
-			for ( const proj of needChangelog ) {
-				argv = await formatAutoArgs( proj, argv, response );
-				await changelogArgs( argv );
-			}
+	// Confirm what projects we're adding a changelog to, and how we want to add them.
+	const promptConfirm = argv.project
+		? true
```
anomiex: If we do this, the behavior is a little more consistent between e.g. `jetpack changelog add plugins/beta` and `jetpack changelog add plugins/jetpack`.
```suggestion
		? { separateChangelogFiles: true }
```

tools/cli/commands/changelog.js
```diff
@@ -828,50 +806,59 @@ async function promptChangelog( argv, needChangelog ) {
...
-
-	if ( totalProjects.length === 1 ) {
-		const response = await inquirer.prompt( {
-			type: 'confirm',
-			name: 'autoPrompt',
-			message: `Add changelog for ${ totalProjects[ 0 ] }?`,
-		} );
-		return response;
+	const response = await inquirer.prompt( prompts );
+	if ( ! response.sameChangelog && ! response.separateChangelogFiles ) {
```
anomiex: Bug
```suggestion
	if ( ! response.sameChangelogFiles && ! response.separateChangelogFiles ) {
```

anomiex (commented):
One bug found, and one suggestion

anomiex (approved):
Let's go for it!

PR #34222, opened 2023-11-20
Remove unnecessary `function_exists()` checks

anomiex:
## Proposed changes:

These should be safe:

* `class_alias` was added in PHP 5.3.
* `error_clear_last` was added in PHP 7.0.
* `inet_pton` was not available on Windows until php 5.3.
* `libxml_disable_entity_loader` was added in php 5.2.11. No need to check when we're not checking for calls to other libxml functions.
* `DOMDocument` requires `libxml_use_internal_errors`, so we shouldn't need to check for it separately.

Others, such as `ignore_user_abort`, are apparently sometimes disabled on hosts so we still need to check them. Similarly, we should probably keep checks for functions from any PHP modules (e.g. dom, intl, mbstring, openssl) that can be disabled.

I'm unsure about these:

* The use of `filter_var` in `projects/packages/connection/src/class-manager.php`. It explicitly calls out the possibility of the Filter module not being compiled in, but we have tons of unguarded calls elsewhere.
* `glob` in `projects/packages/status/src/class-files.php`, the `glob_php` function was added way back in 2011 with a comment that glob "is not always available" but no details.

I also skipped crm, vaultpress, and super-cache since they're full of un-phpcs-ed stuff and other legacy code, and
`projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-jwt.php` since it's a copy-paste of an external library.

### Other information:

## Testing instructions:

* Stuff should still work


jeherve (dismissed):

projects/plugins/jetpack/_inc/lib/class-jetpack-podcast-feed-locator.php
```diff
@@ -49,7 +49,7 @@ private function is_podcast_feed( $file ) {
 	private function safely_load_xml( $xml ) {
 		$disable_entity_loader = PHP_VERSION_ID < 80000;
 
-		if ( $disable_entity_loader && ! function_exists( 'libxml_disable_entity_loader' ) ) {
+		if ( $disable_entity_loader ) {
```
jeherve, line 52: I approved and then noticed the failing tests, seemingly caused by that:
https://github.com/Automattic/jetpack/actions/runs/6937070982/job/18870424528?pr=34222

jeherve (dismissed):

jeherve (approved):

PR #34224, opened 2023-11-21
Adds the WooCommerce HPOS sync module to list of active modules.

zinigor:
This change is needed to make sure we can run checksums on WooCommerce HPOS order database entries.

## Proposed changes:
* Adds the `woocommerce_hpos_orders` Sync module to the list of active modules.

### Other information:

## Jetpack product discussion
p9dueE-7MN-p2

## Testing instructions:

* You will need a Phab diff for testing this, testing instructions will be in the diff.


zinigor:
This is incorrect, the module is already added in the `initialize_woocommerce` method, it just needs better logic.

zinigor:
I need to get back to this, will do that when I get a moment.

zinigor:
This is super stale, going to get implemented as part of this initiative:
p9dueE-8ec-p2

PR #34227, opened 2023-11-21
Whitelist payments endpoints for blaze package

therocket-gr:
Fixes #

## Proposed changes:

Added a GET and POST endpoint whitelisting for the api/v1/payments DSP route. We need those endpoints for any new API call that we want to implement in the DSP side. (For context I am linking to this PR in dsp https://github.tumblr.net/Tumblr/a8c-dsp/pull/628 where we want those endpoints implemented)

### Other information:

## Testing instructions:

At the moment the endpoint is not existent yet (we have used an already whitelisted /subscriptions endpoint to make it work temporarily) so testing this PR requires to edit your local DSP (to add the new endpoint) and test it using the local DSP repo. 

But - I think that - a code review (as this is something we have already implemented a couple of times) is enough

### DIFF
--- a/projects/packages/blaze/changelog/vdimitrakis-SELFSERVE-830-whitelist-payments-endpoints
+++ b/projects/packages/blaze/changelog/vdimitrakis-SELFSERVE-830-whitelist-payments-endpoints
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
+Added whitelisting for payments endpoint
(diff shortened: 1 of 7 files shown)


sbarbosa (approved):
Thanks for the change @therocket-gr.

To add context for future reviewers, this new endpoint will be used for the new changes in the payment system of the widget. 

I locally tested the changes and the redirection is working fine:
Screenshot 2023-11-21 at 14 15 46

PR #34230, opened 2023-11-21
Use spaceship operator in sorting functions

anomiex:
## Proposed changes:

PHP 7 introduced the spaceship operator, `$a <=> $b`, which is basically equivalent to `$a < $b ? -1 : $a > $b ? 1 : 0`. It also won't do the wrong thing with floats like `$a - $b` might.

There's a difference versus `strcmp` in that `strcmp` will always cast arguments to strings while `<=>` will do PHP's normal type comparisons, so something like `strcmp( 1, 10 )` would be different from `1 <=> 10`.

### Other information:

## Testing instructions:

* All the things should still sort in the same orders.

### DIFF
--- a/projects/plugins/jetpack/modules/sharedaddy/sharing-service.php
+++ b/projects/plugins/jetpack/modules/sharedaddy/sharing-service.php
@@ -696,7 +696,7 @@ class Sharing_Service_Total {
 	 * @param object $a Sharing_Service_Total object.
 	 * @param object $b Sharing_Service_Total object.
 	 *
-	 * @return bool
+	 * @return int -1, 0, or 1 if $a is <, =, or > $b
 	 */
 	public static function cmp( $a, $b ) {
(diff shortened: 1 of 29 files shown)


projects/plugins/jetpack/modules/sharedaddy/sharing-service.php
```diff
@@ -696,13 +696,13 @@ public function __construct( $id, $total ) {
 	 * @param object $a Sharing_Service_Total object.
 	 * @param object $b Sharing_Service_Total object.
 	 *
-	 * @return bool
+	 * @return int -1, 0, or 1 if $a is <, =, or > $b
```
anomiex, line 699: The situation with these was weird. Returning `$a < $b` means `usort()` sees 0 or 1, resulting in a reverse-order sort.

jeherve (approved):

PR #34213, opened 2023-11-20
Backport mu-wpcom-plugin 2.0.0, jetpack 12.9-a.7 Changes

ice9js:
## Proposed changes:

Backports changes for mu-wpcom-plugin 2.0.0, jetpack 12.9-a.7

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

N/a
## Does this pull request change what data or activity we track or use?

No
## Testing instructions:

Proofread changes.

PR #34236, opened 2023-11-21
tools: Avoid empty changelog entries in releases

anomiex:
## Proposed changes:

When `tools/changelogger-release.sh` is about to release a project, have it check for whether changelogger is going to prompt for "There are no changes with content for this write" and supply a change file with some content ("Internal updates") to both avoid the prompt and avoid the resulting empty entry in the changelog.

### Other information:

## Jetpack product discussion

https://github.com/Automattic/jetpack/pull/34213#discussion_r1399722631

No

## Testing instructions:

* Check out 6e2ed4566db621514afdcfbec595d927e38be3fa, cherry-pick this change, then run `tools/changelogger-release.sh plugins/jetpack`. There should be no prompts about "There are no changes with content for this write", and `git diff '**/CHANGELOG.md'` should not have any empty entries being added.

### DIFF
--- a/tools/changelogger-release.sh
+++ b/tools/changelogger-release.sh
@@ -174,7 +174,7 @@ for SLUG in "${TO_RELEASE[@]}"; do
 	# Avoid "There are no changes with content for this write. Proceed?" prompts and empty changelog entries.
 	ANY=false
 	for f in "$CHANGES_DIR"/*; do
-		if [[ -n $( sed -n '/^$/,$ { /[^ \t]/ p }' "$CHANGES_DIR"/* ) ]]; then
+		if [[ -n $( sed -n -e '/^$/,$ { /[^ \t]/ p }' "$f" ) ]]; then
 			ANY=true
 			break
 		fi
--- a/tools/changelogger-release.sh
+++ b/tools/changelogger-release.sh
@@ -172,7 +172,14 @@ for SLUG in "${TO_RELEASE[@]}"; do
 	RELEASED[$SLUG]=1
 
 	# Avoid "There are no changes with content for this write. Proceed?" prompts and empty changelog entries.
-	if [[ -z $( sed -sn '/^$/,$ { /[^ \t]/ p }' "$CHANGES_DIR"/* ) ]]; then
+	ANY=false
+	for f in "$CHANGES_DIR"/*; do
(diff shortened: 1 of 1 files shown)


anomiex:
In particular someone on a Mac (and with the stock sed installed rather than GNU sed, if that passes `tools/check-development-environment.sh`) should test this.

tools/changelogger-release.sh
```diff
@@ -171,6 +171,19 @@ for SLUG in "${TO_RELEASE[@]}"; do
 	info "Processing $SLUG..."
 	RELEASED[$SLUG]=1
 
+	# Avoid "There are no changes with content for this write. Proceed?" prompts and empty changelog entries.
+	ANY=false
+	for f in "$CHANGES_DIR"/*; do
+		if [[ -n $( sed -n '/^$/,$ { /[^ \t]/ p }' "$CHANGES_DIR"/* ) ]]; then
```
sdixon194:
```suggestion
		if [[ -n $( sed -n -e '/^$/,$ { /[^ \t]/ p }' "$CHANGES_DIR"/* ) ]]; then
```
Without the `-e`, this sed outputs the following:

Screenshot 2023-11-29 at 12 50 58 PM

Which seems to be from MacOS using a different version of sed as you indicated.

Also is it supposed to be:

```suggestion
		if [[ -n $( sed -n -e '/^$/,$ { /[^ \t]/ p }' "$f" ) ]]; then
```
?

anomiex: Yes, good catches on both!

tools/changelogger-release.sh
```diff
@@ -171,6 +171,19 @@ for SLUG in "${TO_RELEASE[@]}"; do
 	info "Processing $SLUG..."
 	RELEASED[$SLUG]=1
 
+	# Avoid "There are no changes with content for this write. Proceed?" prompts and empty changelog entries.
+	ANY=false
+	for f in "$CHANGES_DIR"/*; do
+		if [[ -n $( sed -n -e '/^$/,$ { /[^ \t]/ p }' "$f" ) ]]; then
```
sdixon194:
```suggestion
		if [[ -n $( sed -n -e '/^$/,$ { /[^ \t]/ p; }' "$f" ) ]]; then
```

So I'm still getting the error, and after doing some more testing, it seems like adding a semicolon here is the only thing that fixes it for some reason?

sdixon194 (approved):
Nice this works for me now, thanks for fixing this!

PR #34240, opened 2023-11-21
My Jetpack: Link Jetpack AI Contact Us button to support email on interstitial page

### LINKED ISSUE #33358: AI Tiered Plans: Update upgrade interstitial to show the next tier available

dhasilva:
Fixes #33358

## Proposed changes:

* Uses a `mailto` link with a specific subject

### Other information:

## Testing instructions:

* In a site with an unlimited plan or 1000 tier, go to the Jetpack AI interstitial page
* Click on `Contact us`
* Your email client should open with the `to` and `subject` fields with the right values

### DIFF
--- a/projects/packages/my-jetpack/_inc/components/product-interstitial/jetpack-ai/more-requests.jsx
+++ b/projects/packages/my-jetpack/_inc/components/product-interstitial/jetpack-ai/more-requests.jsx
@@ -13,9 +13,6 @@ import GoBackLink from '../../go-back-link';
 import jetpackAiImage from '../jetpack-ai.png';
 import styles from './style.module.scss';
 
-// Disabled for now
-const onClick = () => {};
-
 /**
  * JetpackAIInterstitialMoreRequests component
  *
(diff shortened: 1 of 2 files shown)


lhkowalski (approved):
It's working as expected:

Screenshot 2023-11-22 at 09 54 00

---

An idea for a follow-up: now that we understand how the Jetpack Redirect works, we could create a record for the `jetpack-ai-inquiry-about-more-requests` slug and set the target of it as the `mailto:` link. This way, when we have a fancier way to handle the button in the future (like a form), we just update the target of the redirect to the new location, no code changes (and no release) needed.

PR #34242, opened 2023-11-22
Add/prevent batcache when accessing paywalled content

millerf:
## Proposed changes:

* Send no-cache headers so batcache is disabled when user access a paywalled content

### Other information:

### Works only on Atomic (can't see batcache on simple)
* Deploy on your sandbox
* Sandbox your preferred site
* Create a paywalled content
* Access gated content and notice a batcache hit
* Access paywall and notice no cache header

### DIFF
--- a/projects/plugins/jetpack/changelog/add-prevent-batcache-when-accessing-paywalled-content
+++ b/projects/plugins/jetpack/changelog/add-prevent-batcache-when-accessing-paywalled-content
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
+Prevent caching paywalled content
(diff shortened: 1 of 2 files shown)


chrismccluskey (approved):
Tested this and I see the `Cache-Control: no-cache` header both as an editor and as a subscriber. LGTM!

simison:
@chrismccluskey @millerf would it make sense to disable the cache for this view, too?

image

millerf:
> @chrismccluskey @millerf would it make sense to disable the cache for this view, too?
> 
> image

Why?

millerf:
> @chrismccluskey @millerf would it make sense to disable the cache for this view, too?
> 

What makes you think this page is cached?

Screenshot 2023-12-04 at 11 53 57

PR #34243, opened 2023-11-22
Fix Jetpack Social pre publish UI

### LINKED ISSUE #84314: Jetpack "Share this Post" option cannot be disabled, and cannot type a message
### Quick summary

This issue occurs when publishing or scheduling a post.

The Share this post option cannot be disabled, and you cannot type a message - https://share.zight.com/X6unGNyX.

### Steps to reproduce

1. Connect a social media account to blog.
2. Create a post and click on publish
3. On the Pre-publish preferences, go to the "Share this post" section
4. Try disabling/enabling the option
5. Try adding a message

1. The option can be enabled/disabled
2. The message area should accept input

### What actually happened

1. The option cannot be enabled/disabled → shows the same status
2. The message area is not accepting any input, and only accepts the first letter

### Impact

Some (< 50%)

### Available workarounds?

_No response_

### Logs or notes

7328996-zen

manzoorwanijk:
Due to some regression in #34111, reactivity in pre-publish UI for Jetpack Social doesn't work as expected as reported in https://github.com/Automattic/wp-calypso/issues/84314

Fixes https://github.com/Automattic/wp-calypso/issues/84314

## Proposed changes:

* Convert the Jetpack Social Store utils to `usePostMeta` hook to make it reactive

### Other information:

## Testing instructions:

* Follow the steps in #84314
* Confirm that the issue is resolved.


projects/js-packages/publicize-components/src/utils/use-post-meta.js
```diff
@@ -0,0 +1,77 @@
...
+
+		return {
+			isPublicizeEnabled,
+			jetpackSocialOptions,
+			attachedMedia,
+			imageGeneratorSettings,
+			isPostAlreadyShared,
+			shareMessage,
+			shouldUploadAttachedMedia,
+		};
```
pablinos, line 39: Is this stable as it's a new object? I think it should be as we shouldn't get here without the meta changing, but if not, then this will invalidate the `useMemo` below

projects/js-packages/publicize-components/src/utils/use-post-meta.js
```diff
@@ -0,0 +1,77 @@
...
+			updateMeta( 'jetpack_social_options', {
+				...metaValues.jetpackSocialOptions,
+				[ key ]: value,
+			} );
+		},
+		[ metaValues.jetpackSocialOptions, updateMeta ]
+	);
+
+	return useMemo( () => {
+		return {
```
pablinos: We could get rid of this `return`

pablinos (commented):
I much prefer this. Just had a thought about the memoing in the hook.

projects/js-packages/publicize-components/src/utils/use-post-meta.js
manzoorwanijk: Yes, it's stable because of the optimisation of `useSelect` via that dependency array.

projects/js-packages/publicize-components/src/utils/use-post-meta.js
manzoorwanijk: Done.

pablinos (approved):

PR #34244, opened 2023-11-22
Launchpad: Add dynamic titles to task lists

valterlorran:
Related to https://github.com/Automattic/wp-calypso/issues/83880

## Proposed changes:

* The task list title is currently hardcoded on the CustomerHome launchpad component. Meaning that all the task lists has the same name across different intents.
* This PR adds the possibility of defining the title on the back end, which can give as more flexibility on the titles. 
* For now, this will only be used on the task lists displayed on the Customer Home.

### Other information:

## Testing instructions:

* Apply this PR to your sandbox and make sure you are sandboxed
```
bin/jetpack-downloader test jetpack-mu-wpcom-plugin add/launchpad-dynamic-titles
```
* Navigate to the Developer Console: https://developer.wordpress.com/docs/api/console/
* Make a `GET` request to `wpcom/v2 -> /sites/:siteSlug/launchpad?checklist_slug=:checklist_slug`, please test all the cases below to ensure we are returning the title to all use cases in the Customer Home. Verify that you see the `title` key as shown in the screenshot below.
  - `build`
  - `free`
  - `link-in-bio`
  - `link-in-bio-tld`
  - `newsletter`
  - `videopress`
  - `write`
  - `start-writing`
  - `design-first`
  - `intent-build`
  - `intent-write`
  - `intent-free-newsletter`
  - `intent-paid-newsletter`

Screen Shot 2023-11-22 at 10 01 08

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php
@@ -146,6 +146,7 @@ class WPCOM_REST_API_V2_Endpoint_Launchpad extends WP_REST_Controller {
 			'checklist'          => wpcom_get_launchpad_checklist_by_checklist_slug( $checklist_slug ),
 			'is_enabled'         => wpcom_get_launchpad_task_list_is_enabled( $checklist_slug ),
 			'is_dismissed'       => wpcom_launchpad_is_task_list_dismissed( $checklist_slug ),
+			'title'              => wpcom_get_launchpad_checklist_title_by_checklist_slug( $checklist_slug ),
 		);
 	}
 
--- a/projects/packages/jetpack-mu-wpcom/changelog/add-launchpad-dynamic-titles
+++ b/projects/packages/jetpack-mu-wpcom/changelog/add-launchpad-dynamic-titles
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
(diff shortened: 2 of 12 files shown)


projects/packages/jetpack-mu-wpcom/src/features/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-launchpad.php
```diff
@@ -146,6 +146,7 @@ public function get_data( $request ) {
 			'checklist'          => wpcom_get_launchpad_checklist_by_checklist_slug( $checklist_slug ),
 			'is_enabled'         => wpcom_get_launchpad_task_list_is_enabled( $checklist_slug ),
 			'is_dismissed'       => wpcom_launchpad_is_task_list_dismissed( $checklist_slug ),
+			'title'              => wpcom_get_launchpad_checklist_title_by_checklist_slug( $checklist_slug ),
```
daledupreez, line 149: As a general thought, it may be worth building one instance of the task list and then calling functions -- while this pattern is clean, it does feel like we may end up doing duplicate work.
(Definitely not blocking - even if we tackle this, I think it should be in a follow-up.)

daledupreez (approved):
This is looking good to me, though I think we should follow up on the three checklists where we _didn't_ add the titles to work out what they should be. Those are:
* `earn`
* `host-site`
* `subscribers`

Beyond the basic instructions, I also tested what would happen when adding `&_locale=de` to the URL, and I correctly got translated title values returned. (I am not sure how or why they were already present, but I went with it!)

PR #34248, opened 2023-11-22
Module-active checks should always return modules as active on WordPress.com simple

### LINKED ISSUE #34231: Status > Modules: update module checks to be aware of WordPress.com Simple
The Status package includes a `Modules` class that can be very handy to check whether a specific Jetpack module is active:
https://github.com/Automattic/jetpack/blob/c100a09efb6c4c59029fc3e16ac0560bd1cf978c/projects/packages/status/src/class-modules.php#L21-L27

Unfortunately, those tools are not compatible with WordPress.com, where modules are always active and where you do not need to check for Jetpack options (nor should you, that won't work).

Today, we work around that problem by adding a condition along with the module check, like in #34229. 
Ideally, that condition should be baked in the Module class.

pgk:
Fixes #34231, closes https://github.com/Automattic/greenhouse/issues/1959

## Proposed changes:

* As outlined in #34231 we should always return true on module active checks when running Jetpack on WordPress.com

### Other information:

## Testing instructions:

* Note: Original context for this was #34229. There might be other tests that need to be performed in other areas of the plugin.
* Apply PR as a diff on wpcom and sandbox a simple site with a plan that includes VideoPress
* Create a subscription plan. For more info check https://github.com/Automattic/jetpack/pull/32154
* Create a post, add a subscriber-only block and a videopress private video block.
* View as a subscriber. It should work.


jgcaruso (approved):
Works as described.

Subscribed to a wpcom simple site's Paid Content block and I was able to view the private video.

Existing videos continue to work as expected for both subscribers and non-subscribers.

jeherve (approved):

PR #34258, opened 2023-11-22
Promote subscription: Add verbum_subscription_modal setting

lupus2k:
Fixes https://github.com/Automattic/verbum/issues/34

## Proposed changes:
* Add an option to enable/disable the subscription modal on Verbum.

PT: pb5gDS-3tB-p2
Calypso PR: https://github.com/Automattic/wp-calypso/pull/84448

We will add this to the `comments` section of `/settings/discussion/`
image

### Other information:

No

### Calypso
Use this PR: https://github.com/Automattic/wp-calypso/pull/84448

### Developer console
* Follow the instructions of the first comment to apply on your sandbox
* Go to https://developer.wordpress.com/docs/api/console/
* Choose WPCOM API / v1.4 / GET
* Use this URL: `/sites/YOUR_SIMPLE_SITE_BLOG_ID/settings`
* Check for `verbum_subscription_modal`

### DIFF
--- a/projects/packages/sync/src/class-defaults.php
+++ b/projects/packages/sync/src/class-defaults.php
@@ -189,6 +189,7 @@ class Defaults {
 		'wpcom_reader_views_enabled',
 		'wpcom_site_setup',
 		'wpcom_subscription_emails_use_excerpt',
+		'verbum_subscription_modal',
 	);
 
 	/**
--- a/projects/packages/sync/src/class-defaults.php
+++ b/projects/packages/sync/src/class-defaults.php
@@ -189,7 +189,7 @@ class Defaults {
 		'wpcom_reader_views_enabled',
 		'wpcom_site_setup',
 		'wpcom_subscription_emails_use_excerpt',
-		'verbum_subscription_modal',
+		'jetpack_verbum_subscription_modal',
 	);
 
 	/**
--- a/projects/packages/sync/changelog/add-verbum-subscription-modal-settings
+++ b/projects/packages/sync/changelog/add-verbum-subscription-modal-settings
@@ -1,4 +1,4 @@
 Significance: minor
 Type: added
 
-Add verbum_subscription_modal setting to manage subscription modal show/hide on Verbum
+Add jetpack_verbum_subscription_modal setting to manage subscription modal show/hide on Verbum
--- a/projects/packages/sync/changelog/add-verbum-subscription-modal-settings
+++ b/projects/packages/sync/changelog/add-verbum-subscription-modal-settings
@@ -1,0 +1,4 @@
+Significance: minor
+Type: added
+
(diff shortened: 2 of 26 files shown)


jeherve (commented):
Is it too late to update the option name? I'm asking since I'm thinking it would be nice if it were prefixed with `jetpack_`, to avoid confusion later on for any self-hosted on WoA site owners who may be looking at their site options and finding a `verbum` option that may not tell them much.

dhasilva (commented):
Tests well on the console. Commenting as there is a request from Jeremy.

lupus2k:
> Is it too late to update the option name? I'm asking since I'm thinking it would be nice if it were prefixed with `jetpack_`, to avoid confusion later on for any self-hosted on WoA site owners who may be looking at their site options and finding a `verbum` option that may not tell them much.

Hi @jeherve, it's not too late yet, so I'll change the name. The thing is this is not for Jetpack, it is also available for simple sites, should we prefix `jetpack_` or `wpcom_`?

jeherve:
> It's not too late yet, so I'll change the name.

🥳 Thank you!

> The thing is this is not for Jetpack, it is also available for simple sites, should we prefix `jetpack_` or `wpcom_`?

We use the `jetpack_` prefix for things that are available for both platforms, so I think it should be ok here.

lupus2k:
Thanks for your feedback+review @jeherve @dhasilva, I updated the PR and is ready for a second review 🙇

dhasilva (approved):

projects/packages/sync/src/class-defaults.php
```diff
@@ -189,6 +189,7 @@ class Defaults {
 		'wpcom_reader_views_enabled',
 		'wpcom_site_setup',
 		'wpcom_subscription_emails_use_excerpt',
+		'jetpack_verbum_subscription_modal',
```
fgiannar, line 192: From a Sync perspective, is there a WPCOM counterpart diff that adds this option in the Sync options' whitelist on WPCOM?
Also just a double check that you actually need this setting synced :)

fgiannar (changes requested):
Hi there 👋 

Let an inline comment related to the Sync package changes. Many thanks for working on this!

projects/packages/sync/src/class-defaults.php
lupus2k: Thanks for noticing this. 
I can confirm we need this setting synced as we will use this on Jetpack and WPCOM. I'll create the WPCOM diff

lupus2k: Diff D129916-code

fgiannar (approved):
Thanks for handling the WPCOM counterpart for Sync 👍

PR #34259, opened 2023-11-22
My Jetpack: Remove hardcoded tiers from Jetpack AI interstitial

### LINKED ISSUE #34208: Jetpack AI: Remove hardcoded tiers
We have hardcoded the tiers until now, but to make it easier to change the tier list, let's remove those hardcoded values and use a single point of truth from the backend.

dhasilva:
Fixes #34208

## Proposed changes:

* Removes most hardcoded tier references, keeping only `100` as that tier has different text for it.

### Other information:

## Testing instructions:

* Mock the `get_tier_licensed_quantity` value for `500`, `750` or `1000`. At this point you know the drill.
* In your sandbox, enable `USE_STORE_SANDBOX`
* Sandbox the public API
* Locally, on `packages/my-jetpack/src/class-wpcom-products.php`, return true on `is_cache_old` to update the prices at the next request
* Go to the Jetpack AI interstitial page
* Check that the new tiers are included and working
* Remove the `return true` on the cache function after disabling `USE_STORE_SANDBOX` to get the production prices back

### DIFF
--- a/projects/packages/my-jetpack/_inc/components/product-interstitial/index.jsx
+++ b/projects/packages/my-jetpack/_inc/components/product-interstitial/index.jsx
@@ -276,9 +276,9 @@ export function JetpackAIInterstitial() {
 	const { detail } = useProduct( slug );
 	const { onClickGoBack } = useGoBack( { slug } );
 
-	const nextTier = detail?.[ 'ai-assistant-feature' ]?.[ 'next-tier' ]?.value;
+	const nextTier = detail?.[ 'ai-assistant-feature' ]?.[ 'next-tier' ] || null;
 
-	if ( nextTier == null ) {
+	if ( ! nextTier ) {
 		return <JetpackAIInterstitialMoreRequests onClickGoBack={ onClickGoBack } />;
 	}
 
@@ -292,7 +292,7 @@ export function JetpackAIInterstitial() {
 			imageContainerClassName={ styles.aiImageContainer }
 			ctaButtonLabel={ ctaLabel }
 			hideTOS={ true }
-			quantity={ nextTier }
+			quantity={ nextTier.value }
 			directCheckout={ hasRequiredPlan }
 		>
 			<img src={ jetpackAiImage } alt="Jetpack AI" />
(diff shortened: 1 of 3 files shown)


projects/packages/my-jetpack/_inc/components/product-interstitial/index.jsx
```diff
@@ -276,12 +276,9 @@ export function JetpackAIInterstitial() {
 	const { detail } = useProduct( slug );
 	const { onClickGoBack } = useGoBack( { slug } );
 
-	const currentTier = detail?.[ 'ai-assistant-feature' ]?.[ 'current-tier' ]?.value;
 	const nextTier = detail?.[ 'ai-assistant-feature' ]?.[ 'next-tier' ]?.value;
```
CGastrell: I think checking if `next-tier` has any value is enough. But if you really want to get a value from it, I'd suggest using the `limit`, as that will never be `0` so you can rely on `! nextTier`.

Besides, a non existent property on an object will give you `undefined`, not `null` (might also suggest using strict comparison on line 281)

lhkowalski: He may want to get a value from it to set as quantity on the interstitial. This value will be used to determine the `q` parameter on the checkout URL, so it's necessary. Using `value` or `limit` for the `q` parameter will have no difference, unless we start changing the limits of a given tier without changing it's quantity (unlikely).

+1 regarding the strict comparison, or maybe something like:

```
const nextTier = detail?.[ 'ai-assistant-feature' ]?.[ 'next-tier' ] || null;
```

Then

```
if ( !! nextTier ) {
```

And also:

```
quantity={ nextTier.value }
```

lhkowalski:
I tested the code with all the possible tiers and it's working nicely.

When the current tier is free (0):

Screenshot 2023-11-23 at 10 53 34

When the current tier is 100:

Screenshot 2023-11-23 at 10 54 59

When the current tier is 200:

Screenshot 2023-11-23 at 10 55 31

When the current tier is 500:

Screenshot 2023-11-23 at 10 55 56

When the current tier is 750:

Screenshot 2023-11-23 at 10 56 25

When the current tier is 1000:

Screenshot 2023-11-23 at 10 56 53

When the current tier is unlimited (1):

Screenshot 2023-11-23 at 10 54 29

lhkowalski (approved):
I tested all the scenarios again after the last changes, and they are working as expected.

PR #34263, opened 2023-11-23
Search: fix title for product layout

kangzj:
Fixes Automattic/jpop-issues/issues/8682

## Proposed changes:

* Reset display for title in product layout, so that it's properly wrapped

### Other information:

## Jetpack product discussion
n/a

no

## Testing instructions:

* Go to you local test site with Instant Search enabled
* Change layout to 'Product (for WooCommerce stores)' on page `/wp-admin/admin.php?page=jetpack-search-configure`
* Open the search model from the frontend `/?s=`
* Ensure titles are properly wrapped not like what's described in https://github.com/Automattic/jpop-issues/issues/8682

image

### DIFF
--- a/projects/packages/search/changelog/fix-instant-search-title-layout
+++ b/projects/packages/search/changelog/fix-instant-search-title-layout
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Instant Search: fix title layout for product layout
--- a/projects/packages/search/src/instant-search/components/search-result-product.scss
+++ b/projects/packages/search/src/instant-search/components/search-result-product.scss
@@ -52,6 +52,9 @@ $item-spacing: 16px;
 		$column-count: 5;
 		width: calc( ( 100% - ( #{$item-spacing} * #{$column-count} ) ) / #{$column-count} );
 	}
+	.jetpack-instant-search__search-result-title-link {
+		display: initial;
+	}
 }
 
 .jetpack-instant-search__search-result > .jetpack-instant-search__search-result-product-img-link {


dognose24 (commented):
I found the title of the default `RESULT_FORMAT_MINIMAL` is also not aligned with the prefix icon.

|Before|After|
|-|-|
|截圖 2023-11-24 下午3 01 10|截圖 2023-11-24 下午3 14 48|

Could we change the `display: initial` in `search-result.scss` for both `.jetpack-instant-search__search-result-minimal-title-link` and `.jetpack-instant-search__search-result-product-title-link`?

kangzj:
Thanks @dognose24 good point!

The reason I didn't apply that to minimal is that, the check mark exists in the minimal layout too:

https://github.com/Automattic/jetpack/blob/f021fbc0e60493ab827d8ef1ae69cfd37f4b74a1/projects/packages/search/src/instant-search/components/search-result-minimal.jsx#L124

I checked [wpcom forum](https://wordpress.com/forums?s=), which seems to be using `expanded`, but I'm not sure if that's the only place. 

Pinging @wellyshen who originally added the style, can we apply what's suggested by @dognose24 please? Thanks

Either way, I'd like to merge the fix for the product layout, and we could figure out the minimal layout later because the code freeze is really close now. What do you say @dognose24 ?

dognose24:
@kangzj Thanks for your consideration. Sounds good to me! 👍🏼

dognose24 (approved):
LGTM! Let's ship and iterate for the minimal format result. 🙂

kangzj:
@dognose24 Thank you ❤️

PR #34268, opened 2023-11-23
Gutenberg: Dequeue editor assets when they aren't in use

tyxla:
Currently, on WP.com, the pattern inserter is cluttered with asset requests for every pattern preview. This causes the pattern inserter to reach the browser request limits as soon as the user browses through a couple of pattern pages. 

I spotted that this occurs because in Jetpack, we'll enqueue all the editor assets for the admin during the `enqueue_block_assets` action, and that action can be called multiple times, including [when retrieving the iframed editor assets](https://github.com/WordPress/gutenberg/blob/cc1870a49f57f89cb4c97b41782420a36f77e4da/lib/compat/wordpress-6.4/script-loader.php#L146). Because of that, all the Jetpack editor assets will be loaded inside each pattern preview iframe when they won't be utilized there.

This causes a serious bug: each pattern preview comes with hundreds of requests, and we're hitting the browser request limit and it will stop processing those assets. Furthermore, this results in errors in handling these patterns.

To resolve this, I'm suggesting we keep the asset enqueueing, but dequeue when we know that we shouldn't be loading the editor scripts and styles. Right now, I'm only suggesting doing the dequeuing for the scripts, but we'll likely have to do the same thing with styles. Dequeueing is cheap: even if an asset hasn't been enqueued yet, it will silently behave as a noop.

However, this immediately fixes the bug and allows the browser to load all previews as the user browses through pattern pages. These errors are no longer cluttering the console and there is a visible performance improvement on slower machines (can be tested with CPU throttling). I expect this should improve the performance of the pattern inserter in WP.com (which @mtias had mentioned before), but I think we have a few separate issues to address independently in that area.

## Proposed changes:
Gutenberg: Dequeue editor assets when they aren't in use

### Other information:

## Jetpack product discussion
None, but we've been discussing with @mtias that the pattern inserter is particularly slow on WP.com.

None

## Testing instructions:

* Apply this on WP.com
* Open a WP.com simple site. 
* Open the pattern inserter.
* Click on the button to browser all patterns.
* Click through some pattern tabs and verify that the patterns keep loading quicker than before.
* Verify you don't see new INSUFFICIENT_RESOURCES errors or undefined variable errors in the console.
* Test on a JN site too.

## Previews and screenshots

Browsing pattern inserter - before:

https://github.com/Automattic/jetpack/assets/8436925/d03a819a-2e5c-4bed-ab21-87033c89dbe1

Browsing pattern inserter - after:

https://github.com/Automattic/jetpack/assets/8436925/51c115f8-13ea-42db-b589-c8c557dec584

Error with the browser request limit

Screenshot 2023-11-23 at 17 04 20

Error with result from the unprocessed assets

Screenshot 2023-11-23 at 17 04 26

### DIFF
--- a/projects/plugins/jetpack/class.jetpack-gutenberg.php
+++ b/projects/plugins/jetpack/class.jetpack-gutenberg.php
@@ -633,7 +633,7 @@ class Jetpack_Gutenberg {
 
 			return;
 		}
-		
+
 		$status = new Status();
 
 		// Required for Analytics. See _inc/lib/admin-pages/class.jetpack-admin-page.php.
--- a/projects/plugins/jetpack/class.jetpack-gutenberg.php
+++ b/projects/plugins/jetpack/class.jetpack-gutenberg.php
@@ -622,6 +622,18 @@ class Jetpack_Gutenberg {
 			return;
 		}
 
+		/**
+		 * This can be called multiple times per page load in the admin, during the `enqueue_block_assets` action.
+		 * These assets are necessary for the admin for editing but are not necessary for each pattern preview.
+		 * Therefore we dequeue them, so they don't load for each pattern preview iframe.
+		 */
+		if ( ! wp_should_load_block_editor_scripts_and_styles() ) {
+			wp_dequeue_script( 'jp-tracks' );
+			wp_dequeue_script( 'jetpack-blocks-editor' );
+
+			return;
+		}
+		
 		$status = new Status();
 
 		// Required for Analytics. See _inc/lib/admin-pages/class.jetpack-admin-page.php.
--- a/projects/plugins/jetpack/changelog/update-dequeue-editor-assets-perf
+++ b/projects/plugins/jetpack/changelog/update-dequeue-editor-assets-perf
@@ -1,0 +1,4 @@
+Significance: patch
+Type: enhancement
+
+Gutenberg: Dequeue editor assets when they aren't in use


jsnajdr:
I don't understand how this patch works. The pattern preview is an iframe that is not loaded from a URL, but is populated directly with a content string. And all the `<script>` tags come from an editor setting called `__unstableResolvedAssets.scripts`. This setting is serialized, in the `wp-edit-post-js-after` inline script, in the editor initial HTML document. In that document, `wp_should_load_block_editor_scripts_and_styles` is certainly true and we are loading all the Gutenberg assets.

So, what difference does this patch make? Is there a situation where `enqueue_block_editor_assets` is called, but `wp_should_load_block_editor_scripts_and_styles` is false?

tyxla:
> I don't understand how this patch works. The pattern preview is an iframe that is not loaded from a URL, but is populated directly with a content string. And all the `<script>` tags come from an editor setting called `__unstableResolvedAssets.scripts`. This setting is serialized, in the `wp-edit-post-js-after` inline script, in the editor initial HTML document. In that document, `wp_should_load_block_editor_scripts_and_styles` is certainly true and we are loading all the Gutenberg assets.

That's correct so far. Then, the `__unstableResolvedAssets.scripts` setting itself is populated from the `wp_get_iframed_editor_assets()` function, or `_gutenberg_get_iframed_editor_assets_6_4()` if we are in compatibility mode. There, the script string is essentially generated from the following 2 calls:

```
wp_print_head_scripts();
wp_print_footer_scripts();
```

However, these are generic WordPress functions that are called when generating the header and footer scripts to be rendered in the front-end of the sites - specifically during `wp_head()` and `wp_footer()`, which WordPress themes utilize, respectively in their `<head>` tags and right before the closing `<body>` tag. At the same time, the iframe doesn't render a real page, it's a synthetic one and its content and assets are the ones generated by `wp_get_iframed_editor_assets()`, so the `is_admin()` conditional tag is `true`. Therefore, we end up loading a ton of scripts for each iframe, because Jetpack doesn't expect that we'll have a situation where we'll be both in the admin and not wanting to render all admin scripts and styles. 

> So, what difference does this patch make? Is there a situation where `enqueue_block_editor_assets` is called, but `wp_should_load_block_editor_scripts_and_styles` is false?

Yes, we're taking advantage of exactly one such situation. The underlying filter of `wp_should_load_block_editor_scripts_and_styles()` will be `true` for the general admin page load, but it will intentionally be set to `false` by Gutenberg right when these iframed editor assets are being generated:

https://github.com/WordPress/gutenberg/blob/fc5619d6ac11605683f4a72dcea53e92d20f4410/lib/compat/wordpress-6.4/script-loader.php#L145

However, Jetpack wasn't respecting that filter before, and with the proposed change in this PR, it started respecting it. 

I know it's a tricky one, it took me like 2 hours to trace today. Let me know if that makes more sense!

jsnajdr:
I see, it makes sense now :+1: I'll have a closer look in the morning if you don't mind. By the way, I'm wondering if we need to load any scripts at all in the iframe. Don't we need only styles there? The content of the iframe is rendered and managed by the parent React app, there should be no script running _inside_ the iframe.

tyxla:
> I see, it makes sense now 👍 I'll have a closer look in the morning if you don't mind. 

No rush, thank you 🙌 

> By the way, I'm wondering if we need to load any scripts at all in the iframe. Don't we need only styles there? The content of the iframe is rendered and managed by the parent React app, there should be no script running _inside_ the iframe.

A few scripts are running there even in pure Gutenberg, so I believe we should keep the rest of the scripts. After all, theoretically, there might be scripts necessary for the view part of the blocks rendered in one or more of the patterns. 

On a side note, I do want to audit the styles there, too. In my testing, there's 1MB of styles (a big part of it being custom fonts), which we could be able to clean up a bit. It seems like it's not as straightforward though, since those styles could be registered and enqueued in various places and it's not guaranteed they even have a consistent naming scheme. This is something I'd like to dig into separately once we address the scripts part.

projects/plugins/jetpack/class.jetpack-gutenberg.php
```diff
@@ -622,6 +622,18 @@ public static function enqueue_block_editor_assets() {
...
 		}
 
+		/**
+		 * This can be called multiple times per page load in the admin, during the `enqueue_block_assets` action.
+		 * These assets are necessary for the admin for editing but are not necessary for each pattern preview.
+		 * Therefore we dequeue them, so they don't load for each pattern preview iframe.
+		 */
+		if ( ! wp_should_load_block_editor_scripts_and_styles() ) {
+			wp_dequeue_script( 'jp-tracks' );
+			wp_dequeue_script( 'jetpack-blocks-editor' );
```
jsnajdr, line 632: Two questions about this: why do we need to dequeue the two scripts? They are not going to be enqueued in this `enqueue_block_editor_assets` method. Then who else has already enqueued them?

For the iframed assets, the `_wp_get_iframed_editor_assets` is working on a private temporary instance of `WP_Scripts`, which only copies _registered_ scripts from the global instance, but not the _queued_ ones.

Second question: this method is called also in an non-admin context, see the condition here:

https://github.com/Automattic/jetpack/blob/42aec0a2c7b092bf2fb2c8d6fdca1aba428f457c/projects/plugins/jetpack/class.jetpack.php#L721

Apparently we want to enqueue the `jetpack-blocks-editor` script in some context that has `is_admin() == false`. Probably some site editor iframe? I can't be frontend because the `enqueue_block_editor_assets` is executed only in admin. Is it ever called at all?

If the intent is to enqueue `jetpack-blocks-editor` in some page where `is_admin()` is `false`, then your patch is preventing that because the `wp_should_load_block_editor_scripts_and_styles` will check `$current_screen->is_block_editor()`.

jsnajdr: The condition mentioned above was added in #32706 by @BogdanUngureanu. Maybe he knows what the non-admin code path does. And what exactly did that PR fix? The PR description mentions something about correctly rendering the Donations block, but without specific details.

tyxla: > why do we need to dequeue the two scripts? They are not going to be enqueued in this enqueue_block_editor_assets method. Then who else has already enqueued them?

They will be enqueued on the first call of `enqueue_block_editor_assets()`, which is part of the initial editor load, so when that method runs again (during subsequent `enqueue_block_editor_assets()` calls, like for retrieving the editor assets), they'll already be enqueued. But we don't need them, because they only add extra unused scripts that clutter network requests and block the pattern preview loading process. That's why we're dequeuing them. You can refer to my previous comment for more information on how this works: https://github.com/Automattic/jetpack/pull/34268#issuecomment-1824748005

> Apparently we want to enqueue the jetpack-blocks-editor script in some context that has is_admin() == false. Probably some site editor iframe? I can't be frontend because the enqueue_block_editor_assets is executed only in admin. Is it ever called at all?
> The condition mentioned above was added in #32706 by @BogdanUngureanu. Maybe he knows what the non-admin code path does. And what exactly did that PR fix? The PR description mentions something about correctly rendering the Donations block, but without specific details.

I'm not sure about that usage, too. As you mentioned it would be good to get some information from the folks who introduced it. 

> If the intent is to enqueue jetpack-blocks-editor in some page where is_admin() is false, then your patch is preventing that because the wp_should_load_block_editor_scripts_and_styles will check $current_screen->is_block_editor().

Which is fine by design, and if that's the case, then that code should also use the built-in filter of `wp_should_load_block_editor_scripts_and_styles()` to override the default behavior and allow loading of the editor assets outside of the admin context for that particular case. I'd be happy to add that where applicable, if we get a confirmation from the authors. 

Perhaps someone else from @Automattic/t-rex will know?

BogdanUngureanu: Hi @jsnajdr and @tyxla, I've used the two-hook approach because:
- in WP-Admin we had some performance issues in the Site Editor https://github.com/Automattic/jetpack/pull/32641 - basically, the GS UI was sluggish and difficult to use
- However, using the same enqueue hook on the front-end (e.g. p2 post comments) the assets weren't loaded properly.

So, to fix the UI performance we had in the SE, we used the new approach for enqueuing the assets, but to have it also working on front-end (e.g. p2 comments), we fallback to the previous way of loading the assets since on the front-end we don't use an iframed editor (from what I remember).

**WP-Admin Performance**
From what I understood and remember, the problem was that the assets were initially loaded in the root document and migrated by Gutenberg on the front-end to the Gutenberg iframe (DOM manipulation) - it was Gutenberg's way of keeping backward compatibility for the older hook. Because this operation is done on the client side, it caused performance problems.

tyxla: @BogdanUngureanu thanks for elaborating! The approach appears to be causing performance issues for the pattern previews on WP.com though. Can you help us with some testing instructions on how to test the scenarios you described, so we can ensure we're not breaking something?

jeherve (approved):
I have not tested the performance, but the blocks (and tracking) continue to work well on self-hosted, WoA, and simple sites, in the block editor and in the site editor.

projects/plugins/jetpack/class.jetpack-gutenberg.php
BogdanUngureanu: Sure! :)

- Open the Site Editor and then the Global Styles panel
- Go to Typography > Text > Size input.
- Toggle the input to a slider (the icon in the right)
- Move the slider left and right
- You'll notice that the slider UI responds slower before the change I've introduced

You can see a difference in response time in these two videos:

https://github.com/Automattic/jetpack/assets/995955/e35582a4-ad88-4836-82dc-2896507b85ea


https://github.com/Automattic/jetpack/assets/995955/d579ef1e-5d19-4075-98bd-ee18a61d84d6

In the video there are different types of sliders, for the one I've mentioned above, the difference should be more visible because the inputs change faster with a regular slider.

jsnajdr: > They will be enqueued on the first call of `enqueue_block_editor_assets()`, which is part of the initial editor load, so when that method runs again (during subsequent `enqueue_block_editor_assets()` calls, like for retrieving the editor assets), they'll already be enqueued.

I think it works a bit differently: the first call of `enqueue_block_editor_assets` will enqueue the assets for the initial HTML page of the editor.

Second time it runs it is inside `_wp_get_iframed_editor_assets`. There the global `$wp_scripts` object has been replaced temporarily with a newly created instance, where nothing is enqueued. I verified on a simple site that all I need to do is `if ( ! wp_should_load() ) { return; }`. No need to dequeue `jetpack-blocks-editor` because it's not enqueued. Unless there is a _second_ `enqueue_block_assets` registered filter that will add it. Then the best fix is to find it and add the same condition there, too.

jsnajdr: > However, using the same enqueue hook on the front-end (e.g. p2 post comments) the assets weren't loaded properly.

Oh, so here we have a case where we want to enqueue `jetpack-blocks-editor` even on the frontend, in a non-admin environment. So that the P2 editor has Jetpack blocks available.

If this is true, and the P2 editor relies on this Jetpack code to load Jetpack blocks, then the current patch breaks P2. Jetpack blocks won't be available there. I didn't test this yet, but it's something to verify before shipping.

jsnajdr: Verified that P2 works with this patch. The `Isolated_Block_Editor` class (`plugins/p2/.../class-editor-isolated.php`) will load the `jetpack-blocks-editor` script. Because it calls `do_action( 'enqueue_block_editor_assets' )`, and part of that action is enqueueing all scripts that were found in `block.json` files, as the `editorScript` fields. And Jetpack blocks declare `jetpack-blocks-editor` this way.

tyxla: I don't think it will break Jetpack blocks in p2, because we're intentionally enabling that same filter there:

https://github.com/Automattic/p2/blob/9c8502cc17775e863450e6e42c9e720dff096dc8/plugins/p2/src/server/core/editor/class-editor-isolated.php#L15C55-L15C55

which will result in `wp_should_load_block_editor_scripts_and_styles()` returning true as well, ultimately keeping the Jetpack block editor assets still enqueued. 

I'll make sure to test with p2 though, just in case.

tyxla: > Sure! :)

Thanks a bunch for the instructions and videos @BogdanUngureanu! I'll make sure to test those cases and to make sure we don't reintroduce that performance bottleneck.

jsnajdr: The "if is admin then `enqueue_block_assets` otherwise `enqueue_block_editor_assets`" condition is present also in many themes shipped by Core, like here:

https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-content/themes/twentytwenty/functions.php#L440-L444

But it also has an extra "WordPress version is 6.3 or newer" condition. So, something about the behavior of these two actions must have changed in 6.3. @BogdanUngureanu Maybe Jetpack wants to check the 6.3 condition, too? I think it's worth investigating.

jsnajdr (approved):
This works well for me :+1:

Two further changes that should be considered as a follow-up:
1. Instead of trying to dequeue the `jp-tracks` and `jetpack-blocks-editor` scripts, the hook should just `return`. Nobody else should be enqueuing these scripts. And if some hook does, it means that it has the same bug as is being fixed here, and should include the `wp_should_load_block_editor_scripts_and_styles` condition, too.
2. Maybe the `Jetpack::enqueue_block_editor_assets` method shouldn't enqueue `jetpack-blocks-editor` at all? The enqueuing is already triggered by the `editorScript: 'jetpack-blocks-editor'` field in the Jetpack blocks' `block.json` files.

The block loading code is quite convoluted, and looking at this could help clarify and simplify it.

projects/plugins/jetpack/class.jetpack-gutenberg.php
tyxla: @jeherve please correct me if I'm wrong, but IIRC backward compatibility policies of Jetpack with WP core were "current and previous major versions", right? While for themes (and especially core themes), it's a bigger deal, since I expect we strive to support older versions of WordPress.

So, if that's the backward compatibility support for Jetpack, and I understand this correctly, we shouldn't be concerned about WP older than 6.3.

jeherve: > we shouldn't be concerned about WP older than 6.3.

That's correct. Jetpack requires WordPress 6.3 and newer since #34127.

tyxla: > Open the Site Editor and then the Global Styles panel
> Go to Typography > Text > Size input.
> Toggle the input to a slider (the icon in the right)
> Move the slider left and right
> You'll notice that the slider UI responds slower before the change I've introduced
> You can see a difference in response time in these two videos:
> 
> In the video there are different types of sliders, for the one I've mentioned above, the difference should be more visible because the inputs change faster with a regular slider.

Thanks again for the videos and instructions, @BogdanUngureanu, I've just verified that the performance of various sliders in global styles is fine with this patch. 👍

tyxla:
I've given this a bit more testing today, and it worked well for me, including the additional scenarios for Global Styles that came up in the meantime, with the latest WP core and with/without Gutenberg. 

@jeherve is there anything else you'd recommend before shipping this one? I see you added the "[Status] Needs Team Review" label, but I wondered what it means - @jsnajdr is from my team and he approved it already, does it mean we can go ahead and ship? Thanks in advance!

jeherve:
> I see you added the "[Status] Needs Team Review" label, but I wondered what it means - @jsnajdr is from my team and he approved it already, does it mean we can go ahead and ship

Yep, exactly! :) As long as you have one approval and the green button, you're good to go :)

projects/plugins/jetpack/class.jetpack-gutenberg.php
jsnajdr: If Jetpack supports only WordPress 6.3+ that's great, because it means that the condition:
```php
if ( is_admin() ) {
  add_action( 'enqueue_block_assets', array( 'Jetpack_Gutenberg', 'enqueue_block_editor_assets' ) );
} else {
  add_action( 'enqueue_block_editor_assets', array( 'Jetpack_Gutenberg', 'enqueue_block_editor_assets' ) );
}
```
can be simplified to just:
```php
if ( is_admin() ) {
  add_action( 'enqueue_block_assets', array( 'Jetpack_Gutenberg', 'enqueue_block_editor_assets' ) );
}
```

What happened in WordPress 6.3 is that the [post editor got iframed](https://make.wordpress.org/core/2023/07/18/miscellaneous-editor-changes-in-wordpress-6-3/). Without the iframe, all styles and scripts were loaded in the top frame, in the `enqueue_block_editor_assets` action. That's the `else` part of the condition and it's no longer relevant because the action is never executed in a non-admin context.

With WordPress 6.3+ and always-iframed editor, we can target either the
- top editor frame with `enqueue_block_editor_assets`
- top editor frame with `enqueue_block_assets` and `is_admin() && wp_should_load_block_editor_scripts_and_styles()`
- editor iframe content with `enqueue_block_assets` and `is_admin() && ! wp_should_load_block_editor_scripts_and_styles()`
- site frontend with `enqueue_block_assets` and `! is_admin()`.

The `enqueue_block_assets` action runs everywhere, and should enqueue assets, like common block styles, that should be both on frontend and in the editor. Inside the callback you can find out where you are by calling `is_admin()` and `wp_should_load_block_editor_scripts_and_styles()`.

tyxla: Thanks for the research, @jsnajdr!

This also that there are opportunities for follow-ups here, particularly for renaming methods and moving some code so it is more predictable. For example, we could have 2 separate methods - `enqueue_block_editor_assets()` and `enqueue_block_assets()` and each would be hooked to its corresponding action. Since my goal is to fix the performance issue (and potentially more of them as I move forward), I'm not planning to work on that, but it's a worthy enhancement in favor of code comprehensibility.

PR #34271, opened 2023-11-23
RNMobile: Revert set default for video

SiobhyB:
Reverts https://github.com/Automattic/jetpack/pull/34073 to prevent a default src value being set for all video blocks added on mobile.

## Proposed changes:

* Following the changes in https://github.com/Automattic/jetpack/pull/34073, a default src is always set to any video block that's added via mobile. This change reverts that bug.
* Internal discussion here: p1700755582478559/1700745262.003959-slack-C0126N6JH7Z

### Other information:

## Jetpack product discussion

N/A

No, it does not.

## Testing instructions:

With this branch checked out, ensure it's possible to add a video block to the Android or iOS apps.


fluiddot (approved):
LGTM 🎊 ! I confirmed that after this revert, the bug related to adding the Video block is solved. Thanks!

PR #34274, opened 2023-11-23
Fix Paywall block unregister logic

fluiddot:
Fixes https://github.com/Automattic/jetpack/pull/33941/files#r1403640052.

## Proposed changes:

* Get block's name from `metadata` when invoking `unregisterBlockType`.

### Other information:

### Web version
* Create a page.
* Observe in the console that no errors are logged and no exceptions are encountered.
* **[UPDATE]** Observe that the Paywall block is not available (i.e. the block can't be added to the page).

### App version
* Open the app.
* Create a page.
* Observe that no exceptions/crashes are encountered.
* **[UPDATE]** Observe that the Paywall block is not available (i.e. the block can't be added to the page).

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-paywall-block-registration
+++ b/projects/plugins/jetpack/changelog/fix-paywall-block-registration
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
+Fix block name usage when unregistering Paywall block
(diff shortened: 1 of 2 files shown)


derekblank (approved):
I was able to test this feature branch successfully on both web and mobile, and did not note any exceptions or console errors. 🚀

fluiddot:
While testing the fix in the web version, I also noticed that before these changes, the Paywall block was available in Pages. As per this logic, this is unexpected as the Paywall block should be automatically unregistered. This fix will address it.

https://github.com/Automattic/jetpack/blob/8df317df876918638a758ed43dab4b8e2ac35b66/projects/plugins/jetpack/extensions/blocks/paywall/editor.js#L19-L22

simison:
Correct, Paywall block should be available only to posts; no pages nor custom post types either. Thanks for catching! cc @lezama

lezama:
Thanks!

PR #34287, opened 2023-11-24
Change isOverLimit and requireUpgrade evaluation for upgrade prompts

CGastrell:
WIP: Refactor use of isOverLimit and requireUpgrade

Depends on D129877-code

## Proposed changes:

This PR re-implements part of the data reduced on the store. See the diff as it uses a similar yet alternative approach: looks for the highest tier and compares to the current tier for the site. The reducer, as it lacks the list of tier plans, checks if it `isOverLimit` and `nextTier` is `null` to derive the value of `requireUpgrade`.

### Other information:

## Jetpack product discussion
p1700773172852749-slack-C054LN8RNVA

No

## Testing instructions:

There are at least two things we need to test:

- the backend change on the flags, to make sure they are returning the correct values
- the store manipulation on the client, to make sure we are setting the values correctly and using it correctly to show the upgrade messaging on the assistant

---

### Testing the backend change

- Apply this change to your sandbox since it touches a Jetpack file that runs on the backend
- Open the WordPress.com Console to perform requests to `wpcom/v2` `/sites/$wpcom_site/jetpack-ai/ai-assistant-feature`
- Consult the testing instructions for the `site_requires_upgrade` method on the D129877-code change and follow it, but instead of running the method, run a request using the WordPress.com Console; the expected results are the same as in the table listed on the patch

---

### Testing the client change

- Apply this change to your sandbox since it touches a Jetpack file that runs on the backend
- Run this branch on your testing site and make sure it's pointing to your sandbox
- Things we need to check:
   - if the upgrade nudge shows up on the correct situations
   - if the features are disabled/enable (due to requiring an upgrade) on the correct situations
- Consult the testing instructions for the `site_requires_upgrade` method on the D129877-code change and follow it, but instead of running the method, reload the editor on your testing site and check if the upgrade nudge is visible and if the features are enabled or disabled; the following table summarizes the expected behavior on all test scenarios:

| get_tier_licensed_quantity | get_current_period_requests_count | get_all_time_requests_count | upgrade nudge is hidden or visible? | features are enabled or disabled? |
| --- | --- | --- | --- | --- |
| 0 (free plan) | any | < 20 | hidden | enabled |
| 0 (free plan) | any | >= 20 | visible | disabled |
| 1 (unlimited paid plane) | any | any | hidden | enabled |
| 100 | < 100 | any | hidden | enabled |
| 100 | >= 100 | any | visible | disabled |
| 250 | < 250 | any | hidden | enabled |
| 250 | >= 250 | any | visible | disabled |
| 500 | < 500 | any | hidden | enabled |
| 500 | >= 500 | any | visible | disabled |
| 750 | < 750 | any | hidden | enabled |
| 750 | >= 750 | any | visible | disabled |
| 1000 | < 1000 | any | hidden | enabled |
| 1000 | >= 1000 | any | visible (soft-limit in place, not requiring the upgrade to continue using) | enabled |


lhkowalski:
I removed the comments on the code that handles the `ACTION_SET_AI_ASSISTANT_FEATURE_REQUIRE_UPGRADE` action, since the action is still dispatched on a portion of the code, when we receive a 429 response on the completion, meaning we indeed require an upgrade.

I changed the code, though, because it had wrong assumptions about the `hasFeature` flag and coupled `isOverLimit` with the behavior of `requireUpgrade`, that is not true anymore given the last changes:
- I dropped the optimistic update of the `isOverLimit` flag when `requireUpgrade` is false, because we can't know for sure if it's false too when we are in the last tier
- I kept the optimistic update of the `isOverLimit` flag when `requireUpgrade` is true, because in this case we know for sure that `isOverLimit` is also true

lhkowalski (approved):
Code looks good!

PR #34289, opened 2023-11-24
Fix | Publicize remains ON after the post is published

manzoorwanijk:
After #34064, there is some regression that resulted in publicize remaining ON even after the post is published. The reason for that is when a post is published, the API call returns `jetpack_publicize_connections` in the post meta with `enabled` set to `false` for all the connections. But, in #34064, we changed the source of truth for connections from post meta to social store thus the issue.

## Proposed changes:
* Sync the connections from post attributes to the social store after a post is published

### Other information:

## Testing instructions:

* For a site with Jetpack Social enabled, add some social connections
* Create a new post
* Publish the post 
* Open the Jetpack/Social sidebar
* Confirm that publicize and all the connections are disabled

Screenshot 2023-11-24 at 2 24 52 PM


projects/js-packages/publicize-components/src/components/sync-post-data-to-store/index.tsx
```diff
@@ -0,0 +1,8 @@
+import { useSyncPostDataToStore } from './use-sync-post-data-to-store';
+
+/**
+ * Renders the hooks to sync post data to the social store.
+ */
+export function SyncPostDataToStore() {
+	useSyncPostDataToStore();
```
gmjuhasz: Cant' we just use this hook where we add the component to the tree?

gmjuhasz (commented):
This works fine, just had a question, if we could do it by not having to add an empty component that just calls the hook, and just use the hook itself

projects/js-packages/publicize-components/src/components/sync-post-data-to-store/index.tsx
manzoorwanijk: We can't call the hook conditionally and the social module may not be enabled.

projects/js-packages/publicize-components/src/hooks/use-saving-post/index.js
```diff
@@ -30,7 +30,7 @@ export function usePostJustSaved( fn, deps ) {
  * Additionally, it accepts a dependency array which is passed to useEffect hook.
  *
  * @param {Function} fn - Callback function to run when the post is just published.
- * @param {Array} deps  - Depencency array.
+ * @param {Array} [deps]  - Depencency array.
```
pablinos, line 33: That's interesting, isn't `deps` the array here? We're not destucturing it to get `deps` as an element from the array. We're also not using it I think! Shouldn't we pass that on or remove the parameter?

projects/js-packages/publicize-components/src/social-store/resolvers.js
```diff
@@ -70,11 +69,11 @@ export function* getAutoConversionSettings() {
  * @returns {Function} Resolver
  */
 export function getConnections() {
-	return function ( { dispatch } ) {
+	return function ( { dispatch, registry } ) {
```
pablinos, line 72: This clearly works but I'd like to understand more about it. I've not seen resolvers used like this before, and thought they should return action objects as a result of a generator or a promise, not a function.

projects/plugins/social/src/js/editor.js
```diff
@@ -67,6 +68,7 @@ const JetpackSocialSidebar = () => {
 
 	return (
 		<PostTypeSupportCheck supportKeys="publicize">
+			<SyncPostDataToStore />
```
pablinos: We could use the hook directly here, as there are no conditionals. perhaps if we could reorganise the code in the Jetpack version so that the Publicize sidebar panels form their own component that is not rendered if the module is loading/disabled, then we could avoid the conditional hook problem.

pablinos (commented):
The PR needs a trunk merge because I took so long reviewing it, sorry!

I had a couple of thoughts, but otherwise it's ok

projects/js-packages/publicize-components/src/social-store/resolvers.js
pablinos: Ah it's a thunk! I assume the registry is needed to make testing easier?

projects/js-packages/publicize-components/src/hooks/use-saving-post/index.js
manzoorwanijk: Wrapping it in square brackets means it's optional, not required. 😁

projects/js-packages/publicize-components/src/social-store/resolvers.js
manzoorwanijk: Yes, it's a thunk. Yes, registry makes testing easier.

projects/plugins/social/src/js/editor.js
manzoorwanijk: That will need some refactor which we can do in a separate PR. Then we can get rid of that component and use the hook directly.

gmjuhasz (approved):
Thanks for refactoring this, I think it's a lot better like this without the code smell

projects/js-packages/publicize-components/src/hooks/use-saving-post/index.js
pablinos: TIL!

PR #34290, opened 2023-11-24
Refactor WPCOM_JSON_API_Get_Post_v1_1_Endpoint to allow code access

daledupreez:
## Proposed changes:

* This PR refactors `WPCOM_JSON_API_Get_Post_v1_1_Endpoint` to expose the underlying post retrieval logic via a new `fetch_post()` helper function.
* The main driver for this change was some work on WordPress.com, where we have some code calling a subclass via REST to get this content, but where it would be much more effective to simply call relevant code APIs to get the post content as we're already on the server.
  - I am working on the bulk of the WordPress.com refactoring/changes in D128370-code, which no longer includes any Jetpack changes to simplify checking the changes into WordPress.com, but I had the code from this PR running during some earlier testing. You can see that earlier version of the code along with the other server-side refactors in D128370-code?vs=on&id=455554

### Other information:

## Jetpack product discussion

This should have no impact on the Jetpack functionality, so I haven't engaged in a product discussion yet.

No changes to data we track or activity tracking.

## Testing instructions:

* Identify a WordPress.com Simple site and a Jetpack-connected site (which may be a WordPress.com Atomic site), and identify at least one post ID from each site
* Follow the instructions in [this comment](https://github.com/Automattic/jetpack/pull/34290#issuecomment-1825399301) to apply the changes to your WordPress.com sandbox system
* Ensure that requests for `public-api.wordpress.com` are routed to your WordPress.com sandbox
* Open the WordPress.com developer console: https://developer.wordpress.com/docs/api/console/
* Ensure that the dropdowns have the following values: `WP.COM REST API`, `v1.1`, and `GET`
* In the request text entry bar, type in `/sites/$site/posts/$post_ID` and select the relevant GET item that's shown in the dropdown results
* For each of your site and post combinations, replace the `$site` and `$post_ID` values with the relevant site ID and post ID combination, and hit Return to trigger a request. Inspect the results for each, and note the post slug for each post.
* Stop routing requests for `public-api.wordpress.com` to your sandbox, and repeat the requests above.
* Verify that the returned data are identical for each request.
* Re-enable routing of requests for `public-api.wordpress.com` to your sandbox.
* Update the request entry bar to be `/sites/$site/posts/slug:$post_slug`, and select the matching entry in the dropdown
* Using the same site IDs as before, and the post slugs you noted earlier, replace `$site` and `$post_slug` with the various combinations.
* Verify that you get results back for all the calls.
* Stop routing requests for `public-api.wordpress.com` to your sandbox, and repeat the requests above.
* Verify that the returned data are identical for each request.


projects/plugins/jetpack/json-endpoints/class.wpcom-json-api-get-post-v1-1-endpoint.php
```diff
@@ -63,33 +63,51 @@ public function callback( $path = '', $blog_id = 0, $post_id = 0 ) {
 
 		$args = $this->query_args();
 
-		$site = $this->get_platform()->get_site( $blog_id );
-
 		if ( str_contains( $path, '/posts/slug:' ) ) {
+			$site = $this->get_platform()->get_site( $blog_id );
```
sixhours, line 67: I am assuming this moved because we don't use `$site` in additional contexts/elsewhere in this method?

projects/plugins/jetpack/json-endpoints/class.wpcom-json-api-get-post-v1-1-endpoint.php
```diff
@@ -63,33 +63,51 @@ public function callback( $path = '', $blog_id = 0, $post_id = 0 ) {
...
+	/**
+	 * Helper function to fetch the content of a post. User validation
+	 * should be handled by the caller.
+	 *
+	 * @param int    $blog_id The blog ID for the post.
+	 * @param int    $post_id The post ID.
+	 * @param string $context The context we're fetching for.
+	 * @return array|SAL_Post|WP_Error
+	 */
+	public function fetch_post( $blog_id, $post_id, $context ) {
```
sixhours, line 87: Nit, the naming throughout is a little confusing; are we fetching the post or the post content? Maybe `fetch_post_content`/`$post_content` for clarity (or update the comment, either way).

sixhours (approved):
Working for me! Had some minor q's, nothing blocking. 👍

projects/plugins/jetpack/json-endpoints/class.wpcom-json-api-get-post-v1-1-endpoint.php
daledupreez: That's a good question. I _think_ sticking with `fetch_post()` may be better than `fetch_post_content()`, because in my mind "post content" refers to the content of the post (i.e. `$post->post_content`), where we are returning data that's more aligned with a `WP_Post` object.

projects/plugins/jetpack/json-endpoints/class.wpcom-json-api-get-post-v1-1-endpoint.php
daledupreez: Yes -- prior to moving all the other logic to `fetch_post()` we did have other uses of `$site`, but now that we've moved that code, the only remaining place we need `$site` is inside this if block.

daledupreez:
Noting for posterity that I also verified that this change correctly interacts with the changes from D128370-code.

PR #19173, opened 2021-03-15
Use `composer update` rather than `install` when lockfile isn't checked in.

anomiex:
Partially fixes #18753

#### Changes proposed in this Pull Request:

* The CLI needs to detect whether the lockfile is checked in or not.
* It also has a bug where it wasn't doing the installs concurrently like it intended to.
* All the packages' `composer.json` should use `@composer update` in the scripts.

This PR doesn't fix the cli `generate` command's composer.json, as it currently doesn't seem to have logic determining whether composer.lock is checked in or not. Thus this doesn't completely fix #18753.

#### Jetpack product discussion

#18753

#### Does this pull request change what data or activity we track or use?

No.

#### Testing instructions:

* Try `jetpack install` for projects with and without the lock files checked in, and for those without it checked in try it with or without the lock file existing. Ensure the correct command was run.

PR #34297, opened 2023-11-24
Backup: Rewrite Helper_Script_Manager class, add tests

pypt:
Rewrite `Helper_Script_Manager` class, add tests.

## Proposed changes:

* **Improved error reporting.**

    In all cases, if the helper script can't be installed to any of the candidate locations (or cleaned up from them), we'll now get a specific reason as to why.

* **Started reporting back I/O errors.**

    If the helper script can't be installed/deleted due to an I/O reason (e.g. inadequate permissions, or no disk space left), the failure will be reported back as such, together with the specific I/O reason as reported by the PHP warning (if any).

* **Switched from using `WP_Filesystem` to native PHP functions.**

    We're installing executable code here, so if `WP_Filesystem` puts our script on a FTP/SSH server somewhere, it won't be reachable and won't work.

    Also, `WP_Filesystem`'s error reporting capabilities are limited.

* **Made it try installing the helper script to `wp-content/` and `wp-content/uploads/` before WordPress root.**

    Both `wp-content/` and `wp-content/uploads/` are expected to be writable more often, and constructing URLs that point to those locations is supposed to be more predictable.

    Meanwhile, `ABSPATH` (WordPress root) might sometimes point to weird locations on certain hosting providers, and/or not be writable for security reasons.

* **Merged two duplicate `Helper_Script_Manager` class implementations into one.**

    `backup` and `transport-helper` packages had two nearly identical copies of the `Helper_Script_Manager` class.

    I've merged them into one and moved said class to a new `backup-helper-script-manager` package (that `backup` and `transport-helper` packages now depend on).

* **Fixed helper script cleanup code.**

    A regular expression in the cleanup code had a typo:

    https://github.com/Automattic/jetpack/blob/8ecbce229b9038e4f03d303c8bf7418ff8e57523/projects/packages/backup/src/class-helper-script-manager.php#L165

    (note the missing `.` before `*`.), so `cleanup_helper_scripts()` wasn't cleaning up much before.

* **Added a check for `[wp_path]` marker.**

    If the incoming helper script doesn't have the `[wp_path]` marker, installation will fail.

* **Made `delete_helper_script()` return `WP_Error` on failures.**

    If the cleanup of the helper script doesn't go through, we'll get back an HTTP 5xx response complete with a JSON reason why it failed.

* **Made it clean up empty `jetpack-temp` directories.**

    If a helper script cleanup operation leads to `jetpack-temp` being empty (barring the `index.php` and `README` files), it will now get cleaned up (which didn't always happen before).

* **`install_helper_script()` now always returns `abspath` key.**

    Many callers of `install_helper_script()` added `abspath` to the result, so I've added `abspath` in `install_helper_script()` itself.

* **Added tests.**

    A bunch of helper script installation scenarios now get tested via unit tests.

### Other information:

## Jetpack product discussion

peaFOp-20N-p2

No.

## Testing instructions:

Run a Jetpack VaultPress Backup backup without credentials set up, so that the tool is then forced to upload the helper script via Jetpack API and thus trigger this code.


mavegaf:
I created a site using this Jetpack branch version and ran some happy tests. Verified the `jp-helper-<random>.php` file is created ok. I would have liked to find a way to trigger some issues to see how the errors are displayed (I guess in the Transport Server), but I couldn't figure out how to simulate them

pypt:
Those errors will end up on our Logstash, e.g. try searching for an infamous `*install_faied*` (sic) with a `feature=vaultpress-transport` and `severity=error` filters.

This patch gets rid of generic and vague "Failed to install Helper Script" error messages, and tries its best to send back the things that were tried and the reasons why the installation didn't work, e.g. instead of `install_failed`, we'll now be getting something along the lines of:

```json
{"code":"all_locations_failed","message":"Unable to write the helper script to any install locations; tried: directory '/var/www/html/' (URL 'https://www.example.com/'): <reason why it couldn't create jetpack-temp and write the helper script there, e.g. 'permission denied'); directory '/var/www/html/wp-content/' (URL 'https://www.example.com/wp-content/'): <some other reason why it didn't manage to write to wp-content>; ...","data":null}
```

pypt:
IDK why *WordPress.com Tests* are failing, they were working just yesterday. Something to do with `jetpack-sync/src/modules/class-posts.php` I suppose.

pypt:
In 876012f, I've made it try installing the helper script to `wp-content/` and `wp-content/uploads/` before WordPress root.

Both `wp-content/` and `wp-content/uploads/` are expected to be writable more often, and constructing URLs that point to those locations is supposed to be more predictable.

Meanwhile, `ABSPATH` (WordPress root) might sometimes point to weird locations on certain hosting providers, and/or not be writable for security reasons.

Ping @jwebbdev

projects/packages/backup-helper-script-manager/CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## 0.1.0 - 2023-12-06
```
anomiex: I note this "release" won't actually get released, since you also have `projects/packages/backup-helper-script-manager/changelog/modify-backup-helper-script-manager-tests-verbosity` being created.

projects/packages/backup-helper-script-manager/CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
...
+### Added
+- Initial release.
+
+### Fixed
+- Use `composer update` rather than `install` in scripts, as composer.lock isn't checked in.
```
anomiex: This seems like it was copied from somewhere, we did that a long time ago and later since then we removed running `composer update` or `composer install` from the scripts entirely.

projects/packages/backup-helper-script-manager/src/class-helper-script-manager.php
```diff
@@ -0,0 +1,584 @@
...
+namespace Automattic\Jetpack\Backup;
+
+use Exception;
+
+/**
+ * Helper_Script_Manager manages installation, deletion and cleanup of Helper Scripts
+ * to assist with backing up Jetpack Sites.
+ *
+ * Does *not* use WP_Filesystem, because the helper script is something that we'll call over HTTP, so we don't want it
+ * to end up on an FTP server somewhere. Also, PHP provides us with better error reporting than WP_Filesystem.
```
anomiex: Reading this, I feel like it misses the point of `WP_Filesystem`. As I understand it `WP_Filesystem` exists to allow for adding files visible via HTTP to sites where the webserver doesn't have write access to the filesystem, not for uploading to some separate FTP server.

Perhaps I'm missing something? Can the comment be improved?

projects/packages/backup-helper-script-manager/src/class-helper-script-manager.php
```diff
@@ -0,0 +1,584 @@
...
+	/**
+	 * Lines that will be written to README in the helper directory.
+	 */
+	const README_LINES = array(
+		'These files have been put on your server by Jetpack to assist with backups, restores, and scans of your ' .
+		'site content. They are cleaned up automatically when we no longer need them.',
+		'If you no longer have Jetpack connected to your site, you can delete them manually.',
+		'If you have questions or need assistance, please contact Jetpack Support at https://jetpack.com/support/',
+		'If you like to build amazing things with WordPress, you should visit automattic.com/jobs and apply to join ' .
+		'the fun – mention this file when you apply!;',
```
anomiex: Is the semicolon at the end intended to be part of the message?

projects/packages/backup-helper-script-manager/src/class-helper-script-manager.php
```diff
@@ -0,0 +1,584 @@
...
+		$readme_path = trailingslashit( $temp_dir ) . 'README';
+		Throw_On_Errors::t_file_put_contents( $readme_path, implode( "\n\n", static::README_LINES ) );
+
+		$index_path = trailingslashit( $temp_dir ) . 'index.php';
+		Throw_On_Errors::t_file_put_contents( $index_path, static::INDEX_FILE );
+
+		$file_key  = wp_generate_password( 10, false );
+		$file_name = 'jp-helper-' . $file_key . '.php';
+		$file_path = trailingslashit( $temp_dir ) . $file_name;
+		Throw_On_Errors::t_file_put_contents( $file_path, $script_body );
```
mavegaf: In the current implementation, it is validated if the helper file [already exists ](https://github.com/Automattic/jetpack/blob/739375073fc0a3976b296db736c821c6336af7bb/projects/packages/transport-helper/src/class-helper-script-manager.php#L70). Should we add that validation here too to prevent override it?

projects/packages/backup-helper-script-manager/src/class-helper-script-manager.php
```diff
@@ -0,0 +1,584 @@
...
+				'than the max. size (' . static::MAX_FILESIZE . ' bytes)',
+				array( 'status' => 413 )
+			);
+		}
+
+		// Replace '[wp_path]' in the Helper Script with the WordPress installation location. Allows the Helper Script
+		// to find WordPress.
+		$wp_path_marker = '[wp_path]';
+		try {
+			$normalized_abspath = addslashes( Throw_On_Errors::t_realpath( ABSPATH ) );
```
anomiex: `addslashes()` is a poor escaping function for what seems like this use case. For any kind of string that the `[wp_path]` token is inside, it will do the wrong thing for some input.

Ideally you should probably inject an entire statement like `define( WP_PATH, '...' );` at the top of the script, with the `'...'` bit being produced by `var_export()`, and then use that constant in the script itself.

Failing that, you should probably reject any ABSPATH containing `'` or `"` or `$` or ASCII NUL as input here, so all `addslashes()` will do is escape `\`. Nowdocs seem unlikely and the other types of strings all handle escaped backslashes in the same way.

projects/packages/backup-helper-script-manager/src/class-throw-on-errors.php
```diff
@@ -0,0 +1,497 @@
...
+			// On PHP 7.x, all Exception and Error are subclasses of Throwable.
+			try {
+				$result = $callable();
+			} catch ( Throwable $throwable ) {
+				$error_message = $throwable->getMessage();
+			} catch ( Exception $exception ) {
+				$error_message = $exception->getMessage();
+			}
+		} else {
+			// On PHP 5.x, there's only Exception.
```
anomiex: We don't support 5.x anymore. The `composer.json` you're adding here specifically requires PHP >=7.0.

Even when we did, it never hurt to `catch ( Throwable )`, it would just never have caught anything.

projects/packages/backup-helper-script-manager/src/class-throw-on-errors.php
```diff
@@ -0,0 +1,497 @@
...
+		$result        = null;
+		$error_message = null;
+		if ( PHP_MAJOR_VERSION >= 7 ) {
+			// On PHP 7.x, all Exception and Error are subclasses of Throwable.
+			try {
+				$result = $callable();
+			} catch ( Throwable $throwable ) {
+				$error_message = $throwable->getMessage();
+			} catch ( Exception $exception ) {
+				$error_message = $exception->getMessage();
```
anomiex: This will never be reached, since the previous catch for Throwable will always do so instead.

This is the kind of code you'd write to also support PHP 5, but for some reason you're only running it when `PHP_MAJOR_VERSION >= 7`.

projects/packages/backup-helper-script-manager/src/class-throw-on-errors.php
```diff
@@ -0,0 +1,497 @@
+<?php // phpcs:disable Squiz.Commenting.FileComment.Missing
+
+// phpcs:disable WordPress.PHP.DevelopmentFunctions.prevent_path_disclosure_error_reporting
+// phpcs:disable WordPress.PHP.DiscouragedPHPFunctions.runtime_configuration_error_reporting
+// phpcs:disable WordPress.PHP.IniSet.display_errors_Disallowed
+
+namespace Automattic\Jetpack\Backup;
+
+use Exception;
```
anomiex, line 9:
```suggestion
use Exception;
use Throwable;
```

projects/packages/backup-helper-script-manager/src/class-throw-on-errors.php
```diff
@@ -0,0 +1,497 @@
...
+			function () use ( $filename, $data ) {
+				// phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_operations_file_put_contents
+				return file_put_contents( $filename, $data );
+			},
+			$label
+		);
+
+		if ( false === $file_put_contents_result ) {
+			throw new Exception( "Unable to $label" );
+		}
```
anomiex, line 449: No check for `$file_put_contents_result < strlen( $data )`?

projects/packages/backup-helper-script-manager/tests/php/test-class-throw-on-errors.php
```diff
@@ -0,0 +1,504 @@
+<?php // phpcs:disable WordPress.Files.FileName.InvalidClassFileName
+
+// phpcs:disable Generic.Commenting.DocComment.MissingShort
+// phpcs:disable Squiz.Commenting.FileComment.Missing
+// phpcs:disable Squiz.Commenting.FunctionComment.EmptyThrows
```
anomiex, line 5: You could do like `projects/packages/my-jetpack/tests/php/.phpcs.dir.xml` instead of having to manually exclude a bunch of stuff in the tests.

projects/packages/backup/actions.php
```diff
@@ -23,7 +25,8 @@
 }
 
 // Clean up expired Helper Scripts from a scheduled event.
-$add_action( 'jetpack_backup_cleanup_helper_scripts', array( 'Automattic\\Jetpack\\Backup\\Helper_Script_Manager', 'cleanup_expired_helper_scripts' ) );
+$helper_script_manager = new Helper_Script_Manager();
```
anomiex: Instantiating the class here is probably not a good idea. It'll probably force auto-loading the class even if a later plugin is going to provide a later version.

projects/packages/transport-helper/actions.php
```diff
@@ -23,7 +25,8 @@
 }
 
 // Clean up expired Jetpack Helper Scripts from a scheduled event.
-$add_action( 'jetpack_cleanup_helper_scripts', array( 'Automattic\\Jetpack\\Transport_Helper\\Helper_Script_Manager', 'cleanup_expired_helper_scripts' ) );
+$helper_script_manager = new Helper_Script_Manager();
```
anomiex: Same.

anomiex (commented):
Gave it a once-over. Didn't fully review.

mavegaf:
I have done some tests on sites with different configurations, and they worked. I saw some 301, but it seems it was related to my installation, since reinstalling the plugin the errors were gone.

I left a question about verifying if a helper exists to prevent clashes. It is probably very improbable, but it could be an unpredictable behavior if that happens

Can this change affect/be affected by Jetpack Backup plugin? I tested using the same branch for the Jetpack Backup, and I didn't find problems, but I'm not sure how they are related.

projects/packages/backup-helper-script-manager/src/class-helper-script-manager.php
pypt: It's very unlikely that it will already exist (1 out of 62^10 chance on case-sensitive filesystems, such as ext4 by default, or 1 out of 36^10 chance on filesystems mounted as case-insensitive), so I've omitted that check originally, but nonetheless I've readded that check in 7c9ceeebeca4ef4d1bb5cf3dd8fe77c56c3df44b.

projects/packages/backup-helper-script-manager/src/class-helper-script-manager.php
pypt: My understanding is that `WP_Filesystem` is a mechanism that allows one to serve various resources (e.g. media files) from an FTP/SFTP server (given the fact that subclasses `WP_Filesystem_FTPext`, `WP_Filesystem_ftpsocket`, `WP_Filesystem_SSH2` exist, in addition to `WP_Filesystem_Direct` which is just a thin layer over the likes of `file_put_contents()`).

Our helper script is not a resource, but rather an executable PHP file which we intend to be able to access over HTTP after uploading it, so we *do not* want it to accidentally end up on an FTP server somewhere where it won't be accessible to us.

Something like a WordPress plugin editor doesn't use `WP_Filesystem` either, I think it's for the same reason:

https://github.com/WordPress/WordPress/blob/a2bcc36278ea932406a3e0044a6d9f96c8493c10/wp-admin/plugin-editor.php#L167-L171

projects/packages/backup-helper-script-manager/src/class-helper-script-manager.php
pypt: Thanks, possibly this could be improved later, but here I'm just preserving how it was all being done before, in an attempt to not break anything.

Changing up how that `[wp_path]` gets injected would involve us having to change the helper script itself, then have/support two versions of the helper while the plugin changes get gradually rolled out, etc.

I don't see this as a security issue either, as the user with NUL bytes in their `ABSPATH` just won't get their website backed up/restored. So, not quite ready to do all of this now.

projects/packages/backup-helper-script-manager/src/class-helper-script-manager.php
pypt: Thanks, removed in d58cc1d.

projects/packages/backup-helper-script-manager/src/class-throw-on-errors.php
pypt: How come that [our plugin says that we do](https://wordpress.org/plugins/jetpack-backup/)?

Screenshot 2023-12-12 at 14 02 10

projects/packages/backup-helper-script-manager/src/class-throw-on-errors.php
pypt: Thanks, I've assumed that somehow `file_put_contents()` would bother enough to throw a warning if it doesn't manage to write everything, but that might not always be the case.

Added a check in 883a2c9.

pypt:
> Can this change affect/be affected by Jetpack Backup plugin? I tested using the same branch for the Jetpack Backup, and I didn't find problems, but I'm not sure how they are related.

Could you elaborate? This _is_ the Jetpack VaultPress plugin, or is it not?

projects/packages/backup-helper-script-manager/tests/php/test-class-throw-on-errors.php
pypt: I want to ignore those in just a single test file (but not the other), so I think it's more KISS that way.

projects/packages/backup/actions.php
pypt: Could you please elaborate?

I've remade that class into not being a static one to make it more testable, and the constructor doesn't do much of anything at all:

https://github.com/Automattic/jetpack/blob/883a2c962e2348d47507fd8ad3b8f4566cb179d8/projects/packages/backup-helper-script-manager/src/class-helper-script-manager.php#L76-L84

so I figure that it's cheap to just initialize an instance here.

projects/packages/backup-helper-script-manager/CHANGELOG.md
pypt: I did copy-paste it, as that line is everywhere:

Screenshot 2023-12-12 at 14 26 40

projects/packages/backup-helper-script-manager/CHANGELOG.md
pypt: Could you help me here by explaining how it all should be so that the "release" gets "released"?

Like (I would assume) most of A8C, I have no idea how that whole versioning/changelogger thing works/is supposed to work, so I just run `./tools/changelogger-validate-all.sh` repeatedly, and swear/cry/make random changes until it says that it's happy.

pypt:
> Gave it a once-over. Didn't fully review.

Thank you Brad, very useful review!

projects/packages/backup-helper-script-manager/src/class-throw-on-errors.php
jeherve: This is because the Jetpack Backup plugin was last updated on September 20. Since then, multiple changes were made to the plugin, including a bump to the PHP version requirements:
https://github.com/Automattic/jetpack-backup-plugin/blob/trunk/CHANGELOG.md#23-alpha---unreleased

When you'll want to ship the changes you're introducing in this PR to customers, you'll need to ship a new update to the plugin. That will include your changes as well as all the ones currently listed as unreleased, so the plugin will require PHP 7.0+

jeherve:
> Could you elaborate? This _is_ the Jetpack VaultPress plugin, or is it not?

The new package you're introducing, `Automattic/jetpack-backup-helper-script-manager`, is required by the `Automattic/jetpack-backup` package, which in turn is required by 4 different plugins:

- Jetpack
- Jetpack Backup
- Jetpack Protect
- Move to WordPress.com

So your changes can potentially impact all 4 of those plugins.

projects/packages/backup-helper-script-manager/src/class-throw-on-errors.php
pypt: Thanks, removed PHP 5.x hacks in 202ce2b.

> When you'll want to ship the changes you're introducing in this PR to customers, you'll need to ship a new update to the plugin. That will include your changes as well as all the ones currently listed as unreleased, so the plugin will require PHP 7.0+

How do we do it? Is this something that Crew / Garage does, or should we ship that plugin ourselves somehow?

projects/packages/backup-helper-script-manager/src/class-throw-on-errors.php
pypt: Thanks, removed PHP 5.x hacks in 202ce2b.

projects/packages/backup-helper-script-manager/src/class-throw-on-errors.php
pypt: Fixed in 202ce2b.

projects/packages/backup-helper-script-manager/src/class-throw-on-errors.php
jeherve: You can release an update whenever you'd like, by following the steps described in PCYsg-KCk-p2
You can also ask for help in #jetpack-releases if you get stuck at any point.

projects/packages/backup-helper-script-manager/src/class-helper-script-manager.php
mavegaf: Excellent. Thank you!

projects/packages/backup-helper-script-manager/src/class-helper-script-manager.php
anomiex: That's not a good example, as it's using `file_get_contents` (a read operation, if source files couldn't be read then even `require` would break) rather than a write operation like `file_put_contents`.

projects/packages/backup/actions.php
anomiex: The point isn't the instantiating the class, it's that instantiating the class first triggers the autoloader to _load_ the class file.

The way things normally work with the Jetpack Autoloader is something like this:

* Plugin A loads, registers its copy of the Jetpack Autoloader.
* Plugin B loads, registers its copy of the Jetpack Autoloader.
* Later on, something tries to autoload the class (e.g. the hook fires and calls the static method), and the Jetpack Autoloader chooses the latest version of the class from all registered versions.

If you have this instantiation here, it instead could happen like this.

* Plugin A loads, registers its copy of the Jetpack Autoloader.
* Plugin A then instantiates `Helper_Script_Manager`, triggering loading of its copy of the class file.
* Plugin B loads, registers its copy of the Jetpack Autoloader.
* Plugin B then instantiates the `Helper_Script_Manager` manager class, and gets Plugin A's version even if B's is newer because A already triggered the loading.

projects/packages/backup-helper-script-manager/CHANGELOG.md
anomiex: That line is "everywhere" because most of our packages were created before #19173 happened in March 2021 and so had that change. This package is new.

projects/packages/backup-helper-script-manager/CHANGELOG.md
anomiex: Happy to.

For packages (or really any project that doesn't have `.extra.release-branch-prefix` set in its `composer.json`), a release happens when a commit is pushed to the `trunk` or `prerelease` branch where there are no non-dotfiles in the project's `/changelog/` directory.

More specifically, the lack of non-dotfiles means that no "x.y.z-alpha" line gets added to `CHANGELOG.md` when mirroring the build project to its mirror repo. The fact that the top entry in `CHANGELOG.md` is not a prerelease version number (e.g. `-alpha`) triggers a workflow in the mirror that creates a git tag for that version number. The creation of the git tag does two things: (1) Packagist picks up that tag as a release, and (2) another workflow in the mirror repo runs that creates an GitHub release for the version.

So, for the initial version of a package, you can do one of two things:
1. Leave `CHANGELOG.md` empty to start, just the header part (lines 1–6), along with a file in `.../changelog/` with an appropriate entry such as "Initial release". Later on when someone does a release of the package (probably to go along with a release of a plugin using the package) an entry will be created.
2. Create the entry in `CHANGELOG.md` and do not add any file (other than `.gitkeep`) in `.../changelog/`. That should result in all the workflows described above running to create the 0.1.0 tag.

projects/packages/backup-helper-script-manager/src/class-helper-script-manager.php
anomiex: https://developer.wordpress.org/apis/filesystem/#purpose may be a useful explanation of what `WP_Filesystem` is for.

pypt: > That's not a good example, as it's using file_get_contents (a read operation, if source files couldn't be read then even require would break) rather than a write operation like file_put_contents.

I think one is supposed to fetch contents with `WP_Filesystem_Base::get_contents()`.

> https://developer.wordpress.org/apis/filesystem/#purpose may be a useful explanation of what WP_Filesystem is for.

Thanks, TIL, but it does not apply to us in this case, because if there are permissions issues between the webserver's user and the FTP/SSH user, then we'll just install the helper script and do a backup/restore using FTP/SSH credentials (that we collect ourselves), without using `WP_Filesystem` in any way.

Also, if we can't write that helper script somewhere (due to writes being inaccessible to the webserver's user, or for other reasons), we want to know about it (in the form of an error response), instead of having that helper script silently uploaded via FTP/SFTP, so that we could fallback to a backup/restore using credentials.

projects/packages/backup-helper-script-manager/CHANGELOG.md
pypt: Thanks, removed in 8a06cb4.

projects/packages/backup-helper-script-manager/CHANGELOG.md
pypt: OK, thanks, I've purged `CHANGELOG.md`, and edited the entry in `changelog/` in 53a4299. Does it look right?

projects/packages/backup/actions.php
pypt: Hmm.

What would be the way to ensure that plugin A loads its own version of `Helper_Script_Manager` class, and plugin B loads its own version too - not an older one, not a newer one?

projects/packages/backup-helper-script-manager/src/class-helper-script-manager.php
anomiex: > if there are permissions issues between the webserver's user and the FTP/SSH user, then we'll just install the helper script and do a backup/restore using FTP/SSH credentials (that we collect ourselves), without using WP_Filesystem in any way.
>
> Also, if we can't write that helper script somewhere (due to writes being inaccessible to the webserver's user, or for other reasons), we want to know about it (in the form of an error response), instead of having that helper script silently uploaded via FTP/SFTP, so that we could fallback to a backup/restore using credentials.

Thanks, that would be a good explanation to put in the comment instead. 🙂

projects/packages/backup-helper-script-manager/CHANGELOG.md
anomiex: That looks good to me!

projects/packages/backup-helper-script-manager/src/class-helper-script-manager.php
pypt: Thanks, updated comment in a77d279.

projects/packages/backup-helper-script-manager/CHANGELOG.md
pypt: Thank you Brad!

projects/packages/backup-helper-script-manager/src/class-throw-on-errors.php
pypt: Thank you, we'll have a look!

projects/packages/backup/actions.php
anomiex: You'd have to have each plugin have its own version of the class under different names, instead of both using `Automattic\Jetpack\Backup\Helper_Script_Manager`.

There are ways to try to rewrite Composer packages like that on installation (see some details in p6r3EZ-1YO-p2#comment-3498), but that's not something we've done in the monorepo and if we were to try to set something like that up we'd have to be careful not to have it affect everything relying on the current Jetpack Autoloader method.

Or you could have a custom script run during the build of each plugin that copies the one class file out of the package, modifies it, and writes the modified version into the plugin's directories somewhere. That could be tricky though when the new package isn't being used directly by the plugin, but instead by different packages used by the plugin.

Or you could stick with the status quo, where each use had its own copy-pasted version instead of having a package at all.

projects/packages/backup-helper-script-manager/src/class-helper-script-manager.php
```diff
@@ -0,0 +1,598 @@
...
+namespace Automattic\Jetpack\Backup;
+
+use Exception;
+
+/**
+ * Helper_Script_Manager manages installation, deletion and cleanup of Helper Scripts
+ * to assist with backing up Jetpack Sites.
+ *
+ * Does *not* use WP_Filesystem, because the helper script is something that we'll call over HTTP, so we don't want it
+ * to end up on an FTP server somewhere (even if it's the very same host). If there are permissions issues between the
```
anomiex: Thanks for the addition. I'd still suggest removing the old sentence that seems inaccurate though.
```suggestion
 * Does *not* use WP_Filesystem, because if there are permissions issues between the
```

projects/packages/backup/actions.php
pypt: Thanks, I'll add a static wrapper around `Helper_Script_Manager` so that autoloading gets delayed, and at least we could ensure that we always get the latest version instead of `rand()` version.

projects/packages/backup-helper-script-manager/src/class-helper-script-manager.php
pypt: Fixed in 9ed7a1a.

projects/packages/backup/actions.php
pypt: In 2342562, I've moved the implementation of the helper script manager to `Helper_Script_Manager_Impl` (to make it still testable with custom locations etc.), and added a static `Helper_Script_Manager` wrapper that gets lazy-late-loaded in `add_action()` and such.

projects/packages/transport-helper/actions.php
pypt: Fixed: https://github.com/Automattic/jetpack/pull/34297#discussion_r1424331608

pypt:
@mavegaf after some (I think rather minor) reshuffling, it's ready for another review (if you get a minute).

anomiex (commented):
Just want to say that this seems ok to me to merge and iterate on, but I didn't fully review so I'll leave that to the team.

pypt:
Thank you @anomiex, again, very productive review! Learned a lot!

pypt:
@mavegaf pointed out that if the helper script gets written to `wp-content/` (the new default) or `wp-content/uploads/`, the returned URL sometimes points to an `http://` location even though the website uses `https://`, thus requiring us to implement support for HTTP 301 redirects.

In 0b291c6, I've made it use `content_url()` (which seems to check for `is_ssl()` behind the scenes), and I've made it SSL-ize the `wp_upload_dir()['baseurl']` if needed.

mavegaf:
> @mavegaf pointed out that if the helper script gets written to `wp-content/` (the new default) or `wp-content/uploads/`, the returned URL sometimes points to an `http://` location even though the website uses `https://`, thus requiring us to implement support for HTTP 301 redirects.
> 
> In 0b291c6, I've made it use `content_url()` (which seems to check for `is_ssl()` behind the scenes), and I've made it SSL-ize the `wp_upload_dir()['baseurl']` if needed.

Thanks @pypt . I have tested it again, and everything looks good.

mavegaf (approved):
Approved by team ✅

pypt:
Thank you all!

PR #34070, opened 2023-11-10
Subscribe Block: Ignore token on simple sites if logged out

lezama:
Fixes p1699619742836339-slack-C02TCEHP3HA

## Proposed changes:

* When checking subscribe status ignore token from cookie on simple sites

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* Go to '..'
*

PR #34299, opened 2023-11-25
Refactor/wpcom token subscription service into jetpack token subscription service

chrismccluskey:
Fixes 197-gh-automattic/gold

## Proposed changes:

* Merges the logic from `WPCOM_Token_Subscription_Service` into `Jetpack_Token_Subscription_Service`.
* Move `available` logic from `WPCOM_Token_Subscription_Service` to `is_wpcom` function.
* Make `available` logic in `Jetpack_Token_Subscription_Service` into `$this->is_wpcom() || <existing logic>`.
* Move WPCOM-specific functions behind `if ( $this->is_wpcom ) { return ... }` clause.
* Rename all instances of `WPCOM_Token_Subscription_Service` to `Jetpack_Token_Subscription_Service'.
* Delete `WPCOM_Token_Subscription_Service`.

### Other information:

## Testing instructions:

* Apply this patch to your sandbox and sandbox your site (unless using JN - in which case setup constants and global http), public api, and subscribe.
* If running on JN, activate this branch.
* Try all payment blocks work properly when logged in, not logged in and a new subscriber, not logged in but just subscribed, etc..
* Be sure to try this on WPCOM simple if you've tried on JN or vice-versa. 
* Make sure there are no regressions.


projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-jetpack-token-subscription-service.php
```diff
@@ -41,10 +55,26 @@ public function get_site_id() {
...
 	}
+
+	/**
+	 * Returns true if the current authenticated user is subscribed to the current site.
+	 *
+	 * @return boolean
+	 */
+	public function is_current_user_subscribed() {
+		if ( $this->is_wpcom() ) {
+			return false;
```
millerf: Now that I am reading that, I am not sure about the `false` here 🤔

millerf: It traces back to https://github.com/Automattic/jetpack/pull/34070/files by @lezama 

Miguel, the PR description says 
```
Ignore token on simple sites if logged out
```

What if we modify to something like
```
if ( $this->is_wpcom() && !is_user_logged_in() ) {
	return false;
}
```

projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-jetpack-token-subscription-service.php
```diff
@@ -17,13 +17,23 @@
  */
 class Jetpack_Token_Subscription_Service extends Token_Subscription_Service {
 
+	/**
+	 * Is this code executing on WPCOM?
+	 *
+	 * @return bool True if executing on WPCOM.
+	 */
+	private function is_wpcom() {
```
lezama: there's a helper both in Jetpack and WPCOM that does this check that we can use instead

https://github.com/Automattic/jetpack/blob/df6ffd8d211e5596d28e525787cb50a0803f609b/projects/packages/status/src/class-host.php#L66

projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-jetpack-token-subscription-service.php
lezama: in think a better approach would be to define this method in all the wpcom concrete classes so this never gets hit in wpcom.

chrismccluskey: There's a `is_current_user_subscribed` function that overrides this one inside `WPCOM_Online_Subscription_Service`. I think my mistake here was adding this function at all. Instead of `return false`, it should just do what the parent function does. I'll remove this function altogether.

projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-jetpack-token-subscription-service.php
```diff
@@ -17,13 +18,22 @@
...
```
millerf: Could we just remove this and use `new Host() )->is_wpcom_simple()` everywhere else? It only adds unnecessary redundancy...

millerf (dismissed):
LGTM. Added a small request.

projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-jetpack-token-subscription-service.php
```diff
@@ -32,6 +33,10 @@ public static function available() {
 	 * @return int The site ID.
 	 */
 	public function get_site_id() {
+		if ( ( new Host() )->is_wpcom_simple() ) {
+			return get_current_blog_id();
+		}
+
 		return \Jetpack_Options::get_option( 'id' );
```
lezama: this works in WPCOM and Jetpack, I think there's no need for the previous `if`

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -807,7 +806,7 @@ function add_paywall( $the_content ) {
 	}
 
 	require_once JETPACK__PLUGIN_DIR . 'extensions/blocks/premium-content/_inc/subscription-service/include.php';
-	$token_service              = is_wpcom() ? new WPCOM_Token_Subscription_Service() : new Jetpack_Token_Subscription_Service();
+	$token_service              = is_wpcom() ? new Jetpack_Token_Subscription_Service() : new Jetpack_Token_Subscription_Service();
```
lezama: this `if` doesn't make much sense?  😆

I think we should use something like: https://github.com/Automattic/jetpack/blob/1b98ac60b97a2b3e31f42200bc89f34b656160fa/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/include.php#L43 whenever we need a `Jetpack_Token_Subscription_Service`

lezama: [like we do in Jetpack_Memberships](https://github.com/Automattic/jetpack/blob/1b98ac60b97a2b3e31f42200bc89f34b656160fa/projects/plugins/jetpack/modules/memberships/class-jetpack-memberships.php#L591-L592)

projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/include.php
```diff
@@ -70,8 +68,8 @@ function default_service( $service ) {
 		return new WPCOM_Online_Subscription_Service();
 	}
 
-	if ( WPCOM_Token_Subscription_Service::available() ) {
-		return new WPCOM_Token_Subscription_Service();
+	if ( Jetpack_Token_Subscription_Service::available() ) {
+		return new Jetpack_Token_Subscription_Service();
 	}
 
 	if ( Jetpack_Token_Subscription_Service::available() ) {
```
lezama: we are doing twice the same check

projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/include.php
```diff
@@ -70,8 +68,8 @@ function default_service( $service ) {
 		return new WPCOM_Online_Subscription_Service();
```
lezama, line 70: just to understand, what was the difference between `WPCOM_Online_Subscription_Service()`, `WPCOM_Token_Subscription_Service`?

chrismccluskey: `WPCOM_Token_Subscription_Service` is a class which `WPCOM_Online_Subscription_Service` inherited from. `WPCOM_Token_Subscription_Service` used to inherit from `Token_Subscription_Service`, just as `Jetpack_Token_Subscription_Service` does currently.

`WPCOM_Token_Subscription_Service` and `Jetpack_Token_Subscription_Service` were so similar that can be refactored into one class, which is what this PR does. That then makes `WPCOM_Online_Subscription_Service` inherit from `Jetpack_Token_Subscription_Service` instead of `Token_Subscription_Service`.

projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/include.php
chrismccluskey: Thanks, this was not intended.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
chrismccluskey: This can be simplified to `$token_service = new Jetpack_Token_Subscription_Service();` at least. I'll try to refactor it as you've suggested.

projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-jetpack-token-subscription-service.php
chrismccluskey: I'll try this. Previously, I had just copied this logic over from `WPCOM_Token_Subscription_Service` which uses `get_current_blog_id`. Ultimately this gets overridden by `WPCOM_Online_Subscription_Service` anyway, which uses `get_current_blog_id`.

projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/include.php
lezama: thanks for the explanation Chris!

projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-jetpack-token-subscription-service.php
lezama: > Ultimately this gets overridden by WPCOM_Online_Subscription_Service anyway, which uses get_current_blog_id.

I would definitely remove it as it will add confusion for future devs

millerf: TIL it works on simple as well.
```
> print \Jetpack_Options::get_option( 'id' )
29428530
> print_r(get_blog_details())
WP_Site Object
(
    [blog_id] => 29428530
    [site_id] => 1
    [...]
)
```

projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-wpcom-token-subscription-service.php
```diff
@@ -1,55 +0,0 @@
...
-
-	/**
-	 * Is get_key()
-	 *
-	 * @inheritDoc
-	 */
-	public function get_key() {
-	 // phpcs:ignore ImportDetection.Imports.RequireImports.Symbol
-		return defined( 'EARN_JWT_SIGNING_KEY' ) ? EARN_JWT_SIGNING_KEY : false;
-	}
```
chrismccluskey, line 45: It looks like the tests are expecting this code to be available. In this refactor, it's hidden behind `( new Host() )->is_wpcom_simple()`. The problem is that these tests run outside of WPCOM simple, so this `is_wpcom_simple` test will always be false. I think the easy solution here is to make the tests use `WPCOM_Online_Subscription_Service` rather that `Jetpack_Token_Subscription_Service`.

chrismccluskey: Doing this however creates new errors, because the code is now trying to include files from wpcom which don't exist in Jetpack. I'm starting to think that the reason `WPCOM_Token_Subscription_Service` existed in the firstplace was to allow for tests to be run without needing to create many mocks. The tests can just opt to use this intermediate `WPCOM_Token_Subscription_Service` class which then calls `get_key` and provides the a value of a constant - that constant likely already set to a mock value.

millerf: There are multiple ways to approach this:
- you should be able to mock `Host` and return true to `is_wpcom()` (this must be used somewhere in other tests)
- set `IS_WPCOM` for those tests (not recommended)
- set  `TESTING_IN_JETPACK` or any other var to bypass (see [this one](https://github.com/Automattic/jetpack/blob/b97d5af5848e42c1fb4f4272f0b1a9ca6a1c1aec/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php#L435))

I'd use the 1st or 3rd approach...

chrismccluskey: I've started exploring these approaches. I understand that some work will need to be done to refactor these tests if we truly want to make this change. I'll leave some notes here on the approaches as I work through the code on the work required so I can compare the approaches.

TL:DR; I went with `Test_Jetpack_Token_Subscription_Service`, and explained how I came to that conclusion.

**Problem**
Tests which assert that a user can access content under specific conditions are now reporting `false` when `true` was expected. This is due to an early `return false` statement because a key was not returned from `get_key`. We just need to return a string for these tests to pass.

**Mocking `Host`**
Actually mocking `Host` here might be difficult, since the test doesn't define which files get included and `Host` is already present. I checked to see how `Host` is tested, and it seems it uses approach 2 via `Constants::set_constant`.

**Setting `IS_WPCOM`** 
I assume with approach 2 that you mean setting the constant via `define( `IS_WPCOM`, true )`; if so I agree that this is not a good idea. However, `Host` can be manipulated to do what we want with functions like `Constants::set_constant()`, since `Host` uses the testable `Constants` class. I can use in tests, but doing so presents a new problem (one we'd still face even if we could easily mock `Host`): the tests weren't expecting to need to call `is_jetpack_site()` (which doesn't exist) but they now are.

We only want to set `IS_WPCOM` at a certain depth - just enough to get that key and then we want to clear the constant. After all, this is all being done just to get `get_key`, which in `WPCOM_Token_Subscription_Service` is simply returned when `get_key` is called. Now that we have merged the two, we need to short-cut just enough logic (`( new Host() )->is_wpcom_simple() && is_jetpack_site()`) in only `Jetpack_Token_Subscription_Service` and nowhere else. This is a difficult task, because outside of this one specific case where we `get_key`, all other calls to functions called by functions under test need to be run as if `IS_WPCOM` is not defined.

**Refactor the code to early return in certain conditions based on `TESTING_IN_JETPACK`**
Remember that the end goal here is to return `defined( 'EARN_JWT_SIGNING_KEY' ) ? EARN_JWT_SIGNING_KEY : false`. This code (inside `get_key`) is behind `if ( ( new Host() )->is_wpcom_simple() ) {`:

```
	public function get_key() {
		if ( ( new Host() )->is_wpcom_simple() ) {
			// phpcs:ignore ImportDetection.Imports.RequireImports.Symbol
			return defined( 'EARN_JWT_SIGNING_KEY' ) ? EARN_JWT_SIGNING_KEY : false;
		}
		$token = ( new Tokens() )->get_access_token();
		if ( ! isset( $token->secret ) ) {
			return false;
		}
		return $token->secret;
	}
```

I can add a `( defined( 'TESTING_IN_JETPACK' ) && TESTING_IN_JETPACK ) || ` before the check for `is_wpcom_simple`. This would fix the issue for this test, but then I figured it should present an issue when testing `Jetpack_Token_Subscription_Service` directly. However, when I searched for this specific class under test inside tests, I found `Test_Jetpack_Token_Subscription_Service` which `extends Jetpack_Token_Subscription_Service`.

I ran the phpunit tests for this, and they passed. However, knowing now about `Test_Jetpack_Token_Subscription_Service`, and another approach we could take is..

**Mock/Double Jetpack_Token_Subscription_Service**
This class extends the class we want to test, overwriting the `get_key` function to just return a random string, which is all we need for testing. This isn't a very good test if `Jetpack_Token_Subscription_Service` because nothing in that function is being tested. At least with `WPCOM_Token_Subscription_Service` was used, the `EARN_JWT_SIGNING_KEY` would be returned. This means everything  from `$token = ( new Tokens() )->get_access_token();` and below isn't being tested, but it also means that it never was covered in tests to begin with.

This has the downside of needing to change the tests to test `Test_Jetpack_Token_Subscription_Service`, but I those are changes to tests, and not changes to code, so I think that's better.

millerf: > Mock/Double Jetpack_Token_Subscription_Service

> but it also means that it never was covered in tests to begin with.
I don't think it was ever the point of testing this particular code. So all good.

I think it is a good approach.

millerf (dismissed):
LGTM

projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-jetpack-token-subscription-service.php
```diff
@@ -23,7 +24,7 @@ class Jetpack_Token_Subscription_Service extends Token_Subscription_Service {
 	 * @return bool Whether Jetpack_Options class exists.
 	 */
 	public static function available() {
-		return class_exists( '\Jetpack_Options' );
+		return ( new Host() )->is_wpcom_simple() || class_exists( '\Jetpack_Options' );
```
lezama, line 27: is there a case where `Jetpack_Options` doesn't exist?

chrismccluskey: I believe this could be possible on WPCOM, since the WPCOM token subscription service did not require this class to exist.

lezama: I think we can assume it is always loaded, there's a lot of Jetpack code that relies on it

chrismccluskey: I'm not sure if we should make assumptions in this case. If the intention of this PR is to refactor two classes into one, I would ideally like to do so without unintended side effects. Both classes have `available()` conditions, and I want `available()` to return in either of those conditions.

Currently in trunk, the Jetpack_Token class isn't available unless the `Jetpack_Options` class is available. I assume this can happen in broken installations or in cases of data corruption / a typo in the code. Either way, this code is doing a basic check to confirm that the Token service can be used and if `false` is returned, an `Unconfigured_Subscription_Service` will be returned. It might make sense to remove this edge case / flow, but I would like to understand a bit more about how the code currently handles this edge case before I can form much of an opinion on it.

From what I can see, `Unconfigured_Subscription_Service` looks like a simple service that just returns `false` when the `visitor_can_view_content` function is called, and provides the `site_url` when the `access_url` is called. This seems like an easy way to avoid sending PHP error messages to the user. If I make this refactor (changing the line into `return true`, since `( new Host() )->is_wpcom_simple() || true` will always return `true`), and for some reason the `\Jetpack_Options` class isn't available, PHP errors would be displayed to the user.

millerf: >  If the intention of this PR is to refactor two classes into one, I would ideally like to do so without unintended side effects

I agree with @chrismccluskey  here. We can come back to it later.

millerf (approved):

PR #34305, opened 2023-11-26
Fix block registration to prevent console errors in the full site editor

ice9js:
## Proposed changes:

This is an attempt at fixing an issue with the full site editor, where currently all of our blocks are causing the following console error to be thrown:

>The block "jetpack/***" must have a title.

### Other information:

## Jetpack product discussion

p9dueE-7Zq-p2

No.

## Testing instructions:

- Open the full site editor on your test site.
- There should be no errors being logged in the console anymore.
- Jetpack blocks should still be registered correctly and available.

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-block-registration
+++ b/projects/plugins/jetpack/changelog/fix-block-registration
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
+Fixed console errors being thrown for Jetpack blocks inside the Full Site Editor.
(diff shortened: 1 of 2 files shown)


jeherve (commented):
~While this fix makes sense, it doesn't seem to prevent the console errors on my end; I still see them in the site editor.~

Edit: the fix now works on another test site of mine, I think I was running into caching issues. This should be good to go!

cc @monsieur-z who worked on #33546

jeherve (approved):

PR #34306, opened 2023-11-27
Jetpack Google Fonts: Filter out deprecated google fonts data from user global styles

arthur791004:
Fixes https://github.com/Automattic/dotcom-forge/issues/4535, pbxlJb-52N-p2

## Proposed changes:

* The old Google Fonts data, such as the following, was inadvertently stored in the user's global styles. As a result, the Google Fonts cannot be loaded correctly.
  ```json
  [
	{
		"fontFamily": "'Albert Sans'",
		"name": "Albert Sans",
		"slug": "albert-sans",
		"fontFace": {
			"albert-sans-100-900-italic": {
				"origin": "gutenberg_wp_fonts_api",
				"provider": "jetpack-google-fonts",
				"fontFamily": "Albert Sans",
				"fontStyle": "italic",
				"fontWeight": "100 900",
				"fontDisplay": "fallback"
			},
			"albert-sans-100-900-normal": {
				"origin": "gutenberg_wp_fonts_api",
				"provider": "jetpack-google-fonts",
				"fontFamily": "Albert Sans",
				"fontStyle": "normal",
				"fontWeight": "100 900",
				"fontDisplay": "fallback"
			}
		}
	}
  ]
  ```
* This PR adds a new hook to filter out those data from user global styles to resolve the issue

### Other information:

## Testing instructions:

* Go to WoA site
* Install changes from this PR by Jetpack Beta Tester
* Go to the Site Editor
* Update the global styles first
* Open the DevTool, and find the request from the previous step
* Copy the request as followed, `/wp-json/wp/v2/global-styles/<global_styles_id>`.
  image
* Trigger the code with the body below on the console to insert the dirty data. Note that you have to replace the `<global_styles_id>` with the real id on your site.
  ```
  `{"id":<global_styles_id>,"styles":{},"settings":{"typography":{"fontFamilies":{"theme":[{"fontFamily":"'Albert Sans'","name":"Albert Sans","slug":"albert-sans","fontFace":{"albert-sans-100-900-italic":{"origin":"gutenberg_wp_fonts_api","provider":"jetpack-google-fonts","fontFamily":"Albert Sans","fontStyle":"italic","fontWeight":"100 900","fontDisplay":"fallback"},"albert-sans-100-900-normal":{"origin":"gutenberg_wp_fonts_api","provider":"jetpack-google-fonts","fontFamily":"Albert Sans","fontStyle":"normal","fontWeight":"100 900","fontDisplay":"fallback"}}}]}}}}`
  ```
* Refresh the page
* Ensure you're able to see the Google Fonts when you want to change the font family
* Pick the “Albert Sans” font for the “Text”
* Visit your homepage, and ensure the font of the text is rendered correctly

### DIFF
--- a/projects/plugins/jetpack/modules/google-fonts/current/class-jetpack-google-font-face.php
+++ b/projects/plugins/jetpack/modules/google-fonts/current/class-jetpack-google-font-face.php
@@ -75,7 +75,7 @@ class Jetpack_Google_Font_Face {
 
 		$global_styles_font_slug = $this->get_font_slug_from_setting( $global_styles );
 		if ( $global_styles_font_slug ) {
-			$this->fonts_in_use[] = $global_styles_font_slug;
+			$this->add_font( $global_styles_font_slug );
 		}
 
 		if ( isset( $global_styles['blocks'] ) ) {
@@ -83,7 +83,7 @@ class Jetpack_Google_Font_Face {
 				$font_slug = $this->get_font_slug_from_setting( $setting );
 
 				if ( $font_slug ) {
-					$this->fonts_in_use[] = $font_slug;
+					$this->add_font( $font_slug );
 				}
 			}
 		}
@@ -93,7 +93,7 @@ class Jetpack_Google_Font_Face {
 				$font_slug = $this->get_font_slug_from_setting( $setting );
 
 				if ( $font_slug ) {
-					$this->fonts_in_use[] = $font_slug;
+					$this->add_font( $font_slug );
 				}
 			}
 		}
@@ -109,11 +109,20 @@ class Jetpack_Google_Font_Face {
 	 */
 	public function collect_block_fonts( $content, $parsed_block ) {
 		if ( ! is_admin() && isset( $parsed_block['attrs']['fontFamily'] ) ) {
-			$block_font_family    = $parsed_block['attrs']['fontFamily'];
-			$this->fonts_in_use[] = $block_font_family;
+			$block_font_family = $parsed_block['attrs']['fontFamily'];
+			$this->add_font( $block_font_family );
 		}
 
 		return $content;
+	}
+
+	/**
+	 * Add the specify font to the fonts_in_use list.
+	 *
+	 * @param string $font_slug The font slug.
+	 */
+	public function add_font( $font_slug ) {
+		$this->fonts_in_use[] = _wp_to_kebab_case( $font_slug );
 	}
 
 	/**
--- a/projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
+++ b/projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
@@ -5,6 +5,8 @@
  * @package automattic/jetpack
  */
 
+use Automattic\Jetpack\Tracking;
+
 /**
  * Gets the Google Fonts data
  *
@@ -220,9 +222,15 @@ function jetpack_google_fonts_user_global_styles_pre_save( $post_data ) {
 		return $post_data;
 	}
 
+	$prev_count = count( $raw_data['settings']['typography']['fontFamilies'] );
 	$raw_data['settings']['typography']['fontFamilies'] = jetpack_google_fonts_filter_out_deprecated_font_data(
 		$raw_data['settings']['typography']['fontFamilies']
 	);
+
+	if ( $prev_count !== count( $raw_data['settings']['typography']['fontFamilies'] ) ) {
+		$tracking = new Tracking();
+		$tracking->record_user_event( 'jetpack_google_fonts_clean_up_dirty_data' );
+	}
 
 	/**
 	 * Clean up the empty properties
--- a/projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
+++ b/projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
@@ -58,11 +58,15 @@ function jetpack_get_google_fonts_data() {
  * @return object[] The map of the the available Google Fonts.
  */
 function jetpack_get_available_google_fonts_map( $google_fonts_data ) {
+	$font_families = isset( $google_fonts_data ) && isset( $google_fonts_data['fontFamilies'] )
+		? $google_fonts_data['fontFamilies']
+		: array();
+
 	$jetpack_google_fonts_list = array_map(
 		function ( $font_family ) {
 			return $font_family['name'];
 		},
-		$google_fonts_data['fontFamilies']
+		$font_families
 	);
 
 	/** This filter is documented in modules/google-fonts/wordpress-6.3/load-google-fonts.php */
--- a/projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
+++ b/projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
@@ -149,22 +149,15 @@ function jetpack_register_google_fonts_to_theme_json( $theme_json ) {
 add_filter( 'wp_theme_json_data_default', 'jetpack_register_google_fonts_to_theme_json' );
 
 /**
- * Unregister the google fonts data from user's theme json data that were stored by accident.
+ * Filter out the deprecated font families that are from the jetpack-google-fonts provider.
  *
- * @param WP_Theme_JSON_Data $theme_json The theme json data of user.
- * @return WP_Theme_JSON_Data The filtered theme json data.
+ * @param object[] $font_families The font families.
+ * @return object[] The filtered font families.
  */
-function jetpack_unregister_deprecated_google_fonts_from_theme_json_data_user( $theme_json ) {
-	$raw_data = $theme_json->get_data();
-	$origin   = 'custom';
-	if ( empty( $raw_data['settings']['typography']['fontFamilies'][ $origin ] ) ) {
-		return $theme_json;
-	}
-
-	// Filter out the font definitions that are from the jetpack-google-fonts provider.
-	$raw_data['settings']['typography']['fontFamilies'][ $origin ] = array_values(
+function jetpack_google_fonts_filter_out_deprecated_font_data( $font_families ) {
+	return array_values(
 		array_filter(
-			$raw_data['settings']['typography']['fontFamilies'][ $origin ],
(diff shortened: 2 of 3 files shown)


projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
```diff
@@ -148,6 +150,106 @@ function ( $google_fonts_family ) use ( $available_google_fonts_map, $theme_font
...
+	/**
+	 * Clean up the empty properties
+	 */
+	if ( empty( $raw_data['settings']['typography']['fontFamilies'] ) ) {
+		unset( $raw_data['settings']['typography']['fontFamilies'] );
+	}
+
+	if ( empty( $raw_data['settings']['typography'] ) ) {
+		unset( $raw_data['settings']['typography'] );
+	}
```
okmttdhr: Could you clarify the reason for cleaning up empty properties for my learning? Not doing so could lead to errors or other issues?

projects/plugins/jetpack/modules/google-fonts/current/class-jetpack-google-font-face.php
```diff
@@ -109,13 +109,22 @@ public function collect_global_styles_fonts() {
...
 		return $content;
 	}
 
+	/**
+	 * Add the specify font to the fonts_in_use list.
+	 *
+	 * @param string $font_slug The font slug.
+	 */
+	public function add_font( $font_slug ) {
+		$this->fonts_in_use[] = _wp_to_kebab_case( $font_slug );
```
okmttdhr, line 125: badge Is this for the normalization?

arthur791004: YES!

projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
arthur791004: There are user settings; we use them to check whether the value is the same as default or whether global styles are being used. It would be better to clean up empty properties for these checks. Does it make sense?

projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
```diff
@@ -148,6 +150,106 @@ function ( $google_fonts_family ) use ( $available_google_fonts_map, $theme_font
...
+
+add_filter( 'wp_theme_json_data_user', 'jetpack_unregister_deprecated_google_fonts_from_theme_json_data_user' );
+
+/**
+ * Filter out the font families from the jetpack-google-fonts provider when saving the user global styles.
+ *
+ * @param array $post_data The post data to filter.
+ * @return array The filtered post data.
+ */
+function jetpack_google_fonts_user_global_styles_pre_save( $post_data ) {
```
fushar: Do we need this, though? This will always decode and re-encode the saved JSON, for all users, regardless of whether the site actually has bad data.

I think the "more correct" way is to do something like mass-cleanup of all sites with bad user data. Because with this, a site will always have the bad data if it never updates the global styles again.

Since we're already filtering out fonts with deprecated provider, I think we can just leave the bad data as-is. What do you think?

Side note: I'm surprised that the `wp_theme_json_data_user` hook doesn't already filter the `post_data`. It's like we're doing the same logic twice here 😄

projects/plugins/jetpack/modules/google-fonts/current/class-jetpack-google-font-face.php
fushar: Sorry I still don't get it; currently in production, we don't have this "normalization", what could break without this in production?

projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
```diff
@@ -148,6 +150,106 @@ function ( $google_fonts_family ) use ( $available_google_fonts_map, $theme_font
...
+ * @return object[] The filtered font families.
+ */
+function jetpack_google_fonts_filter_out_deprecated_font_data( $font_families ) {
+	return array_values(
+		array_filter(
+			$font_families,
+			function ( $font_family ) {
+				$has_deprecated_google_fonts_data = false;
+				foreach ( $font_family['fontFace'] as $font_face ) {
+					$provider = isset( $font_face['provider'] ) ? $font_face['provider'] : '';
```
fushar, line 168: Nit: why not
```suggestion
					$provider = $font_face['provider'] ?? '';
```

fushar (commented):
Looks good! I also tried updating some random global styles after saving and it still works fine.

|Before|After|
|-|-|
|image<br><br>image|image|

I have several comments; could you check before I approve this PR?

projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
fushar: Another reason is this code actually modifies user data; adding another thing to check if we encounter some other bug in the future 😄 

But not sure. I'm a bit 50:50 here. Maybe we should ask folks outside Lego.

projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
arthur791004: The null coalescing operator (??) was introduced in PHP 7 but the Jetpack plugin supports [5.6 or higher](https://wordpress.org/plugins/jetpack/). I'm unsure whether this feature is available on the client site.

projects/plugins/jetpack/modules/google-fonts/current/class-jetpack-google-font-face.php
arthur791004: There is at least one bug when using the Bitacora theme. The theme JSON assigns the `Alegreya` font to all headings but we don't convert the slug into a kebab casing, so the font is not printed.

projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
arthur791004: Yes, it's not necessary. But I'm unsure how to do a mass-cleanup of all atomic sites with bad user data or how to provide a tool for HEs to run when a user encounters this issue...

Asked p1701172594582909-slack-CDLH4C1UZ

> I'm surprised that the wp_theme_json_data_user hook doesn't already filter the post_data. It's like we're doing the same logic twice here 😄

I'm surprised too! The endpoint to get/set the user global styles loads the `wp-global-styles` post directly. Whereas, the function to get the global styles relies on the `WP_Theme_JSON_Resolver_Gutenberg` class...

projects/plugins/jetpack/modules/google-fonts/current/class-jetpack-google-font-face.php
fushar: Is it this one? (Just wanted to make sure my understanding is correct)

https://github.com/Automattic/themes/blob/ec6164871768088b630b36dc0bc6c2ba39f3bb49/bitacora/theme.json#L235-L236

projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
fushar: TIL about everything you said above, thanks!

projects/plugins/jetpack/modules/google-fonts/current/class-jetpack-google-font-face.php
arthur791004: this one: https://github.com/Automattic/themes/blob/ec6164871768088b630b36dc0bc6c2ba39f3bb49/bitacora/theme.json#L713

If you change the text font of the global styles, the “Alegreya” won't be printed anymore

fushar: Discussed offline: The problematic line is:

https://github.com/Automattic/themes/blob/ec6164871768088b630b36dc0bc6c2ba39f3bb49/bitacora/theme.json#L740

When we change the Text font of the global style, no `alegreya` slug (extracted from `var:preset|font-family|alegreya`) will be in `fonts_in_use` anymore, so the `Alegreya` font used in headings will not be printed (as we're converting it to `alegreya` first when checking whether a font is used).

projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
okmttdhr: Thanks for your explanation! 😸 
One more question for my learning 🙏 : Is it possible to perform the cleanup of empty data in the `jetpack_unregister_deprecated_google_fonts_from_theme_json_data_user` function (or not)? 
I'm curious whether handling the cleanup during the pre-save is sufficient to ensure no empty data is left.

arthur791004: No. Referring to https://github.com/Automattic/jetpack/pull/34306#discussion_r1407641707, they have 2 ways to handle the global styles. For getting & saving via the endpoint, it accesses the global styles post directly so the hook won't be applied 🥲

arthur791004:
Split the cleanup to another PR, https://github.com/Automattic/jetpack/pull/34352

projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
fushar: We decided to extract the filter-on-save logic to another PR.

fushar (approved):
Re-tested and it's still working fine. 🚢

arthur791004:
Sync changes to https://github.com/Automattic/wpcomsh/pull/1587

arthur791004:
I noticed the origin of the actual data is “theme” instead of “custom”. Proposed https://github.com/Automattic/jetpack/pull/34355 to fix the issue.

projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
okmttdhr: I now understand! Thank you so much for this investigation 🙏 🚀

PR #34307, opened 2023-11-27
Adds Creator to My Jetpack

mikestottuk:
## Proposed changes:

Adds Creator to My Jetpack, which are then managed via the settings modules. I've added creator search terms to most of them but some of the earn settings tabs aren't modules so do not show up in search. We can iterate on that later, we may need to look at whether it's worthwhile having a standalone product in future.

### Other information:

Couldn't see any E2E tests that needed to be updated.

## Jetpack product discussion
pf9WZf-9v-p2

No

## Testing instructions:
* Go to `/wp-admin/admin.php?page=my-jetpack` you should see a new card for Creator with a purchase CTA
* Clicking purchase should take you here `/wp-admin/admin.php?page=my-jetpack#/add-creator`
* This loads an interstitial the same style as boost
* Clicking start free, takes you to the settings page `/wp-admin/admin.php?page=jetpack#/settings?term=creator`
* I've added some additional search terms to the modules, not all are able to be searched for though (e.g. Settings -> Monetize -> Collect PayPal payments.
* Purchasing Creator changes the CTA from "Purchase" to "Manage" 

https://github.com/Automattic/jetpack/assets/4077604/77f9e538-6e63-4168-9b67-871dae8242c3

### DIFF
--- a/projects/packages/my-jetpack/_inc/components/product-interstitial/index.jsx
+++ b/projects/packages/my-jetpack/_inc/components/product-interstitial/index.jsx
@@ -246,11 +246,7 @@ export function BoostInterstitial() {
  * @returns {object} CreatorInterstitial react component.
  */
 export function CreatorInterstitial() {
-	return (
-		<ProductInterstitial slug="creator" installsPlugin={ true }>
-			<img src={ boostImage } alt="Creator" />
-		</ProductInterstitial>
-	);
+	return <ProductInterstitial slug="creator" />;
 }
 
 /**
--- a/projects/packages/my-jetpack/_inc/components/product-interstitial/index.jsx
+++ b/projects/packages/my-jetpack/_inc/components/product-interstitial/index.jsx
@@ -241,6 +241,19 @@ export function BoostInterstitial() {
 }
 
 /**
+ * CreatorInterstitial component
+ *
+ * @returns {object} CreatorInterstitial react component.
+ */
+export function CreatorInterstitial() {
+	return (
+		<ProductInterstitial slug="creator" installsPlugin={ true }>
+			<img src={ boostImage } alt="Creator" />
+		</ProductInterstitial>
+	);
+}
+
+/**
  * CRMInterstitial component
  *
  * @returns {object} CRMInterstitial react component.
--- a/projects/packages/my-jetpack/_inc/components/product-interstitial/index.jsx
+++ b/projects/packages/my-jetpack/_inc/components/product-interstitial/index.jsx
@@ -246,7 +246,7 @@ export function BoostInterstitial() {
  * @returns {object} CreatorInterstitial react component.
  */
 export function CreatorInterstitial() {
-	return <ProductInterstitial slug="creator" />;
+	return <ProductInterstitial slug="creator" installsPlugin={ true } />;
 }
 
 /**
--- a/projects/packages/my-jetpack/src/products/class-creator.php
+++ b/projects/packages/my-jetpack/src/products/class-creator.php
@@ -27,13 +27,11 @@ class Creator extends Product {
 	public static $slug = 'creator';
 
 	/**
-	 * Get the plugin slug - ovewrite it and return Jetpack's
+	 * The slug of the plugin associated with this product - Creator functionalities are part of Jetpack's main plugin
 	 *
-	 * @return ?string
+	 * @var string
 	 */
-	public static function get_plugin_slug() {
-		return self::JETPACK_PLUGIN_SLUG;
-	}
+	public static $plugin_slug = self::JETPACK_PLUGIN_SLUG;
 
 	/**
 	 * Get the plugin filename - ovewrite it and return Jetpack's
@@ -43,13 +41,6 @@ class Creator extends Product {
 	public static function get_plugin_filename() {
 		return self::JETPACK_PLUGIN_FILENAME;
 	}
-
-	/**
-	 * The slug of the plugin associated with this product.
-	 *
-	 * @var string
-	 */
-	public static $plugin_slug = 'jetpack-creator';
 
 	/**
 	 * Whether this product requires a user connection
--- a/projects/packages/my-jetpack/src/products/class-creator.php
+++ b/projects/packages/my-jetpack/src/products/class-creator.php
@@ -27,6 +27,15 @@ class Creator extends Product {
 	public static $slug = 'creator';
 
 	/**
+	 * Get the plugin slug - ovewrite it and return Jetpack's
+	 *
+	 * @return ?string
+	 */
+	public static function get_plugin_slug() {
+		return self::JETPACK_PLUGIN_SLUG;
+	}
+
+	/**
 	 * Get the plugin filename - ovewrite it and return Jetpack's
 	 *
 	 * @return ?string
--- a/projects/packages/my-jetpack/src/products/class-creator.php
+++ b/projects/packages/my-jetpack/src/products/class-creator.php
@@ -337,13 +337,7 @@ class Creator extends Product {
 		}
 		if ( is_array( $purchases_data ) && ! empty( $purchases_data ) ) {
 			foreach ( $purchases_data as $purchase ) {
-				if ( str_starts_with( $purchase->product_slug, static::get_wpcom_biyearly_product_slug() ) ) {
-					return true;
-				}
-				if ( str_starts_with( $purchase->product_slug, static::get_wpcom_product_slug() ) ) {
-					return true;
-				}
-				if ( str_starts_with( $purchase->product_slug, static::get_wpcom_monthly_product_slug() ) ) {
+				if ( strpos( $purchase->product_slug, 'jetpack_creator' ) !== false ) {
 					return true;
 				}
 			}
@@ -358,14 +352,8 @@ class Creator extends Product {
 	 */
 	public static function is_upgradable() {
 		$has_required_plan = self::has_required_plan();
-
-		// Mark as not upgradable if user is on unlimited tier or does not have any plan.
-		if ( ! $has_required_plan ) {
-			return true;
-		}
-		return false;
+		return ! $has_required_plan;
 	}
-
 	/**
 	 * Activates the product by installing and activating its plugin
 	 *
--- a/projects/packages/my-jetpack/src/products/class-creator.php
+++ b/projects/packages/my-jetpack/src/products/class-creator.php
@@ -354,24 +354,4 @@ class Creator extends Product {
 		$has_required_plan = self::has_required_plan();
 		return ! $has_required_plan;
 	}
-	/**
-	 * Activates the product by installing and activating its plugin
-	 *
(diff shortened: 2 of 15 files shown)


projects/packages/my-jetpack/_inc/components/product-interstitial/index.jsx
```diff
@@ -240,6 +240,15 @@ export function BoostInterstitial() {
 	);
 }
 
+/**
+ * CreatorInterstitial component
+ *
+ * @returns {object} CreatorInterstitial react component.
+ */
+export function CreatorInterstitial() {
+	return <ProductInterstitial slug="creator" />;
```
jboland88: What happens if we start from a standalone plugin and then buy creator - should we install the main Jetpack plugin in this case since it contains the logic to power creator features?

projects/packages/my-jetpack/src/products/class-creator.php
```diff
@@ -0,0 +1,389 @@
...
+			self::FREE_TIER_SLUG,
+		);
+	}
+
+	/**
+	 * Get the internationalized comparison of free vs upgraded features
+	 *
+	 * @return array[] Protect features comparison
+	 */
+	public static function get_features_by_tier() {
```
jboland88, line 118: It's out of scope for this PR, but it would be great if we could shift to getting these plan features from an API endpoint so that they could be updated independently of a plugin release.

projects/packages/my-jetpack/src/products/class-creator.php
```diff
@@ -0,0 +1,389 @@
...
+			foreach ( $purchases_data as $purchase ) {
+				if ( str_starts_with( $purchase->product_slug, static::get_wpcom_biyearly_product_slug() ) ) {
+					return true;
+				}
+				if ( str_starts_with( $purchase->product_slug, static::get_wpcom_product_slug() ) ) {
+					return true;
+				}
+				if ( str_starts_with( $purchase->product_slug, static::get_wpcom_monthly_product_slug() ) ) {
+					return true;
+				}
```
jboland88: You could likely shorten this up by just doing one `strpos` check for `jetpack_creator` since all the relevant plan slugs should contain that string

projects/packages/my-jetpack/src/products/class-creator.php
```diff
@@ -0,0 +1,389 @@
...
+	 * @return boolean
+	 */
+	public static function is_upgradable() {
+		$has_required_plan = self::has_required_plan();
+
+		// Mark as not upgradable if user is on unlimited tier or does not have any plan.
+		if ( ! $has_required_plan ) {
+			return true;
+		}
+		return false;
```
jboland88: nitpick: could simplify by just returning `! $has_required_plan`

projects/packages/my-jetpack/src/products/class-creator.php
```diff
@@ -0,0 +1,389 @@
...
+
+		$product_activation = parent::do_product_specific_activation( $current_result );
+
+		if ( is_wp_error( $product_activation ) && 'module_activation_failed' === $product_activation->get_error_code() ) {
+			// A bundle is not a module. There's nothing in the plugin to be activated, so it's ok to fail to activate the module.
+			$product_activation = true;
+		}
+
+		// We just "got started" in My Jetpack, so skip the in-plugin experience.
+		update_option( 'jb_get_started', false );
```
jboland88: this looks like a boost specific option?

projects/packages/my-jetpack/src/products/class-creator.php
```diff
@@ -0,0 +1,389 @@
...
+	 * Activates the product by installing and activating its plugin
+	 *
+	 * @param bool|WP_Error $current_result Is the result of the top level activation actions. You probably won't do anything if it is an WP_Error.
+	 * @return boolean|\WP_Error
+	 */
+	public static function do_product_specific_activation( $current_result ) {
+
+		$product_activation = parent::do_product_specific_activation( $current_result );
+
+		if ( is_wp_error( $product_activation ) && 'module_activation_failed' === $product_activation->get_error_code() ) {
```
jboland88: I don't think you'd end up getting the `module_activation_failed ` code here since this class only extends the base `Product` class, which does not return that error.

projects/packages/my-jetpack/src/products/class-creator.php
```diff
@@ -0,0 +1,389 @@
...
+		return false;
+	}
+
+	/**
+	 * Activates the product by installing and activating its plugin
+	 *
+	 * @param bool|WP_Error $current_result Is the result of the top level activation actions. You probably won't do anything if it is an WP_Error.
+	 * @return boolean|\WP_Error
+	 */
+	public static function do_product_specific_activation( $current_result ) {
```
jboland88: I'm not super familiar with the activation routines in My Jetpack - but I don't think this function is needed unless the product is installing and activating a plugin. Granted, we might want to install the main Jetpack plugin in some cases to power creator features?

Additionally, if there are specific modules that creator is using, those could be auto-enabled here.

jboland88 (dismissed):
Overall, the code looks alright.

I think we need to handle the case where creator is added/ purchased when only a standalone plugin is activated. We need to install the main Jetpack plugin in these cases since it contains the functionality to power creator.

projects/packages/my-jetpack/_inc/components/product-interstitial/index.jsx
mikestottuk: Good catch, yes we should

projects/packages/my-jetpack/src/products/class-creator.php
mikestottuk: > Additionally, if there are specific modules that creator is using, those could be auto-enabled here.

I did wonder about auto-enabling, but sites might not want to use Newsletter so I directed them to the settings with ?term=creator instead so they can turn on what they want.

It's a step towards better onboarding, but not ideal IMO

projects/packages/my-jetpack/_inc/components/product-interstitial/index.jsx
mikestottuk: This looks to work already. Just tested with Boost active and then purchased Creator. It landed me in settings with Jetpack active.  I did have Jetpack deactivated (but not deleted). Will try that flow next after I push the code changes.

mikestottuk: @jboland88  I'm not sure how best to test the standalone plugin only -> purchase Creator from My Jetpack -> Jetpack installs & activates until this is merged. When I try on a JN site it shows me the old My Jetpack (I have Boost installed). The flow looks to work for AI. 

It looked to activate Jetpack when I purchased above, with Jetpack already active. 

Unless you're aware of a better approach to testing this.

Screenshot 2023-11-28 at 10 51 36

simison:
@mikestottuk the copy feels a bit vague/marketing'y:

Screenshot 2023-11-29 at 14 05 59

Could it have something more specific about Newsletters, paid memberships, donations, selling products, and blocks? If I'd have to pick I'd go with mentioning Newsletters and paid memberships.

Does the next step elaborate more on specific features included in the plans?

Above said, and I get that this is the "product" bundle, but at the same time I think you'd have more success if we could just have "Newsletters" and "Paid Memberships" at My Jetpack dash as cards. Main CTA wouldn't need to be "purchase", and you'd upsell in the activation flow instead similarly to recommendation flow. Features are free, after all, and we don't surface them much anywhere.

mikestottuk:
> the copy feels a bit vague/marketing'y. Could it have something more specific about Newsletters, paid memberships, donations, selling products, and blocks?

@simison  So are the other products. We can iterate on the copy if we're not seeing CTRs in line with other products.

Screenshot 2023-11-29 at 12 03 14


The next step lists what features are included, see below: 

Screenshot 2023-11-29 at 12 03 23

> I think you'd have more success if we could just have "Newsletters" and "Paid Memberships" at  My Jetpack dash as cards.

Individual features aren't displayed at the "My Jetpack" level, it's intended to be a place where you manage your products that you've purchased & activated. 

> Features are free, after all, and we don't surface them much anywhere.

Yes, most are. WordAds & Pay with PayPal aren't. We should look at surfacing individual features (like Newsletters & Paid Memberships) elsewhere in flows ( e.g. like the Recommendations flow #34350 ) where we could look to feature prompt additional features.

projects/packages/my-jetpack/_inc/components/product-interstitial/index.jsx
robertsreberski: It doesn't seem to work for me, hmmm 🤔 here are test steps I made:

1. Install Jurassic Ninja with Beta Tester
2. Activate Boost plugin with feature branch `add/creator-to-my-jetpack`
3. Go to "My Jetpack" and see Creator card there
4. Click "Get Creator", see interstitial, proceed to checkout, complete purchase
5. Come back to `/wp-admin/admin.php?page=jetpack-beta`
6. But the page is empty - because no Jetpack main plugin is installed

cc @jboland88

robertsreberski: I've gone ahead and debugged and fixed the issue! The main Jetpack plugin gets installed successfully when JP Creator is activated from JP Boost standalone plugin, with this fix applied: https://github.com/Automattic/jetpack/pull/34307/commits/8b698ac662f0ea2971a5aecd0384ff6794a2313e

Currently it installs the newest public version (and thus search terms don't work), but I'm confident it will work as expected once this release will get deployed.

robertsreberski (dismissed):
All LGTM!

projects/packages/my-jetpack/src/products/class-creator.php
```diff
@@ -0,0 +1,366 @@
...
+	public static function get_plugin_filename() {
+		return self::JETPACK_PLUGIN_FILENAME;
+	}
+
+	/**
+	 * The slug of the plugin associated with this product.
+	 *
+	 * @var string
+	 */
+	public static $plugin_slug = 'jetpack-creator';
```
jboland88: Any reason not to set this as the Jetpack plugin slug here instead of overriding the methods above?

projects/packages/my-jetpack/_inc/components/product-interstitial/index.jsx
jboland88: @robertsreberski - should we add `installsPlugin={true}` here?

jboland88 (commented):
@robertsreberski - left a couple questions here. I have not re-tested.

projects/packages/my-jetpack/src/products/class-creator.php
robertsreberski: I was just following the way it was done in Jetpack AI. Indeed, it can be done simpler - will adjust!

projects/packages/my-jetpack/_inc/components/product-interstitial/index.jsx
robertsreberski: We do, thanks!

robertsreberski (approved):
🚀

PR #34310, opened 2023-11-27
Newsletter: update post-publish panel copy for scheduled posts

simison:
Fixes issue reported in pdrWKz-1ll-p2

Before:

## Proposed changes:

* Update copy when post is schedulled, i.e. emails didn't go out yet.
* Remove joined string for translations; makes "subscribers" or "paid subscribers" string part of the whole translated sentence, or otherwise it'll be hard or impossible to translate. Results in more elaborate code unfortunately.
* Move logic of determining exact string to its own function so that it's easier to read

### Other information:

## Testing instructions:

* Have blog with subscriptions enabled, and paid memberships configured (stripe connected etc)

Note copy in post-publish panel by doing:

* Create a new paid post, and schedule it.
* Create another paid post and publish it.
* Create a new free post, and schedule it.
* Create another free post and publish it.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
@@ -172,7 +172,97 @@ function NewsletterPrePublishSettingsPanel( {
 	);
 }
 
+// Determines copy to show in post-publish panel to confirm number and type of subscribers who received the post as email, or will receive in case of scheduled post.
+function getNumberOfSubscribersText( {
+	accessLevel,
+	isScheduledPost,
+	newsletterCategories,
+	newsletterCategoriesEnabled,
+	postCategories,
+	postName,
+	reachCount,
+	subscriptionsCount,
+} ) {
+	// Text when categories are enabled
+	if (
+		newsletterCategoriesEnabled &&
+		newsletterCategories.length &&
+		accessLevel !== accessOptions.paid_subscribers.key
+	) {
+		const formattedCategoryNames = getFormattedCategories( postCategories, newsletterCategories );
+		const formattedSubscriptionsCount = getFormattedSubscriptionsCount( subscriptionsCount );
+		const categoryNamesAndSubscriptionsCount = formattedCategoryNames + formattedSubscriptionsCount;
+
+		if ( isScheduledPost && formattedCategoryNames ) {
+			return sprintf(
+				// translators: %1s is the post name, %2s is the list of categories with subscriptions count
+				__(
+					'<postPublishedLink>%1$s</postPublishedLink> was sent to everyone subscribed to %2$s.',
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
```diff
@@ -172,7 +172,97 @@ function NewsletterPrePublishSettingsPanel( {
...
+} ) {
+	// Text when categories are enabled
+	if (
+		newsletterCategoriesEnabled &&
+		newsletterCategories.length &&
+		accessLevel !== accessOptions.paid_subscribers.key
+	) {
+		const formattedCategoryNames = getFormattedCategories( postCategories, newsletterCategories );
+		const formattedSubscriptionsCount = getFormattedSubscriptionsCount( subscriptionsCount );
+		const categoryNamesAndSubscriptionsCount = formattedCategoryNames + formattedSubscriptionsCount;
```
simison, line 194: This results in strings like **Category1 and Category 2 (5 subscribers)** which still isn't ideal for translating, as we're concennating the text. The use of brackets makes it fine I think, though!

simison:
There's a build error that I'm looking into:

> ERROR in editor.js: Optimization seems to have broken the following translation strings:

**Update:** Solved in https://github.com/Automattic/jetpack/pull/34310/commits/7660fee3cbda81720050725b76a7f4810d13e068; translation strings in ternaries break in minification. See further context p1683198038274859-slack-CBG1CP4EN

projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
```diff
@@ -172,7 +172,105 @@ function NewsletterPrePublishSettingsPanel( {
...
+
+		if ( isScheduledPost && formattedCategoryNames ) {
+			return sprintf(
+				// translators: %1s is the post name, %2s is the list of categories with subscriptions count
+				__(
+					'<postPublishedLink>%1$s</postPublishedLink> was sent to everyone subscribed to %2$s.',
+					'jetpack'
+				),
```
edanzer, line 202: I can't figure quite how to test this with the categories active, but lookin at the logic here, I wonder if we have the tenses reverse?

The first conditional is (isScheduledPost && formattedCategoryNames), which means the post has not been sent yet, but the text says 'was sent'. Should it be 'will be sent here', and 'was sent' in the else if block below? 

That's how the tenses are set up below in the non-category if / elseif statements.

simison: Ah yes, sorry about that! I'll fix.

Newsletter categories are now accessible only via Calypso for Jetpack sites, i.e. _Settings -> Newsletter_.

lezama (approved):

PR #34314, opened 2023-11-27
Fix PHP 8.3 `ReflectionProperty::setValue()` deprecation warnings

anomiex:
## Proposed changes:

In PHP 8.3, calling `ReflectionProperty::setValue()` with a single argument (rather than passing `null` for the `$object`) is deprecated. We do this in a few tests, let's fix it in preparation for running PHP 8.3 in CI.

### Other information:

## Testing instructions:

* Do the tests still pass?
* Unfortunately we can't easily test the PHP 8.3 support yet.


jeherve (dismissed):
Should be okay after a trunk update. 🚢

anomiex:
Needs re-approval after version bumps

jeherve (approved):

PR #33424, opened 2023-09-29
CRM: Add Send_Email action + create a typical real automation scenario for a business

cleacos:
## Proposed changes:

With the goal of having a real example, this PR adds a basic Send_Email functionality. Most of the code was extracted from https://github.com/Automattic/jetpack/pull/33219 and updated/adapted in this one.

**This is the automation we want to implement:**

- On new Contact with the status `Lead`, tag it as `Prospect`.
- When the Contact `Lead` changes to `Customer`, add the `New Client` tag and remove the `Prospect` tag. Finally, send a Customer Welcome email with a special offer.

And these are the two Automation Workflows required:

```php
$workflow_1 = new \Automattic\Jetpack\CRM\Automation\Automation_Workflow(
	[
		'name'         => 'New Contact Lead: Prospect',
		'description'  => 'On a new Contact Lead, apply the tag Prospect',
		'category'     => 'contact',
		'active'       => true,
		'triggers'     => [
			\Automattic\Jetpack\CRM\Automation\Triggers\Contact_Created::get_slug(),
		],
		'initial_step' => 1,
		'steps'        => [
			1 => [
				'slug'       => \Automattic\Jetpack\CRM\Automation\Conditions\Contact_Field_Changed::get_slug(),
				'attributes' => [
					'field'    => 'status',
					'operator' => 'is',
					'value'    => 'Lead'
				],
				'next_step_true'  => 2,
				'next_step_false' => null,
			],
			2 => [
				'slug'       => \Automattic\Jetpack\CRM\Automation\Actions\Add_Remove_Contact_Tag::get_slug(),
				'attributes' => [
					'mode'      => 'append',
					'tag_input' => ['Prospect'],
				],
				'next_step_true'  => null,
			],
		],
	]
);

$workflow_2 = new \Automattic\Jetpack\CRM\Automation\Automation_Workflow(
	[
		'name'         => 'Conversion Contact Lead to Customer',
		'description'  => 'This automation will remove/add tags and send a Welcome Email to the new customer.',
		'category'     => 'contact',
		'active'       => true,
		'triggers'     => [
			\Automattic\Jetpack\CRM\Automation\Triggers\Contact_Status_Updated::get_slug(),
		],
		'initial_step' => 1,
		'steps'        => [
			1 => [
				'slug'       => \Automattic\Jetpack\CRM\Automation\Conditions\Contact_Transitional_Status::get_slug(),
				'attributes' => [
					'operator'            => 'from_to',
					'previous_status_was' => 'Lead',
[...]

PR #34315, opened 2023-11-27
CRM: Resolves 3389 - remove axios dependencies

tbradsha:
CRM: Resolves Automattic/zero-bs-crm#3389 - remove axios dependencies

## Proposed changes:

In Automattic/jetpack#34238 the axios dependency was updated. However, we really don't even need this dependency, as we can just use `fetch`.

This PR removes the axios dependency in favor of `fetch`, adding a small `callAutomationsAPI` wrapper function. I also fixed a few React errors due to items not having a `key` prop and table data not being wrapped in `<thead>` and `<tbody>`.

### Other information:

## Testing instructions:

No functionality should have changed between `trunk` and the `fix/crm/3389-remove_axios_dependencies` branch. Likewise, automations is behind a feature flag.

1. Enable automations feature flag:
    ```
    ```
2. Go to the automations dashboard: `/wp-admin/admin.php?page=jpcrm-automations#/automations`
3. Ensure it is populated with any workflows you'd previously added, and that you can edit them. If you don't have any, there are two example workflows in #33424.

### DIFF
--- a/projects/plugins/crm/src/js/data/query-functions.ts
+++ b/projects/plugins/crm/src/js/data/query-functions.ts
@@ -1,4 +1,3 @@
-import axios from 'axios';
 import { Workflow, ServerPreparedWorkflow } from 'crm/state/automations-admin/types';
 import { getServerPreparedWorkflow } from 'crm/state/automations-admin/util';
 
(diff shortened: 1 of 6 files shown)


projects/plugins/crm/src/js/data/query-functions.ts
```diff
@@ -7,24 +6,43 @@ declare let jpcrmAutomationsInitialState: any;
...
+			'Content-Type': 'application/json',
+		},
+		method: 'GET',
+	};
+	if ( payload ) {
+		options.method = 'POST';
+		options.body = JSON.stringify( payload );
+	}
+	const response = await fetch( v4ApiRoot + relative_url, options );
+	return await response.json();
```
coder-karen, line 28: One thing I'm not sure about here, as we don't appear to have done that before (in this file anyway) is what happens if the response is an error or unexpected result. Perhaps not for this PR anyway :)

coder-karen (dismissed):
This tests well for me .

projects/plugins/crm/src/js/data/query-functions.ts
tbradsha: Yeah, there are definitely improvements to be made...in the future!

coder-karen (approved):

PR #34320, opened 2023-11-28
Launchpad: Add the Legacy Site Setup task list

valterlorran:
Related to: https://github.com/Automattic/wp-calypso/issues/84409

More context: paYKcK-3Hy-p2

## Proposed changes:

* We need to migrate the site setup logic that still displays the old checklist. This PR adds the initial structure for the new checklist called `legacy-site-setup`, which we will use to show the Launchpad in the Customer Home in case the site does not have the `write` intent nor is a hosted site.

### Other information:

## Testing instructions:

* Apply this PR to your sandbox using the command below:
```bash
bin/jetpack-downloader test jetpack-mu-wpcom-plugin add/site-setup-launchpad
```
* Navigate to the Developer Console: https://developer.wordpress.com/docs/api/console/
* Make a GET request to `wpcom/v2` -> `/sites/:siteSlug/launchpad?checklist_slug=legacy-site-setup`
* Ensure you get a response similar to:
Screen Shot 2023-12-05 at 01 32 28

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -13,7 +13,7 @@
 function wpcom_launchpad_get_task_definitions() {
 	$task_definitions = array(
 		// Core tasks.
-		'design_edited'                   => array(
+		'design_edited'                      => array(
 			'get_title'             => function () {
 				return __( 'Edit site design', 'jetpack-mu-wpcom' );
 			},
@@ -26,7 +26,7 @@ function wpcom_launchpad_get_task_definitions() {
 			},
 		),
 		// design_completed checks for task completion while design_selected always returns true.
-		'design_completed'                => array(
+		'design_completed'                   => array(
 			'get_title'            => function () {
 				return __( 'Select a design', 'jetpack-mu-wpcom' );
 			},
@@ -36,7 +36,7 @@ function wpcom_launchpad_get_task_definitions() {
 				return '/setup/update-design/designSetup?siteSlug=' . $data['site_slug_encoded'] . '&flow=' . $flow;
 			},
 		),
-		'design_selected'                 => array(
+		'design_selected'                    => array(
 			'get_title'            => function () {
 				return __( 'Select a design', 'jetpack-mu-wpcom' );
 			},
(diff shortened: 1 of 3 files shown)


projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -579,6 +579,44 @@ function wpcom_launchpad_get_task_definitions() {
 				return '/site-editor/' . $data['site_slug_encoded'];
 			},
 		),
+		'blogname_set'                       => array(
+			'get_title'            => function () {
+				return __( 'Give your site a name', 'jetpack-mu-wpcom' );
+			},
+			'is_complete_callback' => 'wpcom_launchpad_is_legacy_task_option_completed',
```
daledupreez: Might it not be more robust to have the completion and/or visibility logic combine checks against the legacy option and the launchpad option? I am especially thinking of Atomic sites, where we won't have access to code in the `Onboarding` namespace. This specific task feels equivalent to the site title task, so we're already tracking things there.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad.php
```diff
@@ -242,6 +242,18 @@ function wpcom_launchpad_get_task_list_definitions() {
 				'share_site',
 			),
 		),
+		'site-setup'             => array(
```
daledupreez: Might be worth calling this `legacy-site-setup` to make it clearer that it's an older/legacy task list.

daledupreez (commented):
Based on inspection, this all looks OK, but I have a broader concern about relying on the older completion options alone, as doing that means we're also reliant on all the related code needing to stay in place as well. For me, that's especially problematic as I don't believe the completion code is available on Atomic sites, so I am worried that the task list will be empty due to us not being able to tell what state the various tasks are in. We can try just this code as a first iteration, but I'd like to make sure we've thought through the overall process before we commit to that.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
valterlorran: Oh! One of my pushes didn't go through yesterday, the `wpcom_launchpad_is_legacy_task_option_completed` should look like this:
```php
function wpcom_launchpad_is_legacy_task_option_completed( $task ) {
	$option_name = 'onboarding_checklist_task_' . $task['id'];
	return get_option( $option_name, false ) || wpcom_launchpad_is_task_option_completed( $task );
}
```

markbiek (approved):
LGTM!

PR #34322, opened 2023-11-28
Fix warnings from jetpack-mu-wpcom coming soon filter

daledupreez:
## Proposed changes:

* This PR adds an `is_array()` check to a coming soon filter to prevent PHP warnings when the incoming argument is not an array. On WordPress.com, we're seeing cases where the incoming `$unfiltered_input` argument is null

### Other information:

## Testing instructions:

* I think it should be sufficient to handle this one via inspection, as we shouldn't be calling `array_key_exists()` if we don't have an array argument.


sixhours (approved):
LGTM!

PR #34330, opened 2023-11-28
Remove/social obsolete twitter helpers

gmjuhasz:
Fixes https://github.com/orgs/Automattic/projects/733/views/2?pane=issue&itemId=43389559

WPCOM change: D130298-code

Remove sync: D132176-code

## Proposed changes:
* Removed obsolete class and it's tests
* Removed 3rd party that is not needed anymore'

### Other information:

## Jetpack product discussion

https://github.com/orgs/Automattic/projects/733/views/2?pane=issue&itemId=43389559

No.
## Testing instructions:

* Sanity checks that the files are not used anymore


jeherve (dismissed):
We would need to remove the endpoints as well, since they rely on that class:
https://github.com/Automattic/jetpack/blob/7df859ce96c415d521863b54d66078d31fca3050/projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-tweetstorm-gather.php#L32-L34
https://github.com/Automattic/jetpack/blob/c6e4509b63b6f074abcfd713a35f2d2d13993615/projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-tweetstorm-parse.php#L22-L24

(of note, we'll need to remove the code that loads those endpoints from WordPress.com as well, ideally before this is merged)

If we remove the endpoint, we can also remove everything that's related to tweetstorms from the Publicize package as well. See references to `tweetstorm` in the codebase.

gmjuhasz:
@jeherve I removed those too, and removed the imports from wpcom here:
D130298-code

jeherve:
Thank you! Looks like you may have to fix one more changelog issue before this can be merged, though.

gmjuhasz:
> Thank you! Looks like you may have to fix one more changelog issue before this can be merged, though.

Thanks Jeremy! And the Phab change should go in first, right?

jeherve:
Yes, that would make the daily WordPress.com deployment easier.

manzoorwanijk (dismissed):
This looks good to me

projects/plugins/social/changelog/remove-social-obsolete-twitter-helpers
jeherve, line 1: This changelog entry can be removed since there is no more change to the Social plugin.

projects/plugins/social/changelog/remove-social-obsolete-twitter-helpers#2
jeherve, line 1: This changelog entry can be removed since there is no more change to the Social plugin.

projects/plugins/jetpack/_inc/lib/class-jetpack-tweetstorm-helper.php
```diff
@@ -1,1754 +0,0 @@
...
-use Automattic\Jetpack\Status;
-use Twitter\Text\Regex as Twitter_Regex;
-use Twitter\Text\Validator as Twitter_Validator;
-
-/**
- * Class Jetpack_Tweetstorm_Helper
- *
- * @since 8.7.0
- */
-class Jetpack_Tweetstorm_Helper {
```
jeherve, line 20: Since the matching WordPress.com diff isn't merged yet, maybe it would be best to either merge the WordPress.com diff before we merge this, or keep this class around but mark it as deprecated, only to remove it later.

projects/packages/publicize/src/class-publicize-base.php
```diff
@@ -49,13 +49,6 @@ abstract class Publicize_Base {
 	 */
 	public $POST_MESS = '_wpas_mess';
 
-	/**
-	 * Post meta key for flagging when the post is a tweetstorm.
-	 *
-	 * @var string
-	 */
-	public $POST_TWEETSTORM = '_wpas_is_tweetstorm';
```
jeherve, line 57: Since we're not using this anymore, maybe we can remove it from the sync post meta too?

https://github.com/Automattic/jetpack/blob/1039bd3005c64e43089e38e7264866bef53d793e/projects/packages/sync/src/class-defaults.php#L743

gmjuhasz: Good point

gmjuhasz: I created a Phab patch for the wpcom change: D132176-code

So we have to merge that first once again?

projects/plugins/jetpack/_inc/lib/class-jetpack-tweetstorm-helper.php
gmjuhasz: It was merged already when I saw your comment. It was merged two days ago. Should I add these back as deprecated or it should be fine to merge like this?

D130298-code

jeherve: Perfect, no need to add them back now 👍

projects/packages/publicize/src/class-publicize-base.php
jeherve: You can merge it now indeed. 👍

jeherve (dismissed):

spsiddarthan (approved):
Works well!

Issue #19694, opened 2021-04-29
Sharing: create a "Sharing Buttons" block


andrewserong:
Related Epic: #31920

Related to #19659 (essentially the same task for the Like button feature)

This issue is to implement a simple MVP server side rendering block for the "Share this" feature that gets automatically appended to the end of post content. Ideally in full site editing, a user should be able to place this feature where they like within a template part.

As an example, the below screenshot is a mock up of moving out the Like button feature to a Footer template part. Here, it's clear that it would be good for the user to also be able to place the "Share this" feature elsewhere within the page:

![image](https://user-images.githubusercontent.com/14988353/116512626-ae5d0c00-a90b-11eb-8036-e0f6cda3b8b2.png)

## Tasks

* [x] Create a block for the Sharing feature (and come up with a clear name for it)
  * [x] In the editor, use a static approximation of the feature
  * [x] On the front end, add a server side render_block method that calls the existing `sharing_display` method
* [ ] If an FSE theme is active, prevent the automatic `sharing_display` from appending the "Share this" feature to post content (it's registered against [the_content filter here](https://github.com/Automattic/jetpack/blob/e13487d0dbbe4c743739b6f1a3e37c143a8c2614/projects/plugins/jetpack/modules/sharedaddy/sharing-service.php#L21-L21), and the [logic lives here](https://github.com/Automattic/jetpack/blob/e13487d0dbbe4c743739b6f1a3e37c143a8c2614/projects/plugins/jetpack/modules/sharedaddy/sharing-service.php#L700-L700) )
* [ ] Ensure the existing settings to hide the sharing feature on an individual post or globally, still applies to the block version

*******

```[tasklist]
### Tasks
- [ ] https://github.com/Automattic/jetpack/issues/34050
- [ ] https://github.com/Automattic/jetpack/pull/34085
- [ ] https://github.com/Automattic/jetpack/issues/34108
- [ ] https://github.com/Automattic/jetpack/issues/34115
- [ ] https://github.com/Automattic/jetpack/issues/34116
- [ ] https://github.com/Automattic/jetpack/issues/34324
- [ ] https://github.com/Automattic/jetpack/issues/34325
- [ ] https://github.com/Automattic/jetpack/issues/34326
- [ ] https://github.com/Automattic/jetpack/issues/34327
- [ ] https://github.com/Automattic/jetpack/issues/34521
- [ ] https://github.com/Automattic/jetpack/issues/34580
[...]

PR #34340, opened 2023-11-28
Like block: Create scaffolding for new dynamic beta Like button block

### LINKED ISSUE #84685: Like block: Create scaffolding for beta version of the block
The goal of this task is to create scaffolding for the new dynamic Like block. Once the main block structure is in, we can then build on top of it with additional PRs.

ivan-ottinger:
Resolves https://github.com/Automattic/wp-calypso/issues/84685.

## Proposed changes:

* adds the scaffolding for the new dynamic Like block (beta)

Markup on 2023-11-28 at 16:32:57

### Other information:

## Jetpack product discussion

Related project thread: pdDOJh-2JP-p2

No, it does not.

## Testing instructions:

1. Please check out the branch and build it.
3. Create a new page / post, add the Like block to it and save the changes. The block should be working correctly in backend and frontend. At the moment, the frontend displays just a simple text string ("This is where the like button will go.").


ivan-ottinger:
`WordPress.com Tests` are failing, but it doesn't seem to be related to the changes proposed in this PR.

jeherve (approved):
Let's get this merged!

One thing that we'll need to consider as we build it is module status, since on self-hosted and WoA sites, the Likes module can be disabled. I don't know what our approach should be there. Maybe the block should be independent from the module (this is the approach we're taking with the sharing buttons block so far), but in this case we'll need to update the different UI settings around Likes whenever a block-based theme is in use, to let folks know they should use the block, and should not expect the module elements to play a role.

For reference, here is the primary issue for our work on the sharing buttons block. It includes sub-issues for different UI elements.
#19694

ivan-ottinger:
Thanks for the review and sharing the details, Jeremy.

> One thing that we'll need to consider as we build it is module status, since on self-hosted and WoA sites, the Likes module can be disabled. I don't know what our approach should be there. Maybe the block should be independent from the module (this is the approach we're taking with the sharing buttons block so far), but in this case we'll need to update the different UI settings around Likes whenever a block-based theme is in use, to let folks know they should use the block, and should not expect the module elements to play a role.

From my point of view, I also think it would be the best if the new block isn't tied to the module. I would even vote for building the block from scratch (instead of relying on the existing iframe code we currently have). We will discuss this in our team and experiment further.

cc @TimBroddin 

> For reference, here is the primary issue for our work on the sharing buttons block. It includes sub-issues for different UI elements. https://github.com/Automattic/jetpack/issues/19694

That's helpful! Thanks!

PR #34345, opened 2023-11-28
renovate: Filter updates of PHP non-dev deps

anomiex:
## Proposed changes:

For our non-dev PHP deps, we can't update if the upstream package has dropped support for the lowest PHP version we still support. Adjust renovate's config to filter those out.

Also, update the rule that sets `rangeStrategy: widen` for PHP dev deps to do it generally instead of depending on a list of specific deps. Too bad there's no "only widen if necessary for version support" option.

### Other information:

## Jetpack product discussion

None

No

## Testing instructions:

* Merge this in a forked repo, add any necessary secrets, and run it. Or just look at https://github.com/anomiex/jetpack/actions/runs/7022848363/job/19108043107 where I did that.

### DIFF
--- a/.github/renovate-config.js
+++ b/.github/renovate-config.js
@@ -81,5 +81,14 @@ module.exports = {
 			} )(),
 			enabled: false,
 		},
+		// PHP non-dev deps need to work with the oldest PHP versions we support.
+		{
+			matchDatasources: [ 'packagist' ],
+			matchDepTypes: [ 'require' ],
+			constraintsFiltering: 'strict',
+			constraints: {
+				php: `~${ versions.MIN_PHP_VERSION }.0`,
+			},
+		},
 	],
 };
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -26,16 +26,12 @@
 		// Monorepo packages are excluded in renovate-config.js, where we can read
 		// the list of them in code.
 
-		// We need to keep a wide version range to support PHP 7.0.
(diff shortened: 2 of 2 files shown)


.github/renovate-config.js
```diff
@@ -81,5 +81,14 @@ module.exports = {
 			} )(),
 			enabled: false,
 		},
+		// PHP non-dev deps need to work with the oldest PHP versions we support.
+		{
+			matchDatasources: [ 'packagist' ],
```
jeherve, line 86: When we match for `packagist`, wouldn't that also match all our own packages, and not just third-party packages? Would that be an issue?

anomiex: Our own packages are all excluded just above.

OTOH, I just thought that we might want to restrict this to just _plugin_ deps, as that's the only place where we bundle stuff. So I might see if I can rework it later along those lines.

jeherve (approved):

anomiex:
Darn, I forgot to turn off auto-merge...

.github/renovate-config.js
anomiex: Now that I think about it even more, it's probably better to leave this as-is. To make the change I was thinking we'd have to somehow ensure that the lower-version-compatible versions were what the build bundles in with the plugins, and then we'd have to make sure that all the CI tests run with those lower versions too. Better to just depend only on the lower versions in the first place, even in the packages.

PR #34349, opened 2023-11-28
Move/source sanitize to jetpack

chrisfromthelc:
Fixes https://github.com/Automattic/gold/issues/234
Companion diff: D130238-code

## Proposed changes:

* This will move this sanitization to the Memberships endpoint in Jetpack (`members.php`) so this is clearer, and then it can be removed from WPCOM's `connected-accounts.php`.

### Other information:

## Testing instructions:

* TBD


n3f:
The code is looking good, but I haven't tested it.

chrisfromthelc:
> The code is looking good, but I haven't tested it.

I changed from the nested ternaries that was on WPCOM to something a little more straightforward to read here. I can switch back to the ternaries if we really want to keep it that way.

millerf (dismissed):
LGTM

n3f (approved):
LGTM

PR #34353, opened 2023-11-29
Jetpack Menu: Add a link to the Jetpack Manage to Jetpack menu

vitozev:
## Proposed changes:

The Jetpack Manage might be useful for users managing multiple websites. This PR adds the link on the Jetpack menu for visibility. It will be displayed on all Jetpack plugins, as long as the user is connected to WordPress.com, and the site isn't multisite, as Jetpack Cloud isn't fully compatible with multisite installations yet.

image

### Other information:

## Jetpack product discussion

* p1HpG7-pYl-p2

* Yes, clicks on the link are tracked (see PCYsg-pY7-p2#monitoring )

## Testing instructions:

- Start with a brand new site (you can use the link below for spinning up a JN site), using Jetpack plugin only (the instructions aren't valid for standalone plugins, and the link shouldn't appear there)
- You should not see the "Jetpack Menu" menu item by default.
- Connect your site to WordPress.com.
- When you return to wp-admin, you should notice a "Jetpack Manage" link with an external icon next to it, on the last position.
- When clicking on the link, you should be redirected to `https://cloud.jetpack/dashboard?origin=wp-admin`
- You can try installing an additional Jetpack plugin, and the link position should not change.
- You can also try to launch a Multisite network; you should not see the "Jetpack Manage" menu item there.

### DIFF
--- a/projects/packages/my-jetpack/src/class-jetpack-manage.php
+++ b/projects/packages/my-jetpack/src/class-jetpack-manage.php
@@ -39,8 +39,12 @@ class Jetpack_Manage {
 		}
 
 		// Do not display the menu if the user has <= 2 sites.
-		$number_of_sites = ( new Connection_Manager() )->get_connected_user_data( get_current_user_id() )['site_count'];
-		if ( $number_of_sites <= 2 ) {
+		$user_data = ( new Connection_Manager() )->get_connected_user_data( get_current_user_id() );
+		if ( ! isset( $user_data['site_count'] ) ) {
+			return;
+		}
+
+		if ( $user_data['site_count'] <= 2 ) {
 			return;
 		}
 
--- a/projects/packages/my-jetpack/src/class-jetpack-manage.php
+++ b/projects/packages/my-jetpack/src/class-jetpack-manage.php
@@ -1,0 +1,50 @@
+<?php
+/**
+ * Manage the display of an "Jetpack Manage" menu item.
+ *
+ * @package automattic/my-jetpack
+ */
+
+namespace Automattic\Jetpack\My_Jetpack;
+
+use Automattic\Jetpack\Admin_UI\Admin_Menu;
+use Automattic\Jetpack\Connection\Manager as Connection_Manager;
+use Automattic\Jetpack\Redirect;
+
+/**
+ * Jetpack Manage features in My Jetpack.
+ */
+class Jetpack_Manage {
+	/**
+	 * Initialize the class and hooks needed.
+	 */
+	public static function init() {
+		add_action( 'admin_menu', array( self::class, 'add_submenu_jetpack' ) );
+	}
+
+	/**
+	 * The page to be added to submenu
+	 *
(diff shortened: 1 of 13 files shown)


eeeeeevon2023:
1. Let's replace **all of the** external link icons in the Jetpack menu with the one from the WordPress library.  
2. Please also ensure they are vertically aligned with the text (centered). 
3. Make them the same size as in Calypso blue. 20pxx20px
You'll see on Calypso blue they have already updated theirs. We are moving away from Gridicons.
external-link



Screenshot 2023-11-29 at 10 31 47 AM
(attached).

vitozev:
@eeeeeevon2023, this one will be tracked into a separate issue. I'll create it in a bit and will reference this task there.

More info: p1HpG7-pYl-p2#comment-68063

projects/packages/my-jetpack/src/class-jetpack-manage.php
```diff
@@ -0,0 +1,60 @@
...
+		if ( ! isset( $user_data['site_count'] ) ) {
+			return;
+		}
+
+		if ( $user_data['site_count'] <= 2 ) {
+			return;
+		}
+
+		return Admin_Menu::add_menu(
+			__( 'Jetpack Manage', 'jetpack-my-jetpack' ),
```
jeherve, line 53: Should this say "Manage"? I'm asking because the other menu items in the Jetpack plugin and in standalone plugns do not use the Jetpack prefix.

vitozev:
Issue for updating the icons was created: https://github.com/Automattic/jetpack/issues/34395

yashwin (approved):
This works as expected. The code changes look reasonable to me, and the test cases cover all the scenarios. I also tested this with multi-site, and the `Jetpack Manage` menu item wasn't displayed.

> When you return to wp-admin, you should notice a "Jetpack Manage" link with an external icon next to it, right below "My Jetpack".

For me, this is displayed at the end. I hope this is expected. 

Screenshot 2023-12-01 at 9 39 52 AM

darssen:
I've tested with Jetpack plugin and things look good. Link appears in the last position. As I think is expected (although instructions say it should be right below My jetpack)

But I've tested with a standalone like Boost and after connecting, I am unable to see the Jetpack Manage Link. Is this expected?

projects/packages/my-jetpack/src/class-jetpack-manage.php
vitozev: I think Manage might be confusing, it's been discussed here and we'd like to proceed with Jetpack Manage.

p1HpG7-pYl-p2#comment-68026

vitozev:
Thank you, @darssen! You're right about the position: initially it was below My Jetpack, but we decided to place it on the last position. I just updated the instructions.

> But I've tested with a standalone like Boost and after connecting, I am unable to see the Jetpack Manage Link. Is this expected?

Yes, that's expected because if you have a standalone plugin some features might not work as expected (e.g., plugin management).

PR #34358, opened 2023-11-29
Subs / Masterbar: allow all admins to toggle the feature on / off

jeherve:
## Proposed changes:

See https://github.com/Automattic/jetpack/pull/34203#issuecomment-1831628558

All admins should be able to toggle features on and off, just like they can under Jetpack > Settings > Modules.

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

* Go to Jetpack > Settings
* Enable the WordPress.com Toolbar and Subscripions features.
* Go to Users > Add New and create a new admin.
* In a new browser, log in with that user.
* Go to Jetpack > Settings 
    * You should be able to turn the WordPress.com Toolbar and Subscripions features off.

### DIFF
--- a/projects/plugins/jetpack/_inc/client/newsletter/subscriptions-settings.jsx
+++ b/projects/plugins/jetpack/_inc/client/newsletter/subscriptions-settings.jsx
@@ -91,6 +91,9 @@ function SubscriptionsSettings( props ) {
 		);
 	};
 
+	const isDisabled =
+		! isSubscriptionsActive || unavailableInOfflineMode || isSavingAnyOption( [ 'subscriptions' ] );
+
 	return (
 		<SettingsCard { ...props } hideButton module="subscriptions">
 			<SettingsGroup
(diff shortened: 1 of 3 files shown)


darssen:
Hi @jeherve thanks for working on this!

When checking the changes, I can see that now that the toggles for WordPress.com Toolbar and the main toggle for Subscripions are enabled for non-connected Admins.

I have one question though. As you can see below, nested toggles are not enabled for Subscriptions (Newsletter).
image

I am unsure if the nested toggles should be enabled or not. I'm guessing we shouldn't since the idea is to get the same functionality as Jetpack > Settings > Modules. But I feel like it is a bit counter-intuitive.

What do you think?

jeherve:
> I have one question though. As you can see below, nested toggles are not enabled for Subscriptions (Newsletter).

That's a good point. I'm happy to change that too. Done in 7cd56e879b5a864a9f585ff2c7ff41c6507c7aa7

simison:
Pretty small thing but looks like all these toggles could be combined to a single constant?

```js
const isDisabled =
	! isSubscriptionsActive ||
	unavailableInOfflineMode ||
	isSavingAnyOption( [ 'subscriptions' ] );
```

jeherve:
> ```js
> const isDisabled =
> 	! isSubscriptionsActive ||
> 	unavailableInOfflineMode ||
> 	isSavingAnyOption( [ 'subscriptions' ] );
> ```

It certainly wouldn't hurt :) We can even do more extractions in a future PR, but I'm happy to add this to this PR. Done in aadd10dc1f6e1e7f9e971b66925742a9cd2f8edf, thank you.

simison (approved):
Looking good and tested as connected admin and non-connected admin 👍

darssen (approved):
Code looks good. And tests good as well. :)

PR #34174, opened 2023-11-16
Activity Log: add link to Activity Log to Jetpack menu

jeherve:
## Proposed changes:

The Activity Log is a feature that can be very useful for site owners, but that we typically haven't shown a lot until now. This commit has an "Activity Log" menu item to the Jetpack menu, for all Jetpack plugins using "My Jetpack", as soon as the user is connected to WordPress.com.

Just like for the "My Jetpack" menu, we do not show it on Multisite sites since Jetpack Cloud isn't fully compatible with Multisite yet.

image

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

* 668-gh-jetpack-roadmap
* p1HpG7-po1-p2

## Does this pull request change what data or activity we track or use?

* Yes, clicks on the link are tracked (see PCYsg-pY7-p2#monitoring )

## Testing instructions:

* Start with a brand new site, using one of the plugins (either Jetpack or
 standalone)
* You should not see any "Activity Log" menu item by default.
* Connect your site to WordPress.com.
* When you come back to wp-admin, you should notice an "Activity Log" link with an external icon next to it, right below "My Jetpack".
* When clicking on the link, you should be redirected to https://cloud.jetpack/activity-log/yoursite.com
* You can try installing an additional Jetpack plugin, and the link position should not change.
* You can also try to launch a Multisite network; you should not see the "Activity Log" menu item there.

PR #34359, opened 2023-11-29
Admin Page: fix nav menu when Activity Log menu isn't available

jeherve:
Follow-up to #34174

## Proposed changes:

We want to ensure that the admin navigation menu focus will always be correct, regardless of your site's status:

- Can the "My Jetpack" menu be displayed?
- Can the "Activity Log" menu be displayed?

image

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

* Start with a brand new site, using one of the plugins (either Jetpack or
 standalone)
* You should not see any "Activity Log" menu item by default.
* Connect your site to WordPress.com. 
    * During the connection process, stop before you're asked to connect your user, and come back to wp-admin.
* When you come back to wp-admin, you should not see any "Activity Log" menu item.
    * Navigating between the different Jetpack menu pages, the menu focus should be correct.
* Go to the Jetpack menu, and follow the prompts to connect your user to WordPress.com.
* When you come back to wp-admin, you should notice an "Activity Log" link with an external icon next to it, right below "My Jetpack".
    * Focus when switching from one Jetpack menu to the next should be correct.

### DIFF
--- a/projects/plugins/jetpack/_inc/client/main.jsx
+++ b/projects/plugins/jetpack/_inc/client/main.jsx
@@ -647,8 +647,12 @@ class Main extends React.Component {
 
 		if ( this.props.isWoaSite ) {
 			window.wpNavMenuClassChange( { dashboard: 1, settings: 1 } );
-		} else if ( ! this.props.showMyJetpack ) {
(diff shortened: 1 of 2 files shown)


monsieur-z (approved):
Works as intended 👍

PR #23093, opened 2022-02-24
My Jetpack: Remove global notices

renatoagds:
Fixes #22502 

#### Changes proposed in this Pull Request:

* Update `admin` styles to remove global notices.

#### Jetpack product discussion

N/A.

#### Does this pull request change what data or activity we track or use?

No.

#### Testing instructions:

* Goes to `tools/docker/wordpress` in our local env.
* Open `wp-admin/includes/update.php`
* Navigate to `update_nag` method.
* Comment all `if` blocks (probably lines `284-286`, `290-292`, `296-298`).
* Comment current changes applied on this PR.
* Open `My Jetpack` page and you'll see WordPress update version notification.
* Return the code applied on this PR.
* The notice should disappear.

PR #34364, opened 2023-11-29
My Jetpack: do not display any notices from third-parties

### LINKED ISSUE #34360: My Jetpack: do not display other admin notices on "My Jetpack" screen
We currently do not allow other plugins to display admin notices in the main Jetpack dashboard:

https://github.com/Automattic/jetpack/blob/d724ff34599914b17dd003b45c98239b1be3bf87/projects/plugins/jetpack/_inc/client/components/admin-notices/index.jsx#L101-L108

It would be nice to do the same on the My Jetpack screen, to avoid this.

image-1

image-49

It looks like we had taken a first pass at it in #23093.

Reported:

- p8oabR-1oL-p2#comment-7750
- p1HpG7-pNB-p2#comment-67719

jeherve:
Fixes #34360

## Proposed changes:

This is a follow-up to #23093. #23093 mostly works, but some plugins add inline display styles to their notices to be able to style them further. That extra important should fix things.

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

* Find the GoDaddy test site in P8oabR-1n-p2 (it's the first on the page).
* Log in, go to Jetpack > Beta > Jetpack and switch to this branch.
* Go to My Jetpack
    * You should not see any third-party notices at the top of the page.

### DIFF
--- a/projects/packages/my-jetpack/_inc/style.module.scss
+++ b/projects/packages/my-jetpack/_inc/style.module.scss
@@ -12,7 +12,7 @@
 
 	#wpbody-content {
 		& > .notice {
-			display: none;
+			display: none !important;
 		}
 	}
 	#wpwrap {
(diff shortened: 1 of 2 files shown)


darssen (approved):
Followed test instructions and I was able to see how after patching with this branch a 3rd party notice stopped showing in My Jetpack.

PR #34366, opened 2023-11-29
FollowingEndpoint: change permission check

heavyweight:
Fixes #

## Proposed changes:

* The endpoint for retrieving the user's subscriptions and site recommendations is not a site-specific one. But the permission check was a site-specific one. This PR ensures that we have the same permission checks like other related endpoints.
* The issue is most obvious for free users on wpcom. When the permission check runs the current site is the public API 

### Other information:

## Testing instructions:

* Sync this PR to your sandbox
* Go to https://developer.wordpress.com/docs/api/console/
* Login with free wp user:

Ensure that responses are returned for:
```
 wpcom/v2/following/mine
 wpcom/v2/following/recommendations
```


agrullon95 (approved):
Tested both APIs on developer console with new user.
Also tested the blogroll block on site editor/front-end.

PR #34381, opened 2023-11-29
CLI: Use new inquirer-autocomplete-standalone for `jetpack changelog add`

sdixon194:
Fixes: https://github.com/Automattic/garage-backlog/issues/40

## Proposed changes:

This change adds a new package, https://github.com/mokkabonna/inquirer-autocomplete-prompt, that will let users type to autocomplete suggestions, instead of just relying on keyboard navigation when choosing significance and changelog types. This makes the behavior more like the PHP package.

### Other information:

## Jetpack product discussion

pf4qpu-9d-p2

No
## Testing instructions:

Try using `jetpack changelog add` and specify a project, or make some changes to projects and use `jetpack changelog add` to make sure everything works and the changelogger is put through its paces.

### DIFF
--- a/tools/cli/commands/changelog.js
+++ b/tools/cli/commands/changelog.js
@@ -362,7 +362,6 @@ async function changelogAdd( argv ) {
 			)
 		);
 		const response = await promptChangelog( argv, defaultProjects, defaultTypes );
-		console.log( response );
 		for ( const proj of defaultProjects ) {
 			argv = await formatAutoArgs( proj, argv, response );
 			await changelogArgs( argv );
--- a/tools/cli/commands/changelog.js
+++ b/tools/cli/commands/changelog.js
@@ -817,14 +817,14 @@ async function promptChangelog( argv, needChangelog, types ) {
 			},
 		},
 	] );
-	const commands = {
+
+	return {
 		changelogName: changelogName.changelogName,
-		significance: significance,
-		type: type,
+		significance,
+		type,
 		entry: entry.entry,
 		comment: entry.comment,
 	};
-	return { ...commands };
 }
 
 /**
--- a/tools/cli/commands/changelog.js
+++ b/tools/cli/commands/changelog.js
@@ -719,11 +719,12 @@ async function promptVersion( argv ) {
 /**
  * Takes the user input and searches for a matching item in the array.
  *
+ * @param {Array} choices - An array of choices we want to filter matches for.
  * @param {string} input - the user input.
+ *
  * @returns {Array} - the filtered array.
  */
-async function searchSignificance( input = '' ) {
-	const choices = [ 'patch', 'minor', 'major' ];
+async function filterInput( choices, input = '' ) {
 	return choices.filter( choice => choice.toLowerCase().includes( input.toLowerCase() ) );
 }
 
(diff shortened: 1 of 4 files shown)


tools/cli/commands/changelog.js
```diff
@@ -5,6 +5,7 @@ import process from 'process';
 import { fileURLToPath } from 'url';
 import chalk from 'chalk';
 import inquirer from 'inquirer';
+import autoComplete from 'inquirer-autocomplete-standalone';
```
anomiex, line 8: I wonder whether we should use `inquirer-autocomplete-prompt` for now, pending switching from `inquirer` to its rewrite as [`@inquirer/prompts`](https://www.npmjs.com/package/@inquirer/prompts).

Or we could look at switching to [`enquirer`](https://www.npmjs.com/package/enquirer) instead. That one has an autocomplete prompt built in.

tools/cli/commands/changelog.js
```diff
@@ -729,62 +742,67 @@ async function promptChangelog( argv, needChangelog, types ) {
...
-					name: '[minor] Added (or deprecated) functionality in a backwards-compatible manner.',
-				},
-				{
-					value: 'major',
-					name: '[major] Broke backwards compatibility in some way.',
-				},
-			],
+	} );
+	const significance = await autoComplete( {
+		message: 'What is the significance of the change?',
```
anomiex, line 764: IMO we should keep the reference to semver in the question.

The former message was "Significance of the change, in the style of semantic versioning."

tools/cli/commands/changelog.js
```diff
@@ -729,62 +742,67 @@ async function promptChangelog( argv, needChangelog, types ) {
...
+		source: async input => {
+			const significanceArray = Object.keys( significanceChoices );
+			const filteredSignificance = await filterInput( significanceArray, input );
+			return filteredSignificance.map( choice => {
+				return {
+					value: choice,
+					description: significanceChoices[ choice ]
+						? significanceChoices[ choice ]
+						: 'Unknown significance',
```
anomiex, line 773: I don't like the UI this produces, we lose the description of each option inline in the list. This seems to preserve a similar UI to what we had before.
```suggestion
					name: significanceChoices[ choice ],
```

tools/cli/commands/changelog.js
```diff
@@ -729,62 +742,67 @@ async function promptChangelog( argv, needChangelog, types ) {
...
+
+	const type = await autoComplete( {
+		message: 'What type of change?',
+		source: async input => {
+			const typesArray = Object.keys( types );
+			const filteredType = await filterInput( typesArray, input );
+			return filteredType.map( choice => {
+				return {
+					value: choice,
+					description: types[ choice ] ? types[ choice ] : 'Unknown type',
```
anomiex, line 787: Same.
```suggestion
					name: types[ choice ],
```
In this case we'd also want to bring back the `[value]` indicator at the start of the lines ([lines 732–736 from the base version](https://github.com/Automattic/jetpack/blob/44ce10fb1eb5bf6a0ebaa7a936d4868de4c81d1a/tools/cli/commands/changelog.js#L732-L736)), or else have the autocompletion search against the displayed text in addition to the value.

tools/cli/commands/changelog.js
sdixon194: I had looked at `inquirer-autocomplete-prompt` but saw it was only receiving critical bug fixes now and they recommended `inquirer-autocomplete-standalone` - plus we didn't match the version of `inquirer` required by the package, so I went with the standalone.

However doing a cursory look at `enquirer` looks like it might be the better move indeed. It has autocomplete built in and the way it works seems more similar to the existing implementation, so switching might be easier. Let me take a look at that before we move forward here.

sdixon194: https://github.com/Automattic/jetpack/pull/34403

sdixon194:
Closing in favor of https://github.com/Automattic/jetpack/pull/34403

PR #34382, opened 2023-11-29
(Jetpack)(PHP8)(GB 17.1.3) Make the check around `$google_fonts_data`'s array more resilient and return early if key or value is not of the right type

fullofcaffeine:
## Proposed changes:

We've found that in some circumstances (which I couldn't pinpoint yet, exactly) when using PHP8 and Gutenberg 17.1.3+, the `$google_fonts_data['fontFamilies']` would be `null` and cause a fatal. The actual changeset in GB that triggers it (when PHP8 is being used) is this https://github.com/WordPress/gutenberg/pull/55240.This adds a more resilient guard around it, extending the current one, and return early if the shape is not correct.

See p1701220768492429-slack-CDLH4C1UZ and p1701369236916389-slack-C05Q5HSS013 for more context and info.

### Other information:

## Testing instructions:

Steps to reproduce are tricky, see p1701220768492429-slack-CDLH4C1UZ for more info. It needs some WPCOM-specific builds that might not be available in this repo.

### DIFF
--- a/projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
+++ b/projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
@@ -112,7 +112,7 @@ function jetpack_get_theme_fonts_map() {
  */
 function jetpack_register_google_fonts_to_theme_json( $theme_json ) {
 	$google_fonts_data = jetpack_get_google_fonts_data();
-	if ( ! $google_fonts_data ) {
+	if ( ! $google_fonts_data || ! isset( $google_fonts_data['fontFamilies'] ) || ! is_array( $google_fonts_data['fontFamilies'] ) ) {
 		return $theme_json;
 	}
 
--- a/projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
+++ b/projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
@@ -46,7 +46,9 @@ function jetpack_get_google_fonts_data() {
 		}
 	}
 
-	return $data;
+	if ( is_array( $data ) && is_array( $data['fontFamilies'] ) ) {
+		return $data;
+	}
 }
 
 /**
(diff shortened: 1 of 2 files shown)


kraftbj:
Marking as a blocker since the fix has been deployed on WP.com Simple, so need to land this before next daily to ensure we don't regress.

fullofcaffeine:
> Marking as a blocker since the fix has been deployed on WP.com Simple, so need to land this before next daily to ensure we don't regress.

Corresponding diff: D129881-code

fullofcaffeine:
@kraftbj Thanks for adding the changelog!

arthur791004 (dismissed):
Thanks 👍

projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
```diff
@@ -112,7 +112,7 @@ function jetpack_get_theme_fonts_map() {
  */
 function jetpack_register_google_fonts_to_theme_json( $theme_json ) {
 	$google_fonts_data = jetpack_get_google_fonts_data();
-	if ( ! $google_fonts_data ) {
+	if ( ! $google_fonts_data || ! isset( $google_fonts_data['fontFamilies'] ) || ! is_array( $google_fonts_data['fontFamilies'] ) ) {
```
fushar: Idea: the problematic data is returned by `jetpack_get_google_fonts_data()`, which is used in two call sites:

- `jetpack_get_available_google_fonts_map()` (line 58)
- `jetpack_register_google_fonts_to_theme_json()` (here)

I think it will be even more safe if the `jetpack_get_google_fonts_data()` function itself that verifies if `['fontFamilies']` exists and is an array,  so that we don't have to do this multiple times. Because it's this function that parses the JSON from a remote URL, so it should be also its responsibility to provide the correct shape in the return value.

E.g. one possible "bug" is that in line 58, we didn't add the "is_array" check, but here we added that, which is not symmetric.

So something like:

```
if ( ! isset( $data['fontFamilies'] ) || ! is_array( $data['fontFamilies'] ) {
  return null;
}
return $data;
```

What do you think?

fushar (dismissed):

projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
fullofcaffeine: It makes sense! Let's try it out; I'll make the change tomorrow, and I'll run this through the PHP8 pre-migration builds in WPCOM, too.

sdixon194 (approved):

PR #34384, opened 2023-11-30
Add Welcome banner to My Jetpack

IanRamosC:
## Proposed changes:

* This PR adds a new banner to My Jetpack as a Welcome message to Jetpack users. This banner has a close button and is dismissed (after refresh) if the user interacts with any CTA on the page, so the banner is not annoying.

image

### Other information:

## Jetpack product discussion

p1HpG7-pPp-p2#comment-67823

It adds a few more events when the user views and/or clicks to dismiss the banner

## Testing instructions:

* Spin up a new JN site pointing Jetpack Beta to this branch
* Visit My Jetpack and make sure the banner is there
* If you click to close the banner, it will close and won't show again after you refresh the page
* If you click on the main CTAs (Add, Manage, Purchase, Fix), the banner won't show again after you refresh the page (it'll be there if you just go back to the previous page)


IanRamosC:
@keoshi I'd appreciate some input here on the mobile design

projects/packages/my-jetpack/_inc/components/card/index.jsx
```diff
@@ -4,21 +4,27 @@ import PropTypes from 'prop-types';
...
 	const containerClassName = classNames( styles.container, className );
 
+	return <div className={ containerClassName }>{ children }</div>;
+};
+
+const Card = props => {
+	const { title, headerRightContent, className, children } = props;
+
 	return (
-		<div className={ containerClassName }>
```
robertsreberski, line 13: Wouldn't it be enough to just add `classNames( containerClassName, className )` here instead of introducing new component? Don't mind both approaches though.

projects/packages/my-jetpack/_inc/state/actions.js
```diff
@@ -256,6 +256,18 @@ const installStandalonePluginForProduct = productId => async store => {
 	} );
 };
 
+/**
+ * Request to set the welcome banner as dismissed
+ *
+ * @returns {Promise} - Promise which resolves when the banner is dismissed.
+ */
+const dismissWelcomeBanner = () => async () => {
```
robertsreberski: Let's also change the value in the reducer state as well? Currently, we just run side effect, but nothing changes in the app.

I understand we want to dismiss the banner on page refresh, but it shouldn't decide about data flow during runtime of the app. It's more a "view" layer decision what to do with that information that changed.

projects/packages/my-jetpack/_inc/components/my-jetpack-screen/index.jsx
```diff
@@ -124,9 +128,15 @@ export default function MyJetpackScreen() {
 						<div id="jp-admin-notices" className="my-jetpack-jitm-card" />
 					</Col>
 				</Container>
+				<WelcomeBanner
+					onDismissBanner={ dismissWelcomeBanner }
+					dismissedWelcomeBanner={ window?.myJetpackInitialState?.dismissedWelcomeBanner === '1' }
```
robertsreberski: I think it's preferred to use `useSelector` way of accessing state, rather than window object variable. I see that window object is primarily used as a seed for the initial state of the reducer here:
https://github.com/Automattic/jetpack/blob/trunk/projects/packages/my-jetpack/_inc/state/store.js

keoshi:
@IanRamosC Here's a quick mock of how this could show up in smaller devices. Let me know if it needs any adjustments.

image

Figma: https://www.figma.com/file/hABSkFPbbMyVH6sylUVVEk/My-Jetpack%3A-card%2FCTA-consistency?type=design&node-id=3808-10266&mode=design&t=WVKed04OBu0zucFe-11

robertsreberski (commented):
I like the code, and it tests well! I'm just leaving a couple of comments to make it more aligned with Redux approach. I'm happy to talk more about it.

robertsreberski:
Plus: Storybook build fails on github action 🤔 ("Build / Build all projects"). It needs to be resolved before merging.

projects/packages/my-jetpack/_inc/components/card/index.jsx
IanRamosC: Not really, the Card component has a title and a header positioned on the top right that would require some tweaks before I could use it here. Better to export two different components

projects/packages/my-jetpack/_inc/components/welcome-banner/index.jsx
```diff
@@ -0,0 +1,78 @@
...
+ * @returns {object} The WelcomeBanner component.
+ */
+const WelcomeBanner = () => {
+	const { recordEvent } = useAnalytics();
+	const { hasBeenDismissed, dismissWelcomeBanner } = useWelcomeBanner();
+	const [ bannerVisible, setBannerVisible ] = useState( ! hasBeenDismissed );
+
+	useEffect( () => {
+		if ( bannerVisible ) {
+			recordEvent( 'jetpack_myjetpack_welcome_banner_view' );
```
jeherve, line 22: This seems to be triggered a lot of times when the page is initially rendered; maybe we should add some dependencies to avoid triggering it so much?


https://github.com/Automattic/jetpack/assets/426388/9edb7b60-cbd4-4394-b7df-b3ce6c66ed53

jeherve (commented):
This looks pretty. I have a small note on tracking triggers, below.

robertsreberski (dismissed):
Thanks for improvements, looks good to me now! The only comment left is what @jeherve spotted, and I think we're good to go.

robertsreberski (approved):
Dependency array looks good.

jboland88:
Will this banner show for all users when the plugin is updated? Do we want to disable it for users who have been using Jetpack for a while?

PR #34386, opened 2023-11-30
Avoid a PHP Warning in invalid contact form payloads

dd32:
## Proposed changes:

**Data type validation**

A Malformed POST input to a contact form can pass an array of data as the value for a textarea  (which is expected by some other field types), which ultimately causes a PHP Warning within `esc_textarea()`.

```
E_WARNING: htmlspecialchars() expects parameter 1 to be string, array given in wp-includes/formatting.php:4732
```

This simply discards non-stringy data when rendering an `<input>` or `<textarea>` element.

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

* See comment below

### DIFF
--- a/projects/packages/forms/changelog/fix-warning-array-to-string
+++ b/projects/packages/forms/changelog/fix-warning-array-to-string
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Avoid PHP warnings when form has to process malformed data.
--- a/projects/packages/forms/src/contact-form/class-contact-form-field.php
+++ b/projects/packages/forms/src/contact-form/class-contact-form-field.php
@@ -506,6 +506,10 @@ class Contact_Form_Field extends Contact_Form_Shortcode {
 	 * @return string HTML
 	 */
 	public function render_input_field( $type, $id, $value, $class, $placeholder, $required, $extra_attrs = array() ) {
+		if ( ! is_string( $value ) ) {
+			$value = '';
+		}
+
 		$extra_attrs_string = '';
 
 		if ( ! empty( $this->field_styles ) ) {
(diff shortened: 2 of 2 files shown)


jeherve:
@dd32 Thanks for the report! Do you happen to have an example on how we can reproduce this?

dd32:
> Do you happen to have an example on how we can reproduce this?

Thanks for the changelog @jeherve!

The way I came across this was through a request similar to this from a pentester, duplicated into a curl command here for when I was testing:

```
curl https://example.org/report/ --data 'action=1&ak_hp_textarea=555&ak_js=1&contact-form-hash=1&contact-form-id=1&g90-textareaformentry%5B%24testing%5D=1'
```

Notably, the form expected something like this as the POST payload:
```
array(6) {
  'action' =>
  string(1) "1"
  'ak_hp_textarea' =>
  string(3) "555"
  'ak_js' =>
  string(1) "1"
  'contact-form-hash' =>
  string(1) "1"
  'contact-form-id' =>
  string(1) "1"
  'g90-textareaformentry' =>
  string(1) "TextFormEntry"
}
```

but got sent this:
```
array(6) {
  'action' =>
  string(1) "1"
  'ak_hp_textarea' =>
  string(3) "555"
  'ak_js' =>
  string(1) "1"
  'contact-form-hash' =>
  string(1) "1"
  'contact-form-id' =>
  string(1) "1"
  'g90-textareaformentry' =>
  array(1) {
    '$testing' =>
    string(1) "1"
  }
}
```

The array value in `$value` probably comes from this entrypoint:
https://github.com/Automattic/jetpack/blob/75365cddb2dae8814442c70e0314a2545b50aab0/projects/packages/forms/src/contact-form/class-contact-form-field.php#L216-L223
(which again, is expected for things like select, radio, and checkboxes amongst others probably)

jeherve (approved):

PR #34390, opened 2023-11-30
Limit amount of data send by full sync

darssen:
## Proposed changes:

We are seeing many OOMs caused by **Full Sync** which tries to send as much data as possible, ignoring the `max_upload_size` restriction, due to the nature how Jetpack Sync queues items. 

This is an approach to send smaller chunks when in **Full Sync** if they are bigger than a specific threshold only for **Post objects** since they are the ones that feel like might be a problem based on size. 

Prior logic used to chunk Posts only based on defaults using `chunk_size` and `max_chunks`. This PR overrides the `get_next_chunk` method for the `Posts` Module and incorporates existing logic about expanding posts and adding metadata that was being done later down the road in the Full Sync Process. Also, metadata trimming has been added to the process since it did not exist for **Full Sync**.

The idea is to make sure we limit the max amount of posts and their metadata associated to a maximum size. (For the moment it is  `MAX_POST_META_LENGTH + MAX_POST_CONTENT_LENGTH`) All of it after trimming both posts and metadata.

In the edge case of the first post and its associated metadata being bigger than the max size allowed after trimming the content, we allow only that one post to be synced. Otherwise, we will never be able to sync that object. We might need to monitor that edge case to see if it still causes issues.

### Other information:

## Jetpack product discussion

pf5801-gq-p2

## Testing instructions:

Let's start with a self-hosted site without this branch applied (be it your local or a JN)
* Create a bunch of posts (more than 100) with some big content. I suggest creating a very big one and using some plugin to easily duplicate
* Go to Jetpack Debugger and Schedule a Full Sync for posts only.
* Verify the content was synced by chunk sizes of 100 checking the corresponding sync-audit logs (link provided in the Jetpack Debugger but change the queue to `immediate-send`). 

Now move the site to this branch
* Go to Jetpack Debugger 
* Just in case, clean the Full Sync queue by scheduling a Full Sync for options, constants, functions and updates.
* Schedule a Full Sync for posts only.
* Verify the content was synced by smaller chunk sizes than 100 checking the corresponding sync-audit logs (link provided in the Jetpack Debugger but change the queue to `immediate-send`). The specific chunking size will depend on how big your posts are.


projects/packages/sync/src/class-sender.php
```diff
@@ -529,6 +539,10 @@ public function get_items_to_send( $buffer_or_items, $encode = true ) {
 				$skipped_items_ids[] = $key;
 				continue;
 			}
+			if ( $this->action_is_full_sync( $item[0] ) && strlen( wp_json_encode( $item ) ) > self::MAX_SIZE_FULL_SYNC ) {
```
darssen: I initially thought of doing this check after encoding, and comparing it against `$upload_size` instead but I did get some `Fatal error: Maximum execution time of 30 seconds exceeded` when testing.

projects/packages/sync/src/modules/class-module.php
```diff
@@ -346,6 +346,15 @@ public function send_full_sync_actions( $config, $status, $send_until ) {
 
 			$result = $this->send_action( 'jetpack_full_sync_' . $this->name(), array( $objects, $status['last_sent'] ) );
 
+			if ( is_wp_error( $result ) && 'full_sync_max_upload_exceeded' === $result->get_error_code() && $limits['chunk_size'] > 1 ) {
+				// If we hit the max upload size, we need to reduce the chunk size and try again.
+				// Also resetting send_until and chunks_sent to only send one chunk with the new chunk_size.
+				$limits['chunk_size'] = (int) ( $limits['chunk_size'] / 2 );
+				$chunks_sent          = $limits['max_chunks'] - 1;
+				$send_until           = microtime( true ) + Settings::get_setting( 'full_sync_send_duration' );
```
darssen: Not sure if there is a better way to "reset" `$send_until`

projects/packages/sync/src/class-sender.php
fgiannar: That's strange. I agree it's best to check after encoding and avoid encoding the item twice. Let's figure out and fully understand why this results in Fatal

projects/packages/sync/src/modules/class-module.php
fgiannar: I wonder if we actually need to reset `$send_until`. It would make more sense to me to keep it as is since the Full Sync duration should be independent of how we chunk the items.

fgiannar (commented):
Amazing work, Juanma! Thanks for working on this!

I left some initial thoughts to start a discussion. Mostly focusing on the logic for now.

projects/packages/sync/src/modules/class-module.php
darssen: From what I've gathered this `$send_until` is for the current `send` request in https://github.com/Automattic/jetpack/blob/trunk/projects/packages/sync/src/modules/class-full-sync-immediately.php#L381 for each module, and it is just set to 9 microseconds after. So in our case in the next iteration of the while, the check below in LINE 343 will make us return early.
```
if ( $chunks_sent++ === $limits['max_chunks'] || microtime( true ) >= $send_until ) {
	return $status;	
}
```
Not sure how are we getting past this for sync objects with `$limits['max_chunks']` greater than 1 though.

projects/packages/sync/src/modules/class-module.php
```diff
@@ -340,19 +340,18 @@ public function send_full_sync_actions( $config, $status, $send_until ) {
...
 			if ( is_wp_error( $result ) || $wpdb->last_error ) {
 				$status['error'] = true;
 				return $status;
 			}
 			// The $ids are ordered in descending order.
-			$status['last_sent'] = end( $objects );
-			$status['sent']     += count( $objects );
+			$status = $this->set_send_full_sync_actions_status( $status, $objects );
+
+			if ( $chunks_sent++ === $limits['max_chunks'] || microtime( true ) >= $send_until ) {
```
fgiannar:
```suggestion
			if ( ++$chunks_sent === $limits['max_chunks'] || microtime( true ) >= $send_until ) {
```
So that we increment first, compare after :) 

Moving this check in the end of the loop is a performance optimization that will save us 1 DB query plus the overkill of expanding items before sending.

fgiannar (approved):
Many thanks for working on this, Juanma!

The logic is solid, I made sure to fully test in my local env and on a JN site with/without `MAX_SIZE_FULL_SYNC` enforced.
Without the `MAX_SIZE_FULL_SYNC` enforced, we are fully syncing 100 posts at the time as we used to.
With the `MAX_SIZE_FULL_SYNC` enforced, we are fully syncing in chunks < 100, depending on the limit.
I also compared the remote/Cache site posts/tags/relationships and they also match.

Appreciate the unit tests too 👍 

As a reminder, lets make sure to refactor `send_full_sync_actions`  in a follow up PR as we are currently making 1 redundant query/processing of items (not related to the current PR tho)

PR #34391, opened 2023-11-30
Jetpack Social | Fix admin page connections UI

manzoorwanijk:
In #34064, we made some changes to the store and thus admin page doesn't know anything about the connections and it always says "Connect accounts" even if there are accounts connected.

## Proposed changes:

* Update the admin header to feed it the data from initial state.

### Other information:

## Testing instructions:

* Install Jetpack Social - `jetpack install plugins/social`
* Remove all the Social Media connections for the site
* Goto Social admin page - `/wp-admin/admin.php?page=jetpack-social`
* Confirm that you see the "Connect accounts" button
* Now connect one or more social media accounts for the site
* Reload the admin page
* Confirm that you no longer see the "Connect accounts" button

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2023-11-30 at 5 54 02 PM | Screenshot 2023-11-30 at 5 51 46 PM |

### DIFF
--- a/projects/plugins/social/changelog/fix-social-admin-page
+++ b/projects/plugins/social/changelog/fix-social-admin-page
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Fixed broken connections UI
--- a/projects/plugins/social/src/js/components/header/index.js
+++ b/projects/plugins/social/src/js/components/header/index.js
@@ -18,6 +18,7 @@ import StatCards from '../stat-cards';
 import styles from './styles.module.scss';
 
 const Header = () => {
+	const connectionData = window.jetpackSocialInitialState.connectionData ?? {};
 	const {
 		connectionsAdminUrl,
 		hasConnections,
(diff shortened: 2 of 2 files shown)


pablinos (approved):
This fixes it!

pablinos:
@jeffgolenski and/or @ilonagl - Looking at the 'After' screenshot above. Do you think the 'Write a Post' button should become primary once the site has connections? Initially the primary CTA is to connect networks, but after that we want people to write things to share, and the admin page has no primary CTA in this state.

ilonagl:
> @jeffgolenski and/or @ilonagl - Looking at the 'After' screenshot above. Do you think the 'Write a Post' button should become primary once the site has connections? Initially the primary CTA is to connect networks, but after that we want people to write things to share, and the admin page has no primary CTA in this state.

Yes, making that button primary would be the best approach! It helps to guide user's actions.

jeffgolenski:
Good eye, Paul! Yes sir!

manzoorwanijk:
@pablinos I have created a task for that which you can see mentioned above.

PR #34392, opened 2023-11-30
Update share limits bar design

manzoorwanijk:
## Proposed changes:

* Update the share limits bar design to match the Figma design - 25F3UyoNr6TUSWBplnpRyM-fi-4007%3A5874

### Other information:

## Testing instructions:

* Create/edit a post on a site without Jetpack Social plan
* Confirm that the limits bar works as expected on the sidebar of the editor
* Goto Jetpack Social admin page
* Confirm that the bar works as expected

Different scenarios

- Used: 20, Scheduled: 0

   Screenshot 2023-12-06 at 2 33 51 PM

- Used: 25, Scheduled: 0

    Screenshot 2023-12-06 at 2 34 39 PM

- Used: 25, Scheduled: 3

    Screenshot 2023-12-06 at 2 35 36 PM

- Used: 25, Scheduled: 5

    Screenshot 2023-12-06 at 2 36 23 PM

- Used: 25, Scheduled: 10

    Screenshot 2023-12-06 at 2 37 08 PM

- Used: 30, Scheduled: 0

    Screenshot 2023-12-06 at 2 37 46 PM

- Jetpack Social admin page

    Screenshot 2023-12-06 at 2 17 39 PM

    Screenshot 2023-12-06 at 2 40 09 PM

    Screenshot 2023-12-06 at 2 41 18 PM


projects/js-packages/publicize-components/src/components/share-limits-bar/index.tsx
```diff
@@ -1,53 +1,115 @@
...
+				backgroundColor: colorsForScheduled[ noticeType ],
+				label: __( 'scheduled', 'jetpack' ),
 			},
 			{
-				count: scheduledCount,
-				backgroundColor: 'var(--jp-gray-20)',
-				label: __( 'Shares scheduled', 'jetpack' ),
+				count: remaining,
+				backgroundColor: 'var(--jp-gray-off)',
+				label: __( 'left', 'jetpack' ),
```
manzoorwanijk: `left` may be confusing for translators. May be we should add a comment for them that it means "remaining" not the direction "left/right".

projects/js-packages/publicize-components/src/social-store/resolvers.js
```diff
@@ -75,11 +75,7 @@ export function getConnections() {
 		const connections = editor.getEditedPostAttribute( 'jetpack_publicize_connections' );
 
 		dispatch( setConnections( connections || [] ) );
-
-		// For now, we are interested in the initial connections only if the post is scheduled.
-		if ( editor.isCurrentPostScheduled() ) {
-			dispatch( setInitialConnections( connections || [] ) );
-		}
+		dispatch( setInitialConnections( connections || [] ) );
```
pablinos: Whether to subtract the initial connections count is dependent on the post date, and as that can be edited, I moved it to the useScheduledPost hook. The problem is that we need to do different things based on if the post date moves in or out of the 30 day window.

I think we might be able to solve it by storing the initial date in a ref and seeing if the edited date is moving in or out of the 30 day window.

projects/js-packages/publicize-components/src/social-store/selectors/shares-data.ts
```diff
@@ -85,17 +85,10 @@ export function getTotalSharesCount(
...
-	if ( options.excludeInitialConnections ) {
-		count -= getInitialEnabledConnectionsCount( state );
-	}
-
-	if ( ! options.includeEnabledConnections ) {
-		return count;
-	}
-
-	const enabledConnections = getEnabledConnections( state );
+	count -= options.initialEnabledConnectionsCount ?? 0;
```
pablinos: Again, I'm trying move this logic and calculations to the hooks. I think we might not need `getTotalSharesCount` at all now

projects/js-packages/publicize-components/src/social-store/selectors/shares-data.ts
```diff
@@ -149,16 +142,15 @@ export function numberOfSharesRemaining(
...
 		...options,
 	};
 
 	const sharesUsed = getSharesUsedCount( state );
 	const sharesLimit = getShareLimit( state );
 	const scheduledShares = includeScheduled ? getScheduledSharesCount( state ) : 0;
-	const enabledConnections = includeEnabledConnections ? getEnabledConnections( state ) : [];
 
-	return Math.max( sharesLimit - sharesUsed - scheduledShares - enabledConnections.length, 0 );
+	return Math.max( sharesLimit - sharesUsed - scheduledShares - enabledConnectionsCount, 0 );
```
pablinos: Similarly, I think we won't need this either, because the 'enabled' connections are based on various pieces of information from both the social and editor stores, as well as the validation information. Perhaps we should move that into the store, so it's associated with the connection.

projects/plugins/social/src/js/components/header/index.js
```diff
@@ -82,14 +88,11 @@ const Header = () => {
 					{ showShareLimits ? (
 						<>
 							<ShareLimitsBar
-								maxCount={ shareLimit }
-								currentCount={ sharesCount }
+								limit={ shareLimit }
```
pablinos: This will be broken now as I've changed the props. I'll try and update it next.

projects/js-packages/components/components/record-meter-bar/index.tsx
```diff
@@ -18,6 +18,10 @@ type RecordMeterBarItem = {
 	 * Color code for the background color for the item
 	 */
 	backgroundColor: string;
+	/**
+	 * Width of the item in percentage.
+	 */
+	widthPercent?: number;
```
pablinos: We shouldn't need this change now.

projects/js-packages/publicize-components/src/components/form/share-count-notice.tsx
```diff
@@ -30,31 +31,33 @@ export const ShareCountNotice: React.FC = () => {
...
-								query: 'redirect_to=' + encodeURIComponent( window.location.href ),
-							} ) }
-							onClick={ autosaveAndRedirect }
-						/>
-					),
-				}
-			) }
-		</Notice>
+		<PanelRow>
+			<Notice type="warning">
```
pablinos: The designs have the notice as always being a warning.

projects/js-packages/publicize-components/src/components/share-limits-bar/index.tsx
```diff
@@ -1,51 +1,78 @@
...
 			{
-				count: scheduledCount,
-				backgroundColor: 'var(--jp-gray-20)',
-				label: __( 'Shares scheduled', 'jetpack' ),
+				count: remainingCount,
+				backgroundColor: 'var(--jp-gray-off)',
+				label:
+					// translators: This is 'left' referring to the quantity remaning, not the direction.
+					__( 'left', 'jetpack' ),
 			},
```
pablinos: I can't say I'm proud of this! I think there is probably a way to do this more simply

projects/js-packages/publicize-components/src/hooks/use-scheduled-post/index.js
```diff
@@ -0,0 +1,36 @@
...
+				new Date().getTime() -
+					new Date( select( editorStore ).getEditedPostAttribute( 'date' ) ).getTime()
+			) /
+			( 1000 * 3600 * 24 );
+		const isScheduledWithin30Days = isScheduled && daysUntilPublish <= 30;
+
+		return {
+			isScheduled,
+			daysUntilPublish,
+			isScheduledWithin30Days,
```
pablinos: I think we might want to create a ref that is something like `initiallyScheduledWithin30Days`, then we can spot if the date is changing in a way that would affect how we should adjust the counts.

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
```diff
@@ -1,123 +1,135 @@
...
+	);
+}
+
+/**
+ * Wraps the useShareLimits hook with the current post context, so we adjust the share
+ * counts according to the connections selected on the post.
+ *
+ * @returns {ShareLimits} Share limits details
+ */
+export function usePostShareLimits() {
```
pablinos: Maybe this should be in a separate file.

pablinos (commented):
There's still a fair amount to do here:

1. Take into account all the other logic around whether a connection is enabled/disabled. If a connection is broken, invalid (like with media requirements for Instagram), etc. Then we will still count it as 'enabled' for the share counts. The logic for this probably needs to move into the useSocialMediaConnections hook so we can set the `disabled` property on each connection correctly.  This might well be broken on trunk too, so we could also do it as a follow-up
2. We need to fix the logic around a post date being set inside or outside of the 30 day window we look at for scheduled shares. I've added some comments on what we could do.
3. Add the line "This post will be shared to x connection(s)". I was thinking instead of saying "This post won't be shared to any connections" when there aren't any enabled, we could say something like "Enable a connection to share this post".
4. Fix the styling of the bar in the Social admin page
5. Fix the tests!

pablinos:
We also need to update the notice styling so the upgrade link is a button, ande we adde the Learn more link.

I think we should always have an upgrade link whether there is a warning or not.

projects/js-packages/components/components/record-meter-bar/index.tsx
manzoorwanijk: Reverted

projects/js-packages/publicize-components/changelog/update-share-limits-bar-design
```diff
@@ -0,0 +1,4 @@
+Significance: patch
```
pablinos: We seem to always make things a patch. Is this more significant than that? Perhaps the feature branch one should be minor/major

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
```diff
@@ -1,123 +1,103 @@
...
  * @returns {ReturnType<typeof getMessages>} Share limits messages
  */
-export function getMessages( shareLimit: number ) {
+export function getMessages( remainingCount: number ) {
+	const remaining = Number.isFinite( remainingCount )
+		? sprintf(
+				// translators: %d: The number of shares to social media remaining
+				_n(
+					'You have %d auto-share remaining.',
+					'You have %d auto-shares remaining.',
```
pablinos, line 31: I've noticed the designs have 'available' in here. I'm not sure if that's necessary or not?

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
```diff
@@ -1,123 +1,103 @@
...
 	return {
-		default: sprintf(
-			// translators: %1$d is the number of shares allowed in 30 days.
-			__( 'Share limit for 30 days: %1$d.', 'jetpack' ),
-			shareLimit
-		),
+		default: remaining,
 		exceeded: __(
-			'You have exceeded your share limit. Your posts will no longer be shared.',
+			'You have have reached your auto-share. Scheduled posts will not be shared until shares become available.',
```
pablinos: This looks like a typo:
```suggestion
			'You have reached your auto-share limit. Scheduled posts will not be shared until shares become available.',
```

pablinos (commented):
Overall, I think this is good. I've got a few minor faults. Probably the only critical one is the change to the notice text.

Looking at the designs it seems that although the grey in the legend looks darker, the bar is actually a bit lighter, which I think is better a colour because otherwise the light green for the 'scheduled' count, gets a little lost.

When they are not nearing the limit, I think it would better to have a link to upgrade rather than the warning notice. We could commandeer the 'Learn more' link from the designs, and have something like we have now. 'Upgrade to share more' or something.

Also, I know we said to keep it not reactivity but what do you think about adding the 'This post will be shared to x connections' line? We should be able to filter for all the invalid ones too, so we don't have the problem with Instagram looking like it's enabled.


CleanShot 2023-12-06 at 21 11 35@2x

I know this is a variation on the designs, but because we're not using the current post in the meter anymore, I wondered if it should turn red sooner. We change it to yellow when there are 20% of shares remaining, maybe it could go red with 5-10% remaining? This is just a thought though, feel free to stick to the designs!

On the admin page it looks like there is a lot of space between the bottom of the legend an the upgrade nudge. I wonder if we should remove the padding and rely on the upgrade nudge top margin?

image

manzoorwanijk:
> Looking at the designs it seems that although the grey in the legend looks darker, the bar is actually a bit lighter, which I think is better a colour because otherwise the light green for the 'scheduled' count, gets a little lost.

If you inspect that in the dev tools, both are the same - `--jp-gray-off`

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
manzoorwanijk: > "You have 1 available auto-share remaining"

That looks confusing to me due to having both "available" and "remaining"

projects/js-packages/publicize-components/src/hooks/use-share-limits/index.ts
manzoorwanijk: Fixed. Thanks

projects/js-packages/publicize-components/changelog/update-share-limits-bar-design
manzoorwanijk: Updated it to "minor"

manzoorwanijk:
> When they are not nearing the limit, I think it would better to have a link to upgrade rather than the warning notice. We could commandeer the 'Learn more' link from the designs, and have something like we have now. 'Upgrade to share more' or something.

I think it looks good as a warning always. The bar being green is enough to show that they are under the safe limits.

manzoorwanijk:
> I know this is a variation on the designs, but because we're not using the current post in the meter anymore, I wondered if it should turn red sooner. We change it to yellow when there are 20% of shares remaining, maybe it could go red with 5-10% remaining? This is just a thought though, feel free to stick to the designs!

For a limit of 30 shares, 80% sounds like a good choice because they will have only 6 shares left. If we make it 90%, they will have only 3 shares left.

manzoorwanijk:
> I know we said to keep it not reactivity but what do you think about adding the 'This post will be shared to x connections' line? We should be able to filter for all the invalid ones too, so we don't have the problem with Instagram looking like it's enabled.

Done ✅ 
Screenshot 2023-12-07 at 1 42 31 PM

projects/js-packages/publicize-components/src/components/form/styles.module.scss
```diff
@@ -27,8 +29,28 @@
...
+		.record-meter-bar__items {
+			height: 0.5rem;
+			margin-bottom: 0.8em;
+		}
+	
+		.record-meter-bar__legend--item-circle {
+			width: 0.5rem;
+			height: 0.5rem;
+		}
+	}
```
manzoorwanijk: This needed some specificity to make it override the ShareLimitsBar component styles.

manzoorwanijk:
> On the admin page it looks like there is a lot of space between the bottom of the legend an the upgrade nudge. I wonder if we should remove the padding and rely on the upgrade nudge top margin?

I have removed the padding on the admin page.

gmjuhasz (approved):
Let's merge it into the main branch and give it a final test

pablinos:
> > I know we said to keep it not reactivity but what do you think about adding the 'This post will be shared to x connections' line? We should be able to filter for all the invalid ones too, so we don't have the problem with Instagram looking like it's enabled.
> 
> Done ✅ Screenshot 2023-12-07 at 1 42 31 PM

Thanks! Should we have something like "Enable a connection to share this post when it is published" or something along those lines, when there are none selected? That might be repeating what gets displayed in the text above the connections though.

pablinos:
> > I know this is a variation on the designs, but because we're not using the current post in the meter anymore, I wondered if it should turn red sooner. We change it to yellow when there are 20% of shares remaining, maybe it could go red with 5-10% remaining? This is just a thought though, feel free to stick to the designs!
> 
> For a limit of 30 shares, 80% sounds like a good choice because they will have only 6 shares left. If we make it 90%, they will have only 3 shares left.

Yeah I was thinking that it turns red at 100% currently. Maybe as it's not reactive now that would be better to happen at 90%?

pablinos:
> I think it looks good as a warning always. The bar being green is enough to show that they are under the safe limits.

As mentioned in Slack, I think having it as a warning always will make people blind to it. I think we should only 'warn' them when there is something to warn them about. I was worried about jank with the notice moving from below to above the connections, but now it's not reactive, that shouldn't be a problem.

manzoorwanijk:
> Should we have something like "Enable a connection to share this post when it is published" or something along those lines, when there are none selected? That might be repeating what gets displayed in the text above the connections though.

It is visible for the posts that are already published. So, "...when it is published" won't make sense.

pablinos:
> > Looking at the designs it seems that although the grey in the legend looks darker, the bar is actually a bit lighter, which I think is better a colour because otherwise the light green for the 'scheduled' count, gets a little lost.
> 
> If you inspect that in the dev tools, both are the same - `--jp-gray-off`

Sure, but in the designs, the bar at least is #F6F7F7 which is much lighter

manzoorwanijk:
> I was thinking that it turns red at 100% currently. Maybe as it's not reactive now that would be better to happen at 90%?

That will further complicate the logic. May be we can do that in a follow up PR.

manzoorwanijk:
> in the designs, the bar at least is #F6F7F7 which is much lighter

That is `--jp-gray-0` which gets lost in the admin page

Screenshot 2023-12-07 at 3 02 38 PM

manzoorwanijk:
> As mentioned in Slack, I think having it as a warning always will make people blind to it. I think we should only 'warn' them when there is something to warn them about. I was worried about jank with the notice moving from below to above the connections, but now it's not reactive, that shouldn't be a problem.

I have updated it to look like this
Screenshot 2023-12-07 at 2 55 48 PM

manzoorwanijk:
Here is the updated look of the title and the notice

Screenshot 2023-12-07 at 3 46 15 PM
Screenshot 2023-12-07 at 3 58 21 PM

PR #34401, opened 2023-11-30
Contact Form: revalidate fields on blur

### LINKED ISSUE #34281: Enhancement: validate Contact Form fields as user types
### Impacted plugin

Jetpack

### What

When user is fixing errors in the form, validate the field as they type.

### How

_No response_

monsieur-z:
Fixes https://github.com/Automattic/jetpack/issues/34281

## Proposed changes:

https://github.com/Automattic/jetpack/pull/34173 made the form validation more accessible. This PR iterates on it by revalidating the invalid fields on blur (i.e., focus out) so that users receive quicker feedback. We chose not to revalidate on input change because this behavior is confusing/irritating for most inputs.

### Other information:

## Jetpack product discussion

Project thread: pf5801-87-p2

No.

## Testing instructions:

Review in conjunction with https://github.com/Automattic/jetpack/pull/34400.

- Create a new post
- Add a _Contact form_ block to the post
- Add at least a text, Multiple Choice, and Single Choice fields
- Make them required
- View the preview
- Check that when one or more fields are invalid, an error message is displayed above the submit button and below each invalid field
- Fix the error of each field and notice the error message disappears when focusing out
- Make sure you can submit the form successfully


projects/packages/forms/src/contact-form/js/accessible-form.js
```diff
@@ -41,6 +41,9 @@ const initForm = form => {
 		hasInsetLabel: hasFormInsetLabels( form ),
 	};
 
+	// Hold references to the input event listeners.
+	let inputListenerMap = {};
```
monsieur-z, line 45: We need to keep a reference to the event handlers we'll set so we can remove them later on.

sergeymitr:
Hi @monsieur-z.
I noticed a potential issue with multiple choice validation.
It only runs real-time validation if there are other invalid fields. If all other fields are valid, the form can be submitted.
The error still shows up after page reload, so it's not a big problem, but JS validation does ignore the field for some reason.

This doesn't seem to be a new problem though, as it also happens on `trunk`.

Steps to reproduce:
1. Create the contact form as per the testing instructions.
2. Fill up all the fields but leave the "multiple choice" one empty.
3. Submit the form. No JS-triggered errors, form is allowed to be submitted.

monsieur-z:
Good catch @sergeymitr, thanks! I've pushed a fix.

sergeymitr (dismissed):
Works well 👍

sergeymitr (approved):

PR #34403, opened 2023-11-30
CLI: Try using enquirer to handle autocomplete prompts for changelog add

sdixon194:
An alternative to https://github.com/Automattic/jetpack/pull/34381

## Proposed changes:

Uses enquirer instead of inquirer to handle autocompletion when searching for changes.

This just handles the prompts for the `jetpack changelog add` command, and doesn't replace inquirer everywhere. If this is the direction we want to go later, we can look into that.
### Other information:

## Jetpack product discussion

pf4qpu-9d-p2

no
## Testing instructions:

Put the changelogger through its paces:

Try using `jetpack changelog add` and specify a project, or make some changes to projects and use `jetpack changelog add` to make sure everything works and the changelogger is put through its paces.

### DIFF
--- a/tools/cli/commands/changelog.js
+++ b/tools/cli/commands/changelog.js
@@ -747,8 +747,7 @@ async function promptChangelog( argv, needChangelog, types ) {
 		validate: input => {
 			const fileExists = doesFilenameExist( input, needChangelog );
 			if ( fileExists ) {
-				console.log( chalk.red( 'Please choose another file name.' ) );
-				return 'Please choose another file name, or delete the file manually.';
+				return 'Filename exists already. Please choose another file name, or delete the file existing manually.';
 			}
 			return true;
 		},
--- a/tools/cli/commands/changelog.js
+++ b/tools/cli/commands/changelog.js
@@ -779,9 +779,8 @@ async function promptChangelog( argv, needChangelog, types ) {
 		type: 'autocomplete',
 		name: 'type',
 		message: 'Type of change.',
-		suggest: ( input, typeChoices = choices.map( choice => choice.value ) ) => {
-			return Promise.resolve( typeChoices.filter( choice => choice.value.startsWith( input ) ) );
-		},
+		suggest: ( input, choices ) => choices.filter( choice => choice.value.startsWith( input ) ),
+		highlight: v => v,
 		choices: choices,
 	} );
 
--- a/tools/cli/commands/changelog.js
+++ b/tools/cli/commands/changelog.js
@@ -341,14 +341,6 @@ async function changelogAdd( argv ) {
 		return;
 	}
 
-	console.log(
-		chalk.yellow(
-			"When writing your changelog entry, please use the format 'Subject: change description.'\n" +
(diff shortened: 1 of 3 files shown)


tools/cli/commands/changelog.js
```diff
@@ -369,7 +370,8 @@ async function changelogAdd( argv ) {
 
 	for ( const proj of uniqueProjects ) {
 		console.log( chalk.green( `Running changelogger for ${ proj }!` ) );
-		const response = await promptChangelog( argv, proj, projectChangeTypes[ proj ] );
+		const projArray = [ proj ];
+		const response = await promptChangelog( argv, projArray, projectChangeTypes[ proj ] );
```
anomiex: Does it not work to just do like this?
```suggestion
		const response = await promptChangelog( argv, [ proj ], projectChangeTypes[ proj ] );
```

tools/cli/commands/changelog.js
```diff
@@ -734,72 +737,100 @@ async function promptChangelog( argv, needChangelog, types ) {
...
+	// Get the significance.
+	const { significance } = await prompt( {
+		type: 'autocomplete',
+		name: 'significance',
+		message: 'Significance of the change, in the style of semantic versioning.',
+		suggest: ( input, significanceChoices = [ 'patch', 'minor', 'major' ] ) => {
+			return Promise.resolve(
+				significanceChoices.filter( choice => choice.value.startsWith( input ) )
+			);
 		},
```
anomiex: I think you overthought this.
```suggestion
		suggest: ( input, choices ) => choices.filter( choice => choice.value.startsWith( input ) ),
```

Although I note the highlighting enquirer applies is wonky, it doesn't match what's actually being searched. Maybe we should just turn it off.
```suggestion
		suggest: ( input, choices ) => choices.filter( choice => choice.value.startsWith( input ) ),
		highlight: v => v,
```

tools/cli/commands/changelog.js
```diff
@@ -734,72 +737,100 @@ async function promptChangelog( argv, needChangelog, types ) {
...
+	} );
+
+	// Get the type of change.
+	const { type } = await prompt( {
+		type: 'autocomplete',
+		name: 'type',
+		message: 'Type of change.',
+		suggest: ( input, typeChoices = choices.map( choice => choice.value ) ) => {
+			return Promise.resolve( typeChoices.filter( choice => choice.value.startsWith( input ) ) );
 		},
```
anomiex: Same as above
```suggestion
		suggest: ( input, choices ) => choices.filter( choice => choice.value.startsWith( input ) ),
		highlight: v => v,
```

tools/cli/commands/changelog.js
sdixon194: Ah yeah I did indeed.

sdixon194:
Should be ready for review again, also moved where the changelog example text appears so it's right before the actual changelog entry prompt instead of as soon as you run the command now.

anomiex (approved):
Let's do it!

PR #34406, opened 2023-11-30
cli: Fix build execa `stdio` null handling

anomiex:
## Proposed changes:

In stock node `child_process` and in default `execa`, passing null for a stream in `stdio` is equivalent to "pipe" for stdin/stdout/stderr.

But in the wrapper we define in the build command, we instead had it treating null as either "ignore" or "inherit" depending on the `-v` option. This non-standard interpretation crept into `helpers/list-project-files.js` and caused me some confusion when trying to debug something.

Let's bring the wrapper's behavior closer to the default. We still interpret "inherit" according to `-v`, but otherwise it should be the same now.

### Other information:

## Testing instructions:

* Check the output from `jetpack build` with and without `-v`, it should look right.


sdixon194:
Seems to build fine, but needs to resolve conflicts now.

sdixon194 (approved):

PR #34407, opened 2023-11-30
cli: Avoid node bug in process plumbing

anomiex:
## Proposed changes:

In `listProjectFiles` we first run `git ls-files` and process its output, then run a second `git ls-files` with different options and piped to `git check-attr` and process its output.

To keep the promise handling less confusing, we create all those processes at once and only attach the second pipe's output to the processor after the first pipe finishes.

But it seems Node has a bug (or at least unexpected behavior) where it throws away its internal stream buffer for the second pipe's output if the final process exits before the first pipe finishes.

The most logical (if not the most straightforward to code) way to handle that is to not even create the second pipe until the first one finishes.

### Other information:

## Jetpack product discussion

I think this is the cause of p1701368080443519/1701367288.275599-slack-C034JEXD1RD

No

## Testing instructions:

* Build working?

### DIFF
--- a/tools/cli/helpers/list-project-files.js
+++ b/tools/cli/helpers/list-project-files.js
@@ -15,26 +15,25 @@ export async function* listProjectFiles( src, spawn ) {
 	//    ls-files --ignored | check-attr production-include | filter
 	//  } | check-attr production-exclude | filter
 
-	const lsFiles = spawn( 'git', [ '-c', 'core.quotepath=off', 'ls-files' ], {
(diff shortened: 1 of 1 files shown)


sdixon194 (approved):
Seems logical and the build passes

anomiex:
Build passes, but also nothing was built. I pushed #34422 to test it more thoroughly.

anomiex:
Worked, no differences in the artifact.

PR #34412, opened 2023-12-01
Refactor checkout modal

### LINKED ISSUE #32237: Subscriptions Block: extra resources loaded every time the block is loaded on a page
This issue is similar to https://github.com/Automattic/wp-calypso/issues/71048, but I wanted to log it here in the Jetpack repo because of the following changes:

1. The Subscriptions modal is now used for all subscription blocks.
2. This change's impact on performance is significant, since subscription blocks are often loaded on every page of a site.

*************

The Jetpack Subscription block now relies on the Thickbox library:
https://github.com/Automattic/jetpack/blob/a1ad934e39d33cfc741f8c310d01820b87a010cd/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php#L431-L435

This is problematic because the Thickbox library comes with its own set of dependencies, thus potentially significantly impacting the loading speed on a site. Here is a preview of the different items that get added when adding a Subscribe block to your site:

- The block's CSS file.
- Thickbox's js
- Thickbox's css
- Thickbox's loading gif
- jQuery
- jQuery Migrate
- Dashicons' CSS.

<table>
<tr>
 <td><strong>Without the Subscribe Block</strong></td>
 <td><strong>With the Subscribe block</strong></td>
<tr>
 <td>
Screenshot 2023-08-02 at 17 37 26
</td>
 <td>
Screenshot 2023-08-02 at 17 36 54
</td>
</table>

### LINKED ISSUE #71048: Rework the checkout modal
We use `thickbox` library to display the modal iframe:
https://github.com/Automattic/jetpack/blob/b313bd253afdf919adf7a71a866d0ab85b3a7bda/projects/plugins/jetpack/extensions/shared/memberships.js#L19

That has a dependancy on jQuery, which is less than ideal. We want to remove that dependency and switch to another modal system.
Suposedly, navigation modal has a good library:

https://make.wordpress.org/core/2022/01/07/the-new-navigation-block/

**Remember that checkout modal displays for the site visitor, not in Gutenberg editor.** This means view.js in gutenberg components.

### Goal of this task:

1. Figure out how to use the navigation block modal to display an iframe. If that is not possible, research other Gutenberg components and see if there is a modal that fits.
2. Reimplement checkout using that new modal.

Watch out for:
- Mobile layouts
- Cross browser compatibilities

iOS safari MUST WORK FLAWLESSLY.

CC @mtias

chrismccluskey:
Fixes #32237
Fixes https://github.com/Automattic/wp-calypso/issues/71048
Fixes 153-gh-automattic/gold

## Proposed changes:

* A refactor of the checkout modal for memberships which removes the Thickbox and jQuery dependencies.

### Other information:

## Testing instructions:

* Optionally, you can apply D131322-code to your sandbox. This isn't required for this test, but if you choose to do so, make sure you download the patch in the next step on top of this patch (and not to trunk).
* Download this patch to your sandbox.
* Use a test site with payment plans and multiple membership block types.
* Navigate to a site with a payment button / paywalled content.
* Click the button / enter your email and click subscribe to bring up the checkout modal.
* Observe that the checkout modal functions as expected.
* Confirm there is no jquery or thickbox dependency loaded in the browser for these pages.
* Confirm the checkout works for these pages.


chrismccluskey:
Please note that this is a rough proof-of-concept at this point.

projects/plugins/jetpack/modules/memberships/class-jetpack-memberships.php
```diff
@@ -340,7 +340,7 @@ public function should_render_button_preview( $block ) {
 	 * @return string|void - HTML for the button, void removes the button.
 	 */
 	public function render_button( $attributes, $content = null, $block = null ) {
-		Jetpack_Gutenberg::load_assets_as_required( self::$button_block_name, array( 'thickbox', 'wp-polyfill' ) );
```
millerf, line 343: Are you sure removing the polyfill is necessary? (I am actually not sure). I would keep it to be on the safe side...

@jeherve Do you know anything about this one?

millerf (dismissed):
Seems to work fine. I could not find any jQuery commands leftover.

note: the model discards when cliking on the background, but it might need a rebase on trunk?

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -526,11 +526,8 @@ function render_block( $attributes ) {
 	}
 
 	if ( class_exists( '\Jetpack_Memberships' ) ) {
-		// We only want the sites that have newsletter feature enabled to be graced by this JavaScript and thickbox.
-		Jetpack_Gutenberg::load_assets_as_required( __DIR__, array( 'thickbox' ) );
-		if ( ! wp_style_is( 'enqueued' ) ) {
-			wp_enqueue_style( 'thickbox' );
-		}
+		// We only want the sites that have newsletter feature enabled to be graced by this JavaScript.
+		jetpack_gutenberg::load_assets_as_required( __DIR__ );
```
jeherve:
```suggestion
		Jetpack_Gutenberg::load_assets_as_required( __DIR__ );
```

projects/plugins/jetpack/modules/memberships/class-jetpack-memberships.php
jeherve: It shouldn't be necessary; the polyfill is included anyway, even without this explicit addition here:
https://github.com/Automattic/jetpack-production/blob/ec28b519b8f5fe474696391d357451fa0dc0cdfb/_inc/blocks/recurring-payments/view.asset.php

It gets added because the `view.js` file requires a WordPress package:
https://github.com/Automattic/jetpack/blob/10bbdccba045f173f484b23ddb749b9919fe1810/projects/plugins/jetpack/extensions/blocks/recurring-payments/view.js#L1

millerf:
Thanks @jeherve !

chrisfromthelc (approved):
Everything appears to work correctly for me on this.

PR #34414, opened 2023-12-01
Update sharing block controls

### LINKED ISSUE #34115: Sharing block: hook new block buttons to Sharing_Service
Epic: #19694

andrii-lysenko:
Fixes https://github.com/Automattic/jetpack/issues/34115

## Proposed changes:

This is follow up to https://github.com/Automattic/jetpack/pull/34181 so that buttons will have working logic. 
For now, we only using the most important part of redirects.
- Copied `sharing-sources-block` to the block folder leaving only logic inside (no html and styles).
- Js was moved to `view.js`
 

### Other information:

## Testing instructions:

- Add Sharing Buttons block to a post and check it's behaviour.
- View the post and check Sharing Buttons functionality.


projects/plugins/jetpack/extensions/blocks/sharing-buttons/sharing-buttons.php
```diff
@@ -39,3 +39,52 @@ function load_assets( $attr, $content ) {
...
 }
+
+/**
+ * Add sharing JavaScript to the footer of a page.
+ *
+ * @return void
+ */
+function sharing_add_footer() {
+	?>
+	<script>
```
andrii-lysenko: Is it OK to add JS like this to the page? We still need to do some part inline as we might pass some parameters from `attributes`.

projects/plugins/jetpack/extensions/blocks/sharing-button/sharing-button.php
```diff
@@ -7,12 +7,12 @@
...
  */
 
-namespace Automattic\Jetpack\Extensions\Sharing_Button;
+namespace Automattic\Jetpack\Extensions\Sharing_Button_Block;
 
 use Automattic\Jetpack\Blocks;
 use Jetpack_Gutenberg;
 
-require_once JETPACK__PLUGIN_DIR . 'modules/sharedaddy/sharing-sources.php';
+require_once JETPACK__PLUGIN_DIR . 'modules/sharedaddy/sharing-sources-block.php';
```
andrii-lysenko: I copied original `sharing-sources.php` and left only the logic we possible need (URLs and Redirects). Not sure where we should keep it though. As a new Module, as a Package (with only 1 file) or **I can simply put this php file in the block folder** and use it as sort of utils?

jeherve: >  I can simply put this php file in the block folder and use it as sort of utils?

That would be my vote, I like that idea. 👍 I think it would also make it easier for new folks to understand and onboard on the long term.

projects/plugins/jetpack/extensions/blocks/sharing-buttons/sharing-buttons.php
jeherve: Would it be an option to enqueue most of it in the view.js file for the block, and then when parameters need to be passed from PHP to JS, rely on `wp_add_inline_script` or `wp_localize_script` instead?

If the data comes from attributes, would you be able to add it to the block markup so you can access it from the frontend that way?

andrii-lysenko: Hmm, possibly yes. I can try add this into the `save.js` actually.

projects/plugins/jetpack/tools/check-block-assets.php
```diff
@@ -75,7 +75,11 @@
 		'wp-element',
 		'wp-primitives',
 	),
-	'story'          => array(
+	'sharing-buttons' => array(
+		'react',
+		'react-dom',
+	),
```
jeherve: We shouldn't need this. Since we'll be adding such buttons to all posts on a site, we don't want to start loading React on every single post.

I think we would be better off loading individual svg files or an svg sprite, it would still be better for performance.

projects/plugins/jetpack/extensions/blocks/sharing-button/sharing-sources-block.php
```diff
@@ -0,0 +1,1397 @@
+<?php // phpcs:ignore WordPress.Files.FileName.InvalidClassFileName
```
jeherve: Since we're starting fresh, maybe we could rename that file to match the class?

projects/plugins/jetpack/extensions/blocks/sharing-button/sharing-sources-block.php
```diff
@@ -0,0 +1,1397 @@
...
+		 * Filter the way sharing links open.
+		 *
+		 * By default, sharing links open in a new window.
+		 *
+		 * @module sharedaddy
+		 *
+		 * @since 3.4.0
+		 *
+		 * @param bool true Should Sharing links open in a new window. Default to true.
+		 */
```
jeherve: Since all the filters are already defined in the module, you should be able to remove all the docblocks, and replace them with `/** This filter is documented in projects/plugins/jetpack/modules/sharedaddy/sharing-sources.php */`

projects/plugins/jetpack/extensions/blocks/sharing-button/sharing-sources-block.php
```diff
@@ -0,0 +1,1397 @@
...
+	}
+
+	/**
+	 * Get the protocol to use for a sharing service, based on the site settings.
+	 *
+	 * @return string
+	 */
+	public function http() {
+		return is_ssl() ? 'https' : 'http';
+	}
```
jeherve: Maybe this is an opportunity to get rid of some of the legacy things. Here for example maybe we can always use HTTPS at this point?

projects/plugins/jetpack/extensions/blocks/sharing-button/sharing-sources-block.php
```diff
@@ -0,0 +1,1397 @@
...
+	 * Process sharing request. Add actions that need to happen when sharing here.
+	 *
+	 * @param WP_Post $post Post object.
+	 * @param array   $post_data Array of information about the post we're sharing.
+	 *
+	 * @return void
+	 */
+	public function process_request( $post, array $post_data ) {
+		if ( empty( $_POST['jetpack-mastodon-instance'] ) ) {
+			require_once WP_SHARING_PLUGIN_DIR . 'services/class-jetpack-mastodon-modal.php';
```
jeherve: This is something we'll have to be intentional about, to avoid errors if one day that file is removed or renamed.

What do you think about adding a comment in the file, mentioning that the `Jetpack_Mastodon_Modal` is used in the module as well as the block?

projects/plugins/jetpack/extensions/blocks/sharing-button/sharing-sources-block.php
andrii-lysenko: There's a problem that that file contains lots of classes inside. Or we should need at least one?

projects/plugins/jetpack/extensions/blocks/sharing-button/sharing-sources-block.php
andrii-lysenko: Hmm, I thought we would eventually remove the module? Or we will still keep it?

projects/plugins/jetpack/extensions/blocks/sharing-button/sharing-sources-block.php
jeherve: Maybe this is an opportunity to organize this with one class per file?

projects/plugins/jetpack/extensions/blocks/sharing-button/sharing-sources-block.php
jeherve: We will have to keep it around for a little while, we cannot remove it as soon as the block will be available.

andrii-lysenko: I'm still wondering on how many filters we would actually need at the end, so I tried not to delete anything might be useful. I could remove notes on module though, as this block should work on it's own.

projects/plugins/jetpack/extensions/blocks/sharing-button/sharing-sources-block.php
andrii-lysenko: What if I also move this file into `block` directory? To keep things fully separated? What do you think?

projects/plugins/jetpack/extensions/blocks/sharing-button/sharing-sources-block.php
andrii-lysenko: Oh, I might was getting that wrong. I will try to address this.

andrii-lysenko: > Maybe this is an opportunity to organize this with one class per file?

Well, they are quite similar. I was actually wondering if, when finalized we could find a way to combine these into one class. We would need to see first how much logic there would differ. 
As of now, there's only 1 function actually, while before there were a lot of frontend stuff.

projects/plugins/jetpack/extensions/blocks/sharing-button/sharing-sources-block.php
jeherve: Yeah, that'd be okay too for sure. 👍

projects/plugins/jetpack/extensions/blocks/sharing-buttons/view.js
```diff
@@ -1 +1,52 @@
 import './style.scss';
+
+const services = [
```
monsieur-z: I'm wondering if getting that info from the server would make future edits easier.

I'm thinking printing the list in a template like:
```
<script>
window.WHATEVER_NAMESPACE.services = '<?php json_encode( $services) ?>';
</script>
```

Then retrieving the list in JS.

projects/plugins/jetpack/extensions/blocks/sharing-buttons/block.json
monsieur-z, line 1: Nitpicking here: the `name` should probably be `Sharing Buttons` for case consistency.

monsieur-z (changes requested):
It works well!

The only issue I've noticed is that the block is missing some padding on the front-end when adding a background:
Screenshot 2023-12-06 at 5 02 50 PM
Screenshot 2023-12-06 at 5 02 54 PM

I've also a question. When enabling Sharing buttons in the Jetpack settings, this section appears in the nav:
Screenshot 2023-12-06 at 5 00 18 PM

When we insert a new block, should we add by default the buttons that are in the _Enabled services_ section?

projects/plugins/jetpack/extensions/blocks/sharing-buttons/view.js
andrii-lysenko: Yes, I also find this not ideal. However, maybe keeping it in global `windows` making me a bit nervous as well. Maybe we could put it as an `attribute` in block.json? What do you think?

projects/plugins/jetpack/extensions/blocks/sharing-buttons/block.json
andrii-lysenko: Nice catch, thanks!

andrii-lysenko:
@monsieur-z Thanks for reviewing and spotting these. I've addressed the `padding` issue.
As per Sharing Buttons settings, we currently keep these out of scope. At first we would deliver only the block, which is fully independent from current functionality. it should state that this is Experimental, and users would need to handle every block separately for now. As @jeherve noticed, there's a possibility to create templates for posts, so the assumption is that they would set it up there.

andrii-lysenko:
In the future, we would either remove this functionality or make it work with the block I assume.

projects/plugins/jetpack/extensions/blocks/sharing-button/class-sharing-source-block.php
```diff
@@ -0,0 +1,1359 @@
+<?php
+/**
+ * Define all sharing sources.
+ *
+ * @since 11.x
```
jeherve:
```suggestion
 * @since $$next-version$$
```

projects/plugins/jetpack/extensions/blocks/sharing-button/components/social-icons.jsx
```diff
@@ -0,0 +1,410 @@
+/* eslint-disable react/forbid-elements */
+export function getSocialIcon( service ) {
+	let svg_val = '';
+
+	switch ( service ) {
+		case 'facebook':
+			svg_val = (
+				<svg
```
jeherve: I'm afraid we cannot afford to include the SVGs in here. It's going to be messy to keep that list updated. This is why the `Automattic/social-logos` repo was created, so we only have one single source of truth for social media icons, to use across all our products.

Maybe you could use the svgs that come with the exports of Automattic/social-logos instead, or the sprite? We take the sprite approach here for example:
https://github.com/Automattic/jetpack/tree/793fbdc542696c836af5123dcd614b93d7e39a74/projects/plugins/jetpack/modules/theme-tools/social-menu

projects/plugins/jetpack/extensions/blocks/sharing-buttons/view.js
jeherve: Passing the info from PHP to js seems like a good idea to me. This is something we do for a few other blocks as well. The AI Chat block is probably a good example:

https://github.com/Automattic/jetpack/blob/4afbe20c341b366744a77a103e9895ac37b09f83/projects/plugins/jetpack/extensions/blocks/ai-chat/ai-chat.php#L91-L95

projects/plugins/jetpack/extensions/blocks/sharing-button/components/social-icons.jsx
andrii-lysenko: Yeah, I had troubles in images showing up in Simple sites and thought it's because using that lib. But it seem actually not, so I will revert back to social-logos

andrii-lysenko:
@jeherve @monsieur-z Thank you for reviews. 
The padding issue and passing `services` parameters from the backend were addressed.

On a side note, while additional testing I noticed that there's a problem with `save` on Simple/Atomic sites, most likely due to `unfiltered_html `.  We would want to address that in a following PRs in suggested order:
- Duplicate locally SVGs from [`social-logos`](https://github.com/Automattic/social-logos) repo.
- Remove current `save` file and have it moved to a blocks php file instead.
- On a later stage, find a way to re-use the actual files from `social-logos`.

jeherve (approved):
Let's get this merged so you can iterate in follow-up PRs!

monsieur-z (approved):

PR #34419, opened 2023-12-01
Email preview: fix sending preview if content in the editor is different from latest in the database

simison:
## Proposed changes:

* Fire auto-save before sending the email preview. Ensures the preview email has latest content from the post, instead of what might've been saved by editor previously.

### Other information:

## Testing instructions:

* Create a post; good to test without saving it first as a draft, as well as saving it as a draft
* If you save a draft, go ahead and modify the text some more. Using numbers in content makes testing easy.
* Hit publish, hit "Send test email"
* Note auto-save kick on the background
* See that you get the email with latest content
* Without the change, latest unsaved changes aren't in the email

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/email-preview.js
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/email-preview.js
@@ -11,7 +11,8 @@ import {
 	TextControl,
 	Icon,
 } from '@wordpress/components';
-import { useSelect } from '@wordpress/data';
+import { useSelect, useDispatch } from '@wordpress/data';
+import { store as editorStore } from '@wordpress/editor';
 import { useState } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import './email-preview.scss';
@@ -23,15 +24,23 @@ export default function EmailPreview( { isModalOpen, closeModal } ) {
 	const [ emailSending, setEmailSending ] = useState( false );
 	const [ errorMessage, setErrorMessage ] = useState( false );
 	const postId = useSelect( select => select( 'core/editor' ).getCurrentPostId() );
+	const { __unstableSaveForPreview } = useDispatch( editorStore );
 	const [ isSmall ] = useBreakpointMatch( 'sm' );
 	const { tracks } = useAnalytics();
 
-	const sendEmailPreview = () => {
+	const sendEmailPreview = async () => {
 		tracks.recordEvent( 'jetpack_send_email_preview', {
 			post_id: postId,
 		} );
 
 		setEmailSending( true );
+
+		// Save post revision so that we send what they see in the editor, and not what previous draft/revision might've saved
+		// Introduced at GB 16.3 at https://github.com/WordPress/gutenberg/pull/44971
+		if ( typeof __unstableSaveForPreview === 'function' ) {
+			await __unstableSaveForPreview();
+		}
+
 		apiFetch( {
 			path: '/wpcom/v2/send-email-preview/',
 			method: 'POST',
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/subscriptions/email-preview.js
```diff
@@ -23,15 +24,23 @@ export default function EmailPreview( { isModalOpen, closeModal } ) {
...
 		tracks.recordEvent( 'jetpack_send_email_preview', {
 			post_id: postId,
 		} );
 
 		setEmailSending( true );
+
+		// Save post revision so that we send what they see in the editor, and not what previous draft/revision might've saved
+		// Introduced at GB 16.3 at https://github.com/WordPress/gutenberg/pull/44971
+		if ( typeof __unstableSaveForPreview === 'function' ) {
+			await __unstableSaveForPreview();
```
lezama, line 42: nice!

projects/plugins/jetpack/extensions/blocks/subscriptions/email-preview.js
```diff
@@ -23,15 +24,23 @@ export default function EmailPreview( { isModalOpen, closeModal } ) {
...
-	const sendEmailPreview = () => {
+	const sendEmailPreview = async () => {
 		tracks.recordEvent( 'jetpack_send_email_preview', {
 			post_id: postId,
 		} );
 
 		setEmailSending( true );
+
+		// Save post revision so that we send what they see in the editor, and not what previous draft/revision might've saved
+		// Introduced at GB 16.3 at https://github.com/WordPress/gutenberg/pull/44971
```
simison, line 39: @Automattic/jetpack-crew in case you had advice on how to handle Gutenberg functions and backwards compatibility with older WPs in Jetpack; this isn't really issue for .com since we run the latest.

Method introduced in https://github.com/WordPress/gutenberg/pull/44971
It's here, and it's [contents are stable](https://github.com/WordPress/gutenberg/blob/37bca49efb6c8322d62adf642407c50f4270d9db/packages/editor/src/store/actions.js#L304-L322) so in theory I could just copy all that code here instead.

lezama (dismissed):

lezama:
like a charm

Screenshot 2023-12-01 at 11 11 39

projects/plugins/jetpack/extensions/blocks/subscriptions/email-preview.js
anomiex: Jetpack needs to support back to the version of the Gutenberg packages that is bundled in the previous WordPress release, which is currently 6.3.

I guess the case here is that 6.3 (and maybe 6.4 too?) doesn't have any version of the function at all, and you're wanting to copy-paste the implementation into Jetpack so the change you're making here still happens in 6.3? Be sure to include a comment indicating that the back-compat code can be removed when we drop support for the old version. Something along the lines of ``// @todo Remove the `if` check and `else` branch once WP 6.4 is the minimum supported version`` is fine, you don't have to get too fancy. Just be sure the comment clearly mentions an appropriate WordPress version so we can find it easily via a grep when we're making the PRs like #34210.

Or if it's ok that WP 6.3 just doesn't do whatever the new function does, that's fine with us too. Again though a comment mentioning the WP version when the `if` check can be removed would be useful.

lezama: Done!

projects/plugins/jetpack/extensions/blocks/subscriptions/email-preview.js
```diff
@@ -23,15 +24,24 @@ export default function EmailPreview( { isModalOpen, closeModal } ) {
...
+		// @todo Remove the `if` check and `else` branch once WP 6.4 is the minimum supported version
```
anomiex: Err, there's no else branch here (yet?)

simison: Updated 👍 https://github.com/Automattic/jetpack/pull/34419/commits/b0668e57e0323b5293dc74fe94b91ed6e744afcd

anomiex (approved):
Comment looks good to me now. Haven't reviewed the code itself, but someone else did earlier.

PR #34421, opened 2023-12-01
Add page view events to WooCommerce Checkout flow.

senadir:
NOTE: This should be merged until https://github.com/woocommerce/woocommerce-blocks/pull/12010 is merged and assigned a version, since we will use that version to gate keep the work here.

This is part of our work here https://github.com/woocommerce/woocommerce-blocks/pull/12010 to move page view events from WooCommerce Blocks to Jetpack, this ensures all shopper events live together and also now support classic cart/checkout views.

### Other information:

## Testing instructions:

* On a live, Jetpack connected website, install WooCommerce. It would be good to install WooCommerce Blocks from this link [to avoid duplicate events.](https://github.com/woocommerce/woocommerce-blocks/pull/12010#issuecomment-1836325302)
* Using a block theme and running Checkout Block and Cart Block.
* Open Tracks Vigilante
* View a product page, add it to cart, and visit Cart page and then Checkout.
* Make sure the product_view and cart_view events fire.
* Make sure all data is present, checkout_view should be similar to product_checkout.
* Place the order, make sure the order_view event fire, it should be similar to product_purchase (with main different being that products is an array).
* Do the flow again, now with Cart shortcode and Checkout shortcode.
* The same events should fire with the same data.
* Do the flow again but in a classic theme.

### DIFF
--- a/projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
+++ b/projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
@@ -1,0 +1,109 @@
+<?php
+/**
+ * Jetpack_WooCommerce_Analytics_Checkout_Flow
+ *
+ * @package automattic/jetpack
+ * @author Automattic
+ */
+
+/**
+ * Bail if accessed directly
+ */
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Class Jetpack_WooCommerce_Analytics_Checkout_Flow
+ * Class that handles all page view events for the checkout flow (from product view to order confirmation view)
+ */
+class Jetpack_WooCommerce_Analytics_Checkout_Flow {
+
+	use Jetpack_WooCommerce_Analytics_Trait;
+
+	/**
+	 * Jetpack_WooCommerce_Analytics_Checkout_Flow constructor.
+	 */
+	public function __construct() {
+		$this->find_cart_checkout_content_sources();
+
+		// single product page view.
+		add_action( 'woocommerce_after_single_product', array( $this, 'capture_product_view' ) );
+
+		// order confirmed page view
+		add_action( 'woocommerce_thankyou', array( $this, 'capture_order_confirmation_view' ), 10, 1 );
+
+		// cart page view
+		add_action( 'wp_footer', array( $this, 'capture_cart_view' ) );
+
+		// checkout page view
+		add_action( 'wp_footer', array( $this, 'capture_checkout_view' ) );
+	}
+
+		/**
+		 * Track a product page view
+		 */
+	public function capture_product_view() {
+		global $product;
+		if ( ! $product instanceof WC_Product ) {
+			return;
+		}
+
+		$this->record_event(
+			'woocommerceanalytics_product_view',
+			array(),
+			$product->get_id()
+		);
+	}
+
+	/**
+	 * Track the order confirmation page view
+	 */
+	public function capture_order_confirmation_view() {
+		$order_id = absint( get_query_var( 'order-received' ) );
+		if ( ! $order_id ) {
+			return;
+		}
+
+		if ( ! is_order_received_page() ) {
+			return;
+		}
+
+		$this->record_event(
+			'woocommerceanalytics_order_confirmation_view',
+			array()
+		);
+	}
+
+	/**
+	 * Track the cart page view
+	 */
+	public function capture_cart_view() {
+		if ( ! is_cart() ) {
+			return;
+		}
+		$this->record_event(
+			'woocommerceanalytics_cart_view',
+			array()
+		);
+	}
+
+	/**
+	 * Track the checkout page view
+	 */
+	public function capture_checkout_view() {
+		if ( ! is_checkout() ) {
+			return;
+		}
+
+		// Order received page is also a checkout page, so we need to bail out if we are on that page.
+		if ( is_order_received_page() ) {
+			return;
+		}
+
+		$this->record_event(
+			'woocommerceanalytics_checkout_view',
+			array()
+		);
+	}
+}
--- a/projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
+++ b/projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
@@ -201,7 +201,7 @@ class Jetpack_WooCommerce_Analytics_Checkout_Flow {
 			'create_account'         => $create_account,
 			'guest_checkout'         => $guest_checkout,
 			'express_checkout'       => 'null', // TODO: not solved yet.
-			'number_products'        => $cart->get_cart_contents_count(),
+			'products_count'         => $cart->get_cart_contents_count(),
 			'order_value'            => $cart->get_cart_contents_total(),
 			'shipping_options_count' => 'null', // TODO: not solved yet.
 			'coupon_used'            => $coupon_used,
--- a/projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
+++ b/projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
@@ -26,6 +26,8 @@ class Jetpack_WooCommerce_Analytics_Checkout_Flow {
 	 */
 	public function __construct() {
 		$this->find_cart_checkout_content_sources();
+		$this->additional_blocks_on_cart_page     = $this->get_additional_blocks_on_page( 'cart' );
+		$this->additional_blocks_on_checkout_page = $this->get_additional_blocks_on_page( 'checkout' );
 
 		// single product page view.
 		add_action( 'woocommerce_after_single_product', array( $this, 'capture_product_view' ) );
@@ -69,9 +71,35 @@ class Jetpack_WooCommerce_Analytics_Checkout_Flow {
 			return;
 		}
 
+		$order = wc_get_order( $order_id );
+
+		$coupons     = $order->get_coupons();
+		$coupon_used = 0;
+		if ( is_countable( $coupons ) ) {
+			$coupon_used = count( $coupons ) ? 1 : 0;
+		}
+
+		if ( is_object( WC()->session ) ) {
+			$create_account = true === WC()->session->get( 'wc_checkout_createaccount_used' ) ? 'Yes' : 'No';
+		} else {
+			$create_account = 'No';
+		}
+
 		$this->record_event(
 			'woocommerceanalytics_order_confirmation_view',
-			array()
+			array(
+				'coupon_used'      => $coupon_used,
+				'create_account'   => $create_account,
+				'express_checkout' => 'null', // TODO: not solved yet.
+				'guest_checkout'   => $order->get_customer_id() ? 'No' : 'Yes',
+				'oi'               => $order->get_id(),
+				'order_value'      => $order->get_total(),
+				'payment_option'   => $order->get_payment_method(),
+				'products_count'   => $order->get_item_count(),
+				'products'         => $this->format_items_to_json( $order->get_items() ),
+				'order_note'       => $order->get_customer_note(),
+				'shipping_option'  => $order->get_shipping_method(),
+			)
 		);
 	}
 
@@ -82,9 +110,13 @@ class Jetpack_WooCommerce_Analytics_Checkout_Flow {
 		if ( ! is_cart() ) {
 			return;
 		}
+
 		$this->record_event(
 			'woocommerceanalytics_cart_view',
-			array()
+			array_merge(
+				$this->get_cart_checkout_shared_data(),
+				array()
+			)
 		);
 	}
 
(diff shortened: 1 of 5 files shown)


projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
```diff
@@ -0,0 +1,213 @@
...
+
+		$enabled_payment_options = array_keys( $enabled_payment_options );
+
+		$shared_data = array(
+			'products'               => $this->format_items_to_json( $cart->get_cart() ),
+			'create_account'         => $create_account,
+			'guest_checkout'         => $guest_checkout,
+			'express_checkout'       => 'null', // TODO: not solved yet.
+			'number_products'        => $cart->get_cart_contents_count(),
+			'order_value'            => $cart->get_cart_contents_total(),
```
opr, line 205: This could be copied from https://github.com/Automattic/jetpack/blob/479c0d0f490d76f8a7ab18280285f2f0880a785b/projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-universal.php#L230-L233

The current code in this event looks like it's only the subtotal, and not in the same format as the "core" events.

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
```diff
@@ -0,0 +1,213 @@
...
+ */
+class Jetpack_WooCommerce_Analytics_Checkout_Flow {
+
+	use Jetpack_WooCommerce_Analytics_Trait;
+
+	/**
+	 * Jetpack_WooCommerce_Analytics_Checkout_Flow constructor.
+	 */
+	public function __construct() {
+		$this->find_cart_checkout_content_sources();
```
opr, line 28: I forgot why you said this is necessary, the one that runs in the construct of `Jetpack_WooCommerce_Analytics_Universal` isn't always running? Same for two lines below?

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
```diff
@@ -0,0 +1,213 @@
...
+			$products[] = $data;
+		}
+
+		return wp_json_encode( $products );
+	}
+
+	/**
+	 * Get Cart/Checkout page view shared data
+	 */
+	protected function get_cart_checkout_shared_data() {
```
opr, line 174: Why can't this be in the trait? There are a few events in `Jetpack_WooCommerce_Analytics_Universal` that would benefit from this

opr (approved):
I tested and the page view events themselves seem to be tracking well across shortcode/classic shortcode/blocks in both classic and block themes 🥳

There's just a couple of things I spotted that I mentioned inline, but not sure if it makes total sense to fix them here given the wider scope of these changes. I think we should ensure the tracking events are working in this PR (which they are), merge it into [`update/classic-theme-tracking`](https://github.com/Automattic/jetpack/tree/update/classic-theme-tracking) then address any code-cleanup (i.e. what I left in the inline comments) in that PR? What do you think?

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
senadir: Yes, I just didn't want to touch the core events data much, but I'm likely to merge everything.

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
```diff
@@ -0,0 +1,213 @@
...
+				'coupon_used'      => $coupon_used,
+				'create_account'   => $create_account,
+				'express_checkout' => 'null', // TODO: not solved yet.
+				'guest_checkout'   => $order->get_customer_id() ? 'No' : 'Yes',
+				'oi'               => $order->get_id(),
+				'order_value'      => $order->get_total(),
+				'payment_option'   => $order->get_payment_method(),
+				'products_count'   => $order->get_item_count(),
+				'products'         => $this->format_items_to_json( $order->get_items() ),
+				'order_note'       => $order->get_customer_note(),
```
nielslange, line 100: Only the `woocommerceanalytics_order_confirmation_view` event has the `order_note` entry. The `woocommerceanalytics_product_purchase` does not have this entry.

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
```diff
@@ -0,0 +1,213 @@
...
+				'shipping_option'  => $order->get_shipping_method(),
```
nielslange, line 101: Only the `woocommerceanalytics_order_confirmation_view` event has the `shipping_option` entry. The `woocommerceanalytics_product_purchase` does not have this entry.

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
```diff
@@ -0,0 +1,213 @@
...
+		);
+
+		$enabled_payment_options = array_keys( $enabled_payment_options );
+
+		$shared_data = array(
+			'products'               => $this->format_items_to_json( $cart->get_cart() ),
+			'create_account'         => $create_account,
+			'guest_checkout'         => $guest_checkout,
+			'express_checkout'       => 'null', // TODO: not solved yet.
+			'number_products'        => $cart->get_cart_contents_count(),
```
nielslange: The `woocommerceanalytics_product_checkout` event refers to `number_products`, while the `woocommerceanalytics_checkout_view`, `woocommerceanalytics_order_confirmation_view` and `woocommerceanalytics_product_purchase` call this entry `products_count`. 

To avoid confusion, I wonder if we should also call this entry `products_count` in the `woocommerceanalytics_product_checkout` event.

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
```diff
@@ -0,0 +1,213 @@
...
+			'order_value'            => $cart->get_cart_contents_total(),
+			'shipping_options_count' => 'null', // TODO: not solved yet.
```
nielslange, line 206: In only see `shipping_options_count` in the `woocommerceanalytics_checkout_view` event, but not in the `woocommerceanalytics_product_checkout` event. 

However, the `woocommerceanalytics_product_checkout` event contains the entry `shipping_option`, which does not exist in the `woocommerceanalytics_checkout_view` event.

nielslange (commented):
Thanks for working on this, @senadir. I mentioned a few minor inconsistencies between:
- `woocommerceanalytics_checkout_view` and `woocommerceanalytics_product_checkout`
- `woocommerceanalytics_order_confirmation_view` and `woocommerceanalytics_product_purchase`

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
opr: `woocommerceanalytics_product_purchase` refers to a _single_ product (it will be fired _n_ times where _n_ is the number of distinct items purchased).

`woocommerceanalytics_order_confirmation_view` is the whole page view and is fired _once_. 

[We agreed to leave this property off the `woocommerceanalytics_product_purchase` event](https://github.com/woocommerce/woocommerce-blocks/issues/9250#issuecomment-1710616283) because it's not possible to reliably determine which items have which shipping option. This is even harder in the case of multiple packages.

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
```diff
@@ -0,0 +1,213 @@
...
+	/**
+	 * Jetpack_WooCommerce_Analytics_Checkout_Flow constructor.
+	 */
+	public function __construct() {
+		$this->find_cart_checkout_content_sources();
+		$this->additional_blocks_on_cart_page     = $this->get_additional_blocks_on_page( 'cart' );
+		$this->additional_blocks_on_checkout_page = $this->get_additional_blocks_on_page( 'checkout' );
+
+		// single product page view.
+		add_action( 'woocommerce_after_single_product', array( $this, 'capture_product_view' ) );
```
nielslange, line 33: I wonder if we should use hook into `wp_footer` and use `is_product()` to determine if a single product is viewed, instead of hooking into `woocommerce_after_single_product` to call `capture_product_view`. While I don't assume that `woocommerce_after_single_product` will be removed, I'm certain that `wp_footer` will never be removed.

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
```diff
@@ -0,0 +1,213 @@
...
+
+		// order confirmed page view
+		add_action( 'woocommerce_thankyou', array( $this, 'capture_order_confirmation_view' ), 10, 1 );
```
nielslange, line 36: Similar to `woocommerce_after_single_product`, I wonder if we should also hook into `wp_footer` here for the same reason. It's very unlikely that the `wp_footer` hook will ever be removed.

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-trait.php
```diff
@@ -285,6 +341,72 @@ public function get_user_id() {
...
+				if ( 'core/shortcode' === $block['blockName'] ) {
+					return false;
+				}
+				if ( 'checkout' === $cart_or_checkout && 'woocommerce/checkout' !== $block['blockName'] ) {
+					return true;
+				}
+				if ( 'cart' === $cart_or_checkout && 'woocommerce/cart' !== $block['blockName'] ) {
+					return true;
+				}
+				return false;
```
nielslange: Nitpic: I believe adding a few line breaks here will increase the readability.

```suggestion
				if ( ! isset( $block['blockName'] ) ) {
					return false;
				}

				if ( 'woocommerce/classic-shortcode' === $block['blockName'] ) {
					return false;
				}
				
				if ( 'core/shortcode' === $block['blockName'] ) {
					return false;
				}
				
				if ( 'checkout' === $cart_or_checkout && 'woocommerce/checkout' !== $block['blockName'] ) {
					return true;
				}
				
				if ( 'cart' === $cart_or_checkout && 'woocommerce/cart' !== $block['blockName'] ) {
					return true;
				}
				
				return false;
```

nielslange (commented):
Adding a few more suggestions. Nothing critical, though.

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
nielslange: Thanks for clarifying this, @opr.

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
opr: I appreciate where you're coming from with this suggestion @nielslange however I think this is an unnecessary change to make right now.

In the event of `woocommerce_after_single_product` being slated for removal, not only will there be a period of deprecation, but also an announcement/discussion about it which would trigger us to search for uses and amend them. If that comes we could make that change, but for now it adds noise to the PR.

nielslange (approved):
Apart from the suggestions, I mentioned before, the PR works as expected. Thanks again for working on this, @senadir. 🙌

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
senadir: It has been requested by Pedro in https://github.com/woocommerce/woocommerce-blocks/issues/9250 but was not requested in https://github.com/woocommerce/woocommerce-blocks/issues/9729

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
senadir: True, it was requested to rename it. Fixed.

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
senadir: All options related to shipping are still open ended and we haven't fully figured them, so they're for follow up.

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
senadir: I kept this the same it was previously and tried not to touch how events are hooked up for existing ones (product_view, product_checkout, product_purshase).

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
senadir: This has proved to constantly work fine so I didn't want to change it yet, it's also page independent.

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
senadir: I don't recall yet, I think the plugins where coming out without data

projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-checkout-flow.php
senadir: Addressed differently, using `get_cart_total` which saves us the formatting step.

PR #34426, opened 2023-12-01
Repo Gardening: do not gather support refs on Pull Requests

jeherve:
## Proposed changes:

This PR includes 2 changes:

1. We should only gather issue references from issues. However, the event trigger we use (`issue_comment`) targets both PRs and issues. Let's bail early when we're on a Pull Request. See [the documentation here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment-on-issues-only-or-pull-requests-only)
2. When gathering p2 comment references (this was introduced in #33979), let's not change the capitalization of references.

### Other information:

## Jetpack product discussion

Internal reference: p1701430031933879-slack-C03N25JPCE4

* No

## Testing instructions:

This can only be tested in a fork. See these issues:

https://github.com/jeherve/jetpack/issues/126#issuecomment-1841123896
https://github.com/jeherve/jetpack/pull/127


projects/github-actions/repo-gardening/src/tasks/gather-support-references/index.js
```diff
@@ -87,7 +87,11 @@ async function getIssueReferences( octokit, owner, repo, number, issueComments )
 			const correctedId = `${ wrongId[ 1 ] }-zen`;
 			correctedSupportIds.add( correctedId );
 		} else {
-			correctedSupportIds.add( supportId.toLowerCase() );
+			// Switch to lowercase when it's not a p2 comment reference.
+			const standardizedsupportId = supportId.includes( '#comment' )
```
anomiex, line 91: Should we look for the actual `-p2` suffix instead?

sergeymitr: @anomiex the script only gathers `-p2#comment-` strings, so this should work as well.
@jeherve clarified that here: p1701796261528209-slack-C05PV073SG3

sergeymitr (approved):
Tests well 👍

projects/github-actions/repo-gardening/src/tasks/gather-support-references/index.js
anomiex: Yeah, it'll work with the current situation. But it's not exactly doing what the comment says, and might lead to unexpected behavior if we ever start including some other refs that include `#comment`.

jeherve: Do you think I should check for `[a-zA-Z0-9-]+-p2#comment-[0-9]*` instead to be extra sure, like we do here:
https://github.com/Automattic/jetpack/blob/4f919b9df66768f3334b0e29885628fcfa94bac9/projects/github-actions/repo-gardening/src/tasks/gather-support-references/index.js#L53

anomiex: That would work too. Or you could turn it around and only lowercase the ones ending in `-zen`, if it seems more likely that future refs would also want to not be lowercased.

jeherve: 👍 Follow-up in #34483.

PR #34431, opened 2023-12-02
Stats: add passing select UTM parameters

trakos:
## Proposed changes:
This PR adds select UTM parameters to the pixel requests send to wp.com, so that we can show more statistics to site owners.

### Other information:

## Jetpack product discussion
panfyZ-37z-p2

It adds additional parameters to tracking, but we already mention tracking "visiting URL", which parameters are part of.

## Testing instructions:

* Make sure you are logged out
* Add `?utm_source=a_source&utm_medium=a_medium` to any page on your test website
* Verify that the request to pixel.wp.com includes the UTM parameters
* Add a plugin for AMP support, visit any post with `?amp=1&utm_source=test` added to URL, verify that the pixel URL includes UTM parameter

### DIFF
--- a/projects/packages/stats/src/class-tracking-pixel.php
+++ b/projects/packages/stats/src/class-tracking-pixel.php
@@ -27,6 +27,18 @@ class Tracking_Pixel {
 	 */
 	const STATS_ARRAY_TO_STRING_FILTER = 'stats_array';
 
+	const TRACKED_UTM_PARAMETERS = array(
+		'utm_id',
+		'utm_source',
+		'utm_medium',
+		'utm_campaign',
+		'utm_term',
+		'utm_content',
+		'utm_source_platform',
+		'utm_creative_format',
+		'utm_marketing_tactic',
+	);
+
 	/**
 	 * Stats Build View Data.
 	 *
@@ -61,7 +73,16 @@ class Tracking_Pixel {
 		} else {
 			$post = '0';
 		}
-		return compact( 'v', 'blog', 'post', 'tz', 'srv' );
+		$view_data = compact( 'v', 'blog', 'post', 'tz', 'srv' );
+		foreach ( self::TRACKED_UTM_PARAMETERS as $utm_parameter ) {
+			// phpcs:ignore WordPress.Security.NonceVerification.Recommended
+			if ( isset( $_GET[ $utm_parameter ] ) && is_scalar( $_GET[ $utm_parameter ] ) ) {
+				// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.MissingUnslash, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized, WordPress.Security.NonceVerification.Recommended
+				$view_data[ $utm_parameter ] = substr( (string) $_GET[ $utm_parameter ], 0, 255 );
+			}
+		}
+
+		return $view_data;
 	}
 
 	/**
--- a/projects/packages/stats/src/class-tracking-pixel.php
+++ b/projects/packages/stats/src/class-tracking-pixel.php
@@ -258,6 +258,7 @@ _stq.push([ "clickTrackerInit", "%2$s", "%3$s" ]);',
 		 * @param array $kvs Array of options about the site and page you're on.
 		 */
 		$kvs = (array) apply_filters( self::STATS_ARRAY_TO_STRING_FILTER, $kvs );
+		$kvs = array_map( 'strval', $kvs );
 
 		// Encode into JSON object, and then encode it into a string that's safe to embed into Javascript.
 		// We will then use JSON.parse method in JS to read the array.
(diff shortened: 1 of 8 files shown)


projects/packages/stats/src/class-tracking-pixel.php
```diff
@@ -61,7 +73,16 @@ public static function build_view_data() {
 		} else {
 			$post = '0';
 		}
-		return compact( 'v', 'blog', 'post', 'tz', 'srv' );
+		$view_data = compact( 'v', 'blog', 'post', 'tz', 'srv' );
+		foreach ( self::TRACKED_UTM_PARAMETERS as $utm_parameter ) {
+			// phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
jeherve: When we ignore specific PHPCS rules, I often recommend adding a comment explaining why. It makes things a bit easier later on, when coming back to a file that was edited a long time ago ; it feels safer knowing that we added the exception knowingly.

PHPCS allows adding comments like so:

```suggestion
			// phpcs:ignore WordPress.Security.NonceVerification.Recommended -- we clear the parameter as we output it below.
```

projects/packages/stats/src/class-tracking-pixel.php
```diff
@@ -61,7 +73,16 @@ public static function build_view_data() {
...
+			if ( isset( $_GET[ $utm_parameter ] ) && is_scalar( $_GET[ $utm_parameter ] ) ) {
+				// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.MissingUnslash, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized, WordPress.Security.NonceVerification.Recommended
+				$view_data[ $utm_parameter ] = substr( (string) $_GET[ $utm_parameter ], 0, 255 );
```
jeherve: What would you think about adding some of the recommended sanitization methods here? UTM parameter values are limited on Google's end as well I believe, so why not do the same on our end? It would also help avoid issues like PHP notices being logged when someone tries to do something like `?utm_source=<script>alert( 'hello' )</script>`.

```
PHP Notice:  Function wp_add_inline_script was called incorrectly. Do not pass &lt;script&gt; tags to wp_add_inline_script(). Please see <a>Debugging in WordPress</a> for more information. (This message was added in version 4.5.0.) in /wp-includes/functions.php on line 6031
```

```suggestion
				// phpcs:ignore WordPress.Security.NonceVerification.Recommended -- We're sanitizing the value.
				$view_data[ $utm_parameter ] = substr( sanitize_textarea_field( wp_unslash( $_GET[ $utm_parameter ] ) ), 0, 255 );
```

kangzj (commented):
Thanks for working on this. It works well for me 👍 

@jeherve's https://github.com/Automattic/jetpack/pull/34431/files#r1424316039 seems important to address tho.

projects/packages/stats/src/class-tracking-pixel.php
trakos: Thank you for the suggestion! I've added a comment explaining why. UTM parameters are coming from external websites by design, so we don't control them and can't add a nonce here.

projects/packages/stats/src/class-tracking-pixel.php
trakos: I've added the sanitization. It does limit the valid UTM parameters we can handle slightly, but it shouldn't come up in a real-world scenarios too often.

By the way, would you know why `Tracking_Pixel::stats_array_to_string` didn't use `json_encode` directly? I've changed it now that you've highlighted the importance of escaping the data properly. I don't think `addslashes` was sufficient there, and it wasn't checking keys, which could have been overlooked when adding more parameters.

trakos:
Thank you for reviews @kangzj and @jeherve! I've applied the suggestions and changed the serialization in `build_stats_details` method.

Would you know why `Tracking_Pixel::stats_array_to_string` didn't use `json_encode` before? I've changed it now that you've highlighted the importance of escaping the data properly. I don't think `addslashes` was sufficient there, and it wasn't checking keys, which could have been overlooked when adding more parameters.

projects/packages/stats/src/class-tracking-pixel.php
gibrown: > would you know why Tracking_Pixel::stats_array_to_string didn't use json_encode directly?

Just a fun guess because I was curious... json_encode() was added in php5.2 in late 2006. The original stats_array() function on wpcom in mu-plugins was written in Feb 2007!

trakos: Oh, thanks for checking! I didn't realize it was moved from an old code, that would definitely explain it. I think it's worth changing it to json_encode then, just in case we change array keys in the future.

trakos:
When working on wpcom counter-part I've noticed that we need to account for certain cache configs on Atomic: panfyZ-37z-p2#comment-7723, so I've updated this PR.

sotirispl (approved):
I tested it and it works as described. Thank you for working on this!

PR #34438, opened 2023-12-04
AI Assistant: add prompt to attributes

CGastrell:
PoC: add prompt to attributes instead of local state

## Proposed changes:
This PR moves the prompt to the attributes, allowing for the attribute to persist between reloads and drafts.

It now also stores the requestingState to keep consistency between reloads and action buttons behavior.

### Other information:

No

## Testing instructions:
Load the editor, create an AI assistant block, look at the code editor to see the attribute `userPrompt` is set to the prompt you typed. Save draft and reload, see that your last prompt is still there.

See that, after reloading, the input action buttons remain consistent.


dhasilva (approved):
LGTM

PR #34442, opened 2023-12-04
Jetpack AI: Improve event tracking on Usage Panel

### LINKED ISSUE #34443: AI Tiered Plans: improve tracking of events on the usage panel

lhkowalski:
Fixes #34443.

## Proposed changes:

* Add `jetpack_ai_usage_panel_upgrade_button_click` to start tracking clicks on both the Upgrade and Contact Us buttons
* On every event, include the `blog_id`, the `userid` and `username`, along with the current tier `slug` and the current usage on the given tier

### Other information:

Yes. With this change, we start tracking when the user clicks the "Upgrade" button (or the "Contact Us" button) on the usage panel of the AI Assistant.

### Triggering the new event

* Follow the instructions from p1HpG7-pZP-p2 to create a test site with tiers enabled
* Go to the block editor, do a couple of requests using the AI Assistant and then click the upgrade button on the Usage Panel; while testing it, alternate between the two instances of the usage panel available:
   * the one on Jetpack sidebar (on top right, click the Jetpack icon and the open the "AI Assistant" section)
   * the one on the block settings sidebar (this is the most visible one, that appears every time you select an AI Assistant block)
* Follow the upgrade path to buy the plan, then return to the block editor and do some extra AI Assistant requests (just so we have some usage for the site)
* On the Usage Panel again, click the "Get 200 requests" upgrade button to follow a new upgrade path; proceed with the upgrade
* Repeat the actions above for at least one more tier; this should be enough to register some events related to the upgrade button
* Now let's confirm the events are being registered; to do it, you need to go to the "Tracks Live View" panel on "Mission Control"
* On the correct panel, search for the event name `jetpack_ai_usage_panel_upgrade_button_click` and confirm you see results for the event, mentioning your `blog_id` and all the extra info we included on the event
   * Also confirm that the `placement` property is set correctly on the events, either:
      * `jetpack_sidebar` for the clicks made on the usage panel present on the Jetpack Sidebar
      * `block_settings_sidebar` for the clicks made on the usage panel present on block settings sidebar

**Note:** it can take a few minutes for the events to appear on the panel, so make sure you wait for some time


projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-analytics/index.ts
```diff
@@ -0,0 +1,25 @@
+import jetpackAnalytics from '@automattic/jetpack-analytics';
+import { useEffect } from 'react';
```
CGastrell: Should we use this from the `@wordpress/element` package?

projects/plugins/jetpack/extensions/plugins/ai-assistant-plugin/components/usage-panel/index.tsx
```diff
@@ -5,13 +5,14 @@ import { getRedirectUrl } from '@automattic/jetpack-components';
 import { isAtomicSite, isSimpleSite } from '@automattic/jetpack-shared-extension-utils';
 import { Button } from '@wordpress/components';
 import { __, sprintf } from '@wordpress/i18n';
-import React from 'react';
+import React, { useCallback } from 'react';
```
CGastrell: This one as well, I think we have all those wrappers in place so we always trust `@wordpress/element`

vianasw:
Is this adding event tracking on the usage panel in the Jetpack section or is it also adding it to the panel in the block section? If we're adding it to both, it looks like we won't be able to differentiate which usage panel was used.

lhkowalski:
> Is this adding event tracking on the usage panel in the Jetpack section or is it also adding it to the panel in the block section?

It's just on the Usage Panel, on the sidebar. Since I had to introcue a bit more of code (since the `use-analytics` from My Jetpack was tailored to use just on My Jetpack), I decided to keep it more simple and change one place at the time.

We can (should) have a different name for the event on the block, something in the line of `jetpack_ai_assistant_block_upgrade_button_click` or `jetpack_ai_assistant_upgrade_button_click`.

projects/plugins/jetpack/extensions/blocks/ai-assistant/hooks/use-analytics/index.ts
lhkowalski: Thank you for noticing it! Going to update the code now.

CGastrell (dismissed):
LGTM! Events are on track. :shipit:

CGastrell (approved):
Works as described! :shipit:

PR #34446, opened 2023-12-04
CRM: OAuth Connection"s" - update typo

### LINKED ISSUE #34131: CRM: OAuth Connection"s" - update typo
### Impacted plugin

CRM

### What

Drop the "s" on the `OAuth Connections` page

### How

The OAuth Connections page is labelled with plural (Connection"s"), which is a confusing UX - it feels like they should be able to add more than one connection, but it doesn't allow to.

jinnypark:
Fixes #34131

## Proposed changes:

*Fixes plural "OAuth Connections" to "OAuth Connection" on the JPCRM settings page.

### Other information:

I ignored linter erros for translator comments and changing global variable (since the typo involves a global variable).
I'm not 100% sure how exactly changelog should be added for CRM. I updated the existing CRM changelog, but let me know if I should do otherwise.

No

## Testing instructions:

* Go to Jetpack CRM > CRM Settings > OAuth Connection
* Check around settings to ensure all mentions of "Connections" are now "Connection".

### DIFF
--- a/projects/plugins/crm/CHANGELOG.md
+++ b/projects/plugins/crm/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+### Changed
+- OAuth Connection: OAuth Connection"s" - updated typo. [#34131]
 
 ## [6.3.0] - 2023-11-15
 ### Added
--- a/projects/plugins/crm/CHANGELOG.md
+++ b/projects/plugins/crm/CHANGELOG.md
@@ -4,8 +4,6 @@
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
-### Changed
-- OAuth Connection: OAuth Connection"s" - updated typo. [#34131]
 
 ## [6.3.0] - 2023-11-15
 ### Added
--- a/projects/plugins/crm/admin/settings/oauth-connections.page.php
+++ b/projects/plugins/crm/admin/settings/oauth-connections.page.php
@@ -137,7 +137,7 @@ if ( isset( $_GET['edit-provider'] ) && $zbs->oauth->legitimate_provider( $_GET[
 	} else {
 
 		// OAuth not enabled, probably lacking PHP version >= 7.3
-		echo wp_kses( sprintf( __( 'Currently your system does not support OAuth API connections. <a href="%s" target="_blank">Read More</a>', 'zero-bs-crm' ), $zbs->urls['kb-oauth-requirements'] ), $zbs->acceptable_restricted_html );
+		echo wp_kses( sprintf( __( 'Currently your system does not support OAuth API connection. <a href="%s" target="_blank">Read More</a>', 'zero-bs-crm' ), $zbs->urls['kb-oauth-requirements'] ), $zbs->acceptable_restricted_html ); // phpcs:ignore WordPress.WP.I18n.MissingTranslatorsComment
 
 	}
 }
--- a/projects/plugins/crm/admin/settings/oauth-connections.page.php
+++ b/projects/plugins/crm/admin/settings/oauth-connections.page.php
@@ -1,7 +1,7 @@
 <?php
 /*
 !
- * Admin Page: Settings: OAuth Connections
+ * Admin Page: Settings: OAuth Connection
  */
 
 // stop direct access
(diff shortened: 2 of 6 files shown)


projects/plugins/crm/admin/settings/oauth-connections.page.php
```diff
@@ -137,7 +137,7 @@
 	} else {
 
 		// OAuth not enabled, probably lacking PHP version >= 7.3
-		echo wp_kses( sprintf( __( 'Currently your system does not support OAuth API connections. <a href="%s" target="_blank">Read More</a>', 'zero-bs-crm' ), $zbs->urls['kb-oauth-requirements'] ), $zbs->acceptable_restricted_html );
+		echo wp_kses( sprintf( __( 'Currently your system does not support OAuth API connection. <a href="%s" target="_blank">Read More</a>', 'zero-bs-crm' ), $zbs->urls['kb-oauth-requirements'] ), $zbs->acceptable_restricted_html ); // phpcs:ignore WordPress.WP.I18n.MissingTranslatorsComment
```
tbradsha:
```suggestion
		echo wp_kses( sprintf( __( 'Currently your system does not support an OAuth API connection. <a href="%s" target="_blank">Read More</a>', 'zero-bs-crm' ), $zbs->urls['kb-oauth-requirements'] ), $zbs->acceptable_restricted_html ); // phpcs:ignore WordPress.WP.I18n.MissingTranslatorsComment
```

projects/plugins/crm/admin/settings/partials/menu.block.php
```diff
@@ -91,7 +91,7 @@
 	$tabs['mail']          = __( 'Mail', 'zero-bs-crm' );
 	$tabs['maildelivery']  = __( 'Mail Delivery', 'zero-bs-crm' );
 	$tabs['mailtemplates'] = __( 'Mail Templates', 'zero-bs-crm' );
-	$tabs['oauth']         = __( 'OAuth Connections', 'zero-bs-crm' );
+	$tabs['oauth']         = __( 'OAuth Connection', 'zero-bs-crm' ); // phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
```
tbradsha, line 94: Heh, we definitely probably shouldn't have been doing this. I didn't realise `$tabs` was a WP global, but just confirmed.

tbradsha (changes requested):
One tiny typo introduced, but then it'll be good to go.

projects/plugins/crm/admin/settings/oauth-connections.page.php
tbradsha: Although...let's split this out. This is a case we can fully avoid `wp_kses()` and make things better all around. Example (untested code):

```
esc_html_e( 'Currently your system does not support OAuth API connection.', 'zero-bs-crm' );
?>
<a href="<?php echo esc_url( $zbs->urls['kb-oauth-requirements'] ); ?>" target="_blank">
<?php esc_html_e( 'Read More', 'zero-bs-crm' ); ?>
</a>
<?php
```

projects/plugins/crm/admin/settings/oauth-connections.page.php
jinnypark, line 1: I couldn't quite figure out triggering this message to test. I tried testing with PHP version less than 7.2, and got a warning: 
image

jinnypark: I also realized that docs link should be excluded for whitelabels so tried my hand at it, but let me know if I should revert.

projects/plugins/crm/admin/settings/partials/menu.block.php
cleacos: TIL here too

projects/plugins/crm/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+### Changed
+- OAuth Connection: OAuth Connection"s" - updated typo. [#34131]
```
cleacos: > I'm not 100% sure how exactly changelog should be added for CRM. I updated the existing CRM changelog, but let me know if I should do otherwise.

This is not the Jetpack way to introduce the changelog. We don't add it directly to the `CHANGELOG.md` file. We add changelogs using individual files, see this PR example below:

https://github.com/Automattic/jetpack/pull/34605/files#diff-cef089a0a979d8206bff13839ad6665be0ae89a3e6f42df7313e6ddde775ee03.

Normally when you push the branch it performs some checks to see if a changelog file was added, if not, it prompts you to add one on the fly. Also, you can create that file manually before pushing it or better using the `jetpack changelogger add plugins/crm` command. Then, when we release JPCRM, one of the steps merges all the changelog files into the `CHANGELOG.md` file and finally removes them.

cleacos (dismissed):
I left a comment about how to add the changelog correctly. 

The code looks good and the update appears as expected. Well done Jinny!

projects/plugins/crm/CHANGELOG.md
tbradsha: Indeed, remove from here and add an entry via the `changelogger add` tool.

jinnypark: Thanks! Removed this one and added one via changelogger tool.

tbradsha (approved):

PR #34447, opened 2023-12-05
Contact Form: add margin below global error message

monsieur-z:
## Proposed changes:

Add margin below the form error message, matching the space above.

|Before|After|
|-|-|
|Screenshot 2023-12-04 at 7 10 46 PM|Screenshot 2023-12-04 at 7 10 27 PM|

### Other information:

## Jetpack product discussion

See https://github.com/Automattic/jetpack/pull/34173#issuecomment-1830324528.

No.

## Testing instructions:

- Create a new post
- Make sure at least one field is required
- Open the preview
- Submit the form
- Notice the style changes

### DIFF
--- a/projects/packages/forms/src/contact-form/css/grunion.css
+++ b/projects/packages/forms/src/contact-form/css/grunion.css
@@ -790,7 +790,7 @@ on production builds, the attributes are being reordered, causing side-effects
 }
 
 .contact-form__error {
-	margin-bottom: 0.25rem;
+	margin-bottom: var(--wp--style--block-gap, 1.5rem);
 	padding: 0.5em;
 
 	background-color: var(--jetpack--contact-form--error-color);
--- a/projects/packages/forms/changelog/fix-contact-form-error-message-margin
+++ b/projects/packages/forms/changelog/fix-contact-form-error-message-margin
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Contact Form: add margin below global error message


projects/packages/forms/src/contact-form/css/grunion.css
```diff
@@ -790,7 +790,7 @@ on production builds, the attributes are being reordered, causing side-effects
 }
 
 .contact-form__error {
-	margin-bottom: 0.25rem;
+	margin-bottom: var(--wp--style--block-gap, 1.5rem);
```
jeherve, line 793: I don't know if we should use `--wp--style--block-gap`; that will not be set for classic, non block-based themes.

monsieur-z: It's used throughout `grunion.css`. Shoud I update it everywhere?

> that will not be set for classic, non block-based themes.

What'd be the best way to test that?

jeherve: > Shoud I update it everywhere?

I'd think so, yes.

If you use a classic theme, like Twenty Twenty, instead of using a more recent theme like Twenty Twenty Four, you should be able to test for this.

monsieur-z: All instances in the stylesheet implement the `1.5rem` fallback, which is what's eventually used in the Twenty Twenty theme.

I think it's safe as is, but I don't mind getting rid of the variable if you think it's safer down the road.

jeherve: Should we worry about other classic themes, that may not have any value defined?

monsieur-z: I don't think I understand what you mean by that:
> that may not have any value defined?

If I understand correctly, in the worst-case scenario, `var(--wp--style--block-gap, 1.5rem)` will evaluate to `1.5rem`.

jeherve: Oh, I had misunderstood. Yeah, if they all have the fallback already, that should be fine! 🚢

jeherve (approved):

PR #34450, opened 2023-12-05
Like block: Add iframe to the new Like block (beta)

### LINKED ISSUE #84850: Like block: have the block render an iframe

ivan-ottinger:
Resolves https://github.com/Automattic/wp-calypso/issues/84850.

Markup on 2023-12-05 at 15:48:21

## Proposed changes:

* add `<iframe>` render for self-hosted / Atomic and Simple sites
* add condition that will display either the Like button _widget_ or Like button _block_ 

This PR does _not_ address the rendering of the block in the editor. We have a dedicated issue for that matter: https://github.com/Automattic/wp-calypso/issues/84868.

### Other information:

## Jetpack product discussion

Related project thread: pdDOJh-2JP-p2.

No, it does not.

## Testing instructions:

**Self-hosted site**

6. When you open the post in the front-end, the Like block **button** should load and be working correctly.
7. If you enable a classic theme (e.g. Twenty Nineteen), the Like block **widget** should display instead.
8. Please check for regressions, e.g. nothing should change if the `define( 'JETPACK_BLOCKS_VARIATION', 'beta' );`  is not set.

**Simple site**

1. Sync the changes of this PR to your simple site (e.g. by using the script from the https://github.com/Automattic/jetpack/pull/34450#issuecomment-1840386739 ).
2. Apply the changes from the following patch to your sandbox: D130953-code (if the patch hasn't been merged already).
3. Sandbox the test site.
4. Add `define( 'JETPACK_BLOCKS_VARIATION', 'beta' );` to your `0-sandbox.php` file.
6. When you open the post in the front-end, the Like button **block** should load and be working correctly.
7. If you enable a classic theme (e.g. Twenty Nineteen), the Like button **widget** should display instead.
8. Please check for regressions, e.g. nothing should change if the `define( 'JETPACK_BLOCKS_VARIATION', 'beta' );`  is not set.

**Atomic site**
3. When you open the post in the front-end, the Like block **button** should load and be working correctly.
4. If you enable a classic theme (e.g. Twenty Nineteen), the Like block **widget** should display instead, but both buttons render. This is because the block editor is still present and working - hence it displays the Like button block as well. This needs to be addressed and needs further discussion. 
5. Please check for regressions, e.g. nothing should change if the `define( 'JETPACK_BLOCKS_VARIATION', 'beta' );`  is not set.


TimBroddin (approved):
Tested on Simple & self-hosted (Atomic gave me some unrelated trouble with SFTP) and works as intended! 

CleanShot 2023-12-06 at 15 52 01@2x
CleanShot 2023-12-06 at 15 39 09@2x

projects/plugins/jetpack/extensions/blocks/like/like.php
```diff
@@ -31,21 +31,72 @@ function register_block() {
...
+	$title   = esc_html__( 'Like or Reblog', 'jetpack' );
+
+	/**
+	 * Enable an alternate Likes layout.
+	 *
+	 * @since 12.9
+	 *
+	 * @module likes
+	 *
+	 * @param bool $new_layout Enable the new Likes layout. False by default.
```
TimBroddin, line 59: There's an error in this doc line. It's true by default.
My bad - we can fix this in a follow-up PR.

projects/plugins/jetpack/extensions/blocks/like/like.php
```diff
@@ -31,21 +31,72 @@ function register_block() {
...
+			/** This filter is already documented in modules/sharedaddy/sharing-service.php */
+			apply_filters( 'jetpack_sharing_headline_html', '<h3 class="sd-title">%s</h3>', esc_html__( 'Like this:', 'jetpack' ), 'likes' ),
+			esc_html__( 'Like this:', 'jetpack' )
+		);
+	}
+
+	$name    = sprintf( 'like-post-frame-%1$d-%2$d-%3$s', $blog_id, $post_id, $uniqid );
+	$wrapper = sprintf( 'like-post-wrapper-%1$d-%2$d-%3$s', $blog_id, $post_id, $uniqid );
+
+	$html  = "<div class='sharedaddy sd-block sd-like jetpack-likes-widget-wrapper jetpack-likes-widget-unloaded' id='$wrapper' data-src='$src' data-name='$name' data-title='$title'>";
```
obenland, line 91: @ivan-ottinger Let's please run the variables in this line through `esc_attr`

ivan-ottinger: Hey Konstantin 👋🏼 Thanks for the comment. The way the output is built here reflects how it is in the old / current Like widget so that's why I didn't question it. I will escape the the attributes in a separate PR. 👍🏼

ivan-ottinger: Fixed for the block in https://github.com/Automattic/jetpack/pull/34536 and for the widget in D131398-code. 🙂

obenland: Never question doing the right thing 😉
Thanks for the quick turnaround!

ivan-ottinger: Oh, I meant that I did not pay enough attention to that line - as I used it right from the source where it was already working in production (I didn't even notice the values were not escaped).

Thanks for highlighting that. 🙂

PR #23638, opened 2022-03-25
cli: Don't set `shell: true` in runCommand helper

anomiex:
#### Changes proposed in this Pull Request:

Setting `shell: true` causes shell metacharacters in arguments to
potentially be expanded, which is unsafe. Better to not pass it,
especially since neither of the existing callers rely on it anyway.

#### Jetpack product discussion

None

#### Does this pull request change what data or activity we track or use?

No

#### Testing instructions:

* Try doing `jetpack changelog add` and supplying a name containing spaces to "Name your change file". It should fail without this PR, and succeed with it.

PR #34462, opened 2023-12-05
Add the documentation parser package and CLI command.

zinigor:
This ports the PHP documentation parsing package from dotcom, allowing us to customize the way we parse documents in Jetpack and its packages.

## Proposed changes:
* Add a new package for the doc parser.
* Port the parser class.
* Add the runner script to be picked up by Jetpack CLI.
* Add the Jetpack CLI command to run the script.

*** TODO
- [ ] The Jetpack CLI command needs proper project selection.
- [x] The doc-parser package's dependencies get installed in `wp-content` for some reason.

### Other information:

## Jetpack product discussion
* TBA

No.

## Testing instructions:

* Install Jetpack using `jetpack install -r` and `jetpack install packages/doc-parser`
* Run `jetpack docs`


projects/packages/doc-parser/src/class-doc-parser.php
```diff
@@ -0,0 +1,89 @@
...
+
+		if ( empty( $output_file ) ) {
+			$output_file = 'phpdoc.xml';
+		}
+
+		$directory = realpath( $directory );
+		echo PHP_EOL;
+
+		// Get data from the PHPDoc
+		$json = $this->get_phpdoc_data( $directory );
```
anomiex: Is it JSON or XML? If it's JSON, the file name should probably not be "phpdoc.xml". If it's XML, the variable here should probably not be `$json`.

projects/packages/doc-parser/src/class-doc-parser.php
```diff
@@ -0,0 +1,89 @@
...
+			$path  = dirname( $path );
+
+		} else {
+			ob_start();
+			$files = \WP_Parser\get_wp_files( $path );
+			$error = ob_get_clean();
+
+			if ( $error ) {
+				printf( 'Problem with %1$s: %2$s', $path, $error ); // phpcs:ignore
+				exit;
```
anomiex: Error cases should probably `exit(1)` or something so a caller can tell it failed.

tools/cli/commands/docs.js
```diff
@@ -0,0 +1,50 @@
...
+	return yargs;
+}
+
+/**
+ * Handle args for docs command.
+ *
+ * @param {argv}  argv - the arguments passed.
+ */
+export async function docsCli( argv ) {
+	child_process.spawnSync(
```
anomiex: You may want to do something along the lines of https://github.com/Automattic/jetpack/blob/28c6b4d382d4c0f68df97f2d9b2838ecda42e271/tools/cli/commands/changelog.js#L505-L525 to handle the case where `composer update` needs to be run for the program to work.

tools/cli/commands/docs.js
```diff
@@ -0,0 +1,50 @@
...
+		'php',
+		[
+			path.resolve( './projects/packages/doc-parser/runner.php' ),
+			path.resolve( `./projects/plugins/${ argv.$0 }` ),
+		],
+		{
+			cwd: path.resolve( './' ),
+			shell: true,
```
anomiex: Is there a need for `shell: true`? If not, let's not use it. See #23638 for details.

projects/packages/doc-parser/.gitattributes
```diff
@@ -0,0 +1,17 @@
...
+# Files to include in the mirror repo, but excluded via gitignore
+# Remember to end all directories with `/**` to properly tag every file.
+# /src/js/example.min.js  production-include
+
+# Files to exclude from the mirror repo, but included in the monorepo.
+# Remember to end all directories with `/**` to properly tag every file.
+.gitignore        production-exclude
+changelog/**      production-exclude
+phpunit.xml.dist  production-exclude
+.phpcs.dir.xml    production-exclude
```
anomiex:
```suggestion
.phpcs.dir.xml    production-exclude
.phpcs.dir.phpcompatibility.xml    production-exclude
```

projects/packages/doc-parser/.phpcs.dir.xml
```diff
@@ -0,0 +1,37 @@
...
+	</rule>
+	<rule ref="WordPress.WP.AlternativeFunctions.file_system_operations_file_put_contents">
+		<severity>0</severity>
+	</rule>
+	<rule ref="WordPress.Security.EscapeOutput.OutputNotEscaped">
+		<severity>0</severity>
+	</rule>
+	<rule ref="WordPress.WP.AlternativeFunctions.json_encode_json_encode">
+		<severity>0</severity>
+	</rule>
```
anomiex:
```suggestion
	<rule ref="Jetpack-NoWP" />
	<rule ref="Jetpack-Compat-82" />
```

projects/packages/doc-parser/composer.json
```diff
@@ -0,0 +1,63 @@
+{
+	"name": "automattic/jetpack-doc-parser",
+	"description": "A parser for Jetpack Monorepo to gather inline documentation.",
+	"type": "jetpack-library",
+	"license": "GPL-2.0-or-later",
+	"require": {
+		"php": ">=7.0",
```
anomiex: Since you're using `<rule ref="Jetpack-Compat-82" />` for phpcs, this should probably match
```suggestion
		"php": ">=8.2",
```

projects/packages/doc-parser/composer.json
```diff
@@ -0,0 +1,63 @@
...
+		"wordpress/phpdoc-parser": "dev-master"
```
anomiex: Huh, the project has never done a release? That's going to be problematic.

As things are, we could put this in `require-dev` instead of `require` then depend on something like "dev-master#d4d8751d1750c82098865a182dd2d47692e18a6a". But it's not really a `require-dev`.

Or we could put an exception for this package into `.github/files/lint-project-structure.sh`.

Or are we really going to publish this to Packagist? If not, we could move this all somewhere under `tools/`and not have to worry about the project structure linting at all.

projects/packages/doc-parser/package.json
anomiex, line 1: Are we likely to ever have any JS in here? If not, we can delete this.

projects/packages/doc-parser/src/class-doc-parser.php
```diff
@@ -0,0 +1,92 @@
...
+		$json = array();
+		foreach ( $directories as $directory ) {
+			$directory = realpath( $directory );
+			echo PHP_EOL;
+
+			// Get data from the PHPDoc
+			$json = array(
+				...$json,
+				$this->get_phpdoc_data( $directory ),
+			);
```
anomiex: To fix the "PHP lint" jobs on this line, we'll have to add `SKIPS` for appropriate PHP versions in `tools/parallel-lint.sh`.

I wonder if I can make that read from composer.json though...

OTOH, you could avoid all the trouble by just using the normal PHP idiom like this instead:
```suggestion
			$json[] = $this->get_phpdoc_data( $directory );
```

tools/cli/commands/docs.js
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Handle args for docs command.
+ *
+ * @param {argv}  argv - the arguments passed.
+ */
+export async function docsCli( argv ) {
+	let paths;
+	if ( 'all' === argv.project ) {
+		// "All" is a keyword for Jetpack plus packages.
```
anomiex, line 43: Not any of the other plugins?

tools/cli/commands/docs.js
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	const parser_options = [ path.resolve( './projects/packages/doc-parser/runner.php' ), ...paths ];
+
+	let data = child_process.spawnSync( 'php', parser_options, {
+		cwd: path.resolve( './' ),
+		stdio: [ 'inherit', 'inherit', 'ignore' ],
+	} );
+
+	if ( data.status !== 0 ) {
```
anomiex, line 57: There's a bit of a problem here in that this might result in it running a long extraction only to fail at the end, then try `composer update` and re-run the long extraction.

In the changelogger code this was based on the first check is to run with with no arguments, which just displays some help text and exits. That's sufficient to check that Composer has been run.

projects/packages/doc-parser/src/class-doc-parser.php
anomiex: > I wonder if I can make that read from composer.json though...

#34497

projects/packages/doc-parser/src/class-doc-parser.php
zinigor: Yeah, thanks, resolved!

tools/cli/commands/docs.js
zinigor: I figured we'd start with this default, but we should probably discuss this more.
@ice9js  what do you think? Shall we include every other plugin that the Monorepo has in the export?

tools/cli/commands/docs.js
zinigor: I have removed all that when I moved the tool into cli/helpers.

tools/cli/helpers/doc-parser/.gitignore
```diff
@@ -0,0 +1,2 @@
+vendor/
+node_modules/
```
anomiex: No need for `node_modules/` here since there's no package.json and no match for this path in `pnpm-workspace.yaml`, although it doesn't really hurt anything either.
```suggestion
```

tools/cli/helpers/doc-parser/.phpcs.dir.xml
anomiex, line 1: Now that this is under `tools/`, this file is redundant to `tools/.phpcs.dir.xml`.

If it weren't, you'd need a `.phpcs.dir.phpcompatibility.xml` (with contents like `tools/.phpcs.dir.phpcompatibility.xml`) to match what's here for the PHPCompatibility CI job.

tools/cli/helpers/doc-parser/CHANGELOG.md
anomiex, line 1: Not sure if we're going to bother to maintain a changelog for this. CI and such won't complain about missing changelog entries as they only look under `projects/`.

If we're not going to bother, we could remove the changelogger dep and all the config under `.extra.changelogger` from composer.json.

tools/cli/helpers/doc-parser/composer.json
```diff
@@ -0,0 +1,63 @@
...
+			"src/"
+		]
+	},
+	"scripts": {
+		"phpunit": [
+			"./vendor/phpunit/phpunit/phpunit --colors=always"
+		],
+		"test-php": [
+			"@composer phpunit"
+		]
```
anomiex: If we do decide to write tests for this, rigging them up to run would probably mean adding a test-php script in the monorepo root composer.json to do similarly to what it does for test-js to get the CLI's tests to run.

If we're not going to bother, we could remove the `yoast/phpunit-polyfills` dep and these scripts.

tools/cli/helpers/doc-parser/composer.json
```diff
@@ -0,0 +1,63 @@
...
+	},
+	"repositories": [
+		{
+			"type": "path",
+			"url": "../../packages/*",
```
anomiex:
```suggestion
			"url": "../../../../projects/packages/*",
```

tools/cli/helpers/doc-parser/composer.json
```diff
@@ -0,0 +1,63 @@
...
+			"link-template": "https://github.com/Automattic/jetpack-doc-parser/compare/v${old}...v${new}"
+		},
+		"autotagger": true,
+		"branch-alias": {
+			"dev-trunk": "0.1.x-dev"
+		},
+		"textdomain": "jetpack-doc-parser",
+		"version-constants": {
+			"::PACKAGE_VERSION": "src/class-doc-parser.php"
+		},
```
anomiex: `mirror-repo`, `autotagger`, `branch-alias`, `text-domain`, and `version-constants` are not really needed since this isn't going to be released anywhere.

tools/cli/helpers/doc-parser/composer.json
```diff
@@ -0,0 +1,63 @@
...
+		"installer-disable": [
+			"wordpress"
+		]
+	},
+	"suggest": {
+		"automattic/jetpack-autoloader": "Allow for better interoperability with other plugins that use this package."
+	},
```
anomiex: This too is pointless since this isn't going to be a dep of anything.

tools/cli/helpers/doc-parser/src/class-doc-parser.php
```diff
@@ -0,0 +1,94 @@
...
+		$json = array();
+		foreach ( $directories as $directory ) {
+			$directory = realpath( $directory );
+			echo PHP_EOL;
+
+			// Get data from the PHPDoc
+			$json = array(
+				...$json,
+				$this->get_phpdoc_data( $directory ),
+			);
```
anomiex: Hmm. Since we flagged `tools/` with Jetpack-Compat-82 in #34207, we should do the same for parallel-lint. #34550 should fix that.

OTOH, I still think this would be more idiomatic.
```suggestion
			$json[] = $this->get_phpdoc_data( $directory );
```

tools/cli/commands/docs.js
anomiex: I think you'll probably find you need to add it back if you try running this on a fresh monorepo checkout.

tools/cli/helpers/doc-parser/composer.json
```diff
@@ -0,0 +1,43 @@
...
+			"src/"
+		]
+	},
+	"minimum-stability": "dev",
+	"prefer-stable": true,
+	"extra": {
+		"mirror-repo": "Automattic/jetpack-doc-parser",
+		"changelogger": {
+			"link-template": "https://github.com/Automattic/jetpack-doc-parser/compare/v${old}...v${new}"
+		},
```
anomiex: Since you dropped the changelogger dep, at least this part should go away too.
```suggestion
```

tools/cli/helpers/doc-parser/docs.json
anomiex, line 1: I'm a little confused as to what's going on with this file. As it's currently used (perhaps all the code isn't done?) all we do is read the keys under `.parents` and process those docs, doing nothing with the array of other docs.

tools/cli/helpers/doc-parser/composer.json
zinigor: I have removed "repositories" altogether.

tools/cli/helpers/doc-parser/composer.json
zinigor: Thanks, removed!

tools/cli/helpers/doc-parser/composer.json
zinigor: Thanks, good point, removed!

tools/cli/helpers/doc-parser/composer.json
zinigor: I have removed that in a subsequent commit, thanks!

tools/cli/helpers/doc-parser/.phpcs.dir.xml
zinigor: Removed the file entirely, thanks!

tools/cli/helpers/doc-parser/src/class-doc-parser.php
zinigor: You're right, they are equivalent, I'm going to merge your suggestion.

tools/cli/helpers/doc-parser/composer.lock
anomiex, line 1: We shouldn't check in this `composer.lock` as long as we have the `dev-master` dep in `composer.json`. The `composer.lock` will go out of date every time something is pushed to the upstream's master branch, which will make the "Lock files are up to date" check randomly start failing.

zinigor: Removed, thanks!

tools/cli/commands/docs.js
zinigor: I think I got it right this time, can you check again please?

tools/cli/helpers/doc-parser/docs.json
zinigor: Yes, fixed, thank you!

tools/cli/commands/docs.js
```diff
@@ -0,0 +1,77 @@
...
+		// Something is wrong, let's try to run composer update.
+		console.debug( 'Preparing doc-parser package...' );
+		data = child_process.spawnSync( 'composer', [ 'update' ], {
+			cwd: path.resolve( './tools/cli/helpers/doc-parser' ),
+			stdio: [ 'inherit', 'inherit', 'ignore' ],
+		} );
+		if ( data.status !== 0 ) {
+			// Running composer update didn't help, exiting.
+			console.error(
+				"Failed to prepare the doc-parser package. Try running 'jetpack install -v packages/doc-parser'."
```
anomiex: That command won't work.
```suggestion
				`Failed to prepare the doc-parser package. Try running 'composer -d ${ path.resolve( './tools/cli/helpers/doc-parser' ) } update'.`
```

tools/cli/commands/docs.js
anomiex: It'll work, but it has weird behavior in cases where the composer packages are fine but something in the script itself breaks. Stick an `exit(1)` at the end of the runner.php script, for example:
```
$ jetpack docs beta

Extracting PHPDoc from /usr/local/src/automattic/jetpack/projects/plugins/beta.
Data exported to phpdoc.json
Extracting Markdown from readme.md.
Extracting Markdown from docs/quick-start.md.
Extracting Markdown from docs/CONTRIBUTING.md.
Extracting Markdown from docs/CONTRIBUTING.md.
Extracting Markdown from CODE-OF-CONDUCT.md.
Extracting Markdown from docs/monorepo.md.
Extracting Markdown from docs/quick-start.md.
Data exported to markdown.json
Preparing doc-parser package...

Extracting PHPDoc from /usr/local/src/automattic/jetpack/projects/plugins/beta.
Data exported to phpdoc.json
Extracting Markdown from readme.md.
Extracting Markdown from docs/quick-start.md.
Extracting Markdown from docs/CONTRIBUTING.md.
Extracting Markdown from docs/CONTRIBUTING.md.
Extracting Markdown from CODE-OF-CONDUCT.md.
Extracting Markdown from docs/monorepo.md.
Extracting Markdown from docs/quick-start.md.
Data exported to markdown.json
```

sdixon194:
Hmm I tried this again, but I can't get the parser to work. Everytime I run it I get:

> Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 4096 bytes) in /Users/stevedixon/jetpack-dev/jetpack/tools/cli/helpers/doc-parser/vendor/nikic/php-parser/lib/PhpParser/Lexer.php on line 54

tools/cli/commands/docs.js
```diff
@@ -0,0 +1,77 @@
...
+		// Something is wrong, let's try to run composer update.
+		console.debug( 'Preparing doc-parser package...' );
+		data = child_process.spawnSync( 'composer', [ 'update' ], {
+			cwd: path.resolve( './tools/cli/helpers/doc-parser' ),
+			stdio: [ 'inherit', 'inherit', 'ignore' ],
+		} );
+		if ( data.status !== 0 ) {
+			// Running composer update didn't help, exiting.
+			console.error(
+				`Failed to prepare the doc-parser package. Try running 'composer -d ${ path.resolve( './tools/cli/helpers/doc-parser' ) } update'.`
```
anomiex: > Replace `·'./tools/cli/helpers/doc-parser'·` with `⏎↹↹↹↹↹'./tools/cli/helpers/doc-parser'⏎↹↹↹↹`

Personally I always hate when Prettier wants to stick newlines in the middle of backquote-strings, it tends to screw up syntax highlighting in my vim.

It might be better to assign a `const cwd = ` above and use it on line 61 too.

tools/cli/commands/docs.js
```diff
@@ -0,0 +1,78 @@
...
+	if ( data.status !== 0 ) {
+		// Something is wrong, let's try to run composer update.
+		console.debug( 'Preparing doc-parser package...' );
+		data = child_process.spawnSync( 'composer', [ 'update' ], {
+			cwd: path.resolve( './tools/cli/helpers/doc-parser' ),
+			stdio: [ 'inherit', 'inherit', 'ignore' ],
+		} );
+		if ( data.status !== 0 ) {
+			// Running composer update didn't help, exiting.
+			const cwd = path.resolve( './tools/cli/helpers/doc-parser' );
```
anomiex: Feel free to ignore the suggestion, but this might make sense
```suggestion
		const cwd = path.resolve( './tools/cli/helpers/doc-parser' );
		data = child_process.spawnSync( 'composer', [ 'update' ], {
			cwd,
			stdio: [ 'inherit', 'inherit', 'ignore' ],
		} );
		if ( data.status !== 0 ) {
			// Running composer update didn't help, exiting.
```

tools/cli/commands/docs.js
ice9js: That sounds about right - let's start with that and add support for other plugins once we have a better feeling for how to organize things on the site.

sdixon194 (approved):

PR #34474, opened 2023-12-05
Calendly Block: Fix Embed Attributes Being Stripped

### LINKED ISSUE #34445: Calendly Block Inline Embed Custom Colors do not appear on WoA Site
### Impacted plugin

Jetpack

### Quick summary

Calendly's inline embeds can be set to use custom colors; we outline this in our support guides for the block.  However, the custom colors appear properly in the editor, but not on the front end.

### Steps to reproduce

1. Create a Calendly Account
2. Create an event
3. Follow these steps to set custom colors: https://wordpress.com/support/wordpress-editor/blocks/calendly-block/#change-the-look-and-feel-of-the-calendly-embed
4. Embed the new code
5. Publish the site and view the front end

I expected the custom colors I set in Calendly and saw in the editor to appear on the front end.

### What actually happened

The Calendly block displays default colors on the front end.

**Editor**
Markup on 2023-12-04 at 14:30:01

**Live Site**
Markup on 2023-12-04 at 14:30:34

### Impact

Some (< 50%)

### Available workarounds?

Atomic

### Logs or notes

Tested on Atomic site with theme Twenty Twenty Four active. 

Customer report: 7404467-zen
Theme: Creatio

They mentioned:

>i have never had the custom colour showing up for me unfortunately

Aurorum:
Fixes #34445

## Proposed changes:

`esc_url` turns this: `https://calendly.com/user/30min?hide_event_type_details=0&background_color=ac1e1e&text_color=ff1c1c&primary_color=16345f`

Into this: `https://calendly.com/user/30min?embed_domain=a96d-2a00-23c8-3f07-3e01-c57f-94f3-a43c-a1fa.ngrok-free.app&embed_type=Inline&hide_event_type_details=0&=undefined`

I've tried to research why this might be happening, but I can't figure it out. `esc_url_raw` fixes this problem, and seems to be encouraged [here](https://developer.wordpress.org/reference/functions/esc_url_raw/#more-information) since the URL isn't displayed, but I'm curious if there's any downsides that explain why it wasn't used in the first place. 

### Other information:

## Jetpack product discussion

N/A.

N/A.

## Testing instructions:

Follow the instructions listed here to embed a Calendly block with custom colours: https://wordpress.com/support/wordpress-editor/blocks/calendly-block/#change-the-look-and-feel-of-the-calendly-embed

On the front end, those colours should correctly display with this change.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/calendly/calendly.php
+++ b/projects/plugins/jetpack/extensions/blocks/calendly/calendly.php
@@ -112,14 +112,17 @@ function load_assets( $attr, $content ) {
 			esc_attr( $block_id )
 		);
 		$script  = <<<JS_END
-jetpackInitCalendly( %s, '%s' );
+jetpackInitCalendly( %s, %s );
 JS_END;
 		wp_add_inline_script(
 			'jetpack-calendly-external-js',
 			sprintf(
 				$script,
-				wp_json_encode( esc_url_raw( $url ), JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES ),
-				esc_attr( $block_id )
+				wp_json_encode(
+					esc_url_raw( $url ),
+					JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES | JSON_HEX_TAG | JSON_HEX_AMP
+				),
+				wp_json_encode( $block_id, JSON_HEX_TAG | JSON_HEX_AMP )
 			)
 		);
 	}
--- a/projects/plugins/jetpack/extensions/blocks/calendly/calendly.php
+++ b/projects/plugins/jetpack/extensions/blocks/calendly/calendly.php
@@ -114,7 +114,7 @@ function load_assets( $attr, $content ) {
 		$script  = <<<JS_END
 jetpackInitCalendly( '%s', '%s' );
 JS_END;
-		wp_add_inline_script( 'jetpack-calendly-external-js', sprintf( $script, esc_url( $url ), esc_js( $block_id ) ) );
+		wp_add_inline_script( 'jetpack-calendly-external-js', sprintf( $script, esc_url_raw( $url ), esc_js( $block_id ) ) );
 	}
 
 	return $content;
--- a/projects/plugins/jetpack/extensions/blocks/calendly/calendly.php
+++ b/projects/plugins/jetpack/extensions/blocks/calendly/calendly.php
@@ -122,7 +122,7 @@ JS_END;
 					esc_url_raw( $url ),
 					JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES | JSON_HEX_TAG | JSON_HEX_AMP
 				),
-				wp_json_encode( $block_id, JSON_HEX_TAG | JSON_HEX_AMP )
+				wp_json_encode( $block_id, JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES | JSON_HEX_TAG | JSON_HEX_AMP )
 			)
 		);
 	}
--- a/projects/plugins/jetpack/extensions/blocks/calendly/calendly.php
+++ b/projects/plugins/jetpack/extensions/blocks/calendly/calendly.php
@@ -112,9 +112,16 @@ function load_assets( $attr, $content ) {
 			esc_attr( $block_id )
 		);
 		$script  = <<<JS_END
-jetpackInitCalendly( '%s', '%s' );
+jetpackInitCalendly( %s, '%s' );
 JS_END;
-		wp_add_inline_script( 'jetpack-calendly-external-js', sprintf( $script, esc_url_raw( $url ), esc_js( $block_id ) ) );
+		wp_add_inline_script(
+			'jetpack-calendly-external-js',
+			sprintf(
+				$script,
+				wp_json_encode( esc_url_raw( $url ), JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES ),
+				esc_attr( $block_id )
+			)
+		);
 	}
 
 	return $content;
--- a/projects/plugins/jetpack/changelog/fix-calendly-block-attributes
+++ b/projects/plugins/jetpack/changelog/fix-calendly-block-attributes
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
+Calendly block: fix custom colours being stripped.


projects/plugins/jetpack/extensions/blocks/calendly/calendly.php
```diff
@@ -114,7 +114,7 @@ function load_assets( $attr, $content ) {
 		$script  = <<<JS_END
 jetpackInitCalendly( '%s', '%s' );
 JS_END;
-		wp_add_inline_script( 'jetpack-calendly-external-js', sprintf( $script, esc_url( $url ), esc_js( $block_id ) ) );
+		wp_add_inline_script( 'jetpack-calendly-external-js', sprintf( $script, esc_url_raw( $url ), esc_js( $block_id ) ) );
```
anomiex: It seems like people find it hard to get escaping right for inline JS in WordPress code. Then again, WordPress's so-called "escaping" functions are pretty poor to start with.

* The old code was wrong in that it used `esc_url` which changes `&` to `&amp;` incorrectly for this context.
* Both the old and new are wrong in that they don't apply the escaping that's actually needed for inline JavaScript. If the URL happens to contain a `'` (rather than `%27`), things may break.
* The use of `esc_js()` for `$block_id` is wrong too, as that would also do the `&amp;` thing since it's intended for use in tag attributes like `onclick`. But since it's always alphanumeric that never comes up.

Probably what we really need here to do it right is something like this
```js
		$script  = <<<JS_END
jetpackInitCalendly( %s, %s );
JS_END;
		wp_add_inline_script(
			'jetpack-calendly-external-js',
			sprintf(
				$script,
				wp_json_encode( esc_url_raw( $url ), JSON_HEX_TAG | JSON_HEX_AMP ),
				wp_json_encode( $block_id, JSON_HEX_TAG | JSON_HEX_AMP )
			)
		);
```

anomiex (commented):
Just took a quick glance, I don't have a calendly account handy to create a URL to test with. If you can provide a test URL, that would be helpful.

> `esc_url` turns this: `https://calendly.com/user/30min?hide_event_type_details=0&background_color=ac1e1e&text_color=ff1c1c&primary_color=16345f`
>
> Into this: `https://calendly.com/user/30min?embed_domain=a96d-2a00-23c8-3f07-3e01-c57f-94f3-a43c-a1fa.ngrok-free.app&embed_type=Inline&hide_event_type_details=0&=undefined`

I suspect that's not entirely true. More likely it's turning it into `https://calendly.com/user/30min?hide_event_type_details=0&amp;background_color=ac1e1e&amp;text_color=ff1c1c&amp;primary_color=16345f`, then the JS it's passed to is stripping the unrecognized parameter names "amp;background_color", "amp;text_color", and "amp;primary_color" when adding the "embed_domain" parameter.

projects/plugins/jetpack/extensions/blocks/calendly/calendly.php
Aurorum: TIL! And you're completely correct on what's happening with the link. 

The problem is that your solution seems to result in this: `"https:\/\/calendly.com\/user\/30min?hide_event_type_details=0\u0026background_color=ac1e1e\u0026text_color=ff1c1c\u0026primary_color=16345f"`

Would you mind checking the changes which I've committed that handle those? Two things which I'm uncertain of though: 

1) I don't think we need to flag hex tags because that's already done when it's included in `add_query_arg`
2) As I understand it, can we not just use `esc_attr` for `$block_id`? Especially if there's no need to use `esc_js` for inline JavaScript. 

Thanks!! :)

Aurorum: Here's a link for testing, by the way: https://calendly.com/sgdfkopfsdgpoksdfgkpodgfskpfsdkpdfsgdfg/30min (sorry - spam link that I just created for this)

jeherve (dismissed):
This tests well for me. 👍

projects/plugins/jetpack/extensions/blocks/calendly/calendly.php
anomiex: > The problem is that your solution seems to result in this: `"https:\/\/calendly.com\/user\/30min?hide_event_type_details=0\u0026background_color=ac1e1e\u0026text_color=ff1c1c\u0026primary_color=16345f"`

Ah, yeah. The value is valid, but adding `JSON_UNESCAPED_SLASHES` (and maybe `JSON_UNESCAPED_UNICODE` too) to the `json_encode` options should be fine to skip some unnecessary backslashes.

But the `\u0026` bits in there are intended to avoid potential edge cases in old/strange browsers.

> 1. I don't think we need to flag hex tags because that's already done when it's included in `add_query_arg`

In the general case we want hex tags to avoid problems if someone manages to get `</script>` in the value. Yeah, that won't happen for the URL because `esc_url_raw()` strips any angle brackets, but IMO it's better to specify it anyway in case someone copy-pastes the flags from this example for encoding a non-url value.

> 2. As I understand it, can we not just use `esc_attr` for `$block_id`? Especially if there's no need to use `esc_js` for inline JavaScript.

Nope, that has the same `&amp;` problem that `esc_js` does.

Ideally you should only use `esc_attr` when the value is going into an HTML tag's attribute value. And even then depending on the context it might do the wrong thing.

Aurorum: Ah, okay, makes sense. I've changed it to match your recommendations - just with the extra flags to prevent the link breaking. Thanks! :)

projects/plugins/jetpack/extensions/blocks/calendly/calendly.php
```diff
@@ -112,9 +112,19 @@ function load_assets( $attr, $content ) {
...
-		wp_add_inline_script( 'jetpack-calendly-external-js', sprintf( $script, esc_url( $url ), esc_js( $block_id ) ) );
+		wp_add_inline_script(
+			'jetpack-calendly-external-js',
+			sprintf(
+				$script,
+				wp_json_encode(
+					esc_url_raw( $url ),
+					JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES | JSON_HEX_TAG | JSON_HEX_AMP
+				),
+				wp_json_encode( $block_id, JSON_HEX_TAG | JSON_HEX_AMP )
```
anomiex: May as well use the same flags here
```suggestion
				wp_json_encode( $block_id, JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES | JSON_HEX_TAG | JSON_HEX_AMP )
```

Aurorum: Sure, done 👍

anomiex (approved):
LGTM, works. Thanks!

PR #34153, opened 2023-11-15
Gutenberg: Add Top Posts and Pages Block 

Aurorum:
Fixes #32937

## Proposed changes:

Converts the legacy Top Posts & Pages widget into a Gutenberg block! It has the same features; you can still choose to filter out different types of media and customise the display, but the block also allows you to choose the timeframe of stats which to draw on. 

The only exception is that this block is based entirely on views/traffic. The widget allowed determining top content from Likes, but I don't have access to a WordPress.com environment that would allow me to add this, so I'm afraid that somebody else would need to handle that in future. I personally don't feel like that's a massive deal though, at least not for this first version. 

Screenshot 2023-11-17 at 23 34 23

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

N/A

## Does this pull request change what data or activity we track or use?

N/A

## Testing instructions:
I'm hoping that the block should be fairly intuitive to use! Just like the Related Posts block, you should be prompted to activate Stats if the module is disabled. In fact, I've drawn quite heavily on that block - for example, the metadata and layout settings are the same. 

Screenshot 2023-11-18 at 15 26 04

If Stats is active, some posts should automatically display. If there aren't enough posts within the given timeframe, then random posts should be used to fill the rest of the block. There are some settings to filter out content and to choose which metadata information to display, so definitely verify that those work. 

Ultimately though, it should render nicely in the frontend. 

Screenshot 2023-11-18 at 15 28 42

PR #34479, opened 2023-12-06
Top Posts & Pages Block: Refactor Endpoint 

Aurorum:
Addresses https://github.com/Automattic/jetpack/pull/34153#discussion_r1416408893

## Proposed changes:

* Refactors the endpoint used by the Top Posts and Pages block to use a helper function, which also avoids making an extra API request when rendering the block. cc @kangzj 

### Other information:

## Jetpack product discussion

See https://github.com/Automattic/jetpack/pull/34153#discussion_r1416408893

> I'm wondering whether the remote request could be saved by calling a function that returns top posts, which could save us the HTTP request overhead, and hence improve the performance of the block. More importantly, it would work with WP.COM with no issues. To accomplish this, we could refactor the following code to a helper or something be shared by the API and the block.

No.

## Testing instructions:

You'll need to enable Beta blocks to use the Top Posts & Pages block. Confirm that after inserting it, there's still content displaying. You can also make a request here and there should be no difference to what currently happens: `/wp-json/wpcom/v2/top-posts?period=1&number=3&types=post,page`

Changing `number` should ensure the endpoint returns that number of items; `types` can be used to set the content types which are included.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/top-posts/top-posts.php
+++ b/projects/plugins/jetpack/extensions/blocks/top-posts/top-posts.php
@@ -13,6 +13,11 @@ use Automattic\Jetpack\Blocks;
 use Automattic\Jetpack\Connection\Manager as Connection_Manager;
 use Automattic\Jetpack\Status;
 use Jetpack_Gutenberg;
+use Jetpack_Top_Posts_Helper;
+
+if ( ! class_exists( 'Jetpack_Top_Posts_Helper' ) ) {
+	require_once JETPACK__PLUGIN_DIR . '/_inc/lib/class-jetpack-top-posts-helper.php';
+}
 
 /**
  * Registers the block for use in Gutenberg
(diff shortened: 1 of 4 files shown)


Aurorum:
Added a commit which I think should fix the Lint issue

projects/plugins/jetpack/extensions/blocks/top-posts/top-posts.php
```diff
@@ -46,19 +51,15 @@ function load_assets( $attributes ) {
...
-		$attributes['postsToShow'],
-		implode( ',', array_keys( array_filter( $attributes['postTypes'] ) ) )
-	);
+	$period = $attributes['period'];
+	$number = $attributes['postsToShow'];
+	$types  = implode( ',', array_keys( array_filter( $attributes['postTypes'] ) ) );
 
-	$request = wp_remote_get( home_url( $request_url ) );
-	$data    = json_decode( wp_remote_retrieve_body( $request ), true );
+	$data = Jetpack_Top_Posts_Helper::get_top_posts( $period, $number, $types );
```
kangzj, line 60: The `context` of Each item from `$data` is type of `WP_Term` now because it doesn't go thru the serialization and serialization processes, so L104 is failing. I guess we could either force covert `context` to array or access it as an object. The related error is like the following,

```
[06-Dec-2023 21:19:01 UTC] PHP Fatal error:  Uncaught Error: Cannot use object of type WP_Term as array in /usr/local/src/jetpack-monorepo/projects/plugins/jetpack/extensions/blocks/top-posts/top-posts.php:104
Stack trace:
#0 /var/www/html/wp-includes/class-wp-block.php(258): Automattic\Jetpack\Extensions\Top_Posts\load_assets()
#1 /var/www/html/wp-includes/blocks.php(1484): WP_Block->render()
#2 /var/www/html/wp-includes/blocks.php(1522): render_block()
#3 /var/www/html/wp-includes/class-wp-hook.php(324): do_blocks()
#4 /var/www/html/wp-includes/plugin.php(205): WP_Hook->apply_filters()
#5 /var/www/html/wp-includes/post-template.php(256): apply_filters()
#6 /var/www/html/wp-content/themes/twentytwentyone/template-parts/content/content-single.php(23): the_content()
#7 /var/www/html/wp-includes/template.php(792): require('...')
#8 /var/www/html/wp-includes/template.php(725): load_template()
#9 /var/www/html/wp-includes/general-template.php(206): locate_template()
#10 /var/www/html/wp-content/themes/twentytwentyone/single.php(18): get_template_part()
#11 /var/www/html/wp-includes/template-loader.php(106): include('...')
#12 /var/www/html/wp-blog-header.php(19): require_once('...')
#13 /var/www/html/index.php(17): require('...')
#14 {main}
  thrown in /usr/local/src/jetpack-monorepo/projects/plugins/jetpack/extensions/blocks/top-posts/top-posts.php on line 104
```

kangzj (commented):
Nice refactoring. I found an error and added comment inline. Thanks for doing this 🙇

projects/plugins/jetpack/extensions/blocks/top-posts/top-posts.php
Aurorum: Whoops, that's not good. Should be fixed now - thanks!

kangzj (approved):
It works well and looks good 👍

PR #34490, opened 2023-12-06
CRM: Fix quote placeholders on the quote template, client portal, pdf and emails

coder-karen:
## Proposed changes:

* This PR focuses on fixing Quote ID, URL and Edit URL quote placeholders when used in quote templates as well as email templates (quote accepted and proposal notification), and also where displayed in the Client Portal and via PDF.
* It is also the base PR upon which two other PRs with further placeholder fixes are based (see below). Because of that, there are some general 'additional notes' below related to other uncovered placeholders.
* The Quote ID placeholder as well as Quote URL are now not rendered in the Quote Editor when added to a Quote Template, but they are rendered elsewhere (Client Portal, PDF, Quote accepted and proposal notification mail templates).
* The Quote Edit URL placeholder won't display anything in the Quote Editor if added to a Quote Template, nor will it display in the Client Portal or PDF (as a link to edit the Quote URL doesn't make sense to share to a client). Similarly it won't display in the quote notification email, however it will show correctly in the quote accepted email notification. 

There are two further PRs based on this PR:
- https://github.com/Automattic/jetpack/pull/34569
- https://github.com/Automattic/jetpack/pull/34568

Related P2 post: pbhBOz-4fT-p2

## Additional notes:

- I've chosen to leave Quote checkbox and autonumber custom fields as 'won'tfix' (they work in mail templates, not in the quote editor, client portal or pdf). The reason they don't work in the quote editor and therefore preview and pdf is:
  - for the checkbox the relevant checkbox names are different for each checkbox within a group, and that not matching the parent group name (number additions to the names). Solving that would like involve  creating an additional hidden input so the name matches the custom field name, and setting it's value to also be the same as all checked checkbox values via JS. 
  - For autonumber custom fields, these are added after the quote is initially generated so there is a similar problem as with quote ID and quote URL - however as it's a custom field we can't know the custom field name in advance adding extra complication. 
  - In both cases, how likely are people to use these? Hence it seems a wontfix given the complication imo (but as with all placeholder availability we should clarify where it works), and I can also add a closed GH issue just for searchability.
- For the same reason, I've left adding currency formatting and adding a currency type to the quote price custom field (also in mail templates - custom price fields aren't showing a currency value)

## Next Steps

Once all relevant PRs are merged, then I'll document which placeholders work where on a comment on this issue: https://github.com/Automattic/zero-bs-crm/issues/3005
The Quote Template placeholder support doc also needs updating: https://github.com/Automattic/jedi-jetpack-enhancements-of-docs-for-ai/issues/327

### Other information:

## Jetpack product discussion

There are several related issues:
- https://github.com/Automattic/zero-bs-crm/issues/3004 (a parent issue of sorts, so this issue will not close that)
- https://github.com/Automattic/zero-bs-crm/issues/1601
- https://github.com/Automattic/zero-bs-crm/issues/3060

As a bonus, this PR will also fix the following issues:
- https://github.com/Automattic/zero-bs-crm/issues/3384

No.

## Testing instructions:

* In a new Quote Template, add the following placeholders: `##QUOTE-ID##`, `##QUOTE-EDIT-URL##` and `##QUOTE-URL##`. Save that template.
* Create a new quote and both before and after saving notice that both the Quote ID and Quote URL placeholders stay unrendered, whereas the Quote Edit URL placeholder is removed.
* After clicking save, preview the quote in the Client Portal. All fields except the for the Quote Edit URL should display correctly.
* Then download the PDF for the quote, and the same should be visible - all but the Quote Edit URL.
* To test mail templates - `/wp-admin/admin.php?page=zbs-email-template` - test the quote accepted email template by accepting the quote you've just created in the client portal and viewing the admin email. Alternatively, add an error log before the final `$html` return in `zeroBSCRM_quote_generateAcceptNotifHTML()` and check the result. You should notice both the Quote Edit URL and Quote URL should be visible. You can add the Quote ID placeholder and that should also display.
* To test the quote proposal email, add the Quote URL and Quote ID placeholders to the 'You have received a new proposal' template - `/wp-admin/admin.php?page=zbs-email-templates&zbs_template_id=4`. Test this by sending the new quote via email in teh quote editor, or by adding an error_log before the `$html` is returned in `zeroBSCRM_quote_generateNotificationHTML()`.


projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
```diff
@@ -717,29 +717,31 @@ function ZeroBSCRM_get_quote_template() {
 	// } Retrive deets
 	$customer_ID = -1;
 	if ( isset( $_POST['cust_id'] ) ) {
-		$customer_ID = (int) sanitize_text_field( $_POST['cust_id'] );
+		$customer_ID = (int) sanitize_text_field( wp_unslash( $_POST['cust_id'] ) ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
```
tbradsha: Better yet:

```suggestion
		$customer_ID = (int) $_POST['cust_id']; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
```

projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
```diff
@@ -717,29 +717,31 @@ function ZeroBSCRM_get_quote_template() {
...
 	}
-	$quote_template_ID = -1;
+	$quote_template_id = -1;
 	if ( isset( $_POST['quote_type'] ) ) {
-		$quote_template_ID = (int) sanitize_text_field( $_POST['quote_type'] );
+		$quote_template_id = (int) sanitize_text_field( wp_unslash( $_POST['quote_type'] ) );
```
tbradsha:
```suggestion
		$quote_template_id = (int) $_POST['quote_type'];
```

projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
```diff
@@ -792,17 +794,16 @@ function ZeroBSCRM_get_quote_template() {
 				$quote_val = '[QUOTEVALUE]';
 			}
 			if ( empty( $quote_date ) ) {
-				$quote_date = date( 'd/m/Y', time() );
+				$quote_date = gmdate( 'Y-m-d' );
```
tbradsha: Trying to figure out what exactly this changes...both of them seem to result in a blank selector when selecting a quote template.

It seems there are three places `$quote_date` can be set...seems ripe for cleanup, but perhaps out of scope.

projects/plugins/crm/includes/jpcrm-mail-templating.php
```diff
@@ -531,6 +530,11 @@ function zeroBSCRM_quote_generateAcceptNotifHTML( $quoteID = -1, $quoteSignedBy
 				$zbs_biz_youremail = zeroBSCRM_getSetting( 'businessyouremail' );
 				$zbs_biz_yoururl = zeroBSCRM_getSetting( 'businessyoururl' );
 
+				$quote_url                      = zeroBSCRM_portal_linkObj( $quoteID, ZBS_TYPE_QUOTE ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
+				$quote_edit_url                 = jpcrm_esc_link( 'edit', $quoteID, 'zerobs_quote' ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
+				$replacements['quote-url']      = $quote_url;
+				$replacements['quote-edit-url'] = $quote_edit_url;
```
tbradsha: Could collapse into two lines if you wanted; not a blocker.

tbradsha: Also, maybe we should move these further down to where we set the other `$replacements` keys.

projects/plugins/crm/includes/jpcrm-templating-placeholders.php
```diff
@@ -1654,7 +1654,8 @@ public function replace_placeholders(
 		$string = '',
 		$replacements = array(),
 		$replacement_objects = false,
-		$retain_unset_placeholders = false
+		$retain_unset_placeholders = false,
+		$include_non_rendered_quote_keys = true
```
tbradsha: I'd rather avoid object-specific params for a generic function like this. Three alternatives to consider:

1. Somehow leverage the `$retain_unset_placeholders` to perform the same function. At a glance, it seems the purpose of this param is to not render unset placeholders, which seems to have a fair amount of overlap.
2. Generify `$include_non_rendered_quote_keys` to instead accept an array of keys to retain.
3. In a combo of 1 and 2, allow `$retain_unset_placeholders` to accept either a bool or an array of keys to retain.

If you're using the same keys over and over, you could add a wrapper function that passes the appropriate keys.

projects/plugins/crm/includes/jpcrm-templating-placeholders.php
```diff
@@ -1753,11 +1754,16 @@ public function replace_placeholders(
...
 
+					// Replace main key.
+					// In the Quote editor itself we don't want to render the Quote ID or URL placeholders,
+					// So, $include_non_rendered_quote_keys will be set to false in that case. See ZeroBSCRM_get_quote_template().
+					$non_rendered_quote_keys = array( 'quote-ID', 'quote-url' );
+					if ( $include_non_rendered_quote_keys || ( ! $include_non_rendered_quote_keys && ! in_array( $replacement_info['key'], $non_rendered_quote_keys, true ) ) ) {
+
+						$string = str_replace( $replace_string, $replace_with, $string );
 
+					}
```
tbradsha: See above comment about params.

tbradsha (dismissed):
Things seem to work. I do have some comments to consider.

Also, more generally I wonder...is there any reason to render placeholders from Quote Templates (e.g. when loading the template into a quote via `AJAX.php`)? Or would it be more reasonable to simply pass through the exact template content without rendering it, and then only render the quote content once we save an actual quote?

tbradsha:
> I've chosen to leave Quote checkbox and autonumber custom fields as 'won'tfix'

Totally fair.

projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
coder-karen: I think cleaning up when we review the date functionality generally would be good. In this case the change is probably confusing due to how I created this PR then removed changes to put in other PRs (in this case probably didn't change it back correctly). This line becomes `wp_date( get_option( 'date_format' ) )` and I see you've commented over there so I'll follow up there.

projects/plugins/crm/includes/jpcrm-mail-templating.php
coder-karen: Yes, good idea, added in d3edfc15bae81dc48adbb5cf67604636df3cc58e

projects/plugins/crm/includes/jpcrm-templating-placeholders.php
coder-karen: Thanks, yes, that makes sense. I went with option 2 here in another commit: 1f87ec0fcb9204136def853cddbcbf9be642f5b3

projects/plugins/crm/includes/jpcrm-templating-placeholders.php
coder-karen: Got it, followed up there.

coder-karen:
> Also, more generally I wonder...is there any reason to render placeholders from Quote Templates (e.g. when loading the template into a quote via AJAX.php)? Or would it be more reasonable to simply pass through the exact template content without rendering it, and then only render the quote content once we save an actual quote?

That's an interesting question. It would certainly be more consistent. It may be seen as a regression though for people who use particular quote placeholders and then all of a sudden notice what seems to be a broken experience. But maybe something worth more discussion and a follow-up issue?

tbradsha (dismissed):
Looks good!

coder-karen:
Post merging, noting I've added an extra commit to fix one 'Quote Accepted' case which wasn't working for the datetime_str and date_str versions.

tbradsha (approved):
Do it!

PR #34497, opened 2023-12-06
tools: Read `.require.php` in parallel-lint

anomiex:
## Proposed changes:

Since we're specifying `.require.php` in all packages now, let's have parallel-lint read it and automatically skip packages that require a newer version of PHP than what's running.

Also, since we're here, some additional incidental cleanups:
- Remove handling for `php7/` dirs since 7.0 is now the minimum.
- Ignore `tools/docker/data/` so local runs don't complain about it being unreadable.

### Other information:

## Jetpack product discussion

https://github.com/Automattic/jetpack/pull/34462#discussion_r1417906443

## Testing instructions:

* Is CI happy?
* Try making a PR adding use of new syntax in a project that has a `.require.php` for a corresponding version. CI still happy?

### DIFF
--- a/tools/parallel-lint.sh
+++ b/tools/parallel-lint.sh
@@ -24,16 +24,24 @@ done
 
 if $ALL; then
 	SKIPS=()
-	if php -r 'exit( PHP_VERSION_ID < 70000 ? 0 : 1 );'; then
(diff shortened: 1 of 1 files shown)


zinigor (approved):
It looks like the linter is working as before, and the code looks OK.

PR #34503, opened 2023-12-06
Subscriptions: Fix subscriptions new subscriber issue

chrisfromthelc:
Fixes https://github.com/Automattic/gold/issues/238

## Proposed changes:

This PR updates the way the subscriber box responds when a user enters in an email and the page reloads. Previously, the page would simply reload, and it would give no indication that the signup succeeded after the confirmation modal; it would appear as if they could sign up again. On Firefox, it would also put in the email address that was just entered in, further indicating that maybe something went wrong. In reality, the signup succeeded.

With these changes, the user will enter an email and hit Enter or click Subscribe. They will get the plan selection (if plans exist) or the confirmation modal popup. Upon going through those modals, the page reloads, and pulls the email address from the token that will be set during the signup. It changes the email field's placeholder to that email address, and disables the form field. Additionally, the subscribe button text switches to "Manage Subscriptions", and clicking will send the user to `https://wordpress.com/email-subscriptions` so they can manage their subscriptions.

### Other information:

## Testing instructions:

This diff should be tested on both WPCOM and Jetpack/Jurassic Ninja sites.

* Create a Subscribe block on a test site.
* Sign up and select the Free plan (if you have plans on the site). Continue through the modal(s) otherwise.
* Upon page reload, note that the email field should be disabled, and show the email you signed up with. The Subscribe button should have also changed to Manage Subscriptions and go to `https://wordpress.com/email-subscriptions` when clicked.
* Additionally, on Firefox, make sure that the email address is light gray and not black after the page reload.

**Result**
Screenshot 2023-12-06 at 3 19 02 PM

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-new-subscriber-subscribe-box
+++ b/projects/plugins/jetpack/changelog/fix-new-subscriber-subscribe-box
@@ -1,0 +1,4 @@
+Significance: minor
+Type: bugfix
+
+Keeps subscribe box visible but not usable when a subscription token is found in the browser's cookie.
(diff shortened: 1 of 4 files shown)


n3f:
@chrisfromthelc is there some overlap between this PR and https://github.com/Automattic/gold/issues/241?

PR #34510, opened 2023-12-07
Top Posts and Pages Block: Tidy Up Titles

### LINKED ISSUE #34505: Top posts: Title length limit
We should limit the length of the title and possibly wrap it too.

image

Aurorum:
Fixes #34505

## Proposed changes:

Tidies up titles for the Top Posts and Pages block:

- Ensures that they're on separate lines
- Adds wrapping onto the next line
- Limits their length 

cc @kangzj 

### Other information:

## Jetpack product discussion

See original issue.

No.

## Testing instructions:
With Beta blocks enabled, insert a Top Post & Pages block, then create some posts with really long titles. They should wrap appropriately here, and their length should be limited. Also, if you now only display the post title and category, they should remain on two different lines.

**Before:**
Screenshot 2023-12-07 at 08 32 46

**After:**
Screenshot 2023-12-07 at 08 31 43


projects/plugins/jetpack/_inc/lib/class-jetpack-top-posts-helper.php
```diff
@@ -94,7 +94,7 @@ public static function get_top_posts( $period, $items_count, $types ) {
 					'context'   => get_the_category( $post_id ) ? get_the_category( $post_id ) : get_the_tags( $post_id ),
 					'href'      => $post['href'],
 					'date'      => get_the_date( '', $post_id ),
-					'title'     => $post['title'],
+					'title'     => strlen( $post['title'] ) > 90 ? sprintf( '%s...', substr( $post['title'], 0, 90 ) ) : $post['title'],
```
Aurorum: Any preference on what an appropriate character limit might be? 90 felt reasonable, but happy to change this.

Aurorum: Actually, I do wonder if we should limit characters here. I agree that long titles don't look great in a column of three, but they're a lot more acceptable in the List View, and it's probably harder to restrict a limit to only one view without rendering two titles. 

It might just be that while long titles are inadvisable and might not look great, that should be up to the user rather than imposing an arbitrary character count? The wrapping should improve the display anyway.

Any thoughts?!

kangzj: I think you have the point! Let's do that 😄 

Just leaving a note, if we going down the limit length road, we'd probably need a title attribute for the title 😸

Aurorum: Removed the character limit :)

TIL about the title attribute - that makes sense! 

Thanks!!

kangzj (approved):
LGTM ❤️

PR #34512, opened 2023-12-07
Jetpack Boost: React migration - convert premium-tooltip

pyronaur:
## Proposed changes:
Part of the great React migration - Convert Premium Tooltip.

Introduced a new property on `ReactComponent`:

`<ReactComponent ... inline />` - this will attempt to set `display: contents` and fall-back to `display: inline` because we must create a DOM node to bind the React component to, but sometimes that doesn't fit our existing markup expectations.

### Other information:

## Jetpack product discussion
N/A

N/A

## Testing instructions:
🌈


thingalon:
> **Testing instructions:**
> 🌈

Knowing what things to click on / look at to test a PR would be helpful.

thingalon (approved):
👍 Seems to work, looks good to me.

For a future iteration, we should consider replacing all of our tooltips with Jetpack's tooltips from the storybook, but I'm not going to worry about it mid-conversion-to-React.

PR #34518, opened 2023-12-07
Contact Form: revalidate missing value as user types

monsieur-z:
## Proposed changes:

In https://github.com/Automattic/jetpack/pull/34401, we revalidated the input of a Contact Form as it lost focus in order to provide quicker feedback to a user correcting an invalid form.

This PR is a further improvement. When a field is invalid because of a missing value, we revalidate the field as the user types (in the case of simple inputs) or checks a radio button or checkbox to provide even quicker feedback.

We've limited this to the "missing value" kind of error because it's a black-or-white situation that's easy to handle. With an invalid email input, the native error message (at least on Chrome) changes as a user types an email address, which is more distracting than helpful. In these instances, we keep the revalidation on the field blur.

### Other information:

## Jetpack product discussion

Project thread: pf5801-87-p2

No.

## Testing instructions:

- Create a new post
- Add at least a text, Single Choice, and Multiple Choice fields, and mark them as required
- Open the preview
- Submit the form
- Fix the issues
- Notice the "required field" message disappears as you type or check an input

### DIFF
--- a/projects/packages/forms/changelog/update-contact-form-validation-strategy
+++ b/projects/packages/forms/changelog/update-contact-form-validation-strategy
@@ -1,0 +1,4 @@
+Significance: patch
+Type: added
+
+Contact Form: revalidate missing value as user types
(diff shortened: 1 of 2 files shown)


jeherve (approved):
Oh this is nice. 🚢 !

PR #34522, opened 2023-12-07
AI Assistant: restrict stored state value

CGastrell:
AI Assistant block should only store deterministic state values

## Proposed changes:
This PR fixes an issue with the latest change to store the state as a block attribute. Half way states are not easy to recover from without context, so we now restrict to `init` and `done` states in case of a reload in the middle of an assistant session.

### Other information:

No

## Testing instructions:
Go to the editor and use the AI Assistant block. Ask a suggestion (eg: haiku on testing), once it's done, check the Code Editor to see the `requestingState` is now `done`. Save draft and reload, you should get back at where you left. 

Now open the developer tools, Network tab, and disable the connection. Hit "Regenerate" button to get the error. Check again the Code Editor, it should still read 'done'. Re-enable the connection and reload the page, you should be once more where you left and continue to work with the block.

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-ai-state-stored-issue
+++ b/projects/plugins/jetpack/changelog/fix-ai-state-stored-issue
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
+AI Assistant: restrict stored states to init and done to prevent half way states of the block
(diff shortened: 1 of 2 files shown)


dhasilva (approved):
Nice!

PR #34523, opened 2023-12-07
Changelogger: Add New Prompt with Enquirer

sdixon194:
## Proposed changes:

A redo of https://github.com/Automattic/jetpack/pull/34058 now that we have the enquirer package.

It adds a new question when adding a jetpack changelogger to check if site admins would need to know about it. If not, it gets assigned the `other` changelog type.

### Other information:

## Jetpack product discussion

pf4qpu-9d-p2

no
## Testing instructions:

* Use `jetpack changelog add plugins/jetpack` and you should see the new prompt. If you said yes, it should then prompt you for the actual changelog type. If you say no, it should be skipped and the resulting changelog file should be "other."
* If you add a changelog to any other project, the changelogger should work per usual, without asking the question.


tools/cli/commands/changelog.js
```diff
@@ -743,6 +743,17 @@ async function promptChangelog( argv, needChangelog, types ) {
 		},
 	} );
 
+	let userFacing = false;
+	// Ask if this is a user facing change if we detect Jetpack.
+	if ( argv.project === 'plugins/jetpack' ) {
+		userFacing = await prompt( {
+			type: 'confirm',
+			name: 'userFacing',
+			message: 'Is this a Jetpack change that site admins would like to know about? (Y/n)',
```
sdixon194: What if we made this even more explicit about the end goal?

```suggestion
			message: 'Is this a change that should be mentioned in the X.Y release post? (Y/n)',
```
Or something of the sort?

tools/cli/commands/changelog.js
```diff
@@ -743,6 +743,17 @@ async function promptChangelog( argv, needChangelog, types ) {
 		},
 	} );
 
+	let userFacing = false;
+	// Ask if this is a user facing change if we detect Jetpack.
+	if ( argv.project === 'plugins/jetpack' ) {
```
anomiex: I'm pretty sure this will only trigger for `jetpack changelog add plugins/jetpack`. We should do something like this to handle when `jetpack changelog add` detects Jetpack too.
```suggestion
	if ( needChangelog.includes( 'plugins/jetpack' ) ) {
```

tools/cli/commands/changelog.js
```diff
@@ -743,6 +743,17 @@ async function promptChangelog( argv, needChangelog, types ) {
...
+		userFacing = await prompt( {
+			type: 'confirm',
+			name: 'userFacing',
+			message: 'Is this a Jetpack change that site admins would like to know about? (Y/n)',
+			default: 'true',
```
anomiex: When I was working on the other PR, I noted that for 'confirm' it renders better if you do like this instead.
```suggestion
			initial: true,
```
That lets you skip having to manually include "(Y/n)" in the message field.

tools/cli/commands/changelog.js
anomiex: Personally I find the "site admins" question a little clearer than the other, but this is the kind of question where I don't necessarily trust my own opinion.

tools/cli/commands/changelog.js
```diff
@@ -766,15 +777,22 @@ async function promptChangelog( argv, needChangelog, types ) {
...
-		type: 'autocomplete',
-		name: 'type',
-		message: 'Type of change.',
-		suggest: ( input, choices ) => choices.filter( choice => choice.value.startsWith( input ) ),
-		highlight: v => v,
-		choices: typeChoices,
-	} );
+	// Get the type, set it to other if this isn't a user facing change.
+	let typeResponse;
+	if ( ! userFacing.userFacing && argv.project === 'plugins/jetpack' ) {
```
anomiex: Same as above regarding `argv.project`. But really we should only need to check the `userFacing` answer without checking again for plugins/jetpack at all.

tools/cli/commands/changelog.js
```diff
@@ -743,6 +743,17 @@ async function promptChangelog( argv, needChangelog, types ) {
 		},
 	} );
 
+	let userFacing = false;
+	// Ask if this is a user facing change if we detect Jetpack.
+	if ( argv.project === 'plugins/jetpack' ) {
+		userFacing = await prompt( {
```
anomiex: Having to check `userFacing.userFacing` below is really kind of weird. Possibly we should do it like this:
```js
const { userFacing } = await enquirer.prompt( {
    type: 'confirm',
    name: 'userFacing',
    message: '???',
    initial: true,
    skip: ! needChangelog.includes( 'plugins/jetpack' ) ,
} );
```
As things currently are (cf. [enquirer/enquirer#340](https://href.li/?https://github.com/enquirer/enquirer/issues/340)) this will result in `userFacing` being automatically true for everything except Jetpack, while for Jetpack it'll ask and assign true or false. Then the check below can be just `if ( userFacing ) {`.

Or if you'd rather only call `prompt` for Jetpack instead of using `skip`, we could still do like
```js
( { userFacing } = await prompt( {
    // etc
} ) );
```
or
```js
userFacing = await prompt( {
    // etc
} ).userFacing;
```

tools/cli/commands/changelog.js
sdixon194: Oh duh, nice catch.

tools/cli/commands/changelog.js
sdixon194: Is this kind of what you had in mind? It wouldn't let me deconstruct the variable without returning anything other than `undefined`

2d85137ad4879e9d7429588dfe5dd9e309b921da

> Having to check userFacing.userFacing below is really kind of weird. Possibly we should do it like this:

Doesn't solve this problem though :/ Not sure what knowledge I'm missing about destructuring objects, or maybe something with how enquirer returns.

anomiex: > Doesn't solve this problem though :/ Not sure what knowledge I'm missing about destructuring objects, or maybe something with how enquirer returns.

Is this it?
* When you do `const { foo } = something;`, JS sees it as destructuring.
* When you do `{ foo } = something;`, JS thinks that the `{ foo }` is a [block statement](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/block) rather than part of a destructuring assignment, so it complains about not expecting the `=`.
* The extra parens in `( { foo } = something );` make it see it as destructuring again.

anomiex (commented):
Works for me, although I wonder if we should move the new question to after the significance question rather than before it so it's just before the type question that it relates to.

sdixon194:
> Works for me, although I wonder if we should move the new question to after the significance question rather than before it so it's just before the type question that it relates to.

I had it this way initially before moving it to try it out since it seemed a bit cleaner overall UI wise. I moved it back to the original position for now though. Let me know if that seems better.

anomiex (approved):
Makes more sense to me this way, but I'm happy to wait for more opinions if you disagree.

sdixon194:
> Makes more sense to me this way, but I'm happy to wait for more opinions if you disagree.

Nope that's where my gut initially placed it before I messed with it so let's go with this for now, can always move it around if we get feedback 👍

PR #34526, opened 2023-12-07
tools, cli: Replace inquirer with enquirer

anomiex:
## Proposed changes:

We started using enquirer in #34403. Let's finish the job.

This also makes some incidental fixes in the code:
* Remove `[bracket]` prompts where they don't matter.
* `jetpack draft`: Remove unneeded `shell` in child process spawn.
* `jetpack generate`: Don't ask about autotagger twice when selecting a non-conflicting mirror repo now.
* `jetpack rsync`: Don't prompt to create a new dest when there are no saved dests to choose.
* CLI github helper: Better invalid token handling.
* `tools/js-tools/list-gh-secrets.js`: Work correctly if cwd is not the monorepo root.

Mostly I kept selection lists as selection lists, since that has always been the behavior previously. I did change the "which project" lists to autocomplete though, as those are very long.

### Other information:

## Jetpack product discussion

Told you I might get bored and do this, Steve 😉

## Testing instructions:

* Make sure all the prompts still work well.


sdixon194:
I have no idea why, but just running `jetpack test` is failing for me after selecting a project (I can pass through like `jetpack test plugins/jetpack` just fine).

```
Error: ENOENT: no such file or directory, scandir './projects/[object Object]'
    at readdirSync (node:fs:1513:3)
    at dirs (file:///Users/stevedixon/jetpack-dev/jetpack/tools/cli/helpers/projectHelpers.js:6:2)
```

It looks like on this branch in `projectHelpers.js`, when I log out stuff in `allProjects()`, `projectTypes` is returning an array of objects, and on trunk it logs the correct projectTypes array defined in the file. I don't see anything in this PR that modifies `projectTypes` before calling allProjects, so I'm not sure why this is broken for me 🤔

sdixon194:
While we're here, for `draft.js`, should we update the message on line 36?

```
console.log(
	chalkJetpackGreen( 'You are now in draft mode. Some pre-commit and pre-push hooks are disabled, please be careful.' )
);
```

sdixon194:
Also just noting `jetpack changelog` fails when you don't pass it a command initially because it doesn't collect changelog information from `changelogAdd()` before running `changelogArgs()`, but this is also broken in trunk so can be fixed in a follow up.

anomiex:
> I don't see anything in this PR that modifies `projectTypes` before calling allProjects, so I'm not sure why this is broken for me 🤔

It seems that somewhere inside equirer it's mutating the passed array rather than adjusting a copy. We can avoid the problem by making our own copy first.

> While we're here, for draft.js, should we update the message on line 36?

To what?

sdixon194:
> To what?

Oh apologies I provided the example for what I think might be a better fit above:
```
console.log(
	chalkJetpackGreen( 'You are now in draft mode. Some pre-commit and pre-push hooks are disabled, please be careful.' )
);
```

This is what it says currently:

```
		console.log(
			chalkJetpackGreen( 'You are now in draft mode. No nags for you, but be careful.' )
		);
```

> It seems that somewhere inside equirer it's mutating the passed array rather than adjusting a copy. We can avoid the problem by making our own copy first.

Nice find, this does fix it for me!

anomiex:
Ah, ok. I didn't look and just thought that was the existing wording. That suggestion _is_ better.

sdixon194 (approved):

PR #34527, opened 2023-12-07
Fix sharing buttons block icons on simple/atomic sites

### LINKED ISSUE #34521: Sharing Block: Fix icons not saving for Simple/Atomic
> HTML that is not allowed gets stripped from the post content on save. Typically, we'd solve that by only using allowed html in the markup that is saved in the post content, and then rendering that not-allowed markup via PHP instead.
Basically in the `render_block` function, we would render specific PHP content (with the svg) based off the attributes saved with the block

Epic: #19694

andrii-lysenko:
Fixes https://github.com/Automattic/jetpack/issues/34521

## Proposed changes:

This is a follow up for https://github.com/Automattic/jetpack/pull/34414. In order to address the issue, where icons are not saved and not shown on the view on Simple/Atomic sites we are moving the markup to the server side. 
As also noted in a https://github.com/Automattic/jetpack/issues/34521#issuecomment-1846453497 we are copying SVGs from https://github.com/Automattic/social-logos.
This PR also does some refactoring and removing unused code.

### Other information:

## Testing instructions:

* Use `rsync` or the cli command in order to copy changes from this PR to your sandbox.
* Modify your hosts file to point your simple/atomic site to your sandbox
* Create a new post. As an A12C you should be able to add beta blocks. Add Sharing Buttons block and individual buttons inside
* Save the post and view the results. There should be identical behavior to self hosted site like in the original pr: https://github.com/Automattic/jetpack/pull/34181


projects/plugins/jetpack/extensions/blocks/sharing-button/class-sharing-source-block.php
```diff
@@ -53,6 +53,24 @@ public function get_id() {
 		return $this->id;
 	}
 
+	/**
+	 * Get unique sharing ID. Similar to get_id().
+	 *
+	 * @return int
+	 */
+	public function get_class() {
```
monsieur-z, line 61: The method name is a bit misleading.

monsieur-z (approved):
I tested with a simple site and could now see the icons 👍

projects/plugins/jetpack/extensions/blocks/sharing-button/class-sharing-source-block.php
andrii-lysenko: Yeah, I've copied this one from the original Sharing Sources file, just to be on a safe side. I see it could be gone soon.

PR #34529, opened 2023-12-08
[Boost] Migrate Speed Score popups to React

thingalon:
Port the Speed Score popout to React.

A few notes:
- Adds `react-transition-group` to handle transitioning the popout in / out from the side.
- Alters the popout logic to always render the last message shown so that the popout can transition out in a react-friendly manner
- Switches from using a link for the dismiss option, to using a button for a11y compliance.

## Proposed changes:
* Port the Speed Score popout to React

### Other information:

## Jetpack product discussion

pc9hqz-2jy-p2

no

## Testing instructions:

* Make sure you clear any previously popup dismissals, to make sure this appears; `delete from wp_options where option_name='jetpack_boost_ds_dismissed_score_prompt' or option_name='jb_show_score_prompt';`
* Force your speed score to show a large increase, by adding this to `Speed_Score::fetch_speed_score`:
```
return [
	'status' => 'success',
	'scores' => [
		'current' => [
			'desktop' => 100,
			'mobile' => 100,
		],
		'noBoost' => [
			'desktop' => 50,
			'mobile' => 50,
		],
		'isStale' => false,
	]
];
```
* Load the Boost dashboard, make sure the score increase popup appears
* Close the popup with the X button and refresh the score, make sure the popup appears again
* Make sure that clicking either the rate button or the dismiss link makes the dialog go away (and not return even on refresh)
* Do the same for score decrease


projects/plugins/boost/app/assets/src/js/features/speed-score/pop-out/pop-out.tsx
```diff
@@ -0,0 +1,152 @@
...
+					>
+						{ message && message.cta }
+					</a>
+
+					<button
+						className={ classNames( 'jb-link', styles.dismissLink ) }
+						onClick={ handleDismiss }
+					>
+						{ __( 'Do not show me again', 'jetpack-boost' ) }
+					</button>
```
haqadn: Should we start using the RNA button component for these, now that we are in React?

projects/plugins/boost/app/assets/src/js/features/speed-score/pop-out/pop-out.tsx
```diff
@@ -0,0 +1,152 @@
...
+		dismiss();
+		onClose();
+	}
+
+	return (
+		<div id="parent" className={ styles.wrapper }>
+			<CSSTransition
+				in={ hasScoreChanged && ! isDismissed }
+				appear={ true }
+				classNames={ { ...styles } }
```
haqadn, line 123: Does that mean all the CSS classes will be applied to this component, even headline, card, etc?

haqadn (changes requested):
Tested and looks nice. My only suggestion is to use the Button components.

projects/plugins/boost/app/assets/src/js/features/speed-score/pop-out/pop-out.tsx
thingalon: Done in 413a7d91c0. Thanks for the tip!

projects/plugins/boost/app/assets/src/js/features/speed-score/pop-out/pop-out.tsx
thingalon: No, they won't all be used. `CSSTransition` expects an object where keys include things like this to define css classes for different parts of the transition:
 
```
classNames={{
 appear: 'my-appear',
 appearActive: 'my-active-appear',
 appearDone: 'my-done-appear',
 enter: 'my-enter',
 enterActive: 'my-active-enter',
 enterDone: 'my-done-enter',
 exit: 'my-exit',
 exitActive: 'my-active-exit',
 exitDone: 'my-done-exit',
}}
```

[Their docs ](https://reactcommunity.org/react-transition-group/css-transition) recommend just sending the whole style object when using a css module, in which case only styles in the module scss file named things like `.appearDone` and others from the above list will be used.

projects/plugins/boost/app/assets/src/js/features/speed-score/pop-out/pop-out.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+					<a
+						className="jb-button--primary"
+						href={ message && message.ctaLink }
+						target="_blank"
+						rel="noreferrer"
+						onClick={ handleDismiss }
+					>
+						{ message && message.cta }
+					</a>
```
haqadn, line 140: This can also be a `Button`. For full usage info, I recommend running storybook and seeing the example.

projects/plugins/boost/app/assets/src/js/features/speed-score/pop-out/pop-out.module.scss
```diff
@@ -0,0 +1,39 @@
...
+ * Styles applied by CSSTransition when the component is fully visible.
+ */
+.appearDone, .enterDone {
+	right: 0%;
+}
+
+.dismissLink {
+	margin-top: 10px !important;
+	float: right;
+}
```
haqadn: Just noticed. We have a convention of using kebab case in CSS. I think it's probably ok to skip it for CSSTransition. However, I feel more comfortable with consistent case given that it's actually possible.

The `dismissLink` being not part of the CSSTransition, should be changed to kebab case.

projects/plugins/boost/app/assets/src/js/features/ReactComponent.svelte
```diff
@@ -7,11 +7,13 @@
 	let root;
 	export let inline = false;
 	afterUpdate( () => {
-		// Remove `inline` from the props passed to the component.
-		// eslint-disable-next-line @typescript-eslint/no-unused-vars, no-shadow
-		const { this: component, children, inline, ...otherProps } = $$props;
-		root = WPElement.createRoot( container );
-		root.render( React.createElement( component, otherProps, children ) );
+		const { this: component, children, inline, ...props } = $$props;
```
haqadn: I may have messed inline while trying to merge trunk. Please take a look at that as well.

haqadn (changes requested):
There was a missed Button and a couple more nitpicks

projects/plugins/boost/app/assets/src/js/features/speed-score/pop-out/pop-out.tsx
thingalon: From an accessibility perspective, it's a link. You click it to open a webpage.

It's kind of funny. We have a link to a website which we deliberately style to look like a button, and a button which we deliberately style to look like a link.

But functionally, the button-which-looks-like-a-link is used to interact with the page you're on - it's a control for what the site should show you. The link-that-looks-like-a-button takes you to an external website.

projects/plugins/boost/app/assets/src/js/features/speed-score/pop-out/pop-out.module.scss
thingalon: I've removed the camel case in c0e9305aaf, for the sake of this PR.

However, I would like to raise moving away from kebab-case in module scss files because array-style accessors (`style['dismiss-link']`) are harder to read.

thingalon: > However, I would like to raise moving away from kebab-case in module scss files because array-style accessors (style['dismiss-link']) are harder to read.

I just searched the Jetpack Monorepo, and there are by far more instances of `styles[ 'kebab-case' ]` than `styles.camelCase` so I'll comply with the established convention. Consistency over personal preference.

Complying with `kebab-case` in 37e6061b6e.

projects/plugins/boost/app/assets/src/js/features/ReactComponent.svelte
thingalon: Thanks for raising this. It looks like the only change you undid through the merge was @pyronaur renamed `props` to `otherProps` for clarity. I've re-established that.

However, the linter picked up that `inline` inside `afterUpdate` is shadowing `inline` from the parent context, so I have cleaned that up too by destructuring `inline` into `inlineProp` and copying it to the `inline` variable in code.

Either way, this is temp code. So long as it doesn't cause problems during the remainder of the migration we'll be ok.

32485c45c3

projects/plugins/boost/app/assets/src/js/features/speed-score/pop-out/pop-out.module.scss
haqadn: I like camelCase for the same convenience too. :(

projects/plugins/boost/app/assets/src/js/features/ReactComponent.svelte
haqadn: Thank you

haqadn (approved):

PR #34534, opened 2023-12-08
Boost: Refactor image CDN quality settings into React

haqadn:
Fixes #

## Proposed changes:
* Refactor image CDN quality settings component into react

### Other information:

## Testing instructions:
* Upgrade to premium Boost
* Play with the image CDN quality settings to make sure they are working correctly

image


projects/plugins/boost/app/assets/src/js/features/image-cdn/quality-settings/quality-settings.tsx
```diff
@@ -0,0 +1,123 @@
...
+			__( `For more control over image quality, <link>upgrade now!</link>`, 'jetpack-boost' ),
+			{
+				link: (
+					// eslint-disable-next-line jsx-a11y/anchor-has-content, jsx-a11y/anchor-is-valid
+					<a
+						href="#"
+						onClick={ () => {
+							navigate( 'upgrade' );
+						} }
+					/>
```
thingalon: This upgrade link doesn't work. When you click it, it just loads `href="#"`. To make it work, you'd need to take an `event` parameter to your click handler and call `event.preventDefault()`.

Instead, why not... 
```suggestion
					<a href="#upgrade" />
```

If we do have to use a click handler / call navigate gracefully, then we might want a helper function in `$lib/utils/navigate` which returns an anchor tag ready to go with the right props for use in createInterpolateElement to navigate to another page.

thingalon (changes requested):
Great work! This works well and the code is clean.

I've just found one bug.

projects/plugins/boost/app/assets/src/js/features/image-cdn/quality-settings/quality-settings.tsx
thingalon: Lovely - NavigationLink component does the job nicely, fixes the bug and is reusable. 🥇

thingalon (approved):
Nice one; looks clean and works well

PR #34543, opened 2023-12-08
Paywall: fix pending subscription check

### LINKED ISSUE #34540: Subscriber wall: subscriber with uncofirmed email doesn't get acces after login
Use a subscriber which hasn’t confirmed their email, and then use the “already subscriber?” flow to get access to a walled post like https://cicerosletters.wordpress.com/2023/12/08/free-subscribers-only/

Email confirmed, works:

https://github.com/Automattic/jetpack/assets/87168/16941f03-cb82-4921-81d3-8e5d6fb3281c

Email unconfirmed, doesn't work:

https://github.com/Automattic/jetpack/assets/87168/95ccd6a4-ea0f-449c-806a-b6c15e9ef308

They should see the "verify your email" thing instead:

image

For this user, `$this->get_token_property( 'blog_sub' )` is returning "pending"

https://github.com/Automattic/jetpack/blob/adb1d33cff25f2de39b2073d5aaf7884f1e82a15/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php#L158-L165

lezama:
Fixes #34540

## Proposed changes:

Fix paywall status when user hasn’t confirmed their email but used “already subscriber?” flow to log in.

### Other information:

## Testing instructions:

* In a new session go to a subscribers only post in a WPCOM site.
* Subscribe with a new user email but don't confirm the email
* Restart the session and log in via the “already subscriber?” in the post
* Make sure you get:

Screenshot 2023-12-11 at 14 54 09

* Check the subscription flow in Jetpack sites hasn't been affected.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
+++ b/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
@@ -165,6 +165,15 @@ abstract class Token_Subscription_Service implements Subscription_Service {
 	}
 
 	/**
+	 * Returns true if the current authenticated user has a pending subscription to the current site.
+	 *
+	 * @return boolean
+	 */
+	public function is_current_user_pending_subscriber() {
+		return $this->get_token_property( 'blog_sub' ) === 'pending';
+	}
+
+	/**
 	 * Return if the user has access to the content depending on the access level and the user rights
 	 *
 	 * @param string $access_level Post or blog access level.
--- a/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
+++ b/projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
@@ -170,7 +170,7 @@ abstract class Token_Subscription_Service implements Subscription_Service {
 	 * @return boolean
 	 */
 	public function is_current_user_pending_subscriber() {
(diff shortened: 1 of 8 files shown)


simison (dismissed):
With combination of D130841-code this tests well.

simison (approved):

projects/plugins/jetpack/extensions/blocks/premium-content/_inc/subscription-service/class-token-subscription-service.php
```diff
@@ -164,6 +164,15 @@ public function is_current_user_subscribed() {
 		return $this->get_token_property( 'blog_sub' ) === 'active';
 	}
 
+	/**
+	 * Returns true if the current authenticated user has a pending subscription to the current site.
+	 *
+	 * @return boolean
+	 */
+	public function is_current_user_pending_subscriber() {
+		return false;
```
millerf, line 173: I believe a user using this service could be pending, no?
Something like
```
$this->get_token_property( 'blog_sub' ) === 'pending'
```

If not, that definitely needs a comment to clarify...

lezama: this class is abstract, most services that implement this class have their implementation of this method

millerf: Oh. I missed that. It should be said in the filename I think... Thanks.

PR #34544, opened 2023-12-08
Social: Use core endpoint for dismissing notices

gmjuhasz:
This change updates our notice dismissal logic, so it uses `register_setting` and the core endpoint coming with it.
We have to support the deprecated Jetpack endpoint, so I flagged that with the current versions so we can get rid of it 3 versions later.

## Proposed changes:

* Added the handler class + tests
* Changed the initial states so they use the new getter

### Other information:

## Jetpack product discussion

https://github.com/orgs/Automattic/projects/733/views/2?pane=issue&itemId=46761193

No.
## Testing instructions:

* Delete the dismissed_notices with `jetpack docker wp option delete jetpack_social_dismissed_notices`
* With auto-conversion being on, create a new post and upload a 10Mb jpg image which should be invalid but convertible. You should see the notice that it will be converted.
* Click `Got it`, then refresh the page and select the image again. You should not see the notice anymore

Remove the advanced plan.
* Go to the editor and dismiss the advancedPlan upgrade nudge
* Make sure both are stored correctly in the option

`jetpack docker wp option get jetpack_social_dismissed_notices`
You can also check it from the console: https://developer.wordpress.com/docs/api/console/
Run a query like this with your site ID:
CleanShot 2024-01-09 at 11 37 16 png


projects/packages/publicize/src/jetpack-social-settings/class-dismissed-notices.php
```diff
@@ -0,0 +1,85 @@
...
+			return $updated;
+		}
+
+		$notice            = $value['notice'];
+		$reappearance_time = $value['reappearance_time'];
+		$dismissed_notices = get_option( self::DISMISSED_NOTICES_OPTION );
+
+		if ( ! is_array( $dismissed_notices ) ) {
+			$dismissed_notices = array();
+		}
```
manzoorwanijk: Why not this?

```suggestion
		$dismissed_notices = self:: get_dismissed_notices();
```

projects/plugins/jetpack/_inc/lib/admin-pages/class-jetpack-redux-state-helper.php
```diff
@@ -147,7 +148,7 @@ public static function get_initial_state() {
 			'connectUrl'                  => false == $current_user_data['isConnected'] // phpcs:ignore Universal.Operators.StrictComparisons.LooseEqual
 				? Jetpack::init()->build_connect_url( true, false, false )
 				: '',
-			'dismissedNotices'            => self::get_dismissed_jetpack_notices(),
```
manzoorwanijk: Should the `get_dismissed_jetpack_notices` method be removed because it's not used anywhere else?

Or may be add a deprecation notice?

projects/plugins/jetpack/class.jetpack-gutenberg.php
```diff
@@ -748,7 +749,7 @@ public static function enqueue_block_editor_assets() {
 				'isEnhancedPublishingEnabled'     => $publicize->has_enhanced_publishing_feature(),
 				'isSocialImageGeneratorAvailable' => $settings['socialImageGeneratorSettings']['available'],
 				'isSocialImageGeneratorEnabled'   => $settings['socialImageGeneratorSettings']['enabled'],
-				'dismissedNotices'                => $publicize->get_dismissed_notices(),
```
manzoorwanijk, line 770: Should `get_dismissed_notices` method be deprecated?

manzoorwanijk (dismissed):
This works well. I just have some optional code suggestions/questions.

projects/plugins/jetpack/class.jetpack-gutenberg.php
gmjuhasz: Done

projects/plugins/jetpack/_inc/lib/admin-pages/class-jetpack-redux-state-helper.php
gmjuhasz: Done

projects/packages/publicize/src/jetpack-social-settings/class-dismissed-notices.php
gmjuhasz: Good point, done! I left in the if as a guard clause

projects/js-packages/publicize-components/src/components/form/validation-notice.tsx
```diff
@@ -6,13 +6,15 @@ import Notice from '../notice';
...
+	shouldAutoConvert: boolean;
 };
 
 export const ValidationNotice: React.FC< ValidationNoticeProps > = ( {
 	connectionsCount,
 	invalidConnectionIdsCount,
+	shouldAutoConvert,
 } ) => {
-	return (
+	return shouldAutoConvert ? null : (
```
manzoorwanijk: Thank you for making this change.

manzoorwanijk (dismissed):
Thank you for making the suggested changes. This looks good now.

manzoorwanijk (dismissed):

projects/packages/publicize/src/class-publicize-setup.php
```diff
@@ -42,12 +42,21 @@ public static function on_jetpack_feature_publicize_enabled() {
...
+		add_action( 'rest_api_init', array( static::class, 'register_core_options' ) );
+		add_action( 'admin_init', array( static::class, 'register_core_options' ) );
 
 		( new Social_Image_Generator\Setup() )->init();
 	}
 
+	/**
+	 * Registers the core options for the Publicize package.
+	 */
+	public static function register_core_options() {
```
spsiddarthan, line 59: If `register_core_options` is the callback that should get called for both `register_core_options` for both `admin_init` and `rest_api_init` hooks, do we perhaps need just one hook?

projects/packages/publicize/src/class-publicize-setup.php
```diff
@@ -42,12 +42,21 @@ public static function on_jetpack_feature_publicize_enabled() {
 		add_action( 'rest_api_init', array( new Connections_Post_Field(), 'register_fields' ), 5 );
 		add_action( 'rest_api_init', array( new REST_Controller(), 'register_rest_routes' ) );
 		add_action( 'current_screen', array( static::class, 'init_sharing_limits' ) );
-		add_action( 'rest_api_init', array( new Jetpack_Social_Settings\Settings(), 'register_settings' ) );
-		add_action( 'admin_init', array( new Jetpack_Social_Settings\Settings(), 'register_settings' ) );
+
+		add_action( 'rest_api_init', array( static::class, 'register_core_options' ) );
```
spsiddarthan, line 46: I may be wrong about this, but if we are getting and updating options on simple sites, won't one site be overwriting another site's options since update_option and get_option will run in the context of WPCOM? Perhaps, `update_blog_option` is what we need there. 

Do we even need to register these options on simple sites since we are not showing them in https://github.com/Automattic/jetpack/pull/34665/files?

projects/js-packages/publicize-components/src/components/form/validation-notice.tsx
manzoorwanijk: This change was cherry picked in #34730 as a quick fix for the issue

projects/packages/publicize/src/class-publicize-setup.php
gmjuhasz: We discussed this on a call together, but just noting it here too,[ you need both](https://developer.wordpress.org/reference/functions/register_setting/#comment-3225) if you want it on the API too.

projects/packages/publicize/src/class-publicize-setup.php
gmjuhasz: As we tested, `update_option` on wpcom context also runs on the current blog, so it will essentially be an `update_blog_option` for that blog

spsiddarthan (commented):
This tests well, Geri! Requesting @pablinos  to glance over it as well to make sure there is nothing glaring.

spsiddarthan (dismissed):

projects/packages/publicize/src/jetpack-social-settings/class-dismissed-notices.php
```diff
@@ -62,8 +55,8 @@ public function update_dismissed_notices( $updated, $name, $value ) {
 			return $updated;
 		}
 
-		$notice            = $value['notice'];
-		$reappearance_time = $value['reappearance_time'];
+		$notice            = key( $value );
```
spsiddarthan, line 58: Am I correct in saying that `$updated` is `true` or `false`, `$name` is the option name and `$value` is ['instagram' => 0]?

If that's the case, `$notice` would be `instagram` .

`reappearance_time` would be `$value['notice']` which is 0.

I am bit confused by this change, could you please elaborate why we would need this?

gmjuhasz: Sure!

The first part of the change (https://github.com/Automattic/jetpack/pull/34544/commits/f7eb8bc35e764c2323077870717dad627101a459#diff-b9af7ea1c5d7705e15c3aa19241ee6e3f3b8090af8977a9951def8e88907d2d9R31) is to fix the schema, as the previous one would describe our format poorly. We want the option to store an array with key->value pairs, where the key is the name of the notice, and the value is a numeric value, which is the reappearance time.

The second part (https://github.com/Automattic/jetpack/pull/34544/commits/f7eb8bc35e764c2323077870717dad627101a459#diff-5f1e74f983a6c97e6afd793e071e2b976f2f14616174253e77739c59e07de12eR63) was to modify the UI to send the data in this format. That means that for Instagram, it would send `jerpack_social_dismissed_notices: { Instagram: 0}`, which means that we want to update the registered setting `jerpack_social_dismissed_notices`, with the value of `{ Instagram: 0}`.

The third part of the change is where your comment lies, where value would be as you said, and the notice variable stores the name, and the reappearance_time is the number. Then we add it to the array `		$dismissed_notices[ $notice ] = $reappearance_time;` and update the option

gmjuhasz: Testing it with the API console as in the testing instructions should show if it's working well

spsiddarthan (approved):

PR #34548, opened 2023-12-08
Fix: Add Missing Sync Posts Sanity Check

chrisbliss18:
A conditional in `projects/packages/sync/src/modules/class-posts.php` didn't have a variable sanity check. This resulted in failed builds. This PR adds the missing check.

Relevant discussion about the issue here: p1702050275703559-slack-C05PV073SG3

## Testing instructions:

* None required.

No


anomiex (approved):
Fixes the PHP warning and should fix the wpcom tests. I don't know enough about this to say whether this is correct or if it should be `! $sync_module ||` instead of `$sync_module &&`.

PR #34551, opened 2023-12-08
Launchpad: Add sensei setup task

valterlorran:
Related to: https://github.com/Automattic/jetpack/pull/34320
More context: paYKcK-3Hy-p2

## Proposed changes:

* As part of the migration from the Site Setup Checklist to the Launchpad, we need to reproduce the logic of the legacy task so we can enable the Launchpad to be used for new sites only. 
* This PR adds the Sensei Setup task to the Legacy Site Setup checklist and also the visibility logic.
* The completion logic will be tackled in another PR.

### Other information:

## Testing instructions:

* Apply this PR to your sandbox using the command below.
```
bin/jetpack-downloader test jetpack-mu-wpcom-plugin add/launchpad-sensei-setup-task
```
* Create a new site through the `/setup/free` flow, and then add the Business plan
* Once you reach the Customer Home, go to the Developer Console and make a `GET` request to `wpcom/v2` -> ` /sites/:siteSlug/launchpad?checklist_slug=legacy-site-setup`
* Make sure the `Finish Sensei setup` task is not shown

Screen Shot 2023-12-08 at 16 07 03

* Now, add your site to the WoA Developer list.
* Then, install the Sensei LMS plugin, which should change your site to Atomic.
* Use the Jetpack Beta plugin to apply this PR to your site(You may have to updated the files manually, as the plugin doesn't seem to work on my end)
* Go back to the Developer Console and make the request again. Now you should see the `Finish Sensei setup` task in the response.
Screen Shot 2023-12-08 at 16 20 26

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -643,6 +643,16 @@ function wpcom_launchpad_get_task_definitions() {
 				return site_url( '/wp-admin/admin.php?page=wc-admin' );
 			},
 		),
+		'sensei_setup'                       => array(
+			'get_title'            => function () {
+				return __( 'Finish Sensei setup', 'jetpack-mu-wpcom' );
+			},
+			'is_complete_callback' => 'wpcom_launchpad_is_task_option_completed',
+			'is_visible_callback'  => 'wpcom_launchpad_is_sensei_setup_visible',
+			'get_calypso_path'     => function () {
+				return site_url( '/wp-admin/admin.php?page=sensei' );
+			},
+		),
 	);
 
 	$extended_task_definitions = apply_filters( 'wpcom_launchpad_extended_task_definitions', array() );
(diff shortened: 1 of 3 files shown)


markbiek (dismissed):
Working as expected!

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -977,6 +987,21 @@ function wpcom_launchpad_is_woocommerce_setup_visible() {
...
+ *
+ * @return bool True if the sensei setup task should be visible.
+ */
+function wpcom_launchpad_is_sensei_setup_visible() {
+	$is_atomic_site = ( new Automattic\Jetpack\Status\Host() )->is_woa_site();
+	if ( ! $is_atomic_site ) {
+		return false;
+	}
+
+	$active_plugins = get_option( 'active_plugins' );
```
markbiek: Would it be a little cleaner to use `is_plugin_active` instead? Not sure if that will work in this context.

valterlorran: In this case, the option actually returns all the active plugins. Perhaps it would be clearer if we did something like:
```suggestion
	$active_plugins = get_option( 'active_plugins' );
	$is_sensei_active = in_array( 'sensei-lms/sensei-lms.php', $active_plugins, true );
	return $is_sensei_active;
```
WDYT?

markbiek: What I'm wondering is, can we do this?

```
return is_plugin_active( 'sensei-lms/sensei-lms.php' ); 
```

Then we can avoid calling `get_option` entirely. Again, I'm not sure if this is possible but i was curious because it would

valterlorran: Oh, great! I didn't know about this function, but I will try to use it.

markbiek (approved):
LGTM!

PR #34555, opened 2023-12-08
AI Usage bar: change color for over soft limit

### LINKED ISSUE #34319: AI Tiered Plans: use yellow on the usage bar when the site is over limit but tier has soft limit
On the last tier we have a soft limit instead of a hard one, so to signal these we could use yellow on the usage bar instead of red. The user can keep using it, they knows it's over, but the color signal some difference.

CGastrell:
When highest tier is over limit we treat it as a soft limit, yet the color is the same as when it requires an upgrade.

Fixes #34319 

## Proposed changes:
This PR adds an extra prop to know if the site requires an upgrade and uses a yellowish color for the soft limit. The usual red for required upgrades remain.

image

### Other information:

No

## Testing instructions:
* Go to the block editor of your test site and add an AI Assistant block
* Dispatch the following action to set the site as being over limit but not require an upgrade

```
wp.data.dispatch( 'wordpress-com/plans' ).storeAiAssistantFeature( {
    "hasFeature": true,
    "isOverLimit": true,
    "requestsCount": 20,
    "requestsLimit": 20,
    "requireUpgrade": false,
    "upgradeType": "default",
    "usagePeriod": {
        "currentStart": null,
        "nextStart": null,
        "requestsCount": 101
    },
    "currentTier": {
        "value": 100,
	"limit": 100,
    },
    "nextTier": {
        "slug": "jetpack_ai_yearly",
        "limit": 999999999,
        "value": 1,
        "readable-limit": "Unlimited"
    },
   "tierPlansEnabled": true,
} );
```

- see the usage panel on the side, it should be yellow

Change on the above payload `requireUpgrade` value to switch to different cases


lhkowalski (approved):
The code looks good and tests as expected.

_The color is still read for the free plan limit:_

Screenshot 2023-12-11 at 10 53 51

---

_The color is still red for the common tiers:_

Screenshot 2023-12-11 at 10 55 42

---

_The color is now yellow for the 1000 requests plan (the soft limit one):_

Screenshot 2023-12-11 at 10 56 08

---

_And it's stays yellow after the limit, as expected:_

Screenshot 2023-12-11 at 10 56 21

Issue #32937, opened 2023-09-07
Feature request: add a new Top Posts & Pages Block


jp-imagines:
### What

I would love to see a new block to automatically display a certain number of "top posts" by views or likes, perhaps over a certain time period (or possibly additional sorting options for the Query Loop to accomplish the same thing).

Initial request in 6784524-zd-a8c:
> Do you have anything for most popular posts or trending posts? I didn't see that option in the query loop.

### Why

This was previously possible with the legacy Jetpack "Top Posts & Page" widget. See pdf5j4-4C-p2 regarding the decision to leave that widget "as is." For any site using a Site Editing theme, that widget is inaccessible, and similar functionality would otherwise require a plugin.

### How

```[tasklist]
### Tasks
- [ ] https://github.com/Automattic/jetpack/pull/34153
- [x] Deprecate (i.e. hide) [the existing widget](https://github.com/Automattic/jetpack/blob/848a071ff7904dd0ed680c5df21cb1ffef580f06/projects/plugins/jetpack/modules/widgets/top-posts.php) in favor of the block, with `widget_types_to_hide_from_legacy_widget_block`
- [ ] https://github.com/Automattic/jetpack/issues/34484
- [ ] https://github.com/Automattic/jetpack/pull/34479
- [ ] https://github.com/Automattic/jetpack/issues/34505
- [ ] https://github.com/Automattic/jetpack/issues/34506
- [ ] https://github.com/Automattic/jetpack/pull/34558
- [ ] https://github.com/Automattic/jetpack/pull/34717
- [ ] https://github.com/Automattic/jetpack/pull/34614
- [x] Release the block - move from beta to production
```

cuemarie:
📌 ACTIONS
- Feature request kept
- CC: @jeherve perhaps this is something a Jetpack Block could do!

jeherve:
> perhaps this is something a Jetpack Block could do!

Yeah, definitely! I'm surprised this issue doesn't exist yet!

We already have a [Top Posts & Pages widget](https://jetpack.com/support/extra-sidebar-widgets/top-posts-pages-widget/), we should turn it into a block.

jamiepalatnik:
Just wanted to add a +1 to this based on my own user interactions. This also sometimes comes up because users are confused by the Tenaz theme demo content, which has a Trending section but no way to automatically feature trending posts: https://wordpress.com/theme/tenaz

devNigel:
+1 

[...]

PR #34557, opened 2023-12-09
Gutenberg: Add Blog Stats Block

### LINKED ISSUE #17588: Stats: add option to display views for each post on the frontend
#### Is your feature request related to a problem? Please describe.
Would like to be able to show the number of times a post has been viewed since it was published on all site posts, using the data from Jetpack stats. 

#### Describe the solution you'd like
Would like to be able to turn on/off the feature to show post counts on individual posts and post archive pages.

#### Describe alternatives you've considered
There are some plugins available that do this, but they don't pull their data from Jetpack stats, so the post counts start from the point the plugin was activated. 

#### Additional context
Many of these plugins require a shortcode be added to every post, which is very cumbersome if the site has a large number of posts.

Aurorum:
Fixes #17588

## Proposed changes:
Converts the Blog Stats legacy widget into a block so that it can be used with FSE. This is broadly just a 1:1 convert of a pretty simple (albeit popular) widget, but I've added a feature whereby users can display views for a single item rather than their entire site.

Screenshot 2023-12-09 at 19 09 22

## Jetpack product discussion

See the original issue and https://github.com/Automattic/wp-calypso/issues/63584#issuecomment-1845865418

No - it uses Stats data, but in the same way as the Legacy Widget does. 

## Testing instructions:

You'll need to enable Beta blocks for this. When you insert the Blog Stats block, you should be able to insert text that follows the view count, similar to the widget. I thought that it was a nice touch to display the actual view count where possible, but "12,345" should appear as a placeholder when there is no Post ID (eg. widgets/FSE templates). The correct view count should appear when on the front end, so you might want to confirm this happens. 

Screenshot 2023-12-09 at 22 34 38

Also, you'll want to verify that you're encouraged to active the Stats module if it's disabled, and that setting successfully works. I felt it was best not to share the code with the Top Posts block here because it's not that much, and I've also stressed that views are only counted when the module is active, which feels more relevant to this block. 

For context, this is the legacy widget it replaces - it should be a fairly 1:1 conversion, besides the extra feature to display stats for an individual post:
Screenshot 2023-12-09 at 19 13 51

@kangzj super sorry to ping you for another stats block so soon - if you do fancy taking a look at this one (and totally understand if not given you've just reviewed the Top Posts and Pages block!), this one should be 10x simpler :)


projects/plugins/jetpack/modules/widgets/blog-stats.php
```diff
@@ -99,7 +111,7 @@ public function form( $instance ) {
 			<label for="<?php echo esc_attr( $this->get_field_id( 'hits' ) ); ?>"><?php esc_html_e( 'Pageview Description:', 'jetpack' ); ?></label>
 			<input class="widefat" id="<?php echo esc_attr( $this->get_field_id( 'hits' ) ); ?>" name="<?php echo esc_attr( $this->get_field_name( 'hits' ) ); ?>" type="text" value="<?php echo esc_attr( $instance['hits'] ); ?>" />
 		</p>
-		<p><?php esc_html_e( 'Hit counter is delayed by up to 60 seconds.', 'jetpack' ); ?></p>
+		<p><?php esc_html_e( 'Hit counter is delayed by up to 5 minutes.', 'jetpack' ); ?></p>
```
Aurorum, line 114: The caching time has increased since the widget was released (I think it happened when moving functions to the Stats package)

https://github.com/Automattic/jetpack/blob/57221b8fc8d1dcd09b00d956b873c4464205eb8f/projects/packages/stats/src/class-wpcom-stats.php#L31-L36

kangzj:
Wow another productive week uh @Aurorum 🤩  Nice one! I'll take a spin this week. Thank you so much for your contribution ❤️

Aurorum:
Thanks so much! No rush at all. :)

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-blog-stats.php
```diff
@@ -0,0 +1,81 @@
...
+	 * Get the blog stats.
+	 *
+	 * @param \WP_REST_Request $request Request object.
+	 *
+	 * @return array Blog stats.
+	 */
+	public function get_blog_stats( $request ) {
+		$post_id   = $request->get_param( 'post_id' );
+		$post_data = convert_stats_array_to_object(
+			( new WPCOM_Stats() )->get_post_views( $post_id, array( 'fields' => 'views' ) )
```
jeherve, line 60: The one problem we'll face with this implementation is that this will blow up folks' options table when they use the block in a page template. 

Every time we fetch post views for a post, we cache the data in a transient. If we start regularly fetching views for all posts on a site, this will make for a lot of transients. I'm not sure we'd want that to happen.

cc @Automattic/red for their opinion on this.

Aurorum: Ah, that's a very good point - I didn't think of that! What do you think about batching the post views together in a single transient? Something like this:

```php
/**
 * Fetches stats data from WPCOM or local Cache. Caches locally for 5 minutes.
 *
 * Unlike the above function, this caches post data together. It's intended to
 * prevent wp_options from blowing up when retrieving views for a large number
 * of posts at the same time. However, the final response is the same as above.
 *
 * @param array $args Optional query parameters.
 * @param int   $post_id Post ID to acquire stats for.
 *
 * @return array|WP_Error
 */
protected function fetch_post_stats( $args = array(), $post_id ) {
	$endpoint       = $this->build_endpoint();
	$transient_name = self::STATS_CACHE_TRANSIENT_PREFIX . 'batched_posts';
	$stats_cache    = get_transient( $transient_name );

	if ( ! $stats_cache ) {
		$stats_cache = array();
		$expiration  = self::STATS_CACHE_EXPIRATION_IN_MINUTES * MINUTE_IN_SECONDS;
	} else {
		$expiration = get_option( '_transient_timeout_' . $transient_name ) - time();
		$data       = $stats_cache[ $post_id ];

		if ( isset( $data ) ) {
			if ( is_wp_error( $data ) ) {
				return $data;
			}

			return array_merge(
				array( 'cached_at' => time() - $expiration ),
				(array) json_decode( $data, true )
			);
		}
	}

	$wpcom_stats = $this->fetch_remote_stats( $endpoint, $args );

	$stats_cache[ $post_id ] = is_wp_error( $wpcom_stats ) ? $wpcom_stats : wp_json_encode( $wpcom_stats );

	set_transient( $transient_name, $stats_cache, $expiration );

	return $wpcom_stats;
}
```

The problem that I can envisage is the transient holding a lot of information which slows things down - but since it resets every five minutes, I don't think that should be a big issue. The other downside though is that not all posts will be cached for five minutes (some will be cached for less), but I still think that's better than no caching at all. 

If you have other ideas, please let me know!

jeherve: I'm afraid this would be problematic on sites with a lot of posts. 

I'll let folks from @Automattic/red chime in on this. Unfortunately this is a tricky problem.

kangzj: Thanks for working on this! It looks amazing @Aurorum 

Thanks for your feedback @jeherve 

This is a tricky indeed. Additionally, the block could possibly have impact on user's page performance, as the data needs to be refreshed every 5 mins. So we really want to be really cautious about it.

Could we cache in post meta table? Would that potentially blow it up as well?

Another thought is we could go totally async calls to API without any cache - frontend loads the views from an API which doesn't cache data on the site.

looping in more crew friends for advice @Automattic/jetpack-crew 😄

Aurorum: Thanks so much @kangzj! Those both sound like viable alternatives to me. 

Let me know if anyone has a preference, and I’ll try to include it here.

jeherve: > we could go totally async calls to API without any cache - frontend loads the views from an API which doesn't cache data on the site.

👍 This may be the lightest solution, it wouldn't clog databases and shouldn't have as big an impact on the frontend.

Aurorum: Thanks! I've added that now. :)

anomiex: > 👍 This may be the lightest solution, it wouldn't clog databases and shouldn't have as big an impact on the frontend.

As long as it doesn't result in every frontend page view hitting the API or database. If that happens, it'll probably wind up tripping some WordPress.com monitor and we'll have to add some sort of caching back in.

jeherve: Hm, good point.

Aurorum: If it’s more trouble than it’s worth, I can always just remove the feature - seems a popular request (and a bit defeatist!!) though.

Is there a downside to caching in the post meta table?

Aurorum: I've gone with @kangzj's idea of caching in the post meta since it actually looks like that's what the Related Posts feature already does, and it seems to resolve the problem here. 

Let me know if I should change it!

Aurorum:
Trying to fix the tests, but I seem to get this error whenever I try to update the branch (not just this branch, but any Jetpack one).

```
Checking for filename collisions. Just a sec...
No collisions detected.
Checking if changelog files are needed. Just a sec...
Project packages/backup-helper-script-manager is being changed, but no change file in projects/packages/backup-helper-script-manager/changelog/ is touched!
Use `jetpack changelogger add <slug>` to add a change file for each project.
Guidelines: https://github.com/Automattic/jetpack/blob/trunk/docs/writing-a-good-changelog-entry.md
Checking if changelog files are needed. Just a sec...
? Found 46 project(s) that need a changelog. Create and add the same changelog to all of them?
```

Any ideas what I'm doing wrong?! I'm assuming that I'm not supposed to actually add a changelog for each project.

kraftbj:
At first blush, I thought it might be a situation where if you update a package version that is used in a lot of different packages, there will often need a changelog in all of the impacted projects to acknowledge the package dependencies were updated.

But, I don't think that's what's happening here. Or at least not as straightforward. It looks like your branch is a little behind trunk as well, I opened https://github.com/Aurorum/jetpack/pull/2 for you to pull in the latest trunk changes. I'm curious if you accept that PR into your branch, if the issue will continue.

Aurorum:
Welp, that didn't work, but thanks for giving it a go! Looks like I've made a mess of this branch in the process though - sorry about that. 

I've managed to update my branch again using GitHub, but I'm still getting the same error when committing my local changes - same with other branches (I tried with #34717 too). Any other ideas on where I might be going wrong? It was working fine earlier this week, and I don't believe that I've made any changes. :/ 

Thanks!

Aurorum:
Sorry about the labels - I'm still having a bunch of trouble with my dev environment and trying to commit, so I can't run the tests locally. I think I've fixed them though, and that this branch is okay again.

anomiex:
The check compares `origin/trunk` with `HEAD`, and the trunk branch in your fork is 1226 commits behind Automattic/jetpack. Try syncing your fork; GitHub should give you a button for that when visiting https://github.com/Aurorum/jetpack/tree/trunk.

Aurorum:
Ah, that worked perfectly! Thank you.

Aurorum:
Hi folks, just looping back on this old one and realised some tests were broken (sorry!). They should be fixed now, but I'd be grateful if someone could run them again so I can double-check. 

This should be good to review otherwise (sorry - I know the labels are a mess), but let me know if I'm missing anything else!

Aurorum:
Looks like `composer.lock` still needed updating - should be taken care of now!

kangzj:
Thanks for working on this @Aurorum 👍 I'll take a look early tomorrow 😄

kangzj:
Sorry I didn't manage to get onto this one today. Will do first thing tomorrow.

Aurorum:
All good! :)

projects/packages/stats/src/class-wpcom-stats.php
```diff
@@ -399,6 +404,44 @@ protected function fetch_stats( $args = array() ) {
...
+	 */
+	protected function fetch_post_stats( $args, $post_id ) {
+		$endpoint    = $this->build_endpoint();
+		$meta_name   = '_' . self::STATS_CACHE_TRANSIENT_PREFIX;
+		$stats_cache = get_post_meta( $post_id, $meta_name );
+
+		if ( $stats_cache ) {
+			$data = reset( $stats_cache );
+
+			if ( is_wp_error( $data ) ) {
```
kangzj, line 427: I took a look at the returning value of get_post_meta, which wouldn't be an array of WP_Error, so I guess the condition here will never be satisfied.

<strike>We could the block as a check whether $data is still an array; otherwise `key` function would throw out an error. So I guess we could do something like:

```
if( ! is_array($data) ) {
  return new WP_Error(...);
}
```
</strike>

Update:

I guess it might be better if let it continue without cache, if cache fails?

projects/packages/stats/src/class-wpcom-stats.php
```diff
@@ -203,13 +203,18 @@ public function get_file_downloads( $args = array() ) {
...
+	 * @param array $args           Optional query parameters.
+	 * @param bool  $cache_in_meta  Optional should cache in post meta.
 	 * @return array|WP_Error
 	 */
-	public function get_post_views( $post_id, $args = array() ) {
+	public function get_post_views( $post_id, $args = array(), $cache_in_meta = false ) {
 		$this->resource = sprintf( 'post/%d', $post_id );
 
+		if ( $cache_in_meta ) {
+			$this->fetch_post_stats( $args, $post_id );
```
kangzj:
```suggestion
			return $this->fetch_post_stats( $args, $post_id );
```

I believe we should return the result here 😄

kangzj (changes requested):
Thanks for your amazing work @Aurorum 👍 

All looks and works amazingly well except for missing a `return` and another nitpick 👍 👍 👍

projects/packages/stats/src/class-wpcom-stats.php
```diff
@@ -399,6 +404,44 @@ protected function fetch_stats( $args = array() ) {
...
+				return $data;
+			}
+
+			$time       = key( $data );
+			$expiration = self::STATS_CACHE_EXPIRATION_IN_MINUTES * MINUTE_IN_SECONDS;
+
+			if ( ( time() - $time ) < $expiration ) {
+				return array_merge( array( 'cached_at' => $time ), $data[ $time ] );
+			}
+		}
```
kangzj, line 437: Maybe we could try something like:

```PHP
		if ( is_array( $data ) ) {
			  $time       = key( $data );
			  $expiration = self::STATS_CACHE_EXPIRATION_IN_MINUTES * MINUTE_IN_SECONDS;
  
			  if ( ( time() - $time ) < $expiration ) {
				  return array_merge( array( 'cached_at' => $time ), $data[ $time ] );
			  }
			}
		}
```

Aurorum: > I guess it might be better if let it continue without cache, if cache fails?

If we're doing this, would it be preferable to just not cache the error instead? Something like this: 

```php
		$wpcom_stats = $this->fetch_remote_stats( $endpoint, $args );
		if ( ! is_wp_error( $wpcom_stats ) ) {
			update_post_meta( $post_id, $meta_name, array( time() => $wpcom_stats ) );
		}
```

Let me know if I'm misunderstanding you though! :) 

(And thank you for the review!!)

Aurorum: I tested this a bit more and it seems to work well, so I've committed it now. That way we only cache the view count, which is the data we want to cache.

Happy to change it if you prefer a different approach though!

projects/packages/stats/src/class-wpcom-stats.php
Aurorum: Whoops, great spot! :)

projects/packages/stats/src/class-wpcom-stats.php
Aurorum: > I took a look at the returning value of get_post_meta, which wouldn't be an array of WP_Error, so I guess the condition here will never be satisfied.

Ah, totally missed this. I've changed it now so that we only cache the stats data that we're interested in.

Aurorum:
Thanks so much - very much appreciated! :))

projects/packages/stats/src/class-wpcom-stats.php
kangzj: ah sorry! I would prefer the errors being cached please just in case. I was thinking whether WP_Error could be serialized and unserialized correctly, but I just tested, it worked fine!

Could we revert the change please?

Aurorum: No problem - switched it back!

kangzj (approved):
Looks good to me! Thanks for working on this 👍 ❤️

Aurorum:
Thanks so much for your help @kangzj! ❤️

Do we need to create an issue like #32937 on the steps still needed to get the block out of beta (supporting WPcom etc) or will that be okay?

kangzj:
> Do we need to create an issue like https://github.com/Automattic/jetpack/issues/32937 on the steps still needed to get the block out of beta (supporting WPcom etc)

Hi @Aurorum, thanks for your contribution. Yes, it'll be great to track it than forgetting it later 😄 I've filed an issue for that:

https://github.com/Automattic/jetpack/issues/35436

jeherve:
Related diff: D138696-code

Aurorum:
Thanks!! Out of interest, does that handle the WPcom side? If so, can I remove the WPcom check? 

https://github.com/Automattic/jetpack/blob/ce56219691155c4ab0941a31d1d6fa6d4f9367ec/projects/plugins/jetpack/extensions/blocks/blog-stats/blog-stats.php#L25

jeherve:
> does that handle the WPcom side? If so, can I remove the WPcom check?

We should indeed be able to remove that check now.

kangzj:
Thanks @jeherve 👍

PR #34560, opened 2023-12-11
Google Fonts Module: Fix the error when parsing a font family that has a reference to a path

arthur791004:
Related to pbxlJb-5a2-p2

## Proposed changes:

* If the font family is a reference to a path, the Google Fonts module will throw an error as follows in PHP 8.1 when parsing the font families as we assume the value is a string.
  ```
  [08-Dec-2023 15:54:01 UTC] PHP Fatal error:  Uncaught TypeError: preg_match(): Argument #2 ($subject) must be of type string, array given in /home/wpcom/public_html/wp-content/mu-plugins/jetpack-plugin/sun/modules/google-fonts/current/class-jetpack-google-font-face.php:182
  ```
* As a result, this PR proposes a change to ignore those font families because we also get the value from the path they refer to.

### Other information:

## Testing instructions:

* Apply the changes to your sandbox
  ```
  bin/jetpack-downloader test jetpack fix/google-fonts-font-family-ref
  ```
* Go to your WP.com site
* Activate the Assembler theme
* Go to your homepage or the Site Editor
* Ensure the error is gone

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-google-fonts-font-family-ref
+++ b/projects/plugins/jetpack/changelog/fix-google-fonts-font-family-ref
@@ -1,0 +1,4 @@
+Significance: patch
+Type: other
+
+Google Fonts Module: Fix the error when parsing a font family that has a reference to a path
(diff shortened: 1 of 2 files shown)


fushar (approved):
Nice, the fix makes sense, thanks!

|Before|After|
|-|-|
|image|(errors gone)|

PR #34565, opened 2023-12-11
Refactor full sync chunking to not repeat unnecessary DB queries and processing

### LINKED ISSUE #34574: Small improvement in the full sync chunking logic loop
### Impacted plugin

Jetpack

### What

Refactor full sync chunking to not repeat unnecessary DB queries and processing

### How

When working on https://github.com/Automattic/jetpack/pull/34390 we discovered that `get_next_chunk` ( which calls the DB and processes data) is being called unnecessarily in the [while loop](https://github.com/Automattic/jetpack/blob/trunk/projects/packages/sync/src/modules/class-module.php#L342) when the limit of max chunks has been reached.

darssen:
This PR aims to simplify the while loop [here](https://github.com/Automattic/jetpack/blob/trunk/projects/packages/sync/src/modules/class-module.php#L342)
so that we don't need to make unnecessary calls to the DB or any processing that might happen in `get_next_chunk`. The simple scenario would be when the limit of max chunks has been reached.

## Proposed changes:

The simple approach of checking for the condition at the end of the loop is not valid since setting the status to finished is part of the logic and needs to be taken into consideration. 
A refactor of the loop that improves the efficiency while making the code more readable is suggested. 

### Other information:

## Testing instructions:

* Let's start with a self-hosted site with this branch applied (be it your local or a JN)
* Create a bunch of posts (more than 100). I suggest using some plugin to easily duplicate.
* Create also some comments.
* Go to Jetpack Debugger and Schedule a Full Sync only.
* Verify the content was synced by checking the corresponding sync-audit logs (link provided in the Jetpack Debugger but change the queue to immediate-send)


projects/packages/sync/src/modules/class-posts.php
```diff
@@ -896,7 +896,7 @@ public function filter_posts_and_metadata_max_size( $posts, $metadata ) {
 			}
 			// Always allow the first post with its metadata.
 			if ( empty( $filtered_post_ids ) || ( $current_size + $post_content_size + $metadata_size ) <= ( self::MAX_SIZE_FULL_SYNC ) ) {
-				$filtered_post_ids[] = $post->ID;
+				$filtered_post_ids[] = strval( $post->ID );
```
darssen, line 899: This is to make sure that `$filtered_post_ids` will have identical structure than what is coming from DB. This will help later when using `$status['last_sent']` to compare.

fgiannar: Why not add the `strval` part only when comparing?

projects/packages/sync/src/modules/class-module.php
```diff
@@ -338,23 +360,31 @@ public function send_full_sync_actions( $config, $status, $send_until ) {
...
 			}
-			// The $ids are ordered in descending order.
+			// Updated the sent and last_sent status.
 			$status = $this->set_send_full_sync_actions_status( $status, $objects );
-		}
-
-		if ( ! $wpdb->last_error ) {
-			$status['finished'] = true;
+			if ( $last_item === $status['last_sent'] ) {
+				$status['finished'] = true;
```
fgiannar, line 386:
```suggestion
				$status['finished'] = true;
				return $status;
```

Shouldn't we return early here as well?

projects/packages/sync/src/modules/class-module.php
```diff
@@ -338,23 +360,31 @@ public function send_full_sync_actions( $config, $status, $send_until ) {
...
-		while ( $objects = $this->get_next_chunk( $config, $status, $limits['chunk_size'] ) ) {
-			if ( $chunks_sent++ === $limits['max_chunks'] || microtime( true ) >= $send_until ) {
+
+		$last_item = $this->get_last_item( $config );
+
+		while ( $chunks_sent < $limits['max_chunks'] && microtime( true ) < $send_until ) {
+			$objects = $this->get_next_chunk( $config, $status, $limits['chunk_size'] );
+
+			if ( $wpdb->last_error ) {
 				return $status;
```
fgiannar, line 371:
```suggestion
                                 $status['error'] = true;
				return $status;
```
Shouldn't we also update the `$status` here?

fgiannar (approved):
LGTM! Reviewed code and tested with the provided instructions 👍 

Thanks for working on this!

PR #34566, opened 2023-12-11
Social: Fix endpoint backwards compatibility

gmjuhasz:
With #34113 merged, we moved to use core endpoints to fetch the Social features. We also removed the old endpoint handlers, which cause backward compatibility issues as older plugin versions still ship with the newest PHP package where the endpoints are absent.

This change re-adds those endpoints, but the logic uses the new format of the options. I also had to add back the `class-settings.php` files as the initial state is populated from there.

We have to support three versions back, so the plan is to revert this change three versions from now. I annotated the files with the current version, so it's easy to keep track.

**Jetpack version: 12.9
Jetpack Social version: 3.0.0**

## Proposed changes:

* Re-added controllers with updated logic e6f1608a313203f32ee83e9f56265214d8bc704f
* Re-added the settings classes for initial state population f61739777b81aaff0bc8f3693054d33e0c5f66e5

### Other information:

## Jetpack product discussion

https://github.com/orgs/Automattic/projects/733/views/2?pane=issue&itemId=47186800

No.
## Testing instructions:

To see the issue, do the following:
* Check out the last jetpack release `git checkout plugins/jetpack@12.9`
* Check out the publicize package from trunk `git checkout origin/trunk projects/packages/publicize --no-overlay`
* Re-build `jetpack install plugins/social plugins/jetpack`
* Now open the Social admin page or the social settings under Jetpack -> Settings, and in the dev tools network tab you should see that the setting fetches fail because of invalid enpoint. The same happens if you try to update a value

To test this solution:
* Check out the last jetpack release `git checkout plugins/jetpack@12.9`
* Check out the publicize package from branch `git checkout fix/social-store-endpoint-backwards-compatibility projects/packages/publicize --no-overlay`
* Re-build `jetpack install plugins/social plugins/jetpack`
* Now you should have the latest Jetpack release with bleeding edge publicize package.
* Test the Social admin page and Jetpack settings, and confirm that changes are persisted, and everything works as intended.


manzoorwanijk (dismissed):
It works well for both Jetpack and Jetpack Social

projects/packages/publicize/src/auto-conversion-settings/class-rest-settings-controller.php
```diff
@@ -0,0 +1,136 @@
+<?php
+/**
+ * Class used to register REST API auto-conversion settings endpoints.
+ *
+ * Flagged to be removed after deprecation.
```
pablinos, line 5: Let's use the `@deprecated` annotation for this. We can add something to the backlog, to make sure we come back to tidy this up

pablinos (commented):
The testing instructions demonstrates and fixes the problem, but I'm not totally sure why the Social admin page on trunk ends up calling the old endpoints. Shouldn't the trunk version be using the core endpoints?

Either way it shows the conflict where we're loading the most recent version of the PHP package.

gmjuhasz:
> The testing instructions demonstrates and fixes the problem, but I'm not totally sure why the Social admin page on trunk ends up calling the old endpoints. Shouldn't the trunk version be using the core endpoints?

@pablinos it's because you only check out the trunk version from the publicize package, the UI and admin page is still the released version, which is calling the old endpoints, if that makes sense

manzoorwanijk (approved):
I can confirm that this PR fixes the failed API call when used with JP 12.9.

PR #34568, opened 2023-12-11
CRM: Fixing assorted placeholders: currency symbol and currency, notes from templates

coder-karen:
## Proposed changes:

This PR is a child PR of it's parent - https://github.com/Automattic/jetpack/pull/34490 - which is needed as well in order for the changes in this PR to take full effect. The changes in this PR include:

* The Quote Value placeholder will now show a currency symbol in mail templates (in the quote notification and quote accepted email notification).
* The Quote Currency placeholder now shows the currency set in CRM settings. It isn't clear to me what else that should be set to, or if the idea was to be able to set the currency for individual Quotes - if so that's beyond the scope of this PR for now.
* The Quote Notes placeholder will correctly render the Quote Template notes field if the notes field in the quote itself is blank.

### Other information:

## Jetpack product discussion

Parent issue: https://github.com/Automattic/zero-bs-crm/issues/3004
Parent PR: https://github.com/Automattic/jetpack/pull/34490

This PR fixes:
- https://github.com/Automattic/zero-bs-crm/issues/1588

No.

## Testing instructions:

* Test the ##QUOTE-VALUE## placeholder in the quote accepted and quote notification email templates (`/wp-admin/admin.php?page=zbs-email-templates&zbs_template_id=4` and `/wp-admin/admin.php?page=zbs-email-templates&zbs_template_id=2`). These should now display the currency symbol in those emails, whereas before they did not.
* Test the ##QUOTE-CURRENCY## placeholder in the two email templates, as well as adding it to a quote template. You should see the currency set in CRM site settings - `/wp-admin/admin.php?page=zerobscrm-plugin-settings&tab=locale` - displaying as a string (eg 'USD'). This should display in the quote editor, portal, pdf, and the emails.
* To test the ##QUOTE-NOTES## field rendering the quote template notes field if the quote is blank (and test escaping issues), follow the excellent testing steps in the comment of this issue: https://github.com/Automattic/zero-bs-crm/issues/1588

### DIFF
--- a/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
@@ -816,6 +816,11 @@ function ZeroBSCRM_get_quote_template() {
 			$replacements['biz-state']        = $bizState;
 			$replacements['contact-fullname'] = $customerName;
 
+			$settings = $zbs->settings->getAll();
+			if ( $settings['currency'] && $settings['currency']['strval'] ) {
+				$replacements['quote-currency'] = $settings['currency']['strval'];
+			}
+
 			// if DAL3, also replace any custom fields
 			if ( isset( $_POST['quote_fields'] ) && is_array( $_POST['quote_fields'] ) ) {
 
--- a/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
@@ -802,9 +802,6 @@ function ZeroBSCRM_get_quote_template() {
 			if ( empty( $quote_date ) ) {
 				$quote_date = gmdate( 'Y-m-d' );
 			}
-			if ( empty( $quote_notes ) ) {
-				$quote_notes = '[QUOTENOTES]';
-			}
 
 			// HTML is escaped just prior to the complete HTML in this function being returned
 			$working_html = wpautop( $quote_template['content'] );
--- a/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
@@ -759,6 +759,9 @@ function ZeroBSCRM_get_quote_template() {
 			if ( isset( $_POST['quote_fields']['zbscq_date'] ) && ! empty( $_POST['quote_fields']['zbscq_date'] ) ) {
 				$quote_date = sanitize_text_field( wp_unslash( $_POST['quote_fields']['zbscq_date'] ) );
 			}
+			if ( isset( $_POST['quote_fields']['zbscq_notes'] ) && ! empty( $_POST['quote_fields']['zbscq_notes'] ) ) {
+				$quote_notes = sanitize_text_field( wp_unslash( $_POST['quote_fields']['zbscq_notes'] ) );
+			}
 		}
 
 		// } Fill out rest
--- a/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
@@ -788,6 +788,9 @@ function ZeroBSCRM_get_quote_template() {
 			if ( empty( $quote_val ) && ! empty( $quote_template['value'] ) ) {
 				$quote_val = $quote_template['value'];
 			}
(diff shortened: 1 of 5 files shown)


projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
```diff
@@ -797,9 +803,7 @@ function ZeroBSCRM_get_quote_template() {
 				$quote_date = gmdate( 'Y-m-d' );
 			}
 			if ( empty( $quote_notes ) ) {
-				if ( isset( $_POST['quote_fields']['zbscq_notes'] ) ) {
-					$quote_notes = sanitize_text_field( wp_unslash( $_POST['quote_fields']['zbscq_notes'] ) );
-				}
+				$quote_notes = '[QUOTENOTES]';
```
tbradsha: Why set this?

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.QuoteTemplates.php
```diff
@@ -1070,7 +1070,7 @@ private function tidy_quotetemplate($obj=false,$withCustomFields=false){
             $res['owner'] = $obj->zbs_owner;
 
 
-			$res['title']            = $this->stripSlashes( $obj->zbsqt_title );
+			$res['title']            = wp_kses( html_entity_decode( $obj->zbsqt_title, ENT_QUOTES, 'UTF-8' ), $zbs->acceptable_restricted_html );
```
tbradsha, line 1073: 🤔 Why are we converting to HTML entities at any point? Are we saving things wrong?

projects/plugins/crm/includes/jpcrm-templating-placeholders.php
```diff
@@ -1755,6 +1755,16 @@ public function replace_placeholders(
...
+					// If the key is quote-currency and the value is empty, let's use the currency set in the site settings.
+					if ( $key === 'quote-currency' && $replace_with === '' ) {
+
+						$settings = $zbs->settings->getAll();
+						if ( $settings['currency'] && $settings['currency']['strval'] ) {
+
+							$replace_with = $settings['currency']['strval'];
+						}
+					}
+
```
tbradsha: As in the other PR, I don't think we should be putting key-specific fixes here; shouldn't we just pass the proper values to the function?

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.QuoteTemplates.php
coder-karen: Here we're making the same change as on line 1079 below for notes, which fixed the notes rendering issue (see https://github.com/Automattic/jetpack/pull/33614) and also fixes the title escaping issue. You made a suggestion in that PR regarding using a textarea and using sanitize_textarea_field. That could also be a good idea here. Possibly for a follow-up in that case 🤔

projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
coder-karen: Following the previous trend with the other values, however I notice it does print out [QUOTENOTES] if nothing is added to the template notes either. Removed now.

projects/plugins/crm/includes/jpcrm-templating-placeholders.php
coder-karen: I see what you mean. My original intention was to reduce duplication, hence putting it here. I've moved it back out to the relevant functions that need it: e3199b1c2df3d8674cd3d7f49cb9dc0573528a58

projects/plugins/crm/includes/ZeroBSCRM.DAL3.Obj.QuoteTemplates.php
tbradsha: Yeah, for notes it's a textarea, so we'd expect HTML, but for a title we'd expect regular text as-is. But out of scope here.

projects/plugins/crm/includes/jpcrm-templating-placeholders.php
tbradsha: One way to avoid duplication is to handle most of these in preprocessing via a helper function (e.g. have quotes call `render_quote_placeholders()`, which does all the special quote string manipulations.

Wasn't a blocker. :^)

tbradsha (approved):

PR #34569, opened 2023-12-11
CRM: Fixing quote date placeholders in the editor, portal, pdf, and mail templates

coder-karen:
## Proposed changes:

This PR is a child PR of it's parent - https://github.com/Automattic/jetpack/pull/34490 - which is needed as well in order for the changes in this PR to take full effect. The changes in this PR include:

* The Quote Date placeholder now shows the same date format (in the quote editor, client portal, and pdf) when initially set, or left to be set by default. This matches custom date fields as well.
* Custom Quote Date placeholders now render as expected in all areas (editor, portal, pdf, and the two quote email templates).
* Quote Accepted, Last Updated and Created placeholders now render 'as expected' (see additional notes below) in all areas except the quote editor (so portal, pdf, and the two quote email templates). For that, no information is available on initial quote creation so the placeholders are left unrendered there.
* The main date versions of date placeholders (see notes below) render the date based on site settings.

## Additional notes:

- For the 'Last Viewed' placeholder - how do we track a quote has been viewed? Do we need to? I don't see that we are currently doing this so it looks like it may have been added as something that may get implemented in the future, from what I can tell. I would vote for removing this completely.
- In terms of picking the correct date time:
  - The complication is that we have three different placeholders for several main placeholders. These include Quote Accepted, Quote Last Viewed, Quote Last Updated and Quote Created, as well as custom date fields.
  - What I've done to both respect the available placeholder options but also consider the date format itself and let there be some flexibility to site admins can make sure dates are displayed in the format of their choice, is to use the site date settings at `/wp-admin/options-general.php` for the datetime_str and date_str placeholder verisons of dates where three exist, and UTS for the main placeholder. All except for Quote Date which would use site settings.~main date (eg. quote-date, quote-accepted, etc - not considering the 'datetime_str' and 'date_str' versions). That does mean that two placeholders may be duplicated if one option is chosen in particular (`j F Y`), but I don't think that matters so much. It allows more flexibility in options for most.~
  - Also related, the reason the date functions are taking place at the placeholder rendering / Ajax file level is because once the date has rendered, we then save it in that format (affecting the quote editor, preview and portal). The date value sent via $_POST will always be yyyy-mm-dd ([MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date)), so the first place to change this is when building the quote data.
  - For emails, we could fetch the data from the database directly, so we can actually update the date functionality further up so it is saved correctly (eg `tidyquote`). There is a pending date related issue and existing functions that need updating here and further testing needed as this affects more than just palceholders, so I'm leaving that as a separate issue: https://github.com/Automattic/zero-bs-crm/issues/313

### Other information:

## Jetpack product discussion

Parent issue: https://github.com/Automattic/zero-bs-crm/issues/3004
Parent PR: https://github.com/Automattic/jetpack/pull/34490

Addresses:
- https://github.com/Automattic/zero-bs-crm/issues/3288

No.

## Testing instructions:

* Test the ##QUOTE-DATE## placeholder in a quote template, setting the quote date when creating a new quote and also not setting it. In both cases, the formatted date shown in the editor should be the same, and also in the client portal and pdf. This should be the same as the custom date field (bearing in mind each custom date field creates three placeholders, you need the correct one! That would be the first date custom field in the placeholder list (on the system status page) for that custom field).
* Test the custom Quote Date placeholders by creating a new custom quote field - `/wp-admin/admin.php?page=zerobscrm-plugin-settings&tab=customfields` - and checking the placeholder map for the new values (there should be three - regular date, date string, date time string). Add these to a quote template and check the rendering in the editor, portal and pdf. Similarly add these to the two quote email templates -  (`/wp-admin/admin.php?page=zbs-email-templates&zbs_template_id=4` and `/wp-admin/admin.php?page=zbs-email-templates&zbs_template_id=2`) - and check the results. What you should see is the main date rendered as a UTS string, then the datetime and date strings as per the site settings.
* Test the Quote Accepted, Last Updated and Created placeholders (there are three of each, available via the placeholder map). Add these to a quote template and check the rendering in the portal and pdf (will remain unrendered in the quote editor). Similarly, the placeholders will now work in the two quote email templates. 
* To test the date formatting, keep all of the date placeholders you've been testing with in place in a template. Then on the site settings page at `/wp-admin/options-general.php` change the site's date format. Create a new quote with the relevant template, and what you should see is all custom date fields, plus the quote-date placeholder, rendering. The date should be in the format you selected in site settings. All other date fields (Quote Accepted, Quote Created, Quote Last Updated) will not render in the editor (and the Quote Last viewed placeholders will just show 'Quote not yet viewed' text for now.
* Save the quote, and view the quote in the portal and pdf. All fields should now be rendered, in the correct format. The previously rendered date format should match those in the editor, and any date_str and datetime_str placeholders should now render what is in the site settings, and the main placeholder rendering a UTS string. Quote date should render the date as per the site settings.
* For all of the above, where the placeholders are rendering the date in a specific language, you can test changing the site language on the same admin settings page as the date settings. Try creating a new quote and viewing emails / portal / pdf and where relevant the quote editor, and the correct language should be used.

### DIFF
--- a/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
@@ -757,8 +757,8 @@ function ZeroBSCRM_get_quote_template() {
 				$quote_val = sanitize_text_field( wp_unslash( $_POST['quote_fields']['zbscq_value'] ) );
 			}
 			if ( isset( $_POST['quote_fields']['zbscq_date'] ) && ! empty( $_POST['quote_fields']['zbscq_date'] ) ) {
-				$sanitized_date = sanitize_text_field( wp_unslash( $_POST['quote_fields']['zbscq_date'] ) );
-				$quote_date     = jpcrm_uts_to_date_str( strtotime( $sanitized_date ), get_option( 'date_format' ) );
+				$sanitized_date = jpcrm_date_str_to_uts( sanitize_text_field( wp_unslash( $_POST['quote_fields']['zbscq_date'] ) ) );
+				$quote_date     = jpcrm_uts_to_date_str( $sanitized_date );
 			}
 		}
 
@@ -841,7 +841,7 @@ function ZeroBSCRM_get_quote_template() {
 								// Here is where we search and replace placeholders for dates with a date string and date time strings), initially checking the value is similar to that of 'yyyy-mm-dd'.
 								if ( preg_match( '/^[0-9]{4}-(0[1-9]|1[0-2])-(0[1-9]|[1-2][0-9]|3[0-1])$/', $v ) ) {
 
-									// Additional date validation to confirm the date is valid
+									// Additional date validation to confirm the date is valid, before processing (creating placeholder strings for searching and replacing).
 									$date_time = DateTime::createFromFormat( 'Y-m-d', $v );
 									if ( $date_time && $date_time->format( 'Y-m-d' ) === $v ) {
 
--- a/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
@@ -758,7 +758,7 @@ function ZeroBSCRM_get_quote_template() {
 			}
 			if ( isset( $_POST['quote_fields']['zbscq_date'] ) && ! empty( $_POST['quote_fields']['zbscq_date'] ) ) {
 				$sanitized_date = sanitize_text_field( wp_unslash( $_POST['quote_fields']['zbscq_date'] ) );
-				$quote_date     = wp_date( get_option( 'date_format' ), strtotime( $sanitized_date ) );
+				$quote_date     = jpcrm_uts_to_date_str( strtotime( $sanitized_date ), get_option( 'date_format' ) );
 			}
 		}
 
@@ -795,7 +795,7 @@ function ZeroBSCRM_get_quote_template() {
 				$quote_val = '[QUOTEVALUE]';
 			}
 			if ( empty( $quote_date ) ) {
-				$quote_date = wp_date( get_option( 'date_format' ) );
+				$quote_date = jpcrm_uts_to_date_str( time(), get_option( 'date_format' ) );
 			}
 			if ( empty( $quote_notes ) ) {
 				if ( isset( $_POST['quote_fields']['zbscq_notes'] ) ) {
@@ -845,7 +845,7 @@ function ZeroBSCRM_get_quote_template() {
 									$date_time = DateTime::createFromFormat( 'Y-m-d', $v );
 									if ( $date_time && $date_time->format( 'Y-m-d' ) === $v ) {
 
-										$working_html = jpcrm_process_date_variables( $v, $key, $working_html, $custom_field = true );
+										$working_html = jpcrm_process_date_variables( $v, $key, $working_html, $placeholder_str_start = '##QUOTE-' );
 
 									}
 								}
--- a/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
@@ -838,7 +838,7 @@ function ZeroBSCRM_get_quote_template() {
 							if ( isset( $_POST['quote_fields'][ 'zbscq_' . $key ] ) ) {
 								$v = sanitize_text_field( $_POST['quote_fields'][ 'zbscq_' . $key ] );
 
-								// Here is where we search and replace placeholders for dates with a date string and date time strings), initially checking the value is similar to that of a Unix timestamp
+								// Here is where we search and replace placeholders for dates with a date string and date time strings), initially checking the value is similar to that of 'yyyy-mm-dd'.
 								if ( preg_match( '/^[0-9]{4}-(0[1-9]|1[0-2])-(0[1-9]|[1-2][0-9]|3[0-1])$/', $v ) ) {
 
 									// Additional date validation to confirm the date is valid
--- a/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
@@ -757,7 +757,8 @@ function ZeroBSCRM_get_quote_template() {
 				$quote_val = sanitize_text_field( wp_unslash( $_POST['quote_fields']['zbscq_value'] ) );
 			}
 			if ( isset( $_POST['quote_fields']['zbscq_date'] ) && ! empty( $_POST['quote_fields']['zbscq_date'] ) ) {
-				$quote_date = sanitize_text_field( wp_unslash( $_POST['quote_fields']['zbscq_date'] ) );
+				$sanitized_date = sanitize_text_field( wp_unslash( $_POST['quote_fields']['zbscq_date'] ) );
+				$quote_date     = wp_date( get_option( 'date_time' ), strtotime( $sanitized_date ) );
 			}
 		}
 
@@ -794,7 +795,7 @@ function ZeroBSCRM_get_quote_template() {
 				$quote_val = '[QUOTEVALUE]';
 			}
 			if ( empty( $quote_date ) ) {
-				$quote_date = gmdate( 'Y-m-d' );
+				$quote_date = wp_date( get_option( 'date_format' ) );
 			}
 			if ( empty( $quote_notes ) ) {
 				if ( isset( $_POST['quote_fields']['zbscq_notes'] ) ) {
@@ -844,17 +845,7 @@ function ZeroBSCRM_get_quote_template() {
 									$date_time = DateTime::createFromFormat( 'Y-m-d', $v );
 									if ( $date_time && $date_time->format( 'Y-m-d' ) === $v ) {
 
-										$datetime_key       = $key . '_datetime_str';
-										$string_to_datetime = gmdate( 'd F Y H:i:s', strtotime( $v ) );
-										$date_key           = $key . '_date_str';
-										$string_to_date     = gmdate( 'd F Y', strtotime( $v ) );
-
-										$working_html = str_replace( '##QUOTE-' . strtoupper( $datetime_key ) . '##', $string_to_datetime, $working_html );
-										$working_html = str_replace( '##QUOTE-' . strtolower( $datetime_key ) . '##', $string_to_datetime, $working_html );
-										$working_html = str_replace( '##quote-' . strtolower( $datetime_key ) . '##', $string_to_datetime, $working_html );
-										$working_html = str_replace( '##QUOTE-' . strtoupper( $date_key ) . '##', $string_to_date, $working_html );
-										$working_html = str_replace( '##QUOTE-' . strtolower( $date_key ) . '##', $string_to_date, $working_html );
-										$working_html = str_replace( '##quote-' . strtolower( $date_key ) . '##', $string_to_date, $working_html );
+										$working_html = jpcrm_process_date_variables( $v, $key, $working_html, $custom_field = true );
 
 									}
 								}
--- a/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
@@ -859,7 +859,7 @@ function ZeroBSCRM_get_quote_template() {
 					}
 				}
 			}
-			$keys_staying_unrendered = array( 'quote-ID', 'quote-url' );
+			$keys_staying_unrendered = array( 'quote-ID', 'quote-url', 'quote-created', 'quote-created_datetime_str', 'quote-created_date_str', 'quote-accepted', 'quote-accepted_datetime_str', 'quote-accepted_date_str', 'quote-lastupdated', 'quote-lastupdated_datetime_str', 'quote-lastupdated_date_str' );
 			$working_html            = $placeholder_templating->replace_placeholders( array( 'global', 'contact', 'quote' ), $working_html, $replacements, array( ZBS_TYPE_CONTACT => $contact_object ), false, $keys_staying_unrendered ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.VariableNotSnakeCase
 
 			// } replace the rest (#fname, etc)
--- a/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
@@ -836,6 +836,28 @@ function ZeroBSCRM_get_quote_template() {
 							$v = '';
 							if ( isset( $_POST['quote_fields'][ 'zbscq_' . $key ] ) ) {
 								$v = sanitize_text_field( $_POST['quote_fields'][ 'zbscq_' . $key ] );
+
+								// Here is where we search and replace placeholders for dates with a date string and date time strings), initially checking the value is similar to that of a Unix timestamp
+								if ( preg_match( '/^[0-9]{4}-(0[1-9]|1[0-2])-(0[1-9]|[1-2][0-9]|3[0-1])$/', $v ) ) {
+
+									// Additional date validation to confirm the date is valid
+									$date_time = DateTime::createFromFormat( 'Y-m-d', $v );
+									if ( $date_time && $date_time->format( 'Y-m-d' ) === $v ) {
+
+										$datetime_key       = $key . '_datetime_str';
+										$string_to_datetime = gmdate( 'd F Y H:i:s', strtotime( $v ) );
+										$date_key           = $key . '_date_str';
+										$string_to_date     = gmdate( 'd F Y', strtotime( $v ) );
+
+										$working_html = str_replace( '##QUOTE-' . strtoupper( $datetime_key ) . '##', $string_to_datetime, $working_html );
+										$working_html = str_replace( '##QUOTE-' . strtolower( $datetime_key ) . '##', $string_to_datetime, $working_html );
+										$working_html = str_replace( '##quote-' . strtolower( $datetime_key ) . '##', $string_to_datetime, $working_html );
+										$working_html = str_replace( '##QUOTE-' . strtoupper( $date_key ) . '##', $string_to_date, $working_html );
+										$working_html = str_replace( '##QUOTE-' . strtolower( $date_key ) . '##', $string_to_date, $working_html );
+										$working_html = str_replace( '##quote-' . strtolower( $date_key ) . '##', $string_to_date, $working_html );
+
+									}
+								}
 							}
 
 							// allow upper or lower to catch various uses
--- a/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
@@ -758,7 +758,7 @@ function ZeroBSCRM_get_quote_template() {
 			}
 			if ( isset( $_POST['quote_fields']['zbscq_date'] ) && ! empty( $_POST['quote_fields']['zbscq_date'] ) ) {
 				$sanitized_date = sanitize_text_field( wp_unslash( $_POST['quote_fields']['zbscq_date'] ) );
-				$quote_date     = wp_date( get_option( 'date_time' ), strtotime( $sanitized_date ) );
+				$quote_date     = wp_date( get_option( 'date_format' ), strtotime( $sanitized_date ) );
 			}
 		}
 
--- a/projects/plugins/crm/includes/ZeroBSCRM.FormatHelpers.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.FormatHelpers.php
@@ -851,30 +851,27 @@ function zeroBSCRM_html_companyTimeline($companyID=-1,$logs=false,$companyObj=fa
 	}
 
 /**
- * Return an updated HTML string, replacing quote date placeholders with correct date strings based on site settings.
- * @param  string  $value         The value of the date variable to be replaced.
- * @param  string  $key           The key of the date variable to be replaced.
- * @param  string  $working_html  The HTML string to be updated.
- * @param  boolean $custom_field Whether the key in question is a custom quote field.
+ * Return an updated HTML string, replacing date placeholders with correct date strings based on site settings.
+ * @param  string $value                 The value of the date variable to be replaced.
+ * @param  string $key                   The key of the date variable to be replaced.
+ * @param  string $working_html          The HTML string to be updated.
+ * @param  string $placeholder_str_start The string to add to the beginning of the placeholder string (eg. ##QUOTE-).
  *
  * @return string                The updated HTML string.
  */
-function jpcrm_process_date_variables( $value, $key, $working_html, $custom_field = false ) {
-	$placeholder_str_start = '##';
-	if ( $custom_field ) {
-		$placeholder_str_start = '##QUOTE-';
-	}
-	$numeric_date_key       = $key;
-	$string_to_numeric_date = wp_date( get_option( 'date_format' ), strtotime( $value ) );
-	$datetime_key           = $key . '_datetime_str';
-	$string_to_datetime     = wp_date( 'd F Y ' . get_option( 'time_format' ), strtotime( $value ) );
-	$date_key               = $key . '_date_str';
-	$string_to_date         = wp_date( 'd F Y', strtotime( $value ) );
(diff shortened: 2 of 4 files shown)


projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
```diff
@@ -794,7 +795,7 @@ function ZeroBSCRM_get_quote_template() {
 				$quote_val = '[QUOTEVALUE]';
 			}
 			if ( empty( $quote_date ) ) {
-				$quote_date = gmdate( 'Y-m-d' );
+				$quote_date = wp_date( get_option( 'date_format' ) );
```
tbradsha: In general I'd recommend using `jpcrm_uts_to_date_str( $uts )`, but this appears to be equally valid.

projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
```diff
@@ -757,7 +757,8 @@ function ZeroBSCRM_get_quote_template() {
 				$quote_val = sanitize_text_field( wp_unslash( $_POST['quote_fields']['zbscq_value'] ) );
 			}
 			if ( isset( $_POST['quote_fields']['zbscq_date'] ) && ! empty( $_POST['quote_fields']['zbscq_date'] ) ) {
-				$quote_date = sanitize_text_field( wp_unslash( $_POST['quote_fields']['zbscq_date'] ) );
+				$sanitized_date = sanitize_text_field( wp_unslash( $_POST['quote_fields']['zbscq_date'] ) );
+				$quote_date     = wp_date( get_option( 'date_time' ), strtotime( $sanitized_date ) );
```
tbradsha: This gets a bit dicey with timezones...I believe it will cause issues (though haven't tested).

projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
```diff
@@ -836,6 +837,18 @@ function ZeroBSCRM_get_quote_template() {
 							$v = '';
 							if ( isset( $_POST['quote_fields'][ 'zbscq_' . $key ] ) ) {
 								$v = sanitize_text_field( $_POST['quote_fields'][ 'zbscq_' . $key ] );
+
+								// Here is where we search and replace placeholders for dates with a date string and date time strings), initially checking the value is similar to that of a Unix timestamp
+								if ( preg_match( '/^[0-9]{4}-(0[1-9]|1[0-2])-(0[1-9]|[1-2][0-9]|3[0-1])$/', $v ) ) {
```
tbradsha, line 842: > initially checking the value is similar to that of a Unix timestamp

This doesn't check for a Unix timestamp, but instead a `YYYY-MM-DD` (ISO standard).

tbradsha: Also, I'd avoid doing any "smart" processing on date-like content unless the field is really a date type.

projects/plugins/crm/includes/ZeroBSCRM.FormatHelpers.php
```diff
@@ -850,6 +850,44 @@ function zeroBSCRM_html_quoteStatusLabel($quote=array()){
 
 	}
 
+/**
```
tbradsha, line 853: I believe this function will result in timezone issues...I'm extremely paranoid about those, as I think I successfully wiped (most of) them out over the past few years and would hate to introduce more. I'll have to look again tomorrow, but:

* Some fields are stored in `YYYY-MM-DD` format, and to properly print in WP format we'd need to make sure we are not adding/removing an offset that could put us in a different day.
* Some fields are stored in UTS, which means we have to add the proper offset before retrieving.

I'm struggling to read this function (lots of variable soup). Will look with fresher eyes tomorrow.

Also, I'm not quite sure why we're making this quote-specific.

projects/plugins/crm/includes/ZeroBSCRM.FormatHelpers.php
```diff
@@ -850,6 +850,44 @@ function zeroBSCRM_html_quoteStatusLabel($quote=array()){
...
+ *
+ * @return string                The updated HTML string.
+ */
+function jpcrm_process_date_variables( $value, $key, $working_html, $custom_field = false ) {
+	$placeholder_str_start = '##';
+	if ( $custom_field ) {
+		$placeholder_str_start = '##QUOTE-';
+	}
+	$numeric_date_key       = $key;
+	$string_to_numeric_date = wp_date( get_option( 'date_format' ), strtotime( $value ) );
```
tbradsha: What's a numeric date?

projects/plugins/crm/includes/ZeroBSCRM.FormatHelpers.php
```diff
@@ -850,6 +850,44 @@ function zeroBSCRM_html_quoteStatusLabel($quote=array()){
...
+	$datetime_key           = $key . '_datetime_str';
+	$string_to_datetime     = wp_date( 'd F Y ' . get_option( 'time_format' ), strtotime( $value ) );
```
tbradsha: Why `d F Y`?

projects/plugins/crm/includes/ZeroBSCRM.FormatHelpers.php
```diff
@@ -850,6 +850,44 @@ function zeroBSCRM_html_quoteStatusLabel($quote=array()){
...
+	$date_key               = $key . '_date_str';
+	$string_to_date         = wp_date( 'd F Y', strtotime( $value ) );
+
+	$search_replace_pairs = array(
+		$placeholder_str_start . strtoupper( $numeric_date_key ) . '##' => $string_to_numeric_date,
+		$placeholder_str_start . strtolower( $numeric_date_key ) . '##' => $string_to_numeric_date,
+		$placeholder_str_start . strtolower( $numeric_date_key ) . '##' => $string_to_numeric_date,
```
tbradsha: Why is this here once uppercase and once lowercase? Same question with the items below.

projects/plugins/crm/includes/jpcrm-templating-placeholders.php
```diff
@@ -1755,10 +1755,31 @@ public function replace_placeholders(
 					
 					}
 
+					// If this is a Quote date key and is not set (Quote accepted or last viewed), let's print out a message saying the quote isn't accepted or viewed.
+					$possible_empty_quote_keys = array( 'quote-accepted', 'quote-accepted_date_str', 'quote-accepted_datetime_str', 'quote-lastviewed', 'quote-lastviewed_datetime_str', 'quote-lastviewed_date_str' );
+					if ( in_array( $key, $possible_empty_quote_keys, true ) ) {
+						// 82799 is 23hrs, 59mins, 59secs on Jan 1, 1970 (so a blank timestamp, allowing for systems to update the time within that day for any reason)
+						if ( empty( $replace_with ) || strtotime( $replace_with ) <= 82799 ) {
```
tbradsha: Why not `=== 0`? Is there any scenario it would be set to something other than 0 or a "recent" date?

projects/plugins/crm/includes/jpcrm-templating-placeholders.php
```diff
@@ -1755,10 +1755,31 @@ public function replace_placeholders(
...
+					}
+
+					// We want to check if a quote field is a unix timestamp, so this light approach will check if the value is numeric initially to lower the searches,
+					// then check if its _datetime_str equivalent exists (unless it's the 'quote-date' key), confirming our current key is a timestamp.
+					// Then we can convert the replace_with value to a human readable format based on site date settings.
+					if ( str_starts_with( $key, 'quote-' ) && is_numeric( $replace_with ) ) {
+
+						$new_key = '##' . strtoupper( $key ) . '_DATETIME_STR##';
+						if ( array_key_exists( $new_key, $to_replace ) || $key === 'quote-date' ) {
+							$replace_with = wp_date( get_option( 'date_format' ), $replace_with );
```
tbradsha: Timezone warning.

projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
coder-karen: Yes I think when we review all timezone functions generally we might want to swap things out if needed.

projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
coder-karen: I might be better off changing the comment here as I realize it isn't accurate - the date will only ever be in the format `yyyy-mm-dd` when coming from the $_POST data due to how the date input sends date data via POST. So we can still check for that format to prevent extra processing (the regex is still correct in this respect, just the comment is not) - or just check if `is_numeric` and run the datetime function check on that.

> Also, I'd avoid doing any "smart" processing on date-like content unless the field is really a date type.

Not too sure what you mean here?

projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
coder-karen: Not too sure what issues you mean here? I did make a mistake here though - it should have been `get_option( 'date_format' ). Fixing that did resolve issues I was now noticing when using some timezones, so perhaps that is what you meant? Commit: 07c425aee8ed834e67ea946545cf9811736f996a

projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
coder-karen: Following up, I ow gone through and change these due to the timezone issues.

projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
coder-karen: Following up - I have converted these to the jpcrm function now.

projects/plugins/crm/includes/ZeroBSCRM.FormatHelpers.php
coder-karen: Yes, you were right re timezones (and good point about being quote specific). I've addressed both of those now.

projects/plugins/crm/includes/ZeroBSCRM.FormatHelpers.php
coder-karen: My way of saying it's the original date, but that can be a non numeric date too. I've changed it to `$base_date_key`.

projects/plugins/crm/includes/ZeroBSCRM.FormatHelpers.php
coder-karen: So, this is the part that is covered in the notes in that I made the call to use those formats for those two types of placeholders (datetime_str and date_str), as otherwise it seems pointless having those placeholders available. It does mean possible duplication with the base date placeholder depending on what date setting is selected in site settings.

projects/plugins/crm/includes/ZeroBSCRM.FormatHelpers.php
coder-karen: In this case I'm just copying the catch-all method used in `ZeroBSCRM_get_quote_template` to catch various uses. Without having full knowledge of when different cases could arise, I opted for using the same fallback that exists already.

projects/plugins/crm/includes/jpcrm-templating-placeholders.php
coder-karen: In my case, the timestamp for 1 Jan 1970 was 3600 (+1hr for the exact time), so I figured that to cover any timezone offsets but still be 1st Jan then that figure should work.

projects/plugins/crm/includes/jpcrm-templating-placeholders.php
coder-karen: Should be fixed now.

projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
```diff
@@ -758,7 +758,7 @@ function ZeroBSCRM_get_quote_template() {
 			}
 			if ( isset( $_POST['quote_fields']['zbscq_date'] ) && ! empty( $_POST['quote_fields']['zbscq_date'] ) ) {
 				$sanitized_date = sanitize_text_field( wp_unslash( $_POST['quote_fields']['zbscq_date'] ) );
-				$quote_date     = wp_date( get_option( 'date_time' ), strtotime( $sanitized_date ) );
+				$quote_date     = jpcrm_uts_to_date_str( strtotime( $sanitized_date ), get_option( 'date_format' ) );
```
tbradsha: Note that this function defaults to the WP format, so you could just do this (not a blocker):

```suggestion
				$quote_date     = jpcrm_uts_to_date_str( strtotime( $sanitized_date ) );
```

projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
tbradsha: > Not too sure what you mean here?

Just because something looks like a date field doesn't mean it is a date field; the best way to ensure it's a date field is to look at the object definition (each field has a specific type associated). For example, one could have a math problem of `1500-12-30`. Or someone could have an input of 31 February (caught in your condition). But highly unlikely edge cases here, so probably safe enough to ignore.

projects/plugins/crm/includes/ZeroBSCRM.FormatHelpers.php
tbradsha: Ah, but looking in the function from which the fallbacks came, there's a subtle difference; the second and third line are different:

```
$working_html = str_replace( '##QUOTE-' . strtolower( $key ) . '##', $v, $working_html );
$working_html = str_replace( '##quote-' . strtolower( $key ) . '##', $v, $working_html );
```

In your version, the second and third lines are identical. To match the source you'd want to do an uppercase and lowercase variant of `$placeholder_str_start`. Or, since these are placeholders that didn't work well anyway, you could just drop the fallbacks and force the use of the uppercase one. :^)

projects/plugins/crm/includes/jpcrm-templating-placeholders.php
tbradsha: > In my case, the timestamp for 1 Jan 1970 was 3600

To confirm, it stored `3600` in the database when it was not accepted? It seems `0` is the value other queries use elsewhere, so if you're seeing anything else, it sounds like a different bug.

projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
tbradsha: Also, `strtotime()` may also introduce timezone issues.

tbradsha (changes requested):
Unfortunately, `strtotime()` is prone to timezone issues, as it uses the PHP server timezone, which often doesn't match the WP timezone:
```
$date = '2023-11-30';
$uts = strtotime( $date ) // 1701302400 <-- PHP server in UTC timezone, midnight
date_default_timezone_set('America/Los_Angeles');
$uts = strtotime( $date ) // 1701331200 <-- PHP server in PT timezone, midnight
```

In the end, though, we should ignore the PHP server timezone, and only use WP's timezone.

```
jpcrm_date_str_to_uts( $date ); // 1701345600 <-- WP set to UTC-12, midnight
```

If you try the following, you'll notice you always get the same date string, even when changing around WP timezones:

```
	$uts = jpcrm_date_str_to_uts( '2023-11-30' );
	error_log(var_export($uts, true));
	$date_str = jpcrm_uts_to_date_str( $uts );
	error_log(var_export($date_str, true));
```

Alternatively, you can always use midnight UTC:
```
	$uts = jpcrm_date_str_to_uts( $date, false, true );
	error_log(var_export($uts, true));
	$date_str = jpcrm_uts_to_date_str( $uts, false, true );
	error_log(var_export($date_str, true));
```

Just be glad you're not having to deal with browser locale as well. :^D

projects/plugins/crm/includes/ZeroBSCRM.FormatHelpers.php
```diff
@@ -850,6 +850,41 @@ function zeroBSCRM_html_quoteStatusLabel($quote=array()){
...
+ * @param  string $key                   The key of the date variable to be replaced.
+ * @param  string $working_html          The HTML string to be updated.
+ * @param  string $placeholder_str_start The string to add to the beginning of the placeholder string (eg. ##QUOTE-).
+ *
+ * @return string                The updated HTML string.
+ */
+function jpcrm_process_date_variables( $value, $key, $working_html, $placeholder_str_start = '##' ) {
+
+	$base_date_key       = $key;
+	$string_to_base_date = jpcrm_uts_to_date_str( strtotime( $value ), get_option( 'date_format' ) );
```
tbradsha: As best I understand it, this is the intent (`strtotime` comments aside):
* `$string_to_base_date` → date in WP format
* `$string_to_datetime` → datetime in hybrid hard-coded date + WP time format...why not `jpcrm_uts_to_datetime_str()`?
* `$string_to_date` → hard-coded date

Three thoughts:
1. For the datetime one, does this make sense to include? It's always a date being passed, not a datetime, so the time part would be some arbitrary time depending on what functions we used.
2. For the third one, perhaps we should hard-code to the ISO standard instead of an arbitrary "common (in English)" format?
3. We're basically creating undocumented placeholders here (e.g. they won't show in `/wp-admin/admin.php?page=zerobscrm-systemstatus&tab=placeholder-map`). Any reason we need all of these? The only one I see documented is `##QUOTE-DATE##`.

projects/plugins/crm/includes/jpcrm-templating-placeholders.php
```diff
@@ -1755,6 +1755,27 @@ public function replace_placeholders(
...
+					// If this is a Quote date key and is not set (Quote accepted or last viewed), let's print out a message saying the quote isn't accepted or viewed.
+					$possible_empty_quote_keys = array( 'quote-accepted', 'quote-accepted_date_str', 'quote-accepted_datetime_str', 'quote-lastviewed', 'quote-lastviewed_datetime_str', 'quote-lastviewed_date_str' );
+					if ( in_array( $key, $possible_empty_quote_keys, true ) ) {
+						// 82799 is 23hrs, 59mins, 59secs on Jan 1, 1970 (so a blank timestamp, allowing for systems to update the time within that day for any reason)
+						if ( empty( $replace_with ) || strtotime( $replace_with ) <= 82799 ) {
+							$replace_with = str_starts_with( $key, 'quote-accepted' ) ? __( 'Quote not yet accepted', 'zero-bs-crm' ) : __( 'Quote not yet viewed', 'zero-bs-crm' );
+						}
+					}
+
+					// We want to check if a quote field is a unix timestamp, so this light approach will check if the value is numeric initially to lower the searches,
```
tbradsha: ~This is not a Unix timestamp...perhaps that's what lead to the confusion and the desire to add (undocumented) `_DATETIME_STR` and `_DATE_STR` placeholders?~

projects/plugins/crm/includes/jpcrm-templating-placeholders.php
tbradsha: Looking at the code, it should always set it to `0` (not the Unix timestamp + timezone offset equivalent of 1 Jan 1970) if not accepted. I'd stick to `0` for any comparisons.

tbradsha: (in my ideal world, we'd never use `0` for timestamps (or `-1`), but legacy code...)

projects/plugins/crm/includes/jpcrm-templating-placeholders.php
```diff
@@ -1755,6 +1755,27 @@ public function replace_placeholders(
...
+					// then check if its _datetime_str equivalent exists (unless it's the 'quote-date' key), confirming our current key is a timestamp.
+					// Then we can convert the replace_with value to a human readable format based on site date settings.
+					if ( str_starts_with( $key, 'quote-' ) && is_numeric( $replace_with ) ) {
```
tbradsha: Wouldn't this basically assume any number (including floats) in a quote custom field is a date?

projects/plugins/crm/includes/jpcrm-templating-placeholders.php
tbradsha: There's inconsistency with how we show things here vs. with Unix timestamps elsewhere. In other objects, we show the following placeholders:

* `##SOMETHING_DATE##` <-- UTS
* `##SOMETHING_DATE_DATE_STR##` <-- formatted date string
* `##SOMETHING_DATE_DATETIME_STR##` <-- formatted datetime string

projects/plugins/crm/includes/jpcrm-templating-placeholders.php
tbradsha: The answer: yes, but you have a condition three lines down that filters those out.

coder-karen: Yes that's right, my reasoning is that it is saving the `array_key_exists` check for fields that are more likely to be dates (well, numeric), instead of everything, or every quote field. I can easily remove it if it seems excessive.

projects/plugins/crm/includes/ZeroBSCRM.FormatHelpers.php
coder-karen: That sounds like a much better approach :D

projects/plugins/crm/includes/jpcrm-templating-placeholders.php
coder-karen: But, I'll be getting rid of this now given changes based on earlier comments.

tbradsha: It's fine! I'll note that conditions fail fast, so if it was all in one `if` statement and the first condition failed, it wouldn't go on to the next, but nothing wrong with splitting it here.

projects/plugins/crm/includes/jpcrm-templating-placeholders.php
coder-karen: > To confirm, it stored 3600 in the database when it was not accepted? 

That was the result after the str_to_time conversion. I'm switching to using `jpcrm_date_str_to_uts` though now.

projects/plugins/crm/includes/ZeroBSCRM.FormatHelpers.php
coder-karen: We chatted in Slack, following up here to say that I'll follow that suggested approach - using the in-built functions, following the style in the similar invoice date PR: https://github.com/Automattic/zero-bs-crm/pull/2239
So the base date is actually the ISO standard (except for Quote Date itself, which I'll leave with just one available placeholder). 
For point 3, to clarify here too we're only checking for strings to search and replace for quote custom fields in this case, which all have three available placeholder options.

projects/plugins/crm/includes/jpcrm-templating-placeholders.php
coder-karen: That's now fixed in a follow-up commit.

projects/plugins/crm/includes/jpcrm-templating-placeholders.php
```diff
@@ -1758,21 +1758,9 @@ public function replace_placeholders(
...
-					// then check if its _datetime_str equivalent exists (unless it's the 'quote-date' key), confirming our current key is a timestamp.
-					// Then we can convert the replace_with value to a human readable format based on site date settings.
-					if ( str_starts_with( $key, 'quote-' ) && is_numeric( $replace_with ) ) {
-
-						$new_key = '##' . strtoupper( $key ) . '_DATETIME_STR##';
-						if ( array_key_exists( $new_key, $to_replace ) || $key === 'quote-date' ) {
-							$replace_with = gmdate( 'Y-m-d', $replace_with );
-							$string       = jpcrm_process_date_variables( $replace_with, $key, $string );
+						if ( empty( $replace_with ) || jpcrm_date_str_to_uts( $replace_with ) === 0 || jpcrm_datetime_str_to_uts( $replace_with ) === 0 ) {
+							$replace_with = str_starts_with( $key, 'quote-accepted' ) ? __( 'Quote not yet accepted', 'zero-bs-crm' ) : __( 'Quote not yet viewed', 'zero-bs-crm' );
```
tbradsha, line 1763: ❤️

tbradsha (approved):
The only issue I saw was `##QUOTE-DATE##` when in the quote content (as opposed to being imported from the quote template). That'd likely be out of scope.

Huge improvements; thank you!

projects/plugins/crm/includes/ZeroBSCRM.AJAX.php
coder-karen: Following up here for completion as we discussed this in Slack - the $_POST data doesn't have any object related fields associated with keys, only the keys and their values.

PR #34576, opened 2023-12-11
Contact form: more defensive in screen checks, avoid PHP warnings

### LINKED ISSUE #34538: Warning: Attempt to read property 'id' on null in Jetpack Forms
### Impacted plugin

Jetpack

### Quick summary

An user reported an issue with a conflict between one of our plugins and Jetpack Forms, which I think you should fix on your side. 

This is the error:

> Warning: Attempt to read property 'id' on null in .../jetpack-forms/src/contact-form/class-admin.php on line 475

This is the affected line: https://github.com/Automattic/jetpack-forms/blob/40dba57b663df150cb83b9d89b0e703c4d69fac6/src/contact-form/class-admin.php#L475

And I think you should add a check for the screen object like you do here: https://github.com/Automattic/jetpack-forms/blob/40dba57b663df150cb83b9d89b0e703c4d69fac6/src/contact-form/class-admin.php#L452

### Steps to reproduce

One of our plugins opens a tab where the screen is not available, which triggers the error above.

_No response_

### What actually happened

_No response_

### Impact

One

### Available workarounds?

_No response_

### Logs or notes

_No response_

jeherve:
Fixes #34538

## Proposed changes:

Be more defensive when checking whether we are on the right screen, to avoid PHP warnings.

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

I am not quite sure how to reproduce the original issue just yet. We may receive additional information to help with that in #34538.

Until then, you can try the following:

1. Go to Pages > Add New
2. Insert a form in a new page.
3. Publish your page.
4. Access that page in the frontend, and leave a test submission.
5. Go to the Feedback menu in wp-admin.
6. Attempt some of the actions, like marking the submissions as spam; they should work as expected.


andrii-lysenko (dismissed):
The code looks good to me and I haven't experienced any issues while testing.

zinigor (approved):
Looks like a good change to me!

PR #34579, opened 2023-12-11
Launchpad: setup_general task completion callback

### LINKED ISSUE #84058: Set up your site task (setup_general) is always marked as complete
### Quick summary

The "Set up your site" task (`setup_general` on the back end) is always marked as complete in the fullscreen launchpad, which can be incorrect. Furthermore, this task is disabled, which makes it impossible to fix the issue from the launchpad. We need to allow users set up their site title and tagline.

### Steps to reproduce

* Create a new site from (https://wordpress.com/start)
* Pick a domain and a plan (which can both be free)
* When you get to the step that asks what you want to do with your site, pick "Other"
* Pick a design for your site
* When you get to the fullscreen launchpad, you'll notice that the "Set up your site" task is marked as complete, and is not clickable

The task should not be completed, and I should be able to click on it to set up my site title and tagline.

### What actually happened

The task was marked as complete and I couldn't interact with it.

### Impact

Some (< 50%)

### Available workarounds?

Simple, Atomic

### Logs or notes

I've looked at the back end code and I see that this is also an issue on the back end, as we have this task marked as complete in all cases. What we should be doing is marking the task as complete when the user specifies a site title and site description, which would be best-handled on the back end, most likely by hooking into the option-specific `'update_option_[option_name]'` actions. (It looks like we already have similar logic for the more specific `site_title` task.)

We also need to work out whether we can show the relevant screen to update the site settings for the user, as the screen's availability will depend on the Stepper flow that we happen to be in.

andres-blanco:
Fixes https://github.com/Automattic/wp-calypso/issues/84058

## Proposed changes:

This diff will change the completion event of the `setup_general` task to be based on wether the blog title or description changed.

**This should be merged AFTER https://github.com/Automattic/wp-calypso/pull/85101** so the incomplete task shows in the right position.

### Other information:

## Testing instructions:

* Apply this PR to your sandbox
* Sandbox public-api
* Go to wordpress.com/start and create a new site (can be free) with the `other` intent
* Go through the design flow and get to the Fullscreen Launchpad
* Verify that the `Set up your site` task appears as not completed
* Click on the task
* Change title/description and save
* Verify that  the task is marked as complete in the launchpad
Before
image

After
image

*NOTE*: there was also an ordering issue, making the `setup_general` task appear at the top of the list, even if it was incomplete. This is addressed here:

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -1474,7 +1474,6 @@ function wpcom_launchpad_mark_setup_general_task_complete( $old_value, $value ) 
 		return;
 	}
 
-	l( 'CALLING wpcom_launchpad_mark_setup_site_tasks_complete', $old_value, $value );
 	if ( $old_value !== $value ) {
 		wpcom_mark_launchpad_task_complete( 'setup_general' );
 	}
--- a/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
+++ b/projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
@@ -117,7 +117,7 @@ function wpcom_launchpad_get_task_definitions() {
 			'get_title'            => function () {
 				return __( 'Set up your site', 'jetpack-mu-wpcom' );
 			},
-			'is_complete_callback' => '__return_true',
+			'is_complete_callback' => 'wpcom_launchpad_is_task_option_completed',
 			'is_disabled_callback' => '__return_true',
 			'get_calypso_path'     => function ( $task, $default, $data ) {
 				return '/settings/general/' . $data['site_slug_encoded'];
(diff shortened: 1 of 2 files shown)


projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
```diff
@@ -1430,6 +1430,31 @@ function wpcom_launchpad_mark_setup_site_tasks_complete( $old_value, $value ) {
...
+function wpcom_launchpad_mark_setup_general_task_complete( $old_value, $value ) {
+	if ( defined( 'HEADSTART' ) && HEADSTART ) {
+		return;
+	}
+
+	if ( wp_installing() ) {
+		return;
+	}
+
+	l( 'CALLING wpcom_launchpad_mark_setup_site_tasks_complete', $old_value, $value );
```
sixhours: Did you intend to leave a log statement here? Worried we'll pollute the error logs with unnecessary notifications.

sixhours (commented):
This is working as described along with the Calypso sorting PR! Left a note about the log statement, and it looks like it needs a rebase.

projects/packages/jetpack-mu-wpcom/src/features/launchpad/launchpad-task-definitions.php
andres-blanco: nope, not intended at all. I will remove it. Thanks for spotting :)

valterlorran (approved):
LGTM!

PR #34584, opened 2023-12-11
AI Client: Fix one click actions behavior

CGastrell:
Editing the input after using one of the one-click actions from the AI Assistant ends up in inconsistent state of the action buttons

## Proposed changes:
This PR improves the logic behind which we toggle suggestion and prompt actions upon any change on the input

### Other information:

## Jetpack product discussion
p1702318281645179-slack-C054LN8RNVA

No

## Testing instructions:
Before this change, using one of the one-click actions and then editing the input would get you in a weird combination of buttons, allowing you to go "Back to edit" with no visible change and seemingly inconsistent.

With this change, any input on the AI Assistant will get the block on "edit mode", showing the prompt action buttons (Cancel | Generate)

### DIFF
--- a/projects/js-packages/ai-client/changelog/fix-ai-one-click-actions-behavior
+++ b/projects/js-packages/ai-client/changelog/fix-ai-one-click-actions-behavior
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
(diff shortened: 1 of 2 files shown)


dhasilva (approved):
LGTM :brain:

PR #34592, opened 2023-12-12
Subscriptions: add "Send as email" toggle

### LINKED ISSUE #10876: Subscriptions: add Gutenberg support on option to toggle email delivery on a per-post basis
With WordPress 5.0 it became impossible for customers to use the `jetpack_allow_per_post_subscriptions` filter as described here: https://jetpack.com/support/subscriptions/#filters

Previously, a new option would appear in a metabox, like this:

image

Code ref: https://github.com/Automattic/jetpack/blob/d7a1fadb4163d62d388c1b636c05f261fbab31fd/modules/subscriptions.php#L122-L159

Reported in 1626604-zen

### LINKED ISSUE #44263: Newsletter: allow not send blog post as an email
Jetpack issue https://github.com/Automattic/jetpack/issues/10876

This was added to our HappOps Product Support Github repo ( 20-GH-happops ) as an item that is low priority, but likely a quick win. We are often asked by customers for the ability to publish a new post on a blog without notifying their subscribers. 

To name a few situations where this has come up - some customers have their own mailing tool they use to notify subscribers, some customers post frequently and get complaints about too many emails, some want to backdate posts and not notify subscribers, and some customers have private posts they would like to make public without notifying anyone. 

Historically this has only been surfaced to my knowledge on WPCOM Ideas in several different threads:

- Two reports here: pKdGS-Wi-p2
- 17 reports here: pKdGS-78-p2 
- 7 reports here: pKdGS-1IG-p2
- 5 reports here: pKdGS-1z4-p2

The above is definitely not an inclusive list of all instances where this has been requested, as not all are noted here in WPCOM Ideas.

simison:
Resolves https://github.com/Automattic/wp-calypso/issues/44263
Resolves https://github.com/Automattic/jetpack/issues/10876

Adds "Send as email" toggle:

(FYI Below screenshots are outdated)

Screenshot 2023-12-12 at 15 08 01

## Proposed changes:

* Adds a toggle to disable email sending post-by-post basis.
* Register the meta, and sync the setting to WP.com

The feature pre-existed before Gutenberg, so the backend was already handled in Jetpack.

Previously the toggle was "negated", i.e. checking it to "not send email". Clearer UX is to do something when toggle is enabled, hence the meta and toggle handling seem a bit backwards.

**TODO:** PR needs some minor work in pre/post-publish panel copies and to check if post was published previously.
image
image
image

### Other information:

## Testing instructions:

* Create a new post
* In post settings, leave toggle enabled and publish the post.
* Note pre-publish panel toggle — you can try change it, go back to post panel, etc.
* Note copy at pre-publish panel.
* Post gets sent as an email
* The toggle should be disabled now and in truthy setting (the email was already sent)
   image

* Create another post, leave toggle disabled and publish the post
* Note copy at pre-publish panel.
* No email was sent
* The toggle should be disabled now and in falsy setting (no way to send email now that post was published)

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
@@ -215,6 +215,12 @@ function NewsletterPostPublishSettingsPanel( { accessLevel } ) {
 
 	const showMisconfigurationWarning = getShowMisconfigurationWarning( postVisibility, accessLevel );
 
+	const isSendEmailEnabled = useSelect( select => {
+		const meta = select( editorStore ).getEditedPostAttribute( 'meta' );
+		// Meta value is negated, "don't send", but toggle is truthy when enabled "send"
+		return ! meta?.[ META_NAME_FOR_POST_DONT_EMAIL_TO_SUBS ];
+	} );
+
 	return (
 		<>
 			<PluginPostPublishPanel
@@ -233,7 +239,11 @@ function NewsletterPostPublishSettingsPanel( { accessLevel } ) {
 				icon={ <JetpackEditorPanelLogo /> }
 				name="jetpack-subscribe-newsletters-postpublish-panel"
 			>
-				{ ! showMisconfigurationWarning && <SubscribersAffirmation accessLevel={ accessLevel } /> }
+				{ ! isSendEmailEnabled || showMisconfigurationWarning ? (
+					<p>{ __( 'Only published, did not send this post.', 'jetpack' ) }</p>
+				) : (
+					<SubscribersAffirmation accessLevel={ accessLevel } />
+				) }
 			</PluginPostPublishPanel>
 
 			{ ! isStripeConnected && (
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
@@ -18,14 +18,20 @@ import { store as editorStore } from '@wordpress/editor';
 import { useState } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import { external, Icon } from '@wordpress/icons';
-import { accessOptions } from '../../shared/memberships/constants';
+import {
+	accessOptions,
+	META_NAME_FOR_POST_DONT_EMAIL_TO_SUBS,
+} from '../../shared/memberships/constants';
 import { useAccessLevel, isNewsletterFeatureEnabled } from '../../shared/memberships/edit';
 import {
 	NewsletterAccessDocumentSettings,
-	NewsletterAccessPrePublishSettings,
+	NewsletterEmailDocumentSettings,
 } from '../../shared/memberships/settings';
 import SubscribersAffirmation from '../../shared/memberships/subscribers-affirmation';
-import { getShowMisconfigurationWarning } from '../../shared/memberships/utils';
+import {
+	getShowMisconfigurationWarning,
+	MisconfigurationWarning,
+} from '../../shared/memberships/utils';
 import { store as membershipProductsStore } from '../../store/membership-products';
 import metadata from './block.json';
 import EmailPreview from './email-preview';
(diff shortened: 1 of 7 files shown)


projects/plugins/jetpack/extensions/shared/memberships/settings.js
```diff
@@ -367,3 +369,49 @@ export function NewsletterAccessPrePublishSettings( { accessLevel } ) {
...
+		return ! meta?.[ META_NAME_FOR_POST_DONT_EMAIL_TO_SUBS ];
+	} );
+
+	/*
+	// Don't allow changing setting for a post that was already published in the past.
+	const postWasEverPublished = useSelect( select => {
+		const meta = select( editorStore ).getEditedPostAttribute( 'meta' );
+		return ! meta?.[ META_NAME_FOR_POST_WAS_EVER_PUBLISHED ];
+	} );
+	console.log( 'postWasEverPublished:', postWasEverPublished ); // eslint-disable-line
```
simison: This setting felt flaky, so commented it out. If it doesn't work, we can remove for now and add back in separate PR.

jeherve (commented):
This works well for me, but I don't know if the wording for that setting is clear enough. At first, I thought the toggle was related to the preview email below:

image

- Maybe there should be a separator between the 2?
- Maybe the parts that do not apply anymore get removed when you toggle off the email?

Screenshot 2023-12-12 at 18 17 37

- What if instead of a toggle, we had a "none" option as a radio, something like this?

image

simison:
@jeherve thanks! Yep, we're working still on the wording and structure of the pre-publish and post-publish panels for this 👍  It definitely needs some adjustments.

simison:
> What if instead of a toggle, we had a "none" option as a radio, something like this?

That panel controls _access_, which is somewhat separate from whom we send the emails, albeit very related. I.e. the access also affects how you see blog posts, posts in the Reader, etc. Hence the setting should be separate.

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -91,6 +92,20 @@ function register_block() {
 		)
 	);
 
+	register_post_meta(
+		'post',
+		META_NAME_FOR_POST_DONT_EMAIL_TO_SUBS,
```
simison, line 97: Note that this was old meta field I'm re-using for backwards compatibility, but it was never registered like this:

https://github.com/search?q=repo%3AAutomattic%2Fjetpack%20_jetpack_dont_email_post_to_subs&type=code

lezama (dismissed):
Tested and works great and everything makes much more sense than before.

Feel free to keep polishing but Approving in case you want to merge 😄

projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js
```diff
@@ -113,32 +119,24 @@ function NewsletterPrePublishSettingsPanel( { accessLevel, isModuleActive, showP
...
-					<Button variant="secondary" onClick={ showPreviewModal }>
-						{ __( 'Send test email', 'jetpack' ) }
-					</Button>
-				</>
-			) }
+	const postVisibility = useSelect( select => select( editorStore ).getEditedPostVisibility() );
+	const showMisconfigurationWarning = getShowMisconfigurationWarning( postVisibility, accessLevel );
 
-			{ shouldLoadSubscriptionPlaceholder && (
+	const isSendEmailEnabled = useSelect( select => {
```
simison, line 125: In a follow-up, I'd like this in a re-usable React hook.

projects/plugins/jetpack/extensions/shared/memberships/settings.js
```diff
@@ -287,23 +288,38 @@ export function NewsletterAccessDocumentSettings( { accessLevel } ) {
...
-					) : (
-						<SubscribersAffirmation prePublish={ true } accessLevel={ _accessLevel } />
-					) }
-				</PanelRow>
-			) }
+			render={ ( { canEdit } ) => {
+				return (
+					<ToggleControl
+						checked={ isSendEmailEnabled }
+						disabled={ isPostPublished || ! canEdit }
```
simison, line 317: In a follow-up we should check if the post was published already previously and set the toggle disabled. Prior art:

https://github.com/Automattic/jetpack/blob/2f325aecc2ea5a7d9ec599c8ec9d2c7bf7417909/projects/plugins/jetpack/modules/subscriptions.php#L1047-L1062

https://github.com/Automattic/jetpack/blob/2f325aecc2ea5a7d9ec599c8ec9d2c7bf7417909/projects/plugins/jetpack/extensions/blocks/subscriptions/panel.js#L110-L114

I tried it briefly but it felt flaky, so better in separate PR.

lezama (approved):

projects/plugins/jetpack/extensions/shared/memberships/settings.js
simison: https://github.com/Automattic/jetpack/pull/34711

PR #34600, opened 2023-12-12
Disable official buttons option in Sharing Buttons block

### LINKED ISSUE #34580: Sharing Block: Disable `official buttons` setting in Sharing Buttons block
Currently, we are not supporting `official buttons` view in Sharing Buttons block so we wan't to disable this option for now to avoid confusion.

----

Epic: #19694

andrii-lysenko:
Fixes https://github.com/Automattic/jetpack/issues/34580

## Proposed changes:

As this feature is not available yet, we want to mark this option disabled.
I've also added localization to the options.

Also, open to change it so we just remove the button entirely as mentioned https://github.com/Automattic/jetpack/issues/34580#issuecomment-1852331255, as don't have a strong opinion on this

### Other information:

## Testing instructions:

- Add Sharing Buttons block.
- In settings (right block sidebar), `Official buttons` should not be present

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/sharing-buttons/edit.js
+++ b/projects/plugins/jetpack/extensions/blocks/sharing-buttons/edit.js
@@ -38,8 +38,6 @@ export function SharingButtonsEdit( props ) {
 							{ value: 'icon', label: __( 'Icon Only', 'jetpack' ) },
 							/* translators: Sharing: Sharing button option label. */
 							{ value: 'text', label: __( 'Text Only', 'jetpack' ) },
-							/* translators: Sharing: Sharing button option label. */
-							{ value: 'official', label: __( 'Official Buttons', 'jetpack' ), disabled: true },
 						] }
 						value={ styleType }
 						onSelect={ value => setAttributes( { styleType: value } ) }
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/sharing-buttons/edit.js
```diff
@@ -32,10 +32,14 @@ export function SharingButtonsEdit( props ) {
...
-							{ value: 'text', label: 'Text Only' },
-							{ value: 'official', label: 'Official Buttons' },
+							/* translators: Sharing: Sharing button option label. */
+							{ value: 'icon-text', label: __( 'Icon & Text', 'jetpack' ) },
+							/* translators: Sharing: Sharing button option label. */
+							{ value: 'icon', label: __( 'Icon Only', 'jetpack' ) },
+							/* translators: Sharing: Sharing button option label. */
+							{ value: 'text', label: __( 'Text Only', 'jetpack' ) },
+							/* translators: Sharing: Sharing button option label. */
+							{ value: 'official', label: __( 'Official Buttons', 'jetpack' ), disabled: true },
```
andrii-lysenko: I could probably use wording like 'Official Buttons (In Development)' or something. Or leave as is.

jeherve:
> Also, open to change it so we just remove the button entirely

That would be my vote as well, I think it avoids setting expectations too early.

andrii-lysenko:
@jeherve Thanks! Removed the button.

monsieur-z (approved):
Works as expected 🚢 

The comments for translators might not provide enough context if you ask me.

andrii-lysenko:
@monsieur-z Thanks! I used them from Calypso, so my assumption is that it's already has a translation, so I just wanted to keep things in line.

PR #34608, opened 2023-12-13
Jetpack Google Fonts: Fix the fonts provided by theme are still added

arthur791004:
Related to p1702367211316529/1702365341.631869-slack-CRWCHQGUB

## Proposed changes:

* We have to ignore the fonts that are provided by the theme to keep the behavior consistent between the Site Editor & frontend page

| Before | After |
| - | - |
| Screenshot 2023-12-13 at 3 40 29 PM | Screenshot 2023-12-13 at 3 39 51 PM |

### Other information:

## Testing instructions:

* Apply this change to your WoA site via Jetpack Beta Tester
* Active the Adventurer theme
* Go to the Site Editor
* Open the Styles > Typography > Any Element (e.g.: Text) > Font
* Pick the Rubik font
* Pick different font weights
* Make sure that only `font-weight: 900` works
* Go to your frontend page
* Make sure that only `font-weight: 900` works as well


projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
```diff
@@ -94,8 +101,12 @@ function jetpack_get_theme_fonts_map() {
 
 	$theme_fonts_map = array();
 	foreach ( $raw_data['settings']['typography']['fontFamilies'] as $font_family ) {
-		if ( isset( $font_family['name'] ) ) {
-			$theme_fonts_map[ $font_family['name'] ] = true;
+		$font_family_name = isset( $font_family['name'] )
+			? $font_family['name']
+			: Jetpack_Google_Font_Face::get_font_family_name( $font_family );
```
jeherve: The good news is, since Jetpack now requires PHP 7+, we can use the null coalescing operator :)

```suggestion
		$font_family_name = $font_family['name'] ?? Jetpack_Google_Font_Face::get_font_family_name( $font_family );
```

jeherve (commented):
This looks good, I just have a tiny nitpick!

projects/plugins/jetpack/modules/google-fonts/current/load-google-fonts.php
arthur791004: Oh, nice! I saw Jetpack still supported PHP 5.6+ last time

mreishus:
I tried testing it out WPCOM by looking on a client site that was having problems (They select weight 100 in the editor and see it render thin there, but on the homepage it renders thicker than 100). It didn't seem to fix the issue for me.  See more info at D131894-code.

arthur791004:
>  It didn't seem to fix the issue for me

It looks good on my side. I guess the changes might not have effects on your sandbox 😅

image

arthur791004:
>  It didn't seem to fix the issue for me but on the homepage it renders thicker than 100).

It looks good on my side, and the font remains the same (it renders 400 actually) on both the Site Editor and the front page.

image

The purpose of this PR is to keep the fonts remaining the same on both the Site Editor and the front page. It won't solve the problem that the font weights are not provided by the theme.

taipeicoder (approved):
Tested and works as intended:

https://github.com/Automattic/jetpack/assets/797888/7b9265be-6f4a-40ac-b254-611e95d48fa9

Although I feel that as a result of making the Site Editor more "accurate", we are making it also a bit more "buggier" (aka: Users wondering why font weights don't work).

arthur791004:
> we are making it also a bit more "buggier" (aka: Users wondering why font weights don't work).

I don't think it's a bit more “buggier” as the issue exists on the self-hosted sites.

taipeicoder:
> > we are making it also a bit more "buggier" (aka: Users wondering why font weights don't work).
> 
> I don't think it's a bit more “buggier” as the issue exists on the self-hosted sites.

Yes, I meant that we are exposing an existing issue that is more discoverable to users.
But either way users will be confused as to why their font appearances don't work as they want to 😅

mreishus:
Right, I have a user that wants to use 100 weight font, and they can't because their theme ships the font, but not 100 weight. I can't think of a workaround for them.

PR #34610, opened 2023-12-13
Boost: Refactor speed scores into react

haqadn:
Fixes c/b5lGIECr-tr/26-scoresvelte

## Proposed changes:
* Refactored speed score scrion into react
* Moveed styles inside the component
* Fixed style in error-notice
* Moved data-sync interaction inside performance history
* Reduced LoC 😎 by cleaning up redundant css

### Other information:

## Jetpack product discussion
None

No

## Testing instructions:

* Play with speed scores and performance history and make sure they are still working well.


thingalon:
There's a z-ordering issue going on.

image

thingalon:
There's a misalignment with the mouseover on the history graph.

When you mouseover, it focuses on the closest graphed point and shows a tooltip. That's fine, but it shows a grey bar indicating what you are focused on to the right of what is highlighted.

I think that grey bar is meant to be centered on the current graph point, right?

image

projects/plugins/boost/app/assets/src/js/features/speed-score/lib/hooks.ts
```diff
@@ -0,0 +1,128 @@
...
+ * @param {RefreshDependencies} dependencies                         - The dependencies to watch.
+ * @param {Array<boolean>}      dependencies.moduleStates            - An array of booleans that represent the state of the modules.
+ * @param {number}              dependencies.criticalCssCreated      - The timestamp of when the critical CSS was created.
+ * @param {boolean}             dependencies.criticalCssIsGenerating - Whether the critical CSS is currently generating.
+ * @param {RefreshFunction}     loadScore                            - The method to refresh the speed score.
+ */
+export const useDebouncedRefreshScore = (
+	{ moduleStates, criticalCssCreated, criticalCssIsGenerating }: RefreshDependencies,
+	loadScore: RefreshFunction
+) => {
```
thingalon, line 99: 🤔 Given it takes a list of dependencies to use and a function to call... Is there a reason not to just make this a general-purpose debouncing thing a bit like useEffect, but with debouncing?

projects/plugins/boost/tests/e2e/lib/pages/wp-admin/JetpackBoostPage.js
```diff
@@ -18,7 +18,7 @@ export default class JetpackBoostPage extends WpPage {
 	async chooseFreePlan() {
 		const button = this.page.locator( 'text=Start for free' );
 		await button.click();
-		await this.waitForElementToBeVisible( '.jb-section--scores' );
+		await this.waitForElementToBeVisible( '[data-testid="speed-scores"]' );
```
thingalon, line 21: 🤔 Well, it's a solution that works around the generated module class names. I think it makes sense. 👍

thingalon (commented):
Minor suggestion in the review below. Raised a few bugs in testing above.

projects/plugins/boost/app/assets/src/js/features/speed-score/lib/hooks.ts
haqadn: I didn't make this general purpose because it treats `criticalCssIsGenerating` slightly differently, it only triggers if the value is true.

haqadn:
> When you mouseover, it focuses on the closest graphed point and shows a tooltip. That's fine, but it shows a grey bar indicating what you are focused on to the right of what is highlighted.
> 
> I think that grey bar is meant to be centered on the current graph point, right?

The gray bar highlights the entire day of the specific date. The tic is not centered because the measurements were taken at midnight in your site's case. This is actually working accurately, but I see how this can look buggy. Let's discuss this.

thingalon:
> > When you mouseover, it focuses on the closest graphed point and shows a tooltip. That's fine, but it shows a grey bar indicating what you are focused on to the right of what is highlighted.
> > I think that grey bar is meant to be centered on the current graph point, right?
> 
> The gray bar highlights the entire day of the specific date. The tic is not centered because the measurements were taken at midnight in your site's case. This is actually working accurately, but I see how this can look buggy. Let's discuss this.

Ok, if this is the pre-migration behaviour we'll stick with it for now and review it post-migration.

projects/plugins/boost/app/assets/src/js/features/speed-score/lib/hooks.ts
thingalon: Hmm, ok fair enough. It does look like something we can eventually generalize with a bit of effort but I don't want to lose momentum on the migration so I'll ✅  it :)

thingalon (approved):

PR #34612, opened 2023-12-13
Split `PublicizeForm` component

manzoorwanijk:
## Proposed changes:

* This PR breaks down `PublicizeForm` component into smaller ones

### Other information:

## Testing instructions:

* Run `jetpack build --all`
* Goto post editor
* Confirm that everything is same as before


projects/js-packages/publicize-components/src/components/form/instagram-supported-notice.tsx
```diff
@@ -0,0 +1,43 @@
...
+					{ __( 'Connect now', 'jetpack' ) }
+				</Button>,
+				<Button
+					key="learn-more"
+					href={ getRedirectUrl( 'jetpack-social-connecting-to-social-networks' ) }
+					target="_blank"
+					rel="noreferrer noopener"
+				>
+					{ __( 'Learn more', 'jetpack' ) }
+				</Button>,
```
pablinos, line 37: Do you think we should make a `LearnMore` component? There's this Button and a some ExternalLink components that are quite similar.

projects/js-packages/publicize-components/src/components/form/advanced-plan-nudge.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { ExternalLink } from '@wordpress/components';
+import { createInterpolateElement, useCallback } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+import { usePublicizeConfig } from '../../..';
+import useDismissNotice from '../../hooks/use-dismiss-notice';
+import Notice from '../notice';
+
+const MONTH_IN_SECONDS = 30 * 24 * 60 * 60;
+
+export const AdvancedPlanNudge: React.FC = () => {
```
pablinos, line 12: Not for this PR of course, but I think it would be good to research lint rules for TS. I can only find [one instance of `React.FC`](https://github.com/WordPress/gutenberg/blob/d9ea8d072626f5ad1e5d50ab4207590cfa170b4a/packages/components/src/higher-order/with-focus-outside/test/index.tsx#L43) so there's either a conscious decision to rely on inference, or TS isn't being used extensively.

If there are some sylistic rules we can define for this, I think it might help.

pablinos (approved):
This works well. It's a little hard to tell if anything has crept as the code is moved to separate components, but it looks like a straight copy, which is great. We can continue to improve it in updates.

projects/js-packages/publicize-components/src/components/form/instagram-supported-notice.tsx
manzoorwanijk: I did want to do that but I thought to simply copy the components as they are without making any changes so that the review is easier.

projects/js-packages/publicize-components/src/components/form/advanced-plan-nudge.tsx
manzoorwanijk: Yes, lint rules will help.

pnpm-lock.yaml
```diff
@@ -4127,11 +4130,11 @@ packages:
     resolution: {integrity: sha512-q+epW0trjVUUHboliPb4UF9g2msf+w61b32tAkFEwL/IwP0DQWgbCMM0Hbe3e3WXSKz5VcUXbzJQgy8Hkra/Lg==}
     dependencies:
       tunnel: 0.0.6
-      undici: 5.28.2
+      undici: 5.27.2
     dev: false
 
-  /@adobe/css-tools@4.3.2:
-    resolution: {integrity: sha512-DA5a1C0gD/pLOvhv33YMrbf2FK3oUzwNl9oOJqE4XVjuEtt6XIakRcsd7eLiOSPkp1kTRQGICTA8cKra/vFbjw==}
+  /@adobe/css-tools@4.3.1:
```
anomiex, line 4136: It's unfortunate this was downgraded in this PR, as version 4.3.2 includes a security fix. See https://github.com/Automattic/jetpack/security/dependabot/65

manzoorwanijk: I'm sure it happened due to a merge conflict and regenerating the lockfile.

PR #34614, opened 2023-12-13
Top Posts & Pages Endpoint: Remove Unnecessary Parameters 

Aurorum:
Follow up to #34479

## Proposed changes:
It just occurred to me that I messed up a little in #34479 when I refactored the endpoint because it's not necessary to maintain some of the parameters anymore. There's also no issue with now adding a `permission_callback` since the block uses the Helper function directly when rendering it on the front end. 

The endpoint already returns a few extra posts to handle configuring content types within the Block Editor. It's not possible to filter these out from the API request to the Stats endpoint, so this approach prevents the need to make another API request whenever the user changes content types. 

## Jetpack product discussion

N/A.

No, this is a Janitorial PR.

## Testing instructions:

Enable Beta blocks and insert a Top Posts and Pages block. If they load in both the Block Editor and when previewed on the site, then everything should be fine. You can also check that `$is_rendering_block` works by disabling some of the content types to display (eg. display pages instead of posts).

Screenshot 2023-12-13 at 13 58 41


kangzj:
Thanks for working on this! Sorry for the delay - I'll give it a spin early next week.

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-top-posts.php
```diff
@@ -47,7 +47,9 @@ public function register_routes() {
 				array(
 					'methods'             => WP_REST_Server::READABLE,
 					'callback'            => array( $this, 'get_top_posts' ),
-					'permission_callback' => '__return_true',
+					'permission_callback' => function () {
```
kangzj, line 50: 👍

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-top-posts.php
```diff
@@ -107,9 +93,7 @@ public function get_post_types() {
 	 */
 	public function get_top_posts( $request ) {
 		$period = $request->get_param( 'period' );
-		$number = $request->get_param( 'number' );
-		$types  = $request->get_param( 'types' );
-		return Jetpack_Top_Posts_Helper::get_top_posts( $period, $number, $types );
+		return Jetpack_Top_Posts_Helper::get_top_posts( $period, '', '' );
```
kangzj: maybe we could set default values for the two params so that we don't have to pass them here. Also considering the types of the params, could we please use `null` as default values? Thoughts?

kangzj (approved):
Sorry for the late review. The PR will need a rebase/merge to be able to merge. Thanks again for your contribution 👍

kangzj (dismissed):

Aurorum:
Sorry @kangzj - disabling because just about to push a commit to address your comments. :)

projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-top-posts.php
Aurorum: Makes total sense! Added now. :)

kangzj (approved):
let's :shipit:

PR #34615, opened 2023-12-13
Form block: hide 'lead capture' variation for WP.com Atomic sites

simison:
I noticed that the Creative Mail CRM integration in forms was hidden just for simple sites, but not for WP.com sites.

I believe this was a partnership block developed for Jetpack audience, but WP.com didn't have the partnership?

Also moves the thing next to the other lead-capture form.

Screenshot 2023-12-13 at 13 29 05
Screenshot 2023-12-13 at 13 28 43

## Proposed changes:

*

### Other information:

## Testing instructions:

* Go to '..'
*

### DIFF
--- a/projects/packages/forms/src/blocks/contact-form/edit.js
+++ b/projects/packages/forms/src/blocks/contact-form/edit.js
@@ -1,5 +1,9 @@
 import { ThemeProvider } from '@automattic/jetpack-components';
-import { getJetpackData, isSimpleSite } from '@automattic/jetpack-shared-extension-utils';
+import {
+	getJetpackData,
+	isAtomicSite,
+	isSimpleSite,
+} from '@automattic/jetpack-shared-extension-utils';
 import {
 	InnerBlocks,
 	InspectorControls,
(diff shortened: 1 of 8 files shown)


projects/packages/forms/src/blocks/contact-form/edit.js
```diff
@@ -327,7 +331,7 @@ export const JetpackContactFormEdit = forwardRef(
 							instanceId={ instanceId }
 						/>
 					) }
-					{ ! isSimpleSite() && (
+					{ ! ( isSimpleSite() || isAtomicSite() ) && (
```
simison, line 334: It looks like I wrapper too much onto check? Jetpack CRM should stay.

And prolly keep showing Creative Mail if it was already configured.

jeherve (approved):

projects/packages/forms/src/blocks/contact-form/edit.js
jeherve: I think that's fine for now, to be honest. It makes sense to me to keep things simpler on WoA. But you could certainly improve the logic in a follow-up PR to keep promoting Jetpack CRM.

simison: 👍 

Ideally we also promote on simple, and just put them through business plan upgrade.

PR #28944, opened 2023-02-14
Subscribe block: Hide social followers toggle when showSubscribersTotal is not on or isPublicizeEnabled is false

TimBroddin:
Fixes #28725

## Proposed changes:
This PR hides the "Include social followers in count" when "Show subscribers count" is off, or when Publicize is not enabled.

CleanShot 2023-02-14 at 15 06 59@2x

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Does this pull request change what data or activity we track or use?

No

## Testing instructions:
- Apply this PR
- Create a new post
- Add a Subscribe block
- "Include social followers in count" should be hidden when "Show subscribers count" is off, or Publicize is disabled.

PR #34617, opened 2023-12-13
Subscribe block: don't include social followers on subscriber counts by default

simison:
Update "include social follows in count" to be disabled by default.

## Proposed changes:

* Don't include social followers on subscriber counts by default
* Don't hide the social count toggle, simply disable it. Makes for less jumpy UI.

**Before**, enabling social count would set "include social counts" active by default as well

Screenshot 2023-12-13 at 14 39 20

**After**, "include social counts" stays disabled until enabled specifically:

Screenshot 2023-12-13 at 14 47 13

Screenshot 2023-12-13 at 14 47 20

We cannot just switch "default" in the `block.json` to false, because block editor doesn't store values matching the default in block's saved attributes. Hence every block with "include social counts" enabled (which was the default before) _does not_ have this value stored. Changing the default to `false` would switch all those blocks to not include social counts, hence altering their numbers and sites.

Hence we manually check for `undefined` when toggle is touched, and always store both `true`/`false` values for `includeSocialFollowers` in the attributes now. 

Altering the numbers (albeit somewhat vanity number) would be controbersial.

Note that since Twitter connections have been disconnected for a while, we haven't been able to update the social counts for sites anyway. They're stuck in the past counts.

### Other information:

## Testing instructions:

* Insert block, go to settings and enable the "show counts".
* Ensure social counts toggle stays disabled, and value is stored in attributes when looking at block code (three dots menu from the top):
    Screenshot 2023-12-13 at 15 03 24
* Enabling the "include social counts" includes in the code now as well:
   Screenshot 2023-12-13 at 15 03 44
* Manually remove the `includeSocialFollowers` attribute from block markup and save the post.
* When previewing the block on the frontend, log `$include_social_followers` value and see it's `true`.
* Open the block with missing `` atrribute in the editor, and confirm that toggle appears "enabled" (since unset value defaults to enabled). Setting the value on/off still works.
    Screenshot 2023-12-13 at 15 15 45
    Screenshot 2023-12-13 at 15 16 14


simison:
Block fixtures need updating in this one.

doc: https://github.com/Automattic/jetpack/blob/27fcd9758e93cfa7199da3e39831a1302607935f/projects/plugins/jetpack/extensions/shared/test/block-fixtures.md

pkuliga (commented):
The code looks good but it feels a bit confusing when the “Include social followers in count” toggle is disabled, but on, and when I enable the toggle above than it switches to off. I would expect it to be off when disabled.

projects/plugins/jetpack/extensions/blocks/subscriptions/controls.js
```diff
@@ -231,7 +231,15 @@ export default function SubscriptionControls( {
...
 					onChange={ () => {
-						setAttributes( { showSubscribersTotal: ! showSubscribersTotal } );
+						setAttributes( {
+							showSubscribersTotal: ! showSubscribersTotal,
+							// Don't do anything if set previously, but by default set to false. We want to disencourage including social count as it's misleading.
+							// We don't want to rely setting "default" in block.json to falsy, because the default value was previously "true".
+							// Hence users without this set will still get social counts included in the subscriber counter.
+							// Lowering the subscriber count on their behalf with code change would be controversial.
+							includeSocialFollowers:
+								typeof includeSocialFollowers === 'undefined' ? false : includeSocialFollowers,
```
simison, line 241: To address @pkuliga your feedback at https://github.com/Automattic/jetpack/pull/34617#pullrequestreview-1841614648, I need to change the last `includeSocialFollowers` to something, maybe `false`?  My head isn't fresh enough now to wrap around this, and see what exactly needs to be here. Might be we even can't do anything about it since we want to preserve backwards compatibility and default to `false`? Tomorrow's problem!

simison: Actually I'll just merge and we can address in a follow-up! What do you think @pkuliga ?

pkuliga: Sure, let's do this!

pkuliga (approved):

projects/plugins/jetpack/extensions/blocks/subscriptions/controls.js
```diff
@@ -242,16 +250,21 @@ export default function SubscriptionControls( {
 						}
 					} }
 				/>
-				{ showSubscribersTotal && isPublicizeEnabled ? (
+				{ isPublicizeEnabled && (
```
manzoorwanijk, line 253: This will not work for published posts on simple sites with a free plan. See https://github.com/Automattic/jetpack/pull/28944/files#r1715184860

simison: Thanks, so basically needs updating to check if publicize module is enabled/disabled, or if Jetpack social is installed?

manzoorwanijk: Right.

PR #34618, opened 2023-12-13
Jetpack Boost: Strict TypeScript

pyronaur:
Guardrails off.

IIRC we enabled typescript without strict mode a long time ago because we were "ramping up slowly". I think 2 years is slow enough 😛 

On top of that, this addresses missing React component types.

## Proposed changes:
Enables strict mode in TypeScript.

### Other information:

## Testing instructions:

Apply patch, create a function with unspecified types, see squiggly lines:

2023_12_13RaalHoSH 42


thingalon (dismissed):
Try running `pnpm run build-development` and you will see a giant list of errors. E2E tests and Github build fail too.

While I agree with this idea, I'm not willing to put trunk in a state where it can't be compiled.

pyronaur:
@thingalon The project now compiles.

1 - Strict mode is enabled in tsconfig - IDE's should pick up typescript files strictly
2 - Rollup doesn't actually validate TypeScript - it just compiles both in development and production. This now remains unchanged
3 - Updated the dev workflow a bit too to also watch for non-strict TypeScript errors during development by default.
4 - Added commands `pnpm run dev-compile` and `pnpm run dev-validate-ts`. They're both run concurrently in dev mode (default watch for jetpack cli), but you can run them separately in two panes in iTerm :)

haqadn (approved):
🥇

PR #34632, opened 2023-12-13
Contact Form: align half-width fields on same row

### LINKED ISSUE #34567: Form block shows 50% fields in separate lines
### Impacted plugin

Jetpack

### Quick summary

When two 50% fields are added to a Form block, they are shown on different lines instead of side by side on the live view of the page. They are shown side by side in the editor:

Editor view:

imagen

Preview and live view:

imagen

### Steps to reproduce

1. Create a Form block.
2. Add two fields and set the field width to 50%.
3. Save as draft or publish.
4. Check the preview and/or visit the page on the live site.

The two 50% fields should be side by side, as they appear in the editor: 

imagen

### What actually happened

The fields appear stacked:

imagen

### Impact

All

### Available workarounds?

Simple

### Logs or notes

7387565-zd-a8c
Reproduced on test site.

monsieur-z:
Fixes https://github.com/Automattic/jetpack/issues/34567

## Proposed changes:

Make sure _Contact Form_ fields that are 50% wide or less are aligned on the same row for all form styles. It's currently only the case for the _Default_ style.

| Before | After |
|-|-|
|Screenshot 2023-12-13 at 4 24 17 PM|Screenshot 2023-12-13 at 4 44 41 PM|

### Other information:

## Jetpack product discussion

Project thread: pf5801-87-p2

No.

## Testing instructions:

- Create a new post
- Make sure that two consecutive fields have a width of 50% or less
- Open the preview
- For each styles (_Default_, _Animated_, and _Outlined_), check that the fields are on the same row


monsieur-z:
I'm fixing the issue since the behavior should be the same for all styles, and that's the original one.

For some users, however, a horizontal alignment could imply that the 2 fields are somewhat related. That's an assumption we shouldn't make automatically. Editors can use the grouping feature to align fields on the same row.

andrii-lysenko (approved):
Looks good codewise and no problems while testing!

In the same time I agree with your comment regarding grouping and positioning, but, I assume many people are already used to things work in a certain way.

PR #34634, opened 2023-12-14
Fix double cookie token redirect

chrisfromthelc:
WIP related to D131982-code

Fixes https://github.com/Automattic/gold/issues/245

## Proposed changes:

* This changes the confirmation flow to that multiple cookies are not present in the browser for a free subscriber signup. It forces login upon confirmation via the "Email me a login link" flow, and once logged in, the user will see the Subscribe block in the subscribed state.

### Other information:

## Testing instructions:

* Patch D131982-code first, then rsync this patch to your sandbox.
* Add a subscribe block to a page. 
* Sign up as a free subscriber. You'll see the modal to select categories and the notification to confirm your email address.
* Go to your email and confirm the account. You'll be asked to use the "Email me a login link" flow. Log in via that link.
* You should see the Subscribe block now showing in the subscribed state (only the button, no text box). If you check your browser's cookies for the site, you should only see the remaining initial cookie for `jp-premium-content-session`. Deleting this cookie will have no effect on the state of the subscribe block.
* You should also be able to close the browser, and go back to the site; assuming you have not logged out of WPCOM, the subscribe block will still be in the subscribed state.


millerf (approved):
LGTM.

millerf:
I see an issue here. And it is my mistake...

We should go through the "login" flow (on WPCOM) only if the user is logged (so it gets redirected).

It is pretty bad to get through login while the user is actually confirming the subscription.
Screenshot 2023-12-21 at 16 08 25

So we should probably login the user (as he is identified via the subscription confirmation token) before redirecting...

Almost there...Sorry about this oversight.

Actually, once the user is logged in, we should redirect him to the site directly... The token should not interfere...

PR #34638, opened 2023-12-14
Super Cache: Fixed delete cache button trigger hook from admin panel

### LINKED ISSUE #999: Bug: Delete cache button does not trigger wpsc_after_delete_cache_admin_bar hook when used from admin panel
### Impacted plugin

Super Cache

### Quick summary

When deleting cache via button on admin panel side, hook `wpsc_after_delete_cache_admin_bar` is not triggered, as function is called via AJAX call.

### Steps to reproduce

1. Add any action to `wpsc_after_delete_cache_admin_bar`
2. Run cache delete from front site, hook will trigger properly
3. Run cache delete from admin panel
4. Hook will not fire

In that case, hook should be fired with path `/` as cache is purged for entire page then.

### What actually happened

_No response_

### Impact

All

### Available workarounds?

_No response_

### Logs or notes

_No response_

### LINKED ISSUE #34637: Bug: Delete cache button does not trigger wpsc_after_delete_cache_admin_bar hook when used from admin panel
### Impacted plugin

Super Cache

### Quick summary

When deleting cache via button on admin panel side, hook `wpsc_after_delete_cache_admin_bar` is not triggered, as function is called via AJAX call.

### Steps to reproduce

1. Add any action to `wpsc_after_delete_cache_admin_bar`
2. Run cache delete from front site, hook will trigger properly
3. Run cache delete from admin panel
4. Hook will not fire

In that case, hook should be fired with path `/` as cache is purged for entire page then.

### What actually happened

_No response_

### Impact

All

### Available workarounds?

_No response_

### Logs or notes

_No response_

Dalias96:
After that change hook wpsc_after_delete_cache_admin_bar will be fired also when pressing button from admin panel

Fixes Automattic/wp-super-cache#999

## Proposed changes:

*

### Other information:

No, it doesn't.

## Testing instructions:

* Add any action to hook `wpsc_after_delete_cache_admin_bar`
* Delete cache via button from front page
* Delete cache via buttom from admin panel
* Hook should be triggered in both scenarios

### DIFF
--- a/projects/plugins/super-cache/inc/delete-cache-button.php
+++ b/projects/plugins/super-cache/inc/delete-cache-button.php
@@ -86,8 +86,8 @@ function wpsc_admin_bar_delete_cache_ajax() {
 		if ( defined( 'WPSCDELETEERROR' ) ) {
 			return json_decode( constant( 'WPSCDELETEERROR' ) );
 		} else {
-			$req_path = !empty($_POST['path']) ? $_POST['path'] : '/';
-			$referer = !empty($_SERVER['HTTP_REFERER']) ? $_SERVER['HTTP_REFERER'] : admin_url();
+			$req_path = isset( $_POST['path'] ) ? sanitize_text_field( stripslashes( $_POST['path'] ) ) : '';
+			$referer = wp_get_referer();
 			/**
 			 * Hook into the cache deletion process after a successful cache deletion from the admin bar button.
 			 *
--- a/projects/plugins/super-cache/inc/delete-cache-button.php
+++ b/projects/plugins/super-cache/inc/delete-cache-button.php
@@ -86,6 +86,17 @@ function wpsc_admin_bar_delete_cache_ajax() {
 		if ( defined( 'WPSCDELETEERROR' ) ) {
 			return json_decode( constant( 'WPSCDELETEERROR' ) );
 		} else {
+			$req_path = !empty($_POST['path']) ? $_POST['path'] : '/';
+			$referer = !empty($_SERVER['HTTP_REFERER']) ? $_SERVER['HTTP_REFERER'] : admin_url();
+			/**
+			 * Hook into the cache deletion process after a successful cache deletion from the admin bar button.
+			 *
+			 * @since 1.12
+			 *
+			 * @param string $req_path Path of the page where the cache flush was requested.
+			 * @param string $referer  Referer URL.
+			 */
+			do_action( 'wpsc_after_delete_cache_admin_bar', $req_path, $referer );
 			return json_decode( false );
 		}
 	}


projects/plugins/super-cache/inc/delete-cache-button.php
```diff
@@ -86,6 +86,17 @@ function wpsc_admin_bar_delete_cache_ajax() {
 		if ( defined( 'WPSCDELETEERROR' ) ) {
 			return json_decode( constant( 'WPSCDELETEERROR' ) );
 		} else {
+			$req_path = !empty($_POST['path']) ? $_POST['path'] : '/';
```
donnchawp: Can you add spaces in the empty() call here, so it follows the WordPress coding conventions?  It should look like this:
`! empty( $_POST['path'] )`

It's probably worth sanitizing as well, since that's what happens when `wpsc_after_delete_cache_admin_bar` is used further down in the code.  the default path is '' in the other action, so we should probably change that too. 
The line will look like this:
`$req_path = ! empty( $_POST['path'] ) ? sanitize_text_field( stripslashes( $_POST['path'] ) ) : '';`

projects/plugins/super-cache/inc/delete-cache-button.php
```diff
@@ -86,6 +86,17 @@ function wpsc_admin_bar_delete_cache_ajax() {
...
+			$referer = !empty($_SERVER['HTTP_REFERER']) ? $_SERVER['HTTP_REFERER'] : admin_url();
```
donnchawp: It's probably safer to use `wp_get_referer()` rather than accessing $_SERVER directly.

donnchawp (changes requested):
Thank you so much for your PR to add this missing code to the delete button. There are a few small things that need to be changed and then we can get this merged.

projects/plugins/super-cache/inc/delete-cache-button.php
Dalias96: I just copied full line from code below, to keep consistency.

projects/plugins/super-cache/inc/delete-cache-button.php
Dalias96: good point, changed that :)

projects/plugins/super-cache/inc/delete-cache-button.php
```diff
@@ -86,6 +86,17 @@ function wpsc_admin_bar_delete_cache_ajax() {
...
+			/**
+			 * Hook into the cache deletion process after a successful cache deletion from the admin bar button.
+			 *
+			 * @since 1.12
```
Dalias96, line 94: @donnchawp maybe different version should be here?

Dalias96:
@donnchawp any update on the above?

projects/plugins/super-cache/inc/delete-cache-button.php
```diff
@@ -86,6 +86,17 @@ function wpsc_admin_bar_delete_cache_ajax() {
 		if ( defined( 'WPSCDELETEERROR' ) ) {
 			return json_decode( constant( 'WPSCDELETEERROR' ) );
 		} else {
+			$req_path = isset( $_POST['path'] ) ? sanitize_text_field( stripslashes( $_POST['path'] ) ) : '';
```
donnchawp, line 89: PHP Code Sniffer found some (longstanding) issues here that we should look at. The first is the nonce check. Can you add this code right at the top of the function?

```
	if ( ! isset( $_POST['path'] ) ) {
		$_POST['path'] = '/';
	}

	if ( ! isset( $_POST['nonce'] ) ) {
		wp_send_json_error( 'missing nonce.', 401 );
	}

	if ( ! wp_verify_nonce( $_POST['nonce'], 'delete-cache-' . wp_unslash( $_POST['path'] ) . '_1' ) ) { // "1" for admin
		wp_send_json_error( 'Invalid nonce.', 401 );
	}
```
That takes care of the nonce check.

Can you change `stripslashes` on line 89 to `wp_unslash`? That will satisfy the other PHPCS check.

projects/plugins/super-cache/inc/delete-cache-button.php
```diff
@@ -86,6 +86,17 @@ function wpsc_admin_bar_delete_cache_ajax() {
...
+			$referer = wp_get_referer();
```
donnchawp, line 90: PHPCS wants "=" signs to line up on consecutive lines. Can you line them up on lines 89 and 90, please?

projects/plugins/super-cache/inc/delete-cache-button.php
donnchawp: Yeah, can you change that version to this: `$$next-version$$`

donnchawp (changes requested):

projects/plugins/super-cache/inc/delete-cache-button.php
donnchawp: One more change to support returning an error. The file delete-cache-button.js needs the following code around line 24, just before the "complete:" section
```
error: function ( msg ) {
    console.log( 'Error deleting cache' );
},
```

Make sure it aligns with the rest of the code, please.

Sorry about making this patch more complicated. Jetpack has strict code checks, and they notice things that are missing.

Issue #34484, opened 2023-12-06
Top Posts Block: handle sites where Stats is inactive


jeherve:
Since the block relies on the Stats module, we would need to only make the full block available when the Stats module is enabled, and offer a way to turn on the module from the block settings when it is inactive.
We can check the different Pull Requests in #32401 to figure out examples of how this can be done.

*********

Epic: #32937

Aurorum:
Thanks! I thought I did that though - am I misunderstanding the issue?

<img width="761" alt="0F3872FD-22A7-4A13-9E6F-D683C04F14F0" src="https://github.com/Automattic/jetpack/assets/43215253/c9d33bae-93b2-4d89-a4bf-d02387782079">

jeherve:
Oh awesome, I had missed that you implemented it! 💥

PR #34639, opened 2023-12-14
Like block: inactivate block when the Likes module or related site options are disabled

### LINKED ISSUE #85156: Like block: handle cases when Likes are disabled
Currently, when the Likes are disabled, we get unexpected behavior of the Like block:

1. If Likes and Reblogs are disabled in Calypso (in _Tools → Marketing → Sharing Buttons_):

Image

The Like block is stuck at loading as can be seen over here:

Image

2. If Likes are disabled completely as Jetpack module (in _Jetpack → Settings → Sharing_):

Markup on 2023-12-12 at 16:39:23

The Like block is stuck at loading as can be seen over here:

Image

In the cases above, we:

**A) Shouldn't render anything in the front-end**

- can consider displaying information to the user in the block editor - letting them know how to enable the Like block
- can consider hiding the block from the block editor selection of blocks (the block should still stay in the post if it has been added there previously)

-- or --

**B) Make the Like block independent of the above settings completely**

In this case, we will need to make sure the Like block is loading the resources it needs (e.g. `queuehandler.js` and the `master.html` iframe).

Currently, for the Jetpack sites, if the `Likes` module is disabled, `the_content` filter isn't added and hence the `post_like()` never runs. This function is responsible for adding necessary resources for the Like button to load properly.

Image
(this is all inside `projects/plugins/jetpack/modules/likes.php`)

-- or --

**C) Maybe we could do the following steps instead:**

1. do not decouple the Like block from the Like module - since it is so dependent on it
2. if a post or template includes Like block, hide the Like widget automatically; the reason for this is that there’s currently no way to hide the Like widget without messing with the Like module - unless we introduce some new setting
  - additional issue to consider here is that there’s no easy way to check if post _template_ has certain block in it; we can check only if the post _content_ has the Like block in it currently (using `has_block()` or `parse_blocks()`); Maybe we could use `get_block_templates()` that returns all the block templates and then parse the contents of each - to search for the Like block; but it might be too much to do on every front-page load. 🤔  
3. take the steps listed in the option A as well

ivan-ottinger:
Resolves https://github.com/Automattic/wp-calypso/issues/85156.

## Proposed changes:

* check the `likes` module and `disabled_likes` and `disabled_reblogs` options to decide whether the Likes block should be registered or not.

ℹ️ Please note that only for Simple sites:
- the `likes` module is always active and the Like block will be disabled only when both site options above are `true` 
- the Reblog button is available

ℹ️ We have a separate task that will properly communicate to the user how they can enable the Like block: https://github.com/Automattic/jetpack/pull/34639#issuecomment-1855822381.

### Other information:

## Jetpack product discussion

Related project thread: pdDOJh-2JP-p2

No, it does not.

## Testing instructions:

Generally, it shouldn't be possible to reproduce the issues outlined in the description of https://github.com/Automattic/wp-calypso/issues/85156.

**Simple site**

1. Sync the changes of this PR to your simple site (e.g. by using the script from the comment below).
2. Sandbox the test site.
3. Add `define( 'JETPACK_BLOCKS_VARIATION', 'beta' );` to your `0-sandbox.php` file.
4. Add the new Like block to a post on your test site.
5. Navigate to _Tools → Marketing → Sharing Buttons_ in Calypso and try to change the _Reblog & Like_ settings:

Markup on 2023-12-14 at 13:02:21

6. The block editor and the post front-end should display the Like block in all cases apart from the one **when both buttons are disabled**. In that case, the **front-end shouldn't include any block markup** and the block editor should display the following message:

Markup on 2023-12-14 at 11:58:20 

**Self-hosted site**

6. Navigate to _Tools → Marketing → Sharing Buttons_ in Calypso or _Jetpack → Settings → Sharing_ in WP Admin and disable the related Likes toggle (both toggles enable / disable the `likes` Jetpack module).
7. If the toggles are disabled, the Like block shouldn't be available (similar as with the Simple site above).

**Atomic site** (should have the same behavior as a self-hosted site)
3. Since this new block is in beta, make sure your WordPress site has `define( 'JETPACK_BLOCKS_VARIATION', 'beta' );` line in its `wp-config.php`.
4. Add the new Like block to a post on your test site.
5. Navigate to _Tools → Marketing → Sharing Buttons_ in Calypso or _Jetpack → Settings → Sharing_ in WP Admin and disable the related Likes toggle (both toggles enable / disable the `likes` Jetpack module).
6. If **any of the two toggles is disabled**, the Like block shouldn't be available.

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-like-block-disabled-case
+++ b/projects/plugins/jetpack/changelog/fix-like-block-disabled-case
@@ -1,0 +1,4 @@
+Significance: minor
+Type: other
+
+Like block (beta): Make sure the block is not available when the Likes module or Likes / Reblog site options are disabled
--- a/projects/plugins/jetpack/extensions/blocks/like/like.php
+++ b/projects/plugins/jetpack/extensions/blocks/like/like.php
@@ -18,12 +18,16 @@ use Jetpack_Gutenberg;
  * registration if we need to.
  */
 function register_block() {
-	// The Like block is only available when the Likes module is active and either the Likes or Reblogs option is enabled.
-	if ( ! \Jetpack::is_module_active( 'likes' ) || ( get_option( 'disabled_likes' ) && get_option( 'disabled_reblogs' ) ) ) {
+	$is_wpcom = defined( 'IS_WPCOM' ) && IS_WPCOM;
+
+	$is_likes_module_inactive = ! \Jetpack::is_module_active( 'likes' );
+	$is_disabled_on_wpcom     = $is_wpcom && get_option( 'disabled_likes' ) && get_option( 'disabled_reblogs' );
+	$is_disabled_on_non_wpcom = ! $is_wpcom && get_option( 'disabled_likes' );
+	$disable_like_block       = $is_likes_module_inactive || $is_disabled_on_wpcom || $is_disabled_on_non_wpcom;
+
+	if ( $disable_like_block ) {
 		return;
 	}
-
-	$is_wpcom = defined( 'IS_WPCOM' ) && IS_WPCOM;
 
 	Blocks::jetpack_register_block(
 		__DIR__,
--- a/projects/plugins/jetpack/extensions/blocks/like/like.php
+++ b/projects/plugins/jetpack/extensions/blocks/like/like.php
@@ -18,6 +18,11 @@ use Jetpack_Gutenberg;
  * registration if we need to.
  */
 function register_block() {
+	// The Like block is only available when the Likes module is active and either the Likes or Reblogs option is enabled.
+	if ( ! \Jetpack::is_module_active( 'likes' ) || ( get_option( 'disabled_likes' ) && get_option( 'disabled_reblogs' ) ) ) {
+		return;
+	}
+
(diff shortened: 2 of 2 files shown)


simison:
I think in Jetpack folks wanted to always include the block, but show "enable module" CTA, see e.g. https://github.com/Automattic/jetpack/issues/34484

ivan-ottinger:
> I think in Jetpack folks wanted to always include the block, but show "enable module" CTA, see e.g. #34484

Thank you, Mikael! I wasn't aware about the "enable module" CTA and will look into it.

ivan-ottinger:
> I think in Jetpack folks wanted to always include the block, but show "enable module" CTA, see e.g. #34484

Since it requires a bit more work, I created a separate task for it (so we can still launch the i1 without the CTA in place):

https://github.com/Automattic/wp-calypso/issues/85291

Apart from the `likes` module activation, we can also let the users know how to change related site options (in case those are the cause of the Like block not working).

Thanks for the suggestions, Mikael! It will make the block much more user friendly.

TimBroddin (approved):
LGTM!

TimBroddin:
Rebasing this

PR #34643, opened 2023-12-14
Subscribers: pre/post-publish panel, show correct subs count with paywall

simison:
## Proposed changes:

* Show correct subs count with paywall block in the pre/post-publish panel

Previously we would show the paid post reach count, while we actually should show all subscribers -count.

Screenshot 2023-12-14 at 17 30 53

Screenshot 2023-12-14 at 17 30 58

Note "paid subscribers" as access, but email will be sent to 11 subscribers, which are free.

### Other information:

## Testing instructions:

* Create a new post
* See correct numbers for free and paid subscribers in the posts sidebar, and pre/post publish panels
* Create 2nd post with paywall, this time in pre/post publish panel you see "free" subscriber count regardless if the post is gated to "subscribers only" or "paid subscribers only".

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-correct-subs-count-with-paywall
+++ b/projects/plugins/jetpack/changelog/fix-correct-subs-count-with-paywall
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
+Subscribers: pre/post-publish panel, correct subs count with paywall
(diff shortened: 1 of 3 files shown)


simison:
In the access setting sidepanel we specifically don't pass the "has paywall" because it would change it appear as if they have 11 paid subscribers in our example:

https://github.com/Automattic/jetpack/blob/5962b36e8cf189e23f5595db44201f2eae2d53d5/projects/plugins/jetpack/extensions/shared/memberships/settings.js#L157-L172

image

lezama:
✅ tested multiple combinations local and on Simple

lezama (approved):

PR #34649, opened 2023-12-14
Protect: Add onboarding dialog

dkmyta:
## Description:
Adds onboarding dialog

## Proposed changes:

* Adds a custom `useOnboardingPopover` hook with `OnboardingRenderedContext`, used within a new `OnboardingPopover` component, implemented on the `ScanPage`
* Uses the `ActionPopover` js-packages component to add conditional onboarding flows to the Protect UI

### Other information:

## Jetpack product discussion

- https://github.com/Automattic/jetpack-scan-team/issues/331

- No

## Testing instructions:

* Checkout this branch
* Start up your Jurassic Tube and test the following flows (mobile and desktop)
* Ensure no regressions are introduced

**Free with or without threats**
- Your scan results
- Daily automated scanning
- Dismiss or Finish onboarding and verify that the popover no longer load
- Upgrade and proceed
Screen Capture on 2023-12-18 at 15-16-20

**Paid with no threats**
- Your scan results
- Daily automated and manual scanning
Screen Capture on 2023-12-18 at 16-58-42

**Paid with non-fixable threats**
- Your scan results
- Understand severity
- Daily automated and manual scanning
Screen Capture on 2023-12-18 at 15-30-35

**Paid with fixable threats**
- Your scan results
- Auto-fix all threats
- Understand severity
- Daily automated and manual scanning
- Dismiss or Finish onboarding and verify that the popovers no longer load
Screen Capture on 2023-12-18 at 16-56-20


nateweller:
Great amount of work here, it will be awesome to finally have the full onboarding experience :+1: :+1:

This PR already includes:
1. A centralized source of data for onboarding steps
2. Unique identifiers for each onboarding step

I think we can further leverage those properties to simplify some of the implementation.
1. If we saved the UID of each step completed, we can infer everything about their progress and interaction with the onboarding feature through a single stored variable.
2. If we rendered each individual onboarding popover immediately adjacent to its anchor, it would negate all of the prop drilling and potentially the need for dynamic refs and all that magic.

i.e. https://github.com/Automattic/jetpack/pull/34649/files#diff-bda6c61affe8df59ccd5c05e8f7aef66fdcb51842bdddf36fb54912e30f930f7R10

i.e. https://github.com/Automattic/jetpack/pull/27028/files#diff-aec6ac3607a0bc5a7c48849d35e9b1fe7ac05799f7e5f8bb8b77d9fd6af77c7fR10

projects/plugins/protect/src/js/components/scan-onboarding/index.jsx
```diff
@@ -0,0 +1,97 @@
...
+		4: () => {
+			if ( hasRequiredPlan && ! paidScanOnboardingDismissed ) {
+				return popoverArgs.dailyAndManualScans;
+			}
+		},
+	};
+
+	const scanOnboardingPopoverArgs = useMemo( () => {
+		return getCurrentPopoverArgs( scanOnboardingStepHandlers );
+		// eslint-disable-next-line react-hooks/exhaustive-deps
```
nateweller: In general exhaustive deps is important to ensure you're doing things The React Way, so the disables make me suspicious that the problem this memoization is solving might be avoidable higher up somehow. Could this value be provided directly from the onboarding hook instead?

projects/plugins/protect/src/js/components/scan-onboarding/sample-onboarding/index.jsx
nateweller, line 1: I think the `sample-onboarding` dir can be removed?

projects/plugins/protect/src/js/components/scan-onboarding/use-scan-popover-args.jsx
```diff
@@ -0,0 +1,142 @@
...
+	} );
+	const { recordEventHandler } = useAnalyticsTracks();
+	const getScan = recordEventHandler( 'jetpack_protect_onboarding_get_scan_link_click', run );
+
+	const handleDismiss = () => {
+		API.scanOnboardingDismissed();
+	};
+
+	return {
+		yourScanResults: createPopoverArgs( {
```
nateweller: Why is the `createPopoverArgs` pattern necessary here? If it is to inject data from the `useOnboarding` hook, could those properties be exported directly from the hook?

projects/plugins/protect/src/js/components/scan-onboarding/index.jsx
dkmyta: >In general exhaustive deps is important to ensure you're doing things The React Way, so the disables make me suspicious that the problem this memoization is solving might be avoidable higher up somehow.

I see we've previously disabled the eslint rule to avoid unnecessary re-renders and/or if we've wanted to remove items from the dep array that we're certain will not change. In this particular case, I don't believe adding the missing deps will make any significant difference, but I agree that we should follow The React Way and only disable the rule where necessary.

On further review, it appears the reason I added the disable was not for the missing deps but that I've included deps that aren't directly in the `useMemo` but that should force re-renders when they change. Perhaps there is a more appropriate way of doing so?

**UPDATE**: ✅ Done - by integrating the scanOnboardingPopoverArgs with the related methods, we can preserve the existing functionality and eliminate the necessity to disable the eslint rule.

>Could this value be provided directly from the onboarding hook instead?

I structured it this way initially to prioritize reusability. The intention was for `getCurrentPopoverArgs` to be a shared method that takes in a list of directives for how and when to return certain args, and determines which one to display based on the current state of `onboardingStep`, among other variables (`hasRequiredPlan`, `selected`, etc). My thought was that eventually, we may have another set of onboarding steps/rules for the WAF perhaps, we can develop the logic specific to those steps and pass them to this function to determine which one should be returned when, then pass that value to an `ActionPopover` component for display. That is why I have separated the logic between the hook where we have the shared methods and a component where we have methods and data specific to those onboarding flows, however, I am open to adjusting that approach to centralize things if we have no intention of reusing this.

projects/plugins/protect/src/js/components/scan-onboarding/use-scan-popover-args.jsx
dkmyta: This could probably be done better - `createPopArgs` is meant to take the unique args for each popover instance, apply all args that are shared, and return the expected format. This could likely just be used a single time once the current popover args are identified before they are passed to `ActionPopover` component.

**UPDATE**: ✅ Done - instead exported the common props directly from the hook and combined them with the unqiue props using a spread before they are passed to the component.

projects/plugins/protect/src/js/components/scan-onboarding/sample-onboarding/index.jsx
dkmyta: I kept this in here so we could further discuss it. It's certainly not the ideal way to do this, but I wanted to have something that could represent an example of how this can be reused and to document the process of using it. Again, this was strictly for the sake of reusability, and partially because the resulting component was too complex to create a storybook entry for. If we'd prefer to centralize and simplify things and move away from reusability, I'll remove it. If we want to keep that potential, what are your thoughts on how we document the usage better for posterity?

**UPDATE**: ✅ Done - removed this for now, but it is still a discussion we should have at some point before merging.

dkmyta:
>If we saved the UID of each step completed, we can infer everything about their progress and interaction with the onboarding feature through a single stored variable.

This sounds like it could be handy, but I'll admit that I am not quite certain how that could work. Unless it could reduce the complexity of the current set of checks necessary, introducing an additional variable to the mix might also complicate things further.

>If we rendered each individual onboarding popover immediately adjacent to its anchor, it would negate all of the prop drilling and potentially the need for dynamic refs and all that magic

I initially started with something similar to this but found that with the sheer number of potential scenarios that we had to take into consideration (free - no vulns, vulns, and paid - no threats, non-fixable threats, fixable threats, and toggling between the navigation list value rendering things conditionally) and the different popovers and number of popovers that had to display in each of those, the logic for determining when each popover should and shouldn't display got very complicated. I am willing to reassess that with this approach in mind but I worry that I might just end up somewhere similar.

I think perhaps I have put too much effort into making this reusable when it may have only been necessary to address this individual task. Awaiting your opinion before I begin to refactor.

projects/plugins/protect/src/js/hooks/use-onboarding/index.jsx
```diff
@@ -0,0 +1,100 @@
...
+	const completeAllCurrentSteps = useCallback( () => {
+		const stepIds = steps
+			.filter( step => {
+				// If hasRequiredPlan is true, include paid steps, else include free steps
+				return hasRequiredPlan ? step.id.startsWith( 'paid-' ) : step.id.startsWith( 'free-' );
+			} )
+			.reduce( ( carry, step ) => {
+				carry.push( step.id );
+				return carry;
+			}, [] );
```
nateweller:
```suggestion
			.map( step => step.id );
```
Nitpick - good use case for `map`.

projects/plugins/protect/src/js/components/scan-page/onboarding-steps.jsx
```diff
@@ -0,0 +1,106 @@
...
+	} );
+	const { recordEventHandler } = useAnalyticsTracks();
+	const getScan = recordEventHandler( 'jetpack_protect_onboarding_get_scan_link_click', run );
+
+	return <Button variant="link" weight="regular" onClick={ getScan } { ...props } />;
+};
+
+export default [
+	{
+		id: hasRequiredPlan ? 'paid-scan-results' : 'free-scan-results',
```
nateweller: [not blocking] One of the benefits of moving `render_callback` out of this file, and adding context to monitor mounted/rendered step components, was that the conditions for when and where each step is rendered can be left solely to the implementation/UI.

Could we alternatively add two separate objects for "paid-scan-results" and "free-scan-results", and then conditionally render them as applicable for each individually?

This way:
- the UI code defines what is shown on screen based on state;
- the hook controls what the current onboarding state actually is;
- and then this list is dedicated only to defining the content of the steps themselves.

projects/plugins/protect/src/js/hooks/use-onboarding/index.jsx
dkmyta: @nateweller I am realizing now that we could make this even more efficient and only iterate over the array a single time with something along the lines of:
```
const stepIds = steps.reduce( ( carry, step ) => {
	if ( hasRequiredPlan ? step.id.startsWith( 'paid-' ) : step.id.startsWith( 'free-' ) ) {
		carry.push( step.id );
	}
	return carry;
}, [] );
```
Does that appeal to you?

projects/plugins/protect/src/js/hooks/use-onboarding/index.jsx
```diff
@@ -0,0 +1,100 @@
...
+	}, [ currentStep, setOnboardingProgress, progress, completeOnboardingSteps ] );
+
+	/**
+	 * Complete All Current Steps
+	 */
+	const completeAllCurrentSteps = useCallback( () => {
+		const stepIds = steps
+			.filter( step => {
+				// If hasRequiredPlan is true, include paid steps, else include free steps
+				return hasRequiredPlan ? step.id.startsWith( 'paid-' ) : step.id.startsWith( 'free-' );
```
nateweller: [not blocking] Continuation from my above comment, another pro for keeping "what step displays when/where" - this logic is handy, but hidden from sight in the actual view files. Moving this to the UI level will make it (IMO) easier to understand and debug in the future.

projects/plugins/protect/src/js/components/scan-page/onboarding-steps.jsx
dkmyta: Yep! I originally started with that approach but then reconsidered repeating the same code but I can see the tradeoff here of doing that so we can leave the rendering solely to the implementation/UI.

nateweller:
Positioning

Screenshot 2024-01-16 at 2 56 41 PM

projects/plugins/protect/src/js/hooks/use-onboarding/index.jsx
dkmyta: I will look into an alternative but I believe this one might be more of a challenge than the initial instance. We need a way within the method for completing all steps to determine what steps to complete based on the current plan status so that we leave the paid flow incomplete when the free flow is completed and vice versa. We _could_ declare separate methods for completing all paid steps and one that completes all free steps?

dkmyta:
> Positioning

I'll see what can be done here but this behavior is inherent to the `ActionPopver` component. Depending on the height or width of the element it is anchored to and where it's positioned it'll flip-flop back and forth as the screen view changes. Perhaps there is a way to force it to remain in a single position.

nateweller:
I see what you mean! The `ActionPopover` has a couple weird behaviors 🤔 

Screenshot 2024-01-08 at 4 38 45 PM

dkmyta:
>The ActionPopover has a couple weird behaviors 

I mentioned this in Slack but wanted to add it here for posterity. Turns out the original positioning issue is an expected and default behavior of the core WordPress components `Popover` component and that there is a boolean `flip` property we can pass our `ActionPopover` instances to restrict the movement. More info [here](https://developer.wordpress.org/block-editor/reference-guides/components/popover/#flip-boolean)!

nateweller (dismissed):
:shipit: Looks excellent!

PR was out of date, I rebased trunk and pushed.

nateweller (approved):

PR #34650, opened 2023-12-14
Like block: remove condition that decided whether to display Like block or Like widget (Jetpack & WoA sites)

ivan-ottinger:
Partially addresses https://github.com/Automattic/wp-calypso/issues/85153
Related WPCOM patch: D132068-code

## Proposed changes:

* we are basically removing the following change: https://github.com/Automattic/jetpack/blob/ca18c7d2d7e7fef74e25f9a13a14a138b69d7270/projects/plugins/jetpack/modules/likes.php#L470 (introduced in https://github.com/Automattic/jetpack/pull/34450)

The reason for the shift in direction is that we currently cannot replace the Like widget with Like block as originally planned. More context: pdDOJh-2JP-p2#comment-2175.

### Other information:

## Jetpack product discussion

- more context: pdDOJh-2JP-p2#comment-2175
- related project thread: pdDOJh-2JP-p2

No, it does not.

## Testing instructions:

Please test with self-hosted Jetpack and WoA sites. Simple sites have their own patch: D132068-code.

**Self-hosted site**

6. Review the post in the front-end.
7. Both: Like block and Like widget should load (if https://github.com/Automattic/wp-calypso/issues/84852 was already fixed); or at least one should load and the other should be in the loading state - as can be seen in the following screenshot:

Markup on 2023-12-14 at 18:01:51
(this screenshot is from a Simple site, but it illustrates the point)

**Atomic site**
3. Since this new block is in beta, make sure your WordPress site has `define( 'JETPACK_BLOCKS_VARIATION', 'beta' );` line in its `wp-config.php`.
4. Add the new Like block to a post on your test site.
6. Review the post in the front-end.
5. Both: Like block and Like widget should load (if https://github.com/Automattic/wp-calypso/issues/84852 was already fixed); or at least one should load and the other should be in the loading state (same behavior as with the self-hosted Jetpack site).

### DIFF
--- a/projects/plugins/jetpack/changelog/remove-like-block-special-treatment
+++ b/projects/plugins/jetpack/changelog/remove-like-block-special-treatment
@@ -1,0 +1,4 @@
+Significance: minor
+Type: other
+
+Like block (beta): remove the condition that decided whether to display Like block or Like widget
--- a/projects/plugins/jetpack/modules/likes.php
+++ b/projects/plugins/jetpack/modules/likes.php
@@ -464,15 +464,7 @@ class Jetpack_Likes {
 		// Let's make sure that the script is enqueued.
 		wp_enqueue_script( 'jetpack_likes_queuehandler' );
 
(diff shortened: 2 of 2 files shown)


TimBroddin (approved):
Checks out!

PR #31695, opened 2023-07-05
tools: Avoid problems with SVN "peg revisions" in publishing

anomiex:
## Proposed changes:

SVN has a feature called "peg revisions", which allows for specifying a file at a particular revision number by appending the revision number to the filename like `file.txt@12345`. To avoid a filename like `image@2x.png` being misinterpreted as a file named `image` at revision `2x.png`, we need to append another `@` (with no revision number) to the filename passed to SVN. And fortunately it's fine to do this unconditionaly.

See https://svnbook.red-bean.com/nightly/en/svn.advanced.pegrevs.html for details.

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

p1688525237144999-slack-C01U2KGS2PQ

## Does this pull request change what data or activity we track or use?

No

## Testing instructions:

* You'd have to have a plugin tag of some sort ready to tag in SVN. Add a file with an `@` in the name to SVN trunk, then run the script to deploy the tag and see if the file gets deleted correctly. Then also do vice versa.

PR #34651, opened 2023-12-14
tools: Fix `svn status` parsing

anomiex:
## Proposed changes:

While `svn status` normally looks like it only outputs one flag followed by 7 spaces then the filename, in fact six of those seven spaces can be additional flags. Our current parsing breaks if we have any non-space flags after a space flag.

Instead, extract the seven flags, log if any of the last six aren't spaces, and then process as before.

Also it seems `tools/deploy-assets-to-svn.sh` missed the "peg revisions" fix from #31695, add that in there too.

### Other information:

## Jetpack product discussion

p1702497650047609/1702484383.442749-slack-C01U2KGS2PQ

No

## Testing instructions:

* Check that the scripts work as intended with the normal case. I don't know how to produce the unexpected flags 🤷


sdixon194 (approved):
Might as well merge and give it a go for the next release.

PR #34652, opened 2023-12-14
Improve sharing buttons events performance

### LINKED ISSUE #34581: Sharing Block: Improve buttons event handling performance
Currently, we are adding event listener to every Sharing Button on the page. In order to have a better performance we want to revert logic to being subscribed to `document` and check the event target.

----

Epic: #19694

andrii-lysenko:
Fixes https://github.com/Automattic/jetpack/issues/34581

## Proposed changes:

Instead of having listener for each button, we are now have only one on parent element. Additionally:
- Adding `id` to parent element
- Adding `service` dataset link item for easier retrieval

### Other information:

## Testing instructions:

- Add Sharing Buttons block to a post and check it's behaviour.
- View the post and check Sharing Buttons functionality.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/sharing-button/sharing-button.php
+++ b/projects/plugins/jetpack/extensions/blocks/sharing-button/sharing-button.php
@@ -44,7 +44,7 @@ function render_block( $attr, $content, $block ) {
 
 	$style_type  = $block->context['styleType'];
 	$style       = 'style-' . $style_type;
-	$data_shared = 'sharing-' . $attr['service'] . '-' . $post_id . $attr['service'];
+	$data_shared = 'sharing-' . $attr['service'] . '-' . $post_id;
 	$query       = 'share=' . $attr['service'] . '&nb=1';
 
 	$services   = get_services();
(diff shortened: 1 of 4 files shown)


projects/plugins/jetpack/extensions/blocks/sharing-button/sharing-button.php
```diff
@@ -65,7 +65,8 @@ function render_block( $attr, $content, $block ) {
 	Jetpack_Gutenberg::load_assets_as_required( __DIR__ );
 
 	$component  = '<li class="jetpack-sharing-button__list-item">';
-	$component .= '<a rel="nofollow noopener noreferrer" class="' . $sharing_link_class . '" href="' . $link_url . '" target="_blank" data-shared="' . $data_shared . '" aria-label="' . $link_aria_label . '" primary>';
+	$component .= '<a rel="nofollow noopener noreferrer" class="' . $sharing_link_class . '" href="' . $link_url . '" target="_blank" ';
```
monsieur-z: We could escape attributes here to be on the safer side, but maybe that's me being over-cautious and now fully aware of the data flow here.

monsieur-z (dismissed):
Looks good!

I left a comment you may address or discard.

projects/plugins/jetpack/extensions/blocks/sharing-button/sharing-button.php
andrii-lysenko: Thanks! Thats a good point. Do you suggest using `esc_html ` here for params?

andrii-lysenko:
@monsieur-z Thanks for suggestion. Updated.

projects/plugins/jetpack/extensions/blocks/sharing-button/sharing-button.php
monsieur-z: `esc_attr` for attributes, including `aria-label`.

monsieur-z (approved):

PR #33921, opened 2023-11-01
Subscribe block: button-only design

simison:
Part of https://github.com/Automattic/jetpack/issues/35708

## Proposed changes:

* Button-only design. Email is asked in the followed popup that opens.

Screenshot 2023-11-01 at 19 21 01

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Testing instructions:

* Go to '..'
*

PR #34659, opened 2023-12-15
Promote subscription: Add subscription modal on comment

lupus2k:
Related to https://github.com/Automattic/dotcom-forge/issues/4584

## Proposed changes:
* Add subscription modal to users after commenting.
* User can disable this feature in `/settings/newsletter/` 
* Calypso PR to enable toggle for Jetpack users: https://github.com/Automattic/wp-calypso/pull/85331
* The mc stat used for this is `jetpack-subscribe-modal-comm`

## To Do
- [x] Subscription is not matching comments toggle (sync issue?)
- [x] We need to check the message origin
- [x] Add cookie to not show > 5 times
- [x] Check if user is already subscribed
- [x] Add custom source for tracking events 
- [x] Some CSS Tweaks
- [x] Limit to Automatticians

### Other information:

## Jetpack product discussion
P2: pb5gDS-3tB-p2

Not sure

## Testing instructions:
* Follow instructions from here: https://github.com/Automattic/jetpack/pull/34659#issuecomment-1857257562
* Go to a post with a non-site user (admin or editor)
* Comment on the post
* You should see the subscription modal
* Check stats tracking in mc stats

image


projects/plugins/jetpack/modules/comments/subscription-modal-on-comment/class-jetpack-subscription-modal-on-comment.php
```diff
@@ -49,11 +47,14 @@ public static function get_block_template_part_id() {
...
-		if ( ( new Host() )->is_woa_site() && get_option( 'jetpack_verbum_subscription_modal', true ) ) {
-			add_action( 'wp_enqueue_scripts', array( $this, 'enqueue_assets' ) );
-			add_action( 'wp_footer', array( $this, 'add_subscription_modal_to_frontend' ) );
+		// Only Automattitians for now
+		if ( defined( 'AT_PROXIED_REQUEST' ) && AT_PROXIED_REQUEST ) {
+			if ( ( new Host() )->is_woa_site() && get_option( 'jetpack_verbum_subscription_modal', true ) ) {
+				add_action( 'wp_enqueue_scripts', array( $this, 'enqueue_assets' ) );
+				add_action( 'wp_footer', array( $this, 'add_subscription_modal_to_frontend' ) );
+			}
+			add_filter( 'get_block_template', array( $this, 'get_block_template_filter' ), 10, 3 );
```
lupus2k, line 55: I think we can get rid of `get_block_template` since we don't want to show anything if the feature is not enabled or is a self-hosted site.

projects/plugins/jetpack/modules/comments/comments.php
```diff
@@ -614,13 +614,13 @@ public function add_comment_meta( $comment_id ) {
...
-		if ( ! ( new Host() )->is_wpcom_platform() || ( new Host() )->is_wpcom_simple() ) {
+		// Only Automattitians for now
+		if ( ! ( defined( 'AT_PROXIED_REQUEST' ) && AT_PROXIED_REQUEST ) ) {
 			return false;
 		}
 
-		// Only Automattitians for now
-		if ( ! ( defined( 'AT_PROXIED_REQUEST' ) && AT_PROXIED_REQUEST ) ) {
+		// Not allow it to run on self-hosted or simple sites
+		if ( ! ( new Host() )->is_wpcom_platform() || ( new Host() )->is_wpcom_simple() ) {
```
lupus2k, line 618: Why not using `( new Host() )->is_woa_site()` instead?

paulopmt1: My goal was to leave two options so we can remove the left one when we release it to self-hosted. But I don't have strong feelings about using the one you suggested.

lupus2k: Same here 👍

paulopmt1 (dismissed):
Good job on this @lupus2k!

Tested on:

- Simple sites: doesn't interfere as expected.
- Self-hosted sites: doesn't interfere as expected.
- Atomic sites logged in and logged out: Working as expected on our past tests.
- Like and follow buttons are working as expected.
- Internal P2 doesn't seem to use Verbum so it doesn't apply.

candy02058912 (commented):
For the `// phpcs:ignore WordPress.Security.NonceVerification`, could we provide a short explanation of it?

I'm guessing the nonce verification is done at `pre_comment_on_post()`?

lupus2k:
> For the `// phpcs:ignore WordPress.Security.NonceVerification`, could we provide a short explanation of it?
> 
> I'm guessing the nonce verification is done at `pre_comment_on_post()`?

That's what I understood 😉

candy02058912 (dismissed):
Tested on with proxy enabled/disabled logged in/logged out:
✅ Jetpack sites are not affected
✅ WoA shows the subscription modal
✅ Simple sites are not affected

candy02058912 (dismissed):
Functional wise works as expected, but I'm a bit worried about styling issues

This is how the modal looks like with the Peirao theme.

projects/plugins/jetpack/modules/comments/subscription-modal-on-comment/class-jetpack-subscription-modal-on-comment.php
```diff
@@ -0,0 +1,224 @@
...
+		<!-- wp:heading {"textAlign":"center","style":{"typography":{"fontStyle":"normal","fontWeight":"600","fontSize":"26px"},"layout":{"selfStretch":"fit","flexSize":null},"spacing":{"margin":{"top":"4px","bottom":"10px"}}}} -->
+		<h2 class="wp-block-heading has-text-align-center" style="margin-top:4px;margin-bottom:10px;font-size:26px;font-style:normal;font-weight:600">$discover_more_from</h2>
+		<!-- /wp:heading -->
+
+		<!-- wp:paragraph {"align":"center","style":{"typography":{"fontSize":"15px"},"spacing":{"margin":{"top":"4px","bottom":"0px"}}}} -->
+		<p class='has-text-align-center' style='margin-top:4px;margin-bottom:0px;font-size:15px'>$subscribe_text</p>
+		<!-- /wp:paragraph -->
+
+		<!-- wp:group {"style":{"spacing":{"top":"32px","bottom":"32px","left":"32px","right":"32px"},"margin":{"top":"0","bottom":"0"}},"border":{"color":"#dddddd","width":"1px"}},"layout":{"type":"constrained","contentSize":"450px"}} -->
+		<form class="jetpack-subscription-modal__form">
```
simison, line 168: Why not just render subscribe block directly here? 

You're replicating a lot of styles and functionality that you'd get for free.

You could extend block's parameters to allow passing pre-defined email.

simison: That would also allow the block to inherit theme styles directly, so it'll match everything else on the site better.

simison: Example:

https://github.com/Automattic/jetpack/blob/e849efbc01ccb083925969dd38994c80b242a843/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php#L1058

projects/plugins/jetpack/modules/comments/subscription-modal-on-comment/subscription-modal.css
```diff
@@ -0,0 +1,153 @@
+.jetpack-subscription-modal {
```
simison, line 1: This is pretty close to "Subscription modal" feature's CSS, so I'd try add something that clarifies it's separate:

https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.css#L1

Honestly I'd just use the same modal, though. :-)

projects/plugins/jetpack/modules/comments/subscription-modal-on-comment/subscription-modal.css
```diff
@@ -0,0 +1,153 @@
...
+	visibility: hidden;
+	position: fixed;
+    z-index: 50000;
```
simison, line 4: Lots of mixed tabs/spacing in this file.

projects/plugins/jetpack/modules/comments/subscription-modal-on-comment/subscription-modal.css
```diff
@@ -0,0 +1,153 @@
...
+	bottom: 0;
+}
+
+/*
+ * These text-wrap properties still have limited browser
+ * support, but based on feedback still adding them for when
+ * they are supported.
+ */
+.jetpack-subscription-modal__modal-content p {
+	text-wrap: balance;
```
simison: I think you could apply `balance` for the modal's heading, and leave `pretty` for the `p`. I think it's fine to not stack them together as browser support is pretty good at this point, and only Firefox is lacking (but presumably not for long).

More in p58i-g2L-p2

projects/plugins/jetpack/modules/comments/subscription-modal-on-comment/subscription-modal.js
```diff
@@ -0,0 +1,173 @@
+document.addEventListener( 'DOMContentLoaded', function () {
+	const modal = document.getElementsByClassName( 'jetpack-subscription-modal' )[ 0 ];
```
simison, line 2: `[0]` might fatal if the element doesn't exist?

Actually seeing same in [here](https://github.com/Automattic/jetpack/blob/trunk/projects/plugins/jetpack/modules/subscriptions/subscribe-modal/subscribe-modal.js#L4).

projects/plugins/jetpack/modules/comments/subscription-modal-on-comment/subscription-modal.js
```diff
@@ -0,0 +1,173 @@
...
+
+	function reloadOnCloseSubscriptionModal( customUrl ) {
+		const destinationUrl = customUrl ? new URL( customUrl ) : new URL( redirectUrl );
+
+		// Prevent redirect to external sites.
+		if ( destinationUrl.hostname !== window.location.hostname ) {
+			return;
+		}
+
+		localStorage.setItem( 'jetpack-subscription-modal-goto', destinationUrl.hash );
```
simison: I'd add more specificity to name not to confuse with the other "subscribe modal" functionality.

projects/plugins/jetpack/modules/comments/subscription-modal-on-comment/subscription-modal.js
```diff
@@ -0,0 +1,173 @@
...
+			const iframeElement = document.querySelector( '.jetpack-subscription-modal__iframe' );
+			if ( data && data.action === 'close' ) {
+				window.removeEventListener( 'message', handleSubscriptionModalIframeResult );
+				iframeElement.src = 'about:blank';
+				reloadOnCloseSubscriptionModal( subscriptionData.url );
+			}
+		}
+	}
+
+	function showSubscriptionIframe( subscriptionData ) {
```
simison, line 41: I think it's problematic to duplicate these features here, as it becomes even harder to maintain since they're duplicated in many places.

Could we just re-use subscription block's existing functionality here?

projects/plugins/jetpack/modules/comments/subscription-modal-on-comment/class-jetpack-subscription-modal-on-comment.php
lupus2k: I worked on the custom input since we need three features not included in the block:
- Predefined email
- Read only input (if user is logged-in)
- Custom param (app_source) for tracking purpose

Do you think we can extend the block to use this features instead of what we did?

projects/plugins/jetpack/modules/comments/subscription-modal-on-comment/subscription-modal.js
lupus2k: Most of your suggestions will be fixed by using the subscription block as mentioned here (https://github.com/Automattic/jetpack/pull/34659#discussion_r1442843758f). If you think it's a good idea, I can work on this today.

projects/plugins/jetpack/modules/comments/subscription-modal-on-comment/class-jetpack-subscription-modal-on-comment.php
simison: Totally! :-)

We're already adding the custom tracking source here: https://github.com/Automattic/jetpack/pull/34792 (help testing welcome!). It works with D133298-code.

I think it would be good to add another attribute which passes the email address to the block.

"Button only" style could also be helpful, I could finish this next week or you can also take over: https://github.com/Automattic/jetpack/pull/33921

projects/plugins/jetpack/modules/comments/subscription-modal-on-comment/subscription-modal.js
simison: Yeah so as an overall advice I'd look to:

- Consolidate with subscribe block as much as possible
- Consolidate styles/JS with [subscriber modal feature](https://github.com/Automattic/jetpack/tree/83b1208519d32ca61a1f6abad0b7f246a80a928c/projects/plugins/jetpack/modules/subscriptions/subscribe-modal) where you see it makes sense — sometimes it makes sense to copy trivial things, sometimes makes sense to share. It really depends.

All this is fine in follow-ups too, if you'd prefer to get this PR merged already now to get some user-testing/numbers over the weekend. All this is pretty much under-the hood, after all.

projects/plugins/jetpack/modules/comments/subscription-modal-on-comment/class-jetpack-subscription-modal-on-comment.php
lupus2k: That's interesting, thanks 🙇 

Since I didn't know the user's email nor was logged in until I received the comment response, I can update the email address and make the input read-only using javascript like here: https://github.com/Automattic/jetpack/pull/34659/files#diff-3381b2180b188f8bfc6654da80950e0057d5f34b7b4b02eef11f216d33f56360R86

I think we can make it work using only this PR: https://github.com/Automattic/jetpack/pull/34792

projects/plugins/jetpack/modules/comments/subscription-modal-on-comment/subscription-modal.js
lupus2k: I just pushed some little fixes and I'm gonna merge it.
I'll create a follow-up to re-use the subscribe block instead of our custom one.

Thanks for all your feedback @simison

lupus2k: I asked for a new review since I need it after the last changes 🤦‍♂️

DustyReagan (approved):
I tested the functionality on Atomic, and it's working well! It appears to behave just like it does on Simple sites.

PR #34661, opened 2023-12-15
Added preemptive check full sync posts processing

darssen:
## Proposed changes:

This is a follow-up from https://github.com/Automattic/jetpack/pull/34390. Small performance improvement which should solve some memory issues. I added a check to break the loop of adding metadata if we will not add the post anyway.

### Other information:

## Testing instructions:

Same as in https://github.com/Automattic/jetpack/pull/34390


fgiannar (dismissed):
Makes sense 👍

fgiannar (approved):
Re-approving after version bump

PR #34664, opened 2023-12-15
Like block: remove like module dependency

TimBroddin:
## Proposed changes:
- This removes the like block's dependency on the like module being enabled
- It ensures the master iframe gets inserted once
- It doesn't render anything when there's no post_id present 

### Other information:

## Testing instructions:

Depends on this PR for WPCOM: D132336-code 

- Check out the PR and build it.
- Since this new block is in beta, make sure your WordPress site has `define( 'JETPACK_BLOCKS_VARIATION', 'beta' );` line in its `wp-config.php`.
- Launch Jurassic Tube to make your local site available publicly.
- Connect Jetpack to your WordPress.com account and enable Likes feature.
- Add the new Like block to a post on your test site.

Now try different scenarios:
- Go to Jetpack → Settings → Sharing & toggle on/off the Likes
- The master iframe should be inserted once if there's a Like widget/block present

Also test on Simple site, the only difference is the location of the toggle: Tools > Marketing > Sharing Buttons

⚠️ The actual rendering of the widget depends on an API call, which checks among other things ... if the module is enabled. We'll tackle this in a follow up PR. For now  return `true` here: fbhepr%2Skers%2Sjcpbz%2Schoyvp.ncv%2Serfg%2Sjcpbz%2Qwfba%2Qraqcbvagf%2Spynff.jcpbz%2Qwfba%2Qncv%2Qyvfg%2Qcbfg%2Qyvxrf%2Qraqcbvag.cuc%2367%2Q68-og  ⚠️

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/like/like.php
+++ b/projects/plugins/jetpack/extensions/blocks/like/like.php
@@ -20,15 +20,6 @@ use Jetpack_Gutenberg;
 function register_block() {
 	$is_wpcom = defined( 'IS_WPCOM' ) && IS_WPCOM;
 
-	$is_likes_module_inactive = ! \Jetpack::is_module_active( 'likes' );
-	$is_disabled_on_wpcom     = $is_wpcom && get_option( 'disabled_likes' ) && get_option( 'disabled_reblogs' );
-	$is_disabled_on_non_wpcom = ! $is_wpcom && get_option( 'disabled_likes' );
-	$disable_like_block       = $is_likes_module_inactive || $is_disabled_on_wpcom || $is_disabled_on_non_wpcom;
-
-	if ( $disable_like_block ) {
-		return;
-	}
-
 	Blocks::jetpack_register_block(
 		__DIR__,
 		array(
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/like/like.php
```diff
@@ -72,6 +67,19 @@ function render_block( $attr, $content, $block ) {
 	 */
 	$new_layout = apply_filters( 'likes_new_layout', true ) ? '&amp;n=1' : '';
 
+	static $master_iframe_added = false;
```
obenland: Can we call this `$primary_iframe_added`? ([context](https://www.wired.com/story/tech-confronts-use-labels-master-slave/))

TimBroddin: I've updated the PR.

I've another commit ready to push that replaces the name of the function `jetpack_likes_master_iframe` to `jetpack_likes_main_iframe` and updates all references. Is it ok to do this in one go? Or is it too risky, as we might break other plugins that are somehow relying on this function to be there? (*)

(* we could stub it)

obenland: I'd do that separately and deprecate the function so it doesn't break things

JuanLucha:
Hi!
I...
... applied this patch to my sandbox: D132336-code
... set true as the return value of fbhepr%2Skers%2Sjcpbz%2Schoyvp.ncv%2Serfg%2Sjcpbz%2Qwfba%2Qraqcbvagf%2Spynff.jcpbz%2Qwfba%2Qncv%2Qyvfg%2Qcbfg%2Qyvxrf%2Qraqcbvag.cuc%3Se%3Q62rs4q03%2367%2Q68-og
... was switching on/off the Likes toggle in my site.

Everything works great:

- With LIkes ON:
Screenshot 2023-12-18 at 12 06 19
Screenshot 2023-12-18 at 12 09 34

- With Likes OFF:
Screenshot 2023-12-18 at 12 07 00
...and no master iframe.

Great job!

JuanLucha (approved):

PR #34636, opened 2023-12-14
Social: Fix register_settings namespace php error

gmjuhasz:
With the Social store refactor we introduced new registered settings, where to be able to use core API endpoints. That change caused some issues on Atomic sites, this change fixes that.

With a lot of debugging @pablinos found out that the namespace of the settings was set to `general` which is a problem in this case, as any change on `wp-admin/options-general.php` will set our `jetpack_social_autoconvert_images` option to an empty string.

## Proposed changes:

* Updated the namespace of the options
* Fixed a possible bug where the type was set incorrectly
* Added a guard clause so the currently messed up options get restored to the defaults

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

Fixes #34627
## Does this pull request change what data or activity we track or use?

No.
## Testing instructions:

* On trunk Go to your local site on `wp-admin/options-general.php` and click on Save. It should set the option to empty string. Loading the editor will cause the issue.
* Apply this patch. Loading the editor should be fine, as the option should get reset to the defaults.
* Do the first step again on this branch, and confirm that it still works

PR #34666, opened 2023-12-15
Social: Fix auto-conversion cleanup logic

gmjuhasz:
Fixes an issue where a delete_option call is getting called on every API call with Social.

## Proposed changes:

* Fixed the logic so we are mindful of the option being absent and do not try to delete it if it's not there.

### Other information:

No.
## Testing instructions:

*  Without this PR sandbox Jetpack SUN. Looking at the logs while hitting `https://public-api.wordpress.com/` should call https://github.com/Automattic/jetpack/pull/34636/files#diff-fcaff5fe90306f15c873c21bc21a7f091da0b003f9982ac57e16bf52aa90752cR58 delete_option on every call. You can put a log in there to see it in the logs
* Apply this PR with the jetpack-downloader to your sandbox. Logging in the same place should not result in any logs on calling the api.

### DIFF
--- a/projects/packages/publicize/changelog/fix-social-auto-conversion-cleanup-logic
+++ b/projects/packages/publicize/changelog/fix-social-auto-conversion-cleanup-logic
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Social: Fixed issue with auto-conversion option logic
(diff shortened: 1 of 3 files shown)


spsiddarthan (approved):

spsiddarthan (approved):
Works as expected!

PR #34667, opened 2023-12-15
Contact Form: add accessible name to form

### LINKED ISSUE #34504: Enhancement: [Contact Form] Add accessible name to forms
### Impacted plugin

Jetpack

### What

Add an accessible name to forms so screen reader users can identify them more easily

### How

A possible solution is to add an entry in the editor for users to name the form. However, we can't rely on that entry always being filled. We could develop heuristics for a fallback, such as getting the nearest heading or page title.

monsieur-z:
Fixes https://github.com/Automattic/jetpack/issues/34504

## Proposed changes:

Forms with an accessible name are easier to discover and access with a screen reader. One way to define an accessible name is adding an `aria-label` attribute to the `form`, the value being the name.

Adding a title to a form, visible to all users, is considered a best practice (see [this MDN article](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/form_role#description)). We'll therefore look for headings to determine a form accessible name. The heuristic is:
1. Look for the first heading inside the form
2. Look for a heading preceding the form
3. Use the post title

The strategy here is to save the accessible name in a block attribute (recycling the existing `formTitle` attribute, which was declared but wasn't used) as the content of the page changes in the editor. When the page is rendered, we simply get the attribute.

In the editor, we've added a warning notice to nudge users to add a heading to the form.

### Other information:

## Jetpack product discussion

Project thread: pf5801-87-p2

No.

### Prerequisites
- Create a new post
- Open the preview in a separate tab

### Notice
- Select the _Contact Form_ in the editor
- Notice the warning notice in the sidebar
Screenshot 2023-12-19 at 4 15 58 PM

### Heading Inside Form
- Add a heading inside the form
- Notice the warning has disappeared
- Save the post and refresh the preview
- Notice the `form`  has an `aria-label` attribute of which the value is the heading content
- Remove the heading

### Heading Before Form
- Add a heading just before the form
- Save the post and refresh the preview
- Notice the `form`  has an `aria-label` attribute of which the value is the heading content
- Remove the heading

### No Heading
- Notice the `form`  has an `aria-label` attribute of which the value is the post title, followed by `Form`

## Captures
_Form in the editor_
Screenshot 2023-12-19 at 12 35 22 PM

_Form in the site_
Screenshot 2023-12-19 at 12 35 32 PM

_The VoiceOver screen reader shows the form in the landmarks shortcut_
Screenshot 2023-12-19 at 12 36 47 PM

### DIFF
--- a/projects/packages/forms/src/blocks/contact-form/components/jetpack-manage-responses-settings.js
+++ b/projects/packages/forms/src/blocks/contact-form/components/jetpack-manage-responses-settings.js
@@ -1,5 +1,5 @@
 import { getJetpackData } from '@automattic/jetpack-shared-extension-utils';
-import { Button, TextControl } from '@wordpress/components';
+import { Button } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 import { get } from 'lodash';
 import React from 'react';
@@ -7,11 +7,7 @@ import InspectorHint from '../components/inspector-hint';
 
 const RESPONSES_PATH = `${ get( getJetpackData(), 'adminUrl', false ) }edit.php?post_type=feedback`;
 
-const JetpackManageResponsesSettings = ( {
-	formTitle = '',
-	isChildBlock = false,
-	setAttributes,
-} ) => {
+const JetpackManageResponsesSettings = () => {
 	return (
 		<>
 			<InspectorHint>
@@ -23,15 +19,6 @@ const JetpackManageResponsesSettings = ( {
 					{ __( '(opens in a new tab)', 'jetpack-forms' ) }
 				</span>
 			</Button>
-			{ /* Temporarily hiding the Form Title */ }
-			{ false && ! isChildBlock && (
-				<TextControl
-					label={ __( 'Title of the Form', 'jetpack-forms' ) }
-					value={ formTitle }
-					onChange={ value => setAttributes( { formTitle: value } ) }
-					help={ __( 'Optional - not visible to viewers', 'jetpack-forms' ) }
-				/>
-			) }
 		</>
 	);
 };
--- a/projects/packages/forms/src/contact-form/class-contact-form.php
+++ b/projects/packages/forms/src/contact-form/class-contact-form.php
@@ -240,6 +240,8 @@ class Contact_Form extends Contact_Form_Shortcode {
 	 * @return string HTML for the concat form.
 	 */
 	public static function parse( $attributes, $content ) {
+		global $post;
+
 		if ( Settings::is_syncing() ) {
 			return '';
 		}
@@ -352,7 +354,7 @@ class Contact_Form extends Contact_Form_Shortcode {
 			$url                     = apply_filters( 'grunion_contact_form_form_action', "{$url}#contact-form-{$id}", $GLOBALS['post'], $id );
 			$has_submit_button_block = str_contains( $content, 'wp-block-jetpack-button' );
 			$form_classes            = 'contact-form commentsblock';
-			$form_accessible_name    = $attributes['formTitle'];
+			$form_accessible_name    = $attributes['formTitle'] ? $attributes['formTitle'] : $post->post_title;
 			$form_aria_label         = isset( $form_accessible_name ) && ! empty( $form_accessible_name ) ? 'aria-label="' . esc_attr( $form_accessible_name ) . '"' : '';
 
 			if ( $has_submit_button_block ) {
--- a/projects/packages/forms/src/contact-form/class-contact-form.php
+++ b/projects/packages/forms/src/contact-form/class-contact-form.php
@@ -357,7 +357,7 @@ class Contact_Form extends Contact_Form_Shortcode {
 				$form_classes .= ' wp-block-jetpack-contact-form';
 			}
 
-			$r .= "<form role='form' action='" . esc_url( $url ) . "' method='post' class='" . esc_attr( $form_classes ) . "' novalidate>\n";
+			$r .= "<form role='form' action='" . esc_url( $url ) . "' method='post' class='" . esc_attr( $form_classes ) . "' " . $form->get_form_label( $content ) . " novalidate>\n";
 			$r .= $form->body;
 
 			// In new versions of the contact form block the button is an inner block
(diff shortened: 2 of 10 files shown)


projects/packages/forms/src/contact-form/class-contact-form.php
```diff
@@ -1741,4 +1741,95 @@ public static function get_block_alignment_class( $attributes = array() ) {
...
+	/**
+	 * Get the form's accessible name
+	 *
+	 * @param string $content The form's content.
+	 * @return string The form's accessible name.
+	 */
+	public function get_form_label( $content ) {
+		global $post;
+
+		$headings = array_map(
```
andrii-lysenko: Maybe just use list like ["h1", "h2" ...] for easier reading?

andrii-lysenko (commented):
Generally, the code looks good to me. I've left just small comment.
I do know that `Document and loadHTML ` are quite costly, especially one we are having inside the loop, but I'm not too strong in PHP to evaluate the impact. I would prefer some additional eyes on this and maybe ideas avoiding using costly functions.

projects/packages/forms/src/contact-form/class-contact-form.php
```diff
@@ -1741,4 +1741,95 @@ public static function get_block_alignment_class( $attributes = array() ) {
...
+			},
+			range( 1, 6 )
+		);
+
+		// 1. Look for a heading inside the form
+
+		$dom = new \DOMDocument();
+		$dom->loadHTML( $content );
+
+		foreach ( $headings as $heading ) {
```
darssen: Maybe refactor this into a method since we are using it in Step 2 as well.
```php
public function get_aria_label_based_on_headings(DOMDocument $dom, array $headings): string {
        foreach ($headings as $heading) {
            $heading_element = $dom->getElementsByTagName($heading)->item(0);

            if (isset($heading_element)) {
                $id = $heading_element->getAttribute('id');

                if (isset($id) && !empty($id)) {
                    return "aria-labelledby='$id'";
                }

                return "aria-label='" . $heading_element->textContent . "'";
            }
        }
        return '';
    }
```
And then check for it not being empty to return.

projects/packages/forms/src/contact-form/class-contact-form.php
```diff
@@ -1741,4 +1741,95 @@ public static function get_block_alignment_class( $attributes = array() ) {
...
+				return "aria-label='$heading->textContent'";
+			}
+		}
+
+		// 2. Look for a heading as a previous sibling
+
+		if ( has_blocks( $post->post_content ) ) {
+			$blocks = parse_blocks( $post->post_content );
+			// Filter out empty blocks
+			$blocks = array_values(
```
darssen: Not sure I would loop through the array to clean it if we are going to loop it anyway with below's `foreach`.

projects/packages/forms/src/contact-form/class-contact-form.php
```diff
@@ -1741,4 +1741,95 @@ public static function get_block_alignment_class( $attributes = array() ) {
...
+					}
+				)
+			);
+
+			foreach ( $blocks as $index => $block ) {
+				if ( $block['blockName'] !== 'jetpack/contact-form' ) {
+					continue;
+				}
+
+				$prev_block = $blocks[ $index - 1 ];
```
darssen: Would this one give an error if `jetpack/contact-form` is the 1st in the `$blocks` array?

darssen (commented):
Many thanks for working on this and making Jetpack more accessible @monsieur-z! 
Code looks good aside from a few comments that might help.

On the other hand, I am with @andrii-lysenko that `Document and loadHTML`  might be very costly, but can't come up with a different solution.

monsieur-z:
Thank you both for your reviews! I've opted for a new strategy: updating the form accessible name as the post content changes. We can then just retrieve the attribute when we render the page.

projects/packages/forms/src/blocks/contact-form/components/jetpack-manage-responses-settings.js
```diff
@@ -23,15 +19,6 @@ const JetpackManageResponsesSettings = ( {
 					{ __( '(opens in a new tab)', 'jetpack-forms' ) }
 				</span>
 			</Button>
-			{ /* Temporarily hiding the Form Title */ }
```
monsieur-z, line 26: Remove this altogether. A title should be defined in the markup and be visible to all users.

projects/packages/forms/src/contact-form/class-contact-form.php
```diff
@@ -352,12 +352,14 @@ public static function parse( $attributes, $content ) {
 			$url                     = apply_filters( 'grunion_contact_form_form_action', "{$url}#contact-form-{$id}", $GLOBALS['post'], $id );
 			$has_submit_button_block = str_contains( $content, 'wp-block-jetpack-button' );
 			$form_classes            = 'contact-form commentsblock';
+			$form_accessible_name    = $attributes['formTitle'];
```
monsieur-z: Recycling `formTitle` here. Is there any side effect to renaming a block attribute?

projects/packages/forms/src/blocks/contact-form/components/jetpack-manage-responses-settings.js
darssen: Yes, we should be good by removing it. Not sure how long ago that temporarily was there :)

projects/packages/forms/src/contact-form/class-contact-form.php
darssen: Not sure here. We are not using it anywhere else, so looks good. But unsure if there is some type of contract for blocks and their attributes.

darssen (commented):
I really like the new approach since we are only doing the heavy looping to find a heading when editing. 
I tested on self-hosted and it tested good.
Not very familiar with the frontend so will let @andrii-lysenko to review as well for approval.

andrii-lysenko (approved):
I agree with @darssen that this approach is much better and I don't have performance concerns. It also tests well for me on self-hosted and simple 👍

PR #34673, opened 2023-12-15
Sharing buttons block alternate settings screen

### LINKED ISSUE #34326: Sharing: add notice in wp-admin Sharing settings to explain that this doesn't impact the sharing block
In WP-admin > Settings > Sharing, when the sharing module is active and the site uses a block-based theme, we should gray out sharing settings and display a notice inviting folks to go to the site editor to configure sharing on their site using the new sharing block.

image

Related: #34327

Epic: #19694

andrii-lysenko:
Fixes https://github.com/Automattic/jetpack/issues/34326

## Proposed changes:

This PR changes the view of Sharing Settings menu (/wp-admin/options-general.php?page=sharing) when the block is available and block-based theme is used. Following the approach in https://github.com/Automattic/jetpack/pull/34619.

New view should appear like this:
Screenshot 2023-12-15 at 11 00 01

Note: `Learn how to add Sharing Buttons` points to page still not ready.

### Other information:

## Testing instructions:

- Checkout changes locally, build `jetpack build plugins/jetpack` and spin up your site locally
- Check `/wp-admin/options-general.php?page=sharing`.
- You should see new view only when beta blocks enabled: `define( 'JETPACK_BLOCKS_VARIATION', 'beta' );`
- For block-based theme (eg Twenty Twenty-Four) you should see new view.
- For older themes (eg Twenty Ten) old view should appear

### DIFF
--- a/projects/plugins/jetpack/modules/sharedaddy/sharing.php
+++ b/projects/plugins/jetpack/modules/sharedaddy/sharing.php
@@ -8,6 +8,7 @@
 // phpcs:disable Universal.Files.SeparateFunctionsFromOO.Mixed -- TODO: Move classes to appropriately-named class files.
 
 use Automattic\Jetpack\Assets;
+use Automattic\Jetpack\Redirect;
 use Automattic\Jetpack\Status;
 
 if ( ! defined( 'WP_SHARING_PLUGIN_URL' ) ) {
@@ -686,14 +687,12 @@ class Sharing_Admin {
 
 	<!-- Showing Sharing Buttons Block message -->
 	<?php elseif ( current_user_can( 'manage_options' ) ) : ?>
-		
 		<?php
-			$sharer            = new Sharing_Service();
 			$showcase_services = array(
-				$sharer->get_service( 'twitter' ),
-				$sharer->get_service( 'facebook' ),
-				$sharer->get_service( 'email' ),
-				$sharer->get_service( 'reddit' ),
+				new Share_Tumblr( 'tumblr', array() ),
+				new Share_Facebook( 'facebook', array() ),
+				new Share_Email( 'email', array() ),
+				new Share_Reddit( 'reddit', array() ),
 			);
 			?>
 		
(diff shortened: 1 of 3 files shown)


projects/plugins/jetpack/modules/sharedaddy/sharing.php
```diff
@@ -676,6 +685,50 @@ public function management_page() {
...
+						<a href="<?php echo esc_url( admin_url( 'site-editor.php?path=%2Fwp_template' ) ); ?>" class="button button-primary">
+							<?php esc_html_e( 'Go to the site editor', 'jetpack' ); ?>
+						</a>
+						<a href="<?php echo esc_url( Redirect::get_url( 'jetpack-support-sharing-block' ) ); ?>" class="button" target="_blank" rel="noopener noreferrer">
+							<?php esc_html_e( 'Learn how to add Sharing Buttons', 'jetpack' ); ?>
+						</a>
+					</div>
+				</div>
+				<div>
+					<p><?php esc_html_e( 'Sharing Buttons example: ', 'jetpack' ); ?></p>
```
jeherve:
```suggestion
					<p><?php esc_html_e( 'Sharing Buttons example:', 'jetpack' ); ?></p>
```

projects/plugins/jetpack/modules/sharedaddy/sharing.php
```diff
@@ -376,7 +377,15 @@ public function management_page() {
 		do_action( 'pre_admin_screen_sharing' );
 		?>
 
-		<?php if ( current_user_can( 'manage_options' ) ) : ?>
+		<?php
+			$block_availability = Jetpack_Gutenberg::get_cached_availability();
+			$is_block_available = (bool) isset( $block_availability['sharing-buttons'] ) && $block_availability['sharing-buttons']['available'];
+			$current_theme      = wp_get_theme();
+			$is_block_theme     = $current_theme->is_block_theme();
```
jeherve: I think this can be simplified to `wp_is_block_theme()`.

projects/plugins/jetpack/modules/sharedaddy/sharing.php
```diff
@@ -676,6 +685,50 @@ public function management_page() {
 	</div>
 	</div>
 
+	<!-- Showing Sharing Buttons Block message -->
+	<?php elseif ( current_user_can( 'manage_options' ) ) : ?>
```
jeherve: Since we're making changes here, maybe we could extract this new logic to its own method, and extract the legacy settings to its own method as well? It would keep `management_page()` a bit shorter, and would allow us to call the `Sharing_Service()` class only when necessary for example.

jeherve (commented):
Like in #34619, I think we should add an invitation to deactivate the sharing module when it is active but not needed (the sharing block is available and the site uses a block-based theme).

Also like in #34619, maybe there should be an option to stick with the legacy buttons. Let's see what @keoshi thinks in the other PR. 👍

projects/plugins/jetpack/modules/sharedaddy/sharing.php
andrii-lysenko: Thanks for the suggestions. Updated.

jeherve (approved):
Let's merge this to keep moving, but my questions from https://github.com/Automattic/jetpack/pull/34673#pullrequestreview-1786354226 remain. I think we should address them in a follow-up PR.

@andrii-lysenko Could you keep track of those to-dos in an issue?

PR #34682, opened 2023-12-16
Connection: add IDC secret compare endpoint

sergeymitr:
## Proposed changes:
* Add the "compare secrets" endpoint.
* Always send IDC secret if persistent blog ID is in place.

### Other information:

## Jetpack product discussion
D129990-code

No.

## Testing instructions:
The diff does not bring functional changes, so just confirm the connection flow works fine.


projects/packages/identity-crisis/src/class-rest-endpoints.php
```diff
@@ -219,6 +237,31 @@ public static function fetch_url_secret() {
...
+	 * @return WP_Error|\WP_REST_Response
+	 */
+	public static function compare_url_secret( $request ) {
+		$match = false;
+
+		$storage = new URL_Secret();
+
+		if ( $storage->exists() ) {
+			$remote_secret = $request->get_param( 'secret' );
+			$match         = $remote_secret && $storage->get_secret() === $remote_secret;
```
anomiex: Drive-by comment: Should we use [hash_equals](https://www.php.net/hash_equals) here?

sergeymitr: Good point, I'll use that 👍

projects/packages/identity-crisis/src/class-rest-endpoints.php
```diff
@@ -233,4 +276,20 @@ public static function url_secret_permission_check() {
...
+
+	/**
+	 * The endpoint is only available on non-connected sites.
+	 * use `/identity-crisis/url-secret` for connected sites.
+	 *
+	 * @return true|WP_Error
+	 */
+	public static function compare_url_secret_permission_check() {
+		return ( new Connection_Manager() )->is_connected()
+			? new WP_Error(
```
fgiannar, line 288: Another drive-by comment: The error code and message here can be misleading, since this is not the result of a permissions' check. Maybe we could replace it with a more relevant `WP_Error`?

projects/packages/identity-crisis/src/class-rest-endpoints.php
```diff
@@ -74,6 +74,24 @@ public static function initialize_rest_api() {
 				'permission_callback' => array( static::class, 'url_secret_permission_check' ),
 			)
 		);
+
+		// Fetch URL verification secret.
+		register_rest_route(
+			'jetpack/v4',
+			'/identity-crisis/compare-url-secret',
+			array(
+				'methods'             => WP_REST_Server::EDITABLE,
```
fgiannar, line 83: Would it maybe make more sense for this to be a `GET` request? We are not modifying anything, right? Just comparing the secrets.

sergeymitr: Hi @fgiannar, thanks for taking a look!
The idea here is to keep the secret from showing up in the access log, which will happen if we use GET.
So it may be not be quite REST-ful, but safer that way 🙂

projects/packages/identity-crisis/src/class-rest-endpoints.php
sergeymitr: Good point, fixed 👍

bindlegirl (approved):
As far as I can tell, this works

PR #34688, opened 2023-12-18
Fully implement` jetpack_is_potential_blogging_site` function

roo2:
Currently, in wpcom, we have some features which we think only make sense for "bloggers" but not for people building business promotion websites or eCommerce sites. The best example of this is the "blogging prompts" feature.

Currently we only rely on the site_intent, which is a wpcom `site_option` set in onboarding. But this is quite insufficient as I investigated in this post pe7F0s-1ql-p2.

This change adds to the existing `jetpack_is_potential_blogging_site` function to make it as robust as possible to determine whether a site is likely to be a blog or not.

Currently, this function is unused, but I intend on first applying it to the blogging prompts feature in wpcom.

I chose to add this to jetpack rather than wpcom, because I believe it will be a useful function for jetpack features, I can see that `launchpad` in particular may be able to use this function as it is currently checking for `site_intent === 'write'` which was the original method for testing for a blogging site.

### Testing instructions.

I tested this by pushing to my wpcom sandbox and testing via wpsh:
```
pnpm install
jetpack install plugins/jetpack
jetpack build plugins/jetpack
# after setting up `wpcom-sandbox` in my ~/.ssh/config
jetpack rsync jetpack wpdev@wpcom-sandbox:~/public_html/wp-content/mu-plugins/jetpack-plugin/production
```

in wpsh:

```
switch_to_blog(227131231);
return jetpack_is_potential_blogging_site();
restore_current_blog();
```

I tested this with sites I created myself and recently created wpcom sites.

N/A

### DIFF
--- a/projects/plugins/jetpack/_inc/blogging-prompts.php
+++ b/projects/plugins/jetpack/_inc/blogging-prompts.php
@@ -334,6 +334,7 @@ function _jetpack_show_posts_on_front() {
 	if ( 'posts' === get_option( 'show_on_front' ) ) {
 		return true;
 	}
+	return false;
 }
 
 /**
--- a/projects/plugins/jetpack/_inc/blogging-prompts.php
+++ b/projects/plugins/jetpack/_inc/blogging-prompts.php
@@ -259,10 +259,102 @@ function jetpack_is_new_post_screen() {
 }
 
 /**
+ * Determines if the user selected a "blogging" intent or goal when creating the site.
+ *
+ * @return bool
+ */
+function _jetpack_has_selected_blogging_goal_or_intent() {
+	$intent = get_option( 'site_intent', '' );
+	if ( 'write' === $intent || 'design-first' === $intent || 'start-writing' === $intent ) {
+		return true;
+	}
+
+	$goals = get_option( 'goals', array() );
+	if ( is_array( $goals ) && in_array( 'write', $goals, true ) ) {
+		return true;
+	}
+
+	return false;
+}
+
+/**
+ * Determines if the user has started setting up an ecommerce site, or selected "sell" as the site intent.
+ *
+ * @return bool
+ */
+function _jetpack_is_potential_ecommerce_site() {
+	$goals       = get_option( 'goals', array() );
+	$site_intent = get_option( 'site_intent', '' );
+
+	if ( 'sell' === $site_intent || ( is_array( $goals ) && in_array( 'sell', $goals, true ) ) ) {
+		return true;
+	}
+
+	// if user has started setting up woocommerce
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/_inc/blogging-prompts.php
```diff
@@ -258,11 +258,104 @@ function jetpack_is_new_post_screen() {
...
+	return false;
+}
+
 /**
  * Determines if the site might have a blog.
  *
  * @return bool
  */
 function jetpack_is_potential_blogging_site() {
-	return jetpack_has_write_intent() || jetpack_has_posts_page() || jetpack_has_or_will_publish_posts();
```
eoigal, line 267: I feel like we should still use `jetpack_has_or_will_publish_posts` as well as the check on`_jetpack_has_more_posts_than_pages` - we should at least check if the there are any scheduled posts instead of just published posts.

projects/plugins/jetpack/_inc/blogging-prompts.php
```diff
@@ -258,11 +258,104 @@ function jetpack_is_new_post_screen() {
...
+ *
+ * @return bool
+ */
+function _jetpack_has_selected_blogging_goal_or_intent() {
+	$intent = get_option( 'site_intent', '' );
+	if ( 'write' === $intent || 'design-first' === $intent || 'start-writing' === $intent ) {
+		return true;
+	}
+
+	$goals = get_option( 'goals', array() );
```
jeherve, line 272: Is the `goals` option new? Is it still possible to update that option name to give it a `jetpack_` prefix? I worry that with such a general name, it could easily conflict with other plugins installed on the site and using a similar option name.

projects/plugins/jetpack/_inc/blogging-prompts.php
```diff
@@ -258,11 +258,104 @@ function jetpack_is_new_post_screen() {
...
 }
 
+/**
+ * Determines if the user selected a "blogging" intent or goal when creating the site.
+ *
+ * @return bool
+ */
+function _jetpack_has_selected_blogging_goal_or_intent() {
+	$intent = get_option( 'site_intent', '' );
+	if ( 'write' === $intent || 'design-first' === $intent || 'start-writing' === $intent ) {
```
jeherve, line 268: Nitpick: what would you say for using `in_array` here as well, to be consistent with what's below, and maybe to increase readability a bit?

```suggestion
	if ( in_array( $intent, array( 'write', 'design-first', 'start-writing' ), true ) ) {
```

jeherve (commented):
Left a couple of small questions / nitpicks!

PR #34690, opened 2023-12-18
CRM: Ensure segments can be deleted

### LINKED ISSUE #34687: Fatal Error: Jetpack CRM plugin
### Impacted plugin

CRM

### Quick summary

User is unable to delete segments created with Jetpack CRM Extension: Advanced Segments plugin. Running php-errors in CLI provided Fatal error for Jetpack CRM. 

ZD-7465293

### Steps to reproduce

1. Have Jetpack CRM & Jetpack CRM Extension: Advanced Segments installed (with segments created)
2. Go to: Contacts > Segments 
3. Attempt to delete segment

Result: Pop up: "Could not delete!" 

With both the Jetpack CRM & Jetpack Advanced Segments plugin you should be able to properly delete segments. 

### What actually happened

When you attempt to delete segments you get a pop up that says,  "Could not delete!" 

### Impact

One

### Available workarounds?

Atomic

### Logs or notes

Fatal error related to Jetpack CRM:

```
[17-Dec-2023 18:12:01 UTC] PHP Fatal error:  Uncaught Error: Call to undefined function jpcrm_api_forbidden_request() in /srv/htdocs/wp-content/plugins/zero-bs-crm/includes/ZeroBSCRM.AJAX.php:3744
Stack trace:
#0 /wordpress/core/6.4.2/wp-includes/class-wp-hook.php(324): zeroBSCRM_AJAX_enactListViewBulkAction('')
#1 /wordpress/core/6.4.2/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters('', Array)
#2 /wordpress/core/6.4.2/wp-includes/plugin.php(517): WP_Hook->do_action(Array)
#3 /wordpress/core/6.4.2/wp-admin/admin-ajax.php(188): do_action('wp_ajax_enactLi...')
#4 {main}
  thrown in /srv/htdocs/wp-content/plugins/zero-bs-crm/includes/ZeroBSCRM.AJAX.php on line 3744
[17-Dec-2023 18:13:21 UTC] PHP Notice:  Function WP_Scripts::localize was called incorrectly. The $l10n parameter must be an array. To pass arbitrary data to scripts, use the wp_add_inline_script() function instead. Please see <a>Debugging in WordPress</a> for more information. (This message was added in version 5.7.0.) in /wordpress/core/6.4.2/wp-includes/functions.php on line 6031
[17-Dec-2023 18:13:26 UTC] PHP Notice:  Function WP_Scripts::localize was called incorrectly. The $l10n parameter must be an array. To pass arbitrary data to scripts, use the wp_add_inline_script() function instead. Please see <a>Debugging in WordPress</a> for more information. (This message was added in version 5.7.0.) in /wordpress/core/6.4.2/wp-includes/functions.php on line 6031
```

coder-karen:
Fixes https://github.com/Automattic/jetpack/issues/34687

## Proposed changes:

* This PR adds a case and new function to check that a user has permissions to modify segments, which is used to determine whether the bulk actions can be applied on the segments page.
* This fixes an issue with the bulk actions delete option on that page: `/wp-admin/admin.php?page=manage-segments`.
* This is an issue both with and without Advanced Segments installed and active, and also affects segments created both with that extension and without it.
* Note this issue doesn't just affect WoA sites, but also any site with the current version of Jetpack CRM installed an active.

### Other information:

## Jetpack product discussion

https://github.com/Automattic/jetpack/issues/34687

No.

## Testing instructions:

* To replicate the issue, test in trunk or with a Jurassic Ninja site running the most recent release.
* With the Advanced Segments extension installed an active, add several new segments (and add segments without it active): `/wp-admin/admin.php?page=manage-segments`
* With the Advanced Segments extension active or deactivated, try to delete any segment (created with or without the Advanced Segments extension). You will see an 'Invalid permissions' notice.
* To test the fix, apply this PR locally or using the Beta Tester plugin on a test site. Follow the same steps, and you should be able to delete created segments

### DIFF
--- a/projects/plugins/crm/includes/ZeroBSCRM.Permissions.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.Permissions.php
@@ -718,19 +718,6 @@ function zeroBSCRM_permsCustomers() {
 	    return false;
 	}
 
-/**
- * Determine if the current user is allowed to manage segments.
- *
- * @return bool
- */
-function jpcrm_perms_segments() {
-	$current_user = wp_get_current_user();
-	if ( $current_user->has_cap( 'admin_zerobs_customers' ) ) {
-		return true;
-	}
-	return false;
-}
-
 	// LOGS
 
 
--- a/projects/plugins/crm/includes/ZeroBSCRM.Permissions.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.Permissions.php
@@ -577,6 +577,9 @@ function zeroBSCRM_permsObjType( $obj_type_id = -1 ) { // phpcs:ignore WordPress
 
 		case ZBS_TYPE_TASK:
 			return zeroBSCRM_perms_tasks();
+
+		case ZBS_TYPE_SEGMENT:
+			return jpcrm_perms_segments();
 	}
 
 	return false;
@@ -716,6 +719,18 @@ function zeroBSCRM_permsCustomers() {
 	    return false;
 	}
 
+/**
+ * Determine if the current user is allowed to manage segments.
+ *
+ * @return bool
+ */
+function jpcrm_perms_segments() {
+	$current_user = wp_get_current_user();
+	if ( $current_user->has_cap( 'admin_zerobs_customers' ) ) {
+		return true;
+	}
+	return false;
+}
 
 	// LOGS
 
--- a/projects/plugins/crm/includes/ZeroBSCRM.Permissions.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.Permissions.php
@@ -560,6 +560,7 @@ function zeroBSCRM_permsObjType( $obj_type_id = -1 ) { // phpcs:ignore WordPress
 	switch ( $obj_type_id ) {
 		case ZBS_TYPE_CONTACT:
 		case ZBS_TYPE_COMPANY:
+		case ZBS_TYPE_SEGMENT:
 			return zeroBSCRM_permsCustomers();
 
 		case ZBS_TYPE_QUOTE:
@@ -578,8 +579,6 @@ function zeroBSCRM_permsObjType( $obj_type_id = -1 ) { // phpcs:ignore WordPress
 		case ZBS_TYPE_TASK:
 			return zeroBSCRM_perms_tasks();
 
-		case ZBS_TYPE_SEGMENT:
-			return jpcrm_perms_segments();
 	}
 
 	return false;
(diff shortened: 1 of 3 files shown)


projects/plugins/crm/includes/ZeroBSCRM.Permissions.php
```diff
@@ -577,6 +577,9 @@ function zeroBSCRM_permsObjType( $obj_type_id = -1 ) { // phpcs:ignore WordPress
 
 		case ZBS_TYPE_TASK:
 			return zeroBSCRM_perms_tasks();
+
+		case ZBS_TYPE_SEGMENT:
+			return jpcrm_perms_segments();
```
tbradsha: This works. Since we're calling this the same as the contacts permission, another option would be to collapse this into the `ZBS_TYPE_CONTACT` case block (removing the need for a new function).

coder-karen: It seems the function used in the ZBS_TYPE_CONTACTS case is deprecated - `zeroBSCRM_permsCustomers`, but replacing it with `jpcrm_can_user_manage_contacts` is a little messier as we need the contact id, contact etc. So it'd either be changing the segment to use the deprecated function or using a new function. It looks like there a fair few usages of the deprecated function, possibly for the same reason (needing more data to pass in to the function).
I moved it. Commit: dace982a8c724ea2209ac7309620c76b72bb5f0c

projects/plugins/crm/includes/ZeroBSCRM.Permissions.php
```diff
@@ -716,6 +718,18 @@ function zeroBSCRM_permsExport(){
...
+ *
+ * @return bool
+ */
+function jpcrm_perms_segments() {
+	$current_user = wp_get_current_user();
+	if ( $current_user->has_cap( 'admin_zerobs_customers' ) ) {
+		return true;
+	}
+	return false;
+}
```
tbradsha: 🪓

tbradsha (dismissed):
Looks good! Thanks for the quick turnaround!

projects/plugins/crm/includes/ZeroBSCRM.Permissions.php
coder-karen: Thank you - removed: f1d39a6d7b20cf8bbdff92a6c760ab3fafe579c9

tbradsha (approved):

projects/plugins/crm/includes/ZeroBSCRM.Permissions.php
tbradsha: Yeah, the deprecation was a bit controversial; we can ignore it in this case.

PR #34698, opened 2023-12-18
Protect: Remove use of "rulesVersion" in the Firewall settings screen

nateweller:
## Proposed changes:

* Removes `rulesVersion` in favor of just displaying the "last updated" date. The rules version string is not maintained.

### Other information:

## Jetpack product discussion

N/A

No

## Testing instructions:

* Visit the Firewall tab in Jetpack Protect.
* Upgrade to access automatic rules.
* Validate rules are not labelled as "v1.0.0" once installed.

## Screenshots

Before:

Screenshot 2023-12-03 at 3 09 12 PM

After:

Screenshot 2023-12-18 at 11 48 25 AM

### DIFF
--- a/projects/packages/waf/changelog/deprecate-waf-rules-version
+++ b/projects/packages/waf/changelog/deprecate-waf-rules-version
@@ -1,0 +1,5 @@
+Significance: patch
+Type: deprecated
+Comment: Deprecated a function in the WAF package.
+
+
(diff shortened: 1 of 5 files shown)


dkmyta (dismissed):
Performs as described!

Question - do we intend on using this `Waf_Stats` class method at all moving forward? It appears with the removal of this code it's no longer used anywhere. If not, it might just be worth removing to reduce any possible confusion and just creating a new method when we do have use for such a thing.

nateweller:
> Question - do we intend on using this Waf_Stats class method at all moving forward? It appears with the removal of this code it's no longer used anywhere. If not, it might just be worth removing to reduce any possible confusion and just creating a new method when we do have use for such a thing.

Good call - I've marked the method as deprecated, and we can remove it if/when we do a breaking version bump.

ArSn (approved):
Code deprecation LGTM.

PR #34703, opened 2023-12-18
Fix sharing buttons window open

### LINKED ISSUE #34701: Sharing Block: Fix window opening on Simple/Atomic
On Simple/Atomic view.js sites are merged into one that loads early, which make it running prior to page fully load. 
Context: p1702929981921579-slack-C01U2KGS2PQ

andrii-lysenko:
Fixes https://github.com/Automattic/jetpack/issues/34701

## Proposed changes:

In order to fix issue on Simple/Atomic we need run the script only when we certain page content is loaded.

### Other information:

## Testing instructions:

* Pull the branch locally and build `jetpack build plugins/jetpack`
* Use `jetpack rsync` to upload changes to your sandbox
* Create a new post and add Sharing Buttons to it. 
* Open the view with parameter `?concat_js=yes`
* Try share the post. 600*400 window should be opened instead of opening in new tab.

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-sharing-buttons-window-open
+++ b/projects/plugins/jetpack/changelog/fix-sharing-buttons-window-open
@@ -1,0 +1,5 @@
+Significance: patch
+Type: other
+Comment: This feature is not visible for users, so won't affect them. We simply adjusting when does function run.
+
+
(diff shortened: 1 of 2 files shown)


darssen:
Code looks good and tests good as well. :ship:

darssen (approved):

PR #34707, opened 2023-12-19
Fix My plan dashboard js errors due to nested anchor tags 

### LINKED ISSUE #33814: Jetpack > Dashboard > My Plan: JS errors due to nested a tags
### Steps to reproduce

1. Open your browser's JavaScript console.
2. Go to Jetpack > Dashboard > My Plan
3. Notice the errors:

`Warning: validateDOMNesting(...): <a> cannot appear as a descendant of <a>.`

This is because in `projects/plugins/jetpack/_inc/client/my-plan/my-plan-header/index.js`, we use the `ExternalLink` component inside the `Button` component.

This should be fixed:

- Either by updating the `Button` component itself to support defining an `external` prop that would display an External icon and set the link to open in a new tab. **I think that should be the preferred option since it would then be usable in multiple other areas of the dashboard**.
- Either by moving away from `ExternalLink` in that file, and instead follow a pattern similar to this one, with a manually added `Gridicon`:
https://github.com/Automattic/jetpack/blob/9108e0f9632ab11c7aa03f79f27062ceb9bf4b34/projects/plugins/jetpack/_inc/client/at-a-glance/security-bundle/index.jsx#L72-L80

darssen:
Fixes #33814

## Proposed changes:

As explained in the ticket, when navigating to My Plan Dashboard we were getting 
```
Warning: validateDOMNesting(...): <a> cannot appear as a descendant of <a>.
```
This is because we use the `ExternalLink` component inside the `Button` component **passing the `href` to the `Button`**. In this case `Button` creates another `<a>` tag that wraps the one from `ExternalLink`.

**The changes in this PR** move the href from the `Button` to the `ExternalLink` props so that only one `<a>` tag will exist. 
This has been done in all places inside My Plan where such situation existed.

### Other information:

## Testing instructions:

In your testing site, check that before activating this branch the error is still reproduced.

* Open your browser's JavaScript console.
* Notice the errors:

Once the branch is active no errors should appear and the links should appear and work in the same way as before.

### DIFF
--- a/projects/plugins/jetpack/_inc/client/my-plan/my-plan-body.jsx
+++ b/projects/plugins/jetpack/_inc/client/my-plan/my-plan-body.jsx
@@ -106,13 +106,12 @@ class MyPlanBody extends React.Component {
 					<div className="jp-landing__plan-features-text">
 						<h3 className="jp-landing__plan-features-title">{ title }</h3>
 						<p>{ description }</p>
-						<Button
-							onClick={ this.handleButtonClickForTracking( 'view_backup_dash' ) }
-							href={ getRedirectUrl( 'calypso-activity-log', { site: this.props.siteRawUrl } ) }
-							compact
-							rna
-						>
-							<ExternalLink>{ __( 'View your backups', 'jetpack' ) }</ExternalLink>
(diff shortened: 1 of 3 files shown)


jeherve (approved):
This tests well for me. 🚢

PR #34714, opened 2023-12-19
Boost: Update Settings Page to react

dilirity:
## Proposed changes:

* Update Settings Page to react.

### Other information:

## Jetpack product discussion

pc9hqz-2jy-p2

no

## Testing instructions:

* Make sure that everything on the main Settings page works as expected.


projects/plugins/boost/app/assets/src/js/layout/settings-page/settings-page.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+	useEffect( () => {
+		const unsubscribe = isGenerating.subscribe( value => {
+			setIsGeneratingValue( value );
+		} );
+
+		return () => {
+			unsubscribe();
+		};
+	}, [] );
```
haqadn, line 30: This is so cool! If I kinda want a custom hook now:

```javascript
const useSvelteStore = (svelteStore) => {
	const [ value, setValue ] = useState( svelteStore.get() );
	
	useEffect( () => {
		const unsubscribe = svelteStore.subscribe( value => {
			setValue( value );
		} );

		return () => {
			unsubscribe();
		};
	}, [] );

	return value;
}
```

thingalon: I mean if we want to rush to the end of the React migration, there is absolutely nothing magical about Svelte stores. I could write a non-svelte replacement for `writable` in a few lines, we could wrap them into React like so ^ and then we just have a React codebase to tidy up. Just sayin'.

haqadn (changes requested):
The critical CSS states aren't updating as it generates. The UI get's stuck with the mounted state. I suspect it has something to do with embedding React in Svelte.

It might be worth it to ignore this issue and merge it while fixing it in my PR since it will be all React in there and we won't have to deal with the added complexity. Let me know if that's what you too think we should do.

haqadn (approved):
As discussed in DM, giving it a ✅  for merging as is even though the critical CSS UI is currently broken. It will be fixed in #34731

PR #34715, opened 2023-12-19
Admin page: remove alt attribute from decorative icon

jeherve:
## Proposed changes:

The Alt value isn't necessary here, since the icon doesn't bring any additional info on the page for folks using screen readers.

This has the added benefit of getting rid of a React warning:

```
Warning: Failed prop type: Invalid prop `alt` of type `object` supplied to `PlanIcon`, expected `string`.
PlanIcon@wp-content/plugins/jetpack/_inc/build/admin.js?ver=15a830895a6f6d1e3c39:44686:1
MyPlanCard@wp-content/plugins/jetpack/_inc/build/admin.js?ver=15a830895a6f6d1e3c39:51709:20
```

### Other information:

## Jetpack product discussion

* N/A

* No

## Testing instructions:

* Via Store admin, add a legacy Jetpack Backup Daily plan to your site.
* Open your JS console.
    * You should not see the React warning from above.
    * The `alt` attribute from the plan icon should have an empty value (it previously had an invalid object).

image


fgiannar (approved):
LGMT 👍

PR #34716, opened 2023-12-19
Subscriptions: update "verify your email" wall copy

simison:
Updates copy to be more about "confirming their subscription" rather than "confirming email" per-se, and aligns copy with email they received.

See https://github.com/Automattic/wp-calypso/issues/84773#issuecomment-1845026885

Doesn't update the actual modal where we also nudge to check email.

## Proposed changes:

* Update copy
* Add email to the copy

Before
Screenshot 2023-12-19 at 14 47 38

After
Screenshot 2023-12-19 at 14 46 10

### Other information:

## Testing instructions:

* Subscribe to post requiring subscription, but don't confirm your email. 
* Email does not get cached for others to see because the whole page isn't cached (https://github.com/Automattic/jetpack/pull/34242#issuecomment-1838479034)

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
@@ -1114,9 +1114,18 @@ function is_user_auth() {
  * @return string
  */
 function get_paywall_blocks_subscribe_pending() {
-	$access_heading = esc_html__( 'Verify your email to continue reading', 'jetpack' );
+	$subscribe_email = Jetpack_Memberships::get_current_user_email();
 
-	$subscribe_text = esc_html__( 'Please check your inbox to confirm your subscription.', 'jetpack' );
+	/** This filter is documented in modules/contact-form/grunion-contact-form.php */
+	if ( is_wpcom() || false !== apply_filters( 'jetpack_auto_fill_logged_in_user', false ) ) {
+		$current_user    = wp_get_current_user();
+		$subscribe_email = ! empty( $current_user->user_email ) ? $current_user->user_email : '';
+	}
+
+	$access_heading = esc_html__( 'Confirm your subscription to continue reading', 'jetpack' );
+
+	/* translators: %s: email address */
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -1114,9 +1114,18 @@ function is_user_auth() {
...
+	if ( is_wpcom() || false !== apply_filters( 'jetpack_auto_fill_logged_in_user', false ) ) {
+		$current_user    = wp_get_current_user();
+		$subscribe_email = ! empty( $current_user->user_email ) ? $current_user->user_email : '';
+	}
+
+	$access_heading = esc_html__( 'Confirm your subscription to continue reading', 'jetpack' );
 
-	$subscribe_text = esc_html__( 'Please check your inbox to confirm your subscription.', 'jetpack' );
+	/* translators: %s: email address */
+	$subscribe_text = wp_kses( sprintf( __( 'Head to your inbox and confirm your email address %s.', 'jetpack' ), $subscribe_email ), array() );
```
jeherve: Do we need `wp_kses` here? Could we use `esc_html` here instead? 

```suggestion
	$subscribe_text = sprintf( esc_html__( 'Head to your inbox and confirm your email address %s.', 'jetpack' ), $subscribe_email );
```

projects/plugins/jetpack/extensions/blocks/subscriptions/subscriptions.php
```diff
@@ -1114,9 +1114,18 @@ function is_user_auth() {
  * @return string
  */
 function get_paywall_blocks_subscribe_pending() {
-	$access_heading = esc_html__( 'Verify your email to continue reading', 'jetpack' );
+	$subscribe_email = Jetpack_Memberships::get_current_user_email();
+
+	/** This filter is documented in modules/contact-form/grunion-contact-form.php */
+	if ( is_wpcom() || false !== apply_filters( 'jetpack_auto_fill_logged_in_user', false ) ) {
```
lezama, line 1120: I think we should use the `Host` class for these checks, but we can do it in a separate PR

lezama (approved):

PR #34717, opened 2023-12-19
Top Posts & Pages Block: Allow Transform from Legacy Widget

Aurorum:
Part of #32937

## Proposed changes:

* Allows users to easily transform the Top Posts & Pages Legacy widget into blocks. 
* Based on the Contact Info widget, which allows folks to transform the widget into the block, while also creating a Heading block for the title where necessary 

ezgif com-video-to-gif-converted

## Jetpack product discussion

N/A.

No.

## Testing instructions:

* You'll need to remove these lines first so that you can insert the legacy widget

https://github.com/Automattic/jetpack/blob/ec598809607b6bffd312aa6def990ee05952ec61/projects/plugins/jetpack/modules/widgets/top-posts.php#L86-L98

* Then enable Beta blocks
* Go to the Widgets section and insert a legacy widget
* Confirm that you can transform the block (press the calendar icon in the top left)

It should sync with the settings from the Legacy Widget, so that the display settings and post types are aligned. I've also set the default period as 48 hours, since that's what it currently is for the widget.


kangzj:
Thanks for working on this! Sorry for the delay - I'll give it a spin early next week.

kangzj (approved):
Works well and looks good 👍

Aurorum:
Thanks for revisiting these! I've updated the branch for this one.

kangzj (approved):

PR #34732, opened 2023-12-20
Like Block: Add support for Block Hooks API (Draft)

TimBroddin:
## Proposed changes:
- Add the hooked_block_types filter
- When `should_hook_block()` returns true it inserts the Like block inside a `wp:group` block, rendering correctly

CleanShot 2023-12-20 at 10 37 31@2x

### Other information:

## Testing instructions:
1. Apply this PR & make sure `jetpack watch` is running
2. Make `should_hook_block()` return true. (inside `extensions/blocks/like/like.php`)
3. A block should be auto-inserted in the frontend & the editor & look right

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/like/like.php
+++ b/projects/plugins/jetpack/extensions/blocks/like/like.php
@@ -116,11 +116,57 @@ function render_block( $attr, $content, $block ) {
 		. "<div class='likes-widget-placeholder post-likes-widget-placeholder' style='height: 55px;'><span class='button'><span>" . esc_html__( 'Like', 'jetpack' ) . "</span></span> <span class='loading'>" . esc_html__( 'Loading...', 'jetpack' ) . '</span></div>'
 		. "<span class='sd-text-color'></span><a class='sd-link-color'></a>"
 		. '</div>';
-	return sprintf(
+
+	$group_block_start = '<!-- wp:group {"layout":{"type":"constrained"}} --><div class="wp-block-group">';
+	$output            = sprintf(
 		'<div class="%1$s">%2$s</div>',
 		esc_attr( Blocks::classes( Blocks::get_block_feature( __DIR__ ), $attr ) ),
 		$html
 	);
+	$group_block_end   = '</div><!-- /wp:group -->';
+
+	return do_blocks( $group_block_start . $output . $group_block_end );
+}
+
+/**
+ * Filters the list of blocks to be automatically inserted into the post content.
+ *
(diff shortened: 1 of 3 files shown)


projects/plugins/jetpack/extensions/blocks/like/like.php
```diff
@@ -116,11 +116,57 @@ function render_block( $attr, $content, $block ) {
...
+ * to add a new block type (e.g., 'jetpack/like') to the list of hooked blocks.
+ *
+ * @param array                     $hooked_blocks       The current array of blocks to be hooked.
+ * @param string                    $relative_position   The position relative to the anchor block ('before' or 'after').
+ * @param string                    $anchor_block        The anchor block type (e.g., 'core/post-content').
+ * @param WP_Block_Template | array $context             The block template, template part, or pattern that the anchor block belongs to.
+ *
+ * @return array The modified array of blocks to be hooked.
+ */
+function hooked_block_types( $hooked_blocks, $relative_position, $anchor_block, $context ) {
```
jeherve, line 145: Do you think we should make this available for both the sharing block and the like block, maybe via a new method in `Jetpack_Gutenberg`, so we can use the same strategy for both?

TimBroddin: That's certainly a possibility. However, at the moment, I'm still thinking about the logic that should go into `should_hook_block().` 

Do you mean something along these lines? 

```
function init_function() {
   if( should_hook_block() ) {
      Jetpack_Gutenberg::block_hook_insert_after_post_content( 'jetpack/like' ); 
   }
}


add_action( 'init', 'init_function');
```

JuanLucha (approved):
Nice work!
Screenshot 2023-12-20 at 11 53 19

**Note:** I learned a lot about Block Hooks API reviewing this PR 🚀

projects/plugins/jetpack/extensions/blocks/like/like.php
jeherve: I was thinking about moving the whole `hooked_block_types` function to `Jetpack_Gutenberg`, and in there add both the like button and the sharing buttons blocks to the `$hooked_blocks` array, if the blocks are available. This way we hook into the `hooked_block_types` filter from one place only, with the same conditions and the same priority, and we can expand that list of supported Jetpack blocks later on.

TimBroddin:
I'm putting this PR in draft for a bit. I'm going to focus on the other areas first, and will then revisit.

PR #34733, opened 2023-12-20
Social | Hide conversion notice for Simple sites

gmjuhasz:
We plan to turn on the auto-conversion for simple sites, where currently, they cannot turn the feature off. For now we hide the conversion notice for those sites.

### Other information:

## Jetpack product discussion

https://github.com/orgs/Automattic/projects/733/views/2?pane=issue&itemId=48088220

No.
## Testing instructions:

On a Jetpack site
* Auto conversion should work as before, and you should see the conversion notice when the feature is on, and you select an invalid but convertible media.

On a simple site
* You can sandbox this change with the jetpack downloader below, and on Simple sites, the conversion notice should not appear. To have the feature as available you can download this change onto D131678-code

On Atomic sites you can use the Jetpack Beta plugin to test this

### DIFF
--- a/projects/js-packages/publicize-components/changelog/update-social-hide-conversion-notice-wpcom
+++ b/projects/js-packages/publicize-components/changelog/update-social-hide-conversion-notice-wpcom
@@ -1,0 +1,4 @@
+Significance: minor
+Type: changed
+
+Hide conversion notice on Simple sites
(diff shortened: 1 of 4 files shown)


spsiddarthan (approved):
Tests well!

PR #34739, opened 2023-12-20
Backup: Add namespace versioning to Helper_Script_Manager and other classes

pypt:
Add class versioning using PHP namespaces, to make sure that the plugin always loads the expected version of a class.

## Proposed changes:

When two or more plugins are installed that contain a class with the same name (e.g. the main Jetpack plugin and the Jetpack VaultPress Backup plugin both ship `Automattic\Jetpack\Backup\Helper_Script_Manager` class), the Jetpack autoloader https://github.com/Automattic/jetpack/pull/34297#discussion_r1424048710, and under certain circumstances it might even decide to load an older version of the class.

Specifically, we had issues with the Jetpack VaultPress Backup plugin loading an older version of the `Helper_Script_Manager` class using the newer code, in which case an `abspath` property in `Helper_Script_Manager::install_helper_script()` (https://github.com/Automattic/jetpack/pull/34297) wasn't always returned.

* p1703028095475669-slack-CS8UYNPEE

To avoid that behavior, and always load the *exact same* version of the class that has been shipped with the plugin, we're adding simple namespace-based versioning to `Helper_Script_Manager`, backup plugin, and other related code.

All of the classes that make up the package/plugin got moved to either the `Automattic\Jetpack\Backup\V0001` or the `Automattic\Jetpack\Transport_Helper\V0001` namespace. Code that uses said classes refer to them using the versioned namespace (e.g. `Automattic\Jetpack\Backup\V0001\Helper_Script_Manager::install_helper_script()`, so that once the plugin gets installed, it always loads and uses a specific version of its dependencies (and not an older/newer one, which it might be incompatible with).

Older versions of the Jetpack plugins don't use these namespaces (e.g. they refer to the `Helper_Script_Manager` class as `Automattic\Jetpack\Backup\Helper_Script_Manager`), so if the plugin with versioned namespaces gets installed together with an older plugin without those, both should work fine.

If we ever need to change the plugin, the process is somewhat cumbersome yet straightforward:

1. Do the changes that you need;
2. Search for `V0001`, replace it with `V0002` everywhere.

to make sure that older versions of the plugin load and use the `V0001` version of the relevant code, and the newer versions use `V0002`.

Relevant discussions:

* peaFOp-2ar-p2
* https://github.com/Automattic/jetpack/pull/34297#discussion_r1424048710

### Other information:

## Jetpack product discussion

peaFOp-2ar-p2

No.

## Testing instructions:

Just like with https://github.com/Automattic/jetpack/pull/34297, run a Jetpack VaultPress Backup backup without credentials set up, so that the tool is then forced to upload the helper script via Jetpack API and thus trigger this code.

Make sure that the `abspath` property gets always returned in the `install_helper_script()` results.

Also, test with multiple versions/forms of various Jetpack plugins installed, e.g. the newer Jetpack VaultPress Backup plugin (with this patch), and the older main Jetpack plugin (without this patch); make sure that you still always get the `abspath` back.


projects/packages/backup-helper-script-manager/src/class-helper-script-manager-impl.php
```diff
@@ -5,7 +5,11 @@
  * @package automattic/jetpack-backup
  */
 
-namespace Automattic\Jetpack\Backup;
+// After changing this file, consider increasing the version number ("VXXX") in the namespace, in order to ensure that
```
anomiex: Just this file, or all the files using the "Automattic\Jetpack\Backup\V0001" namespace?

projects/packages/backup-helper-script-manager/changelog/backup-helper-script-manager-class-versioning
```diff
@@ -0,0 +1,4 @@
+Significance: patch
```
anomiex: This changes the fully-namespaced names of the classes, which is a breaking change.
```suggestion
Significance: major
```

projects/packages/backup/changelog/backup-helper-script-manager-class-versioning
```diff
@@ -0,0 +1,4 @@
+Significance: patch
```
anomiex: This changes the fully-namespaced names of the classes, which is a breaking change.
```suggestion
Significance: major
```

projects/packages/transport-helper/changelog/backup-helper-script-manager-class-versioning
```diff
@@ -0,0 +1,4 @@
+Significance: patch
```
anomiex: This changes the fully-namespaced names of the classes, which is a breaking change.
```suggestion
Significance: major
```

anomiex (commented):
A quick opengrok search didn't turn up anything outside of the monorepo referring to the namespaces being changed, so :crossed_fingers: hopefully it won't break Simple.

I didn't review much beyond that and the few notes inline.

projects/packages/backup-helper-script-manager/src/class-helper-script-manager-impl.php
pypt: Thanks, updated in 712c9e9.

projects/packages/backup-helper-script-manager/changelog/backup-helper-script-manager-class-versioning
pypt: Here and elsewhere: cfa3b52

jwebbdev:
I tried loading this up in a Jurassic Ninja site along-side 12.9.1 of Jetpack and it caused connection errors to appear and the helper script hit 500s when trying to install remotely.

I'll give it another shot once the latest CI has finished running in case Jetpack Beta isn't getting the most recent versions of things.

jwebbdev:
Seems to be some issues.  Generating a JN site with this version of the plugin fails to be able to form a Jetpack connection.  With or without the Jetpack plugin.

Screenshot 2023-12-20 at 3 47 02 PM

https://jurassic.ninja/create?jetpack-beta&branches.jetpack-backup=modify/backup-helper-script-manager-class-versioning&wp-debug-log&nojetpack

pypt:
Thanks @jwebbdev, good catch.

The plugin was failing because it was trying to load `Automattic\Jetpack\Backup\V0001\WP_REST_Server` in the newly-namespaced `Jetpack_Backup` class.

In ac5daaa and 535e7a0, I've fixed those (not sure if I needed all the `use function` there, but they don't hurt).

In 587c7d3, I've also fixed a namespace in the REST controller setup.

I've tried it with JN:

https://jurassic.ninja/create?jetpack-beta&branches.jetpack-backup=modify/backup-helper-script-manager-class-versioning&wp-debug-log&nojetpack

and it seems to work now, i.e. I was able to connect and create a backup (all without credentials).

Could someone from @Automattic/yamato-backup have another look?

"WordPress.com Tests" are "Failing after 1s — Unexpected termination", but I'm not sure if it's me; asked here:

p1703173071584489-slack-C05Q5HSS013

Initsogar:
@pypt I will start running some tests. Will keep you posted.

projects/packages/backup/src/class-jetpack-backup.php
```diff
@@ -5,14 +5,20 @@
...
+namespace Automattic\Jetpack\Backup\V0001;
+
 if ( ! defined( 'ABSPATH' ) ) {
 	exit;
 }
 
 use Automattic\Jetpack\Admin_UI\Admin_Menu;
 use Automattic\Jetpack\Assets;
-use Automattic\Jetpack\Backup\Initial_State as Backup_Initial_State;
-use Automattic\Jetpack\Backup\Jetpack_Backup_Upgrades;
```
Initsogar, line 15: It seems we are still using `Jetpack_Backup_Upgrades` in this file, should we add the `use` statement?
https://github.com/Automattic/jetpack/blob/cb410855f3e5daf7efb3fd135b4a6ed0bfeb4d40/projects/packages/backup/src/class-jetpack-backup.php#L180

projects/plugins/migration/src/class-wpcom-migration.php
```diff
@@ -61,7 +62,7 @@ function () {
 			},
 			1
 		);
-		\Jetpack_Backup::initialize();
+		My_Jetpack_Backup::initialize();
```
Initsogar: I think it would be okay to keep it as `Jetpack_Backup` to avoid any confusions with `My_Jetpack`, which is another feature in our plugins.

projects/packages/backup/src/class-jetpack-backup.php
pypt: They used to be in two different namespaces:

* `Jetpack_Backup` was in a global namespace, i.e. `\Jetpack_Backup`
* `Jetpack_Backup_Upgrades` was in a `Automattic\Jetpack\Backup` namespace, i.e. `\Automattic\Jetpack\Backup\Jetpack_Backup_Upgrades`

thus the import was needed.

Now that they're both in the same namespace (`Automattic\Jetpack\Backup\V0001`), no import is needed.

Initsogar:
> @pypt I will start running some tests. Will keep you posted.

Tested backing up my site with different scenarios and all went well. Just left a few comments that would be great to double-check.

**Scenarios tested:**
Only backup plugin with this feature branch ✅
Only Jetpack plugin with this feature branch ✅
Backup plugin with this feature branch + Jetpack plugin 12.9.1 ✅
Backup plugin with this feature branch + Jetpack plugin bleeding edge ✅
Backup plugin and Jetpack plugin both with this feature branch ✅

projects/plugins/migration/src/class-wpcom-migration.php
pypt: Good point, renamed back in 3e8934a.

pypt:
Thank you Rafa!

projects/packages/backup/src/class-jetpack-backup.php
Initsogar: Ahhh, gotcha! Thanks for clarification :)!

pypt:
@Automattic/yamato-backup, could someone please review/approve this one?

The ["E2E Tests / Jetpack post-connection e2e tests (pull_request)"](https://github.com/Automattic/jetpack/actions/runs/7348501314/job/20006837157?pr=34739) has been seemingly failing since forever (whenever the test runner decides to run that test):

https://github.com/Automattic/jetpack/actions/workflows/e2e-tests.yml

pypt:
Guess-fix for the failing unrelated tests: https://github.com/Automattic/jetpack/pull/34796

pypt:
Unrelated test failures have been fixed in https://github.com/Automattic/jetpack/pull/34796, so @Automattic/yamato-backup could someone approve this one?

@Initsogar tested the changes in https://github.com/Automattic/jetpack/pull/34739#issuecomment-1866600046 and didn't find any problems, and I haven't changed this PR in any significant way since.

Initsogar (approved):
Looks good to me 👍🏽

pypt:
Some upgrade weirdness:

p1705089030985119-slack-CBG1CP4EN

PR #34744, opened 2023-12-20
Promote Subscription: Replace use of subscription block

lupus2k:
Related to https://github.com/Automattic/jetpack/pull/34659

## Proposed changes:
Use custom input + button for subscription modal instead of using the Subscriptions block.
The input is disabled for logged-in users as the subscription endpoint won't let you use another email when you are logged-in.

## Testing instructions:
* Follow instructions from here: https://github.com/Automattic/jetpack/pull/34659#issuecomment-1857257562
* Go to a post with a non-site user (admin or editor)
* Comment on the post
* You should see the subscription modal


paulopmt1 (approved):
LTGM, nice work @lupus2k!

PR #34746, opened 2023-12-21
[Boost] Fix importmap deferral

thingalon:
Fixes compatibility with Gutenberg plugin / upcoming Gutenberg code.

`<script type="importmap">` blocks generally need to be at the top of the page, before anything that depends on them. They don't run complex code and moving them down the page can break it.

This PR addresses the problem by adding `importmap` to the set of script types excluded when deferring JS.

## Proposed changes:

* Add `type="importmap"` to the set of exclusions from deferred JS.

### Other information:

no

## Testing instructions:

** To replicate **
* Install the Gutenberg plugin
* Turn on Deferred JS
* Use an interactive block, like the Jetpack podcast player
* Spot an error in the console output -- `An import map is added after module script load was triggered.`

Then install this patch and verify the issue is resolved.

### DIFF
--- a/projects/packages/forms/changelog/boost-fix-importmap-deferral
+++ b/projects/packages/forms/changelog/boost-fix-importmap-deferral
@@ -1,0 +1,5 @@
+Significance: patch
+Type: fixed
+Comment: Bumped version numbers
+
+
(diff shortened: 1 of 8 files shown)


haqadn (approved):
Looks good and works 👍🏼

PR #34753, opened 2023-12-21
IDC: add ip_requester option

bindlegirl:
Extending the IDC class so it supports requests from IP addresses. 

## Proposed changes:

* When a request goes out from an IP address, the ip_requester option is created that includes information about the IP and the expiry.
* When IDC logic pings the site for secret validation, any ip_requester info will be included in the response so there is no need for the second request for secret

### Other information:

## Testing instructions:

* TBD, will be described in D132906-code


projects/packages/identity-crisis/src/class-identity-crisis.php
```diff
@@ -1394,4 +1395,65 @@ public static function register_request_body( array $params ) {
...
+						$item['expires_at'] = time() + 300;
+						$updated_value      = true;
+					}
+					$updated_data[] = $item;
+				}
+			}
+			if ( ! $updated_value ) {
+				$updated_data[] = $ip_requester;
+			}
+			if ( $data !== $updated_data ) {
```
sergeymitr: Comparing the arrays might get a bit expensive if we do that on every request to WPCOM.
What do you think if we use some sort of "is changed" flag instead?

projects/packages/identity-crisis/src/class-identity-crisis.php
```diff
@@ -1394,4 +1395,65 @@ public static function register_request_body( array $params ) {
...
+				Jetpack_Options::update_option( 'identity_crisis_ip_requester', $updated_data );
```
sergeymitr: I'm a bit worried about concurrency. If the site is loaded via two different IP's simultaneously, some data might get overwritten.
I can't think of a scenario where it would pose a problem though. Do you think it would be safe?

projects/packages/identity-crisis/src/class-identity-crisis.php
```diff
@@ -1394,4 +1395,65 @@ public static function register_request_body( array $params ) {
 	public static function site_registered( $blog_id ) {
 		update_option( static::PERSISTENT_BLOG_ID_OPTION_NAME, (int) $blog_id, false );
 	}
+
+	/**
+	 * If URL is an IP, add the IP value to the ip_requester option with its expiry value.
+	 *
+	 * @throws Exception Thrown if unable to save the new IP requester.
+	 */
+	public static function set_ip_requester_for_idc() {
```
sergeymitr: Should we add some sort of transient here?
I think we could skip running the code below if it's already been run for that `siteurl` in the last few (five?) minutes.

sergeymitr (changes requested):
Just a couple of potential improvements.

projects/packages/identity-crisis/src/class-identity-crisis.php
bindlegirl: Good idea. Transient added in 3567c969ed36f5b49a5f220776ba85fea3b7ea7b

projects/packages/identity-crisis/src/class-identity-crisis.php
bindlegirl: It may be possible with sites that use multiple IPs and have heavy traffic. I'm just not sure how likely is that to happen. We could add some locking system or queuing but I worry we could be overcomplicating things that will never turn out to be a real problem.

projects/packages/identity-crisis/src/class-identity-crisis.php
bindlegirl: That `if ( $data !== $updated_data )` is not just expensive but plain wrong 😅 
If we even do try to update the option with the identical data, WP will reject it anyway and return false for the update_option. Thanks for catching the bad decision I made here.
I reorganized the whole method so I think this is no longer an issue.

projects/packages/identity-crisis/src/class-identity-crisis.php
sergeymitr: I agree, let's not complicate it any further than we have to 🙂

projects/packages/identity-crisis/src/class-identity-crisis.php
```diff
@@ -1359,11 +1364,16 @@ public static function add_secret_to_url_validation_response( array $response )
...
-	public static function url_is_ip() {
-		$hostname = wp_parse_url( Urls::site_url(), PHP_URL_HOST );
-		$is_ip    = filter_var( $hostname, FILTER_VALIDATE_IP ) !== false ? true : false;
+	public static function url_is_ip( $hostname = null ) {
+
+		if ( ! $hostname ) {
+			$hostname = wp_parse_url( Urls::site_url(), PHP_URL_HOST );
+		}
+
+		$is_ip = filter_var( $hostname, FILTER_VALIDATE_IP ) !== false ? $hostname : false;
```
sergeymitr, line 1376: We should be able to replace it with something like `return filter_var( $hostname, FILTER_VALIDATE_IP ) !== false;` to make the return value `bool` as declared in the PHPDoc.

sergeymitr (approved):
Works well 👍 
I added a potential code improvement, up to you if you'd like to adjust that.

PR #34754, opened 2023-12-21
Social | Improve Quick Share buttons UI

manzoorwanijk:
Quick share buttons in Jetpack/Social sidebar and the post publish panel had some a11y issues which this PR fixes and also updates the design for the share buttons:

## Proposed changes:

* Updates the design for the Quick Share buttons - XPSW2jfYPEVyIAOQOWXijm-fi-3942%3A1676
* Removes the modal and dropdown for the quick share panel
* Also improves the `IconTooltip` component which didn't work well on smaller screens

### Other information:

## Jetpack product discussion
PT: pdrWKz-1h5-p2

No

## Testing instructions:

- Goto Post edit page
- Create a post and hit publish
- On the post publish panel, confirm that you see the quick share buttons inline instead of a dropdown
- Confirm that "Learn more" link points to the Jetpack Social support page
- Close the post publish sidebar
- Open the Jetpack or Social sidebar
- Confirm that you see the quick share buttons under "Share this post" section
- Click on the info icon
- Confirm that you see the text as expected
- Confirm that "Learn more" link points to the Jetpack Social support page

Desktop

Screenshot 2023-12-21 at 7 24 10 PM
Screenshot 2023-12-21 at 7 24 01 PM
Screenshot 2023-12-21 at 7 26 36 PM

Mobile
IMG_5265


projects/js-packages/components/components/icon-tooltip/index.tsx
```diff
@@ -57,7 +58,8 @@ const IconTooltip: React.FC< IconTooltipProps > = ( {
 		focusOnMount: 'container' as const,
 		onClose: hideTooltip,
 		className: 'icon-tooltip-container',
-		inline: true,
+		inline,
+		shift: true,
```
manzoorwanijk: Icon tooltip did not work well on smaller devices. So `shift` prop is necessary to let the popover make the necessary adjustments to the `placement` in order to make sure that it stays visible.

| Before | After |
|--------|--------|
| IMG_5266 | IMG_5267 |
| IMG_5268 | IMG_5264 |

projects/js-packages/components/components/icon-tooltip/style.scss
```diff
@@ -91,6 +82,18 @@ $arrow-color: var( --jp-gray );
...
+	// __content styles here covers both the inline and portal modes
+	.components-popover__content {
+		padding: 24px;
+		width: 304px;
+		white-space: normal;
+		border-radius: 4px;
+		outline: none;
+		border: 1px solid $arrow-color;
+		max-width: 90vw;
+		margin: 0 10px;
```
manzoorwanijk, line 95: On smaller devices, the popover touches the edges, thus some margin along with max-width to fix that

| Before | After |
|--------|--------|
| IMG_5264 | IMG_5265 |
| Screenshot 2023-12-21 at 6 57 29 PM | Screenshot 2023-12-21 at 6 58 22 PM |

projects/js-packages/components/components/icon-tooltip/index.tsx
gmjuhasz: Do you think we should also pass in the `shift` property and default it to false as it was before? If it's a fix in every case then it should be okay like this, just asking to avoid unintentionally messing up the tooltip where others use it

gmjuhasz (commented):
Just added a question regarding the icon-tooltip component. Should we do that in a separate change if there are multiple changes with that component?

projects/js-packages/components/components/icon-tooltip/index.tsx
manzoorwanijk: Yes, the component is broken everywhere on smaller devices without this prop.

manzoorwanijk:
> Just added a question regarding the icon-tooltip component. Should we do that in a separate change if there are multiple changes with that component?

It's comparatively a new component and that bug is for each of its use-case.
So, I think it's fine to make this change here.

projects/js-packages/components/components/icon-tooltip/index.tsx
manzoorwanijk: OK, I have reverted it to a default value of `false` as it was before because that seems to be a safe approach.

spsiddarthan (approved):
Nice one, Manzoor! Works well on simple and Jetpack sites both. 

 It looks like the PR needs a rebase, but good to go otherwise.  I can give another approval in the morning and we can get it merged.

PR #34599, opened 2023-12-12
Jetpack Boost: Refactor Image Size Analysis

pyronaur:
## Proposed changes:
Refactor the entire image guide UI to React.js

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion
N/A

## Does this pull request change what data or activity we track or use?
N/A

## Testing instructions:
Check out the branch, build all the things, play use the image size analysis module and compare to the current trunk _(don't compare to svelte version because transitions don't exist in our React app at the moment - we don't have a way to use transitions right now)_

PR #34755, opened 2023-12-21
Jetpack Boost: Add Action Endpoints

pyronaur:
## Proposed changes:
This adds `useDataSyncAction` and the PHP endpoint handlers, allowing to create custom endpoints that can mutate data if necessary without passing the whole state back & forth.

### Other information:

## Testing instructions:
This patch is scoped to only include the package code - there's no direct implementation to test.

For implementation example, see #34599

* `jetpack_boost_register_action()` utility function: https://github.com/Automattic/jetpack/blob/49cd57ce20b8affed327e8794aaa6ccfe647bfdd/projects/plugins/boost/wp-js-data-sync.php#L32-L43

* Register PHP Endpoint: https://github.com/Automattic/jetpack/blob/49cd57ce20b8affed327e8794aaa6ccfe647bfdd/projects/plugins/boost/app/modules/image-size-analysis/data-sync/init.php#L78

* PHP Endpoint: https://github.com/Automattic/jetpack/blob/49cd57ce20b8affed327e8794aaa6ccfe647bfdd/projects/plugins/boost/app/modules/image-size-analysis/data-sync/Image_Size_Analysis_Fixer_Action.php#L12-L66

* useDataSyncAction implementation: https://github.com/Automattic/jetpack/blob/49cd57ce20b8affed327e8794aaa6ccfe647bfdd/projects/plugins/boost/app/assets/src/js/features/image-size-analysis/lib/image-fixer.ts#L22-L58

* React usage: https://github.com/Automattic/jetpack/blob/49cd57ce20b8affed327e8794aaa6ccfe647bfdd/projects/plugins/boost/app/assets/src/js/features/image-size-analysis/recommendations/table/table.tsx#L43-L50

### DIFF
--- a/projects/js-packages/react-data-sync-client/src/DataSyncHooks.ts
+++ b/projects/js-packages/react-data-sync-client/src/DataSyncHooks.ts
@@ -70,7 +70,13 @@ export function useDataSync<
 	params: Record< string, string | number > = {}
 ): DataSyncHook< Schema, Value > {
 	const datasync = new DataSync( namespace, key, schema );
-	const queryKey = [ key, ...Object.values( params ).sort() ];
+	const queryKey = [
+		key,
+		...Object.entries( params )
+			.sort( ( [ a ], [ b ] ) => a.localeCompare( b ) )
+			.map( ( [ , v ] ) => v ),
+	];
+
 	/**
 	 * Defaults for `useQuery`:
 	 * - `queryKey` is the key of the value that's being synced.
--- a/projects/js-packages/react-data-sync-client/src/DataSyncHooks.ts
+++ b/projects/js-packages/react-data-sync-client/src/DataSyncHooks.ts
@@ -29,9 +29,11 @@ export function DataSyncProvider( props: { children: React.ReactNode } ) {
 /**
  * React Query configuration type for DataSync.
  */
+type DataSyncMutation< Value > = Omit< UseMutationOptions< Value >, 'mutationKey' >;
+type DataSyncQuery< Value > = Omit< UseQueryOptions< Value >, 'queryKey' >;
 type DataSyncConfig< Schema extends z.ZodSchema, Value extends z.infer< Schema > > = {
-	query?: Omit< UseQueryOptions< Value >, 'queryKey' >;
-	mutation?: Omit< UseMutationOptions< Value >, 'mutationKey' >;
+	query?: DataSyncQuery< Value >;
+	mutation?: DataSyncMutation< Value >;
 };
 /**
  * This is what `useDataSync` returns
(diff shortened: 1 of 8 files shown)


projects/js-packages/react-data-sync-client/src/DataSyncHooks.ts
```diff
@@ -12,7 +12,11 @@ import React from 'react';
 import { z } from 'zod';
 import { DataSync } from './DataSync';
 
-const queryClient = new QueryClient();
+/**
+ * @REACT-TODO This is temporary. We need to allow each app to define their own QueryClient.
+ * All of the functions below will have to be moved to a factory wrapper
+ */
+export const queryClient = new QueryClient();
```
thingalon, line 19: 👍 I'm ok with this for now but would like to this on the board to make sure we take care of it before we call the migration done.

Before I add it to the board, a quick discussion: are we sure we need to? Will it cause any harm for DataSync to have its own queryClient that it exclusively works with, and for non-datasync REact Query stuff to live alongside it? I don't see a downside.

projects/js-packages/react-data-sync-client/src/DataSyncHooks.ts
```diff
@@ -63,7 +70,7 @@ export function useDataSync<
 	params: Record< string, string | number > = {}
 ): DataSyncHook< Schema, Value > {
 	const datasync = new DataSync( namespace, key, schema );
-	const queryKey = [ key, ...Object.values( params ) ];
+	const queryKey = [ key, ...Object.values( params ).sort() ];
```
thingalon: 🤔 I'm not 100% sure this is right. I think the key order matters. Imagine this scenario: 

You have two parameters: `foo` and `bar`. If you're sorting by value then `{foo:1,bar:2}` will have the same query cache key as `{foo:2,bar:1}`.

I think we'd want to sort by key for consistency, then pull out the values. Something like: 

```
const queryKey = [ key, ...Object.entries.sort( ( [ a ], [ b ] ) => b-a ).map( [ k, v ] => v ) ];
```

thingalon (changes requested):
One issue I see (inline), but otherwise this is great. Thank you for PRing it. :)

projects/js-packages/react-data-sync-client/src/DataSyncHooks.ts
pyronaur: It _may_ work but it's not correct. Think of queryClient as the namespace. Invalidation happens within a queryClient - so if I `invalidateAll`, I may accidentally invalidate applications I didn't intend to.

The fix is fairly simple - we'll just import `useDataSync` from boost, not from packages directly when we set this up - we can set those variables up in a lib directory statically. (something like `export const { useDataSync } = useDataSyncFactory()` )

projects/js-packages/react-data-sync-client/src/DataSyncHooks.ts
pyronaur: A-ha. Yes, nice catch. Fully agree.

projects/js-packages/react-data-sync-client/src/DataSyncHooks.ts
thingalon: 🤔 It's not a blocker here, but I don't understand. What is a scenario where we want to share a QueryClient with something that isn't DataSync?

Surely it'd be cleaner for DataSync to have its own cache, and any other use of ReactQuery to be kept separate?

projects/js-packages/react-data-sync-client/src/DataSyncHooks.ts
thingalon: Fixed in 39610c3.

thingalon (approved):
🚀

PR #34756, opened 2023-12-21
Contact Form: suppress PHP warning

monsieur-z:
## Proposed changes:

This PR suppresses the following PHP warning triggered by the Contact Form: `Undefined array key "formTitle" in /usr/local/src/jetpack-monorepo/projects/packages/forms/src/contact-form/class-contact-form.php on line 357`

### Other information:

## Jetpack product discussion

n/a

No.

## Testing instructions:

- Make sure you can see the server logs: `jetpack docker tail`
- Create a new post
- Add a _Contact Form_ block
- Open the preview
- Notice you don't see the warning mentioned above
- Go back to the editor
- Add a heading 2 before the form
- In the preview, check that the form has an `aria-label` attribute of which the value is the content of the heading 2

### DIFF
--- a/projects/packages/forms/src/contact-form/class-contact-form.php
+++ b/projects/packages/forms/src/contact-form/class-contact-form.php
@@ -354,7 +354,7 @@ class Contact_Form extends Contact_Form_Shortcode {
 			$url                     = apply_filters( 'grunion_contact_form_form_action', "{$url}#contact-form-{$id}", $GLOBALS['post'], $id );
 			$has_submit_button_block = str_contains( $content, 'wp-block-jetpack-button' );
 			$form_classes            = 'contact-form commentsblock';
-			$form_accessible_name    = $attributes['formTitle'] ? $attributes['formTitle'] : $post->post_title;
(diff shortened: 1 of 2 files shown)


projects/packages/forms/src/contact-form/class-contact-form.php
```diff
@@ -354,7 +354,7 @@ public static function parse( $attributes, $content ) {
 			$url                     = apply_filters( 'grunion_contact_form_form_action', "{$url}#contact-form-{$id}", $GLOBALS['post'], $id );
 			$has_submit_button_block = str_contains( $content, 'wp-block-jetpack-button' );
 			$form_classes            = 'contact-form commentsblock';
-			$form_accessible_name    = $attributes['formTitle'] ? $attributes['formTitle'] : $post->post_title;
+			$form_accessible_name    = isset( $attributes['formTitle'] ) && ! empty( $attributes['formTitle'] ) ? $attributes['formTitle'] : $post->post_title;
```
jeherve: You should be able to simplify this, if you check for `empty` you do not need to check for `isset`:

```suggestion
			$form_accessible_name    = ! empty( $attributes['formTitle'] ) ? $attributes['formTitle'] : $post->post_title;
```

andrii-lysenko (approved):
Looks good and works as described 👍

PR #34759, opened 2023-12-21
Sharing block/wp settings suggest deactivating module

### LINKED ISSUE #34727: Sharing Block: In wp-admin Sharing settings suggest deactivating sharing module and use block instead
Context: https://github.com/Automattic/jetpack/pull/34673#pullrequestreview-1786354226

andrii-lysenko:
Fixes https://github.com/Automattic/jetpack/issues/34727

## Proposed changes:

Adding a small notice below inviting users to disable Old sharing module (note only when block is available)

Screenshot 2023-12-21 at 11 06 18

### Other information:

## Testing instructions:

- Checkout changes locally, build `jetpack build plugins/jetpack` and spin up your site locally
- You should see new view only when beta blocks enabled: `define( 'JETPACK_BLOCKS_VARIATION', 'beta' );` and  for block-based theme (eg Twenty Twenty-Four) 
- Check `/wp-admin/options-general.php?page=sharing`.
- Check the new promt and verify that link is working


keoshi (dismissed):
Looks good to me!

As a tiny tweak we can align the copy with https://github.com/Automattic/jetpack/pull/34619 and go with:

> You are using a block-based theme. You can disable Jetpack’s legacy sharing buttons and add a sharing block to your theme’s template instead.

projects/plugins/jetpack/modules/sharedaddy/admin-sharing.css
```diff
@@ -450,4 +450,18 @@ body.settings_page_sharing .advanced input[type=submit] {
...
\ No newline at end of file
+}
+
+
+.admin-sharing-settings__block-theme-description {
+	margin-top: 1rem;
+	font-style: italic;
+}
+
+a.dops-card__link {
```
monsieur-z, line 461: Is this necessary? Value is `pointer` by default for links.

monsieur-z (dismissed):

projects/plugins/jetpack/modules/sharedaddy/admin-sharing.css
andrii-lysenko: I think it wasn't for some reason here.

monsieur-z (approved):

PR #34760, opened 2023-12-21
AI Helper: Add moderation and max tokens

dereksmart:
## Proposed changes:

Moves the moderation check and sets the max_tokens from OpenAI()->request_chat_completion() to the client. 

### Other information:
In D132096-code we updated the jetpack-ai-query endpoint to not hardcode `openai` as the default feature. A side effect of that was making these checks required for every feature calling `request_chat_completion`, which in turn had side effects of some internal features not working, since they needed more tokens - also, some of the features were already doing their own moderation calls, so making that default behavior in a shared method adds unnecessary latency and requests. 

Once this code is deployed to wpcom, we'll be able to remove the default checks from request_chat_completion method and update the jetpack-ai-query endpoint to do the same. This is being handled in D132925-code. 

## Jetpack product discussion

p1703182024064289-slack-C069HLHBMRP

No

## Testing instructions:

jetpack rsync to wpcom 
I'm not actually sure what feature uses this. 
It sounds like there is nothing currently actually using this. p1703187120335959-slack-C054LN8RNVA

### DIFF
--- a/projects/plugins/jetpack/_inc/lib/class-jetpack-ai-helper.php
+++ b/projects/plugins/jetpack/_inc/lib/class-jetpack-ai-helper.php
@@ -208,7 +208,25 @@ class Jetpack_AI_Helper {
 				),
 			);
 
-			$result = ( new OpenAI( 'openai', array( 'post_id' => $post_id ) ) )->request_chat_completion( $data );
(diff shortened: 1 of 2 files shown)


kthai15 (approved):
Didn't test (also this code might not even be used anymore), but looks good to me.

PR #34763, opened 2023-12-22
[Plugin Installer] Ensure language packs get installed when installing plugins

thingalon:
Fixes https://github.com/Automattic/jpop-issues/issues/8747

When installing Boost or CRM from My Jetpack, appropriate language packs were not being installed. This PR addresses this by allowing the usual admin-hooks that occur after installing a plugin to run when installing via My Jetpack / the plugin installer.

Props to @dilirity for doing the investigation work behind this.

## Proposed changes:
* Load admin-filters before installing plugins, so that WordPress can run post-install processes on installation.

### Other information:

no

## Testing instructions:

* Install Jetpack
* Set the language to French
* Make sure that Jetpack is properly in French
* Use "My Jetpack" to install Jetpack Boost
* Ensure that the appropriate language pack is installed automatically, and the Boost UI is in French.

**Notes**
I've noticed that My Jetpack shows Boost's plan selection page twice if you choose the free version.
- The first time it shows, Jetpack is showing the "Choose a plan" component itself without having installed Boost.
- It then installs Boost when you click on a plan / free
- Then Boost shows its getting started / choose a plan screen.

If you see that, please remember that the first screen will show in the language pack that *Jetpack* is using. The second time it is *Boost* and its language pack. I recommend getting past the "Getting started" process to ensure boost is in French appropriately while testing this issue.

I've filed a report of the duplicate-screen bug here: https://github.com/Automattic/jetpack/issues/34764

### DIFF
--- a/projects/packages/plugins-installer/changelog/fix-language-installation
+++ b/projects/packages/plugins-installer/changelog/fix-language-installation
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Ensure that language packs are installed after installing a new plugin
--- a/projects/packages/plugins-installer/src/class-plugins-installer.php
+++ b/projects/packages/plugins-installer/src/class-plugins-installer.php
@@ -74,6 +74,9 @@ class Plugins_Installer {
 			return new WP_Error( 'not_allowed', __( 'You are not allowed to install plugins on this site.', 'jetpack-plugins-installer' ) );
 		}
 
+		// Initialize admin filters to make sure WordPress post-install hooks run. Handles things like language packs.
+		include_once ABSPATH . '/wp-admin/includes/admin-filters.php';
+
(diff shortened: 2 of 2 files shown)


dilirity (approved):
It works! Great work <3

leogermani:
Hi, I'm not working directly with this for a while now, but if it's worth, here is some context. I remember we had to activate (not isntall) plugins in "quiet" mode because some plugins tried to redirect the user to their settings page after activation and this would break things. Not sure if this is related to this PR at all, but just sharing as it might be valuable context

kraftbj (approved):
Let's see if we can check what @leogermani brought up, but we can do that once this is in trunk... may help to get reports. If we find unexpected redirects, we can see if we can be more surgical about adding specific filters.

PR #34768, opened 2023-12-22
Remove jetpack connection banners from non-Jetpack pages

jboland88:
## Proposed changes:

* This diff removes the Jetpack connection banners from wp-admin pages that are not Jetpack pages.
* The banner currently shows on the wp-admin dashboard and plugins page if Jetpack is not connected (see image below)
Screenshot 2023-12-22 at 9 36 48 AM

## Jetpack product discussion

pbNhbs-922-p2#comment-19163

No

## Testing instructions:

* Apply this patch locally or using the Jetpack Beta plugin. If you are testing on your local install, you may need to reset the Jetpack Options
* DO NOT Connect Jetpack 
* Visit the wp-admin dashboard and plugins page and confirm that the connection banner (shown above) does not appear
* Make sure the full-screen connection component still shows at Jetpack > dashboard and that you can connect the site from here.

### DIFF
--- a/projects/plugins/jetpack/_inc/lib/admin-pages/class.jetpack-admin-page.php
+++ b/projects/plugins/jetpack/_inc/lib/admin-pages/class.jetpack-admin-page.php
@@ -118,29 +118,6 @@ abstract class Jetpack_Admin_Page {
 
 		// Attach page specific actions in addition to the above.
 		$this->add_page_actions( $hook );
-
-		// If the current user can connect Jetpack, Jetpack isn't connected, and is not in offline mode, let's prompt!
-		if ( current_user_can( 'jetpack_connect' ) && $connectable ) {
-			$this->add_connection_banner_actions();
-		}
-	}
-
-	/**
-	 * Hooks to add when Jetpack is not active or in offline mode for an user capable of connecting.
-	 */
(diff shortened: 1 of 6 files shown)


elliottprogrammer (changes requested):
Tested, works great as described. Code looks all good.

However I do see some failing e2e tests checking for the connection banner(s) to be visible, that I believe we should remove/disable.

jboland88:
> Tested, works great as described. Code looks all good.
> 
> However I do see some failing e2e tests checking for the connection banner(s) to be visible, that I believe we should remove/disable.

Thanks @elliottprogrammer ! 
I think I've got the pre-connection tests sorted, but some other tests are failing now 🙃 I'll see if I can get those sorted

jboland88:
@elliottprogrammer - looks like the post-connect E2E test is failing, but I see this failing on other open, recent PRs, it does not appear to be related.

elliottprogrammer:
> @elliottprogrammer - looks like the post-connect E2E test is failing, but I see this failing on other open, recent PRs, it does not appear to be related.

Ahh ok, sounds good, thanks for looking into it! 👍

elliottprogrammer (approved):
LGTM! 👍

PR #22322, opened 2022-01-12
PHPCS: Fix errors in Jetpack root classes, part 1

anomiex:
#### Changes proposed in this Pull Request:

Mostly comment changes and declaring things `public`.
Remove a few unused function args (instead of documenting them),
some bad string concatenation, and do minimal code changes needed
because lines are touched.

All the code changes needed by sniffs will come in later parts.

#### Jetpack product discussion

Part of #21951

#### Does this pull request change what data or activity we track or use?

No

#### Testing instructions:

* Is CI happy?
* Did this break anything?

PR #34770, opened 2023-12-22
Allow is_truthy to properly handle occasional objects.

zinigor:
Fixes a fatal error that can occur on WordPress.com API requests when third party plugins add data to responses. 

## Proposed changes:
* Add an additional check to is_truthy for non-object arguments.

### Other information:

## Jetpack product discussion

p1703178139555289-slack-C02ECE3ML

No.

## Testing instructions:

To test this change you will need to sandbox public-api.wordpress.com, update wpcom with this PR, as well as introduce a new filter into your test Jetpack site:

```php
add_action( 'wpcom_api_site_roles', 'add_bogus_stdclass_roles' );
function add_bogus_stdclass_roles( $roles ) {
    $bogus_role = new WP_Role( 'bogus_role', 'Bogus role' );
    $bogus_role->capabilities = new stdClass();

    $bogus_role->capabilities->pmpro_role_1 = new stdClass();
    $bogus_role->capabilities->pmpro_role_1->read = 1 ;

    $roles[] = $bogus_role;
    return $roles;
}
```

After adding this filter and updating wpcom, use the WordPress.com Developer console to query the site's API using the roles endpoint `v1.1/sites/SITE_ID/roles`. Observe no fatals in the sandbox. Without this patch but with the filter installed, you will see a fatal on your sandbox.


projects/plugins/jetpack/class.json-api.php
```diff
@@ -202,12 +202,17 @@ public function add( WPCOM_JSON_API_Endpoint $endpoint ) {
...
 	/**
-	 * Determine if a string is truthy.
+	 * Determine if a string is truthy. If it's not a string, which can happen with
+	 * not well-formed data coming from Jetpack sites, we still consider it a truthy value.
 	 *
 	 * @param string $value "1", "t", and "true" (case insensitive) are falsey, everything else isn't.
 	 * @return bool
 	 */
 	public static function is_truthy( $value ) {
+		if ( ! is_string( $value ) ) {
```
CGastrell: Haven't backtraced the usage of this, maybe you can clarify. But adding this would mean that values such as `0`, `null` and `false` would return `true` here. Is this what we want?

kraftbj: Nod, this would change the behavior for those particular things, which I would think we should try to maintain. These two functions, `is_truthy` and the `is_falsy` below are interestingly setup (only expecting "1", "t", "true" as strings (thus excluding bool `true` and the number 1) and false only expecting strings as well. (Perhaps, originally, it assumed the string casting would always work so `1`, `true` would convert to strings?)

Overall, I'd like to understand better (just driving by the comment right now) why these functions exist and why they're so narrow-scoped on what's eligible for true/false determination. Maybe they should be depreciated for `is_string_truthy`, etc :) with a doing_it_wrong if a non-string is passed to it? Something that would raise a flag that something unexpected is happening without the fatal.

Also, let's update `is_falsy` below since it could suffer from the same fatal for non-string elements being passed to it.

Perhaps another way could be a try/catch for the casting to a string. We can catch the type of fatal we say for a non-convertible type and return a doing_it_wrong, while allowing for convertible other types through.

projects/plugins/jetpack/class.json-api.php
```diff
@@ -202,12 +202,17 @@ public function add( WPCOM_JSON_API_Endpoint $endpoint ) {
 	}
 
 	/**
-	 * Determine if a string is truthy.
+	 * Determine if a string is truthy. If it's not a string, which can happen with
+	 * not well-formed data coming from Jetpack sites, we still consider it a truthy value.
 	 *
 	 * @param string $value "1", "t", and "true" (case insensitive) are falsey, everything else isn't.
```
anomiex: Pre-existing issue, but since we're here
```suggestion
	 * @param string $value "1", "t", and "true" (case insensitive) are truthy, everything else isn't.
```

projects/plugins/jetpack/class.json-api.php
anomiex: I note an odd edge case, `is_truthy( true )` would work because `(string) true` => `'1'`, but `is_falsy( false )` would not because `(string) false` => `''` and the empty string is not considered falsey by `is_falsy`.

I see `is_falsy` was added in 9ed52fc36e1c124a591dca48a8daddf0d4d9bfb1 to do something like `if ( is_falsy( $value ) ) { $value = false }`, handling cases where someone supplied "false" to the API parameter. Which probably explains the oversight.

`is_truthy` came in a big merge in d47c5440fb0c1c8a9df44b13f89f357aecebf370 and seems to have been used like `(bool) $value` there to only accept certain strings as "true".

The doc block specifying "string" was added much later, in #22322, and I may have guessed wrong at what exactly might be passed in.

zinigor: I have updated both methods, thank you all!
@anomiex that's a great find, but I think it's a separate fix, let's not include it into this PR.

zinigor: @anomiex it looks like your suggestion will actually fix the failing tests, I'm going to try it now.

anomiex (approved):
We may want to come back at some point and re-allow integer 1 and 0, but that can be a followup if you want.

projects/plugins/jetpack/class.json-api.php
```diff
@@ -202,12 +202,21 @@ public function add( WPCOM_JSON_API_Endpoint $endpoint ) {
...
-	 * @param string $value "1", "t", and "true" (case insensitive) are falsey, everything else isn't.
+	 * @param string $value "1", "t", and "true" (case insensitive) are truthy, everything else isn't.
 	 * @return bool
 	 */
 	public static function is_truthy( $value ) {
+		if ( true === $value ) {
+			return true;
+		}
+
+		if ( ! is_string( $value ) ) {
```
spsiddarthan, line 216: @zinigor, should this perhaps be `if ( ! is_string( $value ) ) || is_integer( $value ) )`?

In calypso, we a have screen, where toggling off the check button in the screenshot, sends a payload like `shared : 1` to the end-point. The `is_truthy` function returns false when the value is 1.  

I have tweaked calypso to send true or false to fix our issue, but maybe we should handle this differently for integers? 

Screenshot 2024-01-23 at 5 15 45 PM

PR #34772, opened 2023-12-22
Add interstitial for Stats

jboland88:
## Proposed changes:

* This PR adds a product interstitial for Stats in My Jetpack

image

### Other information:

## Jetpack product discussion

pbNhbs-9bI-p2

No

## Testing instructions:

* Spin up a JN site with Jetpack Beta pointing to this branch
* Visit `/wp-admin/admin.php?page=my-jetpack#/add-stats`
* Confirm the new interstitial matches the screenshot above, and the CTA works
* Test the CTA on the My Jetpack stats card - without a plan it should say "learn more". With a free or tiered stats plan, it should say "Upgrade". With a PWYW or legacy commercial plan, it should say "View"


projects/packages/my-jetpack/src/products/class-stats.php
```diff
@@ -162,23 +173,21 @@ public static function is_upgradable() {
...
 				}
-			}
+			);
+
+			return ! empty( $upgradeable_stats_purchases );
 		}
-		return true;
+
+		// If there are no plans found, don't consider the product as upgradeable
+		return false;
```
CodeyGuyDylan, line 205: Just curious, what do you think the reasoning behind defaulting to `true` was before? I can't think of one, just want to make sure there isn't a good reason we are missing

CodeyGuyDylan (dismissed):
This LGTM, just one question 😄

projects/packages/my-jetpack/src/products/class-stats.php
jboland88: Good question -  I am not quite sure either 🤔 
It seems like the old approach was a blacklist approach and the new approach is more of a whitelist 🤷

CodeyGuyDylan (dismissed):
When I have a Stats commercial license, I see the "Upgrade" button but am taken to Stats PWYW when I click on it, which would be a downgrade
image
image

robertsreberski:
After the change, clicking update when having Commercial license leads to Commercial PWYW:
CleanShot 2024-01-08 at 12 33 38@2x

LGTM

robertsreberski (approved):

PR #34774, opened 2023-12-23
Node: match version used by WordPress itself

kraftbj:
In https://core.trac.wordpress.org/ticket/59663, WordPress Core jumped from being pretty far behind on the LTS version of Node it required to 20.x, and specifically 20.10. To simplify things since we're close, what about us upping a tad to match?

Separately, would need to consider this for Calypso, but wanted to run it here for opinions first.

## Proposed changes:

* Update nvmrc to 20.10 from 20.8.1
* Update package.json for the proper engine.

### Other information:

## Jetpack product discussion

None as of now.

No.

## Testing instructions:

* Run `nvm use` to update to 20.10 (possibly need `nvm install` first).
* Run pnpm install, jetpack install, jetpack build, etc to ensure things build right.
* Verify CI is happy.


anomiex (approved):
Wow, Gutenberg updated to node 20 too! :tada: 

I don't see a problem with us bumping the version, other than the possibility of Calypso not doing so and the mismatch potentially being annoying.

anomiex:
From [their announcement post](https://make.wordpress.org/core/2023/12/20/updating-wordpress-to-use-more-modern-versions-of-node-js-npm-2/):

> Will the required version be updated to 22.x when it’s released in April 2024?
>
> Maybe. 

I wonder if they'll decide to wait until 22.x becomes LTS in October, instead of doing it right away in April. If they do go ahead early, I wonder if we and Calypso should hold back until October as we've done previously.

PR #34157, opened 2023-11-16
Jetpack Google Fonts: Print font faces that are in use instead of ALL

arthur791004:
Fixes #34163

## Proposed changes:

This is a follow-up to #33203

* Turn off Font Face hooks in Core that print **ALL** fonts from theme json. Instead, this PR collects fonts that are used by global styles or block-level settings first and then prints the necessary fonts
  * The hooks registered in Core are [here](https://github.com/WordPress/wordpress-develop/blob/0b8ca16ea3bd9722bd1a38f8ab68901506b1a0e7/src/wp-includes/default-filters.php#L749) and [here](https://github.com/WordPress/wordpress-develop/blob/0b8ca16ea3bd9722bd1a38f8ab68901506b1a0e7/src/wp-admin/includes/admin-filters.php#L174)

* In the follow-up iteration, we can continue to detect the font style & font weight that are in use. The difficult is the global styles or block-level settings might only define the font family, and we need a way to know what's the default font style & font weight.

For example, global styles might define a font weight for ALL headings (h1, h2, ..., h6). Hence, we have to include the specific font weight for the font families that are defined by other headings (e.g.: h1) here. However, we might not be able to know the default font style & font weight for all blocks. Any thoughts?

```json
{
  "h1": {
    "typography": {
      "fontFamily": "var(--wp--preset--font-family--spectral)"
    }
  },
  "heading": {
    "typography": {
      "fontWeight": "600"
    }
  }
}
```

### Other information:

- [ ] Have you written new tests for your changes, if applicable?
- [ ] Have you checked the E2E test CI results, and verified that your changes do not break them?
- [ ] Have you tested your changes on WordPress.com, if applicable (if so, you'll see a generated comment below with a script to run)?

## Jetpack product discussion

* p1700040028362329-slack-C4GAQ900P
* p7DVsv-jib-p2

## Does this pull request change what data or activity we track or use?

* No

### Preparation

**JN Sites**

* Create a [Jurassic Ninja sandbox](https://jurassic.ninja/create/?jetpack-beta&branch=feat/print-font-faces-in-use) against this PR.
* Ensure the WP version is 6.4 and above
* Install [the latest Gutenberg](https://github.com/WordPress/gutenberg/releases), e.g.: v17.0.2
* Go to `/wp-admin/admin.php?page=jetpack_modules` to activate the “Google Fonts (Beta)” module

**Atomic sites**

* Install Jetpack Beta Tester plugin
* Install changes from this branch, feat/print-font-faces-in-use

[...]

PR #34776, opened 2023-12-26
Fix fonts not loading in editor when Gutenberg not activated, or when the post type has any metabox registered for it

### LINKED ISSUE #34777: Editor fonts don't match the frontend for Sensei post types
### Quick summary

When editing a course, lesson or message, the fonts don't match the frontend.

### Steps to reproduce

1. Activate the Course theme (although this issue doesn't appear to be isolated to just the Course theme).
2. Create a course, lesson or message.
3. View it on the frontend.
4. Notice that some of the fonts don't match. The frontend is correct; the editor is not.

The editor and frontend to match.

### What actually happened

The fonts in the editor are incorrect.

### Impact

All

### Available workarounds?

_No response_

### Logs or notes

Identified in p6rkRX-6YK-p2.

Imran92:
Fixes https://github.com/Automattic/jetpack/issues/34777

## Proposed changes:

We noticed that when we installed Sensei on WPCOM, the fonts on in the editors of any of our custom post types (Courses, Lessons, Questions etc) didn't match the Theme's fonts. They always fell back to the default themes. So the editor looked broken.

But the Posts and Pages post type editors look alright and the fonts were loading. So it appeared that there was something wrong with Sensei's CPTs. I created a couple more new CPTs and they were good too. So, the issue was with Sensei's CPTs only, not just any CPT. So I tried to find what's wrong with Sensei's CPT. I created a couple of new CPTs with matching CPT configuration of Sensei's Course and Lesson CPT, but to no avail, the issue wasn't reproduced. It also wasn't reproduced locally. So the issue could be caused from anywhere, Sensei, Calypso, Gutenberg, Jetpack or somewhere very specific to WPCOM.

So, then I tried finding differences visually in the editor. And noticed that all of Sensei's editable post types have one or more metaboxes registered to them (using `add_meta_box`). So I added a metabox to one of my new CPTs with a snippet, and only then, I could replicate, fonts were not loading for that particular CPT in editor now as well.

So, the issue boiled down to - If a CPT has any metabox attached, the fonts don't load.

But out of curiosity, I tried another thing. I deactivated Gutenberg plugin. And after that, fonts stopped loading for all Posts and Pages as well. So now no Gutenberg editor for any post types were loading any fonts and was falling back to default.

So it's not only metabox issue, also Gutenberg had something to do with it. But happens only on wpcom, so based on everything, it looked like it could be on Calypso or Jetpack.

So after digging in, I found this code, where fonts are removed from both frontend and backend ('wp_head' and 'admin_print_styles') but then re-added only for 'wp_head', which means fonts don't get printed on the Admin side. Which explained why we didn't get the fonts in the editor side. That's why I've attached the font print action to 'admin_head' and that fixed the issue.

https://github.com/Automattic/jetpack/blob/47a0f6903d08f429e17c6c1a6c7eb60d4e3309fb/projects/plugins/jetpack/modules/google-fonts/current/class-jetpack-google-font-face.php#L30-L48

Though it got fixed, a question remained. Why did it only happen with CPTs with metaboxes when Gutenberg was activated? And why it started happening all over after Gutenberg was disabled?

Well, this answer is not in Jetpack's code, it's in Gutenberg, and in two different places there -

In Gutenberg, there's this following piece of compat code that manually attaches the fonts to the CSS of the editor.
https://github.com/WordPress/gutenberg/blob/c2a54cd28ecab83d7f770cfbf9e8db3649f0f960/lib/compat/wordpress-6.4/fonts/fonts.php#L65-L78

So even though we removed the font printing in Jetpack, we can't see the effect of it as it gets attached from a separate place.

But that still doesn't explain the metabox. This was probably the final piece in the puzzle. The code above attaches the fonts only to the iframe of the editor. But when the CPT has any metaboxes, Gutenberg editor is not loaded in an iframe, that's why these font styles didn't get attached. You can see the logic for loading iframe below -
https://github.com/WordPress/gutenberg/blob/c2a54cd28ecab83d7f770cfbf9e8db3649f0f960/packages/edit-post/src/components/visual-editor/index.js#L74-L77

So this is how it looks before and after the fix -

#### Before
https://github.com/Automattic/jetpack/assets/6820724/d9c3f2cf-1861-4374-8522-4905dc55c0e6

#### After
https://github.com/Automattic/jetpack/assets/6820724/3d04985d-a87d-489d-b433-5ab24f15558d

### Other information:

## Jetpack product discussion

Could be related https://github.com/Automattic/wp-calypso/issues/84611

n/a

## Testing instructions:

1. Apply this to your WoA site
2. Add a metabox to the `post` Post Type. You can use the snippet added below
```
add_action( 'init', function() {
	add_action( 'add_meta_boxes', 'add_metas', 1 );
} );

function add_metas() {
	add_meta_box( 'random-metabox', 'RandomMetabox', 'add_a_metabox', 'post', 'side', 'high' );
}

function add_a_metabox() {

}
```
3. Install Course theme (You can use any other theme also)
4. Go to Posts -> Add New Post
5. Add a Header block (h1)
6. Make sure the font looks the same as it does in the **After** video above (League Gothic font)
7. Now deactivate Gutenberg and make sure the font still looks the same
8. Now switch to Trunk and apply those changes to your WoA
9. Check that the fonts have changed
10. Enable Gutenberg, the font should still not be the correct one
11. Now remove the snippet. It should be alright again.

### DIFF
--- a/projects/plugins/jetpack/modules/google-fonts/current/class-jetpack-google-font-face.php
+++ b/projects/plugins/jetpack/modules/google-fonts/current/class-jetpack-google-font-face.php
@@ -43,12 +43,11 @@ class Jetpack_Google_Font_Face {
 	 * Turn off hooks to print fonts on wp-admin, except for GB editor pages.
 	 */
 	public function current_screen() {
-		if ( $this->is_block_editor() ) {
-			return;
-		}
-
 		remove_action( 'admin_print_styles', 'wp_print_fonts', 50 );
-		remove_action( 'admin_print_styles', 'wp_print_font_faces', 50 );
+
+		if ( ! $this->is_block_editor() ) {
+			remove_action( 'admin_print_styles', 'wp_print_font_faces', 50 );
+		}
 	}
 
 	/**
--- a/projects/plugins/jetpack/modules/google-fonts/current/class-jetpack-google-font-face.php
+++ b/projects/plugins/jetpack/modules/google-fonts/current/class-jetpack-google-font-face.php
@@ -24,6 +24,7 @@ class Jetpack_Google_Font_Face {
 	public function __construct() {
 		// Turns off hooks to print fonts
 		add_action( 'wp_loaded', array( $this, 'wp_loaded' ) );
+		add_action( 'current_screen', array( $this, 'current_screen' ), 10 );
 
 		// Collect and print fonts in use
 		add_action( 'wp_head', array( $this, 'print_font_faces' ), 50 );
@@ -36,6 +37,18 @@ class Jetpack_Google_Font_Face {
 	public function wp_loaded() {
 		remove_action( 'wp_head', 'wp_print_fonts', 50 );
 		remove_action( 'wp_head', 'wp_print_font_faces', 50 );
+	}
+
+	/**
+	 * Turn off hooks to print fonts on wp-admin, except for GB editor pages.
+	 */
+	public function current_screen() {
+		if ( $this->is_block_editor() ) {
+			return;
+		}
+
+		remove_action( 'admin_print_styles', 'wp_print_fonts', 50 );
+		remove_action( 'admin_print_styles', 'wp_print_font_faces', 50 );
 	}
 
 	/**
@@ -191,5 +204,21 @@ class Jetpack_Google_Font_Face {
 		}
 
 		return $font_family;
+	}
+
+	/**
+	 * Check if the current screen is the block editor.
+	 *
+	 * @return bool
+	 */
+	public function is_block_editor() {
+		if ( function_exists( 'get_current_screen' ) ) {
+			$current_screen = get_current_screen();
+			if ( ! empty( $current_screen ) && method_exists( $current_screen, 'is_block_editor' ) && $current_screen->is_block_editor() ) {
+				return true;
+			}
+		}
+
+		return false;
 	}
 }
--- a/projects/plugins/jetpack/modules/google-fonts/current/class-jetpack-google-font-face.php
+++ b/projects/plugins/jetpack/modules/google-fonts/current/class-jetpack-google-font-face.php
@@ -28,6 +28,7 @@ class Jetpack_Google_Font_Face {
 
 		// Collect and print fonts in use
 		add_action( 'wp_head', array( $this, 'print_font_faces' ), 50 );
+		add_action( 'admin_head', array( $this, 'print_font_faces' ), 50 );
 		add_filter( 'pre_render_block', array( $this, 'collect_block_fonts' ), 10, 2 );
 	}
 
--- a/projects/plugins/jetpack/modules/google-fonts/current/class-jetpack-google-font-face.php
+++ b/projects/plugins/jetpack/modules/google-fonts/current/class-jetpack-google-font-face.php
@@ -24,11 +24,9 @@ class Jetpack_Google_Font_Face {
 	public function __construct() {
 		// Turns off hooks to print fonts
 		add_action( 'wp_loaded', array( $this, 'wp_loaded' ) );
-		add_action( 'admin_init', array( $this, 'admin_init' ), 10 );
 
 		// Collect and print fonts in use
 		add_action( 'wp_head', array( $this, 'print_font_faces' ), 50 );
-		add_action( 'admin_head', array( $this, 'print_font_faces' ), 50 );
 		add_filter( 'pre_render_block', array( $this, 'collect_block_fonts' ), 10, 2 );
 	}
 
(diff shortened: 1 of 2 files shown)


arthur791004:
> In Gutenberg, there's this following piece of compat code that manually attaches the fonts to the CSS of the editor.

But the Core prints the fonts inside the `_wp_get_iframed_editor_assets` function 🤔

https://github.com/WordPress/wordpress-develop/blob/e9f335df2f990866f8b8f1acbe1852c592dec3e6/src/wp-includes/block-editor.php#L368

arthur791004:
Oh yes, the assets won't be added to the editor if it's not iframed, e.g.: `hasMetaBoxes` 🥲

```js
const isToBeIframed =
		( ( hasV3BlocksOnly || ( isGutenbergPlugin && isBlockBasedTheme ) ) &&
			! hasMetaBoxes ) ||
		renderingMode === 'template-only';
```

https://github.com/WordPress/gutenberg/blob/c2a54cd28ecab83d7f770cfbf9e8db3649f0f960/packages/edit-post/src/components/visual-editor/index.js#L74

projects/plugins/jetpack/modules/google-fonts/current/class-jetpack-google-font-face.php
```diff
@@ -28,6 +28,7 @@ public function __construct() {
 
 		// Collect and print fonts in use
 		add_action( 'wp_head', array( $this, 'print_font_faces' ), 50 );
+		add_action( 'admin_head', array( $this, 'print_font_faces' ), 50 );
```
arthur791004: It would be better to keep using the `admin_print_styles` hook. WDYT?

Imran92: @arthur791004  Yap, that would make the most sense and I considered it. But I was concerned that it'd rollback the intended outcome from the https://github.com/Automattic/jetpack/blob/47a0f6903d08f429e17c6c1a6c7eb60d4e3309fb/projects/plugins/jetpack/modules/google-fonts/current/class-jetpack-google-font-face.php#L53 function from this PR (https://github.com/Automattic/jetpack/pull/34157) , that is loading only the fonts that are in use instead of printing all the fonts, only in the editor though. Do you think it's okay if it does? If it's alright, then I can update my PR to remove the codes related to `admin_print_styles` hook here instead (I think I'll remove the `admin_init` function from here)

arthur791004: I think we have to print ALL fonts in the page/post editor as people may change the font of the block there. However, I'd prefer to print them only in the editor (without iframe) if possible as those styles are too heavy and we don't need them outside the editor.

Imran92: @arthur791004 makes sense! I have added codes here https://github.com/Automattic/jetpack/pull/34776/commits/52f1527f6bc6e2208b4d8f5b1293c8a649a2814a to load fonts only in GB editor pages. I switched from `admin_init` hook to `current_screen` hook because otherwise the current screen related information were not loaded.

arthur791004: It seems to print the font CSS twice when the editor is iframed.

image

projects/plugins/jetpack/modules/google-fonts/current/class-jetpack-google-font-face.php
```diff
@@ -40,9 +40,13 @@ public function wp_loaded() {
...
-	 * Turn off hooks to print fonts on wp-admin
+	 * Turn off hooks to print fonts on wp-admin, except for GB editor pages.
 	 */
-	public function admin_init() {
+	public function current_screen() {
+		if ( $this->is_block_editor() ) {
+			return;
+		}
+
 		remove_action( 'admin_print_styles', 'wp_print_fonts', 50 );
```
arthur791004, line 46: We have to turn off this to avoid printing the font CSS twice.

Imran92: I've updated it here https://github.com/Automattic/jetpack/pull/34776/commits/5f45928d03d435d979a8ac01fb36611bce5042f5

arthur791004 (approved):
Tested and looks good to me! Nice fix 🎉

Imran92:
Thanks a ton @arthur791004 ! <3 I haven't added any milestone in this PR. Should I just go ahead and merge it? I am asking this because some of our patterns look broken on WPCOM because of this font issue and this is one of the earlier things they notice after signing up. Is there a way to prioritize releasing this fix on wpcom? Or this will by default be a part of this week's release? Thanks again

arthur791004:
> Should I just go ahead and merge it?

Let's go 🚢

> Is there a way to prioritize releasing this fix on wpcom?

Do you mean the simple sites? If so, yes. Jetpack has the daily release to wpcom. But for atomic sites, we have to wait for the next official release of Jetpack.

PR #34781, opened 2023-12-26
Add the block=1 query param to the iframe url when called from the Likes Block

### LINKED ISSUE #85707: Like block: add `&block=1` parameter to iframe
Context: pdDOJh-2Rs-p2

JuanLucha:
Fixes https://github.com/Automattic/wp-calypso/issues/85707

## Proposed changes:
No functional change is added. The likes widget URL is called in an iframe. When the iframe is called from the new Likes Block, we want to signal that call with a &block=1 at the end of the URL.

### Other information:

No

## Testing instructions:
- Don't apply this PR yet.
- Setup your jt site.
- Make sure you have the like button activated in the jetpack setup screen.
- Go to one of your posts in your jt site.
- Check the like button widget URL in the iframe. It shouldn't have a `&block=1` parameter at the end.
- Apply this PR.
- Activate the beta blocks (`define( 'JETPACK_BLOCKS_VARIATION', 'beta' );`).
- Build the jetpack plugin.
- Create or edit a post and add the Like Block. Publish or update the post.
- Go to the URL of your post and check the HTML code. The iframe of the like widget should include a `&block=1` at the end.

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/like/like.php
+++ b/projects/plugins/jetpack/extensions/blocks/like/like.php
@@ -87,7 +87,7 @@ function render_block( $attr, $content, $block ) {
 		$bloginfo = get_blog_details( (int) $blog_id );
 		$domain   = $bloginfo->domain;
 		$version  = '20231201';
-		$src      = sprintf( '//widgets.wp.com/likes/index.html?ver=%1$d#blog_id=%2$d&amp;post_id=%3$d&amp;origin=%4$s&amp;obj_id=%2$d-%3$d-%5$s%6$s&block=1', $version, $blog_id, $post_id, $domain, $uniqid, $new_layout );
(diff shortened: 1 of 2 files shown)


projects/plugins/jetpack/extensions/blocks/like/like.php
```diff
@@ -87,7 +87,7 @@ function render_block( $attr, $content, $block ) {
 		$bloginfo = get_blog_details( (int) $blog_id );
 		$domain   = $bloginfo->domain;
 		$version  = '20231201';
-		$src      = sprintf( '//widgets.wp.com/likes/index.html?ver=%1$d#blog_id=%2$d&amp;post_id=%3$d&amp;origin=%4$s&amp;obj_id=%2$d-%3$d-%5$s%6$s', $version, $blog_id, $post_id, $domain, $uniqid, $new_layout );
+		$src      = sprintf( '//widgets.wp.com/likes/index.html?ver=%1$d#blog_id=%2$d&amp;post_id=%3$d&amp;origin=%4$s&amp;obj_id=%2$d-%3$d-%5$s%6$s&block=1', $version, $blog_id, $post_id, $domain, $uniqid, $new_layout );
```
phcp: I think the `&` should be encoded like `&amp;` as the others in the same URL, and be placed before the `obj_id`. All the other parameters are reflected in the `obj_id`, so I think this one should follow the same approach and be passed as a boolean variable to the `sprintf`.

phcp (approved):
Nice work!

PR #34783, opened 2023-12-26
Fix like avatars being "cut" by the likes label in mobile view of P2

### LINKED ISSUE #85705: Like widget: fix styling on P2s (mobile)

JuanLucha:
Fixes https://github.com/Automattic/wp-calypso/issues/85705

## Proposed changes:
In P2, the like button has more buttons by its side, which means the widget has less space. This makes the "x likes" label "cut" the last avatar like this:
Screenshot 2023-12-28 at 19 24 59

This PR fixes that making the avatar not visible if there is no space for it to appear in full size. This PR also makes the number of likes, and the "likes" label always be in the same like (in the image, you can see it in two lines instead).

### Other information:

## Jetpack product discussion
Not needed.

No.

## Testing instructions:
- Apply this patch to your sandbox using `bin/jetpack-downloader test jetpack fix/likes-in-mobile-view`.
- Sandbox `widgets.wp.com`, and `s0.wp.com`.
- Go to a P2 post with more than 5 likes.
- Put the browser in mobile mode and check that the avatars are not cut by the "x likes" label.

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-likes-in-mobile-view
+++ b/projects/plugins/jetpack/changelog/fix-likes-in-mobile-view
@@ -1,0 +1,5 @@
+Significance: patch
+Type: other
+Comment: it's just a minor bug fix
+
+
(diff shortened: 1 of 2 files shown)


JuanLucha:
Fixing in WordPress code instead

PR #34787, opened 2023-12-27
Fix/jetpack footer inconsistent behavior

### LINKED ISSUE #31969: Jetpack Footers: inconsistent behavior
### Impacted plugin

Jetpack

### Quick summary

From #31627 @bindlegirl : 

The logged-in About link on the Jetpack dashboard (and other) pages shows up as an external link, although it's not. However, it loads the About page in a new tab, which differs from the Terms and Airline pages.

There are inconsistencies between the MyJetpack footer and other pages. This is what I see:

    Logged out About link on the Jetpack page will load https://jetpack.com/about/?site=nutritious-fairy.jurassic.ninja and MyJetpack https://jetpack.com/about
    Logged in About link on the Jetpack page will load /wp-admin/admin.php?page=jetpack_about and MyJetpack will load https://jetpack.com/about/

Privacy on connected MyJetpack is still an external link.

### Steps to reproduce

View the footers with the Jetpack admin and a My Jetpack page.

_No response_

### What actually happened

_No response_

### Impact

All

### Available workarounds?

Yes, easy to implement

_No response_

### Logs or notes

_No response_

darssen:
Fixes #31969

This is a fallback from #31627 where some inconsistencies between the MyJetpack footer and other pages appeared.
## Proposed changes:

The changes focus on
- Make **Privacy** and **About** links in the footer non-external for connected scenarios.
- Add consistent behavior in the Footer of My Jetpack to the footers in the rest of the Jetpack pages.

### Other information:

## Testing instructions:

- Make sure your site is not connected to Jetpack
- Visit the _Jetpack_ section
- In the footer, check the following links:
  - Jetpack: `https://jetpack.com/`
  - About: `https://jetpack.com/about/`
  - Privacy: `https://automattic.com/privacy/`
  - Terms: `https://wordpress.com/tos/`
  - An Automattic Airline: `https://automattic.com/about/`
- Visit the _My Jetpack_ section
  - In the footer, check the following links:
  - Jetpack: `https://jetpack.com/`
  - About: `https://jetpack.com/about/`
  - Privacy: `https://automattic.com/privacy/`
  - Terms: `https://wordpress.com/tos/`
  - An Automattic Airline: `https://automattic.com/about/`
- Connect your site to Jetpack
- In the footer, check the following links:
  - Jetpack: `https://jetpack.com/`
  - About: `/wp-admin/admin.php?page=jetpack_about`
  - Privacy: `/wp-admin/admin.php?page=jetpack#/privacy`
  - Terms: `https://wordpress.com/tos/`
  - An Automattic Airline: `/wp-admin/admin.php?page=jetpack_about`
- Visit the _My Jetpack_ section
- In the footer, check the following links:
  - Jetpack: `https://jetpack.com/`
  - About: `/wp-admin/admin.php?page=jetpack_about`
  - Privacy: `/wp-admin/admin.php?page=jetpack#/privacy`
  - Terms: `https://wordpress.com/tos/`
  - An Automattic Airline: `/wp-admin/admin.php?page=jetpack_about`
- Click on one of the Learn More buttons in one of the cards like VaultPress Backup
- In the footer, check the following links:
  - Jetpack: `https://jetpack.com/`
  - About: `/wp-admin/admin.php?page=jetpack_about`
  - Privacy: `/wp-admin/admin.php?page=jetpack#/privacy`
  - Terms: `https://wordpress.com/tos/`
  - An Automattic Airline: `/wp-admin/admin.php?page=jetpack_about`


darssen:
I don't like that the [getRedirectUrl](https://github.com/Automattic/jetpack/blob/trunk/projects/js-packages/components/tools/jp-redirect/index.ts#L23) function adds the `site` parameter for all the Jetpack pages aside from My Jetpack. This makes things like redirecting to `tos` being redirected to https://wordpress.com/tos/?site=YOUR_SITE

Not sure about a good way to avoid it.

darssen:
Also just realized that the initial change is only good for main My Jetpack page, but it is not working inside things like [page=my-jetpack#/add-backup](url). One possibility would be to add it to every place `AdminPage` is used. Another would be for `JetpackFooter` to don't need it as a prop, and use inside it maybe a global similar to the one I am using in `MyJetpackScreen` , the problem is that the `myJetpackInitialState` var does not belong outside MyJetpack, so still looking for a more elegant solution.

projects/packages/my-jetpack/_inc/components/my-jetpack-screen/index.jsx
```diff
@@ -120,7 +120,7 @@ export default function MyJetpackScreen() {
 	}
 
 	return (
-		<AdminPage>
+		<AdminPage siteAdminUrl={ window?.myJetpackInitialState?.adminUrl }>
```
andrii-lysenko, line 123: I'm wondering if we could pull off this information from some Redux store.

andrii-lysenko (commented):
Overall, it seems reasonable to me. I agree with your comment and maybe we should consider incapsulate all logic inside the footer for consistent behavior.
Maybe, we could reuse some redux store for the latest state.

darssen:
Thanks for the feedback @andrii-lysenko. 
After following some of the internal suggestions, I have updated the PR adding a tool inside` js-packages/components` to retrieve the `siteAdminUrl` and use it directly in the footer. It uses `window` var that gets populated from different Redux stores. Couldn't think of a more elegant solution.

projects/js-packages/components/tools/get-site-admin-url/index.ts
```diff
@@ -0,0 +1,26 @@
+declare global {
```
andrii-lysenko: Maybe it's a bit cleaner to keep the `declaration` in the separate file, say `types.d.ts`

andrii-lysenko (dismissed):
Thanks for working on this. This does seem reasonable to me. While testing on different envs haven't found any problems. Looks good 👍 
Left a small notice but it's not blocking.

projects/js-packages/components/tools/get-site-admin-url/index.ts
darssen: Thanks for the suggestion, I have updated it.

andrii-lysenko (approved):
Thanks for addressing that. LGTM!

PR #34792, opened 2023-12-27
Subscribe Block: Add source attribute.

lezama:
## Proposed changes:

* Adds a source attribute to the Subscribe Block so we can better understand its usage.

### Other information:

## Testing instructions:

* Pull this into your wpcom sandbox `bin/jetpack-downloader test jetpack update/subscribe-block-usage-tracking`
* Check that `wpcom_memberships_checkout_view` gets the `app_source` attribute assigned correctly

### DIFF
--- a/projects/plugins/jetpack/extensions/blocks/subscriptions/block.json
+++ b/projects/plugins/jetpack/extensions/blocks/subscriptions/block.json
@@ -108,6 +108,10 @@
 		"successMessage": {
 			"type": "string",
 			"default": "Success! An email was just sent to confirm your subscription. Please find the email now and click 'Confirm Follow' to start subscribing."
+		},
+		"source": {
+			"type": "string",
+			"default": "subscribe-block"
 		}
 	},
 	"example": {}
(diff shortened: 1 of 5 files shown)


projects/plugins/jetpack/extensions/blocks/subscriptions/block.json
```diff
@@ -108,6 +108,10 @@
 		"successMessage": {
 			"type": "string",
 			"default": "Success! An email was just sent to confirm your subscription. Please find the email now and click 'Confirm Follow' to start subscribing."
+		},
+		"source": {
```
simison, line 112: Thinking if it would make sense to prefix with `__source` to clarify it's sort of internal attribute?

edanzer:
The code changes in this PR look reasonable, but I've been testing this for a while and I'm not able to get the wpcom_memberships_checkout_view event to fire at all (that I can see), so I'm not able to check the app_source attribute. I think I'm missing something about how/when that event fires.

lupus2k:
> The code changes in this PR look reasonable, but I've been testing this for a while and I'm not able to get the wpcom_memberships_checkout_view event to fire at all (that I can see), so I'm not able to check the app_source attribute. I think I'm missing something about how/when that event fires.

I think you have to sandbox `subscribe.wordpress.com`

PR #34795, opened 2023-12-28
External media: provide default numerical post id for uploads 

### LINKED ISSUE #85373: Image uploads return Invalid parameter(s): post_id when trying to add images to Templates
### Quick summary

When trying to add an image from a 3rd-party service like Pexels or Openverse to a Template in the Site Editor, the image is not added and returns `Invalid parameter(s): post_id`. It appears to only affect Templates as adding to a Page or Post works fine. 

User report: 7457481-zd-a8c

### Steps to reproduce

1. With an FSE theme, open a template for editing such as the Index template
2. Add an Image block, choose Select Image, and select Pexels or Openverse
3. Select an image from Pexels or Openverse and click Select to try and insert it into the Template - see the error message returned for `Invalid parameter(s): post_id` and no image inserted. 

Images from 3rd-party services should be added to Templates just like Posts and Pages. 

### What actually happened

`Invalid parameter(s): post_id` is returned and no image is added to the Template. 

### Impact

Some (< 50%)

### Available workarounds?

Atomic

### Logs or notes

Reproduced on an AT test site only. Unknown if Simple sites are affected.

ramonjd:
Fixes https://github.com/Automattic/wp-calypso/issues/85373

## Proposed changes:
This is a test to ensure that external media uploads get a numerical post id and a fallback of 0.

It checks first for a numerical `id`, then for `wp_id`, which template and template part post types use to store the numerical id.

### Other information:

Simple sites use a different endpoint to Atomic sites. See: https://github.com/Automattic/jetpack/blob/a77f281222d73d7a0db33bf865ee3dde0f98b0cc/projects/plugins/jetpack/extensions/shared/external-media/sources/api.js#L6-L6

The simple site endpoint has a built-in type coercion and fallback, e.g., `(int) $data['post_id'] ?? 0;`

The Atomic endpoint `/wpcom/v2/external-media/copy/` however, expects `post_id` [to be a number.](https://github.com/Automattic/jetpack/blob/f43a439ea2a1fa3e24311cef6a197f73bbd75c04/projects/plugins/jetpack/_inc/lib/core-api/wpcom-endpoints/class-wpcom-rest-api-v2-endpoint-external-media.php#L145).

If Jetpack sends a non-number value to `/wpcom/v2/external-media/copy/` it will fail with the following API response:

```json
{
  "code": "rest_invalid_param",
  "message": "Invalid parameter(s): post_id",
  "data": {
    "status": 400,
    "params": {
      "post_id": "post_id is not of type number."
    },
    "details": {
      "post_id": {
        "code": "rest_invalid_type",
        "message": "post_id is not of type number.",
        "data": {
          "param": "post_id"
        }
      }
    }
  }
}
```

Templates in Gutenberg have ids like `theme//template-name`, so they will cause this failure.

### Before

https://github.com/Automattic/jetpack/assets/6458278/cb4690dd-6b71-46aa-9329-fce6cfc1fa01

### After

https://github.com/Automattic/jetpack/assets/6458278/015ccbd3-e49f-4a41-891d-93b77a7fabf0

## Jetpack product discussion
Related discussion: p1703590834062169-slack-C02FMH4G8

No.

## Testing instructions:

1. Open the site editor and edit a template.
2. Insert an image block
3. Select an external media source, such as Pexels or Openverse
4. Insert an image

You should not receive a 400 error in the API response, e.g., from `/wp-json/wpcom/v2/external-media/copy/openverse`


jeherve (approved):

ramonjd:
Thank you @jeherve!!

PR #34799, opened 2023-12-28
Launchpad: Use Site Title task in the Site Setup Launchpad

valterlorran:
Related to paYKcK-3Hy-p2

## Proposed changes:

* We should reuse the Site Title task instead of duplicating the task with the `blogname_set` task. Since it would have exactly the same end, we should avoid duplicating it.
* I thought about creating a new task so it could facilitate the migration, but since we are adding the Launchpad only for new sites, it is not necessary to keep the same name.

### Other information:

## Testing instructions:

* Sandbox the public-api
* Add the following filter to your `0-sandbox.php` file
```
add_filter( 'enable_legacy_site_setup_launchpad', '__return_true' );
```
* Apply this patch to your sandbox
* Create a new site through the `/setup/free` flow, and leave the site name empty or fill it with `Site Title`.
* Go through the flow until you reach the Customer Home. Then, click on the `Show site setup` button in the main banner.
* The Site setup launchpad should be displayed
* Click on the `Give your site a name` task and add a new title
* Go back to the Customer Home, the task should be marked as complete

### DIFF
--- a/projects/packages/jetpack-mu-wpcom/changelog/update-use-site-title-task-site-setup-launchpad
+++ b/projects/packages/jetpack-mu-wpcom/changelog/update-use-site-title-task-site-setup-launchpad
@@ -1,0 +1,4 @@
+Significance: patch
+Type: changed
+
+Uses the Site Title task instead of the blogname_set task.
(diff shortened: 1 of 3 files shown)


markbiek (approved):
LGTM!

PR #34807, opened 2023-12-29
wp-admin Site Default: Fix/enable the menu toggle when in wp-admin

jeroenpf:
Fixes: https://github.com/Automattic/dotcom-forge/issues/4742

The sidebar can not be toggled by clicking on the My Sites button on a mobile view (<=782 pixels). In Calypso, clicking on the My Sites button while on a mobile display will trigger the sidebar but due to the masterbar adjustments, the toggle is not working in wp-admin with the masterbar module active. This is because the original WordPress toggle is replaced with the My Sites button, but no toggle functionality exists for it.

## Proposed changes:

This PR turns the 'My Sites' button in a sidebar toggle while on a mobile display.

* Clicking on the 'My Sites' admin bar button while on a display that is <= 782 pixels toggles the sidebar.
* Clicking outside of the sidebar while it is opened and on a display that is <= 782 pixels will hide the sidebar.
* Disables the default link behavior while on a smaller viewport size so it becomes a toggle.
* Removes conflicting event handlers from the original toggle element.
* Prevents the click event from propagating when on a mobile viewport size so it does not trigger the original behavior of `.ab-item` elements on the My Sites button. 
* * Remove `aria-haspopup` attribute from the 'My Sites' link to prevent it from triggering the wp-responsive-overlay on small screens (<=480 pixels). The responsive overlay would be blocking clicks inside of the expanded menu and should not be triggered. 

### Other information:

## Testing instructions:

TBD
1. First let's verify the bug. Do this by going to any wp-admin page for a WordPress simple site and Atomic site.
2. Try to click on the My Sites button in the top bar navigation. You should be taken to the Calypso dashboard.
3. Now go back to wp-admin and resize the screen so that the menu disappears.
4. Click on the My Sites menu item and confirm that it still links to the Calypso dashboard.
5. Now it's time to apply the PR. You can follow the instructions in the comment below: https://github.com/Automattic/jetpack/pull/34807#issuecomment-1871926829 - **ALSO ensure that if you test with a simple site, that the url of your simple site is sandboxed (yoursite.wordpress.com should point to your sandbox)**
6. Once the PR has been applied to your sandbox and/or WoA dev site, navigate to wp-admin. If you applied the PR to your sandbox, you should go to wp-admin for a simple site. If you applied it to a WoA dev site, you should go to an Atomic site's wp-admin.
7. With a full screen, ensure that the sidebar is visible and try clicking on the 'My Sites' link again. Ensure that you still navigate to the Calypso dashboard.
8. Go back to wp-admin and resize the viewport until the sidebar is fully hidden. Now again click on the 'My Sites' button and confirm that the sidebar appears. Click on it again and confirm that the sidebar closes.
9. Finally, open the sidebar again by clicking on 'My Sites' and this time ensure that your viewport is scaled to the point that there is space on the right side of the sidebar. (there is a breakpoint where the expanded sidebar does not cover the entire screen). Click outside of the sidebar and confirm that the sidebar closes.

Ideally, these tests are done for both Atomic and WordPress simple sites. Also, check on your mobile phone and/or tablet. Testing in very small viewports is needed as there were some conflicts with the existing behaviour of the menu in relation to the `wp-responsive-overlay` being triggered on small screens. 

Viewports to test:
- `> 960` - Should always show the full menu with full text buttons. 
- `> 782 && <= 960` - Should always show the menu but only with icons.
- `> 480 && <= 782` - Should collapse the menu and it should be possible to toggle it via 'My Sites' button in top menu bar.
- `<= 480` - Should collapse the menu and it should be possible to toggle the menu via the WordPress logo in the top menu bar.

Please test the full menu experience to ensure there are no quirks and it works as you would expect.


katinthehatsite:
Here are my testing notes for Atomic sites:

* on the screens bigger than `> 960` - the sidebar is visible and `My Sites` redirects to Calypso dashboard:

https://github.com/Automattic/jetpack/assets/25575134/08d15e2f-080b-4f71-924a-d87ab5b2b1cb

* on the screens `> 782 && <= 960` - shows the menu only with icons  in `wp-admin` and `My Sites` redirects to Calypso. **Perhaps one thing that I found strange is that once you click `My Sites` and get to Calypso interface, the menu is full and not just icons which looks a bit inconsistent.** See the video below:

https://github.com/Automattic/jetpack/assets/25575134/152bb785-816a-4cfb-8d07-01cb1f6ea3e8

* on the screens `> 480 && <= 782` - the menu can be toggled on and off with `My Sites` and there is no redirect to Calypso:

https://github.com/Automattic/jetpack/assets/25575134/71df1a94-ff5c-490a-b6c7-080d29376a74

However, when I switch the interface to `wp-admin` under `Hosting Configuration > Admin interface style`, this immediately becomes buddy with the hamburger menu:

https://github.com/Automattic/jetpack/assets/25575134/8cbf7f63-15b2-4efc-8943-77b247b00c78

* for the screens `<= 480`, I am seeing the same issue as above with the hamburger menu when the admin interface is set to `wp-admin`:

https://github.com/Automattic/jetpack/assets/25575134/06041c5f-d8cc-4645-9a0e-ca25a65718e1

This is how it currently works on Atomic sites and the screens of  `<= 480`:

https://github.com/Automattic/jetpack/assets/25575134/53463935-8d0c-4811-95c9-4db39c830e4c

The changes work as expected though for when the interface style is set to Calypso:

https://github.com/Automattic/jetpack/assets/25575134/d14d8c24-0434-49c3-b957-74a8bf83f0c0

I have not tested these changes on Simples sites yet, I will do that next.

katinthehatsite:
For the Simple sites, these are the testing steps that I did but I am not seeing the changes to the menu. Am I testing this correctly? 

* Run `bin/jetpack-downloader test jetpack fix/enable-the-menu-toggle` on my sandbox
* Sandbox public api
* Force proxy reconnection (just in case)
* Navigate on a Simple site to `yourSimpleSite/wp-admin`
* Access a smaller screen such as `> 480 && <= 782`. I am expecting to be able to toggle the menu on and off now. This is what I am seeing at the moment:

https://github.com/Automattic/jetpack/assets/25575134/a43b49cb-7d45-422c-8d97-ee7a08d0c0ec

What am I missing on my end in the testing steps?

jeroenpf:
> What am I missing on my end in the testing steps?

I overlooked the fact that you need to point your simple site's domain to your sandbox as well, as it is trying to load javascript files that have changed from that domain. I added it to the instructions.

The problems you found on Atomic sites should now be fixed as well.

katinthehatsite:
Thanks for adjusting the testing instructions. I was able to see the changes now. These are the results of my testing on Simple sites: 

* on screens `<= 480`, I can see that the WordPress icon allows to toggle the sidebar menu on and off:

https://github.com/Automattic/jetpack/assets/25575134/e87f125f-6c51-4860-83c2-2ac5517afd37

* on screens `> 480 && <= 782`, I can see that `My Sites` works as a toggle for the sidebar menu:

https://github.com/Automattic/jetpack/assets/25575134/ff8c7220-7c43-4e39-9a4e-9b4efaa34d46

* on screens `> 782 && <= 960`, the menu is always visible but with buttons. My Sites redirects to Calypso:

https://github.com/Automattic/jetpack/assets/25575134/eab68c56-efa0-4145-9a1f-519638c58c08

* on screens `> 960`, there is a full menu. My Sites link redirects to Calypso:

https://github.com/Automattic/jetpack/assets/25575134/2b1ba34d-4268-4ab2-a931-278743406fa7

The changes on Simple sites seem to work as described 👍

projects/plugins/jetpack/modules/masterbar/masterbar/class-masterbar.php
```diff
@@ -266,7 +266,14 @@ public function maybe_logout_user_from_wpcom( $redirect_to ) {
...
 	public function admin_body_class( $admin_body_classes ) {
-		return "$admin_body_classes jetpack-masterbar";
+
+		$classes = array( 'jetpack-masterbar' );
+
+		if ( get_option( 'wpcom_admin_interface' ) === 'wp-admin' ) {
+			$classes[] = 'wpcom-admin-interface';
+		}
+
+		return implode( ' ', $classes ) . $admin_body_classes;
```
zaguiini: I believe we need a space here:

```suggestion
		return implode( ' ', $classes ) . ' ' . $admin_body_classes;
```

Otherwise the `$admin_body_classes` variable will be concatenated together and produce an incorrect class name.

Another alternative would be to reduce this line to `implode(' ', $classes)` and add `$admin_body_classes` to the `$classes` array so that it's imploded separated by the space.

It'll be even better if we trim the input to make sure we're not adding more or less spaces.

projects/plugins/jetpack/modules/masterbar/admin-menu/admin-menu.js
```diff
@@ -1,6 +1,6 @@
 /* global ajaxurl, jetpackAdminMenu */
 
-( function () {
+( function ( $ ) {
```
zaguiini, line 3: Do we need jQuery here?

We're using `.off` and `.width`, two methods that are easily replicable with vanilla JS:

- `.off`: https://youmightnotneedjquery.com/#off
- `.width`: https://youmightnotneedjquery.com/#get_width

I suppose we're already injecting jQuery somewhere on page load, so if that's true, please disregard. But if we're not, then we can save some bytes.

jeroenpf: Thanks for the feedback. I did not really want to use Jquery but its already part of Core. The reason i need it here is because there are some custom event listeners added with `.on` One such example is the `click.wp-responsive` event (in core). To my knowledge, jquery has its own events registry which can't be manipulated through Vanilla JS.

projects/plugins/jetpack/modules/masterbar/masterbar/class-masterbar.php
jeroenpf: Thanks, thats a good point. When i tested it, the class names were separated by spaces and it worked as expected but i've changed the code just to be sure and make it more explicit.

wojtekn (approved):
It works as expected.

PR #34810, opened 2023-12-29
Notification Accessibility: Turn notification icon into a link

phcp:
Related to https://github.com/Automattic/wp-calypso/issues/85710

## Proposed changes:

* This patch turns the notification icon into a link; with that, the icon is clickable from the screen reader's perspective.

### Other information:

## Testing instructions:

- Apply this PR to your local
- Spin up a JT site
- Open the site on a logged-in window
- Inspect the notification icon, the href point to `https://wordpress.com/notifications`
- Enabled the VoiceOver
- Navigate to the notification icon and click it (ctrl + opt + space)
- It should open the notification area
- Check on different browsers, the behavior should be similar
- Switch to [WordPress.com Toolbar](https://jetpack.com/support/masterbar/) and perform the same tests with that

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-notifications-accessibility
+++ b/projects/plugins/jetpack/changelog/fix-notifications-accessibility
@@ -1,0 +1,4 @@
+Significance: patch
+Type: bugfix
+
+Notification Accessibility: Turn notification icon into a link
--- a/projects/plugins/jetpack/modules/masterbar/masterbar/class-masterbar.php
+++ b/projects/plugins/jetpack/modules/masterbar/masterbar/class-masterbar.php
@@ -611,6 +611,7 @@ class Masterbar {
 					'class' => 'menupop mb-trackable',
 				),
 				'parent' => 'top-secondary',
+				'href'   => 'https://wordpress.com/notifications',
 			)
 		);
 	}
--- a/projects/plugins/jetpack/modules/notes.php
+++ b/projects/plugins/jetpack/modules/notes.php
@@ -199,6 +199,7 @@ class Jetpack_Notifications {
 					'class' => 'menupop',
 				),
 				'parent' => 'top-secondary',
+				'href'   => 'https://wordpress.com/notifications',
 			)
 		);
 	}
--- a/projects/plugins/jetpack/modules/notes.php
+++ b/projects/plugins/jetpack/modules/notes.php
@@ -192,6 +192,7 @@ class Jetpack_Notifications {
 			array(
 				'id'     => 'notes',
 				'title'  => '<span id="wpnt-notes-unread-count" class="' . esc_attr( $classes ) . '">
+					<span class="screen-reader-text">' . esc_html__( 'Notifications', 'jetpack' ) . '</span>
 					<span class="noticon noticon-notification"></span>
 					</span>',
 				'meta'   => array(


ivan-ottinger:
I am reviewing this PR now.

ivan-ottinger (dismissed):
Great job! The links work as expected using VoiceOver. 🚀 

I am approving the PR and leaving one suggestion (in case you would like to add it as a part of this PR):

Currently, the notification icon defined in `projects/plugins/jetpack/modules/notes.php` is missing screen-reader text. As a result, when we tab to that link, VoiceOver will read it as a link only (without any description). The user then doesn't know what link they are on and need to click it to see what will happen.

The following change should do the trick (replicating what we already have for the WordPress.com Admin bar):

Markup on 2024-01-03 at 16:13:52

(basically adding the `<span class="screen-reader-text">' . esc_html__( 'Notifications', 'jetpack' ) . '</span>` line)

ivan-ottinger (approved):
Looks good and works as expected. VoiceOver now reads the new description when the icon is highlighted on WordPress.com Admin bar. 🙂

jsnajdr:
On my WP.com sites I now see that clicking on the bell icon doesn't open the Notifications popup, but does a full page navigation to wordpress.com/read/notifications. Which is not a very good looking page. Seems to be caused by this patch. There is probably a missing `event.preventDefault()` somewhere in a `click` handler that opens the popup. Now it has an additional job to do, to prevent the default navigation to the `href` link.

@ivan-ottinger @phcp can you please verify if this is indeed the case?

ivan-ottinger:
Hey @jsnajdr 👋🏼 Thank you for the report! If I am understanding correctly, this does happen on a Simple site, right?

I am having trouble reproducing the issue in my tests. 🤔 The Notifications popup seems to be opening correctly wherever I try to click the bell icon:

Site front-end:
Markup on 2024-01-08 at 10:34:58

WP Admin:
Markup on 2024-01-08 at 10:34:11

Calypso:
Markup on 2024-01-08 at 10:35:50

Could you maybe share with us where exactly you are at when clicking on the bell icon? Thank you.

jsnajdr:
It's happening to me on a frontend of a Simple site, namely on internal P2s (updateomattic etc.) And, what's probably crucial, it happens only on Firefox. I now tried to test on Chrome (which is not my regular browser) and there the popup works.

ivan-ottinger:
Hmm. That's odd. I cannot reproduce it in Firefox either 🤔 (tested with standard sites and P2s as well):

Markup on 2024-01-08 at 10:56:54

My Firefox is updated to 121.0 which seems to be the latest version and no extensions installed. I wonder if there could be some extension that could be conflicting with the latest changes somehow (?)

ivan-ottinger:
Now we have another report of the same thing with recording here: p1704707862903429-slack-C02TCEHP3HA. 🤔

jsnajdr:
I tracked it down to the `initCheck3PC` function that detects whether third-party cookies are enabled or not. On Firefox, I see a "3PC blocked" message in console and the iframe popup is never enabled. On Chrome, I see "3PC enabled" and the popup works.

I'll try to figure out what has changed in the last few days because this definitely used to work for me.

ivan-ottinger:
Thanks! We are also looking at it in that thread above (https://github.com/Automattic/jetpack/pull/34810#issuecomment-1880690913).

TimBroddin:
This probably quietly failed before, and now because of the introduction of the href attribute, falls back to navigating to wordpress.com/notifications .

We might want to consider reverting this to how it was before, until we have a better solution.

ivan-ottinger:
I can create a quick PR that will revert the related change.

yansern:
Well, there is `event.preventDefault()`, it is likely just executing on the incorrect element as the click event listener is hooked to a different element (at least that's what I see). Maybe just need to add one more `event.preventDefault()` targeting the right element.

I'm looking at `initRedirect()` here. fbhepr%2Skers%2Sjcpbz%2Sjc%2Qpbagrag%2Szh%2Qcyhtvaf%2Sabgrf%2Snqzva%2Qone%2Qi2.wf%3Se%3Q687o6r14%23601%2Q605-og

jsnajdr:
The `href` attribute and the `click` handler are probably fine after all. Even the element is right. It's the 3PC detection that has changed and that is now disabling the iframe popup completely.

PR #34818, opened 2024-01-01
Jetpack Scan Client: Add Package

nateweller:
## Proposed changes:

* Introduces the "scan" JS package and general project structure, based on https://github.com/Automattic/jetpack/pull/34530.
* No functionality is included in this PR, just the project structure. Will follow up with additional PRs to introduce types, strings, library functions, and React hooks from the PoC.

### Other information:

## Jetpack product discussion

peb6dq-1Tk-p2

No

## Testing instructions:

None


projects/js-packages/scan/package.json
```diff
@@ -0,0 +1,68 @@
...
+	"type": "module",
+	"devDependencies": {
+		"@automattic/jetpack-webpack-config": "workspace:*",
+		"@babel/core": "7.23.5",
+		"@babel/plugin-transform-react-jsx": "7.23.4",
+		"@jest/globals": "29.7.0",
+		"@storybook/addon-actions": "7.4.6",
+		"@storybook/blocks": "7.4.6",
+		"@storybook/react": "7.4.6",
+		"@testing-library/react": "13.4.0",
```
anomiex: The rest of the monorepo is already on 14.1.2
```suggestion
		"@testing-library/react": "14.1.2",
```

projects/js-packages/scan/package.json
```diff
@@ -0,0 +1,68 @@
...
+		"@types/jest": "29.5.10",
+		"@types/react": "18.2.33",
+		"@wordpress/babel-plugin-import-jsx-pragma": "4.30.0",
```
anomiex:
```suggestion
		"@wordpress/babel-plugin-import-jsx-pragma": "4.31.0",
```

projects/js-packages/scan/package.json
```diff
@@ -0,0 +1,68 @@
...
+	"types": "./build/index.d.ts",
+	"dependencies": {
+		"@automattic/jetpack-api": "workspace:*",
+		"@automattic/jetpack-base-styles": "workspace:*",
+		"@automattic/jetpack-connection": "workspace:*",
+		"@automattic/jetpack-shared-extension-utils": "workspace:*",
+		"@wordpress/api-fetch": "6.44.0",
+		"@wordpress/element": "5.24.0",
+		"@wordpress/i18n": "4.47.0",
+		"@wordpress/url": "3.48.0",
```
anomiex:
```suggestion
		"@wordpress/api-fetch": "6.45.0",
		"@wordpress/element": "5.25.0",
		"@wordpress/i18n": "4.48.0",
		"@wordpress/url": "3.49.0",
```

projects/js-packages/scan/package.json
```diff
@@ -0,0 +1,68 @@
...
+		"debug": "4.3.4",
+		"react": "^18.2.0",
+		"react-dom": "^18.2.0"
+	},
+	"peerDependencies": {
+		"react": "^18.2.0",
+		"react-dom": "^18.2.0",
+		"@wordpress/i18n": "4.47.0"
```
anomiex, line 66: Why the peer dep on `@wordpress/i18n`?

anomiex (dismissed):
Could use some dependency version bumps, otherwise looks ok from a monorepo structural perspective.

If you know what you're doing with using tsc to build, help with p1704210580568709/1704137203.328069-slack-C054LN8RNVA would be appreciated.

projects/js-packages/scan/package.json
nateweller: Peers will need to use `@wordpress/i18n` to use translatable strings provided by the package, a bit more info here: https://github.com/Automattic/jetpack/pull/34530/files#diff-265b9ba8c9fd1dbc6ab4c5f95c07260e3e4424ac4a08a2eb6a7cf98d735f0a01R29

That is the only reason I have listed it as a peer dep, does it make sense for this use case?

nateweller:
@anomiex Accepted the proposed version bumps, and committed the updated lock file. If you're cool with the internationalization peer dep, then I think this should be ready to go :+1:

projects/js-packages/scan/package.json
anomiex: Hmm. Is the use case here regarding if someone wants to use the package outside of a WordPress context?

Within WordPress things should just work, [@wordpress/dependency-extraction-webpack-plugin](https://www.npmjs.com/package/@wordpress/dependency-extraction-webpack-plugin) should externalize the reference to `@wordpress/i18n` already and WordPress should load the relevant i18n itself.

anomiex (approved):
> If you're cool with the internationalization peer dep, then I think this should be ready to go 👍

I don't understand what's going on with that, so I won't block it. In the normal case of use inside WordPress I don't think it would be needed, but possibly something unusual is going on here.

We can always revisit later if we need to. 🙂

projects/js-packages/scan/.eslintrc.cjs
```diff
@@ -0,0 +1,11 @@
+module.exports = {
```
andrii-lysenko, line 1: Maybe also adding react dependency?
`extends: [ require.resolve( 'jetpack-js-tools/eslintrc/react' ) ],`

andrii-lysenko (approved):
This does look good to me. Thanks for going with `Typescript` from the beginning. 
There's a small comment but it's not blocking.
Also, e2e errors are now fixed in trunk, so should be cleared out after merge.

PR #34820, opened 2024-01-01
WAF: Use string for "needs update" option to prevent unnecessary database updates

### LINKED ISSUE #34765: Enhancement: WAF enclosure for NEEDS_UPDATE_OPTION_NAME update
### Impacted plugin

Jetpack

### What

I would like for the check_for_updates WAF initialiser to properly enclose the update_option

### How

Looking at this line https://github.com/Automattic/jetpack/blob/d1b0594007fef579f424b1acf79765243c754d10/projects/packages/waf/src/class-waf-initializer.php#L202 , that update_option is now being called on each `check_for_updates` function call within each admin_init hook request. 

This translated to an additional UPDATE MySQL query on each WordPress backend request, which is not ideal.

Screenshot 2023-12-22 at 09 57 00

Would it make sense to move that update_option within the IF clause the the top so it only updates itself when the value is different?

```
if ( get_option( self::NEEDS_UPDATE_OPTION_NAME ) ) {
  ...
  update_option( self::NEEDS_UPDATE_OPTION_NAME, 0 );
}
```
CC @nateweller as this is related to this original PR https://github.com/Automattic/jetpack/pull/28899 where that was introduced.

Happy to submit a PR if needed. Thanks

nateweller:
Fixes https://github.com/Automattic/jetpack/issues/34765

## Proposed changes:

* Use a string instead of an integer to save the "jetpack_waf_needs_update" option. This allows the core `update_option` function to validate the new value against the saved one, and abort updates when they are identical. 
* The important change is the update that sets the value to `"0"`, but all other cases have been updated for consistency.

### Other information:

## Jetpack product discussion

https://github.com/Automattic/jetpack/issues/34765

No

## Testing instructions:

* Enable the WAF on a site with Jetpack Protect.
* Use Query Monitor on any admin page and validate the issue: an "UPDATE" is made to "jetpack_waf_needs_update" on every admin page load.
* Apply this patch with the Jetpack Beta plugin.
* Validate the unnecessary update is no longer occurring.

### DIFF
--- a/projects/packages/waf/changelog/fix-waf-update-option
+++ b/projects/packages/waf/changelog/fix-waf-update-option
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Optimize how the web application firewall checks for updates on admin screens.
--- a/projects/packages/waf/src/class-compatibility.php
+++ b/projects/packages/waf/src/class-compatibility.php
@@ -98,7 +98,7 @@ class Waf_Compatibility {
 		}
 
 		// If the option hasn't been added yet, the WAF needs to be updated.
-		return true;
+		return '1';
 	}
 
 	/**
@@ -186,7 +186,7 @@ class Waf_Compatibility {
 		$brute_force_allow_list = Jetpack_Options::get_raw_option( 'jetpack_protect_whitelist', false );
 		if ( false !== $brute_force_allow_list ) {
 			$waf_allow_list = self::merge_ip_allow_lists( $waf_allow_list, $brute_force_allow_list );
-			update_option( Waf_Initializer::NEEDS_UPDATE_OPTION_NAME, 1 );
+			update_option( Waf_Initializer::NEEDS_UPDATE_OPTION_NAME, '1' );
 		}
 
 		return $waf_allow_list;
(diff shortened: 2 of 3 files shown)


ArSn:
@nateweller So if I understand correctly, the "fix" here is to make the option a string rather an integer, because what we load from the database is a string instead of an integer, is that correct? And if so, and given that this is a boolean flag, have we considered using a bool instead?

nateweller:
> @nateweller So if I understand correctly, the "fix" here is to make the option a string rather an integer, because what we load from the database is a string instead of an integer, is that correct? And if so, and given that this is a boolean flag, have we considered using a bool instead?

@ArSn I've applied what I think is a better approach, using booleans in the `get_option` and `update_option` calls.

These are saved in the database as `"1"` and `""`, and `get_option` returns these string values. Initially, I wanted to rely on the built-in duplicate value check inside `update_option` to solve the root issue of the `UPDATE` running every admin page load. Providing a boolean value to `update_option` negates the dupe check, since it ends up checking `true === '1'` and `false === ''`.

However, I could have looked a couple lines up in the `check_for_updates` method, and realized we already have a `get_option` check that we can use. So, I just bumped up the option update into that block.

miguelxpn (approved):

PR #34821, opened 2024-01-01
Add a check to My Jetpack to determine if a user is "new"

jboland88:
## Proposed changes:

* This diff adds a check in the My Jetpack initialization to determine if a user is "new" to Jetpack
* This data point will be used to vary some of the output of My Jetpack to better suit new versus established users

No

## Testing instructions:

* Apply this patch to a test site using the Jetpack Beta plugin (the local config has several Jetpack plugins installed already and will not show a user as "new")
* You can check whether the user is new by checking `window.myJetpackInitialState.userIsNewToJetpack` in the browser console, it will be "1" if we consider the user as "new"
* You should be considered a new user until you do one of the following: connect Jetpack, activate a Jetpack module that is not on by default, install another Jetpack plugin, make a Jetpack purchase.
* Test with different combinations of these actions to make sure the property is getting set correctly


projects/packages/my-jetpack/src/class-initializer.php
```diff
@@ -32,13 +33,23 @@ class Initializer {
...
 	const IDC_CONTAINER_ID = 'my-jetpack-identity-crisis-container';
 
+	const JETPACK_PLUGIN_SLUGS = array(
+		'jetpack-backup',
+		'jetpack-boost',
+		'zero-bs-crm',
+		'jetpack',
+		'jetpack-protect',
+		'jetpack-social',
+		'jetpack-videopress',
```
CodeyGuyDylan, line 50: I think we're missing `jetpack-search` here

CodeyGuyDylan (dismissed):
Tests well and code looks good, I think we're just missing one of the standalone plugin slugs

projects/packages/my-jetpack/src/class-initializer.php
jboland88: aha, good catch, thank you!!

projects/packages/my-jetpack/src/class-initializer.php
```diff
@@ -211,6 +224,64 @@ public static function enqueue_scripts() {
...
+			return false;
+		}
+
+		// check for other Jetpack plugins that are installed on the site (active or not)
+		// If there's more than one Jetpack plugin active, this user is not "new"
+		$plugin_slugs              = array_keys( Plugins_Installer::get_plugins() );
+		$plugin_slugs              = array_map(
+			static function ( $slug ) {
+				$parts = explode( '/', $slug );
+				return str_replace( '.php', '', $parts[ is_countable( $parts ) ? count( $parts ) - 1 : 0 ] );
```
robertsreberski: `explode()` will always return an array, that is countable 🤔 If `count( $parts ) is 1`, then `count( $parts ) - 1 === 0`. If `count( $parts ) is 0`, accessing element `$parts[0]` will also throw an error.

projects/packages/my-jetpack/src/class-wpcom-products.php
```diff
@@ -260,6 +265,9 @@ public static function get_site_current_purchases() {
 
 		$body      = wp_remote_retrieve_body( $response );
 		$purchases = json_decode( $body );
+		// Set a transient for 30 seconds
+		set_transient( self::MY_JETPACK_PURCHASES_TRANSIENT_KEY, $purchases, 30 );
```
robertsreberski: This looks good, just curious, how have you came up with value of `30 seconds`?

Also, the comment wouldn't be necessary if you would use `MINUTE_IN_SECONDS / 2`.

robertsreberski (dismissed):
LGTM for me, I just leave two non-blocking questions and suggestions regarding code.

projects/packages/my-jetpack/src/class-initializer.php
jboland88: > explode() will always return an array, that is countable 🤔

Good note, I've become so used to adding `is_countable` to avoid the PHP 8.1 warning 😅 

>  If count( $parts ) is 0, accessing element $parts[0] will also throw an error

Good thought, I can add an empty check.

projects/packages/my-jetpack/src/class-wpcom-products.php
jboland88: >  just curious, how have you came up with value of 30 seconds?

It's somewhat arbitrary, I wanted something long enough to stick around for the My Jetpack page to load and to handle async requests that needed to fetch products, but not much longer than that so that it would not get stale

robertsreberski (dismissed):
LGTM, no comments after the changes

robertsreberski (approved):
Still LGTM

PR #34825, opened 2024-01-02
Boost: Add sync

haqadn:
Related patch: D133459-code, D133458-code

## Proposed changes:
* Add jetpack sync to boost. This will allow us to determine which modules are active and what errors occurred and empower us to provide better support when things go wrong.

### Other information:

It enables Jetpack Sync with some callables to sync modules states(which modules are enabled), critical css generation status and last speed scores.

## Testing instructions:
* Ensure that options are synced with cache site

### DIFF
--- a/projects/packages/boost-speed-score/src/class-speed-score-history.php
+++ b/projects/packages/boost-speed-score/src/class-speed-score-history.php
@@ -52,7 +52,7 @@ class Speed_Score_History {
 	 *
 	 * @return string
 	 */
-	public function get_option_name() {
+	private function get_option_name() {
 		return static::OPTION_PREFIX . Speed_Score_Request::generate_cache_id_from_url( $this->url );
 	}
 
--- a/projects/packages/boost-speed-score/src/class-speed-score-history.php
+++ b/projects/packages/boost-speed-score/src/class-speed-score-history.php
@@ -52,7 +52,7 @@ class Speed_Score_History {
 	 *
 	 * @return string
 	 */
-	private function get_option_name() {
+	public function get_option_name() {
 		return static::OPTION_PREFIX . Speed_Score_Request::generate_cache_id_from_url( $this->url );
 	}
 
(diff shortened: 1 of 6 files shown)


projects/plugins/boost/app/data-sync/Modules_State_Entry.php
```diff
@@ -59,7 +59,7 @@ public function merge( $value, $partial_value ) {
 		return array_merge( $value, $partial_value );
 	}
 
-	private function get_module_option_name( $module_slug ) {
+	public function get_module_option_name( $module_slug ) {
```
dilirity: Why is this necessary?

projects/packages/boost-speed-score/src/class-speed-score-history.php
```diff
@@ -52,7 +52,7 @@ public function __construct( $url ) {
 	 *
 	 * @return string
 	 */
-	private function get_option_name() {
+	public function get_option_name() {
```
dilirity: Why is this necessary? I don't see it used in the PR.

dilirity (changes requested):
Other than two small things, this is looking great! I had no idea adding Sync would look so simple.

projects/plugins/boost/app/data-sync/Modules_State_Entry.php
haqadn: Good catch as always!

dilirity (approved):
Great!

I noticed you removed syncing modules status and that still works.

Any idea when a sync happens after you toggle an option on or off?
I've been trying to get the modules status in the debugger to change (and get the CSS state to update) after making changes on my JN, but it doesn't show up.

image

image

haqadn:
> Any idea when a sync happens after you toggle an option on or off? I've been trying to get the modules status in the debugger to change (and get the CSS state to update) after making changes on my JN, but it doesn't show up.

> The Sync Listener then listens for these action invocations (during any non-GET request, any request by a logged in user, and when running the Jetpack test suite), captures (serializes) the action and parameters, and adds it to the Sync Queue to be sent to the corresponding WordPress.com cache site by the Sync Sender.
- PCYsg-7Q6-p2

thingalon:
This has been merged? Have we pinged the right people to get our privacy info updated on jetpack.com?

PR #34827, opened 2024-01-02
WooCommerce analytics: Track store managers

### LINKED ISSUE #1256: Include admins in woocommerce_analytics tracking
Per the discussion here peapX7-30s-p2#comment-3985 we want to include admins in all woocommerceanalytics tracking.

So that we can differentiate between admins and shoppers, we'll need to add a new field something like admin=true or false.

jimjasson:
Fixes https://github.com/Automattic/wc-calypso-bridge/issues/1256

## Proposed changes:

In this PR: https://github.com/Automattic/jetpack/pull/32692, the `store_admin` property was added in WooCommerce Analytics. This property only checks for `administrators`, but we would like to track actions from `shop managers` as well.

### Other information:

## Testing instructions:

- [ ] Create a site -- it is faster to test this on a Woo Express Performance Plan site. 
- [ ] Download and install the Jetpack beta plugin. 
- [ ] Download and install the Tracks Vigilante Chrome Extension.
- [ ] In the Jetpack beta plugin, switch over to the `add/store-manager-property` branch.
- [ ] Create a shop manager user. 
- [ ] Log in as the shop manager or use a plugin such as User Switching. 
- [ ] Add a product to the cart. 
- [ ] Using the Tracks Vigilante extension, find the `woocommerceanalytics_add_to_cart` event. Confirm that `store_admin` is set to 1 like so: 

Markup on 2024-01-08 at 15:50:33


projects/plugins/jetpack/modules/woocommerce-analytics/classes/class-jetpack-woocommerce-analytics-trait.php
```diff
@@ -262,7 +262,7 @@ public function get_common_properties() {
 			'ui'                                 => $this->get_user_id(),
 			'url'                                => home_url(),
 			'woo_version'                        => WC()->version,
-			'store_admin'                        => in_array( 'administrator', wp_get_current_user()->roles, true ) ? 1 : 0,
+			'store_admin'                        => in_array( 'administrator', wp_get_current_user()->roles, true ) || in_array( 'shop_manager', wp_get_current_user()->roles, true ) ? 1 : 0,
```
PanosSynetos: Hey @jimjasson - What do you think of having one `in_array` instead of two?

```php
in_array( array( 'administrator', 'shop_manager' ), wp_get_current_user()->roles, true )
```

PanosSynetos (changes requested):
Thanks for wrangling this Jason - One relatively minor comment

PanosSynetos (dismissed):
Hey @jimjasson - Pre-approving this.

Manually tested and code reviewed :)

Before we merge this, could you update the instructions on this file? https://github.com/Automattic/jetpack/pull/32692/files#diff-fcaa0c0fdcd6ceb09a7ac11a67f98d9283ac152c97615a5c16e93bd275f3e093R8-R18

jeherve (approved):

PR #34831, opened 2024-01-02
Show the Reblog button according to the toggle in the sidebar

### LINKED ISSUE #85895: Like block: show the reblog button in the editor according to the value of the sidebar toggle.
### Details

Relates to https://github.com/Automattic/wp-calypso/issues/85273

In the Like block (still in beta), we need to show the reblog button (it was hidden in a previous PR) according to the value in the sidebar toggle:
Screenshot 2024-01-02 at 11 58 41

### Related

https://github.com/Automattic/wp-calypso/issues/85273

JuanLucha:
Fixes https://github.com/Automattic/wp-calypso/issues/85895

## Proposed changes:
The reblog button was hidden in a previous PR until we had the time to code the changes properly. Now it's time to show it again. It will show in the preview (yes, this PR only shows it in the editor view) according to the toggle value in the sidebar.

This PR removes some functions (and their files) as the endpoint should be called on publishing or updating the post, not on changing the settings in the sidebar.

### Other information:

## Jetpack product discussion
No.

No.

## Testing instructions:
- Apply this PR.
- This change only works for the simple sites, so we need to simulate that. To do that, go to `projects/plugins/jetpack/extensions/blocks/like/edit.js` and change line 14 from `const showReblogButton = isSimpleSite() && enabledReblogs;` to `const showReblogButton = enabledReblogs;`. Then in line 37, from `{ isSimpleSite() && (` to `{ (`.
- Enable the beta blocks by including `define( 'JETPACK_BLOCKS_VARIATION', 'beta' );` in `projects/plugins/jetpack/jetpack.php`.
- Setup your JT site (including building the plugin).
- Create a post and include a Like block. The Reblog buttons shouldn't be there.
- Enable the Reblog button in the sidebar option and check that the reblog button is now in the Like block preview.

Thanks for the review! 🙇


projects/plugins/jetpack/extensions/blocks/like/edit.js
```diff
@@ -1,60 +1,29 @@
...
-		reblogSetting: currentReblogSetting,
-		isLoading: fetchingReblog,
-	} = useFetchReblogSetting( blogId );
-	const {
-		setReblog,
-		success: reblogSetSuccessfully,
-		resetSuccess: clearReblogSetStatus,
-		isLoading: settingReblog,
-	} = useSetReblogSetting( blogId );
+	const enabledReblogs = attributes?.enabled_reblogs || false;
```
TimBroddin: Looking at other blocks, I can see that attribute names are always camelCase. I also feel like `reblogsEnabled` or even `displayReblogsButton` is more clear (see for example the related-posts block where every attribute to show/hide information starts with display).

JuanLucha: True. I'll change the case, and the name of the variable to "displayReglogsButton" 👍

TimBroddin (dismissed):
Tested on Simple in both site editor and post editor: 🟢
Tested on self-hosted, doesn't show: 🟢

I added one naming comment, but other than that: very solid PR 🎉

projects/plugins/jetpack/extensions/blocks/like/edit.js
```diff
@@ -10,12 +10,11 @@ import './editor.scss';
 
 function LikeEdit( { attributes, setAttributes } ) {
 	const blockProps = useBlockProps();
-	const enabledReblogs = attributes?.enabled_reblogs || false;
-	const showReblogButton = isSimpleSite() && enabledReblogs;
+	const showReblogButton = attributes?.showReblogButton || false;
```
TimBroddin, line 13: Shouldn't this include the `isSimpleSite` check? What's up with the `|| false?`

TimBroddin (approved):

PR #34840, opened 2024-01-03
Protect: Add WAF standalone mode indicator

dkmyta:
### Description
Adds an indicator within the `FirewallHeader` when standalone mode is enabled.

Related Jetpack PR: https://github.com/Automattic/jetpack/pull/34854

**Note**: I've used a common `Status` component with a `label={ __( Standalone mode, slug ) }` as the indicator across both plugins but I am open to suggestions as to other, more attractive or functional ways of doing so.

## Proposed changes:

* Add an initial state variable that retrieves the value of a constant set when standalone mode is activated
* Add a condition to the `FirewallHeader` to indicate when the mode is active
* Replace our custom status `Text` element with the js-packages `Status` component
* Fix the `currentStatus` logic so the FirewallHeader content is updated appropriately

### Other information:

## Jetpack product discussion

- https://github.com/Automattic/jetpack-scan-team/issues/204

- No

## Testing instructions:

* Checkout this branch
* Start your Jurassic Tube
* Enable Protect and verify that the WAF is `Active` from the Firewall tab
* Update the `auto_prepend_file` PHP directive to point to `wp-content/jetpack-waf/bootstrap.php` to activate standalone mode
* Return to the Firewall tab and confirm that the active status is replaced with `Standalone mode`
* Enable and disable various WAF settings, deactivate the module and view the Firewall section from an unsupported env to ensure that the conditional FirewallHeader content is updated as expected

## Screenshots
WAF active, Standalone mode disabled:
Screen Shot 2024-01-04 at 11 35 24

WAF active, Standalone mode enabled:
Screen Shot 2024-01-04 at 11 34 24

WAF inactive:
Screen Shot 2024-01-04 at 11 36 48


projects/plugins/protect/src/class-jetpack-protect.php
```diff
@@ -454,4 +455,13 @@ public static function get_waf_stats() {
...
 	}
+
+	/**
+	 * Get WAF standalone mode status
+	 *
+	 * @return bool|array True if WAF standalone mode is enabled, false otherwise.
+	 */
+	public static function get_waf_standalone_mode_status() {
+		return defined( 'JETPACK_WAF_RUN' ) && JETPACK_WAF_RUN === 'preload';
+	}
```
nateweller: This might fit nicely in the `waf` package, similar to how we are using `Waf_Runner::get_config()` and `Waf_Stats::get_automatic_rules_last_updated()` to map data from the firewall into the plugin's default state.

projects/plugins/protect/src/js/components/firewall-header/index.jsx
```diff
@@ -164,9 +172,14 @@ const FirewallHeader = ( {
...
-								{ __( 'Active', 'jetpack-protect' ) }
-							</Text>
+							<Status
+								status="active"
+								label={
+									standaloneMode
+										? __( 'Standalone mode', 'jetpack-protect' )
+										: __( 'Active', 'jetpack-protect', /* dummy arg to avoid bad minification */ 0 )
+								}
+							/>{ ' ' }
```
nateweller, line 183: Nice use of `Status` component to remove the need for custom styles here :+1:

This label is a perfect start, I think we can iterate on this to add more context or useful info (will circle back on this one).

nateweller: Perhaps "Standalone mode active", just to keep it clear?

nateweller (changes requested):
Looks and works great! Left some minor comments and the following non-blocking brain dump:

I know we discussed using the header for this, though it may be more relevant to include the mode status closer to information surrounding it:

Screenshot 2024-01-29 at 11 44 43 AM

Screenshot 2024-01-29 at 11 44 50 AM

We could consider just using an icon indicator in the header, such as a small green shield instead of the dot when enhanced protection is configured, with a hover popover that could open the full standalone info popover or direct the user otherwise to information about (1) the current status of standalone mode on their site, and (2) the install or uninstall instructions, depending on said status.

As standalone mode becomes more visible, we should also make sure our public documentation/support has adequate information as well. We will need to ping someone about that at some point or another.

projects/plugins/protect/src/class-jetpack-protect.php
dkmyta: Agreed! Made the change 👍🏻

projects/plugins/protect/src/js/components/firewall-header/index.jsx
dkmyta: >Perhaps "Standalone mode active"

I originally started with that but it felt a tad too long (I suppose especially since I am used to just seeing `Active`, here's an example:
Screen Shot 2024-01-30 at 14 27 07

I am open to the change though as it does avoid confusion - thoughts?

dkmyta:
>I know we discussed using the header for this, though it may be more relevant to include the mode status closer to information surrounding it

I thought about this as well. I do think it's important that we have an indicator upfront and visible in the header but leaving the Standalone mode footer section untouched may cause some confusion - we could also opt to just remove the related footer section if we detect it's enabled or as you said, replace the content of the section based on the status - uninstall instructions or a link to more documentation (once available).

nateweller (dismissed):
:shipit: (pending a rebase)

nateweller (approved):
:shipit: :shipit:

PR #34864, opened 2024-01-05
Add Hosting Provider guess

robertsreberski:
Fixes https://github.com/Automattic/jetpack-roadmap/issues/1039

## Proposed changes:
* Add hosting provider guess using nameservers (from wpcom)
* Implement the initial state value indicating whether user is guessed as coming from hosting provider or not

### Other information:

## Testing instructions:

* Run the site on JN / Ephemeral Site
* Activate Jetpack Beta plugin with this branch, and go to `/wp-admin/admin.php?page=my-jetpack`
* Check that `window.myJetpackInitialState.isUserFromHostPartner` exists
* (currently, there is no way to check whether hosting provider is guess properly)


projects/packages/my-jetpack/src/utils/nameservers.php
```diff
@@ -0,0 +1,108 @@
...
+ * Returns an array of nameservers for the current site.
+ *
+ * @return string
+ */
+function get_nameserver_dns_records() {
+	if ( ! isset( $_SERVER['SERVER_NAME'] ) ) {
+		return array();
+	}
+
+	$domain      = filter_var( wp_unslash( $_SERVER['SERVER_NAME'] ), FILTER_SANITIZE_STRING ); // Gets the domain name of the WordPress site
```
jboland88: github is giving a warning about `FILTER_SANITIZE_STRING` being deprecated in PHP 8.1. Would `santize_text_field` work here?

projects/packages/my-jetpack/src/class-initializer.php
```diff
@@ -280,6 +282,23 @@ static function ( $slug ) {
...
 	}
 
+	/**
+	 * Determines whether the user has come from a host partner we can recognize.
+	 *
+	 * @return string
+	 */
+	public static function is_user_from_host_partner() {
+		require_once __DIR__ . '/utils/nameservers.php';
+		$dns_records = get_nameserver_dns_records();
```
jboland88: Is this worth caching for some period? 
It would be nice not to look this up each time

projects/packages/my-jetpack/src/class-initializer.php
```diff
@@ -280,6 +282,23 @@ static function ( $slug ) {
 		return true;
 	}
 
+	/**
+	 * Determines whether the user has come from a host partner we can recognize.
+	 *
+	 * @return string
+	 */
+	public static function is_user_from_host_partner() {
```
jboland88, line 290: nitpick, would it make sense to call this `is_user_from_known_host`?

jboland88 (changes requested):
Just a few small comments, I think this is a clever solution for giving a basic guess at the host 👍

jeherve (commented):
What would you think about moving some of this to the Status package, where we already have a `Host` class?
https://github.com/Automattic/jetpack/blob/0817af0f0e9a21fdbd76745ba6cdcc7d0f43a678/projects/packages/status/src/class-host.php#L12-L15

robertsreberski:
> What would you think about moving some of this to the Status package, where we already have a Host class?

@jeherve I think it's a good idea, I've just implemented that, let me know what you think!

projects/packages/my-jetpack/src/class-initializer.php
robertsreberski: Implemented caching

projects/packages/my-jetpack/src/utils/nameservers.php
robertsreberski: Weirdly, locally I was getting errors based on older PHP version 🤔 I've updated phpcs and fixed the deprecated warning.

projects/packages/status/src/class-host.php
```diff
@@ -134,4 +135,144 @@ public function get_source_query() {
...
 		return '';
 	}
+
+	/**
+	 * Returns an array of nameservers for the current site.
+	 *
+	 * @param string $domain The domain of the site to check.
+	 * @return string
+	 */
+	private function get_nameserver_dns_records( $domain ) {
```
jeherve: thank you! Do you think you could add some tests for the new method, in `projects/packages/status/tests/php/test-host.php`?

jboland88 (approved):
This tested as described, the new tests are passing and the code LGTM 👍

PR #34865, opened 2024-01-05
Launchpad: Remove the "Set up your Professional Email" task

valterlorran:
More context: p1704463397882879-slack-C01TJT42PMM

## Proposed changes:

* In the discussion above, we relate the reasons for removing this task from the migration.

### Other information:

## Testing instructions:

* Visual inspection is enough in this case since the task was not being used and the key on the task list was not matching the task definition key.


daledupreez (dismissed):
Changes look good to me, but it looks like you need a rebase.

sixhours (approved):
LGTM!

PR #34867, opened 2024-01-05
CRM: 3405 - more robust local install detection

tbradsha:
Resolves Automattic/zero-bs-crm#3405 - more robust local install detection

## Proposed changes:

In WordPress Playground, `$_SERVER['REMOTE_ADDR']` is not set, so it was giving a PHP notice in the CRM. I switched the call to use the pre-existing `zeroBSCRM_getRealIpAddr()` (and added a fallback to that function just in case).

### Other information:

## Testing instructions:

In `trunk` + WP Playground, one gets a PHP notice on the System Status page (`/wp-admin/admin.php?page=zerobscrm-systemstatus&tab=status`).

In the `fix/crm/3405-use_more_robust_IP_check` branch + WP Playground the PHP notice is no more.

### DIFF
--- a/projects/plugins/crm/changelog/fix-crm-3405-use_more_robust_IP_check
+++ b/projects/plugins/crm/changelog/fix-crm-3405-use_more_robust_IP_check
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Backend: Add fallback for dev site detection.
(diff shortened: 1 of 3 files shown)


coder-karen:
I started looking at this but I'm not entirely sure how to test plugin branches with WordPress playground (not finding any information with a quick search anyway).

coder-karen (approved):
Works well.

PR #34869, opened 2024-01-05
CRM: 3404 - address syntax differences between SQLite and MySQL

tbradsha:
Resolves Automattic/zero-bs-crm#3404 - address syntax differences between SQLite and MySQL

## Proposed changes:

Transaction queries were outright failing when using SQLite. The underlying causes:

1. We use `transaction` as a table alias in some of our queries. This is a reserved keyword in both MySQL and SQLite, but apparently SQLite is more strict. This PR changes the table alias to `transactions`.
2. MySQL and SQLite have slightly different syntax for `GROUP_CONCAT()`. I logged the issue [here](https://github.com/WordPress/sqlite-database-integration/issues/70), and in the meantime created `build_group_concat()` to build the proper string depending on the database engine.

### Other information:

## Testing instructions:

The most painless way to do this is to set up SQLite via [the SQLite Database Integration plugin](https://wordpress.org/plugins/sqlite-database-integration/). Install that on a WordPress site, initialise the site, and then you should be good to go.

In `trunk`, the dashboard shows a `null` count, the transaction list shows no transactions even if you create one, and creating/viewing a transaction shows an error (see issue for screenshots).

In the `fix/crm/3404-sqlite_transactions` branch, the dashboard shows the correct count, the transaction list shows transactions, and one can create/view transactions.

### DIFF
--- a/projects/plugins/crm/changelog/fix-crm-3404-sqlite_transactions
+++ b/projects/plugins/crm/changelog/fix-crm-3404-sqlite_transactions
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+Transactions: Better support for SQLite.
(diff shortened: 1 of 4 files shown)


coder-karen (approved):
Tests well for me.

PR #34872, opened 2024-01-05
CRM: 3406 - address SQLite table creation error

tbradsha:
Resolves Automattic/zero-bs-crm#3406 - address SQLite table creation error

## Proposed changes:

If using SQLite, the System Status page (`/wp-admin/admin.php?page=zerobscrm-systemstatus&tab=status`) will show a notice that there were issues creating database tables.

This addresses the above by short-circuiting the InnoDB check if using SQLite, as we already know SQLite has no InnoDB support and the `SHOW ENGINES` query used is not valid in SQLite.

This was a bit complicated, as it was supposed to output the errors but they were always blank. Digging around, I learned that if a database query was determined to be invalid prior to running, it just returns `false` rather than actually running and providing the error. In this case, it seems it was because `SHOW ENGINES` was not deemed valid.

While investigating the cause, I managed to trigger some errors around a `zbscrm_api_keys` table; this was probably incidental because it was the first table created, but said table is never actually used other than hard-coded checks for its existence (don't ask me why), so I removed code related to its creation and detection. Removing it from existing installs is out of scope here.

### Other information:

## Testing instructions:

The most painless way to do this is to set up SQLite via [the SQLite Database Integration plugin](https://wordpress.org/plugins/sqlite-database-integration/). Install that on a WordPress site, initialise the site, and then you should be good to go.

Go to the System Status page: `/wp-admin/admin.php?page=zerobscrm-systemstatus&tab=status`

In `trunk`, you'll see this message:

```
Jetpack CRM experienced errors while trying to build it's database tables. Please contact support sharing the following errors: (#306sql)
```

In the `fix/crm/3406-address_sqlite_table_creation_error` branch, the error is no more.

### DIFF
--- a/projects/plugins/crm/admin/system/system-status.page.php
+++ b/projects/plugins/crm/admin/system/system-status.page.php
@@ -247,10 +247,6 @@ function zeroBSCRM_render_systemstatus_page() {
 
 		global $ZBSCRM_t,$wpdb;
 		$missingTables = array();
-		$tablesExist   = $wpdb->get_results( "SHOW TABLES LIKE '" . $ZBSCRM_t['keys'] . "'" );
-		if ( count( $tablesExist ) < 1 ) {
-			$missingTables[] = $ZBSCRM_t['keys'];
-		}
 
 		// then we cycle through our tables :) - means all keys NEED to be kept up to date :)
 		foreach ( $ZBSCRM_t as $tableKey => $tableName ) {
--- a/projects/plugins/crm/changelog/fix-crm-3406-address_sqlite_table_creation_error
+++ b/projects/plugins/crm/changelog/fix-crm-3406-address_sqlite_table_creation_error
@@ -1,0 +1,4 @@
+Significance: patch
+Type: fixed
+
+
--- a/projects/plugins/crm/includes/ZeroBSCRM.Database.php
+++ b/projects/plugins/crm/includes/ZeroBSCRM.Database.php
@@ -30,7 +30,6 @@ global $wpdb, $ZBSCRM_t;
   $ZBSCRM_t['tags']                   = $wpdb->prefix . "zbs_tags";
   $ZBSCRM_t['taglinks']               = $wpdb->prefix . "zbs_tags_links";
   $ZBSCRM_t['settings']               = $wpdb->prefix . "zbs_settings";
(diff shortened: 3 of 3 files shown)


coder-karen:
> The most painless way to do this is to set up SQLite via [the SQLite Database Integration plugin](https://wordpress.org/plugins/sqlite-database-integration/). Install that on a WordPress site, initialise the site, and then you should be good to go.

What steps / what kind of testing site did you use to be able to do this? I've tried with Jurassic Ninja sites - results in the following error:

Screenshot 2024-01-08 at 08 59 28

Local (by Flywheel) sites - activating the plugin just results in an error saying that 'The SQLite Integration plugin is active, but the SQLite3 class is missing from your server. Please make sure that SQLite is enabled in your PHP installation.'.

Then on a Pressable site, which just resulted in a critical error.

tbradsha:
Jurassic Ninja seems to have more general issues with SQLite prior to even getting to our plugin; probably a custom setup preventing WP from installing.

> Local (by Flywheel) sites - activating the plugin just results in an error saying that 'The SQLite Integration plugin is active, but the SQLite3 class is missing from your server. Please make sure that SQLite is enabled in your PHP installation.'.

Odd; I didn't have any issues...what version of PHP?

> Then on a Pressable site, which just resulted in a critical error.

What was the error there?

coder-karen:
Noting here that we sorted out the SQLite issues - updating my Local site to use PHP 8.0+ helped.

I did test this in two different environments though and both times I am still seeing 'Jetpack CRM experienced errors' message on the system status page. The first environment is the Local test site, testing first on trunk and then using the branch. Then secondly directly in WP playground, initially testing trunk then the branch.

However in the third scenario it did work - directly creating a WP playground site and going straight to the branch. Having trunk active beforehand seemed to cause the error to stay. Is this expected?

coder-karen (approved):
Thinking about this more, it really doesn't matter what happens in trunk before using this branch, as once this is merged the playground is only ever going to use the new code. So 👍 from me.

tbradsha:
Yes, if you go from `trunk` to this branch, I believe it'll continue to show the error because it's stored as a WP option, and as long as that option exists it'll continue to show the message for debug purposes.

PR #34873, opened 2024-01-05
CRM: 3409 - general cosmetic fixes

tbradsha:
Resolves Automattic/zero-bs-crm#3409 - general cosmetic fixes

## Proposed changes:

1. Fix a typo on the error from Automattic/zero-bs-crm#3406 (`it's` → `its`).
2. The `task_offset_fix` migration never ran if the WP timezone is set to UTC or equivalent, because we returned earlier if the timezone offset in seconds was `empty()`. Since UTC's timezone offset in seconds is 0, it would be empty and return early, never marking as complete. We now ensure it's marked complete if empty.
3. Some System Assistant steps had links that were broken when `esc_html` was added. They've been updated to use `wp_kses`.
4. I chopped a tiny pre-DAL3 code block.

### Other information:

## Testing instructions:

Easiest thing here is to look at the code and confirm all seems well.


coder-karen (approved):
Everything here looks ok to me 👍

PR #34881, opened 2024-01-06
WP.com API: Allow Users to Insert Header Code 

Aurorum:
Required for Automattic/wp-calypso#86079

## Proposed changes:

* This adds an endpoint that ultimately allows users to include Header code on their site. 

I thought about adding a separate endpoint like Custom CSS in case it handles a large amount of code, but I decided against it for a few reasons: realistically, I think this will mainly be used for a couple of meta tags; the endpoint already handles large amounts of data (including subscription email HTML and post categories!) and this isn't any larger relative to those, and also I don't think I can add another endpoint on the Calypso side since that's not open-source.  

## Jetpack product discussion

See Automattic/wp-calypso#56023.

No.

## Testing instructions:

I always need help testing endpoints. :) I haven't actually been able to do this from an Atomic site, but I have from a Jetpack one. 

1. Load the Calypso PR and navigate to `/settings/header-code/site`
2. Include some header code 
3. Confirm that it's run when you load the site 

For example - saving this on the Calypso side:
Screenshot 2024-01-06 at 15 26 22

Should lead to this:
Screenshot 2024-01-06 at 15 47 13

### DIFF
--- a/projects/plugins/jetpack/json-endpoints/class.wpcom-json-api-site-settings-endpoint.php
+++ b/projects/plugins/jetpack/json-endpoints/class.wpcom-json-api-site-settings-endpoint.php
@@ -458,7 +458,7 @@ class WPCOM_JSON_API_Site_Settings_Endpoint extends WPCOM_JSON_API_Endpoint {
 						'page_for_posts'                   => (string) get_option( 'page_for_posts' ),
 						'subscription_options'             => (array) get_option( 'subscription_options' ),
 						'jetpack_verbum_subscription_modal' => (bool) get_option( 'jetpack_verbum_subscription_modal', true ),
-						'jetpack_header_code'              => get_option( 'jetpack_header_code', true ),
+						'jetpack_header_code'              => get_option( 'jetpack_header_code', '' ),
 					);
 
 					if ( defined( 'IS_WPCOM' ) && IS_WPCOM ) {
--- a/projects/plugins/jetpack/json-endpoints/class.wpcom-json-api-site-settings-endpoint.php
+++ b/projects/plugins/jetpack/json-endpoints/class.wpcom-json-api-site-settings-endpoint.php
@@ -458,6 +458,7 @@ class WPCOM_JSON_API_Site_Settings_Endpoint extends WPCOM_JSON_API_Endpoint {
 						'page_for_posts'                   => (string) get_option( 'page_for_posts' ),
 						'subscription_options'             => (array) get_option( 'subscription_options' ),
 						'jetpack_verbum_subscription_modal' => (bool) get_option( 'jetpack_verbum_subscription_modal', true ),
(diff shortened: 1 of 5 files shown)


projects/plugins/jetpack/modules/header-code.php
```diff
@@ -0,0 +1,29 @@
...
+function jetpack_insert_head_code() {
+	$option = get_option( 'jetpack_header_code' );
+
+	// Ensure no code is added on Simple sites.
+	if ( defined( 'IS_WPCOM' ) && IS_WPCOM ) {
+		return;
+	}
+
+	if ( ! empty( $option ) && ! is_admin() ) {
+		echo wp_unslash( $option, 'post' ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
Aurorum, line 27: What's the best way of handling escaping for something like this? I wasn't quite sure.

anomiex: In theory, probably some form of `wp_kses()`. But since the intent of this is to allow for `<script>` tags there doesn't seem much point in bothering, since if script tags are allowed an attacker able to set this option could do anything they want anyway.

Pointing that out in a comment might be helpful.

Aurorum: Thanks, will do!

projects/plugins/jetpack/json-endpoints/class.wpcom-json-api-site-settings-endpoint.php
```diff
@@ -1115,6 +1116,13 @@ function ( $category_id ) {
 
 					break;
 
+				case 'jetpack_header_code':
+					if ( update_option( $key, $value ) ) {
```
tyxla, line 1120: Should we perform any safety checks on save? Are we good with any input without any verification?

Aurorum: > Are we good with any input without any verification?

I _think_ so, but happy to be corrected if I'm missing anything obvious. It also seems consistent with what the other options do.

tyxla (commented):
It also may be worth discussing if this field should have a UI in Jetpack, or if it's enough to have it in WP.com.

Aurorum:
Would someone from Jetpack mind checking for a product review decision on that? Happy to include some UI if it's something which Jetpack folks would be keen on.

jeherve:
> Would someone from Jetpack mind checking for a product review decision on that? Happy to include some UI if it's something which Jetpack folks would be keen on.

That is definitely something we would need to discuss before we move forward. I'll comment on the original issue so we can continue this discussion.
https://github.com/Automattic/wp-calypso/issues/56023#issuecomment-1884950243

Aurorum:
I still feel this PR is useful! But also still awaiting a product review…wonder if it might be possible to get a decision on that?

PR #34885, opened 2024-01-08
Fix reply link when comment_registration is set

heavyweight:
This PR fixes an issue when comment_registration and we load the jetpack commenting experience via iframe.
This is done by extending the comment class to override the 'comment_reply_link' return value when the setting is set.

Fixes #
126-gh-Automattic/verbum

## Proposed changes:
* Add a filter to apply the override for the specific scenario.

### Other information:

## Testing instructions:
- On atomic or self-hosted site, set `comment_registration` to be on. This is done in Settings -> Discussion.
- Visit a post of your testing site without being logged in.
- The reply link of the post comments should be "Reply" instead of "Login to Reply".

| Before  | After |
| Screenshot 2024-01-08 at 13 05 39  |  Screenshot 2024-01-08 at 13 15 26 |

### DIFF
--- a/projects/plugins/jetpack/changelog/fix-comment-reply-link
+++ b/projects/plugins/jetpack/changelog/fix-comment-reply-link
@@ -1,0 +1,4 @@
+Significance: patch
+Type: other
+
+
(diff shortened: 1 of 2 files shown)


agrullon95 (approved):
Changes are working on Atomic after applying branch with Jetpack Beta Plugin.
I couldn't replicate the issue on my self-hosted site.

Issue #32373, opened 2023-08-09
Subscriber block: present large subscriber numbers more elegantly


simison:
### Impacted plugin

Jetpack

### What

Present these large numbers more elegantly in subscriber block.

### How

![image](https://github.com/Automattic/jetpack/assets/87168/9956672a-4364-451d-af15-524ddb227800)

> we should present these large numbers more elegantly, like on social media platforms
> 100M instead of writing the whole number, which looks like a geocities visitor counter

Let's also check that the number formats are localized correctly. 

Other places to check and ensure everything is consistent:

- [x] Stats (no need to do anything, see https://github.com/Automattic/jetpack/issues/32373#issuecomment-1878590689)
- [x] Reader https://github.com/Automattic/wp-calypso/pull/85984
- [x] Subscriber management https://github.com/Automattic/wp-calypso/pull/86038
- [x] In the editor the block rendering itself, and the sidepanel
    - [x] Server-side (PHP) https://github.com/Automattic/jetpack/pull/32651
    - [x] Editor side (JS) sidebars https://github.com/Automattic/jetpack/pull/34887
    - [x] Editor side (JS) the block at canvas https://github.com/Automattic/jetpack/pull/34919

<img width="280" alt="image" src="https://github.com/Automattic/jetpack/assets/87168/1be43137-efb8-4197-af9a-fe5e9915257f">

<img width="281" alt="image" src="https://github.com/Automattic/jetpack/assets/87168/d5ed9f27-4da6-4646-8cb7-dc4334b0ec43">

<img width="729" alt="image" src="https://github.com/Automattic/jetpack/assets/87168/3b0d0cfc-c307-4fbc-9013-c2e98a87f7f5">

<img width="1287" alt="image" src="https://github.com/Automattic/jetpack/assets/87168/3271261d-c6c7-480c-904e-47f6787234e4">

<img width="1146" alt="image" src="https://github.com/Automattic/jetpack/assets/87168/adff48d1-c842-4a5f-b362-0d3830742149">

<img width="357" alt="Screenshot 2024-01-03 at 16 35 12" src="https://github.com/Automattic/jetpack/assets/87168/8cc20114-8121-4971-833d-afb0a88545a0">
<img width="252" alt="Screenshot 2024-01-03 at 16 34 52" src="https://github.com/Automattic/jetpack/assets/87168/a06378a6-7e7f-41ad-97a2-23c70797010b">

Slack ref p1691598746482899-slack-C052XEUUBL4

[...]

PR #34887, opened 2024-01-08
Subscriptions: Format subscribers reach numbers

pkuliga:
Part of https://github.com/Automattic/jetpack/issues/32373

## Proposed changes:

It presents large subscribers reach numbers more elegantly by:
- using the `numberFormat` helper with `compact` notation
- using the `text-wrap: balance;` CSS rule

#### Access section

| Before | After |
|--------|--------|
| Screenshot 2024-01-08 at 14 33 27 | Screenshot 2024-01-08 at 14 32 20

#### Subscribe block

| Before | After |
|--------|--------|
| Screenshot 2024-01-08 at 17 01 41 | Screenshot 2024-01-08 at 17 01 15 | 

### Other information:

## Jetpack product discussion

N/A

No

#### Access section

* Either have a site with 1000+ subscribers, or mock the number at `getReachForAccessLevelKey()`:
https://github.com/Automattic/jetpack/blob/90cfd932f1d4a1ba94331cb25ee9c68d09c3cc84/projects/plugins/jetpack/extensions/shared/memberships/settings.js#L45-L52
* Create a post, and publish it making sure the subscriber numbers are displayed correctly in the Access panels.

#### Subscribe block

* Either have a site with 1000+ subscribers, or mock the number:
https://github.com/Automattic/jetpack/blob/7c0bcb1a86f7be478b1315b6f47523482e2d2c41/projects/plugins/jetpack/extensions/blocks/subscriptions/controls.js#L65-L74
* Add a Subscribe block making sure the subscribers number is displayed correctly in the block options.


projects/plugins/jetpack/extensions/shared/memberships/settings.js
```diff
@@ -167,19 +178,18 @@ export function NewsletterAccessRadioButtons( {
...
 						: [] ),
 					{
-						label: `${ accessOptions.subscribers.label } (${ getReachForAccessLevelKey( {
-							accessLevel: accessOptions.subscribers.key,
-							emailSubscribers,
-							paidSubscribers,
-						} ).toLocaleString() })`,
+						label: `${ accessOptions.subscribers.label }\u00A0(${ numberFormat( subscribersReach, {
+							notation: 'compact',
+						} ) })`,
```
edanzer, line 183: Totally fine to address this separately or after this PR, but I'd love to see us find a better and more complete solution for wrapping than use of the \u00A0 character. 

Could be we could do it with CSS. 

But one thing that's odd is there's plenty of space to show each line with no wrapping at all:
wrapping

I think it's some default editor styles that are making it wrap, and I'm not sure if we want to over-write core editor styles, but there is plenty of space there for the full text with out wrapping.

edanzer (dismissed):

projects/plugins/jetpack/extensions/shared/memberships/settings.scss
```diff
@@ -12,6 +12,12 @@
 	text-decoration: none;
 }
 
+.jetpack-newsletter-access-radio-buttons {
+	.components-radio-control__label {
```
simison: Do we need the core class here (which could change later and we wouldn't notice), and could just apply for the whole section?

pkuliga: We could remove it but for performance reasons, I wanted to limit the scope of it. Is it a common thing for core classes to change?

simison: Very unlikely in this case for it to change, since it's component and considered stable API,  but I don't think it's issue for performance, as there wouldn't be much content in this section ever. 

I don't remember if the section has other content anymore (we used to have Stripe connect nudge at least), but would be good for CSS to apply to those too.

simison: Otherwise applying to `label` and `p` tags I think is safe for performance.

pkuliga: Like the idea of using `label` instead of the core class. Thanks! Fix already pushed.

simison (approved):
Looks good and tests well!

Screenshot 2024-01-09 at 13 10 33

projects/plugins/jetpack/extensions/shared/memberships/settings.js
pkuliga: Like the idea of using CSS instead, thanks for the review!

